PR #15433, opened 2025-11-27
docs: Update wrong helm reference slack.md

Tim-Visser:
It's using the wrong ENV variable - should be: DATAHUB_ACTIONS_SLACK_DATAHUB_BASE_URL


deepgarg760:
Thanks for the contribution!!. 
run `./gradlew :datahub-web-react:mdPrettierFormat` to fix the markdown and commit again

PR #15435, opened 2025-11-27
feat(ingestion/hive-metastore): add upstream lineage to hive-metastore

acrylJonny:
# Add Storage Lineage to Hive/Hive Metastore + Code Refactoring

## Summary

Adds storage lineage support to Hive and Hive Metastore connectors, enabling lineage tracking between Hive tables and their underlying storage locations (S3, Azure, GCS, HDFS, DBFS). Also refactors Hive sources into a clean directory structure.

### Storage Lineage (Opt-in Feature)
New configuration options (disabled by default):
- `emit_storage_lineage`: Enable storage lineage extraction
- `hive_storage_lineage_direction`: Set direction (`upstream` or `downstream`)  
- `include_column_lineage`: Enable column-level lineage
- `storage_platform_instance`: Platform instance for storage URNs

**Supported platforms:** S3, Azure (ADLS/ABFS), GCS, HDFS, DBFS, local files

### Code Refactoring
- Moved `hive.py` → `hive/hive_source.py`
- Moved `hive_metastore.py` → `hive/hive_metastore_source.py`
- Created `hive/storage_lineage.py` with shared logic
- Extracted `HiveStorageLineageConfigMixin` to eliminate duplication
- Updated `setup.py` entry points to use fully qualified paths

### Code Quality Improvements
- Converted `HiveStorageLineageConfig` to Pydantic model
- Created `LineageDirection` and `StoragePlatform` StrEnums for type safety
- Fixed unsafe exception handling in `get_db_schema` (now raises `ValueError` for invalid input)
- Improved error reporting in `get_workunits_internal` (specific exceptions + proper logging)
- Removed 50+ redundant comments

## PR Checks

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/sql/hive/hive_source.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/sql/hive/hive_source.py
@@ -274,9 +274,15 @@ class HiveSource(TwoTierSQLAlchemySource):
         if self._COMPLEX_TYPE.match(fields[0].nativeDataType) and isinstance(
             fields[0].type.type, NullTypeClass
         ):
-            assert len(fields) == 1
+            if len(fields) != 1:
+                logger.warning(
+                    f"Expected exactly 1 field for complex type {fields[0].nativeDataType}, "
+                    f"got {len(fields)} fields. Skipping complex type expansion.",
+                    extra={"dataset": dataset_name, "column": column.get("name")},
+                )
+                return fields
+
             field = fields[0]
-            # Get avro schema for subfields along with parent complex field
             avro_schema = get_avro_schema_for_hive_column(
                 column["name"], field.nativeDataType
             )
@@ -285,7 +291,6 @@ class HiveSource(TwoTierSQLAlchemySource):
                 json.dumps(avro_schema), default_nullable=True
             )
 
-            # First field is the parent complex field
             new_fields[0].nullable = field.nullable
             new_fields[0].description = field.description
             new_fields[0].isPartOfKey = field.isPartOfKey
--- a/metadata-ingestion/src/datahub/ingestion/source/sql/hive/storage_lineage.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/sql/hive/storage_lineage.py
@@ -395,11 +395,19 @@ class HiveStorageLineage:
                     ),
                 )
 
-        except (ValueError, TypeError, AttributeError) as e:
+        except ValueError as e:
             logger.warning(
-                f"Failed to create storage dataset MCPs for {storage_location}: {e}",
+                f"Invalid storage location format {storage_location}: {e}",
                 exc_info=True,
             )
+            self.report.report_location_failed(storage_location)
+            return
+        except (TypeError, AttributeError) as e:
+            logger.error(
+                f"Programming error while creating storage dataset MCPs for {storage_location}: {e}",
+                exc_info=True,
+            )
+            self.report.report_location_failed(storage_location)
             return
 
     def get_lineage_mcp(


metadata-ingestion/src/datahub/ingestion/source/sql/hive/hive_metastore_source.py
```diff
@@ -140,6 +145,32 @@ class HiveMetastore(BasicSQLAlchemyConfig):
         description="Simplify v2 field paths to v1 by default. If the schema has Union or Array types, still falls back to v2",
     )
 
+    emit_storage_lineage: bool = Field(
```
sgomezvillamor: do we usually call this `emit_...` or `include_...`?

metadata-ingestion/src/datahub/ingestion/source/sql/hive/hive_metastore_source.py
```diff
@@ -140,6 +145,32 @@ class HiveMetastore(BasicSQLAlchemyConfig):
...
+        default=False,
+        description="Whether to emit storage-to-Hive lineage",
+    )
+    hive_storage_lineage_direction: str = Field(
```
sgomezvillamor: this ability to choose direction, do we have something similar in other sources?

anyway, it could be a `Literal` https://docs.pydantic.dev/1.10/usage/types/#literal-type

acrylJonny: We also have this already in this in the main hive source (i.e. not the hive metastore source)

acrylJonny: Therefore adding it here also for consistency. One of the main reasons for this is because Spark often can show lineage just into the files rather than the metastore depending on how the table is updated.

metadata-ingestion/docs/sources/hive-metastore/hive-metastore_pre.md
sgomezvillamor, line 1: Great docs!

Still a new user may wonder: which one `hive` or `hive-metastore` should I use? is `hive` going to be deprecated? is there feature parity or what's the feature that one has that the other misses?

And for existing users of `hive`: what's the plan? is it going to be deprecated eventually? are new features going to be implemented in the `hive` one too?

metadata-ingestion/docs/sources/hive/hive_pre.md
```diff
@@ -0,0 +1,361 @@
...
+  type: hive
+  config:
+    host_port: prod-hive.company.com:10000
+    platform_instance: "prod-hive"
+```
+
+This creates URNs like:
+
+```
+urn:li:dataset:(urn:li:dataPlatform:hive,database.table,prod-hive)
```
sgomezvillamor, line 219: both `hive` and `hive-metastore` generate URNs with `hive` platform in the URN, right?
so if a user moves from `hive` to `hive-metastore` that won't be a breaking change in the identiyy of the assets, right?

metadata-ingestion/src/datahub/ingestion/source/sql/hive/hive_metastore_source.py
```diff
@@ -780,18 +837,14 @@ def loop_views(
             )
             dataset_snapshot.aspects.append(schema_metadata)
 
-            # add view properties
-            properties: Dict[str, str] = {
-                "is_view": "True",
-            }
+            properties: Dict[str, str] = {"is_view": "True"}
```
sgomezvillamor, line 840: this is not new behaviour

anyway, why not using SubTypes aspect here to track this as a view?

metadata-ingestion/src/datahub/ingestion/source/sql/hive/hive_source.py
```diff
@@ -0,0 +1,367 @@
...
+from datahub.utilities import config_clean
+from datahub.utilities.hive_schema_to_avro import get_avro_schema_for_hive_column
+
+logger = logging.getLogger(__name__)
+
+register_custom_type(HiveDate, DateTypeClass)
+register_custom_type(HiveTimestamp, TimeTypeClass)
+register_custom_type(HiveDecimal, NumberTypeClass)
+
+
```
sgomezvillamor, line 62: with the refactor, it's hard to see if all these patches are new or just existed before

are they properly tested?

any idea if those patches are fixed in higher versions of the libs?

metadata-ingestion/src/datahub/ingestion/source/sql/hive/hive_source.py
```diff
@@ -0,0 +1,367 @@
...
+                    )
+                )
+                coltype = types.NullType  # type: ignore
+            result.append(
+                {
+                    "name": col_name,
+                    "type": coltype,
+                    "nullable": True,
+                    "default": None,
+                    "full_type": orig_col_type,  # pass it through
```
sgomezvillamor, line 102: so `full_type`, this is what we really miss in the original non-patched method?

what would be the impact of missing int?
patch is a heavy tech debt, so wondering if the value worths it

metadata-ingestion/src/datahub/ingestion/source/sql/hive/hive_source.py
```diff
@@ -0,0 +1,367 @@
...
+    # including the view definition. However, for multiline view definitions,
+    # it returns multiple rows (of one column each), each with a part of the definition.
+    # Any whitespace at the beginning/end of each view definition line is lost.
+    rows = connection.execute(f"SHOW CREATE TABLE {full_table}").fetchall()
+    parts = [row[0] for row in rows]
+    return "\n".join(parts)
+
+
+HiveDialect.get_view_names = get_view_names_patched
+HiveDialect.get_view_definition = get_view_definition_patched
```
sgomezvillamor, line 141: HiveDialect, this comes from our fork, right? https://github.com/acryldata/PyHive
May we just fix there?

metadata-ingestion/src/datahub/ingestion/source/sql/hive/storage_lineage.py
```diff
@@ -0,0 +1,489 @@
...
+                        entityUrn=storage_urn, aspect=storage_schema
+                    ),
+                )
+
+        except (ValueError, TypeError, AttributeError) as e:
+            logger.warning(
+                f"Failed to create storage dataset MCPs for {storage_location}: {e}",
+                exc_info=True,
+            )
+            return
```
sgomezvillamor: this exception swallowing pattern, is that what we usually do in other sources when we fail emission?

sgomezvillamor:
codecoverage report shows little coverage in failure/exception scenarios, we could iimprove a little bit there, mainly in hive_source.py

also, concerning the lack of coverage for `dbapi_get_columns_patched`

metadata-ingestion/docs/sources/hive/hive_pre.md
acrylJonny: That is correct. You can move between the two and the URNs generated ultimately are the same but using different approaches (metastore db vs thrift). This hasn't changed.

metadata-ingestion/src/datahub/ingestion/source/sql/hive/hive_metastore_source.py
acrylJonny: happy to make this a view subtype - just didn't want to introduce this as it is a change in the connector. Happy to do so if you think that's wise.

metadata-ingestion/src/datahub/ingestion/source/sql/hive/hive_source.py
acrylJonny: This is pre-existing - https://github.com/datahub-project/datahub/blob/ddd93212bb8fb56cfa343e076abf805985f3822b/metadata-ingestion/src/datahub/ingestion/source/sql/hive.py#L64-L66

metadata-ingestion/src/datahub/ingestion/source/sql/hive/hive_source.py
acrylJonny: This is pre-existing - https://github.com/datahub-project/datahub/blob/ddd93212bb8fb56cfa343e076abf805985f3822b/metadata-ingestion/src/datahub/ingestion/source/sql/hive.py#L605-L612

acrylJonny:
> codecoverage report shows little coverage in failure/exception scenarios, we could iimprove a little bit there, mainly in hive_source.py
> 
> also, concerning the lack of coverage for `dbapi_get_columns_patched`

I'll add these today so that we can get this signed off.

metadata-ingestion/src/datahub/ingestion/source/sql/hive/hive_source.py
sgomezvillamor: I didn't mean this part specifically but the whole file. Hard to identify what's new or not.

Nevertheless, it's ok

metadata-ingestion/src/datahub/ingestion/source/sql/hive/hive_source.py
sgomezvillamor: so this patch is not new? is that what you mean?

metadata-ingestion/docs/sources/hive-metastore/hive-metastore_pre.md
acrylJonny: No - they will both remain for now - but long term we could definitely look to consolidate. Essentially Hive connects to the thrift endpoint so gets all of the information from hive and hive-metastore connects to the backend database of Hive as not everyone is running the thrift server - in fact this is often the case for Databricks, Trino, Presto, etc. But both of the connectors do have a lot of overlap so here what I'm looking to do is consolidate where we have shared logic, so for the moment this is purely a logic consolidation rather than a full connector consolidation.

acrylJonny: What's actually added in this commit is adding the upstream lineage to the hive metastore connector which actually already exists on the hive connector.

metadata-ingestion/src/datahub/ingestion/source/sql/hive/hive_source.py
acrylJonny: correct

metadata-ingestion/src/datahub/ingestion/source/sql/hive/hive_source.py
acrylJonny: This isn't actually a change and this logic exists on the original implementation.

metadata-ingestion/src/datahub/ingestion/source/sql/hive/hive_metastore_source.py
acrylJonny: I'll update this to `include_table_location_lineage` and soft deprecate the `emit_storage_lineage` parameter in hive-metastpre as `include_table_location_lineage` I think is the common pattern - at least it is what is used in [Athena](https://docs.datahub.com/docs/generated/ingestion/sources/athena#config-details)

metadata-ingestion/src/datahub/ingestion/source/sql/hive/hive_metastore_source.py
acrylJonny: I've now made this a strEnum

metadata-ingestion/docs/sources/hive-metastore/hive-metastore_pre.md
sgomezvillamor: thanks for the update

yeah, the refactor (moving files) didn't help to see what's new
basically, mixing refactor and feature dev doesn't help

metadata-ingestion/src/datahub/ingestion/source/sql/hive/hive_metastore_source.py
sgomezvillamor: +1 for Subtypes

metadata-ingestion/src/datahub/ingestion/source/sql/hive/hive_source.py
sgomezvillamor: Again, the mix of refactor and feature development is hitting the reviewers in the face 😅

metadata-ingestion/src/datahub/ingestion/source/sql/hive/hive_metastore_source.py
sgomezvillamor: ok, I saw you did 👌

metadata-ingestion/src/datahub/ingestion/source/sql/hive/hive_metastore_source.py
sgomezvillamor: My comment about the renames was assuming all this was new dev

With the confusion because of mixing refactor and feat dev, have we introduced some breaking change on the configs? if so, we should likely keep them as they were

acrylJonny: Ah, no worries. I have reverted a follow-up change I made to consolidate to keep this as is.

metadata-ingestion/src/datahub/ingestion/source/sql/hive/storage_lineage.py
acrylJonny: I've stopped this from swallowing on all errors to make this cleaner

acrylJonny:
> > codecoverage report shows little coverage in failure/exception scenarios, we could iimprove a little bit there, mainly in hive_source.py
> > also, concerning the lack of coverage for `dbapi_get_columns_patched`
> 
> I'll add these today so that we can get this signed off.

More tests have been added around this

PR #15437, opened 2025-11-28
fix(ingestion/extractor): accept bool type in json-schema extractor

### LINKED ISSUE #15421: Datahub failed when ingest json-schema if schema declare a property to true
**Describe the bug**
When trying to ingest a json-schema to a datahub instance, a bug happened if the schema declare a property to true.

**To Reproduce**

1. With recipe file is:
```yaml
source:
  type: json-schema
  config:
    path:  'debug.json'
    platform:  schemaregistry
    platform_instance: debug
    env: 'PROD'
    use_id_as_base_uri: true
    stateful_ingestion:
      enabled: false # recommended to have this turned on

sink:
    type: datahub-rest
    config:
      server: http://localhost:8080/
      token: .......
```

2. json schema file
```json
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "description": "demo schema for datahub json schema parsing",
    "type": "object",
    "properties": {
      "id":{
        "type": "string"
      },
      "content": true
    },
    "examples": [
      {
        "id": "123456",
        "content": 5
      },
      {
        "id": "123456",
        "content": {
          "some_field": "some_value"
        }
      },
      {
        "id": "123456",
        "content": false
      }
    ]
}
```
3. Use command line:  `datahub ingest -c json_schema.yaml `

**Error**
```
 Failed to process file *******ingest_sources\test1\debug.json
Traceback (most recent call last):
  File "*******.venv\lib\site-packages\datahub\ingestion\source\schema\json_schema.py", line 376, in get_workunits_internal
    yield from self._load_one_file(
  File "*******.venv\lib\site-packages\datahub\ingestion\source\schema\json_schema.py", line 274, in _load_one_file
    meta: models.SchemaMetadataClass = get_schema_metadata(
  File "*******.venv\lib\site-packages\datahub\ingestion\extractor\json_schema_util.py", line 672, in get_schema_metadata
    schema_fields = list(JsonSchemaTranslator.get_fields_from_schema(json_schema))
  File "*******.venv\lib\site-packages\datahub\ingestion\extractor\json_schema_util.py", line 636, in get_fields_from_schema
    yield from JsonSchemaTranslator.get_fields(
  File "*******.venv\lib\site-packages\datahub\ingestion\extractor\json_schema_util.py", line 600, in get_fields
    yield from generator.__get__(cls)(
  File "*******.venv\lib\site-packages\datahub\ingestion\extractor\json_schema_util.py", line 414, in _field_from_complex_type
    JsonSchemaTranslator._get_type_from_schema(field_schema),
  File "*******.venv\lib\site-packages\datahub\ingestion\extractor\json_schema_util.py", line 278, in _get_type_from_schema
    if Ellipsis in schema:
TypeError: argument of type 'bool' is not iterable
```

**Expected Behavior**:
- In the json schema specs a property set to true means any valid json.

- I expect the json schema to be loaded into datahub

Thanks :))

btkcodedev:
closes #15421

## Fix for Boolean Schema Properties in JSON Schema Ingestion

What: Fixed TypeError: argument of type 'bool' is not iterable when JSON schema properties are set to boolean values.

## Root cause
Methods expected Dict but received bool when properties were set to true/false per JSON Schema spec.

## Code Analysis
Main file: `metadata-ingestion/src/datahub/ingestion/extractor/json_schema_util.py`
-> Code expected all schemas to be Dict objects
-> Crashed at Ellipsis in schema and "key" in schema checks when schema was boolean
-> Boolean schemas are valid per JSON Schema: true = accepts any JSON, false = never validates

## Solution
-> Added isinstance(schema, bool) checks before Ellipsis in schema
-> Convert boolean → {} dict at property iteration

This PR prevents crashes throughout the call chain for bool type

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/extractor/json_schema_util.py
+++ b/metadata-ingestion/src/datahub/ingestion/extractor/json_schema_util.py
@@ -2,7 +2,7 @@ import json
 import logging
 import unittest.mock
 from hashlib import md5
-from typing import Any, Callable, Dict, Iterable, List, Optional, Type
+from typing import Any, Callable, Dict, Iterable, List, Optional, Type, Union
 
 import jsonref
 import jsonschema
@@ -273,11 +273,17 @@ class JsonSchemaTranslator:
         pass
 
     @staticmethod
-    def _get_type_from_schema(schema: Dict) -> str:
-        """Returns a generic json type from a schema."""
+    def _get_type_from_schema(schema: Union[Dict, bool]) -> str:
+        """Returns a generic json type from a schema.
+
+        Note: Per JSON Schema specification, boolean schemas are valid:
+        - true: accepts any valid JSON
+        - false: never validates
+        Reference: https://json-schema.org/understanding-json-schema/basics#hello-world!
+        """
         # Handle boolean schemas per JSON Schema spec: true accepts any JSON, false never validates
         if isinstance(schema, bool):
-            return "object"  # Treat both as generic object since DataHub has no "never" type
+            return "object" if schema else "null"
 
         if Ellipsis in schema:
             return "object"
@@ -361,10 +367,8 @@ class JsonSchemaTranslator:
         native_type_override = field_path._get_native_type_override()
         nullable = JsonSchemaTranslator._is_nullable(schema)
 
-        # Handle boolean schemas and recursive Ellipsis
-        if isinstance(schema, bool) or Ellipsis in schema:
-            # Boolean schemas: true = accept any JSON, false = never validates (treat both as {})
-            # Ellipsis: recursive fields, short-circuit by making this just be an object
+        if Ellipsis in schema:
+            # This happens in the case of recursive fields, we short-circuit by making this just be an object
             schema = {}
 
         if datahub_field_type == RecordTypeClass:
@@ -412,12 +416,6 @@ class JsonSchemaTranslator:
             field_path = field_path.expand_type(discriminated_type, schema)
 
             for field_name, field_schema in schema.get("properties", {}).items():
-                # Normalize boolean schemas per JSON Schema spec before processing
-                # true = accepts any JSON, false = never validates
-                # Convert to {} to avoid crashes in downstream code that expects Dict
-                if isinstance(field_schema, bool):
-                    field_schema = {}
-
                 required_field: bool = field_name in schema.get("required", [])
                 inner_field_path = field_path.clone_plus(
                     FieldElement(type=[], name=field_name, schema_types=[])


btkcodedev:
Unit test results:
image

deepgarg760:
@btkcodedev Thanks for the contribution

metadata-ingestion/src/datahub/ingestion/extractor/json_schema_util.py
```diff
@@ -275,6 +275,10 @@ def _field_from_primitive(
     @staticmethod
     def _get_type_from_schema(schema: Dict) -> str:
         """Returns a generic json type from a schema."""
+        # Handle boolean schemas per JSON Schema spec: true accepts any JSON, false never validates
+        if isinstance(schema, bool):
```
deepgarg760, line 285: method signature expects `schema` as `Dict` and here its been checked against bool

btkcodedev: This boolean check is to prevent unexpected crashes similar to the issue. The user mentioned a crash from 
`
    if Ellipsis in schema:
TypeError: argument of type 'bool' is not iterable
`

This is a check to prevent it

sgomezvillamor: then, should we fix the signature? `def _get_type_from_schema(schema: Dict) -> str:`

metadata-ingestion/src/datahub/ingestion/extractor/json_schema_util.py
```diff
@@ -275,6 +275,10 @@ def _field_from_primitive(
     @staticmethod
     def _get_type_from_schema(schema: Dict) -> str:
         """Returns a generic json type from a schema."""
+        # Handle boolean schemas per JSON Schema spec: true accepts any JSON, false never validates
```
sgomezvillamor, line 284: is there any public reference to the json schema spec that we could add here as comment?

metadata-ingestion/src/datahub/ingestion/extractor/json_schema_util.py
sgomezvillamor: `schema: Union[Dict, bool]`

metadata-ingestion/src/datahub/ingestion/extractor/json_schema_util.py
batuan: Hello, you can find the reference here: 
https://json-schema.org/understanding-json-schema/basics#hello-world!
Thanks

deepgarg760:
@btkcodedev , Please take care of the comment from @sgomezvillamor

PR #15443, opened 2025-11-28
feat(ui/ingestion): create new page for ingestion run details

purnimagarg1:
**Linear ticket:**
https://linear.app/acryl-data/issue/CAT-966/turn-ingestion-run-details-into-new-page

**Description:**

Creates a new page for run details of ingestion requests.
Also, adds a new component Breadcrumb in the component library

**Screenshot:**

image

### DIFF
--- a/datahub-web-react/src/app/ingestV2/executions/components/ExecutionsTable.tsx
+++ b/datahub-web-react/src/app/ingestV2/executions/components/ExecutionsTable.tsx
@@ -87,17 +87,14 @@ export default function ExecutionsTable({
     const hadleViewDetails = useCallback(
         (urn: string) => {
             if (showIngestionOnboardingRedesignV1) {
-                const selectedRequest = executionRequests.find((request) => request.urn === urn);
                 history.push(PageRoutes.INGESTION_RUN_DETAILS.replace(':urn', urn), {
                     fromUrl: tabUrlMap[TabType.RunHistory],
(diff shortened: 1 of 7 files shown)


datahub-web-react/src/app/ingestV2/executions/components/ExecutionsTable.tsx
```diff
@@ -81,6 +84,19 @@ export default function ExecutionsTable({
         setExecutionInfoToCancel(undefined);
     }, [handleCancelExecution, executionInfoToCancel]);
 
+    const hadleViewDetails = useCallback(
```
chriscollins3456, line 87: typo here

datahub-web-react/src/app/ingestV2/executions/components/ExecutionsTable.tsx
```diff
@@ -81,6 +84,19 @@ export default function ExecutionsTable({
...
+        (urn: string) => {
+            if (showIngestionOnboardingRedesignV1) {
+                history.push(PageRoutes.INGESTION_RUN_DETAILS.replace(':urn', urn), {
+                    fromUrl: tabUrlMap[TabType.RunHistory],
+                });
```
chriscollins3456, line 92: one thing we may need to consider is that if they go back and previously had filters or a query we would lose them with this approach. def not a huge deal and not blocking this, but something to consider

PR #15446, opened 2025-11-28
feat(ingestion): improve create/update source handling

v-tarasevich-blitz-brain:
This PR improves handling of the new create/update  source flow

https://github.com/user-attachments/assets/1d932b10-840d-4d84-b283-394e74571742

### DIFF
--- a/datahub-web-react/src/app/ingestV2/source/IngestionSourceList.tsx
+++ b/datahub-web-react/src/app/ingestV2/source/IngestionSourceList.tsx
@@ -20,7 +20,7 @@ import IngestionSourceRefetcher from '@app/ingestV2/source/IngestionSourceRefetc
 import IngestionSourceTable from '@app/ingestV2/source/IngestionSourceTable';
 import RecipeViewerModal from '@app/ingestV2/source/RecipeViewerModal';
 import { IngestionSourceBuilderModal } from '@app/ingestV2/source/builder/IngestionSourceBuilderModal';
-import { DEFAULT_EXECUTOR_ID, SourceBuilderState, StringMapEntryInput } from '@app/ingestV2/source/builder/types';
+import { SourceBuilderState } from '@app/ingestV2/source/builder/types';
 import {
     addToListIngestionSourcesCache,
     removeFromListIngestionSourcesCache,
@@ -28,6 +28,7 @@ import {
 } from '@app/ingestV2/source/cacheUtils';
 import {
     buildOwnerEntities,
+    getIngestionSourceMutationInput,
     getIngestionSourceSystemFilter,
     getSortInput,
     mapSourceTypeAliases,
@@ -338,13 +339,6 @@ export const IngestionSourceList = ({
         setFocusSourceUrn(undefined);
     };
 
-    const formatExtraArgs = (extraArgs): StringMapEntryInput[] => {
-        if (extraArgs === null || extraArgs === undefined) return [];
-        return extraArgs
-            .filter((entry) => entry.value !== null && entry.value !== undefined && entry.value !== '')
-            .map((entry) => ({ key: entry.key, value: entry.value }));
-    };
-
     const createOrUpdateIngestionSource = (
         input: UpdateIngestionSourceInput,
         resetState: () => void,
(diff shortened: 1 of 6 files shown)


datahub-web-react/src/app/ingestV2/source/IngestionSourceList.tsx
```diff
@@ -452,6 +470,25 @@ export const IngestionSourceList = ({
         }
     };
 
+    // Handle executing or refetching of a created or updated source via the separated page
```
chriscollins3456, line 467: interesting we have to handle it in this way.. i guess because we were executing or refetching on this page already since we were using a modal?

datahub-web-react/src/app/ingestV2/source/IngestionSourceList.tsx
```diff
@@ -452,6 +470,25 @@ export const IngestionSourceList = ({
...
+    const [isCreatedOrUpdatedSourceFromLocationHandled, setIsCreatedOrUpdatedSourceFromLocationHandled] =
+        useState<boolean>(false);
+    useEffect(() => {
+        if (createdOrUpdatedSourceUrnFromLocation && !isCreatedOrUpdatedSourceFromLocationHandled) {
+            setIsCreatedOrUpdatedSourceFromLocationHandled(true);
+            if (shouldRunCreatedOrUpdatedSourceFromLocation) {
+                executeIngestionSource(createdOrUpdatedSourceUrnFromLocation);
+            } else {
+                setSourcesToRefetch((prev) => new Set([...prev, createdOrUpdatedSourceUrnFromLocation]));
```
chriscollins3456, line 476: the way this works is that we fetch once and determine if we want to keep refetching right? like just cuz we updated a source doesn't mean we continue to keep refetching (unless we should because it's in the middle of a run) right?

datahub-web-react/src/app/ingestV2/source/multiStepBuilder/IngestionSourceCreatePage.tsx
```diff
@@ -43,31 +53,95 @@ export function IngestionSourceCreatePage() {
...
 
+    const { defaultOwnershipType } = useOwnershipTypes();
+
     const onSubmit = useCallback(
         async (data: SourceBuilderState | undefined) => {
-            if (data) {
-                await createIngestionSource({
-                    type: data.type as string,
+            if (!data) return undefined;
+            const shouldRun = true; // TODO:: set a real value
```
chriscollins3456, line 56: yeah we can update this once we add that next button to save & run

datahub-web-react/src/app/ingestV2/source/multiStepBuilder/IngestionSourceUpdatePage.tsx
```diff
@@ -49,40 +65,92 @@ export function IngestionSourceUpdatePage() {
...
+
+            const shouldRun = true; // TODO:: set a real value
+            try {
                 const source = ingestionSourceData?.ingestionSource as IngestionSource | undefined;
-                await updateIngestionSource(
-                    urn,
-                    {
-                        type: data.type as string,
-                        name: data.name as string,
+                const input = {
```
chriscollins3456: is it possible to combine some of this logic between `IngestionSourceCreatePage` and `IngestionSourceUpdatePage`?

chriscollins3456:
merging in master because it looks like not all tests ran ie. cypress

datahub-web-react/src/app/ingestV2/source/IngestionSourceList.tsx
v-tarasevich-blitz-brain: Yes, we do that using modals too

datahub-web-react/src/app/ingestV2/source/IngestionSourceList.tsx
v-tarasevich-blitz-brain: I've saved the same logic we have for modals. I see that some data we put to cache could contain empty values so we need to refetch it to get the full fresh data

datahub-web-react/src/app/ingestV2/source/multiStepBuilder/IngestionSourceUpdatePage.tsx
v-tarasevich-blitz-brain: I think I can extract a few common functions/hooks

PR #15457, opened 2025-12-01
feat(ingest/fivetran): make ingestion limits configurable

sgomezvillamor:
Makes Fivetran ingestion limits configurable to allow controlled increase when needed.

## Summary
- Added configurable limits to FivetranLogConfig: max_jobs_per_connector, max_table_lineage_per_connector, max_column_lineage_per_connector
- Default values remain the same (500, 120, 1000 respectively)
- Added validation to ensure positive values
- Updated documentation with configuration examples
- Added comprehensive unit tests

## Checklist
- [x] Tests added
- [x] Documentation updated


metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran.py
```diff
@@ -475,11 +474,14 @@ def _get_connector_workunits(
         yield datajob
 
         # Map Fivetran's job/sync history entity with Datahub's data process entity
-        if len(connector.jobs) >= MAX_JOBS_PER_CONNECTOR:
+        if (
```
skrydal, line 530: Correct me if I am wrong, but if we are already using `self.config.fivetran_log_config.max_jobs_per_connector` in the query used to retrieve the jobs, then this condition will be never met, isn't that right?

metadata-ingestion/tests/unit/fivetran/test_fivetran_config.py
```diff
@@ -0,0 +1,136 @@
...
+    FivetranLogConfig,
+    FivetranSourceConfig,
+    SnowflakeDestinationConfig,
+)
+
+
+class TestFivetranConfig:
+    """Test cases for Fivetran configuration with configurable limits."""
+
+    def test_fivetran_log_config_default_limits(self):
```
skrydal: Is this test meaningful?

skrydal: I would skip the entire suit

metadata-ingestion/tests/unit/fivetran/test_fivetran_query.py
```diff
@@ -0,0 +1,79 @@
...
+        """Test that FivetranLogQuery uses custom limits when provided."""
+        query = FivetranLogQuery(
+            max_jobs_per_connector=1234,
+            max_table_lineage_per_connector=567,
+            max_column_lineage_per_connector=8901,
+        )
+
+        assert query.max_jobs_per_connector == 1234
+        assert query.max_table_lineage_per_connector == 567
+        assert query.max_column_lineage_per_connector == 8901
```
skrydal: In my opinion those 2 tests are not necessary

metadata-ingestion/tests/unit/fivetran/test_fivetran_query.py
```diff
@@ -0,0 +1,79 @@
...
+        query.set_schema("test_schema")
+
+        sql = query.get_column_lineage_query(
+            connector_ids=["connector1", "connector2"],
+        )
+
+        assert (
+            f"QUALIFY ROW_NUMBER() OVER (PARTITION BY connection_id ORDER BY created_at DESC) <= {custom_limit}"
+            in sql
+        )
```
skrydal, line 48: Those tests are on spot. I would consider adding one integration test if possible (not necessary).

metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran.py
sgomezvillamor: To be honest, I didn’t think too deeply about this part. In this PR, I mainly made the limits configurable:
* The limits already existed; they were just hardcoded.
* The _if_ that triggers the report warning also already existed.

So I didn’t introduce new behavior. Once I’m done with my defensive reasoning 🙂, I’m happy to deep dive into your comment because it looks very reasonable.

It’s also worth noting that I’ve actually seen the report warning in practice, so there are cases where lineage gets truncated and the warning is shown.

Given that:
* the query returns <= limit
* the warning is shown when lineage >= limit

We may conclude that the current check is over-protective. It should probably be == to detect when we’ve actually hit the limit (which would indicate potential truncation), rather than >=.

I think being a bit over-protective isn’t harmful here. Do you think it’s better to be precise instead?

metadata-ingestion/tests/unit/fivetran/test_fivetran_config.py
sgomezvillamor: Well, it's meaningful.

I agree that the test is just covering pydantic field defaults ,and because of that it may be considered redundant or unnecessary.

Removing test file.

metadata-ingestion/tests/unit/fivetran/test_fivetran_query.py
sgomezvillamor: Removing...

metadata-ingestion/tests/unit/fivetran/test_fivetran_query.py
sgomezvillamor: Since Fivetran is SaaS, the options for integration testing are basically mocking or connector tests.

Mocking would require a fairly complex setup, and connector tests live in a different repo and are not easily accessible from here.

Given that, I think these unit tests are a good compromise to prevent the feature from being accidentally broken. I’m fine with adding integration tests in the connector-tests repo, but I’d still keep these unit tests here.

And more generally, is the goal to have an integration test for every single feature? 🤔

metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran.py
skrydal: Ok I from what I see the queries actually use condition `<=`, while this if uses `>=`, therefor common part is `=`. The message is not precise, but you can't actually hit the `>` case given the queries, it is logically impossible, right?

metadata-ingestion/tests/unit/fivetran/test_fivetran_query.py
skrydal: I think the more meaningful integration tests - the better. Preferably without mocking.
This case is especially tempting, as unit tests covering this case are close to trivial (you can basically just check query), while with the integration test you can check actual data lengths.
They belong to a different repo, that's true.

metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran.py
sgomezvillamor: Yes, that’s exactly what I mean in my reply: using `==` should be sufficient here; `>=` feels a bit over-protective. Anyway, no big deal.

PR #15471, opened 2025-12-03
feat(ui/ingestion): build new source selection screen

purnimagarg1:
**Linear ticket:**
https://linear.app/acryl-data/issue/CAT-973/build-new-source-selection-screen-step-1-of-create-new-source

**Screenshot:**

image

### DIFF
--- a/datahub-web-react/src/alchemy-components/components/Card/Card.tsx
+++ b/datahub-web-react/src/alchemy-components/components/Card/Card.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import { Tooltip } from '@components';
+import React, { useEffect, useRef, useState } from 'react';
 
 import {
     CardContainer,
@@ -12,14 +13,13 @@ import { CardProps } from '@components/components/Card/types';
 import { Pill } from '@components/components/Pills';
 
 export const cardDefaults: CardProps = {
-    title: 'Title',
     iconAlignment: 'horizontal',
     isEmpty: false,
     isCardClickable: true,
 };
 
 export const Card = ({
-    title = cardDefaults.title,
+    title,
     iconAlignment = cardDefaults.iconAlignment,
     subTitle,
     percent,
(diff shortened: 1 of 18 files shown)


datahub-web-react/src/alchemy-components/components/Card/Card.tsx
```diff
@@ -34,7 +34,29 @@ export const Card = ({
...
+            const isOverflowing = element.scrollHeight > element.clientHeight;
+            setShowSubtitleTooltip(isOverflowing);
+        });
+    }, []);
+
+    const subtitleElement = (
+        <SubTitle ref={subtitleRef} $noOfSubtitleLines={noOfSubtitleLines}>
+            {subTitle}
+        </SubTitle>
+    );
```
chriscollins3456, line 58: maybe we can move this to its own separate component to keep this `Card` implementation cleaner

datahub-web-react/src/alchemy-components/components/Card/Card.tsx
```diff
@@ -34,7 +34,29 @@ export const Card = ({
     style,
     isCardClickable = cardDefaults.isCardClickable,
     dataTestId,
+    noOfSubtitleLines,
+    iconStyles,
+    pillLabel,
```
chriscollins3456, line 39: one thing i've been thinking about is adding in all of these very specific props to our component library components - ultimately making it harder to use these components because it becomes very confusing on what props are needed and what does what.

maybe part of this is an issue with how design has created these cards in the design and that they differ too much from our main component in the library.

in an ideal world these components are more compositional with props that don't affect other props and there's only a few `variant`s that we have

chriscollins3456: okay talked to Paulina and i don't believe we need to pass in `iconStyles` but instead always have the icon top aligned like you have in `alignSelf: 'start',` that you're passing in. so that's nice and one less prop!

chriscollins3456: wondering if there's a better way to handle `noOfSubtitleLines` that is less specific but a more generic way to handle the subtitle content...

datahub-web-react/src/app/ingestV2/source/multiStepBuilder/steps/step1SelectSource/sources.json
```diff
@@ -0,0 +1,450 @@
+[
+    {
```
chriscollins3456, line 2: would it be possible to reuse the original sources.json here? and just add the new items. i'd rather not have two if we can help it because both will be sitting around for a while and people will get confused on which one to add a new source to

datahub-web-react/src/app/ingestV2/source/multiStepBuilder/steps/step1SelectSource/SelectSourceStep.tsx
```diff
@@ -1,24 +1,155 @@
...
+    align-items: center;
+    gap: 8px;
+`;
+
+const RightSection = styled.div`
+    display: flex;
+    cursor: pointer;
+`;
+
 export function SelectSourceStep() {
```
chriscollins3456, line 59: nice! one thing we'll have to build after this PR is going to be the ability to search/filter for a source by name

chriscollins3456 (approved):
overall looking good!

i have a few comments to think about and a few things I think we should do in a followup, but i think we're okay to merge to keep this train rolling

PR #15472, opened 2025-12-03
feat(kafka): add lightweight Kafka connectivity validation

max-datahub:
## Summary

Add lightweight Kafka connectivity validation for fail-fast startup checks in Python connectors and services (executor and integrations-service).

## Changes

- **New Exception**: `KafkaConnectivityError` for connectivity failures
- **New Function**: `validate_kafka_connectivity()` that validates Kafka cluster availability without enumerating all topics
- **Performance Optimization**: Uses `list_topics(topic="")` to get cluster metadata only, not full topic list
- **Configurable Timeout**: Uses `max(10, client_timeout_seconds)` with 10-second minimum
- **Broker Validation**: Explicitly checks that at least one broker is available

## Performance Impact

Listing cluster metadata should only take 1-2 seconds

## Testing

Added comprehensive unit test coverage:
- ✅ Successful connection scenario
- ✅ Connection failure scenario  
- ✅ No brokers available scenario

All tests passing.

## Use Case

Enables executor and integrations-service to fail-fast on startup if Kafka is unreachable, without causing performance degradation in large Kafka deployments.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/kafka/kafka.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/kafka/kafka.py
@@ -74,6 +74,12 @@ from datahub.utilities.str_enum import StrEnum
 logger = logging.getLogger(__name__)
 
 
+class KafkaConnectivityError(Exception):
+    """Exception raised when Kafka connectivity validation fails."""
+
+    pass
+
+
 class KafkaTopicConfigKeys(StrEnum):
     MIN_INSYNC_REPLICAS_CONFIG = "min.insync.replicas"
     RETENTION_SIZE_CONFIG = "retention.bytes"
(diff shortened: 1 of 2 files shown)


metadata-ingestion/src/datahub/ingestion/source/kafka/kafka.py
```diff
@@ -123,6 +129,54 @@ def get_kafka_admin_client(
     return client
 
 
+def validate_kafka_connectivity(connection: KafkaConsumerConnectionConfig) -> None:
```
deepgarg760, line 132: would like to know the caller of this function

max-datahub: @deepgarg760 I plan to open a separate PR for the executor-coordinator and integrations-service (which are not part of OSS) that leverage the connectivity check during startup. Currently, those services start "successfully", even if they have no Kafka connectivity

PR #15475, opened 2025-12-03
feat(ingest): add convert to lowercase for s3

kevinkarchacryl:
Adding convert_urns_to_lowercase for S3/GCS


treff7es (approved):

metadata-ingestion/src/datahub/ingestion/source/s3/source.py
```diff
@@ -461,6 +461,9 @@ def ingest_table(
         # remove protocol and any leading or trailing slashes
         browse_path = re.sub(URI_SCHEME_REGEX, "", table_data.table_path).strip("/")
 
+        if self.source_config.convert_urns_to_lowercase:
```
sgomezvillamor, line 464: is this needed and consistent with other sources?

if browse paths should be lowercased, that should be managed by the auto processors, no?

kevinkarchacryl: A lot of sources do things differently. See [hive](https://github.com/datahub-project/datahub/blob/4074d1ff0e72ac9b95917031e8ba93c5fbb94f01/metadata-ingestion/src/datahub/ingestion/source/sql/hive.py#L280) or [mssql](https://github.com/datahub-project/datahub/blob/4074d1ff0e72ac9b95917031e8ba93c5fbb94f01/metadata-ingestion/src/datahub/ingestion/source/sql/mssql/source.py#L1018).

PR #15479, opened 2025-12-04
feat(datahub-frontend): add client-cert-based authentication for oidc

manuschillerdev:
Adds support for **client-certificate-based authentication** (`private_key_jwt`, [RFC 7523](https://datatracker.ietf.org/doc/html/rfc7523)) to `datahub-frontend`'s OIDC SSO flow, as an alternative to the existing `client_secret_basic` /
`client_secret_post` methods.

Instead of a shared client secret, DataHub signs a JWT assertion with an RSA private key held on disk; the IdP (Azure AD / Entra ID, Okta, Keycloak, …) verifies it against the registered public certificate. This aligns OIDC auth with the
 broader push toward credential-less, PKI-backed service identity — short-lived certs rotate without config churn and unify the Kafka + OIDC PKI story into one identity.

Part of the TLS / mTLS RFC: **datahub-project/datahub#16975** ([PEM-first TLS for DataHub outbound connections](https://github.com/datahub-project/datahub/blob/master/docs/rfcs/active/16975-tls-kafka-oidc.md)).

Related: datahub-project/datahub#16975

## Checklist

- [x] The PR conforms to DataHub's [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [PR Title
Format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#pr-title-format))
- [x] Links to related issues / RFC
- [x] Docs related to the changes have been added (`configure-oidc-react.md`)
- [ ] For any breaking change/potential downtime/deprecation/big changes an entry has been made in [Updating DataHub](https://github.com/datahub-project/datahub/blob/master/docs/how/updating-datahub.md) — N/A, additive config only


billpratt:
Hi @manuschillerdev and @gabe-lyons Is this still in development? I'm looking for the same.

billpratt:
`private_key_jwt` support is a real gap for orgs that can't use
`client_secret_*`. One thing to flag before this merges: as written, the private key must be
loadable as a local `RSAPrivateKey` (`PrivateKeyJwtUtils.loadPrivateKey()`), which means it can't
work with non-exportable/HSM-backed keys in a cloud KMS — e.g. Azure Key Vault or AWS KMS keys
configured so the private key material can never leave the vault/HSM. That's a common security
requirement (and the posture our org needs).

Proposal: introduce a small signing SPI so the file-based loader becomes just the default
backend, with room for remote-signing backends that never expose the key:

```java
public interface PrivateKeyJwtSigner {
  byte[] sign(byte[] signingInput) throws JOSEException;   // raw signature bytes
  JWSAlgorithm algorithm();                                 // e.g. RS256
  String kid();
  Base64URL x5tS256();
  List<Base64> x5cChain();
}
```

- `FilePrivateKeyJwtSigner` — wraps this PR's existing loader unchanged (remains the default).
- `AzureKeyVaultPrivateKeyJwtSigner` — signs via `CryptographyClient.sign(alg, digest)`, scoped to
  the certificate's key ID; the private key never leaves Key Vault/Managed HSM. Auth via
  `WorkloadIdentityCredential` (no secret).
- Same shape works for AWS KMS (`KmsClient.sign(..., MessageType.RAW, RSASSA_PKCS1_V1_5_SHA_256)`
  via IRSA) or GCP Cloud KMS later — this needs to be cloud-agnostic, not Azure-only.

`CustomOidcAuthenticator` would delegate signing to whichever `PrivateKeyJwtSigner` is configured
(new `AUTH_OIDC_PRIVATE_KEY_JWT_SIGNER_TYPE=file|azure-keyvault|aws-kms`), rather than constructing
`RSASSASigner` directly. Happy to contribute the Key Vault implementation (and a KMS sketch) if
this shape is acceptable — wanted to check on the approach first given the scope of the change.

david-leifker (approved):

billpratt:
@david-leifker curious to get your thoughts on [my earlier comment](https://github.com/datahub-project/datahub/pull/15479#issuecomment-5516889949) about a pluggable signing SPI.

Short version: as implemented, the private key has to be loadable into the JVM as an `RSAPrivateKey`, so this won't work with non-exportable / HSM-backed keys in a cloud KMS (Azure Key Vault, AWS KMS, GCP Cloud KMS). A small `PrivateKeyJwtSigner` interface would keep the file-based loader as the default while leaving room for remote-signing backends where the key never leaves the vault.

Totally fine if the preference is to land this PR as-is and do the SPI as a follow-up — a `FilePrivateKeyJwtSigner` refactor is probably easier to review against merged code anyway. Mostly want to know whether the abstraction is a direction the team would accept before I put up a PR for it.

manuschillerdev:
hey @billpratt 
I had a pretty busy phase at work but want to resume work on my PRs now. I'll take a look at your ideas tomorrow - thanks for the ping!

manuschillerdev:
@david-leifker @gabe-lyons I rebased the PR and fixed some minor issues.

  I verified the sign-in flow against a local Keycloak setup with the following configuration:

  - private_key_jwt client authentication
  - RS256 signing
  - PEM-encoded private key and certificate
  - No client secret

  The browser completed the authorization-code flow and signed in successfully. Keycloak recorded the token exchange with client_auth_method: client-jwt.

## Screenshots and sanitized traces are attached below.
* [keycloak-events.json](https://github.com/user-attachments/files/32161778/keycloak-events.json)
* [datahub-oidc-success.log](https://github.com/user-attachments/files/32161834/datahub-oidc-success.log)
  
01-keycloak-login
02-datahub-login-success

so to the best of my knowledge, this should work :)

PR #15489, opened 2025-12-05
feat(ingestion/grafana) add option to pass grafana user email as dashboard owner

daha:
This MR add a option like for the powerbi source to use the user email for the corpuser URN or just the part before @. Default to the part before @ as that is the behavior before this change. 

Also so not add the unique dashboard id as a dashboard owner. I really do not understand why that was added in the first place. If someone really need it, it can be added by a transformer as the dashboard id is part of the dashboard URN. 

```
yaml
*Example Grafana recipe configuration*

source:
  type: grafana
  config:
    url: "https://your-grafana.com/"
    
    # Example 1: Default behavior (strip emails to usernames)
    # This creates: urn:li:corpuser:john
    remove_email_suffix: true  # or omit (this is default)
    
    # Example 2: Keep full email as corporate user identifier
    # This creates: urn:li:corpuser:john@company.com
    # Use this if your company's SSO/LDAP uses full emails as user IDs
    remove_email_suffix: false
```

### DIFF
--- a/metadata-ingestion/docs/sources/grafana/grafana_pre.md
+++ b/metadata-ingestion/docs/sources/grafana/grafana_pre.md
@@ -1,14 +1,14 @@
 ### Concept Mapping
 
-| Source Concept              | DataHub Concept                                           | Notes                                                                    |
-| --------------------------- | --------------------------------------------------------- | ------------------------------------------------------------------------ |
-| `"grafana"`                 | [Data Platform](../../metamodel/entities/dataPlatform.md) |                                                                          |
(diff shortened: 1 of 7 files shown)


deepgarg760:
@daha , Good work!! 
Please add documentation regarding change in the owner type

daha:
> @daha , Good work!! Please add documentation regarding change in the owner type

Where should that information be added, @deepgarg760?

daha:
I have renamed the configuration option, used the one used in powerbi source instead, which I think has a better name. I have updated the docs for the source a bit. @deepgarg760

PR #15494, opened 2025-12-06
fix: make published GMS and frontend ports configurable with quickstart

daha:
I noticed that the docker-compose file which I get when running did not allow to change the published port for GMS and frontend like many of the other docker images support, like https://docs.datahub.com/docs/troubleshooting/quickstart#common-problems suggest with `DATAHUB_MAPPED_GMS_PORT`. This change make the published ports for GMS and frontend configurable with environment variables when running `datahub docker quickstart`.


deepgarg760:
LGTM!!

PR #15498, opened 2025-12-08
feat(databricks): added support for databricks unified auth.

brock-acryl:
Add support for databricks unified authentication

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/unity/connection.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/unity/connection.py
@@ -57,51 +57,12 @@ class UnityCatalogConnectionConfig(ConfigModel):
     def __init__(self, **data: Any):
         super().__init__(**data)
 
-    def _resolve_token(self) -> Optional[str]:
-        """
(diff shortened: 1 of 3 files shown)


metadata-ingestion/src/datahub/ingestion/source/unity/connection.py
```diff
@@ -44,14 +57,62 @@ class UnityCatalogConnectionConfig(ConfigModel):
...
+            except Exception:
+                pass
+
+        except ImportError:
+            pass
+        except Exception:
+            pass
+
+        return None
+
```
ligfx: I don't believe this is needed at all, as the Databricks library is doing it internally. Works for me just fine after removing this.

ligfx: Alternately, the other way to do this is:

```python
if config.oauth_token() and config.oauth_token().access_token:
    return config.oauth_token().access_token
```

`oauth_service_principal` returns `None` in my testing.

ligfx:
Closing in favor of https://github.com/datahub-project/datahub/pull/15824

PR #15514, opened 2025-12-10
fix(ingestion/dremio): fix schema_resolver for non-standard uri length for dremio by implementing custom schema resolver

acrylJonny:

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_source.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_source.py
@@ -76,12 +76,12 @@ logger = logging.getLogger(__name__)
 
 
 class DremioSchemaResolver(SchemaResolver):
-    """Custom schema resolver for Dremio that handles the 'dremio.' prefix correctly.
+    """Custom schema resolver for Dremio that handles the 'dremio.' prefix/infix correctly.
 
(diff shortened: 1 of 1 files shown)


metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_source.py
```diff
@@ -73,6 +75,57 @@
 logger = logging.getLogger(__name__)
 
 
+class DremioSchemaResolver(SchemaResolver):
+    """Custom schema resolver for Dremio that handles the 'dremio.' prefix correctly.
+
+    Dremio always uses 'dremio' as the database, but SQL queries may reference tables
+    without this prefix (e.g., MySource.sales.orders). This resolver ensures that
```
skrydal, line 82: imho to be 100% precise, once we get to URN building it might not be a _prefix_ but could be an _infix_ as well (when there is a platform_instance defined). Although either description could confuse people.

skrydal (approved):
LGTM

metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_source.py
acrylJonny: I'll update the description

PR #15518, opened 2025-12-10
feat(cli): add --env option to ingest deploy command

pedro93:
## Summary
- Add `--env` option to `datahub ingest deploy` command to pass environment variables when deploying ingestion recipes
- Environment variables are specified as comma-separated KEY=VALUE pairs (e.g., `--env "VAR1=value1,VAR2=value2"`)
- Values are stored in `extraArgs` with `env:` prefix and made available to the executor at runtime

## Test plan
- [x] Added unit test `test_deploy_source_vars_with_env_vars` in `tests/unit/utilities/test_ingest_utils.py`
- [x] All existing tests pass
- [x] Linting passes
- [x] Tested locally, see screenshot:

Executed:
```bash
datahub ingest deploy --name "Snowflake Integration" --schedule "0 5 * * *" --time-zone "Europe/London" -c snowflake-demo.yaml --extra-env "SNOWFLAKE_WAREHOUSE=COMPUTE_WH,SNOWFLAKE_ROLE=ANALYST"
```

Generated the following in a local instance

Screenshot 2025-12-10 at 21 34 18


pedro93:
> LGTM
> 
> Just wondering if there is any security concern. eg Should we advise users not storing sensitive data in these env vars and use datahub secrets instead?

These env vars are usually not sensitive but configurations I don't think it's an issue.

Issue #15434, opened 2025-11-27
Support batch authorization via Authorizer


gallyamb:
As of now `Authorizer` supports only one authorization request. It's ok when authorization data is accessed in-memory, but for `Authorizer`s backed by a separate service such design adds unnecessary overhead. It'd be better to check for all grants in one network request, rather than doing multiple smaller requests

I propose adding a new method to `Authorizer` with a `BatchAuthorizationRequest` (the rest is omitted for clarity):

```java
public interface Authorizer extends Plugin {
  /** Authorizes an action based on the actor, the resource, and required privileges. */
  default AuthorizationResult authorize(@Nonnull final AuthorizationRequest request) {
    return new AuthorizationResult(request, AuthorizationResult.Type.DENY, "Not Implemented.");
  }

  /**
   * Authorizes an actions based on the actor, the resource, and required privileges
   */
  default BatchAuthorizationResult authorizeBatch(
      @Nonnull final BatchAuthorizationRequest batchRequest) {
    List<AuthorizationResult> results = batchRequest.getIndividualRequests()
        .map(this::authorize)
        .toList();
    return new BatchAuthorizationResult(batchRequest, results);
  }
}
```

Then `com.datahub.authorization.AuthUtil#isAuthorized(com.datahub.authorization.AuthorizationSession, com.datahub.authorization.DisjunctivePrivilegeGroup, com.datahub.authorization.EntitySpec)` will take care of calculating the final decision based on the returned results (taking into account disjunctive/conjunctive groups)

Existing `Authorizer` implementations wouldn't notice any difference, while new ones may take advantage of the batch form

gallyamb:
`AuthorizationContext` will also cache the results as now and will only forward requests that haven't been cached. Therefore, there will be no performance issues compared to the current state

gallyamb:
UPDATED: 2025.12.10

I've implemented batched authorization. And there are some differences compared to the proposal above:
1. the `BatchAuthorizationResult` contains a `Map<String, AuthorizationResult> results` field
2. we will provide a default `LazyHashMap` implementation, that would not actually compute a result at a moment of calling
3. therefore, this way implementations of the `Authorizer` that do not support batch authorization still would authorize one request at a time, while `Authorizer`s with batch support may do their authorization logic in a one request
4. `AuthorizationContext`, that is responsible for caching the authorization results for current request, will decorate a result with it's own `LazyHashMap`, that looks up for a cached value and only after cache miss calls the actual authorization code
[...]

PR #15523, opened 2025-12-11
feat (authorizer): add support for batch authorization via Authorizer #15434

gallyamb:
This PR addresses the issue #15434

I'm proposing to extend `Authorizer` interface with ability to authorize multiple authorization requests at a one invocation of the authorizer. This will allow `Authorizer` implementations to become more efficient in cases, where latency of the one authorization request processing is almost the same compared to the latency of multiple simultaneous authorization requests processing (e.g. when `Authorizer` authorizes requests via separate http/grpc service with batch auth endpoint)

------------

## Summary by CodeRabbit

* **New Features**
  * Added batch authorization support for checking multiple privileges simultaneously.
  * Implemented lazy evaluation of authorization results for improved performance.

* **Tests**
  * Updated authorization test infrastructure to support new batch-based authorization flows.

* **Chores**
  * Refactored authorization system to use batch operations internally.
  * Simplified authorization dependency injection across application components.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


gallyamb:
@coderabbitai review

metadata-operation-context/src/main/java/io/datahubproject/metadata/context/OperationContext.java
```diff
@@ -352,22 +351,18 @@ public AspectRetriever getAspectRetriever() {
...
-  public AuthorizationResult authorize(
-      @Nonnull String privilege,
+  public BatchAuthorizationResult authorize(
+      @Nonnull Set<String> privileges,
       @Nullable EntitySpec resourceSpec,
       @Nonnull Collection<EntitySpec> subResources) {
     return authorizationContext.authorize(
-        getSessionActorContext(), privilege, resourceSpec, subResources);
+        getSessionActorContext(), privileges, resourceSpec, subResources);
   }
```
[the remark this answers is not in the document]
gallyamb: > If that’s intentional

I need help here from maintainers. Is it really intentional?

metadata-service/openapi-servlet/src/test/java/io/datahubproject/openapi/operations/kafka/KafkaControllerTest.java
```diff
@@ -136,7 +136,7 @@ public void setupMocks() {
     AuthenticationContext.setAuthentication(authentication);
 
     // Setup AuthorizerChain to allow access
-    when(authorizerChain.authorize(any()))
+    when(authorizer.authorize(any()))
         .thenReturn(new AuthorizationResult(null, AuthorizationResult.Type.ALLOW, ""));
   }
```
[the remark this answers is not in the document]
gallyamb: That is not required in test, so it'll just confuse people

metadata-service/openapi-servlet/src/test/java/io/datahubproject/openapi/operations/v1/TraceControllerTest.java
```diff
@@ -273,16 +272,14 @@ public SystemInfoService systemInfoService() {
...
 
       Authentication authentication = mock(Authentication.class);
       when(authentication.getActor()).thenReturn(new Actor(ActorType.USER, "datahub"));
-      when(authorizerChain.authorize(any()))
-          .thenReturn(new AuthorizationResult(null, AuthorizationResult.Type.ALLOW, ""));
       AuthenticationContext.setAuthentication(authentication);
 
-      return authorizerChain;
+      return authorizer;
     }
```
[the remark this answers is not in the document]
gallyamb: Well, this test completes successfully. Have no idea why, though 😅

metadata-service/openapi-servlet/src/test/java/io/datahubproject/openapi/v3/controller/RelationshipControllerTest.java
```diff
@@ -88,16 +87,14 @@ public ElasticSearchGraphService graphService() {
...
 
       Authentication authentication = mock(Authentication.class);
       when(authentication.getActor()).thenReturn(new Actor(ActorType.USER, "datahub"));
-      when(authorizerChain.authorize(any()))
-          .thenReturn(new AuthorizationResult(null, AuthorizationResult.Type.ALLOW, ""));
       AuthenticationContext.setAuthentication(authentication);
 
-      return authorizerChain;
+      return authorizer;
     }
```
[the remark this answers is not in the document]
gallyamb: This is a common pattern among many other tests. I don't think that this should be changed within this PR

metadata-service/openapi-entity-servlet/src/main/resources/JavaSpring/apiController.mustache
```diff
@@ -5,7 +5,7 @@ import io.datahubproject.metadata.context.OperationContext;
 import com.linkedin.metadata.entity.EntityService;
 import com.linkedin.metadata.search.SearchService;
 import io.datahubproject.openapi.v1.entities.EntitiesController;
-import com.datahub.authorization.AuthorizerChain;
+import com.datahub.plugins.auth.authorization.Authorizer;
```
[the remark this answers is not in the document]
gallyamb: I think we need maintainer's opinion here. I agree that qualifier is a better choice. But I think that it should be `"primaryAuthorizer"`, not an `"authorizerChain"`

On the other side, using `Primary` leads to _less_ code changes without actually affecting datahub's functionality. It's just an another configuration option compared to the manual injected beans qualification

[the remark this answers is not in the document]
gallyamb: I meant not just to use qualifier with `"primaryAuthorizer"`, but also specify that bean name at the `AuthorizerChain` bean factory method

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/policy/DeletePolicyResolver.java
```diff
@@ -30,6 +30,9 @@ public CompletableFuture<String> get(final DataFetchingEnvironment environment)
             try {
               _entityClient.deleteEntity(context.getOperationContext(), urn);
               if (context.getAuthorizer() instanceof AuthorizerChain) {
+                // TODO: either we do not need to invalidate cache manually
```
david-leifker, line 33: Agree this does nothing useful in the case of multiple GMS instances. The policies would be updated after the given refresh interval.

metadata-service/factories/src/main/java/com/linkedin/gms/factory/auth/AuthorizerChainFactory.java
```diff
@@ -46,6 +47,7 @@ public class AuthorizerChainFactory {
 
   @Bean(name = "authorizerChain")
   @Scope("singleton")
+  @Primary
```
david-leifker: Why was this required? Is there another bean creating this bean?

metadata-service/factories/src/test/java/com/linkedin/metadata/ingestion/validation/ExecutionIngestionAuthSystemUserTest.java
```diff
@@ -58,28 +58,28 @@ public class ExecutionIngestionAuthSystemUserTest extends AbstractTestNGSpringCo
   @Autowired
   private OperationContext systemOperationContext;
 
-  @Autowired private AuthorizerChain authorizerChain;
```
david-leifker, line 67: While the AuthorizerChain is an Authorizer, I think the intent is more clear that there are multiple authorizers involved with granting access. From what I can tell this is just a cosmetic change and the AuthorizerChain is still being injected as the authorizer. Please revert this to continue to use the subclass AuthorizerChain here but also in the other classes (controllers, etc).

metadata-service/openapi-entity-servlet/src/main/resources/JavaSpring/apiController.mustache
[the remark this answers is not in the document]
david-leifker: I would re-introduce the use of the AuthorizerChain class and not the more generic Authorizer class. Fine with changing the qualifier to something like `primaryAuthorizerChain` in the future case where there may be different chains.

metadata-service/factories/src/test/java/com/linkedin/metadata/ingestion/validation/ExecutionIngestionAuthSystemUserTest.java
david-leifker: By design there are practically always multiple authorizers involved which I think further supports keeping the "chain" naming.

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/policy/DeletePolicyResolver.java
gallyamb: Should I resolve the TODOs and remove cache invalidation in this PR? Or just transform TODOs to the notes like "this is only useful for single-instance deployments" to make the intent clearer?

metadata-service/factories/src/test/java/com/linkedin/metadata/ingestion/validation/ExecutionIngestionAuthSystemUserTest.java
gallyamb: My intent here was not to only cosmetically "improve" the code base, but to decouple the datahub's components from the specific `Authorizer` implementation, leaving the decision to the configuration (that's why `@Primary` is used and `primaryAuthorizerChain` qualifier is the alternative to this)

But after a thinking a while I feel that we should not only revert to the using the `AuthorizerChain`, but also stop implementing the plugin `Authorizer` interface on the `AuthorizerChain`. Now `AuthorizerChain` have a method `authorize(AuthorizationRequest)`, that _should not_ be used within datahub, because it's just a proxy to the batch authorization method. Having that method available to the callers gives a space to misuse it instead of the batch authorization

I see two ways of achieving desired state:
- remove `implements Authorizer` from the `AuthorizerChain` and replace all `Authorizer` usages within datahub
- introduce internal `Authorizer`/`AuthorizerChain` interface, that has only batch authorization available, replace usages of the plugin `Authorizer` interface within the whole code base (except `DataHubAuthorizer`, ofc)

Does this resonate with you?

li-utils/src/main/java/com/datahub/test/authorization/ConstantAuthorizationResultMap.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.AbstractMap;
+import java.util.Map;
+import java.util.Set;
+import javax.annotation.Nonnull;
+
+/**
+ * Custom implementation of the {@link Map} for authorization results, that always returns constant
+ * decision
+ */
+public class ConstantAuthorizationResultMap extends AbstractMap<String, AuthorizationResult> {
```
RyanHolstien, line 15: Remove test class from production code

li-utils/src/test/java/com/datahub/authorization/ConstantAuthorizationResultMapTest.java
```diff
@@ -0,0 +1,35 @@
+package com.datahub.authorization;
+
+import static org.testng.Assert.*;
+
+import java.util.Random;
+import org.apache.commons.lang3.RandomStringUtils;
+import org.testng.annotations.Test;
+
+public class ConstantAuthorizationResultMapTest {
```
RyanHolstien: This is not a useful test, please remove.

li-utils/src/main/java/com/datahub/test/authorization/ConstantAuthorizationResultMap.java
gallyamb: Want to disagree with you. This is not test class, because I could imagine an `Authorizer` implementation, that returns exactly this result map: e.g. an authorizer, that allows all authorization requests for a specific actor, while other requests will be actually processed through authorization logic

I can do a little refactoring to replace `Authorizer#SYSTEM` implementation with batch method using this class. Are you ok with that?

li-utils/src/test/java/com/datahub/authorization/ConstantAuthorizationResultMapTest.java
gallyamb: What kind of test do you think would be a useful for that class (in case the above comment would not end up with moving `ConstantAuthorizationResultMap` to the tests sources)?

li-utils/src/main/java/com/datahub/test/authorization/ConstantAuthorizationResultMap.java
gallyamb: E.g. this could look like that
```java
  Authorizer SYSTEM =
      new Authorizer() {
        @Nonnull
        @Override
        public BatchAuthorizationResult authorizeBatch(
            @Nonnull BatchAuthorizationRequest batchRequest) {
          return new BatchAuthorizationResult(
              batchRequest,
              Constants.SYSTEM_ACTOR.equals(batchRequest.getActorUrn())
                  ? ConstantAuthorizationResultMap.builder()
                      .type(AuthorizationResult.Type.ALLOW)
                      .batchRequest(batchRequest)
                      .message("Granted by system actor " + batchRequest.getActorUrn())
                      .build()
                  : ConstantAuthorizationResultMap.builder()
                      .type(AuthorizationResult.Type.DENY)
                      .batchRequest(batchRequest)
                      .message("Only system user is allowed.")
                      .build());
        }
      };
```

RyanHolstien: This class is being used exclusively in test classes for mocking no-op results. No-oping the authorization result should not be masked in a custom map implementation, it should be done in the authorizer itself. I.e. why would an authorizer return this map rather than actually returning the allowed results that have been processed?

gallyamb: Fair enough. Will move this and `PredefinedAuthorizationResultMap` classes to the test sources

But I'm wondering how to do it the datahub-way? I saw, for example, the `io.datahubproject.test.metadata.context.TestAuthSession`, that is definitely not production class, but it stays in the main source set. Should I move those classes to the `com.datahub.test.authorization` package while keeping them in the main source set? If not, then I'll had to add `testImplementation` dependency on the test source set of the li-utils project to all other project, where they being used now?

RyanHolstien: You bring up a good point, we have some anti-patterns around this. I think moving them to that same test package is sufficient for now and we can take it as a backlog item to clean these test references in main sources to reduce the amount of unrelated refactoring that gets brought into this PR. As long as it gets put in a spot where it's clear that the intent is for tests.

metadata-service/factories/src/test/java/com/linkedin/metadata/ingestion/validation/ExecutionIngestionAuthSystemUserTest.java
gallyamb: For now just reverted the changes to original state

datahub-graphql-core/src/test/java/com/linkedin/datahub/graphql/resolvers/ingest/IngestionAuthUtilsTest.java
```diff
@@ -2,113 +2,105 @@
...
     QueryContext mockContext = Mockito.mock(QueryContext.class);
     when(mockContext.getOperationContext()).thenReturn(mock(OperationContext.class));
 
-    AuthorizationResult result = Mockito.mock(AuthorizationResult.class);
-    Mockito.when(result.getType()).thenReturn(AuthorizationResult.Type.DENY);
     Mockito.when(
             mockContext
                 .getOperationContext()
                 .authorize(
-                    argThat(allowedPrivileges::contains),
```
david-leifker: This removes test logic, please revert.

metadata-auth/auth-api/src/main/java/com/datahub/authorization/AuthUtil.java
```diff
@@ -528,31 +528,28 @@ public static boolean isAuthorized(
...
-
-    // Each privilege in a group _must_ all be true to permit the operation.
-    for (final String privilege : requiredPrivileges.getRequiredPrivileges()) {
-      // Create and evaluate an Authorization request.
-      if (isDenied(session, privilege, resourceSpec, subResources)) {
-        // Short circuit.
+  private static boolean isGroupAuthorized(
+      ConjunctivePrivilegeGroup group, Map<String, AuthorizationResult> privilegeResults) {
+    for (String privilege : group.getRequiredPrivileges()) {
+      AuthorizationResult authorizationResult = privilegeResults.get(privilege);
```
david-leifker: This introduced a possible bypass there an empty privilege group would grant access. Corrected would be the following, restoring the old behavior to deny access on an empty group

```
private static boolean isGroupAuthorized(
    ConjunctivePrivilegeGroup group, Map<String, AuthorizationResult> privilegeResults) {
  
  // Deny empty privilege groups - restores original security check
  if (group.getRequiredPrivileges().isEmpty()) {
    return false;
  }
  
  for (String privilege : group.getRequiredPrivileges()) {
    AuthorizationResult authorizationResult = privilegeResults.get(privilege);
    
    // Null safety - deny if privilege result is missing
    if (authorizationResult == null 
        || authorizationResult.getType() == AuthorizationResult.Type.DENY) {
      return false;
    }
  }
  return true;
}
```

gallyamb:
Hi @david-leifker @RyanHolstien!

I've fixed the comments you've left. Could you review this PR again, please?

Issue #15524, opened 2025-12-11
ownership extraction in grafana source does not work


daha:
**Describe the bug**
Owners are not ingested properly for dashboards in the grafana source. 

The problem is that the code is looking under dashboard.meta (in the dashboard json) but meta is located at the same level as dashboard, https://github.com/grafana/grafana/blob/main/pkg/api/dashboard.go#L261C2-L264C3 so the code does not correctly ingest owners, will probably not map in dashboards correctly to folders either. 

**To Reproduce**

Use the grafana source to some Grafana instance, where dashboards has been created manually. 

**Expected behavior**
It should extract owners. 

**Screenshots**

**Desktop (please complete the following information):**

- OS: Linux debian 12.12
- Version Datahub version (HEAD, or any older version with grafana source)

**Additional context**

PR #15525, opened 2025-12-11
fix(ingestion/grafana): fix broken ownership ingestion

daha:
`meta` where folder_id and created_by is extracted from is located at the top level not under dashboard, from [`/api/dashboards/uid/:uid`](https://grafana.com/docs/grafana/latest/developer-resources/api-reference/http-api/dashboard/#get-dashboard-by-uid). https://github.com/grafana/grafana/blob/4ab198b201acc42077463518d1697ed79a0ce77e/pkg/api/dashboard.go#L261C2-L264C3

Fixes: #15524 

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/grafana/models.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/grafana/models.py
@@ -240,7 +240,6 @@ class Dashboard(_GrafanaBaseModel):
         result.setdefault("version", None)
         result.setdefault("timezone", None)
         result.setdefault("refresh", None)
-        result.setdefault("created_by", None)
 
     @staticmethod
     def _cleanup_dashboard_metadata(result: Dict[str, Any]) -> None:
@@ -276,12 +275,13 @@ class Dashboard(_GrafanaBaseModel):
             else:
                 panels = _panel_data
 
-            meta = dashboard_data.get("meta", {})
-            folder_id = meta.get("folderId") if meta else None
-
             result = {**dashboard_data, "panels": panels}
-            if folder_id is not None:
-                result["folder_id"] = folder_id
+
+            if (meta := data.get("meta")) is not None:
+                # We only want to set folder_id and created_by from meta if we get it
+                # from json and not when creating Dashboard(uid=...) like in tests.
+                result["folder_id"] = meta.get("folderId")
+                result["created_by"] = meta.get("createdBy")
 
             cls._set_dashboard_defaults(result)
             cls._cleanup_dashboard_metadata(result)
--- a/metadata-ingestion/src/datahub/ingestion/source/grafana/entity_mcp_builder.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/grafana/entity_mcp_builder.py
@@ -206,7 +206,7 @@ def build_dashboard_mcps(
     )
 
     # Ownership aspect
-    if dashboard.uid and ingest_owners:
+    if ingest_owners:
         owner = _build_ownership(dashboard, remove_email_suffix)
         if owner:
             mcps.append(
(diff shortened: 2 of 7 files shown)


daha:
A comment I have run this code towards a production instance of Grafana Enterprise and with this PR owners are finally ingested properly.

deepgarg760:
@daha , Thanks for raising this

metadata-ingestion/src/datahub/ingestion/source/grafana/models.py
```diff
@@ -276,12 +275,13 @@ def extract_dashboard_data(cls, data: Any) -> Dict[str, Any]:
...
-            meta = dashboard_data.get("meta", {})
-            folder_id = meta.get("folderId") if meta else None
-
             result = {**dashboard_data, "panels": panels}
-            if folder_id is not None:
-                result["folder_id"] = folder_id
+
+            if (meta := data.get("meta")) is not None:
+                # We only want to set folder_id and created_by from meta if we get it
+                # from json and not when creating Dashboard(uid=...) like in tests.
```
sgomezvillamor, line 282:
```suggestion
                # "meta" is root field in the json
                # https://grafana.com/docs/grafana/latest/developer-resources/api-reference/http-api/dashboard/#get-dashboard-by-uid
```

> "not when creating Dashboard(uid=...) like in tests"? 
Do you mean this is wrong?
https://github.com/datahub-project/datahub/blob/3e00d3f655caa3a11d365596dd99072844840748/metadata-ingestion/tests/unit/grafana/test_grafana_entity_mcp_builder.py#L34-L46

sgomezvillamor:
@acrylJonny can you also have a look to this fix in grafana source?

metadata-ingestion/src/datahub/ingestion/source/grafana/models.py
daha: meta is not mentioned in the dashboard model, I think it is a lot clearer to define the Dashboard with folder_id directly, than trying to mimic parts of how it is transformed when loading from json. 
Either should both created_by and folder_id be defined directly in Dashboard or both createdBy and folderId be defined in meta, but the later is a bit confusing as it is transformed into a Dashboard with folder_id and created_by, and meta is removed. 

EDIT: 
At 3e00d3f It works to populate folder_id only form meta in Dashboard, it does not populate created_by that way. It is also looking for meta at the wrong location, so it does not work when it loads data from json, so there is a miss match between the unit tests and real usage.  

I can possibly update the code a little to allow to create Dashboard with `meta={"folderId": "123"},` if there are forks or transformers or similar out there relying on that creation pattern.

acrylJonny:
> @acrylJonny can you also have a look to this fix in grafana source?

I think this looks good. I'd like to see a test for the edge case test where there is a missing meta field. After that I think we're all good.

daha:
> > @acrylJonny can you also have a look to this fix in grafana source?
> 
> I think this looks good. I'd like to see a test for the edge case test where there is a missing meta field. After that I think we're all good.

I have created a few tests with missing createdBy, and test with missing folderId, and minor updates to some of the other tests regarding created_by and folder_id.

acrylJonny:
> > > @acrylJonny can you also have a look to this fix in grafana source?
> > 
> > 
> > I think this looks good. I'd like to see a test for the edge case test where there is a missing meta field. After that I think we're all good.
> 
> I have created a few tests with missing createdBy, and test with missing folderId, and minor updates to some of the other tests regarding created_by and folder_id.

Thank you for that. I think we're looking good here. Thank you for the contribution.

acrylJonny (approved):
Looks good to me

metadata-ingestion/src/datahub/ingestion/source/grafana/models.py
sgomezvillamor: If the tests are not representative (eg if there are tests construct Dashboard objects directly with folder_id) of production behavior, that should be fixed.

`Dashboard` is a model entity for grafana source and should not be used anywhere else than in this source. So no risk of breaking change impacting others.

PR #15528, opened 2025-12-11
fix(ingestion): fix updated secret field

v-tarasevich-blitz-brain:
Linear: https://linear.app/acryl-data/issue/CAT-998/ensure-we-still-support-secret-creation-and-selection-when-configuring

Fixed the error after creating of a new secret and added tests


chriscollins3456 (approved):
good fix

PR #15532, opened 2025-12-11
feat(context): Context Base V1 Fix Ups

jjoyce0510:
# Summary

As we worked towards townhall, we continued to press on support for ingesting "external" documents from tools like Notion. This PR mainly includes fixups that help us more gracefully handle the presence of both external documents, and those created natively on DataHub. 

Specifics: 

1. Support retrieving the platform / platform instance for a document via GraphQL.  Map it in the same way we do for Data Jobs. 
2. [bug fix] Filter out context documents that are NOT supposed to shown in Global Context from the searchDocuments API. This is important so that we can record "hidden" memories that are not shown in the global context sidebar. Previously, we had this setting but it was NOT being honored correctly. 
3. By default, filter out documents that are not supposed to be shown in global context OR are unpublished from the global search + scroll APIs -> This is important because we DO want external notion documents AND published context documents to show in search results and scroll results, but NOT those which are either a) hidden from global context by setting or b) unpublished (those are only shown to the user who owns them in the context docs sidebar). Added new python smoke tests here. 
4. Correctly show custom document "types" in the dropdown type selector (if ingested from the backend) 
5. When fetching "related" documents for an entity (e.g. snowflake table), include native + external documents. 
6. Adding Notion as a data platform + datahub

### DIFF
--- a/datahub-web-react/src/app/entityV2/document/__tests__/DocumentEntity.test.tsx
+++ b/datahub-web-react/src/app/entityV2/document/__tests__/DocumentEntity.test.tsx
@@ -1,0 +1,1540 @@
+import { MockedProvider } from '@apollo/client/testing';
+import { render, screen, waitFor } from '@testing-library/react';
+import React from 'react';
+import { MemoryRouter } from 'react-router';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import EntityContext from '@app/entity/shared/EntityContext';
+import { PreviewType } from '@app/entityV2/Entity';
+import { DocumentNativeProfile } from '@app/entityV2/document/DocumentNativeProfile';
+import { Preview } from '@app/entityV2/document/preview/Preview';
+import EntitySidebarContext, { entitySidebarContextDefaults } from '@app/sharedV2/EntitySidebarContext';
+import CustomThemeProvider from '@src/CustomThemeProvider';
+import { mocks } from '@src/Mocks';
+
+import { Document, DocumentSourceType, DocumentState, EntityType } from '@types';
+
+// Mock entity registry with all required methods
+const mockEntityRegistry = {
+    getEntityUrl: (_entityType: EntityType, urn: string) => `/document/${urn}`,
+    getPathName: (_entityType: EntityType) => 'document',
+    getIcon: () => null,
+    getEntity: () => ({
+        type: EntityType.Document,
+        getCollectionName: () => 'Documents',
+        getEntityName: () => 'Document',
+        getPathName: () => 'document',
+        getGraphName: () => 'document',
+        supportedCapabilities: () => new Set(),
+    }),
+    getDisplayName: () => 'Test Document',
+    getCollectionName: () => 'Documents',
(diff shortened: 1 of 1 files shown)


datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/types/knowledge/DocumentMapper.java
```diff
@@ -72,14 +72,28 @@ public static Document map(@Nullable QueryContext context, final EntityResponse
...
+      // Only set dataPlatformInstance if there's an actual instance (to avoid null urn/type errors)
+      if (dataPlatformInstance.hasInstance()) {
+        result.setDataPlatformInstance(value);
+      }
+    } else {
+      // No DataPlatformInstance aspect - default to DataHub platform for internal documents
+      result.setPlatform(
+          com.linkedin.datahub.graphql.generated.DataPlatform.builder()
+              .setType(EntityType.DATA_PLATFORM)
+              .setUrn("urn:li:dataPlatform:datahub")
```
jjoyce0510: Add unit tests

datahub-web-react/src/app/document/hooks/useSearchDocuments.ts
```diff
@@ -13,9 +13,20 @@ export interface SearchDocumentsInput {
     count?: number;
     fetchPolicy?: 'cache-first' | 'cache-and-network' | 'network-only';
     includeParentDocuments?: boolean;
+    /**
```
jjoyce0510, line 16: TODO: Clean this up. It feels like we should just require 

sourceType or null.

datahub-web-react/src/app/entityV2/shared/tabs/Documentation/components/AddContextDocumentPopover.tsx
```diff
@@ -159,6 +159,9 @@ export const AddContextDocumentPopover: React.FC<AddContextDocumentPopoverProps>
             hideActionsMenu
             maxHeight={400}
             searchDisabled={isCreating}
+            // Search all document types (both native and external/ingested)
+            // to allow linking documents from third-party sources like Notion
+            sourceType={null}
```
jjoyce0510: Maybe just make this more explicit. 

sourceTypes: [NATIVE, EXTERNAL] 

tis is actually more fool proof.

datahub-web-react/src/app/previewV2/EntityHeader.tsx
```diff
@@ -94,17 +94,28 @@ const EntityHeader: React.FC<EntityHeaderProps> = ({
 }) => {
     const linkProps = useEmbeddedProfileLinkProps();
 
+    // Handle click on the link - run custom onClick before navigation
+    const handleLinkClick = React.useCallback(
```
jjoyce0510: The previous behavior straight up would not work. The link was intercepting the clicks.

shirshanka (approved):
LGTM

PR #15536, opened 2025-12-11
feat(context): Remove "drafts" concept pre-release

jjoyce0510:
We're removing the draftOf field which has not yet been used before the v1 release.
We do not have immediate plans of implementing this.


chriscollins3456 (approved):
nice rice

PR #15539, opened 2025-12-12
fix(ui): Adding a fallback icon for failed images on Lineage cards

sakethvarma397:
The platform icons can fail to load because of CORS errors or missing resources. Adding fallback icons for such cases so that
- We have a better UX of seeing the placeholder icon instead of a failed image logo. We've used this placeholder icon for EntityHeader previously.
- The screenshot functionality doesn't break ( It currently fails if any of the platform icons fail to load)

reactflow_2025-12-12_125736

### DIFF
--- a/datahub-web-react/src/app/lineageV3/components/LineageCard.tsx
+++ b/datahub-web-react/src/app/lineageV3/components/LineageCard.tsx
@@ -126,7 +126,7 @@ const PlatformIcon = ({ src, alt }: { src: string; alt: string }) => {
     const [errored, setErrored] = React.useState(false);
 
     if (errored) {
-        return <Icon size="4xl" source="phosphor" icon="Placeholder" />;
+        return <Icon size="lg" source="phosphor" icon="Placeholder" />;
     }
 
     return <StyledPlatformIcon src={src} alt={alt} onError={() => setErrored(true)} />;


datahub-web-react/src/app/lineageV3/components/LineageCard.tsx
```diff
@@ -116,11 +116,22 @@ const ColumnButtonWrapper = styled.div`
...
     height: 16px;
     width: 16px;
 `;
 
+// Defaults to phosphor icon if the img fails to load for any reason
+const PlatformIcon = ({ src, alt }: { src: string; alt: string }) => {
+    const [errored, setErrored] = React.useState(false);
+
+    if (errored) {
+        return <Icon size="4xl" source="phosphor" icon="Placeholder" />;
```
chriscollins3456: should this be 4xl? it looks larger than other platform icons - it should be the same size at 16px which it looks like is `lg` while `4xl` is 24px

chriscollins3456 (commented):
let's double check sizes

datahub-web-react/src/app/lineageV3/components/LineageCard.tsx
sakethvarma397: Yes, lg is 16px. I'm reducing the size.

PR #15543, opened 2025-12-12
feat(ingestion): add discard changes confirmation

v-tarasevich-blitz-brain:
Linear: https://linear.app/acryl-data/issue/CAT-1024/add-modal-to-confirm-user-wants-to-leave-in-createedit-source-after-an

This PR adds confirmation to discard unsaved changes when an user presses Cancel, clicks another link, closes the tab
Also added a few fixes for filters

https://github.com/user-attachments/assets/edd9aeaa-3579-40f3-a0e8-b4c920ebb3f3

Updated confirmation modal:

image

image

### DIFF
--- a/datahub-web-react/src/app/ingestV2/source/multiStepBuilder/IngestionSourceCreatePage.tsx
+++ b/datahub-web-react/src/app/ingestV2/source/multiStepBuilder/IngestionSourceCreatePage.tsx
@@ -1,6 +1,7 @@
 import { useApolloClient } from '@apollo/client';
+import { Text } from '@components';
 import { message } from 'antd';
-import React, { useCallback } from 'react';
+import React, { useCallback, useState } from 'react';
 import { useHistory } from 'react-router';
 
 import analytics, { EventType } from '@app/analytics';
@@ -23,6 +24,7 @@ import {
     getIngestionSourceSystemFilter,
     getNewIngestionSourcePlaceholder,
 } from '@app/ingestV2/source/utils';
+import { DiscardUnsavedChangesConfirmationProvider } from '@app/sharedV2/confirmation/DiscardUnsavedChangesConfirmationContext';
 import { useOwnershipTypes } from '@app/sharedV2/owners/useOwnershipTypes';
 import { PageRoutes } from '@conf/Global';
 
@@ -53,6 +55,7 @@ const STEPS: IngestionSourceFormStep[] = [
 export function IngestionSourceCreatePage() {
     const history = useHistory();
     const client = useApolloClient();
+    const [isSubmitting, setIsSubmitting] = useState<boolean>(false);
 
     const createIngestionSource = useCreateSource();
 
@@ -67,6 +70,7 @@ export function IngestionSourceCreatePage() {
     const onSubmit = useCallback(
         async (data: MultiStepSourceBuilderState | undefined, options: SubmitOptions | undefined) => {
             if (!data) return undefined;
+            setIsSubmitting(true);
             const shouldRun = options?.shouldRun;
             const input = getIngestionSourceMutationInput(data);
 
@@ -117,6 +121,7 @@ export function IngestionSourceCreatePage() {
                 }
             }
 
+            setIsSubmitting(false);
             return undefined;
         },
         [createIngestionSource, history, client, defaultOwnershipType],
@@ -126,5 +131,20 @@ export function IngestionSourceCreatePage() {
         history.push(PageRoutes.INGESTION);
     }, [history]);
 
-    return <IngestionSourceBuilder steps={STEPS} onSubmit={onSubmit} onCancel={onCancel} initialState={initialState} />;
+    return (
+        <DiscardUnsavedChangesConfirmationProvider
+            enableRedirectHandling={!isSubmitting}
+            confirmationModalTitle="You have unsaved change"
+            confirmationModalText={
+                <>
+                    <Text type="span">You have unsaved changes to your new source. </Text>
+                    <Text type="span" weight="bold">
+                        Are you sure you want to leave and discard your unsaved changes?
+                    </Text>
+                </>
+            }
+        >
+            <IngestionSourceBuilder steps={STEPS} onSubmit={onSubmit} onCancel={onCancel} initialState={initialState} />
+        </DiscardUnsavedChangesConfirmationProvider>
+    );
 }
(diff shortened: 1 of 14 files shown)


datahub-web-react/src/app/sharedV2/confirmation/DiscardUnsavedChangesConfirmationContext.tsx
```diff
@@ -0,0 +1,121 @@
...
+    }, [history, lastRedirectLocation]);
+
+    return (
+        <DiscardUnsavedChangesConfirmationContext.Provider value={{ setIsDirty, showConfirmation }}>
+            {children}
+
+            <ConfirmationModal
+                isOpen={isConfirmationShown}
+                modalTitle={confirmationModalTitle ?? 'Discard unsaved changes?'}
+                modalText={confirmationModalText ?? 'Your changes will be lost.'}
```
chriscollins3456: We can make this a little more verbose by default like "Your changes will be lost. Are you sure you want to leave?"

datahub-web-react/src/app/ingestV2/source/multiStepBuilder/IngestionSourceUpdatePage.tsx
```diff
@@ -135,19 +141,56 @@ export function IngestionSourceUpdatePage() {
...
+        ) => {
+            const defaultsForFields = {
+                isConnectionDetailsValid: null,
+                owners: null,
+                ingestionSource: null,
+            };
+            const initialStateWithoutOwners = {
+                ...(initialStateToCheck ?? {}),
+                ...defaultsForFields,
+            };
```
chriscollins3456: is this correct? should we have the default spread first and then after add `...initialStateToCheck ??{})?

datahub-web-react/src/app/ingestV2/source/multiStepBuilder/IngestionSourceForm.tsx
```diff
@@ -1,9 +1,13 @@
...
 
+import { useDiscardUnsavedChangesConfirmationContext } from '@app/sharedV2/confirmation/DiscardUnsavedChangesConfirmationContext';
 import { useMultiStepContext } from '@app/sharedV2/forms/multiStepForm/MultiStepFormContext';
 
 export function IngestionSourceForm() {
-    const { getCurrentStep } = useMultiStepContext();
+    const { getCurrentStep, isDirty } = useMultiStepContext();
+
+    const { setIsDirty } = useDiscardUnsavedChangesConfirmationContext();
+    useEffect(() => setIsDirty(isDirty()), [isDirty, setIsDirty]);
```
chriscollins3456, line 10: hmm it feels weird to have two pieces of state here from different placed and to update with a `useEffect`. 

what do you think about simplifying things and just keep it in the one context it needs to be in? since we need to know `isDirty` in the unsaved changed context, could we update that piece of state inside of the multistep context instead of having this useEffect handle that situation in a seemingly random component?

chriscollins3456: also would it actually update when you do or don't make changes since `isDirty` is a function and not a value? maybe so since that function gets updated with a useMemo dependency on state

datahub-web-react/src/app/ingestV2/source/multiStepBuilder/IngestionSourceUpdatePage.tsx
```diff
@@ -135,19 +141,56 @@ export function IngestionSourceUpdatePage() {
         history.push(ingestionSourcesListBackUrl ?? PageRoutes.INGESTION);
     }, [history, ingestionSourcesListBackUrl]);
 
+    const isDirtyChecker = useCallback(
```
chriscollins3456, line 145: also just curious why we need this isDirtyChecker when we can just do a deepEqual on the initial state or the current state? i could be missing something of course

chriscollins3456 (commented):
cool beans! i have a few questions about how we could potentially simplify this and some comments. let me know what you think

datahub-web-react/src/app/ingestV2/source/multiStepBuilder/IngestionSourceUpdatePage.tsx
v-tarasevich-blitz-brain: I'll rename it. Here we have to exclude a few fields from deepEqual comparison

datahub-web-react/src/app/ingestV2/source/multiStepBuilder/IngestionSourceUpdatePage.tsx
v-tarasevich-blitz-brain: we need that as some fields could have diff without a real changes

datahub-web-react/src/app/ingestV2/source/multiStepBuilder/IngestionSourceCreatePage.tsx
```diff
@@ -126,5 +131,20 @@ export function IngestionSourceCreatePage() {
         history.push(PageRoutes.INGESTION);
     }, [history]);
 
-    return <IngestionSourceBuilder steps={STEPS} onSubmit={onSubmit} onCancel={onCancel} initialState={initialState} />;
+    return (
+        <DiscardUnsavedChangesConfirmationProvider
+            enableRedirectHandling={!isSubmitting}
+            confirmationModalTitle="Save your new source"
```
chriscollins3456: i think this is wrong (although i know that's what the design says) - let's change it to simply "You have unsaved change"

datahub-web-react/src/app/ingestV2/source/multiStepBuilder/IngestionSourceUpdatePage.tsx
```diff
@@ -136,19 +142,68 @@ export function IngestionSourceUpdatePage() {
...
-            onSubmit={onSubmit}
-            onCancel={onCancel}
-            initialState={{
-                ...mapSourceTypeAliases(removeExecutionsFromIngestionSource(ingestionSourceData.ingestionSource)),
-                ...{ isEditing: true, ingestionSource: ingestionSourceData.ingestionSource as IngestionSource },
-            }}
-        />
+        <DiscardUnsavedChangesConfirmationProvider
+            enableRedirectHandling={!isSubmitting}
+            confirmationModalTitle="Save your source"
```
chriscollins3456: let's make the same copy change here

PR #14580, opened 2025-08-28
refactor(ingestion): Updated Redshift lineage_v1 refs and removed v1 implementation

askumar27:
Refactored code to remove redshift lineage v1 implementation. Redshift lineage v2 has been default for sometime through config. 

Summary:
- Deleted `lineage.py` (v1 implementation)
- Updated `lineage_v2.py` with concrete implementation of refs from v1
- Updated `redshift.py` removed refs from v1
- Updated unit tests
- Removed v1 back setting from config
- Updated docs for upcoming breaking change
- Passing Unit and Connecter Integration Tests

PR #15545, opened 2025-12-12
fix(redshift): fix lineage extraction ignoring disabled flags (regression from #14580)

sgomezvillamor:
**Summary**

Fixes a critical performance regression introduced in PR #14580 where the removal of lineage v1 inadvertently caused lineage extraction to run even when all lineage flags were disabled in the config.

**Problem**

After PR #14580 removed lineage v1, the config check that wrapped lineage extraction was accidentally removed. This caused:

- Lineage extraction to run for 100+ minutes even when disabled
- Customer reports 7x slower ingestion (17 min → 2+ hours)
- Memory usage increased 3.5x (479 MB → 1.68 GB)
- Thousands of SQL queries parsed unnecessarily

*Customer Impact:* Ingestion degraded from 30 minutes (v1.2.0.6) to 2-8 hours (v1.2.0.9+) due to this bug.

**Solution**

Re-add the config flag check before calling `extract_lineage_v2()`:

```python
if (
    self.config.include_table_lineage
    or self.config.include_view_lineage
    or self.config.include_copy_lineage
    or self.config.include_unload_lineage
    or self.config.include_share_lineage
    or self.config.include_table_rename_lineage
):
    with self.report.new_stage(LINEAGE_EXTRACTION):
        yield from self.extract_lineage_v2(...)

### CODE REVIEW

File: metadata-ingestion/src/datahub/ingestion/source/redshift/redshift.py [hunk 1]
```diff
@@ -427,11 +427,26 @@ def _extract_metadata(
...
+            # This addresses a regression introduced in PR #14580 where lineage v1 removal
+            # inadvertently caused lineage extraction to run even when all flags were disabled.
+            if (
+                self.config.include_table_lineage
+                or self.config.include_view_lineage
+                or self.config.include_copy_lineage
+                or self.config.include_unload_lineage
+                or self.config.include_share_lineage
+                or self.config.include_table_rename_lineage
+            ):
```
askumar27, line 440: [COSMETIC] Maybe we can use `any` - more pythonic :) 

```
if any([
      self.config.include_table_lineage,
      self.config.include_view_lineage,
      self.config.include_copy_lineage,
      self.config.include_unload_lineage,
      self.config.include_share_lineage,
      self.config.include_table_rename_lineage,
  ]):
```

[MINOR] Add a test case to prevent future regression? Add this check as a `@property` in `config.py`  or a helper function - can be helpful for testing.
```
@property
def any_lineage_enabled(self) -> bool:
```

PR #15547, opened 2025-12-12
feat(ingestion): add analytics events

v-tarasevich-blitz-brain:
Linera:
- https://linear.app/acryl-data/issue/CAT-961/add-telemetry-for-test-connection-button
- https://linear.app/acryl-data/issue/CAT-1005/capture-events-on-source-create-and-source-edit

This PR adds/updates analytics events:
- when an user clicks Test connection button
- when an user closes Test connection modal
- when an user creates an Ingestion Source
- when an user updates an Ingestion Source


chriscollins3456 (approved):
nice!

PR #15670, opened 2025-12-15
feat(bigquery): add pushdown_allow_usernames and pushdown_deny_usernames for query filtering

dinesh-verma-datahub:
## Summary

Add support for filtering BigQuery queries by user email patterns at the query level, similar to the existing Snowflake implementation. This enables users to include or exclude specific users from lineage, usage, and query extraction.

## Improvements

- **Performance improvement**: Filter out high-volume service accounts or bot users directly at the BigQuery query level, reducing data processed
- **Test isolation**: Allows connector-tests to filter queries by username, preventing test data pollution from concurrent test runs and enabling more deterministic validation
- **Feature parity**: Aligns BigQuery with Snowflake's existing `pushdown_allow_usernames` and `pushdown_deny_usernames` capabilities

### New Config Options

| Config | Description |
|--------|-------------|
| `pushdown_deny_usernames` | List of user emails (SQL LIKE patterns) to **exclude** from extraction |
| `pushdown_allow_usernames` | List of user emails (SQL LIKE patterns) to **include** in extraction |

### Example Usage

source:
  type: bigquery
  config:
    use_queries_v2: true  # Required
    pushdown_deny_usernames:
      - "service_%@%.iam.gserviceaccount.com"  # Exclude service accounts
      - "%_bot@example.com"                     # Exclude bot users
    pushdown_allow_usernames:
      - "%@mycompany.com"                       # Only include company users### Implementation Details

- Uses `LOWER()` for case-insensitive matching (BigQuery doesn't support `ILIKE`)
- Deny patterns use AND logic (all must match to exclude)
- Allow patterns use OR logic (any can match to include)
- SQL injection protection via single quote escaping
- Only applicable when `use_queries_v2: true` is enabled

### DIFF
--- a/metadata-ingestion/docs/sources/bigquery/bigquery_pre.md
+++ b/metadata-ingestion/docs/sources/bigquery/bigquery_pre.md
@@ -123,6 +123,7 @@ DataHub's BigQuery connector supports two approaches for extracting lineage and 
   - Query popularity statistics and rankings
   - Multi-region support via `region_qualifiers`
   - Table and column-level usage statistics
+  - User filtering with patterns (`pushdown_deny_usernames`, `pushdown_allow_usernames`)
 - **Requirements**:
   - `bigquery.jobs.listAll` permission on target projects
   - No additional Cloud Logging permissions needed


dinesh-verma-datahub:
@skrydal Thanks, closing this, will take the suggested approach and raise a new PR.

PR #15686, opened 2025-12-16
feat(ui/login): implement redesign of login page

purnimagarg1:
**Linear ticket:**
https://linear.app/acryl-data/issue/CAT-1019/redesign-the-regular-login-page-to-match-designs-of-new-sign-up-page

**Description:**

Implements the login page redesign

Depends on this [PR](https://github.com/datahub-project/datahub/pull/15684)

**Screenshot:**

image

### DIFF
--- a/datahub-web-react/src/app/auth/shared/ModalHeader.tsx
+++ b/datahub-web-react/src/app/auth/shared/ModalHeader.tsx
@@ -21,7 +21,11 @@ const HeaderText = styled.div`
     gap: 4px;
 `;
 
-export default function ModalHeader() {
+interface Props {
+    subHeading?: string;
+}
+
+export default function ModalHeader({ subHeading }: Props) {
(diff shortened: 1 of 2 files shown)


datahub-web-react/src/app/auth/loginV2/LoginModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+        setIsSubmitDisabled(hasErrors || !isTouched);
+    };
+
+    const handleSSOLogin = () => {
+        window.location.href = resolveRuntimePath('/sso');
+    };
+
+    return (
+        <Modal
+            title={<ModalHeader />}
```
chriscollins3456, line 80: ugh i know the designs say to reuse the "Before we get started we just have a few questions" but i don't think that makes sense for the login page.

maybe we can just remove that copy for login

PR #15698, opened 2025-12-17
Add token refresh for UC Lineage ingestion using system tables

Anshul759:
If using system tables to query lineage, the sql connection fails. 
**Reason -** sql connection requires PAT to connect to databricks and run queries. As the token is valid for only 60mins, and some of the ingestions run for long duration, the sql connection expires and the process fails with authentication error. 
**Fix -** To use azure credentials to generate tokens. Have updated the code to generate PAT using azure credentials. The validity of PAT is 60 mins here too, but we can refresh the PAT before expiring so that the process continues without any halt

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
@@ -1404,11 +1404,9 @@ class UnityCatalogApiProxy(UnityCatalogProxyProfilingMixin):
         try:
             # Initialize or refresh token before executing query
             # This lazy initialization ensures we only fail when SQL is actually needed
-            fresh_token = self._token_provider.get_token()
-            if fresh_token != self._sql_connection_params["access_token"]:
-                logger.debug("Updating SQL connection token")
-                self._sql_connection_params["access_token"] = fresh_token
-
+            self._sql_connection_params["access_token"] = (
+                self._token_provider.get_token()
+            )
             with (
                 connect(**self._sql_connection_params) as connection,
                 connection.cursor() as cursor,


metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
```diff
@@ -391,6 +406,79 @@ def check_basic_connectivity(self) -> bool:
...
         )
 
+    def _should_refresh_token(self) -> bool:
+        """Check if the token should be refreshed based on expiration time."""
+        if self._token_expiration is None:
+            return True
+
+        # Refresh token if it expires within 5 minutes
+        remaining_time = self._token_expiration - datetime.now(timezone.utc)
+        return remaining_time < timedelta(minutes=10)
```
kyungsoo-datahub: Do we want to refresh within 10 mins before expiration?  Line 414 says that it would be within 5 mins.

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
```diff
@@ -391,6 +406,79 @@ def check_basic_connectivity(self) -> bool:
...
+            self._current_token = str(token)
+            # Set default expiration
+            self._token_expiration = datetime.now(timezone.utc) + timedelta(hours=1)
+
+        return self._current_token or ""
+
+    def _refresh_workspace_client(self):
+        """Refresh the workspace client with new credentials."""
+        if self._azure_auth:
+            logger.info("Refreshing workspace client with new Azure credentials")
```
kyungsoo-datahub: It would be good to be debug level logging.

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
```diff
@@ -391,6 +406,79 @@ def check_basic_connectivity(self) -> bool:
...
+            # Create a new workspace client with fresh credentials
+            self._workspace_client = WorkspaceClient(
```
kyungsoo-datahub: Based on the brief investigation, recreating client is redundant. authenticate() will refresh token if it's expired. Please, check this.

authenticate() -> _header_factory() (config.py) -> OAuthCredentialsProvider.__call__() (credentials_provider.py) -> refreshed_headers() (credentials_provider.py) -> inner.token() (credentials_provider.py) -> token() (oath.py) -> _blocking_token() (oath.py) -> refresh() (oath.py)

Please, check token() and _blcoking_token() functions: https://github.com/databricks/databricks-sdk-py/blob/main/databricks/sdk/oauth.py#L280-L333

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
```diff
@@ -391,6 +406,79 @@ def check_basic_connectivity(self) -> bool:
...
+                host=self._workspace_client.config.host,
+                azure_tenant_id=self._azure_auth.tenant_id,
+                azure_client_id=self._azure_auth.client_id,
+                azure_client_secret=self._azure_auth.client_secret.get_secret_value(),
+                product="datahub",
+                product_version=nice_version_name(),
+            )
+
+    def _refresh_sql_connection_token(self):
```
kyungsoo-datahub: I'm wondering if it's safe with multi threads.

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
Anshul759: Done

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
Anshul759: Have used authenticate function, so removed this part of the code

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
Anshul759: Thanks for the suggestion. authenticate function does refresh the token when expired.
For Testing, ran pipeline for around 3 hours without any error - 
image

Anshul759:
@kyungsoo-datahub 
1. Have added unit test cases
2. Seperated the token management in a different class
3. Used inbuild authenticate function to refresh token instead of using custom logic

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
```diff
@@ -1404,9 +1393,10 @@ def _execute_sql_query(self, query: str, params: Sequence[Any] = ()) -> List[Row
 
         try:
             # Refresh token if needed before executing query
-            if self._azure_auth and self._should_refresh_token():
-                logger.info("Refreshing SQL connection token as it is about to expire")
-                self._refresh_sql_connection_token()
+            fresh_token = self._token_provider.get_token()
```
kyungsoo-datahub, line 484: minor: Can we just do like

```
self._sql_connection_params["access_token"] = self._token_provider.get_token()
```
?

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
```diff
@@ -213,18 +279,14 @@ def __init__(
...
-            if azure_auth
-            else self._workspace_client.config.token
-        )
+        initial_token = self._token_provider.get_token()
         self._sql_connection_params = {
             "server_hostname": self._workspace_client.config.host.replace(
                 "https://", ""
             ),
             "http_path": f"/sql/1.0/warehouses/{self.warehouse_id}",
             "access_token": initial_token,
```
kyungsoo-datahub: Don't think we need intial_token variable. We can assign the value directly from the function
```
"access_token": self._token_provider.get_token(),
```

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
```diff
@@ -160,16 +160,93 @@ class TableLineageInfo:
...
+                azure_client_id=self._azure_auth.client_id,
+                azure_client_secret=self._azure_auth.client_secret.get_secret_value(),
+                product="datahub",
+                product_version=nice_version_name(),
+            )
+        else:
+            self._workspace_client = WorkspaceClient(
+                host=self._workspace_url,
+                token=self._personal_access_token,
+                product="datahub",
```
kyungsoo-datahub, line 199: It would be nice to use ```DATABRICKS_USER_AGENT_ENTRY``` instead of "datahub" for consistency.

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
```diff
@@ -160,16 +160,93 @@ class TableLineageInfo:
...
+        self._azure_auth = azure_auth
+
+        # Initialize workspace client
+        if self._azure_auth:
+            self._workspace_client = WorkspaceClient(
+                host=self._workspace_url,
+                azure_tenant_id=self._azure_auth.tenant_id,
+                azure_client_id=self._azure_auth.client_id,
+                azure_client_secret=self._azure_auth.client_secret.get_secret_value(),
+                product="datahub",
```
kyungsoo-datahub, line 192: It would be nice to use ```DATABRICKS_USER_AGENT_ENTRY``` instead of "datahub" for consistency.

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
```diff
@@ -160,16 +160,93 @@ class TableLineageInfo:
...
+        The SDK's authenticate() method automatically handles token refresh
+        when needed, so we simply call it to get a current valid token.
+
+        Returns:
+            A valid authentication token
+        """
+        if self._azure_auth:
+            # The SDK automatically refreshes the token if it's expired
+            # through the Refreshable class mechanism
+            token = self._workspace_client.config.authenticate()
```
kyungsoo-datahub, line 220: What happens if .authenticate() returns None?

If authenticate() == None can break the logic, please, add the test for it as well after handling None case.

metadata-ingestion/tests/unit/test_unity_catalog_proxy.py
```diff
@@ -1872,3 +1874,282 @@ def test_get_query_history_via_system_tables_time_parameters(
...
+        assert mock_workspace_client.call_count == 1
+        call_kwargs = mock_workspace_client.call_args[1]
+        assert call_kwargs["host"] == "https://test.databricks.com"
+        assert call_kwargs["azure_tenant_id"] == "test_tenant"
+        assert call_kwargs["azure_client_id"] == "test_client"
+        assert call_kwargs["azure_client_secret"] == "test_secret"
+        assert call_kwargs["product"] == "datahub"
+
+    @patch("datahub.ingestion.source.unity.proxy.WorkspaceClient")
+    def test_get_workspace_client(self, mock_workspace_client):
```
kyungsoo-datahub: This one is trivial test. Maybe we can remove it.

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
Anshul759: Updated

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
Anshul759: Done

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
Anshul759: Done

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
Anshul759: Have added the code to handle that case, it would be logging a error if in case the token in None

metadata-ingestion/tests/unit/test_unity_catalog_proxy.py
Anshul759: Removed

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
Anshul759: Can you please describe the expected change here, i do not see a related code near the line tagged

kyungsoo-datahub: This is for:
```
            fresh_token = self._token_provider.get_token()
            if fresh_token != self._sql_connection_params["access_token"]:
                logger.debug("Updating SQL connection token")
                self._sql_connection_params["access_token"] = fresh_token
```

My suggestion is that we can assign the value in any case without comparing it.

```
self._sql_connection_params["access_token"] = self._token_provider.get_token()
```

Anshul759: Done

Anshul759:
@kyungsoo-datahub Thanks for the review. Have update the code with the suggestion!!

Anshul759:
Seems like https://github.com/datahub-project/datahub/pull/15824 already adds the refresh connection token and this PR is no longer required.

PR #15702, opened 2025-12-17
feat(mssql): auto-enable use_odbc when source type is mssql-odbc

alfiyas-datahub:
## Summary

When users configure the source type as `mssql-odbc`, they currently also need to manually set `use_odbc: true` in their recipe config. This is redundant and confusing - the source type name already implies ODBC should be used.

This PR automatically enables `use_odbc` when the source type is `mssql-odbc`, eliminating the need for redundant configuration.

## Changes

- Auto-detection in `SQLServerSource.create()`**: When source type is `mssql-odbc` and `use_odbc` is not explicitly set, automatically set it to `True`
- Updated recipe docs**: Removed unnecessary `use_odbc: "True"` from the ODBC example recipe
- Updated integration test**: Removed redundant `use_odbc: true` from test config
- Added unit tests**: Comprehensive tests covering:
- Auto-enabling for `mssql-odbc` type
- Handling when `pipeline_config` is None

## Test Plan

- Added unit tests for all scenarios
- Existing integration tests pass
- Manual verification with `mssql-odbc` source type

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/sql/mssql/source.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/sql/mssql/source.py
@@ -108,8 +108,6 @@ class SQLServerConfig(BasicSQLAlchemyConfig):
     include_descriptions: bool = Field(
         default=True, description="Include table descriptions information."
     )
-    # Internal field to track ODBC mode - set by create() based on source type
-    is_odbc_mode: HiddenFromDocs[bool] = Field(default=False)
     # Soft deprecation: use_odbc is removed, source type determines ODBC mode
     _use_odbc_removed = pydantic_removed_field("use_odbc")
     uri_args: Dict[str, str] = Field(
@@ -156,7 +154,8 @@ class SQLServerConfig(BasicSQLAlchemyConfig):
     def validate_uri_args(
         cls, v: Dict[str, Any], info: ValidationInfo, **kwargs: Any
     ) -> Dict[str, Any]:
-        is_odbc = info.data.get("is_odbc_mode", False)
+        # Use Pydantic validation context to get is_odbc flag
+        is_odbc = info.context.get("is_odbc", False) if info.context else False
         if is_odbc and not info.data["sqlalchemy_uri"] and "driver" not in v:
             raise ValueError(
                 "uri_args must contain a 'driver' option when using mssql-odbc source type"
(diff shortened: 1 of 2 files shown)


metadata-ingestion/src/datahub/ingestion/source/sql/mssql/source.py
```diff
@@ -331,6 +331,16 @@ def _populate_column_descriptions(self, conn: Connection, db_name: str) -> None:
...
     def create(cls, config_dict: Dict, ctx: PipelineContext) -> "SQLServerSource":
+        # Auto-enable use_odbc when source type is 'mssql-odbc'.
+        # The source type name 'mssql-odbc' already implies ODBC should be used,
+        # so we infer use_odbc=True to avoid requiring redundant configuration.
+        # This only applies when use_odbc is not explicitly set in the config.
+        source_type = getattr(
+            getattr(ctx.pipeline_config, "source", None), "type", None
+        )
+        if source_type == "mssql-odbc" and "use_odbc" not in config_dict:
+            config_dict = {**config_dict, "use_odbc": True}
```
kyungsoo-datahub: What happens if someone enabled mssql-odbc but explicitly set use_odbc false. If mssql odbc doesn't work at all with use_odbc false, let's warn that case and enable use_odbc as well.

metadata-ingestion/src/datahub/ingestion/source/sql/mssql/source.py
```diff
@@ -236,16 +239,23 @@ class SQLServerSource(SQLAlchemySource):
...
     - Column types associated with each table/view
     - Table, row, and column statistics via optional SQL profiling
-    We have two options for the underlying library used to connect to SQL Server: (1) [python-tds](https://github.com/denisenkom/pytds) and (2) [pyodbc](https://github.com/mkleehammer/pyodbc). The TDS library is pure Python and hence easier to install.
-    If you do use pyodbc, make sure to change the source type from `mssql` to `mssql-odbc` so that we pull in the right set of dependencies. This will be needed in most cases where encryption is required, such as managed SQL Server services in Azure.
+
+    Two source types are available:
+    - `mssql`: Uses [python-tds](https://github.com/denisenkom/pytds) library (pure Python, easier to install)
+    - `mssql-odbc`: Uses [pyodbc](https://github.com/mkleehammer/pyodbc) library (required for encryption, Azure managed services)
+
+    If you need encryption (e.g., for Azure SQL), use source type `mssql-odbc` and configure `uri_args` with your ODBC driver settings.
```
sgomezvillamor, line 247: Let's double check that this info is clearly documented in the user-facing docs https://docs.datahub.com/docs/generated/ingestion/sources/mssql

alfiyas-datahub: Verified ✅ - the docs clearly explain both source types (mssql vs mssql-odbc) and when to use each for encryption/Azure.

PR #15703, opened 2025-12-17
fix(sql-parsing): Restore # prefix for MSSQL temporary tables in lineage extraction

dinesh-verma-datahub:
## Summary

Fixes incorrect lineage extraction for MSSQL temporary tables. When parsing MSSQL queries containing temporary tables (e.g., `#mytemptable`), SQLGlot strips the `#` or `##` prefix but sets internal flags (`temporary` or `global_`) on the identifier. This caused downstream temp table detection to fail, incorrectly including temporary tables in lineage when their names matched actual tables.

Linear: https://linear.app/acryl-data/issue/ING-1053/mssql-lineage-incorrectly-includes-local-temporary-tables-eg

## Problem

When running MSSQL lineage extraction with queries like:
SELECT key INTO #mytemptable FROM myrawtableThe temporary table `#mytemptable` was being treated as a regular table because:
1. SQLGlot parses `#mytemptable` and strips the `#` prefix, storing just `mytemptable`
2. SQLGlot sets `temporary=True` flag on the identifier (not visible in `.name`)
3. DataHub's `_TableName.from_sqlglot_table()` used only `.name`, losing the temp indicator
4. Downstream `is_temp_table()` checks for `startswith("#")` which now fails
5. If a real table named `mytemptable` exists, it incorrectly appears in lineage

## Solution

Modified `_TableName.from_sqlglot_table()` in `metadata-ingestion/src/datahub/sql_parsing/_models.py` to:
1. Check for `temporary` flag on the identifier → restore `#` prefix (local temp)
2. Check for `global_` flag on the identifier → restore `##` prefix (global temp)
3. Handle both simple table names and multi-part names (e.g., `db.schema.#temp`)
4. Avoid double-prefixing if the prefix is already present

## Testing

Added comprehensive unit tests in `metadata-ingestion/tests/unit/sql_parsing/test_table_name.py`:
- MSSQL local temp tables (`#temp`) in SELECT INTO and CREATE TABLE
- MSSQL global temp tables (`##temp`)
- Multi-part temp table names
- Other dialects (PostgreSQL, Snowflake, BigQuery, Hive, Spark) to ensure no unintended prefixing
- Double-prefix prevention
- Integration with `is_temp_table()` helper

### DIFF
--- a/datahub-web-react/src/alchemy-components/components/Table/Table.tsx
+++ b/datahub-web-react/src/alchemy-components/components/Table/Table.tsx
@@ -259,6 +259,10 @@ export const Table = <T,>({
                                                 ? column.cellWrapper(content, row)
                                                 : content;
 
+                                            const clickable = column.isCellClickable
+                                                ? column.isCellClickable(row)
+                                                : !!column.onCellClick;
+
                                             return (
                                                 <TableCell
                                                     key={column.key}
@@ -271,11 +275,11 @@ export const Table = <T,>({
                                                     isGroupHeader={canExpand}
                                                     isExpanded={isExpanded}
                                                     onClick={() => {
-                                                        if (column.onCellClick) {
+                                                        if (clickable && column.onCellClick) {
                                                             column.onCellClick(row);
                                                         }
                                                     }}
-                                                    style={{ cursor: column.onCellClick ? 'pointer' : 'default' }}
+                                                    style={{ cursor: clickable ? 'pointer' : 'default' }}
                                                     className={column.cellWrapper ? 'hoverable-cell' : undefined}
                                                     data-testid={column.key}
                                                 >
--- a/datahub-web-react/src/alchemy-components/components/Table/types.ts
+++ b/datahub-web-react/src/alchemy-components/components/Table/types.ts
@@ -14,6 +14,7 @@ export interface Column<T> {
     alignment?: AlignmentOptions;
     tooltipTitle?: string;
     onCellClick?: (record: T) => void;
+    isCellClickable?: (record: T) => boolean;
     cellWrapper?: (content: React.ReactNode, record: T) => React.ReactNode;
 }
 
--- a/datahub-web-react/src/app/ingestV2/shared/components/columns/StatusColumn.tsx
+++ b/datahub-web-react/src/app/ingestV2/shared/components/columns/StatusColumn.tsx
@@ -1,10 +1,13 @@
 import { LoadingOutlined } from '@ant-design/icons';
 import { Icon, Pill } from '@components';
-import { Button } from 'antd';
 import React from 'react';
 import styled from 'styled-components';
 
-import { EXECUTION_REQUEST_STATUS_LOADING, EXECUTION_REQUEST_STATUS_RUNNING } from '@app/ingestV2/executions/constants';
+import {
+    EXECUTION_REQUEST_STATUS_LOADING,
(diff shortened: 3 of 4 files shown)


metadata-ingestion/src/datahub/sql_parsing/_models.py
```diff
@@ -5,6 +5,48 @@
 from pydantic import BaseModel
 
 
+def _restore_temp_table_prefix(
+    table_name: str, identifier: sqlglot.exp.Expression
+) -> str:
+    """Restore MSSQL temp table prefix (# or ##) based on SQLGlot identifier flags.
```
treff7es: if this is for mssql only why don't we check the dialect before?
Is this something should be fixed on sqlglot or not?

dinesh-verma-datahub: > Hi @treff7es, great questions! After a deep dive into the codebase, here's my analysis:
>
> ### 1. Why don't we check the dialect before?
>
> Current approach: The temporary and global flags are only set by SQLGlot when parsing MSSQL dialect, so the fix is effectively a no-op for other dialects. However, I agree an explicit dialect check would be cleaner.
>
> ### 2. Should this be fixed elsewhere in DataHub's SQLGlot-related code?
>
> After reviewing the architecture, I found a better location: sqlglot_lineage.py in _extract_table_names().
>
> Why this is better:
> - _extract_table_names is called from _table_level_lineage, which already has dialect parameter
> - It's the central location for table extraction in lineage
> - Follows the existing pattern in query_types.py which has _is_temp_table(table, dialect)
> - Explicit dialect check (only apply fix for MSSQL)
>
> Proposed change:
>
> def _extract_table_names(>     iterable: Iterable[sqlglot.exp.Table],>     dialect: sqlglot.Dialect,  # ← Add dialect parameter> ) -> OrderedSet[_TableName]:>     # Restore MSSQL temp table prefix if needed>     ...>
>
> This would require updating the 4 call sites to pass dialect, but it's a cleaner architectural fit.
>
> ---
>
> Would you prefer:
> 1. Option A: Keep current approach in _models.py (implicit dialect check via flags) — simpler, already tested
> 2. Option B: Refactor to sqlglot_lineage.py with explicit dialect check — cleaner architecture, addresses your concern
>


> Let me know your preference and I'll implement or any other suggestions!

treff7es: I don't really get the two option because the comment only talks about one better approach which seems to be the prerred way.

dinesh-verma-datahub: @treff7es I have made the changes, please check.

metadata-ingestion/tests/unit/sql_parsing/test_table_name.py
```diff
@@ -0,0 +1,803 @@
...
+
+        tables = list(parsed.find_all(sqlglot.exp.Table))
+        table_names = [
+            _table_name_from_sqlglot_table(t, self._get_mssql_dialect()).table
+            for t in tables
+        ]
+
+        assert "#mytemptable" in table_names
+
+    def test_mssql_local_temp_create_table(self):
```
treff7es, line 233: A parametrized test would be much more readable

dinesh-verma-datahub: sure, i will keep it in mind for future tests.

PR #15707, opened 2025-12-17
feat(frontend): Allow external scripts in custom themes

DiogoVala:
Functionality allows external scripts to be specified in themes and loaded at runtime.
Useful to embed functionality without modifying datahub code. In our case we are using it for a feedback widget which sits on top of Datahub and allows users to suggest improvements or point out issues.


DiogoVala:
> Hey @DiogoVala - given the potential security risks of loading arbitrary script URLs, we cannot accept this change as is. Can you consider a more secure approach to achieve the same outcome?

Thank you for the feedback @gabe-lyons 
This feature requires the Frontend to be rebuilt so that the script URL takes effect. This means the only risk is if the script somehow changes.
This feature is meant to offer more flexibility to the team managing datahub and it should only be used to load scripts owned by the team, but if you see this being too high of a security risk, would you accept having an integrity SHA to compare the script to?

PR #15711, opened 2025-12-17
feat(ui): update design of reset credentials page

purnimagarg1:
**Linear ticket:**
https://linear.app/acryl-data/issue/CAT-1050/update-design-for-the-resetcredentials-component-to-match-loginsignup


chriscollins3456 (approved):
in a followup we will want to handle submitting the form on enter as well. i have another PR adding that for sign up and login that should be up later today on OSS

PR #15714, opened 2025-12-17
feat(kafka): automatically upsize topic partition counts

rob-1019:
Extends CreateKafkaTopicsStep to detect and automatically increase partition counts for existing topics that have fewer partitions than configured.

Previously, the upgrade step only created new topics and skipped existing ones. Now it checks existing topics and increases their partition count if they're under-provisioned. The step fails if partition count checks fail, ensuring operators are alerted to configuration issues. Note that Kafka does not support reducing partition counts, so topics with more partitions than configured are left unchanged with a warning.

Changes:
- Add getCurrentPartitionCount() method to check existing partition counts
- Track partitionsToIncrease map for topics that need upsizing
- Track failedTopics list and throw exception if any checks fail
- Add comprehensive test coverage for upsizing and failure scenarios
- Update kafka-config.md to document automatic partition upsizing behavior

- [X] For any breaking change/potential downtime/deprecation/big changes an entry has been made in [Updating DataHub](https://github.com/datahub-project/datahub/blob/master/docs/how/updating-datahub.md)


datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/system/kafka/CreateKafkaTopicsStep.java
```diff
@@ -120,8 +128,38 @@ public Function<UpgradeContext, UpgradeStepResult> executable() {
...
+              } else if (currentPartitions > desiredPartitions) {
+                log.warn(
+                    "Topic {} has {} partitions but configuration specifies {}. "
+                        + "Kafka does not support reducing partition count - leaving unchanged",
+                    topicName,
+                    currentPartitions,
+                    desiredPartitions);
+                topicsToSkip.add(topicName);
+              } else {
+                log.info(
```
deepgarg760: This logging is not required as this a default behavior before this change also

datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/system/kafka/CreateKafkaTopicsStep.java
```diff
@@ -120,8 +128,38 @@ public Function<UpgradeContext, UpgradeStepResult> executable() {
...
+                    desiredPartitions);
+                partitionsToIncrease.put(topicName, NewPartitions.increaseTo(desiredPartitions));
+              } else if (currentPartitions > desiredPartitions) {
+                log.warn(
+                    "Topic {} has {} partitions but configuration specifies {}. "
+                        + "Kafka does not support reducing partition count - leaving unchanged",
+                    topicName,
+                    currentPartitions,
+                    desiredPartitions);
+                topicsToSkip.add(topicName);
```
deepgarg760: `topicsToSkip.add(topicName);` should be outside of the if/else condition as it was before

datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/system/kafka/CreateKafkaTopicsStep.java
```diff
@@ -147,20 +185,48 @@ public Function<UpgradeContext, UpgradeStepResult> executable() {
...
-        result.all().get(); // Wait for all topics to be created
+        // Check if any topics failed to be checked for partition count
+        if (!failedTopics.isEmpty()) {
+          String errorMessage =
+              String.format(
+                  "Failed to check partition count for %d topics: %s. "
+                      + "These topics may not have the correct partition configuration.",
+                  failedTopics.size(), failedTopics);
+          log.error(errorMessage);
+          throw new RuntimeException(errorMessage);
```
deepgarg760, line 225: IMO, we should not throw exception here and log the warning is sufficient. It should not be blocker for topic creation

datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/system/kafka/CreateKafkaTopicsStep.java
```diff
@@ -120,8 +128,38 @@ public Function<UpgradeContext, UpgradeStepResult> executable() {
 
           // Check if topic already exists
           if (existingTopics.contains(topicName)) {
-            log.info("Topic {} already exists - skipping creation", topicName);
-            topicsToSkip.add(topicName);
+            // For existing topics, check if partition count needs to be increased
```
deepgarg760: IMO, this new block can have its own separate function for code readability

deepgarg760:
IMO, this should be controlled by feature flag like: DATAHUB_AUTO_INCREASE_PARTITIONS

datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/system/kafka/CreateKafkaTopicsStep.java
```diff
@@ -120,8 +128,38 @@ public Function<UpgradeContext, UpgradeStepResult> executable() {
...
+            try {
+              int currentPartitions = getCurrentPartitionCount(adminClient, topicName);
```
deepgarg760: Instead of doing single API call for each existing, can this be batch. The goal is to have single API call for all existing topics

datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/system/kafka/CreateKafkaTopicsStep.java
rob-1019: Cleaned this up a bit. PTAL.

datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/system/kafka/CreateKafkaTopicsStep.java
rob-1019: done

datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/system/kafka/CreateKafkaTopicsStep.java
rob-1019: Feel pretty strongly that failure to configure these items during a setup job should be fatal if the user has this functionality enabled. Same for topic creation and most other things during setup jobs do - its better to fail loudly. Failures may get auto-triaged by restart attempts, quiet ignores can't be.

datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/system/kafka/CreateKafkaTopicsStep.java
rob-1019: Take a look and let me know if that doesn't address core concern here. Made the log messages a bit more consistant to ease grepping for the various categories of outcome.

datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/system/kafka/CreateKafkaTopicsStep.java
rob-1019: Done.

rob-1019:
> IMO, this should be controlled by feature flag like: DATAHUB_AUTO_INCREASE_PARTITIONS

Excellent suggestion. Done. Overloading that was naughty.

datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/system/kafka/CreateKafkaTopicsStep.java
deepgarg760, line 35: It would be great if you also convert these variables in Java format since prefix the variables with underscore should not there

PR #15721, opened 2025-12-18
feat(ingestion): add telemetry events for the new ingestion flow

purnimagarg1:
**Linear ticket:**
https://linear.app/acryl-data/issue/CAT-1037/add-telemetry-events-related-to-new-ingestion-flows

### DIFF
--- a/datahub-web-react/src/app/homeV2/layout/navBarRedesign/NavSidebar.tsx
+++ b/datahub-web-react/src/app/homeV2/layout/navBarRedesign/NavSidebar.tsx
@@ -309,10 +309,12 @@ export const NavSidebar = () => {
                         selectedIcon: <Plugs weight="fill" />,
                         link: ingestionLink,
                         onClick: () => {
-                            analytics.event({
-                                type: EventType.EnterIngestionFlowEvent,
-                                entryPoint: 'nav_menu',
-                            });
+                            if (ingestionLink === PageRoutes.INGESTION_CREATE) {
+                                analytics.event({
+                                    type: EventType.EnterIngestionFlowEvent,
+                                    entryPoint: 'nav_menu',
+                                });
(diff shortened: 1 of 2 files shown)


datahub-web-react/src/app/homeV2/layout/navBarRedesign/NavSidebar.tsx
```diff
@@ -308,6 +308,12 @@ export const NavSidebar = () => {
                         icon: <Plugs />,
                         selectedIcon: <Plugs weight="fill" />,
                         link: ingestionLink,
+                        onClick: () => {
+                            analytics.event({
+                                type: EventType.EnterIngestionFlowEvent,
+                                entryPoint: 'nav_menu',
+                            });
+                        },
```
chriscollins3456, line 318: hmm i think this shouldn't always call this event - it's only if we're sending them to create a new source i think? so it should be probably dependent on `ingestionLink` if we're going to the create flow then we hit this event otherwise don't do a new event

datahub-web-react/src/app/ingestV2/source/builder/RecipeForm/TestConnection/TestConnectionButton.tsx
```diff
@@ -77,6 +77,7 @@ function TestConnectionButton({
     const [isModalVisible, setIsModalVisible] = useState(false);
     const [pollingInterval, setPollingInterval] = useState<null | NodeJS.Timeout>(null);
     const [testConnectionResult, setTestConnectionResult] = useState<null | TestConnectionResult>(null);
+    const [duration, setDuration] = useState<number>(0);
```
chriscollins3456: should this be undefined to start and we only set it after we know the duration?

PR #15722, opened 2025-12-18
fix(ui): update button styles for login/signup screens

purnimagarg1:
**Linear ticket:**
https://linear.app/acryl-data/issue/CAT-1061/update-button-styling-on-the-new-loginsign-up-pages

**Description:**

When appConfig and flags are not available before login, themeV1 is considered by default and its primary color gets used in gradient of button styles.
This fixes using only v2 primary color in such instances in button component.
This is a temp fix for button component. We can have a better solution app-wide to have theme V2 by default instead of V1 now.

**Screenshots:**

image

image

### DIFF
--- a/datahub-web-react/src/alchemy-components/components/Button/utils.ts
+++ b/datahub-web-react/src/alchemy-components/components/Button/utils.ts
@@ -132,6 +132,8 @@ const getButtonVariantStyles = (
     theme?: Theme,
 ): CSSObject => {
     const isPrimary = color === 'violet' || color === 'primary';
+    // Adding a hack here for login/signup pages where v1 styles are still loaded by default
+    // Once we move to remove v1 styles we can revert this hack and always use styles from theme only
     const resolvedPrimaryColor =
         theme?.styles?.['primary-color'] === themeV1PrimaryColor
             ? themeV2PrimaryColor


datahub-web-react/src/alchemy-components/components/Button/utils.ts
```diff
@@ -129,7 +132,12 @@ const getButtonVariantStyles = (
     theme?: Theme,
 ): CSSObject => {
     const isPrimary = color === 'violet' || color === 'primary';
-    const primaryGradient = `radial-gradient(115.48% 144.44% at 50% -44.44%, ${theme?.styles?.['primary-color-gradient'] || '#705EE4'} 38.97%, ${theme?.styles?.['primary-color'] || '#533FD1'} 100%)`;
+    const resolvedPrimaryColor =
+        theme?.styles?.['primary-color'] === themeV1PrimaryColor
+            ? themeV2PrimaryColor
+            : (theme?.styles?.['primary-color'] ?? themeV2PrimaryColor);
+
+    const primaryGradient = `radial-gradient(115.48% 144.44% at 50% -44.44%, ${theme?.styles?.['primary-color-gradient'] || '#705EE4'} 38.97%, ${resolvedPrimaryColor} 100%)`;
```
chriscollins3456, line 142: yeah it's hacky but it works

like you said we should just have v2 styling by default soon and we can remove this.

i'm going to add a comment above this with an explanation as to why we're doing this

chriscollins3456 (approved):

PR #15725, opened 2025-12-18
chore(ingestion): validate connector deps

sgomezvillamor:
## Summary

Adds a new workflow job to validate that each ingestion plugin can be installed independently with all its dependencies. This helps catch common mistakes when adding new sources or updating dependencies.

This brings the validation logic from the `connector-tests` repo directly into the OSS repo, making it easier to catch dependency issues earlier in the development cycle.

## Key Features

- **Single Job Validation**: Tests all 86+ plugins sequentially in a single job for efficient resource usage
- **Isolated Testing**: Each plugin is installed in its own virtual environment to ensure clean dependency resolution
- **Smart Triggering**: Only runs when `metadata-ingestion/setup.py` is modified
- **Dynamic Plugin Discovery**: Uses `datahub check plugins` to automatically discover all source plugins
- **System Dependencies**: Installs required system libraries (Kerberos, LDAP, etc.) via `install_deps.sh`
- **Comprehensive Summary**: Provides a detailed report of successful and failed plugin installations

## Workflow Structure

**validate-plugin-deps** job:
- Checks if `setup.py` was modified (skips all steps if not)
- Installs system dependencies using `metadata-ingestion/scripts/install_deps.sh`
- Installs the package using `installDev` to ensure CLI entry points are available
- Discovers all plugins using `datahub check plugins`
- For each plugin:
  - Creates a fresh virtual environment
  - Installs the plugin with all its dependencies using `uv pip install -e .[plugin-name]`
  - Cleans up the venv after validation
- Provides a summary report at the end

## Example Output

```
✅ Found 86 plugins to validate:
  - abs
  - athena
  - bigquery
  ...

Validating 86 plugins...

::group::Validating plugin: bigquery
✅ Plugin 'bigquery' installed successfully with all dependencies
::endgroup::

==========================================
Validation Summary:
  Total plugins: 86
  ✅ Successful: 86
  ❌ Failed: 0
==========================================
✅ All plugins validated successfully!
```

## Implementation Details

- Uses `installDev` instead of `installPackageOnly` to ensure CLI entry points are installed
- Uses `venv/bin/python -m datahub check plugins` to get the plugin list
- Creates isolated venvs for each plugin to prevent dependency conflicts
- Uses system `uv` command with `--python` flag to install into specific venvs
- Installs system dependencies (libkrb5-dev, libldap2-dev, etc.) required for plugins like nifi, ldap, and hive-metastore

## Testing

This PR can be tested by:
1. Manually triggering the workflow via workflow_dispatch
2. Making a test change to `metadata-ingestion/setup.py` to trigger the validation

## Related

- Similar validation logic exists in acryldata/connector-tests (but uses pre-built wheels)
- Helps prevent issues like missing dependencies in plugin definitions
- Catches build-time errors for plugins requiring system libraries


metadata-ingestion/setup.py
```diff
@@ -13,6 +13,8 @@
     else ""
 )
 
+# random update to trigger connector dependency validation
+
```
skrydal: can we remove that line?

skrydal (approved):
LGTM beside the one comment line.

metadata-ingestion/setup.py
sgomezvillamor: sure, the plan is to remove it once approved and before merging

PR #15729, opened 2025-12-19
feat(ingestion): adding ingestion security guidelines

gabe-lyons:
Adding documentation to help guide new users about how to choose their ingestion approach


jamesmayfield-star:
PR looks good! Language is concise and clear. Only thing I might consider adding at the bottom of the doc is a link to our ingestion architecture because folks may be interested in learning more as a logical next step: https://docs.datahub.com/docs/architecture/metadata-ingestion

PR #15730, opened 2025-12-19
fix(sql-parsing): split MSSQL statements when previous expression ends with parentheses

alfiyas-datahub:
## Summary

- Fixed SQL statement splitter incorrectly treating closing parentheses from function calls (e.g., `GETDATE()`) as CTE indicators.
- The `_has_preceding_cte()` method now checks that the statement actually starts with `WITH` before assuming a CTE context.
- Added regression test for this scenario.


metadata-ingestion/tests/unit/sql_parsing/test_split_statements.py
```diff
@@ -244,3 +244,43 @@ def test_split_statement_with_end_keyword_in_bracketed_identifier_with_escapes()
...
+    test_sql = """\
+    WITH temp_cte AS (
+        SELECT * FROM source WHERE date > GETDATE()
+    )
+    SELECT * FROM temp_cte"""
+
+    statements = [s.strip() for s in split_statements(test_sql)]
+
+    assert len(statements) == 1
+    assert statements[0] == test_sql
```
kyungsoo-datahub: ```test_sql.strip()``` would fix the test failure.

metadata-ingestion/src/datahub/sql_parsing/split_statements.py
```diff
@@ -267,8 +267,13 @@ def _process_normal(self, most_recent_real_char: str) -> Iterator[str]:
             self.current_statement.append(c)
 
     def _has_preceding_cte(self, most_recent_real_char: str) -> bool:
```
kyungsoo-datahub, line 269: Current version has

``` 
        # If current statement starts with INSERT/UPDATE/DELETE, the closing paren
        # is likely from the DML statement itself (e.g., WHERE clause), not a CTE
        current_text_stripped = current_text.strip()
        if current_text_stripped.startswith(("INSERT", "UPDATE", "DELETE")):
            return False
```
 
This PR is more accurate. Please, use this PR's change.

PR #15733, opened 2025-12-19
fix(ingestion/schema-resolver): add enabler for sql parsing for more than 3 parts and implement in dremio

acrylJonny:


metadata-ingestion/src/datahub/sql_parsing/_models.py
```diff
@@ -37,6 +37,21 @@ class _TableName(_FrozenModel):
     database: Optional[str] = None
     db_schema: Optional[str] = None
     table: str
+    parts: Optional[tuple] = None
```
sgomezvillamor: `databse`, `db_schema` and `table` are self-explained

⚠️ however this new `parts` field would require some docs about the purpose

💡 additionally, we transform some comments into actual code to emphasize the _identity_ constraint

eg

```
@property
def identity(self) -> tuple:
    return (self.database, self.db_schema, self.table) # parts excluded on purpose
```

metadata-ingestion/src/datahub/sql_parsing/_models.py
```diff
@@ -81,8 +95,18 @@ def from_sqlglot_table(
             table_name = ".".join(parts)
         else:
             table_name = table.this.name
+
+        # Store parts as string tuple for connector-specific handling
+        parts_tuple = None
+        if hasattr(table, "parts") and table.parts:
```
sgomezvillamor: why the `hasattr`? in which cases `sqlglot.exp.Table` have/miss this `parts` field?

similar with the `gettattr` below, quite confusing

maybe AI agent being excessively/unnecessarily conservative?

metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_source.py
```diff
@@ -74,39 +74,40 @@
 
 logger = logging.getLogger(__name__)
 
+# Dremio uses 'dremio' as the default database name in all SQL contexts
```
sgomezvillamor, line 77: where this constraint comes from? us, sqlglot, dremio itself?

metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_source.py
```diff
@@ -74,39 +74,40 @@
...
-            table_name_parts = [
-                "dremio",
-                table.database,
-                table.db_schema,
-                table.table,
-            ]
+        if table.parts and len(table.parts) > 3:
+            # Use parts for full hierarchy in multi-part tables
+            table_name_parts = [DREMIO_DATABASE_NAME] + list(table.parts)
+            table_name = ".".join(filter(None, table_name_parts))
```
sgomezvillamor: would it make sense to have a method in `_TableName` providing this or part of this logic?

```
# _TableName
@property
def name_with_parts(self): # I'm sure you can find a better name
    return ".".join(filter(None, list(self.parts)))

# here
table_name = DREMIO_DATABASE_NAME + "." + table.name_with_parts
```


And I also think part of this could also be pushed to the _TableName

```
                table_name_parts = [
                    DREMIO_DATABASE_NAME,
                    table.database,
                    table.db_schema,
                    table.table,
                ]
            else:
                table_name_parts = [
                    table.database or DREMIO_DATABASE_NAME,
                    table.db_schema,
                    table.table,
                ]
```

metadata-ingestion/src/datahub/sql_parsing/_models.py
```diff
@@ -37,6 +37,21 @@ class _TableName(_FrozenModel):
...
+
+    def __eq__(self, other: object) -> bool:
+        # Parts field is excluded - table identity is database.schema.table only
+        if not isinstance(other, _TableName):
+            return False
+        return (
+            self.database == other.database
+            and self.db_schema == other.db_schema
+            and self.table == other.table
+        )
```
sgomezvillamor: Bart also found this

>  *BLOCKERS - Must Fix Before Merge:**
> 1. qualified() method loses the parts field (line 74-78 in _models.py)
>    - When creating a new _TableName instance with defaults applied, the parts field isn't preserved
>    - This breaks multi-part table handling - the entire feature falls apart here
>    - Fix: Add parts=self.parts when returning the new instance

metadata-ingestion/src/datahub/sql_parsing/_models.py
```diff
@@ -37,6 +37,20 @@ class _TableName(_FrozenModel):
     database: Optional[str] = None
     db_schema: Optional[str] = None
     table: str
+    parts: Optional[Tuple[str, ...]] = None
```
sgomezvillamor, line 40: please, add docs around this field

PR #15753, opened 2025-12-22
feat(validation): add aspect size validation

rob-1019:

## Summary

Implements serialized aspect size validation to protect against large aspects from being created or consumed by mcp-consumer. This provides validation at two checkpoints with a choice of remediation strategies and protects against aspects that are either already too large or would become too large through the application of a patch. Additionally, adds extra logging at the INFO level to document the start of each mcp message as a companion to the per-message event completion we already log at the INFO level.

All options (except the additional logging) added by this PR are disabled by default and must be explicitly enabled.

This introduces a pattern (AspectSerializationHook) to front-load JSON serialization for the database upsert that mce-consumer does when upserting aspects so that it can be size-checked ahead of time and then the serialized value re-used during the database commit. This drives the bulk of the code-delta and is present to optimize a hot-path operation as to avoid duplicating one of the most expensive portions of mcp-processing - the serialization of the resulting aspect which otherwise would happen twice when the post-patch size aspect size validation is enabled - once when we checked for size prior to inserting it to establish an abort path and again when we inserted it into the database.

### Validation Checkpoints

1. **Pre-patch validation** (`datahub.validation.aspectSize.prePatch`)
   - Validates existing aspects from database BEFORE patch application
   - Catches oversized aspects already in the database
   - Remediation: `REPLACE_WITH_PATCH` (delete old, continue) or `IGNORE` (skip write)

2. **Post-patch validation** (`datahub.validation.aspectSize.postPatch`)
   - Validates aspects AFTER patch/merge, before database write
   - Catches aspects that become oversized after applying patches
   - Remediation: `DELETE` (remove from DB) or `IGNORE` (skip write)

### Remediation Strategies (Default: IGNORE)

- **`IGNORE`** (default for both checkpoints): Acknowledges MCP but skips write, logs WARNING without deleting.
- **`REPLACE_WITH_PATCH`** (pre-patch only): Deletes oversized existing aspect, continues with patch as insert (no merge). If new patch is also oversized, post-patch validation will catch it.
- **`DELETE`** (post-patch only): Hard deletes oversized aspect from database, logs WARNING, skips write

## Logging

This PR adds structured WARNING logs when oversized aspects are detected:

```
Oversized pre-patch aspect <remediation description>: urn=urn:li:dataset:..., aspect=datasetProfile, size=20000000 serialized bytes, threshold=15728640 serialized bytes
```

This PR additionally adds structured INFO logs when MCP events are started - mirroring the completion events for MCP we already log at the info level.

## Configuration

**Disabled by default - zero impact on existing deployments.** These are **debugging flags** for identifying oversized aspects in production:

```yaml
datahub:
  validation:
    aspectSize:
      prePatch:
        enabled: false  # Enable when troubleshooting crashes or investigating oversized aspects
        maxSizeBytes: 15728640 
        oversizedRemediation: IGNORE  #  or REPLACE_WITH_PATCH
      postPatch:
        enabled: false  # Enable when troubleshooting crashes or investigating oversized aspects
        maxSizeBytes: 15728640
        oversizedRemediation: IGNORE  # or DELETE
```

Environment variables:
- `DATAHUB_VALIDATION_ASPECT_SIZE_PRE_PATCH_ENABLED`
- `DATAHUB_VALIDATION_ASPECT_SIZE_PRE_PATCH_MAX_SIZE_BYTES`
- `DATAHUB_VALIDATION_ASPECT_SIZE_PRE_PATCH_OVERSIZED_REMEDIATION`
- `DATAHUB_VALIDATION_ASPECT_SIZE_POST_PATCH_ENABLED`
- `DATAHUB_VALIDATION_ASPECT_SIZE_POST_PATCH_MAX_SIZE_BYTES`
- `DATAHUB_VALIDATION_ASPECT_SIZE_POST_PATCH_OVERSIZED_REMEDIATION`

See [docs/deploy/environment-vars.md](https://github.com/datahub-project/datahub/blob/master/docs/deploy/environment-vars.md#aspect-size-validation) for full configuration reference.

### Signature Change: Why It Was Necessary

To support the hook pattern without duplicate serialization, we extended the `EbeanSystemAspect` constructor to accept serialization hooks:

```java
// Before: 8 parameters
public EbeanSystemAspect(ebeanAspectV2, urn, aspectName, entitySpec, aspectSpec,
                         recordTemplate, systemMetadata, auditStamp)

// After: 11 parameters
public EbeanSystemAspect(ebeanAspectV2, urn, aspectName, entitySpec, aspectSpec,
                         recordTemplate, systemMetadata, auditStamp,
                         serializationHooks, prePatchValidationConfig, aspectDao)
```

We pass both `SystemAspect` (context) and `EntityAspect` (serialized form) to hooks:
- `SystemAspect` provides context: URN, aspect spec, RecordTemplate object
- `EntityAspect` provides the JSON string that was ALREADY serialized for DB write
- This is NOT duplicate serialization - hooks reuse the JSON created for the DB write


docs/advanced/mcp-mcl.md
```diff
@@ -225,6 +225,26 @@ The writes to the elasticsearch are asynchronous by default. A writer can add a
...
+
+```yaml
+datahub:
+  validation:
+    aspectSize:
+      prePatch:
+        enabled: false # Validates existing aspects from DB before patch application
+        maxSizeBytes: 15728640
+        oversizedRemediation: IGNORE # IGNORE (skip write) or REPLACE_WITH_PATCH (delete old, continue as insert)
+      postPatch:
```
david-leifker, line 243: We should be consistent with our max bytes. See https://github.com/datahub-project/datahub/blob/master/metadata-service/configuration/src/main/resources/application.yaml#L98

entity-registry/src/main/java/com/linkedin/metadata/aspect/AspectSerializationHook.java
```diff
@@ -0,0 +1,49 @@
...
+ * required step, not optional. The JSON string in {@code serializedAspect} is this already-created
+ * serialization, so hooks can inspect it at zero additional cost. Since JSON serialization is one
+ * of the most expensive operations in MCP processing, performing a second serialization just for
+ * validation would effectively double this cost. The hook pattern makes validation essentially free
+ * by reusing the required serialization.
+ *
+ * <p><b>Implementation Note:</b> Hooks are registered as Spring beans and auto-injected via
+ * {@code @Autowired List<AspectSerializationHook>} in {@code EntityAspectDaoFactory}.
+ */
+public interface AspectSerializationHook {
```
david-leifker: In our code base "hook" typically refers to MCL hooks or other classes involved with ingestion consumers. Because of this prior pattern, I'd suggest a different name such as another Validator or similar naming.

entity-registry/src/main/java/com/linkedin/metadata/aspect/EntityAspect.java
```diff
@@ -271,18 +275,36 @@ public SystemAspect setDatabaseAspect(@Nonnull SystemAspect databaseAspect) {
...
+              Optional.ofNullable(auditStamp)
+                  .map(AuditStamp::getImpersonator)
+                  .map(Urn::toString)
+                  .orElse(null));
+
+      // Call serialization hooks after aspect is serialized to JSON
+      // We pass both SystemAspect (this) and EntityAspect (serialized form) to hooks:
+      // - SystemAspect provides context: URN, aspect spec, RecordTemplate object
+      // - EntityAspect provides the JSON string that was ALREADY serialized for DB write (line 283)
+      // This is NOT duplicate serialization - hooks reuse the JSON created for the DB write,
```
david-leifker: Nice!

metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectDao.java
```diff
@@ -102,6 +106,14 @@ public EbeanAspectDao(
     this.metricUtils = metricUtils;
   }
 
+  public void setSerializationHooks(@Nonnull List<AspectSerializationHook> hooks) {
```
david-leifker: Is there a reason where we would set the hooks more then once? Why the setter pattern here instead of passing in the list via the constructor?

metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanSystemAspect.java
```diff
@@ -129,6 +164,64 @@ public EbeanSystemAspect forUpdate(
       this.ebeanAspectV2 = ebeanAspectV2;
       this.urn = UrnUtils.getUrn(ebeanAspectV2.getUrn());
       this.aspectName = ebeanAspectV2.getAspect();
+
+      // Pre-patch validation: check existing aspect size from DB
+      if (prePatchValidationConfig != null
```
david-leifker: Where is the similar logic for cassandra?

metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanSystemAspect.java
```diff
@@ -129,6 +164,64 @@ public EbeanSystemAspect forUpdate(
...
+          && prePatchValidationConfig.getPrePatch() != null
+          && prePatchValidationConfig.getPrePatch().isEnabled()) {
+        String rawMetadata = ebeanAspectV2.getMetadata();
+        if (rawMetadata != null) {
+          long actualSize = rawMetadata.length();
+          long threshold = prePatchValidationConfig.getPrePatch().getMaxSizeBytes();
+
+          if (actualSize > threshold) {
```
david-leifker: I think you'll want to pull this logic into a util class for re-use by cassandra code path.

metadata-service/configuration/src/main/java/com/linkedin/metadata/config/AspectSizeValidationConfig.java
```diff
@@ -0,0 +1,40 @@
...
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+public class AspectSizeValidationConfig {
+  /**
+   * Validates existing aspect in DB before patch application (measures: raw JSON string character
+   * count from database). Use to catch pre-existing oversized aspects. IGNORE remediation (default)
+   * logs a warning and rejects the MCP without deleting.
+   */
+  private AspectCheckpointConfig prePatch =
```
david-leifker: These configuration classes are directly populated by application.yaml. The defaults here are misleading and can easily become null when not expected. Because of this we want to avoid including default values in these configuration classes and make sure the defaults are specified once in application.yaml

metadata-service/configuration/src/main/java/com/linkedin/metadata/config/DataHubConfiguration.java
```diff
@@ -33,6 +33,8 @@ public class DataHubConfiguration {
 
   private S3Configuration s3;
 
+  private ValidationConfiguration validation = new ValidationConfiguration();
```
david-leifker: Similar to above avoid defaults in these configuration classes.

metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanSystemAspect.java
```diff
@@ -129,6 +164,64 @@ public EbeanSystemAspect forUpdate(
...
+                remediation.logLabel,
+                urn,
+                aspectName,
+                actualSize,
+                threshold);
+
+            // Handle oversized aspect according to remediation strategy
+            if (remediation == OversizedAspectRemediation.REPLACE_WITH_PATCH && aspectDao != null) {
+              try {
+                aspectDao.deleteAspect(urn, aspectName, 0L);
```
david-leifker: This isn't the proper way to delete an aspect. In order to process the aspect delete it must go back through the entity service. You'll want to restructure the logic to avoid the write (leave the SQL row), collect the deletes, and then execute those deletes outside the current transaction.

david-leifker: This leaves Elasticsearch indices, consumer hooks, and potentially graph edges unhandled leading to an inconsistent state.

metadata-service/configuration/src/main/java/com/linkedin/metadata/config/OversizedAspectRemediation.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Defines how to handle aspects that exceed the configured size threshold during MCP processing.
+ */
+public enum OversizedAspectRemediation {
+  /**
+   * Hard delete the oversized aspect from the database and skip the write. This prevents oversized
+   * aspects from accumulating in the database.
+   */
+  DELETE("deleted"),
```
david-leifker, line 22: See implementation issue above, the write can be skipped, but the aspect needs to be deleted from a `system` perspective, not just the SQL row.

metadata-service/configuration/src/main/java/com/linkedin/metadata/config/OversizedAspectRemediation.java
```diff
@@ -0,0 +1,35 @@
...
+   * oversized aspects to accumulate.
+   */
+  IGNORE("write skipped"),
+
+  /**
+   * (Pre-patch only) Replace the oversized existing aspect with the new patch as if it were an
+   * insert (no merge). Deletes the old oversized aspect and proceeds with the write operation. This
+   * allows the operation to succeed while cleaning up the oversized data.
+   */
+  REPLACE_WITH_PATCH("deleted and replaced with patch");
```
david-leifker: This is not going to work. A patch will often only contain a very little amount of data and likely would not make a valid aspect (missing required fields, etc). This option should be removed.

metadata-service/configuration/src/main/java/com/linkedin/metadata/config/ValidationConfiguration.java
```diff
@@ -0,0 +1,10 @@
+package com.linkedin.metadata.config;
+
+import lombok.Data;
+
+/** Validation configuration for DataHub operations. */
+@Data
+public class ValidationConfiguration {
+  /** Aspect size validation configuration (applies to ALL aspect writes: REST/GraphQL/MCP) */
+  private AspectSizeValidationConfig aspectSize = new AspectSizeValidationConfig();
```
david-leifker: Avoid defaults in config class

metadata-service/factories/src/main/java/com/linkedin/gms/factory/entity/EntityAspectDaoFactory.java
```diff
@@ -29,6 +35,14 @@ protected AspectDao createEbeanInstance(
     if (configurationProvider.getDatahub().isReadOnly()) {
       ebeanAspectDao.setWritable(false);
     }
+    if (serializationHooks != null && !serializationHooks.isEmpty()) {
```
david-leifker: Unless these are changing, I'd overload the constructor and pass them in above. The reason the setWritable() is using a setter is because the write mode can change at runtime.

metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanSystemAspect.java
rob-1019: Added

metadata-service/configuration/src/main/java/com/linkedin/metadata/config/OversizedAspectRemediation.java
rob-1019: PTAL - now uses ThreadLocal to collect deletion requests during validation, then calls `EntityService.deleteAspect()` after the db transaction commits.

metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectDao.java
rob-1019: no, changed to constructor injection with `final`

metadata-service/configuration/src/main/java/com/linkedin/metadata/config/OversizedAspectRemediation.java
rob-1019: Seems obvious in retrospect - removed.

entity-registry/src/main/java/com/linkedin/metadata/aspect/AspectPayloadValidator.java
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * <p><b>Naming:</b> "Validator" terminology is used for consistency with existing {@code
+ * AspectPayloadValidators} in {@code
+ * metadata-io/src/main/java/com/linkedin/metadata/aspect/validation/}. The term "Hook" is reserved
+ * for MCL hooks and ingestion consumers in this codebase.
+ *
+ * <p><b>Implementation Note:</b> Validators are registered as Spring beans and auto-injected via
+ * {@code @Autowired List<AspectPayloadValidator>} in {@code EntityAspectDaoFactory}.
+ */
+public interface AspectPayloadValidator {
```
david-leifker: This is going to be very confusing with the existing AspectPayloadValidator abstract class in `com.linkedin.metadata.aspect.plugins.validation`. Maybe we can just extend that one, maybe even in the same class?

david-leifker: The other validator classes are already injected as well so it might simplify the changes to get to just the validation logic being injected.

metadata-io/src/main/java/com/linkedin/metadata/entity/validation/AspectOperationContext.java
```diff
@@ -0,0 +1,77 @@
...
+ * <pre>
+ * try {
+ *   AspectOperationContext.set("isRemediationDeletion", true);
+ *   // ... operation that loads aspects ...
+ * } finally {
+ *   AspectOperationContext.clear(); // Always cleanup
+ * }
+ * </pre>
+ */
+public class AspectOperationContext {
```
david-leifker: This should probably be a simple context added to the existing OperationContext. While the OperationContext does thread this through method parameters, its purpose is exactly the same, to pass context from the initial request hitting the API to the low level DAO classes.

david-leifker: Nothing wrong with this approach, but doesn't follow existing conventions.

metadata-io/src/main/java/com/linkedin/metadata/entity/validation/AspectSizeExceededException.java
```diff
@@ -0,0 +1,36 @@
...
+import lombok.Getter;
+
+/**
+ * Exception thrown when an aspect exceeds configured size limits during validation.
+ *
+ * <p>Includes the validation point (PRE_DB_PATCH or POST_DB_PATCH), actual size, threshold, and
+ * aspect identity for debugging.
+ */
+@Getter
+public class AspectSizeExceededException extends RuntimeException {
```
david-leifker, line 13: This should probably be extending the existing AspectValidationException most likely. There is special handling for batch operations when validation errors occur and allows us to have a uniform exception handling for validation (i.e. returning error code when fatal)

david-leifker (commented):
The business logic around the validation, actions, and configuration all look good. Mostly we need to align the other logic into existing conventions instead of creating new code paths.

entity-registry/src/test/java/com/linkedin/metadata/aspect/EntityAspectTest.java
```diff
@@ -302,7 +303,8 @@ public void testGetSystemMetadata() {
             null,
             null,
             entityRegistry.getEntitySpec(DATASET_ENTITY_NAME),
-            null);
+            null,
+            Collections.emptyList());
```
RyanHolstien: There should be a test here where we inject the validator I would think

metadata-io/src/main/java/com/linkedin/metadata/entity/AspectSizePayloadValidator.java
```diff
@@ -0,0 +1,108 @@
...
+
+    String metadata = serializedAspect.getMetadata();
+    if (metadata == null) {
+      return; // No metadata to validate
+    }
+
+    long actualSize = metadata.length();
+    long threshold = config.getPostPatch().getMaxSizeBytes();
+
+    if (actualSize > threshold) {
```
RyanHolstien, line 247: Metrics would probably also be helpful here.

metadata-io/src/main/java/com/linkedin/metadata/entity/EntityServiceImpl.java
```diff
@@ -1037,305 +1047,429 @@ private IngestAspectsResult ingestAspectsToLocalDB(
...
+                            if (retentionService != null) {
+                              opContext.withSpan(
+                                  "retentionService",
+                                  () -> {
+                                    List<RetentionService.RetentionContext> retentionBatch =
+                                        upsertResults.stream()
+                                            // Only consider retention when there was a previous
+                                            // version
+                                            .filter(
+                                                upsertResult ->
```
RyanHolstien, line 1355: nit: Generally prefer to not do minor renames like this since it makes the PR harder to review with no actual benefit.

metadata-io/src/main/java/com/linkedin/metadata/entity/validation/AspectValidationContext.java
```diff
@@ -0,0 +1,74 @@
...
+ *   // After commit, process deletions
+ *   List&lt;AspectDeletionRequest&gt; deletions = AspectValidationContext.getPendingDeletions();
+ *   processDeletions(deletions);
+ * } finally {
+ *   AspectValidationContext.clearPendingDeletions(); // Always cleanup
+ * }
+ * </pre>
+ */
+public class AspectValidationContext {
+  private static final ThreadLocal<List<AspectDeletionRequest>> PENDING_DELETIONS =
```
RyanHolstien: Why are we doing a ThreadLocal static call here instead of injecting this context? This feels hacky.

metadata-io/src/main/java/com/linkedin/metadata/entity/EntityServiceImpl.java
```diff
@@ -1037,305 +1047,429 @@ private IngestAspectsResult ingestAspectsToLocalDB(
...
+                e.getMessage());
+            // Fall through to finally block to process deletions, then re-throw
+            throw e;
+          } finally {
+            // Process pending deletions whether transaction succeeded or failed.
+            // For DELETE remediation, this is where actual deletion happens.
+            // Must run in finally block to ensure deletions are processed even when
+            // AspectSizeExceededException is thrown during validation.
+            List<AspectDeletionRequest> pendingDeletions =
+                AspectValidationContext.getPendingDeletions();
```
RyanHolstien: Feels weird for the AspectValidationContext to be responsible for storing the pending deletions, but not for handling them. Having the processing in a private method local to this class doesn't seem quite right from an abstraction perspective. I would think you would just pass in what is needed from this class for that one to store so you don't need this chain of calls and so that the AspectValidationContext can do all of the processing itself. Also why is it static when everything else is being injected with Spring? The pending deletions should be localized to the current operation context

metadata-service/factories/src/main/java/com/linkedin/gms/factory/entity/EntityAspectDaoFactory.java
```diff
@@ -2,30 +2,51 @@
...
   @Bean(name = "entityAspectDao")
   @ConditionalOnProperty(name = "entityService.impl", havingValue = "ebean", matchIfMissing = true)
   @Nonnull
   protected AspectDao createEbeanInstance(
       @Qualifier("ebeanServer") final Database server,
       final ConfigurationProvider configurationProvider,
       final MetricUtils metricUtils) {
+    log.debug(
+        "Creating EntityAspectDao with {} AspectPayloadValidators: {}",
+        payloadValidators != null ? payloadValidators.size() : 0,
```
RyanHolstien: nit: Lots of ternaries in this when you could just set it to:
List<> payLoadValidatorsList = payloadValidators != null ? payloadValidators : Collections.singletonList();

metadata-io/src/test/java/com/linkedin/metadata/entity/validation/ValidationPointTest.java
```diff
@@ -0,0 +1,28 @@
+package com.linkedin.metadata.entity.validation;
+
+import static org.testng.Assert.*;
+
+import org.testng.annotations.Test;
+
+public class ValidationPointTest {
+
+  @Test
+  public void testEnumValues() {
```
RyanHolstien: I think this might be the most AI generated test of all time 😅

RyanHolstien:
I'm a bit confused as to why this isn't using existing abstractions. Why are we injecting in new hook points when "Pre Patch" should be equivalent to validateProposedAspects and "Post Patch" should be equivalent to validatePreCommitAspects? Why are we creating a new OperationContext concept that instead of using Spring DI is using a static ThreadLocal?

metadata-io/src/main/java/com/linkedin/metadata/entity/AspectSizePayloadValidator.java
rob-1019: Will do.

entity-registry/src/main/java/com/linkedin/metadata/aspect/AspectPayloadValidator.java
rob-1019: Renaming to disambiguate AspectPayloadValidator's to SystemAspectValidator. I'm expecting the old AspectPayloadValidator interfaces to be pre-serialization and thus not to have the post-serialization byte-size we want to validate here. We don't want to pay the performance cost to serialize twice since this is in the hot path.

metadata-io/src/main/java/com/linkedin/metadata/entity/EntityServiceImpl.java
rob-1019: Scoping issue so I resolved by making the more tightly scoped release the more specifically named one.

```
error: variable result is already defined in method ingestAspectsToLocalDB(OperationContext,AspectsBatch,boolean)
```

rob-1019:
> Why are we injecting in new hook points when "Pre Patch" should be equivalent to validateProposedAspects and "Post Patch" should be equivalent to validatePreCommitAspects?
prePatch runs before serialization has occurred and validateProposedAspects runs after - since we are trying to skip the deserialization of oversized json thats too late for our purposes. Similarly on the other side validatePreCommitAspects runs too early since it doesn't have the serialized representation of the object yet and thus we can't check to see if its oversized. We could eagerly serialize it there and check but we'd need to pass it around afterward or pay the serialization cost twice - which we can't really afford on the path.

> Why are we creating a new OperationContext concept that instead of using Spring DI is using a static ThreadLocal?
Ahh yes, much better.

metadata-io/src/test/java/com/linkedin/metadata/entity/validation/ValidationPointTest.java
rob-1019: Another good reason to add net-new metrics - they give something better to jinn the codecov results. Goodhart's law strikes again.

metadata-io/src/main/java/com/linkedin/metadata/entity/validation/AspectValidationContext.java
rob-1019: Concur - suggestion to use OperationContext was much appreciated.

metadata-io/src/main/java/com/linkedin/metadata/entity/validation/AspectSizeExceededException.java
rob-1019: requires BatchItem context that doesn't exist at the validator level - wrapping it at the applyUpsert() instead

metadata-service/factories/src/main/java/com/linkedin/gms/factory/entity/EntityAspectDaoFactory.java
rob-1019: tidied

metadata-io/src/main/java/com/linkedin/metadata/entity/EntityServiceImpl.java
rob-1019: yep, your suggestion to use OperationContext fixed this one too. Prior implementation was static to provide per-thread global state akin to what is effectively and explicitly provided by OperationContext.

entity-registry/src/test/java/com/linkedin/metadata/aspect/EntityAspectTest.java
```diff
@@ -351,4 +355,41 @@ public void testGetSystemMetadataFromEntityAspect() {
     SystemMetadata cachedResult = systemAspect.getSystemMetadata();
     assertSame(result, cachedResult, "Second call should return the same cached object");
   }
+
+  @Test
+  public void testPayloadValidatorInjection() {
+    // Test that SystemAspectValidators injected via payloadValidators list are actually invoked
+    // during withVersion() - addresses RyanHolstien feedback
```
RyanHolstien: Doesn't need to specify this in the code comment 😄

entity-registry/src/testFixtures/java/com/linkedin/test/metadata/aspect/batch/TestSystemAspect.java
```diff
@@ -83,4 +84,10 @@ public Optional<SystemAspect> getDatabaseAspect() {
   public SystemAspect setDatabaseAspect(@Nonnull SystemAspect databaseAspect) {
     return this;
   }
+
+  @Nullable
+  @Override
+  public Object getOperationContext() {
+    return null; // Not applicable for test implementation
```
RyanHolstien: nit: feels weird to override this, can't it just return the not set field?

metadata-service/factories/src/main/java/com/linkedin/gms/factory/entity/EntityAspectDaoFactory.java
```diff
@@ -2,30 +2,52 @@
...
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.DependsOn;
 
+@Slf4j
 @Configuration
 public class EntityAspectDaoFactory {
 
+  @Autowired(required = false)
+  private List<SystemAspectValidator> payloadValidators;
```
RyanHolstien: nit: Naming inconsistency, payload validators vs system aspect

metadata-service/configuration/src/main/java/com/linkedin/metadata/config/AspectSizeValidationConfig.java
```diff
@@ -0,0 +1,55 @@
...
+ * Configuration for aspect size validation that applies to ALL aspect writes (REST API, GraphQL,
+ * MCP, etc.). Will be nested under datahub.validation.aspectSize in application.yaml.
+ *
+ * <p>All defaults are specified in application.yaml via environment variables. No defaults in Java
+ * code to avoid confusion when values become null unexpectedly during config refresh.
+ */
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+public class AspectSizeValidationConfig {
```
RyanHolstien: nit: Differing naming from standard, Config vs Configuration and mismatch between name of class and config property makes it less discoverable from application.yaml

metadata-service/configuration/src/main/java/com/linkedin/metadata/config/AspectSizeValidationConfiguration.java
```diff
@@ -0,0 +1,55 @@
...
+   * Validates aspect after patch application, in DAO before DB write (measures: serialized JSON
+   * character count, same unit as prePatch). Use to catch bloat from patch application. Validation
+   * happens on JSON already created for DB write - zero additional serialization cost.
+   */
+  private AspectCheckpointConfig postPatch;
+
+  /** Metrics configuration for size distribution tracking. */
+  private MetricsConfig metrics;
+
+  @Data
```
RyanHolstien, line 35: I don't mind this, but it does differ from the standard set up where each is a full standalone class generally.

entity-registry/src/main/java/com/linkedin/metadata/aspect/EnvelopedSystemAspect.java
```diff
@@ -93,6 +93,12 @@ public EntityAspect withVersion(long version) {
     throw new NotImplementedException();
   }
 
+  @Nullable
+  @Override
+  public Object getOperationContext() {
+    return null; // Not applicable for this implementation
```
RyanHolstien: nit: Same here feels weird to explicitly override a getter to return null rather than just returning the empty unset field.

metadata-io/src/main/java/com/linkedin/metadata/entity/EntityServiceImpl.java
```diff
@@ -327,6 +340,9 @@ static SystemAspect applyUpsert(ChangeMCP changeMCP, SystemAspect latestAspect)
       } else {
         // insert
         return EbeanSystemAspect.builder()
+            .payloadValidators(payloadValidators)
```
RyanHolstien: Any reasoning behind separating out the validators rather than passing them around injected into the operation context? I don't think it makes a big difference, just curious

metadata-io/src/main/java/com/linkedin/metadata/entity/cassandra/CassandraAspectDao.java
```diff
@@ -89,9 +100,16 @@ public SystemAspect getLatestAspect(
     validateConnection();
     return Optional.ofNullable(getAspect(urn, aspectName, ASPECT_LATEST_VERSION))
         .map(
-            a ->
-                EntityAspect.EntitySystemAspect.builder()
-                    .forUpdate(a, opContext.getEntityRegistry()))
+            a -> {
+              // Pre-patch validation if this is for update
+              if (forUpdate) {
```
RyanHolstien: Shouldn't this be a hook point for the list of validators rather than calling the specific one?

metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanSystemAspect.java
```diff
@@ -129,6 +178,17 @@ public EbeanSystemAspect forUpdate(
       this.ebeanAspectV2 = ebeanAspectV2;
       this.urn = UrnUtils.getUrn(ebeanAspectV2.getUrn());
       this.aspectName = ebeanAspectV2.getAspect();
+
+      // Pre-patch validation: check existing aspect size from DB using utility
+      // Get context from OperationContext (may contain isRemediationDeletion flag)
+      AspectSizeValidator.validatePrePatchSize(
```
RyanHolstien: Same with this, would expect this to be a hook point for pre-patch enabled validators rather than an explicit call to an impl class.

metadata-io/src/main/java/com/linkedin/metadata/entity/validation/AspectSizeValidator.java
```diff
@@ -0,0 +1,229 @@
...
+   * @param rawMetadata serialized aspect JSON from database (may be null for new aspects)
+   * @param urn entity URN
+   * @param aspectName aspect name
+   * @param config aspect size validation configuration (may be null if validation disabled)
+   * @param opContext optional operation context (may contain remediation deletion flag to skip
+   *     validation)
+   * @param metricUtils optional metrics utility for emitting validation metrics
+   * @throws AspectSizeExceededException if aspect exceeds configured size threshold
+   */
+  public static void validatePrePatchSize(
```
RyanHolstien: Was this intended as an extension point or is static util the right call? A little confused on the pattern difference. Way I was kind of looking at the PR was that you were essentially adding a couple new hook points for abstract validators to run at, but this is limited to whatever functionality we put into this static util.

RyanHolstien (commented):
Overall looking much cleaner from my perspective, a couple questions and minor convention nitpicks.

metadata-io/src/main/java/com/linkedin/metadata/entity/validation/AspectSizeValidator.java
rob-1019: take a look now - refactored it to use the validator list pattern like post-patch

metadata-io/src/main/java/com/linkedin/metadata/entity/EntityServiceImpl.java
rob-1019: no, we could pretext it and say it keeps OperationsContext oriented on request-specific state but that's pretty weak-sauce on my part. Leaving at is but let me know if you want me to tweak.

entity-registry/src/main/java/com/linkedin/metadata/aspect/EnvelopedSystemAspect.java
rob-1019: added default implementation to SystemAspect interface

smoke-test/tests/openapi/v1/test_tracking.py
```diff
@@ -250,15 +250,19 @@ def test_tracking_api_kafka(auth_session):
         if msg is None:
             time.sleep(0.1)  # Small delay between polls
             continue
-        if msg.error():
```
RyanHolstien: Weird that this file got changed 🤔 , but looks like just clean up. Seems fine

PR #15755, opened 2025-12-23
Add mTLS Support and Truststore Handling to GMS

relaxedboi:
Summary
This PR adds support for mutual TLS (mTLS) and truststore-based client certificate validation in the GMS Jetty server, while keeping the default behavior as HTTP‑only when no SSL configuration is provided.

Key Changes
Added support for server.ssl.client-auth (NONE, WANT, NEED) to enable optional or required client certificates.
Added loading of truststore properties (trust-store, trust-store-password, trust-store-type) to validate client certificates when mTLS is enabled.
HTTPS is enabled only when a keystore is configured; truststore is used only when present.
Added warnings when client-auth=NEED is enabled but no truststore is provided.
Default mode remains HTTP-only when no SSL-related config is passed.
Result
GMS now cleanly supports:

HTTP (default)
TLS (server authentication)
mTLS (mutual certificate authentication)

### DIFF
--- a/metadata-ingestion/docs/sources/dbt/dbt.md
+++ b/metadata-ingestion/docs/sources/dbt/dbt.md
@@ -293,6 +293,24 @@ source:
       test_results: No
 ```
 
+:::note Tests not showing up in the Assertion UI?
+
+Double check you are using the same `job_id` for your `test_results: Only` and `test_results: No` recipes. Otherwise you might end up reporting `tests_results` for tests that haven't had their `test_definitions` ingested yet. This can lead to orphaned assertions that do not show up under your dataset.
+
+If you choose not to share a `job_id`, you should adjust your recipe to also ingest the `test_definitions`
+
+```yaml
+entities_enabled:
+  models: No
+  sources: No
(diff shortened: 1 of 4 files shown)


deepgarg760:
@relaxedboi What use case we are trying to achieve here with mTLS ?

relaxedboi:
> @relaxedboi What use case we are trying to achieve here with mTLS ?

Wanted to add mTLS as a good‑to‑have option. With mTLS enabled, we gain an additional layer of authentication between clients and GMS, ensuring that both sides verify each other’s identity before any request is processed. This strengthens security for deployments that want tighter control over service‑to‑service communication, while still keeping mTLS optional for those who don’t need it.

deepgarg760:
@relaxedboi , add few testcases for code coverage

relaxedboi:
> @relaxedboi , add few testcases for code coverage

I have covered the lines that I have added, but report is not picking the tests.

relaxedboi:
@deepgarg760 All the checks are now success.

PR #15764, opened 2025-12-25
feat(docker): add Alpine-based variants for ingestion and actions images

esteban:
Introduce lightweight Alpine Linux variants for datahub-ingestion and
datahub-actions Docker images, providing smaller image sizes and reduced
attack surface.

- `datahub-ingestion:*-slim-alpine` - Lightweight ingestion image
- `datahub-ingestion:*-locked-alpine` - Locked variant with PyPI Index disabled
- `datahub-actions:*-slim-alpine` - Lightweight actions image
- `datahub-actions:*-locked-alpine` - Locked variant with PyPI Index disabled

| Feature                | Ubuntu Full | Ubuntu Slim | Alpine Slim |
|------------------------|-------------|-------------|-------------|
| Base connectors        | Yes         | Yes         | Yes         |
| Cloud storage (S3/GCS) | Yes         | Yes         | Yes         |
| Snowflake/BigQuery     | Yes         | Yes         | No          |
| MySQL/Postgres         | Yes         | Yes         | No          |
| dbt/Looker/Tableau     | Yes         | Yes         | No          |
| JRE (Spark/Trino)      | Yes         | No          | No          |
| Oracle client          | Yes         | No          | No          |
| Allows PIP extra index | Yes         | Yes         | No*         |

| Feature                | Ubuntu Full | Ubuntu Slim | Alpine Slim |
|------------------------|-------------|-------------|-------------|
| Core actions           | Yes         | Yes         | Yes         |
| Kafka/Executor/Slack   | Yes         | Yes         | Yes         |
| Tag/Term propagation   | Yes         | Yes         | Yes         |
| Snowflake propagation  | Yes         | Yes         | No**        |
| Bundled CLI venvs      | Yes         | Yes         | Yes         |
| Allows PIP extra index | Yes         | Yes         | No*         |

| Testing level          | Ubuntu Full | Ubuntu Slim | Alpine Slim |
|------------------------|-------------|-------------|-------------|
| CI/CD                  | Yes         | Yes         | No***       |
| e2e                    | Yes         | Yes         | No***       |

*locked variants have PIP extra index disabled
**excluded due to spacy/blis musl compatibility
***manual testing has been performed on critical functionality, including additional variants CI/CD is TBD.

Alpine variants are built:
- On releases: automatically included
- On PRs: opt-in via `build-alpine-variant` label
- Locally: `./gradlew :docker:... -PincludeAlpineVariants=true`

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -49,6 +49,87 @@ Do not use `latest` or `debug` tags for any of the image as those are not suppor
 - [acryldata/datahub-postgres-setup](https://hub.docker.com/r/acryldata/datahub-postgres-setup/)
 - [acryldata/datahub-actions](https://hub.docker.com/r/acryldata/datahub-actions). Do not use `acryldata/acryl-datahub-actions` as that is deprecated and no longer used.
 
+## Image Variants
+
+The `datahub-ingestion` and `datahub-actions` images are available in multiple variants optimized for different use cases:
+
+| Variant          | Base OS      | Image Size | Use Case                                |
+| ---------------- | ------------ | ---------- | --------------------------------------- |
+| `full` (default) | Ubuntu 24.04 | Largest    | All connectors, maximum compatibility   |
+| `slim`           | Ubuntu 24.04 | Medium     | Common connectors, good balance         |
+| `slim-alpine`    | Alpine 3.22  | Small      | Minimal footprint, security-focused     |
+| `locked`         | Ubuntu 24.04 | Medium     | Air-gapped environments (pypi disabled) |
+| `locked-alpine`  | Alpine 3.22  | Smallest   | Air-gapped slim-alipine (pypi disabled) |
+
+### Variant Tag Format
+
+```
+acryldata/datahub-ingestion:v0.x.y          # full (default)
+acryldata/datahub-ingestion:v0.x.y-slim     # slim
+acryldata/datahub-ingestion:v0.x.y-slim-alpine
+acryldata/datahub-ingestion:v0.x.y-locked
+acryldata/datahub-ingestion:v0.x.y-locked-alpine
+```
+
+### datahub-ingestion Feature Matrix
+
+| Feature               | full | slim | slim-alpine | locked | locked-alpine |
+| --------------------- | :--: | :--: | :---------: | :----: | :-----------: |
+| Core CLI & REST/Kafka | Yes  | Yes  |     Yes     |  Yes   |      Yes      |
+| S3 / GCS / Azure Blob | Yes  | Yes  |     Yes     |  Yes   |      Yes      |
+| Snowflake             | Yes  | Yes  |      -      |   -    |       -       |
+| BigQuery              | Yes  | Yes  |      -      |   -    |       -       |
+| Redshift              | Yes  | Yes  |      -      |   -    |       -       |
+| MySQL / PostgreSQL    | Yes  | Yes  |      -      |   -    |       -       |
+| ClickHouse            | Yes  | Yes  |      -      |   -    |       -       |
+| dbt                   | Yes  | Yes  |      -      |   -    |       -       |
+| Looker / LookML       | Yes  | Yes  |      -      |   -    |       -       |
+| Tableau / PowerBI     | Yes  | Yes  |      -      |   -    |       -       |
+| Superset              | Yes  | Yes  |      -      |   -    |       -       |
+| Glue                  | Yes  | Yes  |      -      |   -    |       -       |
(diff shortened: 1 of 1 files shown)


gradle/docker/docker.gradle
```diff
@@ -292,11 +292,19 @@ project.afterEvaluate {
           bake_spec_target.name = "${project.name}-\${variants.tagSuffix}"
           bake_spec_target.matrix = [variants: []]
 
+          // Check if any variant has an explicit target
+          def hasExplicitTargets = extension.variants.get().any { variant, variantSpec -> variantSpec.target != null }
```
chakru-r: While this works, why do we need to add explicit targets here when the target is already a function of APP_ENV -- the last line in datahub-actions/Dockerfile --  final-${APP_ENV} from [this](https://github.com/datahub-project/datahub/pull/15764/files#diff-2a511c8e1ac6dc91843bd9b857c42ce9afd9b2154ac3ff6a1607fd049f324122R334) line

.github/workflows/docker-unified.yml
```diff
@@ -34,6 +34,8 @@ env:
   DOCKER_CACHE: "DEPOT"
   DEPOT_PROJECT_ID: "s0gr1cr3jd"
   HAS_DEPOT_LABEL: ${{ github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'depot') }}
+  # Include Alpine variants for releases, or when PR has 'build-alpine-variant' label
+  INCLUDE_ALPINE_VARIANTS: ${{ github.event_name == 'release' || (github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'build-alpine-variant')) }}
```
chakru-r, line 38: If we are publishing all variants, isnt it simpler to do the same for on pr publishing? Anyway local builds will not build all variants, but just one default variant (that can be switched to any one variant for testing)

esteban: My reasoning was not to add extra build time for regular PRs unless the alpine variant is requested. Some of that build time might not be relevant due caching after the initial PR is published for alpine artifact but for most of the devs waiting for the alpine variant seems unnecessary.

gradle/docker/docker.gradle
esteban: If I understand the bake matrix build correctly thats necessary otherwise bake will build only the last stage and use `${APP_ENV}` for interpolation, and for the matrix build you need to have the explicit targets to be defined. When I was implementing this I ran into the situation that the final images were ubuntu based instead of alpine. After I moved to explicit targets to include them in `bake_spec_target.target` the final image was build correctly.

chakru-r: There was a bug in how variant specific args were being merged -- I just pushed a commit that just reverses the order in which build args (common ones) and variant args) are merged so that variant args are last and retained. With this APP_ENV is retained correctly and the last stage will get picked correctly because of APP_ENV 
```
FROM final-${APP_ENV} AS final
```

Take a look at the last commit I pushed to your PR -- I removed the explicit target override (which works, but not really needed) and am relying on APP_ENV being correct. I verified a build with actions images, and see correct image being published for each tag. 
Take a look -- if you see something wrong, just revert my commit from your PR.

.github/workflows/docker-unified.yml
chakru-r: OK - since most PRs are not run with PR labels, so we wont incur this cost. For PRs without pr-publish, Only the default variant is built-- since that is sufficnet for smoke-tests. We can take this in another PR.

docker/datahub-ingestion/Dockerfile
```diff
@@ -133,40 +137,200 @@ USER datahub
...
+    bash \
+    curl \
+    ca-certificates \
+    libldap \
+    libsasl \
+    cyrus-sasl-gssapiv2 \
+    krb5-libs \
+    unixodbc \
+    libffi \
+    && apk --no-cache add librdkafka --repository=${ALPINE_REPO_URL}/edge/community
```
RyanHolstien, line 225: Any stability concerns with these dev and edge builds?

gradle/docker/docker.gradle
esteban: +1 lgtm, @chakru-r

docker/datahub-ingestion/Dockerfile
esteban: Not really, edge has been frequently up to date and stable enough and is not that we are bringing the whole world from edge but just specific packages when needed.

esteban:
@david-leifker added notes to docker/README.md

PR #15774, opened 2025-12-29
fix(ui) Refresh user data after user signs up

chriscollins3456:
This fixes a bug where if a user signs up using an invite token granting them a role, the UI would not reflect that role in what the user can see until after some time (~15-20 seconds) passes and they refresh the page. After digging deeper I believe this is due to caching in the authorization layer, but busting that cache is complex and could lead to performance issues.

This solution simply checks if a user has just signed up using a new key: value in localStorage and refetches the user using the getMe query every 5 seconds until a max of 60 seconds ensuring we get the updated permissions without the user having to refresh the page.


chriscollins3456:
closing in favor of https://github.com/datahub-project/datahub/pull/15782

PR #15793, opened 2026-01-03
feat(ingest/keycloak): Add Keycloak ingestion source to ingest groups, users and memberships

DucNgoQuang:
- Ingests users, groups, and memberships for a specific Keycloak client.
- Uses `python-keycloak` for metadata extraction.
- Supports direct memberships only (ignores parent group hierarchies).
- Maps User URNs by using email then username .


deepgarg760:
@DucNgoQuang , Thanks for raising this

PR #15794, opened 2026-01-05
fix(ui): Improve More filters UX

sakethvarma397:
Add truncation for long filter options like structured properties, and add a tooltip to see the full name

Screenshot 2026-01-12 at 6 18 49 PM

### DIFF
--- a/datahub-web-react/src/app/searchV2/filters/MoreFilterOption.tsx
+++ b/datahub-web-react/src/app/searchV2/filters/MoreFilterOption.tsx
@@ -1,5 +1,6 @@
 import { RightOutlined } from '@ant-design/icons';
 import { Tooltip } from '@components';
+import { Typography } from 'antd';
 import React, { useRef } from 'react';
 import styled from 'styled-components';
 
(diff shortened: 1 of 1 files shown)


datahub-web-react/src/app/searchV2/filters/MoreFilterOption.tsx
```diff
@@ -69,11 +77,15 @@ export default function MoreFilterOption({ filter, filterPredicates, activeFilte
                 data-testid={`more-filter-${displayName?.replace(/\s/g, '-')}`}
                 ref={labelRef}
             >
-                <IconNameWrapper>
-                    {filterIcon && <IconWrapper>{filterIcon}</IconWrapper>}
-                    {displayName} {numActiveFilters ? `(${numActiveFilters}) ` : ''}
-                </IconNameWrapper>
-                <StyledRightOutlined />
+                <Tooltip title={displayName} placement="left">
```
chriscollins3456, line 75: this would then add a tooltip on all filter options right? do we want that or just want tooltips on truncated options?

i think ant design's `Text` component allows you to do this if you look at other examples we have all over the app with the `ellipses` property we pass in

sakethvarma397: Ah great it detects when to show the tooltip

PR #15803, opened 2026-01-06
docs(ingestion): adding additional details to security comparison

gabe-lyons:
Adding additional context for security and networking concerns for each ingestion approach.


maggiehays:
@gabe-lyons link to the new image as WIP https://github.com/datahub-project/static-assets/blob/main/imgs/remote-executor/metadata-ingestion-security-comparison.png

PR #15813, opened 2026-01-06
feat(ingestion/deps): add upper bounds to dependency versions in setup.py

kyungsoo-datahub:
- Add upper bounds to all dependencies using next major version (e.g., <3.0.0)
- For 0.x packages, use tight bounds (<=current) or <1.0.0 where no 1.x exists
- Add comments documenting automatic dependency chains:
  - ex> numpy<2 -> feast<=0.47 -> pyarrow<18.1 (resolved automatically)
- Keep explicit constraints for deliberate choices (sqlalchemy<2, numpy<2)

### DIFF
--- a/metadata-ingestion/setup.py
+++ b/metadata-ingestion/setup.py
@@ -46,7 +46,7 @@ framework_common = {
     "psutil>=5.8.0,<8.0.0",
     "Deprecated<2.0.0",
     "humanfriendly<11.0.0",
-    "packaging<25.0.0",
+    "packaging<26.0.0",
     "aiohttp<4",
     "cached_property<3.0.0",
     "ijson<4.0.0",
@@ -580,7 +580,7 @@ plugins: Dict[str, Set[str]] = {
     },
     "datahub-debug": {"dnspython==2.7.0", "requests<3.0.0"},
     "mode": {"requests<3.0.0", "python-liquid<2", "tenacity>=8.0.1,<9.0.0"} | sqlglot_lib,
-    "mongodb": {"pymongo>=4.8.0,<5.0.0", "packaging<25.0.0"},
+    "mongodb": {"pymongo>=4.8.0,<5.0.0", "packaging<26.0.0"},
     "mssql": sql_common | mssql_common,
     "mssql-odbc": sql_common | mssql_common | {"pyodbc<6.0.0"},
     "mysql": sql_common | mysql | aws_common,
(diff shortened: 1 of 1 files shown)


metadata-ingestion/setup.py
```diff
@@ -15,35 +15,35 @@
...
-    "sentry-sdk>=1.33.1",
+    "sentry-sdk>=1.33.1,<3.0.0",
 }
 
 framework_common = {
     # Avoiding click 8.2.0 due to https://github.com/pallets/click/issues/2894
-    "click>=7.1.2, !=8.2.0",
+    "click>=7.1.2,!=8.2.0,<9.0.0",
     "click-default-group",
     "PyYAML",
```
askumar27: [MINOR] There are still unbounded packages, plan to upper-bound all the packages right ?

metadata-ingestion/setup.py
```diff
@@ -307,15 +310,16 @@
     # especially for AWS-based catalogs and warehouses, the properties `profile_name`, `region_name`,
     # `aws_access_key_id`, `aws_secret_access_key`, and `aws_session_token` were deprecated and removed in version
     # 0.8.0.
-    "pyiceberg[glue,hive,dynamodb,snappy,hive,s3fs,adlfs,pyarrow,zstandard]>=0.8.0",
+    "pyiceberg[glue,hive,dynamodb,snappy,hive,s3fs,adlfs,pyarrow,zstandard]>=0.8.0,<=0.10.0",
```
askumar27: `pyiceberg` is still actively developed - should we strict pin to the latest version ? same with `databricks-sdk`

metadata-ingestion/setup.py
```diff
@@ -490,71 +494,73 @@
...
     "dynamodb": aws_common | classification_lib,
     # Starting with 7.14.0 python client is checking if it is connected to elasticsearch client. If its not it throws
     # UnsupportedProductError
     # https://www.elastic.co/guide/en/elasticsearch/client/python-api/current/release-notes.html#rn-7-14-0
     # https://github.com/elastic/elasticsearch-py/issues/1639#issuecomment-883587433
     "elasticsearch": {"elasticsearch==7.13.4", *cachetools_lib},
     "excel": {
-        "openpyxl>=3.1.5",
+        "openpyxl>=3.1.5,<4.0.0",
         "pandas",
```
askumar27: Let add a upper bound here as well

metadata-ingestion/setup.py
kyungsoo-datahub: Done. Thanks.

metadata-ingestion/setup.py
kyungsoo-datahub: Added. Thanks.

PR #15816, opened 2026-01-07
feat(auth): add VIEW_ENTITY_PAGE authorization for Views

pedro93:
## Summary
- Adds `VIEW_ENTITY_PAGE` authorization support for Views when Access Authorization is enabled
- Previously, Views did not respect access control policies because the `dataHubView` entity type was missing from the authorization framework
- Users can now create Metadata Policies targeting Views with "View Entity Page" privilege to control visibility

## Changes
- Add `DATAHUB_VIEW_ENTITY_NAME` to `VIEW_RESTRICTED_ENTITY_TYPES` in AuthUtil.java
- Add `canView` authorization filter to `DataHubViewType.batchLoad()`
- Add `DATAHUB_VIEW_PRIVILEGES` resource definition with `VIEW_ENTITY_PAGE` and `EXISTS_ENTITY` privileges

## Test plan
- [x] Existing `DataHubViewTypeTest` tests pass
- [x] `metadata-utils` and `metadata-auth:auth-api` compile successfully
- [x] Manual testing: Create a Metadata Policy for Views with "View Entity Page" privilege and verify Views respect it

### DIFF
--- a/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/types/view/DataHubViewType.java
+++ b/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/types/view/DataHubViewType.java
@@ -58,7 +58,7 @@ public class DataHubViewType
                   .collect(Collectors.toSet()),
               ASPECTS_TO_FETCH);
 
-      final List<EntityResponse> gmsResults = new ArrayList<>(urns.size());
+      final List<EntityResponse> gmsResults = new ArrayList<>();
       for (Urn urn : viewUrns) {
         gmsResults.add(entities.getOrDefault(urn, null));
       }
--- a/datahub-graphql-core/src/test/java/com/linkedin/datahub/graphql/types/view/DataHubViewTypeTest.java
+++ b/datahub-graphql-core/src/test/java/com/linkedin/datahub/graphql/types/view/DataHubViewTypeTest.java
@@ -4,7 +4,13 @@ import static com.linkedin.metadata.utils.CriterionUtils.buildCriterion;
 import static org.mockito.ArgumentMatchers.any;
 import static org.testng.Assert.*;
 
+import com.datahub.authentication.Actor;
+import com.datahub.authentication.ActorType;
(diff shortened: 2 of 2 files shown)


datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/types/view/DataHubViewType.java
```diff
@@ -53,10 +53,12 @@ public List<DataFetcherResult<DataHubView>> batchLoad(
...
               context.getOperationContext(),
               DATAHUB_VIEW_ENTITY_NAME,
-              new HashSet<>(viewUrns),
+              viewUrns.stream()
+                  .filter(urn -> canView(context.getOperationContext(), urn))
+                  .collect(Collectors.toSet()),
               ASPECTS_TO_FETCH);
 
-      final List<EntityResponse> gmsResults = new ArrayList<>();
+      final List<EntityResponse> gmsResults = new ArrayList<>(urns.size());
```
RyanHolstien: nit: Why change this? If anything the size of it should be viewUrns, but this isn't necessary to change.

RyanHolstien: That being said this bit of code is weird, intentionally injecting nulls into the list feels wrong.

pedro93: It was a micro-optimization claude did to pre-fill the array so memory increases don't have to happen in the loop below

RyanHolstien: But it's not an optimization because it's prefilling the size to a different list size than what is being inserted into it. Per the loop directly underneath it is inserting up to size of viewUrns NOT urns. Generally it's unnecessary as well: https://stackoverflow.com/a/16096863

david-leifker (approved):

PR #15819, opened 2026-01-07
fix(ingestion): handle oversized viewProperties and check all aspect types

kyungsoo-datahub:
- Add ensure_view_properties_size() to truncate oversized viewLogic/formattedViewLogic
  - Change elif to if to check ALL aspect types per workunit (fixes MCE handling)
  - Add unit tests for viewProperties truncation and MCE multi-aspect checking

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/api/auto_work_units/auto_ensure_aspect_size.py
+++ b/metadata-ingestion/src/datahub/ingestion/api/auto_work_units/auto_ensure_aspect_size.py
@@ -1,7 +1,8 @@
 import json
 import logging
 import os
-from typing import TYPE_CHECKING, Iterable, List
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Iterable, List, Optional
 
 from datahub.emitter.rest_emitter import INGEST_MAX_PAYLOAD_BYTES
 from datahub.emitter.serialization_helper import pre_json_transform
@@ -18,6 +19,15 @@ from datahub.metadata.schema_classes import (
 
 if TYPE_CHECKING:
     from datahub.ingestion.api.source import SourceReport
+
+
+@dataclass
+class _TruncationResult:
+    """Result of a field truncation operation."""
+
+    truncated_value: str
+    original_size: int
+    retained_length: int  # Length of the retained content (0 if removed entirely)
 
 
 # TODO: ordering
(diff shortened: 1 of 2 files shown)


metadata-ingestion/src/datahub/ingestion/api/auto_work_units/auto_ensure_aspect_size.py
```diff
@@ -371,6 +372,122 @@ def ensure_query_properties_size(
...
+                    f"Removing viewLogic entirely."
+                )
+                view_properties.viewLogic = (
+                    f"[removed due to size - original was {view_logic_size} chars]"
+                )
+                self.report.warning(
+                    title="View properties truncated due to size constraint",
+                    message="View properties contained too much data and would have caused ingestion to fail",
+                    context=f"viewLogic was removed for {entity_urn} (original was {view_logic_size} chars)",
+                )
```
sgomezvillamor: we are applying same logic in two fields: `view_properties.formattedViewLogic` first and then `view_properties.viewLogic` next
may we have a function to avoid the code duplication?

metadata-ingestion/src/datahub/ingestion/api/auto_work_units/auto_ensure_aspect_size.py
```diff
@@ -380,16 +497,21 @@ def ensure_aspect_size(
         on GMS side and failure of the entire ingestion. This processor will attempt to trim suspected aspects.
         """
         for wu in stream:
-            # logger.debug(f"Ensuring size of workunit: {wu.id}")
+            urn = wu.get_urn()
 
+            # Use 'if' instead of 'elif' to check ALL aspect types.
```
sgomezvillamor, line 523: nice catch!

metadata-ingestion/tests/unit/api/source_helpers/test_ensure_aspect_size.py
```diff
@@ -794,3 +795,177 @@ def test_wu_processor_triggered_by_query_properties_aspect(
...
+    assert final_size < INGEST_MAX_PAYLOAD_BYTES
+
+
+@patch(
+    "datahub.ingestion.api.auto_work_units.auto_ensure_aspect_size.EnsureAspectSizeProcessor.ensure_schema_metadata_size"
+)
+@patch(
+    "datahub.ingestion.api.auto_work_units.auto_ensure_aspect_size.EnsureAspectSizeProcessor.ensure_view_properties_size"
+)
+def test_mce_workunit_checks_all_aspects_not_just_first_elif_regression(
```
sgomezvillamor, line 963: beyond this test that is testing multiple aspects are checked, we don't have any test using `proper_view_properties`

IMO we miss a positive test where the view properties aspect is not updated if proper (short enough) view properties is passed.

metadata-ingestion/src/datahub/ingestion/api/auto_work_units/auto_ensure_aspect_size.py
kyungsoo-datahub: Thanks!

metadata-ingestion/src/datahub/ingestion/api/auto_work_units/auto_ensure_aspect_size.py
kyungsoo-datahub: Thank you for the suggestion. Revised to remove code duplication.

metadata-ingestion/tests/unit/api/source_helpers/test_ensure_aspect_size.py
kyungsoo-datahub: test_ensure_size_of_proper_view_properties test added.

PR #15821, opened 2026-01-07
docs(structured-properties): clarify breaking schema changes behavior

jewright:
Expands the documentation for breaking schema changes on Structured Properties to better explain what happens to existing values.

- Clarifies that existing values remain retrievable via GraphQL and the API
- Clarifies that existing values are excluded from search results, filters, and aggregations
- Adds guidance on when to use versioning vs creating a new property
- Adds a "When would you use this?" section with common scenarios

### DIFF
--- a/docs/api/tutorials/structured-properties.md
+++ b/docs/api/tutorials/structured-properties.md
@@ -1563,7 +1563,7 @@ ENABLE_STRUCTURED_PROPERTIES_SYSTEM_UPDATE=true
 
 This section demonstrates how to make backwards incompatible schema changes to a Structured Property definition.
 
-Breaking schema changes require setting a `version` string in the Structured Property definition. 
+Breaking schema changes require setting a `version` string in the Structured Property definition.
 
 The format `yyyyMMddhhmmss` (for example, `20240614080000`) is a recommended convention, but not a strict requirement. Semantic versioning or other formats also work. Note that only the most recently created version is active.
 


docs/api/tutorials/structured-properties.md
```diff
@@ -1561,18 +1561,26 @@ ENABLE_STRUCTURED_PROPERTIES_SYSTEM_UPDATE=true
 
 ## Update Structured Property With Breaking Schema Changes
 
-This section will demonstrate how to make backwards incompatible schema changes. Making backwards incompatible
-schema changes will remove previously written data.
+This section demonstrates how to make backwards incompatible schema changes to a Structured Property definition.
 
-Breaking schema changes are implemented by setting a version string within the Structured Property definition. This
-version must be in the following format: `yyyyMMddhhmmss`, i.e. `20240614080000`
+Breaking schema changes require setting a `version` string in the Structured Property definition. The version must follow this format: `yyyyMMddhhmmss` (for example, `20240614080000`).
```
david-leifker: Not strictly true, this is convention over a requirement. I made this up, however even semantic versioning can work. There is no implicit ordering, there can only be one active version (last one created).

docs/api/tutorials/structured-properties.md
```diff
@@ -1561,18 +1561,26 @@ ENABLE_STRUCTURED_PROPERTIES_SYSTEM_UPDATE=true
...
 
-:::note IMPORTANT NOTES
-Old values will not be retrieve-able after the new Structured Property definition is applied.
+:::caution IMPORTANT
+
+Making a breaking schema change will **invalidate existing values** for this Structured Property in search.
+
+- Existing values will still be retrievable via GraphQL and the API
```
david-leifker: In the UI is a side-effect of GraphQL returning them.

david-leifker (approved):

PR #15823, opened 2026-01-07
fix(system-update): fix regressions for postgres, cdc in sql setup steps

chakru-r:
DDLs for postgres are transactional and the cdc user creation was getting rolled back because of the missing commit. This worked on mysql where ddls are implicit commit even if connection has autoCommit=false.

There still is a follow on PR required to remove the mysql-setup/postgres-setup containers since system update does that work. Disabled CDC setup in the setup containers to favor the one in system update.

### DIFF
--- a/datahub-upgrade/src/test/java/com/linkedin/datahub/upgrade/sqlsetup/CreateCdcUserStepTest.java
+++ b/datahub-upgrade/src/test/java/com/linkedin/datahub/upgrade/sqlsetup/CreateCdcUserStepTest.java
@@ -482,6 +482,8 @@ public class CreateCdcUserStepTest {
     verify(mockDatabase).dataSource();
     verify(mockConnection, times(6)).prepareStatement(anyString());
     verify(mockPreparedStatement, times(6)).executeUpdate();
+    // Verify commit is called to persist the transaction
+    verify(mockConnection).commit();
   }
 
   @Test
--- a/datahub-upgrade/src/test/java/com/linkedin/datahub/upgrade/sqlsetup/DatabaseOperationsTest.java
+++ b/datahub-upgrade/src/test/java/com/linkedin/datahub/upgrade/sqlsetup/DatabaseOperationsTest.java
@@ -85,7 +85,8 @@ public class DatabaseOperationsTest {
   public void testPostgresCreateCdcUserSql() {
     String result = postgresOps.createCdcUserSql("cdcuser", "cdcpass");
     assertTrue(result.contains("CREATE USER \"cdcuser\""));
-    assertTrue(result.contains("ALTER USER \"cdcuser\" WITH REPLICATION"));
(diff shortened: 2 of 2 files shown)


docker/quickstart/docker-compose.quickstart-profile.yml
```diff
@@ -49,7 +49,7 @@ services:
       EBEAN_DATASOURCE_HOST: mysql:3306
       EBEAN_DATASOURCE_PASSWORD: datahub
       EBEAN_DATASOURCE_URL: jdbc:mysql://mysql:3306/datahub?verifyServerCertificate=false&useSSL=true&useUnicode=yes&characterEncoding=UTF-8&enabledTLSProtocols=TLSv1.2
-      EBEAN_DATASOURCE_USERNAME: datahub
+      EBEAN_DATASOURCE_USERNAME: root
```
RyanHolstien, line 52: This seems off, why are we switching to root here?

docker/profiles/docker-compose.prerequisites.yml
```diff
@@ -193,7 +193,7 @@ services:
   mysql-setup-cdc:
     <<: *mysql-setup-common
     environment: &mysql-setup-env-cdc
-      CDC_MCL_PROCESSING_ENABLED: ${CDC_MCL_PROCESSING_ENABLED:-true}
+      CDC_MCL_PROCESSING_ENABLED: ${CDC_MCL_PROCESSING_ENABLED:-false}
```
RyanHolstien, line 196: Seems counterintuitive that this would be disabled when running under the cdc profile?

docker/profiles/docker-compose.gms.yml
```diff
@@ -2,11 +2,15 @@
 # Common Environment Variables
 #################################
 x-primary-datastore-mysql-env: &primary-datastore-mysql-env
+  EBEAN_DATASOURCE_USERNAME: ${DATAHUB_EBEAN_DATASOURCE_ROOT_USERNAME:-root}
```
RyanHolstien, line 5: Why root for mysql, but datahub for postgres?

docker/profiles/docker-compose.prerequisites.yml
chakru-r: This is due becase mysql|postgres-setup containers that are now no longer required since this functionality moved to system-update (which does have CDC_MCL_PROCESSING_ENABLED to true.
As I mentioned, there still is some work to remove the mysql|postgresl setup containers completely as well.

docker/profiles/docker-compose.gms.yml
chakru-r: mysql has been using root as superuser all throughout and postgres using datahub. As part of move to system update, root ended up getting used for both mysql and postgres for the system update container (but postgres db container is still initialised with datahub as the user and this is now  fixing this by moving them to per-db type env. This has been one of the causes of the nightly test failures that was testing other profiles.

PR #15824, opened 2026-01-07
feat(databricks): add support for OAuth and unified auth

ligfx:
Adds support for authenticating to Databricks via [service principal OAuth client IDs/secrets](https://docs.databricks.com/aws/en/dev-tools/auth/oauth-m2m#step-2-use-oauth-authorization). Also adds support for connecting via [unified auth](https://docs.databricks.com/aws/en/dev-tools/auth/unified-auth) (e.g. via a Databricks CLI login or environment variables), and fixes the Azure auth implementation which was broken for anything that needed to run SQL (like profiling, or getting lineage from system tables).

### DIFF
--- a/docs/how/updating-datahub.md
+++ b/docs/how/updating-datahub.md
@@ -56,6 +56,7 @@ This file documents any backwards-incompatible changes in DataHub and assists pe
 - #15714: Kafka topic partition counts can now automatically be increased during upgrades if configured values exceed existing partition counts. Set `DATAHUB_AUTO_INCREASE_PARTITIONS=true` to enable.
 - (CLI) Added `--extra-env` option to `datahub ingest deploy` command to pass environment variables as comma-separated KEY=VALUE pairs (e.g., `--extra-env "VAR1=value1,VAR2=value2"`). These are stored in the ingestion source configuration and made available to the executor at runtime.
 - #14968: Added an ingestion source for IBM Db2 databases.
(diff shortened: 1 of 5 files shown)


metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
```diff
@@ -1295,10 +1266,8 @@ def _execute_sql_query(self, query: str, params: Sequence[Any] = ()) -> List[Row
             return []
 
         # Log connection parameters (with masked token)
-        masked_params = {**self._sql_connection_params}
-        if "access_token" in masked_params:
-            masked_params["access_token"] = "***MASKED***"
-        logger.debug(f"Using connection parameters: {masked_params}")
+        sql_connection_params = get_sql_connection_params(self._workspace_client)
+        logger.debug(f"Using connection parameters: {sql_connection_params}")
```
askumar27, line 1270: This might leak the credential in the logs ? we were masking the access token explicitly before.

ligfx: Not a concern anymore, credentials are no longer passed directly in the SQL connection params. See the definition of `get_sql_connection_params` — it returns among other things a credentials provider function which is called on the fly to get credentials as needed. When logged it just prints the function name (actually an anonymous lambda in the current implementation).

PR #15828, opened 2026-01-08
feat(tableau): implement exponential backoff for retry logic on InternalServerError

NehaGslab:
…nalServerError

### DIFF
--- a/metadata-ingestion/tests/integration/tableau/test_tableau_ingest.py
+++ b/metadata-ingestion/tests/integration/tableau/test_tableau_ingest.py
@@ -1148,9 +1148,11 @@ def test_retry_on_error(pytestconfig, tmp_path, mock_datahub_graph):
             mock_sdk.return_value = mock_client
 
             reporter = TableauSourceReport()
+            mock_config = mock.MagicMock()
+            mock_config.max_retries = 3  # Set max_retries for backoff calculation
             tableau_source = TableauSiteSource(
                 platform="tableau",
-                config=mock.MagicMock(),
+                config=mock_config,
                 ctx=mock.MagicMock(),
                 site=mock.MagicMock(spec=SiteItem, id="Site1", content_url="site1"),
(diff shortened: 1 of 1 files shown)


metadata-ingestion/src/datahub/ingestion/source/tableau/tableau.py
```diff
@@ -1496,7 +1496,17 @@ def get_connection_object_page(
             if ise.code in RETRIABLE_ERROR_CODES:
                 if retries_remaining <= 0:
                     raise ise
-                logger.info(f"Retrying query due to error {ise.code}")
+
+                # Add exponential backoff to avoid hammering the server
+                backoff_time = min(
```
sgomezvillamor, line 1511: we have retries in:

```
        except REAUTHENTICATE_ERRORS as e:
        except InternalServerError as ise:
        except OSError:
```

however backoff is only applied in:

```
        except InternalServerError as ise:
```

is there any reason we should not add the backoff in all cases? if so, we should left a comment with the rationale for that

NehaGslab: Added it for REAUTHENTICATE_ERRORS and OSError as well.

Dheerajandsoni: @sgomezvillamor is this fix deployed? if yes, in which latest stable version, I will get the fix?

sgomezvillamor: This is a fix in CLI, so it does not require _backend deployment_.

The PR was merged 2 weeks ago, ~~so still not included in `acryl-datahub` releases~~  https://pypi.org/project/acryl-datahub/#history so it should be included in eg `1.3.1.7`
Of course, if you use a dev datahub cli built from master, it will include the fix.

PR #15845, opened 2026-01-10
fix(structured-properties): allow updating properties that already have badge enabled

pedro93:
## Summary

Fixes a bug where updating any setting on a structured property that already has `showAsAssetBadge=true` would fail with:

```
Cannot have more than one property set with show as badge. Property urns currently set: [urn:li:structuredProperty:product_type]
```

The error occurs even though the property being updated IS the one with the badge - no new badge is being added.

### Root Cause

`ShowPropertyAsBadgeValidator` searches for all properties with `showAsAssetBadge=true`, but never excludes the current entity being updated from the conflict check. So when updating property X which already has the badge, it finds X in the search results and incorrectly treats it as a conflict.

### Fix

Filter out the current entity URN from the search results before checking for badge conflicts. Validation now only fails when a *different* property already has the badge enabled.

### Additional Changes

Added PR guidelines section to `CLAUDE.md` to ensure Claude Code follows the project's PR template when creating pull requests.

## Related Issues

- Fixes CUS-6349
- Related to CUS-7363 (same underlying bug)

## Test Plan

- [x] Added unit test `testValidUpsertWhenUpdatingSamePropertyWithBadge()` that verifies updating a property that already has the badge succeeds
- [x] Existing tests continue to pass (4 tests total)
- [ ] Manual verification: Update any setting on a structured property that has `showAsAssetBadge=true` enabled

## Checklist
- [x] The PR conforms to DataHub's [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (especially [PR Title Format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#pr-title-format))

### DIFF
--- a/metadata-io/src/main/java/com/linkedin/metadata/structuredproperties/validation/ShowPropertyAsBadgeValidator.java
+++ b/metadata-io/src/main/java/com/linkedin/metadata/structuredproperties/validation/ShowPropertyAsBadgeValidator.java
@@ -87,41 +87,43 @@ public class ShowPropertyAsBadgeValidator extends AspectPayloadValidator {
                 .build();
         // Only need to get first set, if there are more then will have to resolve bad state
         ScrollResult scrollResult = scrollIterator.next();
-        // Filter out the current entity being updated - it's allowed to keep its own badge
-        List<SearchEntity> otherBadgeEntities =
-            scrollResult.getEntities().stream()
-                .filter(entity -> !entity.getEntity().equals(mcpItem.getUrn()))
-                .collect(Collectors.toList());
-        if (CollectionUtils.isNotEmpty(otherBadgeEntities)) {
-          if (otherBadgeEntities.size() > 1) {
-            // If it's greater than one, don't bother querying DB since we for sure are in a bad
-            // state
-            exceptions.addException(
-                mcpItem,
-                StructuredPropertyUtils.ONLY_ONE_BADGE
-                    + otherBadgeEntities.stream()
-                        .map(SearchEntity::getEntity)
-                        .collect(Collectors.toList()));
-          } else {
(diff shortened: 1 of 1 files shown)


metadata-io/src/main/java/com/linkedin/metadata/structuredproperties/validation/ShowPropertyAsBadgeValidator.java
```diff
@@ -86,14 +87,19 @@ public static Stream<AspectValidationException> validateSettingsUpserts(
                 .build();
         // Only need to get first set, if there are more then will have to resolve bad state
         ScrollResult scrollResult = scrollIterator.next();
-        if (CollectionUtils.isNotEmpty(scrollResult.getEntities())) {
```
RyanHolstien, line 89: By shifting this you are removing a null check, are you sure this can never be null? Presumably it existed for a reason

metadata-io/src/main/java/com/linkedin/metadata/structuredproperties/validation/ShowPropertyAsBadgeValidator.java
```diff
@@ -86,14 +87,19 @@ public static Stream<AspectValidationException> validateSettingsUpserts(
...
-          if (scrollResult.getEntities().size() > 1) {
+        // Filter out the current entity being updated - it's allowed to keep its own badge
+        List<SearchEntity> otherBadgeEntities =
+            scrollResult.getEntities().stream()
```
RyanHolstien: Additional filtering should occur after verifying that the initial list is not empty or null, no?

PR #15848, opened 2026-01-11
Adding mTLS/SSL support to Rest.li clients and HTTP client configuration

relaxedboi:


david-leifker:
https://github.com/datahub-project/datahub/pull/16831 - Should have addressed this fix

PR #15855, opened 2026-01-12
feat(dbt): add support for ingesting dbt exposures

alfiyas-datahub:
This change adds support for dbt exposures, which represent downstream consumers of dbt models like dashboards, notebooks, ML models, and applications.

Changes:
- Add DBTExposure dataclass to represent exposure metadata
- Add extract_dbt_exposures() to parse exposures from manifest.json
- Add create_exposure_mcps() to generate Dashboard MCPs with lineage
- Add exposures config option in DBTEntitiesEnabled
- Add GraphQL fields for dbt Cloud exposure ingestion
- Add unit tests for exposure extraction and configuration

Exposures are mapped to Dashboard entities with:
- DashboardInfoClass with datasets field for lineage
- Ownership from exposure owner email/name
- Tags with configurable prefix
- SubTypes based on exposure type (Dashboard, Notebook, ML Model, etc.)
- Custom properties including maturity, meta fields

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
@@ -1048,19 +1048,11 @@ class DBTExposure:
     def get_urn(
         self,
         platform_instance: Optional[str],
-        env: str,
     ) -> str:
         """Generate a Dashboard URN for this exposure."""
-        # DashboardUrn has no env field; encode env in the id when not default so
-        # the same exposure in different environments gets a distinct URN.
-        name_for_urn = (
-            f"{env}.{self.unique_id}"
-            if env and env != mce_builder.DEFAULT_ENV
-            else self.unique_id
-        )
         return DashboardUrn.create_from_ids(
             platform=DBT_PLATFORM,
-            name=name_for_urn,
+            name=self.unique_id,
             platform_instance=platform_instance,
         ).urn()
 
@@ -1471,7 +1463,6 @@ class DBTSourceBase(StatefulIngestionSourceBase):
         for exposure in sorted(exposures, key=lambda e: e.unique_id):
             exposure_urn = exposure.get_urn(
                 platform_instance=self.config.platform_instance,
-                env=self.config.env,
             )
 
             # Platform instance aspect
--- a/metadata-ingestion/tests/unit/dbt/test_dbt_source.py
+++ b/metadata-ingestion/tests/unit/dbt/test_dbt_source.py
@@ -1889,7 +1889,7 @@ def test_dbt_exposure_get_urn():
         type="dashboard",
     )
 
-    urn = exposure.get_urn(platform_instance=None, env="PROD")
+    urn = exposure.get_urn(platform_instance=None)
     assert urn == "urn:li:dashboard:(dbt,exposure.my_project.weekly_dashboard)"
 
 
@@ -1900,7 +1900,7 @@ def test_dbt_exposure_get_urn_with_platform_instance():
         type="dashboard",
     )
 
-    urn = exposure.get_urn(platform_instance="my_instance", env="PROD")
+    urn = exposure.get_urn(platform_instance="my_instance")
     assert (
         urn == "urn:li:dashboard:(dbt,my_instance.exposure.my_project.weekly_dashboard)"
     )


CrisNavasM:
Will this be included on the next version? @alfiyas-datahub

metadata-ingestion/docs/sources/dbt/dbt.md
```diff
@@ -418,3 +418,103 @@ If these conditions are not met, warnings will appear in the ingestion report:
...
+dbt exposures are ingested as **Dashboard** entities in DataHub, with the exposure type preserved as a subtype:
+
+| dbt Exposure Type | DataHub SubType |
+| ----------------- | --------------- |
+| `dashboard`       | Dashboard       |
+| `notebook`        | Notebook        |
+| `ml`              | ML Model        |
+| `application`     | Application     |
+| `analysis`        | Analysis        |
+
```
sgomezvillamor, line 465: I’m curious about the rationale behind this mapping.
`Notebook`, `ML Model`, and `Application` are already native entity types. Was modelling those exposure as first-class entities considered instead of representing them as subtypes?
Not asking for any change here — just trying to understand the reasoning behind this decision.

metadata-ingestion/docs/sources/dbt/dbt.md
```diff
@@ -418,3 +418,103 @@ If these conditions are not met, warnings will appear in the ingestion report:
...
+```yaml
+source:
+  type: dbt
+  config:
+    manifest_path: _path_to_manifest_json
+    catalog_path: _path_to_catalog_json
+    target_platform: postgres
+    entities_enabled:
+      exposures: Yes # Default - emit exposures
```
sgomezvillamor, line 478: cool!

metadata-ingestion/docs/sources/dbt/dbt.md
alfiyas-datahub: Though ML models/notebooks may already exist in DataHub from their native sources (MLflow, JupyterHub, etc.), exposures are just "pointers" to them, not the entities themselves. Dashboard was chosen as the appropriate existing entity type since exposures typically represent downstream consumers of dbt models.

treff7es:
@alfiyas-datahub fyi, there are a couple of merge conflicts

metadata-ingestion/docs/sources/dbt/dbt.md
```diff
@@ -436,3 +436,103 @@ If these conditions are not met, warnings will appear in the ingestion report:
...
+#### Owner Resolution
+
+Exposure owners are mapped to DataHub users:
+
+- **`owner.email`** (recommended): Uses the full email address for the user URN (e.g., `analytics@company.com` → `urn:li:corpuser:analytics@company.com`)
+- **`owner.name`** (fallback): Converts name to lowercase with underscores (e.g., `"John Doe"` → `urn:li:corpuser:john_doe`)
+
+:::note
+When only `owner.name` is provided (without email), the generated URN may not match existing users in DataHub. We recommend always providing `owner.email` for accurate user matching.
+:::
```
sgomezvillamor, line 515: is this behaviour on ownership specific to dbt exposures?
* if it is specific: why is it different from other assets?
* if it is the same: why do we emphasize as it was different?

Also we have a config: `enable_owner_extraction` I understand that applies to all dbt assets generated, does it cover dbt exposures too?

metadata-ingestion/docs/sources/dbt/dbt.md
```diff
@@ -436,3 +436,103 @@ If these conditions are not met, warnings will appear in the ingestion report:
...
+    url: https://bi.company.com/dashboards/123
+    depends_on:
+      - ref('orders')
+```
+
+Will create a Dashboard entity in DataHub with:
+
+- **Title**: `weekly_metrics_dashboard`
+- **Description**: `Weekly business metrics for leadership`
+- **SubType**: `Dashboard`
```
sgomezvillamor:
```suggestion
- **SubType**: `Dashboard`, `Notebook`, `ML Model`, `Application`, `Analysis`
```

metadata-ingestion/docs/sources/dbt/dbt.md
```diff
@@ -436,3 +436,103 @@ If these conditions are not met, warnings will appear in the ingestion report:
 > **Note: Limitations**
 >
 > Column-level lineage is currently only supported for Snowflake semantic views, as it relies on parsing the Snowflake-specific DDL.
+
+### Exposures
```
sgomezvillamor, line 440: Just to confirm, dbt Exposures is dbt cloud and oss feature, right?

metadata-ingestion/docs/sources/dbt/dbt.md
```diff
@@ -436,3 +436,103 @@ If these conditions are not met, warnings will appear in the ingestion report:
...
+
+DataHub supports ingesting [dbt exposures](https://docs.getdbt.com/docs/build/exposures) - downstream consumers of your dbt models such as dashboards, notebooks, ML models, and applications.
+
```
sgomezvillamor, line 443: In `dbt-cloud` we have

> Create a [service account token](https://docs.getdbt.com/docs/dbt-cloud-apis/service-tokens) with the "Metadata Only" permission. This is a read-only permission.

Just to confirm: no additional permission is required for exposures?

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_cloud.py
```diff
@@ -329,6 +340,11 @@ def infer_metadata_endpoint(access_url: str) -> Optional[str]:
 class DBTCloudSource(DBTSourceBase, TestableSource):
     config: DBTCloudConfig
     report: DBTSourceReport  # nothing cloud-specific in the report
+    _exposures: List[DBTExposure]
```
sgomezvillamor: If both `DBTCloudSource` and `DBTCoreSource` use `_exposures: List[DBTExposure]` with the same type and semantics, it should be defined once in `DBTSourceBase`.
Any reason for not doing that?

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
```diff
@@ -990,6 +1002,44 @@ def set_columns(self, schema_fields: List[SchemaField]) -> None:
...
+class DBTExposure:
+    """
+    Represents a dbt exposure - a downstream consumer of dbt models.
+    Exposures can be dashboards, notebooks, ML models, applications, or analysis tools.
+    See https://docs.getdbt.com/docs/build/exposures
+    """
+
+    name: str
+    unique_id: str  # e.g., "exposure.my_project.my_dashboard"
+    type: str  # dashboard, notebook, ml, application, analysis
```
sgomezvillamor: The code comment could become actual code by using a literal

```suggestion
    type: Literal["dashboard", "notebook", "ml", "application", "analysis"]
```

are there other types we are not covering?

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
```diff
@@ -990,6 +1002,44 @@ def set_columns(self, schema_fields: List[SchemaField]) -> None:
...
+    owner_name: Optional[str] = None
+    owner_email: Optional[str] = None
+    description: Optional[str] = None
+    url: Optional[str] = None
+    maturity: Optional[str] = None  # high, medium, low
```
sgomezvillamor: maybe literal too?

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
```diff
@@ -990,6 +1002,44 @@ def set_columns(self, schema_fields: List[SchemaField]) -> None:
...
+        platform_instance: Optional[str],
+        env: str,
+    ) -> str:
+        """Generate a Dashboard URN for this exposure."""
+        # DashboardUrn.create_from_ids handles platform_instance prefix automatically
+        return DashboardUrn.create_from_ids(
+            platform=DBT_PLATFORM,
+            name=self.unique_id,
+            platform_instance=platform_instance,
+        ).urn()
```
sgomezvillamor, line 1057: why not including the `env`?

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_cloud.py
```diff
@@ -329,6 +340,11 @@ def infer_metadata_endpoint(access_url: str) -> Optional[str]:
 class DBTCloudSource(DBTSourceBase, TestableSource):
     config: DBTCloudConfig
     report: DBTSourceReport  # nothing cloud-specific in the report
```
sgomezvillamor, line 344: should we include in the `DBTSourceReport` some metric about the new entity? eg num of exposures, even num of exposures by subtype, 
can you check how are we doing for others?

metadata-ingestion/docs/sources/dbt/dbt.md
alfiyas-datahub: It’s the same behaviour as for other dbt assets, not exposure-specific: enable_owner_extraction applies to all of them, including exposures. The doc made it sound exposure-specific, so I’ve updated it to say the behaviour and config are shared and that exposures are included.

metadata-ingestion/docs/sources/dbt/dbt.md
alfiyas-datahub: Yes. Exposures are supported in both dbt Core (OSS) and dbt Cloud. In OSS you define them in the project (e.g. in exposures.yml) and they show up in manifest.json; dbt Cloud also exposes them via its API.

metadata-ingestion/docs/sources/dbt/dbt.md
alfiyas-datahub: Yes. Exposures are metadata (name, type, owner, depends_on, etc.), and we only read them via the dbt Cloud API. The “Metadata only” token is read-only and is what we use for that. So no extra permission should be needed for exposures; the same token used for models/sources is enough.

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_cloud.py
alfiyas-datahub: I’ve refactored so _exposures is defined and initialized once on DBTSourceBase.

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_cloud.py
alfiyas-datahub: Updated reporting exposures the same way we do for other entity types: the report has num_exposures_emitted (total) and num_exposures_by_type (counts by subtype: dashboard, notebook, ml, application, analysis)

metadata-ingestion/docs/sources/dbt/dbt.md
sgomezvillamor: Thanks for making this explicitly generic.
Aligned with that, I would move the `Owner Resolution` section out of the `Exposures` section.

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
```diff
@@ -306,6 +309,10 @@ class DBTSourceReport(StaleEntityRemovalSourceReport):
     queries_failed_list: LossyList[str] = field(default_factory=LossyList)
     query_timestamps_fallback_used: bool = False
 
+    # Exposure entity emission statistics
+    num_exposures_emitted: int = 0
+    num_exposures_by_type: Dict[str, int] = field(default_factory=dict)
```
sgomezvillamor: have you tried something like:

```
    num_exposures_by_type: Dict[str, int] = field(default_factory=lambda: defaultdict(int))
```

so you can do:

```
                for e in exposures:
                    self.report.num_exposures_by_type[e.type] += 1
```
?

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
```diff
@@ -990,6 +1007,61 @@ def set_columns(self, schema_fields: List[SchemaField]) -> None:
...
+    dbt_file_path: Optional[str] = None
+
+    def get_urn(
+        self,
+        platform_instance: Optional[str],
+        env: str,
+    ) -> str:
+        """Generate a Dashboard URN for this exposure."""
+        # DashboardUrn has no env field; encode env in the id when not default so
+        # the same exposure in different environments gets a distinct URN.
```
sgomezvillamor: is concatenating env and name standard approach for other sources?

alfiyas-datahub: - Tableau, PowerBI, Superset, Metabase, Qlik, Sigma, Grafana: None of them encode env in the dashboard name. They rely solely on platform_instance for environment differentiation.
- Standard approach: Use platform_instance only → users must set different platform_instance for different environments.
- Current approach: Encode env when non-default → works for users who rely on env config.

Happy to change if you feel strongly about aligning with other sources. We could document that users should use different platform_instance values for different environments instead.

sgomezvillamor: Let's align with other sources.
If we ever need to fix this, we will fix for all sources at once.

PR #15859, opened 2026-01-12
fix(ingest/db2): the ibm_db package is not installable on Linux ARM

ligfx:


metadata-ingestion/setup.py
sgomezvillamor, line 1: could you please also add `db2` to the `all_exclude_plugins` list with a comment explaining the exclusion?

sgomezvillamor:
We may also add a note in `/metadata-ingestion/docs/sources/db2/db2_pre.md`

ligfx:
> We may also add a note in `/metadata-ingestion/docs/sources/db2/db2_pre.md`

Good catch, added a note about architecture support in the docs.

> could you please also add db2 to the all_exclude_plugins list with a comment explaining the exclusion?

I don't understand the rationale for exclusion, so I'll leave that for another PR by someone who does. This PR is focused on fixing the issue when trying to install Db2 on Linux ARM.

PR #15860, opened 2026-01-12
fix(ingestion): Use separate conditional for MCE processor

redblackcoder:
When MCP batch processing is enabled through `metadataChangeProposal.consumer.batch.enabled`, it unexpectantly disables the bean MCE processor which results in any MCE messages not getting processed.

This change introduce a separate conditional class for MCE processor so that batch processing for MCP can be enabled without impacting MCEs.


deepgarg760:
@redblackcoder , check for env `MCE_CONSUMER_ENABLED` should be removed from others conditions then

redblackcoder:
> @redblackcoder , check for env `MCE_CONSUMER_ENABLED` should be removed from others conditions then

@deepgarg760 I left that varible in the other condition for ensure no breaking change. Env var `MCE_CONSUMER_ENABLED` is sufficient to enable both MCE and MCP consumers, as the code is currently. Removing it from MCP conditional would break that assumption, and may stop MCP processing for some setups.

RyanHolstien:
While this is fine, it's concerning to me that this is an issue for you. MCE has been an unused event for years at this point and should not be still in use, should be considered deprecated. It can be removed from the product at any moment.

redblackcoder:
> While this is fine, it's concerning to me that this is an issue for you. MCE has been an unused event for years at this point and should not be still in use, should be considered deprecated. It can be removed from the product at any moment.

We are using hive ingestion to capture hive table schemas, it emits MCEs and was broken when I enabled batch processing - https://github.com/datahub-project/datahub/blob/master/metadata-ingestion/src/datahub/ingestion/source/sql/hive/hive_metadata_processor.py#L533
This should migrate to emit MCPs instead.

deepgarg760:
> > While this is fine, it's concerning to me that this is an issue for you. MCE has been an unused event for years at this point and should not be still in use, should be considered deprecated. It can be removed from the product at any moment.
> 
> We are using hive ingestion to capture hive table schemas, it emits MCEs and was broken when I enabled batch processing - https://github.com/datahub-project/datahub/blob/master/metadata-ingestion/src/datahub/ingestion/source/sql/hive/hive_metadata_processor.py#L533 This should migrate to emit MCPs instead.

can you share the recipe being used for hive ingestion

PR #15862, opened 2026-01-12
feat(ingest): Retry more often on 429 for rate limiting

asikowitz:
To prevent failures when gms is rate limiting ingestion, support more retries when receiving a 429. I didn't want to overwrite `total` so instead I keep track of `_429_count` and only truly decrement `total` some of the time.


david-leifker (approved):

askumar27 (approved):
[nit] Bart complained missing protection and suggested adding checks for `DATAHUB_REST_EMITTER_429_RETRY_MULTIPLIER=0` avoid code crashing with `ZeroDivisionError`

PR #15864, opened 2026-01-13
fix(dagster): derive Dagster job URL from location_name

aviraj-gour:
Linear Ticket: ING-1343
Dagster job URLs usually follow this structure: http://dagster-ui-base-url/locations/<code_location_name>/jobs/<job_name>

## Problem
Currently, get_dagster_environment resolves the Dagster code location from the Python module path obtained via the job’s code_pointer. However, the module path represents the package structure and does not always match the actual Dagster location_name.

In Dagster, the location_name is user-defined and configurable (e.g. via workspace.yaml), so deriving it from the module can result in incorrect URLs (such as appending .definitions).

## Solution

Instead of resolving the code location from the module path, this change derives it directly from the Dagster runtime metadata:
`context.dagster_run.remote_job_origin.location_name`

This value represents the actual location name used by Dagster UI

## Debugging / Validation

To verify the issue, I added debug logging inside get_dagster_environment.
Screenshot 2026-01-13 at 9 17 56 AM

Deployed dagster with correct workspace.yaml:
Screenshot 2026-01-13 at 9 19 32 AM

Check UI url and module name are different:
Screenshot 2026-01-13 at 9 20 00 AM

### Output of print:
`code_pointer: ModuleCodePointer(module='project.definitions', fn_name='defs', working_directory='/Users/avi/Documents/project')`

`location_name: my_custom_location_name`

### DIFF
--- a/.github/workflows/metadata-ingestion.yml
+++ b/.github/workflows/metadata-ingestion.yml
@@ -47,6 +47,7 @@ jobs:
             "testIntegrationBatch3",
             "testIntegrationBatch4",
             "testIntegrationBatch5",
+            "testIntegrationBatchRecording", # Recording tests - isolated execution to avoid module patching interference
           ]
         include:
           # Version compatibility tests.
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -205,6 +205,112 @@ failure_log:
     filename: ./path/to/failure.json
 ```
 
+#### ingest --record (Beta)
+
+:::note Beta Feature
+Recording and replay is currently in beta. The feature is stable for debugging purposes but the archive format may change in future releases.
+:::
+
+The `--record` option enables recording of all HTTP requests and database queries during ingestion. This creates an encrypted archive that can be replayed offline for debugging.
+
+```shell
+# Record an ingestion run with password protection
+datahub ingest -c ./recipe.yaml --record --record-password mysecret
+
+# Record to a specific local directory
+export INGESTION_ARTIFACT_DIR=/path/to/recordings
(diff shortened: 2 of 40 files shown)


metadata-ingestion-modules/dagster-plugin/src/datahub_dagster_plugin/client/dagster_generator.py
```diff
@@ -253,6 +253,7 @@ class DagsterEnvironment:
     is_branch_deployment: bool = False
     branch: Optional[str] = "prod"
     module: Optional[str] = None
+    location_name: Optional[str] = None
 
 
 def job_url_generator(dagster_url: str, dagster_environment: DagsterEnvironment) -> str:
```
treff7es, line 259: No test coverage for job_url_generator

aviraj-gour: @treff7es I didn’t add tests because the Dagster plugin currently doesn’t have unit tests for most of its code, and this was just a small fix. 
That said, if you’d like, I can add a unit test specifically for job_url_generator (please [check](https://github.com/datahub-project/datahub/tree/master/metadata-ingestion-modules/dagster-plugin/tests)).

treff7es: please, add unit tests if possible to make sure we can add tests later

aviraj-gour: Added test for job_url_generator.

PR #15867, opened 2026-01-13
fix(openlineage): Handle null fields in Trino OpenLineage COMPLETE events

alfiyas-datahub:
## Summary

Fixes HTTP 500 errors when DataHub GMS processes Trino OpenLineage COMPLETE events that contain `SchemaDatasetFacet` or `ColumnLineageDatasetFacet` with null fields.

## Problem

When Trino sends OpenLineage COMPLETE events to DataHub:
- `SchemaDatasetFacet.getFields()` can return `null` (valid per OpenLineage spec)
- `ColumnLineageDatasetFacet.getFields()` can return `null` (valid per OpenLineage spec)
- DataHub's `OpenLineageToDataHub` converter called methods on these null values, causing `NullPointerException`
- The NPE bubbled up as HTTP 500 errors to Trino clients

Additionally, the Trino producer URL (`https://github.com/trinodb/trino`) was not recognized, preventing orchestrator detection.

### DIFF
--- a/metadata-integration/java/openlineage-converter/src/main/java/io/datahubproject/openlineage/converter/OpenLineageToDataHub.java
+++ b/metadata-integration/java/openlineage-converter/src/main/java/io/datahubproject/openlineage/converter/OpenLineageToDataHub.java
@@ -420,9 +420,15 @@ public class OpenLineageToDataHub {
     }
 
     OpenLineage.ColumnLineageDatasetFacet columnLineage = dataset.getFacets().getColumnLineage();
+    // Per OpenLineage spec
+    // (https://openlineage.io/spec/facets/1-2-0/ColumnLineageDatasetFacet.json),
+    // "fields" is a required property. However, the OpenLineage Java client may return null
+    // when the fields object is empty {} or when producers (like Trino) omit it.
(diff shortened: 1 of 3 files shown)


sgomezvillamor:
> * `SchemaDatasetFacet.getFields()` can return `null` (valid per OpenLineage spec)

Is this the latest spec? https://openlineage.io/spec/facets/1-2-0/SchemaDatasetFacet.json
In that case, "fields" is not required.
We may add the link as doc

> * `ColumnLineageDatasetFacet.getFields()` can return `null` (valid per OpenLineage spec)

However here https://openlineage.io/spec/facets/1-2-0/ColumnLineageDatasetFacet.json, "fields" is a required field

>> The "required": ["fields"] array explicitly lists fields as a required property, meaning any valid instance of ColumnLineageDatasetFacet must include the fields property.
>> Note that:
>>
>> fields must be present
>> fields must be an object (not null, not an array)
>> fields can be an empty object {} if there are no column-level lineages to report

Maybe the empty object case results in `ColumnLineageDatasetFacet.getFields() == null`?

Let's double check and add some comment in the code with the conclusion and references.

Also, it would be nice to add some test, if possible.

PR #15875, opened 2026-01-14
fix(airflow): emit DataFlow from task handler in Airflow 3 distributed deployments

treff7es:
## Summary
- Fixes issue where pipelines (DataFlow) are not emitted in distributed Airflow 3 deployments (Kubernetes, Astronomer)
- Tasks were emitted with a pointer to the pipeline, but the pipeline itself was not emitted, causing 404 errors in the UI

## Root Cause
In distributed Airflow 3 deployments:
- The `on_dag_run_running` hook runs on the **scheduler process**
- The DataHub listener is only initialized on **worker processes** (lazy initialization requires task execution context)
- The scheduler's `on_dag_run_running` never triggers DataFlow emission because the listener isn't initialized there
- Connection.get() in Airflow 3 requires the task execution context (SUPERVISOR_COMMS channel), which only exists on workers during task execution. The scheduler doesn't have this context, so connection retrieval fails and _get_emitter() returns None -> In long term we should move to Airflow config based DataHub emitter setup

## Fix
Always emit DataFlow from the task instance handler as a fallback. DataHub's UPSERT semantics make duplicate emissions safe and idempotent.

## Testing
- Unit tests pass (10 passed, 15 skipped)
- Lint passes

## Checklist

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/metadata-ingestion-modules/airflow-plugin/src/datahub_airflow_plugin/airflow3/datahub_listener.py
+++ b/metadata-ingestion-modules/airflow-plugin/src/datahub_airflow_plugin/airflow3/datahub_listener.py
@@ -1059,9 +1059,13 @@ class DataHubListener:
 
         # Airflow 3.0+ doesn't need task holder
 
-        # If we don't have the DAG listener API, emit DAG start event
-        if not HAS_AIRFLOW_DAG_LISTENER_API:
-            self.on_dag_start(dagrun)
+        # Always emit DataFlow from task instance handler as a fallback.
+        # In distributed Airflow deployments (Kubernetes, Astronomer), the on_dag_run_running
+        # hook runs on the scheduler process, but the DataHub listener is only initialized
+        # on worker processes. This means the scheduler's on_dag_run_running never triggers
+        # DataFlow emission. By emitting here, we ensure the DataFlow exists before the
+        # DataJob references it. DataHub's UPSERT semantics make duplicate emissions safe.
+        self.on_dag_start(dagrun)
 
         # Generate and emit datajob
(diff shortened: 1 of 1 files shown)


krupakar010:
Hi @treff7es, I understand this emits the dag run start events for every task run event , do you consider moving this to airflow schedulers instead of doing the emit from workers ?

treff7es:
> Hi @treff7es, I understand this emits the dag run start events for every task run event , do you consider moving this to airflow schedulers instead of doing the emit from workers ?

yes, but to do that we need to get rid of our current Connection based config. In Airflow 3.x the Connection is not available on the Scheduler where these events get triggered :(

Connection.get() in Airflow 3 requires the task execution context (SUPERVISOR_COMMS channel), which only exists on workers during task execution. The scheduler doesn't have this context, so connection retrieval fails and _get_emitter() returns None.

sgomezvillamor (approved):
If this is a limitation "In distributed Airflow 3 deployments", would it make sense to identify that scenario somehow and so keep the emission of the dag behind an `if` condition?

treff7es:
> If this is a limitation "In distributed Airflow 3 deployments", would it make sense to identify that scenario somehow and so keep the emission of the dag behind an `if` condition?

I would keep as it is for now because:
  1. Edge cases are tricky - Even LocalExecutor in Airflow 3 can behave differently depending on deployment
  2. UPSERT is cheap - DataHub UPSERT is idempotent and fast; the overhead is minimal
  
The real solution here to get rid or at least to support setting up emitter based on Airflow Config

PR #15877, opened 2026-01-14
feat(ldap): add tls_verify config option to enable TLS certificate validation

alfiyas-datahub:
Addresses security vulnerability (CWE-295) where TLS certificate validation was disabled by default, allowing Man-in-the-Middle attacks.

- Add tls_verify config option (default: true)
- When tls_verify=true, enforces certificate validation using OPT_X_TLS_DEMAND
- Add unit tests for the new configuration option

Security Note: It is recommended to set tls_verify: true in production environments to prevent MITM attacks.

Closes: ING-1380


deepgarg760 (approved):

metadata-ingestion/src/datahub/ingestion/source/ldap.py
```diff
@@ -154,6 +157,14 @@ class LDAPSourceConfig(StatefulIngestionConfigBase, DatasetSourceConfigMixin):
         description="Use email for users' usernames instead of username (disabled by default). \
             If enabled, the user and group urn would be having email as the id part of the urn.",
     )
+
+    tls_verify: bool = Field(
+        default=False,
```
sgomezvillamor: Please, double check on final decission whether this should be True by default.
In that case, the breaking change notice needs to be added in the `Breaking changes` section in `docs/how/updating-datahub.md`.

sgomezvillamor (approved):

docs/how/updating-datahub.md
```diff
@@ -33,6 +33,7 @@ This file documents any backwards-incompatible changes in DataHub and assists pe
 
 ### Breaking Changes
 
+- (LDAP) The LDAP ingestion source now enforces TLS certificate verification by default (`tls_verify: true`). This prevents Man-in-the-Middle attacks (CWE-295) but may break existing configurations using self-signed certificates. To restore the previous behavior, explicitly set `tls_verify: false` in your recipe.
```
```suggestion
- #15877 The LDAP ingestion source now enforces TLS certificate verification by default (`tls_verify: true`). This prevents Man-in-the-Middle attacks (CWE-295) but may break existing configurations using self-signed certificates. To restore the previous behavior, explicitly set `tls_verify: false` in your recipe.
```

PR #15878, opened 2026-01-14
fix(ingestion): Ensure minimal version of pyiceberg dependency

skrydal:


metadata-ingestion/setup.py
```diff
@@ -308,7 +308,9 @@
     # especially for AWS-based catalogs and warehouses, the properties `profile_name`, `region_name`,
     # `aws_access_key_id`, `aws_secret_access_key`, and `aws_session_token` were deprecated and removed in version
     # 0.8.0.
-    "pyiceberg[glue,hive,dynamodb,snappy,hive,s3fs,adlfs,pyarrow,zstandard]>=0.8.0",
+    # - Versions 0.7.0 - 0.8.1 use variable DEPRECATED_BOTOCORE_SESSION instead of BOTOCORE_SESSION, the latter is
+    #   expected by the connector
+    "pyiceberg[glue,hive,dynamodb,snappy,hive,s3fs,adlfs,pyarrow,zstandard]>=0.9.0",
```
kyungsoo-datahub, line 313: Just qq, then, is it better to use > 0.8.1?

skrydal: I don't think we can expect, potential, 0.8.2 to have `BOTOCORE_SESSION`, it seems like a change requiring, at least, minor version increment.

PR #15879, opened 2026-01-15
docs(assertions): Add docs for query attribution

AdrianMachado:
## Summary

Adding docs on Observe query attribution

### DIFF
--- a/docs/managed-datahub/observe/assertion-query-attribution.md
+++ b/docs/managed-datahub/observe/assertion-query-attribution.md
@@ -1,4 +1,12 @@
+---
+description: Learn how to attribute queries to DataHub Cloud Observe to better track costs
+---
+
+import FeatureAvailability from '@site/src/components/FeatureAvailability';
+
 # Assertion Query Attribution
+
+<FeatureAvailability saasOnly />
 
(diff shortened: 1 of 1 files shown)


docs/managed-datahub/observe/assertion-query-attribution.md
```diff
@@ -0,0 +1,37 @@
...
+SELECT *
+FROM users
+```
+
+## Snowflake Query Tag
+
+For queries issued against Snowflake, a [Snowflake Query Tag](https://select.dev/posts/snowflake-query-tags) is added to the SQL statement.
+
+```sql
+ALTER SESSION SET query_tag='{"query_source": "datahub_observe", "assertion_urn": "urn:li:assertion:507e3dec-8fed-4809-9cdd-cf2a4a06a249"}'
```
jayacryl, line 28: Unrelated and scope creep, but I just had a thought, would it be helpful for folks to see assertion_type here? Ie understanding custom SQL vs column_value vs column_metric checks. Curious to hear your thoughts.

AdrianMachado: Its not a hard thing to do - but some platforms have text limits so we can't bloat this infinitely. I think its best to wait for a customer to actually ask for this

PR #15898, opened 2026-01-16
feat(snowflake): added retry logic to snowflake ingestion, to make ingestion more stable

rajatoss:
# Automatic Retry Logic for Snowflake ACCOUNT_USAGE Queries

## Summary
Adds automatic retry logic for Snowflake ingestion when querying ACCOUNT_USAGE system views, resolving intermittent ingestion failures caused by transient unavailability during Snowflake's internal refresh cycles.

Fixes: https://linear.app/acryl-data/issue/ING-1370/snowflake-ingestion-intermittently-fails-due-to-hardcoded-tag

## Problem
DataHub's Snowflake ingestion intermittently fails with authorization errors when querying Snowflake's ACCOUNT_USAGE schema:

**Root cause:** Snowflake's ACCOUNT_USAGE system views (like TAG_REFERENCES, QUERY_HISTORY, etc.) are periodically refreshed by Snowflake and can be temporarily unavailable during these refresh cycles. According to Snowflake support, this is expected behavior and clients should implement retry logic.

These transient failures cause entire ingestion runs to fail, requiring manual reruns and creating operational overhead.

## Solution
Implemented targeted retry logic specifically for ACCOUNT_USAGE queries.

**Exponential backoff:** Retries up to 4 times with increasing delays:
- 1st retry: 20 seconds
- 2nd retry: 40 seconds  
- 3rd retry: 60 seconds

**Total coverage:** ~2 minutes of retry time before failing.

**Safe retry design:** Only retries queries to ACCOUNT_USAGE views (which are read-only), avoiding potential side effects from retrying write operations.


metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_connection.py
```diff
@@ -441,19 +449,49 @@ def get_query_no(self) -> int:
...
+        @retry(
+            retry=retry_if_exception(
+                lambda e: _is_retryable_account_usage_error(e, query)
+            ),
+            stop=stop_after_attempt(3),
+            wait=wait_exponential(multiplier=1, min=2, max=10),
+            before_sleep=before_sleep_log(logger, logging.WARNING),
+            reraise=True,
+        )
+        def _execute():
```
sgomezvillamor: Probably personal taste, but nested method definitions hurt readability.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_connection.py
```diff
@@ -472,3 +510,34 @@ def _is_permission_error(e: Exception) -> bool:
     # 002003 (02000): SQL compilation error: Database/SCHEMA 'XXXX' does not exist or not authorized.
     # Insufficient privileges to operate on database 'XXXX'
     return "Insufficient privileges" in msg or "not authorized" in msg
+
+
+def _is_retryable_account_usage_error(e: BaseException, query: str) -> bool:
```
sgomezvillamor, line 509: Some questions here:

* I was expecting a more generic logic to determine retry, however this is very specific on specific schema AND specific error. Is there any reason why we shouldn't have a more generic retry strategy?
* how this matches with the logic in `_is_permission_error`, which returns true and raises exception if `"not authorized" in msg` ?

rajat-datahub: @sgomezvillamor please check if you are okay with these replies or I should add a general retry.

sgomezvillamor:
> Exponential backoff: Retries up to 3 times with increasing delays:
> 1st retry: 2 seconds
> 2nd retry: 4 seconds
> 3rd retry: 8 seconds

> Snowflake's ACCOUNT_USAGE system views (like TAG_REFERENCES, QUERY_HISTORY, etc.) are periodically refreshed by Snowflake and can be temporarily unavailable during these refresh cycles

Depending on how long the unavailability can last, 14 seconds may not be enough. Should we increase the number of retries or the delay between them?

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_connection.py
sgomezvillamor: I took your 👍 on my comment to mean you were going to address it 😃

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_connection.py
```diff
@@ -441,19 +449,52 @@ def get_query_no(self) -> int:
...
                 logger.info(
                     f"Query #{query_num} got {resp.rowcount} row(s) back from Snowflake",
                     stacklevel=2,
                 )
             return resp
+        except RetryError as retry_err:
+            # If all retries failed, raise the original exception with its original traceback.
+            exc = retry_err.last_attempt.exception()
+            assert exc is not None
+            raise exc  # noqa: B904
```
sgomezvillamor: this may address the `ruff` concern

```suggestion
            raise exc from None
```

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_connection.py
```diff
@@ -441,19 +449,52 @@ def get_query_no(self) -> int:
...
+
+    def query(self, query: str) -> Any:
+        try:
+            # We often run multiple queries in parallel across multiple threads,
+            # so we need to number them to help with log readability.
+            query_num = self.get_query_no()
+            logger.info(f"Query #{query_num}: {query.rstrip()}", stacklevel=2)
```
sgomezvillamor: why the stacklevel 2?

metadata-ingestion/tests/unit/snowflake/test_snowflake_connection_retry.py
```diff
@@ -0,0 +1,252 @@
...
+    def test_is_retryable_account_usage_error_with_different_error_code(self):
+        """Test that ACCOUNT_USAGE queries with different errors are not retryable."""
+        query = "SELECT * FROM snowflake.account_usage.query_history"
+        error = MockSnowflakeError(
+            "001003: SQL compilation error: syntax error line 1 at position 15"
+        )
+
+        result = _is_retryable_account_usage_error(error, query)
+
+        assert result is False
```
sgomezvillamor: The tests around _is_retryable_account_usage_error would be a great fit for parameterized tests, it could reduce duplication and improve readability.

rajatoss:
> > Exponential backoff: Retries up to 3 times with increasing delays:
> > 1st retry: 2 seconds
> > 2nd retry: 4 seconds
> > 3rd retry: 8 seconds
> 
> > Snowflake's ACCOUNT_USAGE system views (like TAG_REFERENCES, QUERY_HISTORY, etc.) are periodically refreshed by Snowflake and can be temporarily unavailable during these refresh cycles
> 
> Depending on how long the unavailability can last, 14 seconds may not be enough. Should we increase the number of retries or the delay between them?

updated to:
Exponential backoff: Retries up to 4 times with increasing delays:

1st retry: 20 seconds
2nd retry: 40 seconds
3rd retry: 60 seconds
Total coverage: ~2 minutes of retry time before failing.

sgomezvillamor:
> 1st retry: 20 seconds
> 2nd retry: 40 seconds
> 3rd retry: 60 seconds
> Total coverage: ~2 minutes of retry time before failing.

2min of total coverage is reasonable, but a 20-second delay before the first retry is a pretty heavy penalty.

Something like `wait_exponential(multiplier=8, min=1, max=60)` would give a sequence of 1, 16, 32, 60.

In general, I’d aim for a moderate multiplier, and if we want broader coverage, increase the number of attempts rather than making early retries so expensive.

rajatoss:
> > 1st retry: 20 seconds
> > 2nd retry: 40 seconds
> > 3rd retry: 60 seconds
> > Total coverage: ~2 minutes of retry time before failing.
> 
> 2min of total coverage is reasonable, but a 20-second delay before the first retry is a pretty heavy penalty.
> 
> Something like `wait_exponential(multiplier=8, min=1, max=60)` would give a sequence of 1, 16, 32, 60.
> 
> In general, I’d aim for a moderate multiplier, and if we want broader coverage, increase the number of attempts rather than making early retries so expensive.

I felt from warehouse token consumption point of view, increasing the number of attempts would lead to more queries being executed on the warehouse. But yes, the initial one should be a little closer to the 10-second mark. I will add this in my other merge request.

PR #15899, opened 2026-01-16
feat(ingest/teradata): Add ownership extraction from table and view creators

aviraj-gour:
Linear Issue: ING-945

Adds ownership extraction capability to the Teradata connector that derives ownership information from table and view creator metadata.

## Changes
- Implemented ownership extraction from Teradata `DBC.TablesV` and `DBC.ViewsV` creator columns
- Added configuration flags to control ownership extraction (disabled by default for safe rollout)
- Added comprehensive tests for ownership extraction and configuration validation

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
@@ -983,7 +983,8 @@ ORDER by DataBaseName, TableName;
 
     def _get_creator_for_entity(self, schema: str, entity_name: str) -> Optional[str]:
         """Get creator name for a table or view."""
-        return self._table_creator_cache.get((schema, entity_name))
+        with self._tables_cache_lock:
+            return self._table_creator_cache.get((schema, entity_name))
 
     def _emit_ownership_if_available(
         self,
@@ -1401,12 +1402,11 @@ ORDER by DataBaseName, TableName;
                         database_counts[table.database]["tables"] += 1
 
                     with self._tables_cache_lock:
-                        if table.database not in self._tables_cache:
-                            self._tables_cache[table.database] = []
                         self._tables_cache[table.database].append(table)
-                        if entry.CreatorName is not None:
+                        creator_name = (entry.CreatorName or "").strip()
+                        if creator_name:
                             self._table_creator_cache[(table.database, table.name)] = (
-                                entry.CreatorName.strip()
+                                creator_name
                             )
 
                 for database, counts in database_counts.items():


metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -956,6 +981,66 @@ def cached_get_table_properties(
                 break
         return description, properties, location
 
+    def _get_creator_for_entity(self, schema: str, entity_name: str) -> Optional[str]:
+        """Get creator name for a table or view."""
+        if (schema, entity_name) in self._table_creator_cache:
+            return self._table_creator_cache[(schema, entity_name)]
+        return None
```
treff7es:
```suggestion
  def _get_creator_for_entity(self, schema: str, entity_name: str) -> Optional[str]:                            
      return self._table_creator_cache.get((schema, entity_name))                                               
```

metadata-ingestion/tests/unit/test_teradata_source.py
```diff
@@ -1627,3 +1688,228 @@ def _create_mock_entry(
...
+            mock_engine.execute.return_value = [mock_entry]
+            mock_get_engine.return_value = mock_engine
+
+            source.cache_tables_and_views()
+
+            assert (
+                source._table_creator_cache[("test_db", "test_table")] == "creator_user"
+            )
+
+    def test_creator_cache_thread_safety(self):
```
treff7es: This tests that a Lock exists, not that thread safety works. It provides no value.

metadata-ingestion/tests/unit/test_teradata_source.py
```diff
@@ -1627,3 +1688,228 @@ def _create_mock_entry(
...
+                process_method(
+                    f"test_db.{entity_name}",
+                    mock_inspector,
+                    "test_db",
+                    entity_name,
+                    mock_sql_config,
+                )
+            )
+
+            assert len(work_units) >= 1
```
treff7es: This doesn't test much, just the existence of the workunit. It doesn't validate the ownership work unit contains correct data

metadata-ingestion/tests/unit/test_teradata_source.py
```diff
@@ -1627,3 +1688,228 @@ def _create_mock_entry(
...
+
+    def test_creator_cache_with_whitespace_trimming(self):
```
treff7es: Test name says "whitespace trimming" but the mock uses creator_name="creator_user" with no whitespace to trim.

metadata-ingestion/tests/unit/test_teradata_source.py
aviraj-gour: removed.

metadata-ingestion/tests/unit/test_teradata_source.py
aviraj-gour: not needed removed

metadata-ingestion/tests/unit/test_teradata_source.py
aviraj-gour: Thanks, now checking work_units aspectName should be ownership

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -1321,6 +1404,10 @@ def cache_tables_and_views(self) -> None:
                         if table.database not in self._tables_cache:
                             self._tables_cache[table.database] = []
                         self._tables_cache[table.database].append(table)
+                        if entry.CreatorName is not None:
```
treff7es: Here, you let an empty string for CreatorName, and even if it is not empty after the `entry.CreatorName.strip()`, it can be an empty string, which can result in an invalid urn.
Can you make sure we don't let an empty name after strip?

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -956,6 +981,64 @@ def cached_get_table_properties(
                 break
         return description, properties, location
 
+    def _get_creator_for_entity(self, schema: str, entity_name: str) -> Optional[str]:
+        """Get creator name for a table or view."""
+        return self._table_creator_cache.get((schema, entity_name))
```
treff7es:
```suggestion
      with self._tables_cache_lock:                                                                             
          return self._table_creator_cache.get((schema, entity_name))                                           ```

File: metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py [hunk 7]
```diff
@@ -1321,6 +1404,10 @@ def cache_tables_and_views(self) -> None:
                         if table.database not in self._tables_cache:
                             self._tables_cache[table.database] = []
                         self._tables_cache[table.database].append(table)
```
treff7es: Since _tables_cache is a defaultdict(list), the existence check is unnecessary.

```suggestion
self._tables_cache[table.database].append(table)
```

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
aviraj-gour: Thanks, fixed now

PR #15916, opened 2026-01-18
fix(ingestion): prevent platform instance duplication in browse paths

sgomezvillamor:
## Problem

When using SDK v2 ContainerKey objects with `parent_container`, the SDK automatically includes the platform instance in the browse path (via `_set_container`). The `auto_browse_path_v2` processor then prepends the platform instance again, causing duplicate platform instance entries in browse paths.

This was discovered when working on the Fabric OneLake connector, which uses SDK v2 Container/Dataset classes with ContainerKey inheritance.

## Solution

Modified `_prepend_platform_instance` to check if the platform instance is already the first entry before prepending it. This prevents duplication while ensuring the platform instance is present when needed.

## Changes

- Modified `_prepend_platform_instance` in `source_helpers.py` to detect existing platform instance entries
- Added comprehensive unit test `test_auto_browse_path_v2_prevents_platform_instance_duplication` covering all scenarios

## Testing

- All existing tests pass (12/12)
- New test verifies:
  - Platform instance already present (no duplication)
  - Platform instance not present (should prepend)
  - Empty path (should add platform instance)
  - No platform/instance (should return unchanged)


sgomezvillamor:
> LGTM, if you revert the change from `source_helpers.py`, does the test fail?

yes

PR #15917, opened 2026-01-18
fix(ingest/sql): Fix test connection with stateful_ingestion enabled

asikowitz:
This previously failed because (i) `pipeline_name` was not set in the `PipelineContext` created for the source and (ii) the pipeline context didn't have a graph. We could make a fake pipeline name but not sure that would solve the graph issue. This seemed like the easiest option.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/sql/sql_common.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/sql/sql_common.py
@@ -396,6 +396,15 @@ class SQLAlchemySource(StatefulIngestionSourceBase, TestableSource):
     def test_connection(cls, config_dict: dict) -> TestConnectionReport:
         test_report = TestConnectionReport()
         try:
+            if config_dict.get("stateful_ingestion", {}).get("enabled", False):
+                # Connection test should only care about config for connecting to the source
+                # Ingestion-only details like whether stateful_ingestion is enabled are irrelevant
+                # And specifically, stateful_ingestion requires a connection to DataHub
+                # which should not be required to test a connection
+                config_dict = {
+                    **config_dict,
+                    "stateful_ingestion": {"enabled": False},
+                }
             source = cast(
                 SQLAlchemySource,
                 cls.create(config_dict, PipelineContext(run_id="test_connection")),
--- a/metadata-ingestion/tests/integration/mysql/test_mysql.py
+++ b/metadata-ingestion/tests/integration/mysql/test_mysql.py
@@ -88,6 +88,7 @@ def test_mysql_ingest_no_db(
                 "database": "northwind",
                 "username": "root",
                 "password": "example",
+                "stateful_ingestion": {"enabled": "true"},
             },
             True,
         ),
--- a/metadata-ingestion/tests/integration/postgres/postgres_test_connection.yml
+++ b/metadata-ingestion/tests/integration/postgres/postgres_test_connection.yml
@@ -1,0 +1,16 @@
+source:
+  type: postgres
+  config:
+    host_port: 'localhost:5432'
+    initial_database: 'postgrestest'
+    username: 'postgres'
+    password: 'example'
+    include_tables: true
+    database: postgres
+    profiling:
+      enabled: true
+      profile_table_level_only: true
+    include_views: true
+    stateful_ingestion:
+      enabled: true
+pipeline_name: 'urn:li:dataHubExecutionRequest:test'
--- a/metadata-ingestion/tests/integration/postgres/test_postgres.py
+++ b/metadata-ingestion/tests/integration/postgres/test_postgres.py
@@ -1,3 +1,4 @@
+import json
 import subprocess
 
 import pytest
@@ -51,7 +52,7 @@ def test_postgres_ingest_with_db(
     config_file = (
         test_resources_dir / "postgres_to_file_with_db_estimate_row_count.yml"
     ).resolve()
-    print("Config file: {config_file}")
+    print(f"Config file: {config_file}")
 
     run_datahub_cmd(["ingest", "-c", f"{config_file}"], tmp_path=tmp_path)
 
@@ -72,7 +73,7 @@ def test_postgres_ingest_with_all_db(
     config_file = (
         test_resources_dir / "postgres_all_db_to_file_with_db_estimate_row_count.yml"
     ).resolve()
-    print("Config file: {config_file}")
+    print(f"Config file: {config_file}")
 
     run_datahub_cmd(["ingest", "-c", f"{config_file}"], tmp_path=tmp_path)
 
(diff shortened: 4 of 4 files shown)


metadata-ingestion/src/datahub/ingestion/source/sql/sql_common.py
```diff
@@ -395,6 +395,12 @@ def _add_default_options(self, sql_config: SQLCommonConfig) -> None:
     def test_connection(cls, config_dict: dict) -> TestConnectionReport:
         test_report = TestConnectionReport()
         try:
+            if config_dict.get("stateful_ingestion", {}).get("enabled", False):
+                # Disable stateful ingestion for test connection
+                config_dict = {
+                    **config_dict,
+                    "stateful_ingestion": {"enabled": False},
+                }
```
sgomezvillamor, line 407: - `self.domain_registry` in line 366 also requires `DataHubGraph`
- we can make this unconditionally
- in the comment, the why is more important than the how

```suggestion
            # Disable settings that require DataHubGraph 
            config_dict = {
                **config_dict,
                "stateful_ingestion": {"enabled": False},
                "domain": None,
            }
```

Still wondering if we could have an _allow_ approach instead of the _deny_ one, that is, keeping very few configs instead of keeping everything and disabling some, so we don't need to add new configs in the future.

asikowitz: I tried to do something like that but couldn't find anything satisfying. Feel free to make your own fix, but I want a fix as soon as possible. I think this one is sufficient as a stopgap.

sgomezvillamor: Makes sense.

The _allow_ approach was a nice-to-have, not a blocking request. Same for everything else, except for updating the comment to capture the _why_. Feel free to use my suggestion or go with your own.

metadata-ingestion/src/datahub/ingestion/run/connection.py
```diff
@@ -2,34 +2,36 @@
...
 from datahub._version import __version__
 from datahub.ingestion.api.source import TestableSource, TestConnectionReport
+from datahub.ingestion.graph.config import ClientMode
+from datahub.ingestion.run.pipeline_config import BasePipelineConfig
 from datahub.ingestion.source.source_registry import source_registry
 
 logger = logging.getLogger(__name__)
 
 
-class ConnectionManager:
```
asikowitz, line 10: Saw no reason for the class

metadata-ingestion/src/datahub/ingestion/run/pipeline.py
```diff
@@ -181,21 +181,14 @@ def __init__(
         ):
             self.graph: Optional[DataHubGraph] = None
             with _add_init_error_context("connect to DataHub"):
-                if self.config.datahub_api:
-                    self.config.datahub_api.client_mode = ClientMode.INGESTION
-                    self.graph = exit_stack.enter_context(
-                        DataHubGraph(self.config.datahub_api)
-                    )
+                graph = self.config.make_graph(ClientMode.INGESTION)
```
asikowitz: Matching implementation with test connection code

metadata-ingestion/src/datahub/ingestion/run/pipeline_config.py
```diff
@@ -84,15 +86,44 @@ def _generate_run_id(source_type: Optional[str] = None) -> str:
     return f"{source_type}-{current_time}-{random_suffix}"
 
 
-class PipelineConfig(ConfigModel):
+class BasePipelineConfig(BaseModel):
```
sgomezvillamor: This `BasePipelineConfig(BaseModel)` can be a problematic design choice. It looks hacky and it may introduce validation inconsistences: BasePipelineConfig accepts extra fields, PipelineConfig rejects them. This creates confusion about what's allowed and beyond that, Pydantic's config merging behavior with multiple BaseModel inheritance can be unpredictable.

I think we should separate responsibilities here: let config to handle validation while introducing a factory pattern to handle context creation.

I have generated something with claude following this approach for illustration purposes (show me the code 😄), see below:

----

```
# pipeline_config.py

class PipelineContextFactory:
  """Factory for creating PipelineContext from various sources."""

  @staticmethod
  def from_pipeline_config(
      config: PipelineConfig,
      dry_run: bool = False,
      preview_mode: bool = False,
  ) -> PipelineContext:
      """Create context from full pipeline config."""
      graph = PipelineContextFactory._make_graph(
          config.datahub_api, ClientMode.INGESTION
      )
      return PipelineContext(
          run_id=config.run_id,
          pipeline_name=config.pipeline_name,
          graph=graph,
          dry_run=dry_run,
          preview_mode=preview_mode,
          pipeline_config=config,
      )

  @staticmethod
  def from_test_connection_dict(
      pipeline_config_dict: dict,
  ) -> PipelineContext:
      """Create minimal context for test connection.

      Accepts partial recipe dicts with optional fields.
      Designed for test_connection scenarios where full config validation
      would reject valid test configs.
      """
      # Parse only the fields we need, ignore extra fields
      run_id = pipeline_config_dict.get("run_id", "test_connection")
      pipeline_name = pipeline_config_dict.get("pipeline_name")
      datahub_api_dict = pipeline_config_dict.get("datahub_api")

      graph: Optional[DataHubGraph] = None
      if datahub_api_dict:
          try:
              datahub_api = DatahubClientConfig.model_validate(datahub_api_dict)
              graph = PipelineContextFactory._make_graph(
                  datahub_api, ClientMode.INGESTION
              )
          except Exception as e:
              logger.warning(
                  f"Failed to create DataHubGraph for test connection: {e}. "
                  f"Continuing without graph.",
                  exc_info=True
              )

      return PipelineContext(
          run_id=run_id,
          pipeline_name=pipeline_name,
          graph=graph,
          dry_run=True,
          preview_mode=True,
          pipeline_config=None,  # Not needed for test connection
      )

  @staticmethod
  def _make_graph(
      datahub_api: Optional[DatahubClientConfig],
      client_mode: ClientMode,
  ) -> Optional[DataHubGraph]:
      """Create DataHubGraph if config provided."""
      if not datahub_api:
          return None

      try:
          datahub_api.client_mode = client_mode
          graph = DataHubGraph(datahub_api)
          logger.debug(f"Created DataHubGraph in {client_mode} mode")
          return graph
      except Exception as e:
          logger.error(f"Failed to create DataHubGraph: {e}", exc_info=True)
          raise


# PipelineConfig stays unchanged

class PipelineConfig(ConfigModel):  # Just ConfigModel, no BasePipelineConfig
  source: SourceConfig
  run_id: str = DEFAULT_RUN_ID
  datahub_api: Optional[DatahubClientConfig] = None
  pipeline_name: Optional[str] = None
  # ... rest of fields

Usage in connection.py

def _run_test_connection(
  source_cls: type[TestableSource],
  pipeline_config: dict
) -> TestConnectionReport:
  with contextlib.ExitStack() as stack:
      ctx = PipelineContextFactory.from_test_connection_dict(pipeline_config)
      if ctx.graph:
          stack.enter_context(ctx.graph)

      return source_cls.test_connection_with_ctx(
          pipeline_config.get("source", {}).get("config", {}),
          ctx
      )

Usage in pipeline.py

class Pipeline:
  def __init__(self, config: PipelineConfig, ...):
      with contextlib.ExitStack() as exit_stack:
          ctx = PipelineContextFactory.from_pipeline_config(
              config,
              dry_run=dry_run,
              preview_mode=preview_mode
          )
          if ctx.graph:
              self.graph = exit_stack.enter_context(ctx.graph)

          self.ctx = ctx
```

Benefits of factory class:
* No inheritance complexity - Single PipelineConfig class inheriting only from ConfigModel
* Clear separation of concerns - Factory handles context creation, config handles validation
* Explicit intent - from_test_connection_dict() vs from_pipeline_config() makes usage clear
* Better error handling - Factory can handle partial dicts, graph creation failures, etc.
* Resource management - Caller controls when/how graph is closed (via ExitStack) -- of course, this is also fixable in your proposal 😄

asikowitz: To be clear, I tested that the inheritance worked and we still disallowed extra fields for `PipelineConfig`. But sure, this works. I think `from_test_connection_dict` is inaccurate naming though -- test connection and regular ingestion have the same exact pipeline config dict passed in. But test connection might not have everything filled out yet, which is why we don't want to actually instantiate PipelineConfig fully. I don't think the AI is really understanding that. So imo it's awkward to manually read the same fields that PipelineConfig is parsing. If in the future we change PipelineConfig fields, we might miss changing it for test connection, even though it will very likely change for test connection as well

PR #15918, opened 2026-01-18
RFC: DataHub CLI Profile Management

shirshanka:
# RFC: DataHub CLI Profile Management

This RFC proposes adding multi-profile support to the DataHub CLI to enable users to manage connections to multiple DataHub instances and easily switch between them.

## Summary

Replace the single-profile `~/.datahubenv` with a flexible `~/.datahub/config.yaml` system while maintaining full backward compatibility.

## Key Features

- 🔄 **Multi-profile configuration**: Manage dev, staging, prod profiles in one file
- 🎯 **Flexible profile selection**: Via `--profile` flag, env var, or current setting
- 🔐 **Secure credentials**: Environment variable interpolation (`${VAR}` syntax)
- ⚠️ **Production safety**: Confirmation prompts for destructive operations
- ✅ **Backward compatible**: Existing `.datahubenv` continues to work
- 🛠️ **Profile management commands**: `list`, `add`, `use`, `show`, `test`, etc.

## Motivation

Current pain points this addresses:
- Users must manually edit config files when switching environments
- No safety rails for production operations
- Poor credential management guidance
- Difficult to maintain multiple customer/project configurations

## What's in This RFC

The RFC document (`docs/rfcs/active/000-cli-profile-management.md`) includes:

- **Detailed design** with architecture diagrams and code examples
- **Complete requirements** (functional, non-functional, extensibility)
- **Profile precedence rules** (flag > env var > current > default > legacy)
- **Security features** and best practices
- **Migration strategy** from legacy config
- **Rollout plan** (4 phases with timeline)
- **Future enhancements** (10 potential extensions)
- **Alternatives considered** (5 different approaches analyzed)

## Review Checklist

- [x] RFC follows template structure
- [x] Requirements are clear and complete
- [x] Alternatives are considered
- [x] Rollout strategy is sound
- [x] Backward compatibility is maintained
- [x] Drawbacks are honestly assessed

## Feedback Requested

- Overall approach and design
- Profile precedence order
- Configuration file format (YAML)
- Production safety mechanisms
- Migration strategy from `.datahubenv`
- Profile naming conventions
- Environment variable interpolation syntax

## Implementation Status

This RFC is based on a working proof-of-concept implementation with:
- Full profile system implementation
- 97 passing unit tests (config models, profile management, CLI commands)
- All CLI commands updated to support profiles
- Comprehensive documentation

The implementation PR will follow separately after RFC approval.

---

**RFC Document**: `docs/rfcs/active/000-cli-profile-management.md`

[...]

PR #15919, opened 2026-01-18
feat: Add multi-profile support to DataHub CLI

shirshanka:
# DataHub CLI Profile Management Implementation

This PR implements multi-profile support for the DataHub CLI as proposed in RFC #15918.

## Summary

Adds comprehensive profile management system enabling users to manage connections to multiple DataHub instances (dev, staging, production) and easily switch between them.

### Profile Management
- 🔄 **Multi-profile configuration** in `~/.datahub/config.yaml`
- 🎯 **Flexible profile selection**: `--profile` flag, `DATAHUB_PROFILE` env var, or `current_profile` setting
- 📋 **Profile commands**: `list`, `current`, `show`, `use`, `add`, `remove`, `test`, `validate`, `export`

### Security & Safety
- 🔐 **Environment variable interpolation**: `${VAR}`, `${VAR:-default}`, `${VAR:?error}` syntax
- ⚠️ **Production safety**: `require_confirmation` flag for destructive operations
- 🙈 **Token redaction** in CLI output

### Backward Compatibility
- ✅ **Full backward compatibility** with existing `~/.datahubenv`
- 🔄 **Automatic detection** and fallback to legacy configuration
- 🌐 **Environment variable overrides** (`DATAHUB_GMS_URL`, etc.) continue to work

### ✅ ALL CLI Commands Now Support Profiles

**Core Commands:**
- `datahub ingest` - All subcommands (run, deploy, mcps, list-source-runs, list-runs, show, rollback)
- `datahub get` - Retrieve metadata with profile support
- `datahub put` - Write metadata using specified profile
- `datahub delete` - Delete metadata with confirmation prompts
- `datahub exists` - Check entity existence
- `datahub graphql` - Execute GraphQL queries and mutations

**Management Commands:**
- `datahub state` - Inspect ingestion state
- `datahub timeline` - View entity timelines
- `datahub check` - All check commands (sql-lineage, server-config, restore-indices, kafka-consumer-offsets)

**Catalog Commands:**
- `datahub iceberg` - All Iceberg catalog operations (create, update, list, get, delete)

**Entity-Specific Commands:**
- `datahub dataset` - Dataset operations (get, add-sibling, sync)
- `datahub dataproduct` - All dataproduct management (upsert, update, get, delete, add/remove assets/owners, set-description)
- `datahub datacontract` - Upsert and delete data contracts
- `datahub user` - User management (upsert, add)
- `datahub group` - Group management (upsert)
- `datahub forms` - Form operations (get)
- `datahub properties` - Structured properties (upsert, get, list)
- `datahub assertions` - Assertion management (upsert)

**Profile Management Commands:**
- `datahub profile list` - List all profiles
- `datahub profile current` - Show active profile
- `datahub profile show` - Display profile details
- `datahub profile use` - Set current profile
- `datahub profile add` - Add/update profile
- `datahub profile remove` - Delete profile
- `datahub profile test` - Test connection
- `datahub profile validate` - Validate config file
- `datahub profile export` - Export profile (redacted)

### Comprehensive User Documentation Added
- 📖 **Complete profile guide**: `docs/how/manage-cli-profiles.md` (~500 lines)
  - Why use profiles (problem/solution format)
  - Quick start (3-step tutorial)
  - Common use cases (dev/staging/prod, secure tokens, customer instances)
  - All profile commands with examples
  - Profile selection priority explained
  - Production safety features
  - Configuration file format reference
  - Migration from legacy config
  - Troubleshooting guide with solutions
  - Best practices and security recommendations
  - Advanced usage patterns

- 📝 **Main CLI docs updated**: `docs/cli.md`
  - Added profile section with quick reference
  - Added `DATAHUB_PROFILE` environment variable
  - Links to comprehensive guide

### Architecture
- **New Pydantic models**: `ProfileConfig` and `DataHubConfig` in `configuration/config.py`
- **Profile precedence**: `--profile` flag > `DATAHUB_PROFILE` env > `current_profile` > "default" profile > legacy config
- **Updated core function**: `get_default_graph()` now accepts `profile` parameter
- **Global CLI integration**: `--profile` flag on main `datahub` command
- **Universal CLI support**: All 20+ CLI command files updated with profile support

### Profile Management Commands
```bash
datahub profile list              # List all profiles
datahub profile current           # Show active profile
datahub profile show <name>       # Display profile details
datahub profile use <name>        # Set current profile
datahub profile add <name> ...    # Add/update profile
datahub profile remove <name>     # Delete profile
datahub profile test <name>       # Test connection
datahub profile validate          # Validate config file
datahub profile export <name>     # Export profile (redacted)
```

### Production Safety
Profiles with `require_confirmation: true` prompt for confirmation before destructive operations:
- User must type profile name in UPPERCASE to confirm
- Shows full context (profile, server, operation details)
- Can be bypassed with `--force` flag for automation

### Example Configuration
```yaml
version: "1.0"
current_profile: dev

profiles:
  dev:
    server: http://localhost:8080
    token: ${DATAHUB_DEV_TOKEN}
    description: "Local development"

  prod:
    server: https://datahub.company.com
    token: ${DATAHUB_PROD_TOKEN}
    timeout_sec: 120
    require_confirmation: true
    description: "Production - USE WITH CAUTION"
```

### Comprehensive Test Suite: 97 Tests Passing ✅
- **25 tests** for configuration models and environment variable interpolation
- **44 tests** for profile management functions  
- **28 tests** for CLI commands

### Test Coverage
- Configuration model validation
- Profile loading with precedence handling
- CLI command functionality
- Environment variable interpolation
- Error handling and edge cases
- Token redaction
- Production confirmation prompts
- Backward compatibility
- Migration scenarios

### New Files (5)
- `src/datahub/configuration/config.py` - Pydantic models for profile configuration
- `src/datahub/cli/profile_cli.py` - Profile management CLI commands
- `tests/unit/test_config.py` - Config model tests
- `tests/unit/cli/test_config_utils_profiles.py` - Profile function tests
- `tests/unit/cli/test_profile_cli.py` - CLI command tests

### New Documentation (2)
- `docs/how/manage-cli-profiles.md` - Comprehensive profile management guide
- `docs/cli.md` (updated) - Main CLI docs with profile section

### Modified Core Files (4)
- `src/datahub/configuration/env_vars.py` - Add `get_active_profile()`
- `src/datahub/cli/config_utils.py` - Profile loading and management functions
- `src/datahub/ingestion/graph/client.py` - Add profile param to `get_default_graph()`
- `src/datahub/entrypoints.py` - Global `--profile` flag

### Modified CLI Commands (20 files)
All CLI commands now support profile parameter:
- `cli/delete_cli.py` - Delete operations with confirmation
- `cli/get_cli.py` - Get metadata
- `cli/put_cli.py` - Put metadata
- `cli/ingest_cli.py` - All ingest commands
- `cli/graphql_cli.py` - GraphQL operations
- `cli/exists_cli.py` - Entity existence checks
- `cli/state_cli.py` - State inspection
- `cli/timeline_cli.py` - Timeline queries
- `cli/check_cli.py` - System checks
- `cli/iceberg_cli.py` - Iceberg catalog
- `cli/specific/dataset_cli.py` - Dataset operations
- `cli/specific/dataproduct_cli.py` - Data product management
- `cli/specific/datacontract_cli.py` - Data contracts
- `cli/specific/user_cli.py` - User management
- `cli/specific/group_cli.py` - Group management
- `cli/specific/forms_cli.py` - Forms
- `cli/specific/structuredproperties_cli.py` - Structured properties
- `cli/specific/assertions_cli.py` - Assertions

## Breaking Changes

**None** - This is a fully backward-compatible feature addition.

## Migration

Existing users do not need to change anything. The CLI automatically detects and uses `~/.datahubenv` if present. Users can optionally migrate to the new profile system when ready.

## Related

- RFC: #15918
- Implements: Multi-profile configuration management
- Replaces: Single-profile `.datahubenv` limitation
- Adds: Profile management commands
- Improves: Security (env var interpolation), Safety (confirmation prompts), UX (easy switching)

## Checklist

- [x] Implementation complete (all CLI commands)
- [x] All tests passing (97/97)
- [x] Backward compatibility verified
- [x] Documentation complete (user guide + CLI reference)
- [x] Code linted and formatted
- [x] No breaking changes
- [x] Related RFC: #15918
- [x] Universal profile support across ALL CLI commands

---

**Ready for review!** 🚀

This PR provides complete profile support across the entire DataHub CLI, enabling seamless multi-environment workflows.

### DIFF
--- a/metadata-ingestion/tests/unit/cli/test_config_utils.py
+++ b/metadata-ingestion/tests/unit/cli/test_config_utils.py
@@ -159,7 +159,8 @@ class TestConfigUtils:
         with (
             patch.dict(os.environ, {}, clear=True),
             patch.object(config_utils, "_should_skip_config", return_value=False),
-            patch.object(config_utils, "_ensure_datahub_config"),
+            patch.object(config_utils, "_load_new_config", return_value=None),
+            patch.object(os.path, "isfile", return_value=True),
             patch.object(
                 config_utils, "get_raw_client_config", return_value=test_config
             ),


metadata-ingestion/src/datahub/configuration/config.py
```diff
@@ -0,0 +1,172 @@
...
+        description: "Local development"
+
+      prod:
+        server: https://datahub.company.com
+        token: ${DATAHUB_PROD_TOKEN}
+        require_confirmation: true
+        description: "Production - USE WITH CAUTION"
+    """
+
+    version: str = Field(default="1.0", description="Configuration file version")
```
skrydal, line 100: I have mixed feelings about this version field. First of all I would suggest renaming it to something which indicates that we are talking about version of the _schema_ of the config (if I understood correctly) and not another version of config values (i.e. incremented after adding a new profile).
My second point would be - we are really dependent here on the underlying `DatahubClientConfig`, which can be changed and there is no mechanism to increment version in such case. We, ideally, need a clear policy related to incrementing version, as well as handlers (version 1 to 2, version 2 to 3, etc.). I understand this field is left here for future, but maybe we can design better relation to the actual config model. It's not necessarily a merge blocker.

metadata-ingestion/src/datahub/configuration/config.py
```diff
@@ -0,0 +1,172 @@
...
+    @field_validator("server", "token", mode="before")
+    @classmethod
+    def interpolate_string_fields(cls, v: Optional[str]) -> Optional[str]:
+        """Interpolate environment variables in string fields."""
+        if v is None:
+            return v
+        return interpolate_env_vars(v)
+
+
+class DataHubConfig(DatahubClientConfig):
```
skrydal, line 78: Why do we make DataHubConfig inherit from `DatahubClientConfig` and then override (with a `None`) a single parameter (`server`)? Until I have noticed that `server` parameter is overridden, I thought it is to have the capability to configure connection details without any profiles (which I would not encourage in first place). For me current approach doesn't make any sense and I suggest inheriting from ConfigModel instead. Also please remove `server` field then.

```suggestion
class DataHubConfig(ConfigModel):
```

metadata-ingestion/tests/unit/test_config.py
```diff
@@ -0,0 +1,273 @@
...
+    def test_profile_with_confirmation(self) -> None:
+        """Test profile with require_confirmation flag."""
+        profile = ProfileConfig(
+            server="https://prod.datahub.com",
+            token="prod_token",
+            require_confirmation=True,
+            description="Production",
+        )
+        assert profile.require_confirmation is True
+        assert profile.description == "Production"
```
skrydal, line 101: I think tests like these are simply too trivial. There are other tests in this suite which are actually meaningful and test our logic. This test effective "tests" `pydantic` library.

metadata-ingestion/src/datahub/cli/ingest_cli.py
```diff
@@ -776,12 +797,19 @@ def show(
     default="./rollback-reports",
     help="Path to directory where rollback reports will be saved to",
 )
+@click.pass_context
 @upgrade.check_upgrade
 def rollback(
```
skrydal, line 802: It seems that currently we add confirmation check only to `delete by-filter` command, should we add it, as a rule of thumb, to every command which currently supports `dry_run`?

metadata-ingestion/src/datahub/cli/delete_cli.py
```diff
@@ -427,9 +432,39 @@ def by_filter(
...
+
+        if not config_utils.confirm_destructive_operation(
+            operation=operation_desc,
+            profile_name=profile_name,
+            force=force,
+            extra_info=extra_info,
+        ):
+            click.echo("Operation cancelled.")
+            return
+
```
skrydal, line 467: While, of course, part of this code-block is specific to this command, maybe we could make confirmation into a decorator so we can apply it easily to other commands? WDYT?

metadata-ingestion/src/datahub/cli/delete_cli.py
```diff
@@ -392,6 +394,9 @@ def by_filter(
 ) -> None:
     """Delete metadata from datahub using a single urn or a combination of filters."""
 
+    # Get profile from context
```
skrydal, line 397: I think we should avoid comments like this for a single, obvious lines

metadata-ingestion/src/datahub/cli/config_utils.py
```diff
@@ -28,6 +29,8 @@
 CONDENSED_DATAHUB_CONFIG_PATH = "~/.datahubenv"
 DATAHUB_CONFIG_PATH: str = os.path.expanduser(CONDENSED_DATAHUB_CONFIG_PATH)
 DATAHUB_ROOT_FOLDER: str = os.path.expanduser("~/.datahub")
+DATAHUB_CONFIG_DIR: str = os.path.expanduser("~/.datahub")
```
skrydal, line 32: why can't we use `DATAHUB_ROOT_FOLDER` instead (1 line above)?

metadata-ingestion/src/datahub/cli/config_utils.py
```diff
@@ -132,6 +156,476 @@ def _ensure_datahub_config() -> None:
...
+        click.echo()
+    else:
+        click.secho("✗ Confirmation failed. Operation cancelled.", fg="red")
+        click.echo()
+
+    return confirmed
+
+
+# ============================================================================
+# Legacy Functions (for backward compatibility)
```
skrydal, line 625: We are not writing the old config anymore, effectively we are retaining function which is not used anywhere but in tests. Maybe it would be better to remove it? Such code, saved for later, tends to create confusion.

metadata-ingestion/src/datahub/cli/config_utils.py
```diff
@@ -100,28 +103,49 @@ def require_config_from_env() -> Tuple[str, Optional[str]]:
...
 
-    if _should_skip_config():
-        raise MissingConfigError(
-            "You have set the skip config flag, but no GMS host or token was provided in env variables."
-        )
+    # If env vars are set, they will override profile settings in load_profile_config
+    if gms_host_env and not profile:
+        # For backward compatibility: if env vars are set and no profile specified,
+        # just use env vars
+        return DatahubClientConfig(server=gms_host_env, token=gms_token_env)
```
skrydal, line 131: This could be implemented as one condition: `if gms_host_env and (_should_skip_config() or not profile)`.
However I see serious change from past behavior. In past if `gms_host_env `was not set but skip config was set an exception was raised. Now this exception is raised under those conditions + if `load_profile_config` throws `MissingConfigError`. I think this was not intended. I would refactor it to:

```suggestion
    if _should_skip_config():
        if gms_host_env:
            return DatahubClientConfig(server=gms_host_env, token=gms_token_env)
        else:
            raise MissingConfigError()
     if gms_host_env and not profile:
         return DatahubClientConfig(server=gms_host_env, token=gms_token_env)
```
Also call to `_should_skip_config` would need to be removed from the except block.

This way we still have redundant conditional logic, but it is clear at first glance what is the intended logic. Moreover there is no behavior change.

metadata-ingestion/src/datahub/cli/config_utils.py
```diff
@@ -132,6 +156,476 @@ def _ensure_datahub_config() -> None:
...
+                server=profile_config.server,
+                token=profile_config.token,
+                timeout_sec=profile_config.timeout_sec,
+                retry_status_codes=profile_config.retry_status_codes,
+                retry_max_times=profile_config.retry_max_times,
+                extra_headers=profile_config.extra_headers,
+                ca_certificate_path=profile_config.ca_certificate_path,
+                client_certificate_path=profile_config.client_certificate_path,
+                disable_ssl_verification=profile_config.disable_ssl_verification,
+            )
```
skrydal, line 288: `get_profile` function already returns `ProfileConfig` which is a subclass of `DatahubClientConfig`. Is listing explicit parameters here, and recreating the object, intentional? This is quite verbose and we are guaranteed to forget to add parameters here after changing `DatahubClientConfig`...


```suggestion
            client_config = config.get_profile(selected_profile)
```

PR #15926, opened 2026-01-20
feat(sdk): automatically parse lineage from view definitions

alfiyas-datahub:
## Summary

When a Dataset is created with a `view_definition` containing SQL, the SDK now automatically parses the SQL to extract upstream table references and populates the upstreams lineage. This eliminates the need for users to manually specify both `view_definition` AND `upstreams` separately.

## Key Changes

- Added `_parse_and_set_view_lineage()` method using sqlglot for SQL parsing
- Modified `set_view_definition()` to auto-parse lineage with `parse: Optional[bool] = None`
- Auto-disabled in ingestion framework via `is_ingestion_attribution()` check
- Explicit upstreams are preserved (not overwritten by auto-parsing)
- Supports multiple SQL dialects (Snowflake, BigQuery, MSSQL, Athena, etc.)
- Filters out CTEs from upstream tables
- Graceful error handling for invalid SQL

Closes: ING-1197

### DIFF
--- a/metadata-ingestion/src/datahub/sdk/dataset.py
+++ b/metadata-ingestion/src/datahub/sdk/dataset.py
@@ -450,6 +450,50 @@ class Dataset(
     A dataset represents a collection of data, such as a table, view, or file.
     This class provides methods for managing dataset metadata including schema,
     lineage, and various aspects like ownership, tags, and terms.
+
+    **Automatic View Lineage Parsing**
+
+    When a `view_definition` SQL string is provided, the SDK can automatically
+    parse it to extract upstream table references using sqlglot. This behavior
+    is controlled by the `parse_view_lineage` parameter:
+
+    - None (default): Auto-parse in SDK mode, skip in ingestion framework
+      (where SqlParsingAggregator handles lineage more accurately)
+    - True: Force parsing (even within the ingestion framework)
+    - False: Never parse automatically
+
+    **Features:**
+
+    - Supports all major SQL dialects (Snowflake, BigQuery, Postgres, etc.)
+    - Filters out CTEs (Common Table Expressions) from upstream tables
+    - Gracefully handles parse errors without raising exceptions
+
+    **Limitations:**
+
+    - Table-level lineage only (no column-level lineage)
+    - No schema resolution: table names are taken as-is from the SQL.
+      For fully-qualified references, ensure your view definition includes
+      the database and schema (e.g., db.schema.table)
+    - Best-effort extraction; for production-grade lineage with schema resolution
+      and additional context (e.g., temporary tables), use SqlParsingAggregator
+
+    Example::
+
+        # Auto-parsing enabled by default
+        view = Dataset(
+            platform="snowflake",
+            name="analytics.reporting.sales_summary",
+            view_definition="SELECT * FROM analytics.raw.sales WHERE year = 2024",
+        )
+        # view.upstreams now contains analytics.raw.sales
+
+        # Disable auto-parsing
+        view = Dataset(
+            platform="snowflake",
+            name="analytics.reporting.sales_summary",
+            view_definition="SELECT * FROM analytics.raw.sales",
+            parse_view_lineage=False,  # Must set upstreams manually
+        )
     """
 
     __slots__ = ()
--- a/metadata-ingestion/tests/unit/sdk_v2/dataset_golden/test_dataset_complex_golden.json
+++ b/metadata-ingestion/tests/unit/sdk_v2/dataset_golden/test_dataset_complex_golden.json
@@ -114,7 +114,7 @@
     "aspect": {
         "json": {
             "materialized": false,
-            "viewLogic": "SELECT * FROM my_other_table",
+            "viewLogic": "SELECT * FROM my_db.my_schema.my_other_table",
             "viewLanguage": "SQL"
         }
     }
@@ -132,7 +132,7 @@
                         "time": 0,
                         "actor": "urn:li:corpuser:unknown"
                     },
-                    "dataset": "urn:li:dataset:(urn:li:dataPlatform:snowflake,my_other_table,PROD)",
+                    "dataset": "urn:li:dataset:(urn:li:dataPlatform:snowflake,my_db.my_schema.my_other_table,PROD)",
                     "type": "VIEW"
                 }
             ]
--- a/metadata-ingestion/tests/unit/sdk_v2/test_dataset.py
+++ b/metadata-ingestion/tests/unit/sdk_v2/test_dataset.py
@@ -126,7 +126,7 @@ def _build_complex_dataset() -> Dataset:
             GlossaryTermUrn("AccountBalance"),
         ],
         domain=DomainUrn("Marketing"),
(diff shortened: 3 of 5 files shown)


sgomezvillamor:
I haven’t reviewed the code yet; this feedback is based only on the feature described in the PR description.
Overall, while this looks nice from an SDK point of view, I’m not fully convinced about its feasibility.

My main concerns are:
* Beyond parsing, we usually need schema resolution (see `SchemaResolver` in `SqlParsingAggregator`) to accurately infer lineage. Is that something we can safely skip when parsing views?
* Lineage resolution often requires additional context beyond the view definition itself (for example, temporary tables). That’s why we currently resolve lineage in the SQL aggregator, where that extra context is available. Again, is that something we can skip when parsing views?
* Given these points, are we comfortable with potentially lower-accuracy lineage?

In any case, we should make sure this mechanism is disabled when the SDK is used within the ingestion framework, which fully delegates lineage resolution to the `SqlParsingAggregator`.

metadata-ingestion/src/datahub/sql_parsing/sqlglot_utils.py
```diff
@@ -20,30 +21,6 @@
...
-    # At least it allows to parse simple SQL queries and built linage for them
-    elif platform == "salesforce":
-        return "databricks"
-    elif platform in {"mysql", "mariadb"}:
-        # In sqlglot v20+, MySQL is now case-sensitive by default, which is the
-        # default behavior on Linux. However, MySQL's default case sensitivity
-        # actually depends on the underlying OS.
-        # For us, it's simpler to just assume that it's case-insensitive, and
-        # let the fuzzy resolution logic handle it.
-        # MariaDB is a fork of MySQL, so we reuse the same dialect.
```
sgomezvillamor, line 41: after the moving the method, let's keep these comments

metadata-ingestion/tests/unit/sdk_v2/test_dataset.py
```diff
@@ -466,3 +466,304 @@ def test_structured_properties() -> None:
...
+
+    view = Dataset(
+        platform="snowflake",
+        name="db.schema.my_view",
+        view_definition="SELECT * FROM source_table",
+    )
+
+    # Should gracefully handle unexpected error during auto-parsing
+    assert view.view_definition is not None
+    assert view.upstreams is None
```
sgomezvillamor, line 777: in all these error cases (sqlglot not installed, sqlglot error, unexpected error), we have no other feedback that `view.upstreams is None`
should we also check that the logger has a message reporting the issue?

metadata-ingestion/tests/unit/sdk_v2/test_dataset.py
```diff
@@ -466,3 +466,304 @@ def test_structured_properties() -> None:
...
+            name="db.schema.my_view",
+            view_definition="SELECT * FROM db.schema.source_table",
+            # parse_view_lineage=None (default) - auto-skips in ingestion mode
+        )
+
+        # View definition should be set, but lineage should NOT be auto-parsed
+        assert view.view_definition is not None
+        assert view.view_definition.viewLogic == "SELECT * FROM db.schema.source_table"
+        assert view.upstreams is None  # Not auto-parsed in ingestion mode
+
```
sgomezvillamor, line 818: we may also assert that parsing method has not been called
otherwise, `assert view.upstreams is None` could be also the case of error scenarios during parsing

metadata-ingestion/tests/unit/sdk_v2/test_dataset.py
```diff
@@ -466,3 +466,304 @@ def test_structured_properties() -> None:
...
+        ],
+    )
+
+    # Auto-parsing should happen by default
+    assert view.view_definition is not None
+    assert "source_table" in view.view_definition.viewLogic
+    assert view.upstreams is not None
+    assert len(view.upstreams.upstreams) == 1
+    assert "source_table" in view.upstreams.upstreams[0].dataset
+
```
sgomezvillamor, line 521: we may also add a test of eg cross database references

```
def test_view_definition_auto_parses_lineage_cross_database() -> None:
    """
    When view_definition is set, lineage should be auto-parsed by default and handle cross database references properly
    """
    view = Dataset(
        platform="snowflake",
        name="db.schema.my_view",
        view_definition="SELECT id, name FROM another_db.another_schema.source_table WHERE active = true",
        schema=[
            ("id", "int64"),
            ("name", "string"),
        ],
    )

    # Auto-parsing should happen by default
    assert view.view_definition is not None
    assert "source_table" in view.view_definition.viewLogic
    assert view.upstreams is not None
    assert len(view.upstreams.upstreams) == 1
    assert "another_db.another_schema.source_table" in view.upstreams.upstreams[0].dataset
```

not sure about this, but you get the idea (validate full urn or full name)

```
    assert "another_db.another_schema.source_table" in view.upstreams.upstreams[0].dataset
```


This makes me think that, in previous test,

```
    assert "source_table" in view.upstreams.upstreams[0].dataset
```

it should be more precise by checking full urn or full name too

metadata-ingestion/tests/unit/sdk_v2/dataset_golden/test_dataset_complex_golden.json
```diff
@@ -119,6 +119,26 @@
...
+    "aspectName": "upstreamLineage",
+    "aspect": {
+        "json": {
+            "upstreams": [
+                {
+                    "auditStamp": {
+                        "time": 0,
+                        "actor": "urn:li:corpuser:unknown"
+                    },
+                    "dataset": "urn:li:dataset:(urn:li:dataPlatform:snowflake,my_other_table,PROD)",
```
sgomezvillamor: entity urn is `urn:li:dataset:(urn:li:dataPlatform:snowflake,my_instance.my_db.my_schema.my_table,PROD) ` - it includes platform instance, db and schema
however referenced dataset is `urn:li:dataset:(urn:li:dataPlatform:snowflake,my_other_table,PROD)` - it misses platform instance, db and schema, isn't that an issue

metadata-ingestion/src/datahub/sdk/dataset.py
```diff
@@ -762,9 +787,170 @@ def set_view_definition(self, view_definition: ViewDefinitionInputType) -> None:
...
+        - Views cannot reference session-scoped temporary tables (they're persistent)
+
+        Features:
+        - Supports all major SQL dialects (Snowflake, BigQuery, Postgres, etc.)
+        - Filters out CTEs (Common Table Expressions) from upstream tables
+        - Gracefully handles parse errors without raising exceptions
+
+        Limitations:
+        - Table-level lineage only (no column-level lineage)
+        - Best-effort extraction; for production-grade lineage use SqlParsingAggregator
```
sgomezvillamor, line 862: Features and limitations should have better visibility
May we add a docstring for dataset entity and place there a paragraph about parsing view definitions?

sgomezvillamor (commented):
All comments are minor, except the one about the golden files — that’s actually the only thing blocking my approval. Could you please double-check it?

PR #15928, opened 2026-01-20
fix(ci): add auto-retry for additional workflows

chakru-r:

### DIFF
--- a/.github/workflows/workflow-monitor.yml
+++ b/.github/workflows/workflow-monitor.yml
@@ -113,8 +113,8 @@ jobs:
       - name: Retrigger workflows known to have flaky tests on first failure
         if: |
           (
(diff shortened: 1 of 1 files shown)


.github/workflows/workflow-monitor.yml
```diff
@@ -110,9 +110,12 @@ jobs:
             echo "Response: $BODY"
           fi
 
-      - name: Retrigger docker-unified on first failure
+      - name: Retrigger workflows known to have flaky tests on first failure
         if: |
-          github.event.workflow_run.name == 'Docker Build, Scan, Test' &&
+          (
```
devashish2203, line 115: nit: As we were discussing yesterday, there should be a way to only retrigger whe the job for the test run fails. 
Currently it retriggers if the Build Image step itself fails.

devashish2203 (approved):

PR #15930, opened 2026-01-20
fix(ingest/airflow): Remove kill switch variable from Airflow 2.x support

treff7es:
The DataHub Airflow plugin's kill switch used Variable.get() to check if the plugin should be disabled:       
                                                                                                                
  if Variable.get("datahub_airflow_plugin_disable_listener", "false").lower() == "true":                        
      return True                                                                                               
                                                                                                                
  Variable.get() uses the @provide_session decorator which creates a new database session and commits it.       
  Listener hooks execute during SQLAlchemy's after_flush event (before the main transaction commits). This      
  nested commit corrupts the outer transaction state, causing data loss—specifically, TaskInstanceHistory       
  records were not being persisted for retried tasks.                                                           
                                                                                                                
  See upstream issue: https://github.com/apache/airflow/pull/48780                                              
                                                                                                                
  Solution                                                                                                      
                                                                                                                
  Changed the kill switch to use environment variables instead of database queries:                             
                                                                                                                
  if os.getenv("AIRFLOW_VAR_DATAHUB_AIRFLOW_PLUGIN_DISABLE_LISTENER", "false").lower() == "true":               
      return True                                                                                               
                                                                                                                
  Migration                                                                                                     
                                                                                                                
  To disable the plugin, use an environment variable instead of an Airflow Variable:                            
                                                                                                                
  # Old way (no longer works)                                                                                   
  airflow variables set datahub_airflow_plugin_disable_listener true                                            
                                                                                                                
  # New way                                                                                                     
  export AIRFLOW_VAR_DATAHUB_AIRFLOW_PLUGIN_DISABLE_LISTENER=true                                               

<!--

Thank you for contributing to DataHub!

- [ ] The PR conforms to DataHub's [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [PR Title Format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#pr-title-format))

-->


docs/lineage/airflow.md
```diff
@@ -413,24 +413,20 @@ Set the `datahub.enabled` configuration property to `False` in the `airflow.cfg`
...
+#### 2. Disable via Environment Variable (Kill-Switch)
 
-If a restart is not possible and you need a faster way to disable the plugin, you can use the kill-switch. Create and set the `datahub_airflow_plugin_disable_listener` Airflow variable to `true`. This ensures that the listener won't process anything.
-
-#### Command Line
+If a restart is not possible and you need a faster way to disable the plugin, you can use the kill-switch. Set the `AIRFLOW_VAR_DATAHUB_AIRFLOW_PLUGIN_DISABLE_LISTENER` environment variable to `true`. This ensures that the listener won't process anything.
 
 ```shell
-airflow variables set datahub_airflow_plugin_disable_listener true
+export AIRFLOW_VAR_DATAHUB_AIRFLOW_PLUGIN_DISABLE_LISTENER=true
```
sgomezvillamor, line 421: Even if not critical, IMO this should be listed as a breaking change.

metadata-ingestion-modules/airflow-plugin/src/datahub_airflow_plugin/airflow2/datahub_listener.py
```diff
@@ -22,7 +22,9 @@
 # Import Airflow 2.x specific shims (clean, no cross-version complexity)
 import airflow.version
 import packaging.version
-from airflow.models import Variable
```
sgomezvillamor, line 25: maybe we can ban the import via ruff?
not sure if ruff configured in this module though

PR #15932, opened 2026-01-20
DRAFT - feat(ingestion/usage): performance improvement for query processing and sharded table metadata retrieval in bigquery - DO NOT MERGE

acrylJonny:

### DIFF
--- a/metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
+++ b/metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
@@ -31,6 +31,7 @@ import sqlglot.optimizer.unnest_subqueries
 from pydantic import field_serializer, field_validator
 
 from datahub.cli.env_utils import get_boolean_env_variable
+from datahub.configuration.env_vars import get_sql_parse_cache_size
 from datahub.ingestion.graph.client import DataHubGraph
 from datahub.metadata.schema_classes import (
     ArrayTypeClass,
@@ -176,7 +177,7 @@ def _table_name_from_sqlglot_table(
 
 Urn = str
 
-SQL_PARSE_RESULT_CACHE_SIZE = 100000
+SQL_PARSE_RESULT_CACHE_SIZE = get_sql_parse_cache_size()
 SQL_LINEAGE_TIMEOUT_ENABLED = get_boolean_env_variable(
     "SQL_LINEAGE_TIMEOUT_ENABLED", True
 )
--- a/metadata-ingestion/src/datahub/configuration/env_vars.py
+++ b/metadata-ingestion/src/datahub/configuration/env_vars.py
@@ -179,6 +179,11 @@ def get_sql_agg_query_log() -> str:
     return os.getenv("DATAHUB_SQL_AGG_QUERY_LOG", "DISABLED")
 
 
+def get_sql_parse_cache_size() -> int:
+    """SQL parse result cache size (number of entries)."""
+    return int(os.getenv("DATAHUB_SQL_PARSE_CACHE_SIZE", "100000"))
+
+
 def get_dataset_urn_to_lower() -> str:
     """Convert dataset URNs to lowercase."""
     return os.getenv("DATAHUB_DATASET_URN_TO_LOWER", "false")
--- a/metadata-ingestion/src/datahub/sql_parsing/sqlglot_utils.py
+++ b/metadata-ingestion/src/datahub/sql_parsing/sqlglot_utils.py
@@ -9,14 +9,15 @@ import sqlglot
 import sqlglot.errors
 import sqlglot.optimizer.eliminate_ctes
 
+from datahub.configuration.env_vars import get_sql_parse_cache_size
 from datahub.sql_parsing.fingerprint_utils import generate_hash
 
 assert SQLGLOT_PATCHED
 
 logger = logging.getLogger(__name__)
 DialectOrStr = Union[sqlglot.Dialect, str]
-SQL_PARSE_CACHE_SIZE = 100000
-FORMAT_QUERY_CACHE_SIZE = 100000
+SQL_PARSE_CACHE_SIZE = get_sql_parse_cache_size()
+FORMAT_QUERY_CACHE_SIZE = get_sql_parse_cache_size()
 
 
 def _get_dialect_str(platform: str) -> str:


metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_audit.py
```diff
@@ -43,6 +44,19 @@ class BigqueryTableIdentifier:
...
+        """Get the compiled regex pattern for shard detection (cached)."""
+        return re.compile(pattern_str, re.IGNORECASE)
+
+    @staticmethod
+    def _get_shard_pattern() -> "re.Pattern[str]":
+        """Get the compiled regex pattern for shard detection."""
+        return BigqueryTableIdentifier._get_compiled_shard_pattern(
+            BigqueryTableIdentifier._BIGQUERY_DEFAULT_SHARDED_TABLE_REGEX
+        )
+
```
sgomezvillamor: this looks over-engineered. why not just compile once in a shared global var?

```
_BIGQUERY_DEFAULT_SHARDED_TABLE_REGEX_PATTERN = re.compile(
    BigqueryTableIdentifier._BIGQUERY_DEFAULT_SHARDED_TABLE_REGEX,
    re.IGNORECASE
)
```

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_schema_gen.py
```diff
@@ -530,13 +530,37 @@ def _process_schema(
...
+                # Fast shard detection without object creation
+                match = shard_pattern.match(table_id)
+                if match:
+                    base_name = match[1] or ""
+                    if base_name and table_id.endswith(match[3]):
+                        base_name = table_id[: -len(match[3])].rstrip("_")
+                    if not base_name or base_name.endswith("."):
+                        base_name = dataset_name
+
+                    # Skip if we've already processed this base table
```
sgomezvillamor, line 550: Everything that we can skip is great but, what are we skipping at this point?

On the other side, we have a method 

```
    def get_core_table_details(
        self, dataset_name: str, project_id: str, temp_table_dataset_prefix: str
    ) -> Dict[str, TableListItem]:
```

where we process shards, not sure if we are breaking something there

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_audit.py
sgomezvillamor: on the other side, shared suffix is configurable, right? I'm surprised we use the `_BIGQUERY_DEFAULT_SHARDED_TABLE_REGEX` for a pattern that is configruable 🤔

metadata-ingestion/src/datahub/sql_parsing/schema_resolver.py
sgomezvillamor, line 1: The `SchemaResolver` have two operational modes:
* _prefetch_: where the resolver is initialized in batch with all URNs and schemas got from DataHub for a given platform and platform instance, once initialized, no more accesses to DataHub
* _lazy_ or _on-demand_: where the resolver is not initialized and instead every URN will be individually attempt once in DataHub

My guess is _prefetch_ mode should be more efficient in most of the cases. I have pending to check what's the operational mode for most of the sources and review that. IMO _lazy_ model would only make sense in those cases where we have very selective dataset filtering.

This `BatchSchemaFetcher` adds batching to the _lazy_ mode

```
        """
        Resolve schema info for a URN, fetching from DataHub if needed.
        Cache precedence policy:
        1. If in cache (from ingestion or previous fetch), return immediately
        2. If batch fetcher available, queue for batch fetch
        3. Otherwise, fetch immediately from DataHub
        4. If no DataHub graph, cache as None
        """
```

While that may be nice, the caller is expeting immediate result, what's the value of the request being queued and being resolved later?



IMO a small improvement we could do is by trying all casing strategies at once, instead of the fallback strategy, which may result on 3 datahub api calls

```
    def resolve_table(self, table: _TableName) -> Tuple[str, Optional[SchemaInfo]]:
        urn = self.get_urn_for_table(table)

        schema_info = self._resolve_schema_info(urn)
        if schema_info:
            self._track_cache_hit()
            return urn, schema_info

        urn_lower = self.get_urn_for_table(table, lower=True)
        if urn_lower != urn:
            schema_info = self._resolve_schema_info(urn_lower)
            if schema_info:
                self._track_cache_hit()
                return urn_lower, schema_info

        # Our treatment of platform instances when lowercasing urns
        # is inconsistent. In some places (e.g. Snowflake), we lowercase
        # the table names but not the platform instance. In other places
        # (e.g. Databricks), we lowercase everything because it happens
        # via the automatic lowercasing helper.
        # See https://github.com/datahub-project/datahub/pull/8928.
        # While we have this sort of inconsistency, we should also
        # check the mixed case urn, as a last resort.
        urn_mixed = self.get_urn_for_table(table, lower=True, mixed=True)
        if urn_mixed not in {urn, urn_lower}:
            schema_info = self._resolve_schema_info(urn_mixed)
            if schema_info:
                self._track_cache_hit()
                return urn_mixed, schema_info

        logger.debug(
            f"Schema resolution failed for table {table}. Tried URNs: "
            f"primary={urn}, lower={urn_lower}, mixed={urn_mixed}"
        )
        self._track_cache_miss()

        if self._prefers_urn_lower():
            return urn_lower, None
        else:
            return urn, None
```

WDYT?

metadata-ingestion/src/datahub/sql_parsing/sql_parsing_aggregator.py
```diff
@@ -197,10 +201,25 @@ def get_subjects(
                     )
         return list(query_subject_urns)
 
-    def make_query_properties(self) -> models.QueryPropertiesClass:
+    def get_formatted_query(self, platform: str, format_queries: bool) -> str:
```
sgomezvillamor, line 204: we have a format query cache

```
@functools.lru_cache(maxsize=FORMAT_QUERY_CACHE_SIZE)
def try_format_query(
    expression: sqlglot.exp.ExpOrStr, platform: DialectOrStr, raises: bool = False
) -> str:
```

but usually, minimum hits here

```
        "format_query_cache_stats": {
          "hits": 0,
          "misses": 2955,
          "maxsize": 1000,
          "currsize": 1000
        },
```

I guess the access pattern is not cache-friendly as well as we have tiny caches.

Maybe you can test this lazy format approach on top of this PR to get some numbers https://github.com/datahub-project/datahub/pull/15953 Otherwise, we cannot blindly get this in and pray 😅

metadata-ingestion/src/datahub/sql_parsing/sql_parsing_aggregator.py
```diff
@@ -937,16 +959,19 @@ def add_preparsed_query(
...
-        formatted_query = self._maybe_format_query(parsed.query_text)
-
         # Register the query's usage.
         if not self._usage_aggregator:
             pass  # usage is not enabled
         elif parsed.timestamp is None:
             self.report.usage_skipped_missing_timestamp += 1
         else:
             upstream_fields = parsed.column_usage or {}
+            usage_query_text = (
```
sgomezvillamor, line 969: we removed the `try_format_query` few lines above and we add it here ?!

instead , should usage also work with raw query and left formatting to the very last?

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
```diff
@@ -176,7 +176,7 @@ def _table_name_from_sqlglot_table(
 
 Urn = str
 
-SQL_PARSE_RESULT_CACHE_SIZE = 1000
+SQL_PARSE_RESULT_CACHE_SIZE = 100000
```
sgomezvillamor: before trying new strategies with lazy formatting, could we make this configruable via env var and test with larger cache sizes?

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_audit.py
acrylJonny: Fair point - I've made it simpler now

gabe-lyons:
@acrylJonny I'm going to close this as @sgomezvillamor did some more in-depth testing and either incorporated these changes or discarded ones that did not provide a performance improvement.

acrylJonny:
> @acrylJonny I'm going to close this as @sgomezvillamor did some more in-depth testing and either incorporated these changes or discarded ones that did not provide a performance improvement.

All good. There have been other itemized PRs introducing the changes.

PR #15936, opened 2026-01-20
feat(ingest/config_loader): support HTTP urls with username/password

ligfx:


metadata-ingestion/src/datahub/configuration/config_loader.py
```diff
@@ -229,7 +229,12 @@ def load_config_file(
         ):  # URLs will return http/https
             # If the URL is remote, we need to fetch it.
             try:
-                response = requests.get(str(config_file))
+                response = requests.get(
```
treff7es, line 232: The code doesn't call response.raise_for_status(), so HTTP error responses (401, 403, 404, 500) will silently 
  succeed and the error page HTML/JSON will be parsed as configuration.

metadata-ingestion/src/datahub/configuration/config_loader.py
```diff
@@ -229,7 +229,12 @@ def load_config_file(
...
+                    str(config_file),
+                    auth=(url_parsed.username or "", url_parsed.password or "")
```
treff7es: urllib.parse.urlparse() returns percent-encoded strings. A URL like                                           
  https://user%40domain:p%40ss@host/config.yaml will send literal user%40domain instead of user@domain. 

```
  from urllib.parse import unquote                                                                              
  username = unquote(url_parsed.username) if url_parsed.username else None                                      
  password = unquote(url_parsed.password) if url_parsed.password else None                                      
```

treff7es (approved):

metadata-ingestion/src/datahub/configuration/config_loader.py
ligfx: Separate issue, handled in #15935

PR #15937, opened 2026-01-20
fix(CLI): add complex types to schemas in dataset upload

kevinkarchacryl:
When you do `datahub dataset get --urn "urn:li:dataset:(urn:li:dataPlatform:hive,example_table,PROD)" --to-file dataset.yaml` your schema fields might have complex types like

  - id: '[version=2.0].[type=struct].[type=array].[type=string].myField'
    type: array
    nativeDataType: array<string>

But when you try to upload the unchanged dataset you get the error 
`Update failed for id <built-in function id>. due to Type array is not a valid primitive type `

This PR adds support for any type that can be on a schema field.

### DIFF
--- a/metadata-ingestion/src/datahub/api/entities/dataset/dataset.py
+++ b/metadata-ingestion/src/datahub/api/entities/dataset/dataset.py
@@ -258,7 +258,10 @@ class SchemaFieldSpecification(StrictModel):
             "timestamp": models.TimeTypeClass,
         }
 
-        type_lower = self.type.lower() if self.type else self.type
+        if not self.type:
+            raise ValueError("Type cannot be None or empty")
+
+        type_lower = self.type.lower()
 
         if (
             type_lower not in set(get_args(PrimitiveType))


metadata-ingestion/src/datahub/api/entities/dataset/dataset.py
```diff
@@ -249,7 +258,21 @@ def get_datahub_type(self) -> models.SchemaFieldDataTypeClass:
...
+            return models.SchemaFieldDataTypeClass(type=models.MapTypeClass())
+        elif type == "union":
+            return models.SchemaFieldDataTypeClass(type=models.UnionTypeClass())
+        elif type == "record":
+            return models.SchemaFieldDataTypeClass(type=models.RecordTypeClass())
+
+        elif type == "date":
+            return models.SchemaFieldDataTypeClass(type=models.DateTypeClass())
+        elif type in ["time", "timestamp"]:
+            return models.SchemaFieldDataTypeClass(type=models.TimeTypeClass())
```
skrydal: Could we apply small refactor here to use a dictionary instead of if/elif/else like that? Not a strong opinion as this is just following existing pattern.

metadata-ingestion/tests/unit/cli/dataset/test_dataset_cmd.py
```diff
@@ -80,6 +80,34 @@ def malformed_yaml_file():
...
+  platform: hive
+  schema:
+    fields:
+    - id: ip
+      type: array
+      description: The IP address
+    - id:  user_id     # Extra spaces
+      type: time
+      description:    The user ID   # Extra spaces
+    """
```
skrydal, line 98: This could be saved as a file in tests folder to avoid the whole writing logic.

metadata-ingestion/tests/unit/cli/dataset/test_dataset_cmd.py
```diff
@@ -353,7 +381,25 @@ def test_run_upsert_fail(self, mock_get_default_graph, invalid_value_yaml_file):
...
+        mock_get_default_graph.return_value.__enter__.return_value = mock_graph
+
+        runner = CliRunner()
+        result = runner.invoke(dataset, ["upsert", "-f", str(complex_type_yaml_file)])
+
+        assert result.exit_code == 0, f"Expected success but got: {result.output}"
+
+        assert mock_graph.emit.called, "Expected emit to be called for dataset upsert"
+
+        assert "Update succeeded" in result.output
```
skrydal, line 408: It is a nice test, I wonder though whether we could try to see what MCPs were sent by the CliRunner, to make sure metadata has proper format. Could be even done via golden files.

metadata-ingestion/tests/unit/cli/dataset/test_dataset_cmd.py
kevinkarchacryl: I think I'd rather stick with the pattern for this one

kevinkarchacryl:
@skrydal made the requested changes

metadata-ingestion/tests/unit/cli/dataset/test_dataset_cmd.py
```diff
@@ -353,7 +381,34 @@ def test_run_upsert_fail(self, mock_get_default_graph, invalid_value_yaml_file):
...
+        mock_graph.exists.return_value = True
+        mock_get_default_graph.return_value.__enter__.return_value = mock_graph
+
+        emitted_items = []
+
+        def capture_emit(item, *args, **kwargs):
+            emitted_items.append(item)
+            return None
+
+        mock_graph.emit.side_effect = capture_emit
```
skrydal, line 401: Interesting. Typically it is retrieved via `call_args_list` field, this one works too, though.

skrydal (approved):
LGTM

PR #15939, opened 2026-01-21
Normalize dataset names by trimming leading/trailing whitespace in DatasetUrn

relaxedboi:
This PR updates DatasetUrn to normalize dataset names by removing leading and trailing whitespace. This ensures more consistent URN creation, comparison, and serialization across the system.


david-leifker:
Sorry for the late response on this — thank you for the contribution and for adding tests around the DB2 padded-name case.

We're going to take a different approach than trimming inside `DatasetUrn`. Dataset URNs are identity (primary key). Silently rewriting `"  foo  "` to `"foo"` in the constructor would change which entity a write lands on, and most GMS paths parse with `Urn.createFromString` rather than `DatasetUrn`, so the trim would also be inconsistent across APIs. Because this has not been a common production issue, we don't think a data migration is needed.

**Suggested alternative:** validate component whitespace in `UrnValidationUtil` (so GraphQL / OpenAPI / Rest.li all get the same behavior), using the existing `strict` / warn path:

- Today, `validateUrn(..., strict)` already `log.warn`s (when `STRICT_URN_VALIDATION_ENABLED` is unset/false) for illegal component characters and commas in simple URNs, and only throws when strict is on.
- Leading/trailing whitespace is currently checked only on the **full URN string**, so internal padding like `urn:li:dataset:(urn:li:dataPlatform:hive,  table  ,PROD)` is not caught.
- Extend component validation to detect leading/trailing whitespace on URN key parts. Warn by default; throw when `strict` is true. Do **not** mutate the URN.

That surfaces padded names in logs without splitting or merging entities. Sources that emit padded CHAR names (e.g. DB2) can still trim at ingest if they want a clean URN.

Happy to review a follow-up PR along those lines. Closing this one in favor of that approach.

PR #15940, opened 2026-01-21
feat(dynamodb): Add support for ingesting AWS resource tags from DynamoDB tables

aviraj-gour:
linear issue: ING-1360
This PR adds support for extracting and ingesting AWS resource tags from DynamoDB tables during metadata ingestion.

The logic in this PR follows a simple overwrite behavior. More advanced handling will be introduced via transformations in a future iteration.

Once the Transformer PR is available, the transformer will be pluggable, and the separation between AWS-sourced tags and UI-managed tags will be configurable through the ingestion recipe.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/autogenerated/capability_summary.json
+++ b/metadata-ingestion/src/datahub/ingestion/autogenerated/capability_summary.json
@@ -1,5 +1,5 @@
 {
-  "generated_at": "2026-02-01T19:48:54.520487+00:00",
+  "generated_at": "2026-02-03T05:09:29.378106+00:00",
   "generated_by": "metadata-ingestion/scripts/capability_summary.py",
   "plugin_details": {
     "abs": {


metadata-ingestion/src/datahub/ingestion/source/dynamodb/dynamodb.py
```diff
@@ -114,6 +118,12 @@ class DynamoDBConfig(
         default=AllowDenyPattern.allow_all(),
         description="Regex patterns for tables to filter in ingestion. The table name format is 'region.table'",
     )
+    extract_table_tags: Optional[bool] = Field(
```
treff7es:
```suggestion
    extract_table_tags:bool = Field(
```

Here I don't see when the None would make sense

metadata-ingestion/src/datahub/ingestion/source/dynamodb/dynamodb.py
```diff
@@ -584,3 +608,66 @@ def _get_domain_wu(
...
+                        entity_urn=dataset_urn, aspect_type=GlobalTagsClass
+                    )
+                    if current_tags and getattr(current_tags, "tags", None):
+                        tags_to_add.extend(
+                            tag_assoc.tag for tag_assoc in current_tags.tags
+                        )
+                except Exception as merge_err:
+                    self.report.report_warning(
+                        title="DynamoDB Tags",
+                        message="Failed to merge existing tags; proceeding with AWS tags only",
```
treff7es:
```suggestion
                        message="Failed to merge existing tags; proceeding with AWS tags only. Manual tags may be lost",
```

metadata-ingestion/src/datahub/ingestion/source/dynamodb/dynamodb.py
```diff
@@ -584,3 +608,66 @@ def _get_domain_wu(
...
+        """
+        try:
+            tags_kv = self._get_dynamodb_table_tags(
+                dynamodb_client=dynamodb_client,
+                table_arn=table_arn,
+            )
+            tags_to_add: List[str] = []
+
+            for tag_dict in tags_kv:
+                key, val = tag_dict["Key"], tag_dict["Value"]
```
treff7es: tag_dict["Key"] and tag_dict["Value"] access without validation. If AWS API response format changes, 
  this will raise KeyError masked as generic failure.

Use tag_dict.get("Key") with validation, or add try-catch with specific error message

metadata-ingestion/src/datahub/ingestion/source/dynamodb/dynamodb.py
```diff
@@ -584,3 +608,66 @@ def _get_domain_wu(
...
+                tag = f"{key}:{val}" if val else key
+                tags_to_add.append(make_tag_urn(tag))
+
+            # Merge in existing tags to avoid clobbering manually added tags when graph is available.
+            if self.ctx.graph:
+                try:
+                    current_tags = self.ctx.graph.get_aspect(
+                        entity_urn=dataset_urn, aspect_type=GlobalTagsClass
+                    )
+                    if current_tags and getattr(current_tags, "tags", None):
```
treff7es: getattr(current_tags, "tags", None) is overly defensive. GlobalTagsClass always has .tags attribute.


```suggestion
                    if current_tags and current_tags.tags:
```

metadata-ingestion/src/datahub/ingestion/source/dynamodb/dynamodb.py
```diff
@@ -584,3 +608,66 @@ def _get_domain_wu(
...
+        merge with existing tags (including manual UI tags) when a graph client is available.
+        """
+        try:
+            tags_kv = self._get_dynamodb_table_tags(
+                dynamodb_client=dynamodb_client,
+                table_arn=table_arn,
+            )
+            tags_to_add: List[str] = []
+
+            for tag_dict in tags_kv:
```
treff7es: This can be a list comprehension here:
```
  tag_urns = [                                                                                                  
      make_tag_urn(f"{t['Key']}:{t['Value']}" if t['Value'] else t['Key'])                                      
      for t in tags_kv                                                                                          
  ]                                                                                                             
```

treff7es:
It would be nice to have tests for edge cases, such as special characters and key-only tags.

metadata-ingestion/src/datahub/ingestion/source/dynamodb/dynamodb.py
aviraj-gour: removed optional, thanks.

metadata-ingestion/src/datahub/ingestion/source/dynamodb/dynamodb.py
aviraj-gour: updated with report_warning if key not available

metadata-ingestion/src/datahub/ingestion/source/dynamodb/dynamodb.py
aviraj-gour: added a logs so it would be better without comprehension

metadata-ingestion/src/datahub/ingestion/source/dynamodb/dynamodb.py
```diff
@@ -584,3 +608,76 @@ def _get_domain_wu(
...
+                    if current_tags and current_tags.tags:
+                        tags_to_add.extend(
+                            tag_assoc.tag for tag_assoc in current_tags.tags
+                        )
+                except Exception as merge_err:
+                    self.report.report_warning(
+                        title="DynamoDB Tags",
+                        message="Failed to merge existing tags; proceeding with AWS tags only, Manual tags may be lost.",
+                        context=f"dataset_urn: {dataset_urn}; error={merge_err}",
+                    )
```
ligfx: Is it possible to send these tags as a patch update rather than requiring a graph client?

metadata-ingestion/src/datahub/ingestion/source/dynamodb/dynamodb.py
```diff
@@ -114,6 +118,12 @@ class DynamoDBConfig(
         default=AllowDenyPattern.allow_all(),
         description="Regex patterns for tables to filter in ingestion. The table name format is 'region.table'",
     )
+    extract_table_tags: bool = Field(
+        default=False,
+        description=(
+            "When enabled, extracts AWS resource tags for DynamoDB tables and applies them as DataHub tags, AWS table tags may reappear if they are deleted and manual tags will lost if datahub api not available"
```
kyungsoo-datahub: nitpick: will lost -> will be lost , datahub api  -> DataHub API

metadata-ingestion/src/datahub/ingestion/source/dynamodb/dynamodb.py
kyungsoo-datahub: +1. PATCH is possible and would be a cleaner approach unless there is a special reason to use UPSERT. It would eliminate the graph client dependency.

aviraj-gour: @kyungsoo-datahub @ligfx when using patch method to update a tag, getting into this problem, [slack](https://datahub.slack.com/archives/C09SXGVD605/p1769509920287379)

aviraj-gour: @treff7es @kyungsoo-datahub @ligfx Reverting back to graph client approach, with patch I am not able to update descriptions, please refer this [slack](https://datahub.slack.com/archives/C02ENS87APK/p1769512278093429)

metadata-ingestion/src/datahub/ingestion/source/dynamodb/dynamodb.py
```diff
@@ -584,3 +608,47 @@ def _get_domain_wu(
...
+                entity_urn=dataset_urn,
+                tags=tags,
+            )
+
+        except Exception as e:
+            self.report.report_warning(
+                title="DynamoDB Tags",
+                message="Failed to extract tags for table.",
+                context=f"dataset_urn: {dataset_urn}; error={e}",
+            )
```
treff7es: self.report.warning(
                title="DynamoDB Tags",
                message="Failed to extract tags for table.",
                context=f"dataset_urn: {dataset_urn}; error={e}",
                exc=e,
            )

metadata-ingestion/src/datahub/ingestion/source/dynamodb/dynamodb.py
```diff
@@ -584,3 +608,47 @@ def _get_domain_wu(
...
+                entity_type="dataset",
+                entity_urn=dataset_urn,
+                tags=tags,
+            )
+
+        except Exception as e:
+            self.report.report_warning(
+                title="DynamoDB Tags",
+                message="Failed to extract tags for table.",
+                context=f"dataset_urn: {dataset_urn}; error={e}",
```
treff7es, line 653: Why can this failure happen? 
If in most cases this is due to permission, then I think itis  worth adding help to the user on how to fix it, like:
`This may be due to missing 'dynamodb:ListTagsOfResource' IAM permission.`

metadata-ingestion/src/datahub/ingestion/source/dynamodb/dynamodb.py
```diff
@@ -584,3 +608,47 @@ def _get_domain_wu(
...
+        try:
+            aws_tags_kv = self._get_dynamodb_table_tags(
+                dynamodb_client=dynamodb_client,
+                table_arn=table_arn,
+            )
+
+            tags = [
+                f"{tag_dict['Key']}:{tag_dict['Value']}"
+                if tag_dict.get("Value")
+                else f"{tag_dict['Key']}"
```
treff7es, line 637: Uses tag_dict['Key'] directly without checking if Key exists, while using .get("Value") for Value -  inconsistent defensive coding.
Consider adding `if tag_dict.get("Key")` filter or use consistent access pattern.

metadata-ingestion/src/datahub/ingestion/source/dynamodb/dynamodb.py
```diff
@@ -187,6 +200,7 @@ class DynamoDBSource(StatefulIngestionSourceBase):
 
     def __init__(self, ctx: PipelineContext, config: DynamoDBConfig, platform: str):
         super().__init__(config, ctx)
+        self.ctx = ctx
```
treff7es: Isn't self.ctx is already set by StatefulIngestionSourceBase.__init__

metadata-ingestion/tests/unit/dynamodb/test_dynamodb.py
```diff
@@ -0,0 +1,181 @@
...
+            "table_arn": "arn:aws:dynamodb:us-west-2:123456789012:table/test_table",
+        }
+
+    @staticmethod
+    def create_dynamodb_source(ctx, config):
+        """Helper method to create DynamoDB source instance."""
+        return DynamoDBSource(ctx=ctx, config=config, platform="dynamodb")
+
+    @staticmethod
+    def get_tag_urns_from_workunits(workunits):
```
treff7es, line 65: get_tag_urns_from_workunits() contains assertions that should be in test methods. Move assertions to test methods or add descriptive messages

metadata-ingestion/src/datahub/ingestion/source/dynamodb/dynamodb.py
aviraj-gour: For some reason type hints were not working, removed it not part of my current changes

metadata-ingestion/src/datahub/ingestion/source/dynamodb/dynamodb.py
aviraj-gour: Thanks, missed it,
added if statement

metadata-ingestion/tests/unit/dynamodb/test_dynamodb.py
aviraj-gour: removed assert from method.

PR #15945, opened 2026-01-22
fix(bigquery): Optimize dataset filtering to reduce unnecessary API calls

rajatoss:
### Summary

This PR optimizes the BigQuery ingestion source by applying `dataset_pattern` filters earlier in the pipeline, significantly reducing unnecessary API calls and improving ingestion performance for large BigQuery deployments.

### Problem

Previously, the BigQuery source:
1. Fetched **all** datasets in a project via `list_datasets()` API
2. Made individual `get_dataset()` API calls for **each** dataset to fetch metadata (description, created, modified)
3. Only then applied `dataset_pattern` filters to exclude unwanted datasets

This was inefficient for users with many datasets who only want to ingest a subset. For example, a project with 100 datasets where only 5 match the `dataset_pattern` would still make 101 API calls (1 list + 100 individual get calls) before filtering.

### Solution

The fix applies filters at two points:

1. **Early filtering in `get_datasets_for_project_id()`**: Added a `dataset_filter` callback parameter that filters datasets immediately after listing, before fetching any metadata. Datasets that don't match the pattern are skipped entirely.

2. **Batch metadata fetching via INFORMATION_SCHEMA**: Instead of calling `get_dataset()` individually for each dataset, metadata is now fetched in batch using `INFORMATION_SCHEMA.SCHEMATA` queries grouped by location. This significantly reduces API calls.

### Changes

| File | Change |
|------|--------|
| `bigquery_schema.py` | Added `dataset_filter` parameter for early filtering; batch metadata fetch via INFORMATION_SCHEMA |
| `common.py` | Added `is_dataset_allowed()` method to `BigQueryFilter` class |
| `bigquery_schema_gen.py` | Added `filters` parameter; passes `dataset_filter` callback when fetching datasets |
| `lineage.py` | Added `filters` parameter; uses early dataset filtering; removed redundant pattern check |
| `bigquery.py` | Passes `filters` to `BigqueryLineageExtractor` and `BigQuerySchemaGenerator` |
| `bigquery_test_connection.py` | Passes `filters` to `BigqueryLineageExtractor` |

### Performance Impact

| Scenario | Before | After |
|----------|--------|-------|
| 100 datasets, 5 match filter | 101 API calls | ~2-3 API calls (1 list + batch queries per location) |
| 1000 datasets, 50 match filter | 1001 API calls | ~2-5 API calls |

### Checklist
- [x] Links to related issues: ING-1342
- [x] Tests updated/added where appropriate
- [x] Documentation updated in `docs/how/updating-datahub.md`

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_schema.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_schema.py
@@ -302,34 +302,35 @@ class BigQuerySchemaApi:
                 self.bq_client.list_datasets(project_id, max_results=maxResults)
             )
 
-            filtered_datasets: List[BigqueryDataset] = []
-            for d in datasets:
-                if dataset_filter is not None and not dataset_filter(d.dataset_id):
-                    logger.debug(
-                        f"Skipping dataset {project_id}.{d.dataset_id} due to dataset_pattern filter"
-                    )
-                    continue
-
-                location = (
-                    d._properties.get("location")
-                    if hasattr(d, "_properties") and isinstance(d._properties, dict)
-                    else None
+        filtered_datasets: List[BigqueryDataset] = []
+        for d in datasets:
+            if dataset_filter is not None and not dataset_filter(d.dataset_id):
+                logger.debug(
+                    f"Skipping dataset {project_id}.{d.dataset_id} due to dataset_pattern filter"
                 )
-                filtered_datasets.append(
-                    BigqueryDataset(
-                        name=d.dataset_id,
-                        location=location,
-                        labels=d.labels,
-                    )
+                continue
+
+            location = (
+                d._properties.get("location")
+                if hasattr(d, "_properties") and isinstance(d._properties, dict)
+                else None
+            )
+            filtered_datasets.append(
+                BigqueryDataset(
+                    name=d.dataset_id,
+                    location=location,
+                    labels=d.labels,
                 )
+            )
 
-            if not filtered_datasets:
-                return []
+        if not filtered_datasets:
+            return []
 
-            # Batch fetch metadata (description, created, modified) per location
+        # Batch fetch metadata (description, created, modified) per location
+        with self.report.enrich_datasets_timer:
             self._enrich_datasets_with_metadata(project_id, filtered_datasets)
 
-            return filtered_datasets
+        return filtered_datasets
 
     def _enrich_datasets_with_metadata(
         self, project_id: str, datasets: List[BigqueryDataset]
--- a/metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_report.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_report.py
@@ -34,6 +34,7 @@ class BigQuerySchemaApiPerfReport(Report):
     list_projects_timer: PerfTimer = field(default_factory=PerfTimer)
     list_projects_with_labels_timer: PerfTimer = field(default_factory=PerfTimer)
     list_datasets_timer: PerfTimer = field(default_factory=PerfTimer)
+    enrich_datasets_timer: PerfTimer = field(default_factory=PerfTimer)
 
     get_columns_for_dataset_sec: float = 0
     get_tables_for_dataset_sec: float = 0


metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_schema.py
```diff
@@ -278,13 +278,22 @@ def get_projects_with_labels(self, labels: FrozenSet[str]) -> List[BigqueryProje
...
         with self.report.list_datasets_timer:
             self.report.num_list_datasets_api_requests += 1
             datasets = self.bq_client.list_datasets(project_id, max_results=maxResults)
             result = []
             for d in datasets:
+                if dataset_filter is not None and not dataset_filter(d.dataset_id):
+                    logger.debug(
+                        f"Skipping dataset {project_id}.{d.dataset_id} due to dataset_pattern filter"
+                    )
+                    continue
```
sgomezvillamor: We introduced individual `self.bq_client.get_dataset` calls to fetch the dataset description, creation time, and last modified time, which are not provided in the `list_datasets` call. However, doing this one by one is quite inefficient, as the comment in the comment points out.

Being able to filter datasets upfront helps mitigate this.

Could you split the PR into two so we can get this part merged ASAP? This may have significant impact on those cases where dataset pattern is being used and are very selective.

Minor comment here:

```
dataset_filter=lambda dataset_name: self.filters.is_dataset_allowed(
    dataset_name, project_id
),
```

Since all parameters are strings, it’s easy to mix them up. Please use named arguments to make it clearer.

rajatoss: done

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_schema.py
```diff
@@ -278,59 +291,94 @@ def get_projects_with_labels(self, labels: FrozenSet[str]) -> List[BigqueryProje
...
-    # This is not used anywhere
-    def get_datasets_for_project_id_with_information_schema(
-        self, project_id: str
-    ) -> List[BigqueryDataset]:
+            if not filtered_datasets:
+                return []
+
+            # Batch fetch metadata (description, created, modified) per location
+            self._enrich_datasets_with_metadata(project_id, filtered_datasets)
+
```
sgomezvillamor: we could wrap this call with a timer separately from the list datasets one

```
        with self.report.list_datasets_timer:
            ...
            datasets = list(
                self.bq_client.list_datasets(project_id, max_results=maxResults)
            )

        with self.report.enrich_datasets_timer:
            ...
            self._enrich_datasets_with_metadata(...)
```

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_schema.py
```diff
@@ -278,59 +291,94 @@ def get_projects_with_labels(self, labels: FrozenSet[str]) -> List[BigqueryProje
...
+                self.bq_client.list_datasets(project_id, max_results=maxResults)
+            )
+
+            filtered_datasets: List[BigqueryDataset] = []
             for d in datasets:
-                # TODO: Fetch dataset description individually impacts overall performance if the number of datasets is high (hundreds); instead we should fetch in batch for all datasets.
-                # https://cloud.google.com/python/docs/reference/bigquery/latest/google.cloud.bigquery.client.Client#google_cloud_bigquery_client_Client_get_dataset
-                # https://cloud.google.com/python/docs/reference/bigquery/latest/google.cloud.bigquery.dataset.Dataset
-                dataset = self.bq_client.get_dataset(d.reference)
+                if dataset_filter is not None and not dataset_filter(d.dataset_id):
```
sgomezvillamor: double check if we have report counters about filtered datasets that we should increase now here

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_schema_gen.py
```diff
@@ -374,8 +380,10 @@ def _process_project(
                 )
             else:
                 action_mesage = (
-                    "Does your service account have `bigquery.datasets.get` permission ? "
-                    "Assign predefined role `roles/bigquery.metadataViewer` to your service account."
+                    "Does your service account have `bigquery.datasets.get` and "
```
sgomezvillamor, line 428: may we add `INFORMATION_SCHEMA.SCHEMATA` as a new requirement in https://docs.datahub.com/docs/generated/ingestion/sources/bigquery#prerequisites?
or is it granted with the required permissions we list there?

rajatoss: already granted with the required permission we list.

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_schema.py
rajatoss: we don't have a report counter about filtered datasets.

PR #15949, opened 2026-01-22
(fix: auth) Ignore case when checking policy conditions (closes #15948)

### LINKED ISSUE #15948: The dataFlow type cannot be configured through Policies UI
**Describe the bug**
When adding `Data Pipeline` to the list of object types that this policy is applicable to, the value in the filter is equal to `dataFlow`. But when authorization request is handled, the `EntitySpec` contains the `dataflow` constant, thus preventing aforementioned policy to be applied 

**To Reproduce**

1. Go to Policies UI
2. Click on `Create new policy`
3. Add `Data Pipeline` object type to the list of applicable object types `Resource Type`
4. Add `Edit Entity` privilege
5. Save the policy
6. Try to create new `dataFlow`
7. See error `401`

**Expected behavior**
The policy should be applied to the `dataFlow` correctly

- OS: MacOS
- Browser: Chrome 143.0.7499.170

gallyamb:
Tried to address related issue. Looking forward for community opinion

The approach in the PR resolves this issue and any potential future issues with inconsistent object types' literal cases

#15948

### DIFF
--- a/metadata-service/auth-impl/src/main/java/com/datahub/authorization/PolicyEngine.java
+++ b/metadata-service/auth-impl/src/main/java/com/datahub/authorization/PolicyEngine.java
@@ -297,7 +297,7 @@ public class PolicyEngine {
       case EQUALS:
         return filterValues.stream()
             .anyMatch(
-                expectedValue -> filterValues.stream().anyMatch(expectedValue::equalsIgnoreCase));
+                expectedValue -> fieldValues.stream().anyMatch(expectedValue::equalsIgnoreCase));
       case STARTS_WITH:
         return fieldValues.stream()
             .anyMatch(
@@ -307,7 +307,7 @@ public class PolicyEngine {
       case NOT_EQUALS:
         return filterValues.stream()
             .noneMatch(
-                expectedValue -> filterValues.stream().anyMatch(expectedValue::equalsIgnoreCase));
+                expectedValue -> fieldValues.stream().anyMatch(expectedValue::equalsIgnoreCase));
       default:
         log.error("Unsupported condition {}", condition);
         return false;


gallyamb:
I'll check this again and come back after weekends. Thank you!

gallyamb:
@RyanHolstien checked the urn, it's constructed with the following pattern `urn:li:dataFlow:(airflow,{instance_name}.{dag_name},prod)`, so yep, `dataFlow` type name spelled correctly

What is the possible reason then for the initial buggy behavior? Where I need to look? Any ideas are appreciated, I'll debug it more thoroughly

RyanHolstien:
> @RyanHolstien checked the urn, it's constructed with the following pattern `urn:li:dataFlow:(airflow,{instance_name}.{dag_name},prod)`, so yep, `dataFlow` type name spelled correctly
> 
> What is the possible reason then for the initial buggy behavior? Where I need to look? Any ideas are appreciated, I'll debug it more thoroughly

Would need to debug in DataFlowType.update to see how the EntitySpec is getting created. Should be using PoliciesConfig.DATA_FLOW_PRIVILEGES.getResourceType() which should be "dataFlow" and then matching on the urn's entity type.

gallyamb:
Oh, that's crazy. The reason why this request does not work for us is because the openapi v3 scheme is [generated](https://github.com/datahub-project/datahub/blob/55817d14b842818131fd5ccb1eb80a74378ac9cc/metadata-service/openapi-servlet/src/main/java/io/datahubproject/openapi/v3/OpenAPIV3Generator.java#L224) with lowercasing the entity type name. But this path parameter is [taken as is](https://github.com/datahub-project/datahub/blob/55817d14b842818131fd5ccb1eb80a74378ac9cc/metadata-service/openapi-servlet/src/main/java/io/datahubproject/openapi/controller/GenericEntitiesController.java#L550) while authorizing the request. Then we just used the declared path, that led us to the authorization error

It's not a blocker, because even the openapi scheme is generated incorrectly, the request with correct entity type name is successfully handled by the GMS

Closing this PR

PR #15951, opened 2026-01-22
fix(frontend): prevent open redirect in redirectTrailingSlash

david-leifker:
# Fix: Open Redirect in redirectTrailingSlash

## Issue

The `redirectTrailingSlash()` method in `Application.java` did not sanitize the `path` parameter before constructing the redirect URL. This allowed URLs with leading slashes (e.g., `//external.com/`) to produce protocol-relative redirects.

## Changes

**File:** `datahub-frontend/app/controllers/Application.java`

- Strip leading slashes from path parameter
- Normalize `basePath="/"` to empty string to prevent double-slash in redirect URL
- Handle null/empty path cases

## Before & After

| Input | Before | After |
|-------|--------|-------|
| `//example.com/` | `//example.com` | `/example.com` |
| `///path/` | `///path` | `/path` |
| `test/` | `/test` | `/test` |

### DIFF
--- a/datahub-frontend/app/controllers/Application.java
+++ b/datahub-frontend/app/controllers/Application.java
@@ -130,8 +130,16 @@ public class Application extends Controller {
    */
   @Nonnull
   public Result redirectTrailingSlash(@Nullable String path) {
+    String redirectBase = basePath.equals("/") ? "" : basePath;
 
-    return movedPermanently("/" + path);
+    if (path == null || path.isEmpty()) {
+      return movedPermanently(redirectBase + "/");
+    }
+    String sanitizedPath = path.replaceFirst("^/+", "");
+    if (sanitizedPath.isEmpty()) {
+      return movedPermanently(redirectBase + "/");
+    }
+    return movedPermanently(redirectBase + "/" + sanitizedPath);
   }
 
   /**
--- a/datahub-frontend/test/app/ApplicationTest.java
+++ b/datahub-frontend/test/app/ApplicationTest.java
@@ -622,6 +622,64 @@ public class ApplicationTest extends WithBrowser {
   }
 
   @Test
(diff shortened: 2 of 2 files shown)


datahub-frontend/app/controllers/Application.java
```diff
@@ -130,8 +130,19 @@ public Result index(@Nullable String path) {
...
+      return movedPermanently(redirectBase + "/");
+    }
+    String sanitizedPath = path;
+    while (sanitizedPath.startsWith("/")) {
+      sanitizedPath = sanitizedPath.substring(1);
+    }
+    if (sanitizedPath.isEmpty()) {
+      return movedPermanently(redirectBase + "/");
+    }
+    return movedPermanently(redirectBase + "/" + sanitizedPath);
```
esteban, line 142: maybe replace with somethinglike `replaceFirst("^/+", "")` instead?

esteban (approved):
Just comment

datahub-frontend/app/controllers/Application.java
```diff
@@ -130,8 +130,19 @@ public Result index(@Nullable String path) {
...
   @Nonnull
   public Result redirectTrailingSlash(@Nullable String path) {
+    String redirectBase = basePath.equals("/") ? "" : basePath;
 
-    return movedPermanently("/" + path);
+    if (path == null || path.isEmpty()) {
+      return movedPermanently(redirectBase + "/");
+    }
+    String sanitizedPath = path;
+    while (sanitizedPath.startsWith("/")) {
```
RyanHolstien: Can cause a lot of loop iterations if someone throws in a ton of forward slashes, but I guess it's bound to 2k so probably not too impactful? Could be covered by a regex though rather than loop.

PR #15953, opened 2026-01-22
feat(sql-aggregator): performance test and micro optimizations

sgomezvillamor:
A profile performance analysis showed that a large portion of the cost comes from `copy` / `deep_copy` calls or costly `.sql` calls. Most of these are required to avoid in-place mutations or happen inside the `sqlglot` library itself, but we identified a few that were unnecessary and removed them.

Additionally, as part of column-level lineage, we were computing join information. Extracting this join data can be expensive, and it’s currently not used anywhere in the platform. Because of that, we added a mechanism to skip it via an environment variable (disabled by default so far).

Beyond these micro-optimizations, maybe the most valuable is the performance testing, which will prevent performance regressions.

**Benchmark**

Initial baseline

```
====================================================================================================
Performance Benchmark Results
====================================================================================================

|   Query Count   |  Elapsed Time (s)  |  Avg Time/Query (ms)   |  Throughput (q/s)  |
| --------------- | ------------------ | ---------------------- | ------------------ |
|             100 |               1.58 |                  15.84 |              63.12 |
|           1,000 |              14.82 |                  14.82 |              67.50 |
|          10,000 |             146.31 |                  14.63 |              68.35 |

====================================================================================================
```

After micro enhancements

```
====================================================================================================
Performance Benchmark Results
====================================================================================================

|   Query Count   |  Elapsed Time (s)  |  Avg Time/Query (ms)   |  Throughput (q/s)  |
| --------------- | ------------------ | ---------------------- | ------------------ |
|             100 |               1.45 |                  14.51 |              68.92 |
|           1,000 |              13.33 |                  13.33 |              75.01 |
|          10,000 |             133.03 |                  13.30 |              75.17 |

====================================================================================================
```

**Summary** 

Based on these benchmarks, the changes result in roughly a 5-10% performance improvement. 

**Performance testing**

Beyond these micro optimizations, a performance test has been introduced to prevent performance regressions.

### DIFF
--- a/metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
+++ b/metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
@@ -32,8 +32,8 @@ from pydantic import field_serializer, field_validator
 
 from datahub.cli.env_utils import get_boolean_env_variable
 from datahub.configuration.env_vars import (
-  get_sql_agg_skip_joins,
-  get_sql_parse_cache_size,
+    get_sql_agg_skip_joins,
+    get_sql_parse_cache_size,
 )
 from datahub.ingestion.graph.client import DataHubGraph
 from datahub.metadata.schema_classes import (
--- a/metadata-ingestion/tests/unit/test_athena_source.py
+++ b/metadata-ingestion/tests/unit/test_athena_source.py
@@ -1135,9 +1135,15 @@ def test_sanitize_identifier_error_handling_in_generate_partition_profiler_query
     assert result == (None, None), "Should return None, None when sanitization fails"
 
     # Verify that a warning log message was generated
-    assert len(caplog.records) == 1
-    log_record = caplog.records[0]
-    assert log_record.levelname == "WARNING"
+    # Filter for WARNING level logs from the athena source (ignore INFO logs from other modules)
+    warning_records = [
+        r for r in caplog.records if r.levelname == "WARNING" and "athena" in r.name
+    ]
+    assert len(warning_records) == 1, (
+        f"Expected 1 WARNING log from athena source, but got {len(warning_records)}. "
+        f"All records: {[(r.levelname, r.name, r.message) for r in caplog.records]}"
+    )
+    log_record = warning_records[0]
     assert (
         "Failed to generate partition profiler query for valid_schema.valid_table due to unsafe identifiers"
         in log_record.message


metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
```diff
@@ -970,13 +971,17 @@ def _column_level_lineage(
...
 
     joins: Optional[List[_JoinInfo]] = None
-    try:
-        # List join clauses.
-        joins = _list_joins(dialect=dialect, root_scope=root_scope)
-        logger.debug("Joins: %s", joins)
-    except Exception as e:
-        # This is a non-fatal error, so we can continue.
-        logger.debug("Failed to list joins: %s", e)
+    if get_sql_agg_skip_joins():
```
treff7es: Won't these info level logs flood our logs on info level?

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
```diff
@@ -970,13 +971,17 @@ def _column_level_lineage(
...
+        logger.info("Skipping join processing in column-level lineage")
+    else:
+        logger.info("Processing join clauses in column-level lineage")
+        try:
+            joins = _list_joins(dialect=dialect, root_scope=root_scope)
+            if logger.getEffectiveLevel() <= logging.DEBUG:
+                logger.debug("Joins: %s", joins)
+        except Exception as e:
+            # This is a non-fatal error, so we can continue.
```
treff7es, line 983: If it is non fatal why we changed it to error?

metadata-ingestion/tests/performance/sql_parsing/test_sql_aggregator.py
```diff
@@ -0,0 +1,257 @@
...
+    print("=" * 100)
+
+    # Basic assertions to ensure tests ran successfully
+    assert len(results) == total_combinations, (
+        f"Expected {total_combinations} results, got {len(results)}"
+    )
+    assert all(r["elapsed_time"] > 0 for r in results), (
+        "All configurations should have positive elapsed time"
+    )
+    assert all(r["throughput"] < MAX_THROUGHPUT_THRESHOLD for r in results), (
```
treff7es: Is this correct? Why is it a problem if I have very fast query processing and it exceeds this threshold? 
This means that if query processing is fast, then this assertion will fail.
Sholdn't this be something like `all(r["throughput"] > MIN_THROUGHPUT_THRESHOLD`

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
```diff
@@ -1358,7 +1366,10 @@ def _extract_select_from_create(
 def _extract_select_from_update(
     statement: sqlglot.exp.Update,
 ) -> sqlglot.exp.Select:
-    statement = statement.copy()
```
treff7es, line 1362: Why don't we just remove this?

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
sgomezvillamor: indeed! I wanted to have some visibility on the setting, but certainly here can be very repetitive

I moved it to sql parsing aggregator constructor in commit 680ea927932563356d670c4102a8bf6ad4d0fe1d

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
sgomezvillamor: Any error in parsing and lineage is non-fatal because we continue anyway
No strong opinion, I set it back to debug in commit 680ea927932563356d670c4102a8bf6ad4d0fe1d

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
sgomezvillamor: It’s mostly “just in case.” I wanted to preserve this detailed knowledge and make it available in case mutations are added in the future, especially with AI agents writing the code.

metadata-ingestion/tests/performance/sql_parsing/test_sql_aggregator.py
sgomezvillamor: when you have an intent and end up implementing the opposite 🤦 

addressed in commit 7cfbbf1f1c0a639e905350c4c2c94464004701bf

thanks for catching!

PR #15957, opened 2026-01-22
feat(ingestion/matillion): matillion connector

acrylJonny:


acrylJonny:
Branch renamed and PR moved to [#15966](https://github.com/datahub-project/datahub/pull/15966)

PR #15958, opened 2026-01-22
fix(docs): remove use of tabs where deeplinking is needed

ligfx:


docs/ui-ingestion.md
```diff
@@ -302,19 +299,13 @@ While the UI-based forms handle most common ingestion scenarios, advanced users
 
 For these advanced use cases, DataHub supports direct YAML recipe configuration. For detailed information about YAML-based configuration, including syntax and examples, see the [Recipe Overview Guide](metadata-ingestion/recipe_overview.md).
 
-<Tabs>
-   <TabItem value="cli" label="CLI">
-
-You can deploy recipes using the CLI as mentioned in the [CLI documentation for uploading ingestion recipes](./cli.md#ingest-deploy).
+You can deploy recipes using the CLI as mentioned in the [CLI documentation for uploading ingestion recipes](./cli.md#ingest-deploy):
```
yoonhyejin: Do we also want to add titles here -- CLI / GraphQL?

ligfx: Hmm, I had decided not to before, but yeah with fresh eyes I think that makes sense. We might want to send people links to these sections.

PR #15966, opened 2026-01-23
feat(ingestion/matillion): matillion connector

acrylJonny:


datahub-web-react/src/app/ingestV2/source/builder/sources.json
```diff
@@ -139,6 +139,16 @@
         "category": "ETL / ELT",
         "isPopular": true
     },
+    {
+        "urn": "urn:li:dataPlatform:matillion-dpc",
```
treff7es: **File:** `datahub-web-react/src/app/ingestV2/source/builder/sources.json`
**Line:** 143

## 🔴 Platform URN inconsistency — UI lookups will not match emitted entities

The platform URN declared here does not match what the connector actually emits, so the frontend logo/recipe lookup won't resolve to ingested entities.

**Emitted by the connector:**
- `metadata-ingestion/src/datahub/ingestion/source/matillion_dpc/constants.py:7` — `MATILLION_PLATFORM = "matillion"`
- `metadata-service/configuration/src/main/resources/bootstrap_mcps/data-platforms.yaml:252` — registers `urn:li:dataPlatform:matillion`

**Declared in the frontend:**
- `datahub-web-react/src/app/ingestV2/source/builder/sources.json:143` — `"urn": "urn:li:dataPlatform:matillion-dpc"`

The `"matillion-dpc"` value is used as the **source type id** (correctly, in `setup.py` and `@platform_name(id="matillion-dpc")`), but the **platform URN** must be the platform identifier, which is `matillion`.

### Fix (pick one consistently)

**Option A (simpler) — keep platform as `matillion`, fix the frontend URN:**
```diff
-     "urn": "urn:li:dataPlatform:matillion-dpc",
+     "urn": "urn:li:dataPlatform:matillion",
      "name": "matillion-dpc",
```

**Option B — rename the platform to `matillion-dpc` everywhere:** update `MATILLION_PLATFORM`, `data-platforms.yaml` (urn, name, displayName entries), all four `DatasetContainerSubTypes`/`FlowContainerSubTypes` constants, golden files (URNs embed the platform), and any docs/recipes that reference it.

Option A is much less invasive. After the fix please regenerate the capability summary / sources.json via the standard regen tasks so downstream consumers see the corrected URN.

metadata-ingestion/examples/recipes/matillion_dpc_to_datahub.dhub.yaml
```diff
@@ -0,0 +1,46 @@
...
+source:
+  type: matillion-dpc
+  config:
+    api_config:
+      client_id: "${MATILLION_CLIENT_ID}"
+      client_secret: "${MATILLION_CLIENT_SECRET}"
+      region: US1 # EU1 or US1
+
+    env: PROD
+    include_lineage: true
```
treff7es: **File:** `metadata-ingestion/examples/recipes/matillion_dpc_to_datahub.dhub.yaml`
**Line:** 11

## 🔴 Example recipes reference config fields that don't exist — first ingestion attempt will fail

`MatillionSourceConfig` sets `model_config = ConfigDict(extra="forbid")` (`config.py:309`). Any key in a user's recipe that isn't a declared field raises a pydantic validation error before ingestion can even start. The current example recipes will all fail this validation.

### Field references in the PR that don't exist in `MatillionSourceConfig`

| Location | Bad field | Status |
|---|---|---|
| `metadata-ingestion/examples/recipes/matillion_dpc_to_datahub.dhub.yaml:11` | `include_lineage: true` | not a config field |
| `metadata-ingestion/examples/recipes/matillion_dpc_to_datahub.dhub.yaml:12` | `include_column_lineage: true` | not a config field |
| `metadata-ingestion/examples/recipes/matillion_dpc_to_datahub.dhub.yaml` (commented) | `project_pattern` / `pipeline_pattern` (singular) | actual fields are plural: `project_patterns` / `pipeline_patterns` |
| `datahub-web-react/src/app/ingestV2/source/builder/sources.json:148` (recipe string) | `include_pipeline_executions: true` | not a config field |
| `metadata-ingestion/src/datahub/ingestion/source/matillion_dpc/matillion.py:111` | `@capability(LINEAGE_COARSE, "... can be disabled via configuration include_lineage")` | references non-existent flag in user-facing capability text |
| `metadata-ingestion/src/datahub/ingestion/source/matillion_dpc/matillion.py:115` | `@capability(LINEAGE_FINE, "... can be disabled via configuration include_column_lineage")` | same |

The capability strings get baked into `metadata-ingestion/src/datahub/ingestion/autogenerated/capability_summary.json` and surface in the integrations docs, so the inaccuracy is user-visible there too.

### Fix — choose one direction

**Option A (recommended) — make the docs match the code.** Replace the non-existent fields with the actual ones the code reads:

```yaml
# examples/recipes/matillion_dpc_to_datahub.dhub.yaml
source:
  type: matillion-dpc
  config:
    api_config:
      client_id: "${MATILLION_CLIENT_ID}"
      client_secret: "${MATILLION_CLIENT_SECRET}"
      region: US1
    env: PROD
    parse_sql_for_lineage: true            # this one IS a real field
    include_streaming_pipelines: true
    include_unpublished_pipelines: true
    max_executions_per_pipeline: 10
    extract_projects_to_containers: true
    # project_patterns / pipeline_patterns / environment_patterns / streaming_pipeline_patterns are real
    stateful_ingestion:
      enabled: true
```

Update the two `@capability` strings in `matillion.py:111` and `:115` to reference real toggles (e.g., `parse_sql_for_lineage` for `LINEAGE_FINE`, and either drop the toggle reference for `LINEAGE_COARSE` or add a real `include_lineage` field).

Update the recipe string at `sources.json:148` to drop `include_pipeline_executions` (use `max_executions_per_pipeline` if execution gating is needed).

**Option B — add the missing fields to `MatillionSourceConfig`.** If the intent is for `include_lineage` / `include_column_lineage` / `include_pipeline_executions` to be real toggles, define them in `config.py` and gate the relevant code paths on them. Currently lineage is always on (no flag) and execution emission is gated by `max_executions_per_pipeline`, so this is the heavier path but maps better to what the docs promise.

Either way, please add an end-to-end test that loads `examples/recipes/matillion_dpc_to_datahub.dhub.yaml` via `MatillionSourceConfig.model_validate` so future drift between the example recipe and the schema fails CI.

datahub-web-react/src/app/ingestV2/source/builder/sources.json
acrylJonny: Fixed in the latest commit. Changed `"urn": "urn:li:dataPlatform:matillion-dpc"` → `"urn": "urn:li:dataPlatform:matillion"` in `ingestV2/source/builder/sources.json` to align with the platform registered in `data-platforms.yaml` and `MATILLION_PLATFORM = "matillion"` in the connector constants. The source `name` stays `"matillion-dpc"` (used as the source type id for recipe lookup).

metadata-ingestion/examples/recipes/matillion_dpc_to_datahub.dhub.yaml
acrylJonny: Fixed across all affected locations:

- `metadata-ingestion/examples/recipes/matillion_dpc_to_datahub.dhub.yaml` — removed `include_lineage` and `include_column_lineage` (not declared fields); fixed commented `project_pattern` / `pipeline_pattern` → `project_patterns` / `pipeline_patterns`
- `ingestV2/source/builder/sources.json` recipe string — removed `include_pipeline_executions`
- `ingest/source/conf/matillion-dpc/matillion-dpc.ts` placeholder recipe — removed `include_pipeline_executions`, `include_lineage`, `include_column_lineage`
- `matillion.py` capability strings — updated `@capability(LINEAGE_COARSE, ...)` to remove the reference to non-existent `include_lineage`; updated `@capability(LINEAGE_FINE, ...)` to reference the real field `parse_sql_for_lineage`

The example recipe now validates cleanly against `MatillionSourceConfig` with `extra="forbid"`.

treff7es (approved):
lgtm

PR #15967, opened 2026-01-23
chore(ingestion/dataplex): clean up gitignore and readme in connector code

acrylJonny:


metadata-ingestion/src/datahub/ingestion/source/dataplex/README.md
sgomezvillamor, line 1: I guess you are removing this as it is mostly duplicating user-facing docs, right?

acrylJonny: correct - no other source has these in there. It feels as though these were committed by mistake

PR #15978, opened 2026-01-26
feat(ingestion/bigquery) sharding optimizations

acrylJonny:
## Summary

Significant performance improvements for BigQuery datasets with many date-sharded tables (e.g., `logs_20240101`, `logs_20240102`, ...). These optimizations reduce processing time and API calls for datasets containing sharded tables.

### 1. Pre-compile Sharded Table Regex Pattern

**Before:** Regex pattern recompiled on every table check (~1-2ms per compilation)

**After:** Compile once at class level, reuse across all tables
- Added `BigqueryTableIdentifier._get_shard_pattern()` static method
- Pattern compiled once and cached
- **Performance gain:** 10-20 seconds saved per large dataset with 10,000 tables

### 2. Track `seen_base_tables` to Skip Duplicate Shards Early

**Before:** Process all 365 daily shards of `logs_YYYYMMDD`, then pick latest

**After:** Detect base table on first shard, skip subsequent shards immediately
- Fast regex-based deduplication before object creation
- Eliminates redundant `BigqueryTableIdentifier` instantiation
- Eliminates redundant pattern filtering checks
- **Performance gain:** For 365 daily shards → process 1 table instead of 365

### 3. Dynamic Batch Sizing Based on Table Count

**Rationale:** Sharded tables share schemas, so larger batches are efficient

**Implementation:**
- 200+ tables: 3x batch size (capped at 500)
- 100-200 tables: 2x batch size (capped at 300)
- <100 tables: default batch size

**Performance gain:** Dataset with 500 sharded tables: 17 → 6 API calls to BigQuery Information Schema

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_audit.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_audit.py
@@ -54,13 +54,7 @@ class BigqueryTableIdentifier:
         """
         Extract base table name from a sharded table regex match.
 
-        The regex pattern _BIGQUERY_DEFAULT_SHARDED_TABLE_REGEX captures:
-        - Group 1: Optional full prefix including separator (e.g., "table_" or "table$")
-        - Group 2: Base name without separator (e.g., "table")
-        - Group 3: Date shard (e.g., "20240101")
-
-        This method extracts group[1] for the prefix, then strips the shard
-        from the table_id to get the clean base name.
+        Regex groups: [1] = prefix with separator, [2] = base name, [3] = shard suffix.
         """
         base_name = match[1] or ""
         shard = match[3]


metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_audit.py
```diff
@@ -43,6 +43,14 @@ class BigqueryTableIdentifier:
...
     _BQ_SHARDED_TABLE_SUFFIX: str = "_yyyymmdd"
 
+    @staticmethod
+    def _get_shard_pattern() -> "re.Pattern[str]":
+        """Get the compiled regex pattern for shard detection."""
+        return re.compile(
+            BigqueryTableIdentifier._BIGQUERY_DEFAULT_SHARDED_TABLE_REGEX,
+            re.IGNORECASE,
+        )
+
```
sgomezvillamor, line 69: two concerns:
* since this is a method, we are still compiling the pattern on every method call, right?
* what about `sharded_table_pattern` config? shouldn't we compile that one instead - to be honest, no idea why in the code we were considering only `_BIGQUERY_DEFAULT_SHARDED_TABLE_REGEX `

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_schema_gen.py
```diff
@@ -530,13 +539,37 @@ def _process_schema(
...
+                if match:
+                    base_name = match[1] or ""
+                    if base_name and table_id.endswith(match[3]):
+                        base_name = table_id[: -len(match[3])].rstrip("_")
+                    if not base_name or base_name.endswith("."):
+                        base_name = dataset_name
+
+                    # Skip if we've already processed this base table
+                    if base_name in seen_base_tables:
+                        continue
```
sgomezvillamor: may we add a log trace for this skip case? maybe even a log info trace 🤔 

and can you check if we have counters about sharded tables that we should update here too

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_schema_gen.py
```diff
@@ -530,13 +539,37 @@ def _process_schema(
...
+                    seen_base_tables.add(base_name)
+
+                    # Use base name for identifier
+                    table_id = (
+                        base_name + BigqueryTableIdentifier._BQ_SHARDED_TABLE_SUFFIX
```
sgomezvillamor: again, I don't get why using the constant here and ignoring the `sharded_table_pattern` config
can you double check?

metadata-ingestion/tests/unit/bigquery/test_bigquery_source.py
```diff
@@ -1446,3 +1446,141 @@ def test_get_projects_with_project_labels(
...
+        "regular_table",
+    ]
+
+    sharded_count = 0
+    for table_id in table_ids:
+        match = pattern.match(table_id)
+        if match:
+            shard = match[3]
+            if shard:
+                sharded_count += 1
```
sgomezvillamor: this is testing custom code, not referencing any code in sources
so not relevant test and so not acceptable

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_schema_gen.py
```diff
@@ -1268,14 +1318,58 @@ def get_core_table_details(
...
+            match = shard_pattern.match(table_id)
+
+            if match:
+                base_name = match[1] or ""
+                shard = match[3]
+
+                if base_name and table_id.endswith(shard):
+                    base_name = table_id[: -len(shard)].rstrip("_")
+                if not base_name or base_name.endswith("."):
+                    base_name = dataset_name
```
sgomezvillamor: there is some duplicated code here that maybe could be wrapped in a function to avoid duplication

also, part of this logic is replicated in some tests, so instead of duplicating the logic in the tests, the function could be tested

metadata-ingestion/tests/unit/bigquery/test_bigquery_source.py
```diff
@@ -1446,3 +1446,141 @@ def test_get_projects_with_project_labels(
...
+
+            if base_name and table_id.endswith(shard):
+                base_name = table_id[: -len(shard)].rstrip("_")
+
+            if base_name in seen_base_tables:
+                continue
+            seen_base_tables.add(base_name)
+            unique_tables.append(base_name + "_yyyymmdd")
+        else:
+            unique_tables.append(table_id)
```
sgomezvillamor: same issue: testing custom logic instead of logic implemented in sources

metadata-ingestion/tests/unit/bigquery/test_bigquery_source.py
```diff
@@ -1446,3 +1446,141 @@ def test_get_projects_with_project_labels(
...
+            base_batch_size * DYNAMIC_BATCH_HIGH_MULTIPLIER,
+            DYNAMIC_BATCH_MAX_SIZE_HIGH,
+        )
+    elif table_count > DYNAMIC_BATCH_LOW_TABLE_THRESHOLD:
+        max_batch_size = min(
+            base_batch_size * DYNAMIC_BATCH_LOW_MULTIPLIER,
+            DYNAMIC_BATCH_MAX_SIZE_LOW,
+        )
+    else:
+        max_batch_size = base_batch_size
```
sgomezvillamor: let's wrap this in a function so it can be used in sources and so tested here

sgomezvillamor:
Do we still provide the max shard id, or would removing it be a breaking change?

If shards are now handled earlier in the code flow, does the logic that handles them later become legacy? is it still necessary, or could it be removed?

How much risk is there of introducing issues in shard handling with this change? did tests exist around this feature? or have you been able to manually test new approach?

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_audit.py
acrylJonny: Fixed! Changed from @staticmethod _get_shard_pattern() to @classmethod get_shard_pattern() with class-level caching:
Pattern is compiled once and cached in _compiled_shard_pattern
Returns same cached object on subsequent calls (verified with is check in tests)
Added recompile_shard_pattern() method called when config changes
Now respects sharded_table_pattern config via BigqueryTableIdentifier._BIGQUERY_DEFAULT_SHARDED_TABLE_REGEX

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_schema_gen.py
acrylJonny: I've added logging and metrics:
- num_sharded_tables_scanned 
- num_sharded_tables_deduped 
to BigQueryV2Report

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_schema_gen.py
acrylJonny: I've updated this to now use actual shard table_id from the matched table:
Essentially this changes from `base_name + BigqueryTableIdentifier._BQ_SHARDED_TABLE_SUFFIX` to `table=table_id` - (the actual shard like `events_20240101`)
This should match the behaviour of `get_core_table_details()` and should also respect `sharded_table_pattern` config

metadata-ingestion/tests/unit/bigquery/test_bigquery_source.py
acrylJonny: I've gone and completely rewritten these tests to call the actual code

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_schema_gen.py
acrylJonny: I've extracted this into a helper method called `extract_base_table_name `. Unit test also added for this

metadata-ingestion/tests/unit/bigquery/test_bigquery_source.py
acrylJonny: Also rewritten, as above

metadata-ingestion/tests/unit/bigquery/test_bigquery_source.py
acrylJonny: Now wrapped in `calculate_dynamic_batch_size` with unit test added also

acrylJonny:
> Do we still provide the max shard id, or would removing it be a breaking change?
> 
> If shards are now handled earlier in the code flow, does the logic that handles them later become legacy? is it still necessary, or could it be removed?
> 
> How much risk is there of introducing issues in shard handling with this change? did tests exist around this feature? or have you been able to manually test new approach?

Yes, we still provide max_shard_id. The changes do not affect the max_shard_id field at all - it's still populated during the full discovery path in _make_bigquery_table() and is used in two critical places. 1. to set custom properties is_sharded and populate DatasetSubTypes.SHARDED_TABLE for downstream consumers, and, 2. by the profiler to identify which shard to profile.

What was introduced was effectively a new sharding early dedup detection. In the old path we were reading all of the partitions and then picking the max_id, whereas now we just grab the max_shard_id earlier and use it.
Full discovery and SQL-level handling still remains

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery.py
```diff
@@ -126,6 +126,8 @@ def __init__(self, ctx: PipelineContext, config: BigQueryV2Config):
         BigqueryTableIdentifier._BIGQUERY_DEFAULT_SHARDED_TABLE_REGEX = (
             self.config.sharded_table_pattern
         )
+        # Recompile the pattern after config is set
+        BigqueryTableIdentifier.recompile_shard_pattern()
```
sgomezvillamor: woow, I have just realized that we were updating `BigqueryTableIdentifier._BIGQUERY_DEFAULT_SHARDED_TABLE_REGEX` from config here

```
        BigqueryTableIdentifier._BIGQUERY_DEFAULT_SHARDED_TABLE_REGEX = (
            self.config.sharded_table_pattern
        )
        # Recompile the pattern after config is set
        BigqueryTableIdentifier.recompile_shard_pattern()
```

I think this is horrible pattern we should get rid of it. This could be much more readable with something like this:

```
from pydantic import BaseModel, model_validator
from typing import Optional, ClassVar
import re


class CompiledPattern:
    """Provides compiled regex patterns."""
    
    _pattern: ClassVar[Optional[re.Pattern]] = None
    
    @staticmethod
    def set_pattern(pattern: str, flags: int = 0) -> None:
        """Set and compile the pattern (called during model initialization)."""
        CompiledPattern._pattern = re.compile(pattern, flags)
    
    @staticmethod
    def get_pattern() -> re.Pattern:
        """Get the compiled pattern."""
        if CompiledPattern._pattern is None:
            raise ValueError("Pattern not initialized. Call set_pattern first.")
        return CompiledPattern._pattern


class MyModel(BaseModel):
    pattern: str
    
    @model_validator(mode='after')
    def compile_pattern(self):
        """Compile the regex pattern once after validation."""
        CompiledPattern.set_pattern(self.pattern, re.IGNORECASE)
        return self


# Usage
model = MyModel(pattern=r"\d{8}$")

# Now other classes can use it
class OtherClass:
    def do_something(self, text: str):
        pattern = CompiledPattern.get_pattern()
        return pattern.match(text)

other = OtherClass()
match = other.do_something("table_20231231")  # Uses the pattern from MyModel
```

This way we preserve the static access (which can be problematic if accessed before initialization).

IMO Ideal solution would be to pass down the compiled pattern: `BigQueryIdentifierBuilder` --> `BigQueryIdentifier` constructor. I guess we ended up with current approach as a quick way to make a hardcoding to become configurable.

acrylJonny: @sgomezvillamor, shall I make these changes as part of this PR or would you prefer a follow-up PR for this

sgomezvillamor: go for it please
we shouldn't built on top of that pattern :-)

acrylJonny: cool cool - I've added that now

metadata-ingestion/tests/unit/bigquery/test_bigquery_source.py
```diff
@@ -1446,3 +1444,175 @@ def test_get_projects_with_project_labels(
...
+) -> None:
+    if shard1.isdigit() and shard2.isdigit():
+        is_newer = int(shard1) > int(shard2)
+    else:
+        is_newer = shard1 > shard2
+
+    assert is_newer == expected_is_newer, f"Failed for: {description}"
+
+
+def test_shard_pattern_is_compiled_once_and_cached():
```
sgomezvillamor: test name does not match test implementation

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_audit.py
```diff
@@ -29,22 +29,43 @@
...
+            base_name = table_id[: -len(shard)].rstrip("_")
+
+        if not base_name or base_name.endswith("."):
+            base_name = dataset_name
+
+        return base_name
+
+    @staticmethod
+    def get_table_and_shard(
+        table_name: str, shard_matcher: Optional[BigQueryShardPatternMatcher] = None
```
sgomezvillamor, line 72: why `shard_matcher` optional? not against that but:
either we make this `shard_matcher` required or we need to double check that all calls to `get_table_and_shard` are correctly passing the `shard_matcher`

metadata-ingestion/tests/unit/bigquery/test_bigquery_source.py
```diff
@@ -1446,3 +1444,175 @@ def test_get_projects_with_project_labels(
...
+)
+def test_numeric_shard_comparison(
+    shard1: str, shard2: str, expected_is_newer: bool, description: str
+) -> None:
+    if shard1.isdigit() and shard2.isdigit():
+        is_newer = int(shard1) > int(shard2)
+    else:
+        is_newer = shard1 > shard2
+
+    assert is_newer == expected_is_newer, f"Failed for: {description}"
```
sgomezvillamor, line 1473: this should be calling `is_newer` function, right?

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_schema_gen.py
```diff
@@ -119,6 +120,37 @@
...
+        return min(
+            base_batch_size * DYNAMIC_BATCH_LOW_MULTIPLIER,
+            DYNAMIC_BATCH_MAX_SIZE_LOW,
+        )
+    else:
+        return base_batch_size
+
+
+def is_shard_newer(shard: str, stored_shard: str) -> bool:
+    if shard.isdigit() and stored_shard.isdigit():
```
sgomezvillamor, line 160: The is_shard_newer function compares numeric shards incorrectly when they have different lengths. For example, comparing "2024" (year) vs "20230101" (date) does int("2024") > int("20230101") which gives False - semantically wrong! Mixed shard patterns will have unpredictable behavior.

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_schema_gen.py
```diff
@@ -1268,14 +1350,49 @@ def get_core_table_details(
...
         for table in self.schema_api.list_tables(dataset_name, project_id):
+            table_id = table.table_id
+
+            match = self.shard_matcher.match(table_id)
+
+            if match:
+                base_name = BigqueryTableIdentifier.extract_base_table_name(
+                    table_id, dataset_name, match
+                )
+                shard = match[3]
```
sgomezvillamor, line 1375: The `extract_base_table_name` method uses magic regex group numbers (`match[1]`, `match[3]`) with no documentation. Add comprehensive docs explaining what these groups represent.

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_schema_gen.py
```diff
@@ -119,6 +120,37 @@
 SNAPSHOT_TABLE_REGEX = re.compile(r"^(.+)@(\d{13})$")
 CLUSTERING_COLUMN_TAG = "CLUSTERING_COLUMN"
 
+# Dynamic batch sizing constants for sharded table optimization
```
sgomezvillamor, line 123: magic numbers, may you briefly document why those values were chosen and what's the intent?

PR #15997, opened 2026-01-28
refactor(sql-parsing): batch schema fetches across URN variants in resolve_table

acrylJonny:
## Summary
`resolve_table()` previously resolved the three URN variants (`urn`, `urn_lower`, `urn_mixed`) by
making sequential individual API calls — up to 3 round trips per table. This PR replaces that
with a single batched `get_entities()` call that fetches all uncached variants at once.

### Changes
**`schema_resolver.py`**
- Refactors `resolve_table()` to collect all URN variants into an ordered list, check the cache
  for all of them, then fetch any misses in a single `get_entities()` batch call
- Priority order is preserved: exact URN wins over `urn_lower`, which wins over `urn_mixed`
- Batch errors are caught and all fetched URNs are cached as `None` to prevent repeated API calls
  on subsequent lookups of the same table
- Extracts `add_schema_metadata_from_fetch()` as a public helper used by the batch flow
**`test_schemaresolver.py`**
- Adds `TestResolveTableBatching` covering: single-call batching, cache-first behaviour,
  error handling with None caching, partial batch matches, and no-fallback on failure


acrylJonny:
> I feel this PR is significantly over-engineered for what it currently delivers.
> 
> If I understand correctly, the main improvement is reducing three sequential API calls to a single batch call per table when checking URN casing variations. However, the implementation introduces a new `BatchSchemaFetcher`, which adds quite a bit of complexity to the interfaces. In my view, the same ~3× performance gain could be achieved with a simpler internal optimization inside `resolve_table()`: generate all URN variations upfront, check the cache, and then call `graph.get_entities()` once with all variations.
> 
> Alternatively, if the real goal is to batch across multiple tables (to unlock higher-level batching), that likely needs a different design. For example, introducing a `resolve_tables()` method in `SchemaResolverInterface`, so callers can accumulate tables and resolve them together. This would be a separate, though complementary, improvement and probably deserves its own PR.

Your assessment was very fair: the PR was over-engineered with the BatchSchemaFetcher abstraction. I've now implemented what you suggested - a simpler internal optimization inside resolve_table().

metadata-ingestion/src/datahub/sql_parsing/schema_resolver.py
```diff
@@ -143,44 +143,73 @@ def resolve_urn(self, urn: str) -> Tuple[str, Optional[SchemaInfo]]:
...
-        # Our treatment of platform instances when lowercasing urns
-        # is inconsistent. In some places (e.g. Snowflake), we lowercase
-        # the table names but not the platform instance. In other places
-        # (e.g. Databricks), we lowercase everything because it happens
-        # via the automatic lowercasing helper.
-        # See https://github.com/datahub-project/datahub/pull/8928.
-        # While we have this sort of inconsistency, we should also
-        # check the mixed case urn, as a last resort.
+        # Check mixed case due to inconsistent platform instance lowercasing across connectors.
+        # See https://github.com/datahub-project/datahub/pull/8928
```
sgomezvillamor: keep these breadcrumbs of knowledge

```suggestion
        # Our treatment of platform instances when lowercasing urns
        # is inconsistent. In some places (e.g. Snowflake), we lowercase
        # the table names but not the platform instance. In other places
        # (e.g. Databricks), we lowercase everything because it happens
        # via the automatic lowercasing helper.
        # See https://github.com/datahub-project/datahub/pull/8928.
        # While we have this sort of inconsistency, we should also
        # check the mixed case urn, as a last resort.
```

metadata-ingestion/src/datahub/sql_parsing/schema_resolver.py
```diff
@@ -143,44 +143,73 @@ def resolve_urn(self, urn: str) -> Tuple[str, Optional[SchemaInfo]]:
...
+            urns_to_fetch = [u for u in urns_to_try if u not in self._schema_cache]
+
+            if urns_to_fetch:
+                try:
+                    entity_results = self.graph.get_entities(
+                        entity_name="dataset",
+                        urns=urns_to_fetch,
+                        aspects=[SchemaMetadataClass.ASPECT_NAME],
+                        with_system_metadata=False,
+                    )
```
sgomezvillamor, line 186: Please correct me if I’m misunderstanding, but I have a couple of concerns:

* Previously, the three resolution attempts were performed exclusively against the cache. Now, we fall back to calling the graph if the entity isn’t found in the cache. What could be the impact of this change?
* Related to that: is there anything preventing repeated graph calls for the same missing URN? Continuously querying the graph for entities that don’t exist could introduce a significant performance penalty.

metadata-ingestion/src/datahub/sql_parsing/schema_resolver.py
```diff
@@ -143,44 +143,73 @@ def resolve_urn(self, urn: str) -> Tuple[str, Optional[SchemaInfo]]:
...
+                            if SchemaMetadataClass.ASPECT_NAME in entity_aspects:
+                                aspect_value, _ = entity_aspects[
+                                    SchemaMetadataClass.ASPECT_NAME
+                                ]
+                                if isinstance(aspect_value, SchemaMetadataClass):
+                                    schema_metadata = aspect_value
+
+                        self.add_schema_metadata_from_fetch(fetch_urn, schema_metadata)
+
+                except (TimeoutError, ConnectionError, OSError) as e:
```
sgomezvillamor: Overly Narrow Exception Handler?
I think graph can raise many different exceptions, also including HTTPError, JSONDecodeError, ValueError, ...

metadata-ingestion/tests/unit/sql_parsing/test_schemaresolver.py
```diff
@@ -11,178 +15,235 @@
...
 
 
-def test_get_urn_for_table_not_lower_should_keep_capital_letters():
-    schema_resolver = SchemaResolver(
-        platform="mssql",
-        platform_instance="Uppercased-Instance",
-        env="PROD",
-        graph=None,
-    )
+class TestResolveTableBatching:
```
sgomezvillamor, line 142: Missing tests?
* Test when batch fetch returns schemas for SOME URN variations but not others
* Test when some URN variations are cached as None (previously checked) and others need fetching.

metadata-ingestion/tests/unit/sql_parsing/test_schemaresolver.py
```diff
@@ -11,178 +15,235 @@
...
+    def test_uses_cache_first(self, schema_resolver, mock_graph):
+        urn = "urn:li:dataset:(urn:li:dataPlatform:snowflake,prod.db.schema.table,PROD)"
+        schema_resolver.add_schema_metadata(
+            urn, create_mock_schema([("cached_col", "string")])
+        )
 
+        table = _TableName(database="db", db_schema="schema", table="table")
+        resolved_urn, schema_info = schema_resolver.resolve_table(table)
 
-def test_match_columns_to_schema():
```
sgomezvillamor, line 109: Are below tests updated because they are paremetrized now in `test_multi_part_table_names` and `test_equality_rules_by_part_count`?

metadata-ingestion/src/datahub/sql_parsing/schema_resolver.py
```diff
@@ -143,44 +143,73 @@ def resolve_urn(self, urn: str) -> Tuple[str, Optional[SchemaInfo]]:
...
+                    logger.warning(
+                        f"Batch fetch failed due to network issue: {e}. "
+                        f"Falling back to individual fetches for {len(urns_to_fetch)} URNs.",
+                        exc_info=True,
+                    )
+                    for fetch_urn in urns_to_fetch:
+                        if fetch_urn not in self._schema_cache:
+                            self._resolve_schema_info(fetch_urn)
```
sgomezvillamor: why do we call again `_resolve_schema_info` which fetches from cache and fallbacks to graph call again?

sgomezvillamor:
@acrylJonny could you please reply to the review comments directly?

When there’s no response, I have to infer whether concerns were addressed by digging through the commits, which makes the review much more time-consuming.

Given the volume of PRs I’m reviewing, a bit more back-and-forth in the comments would really help keep the process efficient.

acrylJonny:
> @acrylJonny could you please reply to the review comments directly?
> 
> When there’s no response, I have to infer whether concerns were addressed by digging through the commits, which makes the review much more time-consuming.
> 
> Given the volume of PRs I’m reviewing, a bit more back-and-forth in the comments would really help keep the process efficient.

Sure - I addressed the comments but forgot to add a response so will do that today

metadata-ingestion/src/datahub/sql_parsing/schema_resolver.py
acrylJonny: I've put these back now

metadata-ingestion/src/datahub/sql_parsing/schema_resolver.py
acrylJonny: That's fair on the performance concerns. Let me clarify the behavior changes:

**Previous behavior:** All three resolution attempts were cache-only. If not found in cache, the method would return `None` without fetching from DataHub.

**New behavior:** 
1. First, we check all URN variations against the cache (lines 166-171)
2. If none are cached, we batch-fetch ONLY the uncached URNs from the graph (line 174: `urns_to_fetch = [u for u in urns_to_try if u not in self._schema_cache]`)
3. We cache the results (including `None` for missing schemas) to prevent repeated lookups (lines 197, 311-312)

metadata-ingestion/src/datahub/sql_parsing/schema_resolver.py
acrylJonny: I've expanded in lines 199-229 to cover:
Network errors: TimeoutError, ConnectionError, OSError
HTTP errors: HTTPError
Data parsing errors: json.JSONDecodeError, ValueError, KeyError
DataHub-specific errors: GraphError, AssertionError
The fallback mechanism (lines 344-383) uses the same comprehensive exception handling.

metadata-ingestion/tests/unit/sql_parsing/test_schemaresolver.py
acrylJonny: Requested tests added

metadata-ingestion/src/datahub/sql_parsing/schema_resolver.py
acrylJonny: The problematic code has now been removed

metadata-ingestion/tests/unit/sql_parsing/test_schemaresolver.py
acrylJonny: Updated

Old tests: Had separate hardcoded test functions for different table name scenarios
New tests:
test_multi_part_table_names (lines 344-366) - Parameterized test for various multi-part table name patterns (4, 5, 7 parts)
test_equality_rules_by_part_count (lines 403-427) - Parameterized test for equality behavior based on part count

acrylJonny:
> > @acrylJonny could you please reply to the review comments directly?
> > When there’s no response, I have to infer whether concerns were addressed by digging through the commits, which makes the review much more time-consuming.
> > Given the volume of PRs I’m reviewing, a bit more back-and-forth in the comments would really help keep the process efficient.
> 
> Sure - I addressed the comments but forgot to add a response so will do that today

@sgomezvillamor - these comments should have now been addressed

rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `264bcd0`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/23441575585)

acrylJonny:
> The feature itself is valuable. However, I have some concerns about the implementation approach:
> 
> **1. Code Clarity and Documentation**
> 
> The caching logic at line 174 is critical for preventing repeated failed lookups, but it's not immediately obvious:
> 
> ```
>   if self.graph:
>       urns_to_fetch = [u for u in urns_to_try if u not in self._schema_cache]
> ```
> 
> This line is doing important work - skipping URNs that are cached as `None` to avoid repeated failed attempts. A code comment or better naming here would make this behavior explicit and help future maintainers understand why we're checking cache membership rather than just checking for non-None values.
> 
> **2. Code Complexity for a Simple Feature**
> 
> The implementation introduces significant complexity for what should be a straightforward batch fetch operation. Consider:
> 
> * 190+ lines of new code with extensive error handling duplication
> * Nested try-except blocks with ~40 lines of duplicated exception handling across 4 different exception types
> * New method `add_schema_metadata_from_fetch()` that seems redundant when `add_schema_metadata()` already exists, and the return value isn't even used
> 
> Suggestion: Extract the batch fetch logic into a dedicated helper method with clear naming (e.g., `_batch_fetch_uncached_urns()`) that encapsulates the fetch + error handling + caching logic. This would make the main `resolve_table()` flow much easier to follow.
> 
> **3. Method Design: `add_schema_metadata_from_fetch()`**
> 
> The new method has a `bool` return value indicating whether caching occurred, but this return value is never used in the codebase. Additionally, having two very similar methods (`add_schema_metadata` vs `add_schema_metadata_from_fetch`) with slightly different precedence rules creates cognitive overhead. Could we consolidate these or make the distinction more obvious?
> 
> **4. Test Refactoring Risk**
> 
> The PR includes 321 lines of test changes vs 214 deletions, with significant refactoring of existing tests using parametrization. While parametrization can be beneficial, mixing it with new feature tests increases review complexity and regression risk. Would it be clearer to add new tests for the batch fetch behavior alongside the existing tests, rather than refactoring everything at once?
> 
> I may have asked for parametrized tests in previous comments, however looking at current status, I'm concerned on significant refactor of tests introducing some breaking change. So I would prefer a more conservative approach by just introducing new tests for the new feature.
> 
> **5. Missing Test Coverage**
> 
> From the review analysis:
> 
> * Critical gap: No test for batch fetch with partial success (some URNs exist in DataHub, others don't)
> * Critical gap: No test verifying that None-cached URNs are not re-fetched on subsequent calls
> 
> These are core behaviors that should be explicitly tested.
> 
> **6. Error Handling Duplication**
> 
> The error handling block (lines 199-229) repeats nearly identical logic for 4 exception types. This could be simplified to reduce maintenance burden and improve readability.
> 
> Overall: Love the optimization goal, but the implementation should be simpler, more testable, and easier to maintain.

Thanks for the detailed review — addressed all points:

**1. Cache filter comment (line 174)**
Added a 3-line comment explaining that `u not in self._schema_cache` intentionally skips both hit entries *and* `None`-cached entries, since a stored `None` means a prior lookup already failed and we want to avoid re-fetching.

**2. Error handling duplication**
Consolidated the 4 near-identical `except` blocks into one in both `resolve_table` and `_fallback_fetch_schemas`. To preserve the diagnostic value that the original per-type messages had, I included `type(e).__name__` in the log message — this is actually more precise than before (e.g. `ConnectionError` vs `TimeoutError` are now distinguishable in logs, where previously both were just "network issue").

**3. `add_schema_metadata_from_fetch()` — unused return value + two-method confusion**
Removed the unused `bool` return value (now `-> None`). Rather than merging the two methods — they have genuinely different semantics that serve different callers — I've given both cross-referencing docstrings that explicitly state when to use each and how their cache precedence rules differ.

**4. Test refactoring risk**
Reverted all refactoring of existing tests. The `TestTableNameParts` class and the `test_get_urn_for_table_*` functions are now restored verbatim from master. The only net-new test code is the `TestResolveTableBatching` class and the `create_mock_schema` helper.

**5. Missing test coverage**
Both flagged gaps are covered in `TestResolveTableBatching`:
- `test_partial_urn_match_from_batch` — batch returns a schema for only one of the URN case variants
- `test_caches_none_to_prevent_repeated_lookups` — calls `resolve_table` twice for a missing table and verifies `get_entities` is only called once

**6. Exception handling — same action for all types**
All 9 exception types share the same fallback action (`_fallback_fetch_schemas`) because that method uses a different API (`get_aspect` vs `get_entities`), making it reasonable to attempt regardless of the batch failure cause. The `type(e).__name__` addition makes them distinguishable in logs without needing separate handlers.

metadata-ingestion/src/datahub/sql_parsing/schema_resolver.py
```diff
@@ -217,9 +266,42 @@ def _resolve_schema_info(self, urn: str) -> Optional[SchemaInfo]:
...
+        schemas are authoritative and fresh from the source system.
+        Use this for schemas discovered during ingestion.
+
+        See also: add_schema_metadata_from_fetch (for schemas retrieved on-demand
+        from the DataHub API, which must not override ingestion-provided schemas).
+        """
         schema_info = _convert_schema_aspect_to_info(schema_metadata)
         self._save_to_cache(urn, schema_info)
 
+    def add_schema_metadata_from_fetch(
```
sgomezvillamor, line 273: As per my understanding, this method is not needed, it's the case of unnecessary Defensive Code
                                     
We have

```
  urns_to_fetch = [u for u in urns_to_try if u not in self._schema_cache]
```
                                                                                                                                                                                                                                
Since we filter out cached URNs before fetching, the "never overwrites non-None entry" logic in `add_schema_metadata_from_fetch` is sort of dead code: If `urn` was in `self._schema_cache`, it wouldn't be in `urns_to_fetch`, so we'd never call this method for that URN.

metadata-ingestion/src/datahub/sql_parsing/schema_resolver.py
```diff
@@ -248,6 +330,39 @@ def with_temp_tables(
     def _save_to_cache(self, urn: str, schema_info: Optional[SchemaInfo]) -> None:
         self._schema_cache[urn] = schema_info
 
+    def _fallback_fetch_schemas(self, urns: List[str]) -> None:
```
sgomezvillamor: This fallback retry strategy looks wrong to me.

In current proposal:
1. batch fetch 3 urns
2. if fails, retry each individually (3 separate calls)

Previously and overall we:
1. try 3 urns individually and sequentially
(no retry at all)

I would continue with the current standard in this layer: no retry

The `DatahubRestEmitter` (which underlies `DataHubGraph`) already has comprehensive retry logic built in

metadata-ingestion/src/datahub/sql_parsing/schema_resolver.py
acrylJonny: Aaah, youre right, there was dead code code in there. I've removed the check to simplify things.

metadata-ingestion/src/datahub/sql_parsing/schema_resolver.py
acrylJonny: Note taken here. I've remove the fallback and now when the batch call fails we now just cache None to prevent the repeated lookups.

metadata-ingestion/src/datahub/sql_parsing/schema_resolver.py
```diff
@@ -167,22 +158,79 @@ def resolve_table(self, table: _TableName) -> Tuple[str, Optional[SchemaInfo]]:
...
+
+        urns_to_try = [urn]
+        if urn_lower != urn:
+            urns_to_try.append(urn_lower)
         if urn_mixed not in {urn, urn_lower}:
-            schema_info = self._resolve_schema_info(urn_mixed)
-            if schema_info:
-                self._track_cache_hit()
-                return urn_mixed, schema_info
+            urns_to_try.append(urn_mixed)
```
sgomezvillamor, line 166:
```suggestion
        urns_to_try = set(urn, urn_lower, urn_mixed)
```

isn't this the same?

acrylJonny: Not quite the same, unfortunately — there are a few issues:

1. There would be a syntax error - `set()` takes a single argument, so the `set(urn, urn_lower, urn_mixed)` would raise an error.

2. Order is semantically significant. `urns_to_try` is iterated in three order-sensitive places:

- Line 168: the first cache hit returns immediately — urn (exact case) must always beat urn_lower / urn_mixed
- Line 177: `urns_to_fetch` preserves the order for the get_entities call
- Line 221: the post-fetch winner is also whichever comes first in the list
- Sets have no defined iteration order, so `{urn, urn_lower, urn_mixed}` could return a lowercase or mixed-case variant even when the exact-case URN is in the cache.

3. Deduplication avoids redundant work. When all three variants are equal (e.g. a fully-lowercase table + platform), the list collapses to a single element. That matters at line 181 — `graph.get_entities()` would otherwise receive duplicate URNs in the batch request.

PR #15998, opened 2026-01-28
tests(cypress): add basic lineage tests

v-tarasevich-blitz-brain:
Linear: https://linear.app/acryl-data/issue/CAT-1299/create-lineage-cypress-tests

This PR adds basic lineage cypress tests


smoke-test/tests/cypress/cypress/e2e/lineageV2/v2_lineage_graph.js
```diff
@@ -158,4 +189,114 @@ describe("lineage_graph", () => {
     // verify that is is not disabled
     cy.contains("Set Upstreams").should("not.be.disabled");
   });
+
+  it.only("displays complete lineage graph with all node types", () => {
```
asikowitz: Remove `.only`

smoke-test/tests/cypress/cypress/e2e/lineageV2/v2_lineage_graph.js
```diff
@@ -158,4 +196,151 @@ describe("lineage_graph", () => {
     // verify that is is not disabled
     cy.contains("Set Upstreams").should("not.be.disabled");
   });
+
+  it("displays complete lineage graph with all node types", () => {
```
chriscollins3456, line 200: can we also add these same tests for lineage v3? we need to make sure `LINEAGE_GRAPH_V3` or (`lineageGraphV3`) is enabled in app config and we can basically run these exact tests hopefully should be simple to update

chriscollins3456 (approved):
nice!

PR #16002, opened 2026-01-28
feat(gcs): adds WIF Workload Identification Federation support

bryanprosser-acryl:
This PR adds in support for Google' Workload Identity Federation as an alternative to the existing HMAC authentication for connecting to GCP. It also adds unit tests and updates the documentation to reflect the additional config options.

### DIFF
--- a/metadata-ingestion-modules/dagster-plugin/src/datahub_dagster_plugin/client/dagster_generator.py
+++ b/metadata-ingestion-modules/dagster-plugin/src/datahub_dagster_plugin/client/dagster_generator.py
@@ -70,6 +70,7 @@ from datahub.metadata.schema_classes import (
     QueryStatementClass,
     QuerySubjectClass,
     QuerySubjectsClass,
+    StatusClass,
     SubTypesClass,
     TagAssociationClass,
     UpstreamClass,
@@ -808,6 +809,13 @@ class DagsterGenerator:
         )
         for mcp in dataset.generate_mcp():
             graph.emit_mcp(mcp)
+
+        graph.emit_mcp(
+            MetadataChangeProposalWrapper(
+                entityUrn=dataset_urn.urn(),
+                aspect=StatusClass(removed=False),
+            )
+        )
 
         if schema:
             mcp = self.convert_table_schema_to_schema_metadata(
--- a/metadata-ingestion-modules/dagster-plugin/tests/unit/test_dagster.py
+++ b/metadata-ingestion-modules/dagster-plugin/tests/unit/test_dagster.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import pathlib
 import tempfile
 import uuid
@@ -27,8 +28,13 @@ from freezegun import freeze_time
 
 from datahub.emitter.mcp import MetadataChangeProposalWrapper
 from datahub.ingestion.graph.client import DatahubClientConfig
+from datahub.metadata.schema_classes import StatusClass
 from datahub.testing.compare_metadata_json import assert_metadata_files_equal
-from datahub_dagster_plugin.client.dagster_generator import DatahubDagsterSourceConfig
+from datahub_dagster_plugin.client.dagster_generator import (
+    DagsterEnvironment,
+    DagsterGenerator,
+    DatahubDagsterSourceConfig,
+)
 from datahub_dagster_plugin.sensors.datahub_sensors import (
     DatahubSensors,
     make_datahub_sensor,
@@ -551,3 +557,36 @@ def test_success_run_still_emits_lineage(mock_emit: Mock, mock_uuid: Mock) -> No
     assert "urn:li:dataset:(urn:li:dataPlatform:snowflake,tableA,PROD)" in [
         str(d) for d in io_aspect.inputDatasets
     ]
+
+
+def test_emit_asset_emits_status_aspect() -> None:
+    """Verify that emit_asset emits StatusClass(removed=False) so that
+    soft-deleted entities are automatically restored on re-ingestion.
+    """
+    mock_graph = Mock()
+    config = DatahubDagsterSourceConfig(
+        datahub_client_config=DatahubClientConfig(server="http://localhost:8081"),
+    )
+    generator = DagsterGenerator(
+        logger=logging.getLogger(__name__),
+        config=config,
+        dagster_environment=DagsterEnvironment(repository=None),
+    )
+    generator.emit_asset(
+        graph=mock_graph,
+        asset_key=["my_asset"],
+        description=None,
+        properties=None,
+    )
+
+    emitted = [c.args[0] for c in mock_graph.emit_mcp.call_args_list]
+    status_mcps = [
+        m
+        for m in emitted
+        if isinstance(m, MetadataChangeProposalWrapper)
+        and isinstance(m.aspect, StatusClass)
+    ]
+    assert len(status_mcps) == 1
+    status_aspect = status_mcps[0].aspect
+    assert isinstance(status_aspect, StatusClass)
+    assert status_aspect.removed is False


gabe-lyons:
Linear: ING-1908

metadata-ingestion/docs/sources/gcs/gcs.md
sgomezvillamor, line 1: Docs look good overall, but please check the structure/organization, this is something we've been improving lately.
Have a look at https://github.com/datahub-project/datahub/blob/master/metadata-ingestion/docs/sources/CLAUDE.md for the structure expectations.

That will also address the conflicts with master in the PR

metadata-ingestion/docs/sources/gcs/gcs.md
```diff
@@ -1,3 +1,74 @@
...
+source:
+  type: gcs
+  config:
+    credential:
+      hmac_access_id: <your-access-id>
+      hmac_access_secret: <your-access-secret>
+```
+
+#### Workload Identity Federation
+
```
sgomezvillamor: we could mention some of the use cases for this WIF authentication:

this is what I have found

> Kubernetes workloads (GKE) that need to avoid storing service account keys, Cross-cloud authentication (AWS/Azure → GCP), Temporary credential rotation for security compliance

metadata-ingestion/src/datahub/ingestion/source/gcs/gcs_source.py
```diff
@@ -37,11 +42,105 @@
...
+    ) -> "S3ServiceResource":
+        resource = super().get_s3_resource(verify_ssl=verify_ssl)
+        creds = getattr(self, "_gcs_oauth_credentials", None)
+        project_id = getattr(self, "_gcs_oauth_project_id", None)
+        if creds is not None:
+            _register_gcs_oauth_before_send(resource.meta.client, creds, project_id)
+        return resource
+
+
+class GCSAuthType(StrEnum):
```
sgomezvillamor, line 125: WIF is a GCP-wide auth mechanism, so other sources may eventually need it too (BigQuery, Dataplex, VertexAI, etc.). Am I right?

We already have some shared GCP code here: https://github.com/datahub-project/datahub/tree/0327b16edcd68411f6c1e334c1af12b09e306c95/metadata-ingestion/src/datahub/ingestion/source/common
Can we place the implementation there and keep it as generic as possible? The initial integration will be GCS, but the code should be ready to support other sources down the line.

metadata-ingestion/src/datahub/ingestion/source/gcs/gcs_source.py
```diff
@@ -105,31 +294,150 @@ def __init__(self, config: GCSSourceConfig, ctx: PipelineContext):
...
+                path_specs=s3_path_specs,
+                aws_config=AwsConnectionConfig(
+                    aws_endpoint_url="https://storage.googleapis.com",
+                    aws_access_key_id=self.config.credential.hmac_access_id,
+                    aws_secret_access_key=self.config.credential.hmac_access_secret.get_secret_value(),
+                    aws_region="auto",
+                ),
+                env=self.config.env,
+                max_rows=self.config.max_rows,
+                number_of_files_to_sample=self.config.number_of_files_to_sample,
```
sgomezvillamor, line 262: are we missing `platform` and `platform_instance`?

they were set previously in one of the two intantiations of `DataLakeSourceConfig`, so please double check

```
            platform=PLATFORM_GCS,  # Ensure GCS platform is used for correct container subtypes
            platform_instance=self.config.platform_instance,
```

sgomezvillamor: same concern for 

```
  convert_urns_to_lowercase=self.config.convert_urns_to_lowercase,
```

metadata-ingestion/src/datahub/ingestion/source/gcs/gcs_source.py
```diff
@@ -105,31 +294,150 @@ def __init__(self, config: GCSSourceConfig, ctx: PipelineContext):
...
+        # Convert all formats to a file path (load_credentials_from_file works reliably)
+        wif_config_file = None
+
+        if self.config.gcp_wif_configuration:
+            wif_config_file = self.config.gcp_wif_configuration
+            logger.info(
+                "Using Workload Identity Federation configuration from file: %s",
+                wif_config_file,
+            )
+        elif self.config.gcp_wif_configuration_json:
```
sgomezvillamor: Given these temp files have sensitive information, we may restrict permissions to user only

sgomezvillamor:
WIF config files can contain sensitive credentials, so writing them to disk isn't ideal for very strict security environments — it expands the attack surface (file permissions, disk forensics, etc.). This is a pattern we've been flagged about before and we should move away from if possible.

The current mitigations (unlinking after use, etc.) help, but files are still accessible while running and may be left behind on failure.

At minimum, let's tighten permissions on those files.

Worth exploring whether there's a way to auth from a dict directly without touching disk, or using `os.memfd_create()` on Linux to avoid it entirely.

metadata-ingestion/docs/sources/gcs/gcs.md
bryanprosser-acryl: Restructured the GCS docs to follow the new _pre.md / _post.md / _recipe.yml format per docs/sources/AGENTS.md. This also resolved the merge conflicts with master. Added WIF use-case examples to gcs_pre.md covering the scenarios you suggested (GKE workloads, cross-cloud auth, credential rotation for compliance).

metadata-ingestion/src/datahub/ingestion/source/gcs/gcs_source.py
bryanprosser-acryl: Moved the WIF implementation out of gcs_source.py into a new shared module: source/common/gcp_wif_config.py. This exposes two reusable exports:

GCPWIFConfig — a Pydantic config mixin with the three input options (file path, JSON dict, JSON string) and format validators. BigQuery, Dataplex, VertexAI can inherit from this directly when they need WIF support.
load_wif_credentials(config) — loads google.auth.credentials.Credentials from any of the three options, applies the cloud-platform scope, and returns (credentials, project_id). This is the exact interface those other sources would need to pass to their google-cloud clients.
GCSSource now just calls load_wif_credentials(self.config) — the GCS-specific part (bearer token injection into boto3) stays in gcs_source.py since that's unique to the S3-compatible XML API approach.

metadata-ingestion/src/datahub/ingestion/source/gcs/gcs_source.py
bryanprosser-acryl: Fixed both issues

metadata-ingestion/src/datahub/ingestion/source/gcs/gcs_source.py
bryanprosser-acryl: Eliminated temp file usage entirely. Switched from load_credentials_from_file() to google.auth.load_credentials_from_dict(), which loads WIF credentials directly from a Python dict in memory — no file is ever written. This applies to all three input options (file path is read once into a dict; the dict and JSON string options never touch disk at all).

bryanprosser-acryl:
@sgomezvillamor 
Updated the PR based on your comments above - main changes made:

1. Docs structure/organisation

Restructured the GCS docs to follow the new _pre.md / _post.md / _recipe.yml format per docs/sources/AGENTS.md. This also resolved the merge conflicts with master. Added WIF use-case examples to gcs_pre.md covering the scenarios you suggested (GKE workloads, cross-cloud auth, credential rotation for compliance).

2. Shared GCP location for WIF

Moved the WIF implementation out of gcs_source.py into a new shared module: source/common/gcp_wif_config.py. This exposes two reusable exports:

GCPWIFConfig — a Pydantic config mixin with the three input options (file path, JSON dict, JSON string) and format validators. BigQuery, Dataplex, VertexAI can inherit from this directly when they need WIF support.
load_wif_credentials(config) — loads google.auth.credentials.Credentials from any of the three options, applies the cloud-platform scope, and returns (credentials, project_id). This is the exact interface those other sources would need to pass to their google-cloud clients.
GCSSource now just calls load_wif_credentials(self.config) — the GCS-specific part (bearer token injection into boto3) stays in gcs_source.py since that's unique to the S3-compatible XML API approach.

3. Missing platform, platform_instance, convert_urns_to_lowercase

These were missing from the WIF branch of create_equivalent_s3_config. Fixed — both DataLakeSourceConfig instantiation paths (HMAC and WIF) now pass all three fields.

4. Temp file security / avoid writing credentials to disk

Eliminated temp file usage entirely. Switched from load_credentials_from_file() to google.auth.load_credentials_from_dict(), which loads WIF credentials directly from a Python dict in memory — no file is ever written. This applies to all three input options (file path is read once into a dict; the dict and JSON string options never touch disk at all).

bryanprosser-acryl:
I don't know why the vertica plugin in failing

metadata-ingestion/src/datahub/ingestion/source/common/gcp_wif_config.py
```diff
@@ -0,0 +1,164 @@
...
+        ValueError: If no WIF configuration is provided or if credential loading fails.
+    """
+    if not any(
+        [
+            wif_config.gcp_wif_configuration,
+            wif_config.gcp_wif_configuration_json,
+            wif_config.gcp_wif_configuration_json_string,
+        ]
+    ):
+        raise ValueError("No valid WIF configuration provided")
```
sgomezvillamor, line 135: this could be moved up as a pydantic validation
you could also add the validation that only one is set

metadata-ingestion/src/datahub/ingestion/source/common/gcp_wif_config.py
```diff
@@ -0,0 +1,164 @@
...
+            logger.info(
+                "Using Workload Identity Federation configuration from JSON content"
+            )
+        else:
+            wif_config_dict = json.loads(wif_config.gcp_wif_configuration_json_string)  # type: ignore[arg-type]
+            logger.info(
+                "Using Workload Identity Federation configuration from JSON string"
+            )
+
+        credentials, project_id = load_credentials_from_dict(wif_config_dict)
```
sgomezvillamor, line 159: 👍

metadata-ingestion/src/datahub/ingestion/source/gcs/gcs_source.py
```diff
@@ -70,16 +158,54 @@ class GCSSourceConfig(
...
+                if len(provided_options) == 0:
+                    raise ValueError(
+                        "One of gcp_wif_configuration (file path), gcp_wif_configuration_json (JSON content), "
+                        "or gcp_wif_configuration_json_string (JSON string) is required when auth_type is 'workload_identity_federation'"
+                    )
+                elif len(provided_options) > 1:
+                    raise ValueError(
+                        "Cannot specify multiple WIF configuration options. Use only one of: "
+                        "gcp_wif_configuration, gcp_wif_configuration_json, or gcp_wif_configuration_json_string."
+                    )
```
sgomezvillamor: oh.... I seee, the validation I was suggesting is here already 👍

metadata-ingestion/src/datahub/ingestion/source/gcs/gcs_source.py
```diff
@@ -105,34 +231,76 @@ def __init__(self, config: GCSSourceConfig, ctx: PipelineContext):
...
+                aws_secret_access_key="not-used",
+            )
+            # Pydantic v2 models freeze field assignment after __init__; object.__setattr__
+            # bypasses this to attach runtime-only state that is not part of the config schema.
+            object.__setattr__(
+                aws_config, "_gcs_oauth_credentials", self._wif_credentials
+            )
+            object.__setattr__(
+                aws_config, "_gcs_oauth_project_id", self._wif_project_id
+            )
```
sgomezvillamor: this looks hacky, and I haven't found any reference to these fields in internet or ALL github 🤔 

how much is this needed?
what's the error you get if not set?

we need more contextual information in the comment in case we ever need to revisit this

sgomezvillamor:
> I don't know why the vertica plugin in failing

It was fixed with https://github.com/datahub-project/datahub/commit/5b1a31698018bff2b74186a924b41271c0141247, so update branch to fix

bryanprosser-acryl:
> Overall LGTM
> 
> I'm a little bit concerned about the `object.__setattr__` calls for `_gcs_oauth_credentials` and `_gcs_oauth_project_id` Approving to unblock but please let's provide more info or context on that.

Updated this to be pydantic v2 way to declare instance-level runtime state that isn't part of the config schema.

PR #16012, opened 2026-01-29
[RFC] Agents and Related Assets in Metadata Model

shirshanka:
## Summary

This RFC proposes adding support for AI agents and related agentic assets to DataHub's metadata model.

## Overview

As organizations build agent platforms and deploy AI agents into production, they need a centralized catalog to discover, understand, and govern these assets - similar to how DataHub catalogs data assets today.

This RFC covers:
- **Agent Entity**: Catalog AI agents with goals, status, relationships, and trust metrics
- **AgentSkill Entity**: High-level capabilities (prompts + tools + expertise) with git-based registry
- **AgentTool Entity**: Catalog MCP tools, APIs, and functions that agents can invoke
- **Vector Store Support**: Model vector databases as DataPlatforms with vector column types
- **Relationships**: Connect agents to skills, tools, ML models, datasets, and code repositories
- **Observability**: Integration with OpenTelemetry traces and evaluation platforms

## Key Design Decisions

✅ **Skills follow git as source of truth** - Skills stored in git repos (agentskills.io standard), DataHub provides discovery/catalog layer

✅ **No new VectorStore entity** - Vector stores are DataPlatforms (like Snowflake), with support for vector column types

✅ **Framework agnostic** - Works with LangGraph, LangChain, custom frameworks

✅ **Trust-first** - Evaluation metrics and reliability indicators are core requirements

✅ **Purely additive** - No breaking changes to existing entities or APIs

## AgentSkill Entity

The RFC now includes **AgentSkill** as a first-class entity representing reusable capability bundles:
- **Git as Source of Truth**: Skills defined in git repositories following [agentskills.io](https://agentskills.io) open standard
- **DataHub as Discovery Layer**: Ingests skill metadata for search, lineage, relationships, governance
- **Mirrors Data Assets Pattern**: Like source systems own data and DataHub catalogs it, git repos own skills and DataHub catalogs them
- **Skills vs Tools**: Skills are high-level (domain expertise + prompts), Tools are low-level (API calls, functions)

## Open Questions for Community

The RFC includes 9 unresolved questions seeking community input:
1. Knowledge graph modeling approach
2. Service user entity integration for agent identity
3. Conversation entity necessity
4. Agent version management strategy
5. Evaluation criteria standardization
6. Tool schema validation approach
7. OTEL semantic conventions alignment
8. Agent type taxonomy
9. Time series metric granularity

## What's Next

Per the [RFC process](https://github.com/datahub-project/datahub/blob/master/docs/rfc.md):
1. Community discussion and feedback on this PR
2. Address questions and iterate on design
3. Final comment period (7 days) before acceptance
4. Once accepted, update filename with PR number and mark as "active"

## Questions?

Please review the full RFC and share your thoughts on:
- The overall approach (especially the new AgentSkill entity)
- Git-based skill registry architecture
- Specific design decisions
- The unresolved questions
- Any concerns or alternative approaches

Looking forward to your feedback!

### DIFF
--- a/docs/rfcs/active/16012-agents-metadata-model.md
+++ b/docs/rfcs/active/16012-agents-metadata-model.md
@@ -190,7 +190,7 @@ The following are explicitly out of scope for this RFC (may be addressed in futu
 
 An Agent represents an AI system that can perceive, reason, and act to achieve specific goals.
 
-**URN Format**: `urn:li:agent:<agent_id>`
+**URN Format**: `urn:li:agent:<agent_id>` (agent ID includes version to ensure uniqueness; see [Unresolved Question #4](#4-agent-version-management))
 
 **Core Properties**:
 
@@ -348,6 +348,16 @@ Skills are authored in git (SKILL.md files), ingested into DataHub for discovery
 - Preferred ingestion methods for agents and skills (git scanning, manifest files, API)?
 - How should evaluation metrics flow into DataHub?
 
+### Static Lineage vs Runtime Lineage
+
+There are two distinct types of agent-to-data relationships:
+
+**Static lineage** represents the declared/configured relationships between agents and their dependencies — which tools an agent *can* use, which datasets it *has access to*, which models it *is configured with*. These are modeled as relationships in the metadata graph and are the focus of this RFC.
+
+**Runtime lineage** captures what actually happened during a specific agent invocation — which tools were *actually called*, which datasets were *actually queried*, and what data flowed where. This is more of an audit log than a lineage graph, since it is high-cardinality, temporal, and input-dependent.
+
+**Our approach**: DataHub models static lineage in the metadata graph. Runtime traces are captured externally (via OpenTelemetry or similar) and linked from the agent entity. We could explore adding a timeseries aspect to capture runtime observations (e.g., tool calls, datasets accessed), but if the primary use cases are aggregation-oriented (e.g., "how often does this agent call tool X?" or "which datasets did this agent access last week?"), it may be more practical to store pre-aggregated summaries rather than raw high-scale event data.
+
 ### Key Concepts
 
 **Agent Card**: Similar to "dataset card" - a profile page showing:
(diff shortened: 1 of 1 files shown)


jjoyce0510:
Thinking about this a bit.. If we are working towards inference time "lineage" and connecting data intensive agents to structured + unstructured data sources, should we also include some sort of agent > tool call > dataset or document lineage path? (Maybe with a query in there as well)

AdrianMachado:
> Thinking about this a bit.. If we are working towards inference time "lineage" and connecting data intensive agents to structured + unstructured data sources, should we also include some sort of agent > tool call > dataset or document lineage path? (Maybe with a query in there as well)

Yeah, I was about to say this too. But more like an audit log than being in the lineage graph tbh

docs/rfcs/active/16012-agents-metadata-model.md
```diff
@@ -0,0 +1,574 @@
...
+
+3. **Agent Skill Entity** (high-level capabilities):
+
+   - Name, description, instructions (specialized prompts and context)
+   - Source repository (git URL)
+   - Domain category and when to use
+   - Which tools and models the skill requires
+   - Which agents have this skill
+
+4. **Relationships**:
```
sgomezvillamor, line 139: What about Agent-to-Agent relationship? Agent composition (Agent A calls Agent B) is common pattern.
Or will this use case be modelled via the Agent Tool? Agent-to-Tool-toAgent

docs/rfcs/active/16012-agents-metadata-model.md
```diff
@@ -0,0 +1,574 @@
...
+
+---
+
+### 4. Agent Version Management
+
+**Question**: How to handle agent versions?
+
+**Options**:
+
+- **A**: Version in URN: `urn:li:agent:customer-support:v1.2.3`
```
sgomezvillamor: Version in URN becomes critical

Different versions can run simultaneously, potentially using different models or relationships. While running in parallel, they may also produce different usage metrics.

Related to this, although I’m not sure it applies in this specific case, the environment is another classic discriminator that is part of the identity for most of our entities.

georgehurley-max:
The way we have been thinking about the Agent Entity is quite closely aligned to the Data Product concept in DataHub today. An Agent is a collection of context, skills, MCPs with the agent entity bringing these together. However, different to data products, an MCP for example may be used by several agents. 

Additionally, it would be great to give users visability of the evals as a measure of "quality" for the overall agent. Then similar again to data products, what is the ownership, domain etc - the DataHub bread and butter!

docs/rfcs/active/16012-agents-metadata-model.md
shirshanka: I was thinking we would use the VersionSet concept that we already introduced as part of logical datasets work - do you think that would work? The urn would just need to make sure that the version was encoded in some form - either a hash or a readable string.

sgomezvillamor: Sure, `VersionSet` as an aspect makes sense for holding metadata and enabling filtering.
Anyway and regardless of whether the version is a hash or a human-readable string, it should still be part of the entity identity.

jjoyce0510:
> Additionally, it would be great to give users visability of the evals as a measure of "quality" for the overall agent. Then similar again to data products, what is the ownership, domain etc - the DataHub bread and butter!

George -- How would these quality metrics be recorded or produced?

We are seeing a range of tools emerging to specifically help with fine-grained AI tracing and quality evaluation. 
Do you see this as solving a different problem?

SajSheth08:
Agents should be a type of asset that can be cataloged in DataHub. This would enable DataHub to have end to end lineage and discoverability for all things AI and Agentic.

PR #16031, opened 2026-01-30
feat(ingest/clickhouse): add column-level lineage support

kyungsoo-datahub:
Implement column-level lineage (CLL) for ClickHouse views by extracting view definitions from system.tables.create_table_query. Previously, view definitions were empty because the ClickHouse SQLAlchemy dialect didn't implement get_view_definition, preventing the SQL parser from extracting column lineage.

Changes:
- Add get_view_definition() to extract CREATE VIEW statements from ClickHouse system tables
- Monkey-patch ClickHouseDialect with the new method
- Add LINEAGE_FINE capability declaration
- Add unit tests for the new functionality

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/sql/clickhouse.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/sql/clickhouse.py
@@ -809,7 +809,10 @@ ORDER BY event_time ASC
                 query_hash=str(row.get("normalized_query_hash", "")),
             )
         except Exception as e:
-            logger.debug(f"Failed to parse query log row: {e}")
+            self.report.warning(
+                "Failed to parse query log row",
+                context=f"query_id={row.get('query_id', 'unknown')}: {e}",
+            )
             return None
 
     def get_workunits_internal(self) -> Iterable[Union[MetadataWorkUnit, SqlWorkUnit]]:


gabe-lyons:
@kyungsoo-datahub does this include table<>table lineage from the audit log?https://clickhouse.com/docs/cloud/security/audit-logging/database-audit-log

metadata-ingestion/src/datahub/ingestion/autogenerated/capability_summary.json
```diff
@@ -365,7 +369,7 @@
           "supported": true
         }
       ],
-      "classname": "datahub.ingestion.source.sql.clickhouse.ClickHouseSource",
+      "classname": "datahub.ingestion.source.sql.clickhouse.source.ClickHouseSource",
```
skrydal: What's the point of introducing the `__init__.py` file, exporting `ClickHouseSource` under old path, yet changing it here to contain `source.` infix?

metadata-ingestion/src/datahub/ingestion/source/sql/clickhouse.py
```diff
@@ -152,6 +189,78 @@ class ClickHouseConfig(
...
+            if not _is_valid_username(username):
+                raise ValueError(
+                    f"Invalid username '{username}' in query_log_deny_usernames. "
+                    "Usernames must contain only alphanumeric characters, "
+                    "underscores, and hyphens."
+                )
+        return v
+
+    @cached_property
+    def _compiled_temporary_tables_pattern(self) -> List[re.Pattern[str]]:
```
skrydal, line 227: good idea

metadata-ingestion/src/datahub/sql_parsing/sql_parsing_aggregator.py
```diff
@@ -1180,7 +1181,32 @@ def _process_view_definition(
         elif parsed.debug_info.column_error:
             self.report.num_views_column_failed += 1
 
-        query_fingerprint = self._view_query_id(view_urn)
+        # Determine the actual downstream URN.
+        # For ClickHouse MVs with TO clause, the SQL parser identifies the TO table
+        # as the output, which differs from the view_urn. Use that instead.
+        # Only apply this for ClickHouse MVs to avoid impacting other databases.
```
skrydal: I find the abbreviation `MV` to be confusing, could you just use full term (materialized view) instead?

metadata-ingestion/src/datahub/ingestion/source/sql/sql_common.py
```diff
@@ -1222,12 +1231,21 @@ def _process_view(
...
+                f"[VIEW-DEF] Adding to aggregator: {dataset_name} -> {dataset_urn}"
+            )
             self.aggregator.add_view_definition(
                 view_urn=dataset_urn,
                 view_definition=view_definition,
                 default_db=default_db,
                 default_schema=default_schema,
             )
+        elif not view_definition:
+            logger.warning(f"[VIEW-DEF] Empty view definition for {dataset_name}")
```
skrydal: Imho this would be clearer
```suggestion
            logger.warning(f"[VIEW-DEF] No view definition for {dataset_name}")
```

metadata-ingestion/src/datahub/ingestion/source/sql/sql_common.py
```diff
@@ -1222,12 +1231,21 @@ def _process_view(
...
+        elif not self.config.include_view_lineage:
+            logger.debug(
+                f"[VIEW-DEF] Skipping {dataset_name}: include_view_lineage=False"
+            )
```
skrydal: I think it would be wasteful to log this line for every single view if that flag is set to `false`. It would make logs less readable. Why not just log it once before iterating through the views?

metadata-ingestion/src/datahub/sql_parsing/sql_parsing_aggregator.py
```diff
@@ -1180,7 +1181,32 @@ def _process_view_definition(
         elif parsed.debug_info.column_error:
             self.report.num_views_column_failed += 1
 
-        query_fingerprint = self._view_query_id(view_urn)
+        # Determine the actual downstream URN.
+        # For ClickHouse MVs with TO clause, the SQL parser identifies the TO table
```
skrydal: What is TO table?

metadata-ingestion/src/datahub/sql_parsing/sql_parsing_aggregator.py
```diff
@@ -1180,7 +1181,32 @@ def _process_view_definition(
...
+            is_clickhouse_mv_with_to
+            and parsed.out_tables
+            and len(parsed.out_tables) == 1
+        ):
+            parsed_downstream = parsed.out_tables[0]
+            if parsed_downstream != view_urn:
+                downstream_urn = parsed_downstream
+                logger.debug(f"ClickHouse MV {view_urn} has TO table: {downstream_urn}")
+
+        query_fingerprint = self._view_query_id(downstream_urn)
```
skrydal: We should avoid leaking source-specific logic into this universal class. Is there no way of injecting that from the source itself?

metadata-ingestion/src/datahub/sql_parsing/sql_parsing_aggregator.py
```diff
@@ -1203,7 +1229,89 @@ def _process_view_definition(
...
+                        ColumnRef(
+                            table=mv_urn,
+                            column=column_name,
+                        )
+                    ],
+                    logic=None,
+                )
+            )
+
+        return mv_to_table_cll
```
skrydal: Same concern as above

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
```diff
@@ -707,6 +785,7 @@ def _sqlglot_force_column_normalizer(
             # - running the full pre-type annotation optimizer
 
             # logger.debug("Schema: %s", sqlglot_db_schema.mapping)
+            logger.debug("[CLL] Starting sqlglot optimizer")
```
skrydal: will this be logged for every single query being parsed?

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
```diff
@@ -496,6 +549,19 @@ def _maybe_resolve_tsql_alias(
         return table
 
     # Generate table-level lineage.
+    # ClickHouse-specific: For CREATE MATERIALIZED VIEW ... TO target_table,
```
skrydal: Unlike `sql_parsing_aggregator`, here we, unfortunately, already introduced logic leak from sources into the generic library. Please add a check for the dialect (`clickhouse`) when introducing new filters which are clickhouse specific (several lines down in the code).
I have hard time understanding filter `and not clickhouse_to_tables` logic. Could you explain it?

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
```diff
@@ -551,6 +620,15 @@ def _maybe_resolve_tsql_alias(
...
                 if not isinstance(table.parent, sqlglot.exp.Drop)
+                # ClickHouse: Filter out dictionary references in DICTGET() and similar functions
+                and not _is_in_clickhouse_dict_function(table)
+                # ClickHouse: Filter out TO table in CREATE MATERIALIZED VIEW ... TO target_table
+                and not isinstance(table.parent, sqlglot.exp.ToTableProperty)
+                # ClickHouse: Filter out MV name (in Schema) when TO table exists
+                and not (
+                    clickhouse_to_tables
+                    and isinstance(table.parent, sqlglot.exp.Schema)
+                )
```
skrydal: Shouldn't this happen only if the dialect is clickhouse and not always?

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
```diff
@@ -822,22 +908,42 @@ def _select_statement_cll(
...
-                statement,
-                dialect=dialect,
-                scope=root_scope,
-                trim_selects=False,
-                # We don't need to pass the schema in here, since we've already qualified the columns.
-            )
+            # Log progress every 10 columns or for first/last column
+            if col_idx == 0 or col_idx == total_columns - 1 or col_idx % 10 == 0:
+                logger.debug(
+                    f"[CLL] Processing column {col_idx + 1}/{total_columns}: {output_col}"
```
skrydal: I have concerns about this lines being too verbose.

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
```diff
@@ -881,11 +987,22 @@ def _select_statement_cll(
...
             )
 
+        logger.debug(
+            f"[CLL] Finished processing all columns, generated {len(column_lineage)} lineage entries"
+        )
         # TODO: Also extract referenced columns (aka auxiliary / non-SELECT lineage)
     except (sqlglot.errors.OptimizeError, ValueError, IndexError) as e:
+        logger.warning(
+            f"[CLL] sqlglot exception during CLL extraction for {output_table}: {e}"
+        )
```
skrydal: Considering that we are raising another exception immediately, do we want to really add a log line here?

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
```diff
@@ -1813,20 +2001,45 @@ def _sqlglot_lineage_inner(
...
-        except KeyError as e:
-            # When this happens, it's usually because of things like PIVOT where we can't
-            # really go up the scope chain.
-            logger.debug(
-                f"Failed to translate column lineage to urns: {e}", exc_info=True
-            )
-            debug_info.column_error = e
+            if not column_lineage_urns:
+                # All column lineages were filtered out
+                column_lineage_urns = None
```
skrydal: In case all columns are filtered, why do we need to set column_lineage_urns to None, but only for `clickhouse` dialect, and for others we return `[]` in such case?

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
skrydal, line 1727: I don't think we should be mixing source-specific logic into a function which previously lacked such mixes. Could you refactor that? Maybe clickhouse logic could be extracted into a separate function, or table-name-urn-mapping could be injected as a general mapper instead?

metadata-ingestion/tests/unit/sql_parsing/test_sqlglot_lineage.py
skrydal, line 1: Those are nice tests, could you add an "ascii-picture" displaying expected lineage to the comments? Might be easier to read than the golden files...

metadata-ingestion/tests/unit/test_clickhouse_source.py
```diff
@@ -89,3 +98,408 @@ def test_clickhouse_uri_https_backward_compatibility():
...
+    mock_connection.execute.return_value.fetchone.return_value = None
+
+    result = get_view_definition(
+        mock_dialect, mock_connection, "nonexistent_view", schema="db1"
+    )
+
+    assert result == ""
+
+
+def test_get_view_definition_without_schema():
```
skrydal: How is this test effectively different than `test_get_view_definition_with_schema`? We are not executing the query, we are mocking the connection.

metadata-ingestion/tests/unit/test_clickhouse_source.py
```diff
@@ -89,3 +98,408 @@ def test_clickhouse_uri_https_backward_compatibility():
...
+        }
+    )
+
+    patterns = config._compiled_temporary_tables_pattern
+
+    # Default patterns
+    assert len(patterns) == 4
+
+    # Test pattern matching
+    assert patterns[0].match("_temp_table")  # ^_.*
```
skrydal: Do we really care, functionally, in what order patterns are executed?
The test, as it is, effectively tests whether default is:
```
default=[
            r"^_.*",  # Tables starting with underscore
            r".*\.tmp_.*",  # Tables with tmp_ prefix
            r".*\.temp_.*",  # Tables with temp_ prefix
            r".*\._inner.*",  # Inner tables for materialized views
        ]
```
and not (for example):
```
default=[
            r".*\.tmp_.*",  # Tables with tmp_ prefix
            r"^_.*",  # Tables starting with underscore
            r".*\.temp_.*",  # Tables with temp_ prefix
            r".*\._inner.*",  # Inner tables for materialized views
        ]
```

Which is the same from our business logic point of view, why would we want to ensure that?

skrydal: Why not using function checking for temporary tables itself?

metadata-ingestion/tests/unit/test_clickhouse_source.py
```diff
@@ -89,3 +98,408 @@ def test_clickhouse_uri_https_backward_compatibility():
...
+            "host_port": "localhost:8123",
+            "query_log_deny_usernames": [
+                "system",
+                "default",
+                "admin-user",
+                "test_user123",
+            ],
+        }
+    )
+    assert config.query_log_deny_usernames == [
```
skrydal: I would consider making this check ignore order of the list.

metadata-ingestion/tests/unit/test_clickhouse_source.py
```diff
@@ -89,3 +98,408 @@ def test_clickhouse_uri_https_backward_compatibility():
...
+    config = ClickHouseConfig.model_validate(
+        {
+            "host_port": "localhost:8123",
+            "include_query_log_lineage": True,
+            "start_time": "2024-01-01T00:00:00Z",
+            "end_time": "2024-01-08T00:00:00Z",
+        }
+    )
+
+    with patch.object(ClickHouseSource, "__init__", lambda x, y, z: None):
```
skrydal: Why do we need to patch init function?

metadata-ingestion/tests/unit/test_clickhouse_source.py
```diff
@@ -89,3 +98,408 @@ def test_clickhouse_uri_https_backward_compatibility():
...
+        source = ClickHouseSource.__new__(ClickHouseSource)
+        source.config = config
+
+        query = source._build_query_log_query()
+
+        # Verify basic query structure
+        assert "SELECT" in query
```
skrydal: Imho if you want to check query generation, which is determined by the variables you define explicitly in this unit test, it would be better to just check entire string.

metadata-ingestion/tests/unit/test_clickhouse_source.py
```diff
@@ -89,3 +98,408 @@ def test_clickhouse_uri_https_backward_compatibility():
...
+        row = {
+            "query": "SELECT 1",
+            "event_time": datetime(2024, 1, 15, 10, 30, 0),  # naive datetime
+        }
+
+        result = source._parse_query_log_row(row)
+
+        # Should still work, timestamp may be treated as local time
+        assert result is not None
+        assert result.query == "SELECT 1"
```
skrydal: I am not a big fan of testing what query was generated. Why don't we just create an integration test with clickhouse docker image and just try queries directly?

metadata-ingestion/setup.py
```diff
@@ -529,7 +529,7 @@
     },
     "bigquery-slim": bigquery_common,
     "bigquery-queries": sql_common | bigquery_common | sqlglot_lib,
-    "clickhouse": sql_common | clickhouse_common,
+    "clickhouse": sql_common | usage_common | clickhouse_common | sqlglot_lib,
```
skrydal: Is there any particular reason for introducing `sqlglot_lib` directly to `clickhouse`? It is already included by `sql_common` and clickhouse source does not use sqlglot directly.
Also the usage_common is part of separate extras - `clickhouse-usage`. Why is it necessary now in `clickhouse`?

metadata-ingestion/tests/unit/test_clickhouse_source.py
```diff
@@ -89,3 +98,408 @@ def test_clickhouse_uri_https_backward_compatibility():
...
+            "include_usage_statistics": True,
+            "query_log_table": "custom.query_log_view",
+            "query_log_deny_usernames": ["system", "default"],
+        }
+    )
+
+    assert config.include_query_log_lineage is True
+    assert config.include_usage_statistics is True
+    assert config.query_log_table == "custom.query_log_view"
+    assert config.query_log_deny_usernames == ["system", "default"]
```
skrydal: I feel those tests are too trivial. We are effectively testing `pydantic` library...

metadata-ingestion/src/datahub/ingestion/autogenerated/capability_summary.json
kyungsoo-datahub: This was from an earlier iteration where I explored creating a clickhouse/ package directory. After the rebase, this change is no longer present - the source remains as a single clickhouse.py file and capability_summary.json correctly references datahub.ingestion.source.sql.clickhouse.ClickHouseSource without the source. infix.

metadata-ingestion/src/datahub/sql_parsing/sql_parsing_aggregator.py
kyungsoo-datahub: Revised.

metadata-ingestion/src/datahub/ingestion/source/sql/sql_common.py
kyungsoo-datahub: You're right, that was debugging code left in during implementation. It's already been removed in the latest version - sorry for including incomplete code in the earlier review. The [VIEW-DEF] prefixed logs have all been cleaned up.

metadata-ingestion/src/datahub/sql_parsing/sql_parsing_aggregator.py
kyungsoo-datahub: ClickHouse materialized views have a unique syntax: CREATE MATERIALIZED VIEW mv TO target_table AS SELECT... where data flows to target_table, not the view itself. This is different from other databases where views are the output.

metadata-ingestion/src/datahub/sql_parsing/sql_parsing_aggregator.py
kyungsoo-datahub: I agree with you. Moved to clickhouse.py.

metadata-ingestion/src/datahub/sql_parsing/sql_parsing_aggregator.py
kyungsoo-datahub: Moved to clickhouse.py. Thanks.

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
kyungsoo-datahub: You're right, this would have been too verbose. These debug logs ([CLL] prefixed) were added during development to trace column-level lineage issues and have been removed in the latest version.

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
kyungsoo-datahub: Good point. _clickhouse_extract_to_tables already has a dialect check (line 453-454) and returns empty for non-ClickHouse. For _is_clickhouse_non_input, while it effectively returns False for non-ClickHouse (since DICTGET and ToTableProperty are ClickHouse-specific).

kyungsoo-datahub: ClickHouse MV syntax:

CREATE MATERIALIZED VIEW mv_name TO target_table AS SELECT * FROM source
- Regular MATERIALIZED VIEW (no TO): mv_name is the output  
- MATERIALIZED VIEW with TO clause: target_table is the output, not mv_name                                                          
                                                                                                                        
The condition and ```not clickhouse_to_tables means```: "Only use the view name as output if there's no TO table." When a TO table exists, we use that instead (via clickhouse_to_tables). 

The _is_clickhouse_non_input filter removes target_table and mv_name from inputs—only source should be an input.

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
kyungsoo-datahub: Added explicit dialect check. Thank you for pointing out.

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
kyungsoo-datahub: Thank you for checking. Removed.

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
kyungsoo-datahub: No, it's for development logs. Removed.

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
kyungsoo-datahub: Revised. 

The behavior is consistent:
- If column_lineage is empty initially -> column_lineage_urns stays None
- If column_lineage is non-empty but gets filtered to [] -> column_lineage_urns becomes [] (same as other dialects)

There's no special handling that sets None when columns are filtered out.

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
kyungsoo-datahub: I agree with you. Moved to clickhouse.py. Thank you for the suggestions.

metadata-ingestion/tests/unit/sql_parsing/test_sqlglot_lineage.py
kyungsoo-datahub: Thank you for the suggesion. Each test has an ASCII diagram in its docstring showing the expected lineage as suggested.

metadata-ingestion/tests/unit/test_clickhouse_source.py
kyungsoo-datahub: You're right - they were testing mocked behavior, not real functionality. Removed.

metadata-ingestion/tests/unit/test_clickhouse_source.py
kyungsoo-datahub: Revised. Thanks.

metadata-ingestion/tests/unit/test_clickhouse_source.py
kyungsoo-datahub: The __init__ is patched to avoid needing a full PipelineContext (database connections, graph, etc.) when we only need to test _build_query_log_query(), which just uses self.config. Please, let me know if you have other suggestions. Thanks.

metadata-ingestion/tests/unit/test_clickhouse_source.py
kyungsoo-datahub: Thank you for the suggestion. Revised.

metadata-ingestion/tests/unit/test_clickhouse_source.py
kyungsoo-datahub: Thank you for the suggestion. They're removed.

metadata-ingestion/setup.py
kyungsoo-datahub: Revised to ```"clickhouse": sql_common | clickhouse_common,```

metadata-ingestion/tests/unit/test_clickhouse_source.py
kyungsoo-datahub: Agreed, these were removed. The current test file doesn't have these trivial config tests.

metadata-ingestion/tests/unit/sql_parsing/test_sqlglot_lineage.py
skrydal: Thank you, it is much clearer now.

metadata-ingestion/tests/unit/test_clickhouse_source.py
skrydal: I see it was removed.

metadata-ingestion/src/datahub/ingestion/source/sql/clickhouse.py
```diff
@@ -408,27 +541,279 @@ class ClickHouseSource(TwoTierSQLAlchemySource):
...
+                user=CorpUserUrn(user) if user else None,
+                # Don't pass current_database as default_db. ClickHouse uses 2-level
+                # naming (database.table), but sqlglot expects 3-level (database.schema.table).
+                # Passing current_database causes sqlglot to prepend it to already-qualified
+                # names, creating incorrect URNs like "default.analytics_marts.table".
+                default_db=None,
+                query_hash=str(row.get("normalized_query_hash", "")),
+            )
+        except Exception as e:
+            logger.debug(f"Failed to parse query log row: {e}")
```
skrydal: Shouldn't we put it in the report?

metadata-ingestion/src/datahub/ingestion/source/sql/sql_common.py
```diff
@@ -1166,6 +1166,24 @@ def get_view_default_db_schema(
         # into a dataset identifier string and then parsed back out.
         return self.get_db_schema(dataset_identifier)
 
+    def _add_view_to_aggregator(
```
skrydal, line 1169: awesome

metadata-ingestion/src/datahub/ingestion/source/sql/clickhouse.py
kyungsoo-datahub: Good call. Revised.

metadata-ingestion/src/datahub/ingestion/source/sql/sql_common.py
kyungsoo-datahub: Thanks!

PR #16032, opened 2026-01-30
feat(snowflake agent): add generators for snowflake sql agent

nwadams:
```
datahub agent create snowflake \
  --sf-account YOUR_ACCOUNT \
  --sf-user YOUR_USER \
  --sf-role YOUR_ROLE \
  --sf-warehouse YOUR_WAREHOUSE \
  --sf-database YOUR_DATABASE \
  --sf-schema YOUR_SCHEMA \
  --datahub-url https://your-datahub.acryl.io \
  --datahub-token YOUR_TOKEN \
  --enable-mutations
```
snowflake-sql

snowflake-intellgence

### DIFF
--- a/metadata-ingestion/tests/unit/test_entrypoints.py
+++ b/metadata-ingestion/tests/unit/test_entrypoints.py
@@ -1,0 +1,74 @@
+import pytest
+from click.testing import CliRunner
+
+
+def test_agent_shim_command_behavior():
+    """Test that the agent shim command displays the correct error message."""
+    from datahub.cli.cli_utils import make_shim_command
+
+    shim_command = make_shim_command("agent", "run `pip install datahub-agent-context`")
+
+    runner = CliRunner()
+    result = runner.invoke(shim_command, [])
+
+    assert result.exit_code == 1
+    assert "missing dependencies" in result.output
+    assert "run `pip install datahub-agent-context`" in result.output
+
+
+def test_agent_command_exists():
+    """Test that agent command is registered in the CLI."""
+    import datahub.entrypoints
+
+    # Verify agent command was added (either real or shim)
+    assert "agent" in datahub.entrypoints.datahub.commands
+
+
+def test_agent_command_shows_error_or_help():
+    """Test that agent command either works or shows helpful error."""
+    import datahub.entrypoints
+
+    runner = CliRunner()
+    result = runner.invoke(datahub.entrypoints.datahub, ["agent", "--help"])
+
+    # Either the command works (exit code 0) or it's a shim (exit code 1 with helpful message)
+    if result.exit_code == 1:
+        assert "missing dependencies" in result.output
+        assert "pip install datahub-agent-context" in result.output
+    else:
+        # Real command should show help
+        assert result.exit_code == 0
+
+
+@pytest.mark.parametrize(
+    "command_name",
+    ["agent", "actions", "lite"],
+)
+def test_optional_commands_exist(command_name):
+    """Test that optional commands (agent, actions, lite) are always registered."""
+    import datahub.entrypoints
+
+    # Verify the command exists in the CLI
+    assert command_name in datahub.entrypoints.datahub.commands
+
+
+@pytest.mark.parametrize(
+    "command_name,install_message",
+    [
+        ("agent", "pip install datahub-agent-context"),
+        ("actions", "pip install acryl-datahub-actions"),
+        ("lite", "pip install 'acryl-datahub[datahub-lite]'"),
+    ],
+)
+def test_shim_commands_show_helpful_error(command_name, install_message):
+    """Test that shim commands created by make_shim_command show helpful error messages."""
+    from datahub.cli.cli_utils import make_shim_command
+
+    shim_command = make_shim_command(command_name, f"run `{install_message}`")
+
+    runner = CliRunner()
+    result = runner.invoke(shim_command, [])
+
+    assert result.exit_code == 1
+    assert "missing dependencies" in result.output
+    assert install_message in result.output


alexsku (commented):
CLAUDE: 

## PR Review Summary

### Critical Issues

**1. Cortex Agent doesn't respect `--no-enable-mutations` flag**
`generators/cortex_agent.py` - The agent YAML always includes all 20 tools even when mutations are disabled. The agent will reference UDFs that don't exist.

**2. Unused `--datahub-ips` parameter**
`snowflake.py:193-196`, `agent_cli.py:53-57` - This option is defined with a default value but never used. Network rules use `HOST_PORT` with domain name instead. Should either be removed or implemented.

### Should Fix

**3. Token exposure in generated SQL files**
`generators/configuration.py:77-83` - The DataHub token is written in plaintext to `00_configuration.sql`. Consider adding a warning about file sensitivity or suggesting users add generated files to `.gitignore`.

**4. Missing URL validation**
`snowflake.py:237-254` - No validation that `--datahub-url` is a valid URL format before use.

**5. Brittle GRANT statement generation**
`generate_udfs.py:138-202` - Large if/elif chain must stay synchronized with UDF signatures. Consider generating GRANT signatures programmatically from the parameter definitions.

**6. Dead `error_count` code**
`snowflake.py:44, 66-68` - `error_count` is initialized to 0 but never incremented, making the error check dead code.

**7. Broken README link**
`README.md:176` - Relative path `../../../../../datahub-agent-context/README.md` likely doesn't resolve correctly.

**8. Hardcoded read-only UDF count**
`generate_udfs.py:123` - `read_ops_count = 9` is a magic number. Should be calculated dynamically.

### Non-blocking

**9. `type: ignore` on import**
`agent_cli.py:138` - The `# type: ignore[import-not-found]` suggests static analysis can't resolve the import. Worth investigating but not blocking.

alexsku (approved):

metadata-ingestion/src/datahub/ai/snowflake/generators/cortex_agent.py
```diff
@@ -0,0 +1,650 @@
...
+    - tool_spec:
+        type: "generic"
+        name: "get_me"
+        description: "Get information about the authenticated user (profile, groups, privileges). This tool takes no parameters."
+        input_schema:
+          type: "object"
+          properties: {{}}
+
+    # SQL Executor
+    - tool_spec:
```
jjoyce0510: We have to provide a function for this? Snowflake doesn't offer it built in ?

datahub-agent-context/src/datahub_agent_context/snowflake/generators/stored_procedure.py
```diff
@@ -0,0 +1,87 @@
...
+--
+-- Prerequisites:
+-- - Run 00_configuration.sql first to set variables
+-- ============================================================================
+
+USE DATABASE IDENTIFIER($SF_DATABASE);
+USE SCHEMA IDENTIFIER($SF_SCHEMA);
+USE WAREHOUSE IDENTIFIER($SF_WAREHOUSE);
+
+-- Create the JavaScript stored procedure (using CREATE OR REPLACE, so no need to drop first)
```
jjoyce0510, line 20: Interesting. SO we actually provide the execute SQL tool, there's no existing tool we can use off the shelf for cortex?

metadata-ingestion/src/datahub/ai/snowflake/snowflake.py
```diff
@@ -0,0 +1,511 @@
...
+            for script_name, script_content in scripts:
+                click.echo(f"\n📝 Executing {script_name}...")
+                success = execute_sql_in_snowflake(conn, script_content, script_name)
+                if success:
+                    click.echo(f"  ✓ {script_name} completed successfully")
+                else:
+                    click.echo(f"  ✗ {script_name} failed", err=True)
+                    all_success = False
+                    break
+
```
jjoyce0510: IMO this function is a bit too long. I'd rather see it split into simple, well-tested names..

metadata-ingestion/tests/unit/ai/snowflake/test_generate_udfs.py
```diff
@@ -0,0 +1,226 @@
...
+
+        # Should have SHOW statements for all UDFs
+        assert "SHOW FUNCTIONS LIKE 'SEARCH_DATAHUB';" in sql
+        assert "SHOW FUNCTIONS LIKE 'GET_ENTITIES';" in sql
+        assert "SHOW FUNCTIONS LIKE 'GET_ME';" in sql
+
+    def test_generate_udfs_sql_verification_query(self) -> None:
+        """Test that verification query lists all functions."""
+        sql = generate_datahub_udfs_sql(include_mutations=True)
+
```
jjoyce0510, line 221: nice tests!

metadata-ingestion/src/datahub/entrypoints.py
```diff
@@ -165,6 +166,7 @@ def init(use_password: bool = False) -> None:
     click.echo(f"Written to {DATAHUB_CONFIG_PATH}")
 
 
+datahub.add_command(agent)
```
jjoyce0510: awesome...

jjoyce0510: any reason to do a try like we do below for datahub actions?

will users always have this installed?

jjoyce0510: is this code published in BOTH acryl-datahub and datahub-agent-context independently? 

any issues if users have installed both at the same time? just curious..

nwadams: As of now it is bundled as part of the acryl-datahub package. We could make it a separate optional addon that gets added like datahub actions if we want? Thoughts?

This code is just published in acryl-datahub. We depend on `datahub-agent-context ` at snowflake runtime, but not at the sql generation stage.

metadata-ingestion/src/datahub/ai/snowflake/snowflake.py
nwadams: will do

datahub-agent-context/src/datahub_agent_context/snowflake/generators/stored_procedure.py
nwadams: Apparently. This was one of the things marty discovered, the only direct way to execute sql as part of the agent natively is the a analyst with a semantic view set up. It kind of sucks.

PR #16034, opened 2026-01-31
fix(preset): fix Preset/Superset dashboard lineage and database alias transformation

kyungsoo-datahub:
- Empty chart arrays: Fetch position_json from detail API (/api/v1/dashboard/{id})                                    
- Connection pool exhaustion: Set pool size to max_threads + 10                                                       
- KeyError in SQL parsing: Use .get() for tables not in mapping                                                       
- Poor error messages: Show missing table_name, database_id, dataset_id                                               
- database_alias not applied to SQL-parsed URNs: Add _apply_database_alias_to_urn() to transform lineage URNs

### DIFF
--- a/metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
+++ b/metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
@@ -1810,20 +1810,8 @@ def _sqlglot_lineage_inner(
     # TODO: Can we generate a common WHERE clauses section?
 
     # Convert TableName to urns.
-    in_urns = sorted(
-        {
-            table_name_urn_mapping[table]
-            for table in tables
-            if table in table_name_urn_mapping
-        }
-    )
-    out_urns = sorted(
-        {
-            table_name_urn_mapping[table]
-            for table in modified
-            if table in table_name_urn_mapping
-        }
-    )
+    in_urns = sorted({table_name_urn_mapping[table] for table in tables})
+    out_urns = sorted({table_name_urn_mapping[table] for table in modified})
     column_lineage_urns = None
     if column_lineage:
         try:


treff7es:
@kyungsoo-datahub it seems like tests are failing

metadata-ingestion/src/datahub/ingestion/source/superset.py
```diff
@@ -344,7 +344,13 @@ def login(self) -> requests.Session:
             allowed_methods=RETRY_ALLOWED_METHODS,
             raise_on_status=False,
         )
-        adapter = HTTPAdapter(max_retries=retry_strategy)
+        # Set connection pool size to match max_threads to avoid "Connection pool is full" warnings
+        pool_size = self.config.max_threads + 10  # Add buffer for concurrent requests
```
treff7es: Can we extract out `10` to const?

metadata-ingestion/src/datahub/ingestion/source/superset.py
```diff
@@ -1155,6 +1201,39 @@ def gen_dataset_urn(self, datahub_dataset_name: str) -> str:
...
+            # Replace at start of dataset name (after the comma following platform)
+            urn = urn.replace(f",{old_db}.", f",{new_db}.")
+            urn = urn.replace(f",{old_db_lower}.", f",{new_db}.")
+
+        # Diagnostic logging for database_alias URN transformation
+        if original_urn != urn:
+            logger.info(
+                f"[DATABASE_ALIAS] _apply_database_alias_to_urn: transformed {original_urn} -> {urn}"
+            )
+        elif "clickhouse" in original_urn.lower():
```
treff7es: Why does this only need ClickHouse?

metadata-ingestion/src/datahub/ingestion/source/superset.py
kyungsoo-datahub: Revised. Thanks.

metadata-ingestion/src/datahub/ingestion/source/superset.py
kyungsoo-datahub: It was introduced for dev debugging. Removed. Thank you for catching this.

PR #16035, opened 2026-01-31
fix(redshift): fix RTRIM bug causing malformed SQL keywords

kyungsoo-datahub:
## Changes

  **1. RTRIM Fix** (`query.py`)
  - Changed `LISTAGG(RTRIM(text))` → `RTRIM(LISTAGG(text))` in 5 SQL queries
  - Preserves spaces at 200-char segment boundaries

  **2. Query Preprocessing** (`redshift_preprocessing.py` - new)
  - Fixes compound keywords: `casewhen`→`case when`, `groupby`→`group by`, etc.
  - Fixes AWS DMS: missing FROM clause, password redaction issues

  **3. Sigma Temp Table Detection** (`lineage.py`)
  - Filters `sigma.t_mat_*` tables from lineage

  **4. Reporting** (`report.py`)
  - Added CLL parse failure counter
  - Added permission denied tracking


gabe-lyons:
@kyungsoo-datahub are these changes really redshift only? that is so strange that sigma would special case for redshift.

kyungsoo-datahub:
> gabe



> @kyungsoo-datahub are these changes really redshift only? that is so strange that sigma would special case for redshift.

Yes, Redshift only for now. Sigma handles SQL dialects differently per warehouse
(https://help.sigmacomputing.com/docs/region-warehouse-and-feature-support). We only observed these malformations in Redshift query logs. PostgreSQL-based platforms (PostgreSQL, AlloyDB) might have similar issues - if reported, we can extend the fix.

ligfx:
@kyungsoo-datahub it's incredibly unlikely that Sigma is generating invalid SQL for Redshift, and that Redshift would agree to actually execute it.

It's more likely that we're messing up the query text when reading it from Redshift. For instance, see`redshift/query.py`, where we `RTRIM` query parts before concatenating them with `LISTAGG(..., '')`: https://github.com/datahub-project/datahub/blob/8653848a0bc83022568c2699078ecc883c774ff2/metadata-ingestion/src/datahub/ingestion/source/redshift/query.py#L717-L723

ligfx:
@kyungsoo-datahub note that the current usage of the `STL_QUERY` table from before this PR can break queries anywhere, not just on well-defined places that we can handle with pre-processing 😕 Any query over 4,000 characters will be truncated and our lineage parsing will fail.

I'd recommend converting the current usages of `STL_QUERY` to use `STL_QUERYTEXT` instead. It'll require some stitching together of query text parts in Python, but that should fix the parsing issue once and for all. Something like this should work:

```python
cursor.execute("""
    select
        listagg(text) within group (order by sequence asc)
    from (
        select
            *,
            sequence / 327 as sequencebucket
        from stl_querytext
    )
    where query = ?
    group by query, sequencebucket
""", (my_query_id,))
return ''.join(row[0] for row in cursor.fetchall()).rstrip()
```

PR #16040, opened 2026-02-02
fix(airflow): use configured cluster environment for dataset URNs in lineage

aviraj-gour:
## Summary

Dataset URNs in Airflow lineage were always using PROD as the environment, even when the Airflow cluster was configured with a different environment (e.g., DEV). This fix ensures the configured cluster environment is used instead.

Linear issue: CUS-7620

## Changes

Updated both Airflow 2.x and 3.x plugins to use the configured cluster environment when generating dataset URNs:

- SQL parser patches (affects PostgreSQL, MySQL, Snowflake, Redshift, BigQuery, and other SQL operators)
- Athena OpenLineage patch
- Teradata OpenLineage patch

The fix uses this pattern:

```
def get_configured_env() -> str:
    """
    Get the configured DataHub cluster/environment name.

    Uses cached config from the listener when available, otherwise reads
    directly from config.

    Returns:
        The configured cluster name
    """
    from datahub_airflow_plugin.datahub_listener import get_airflow_plugin_listener

    listener = get_airflow_plugin_listener()
    if listener and listener.config:
        return listener.config.cluster

    # Fallback: listener disabled or failed to initialize
    return get_lineage_config().cluster
```

## Behaviour

Check enviroment Filter on top
### Before env for dataset in bigquery set to  PROD by default
Screenshot 2026-02-03 at 6 15 23 PM

### Now it correctly picks env from airflow config

Screenshot 2026-02-03 at 6 15 56 PM

### DIFF
--- a/metadata-ingestion-modules/airflow-plugin/src/datahub_airflow_plugin/_config.py
+++ b/metadata-ingestion-modules/airflow-plugin/src/datahub_airflow_plugin/_config.py
@@ -1,3 +1,4 @@
+import logging
 from enum import Enum
 from typing import TYPE_CHECKING, List, Optional, Union
 
@@ -7,6 +8,8 @@ from pydantic import Field
 import datahub.emitter.mce_builder as builder
 from datahub.configuration.common import AllowDenyPattern, ConfigModel
 from datahub_airflow_plugin._airflow_version_specific import IS_AIRFLOW_3_OR_HIGHER
+
+logger = logging.getLogger(__name__)
 
 if TYPE_CHECKING:
     from datahub_airflow_plugin.hooks.datahub import (
@@ -246,4 +249,7 @@ def get_configured_env() -> str:
         return listener.config.cluster
 
     # Fallback: listener disabled or failed to initialize
+    logger.debug(
+        "Listener or config not available, falling back to get_lineage_config()"
+    )
     return get_lineage_config().cluster


metadata-ingestion-modules/airflow-plugin/src/datahub_airflow_plugin/_config.py
```diff
@@ -227,3 +227,23 @@ def get_lineage_config() -> DatahubLineageConfig:
...
+    """
+    Get the configured DataHub cluster/environment name.
+
+    Uses cached config from the listener when available, otherwise reads
+    directly from config.
+
+    Returns:
+        The configured cluster name
+    """
+    from datahub_airflow_plugin.datahub_listener import get_airflow_plugin_listener
```
treff7es, line 245: Import should go to the top if there is no strong reason to be here

metadata-ingestion-modules/airflow-plugin/src/datahub_airflow_plugin/_config.py
```diff
@@ -227,3 +227,23 @@ def get_lineage_config() -> DatahubLineageConfig:
...
+
+    listener = get_airflow_plugin_listener()
+    if listener and listener.config:
+        return listener.config.cluster
+
+    # Fallback: listener disabled or failed to initialize
+    return get_lineage_config().cluster
```
treff7es, line 255: maybe we should log a debug message about the fallback

treff7es (approved):

metadata-ingestion-modules/airflow-plugin/src/datahub_airflow_plugin/_config.py
aviraj-gour: to avoid circular imports, `datahub_listener.py` imports function from `_config.py`

metadata-ingestion-modules/airflow-plugin/src/datahub_airflow_plugin/_config.py
aviraj-gour: added.

PR #16044, opened 2026-02-02
feat(dbt): Extract and emit stats from catalog.json

alfiyas-datahub:
## Summary

This PR enables DataHub to extract and display table statistics (row count, storage size) from dbt's catalog.json file in the Stats tab. Previously, these stats were available in the catalog file but were not being ingested or displayed.
Problem

When using the dbt CLI ingestion source (dbt_common/dbt_core), the stats field from catalog.json (containing num_rows and num_bytes) was being ignored. This meant users could not see row counts and table sizes in the DataHub Stats tab for their dbt models, even though this information was available in the dbt artifacts.

## Solution

Added row_count and size_in_bytes fields to the DBTNode dataclass to store catalog stats
Extract stats from catalog.json in extract_dbt_entities() function - parses num_rows and num_bytes stats, respecting the include flag
Emit DatasetProfile aspect when stats are available, which the UI already supports for displaying in the Stats tab

Closes: ING-1401


gabe-lyons:
@alfiyas-datahub how was this tested? can you show screenshots of dbt + snowflake when both have enabled profiling? How would this interference work?

alfiyas-datahub:
> @alfiyas-datahub how was this tested? can you show screenshots of dbt + snowflake when both have enabled profiling? How would this interference work?

@gabe-lyons The feature was tested by ingesting a dbt catalog.json containing num_rows: 100 and num_bytes: 2048 stats with catalog_stats: Yes enabled, and verified via GraphQL that the DatasetProfile aspect correctly stores rowCount: 100, sizeInBytes: 2048, and columnCount: 7. The attached screenshot shows the current stock v1.3.0 UI displaying "100 Rows" and "7 Columns" - after this PR merges, "Columns" will be replaced with "Storage Size" (2 KB). Regarding dbt + Snowflake profiling interference: there is none, because DatasetProfile is a timeseries aspect where each ingestion adds a new timestamped entry rather than overwriting, so profiles from both sources coexist as separate historical data points with the UI showing the most recent.
Screenshot 2026-02-10 at 12 09 42 PM

gabe-lyons:
@alfiyas-datahub what happens if dbt and bigquery both ingest profiling?

alfiyas-datahub:
> @alfiyas-datahub what happens if dbt and bigquery both ingest profiling?

@gabe-lyons No interference. DatasetProfile is a timeseries aspect, so each ingestion (dbt or BigQuery) adds a new timestamped entry rather than overwriting. Both profiles coexist as separate historical data points. The UI shows the most recent profile in "Latest" and all data points in the historical chart.

rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `e0e29fa`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/23180245431)

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
```diff
@@ -2257,6 +2280,35 @@ def create_dbt_platform_mces(
...
 
+            # Dataset profile (stats from catalog.json).
+            if (
+                self.config.entities_enabled.can_emit_node_type(node.node_type)
+                and self.config.entities_enabled.can_emit_catalog_stats
+            ):
+                if node.row_count is not None or node.size_in_bytes is not None:
+                    # Use catalog's generated_at timestamp if available, else fallback to now
+                    profile_timestamp = (
+                        self.report.catalog_generated_at or datetime.now()
```
kyungsoo-datahub: `datetime.now()` returns a naive (no timezone) datetime, but catalog_generated_at is parsed via dateutil.parser.parse which preserves UTC from the Z suffix in catalog.json's generated_at. This creates a naive/aware mismatch in the fallback path.

Use datetime.now(tz=timezone.utc) instead.

datahub-web-react/src/app/entityV2/shared/tabs/Dataset/Stats/StatsTabV2/highlights/LatestStats.tsx
```diff
@@ -40,14 +41,14 @@ const LatestStats = () => {
                     dataTestId="rows-card"
                 />
                 <Card
-                    title={columnCount !== undefined ? formatNumberWithoutAbbreviation(columnCount) : ''}
-                    subTitle={pluralize(columnCount || 0, 'Column')}
+                    title={formattedSize ? `${formattedSize.number} ${formattedSize.unit}` : ''}
```
kyungsoo-datahub: This replaces the "Columns" card with "Storage Size" for all datasets, not just dbt. Two concerns:

  1. Column count is removed from the highlights for every source (Snowflake, BigQuery, Postgres, etc.), not just dbt.
  2. Storage size is already displayed in its own dedicated section via StorageSizeGraph — adding it to the highlight cards duplicates existing UI.

  Consider keeping the Columns card as-is. The backend DatasetProfile emission is the valuable part of this PR, and the Stats tab already renders sizeInBytes through the Storage Size graph section.

metadata-ingestion/docs/sources/dbt/dbt_pre.md
```diff
@@ -2,6 +2,79 @@
 
 The `dbt` module ingests metadata from Dbt into DataHub. It is intended for production ingestion workflows and module-specific capabilities are documented below.
 
+- [dbt manifest file](https://docs.getdbt.com/reference/artifacts/manifest-json)
```
kyungsoo-datahub: This part(line 5 - line 41) is already in https://github.com/datahub-project/datahub/blob/master/metadata-ingestion/docs/sources/dbt/dbt_pre.md?plain=1#L210-L221

metadata-ingestion/docs/sources/dbt/dbt_pre.md
```diff
@@ -2,6 +2,79 @@
...
+
+```yaml
+source:
+  type: dbt
+  config:
+    manifest_path: target/manifest.json
+    catalog_path: target/catalog.json
+    # Control catalog stats emission (default: YES)
+    entities_enabled:
+      catalog_stats: "NO" # or "YES" (default)
```
kyungsoo-datahub: It's a bit confusing. Maybe `catalog_stats: "YES" # set to "NO" to disable` better.

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
```diff
@@ -2359,6 +2382,36 @@ def create_dbt_platform_mces(
...
+                            type=PartitionTypeClass.FULL_TABLE,
+                        ),
+                    )
+                    yield MetadataChangeProposalWrapper(
+                        entityUrn=node_datahub_urn,
+                        aspect=dataset_profile,
+                    ).as_workunit()
+                    self.report.catalog_stats_extracted += 1
+                else:
+                    self.report.catalog_stats_skipped_no_data += 1
```
kyungsoo-datahub, line 2413: QQ: What's the value of this data?

kyungsoo-datahub (approved):
LGTM with minor comments.

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
alfiyas-datahub: **catalog_stats_extracted**: Number of datasets that had stats (row_count/size_in_bytes) successfully emitted

**catalog_stats_skipped_no_data**: Number of datasets skipped because catalog had no stats (e.g., views, ephemeral models, or warehouses that don't provide stats)

PR #16058, opened 2026-02-03
feat(snowflake): add support for external DMF assertion ingestion

rajatoss:
## Summary
- Add new config option `include_external_dmf_assertions` to ingest user-created Snowflake Data Metric Functions (DMFs) as external assertions in DataHub
- External DMFs are ingested with `AssertionInfo` aspects using CUSTOM type and EXTERNAL source, including column information from `ARGUMENT_NAMES`
- Generate stable, deterministic GUIDs for external DMFs using Snowflake's `REFERENCE_ID` field
- Differentiate between DataHub-created DMFs (prefixed `datahub__*`) which extract GUID from the name, and external DMFs which generate GUIDs from reference IDs

## Test Plan
- **Manual testing**: Verified end-to-end functionality against a live Snowflake environment with both DataHub-created and external DMFs
- **Unit tests** added in `test_snowflake_assertion.py` covering:
  - Pydantic model parsing of `ARGUMENT_NAMES` from JSON strings
  - Query generation with and without external DMF filtering
  - GUID generation determinism and uniqueness based on `REFERENCE_ID`
  - AssertionInfo creation with correct types, sources, and field URNs for single vs multi-column DMFs
  - Mixed processing of both DataHub and external DMFs
- Run `./gradlew :metadata-ingestion:testQuick` to verify tests pass
- Run `./gradlew :metadata-ingestion:lintFix` to verify code quality

## Checklist

## Connector Tests Run for the PR: 
https://github.com/acryldata/connector-tests/actions/runs/21699950462
## Breaking Changes
None - this is an additive feature with a new opt-in config flag (`include_external_dmf_assertions`) that defaults to `false`.


gabe-lyons:
@AdrianMachado want to take a look here?

docs/assertions/snowflake/snowflake_dmfs.md
AdrianMachado:
```suggestion
::: NOTE
```

Sorry, tacking on a mistake I noticed

docs/assertions/snowflake/snowflake_dmfs.md
```diff
@@ -208,6 +208,103 @@ either via CLI or the UI visible as normal assertions.
 
 `datahub ingest -c snowflake.yml`
 
+## Ingesting External (User-Created) DMFs
+
+In addition to DataHub-created DMFs, you can also ingest results from your own custom Snowflake Data Metric Functions. This allows you to monitor data quality assertions that were created outside of DataHub.
+
+### Enabling External DMF Ingestion
+
+To ingest external DMFs, add the `include_external_dmf_assertions` flag to your Snowflake recipe:
```
AdrianMachado: Can you update [this doc](https://docs.datahub.com/docs/generated/ingestion/sources/snowflake) with the flag and also link the text: "assertions created using Datahub assertions CLI" to this page? [text I am referring to](https://docs.datahub.com/docs/generated/ingestion/sources/snowflake#:~:text=created%20using%20Datahub%20assertions%20CLI)

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_assertion.py
```diff
@@ -87,34 +114,122 @@ def _gen_platform_instance_wu(self, urn: str) -> MetadataWorkUnit:
...
-                        result=AssertionResult(
-                            type=(
-                                AssertionResultType.SUCCESS
-                                if status
-                                else AssertionResultType.FAILURE
-                            )
-                        ),
-                    ),
+            if assertee not in discovered_datasets:
+                return []
```
AdrianMachado, line 201: Nice change to early return

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_config.py
```diff
@@ -499,6 +509,15 @@ def validate_include_column_lineage(cls, v, info):
...
         return v
 
+    @field_validator("include_external_dmf_assertions", mode="after")
+    @classmethod
+    def validate_include_external_dmf_assertions(cls, v, info):
+        if not info.data.get("include_assertion_results") and v:
+            raise ValueError(
+                "include_assertion_results must be True for include_external_dmf_assertions to be set."
+            )
+        return v
```
AdrianMachado, line 521: Do we also need to check the `start_time` and `end_time` are set, or are their default values?

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_assertion.py
```diff
@@ -87,34 +114,122 @@ def _gen_platform_instance_wu(self, urn: str) -> MetadataWorkUnit:
...
+        custom_properties: Dict[str, str] = {
+            "snowflake_dmf_name": dmf_name,
+        }
+        # Store all columns in custom properties regardless of count
+        if argument_names:
+            custom_properties["snowflake_dmf_columns"] = ",".join(argument_names)
+
+        assertion_info = AssertionInfoClass(
+            type=AssertionTypeClass.CUSTOM,
+            customAssertion=CustomAssertionInfoClass(
```
AdrianMachado, line 162: Is it possible for you to include an `externalUrl` or is it too difficult to derive with the data you have?

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_assertion.py
```diff
@@ -87,34 +114,122 @@ def _gen_platform_instance_wu(self, urn: str) -> MetadataWorkUnit:
...
+    ) -> MetadataWorkUnit:
+        """Create AssertionInfo for external DMFs."""
+        # Field URN is only set for single-column DMFs. Multi-column DMFs are
+        # treated as table-level assertions with columns stored in custom properties.
+        field_urn: Optional[str] = None
+        if argument_names and len(argument_names) == 1:
+            field_urn = make_schema_field_urn(dataset_urn, argument_names[0])
+
+        custom_properties: Dict[str, str] = {
+            "snowflake_dmf_name": dmf_name,
```
AdrianMachado, line 153: We can probably also toss the reference ID in here too

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_assertion.py
```diff
@@ -87,34 +114,122 @@ def _gen_platform_instance_wu(self, urn: str) -> MetadataWorkUnit:
...
+
+        assertion_info = AssertionInfoClass(
+            type=AssertionTypeClass.CUSTOM,
+            customAssertion=CustomAssertionInfoClass(
+                type="Snowflake Data Metric Function",
+                entity=dataset_urn,
+                field=field_urn,
+            ),
+            source=AssertionSourceClass(
+                type=AssertionSourceTypeClass.EXTERNAL,
```
AdrianMachado, line 168: Thanks for doing this

docs/assertions/snowflake/snowflake_dmfs.md
```diff
@@ -208,6 +208,103 @@ either via CLI or the UI visible as normal assertions.
...
+CREATE DATA METRIC FUNCTION my_null_check_threshold(ARGT TABLE(col VARCHAR))
+RETURNS NUMBER AS
+$$
+  SELECT CASE WHEN COUNT(*) <= 10 THEN 1 ELSE 0 END
+  FROM ARGT WHERE col IS NULL
+$$;
+-- Returns: 1 if ≤10 nulls (pass), 0 if >10 nulls (fail)
+```
+
+### How External DMFs Differ from DataHub-Created DMFs
```
sgomezvillamor, line 307: * I’m having trouble understanding the **operational and functional differences** between *External DMFs* and *DataHub-created DMFs* (not the differences in the aspects we emit). So far, the only explanation I’ve found in the docs is: “This allows you to monitor data quality assertions that were created outside of DataHub.” I think this could use more elaboration to make the distinction clearer.

* Also, the term **“external”** feels a bit ambiguous. From a DataHub-centric perspective, it means “outside of DataHub,” but someone coming from a Snowflake-centric mindset might interpret it differently. That ambiguity makes the concept harder to grasp. Would terminology like **Snowflake-created vs DataHub-created**, or **Snowflake-native vs DataHub-managed DMFs**, be clearer?

sgomezvillamor: I’m not super familiar with assertion modeling, so please take these comments as tentative. It might be worth double-checking with Jay as well.

docs/assertions/snowflake/snowflake_dmfs.md
```diff
@@ -208,6 +208,103 @@ either via CLI or the UI visible as normal assertions.
...
+
+| Aspect             | DataHub-Created DMFs                                    | External DMFs                             |
+| ------------------ | ------------------------------------------------------- | ----------------------------------------- |
+| **Naming**         | Prefixed with `datahub__`                               | Any name                                  |
+| **Definition**     | Created via `datahub assertions compile`                | Created manually by user                  |
+| **Assertion Type** | Based on assertion definition (Freshness, Volume, etc.) | CUSTOM                                    |
+| **Source**         | INFERRED                                                | EXTERNAL                                  |
```
sgomezvillamor: Regarding the `Source`:

https://docs.datahub.com/docs/generated/metamodel/entities/assertion#assertion-source

> The assertionInfo aspect includes an AssertionSource that identifies the origin of the assertion:
> 
> - NATIVE: Defined directly in DataHub (DataHub Cloud feature)
> - EXTERNAL: Ingested from external tools (Great Expectations, dbt, Snowflake, etc.)
> - INFERRED: Generated by ML-based inference systems (DataHub Cloud feature)
>
> External assertions should have a corresponding dataPlatformInstance aspect that identifies the specific platform instance they originated from.

My concerns:
* INFERRED seems to be limited to ML-based inference systems. We may be deviating original purpose.
* We are missing the `DataPlatformInstance` aspect. Which is the one that identifies the origin, according to the docs.

docs/assertions/snowflake/snowflake_dmfs.md
```diff
@@ -208,6 +208,103 @@ either via CLI or the UI visible as normal assertions.
...
+**External DMFs must return `1` for SUCCESS and `0` for FAILURE.**
+
+DataHub interprets the `VALUE` column from Snowflake's `DATA_QUALITY_MONITORING_RESULTS` table as:
+
+- `VALUE = 1` → Assertion **PASSED**
+- `VALUE = 0` → Assertion **FAILED**
+
+This is because DataHub cannot interpret arbitrary return values (e.g., "100 null rows" - is that good or bad?). You must build the pass/fail logic into your DMF.
+
+#### Example: Writing External DMFs Correctly
```
sgomezvillamor, line 270: What if wrong? what's the impact?
How users will find out or identify these cases?

sgomezvillamor (commented):
I mainly checked new feature from the new docs and left some comments on the modelling.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_assertion.py
rajatoss: Snowflake DMFs don't have their own dedicated URL. We already have the urls for the tables on which they are applied, i.e
image

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_assertion.py
rajatoss: added, thanks

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_config.py
rajatoss: There are default values for both `start_time` and `end_time`.

docs/assertions/snowflake/snowflake_dmfs.md
rajatoss: done, added this.

docs/assertions/snowflake/snowflake_dmfs.md
rajatoss: Updated with better documentation.

docs/assertions/snowflake/snowflake_dmfs.md
rajatoss: Updated, yes it should be native for datahub created dmfs. 
In code we already emits `DataPlatformInstance` aspect for every unique assertion URN (both DataHub-created and external DMFs). 

Added in the doc.

docs/assertions/snowflake/snowflake_dmfs.md
rajatoss: updated the doc, added a warning with how they will be interpreted if user doesn't configure it correctly and also how they should configure them correctly.

rajatoss:
cc: @jayacryl if i could get an overall review for this, as it is around dmf, @sgomezvillamor suggested it might be a good idea to double check with you.

docs/assertions/snowflake/snowflake_dmfs.md
sgomezvillamor: Here we are very restrictive on the passing value

> DataHub interprets the `VALUE` column from Snowflake's `DATA_QUALITY_MONITORING_RESULTS` table as:
> 
> - `VALUE = 1` → Assertion **PASSED**
> - `VALUE = 0` → Assertion **FAILED**

However, here we are more relaxed 

> - `VALUE = 0` → **FAILED**
> - `VALUE != 0` (any non-zero value like 5, 100, -1) → **PASSED**

Could you clarify?
Could we follow the 1/0 for passed/failed approach? if so, should we consider the invalid state?
- `VALUE = 1` → Assertion **FAILED**
- `VALUE = 0` → Assertion **PASSED**
- `VALUE != 1 and VALUE != 0` --> Assertion **INVALID**

docs/assertions/snowflake/snowflake_dmfs.md
```diff
@@ -208,6 +208,123 @@ either via CLI or the UI visible as normal assertions.
...
+| **Source**         | NATIVE (defined in DataHub)                        | EXTERNAL                                  |
+| **URN Generation** | Extracted from DMF name (`datahub__<guid>`)        | Generated from Snowflake's `REFERENCE_ID` |
+
+### How External DMFs Appear in DataHub UI
+
+External DMFs appear in DataHub with:
+
+- **Assertion Type**: CUSTOM
+- **Source**: EXTERNAL
+- **Platform Instance**: Snowflake platform instance (if configured)
```
sgomezvillamor, line 323: I miss some test where we validate that the `DataPlatformInstance` aspect is being generated.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_assertion.py
```diff
@@ -87,34 +114,125 @@ def _gen_platform_instance_wu(self, urn: str) -> MetadataWorkUnit:
...
+        DMF-table-column association.
+        """
+        guid_dict: Dict[str, str] = {
+            "platform": "snowflake",
+            "reference_id": result.REFERENCE_ID,
+        }
+        if self.config.platform_instance:
+            guid_dict["instance"] = self.config.platform_instance
+
+        return datahub_guid(guid_dict)
```
sgomezvillamor: what about defining a `SnowflakeExternalDMF(DatahubKey)` class instead?

https://github.com/datahub-project/datahub/blob/master/metadata-ingestion/src/datahub/emitter/mcp_builder.py

docs/assertions/snowflake/snowflake_dmfs.md
rajatoss: I've updated the implementation to be stricter about the return values:

Updated behavior:

`VALUE = 1 `→ SUCCESS (assertion passed)
`VALUE = 0 `→ FAILURE (assertion failed)
`VALUE != 0 and VALUE != 1`→ ERROR (invalid DMF return value)
When an invalid value is encountered, we also log a warning:

We're keeping the 1 = PASS, 0 = FAIL convention to maintain compatibility with DataHub-compiled DMF assertions, which use this convention in metric_operator_sql_generator.py, otherwise they will start showing incorrect results.

docs/assertions/snowflake/snowflake_dmfs.md
rajatoss: added them.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_assertion.py
rajatoss: created a class for this

docs/assertions/snowflake/snowflake_dmfs.md
AdrianMachado: Yeah, I think snowflake-native vs Datahub-managed makes a lot of sense and is more clear @rajatoss

docs/assertions/snowflake/snowflake_dmfs.md
AdrianMachado: cc @jayacryl around the source choice here

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_assertion.py
AdrianMachado: OK, that's sufficient. If they don't have dedicated URLs in snowflake, then its fine.

I am curious as to how users of snowflake actually track DMF failure then @rajatoss

docs/assertions/snowflake/snowflake_dmfs.md
```diff
@@ -208,6 +208,127 @@ either via CLI or the UI visible as normal assertions.
...
+$$
+  SELECT CASE WHEN COUNT(*) <= 10 THEN 1 ELSE 0 END
+  FROM ARGT WHERE col IS NULL
+$$;
+-- Returns: 1 if ≤10 nulls (pass), 0 if >10 nulls (fail)
+```
+
+### How External DMFs Differ from DataHub-Created DMFs
+
+| Aspect             | DataHub-Created DMFs                               | External DMFs                             |
```
jayacryl, line 309:
```suggestion
| Aspect             | DataHub-Managed DMFs                               | Externally Managed DMFs                             |
```

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_assertion.py
```diff
@@ -56,20 +96,19 @@ def __init__(
     def get_assertion_workunits(
         self, discovered_datasets: List[str]
     ) -> Iterable[MetadataWorkUnit]:
+        include_external = self.config.include_external_dmf_assertions
```
jayacryl: to sergio's point do we want to rename the flag to either:  `include_externally_managed_dmfs`, or the inverse: `datahub_managed_only`?

jayacryl (approved):

jayacryl:
Very minor optional comments. Great work Rajat!

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_assertion.py
rajatoss: yes that makes it more clear, renamed to `include_externally_managed_dmfs`

docs/assertions/snowflake/snowflake_dmfs.md
```diff
@@ -208,6 +208,127 @@ either via CLI or the UI visible as normal assertions.
...
+
+DataHub interprets the `VALUE` column from Snowflake's `DATA_QUALITY_MONITORING_RESULTS` table as:
+
+- `VALUE = 1` → Assertion **PASSED**
+- `VALUE = 0` → Assertion **FAILED**
+
+This is because DataHub cannot interpret arbitrary return values (e.g., "100 null rows" - is that good or bad?). You must build the pass/fail logic into your DMF.
+
+::: warning What if my DMF returns other values?
+If your DMF returns values other than 0 or 1, DataHub will mark the assertion result as **ERROR**:
```
tkdrahn, line 258: We tried to use this feature but we get a lot of Errors due to this limitation. Any reason we can’t pull from DATA_QUALOTY_MONITORING_EXPECTATION_STATUS which has a Boolean expectation_violated field?

rajatoss: @tkdrahn EXPECTATION_STATUS only contains results for DMFs that have [expectations](https://docs.snowflake.com/en/user-guide/data-quality-expectations) attached. If a DMF doesn't have an expectation configured, it again becomes a problem.

and  in the doc we do mention that a customer can define the condition when creating data metric function like:
```
CREATE DATA METRIC FUNCTION my_null_check_threshold(ARGT TABLE(col VARCHAR))
RETURNS NUMBER AS
$$
  SELECT CASE WHEN COUNT(*) <= 10 THEN 1 ELSE 0 END
  FROM ARGT WHERE col IS NULL
$$;
-- Returns: 1 if ≤10 nulls (pass), 0 if >10 nulls (fail)
```

which is what `EXPECTATION_STATUS` produces.

tkdrahn: Does it make sense to even ingest DMFs that don’t have an expectation on them? In that scenario it’s not really an Assertion, it seems closer to Profiling metadata

rajatoss: datahub compiled dmf's doesn't have the expectations attached to them, in that case they will get ignored.
We can obviously do that with some conditional logic but for right now can we be okay with the above suggested method for the customers, maybe we can have a ticket to take this up in neartime?

PR #16062, opened 2026-02-03
fix(kafka): Unpack MCE records into MCPs in Kafka sink

max-datahub:
## Summary

Fixes feature parity issue between `datahub-kafka` and `datahub-rest` sinks where MCE-based sources would lose aspects when using the Kafka sink.

**Problem**: Sources that emit MetadataChangeEvent (MCE) records (e.g., Azure AD/EntraID, Okta) were missing certain aspects (like GroupMembership) when using the `datahub-kafka` sink, but worked correctly with `datahub-rest` sink.

**Root Cause**: The Kafka sink was emitting MCE records directly to the `MetadataChangeEvent_v4` topic, which requires the server-side MCE consumer to be enabled (`MCE_CONSUMER_ENABLED=true`). This consumer is disabled by default and the MCE format is deprecated. In contrast, the REST sink automatically unpacks MCEs into individual MetadataChangeProposal (MCP) records—one per aspect—which are processed by the always-enabled MCP consumer.

**Solution**: Modified the Kafka sink to detect MCE records and unpack them into individual MCPs using the existing `mcps_from_mce()` function, matching the REST sink behavior. MCPs are sent to the MCP topic which is always enabled.

## Affected Sources

This fix benefits all sources that emit MCE records:

| Source | Impact |
|--------|--------|
| **azure-ad** | User group memberships now ingested correctly |
| **okta** | User group memberships now ingested correctly |
| **metabase** | All chart aspects now ingested correctly |
| **mode** | All dashboard/chart aspects now ingested correctly |
| **superset** | All dashboard aspects now ingested correctly |
| **redash** | All dashboard aspects now ingested correctly |
| **dbt** | Snapshot-based aspects now ingested correctly |
| **glue** | Snapshot-based aspects now ingested correctly |
| **delta-lake** | Snapshot-based aspects now ingested correctly |
| **file** | MCE-format files now ingested correctly |

## Changes

- `datahub_kafka.py`: Added MCE unpacking logic in `write_record_async()` using pre-existing `mcps_from_mce()`
- `test_kafka_sink.py`: Updated `test_kafka_sink_write()` to verify MCP emission
- `test_kafka_sink.py`: Added `test_kafka_sink_mce_with_multiple_aspects()` to validate multi-aspect MCE handling

## Impact

- ✅ Fixes aspect ingestion for all MCE-based sources via Kafka sink
- ✅ Makes Kafka sink behavior consistent with REST sink
- ✅ Aligns with DataHub's direction (MCP is the standard, MCE is deprecated)
- ✅ Works with all DataHub versions (no server-side changes needed)
- ✅ Backward compatible—MCPs are still emitted as before
- ✅ No configuration changes required

## Checklist

- [x] Updated existing test to verify MCE unpacking
- [x] Added new test with multiple aspects (simulating user + group membership scenario)
- [ ] Manual testing with Azure AD/EntraID ingestion via Kafka sink

### DIFF
--- a/metadata-ingestion/tests/unit/test_kafka_sink.py
+++ b/metadata-ingestion/tests/unit/test_kafka_sink.py
@@ -198,7 +198,13 @@ class KafkaSinkTest(unittest.TestCase):
         )
 
         mce = MetadataChangeEvent(proposedSnapshot=user_snapshot)
-        re = RecordEnvelope(record=mce, metadata={})
+        re: RecordEnvelope[
+            Union[
+                MetadataChangeEvent,
+                MetadataChangeProposal,
+                MetadataChangeProposalWrapper,
+            ]
+        ] = RecordEnvelope(record=mce, metadata={})
         kafka_sink.write_record_async(re, callback)
 
         assert mock_mcp_producer.produce.call_count == 2
@@ -233,7 +239,13 @@ class KafkaSinkTest(unittest.TestCase):
         )
 
         mce = MetadataChangeEvent(proposedSnapshot=user_snapshot)
-        re = RecordEnvelope(record=mce, metadata={})
+        re: RecordEnvelope[
+            Union[
+                MetadataChangeEvent,
+                MetadataChangeProposal,
+                MetadataChangeProposalWrapper,
+            ]
+        ] = RecordEnvelope(record=mce, metadata={})
         kafka_sink.write_record_async(re, callback)
 
         # No MCPs should be produced
@@ -270,6 +282,54 @@ class KafkaSinkTest(unittest.TestCase):
         callback.kafka_callback(None, mock_message)
         mock_w_callback.on_success.assert_called_once()
         assert mock_w_callback.on_success.call_args[0][0] == mock_re
+
+
+def test_aggregating_callback_all_success():
+    """Cover _AggregatingKafkaCallback: all N deliveries succeed."""
+    mock_inner = MagicMock(spec=_KafkaCallback)
+    agg = _AggregatingKafkaCallback(total=3, inner=mock_inner)
+
+    agg.kafka_callback(None, MagicMock())
+    mock_inner.kafka_callback.assert_not_called()
+
+    agg.kafka_callback(None, MagicMock())
+    mock_inner.kafka_callback.assert_not_called()
+
+    agg.kafka_callback(None, MagicMock())
+    mock_inner.kafka_callback.assert_called_once()
+    # The single call should be a success (err=None)
+    assert mock_inner.kafka_callback.call_args[0][0] is None
+
+
+def test_aggregating_callback_first_fails():
+    """Cover _AggregatingKafkaCallback: first delivery fails, rest succeed."""
+    mock_inner = MagicMock(spec=_KafkaCallback)
+    agg = _AggregatingKafkaCallback(total=3, inner=mock_inner)
+
+    mock_error = MagicMock()
+    agg.kafka_callback(mock_error, MagicMock())
+    # Failure fires immediately
+    mock_inner.kafka_callback.assert_called_once()
+    assert mock_inner.kafka_callback.call_args[0][0] is mock_error
+
+    # Subsequent deliveries (success or failure) should not fire again
+    agg.kafka_callback(None, MagicMock())
+    agg.kafka_callback(None, MagicMock())
+    assert mock_inner.kafka_callback.call_count == 1
+
+
+def test_aggregating_callback_last_fails():
+    """Cover _AggregatingKafkaCallback: last delivery fails after others succeed."""
+    mock_inner = MagicMock(spec=_KafkaCallback)
+    agg = _AggregatingKafkaCallback(total=2, inner=mock_inner)
+
+    agg.kafka_callback(None, MagicMock())
+    mock_inner.kafka_callback.assert_not_called()
+
+    mock_error = MagicMock()
+    agg.kafka_callback(mock_error, MagicMock())
+    mock_inner.kafka_callback.assert_called_once()
+    assert mock_inner.kafka_callback.call_args[0][0] is mock_error
 
 
 def _run_concurrent_callbacks(


metadata-ingestion/src/datahub/ingestion/sink/datahub_kafka.py
```diff
@@ -114,10 +115,27 @@ def write_record_async(
         kafka_callback = _KafkaCallback(self.report, record_envelope, write_callback)
         try:
             record = record_envelope.record
```
treff7es, line 154: A single _KafkaCallback instance is reused for all N MCPs unpacked from one MCE. Each Kafka delivery triggers
  the same callback, so:

  - report_record_written() is called N times for 1 logical record, inflating total_records_written
  - on_success() fires N times — upstream callbacks (like LoggingCallback) log the same workunit N times
  - On partial failure, the DeadLetterQueueCallback receives the original MCE envelope (not the failed MCP),
  making retry imprecise — the isinstance(record, MetadataChangeProposalClass) check in sink_callback.py:59 will
   be False

  The REST sink avoids this because MCE unpacking happens inside _emit_batch_wrapper, where callbacks are wired
  at a different level.


Either use an aggregating callback that fires on_success/on_failure only once after all
  deliveries complete, or create per-MCP RecordEnvelope wrappers so each callback references the correct record.

metadata-ingestion/src/datahub/ingestion/sink/datahub_kafka.py
```diff
@@ -114,10 +115,27 @@ def write_record_async(
...
+            # If the record is an MCE, unpack it into multiple MCPs (one per aspect).
+            # This ensures that all aspects (including GroupMembership) are emitted
+            # as separate messages and properly handled by downstream consumers.
+            # This matches the behavior of the REST sink (datahub_rest.py).
+            if isinstance(record, MetadataChangeEvent):
+                logger.debug(
+                    f"Unpacking MCE for {record.proposedSnapshot.urn} into individual MCPs"
+                )
+                mcps = list(mcps_from_mce(record))
+                for mcp in mcps:
```
treff7es, line 179: If an MCE has an empty aspects list, mcps_from_mce() yields nothing.

Add a guard:
```
  if not mcps:
      logger.warning(f"MCE for {record.proposedSnapshot.urn} had zero aspects")
      write_callback.on_success(record_envelope, {"msg": "MCE had zero aspects"})
      return
```

metadata-ingestion/src/datahub/ingestion/sink/datahub_kafka.py
```diff
@@ -114,10 +115,27 @@ def write_record_async(
...
+                    self.emitter.emit(
```
treff7es, line 180: If emit() throws an exception on the 2nd of 3 MCPs (e.g., BufferError when the producer queue is full), the
  1st MCP was already handed to Kafka and the 3rd is never attempted. The except block reports one failure, but
  there's no indication which MCPs succeeded or were skipped.

treff7es: I'm not sure if this is problem.

metadata-ingestion/src/datahub/ingestion/sink/datahub_kafka.py
```diff
@@ -114,10 +115,27 @@ def write_record_async(
...
+
+            # If the record is an MCE, unpack it into multiple MCPs (one per aspect).
+            # This ensures that all aspects (including GroupMembership) are emitted
+            # as separate messages and properly handled by downstream consumers.
+            # This matches the behavior of the REST sink (datahub_rest.py).
+            if isinstance(record, MetadataChangeEvent):
+                logger.debug(
+                    f"Unpacking MCE for {record.proposedSnapshot.urn} into individual MCPs"
+                )
+                mcps = list(mcps_from_mce(record))
```
treff7es, line 166: Why is this needed? Can't the mcps_from_mce(record) be iterated directly?

metadata-ingestion/tests/unit/test_kafka_sink.py
```diff
@@ -68,12 +67,14 @@ def test_kafka_sink_mcp(self, mock_producer, mock_callback):
     @patch("datahub.emitter.kafka_emitter.SerializingProducer", autospec=True)
     @patch("datahub.ingestion.sink.datahub_kafka._KafkaCallback", autospec=True)
     def test_kafka_sink_write(self, mock_k_callback, mock_producer, mock_context):
+        from datahub.emitter.kafka_emitter import MCP_KEY
```
treff7es: import should be at module level

metadata-ingestion/src/datahub/ingestion/sink/datahub_kafka.py
max-datahub: Added `_AggregatingKafkaCallback` (lines 93–126 in `datahub_kafka.py`).

This class wraps the existing `_KafkaCallback` and tracks N delivery results. It fires `on_success` exactly once after all N MCPs are delivered successfully, or `on_failure` exactly once on the first failure. This preserves the 1-record-in = 1-callback-out contract that the REST sink maintains via `BatchPartitionExecutor`.

The class uses a `threading.Lock` because Kafka delivery callbacks are invoked from librdkafka's background thread. Without the lock, concurrent callbacks could both read the same `_remaining` value, causing the inner callback to fire twice or never. While unlikely under CPython's GIL (we verified this empirically), the race is reproducible with forced GIL yields and would be a real problem on free-threaded Python 3.13+ (PEP 703) or alternative runtimes.

**New tests:**

- `test_kafka_sink_mce_with_multiple_aspects` — verifies that `on_success` fires only after all N deliveries complete (not after the first)
- `test_kafka_sink_mce_partial_failure` — verifies that `on_failure` fires exactly once on partial delivery failure
- `test_aggregating_callback_thread_safety` — barrier-based stress test (200 iterations × 20 threads) confirming the lock prevents double/zero fires
- `test_aggregating_callback_thread_safety_with_failure` — same stress test for the failure path

metadata-ingestion/src/datahub/ingestion/sink/datahub_kafka.py
max-datahub: Added the guard at lines 167–174 in `datahub_kafka.py`:

```python
if not mcps:
    logger.warning(
        f"MCE for {record.proposedSnapshot.urn} produced zero MCPs"
    )
    write_callback.on_success(
        record_envelope, {"msg": "MCE had zero aspects"}
    )
    return
```

This also prevents constructing an `_AggregatingKafkaCallback` with `total=0`, which would never fire.

**New test:** `test_kafka_sink_mce_empty_aspects` — verifies that an MCE with zero aspects calls `on_success` immediately with no Kafka produce calls.

metadata-ingestion/src/datahub/ingestion/sink/datahub_kafka.py
max-datahub: The `list()` materialization is now justified by two requirements:

1. The `_AggregatingKafkaCallback` needs to know the total count upfront (`total=len(mcps)`)
2. The empty-aspects guard needs to check `if not mcps`

metadata-ingestion/tests/unit/test_kafka_sink.py
max-datahub: Done

PR #16063, opened 2026-02-03
refactor(dataplex): streamline Dataplex ingestion by removing entity …

NehaGslab:
- Refactor Dataplex ingestion by removing entity processing
- Removed the EntitiesFilterConfig and related entity processing logic.
- Updated EntriesFilterConfig to simplify entry filtering.
- Refactored helper functions and data structures to focus on entries instead of entities.
- Adjusted lineage extraction to work with entries, including renaming relevant classes and methods.
- Enhanced documentation to reflect changes in the Dataplex source functionality.
- Updated unit test cases.
- The lineage map was using only the table name as the key instead of the full dataset path.
This caused tables with identical names but in different datasets to collide and share the same lineage entry.
The fix uses the full dataset_id as the key, ensuring each table has its own unique lineage entry.
- update permissions and API details in Dataplex documentation; remove deprecated sections and clarify entry handling
- enhance integration tests for Dataplex source connector using Entries API; add new golden files and update test structure
- implement lineage handling improvements in tests; replace freeze_time with time_machine and add tests for lineage map key collision


metadata-ingestion/docs/sources/dataplex/dataplex_pre.md
```diff
@@ -94,26 +79,22 @@ Default GCP Role: [roles/datalineage.viewer](https://docs.cloud.google.com/iam/d
...
-   - **Pub/Sub**: topics and subscriptions
-   - **Cloud Storage**: buckets
-   - **Bigtable**: instances, clusters, and tables
-   - **Vertex AI**: models, datasets, and feature stores
-   - **Dataform**: repositories and workflows
-   - **Dataproc Metastore**: services and databases
+**Supported services include:**
 
-2. **Lakes/Zones Entities API** (Optional, default disabled): Extracts entities from Dataplex lakes and zones. Use this if you are using the legacy Data Catalog and need lake/zone information not available in the Entries API. Generally, you should use one or the other unless you have non-overlapping objects across the two APIs. See [API Selection Guide](#api-selection-guide) below for detailed guidance on when to use each API as using both APIs can cause loss of custom properties.
+- **BigQuery**: datasets, tables, models, routines, connections, and linked datasets
```
pdruley, line 88: We list the supported services in two places, here and in lineage. I think we should add a note that only BigQuery and GCS have been properly tested in both places.

NehaGslab: added notes in both places.

pdruley (approved):
looks great but you'll need ingestion team review as well.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_config.py
sgomezvillamor, line 1: We usually remove configs by using `pydantic_removed_field` decorator and report them as breaking changes.

https://github.com/datahub-project/datahub/blob/cb0eaf34d1c01281976d8601033326b6bb33db8f/metadata-ingestion/src/datahub/configuration/validate_field_removal.py#L12-L15

Just raising for your consideration and awareness
Given this connector is INCUBATING and adoption may be very low, we may directly remove those configs.

Anyway, we should drop a line in `docs/how/updating-datahub.md` about the breaking change.

metadata-ingestion/docs/sources/dataplex/dataplex_pre.md
```diff
@@ -130,28 +115,21 @@ This ingestion source maps the following Dataplex Concepts to DataHub Concepts:
 | Dataplex Concept          | DataHub Concept                                                                     | Notes                                                                        |
 | :------------------------ | :---------------------------------------------------------------------------------- | :--------------------------------------------------------------------------- |
 | Entry (Universal Catalog) | [`Dataset`](https://docs.datahub.com/docs/generated/metamodel/entities/dataset)     | From Universal Catalog. Uses source platform URNs (e.g., `bigquery`, `gcs`). |
```
sgomezvillamor, line 117: Do all entries map to `Dataset`?

I understand we have all the entries below and they should map to different entities: `Dataset`, `DataFlow`, `DataJob`, ...

> - **BigQuery**: datasets, tables, models, routines, connections, and linked datasetsExpand commentComment on line R88Resolved
> - **Cloud SQL**: instances
> - **AlloyDB**: instances, databases, schemas, tables, and views
> - **Spanner**: instances, databases, and tables
> - **Pub/Sub**: topics and subscriptions
> - **Cloud Storage**: buckets
> - **Bigtable**: instances, clusters, and tables
> - **Vertex AI**: models, datasets, and feature stores
> - **Dataform**: repositories and workflows
> - **Dataproc Metastore**: services and databases

metadata-ingestion/tests/unit/dataplex/test_dataplex_lineage.py
```diff
@@ -260,40 +233,34 @@ def search_links_side_effect(request):
 
 
 def test_build_lineage_map(lineage_extractor: DataplexLineageExtractor) -> None:
```
sgomezvillamor, line 235: considering:

> The lineage map was using only the table name as the key instead of the full dataset path.
> This caused tables with identical names but in different datasets to collide and share the same lineage entry.
> The fix uses the full dataset_id as the key, ensuring each table has its own unique lineage entry.

do we have a test covering this scenario?

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_lineage.py
```diff
@@ -89,60 +85,53 @@ def __init__(
...
         self.config = config
         self.report = report
         self.lineage_client = lineage_client
-        self.dataplex_client = dataplex_client
         self.redundant_run_skip_handler = redundant_run_skip_handler
 
-        # Map entity IDs to their upstream dependencies to enable efficient lineage lookups
+        # Map entry IDs to their upstream dependencies to enable efficient lineage lookups
         # during workunit generation without re-querying the Lineage API
         self.lineage_map: Dict[str, Set[LineageEdge]] = collections.defaultdict(set)
```
sgomezvillamor: At the very least, we should mention/doc here that the key is expected to be the full dataset id (fully qualified name)

We could even emphasize that by renaming the map: eg `lineage_by_full_dataset_id`

Or we could use proper types. A naive approach could be

```
FullyQualifiedDatasetName = str

...
        # and then use FullyQualifiedDatasetName everywhere instead of str

        self.lineage_map: Dict[FullyQualifiedDatasetName, Set[LineageEdge]] = collections.defaultdict(set)
```

or actual proper type

```
class FullyQualifiedDatasetName:
   project: str
   dataset: str
   name: str
```

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex.py
```diff
@@ -228,7 +212,7 @@ def _emit_final_batch(
             cached_mcps: List of cached MCPs to emit
             total_emitted: Total count of MCPs emitted so far
             project_id: GCP project ID
-            resource_type: Type of resource being emitted ("entities" or "entries")
+            resource_type: Type of resource being emitted
```
sgomezvillamor: given we only process entries now, does this parameter make sense anymore? if no, we can remove it

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_config.py
NehaGslab: Yes, since connector is new and adoption is low I did not follow this path and directly removed fields. Sure will add update file: updating-datahub.md with info.

metadata-ingestion/docs/sources/dataplex/dataplex_pre.md
NehaGslab: We have only tested Bigquery and Cloud Storage. But yes according to code we have only created Datasets.

metadata-ingestion/tests/unit/dataplex/test_dataplex_lineage.py
NehaGslab: Yes tests exists in TestLineageMapKeyCollision class in metadata-ingestion/tests/unit/dataplex/test_dataplex_lineage.py file

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_lineage.py
NehaGslab: Agreed, we can rename  lineage_map to lineage_by_full_dataset_id to be explicit.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex.py
NehaGslab: Done

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_lineage.py
NehaGslab: Done

PR #16067, opened 2026-02-03
feat(ingestion/stored-proc): Add stored procedure lineage with improved pl/sql parsing and function support

acrylJonny:
### Summary
This PR adds comprehensive stored procedure lineage extraction for Oracle, including procedure-to-procedure lineage, improved PL/SQL parsing, and proper support for distinguishing functions from procedures.

### Changes
**Oracle Connector Enhancements:**

- Added complete stored procedure and function lineage extraction from SQL definitions
- Added procedure-to-procedure lineage support via ALL_DEPENDENCIES system view
- Fixed URN format consistency between stored procedures and tables/views
- Added distinct subtypes for functions (Function) vs procedures (Stored Procedure)
- Added proper container hierarchy (Database → Schema → Flow) for stored procedures
- Fixed handling of overloaded procedures with hash-based identifiers
- Fixed two-tier source support (no leading dots in URNs when database name is empty)

**PL/SQL Parsing Improvements:**

- Added Oracle PL/SQL control flow keywords: WHILE, LOOP, END LOOP, END IF, ELSIF, EXCEPTION, WHEN OTHERS, EXIT, CONTINUE, GOTO, RETURN
- Fixed FOR UPDATE/FOR SHARE clause handling to prevent incorrect statement splitting
- Comprehensive test coverage with 13 new unit tests for Oracle-specific lineage scenarios
- Reusable Components:
- Added FUNCTION subtype to JobContainerSubTypes for cross-connector use
- Added procedure registry pattern in sql_common.py for resolving overloaded procedures
- Enhanced stored_procedures/base.py with dependency parsing for procedure-to-procedure lineage
- All changes benefit other SQL connectors (PostgreSQL, MySQL, Snowflake, SQL Server)

**Container Organization:**
Functions and procedures are organized in the same {schema}.stored_procedures container (consistent with PostgreSQL, MySQL, and Snowflake)
Individual entities retain distinct subtypes for filtering in the DataHub UI

### DIFF
--- a/metadata-ingestion/tests/unit/sql_parsing/test_oracle_stored_procedure_lineage.py
+++ b/metadata-ingestion/tests/unit/sql_parsing/test_oracle_stored_procedure_lineage.py
@@ -693,7 +693,7 @@ def test_oracle_parse_dependencies_empty_string():
 
 
 def test_oracle_parse_dependencies_malformed_format():
-    """Test that completely malformed dependency strings raise ValueError."""
+    """Test that malformed dependency strings are silently skipped and return empty list."""
     database_key = DatabaseKey(
         database="ORCL", platform="oracle", instance=None, env="PROD"
     )
@@ -701,23 +701,46 @@ def test_oracle_parse_dependencies_malformed_format():
         database="ORCL", schema="hr", platform="oracle", instance=None, env="PROD"
     )
 
-    # Completely malformed - no parentheses
-    with pytest.raises(ValueError, match="Failed to parse any valid dependencies"):
-        _parse_oracle_procedure_dependencies(
-            "INVALID_FORMAT", database_key, schema_key, None
-        )
+    # Completely malformed - no parentheses (returns empty, not an error)
+    result = _parse_oracle_procedure_dependencies(
+        "INVALID_FORMAT", database_key, schema_key, None
+    )
+    assert result == []
 
-    # Malformed - wrong format
-    with pytest.raises(ValueError, match="Failed to parse any valid dependencies"):
-        _parse_oracle_procedure_dependencies(
-            "SCHEMA.NAME.TOOMANY (PROCEDURE)", database_key, schema_key, None
-        )
+    # Malformed - wrong format (returns empty, not an error)
+    result = _parse_oracle_procedure_dependencies(
+        "SCHEMA.NAME.TOOMANY (PROCEDURE)", database_key, schema_key, None
+    )
+    assert result == []
 
-    # Malformed - invalid type
-    with pytest.raises(ValueError, match="Failed to parse any valid dependencies"):
-        _parse_oracle_procedure_dependencies(
-            "HR.PROC (INVALID_TYPE)", database_key, schema_key, None
-        )
+    # Malformed - invalid type (returns empty, not an error)
+    result = _parse_oracle_procedure_dependencies(
+        "HR.PROC (INVALID_TYPE)", database_key, schema_key, None
+    )
+    assert result == []
+
+
+def test_oracle_parse_dependencies_only_tables_and_views():
+    """Test that dependencies with only tables/views (no procedures) return empty list.
+
+    This is valid behavior - table/view lineage is extracted by the SQL parser
+    from procedure code, not from system metadata dependencies.
+    """
+    database_key = DatabaseKey(
+        database="ORCL", platform="oracle", instance=None, env="PROD"
+    )
+    schema_key = SchemaKey(
+        database="ORCL", schema="hr", platform="oracle", instance=None, env="PROD"
+    )
+
+    # Only tables and views - should return empty list (not an error!)
+    result = _parse_oracle_procedure_dependencies(
+        "HR.EMPLOYEES (TABLE), HR.DEPARTMENTS (VIEW), HR.SALARY_DATA (MATERIALIZED VIEW)",
+        database_key,
+        schema_key,
+        None,
+    )
+    assert result == []
 
 
 def test_oracle_parse_dependencies_partial_malformed():


metadata-ingestion/src/datahub/sql_parsing/split_statements.py
```diff
@@ -22,6 +22,20 @@
...
+    r"END\s+IF",
+    "ELSIF",
+    "EXCEPTION",
+    r"WHEN\s+OTHERS",  # Exception handler (specific to avoid matching CASE/MERGE)
+    "EXIT",  # Exit loop
+    "CONTINUE",  # Continue to next iteration
+    "GOTO",  # Transfer control to label
+    "RETURN",  # Safe - only used in functions/procedures
+    # Note: FOR and DECLARE are intentionally excluded as they conflict with standard SQL
+    # (FOR UPDATE/SHARE, DECLARE variables are handled separately)
```
ligfx: `split_statements.py` is solely for splitting MSSQL queries, since they don't require semicolons between statements and it confuses SQLGlot. What's the intention of these changes?

acrylJonny: You are partially correct. While split_statements.py was originally designed for MSSQL/T-SQL, it's now used generically for stored procedure lineage extraction across multiple platforms. There are unit tests that demonstrate this for a PL/SQL stored proc.
I've updated the code comments to make this clearer.

ligfx: Oh eesh 😬 This should probably be cleaned up at some point but yeah makes sense.

metadata-ingestion/src/datahub/ingestion/source/sql/stored_procedures/base.py
```diff
@@ -72,9 +88,14 @@ def to_urn(self, database_key: DatabaseKey, schema_key: Optional[SchemaKey]) ->
 
 
 def _generate_flow_workunits(
-    database_key: DatabaseKey, schema_key: Optional[SchemaKey]
+    database_key: DatabaseKey, schema_key: Optional[SchemaKey], subtype: str
```
kyungsoo-datahub, line 90: I don't see that subtype is used inside this function. What's this for?

metadata-ingestion/src/datahub/ingestion/source/sql/stored_procedures/base.py
```diff
@@ -239,17 +275,17 @@ def generate_procedure_lineage(
 def generate_procedure_container_workunits(
     database_key: DatabaseKey,
     schema_key: Optional[SchemaKey],
+    subtype: str,
```
kyungsoo-datahub, line 277: Same here. This subtype is not being used _generate_flow_workunits. Do we need to pass this value?

metadata-ingestion/src/datahub/ingestion/source/sql/sql_common.py
```diff
@@ -1569,30 +1570,62 @@ def _process_procedures(
...
-                schema_key=gen_schema_key(
-                    db_name=db_name,
-                    schema=schema,
-                    platform=self.platform,
-                    platform_instance=self.config.platform_instance,
-                    env=self.config.env,
-                ),
+                database_key=database_key,
+                schema_key=schema_key,
+                subtype=JobContainerSubTypes.STORED_PROCEDURE,
```
kyungsoo-datahub: This subtype type is wrong for generate_procedure_container_workunits which is for _generate_flow_workunits. They're creating Dataflow which is using FlowContainerSubTypes. But we're passing JobContainerSubTypes which is for Datajob. Also, it's not being used inside the _generate_flow_workunits.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema_gen.py
```diff
@@ -705,6 +706,7 @@ def _process_procedures(
                         db_name=db_name,
                         schema_name=snowflake_schema.name,
                     ),
+                    subtype=JobContainerSubTypes.STORED_PROCEDURE,
```
kyungsoo-datahub: It's not being used inside generate_procedure_container_workunits. And also, this one is for Datajob but the function requires Dataflow subtype (FlowContainerSubTypes).

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
```diff
@@ -897,6 +927,78 @@ def __getattr__(self, item: str) -> Any:
...
+            ),
+            job_id=job_id,
+            cluster=database_key.env or DEFAULT_ENV,
+            platform_instance=database_key.instance,
+        )
+
+        input_jobs.append(dep_job_urn)
+
+    if not input_jobs and deps:
+        raise ValueError(
```
kyungsoo-datahub: Based on the logic, input_jobs can be empty if all dependencies are from other than "PROCEDURE|FUNCTION|PACKAGE", which can be tables and views. Is it error case for this function?

metadata-ingestion/src/datahub/sql_parsing/split_statements.py
```diff
@@ -253,6 +268,10 @@ def _process_normal(self, most_recent_real_char: str) -> Iterator[str]:
             is_force_new_statement_keyword
             and not self._has_preceding_cte(most_recent_real_char)
             and not self._is_part_of_merge_query()
+            and not (
+                keyword.upper() in ("UPDATE", "SHARE")
```
kyungsoo-datahub, line 272: is_force_new_statement_keyword prevents "SHARE". Do we want to handle SHARE keyword case here?

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
```diff
@@ -33,18 +50,28 @@
...
     BasicSQLAlchemyConfig,
 )
-from datahub.ingestion.source.sql.stored_procedures.base import BaseProcedure
+from datahub.ingestion.source.sql.sql_utils import (
+    gen_database_key,
+    gen_schema_key,
+)
+from datahub.ingestion.source.sql.stored_procedures.base import (
+    BaseProcedure,
+    _get_procedure_flow_name,
```
kyungsoo-datahub: This _get_procedure_flow_name is private function. Would be nice to remove underscore in front of the function name to follow convention.

metadata-ingestion/src/datahub/ingestion/source/sql/stored_procedures/base.py
acrylJonny: Good catch! The parameter was unused and the hardcoded value was MSSQL-specific which was incorrect for Oracle/Snowflake. I've fixed this by:

- Adding a generic `FlowContainerSubTypes.PROCEDURE_CONTAINER` enum
- Updated to use the subtype parameter instead of hardcoding
- Updated all callers to pass the correct generic subtype
- Removed the duplicate `MSSQL_PROCEDURE_CONTAINER` enum entirely and unified everything to use `PROCEDURE_CONTAINER`

metadata-ingestion/src/datahub/ingestion/source/sql/stored_procedures/base.py
acrylJonny: Same as for above fix

metadata-ingestion/src/datahub/ingestion/source/sql/sql_common.py
acrylJonny: Fixed as part of addressing comments the first two items above. All callers now pass `FlowContainerSubTypes.PROCEDURE_CONTAINER` instead of `JobContainerSubTypes.STORED_PROCEDURE`

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema_gen.py
acrylJonny: Twinned with above comment

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
acrylJonny: You're absolutely right - if a procedure depends only on tables/views (not other procedures), `input_jobs` will be empty, and that's completely valid. The SQL parser extracts table/view lineage automatically from the procedure code. The `additional_input_jobs parameter` is specifically for procedure-to-procedure dependencies from system metadata.
I've removed the incorrect `ValueError` and clarified the docstring to explain this only handles procedure-to-procedure dependencies.

metadata-ingestion/src/datahub/sql_parsing/split_statements.py
acrylJonny: The code already handles SHARE. The check on here specifically prevents `SHARE` from starting a new statement when it's part of a `FOR UPDATE`/`FOR SHARE` locking clause. The `_is_for_update_or_share()` method checks for the `FOR` keyword before `UPDATE`/`SHARE`.

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
acrylJonny: I've updated the method name for this

PR #16068, opened 2026-02-03
Fix SchemaField permission issue

yonglingsong:
When user look at a field in a dataset, the name is now a hyperlink which throws an error. The issue is due to permission check on schemaField do not allow when user have the parent dataset permission. But we do have a separate permission for schemaFiled. Extend the permission check on schemaField using parent's permission.


yonglingsong:
@david-leifker When I test the new changes, I found the issue actually not exists any more. When our clients report this issue, when they load column in UI, UI have a hyperlink for schemaField, and they get 401 error. But current new UI do not have any hyperlink which introduce 401 error any more. I also do a quick test on graphQL, load schemaField does not do the authorization check. Look like it only happens when the UI have this wrong hyperlink. So I will close this PR. Thanks again for review.

PR #16079, opened 2026-02-04
test(ingest): add PEP 508 validation for setup.py dependencies

treff7es:
## Summary
- Adds a unit test that validates all `extras_require` entries in setup.py are valid PEP 508 requirement specifiers
- Uses the official `packaging.requirements.Requirement` parser for validation
- Catches issues like #16032 where f-strings were composed incorrectly (`pkg{ver}[extra]` instead of `pkg[extra]{ver}`)

## Motivation
The PyPI release for v1.4.0rc1 failed because an invalid requirement string was introduced:
```
datahub-agent-context==1.4.0rc1[snowflake]  # WRONG - extras after version
datahub-agent-context[snowflake]==1.4.0rc1  # RIGHT - extras before version
```

This test would have caught the issue before merge.

## Test details
- Mocks `setuptools.setup()` to capture `extras_require` without side effects
- Forces a release version (since dev versions have empty `_self_pin`)
- Validates each requirement string with PEP 508 parser
- Runs in ~0.7s as part of `testQuick`

## Test plan
- [x] Test passes on current master
- [x] Test catches the buggy setup.py from PR #16032
- [x] ruff check passes
- [x] mypy passes

### DIFF
--- a/metadata-ingestion/tests/unit/test_setup_py_requirements.py
+++ b/metadata-ingestion/tests/unit/test_setup_py_requirements.py
@@ -1,0 +1,88 @@
+"""
+Validates that all dependency strings in setup.py are valid PEP 508 requirement specifiers.
+
+Uses the official `packaging.requirements.Requirement` parser to validate all
+extras_require entries, catching issues like:
+- `package==1.0[extras]` (invalid - extras must come before version)
+- Malformed version specifiers
+- Invalid package names
+"""
+
+import subprocess
+import sys
+from pathlib import Path
+
+import pytest
+from packaging.requirements import InvalidRequirement, Requirement
+
+
+def test_all_extras_require_are_valid_pep508() -> None:
+    """
+    Validate all extras_require entries using the official PEP 508 parser.
+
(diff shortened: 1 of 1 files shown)


metadata-ingestion/tests/unit/test_setup_py_requirements.py
```diff
@@ -0,0 +1,88 @@
...
+
+exec(compile(setup_content, 'setup.py', 'exec'))
+
+# Output all requirements as JSON lines
+import json
+for extra, deps in captured.get('extras_require', {}).items():
+    for dep in deps:
+        print(json.dumps({'extra': extra, 'dep': dep}))
+"""
+    result = subprocess.run(
```
sgomezvillamor, line 57: why do we need to handle this in a subprocess?
is for the mocking? can't we just restore the mocking once the test is done?

treff7es: The subprocess approach guarantees complete isolation with no risk of test pollution.

Issue #16026, opened 2026-01-30
Multi-project LookML ingestion doesn't detect local dependencies


marcelo-astrafy:
**Context**
I tried ingesting LookML with remote dependency.  My `manifest.lkml` includes:
```lkml
remote_dependency: astrafy_kit (
  url: "https://gitlab.com/a237/looker/looker-toolkit.git"
  ref: "master"
)
```
In my DataHub ingestion recipe, I have this:
```yaml
pipeline_name: "${PIPELINE_NAME}"

source:
  type: lookml
  config:
    env: "${ENV}"
    parse_table_names_from_sql: true
    stateful_ingestion:
      enabled: true
    base_folder: "./${LOOKML_REPO_NAME}"
    project_name: "${LOOKML_PROJECT_NAME}"
    connection_to_platform_map:
      astrafy-lk-prd:
        platform: bigquery
        default_db: bqdts_web_analytics
      astrafy-lk-dev:
        platform: bigquery
        default_db: bqdts_web_analytics
    emit_reachable_views_only: true
    process_refinements: true
    allow_partial_lineage_results: false
    lookml_constants:
      schema_project: "${LOOKML_SCHEMA_PROJECT}"
      table_prefix: "${LOOKML_TABLE_PREFIX}"
    project_dependencies:
      astrafy_kit: "./looker-toolkit"

transformers:
  - type: pattern_add_dataset_dataproduct
    config:
      is_container: true
      dataset_to_data_product_urns_pattern:
        rules:
          '.*': 'urn:li:dataProduct:${DATA_PRODUCT_URN}'
  - type: pattern_add_dataset_domain
    config:
      domain_pattern:
        rules:
          '.*': 
            - 'urn:li:domain:${DOMAIN_URN}'

sink:
  type: "datahub-rest"
  config:
    server: "http://datahub-datahub-gms.datahub.svc.cluster.local:8080/"
```
The looker-toolkit is cloned locally and found at `./looker-toolkit`

**Issue**
When I run the ingestion, I get an error related to `deploy_key_filled_from_deploy_key_file in datahub/configuration/git.py`, as if DataHub is treating my local path as a Git repo config and looking for a deploy key.
```
[2026-01-29T17:13:52.879726Z] INFO - [base]   File "/app/lib/python3.11/site-packages/datahub/configuration/common.py", line 159, in _track_nesting_context
[2026-01-29T17:13:52.879929Z] INFO - [base]     instance = handler(data)
[...]

PR #16080, opened 2026-02-04
fix(lookml): project_dependencies union deserialization and config validation

sgomezvillamor:
Fixes LookML ingestion when `project_dependencies` uses a local path.

The default `union_mode` is `smart`, which means Pydantic may attempt to validate both types in the union. In this case, a string path could be incorrectly treated as `GitInfo`, leading to an `AttributeError`. To prevent this, the `GitInfo` validator now safely handles non-`dict` inputs.

Additionally, this PR sets the union mode to `left_to_right`, so string paths are validated as `DirectoryPath` first, and only treated as `GitInfo` otherwise. As documented, local paths must exist at validation time.

Unit tests have been added for both the config and the validators.

It’s unclear when this regression was introduced, or whether this scenario ever worked reliably. The new tests should prevent similar issues in the future, both for this case and other complex LookML config validations.

This should address https://github.com/datahub-project/datahub/issues/16026

### DIFF
--- a/metadata-ingestion/tests/unit/lookml/test_lookml_config.py
+++ b/metadata-ingestion/tests/unit/lookml/test_lookml_config.py
@@ -8,7 +8,6 @@ from datahub.configuration.git import GitInfo
 from datahub.ingestion.source.looker.looker_config import LookerConnectionDefinition
 from datahub.ingestion.source.looker.lookml_config import LookMLSourceConfig
 
-
 # ---- GitInfo validator tests ----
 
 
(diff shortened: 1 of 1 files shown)


metadata-ingestion/src/datahub/configuration/git.py
```diff
@@ -112,6 +112,8 @@ class GitInfo(GitReference):
     def deploy_key_filled_from_deploy_key_file(
         cls, values: Dict[str, Any]
     ) -> Dict[str, Any]:
+        if not isinstance(values, dict):
+            return values
```
treff7es, line 114: Does this pass on mypy? The return type should be `Any` and not `Dict[str, Any]` and the same true for the input values as well

sgomezvillamor: @cursoragent can you address this? values should be Any and return type should be Any

sgomezvillamor: @cursoragent fix lint issues

```
  |
1 | import pathlib
2 | from copy import deepcopy
3 | from typing import Any, Dict, Optional, Union
  |                         ^^^^ F401
4 |
5 | from pydantic import (
  |
  = help: Remove unused import: `typing.Dict`
```

PR #16085, opened 2026-02-04
fix(documents): prevent duplicate document creation on page refresh

annadoesdesign:
Fixes a race condition where the document tree loading state would become false before tree initialization completed, causing ContextDocumentsPage to incorrectly detect zero documents and auto-create duplicates.

The issue occurred because:
- loadingRoot tracked only the GraphQL query state
- Tree initialization (sorting, checking for children) was async
- ContextDocumentsPage would check rootNodes.length === 0 between query completion and tree initialization, triggering unwanted document creation

Solution:
- Added isInitializing state to track full initialization lifecycle
- Updated loading flag to include both query and initialization states
- Ensures ContextDocumentsPage waits for complete tree population

Related files:
- datahub-web-react/src/app/document/hooks/useLoadDocumentTree.ts

### DIFF
--- a/datahub-web-react/src/app/document/hooks/useLoadDocumentTree.ts
+++ b/datahub-web-react/src/app/document/hooks/useLoadDocumentTree.ts
@@ -1,4 +1,4 @@
-import { useCallback, useEffect } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 
 import { DocumentTreeNode, useDocumentTree } from '@app/document/DocumentTreeContext';
 import { useSearchDocuments } from '@app/document/hooks/useSearchDocuments';
@@ -19,6 +19,9 @@ import { Document, DocumentSourceType } from '@types';
 export function useLoadDocumentTree() {
     const { initializeTree, setNodeChildren, getRootNodes } = useDocumentTree();
     const [searchDocumentsQuery] = useSearchDocumentsLazyQuery();
+
+    // Track whether tree initialization is in progress (prevents race condition)
+    const [isInitializing, setIsInitializing] = useState(true);
 
     // Load root documents
     const { documents: rootDocuments, loading: loadingRoot } = useSearchDocuments({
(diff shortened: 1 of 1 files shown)


datahub-web-react/src/app/document/hooks/useLoadDocumentTree.ts
```diff
@@ -136,16 +139,21 @@ export function useLoadDocumentTree() {
...
+            setIsInitializing(false);
         }
         // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [loadingRoot, rootDocuments]);
 
     return {
         loadChildren,
         checkForChildren,
-        loading: loadingRoot,
+        loading: loadingRoot || isInitializing,
```
jjoyce0510, line 157: Not sure if this is really necessary here. The above changes are not as clean as they could be, but they will work.

jjoyce0510: Like why not just keep loading: loadingRoot. I don't see why we need two variables here instead of one. Maybe AI can explain..

annadoesdesign: @jjoyce0510  Good question! The two variables track fundamentally different things:

**`loadingRoot`** - Tracks when the GraphQL query returns
**`isInitializing`** - Tracks when the async post-processing completes

The race condition occurs in this gap:

```
Time    loadingRoot    initializeAsync()    rootNodes
─────────────────────────────────────────────────────────
T1      true           not started          []          
T2      false          STARTS               []    ← Gap begins
T3      false          await checkChildren  []    ← Still in gap
T4      false          initializeTree()     []    ← Still in gap  
T5      false          DONE                 [nodes...]  ← Gap ends
```

Between T2-T4, `loadingRoot` is `false` but the tree isn't populated yet. `ContextDocumentsPage` sees `loading: false` + `rootNodes.length === 0` and incorrectly concludes "no documents exist," triggering auto-creation of a duplicate.

The async work creating this gap is the `await checkForChildren()` call (another GraphQL query) that runs *after* the root documents query completes. Only after that returns can we call `initializeTree()`.

We can't just use `loadingRoot` because it becomes `false` before all the async initialization work finishes. The `isInitializing` state bridges that gap so `loading` remains `true` until the tree is actually ready to use.

jjoyce0510:
Unrelated flaky unit test fixed in another PR https://github.com/datahub-project/datahub/pull/16092. Merging thru.

PR #16089, opened 2026-02-04
feat(report): extend daily report to include additional stats

shirshanka:

### DIFF
--- a/metadata-service/factories/src/main/java/com/linkedin/gms/factory/telemetry/DailyReport.java
+++ b/metadata-service/factories/src/main/java/com/linkedin/gms/factory/telemetry/DailyReport.java
@@ -2,9 +2,12 @@ package com.linkedin.gms.factory.telemetry;
 
 import static com.linkedin.gms.factory.telemetry.TelemetryUtils.*;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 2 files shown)


metadata-service/factories/src/main/java/com/linkedin/gms/factory/telemetry/DailyReport.java
```diff
@@ -225,4 +271,161 @@ public void ping(String eventName, JSONObject properties) {
...
+  private void addMetadataAnalytics(AnalyticsService analyticsService, JSONObject report) {
+    // Collect entity counts by type
+    Map<String, Integer> entityCounts = collectEntityCounts(analyticsService);
+
+    // Calculate total assets - use long to avoid overflow
+    long totalAssets = entityCounts.values().stream().mapToLong(Integer::longValue).sum();
+    report.put("total_assets", anonymizeToBucket((int) Math.min(totalAssets, Integer.MAX_VALUE)));
+
+    // Add entity type breakdown
+    JSONObject entityTypeBreakdown = new JSONObject();
```
jjoyce0510: I'm not sure if mixpanel will like tis nested map format in case we want to filter etc.

Can we actually have these properties at the top level of the report?

jjoyce0510 (approved):

PR #16092, opened 2026-02-05
fix(): Adding fixes for service account tests

jjoyce0510:
- Reduce flakiness in unit tests
- Simplify cypress tests


smoke-test/tests/cypress/cypress/e2e/settingsV2/v2_manage_service_accounts.js
```diff
@@ -149,8 +148,7 @@ describe("manage service accounts", () => {
     deleteServiceAccount();
   });
 
-  // Test with inviteUsersEnabled = true (new UI pattern - button in header)
-  it("create, generate token, and delete service account (new UI with invite users)", () => {
+  it("create, generate token, and delete service account", () => {
```
asikowitz, line 32: Does cypress handle tests with the same name properly?

jjoyce0510: oop good catch. let me see

PR #16100, opened 2026-02-05
feat(snowflake): Add metadata pattern pushdown and table type filtering for improved performance

rajatoss:
## Summary

- Adds `push_down_metadata_patterns` config option to push database, schema, table, and view patterns directly to Snowflake SQL queries using RLIKE operator for significant performance improvements on large environments
- Adds `exclude_external_tables` and `exclude_dynamic_tables` config options to skip ingestion of external and dynamic tables at the query level
- Adds comprehensive documentation for the metadata pattern pushdown feature including Snowflake RLIKE syntax differences vs Python regex

## Checklist

## New Configuration Options

| Option | Type | Default | Description |
|--------|------|---------|-------------|
| `push_down_metadata_patterns` | bool | `false` | Pushes database_pattern, schema_pattern, table_pattern, and view_pattern filtering to Snowflake queries using RLIKE |
| `exclude_external_tables` | bool | `false` | Excludes external tables from ingestion to speed up ingestion |
| `exclude_dynamic_tables` | bool | `false` | Excludes dynamic tables from ingestion to speed up ingestion |

## Test Plan

- [x] Unit tests added for all new SQL query generation methods (500+ lines of new tests)
- [x] Tests cover pattern filtering for databases, schemas, tables, and views
- [x] Tests cover exclude flags for external and dynamic tables (including combined flags)
- [x] Tests verify COALESCE NULL handling for is_dynamic column
- [x] All existing tests pass
- [x] Linting passes

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/.github/scripts/utils/clients/gh_client.py
+++ b/.github/scripts/utils/clients/gh_client.py
@@ -29,8 +29,8 @@ def handle_rate_limit(
             for attempt in range(1, max_retries + 1):
                 last_resp = func(*args, **kwargs)
 
-                remaining = last_resp.headers.get("x-ratelimit-remaining")
-                if last_resp.status_code in [403, 429] and remaining == "0":
+                is_rate_limited = last_resp.headers.get("retry-after") or last_resp.headers.get("x-ratelimit-remaining")
+                if last_resp.status_code in [403, 429] and is_rate_limited:
                     retry_after = last_resp.headers.get("retry-after")
                     reset_timestamp = last_resp.headers.get("x-ratelimit-reset")
 
--- a/.github/scripts/utils/workflow_metrics.py
+++ b/.github/scripts/utils/workflow_metrics.py
@@ -54,7 +54,7 @@ class JobMetrics:
     completed_at: str | None
     duration_seconds: int | None
     runner_name: str | None
-    rerun_count: int
+    run_attempt: int
     steps: list[StepMetrics]
 
     @staticmethod
@@ -87,7 +87,7 @@ class JobMetrics:
             completed_at=data.get("completed_at"),
             duration_seconds=duration_seconds(started, completed),
             runner_name=data.get("runner_name"),
-            rerun_count=data.get("run_attempt", 1),
+            run_attempt=data.get("run_attempt", 1),
             steps=[StepMetrics.from_api(s) for s in data.get("steps", [])],
         )
 
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -4,9 +4,14 @@ on:
     branches:
       - master
       - releases/**
+      - hotfixes/**
     paths-ignore:
       - "docs/**"
       - "**.md"
+      - ".github/**"
+      - "!.github/workflows/build-and-test.yml"
+      - "!.github/actions/ensure-codegen-updated"
+      - "!.github/scripts/send_failed_tests_to_posthog.py"
   pull_request:
     branches:
       - "**"
--- a/.github/workflows/dagster-plugin.yml
+++ b/.github/workflows/dagster-plugin.yml
@@ -13,7 +13,7 @@ on:
     branches:
       - master
     paths:
-      - ".github/**"
+      - ".github/workflows/dagster-plugin.yml"
       - "metadata-ingestion-modules/dagster-plugin/**"
       - "metadata-ingestion/**"
       - "metadata-models/**"
--- a/.github/workflows/docker-unified.yml
+++ b/.github/workflows/docker-unified.yml
@@ -15,6 +15,16 @@ on:
     types: [opened, synchronize, reopened]
     branches:
       - "**"
+    paths-ignore:
+      - ".github/**"
+      - "!.github/workflows/docker-unified.yml"
+      - "!.github/scripts/send_failed_tests_to_posthog.py"
+      - "!.github/scripts/docker_helpers.sh"
+      - "!.github/actions/ci-optimization"
+      - "!.github/scripts/check_python_package.py"
+      - "!.github/scripts/parse_failed_cypress_tests.py"
+      - "!.github/scripts/parse_failed_pytest_tests.py"
+      - "!.github/scripts/docker_logs.sh"
   release:
     types: [published]
 
--- a/.github/workflows/gx-plugin.yml
+++ b/.github/workflows/gx-plugin.yml
@@ -13,7 +13,7 @@ on:
     branches:
       - master
     paths:
-      - ".github/**"
+      - ".github/workflows/gx-plugin.yml"
       - "metadata-ingestion-modules/gx-plugin/**"
       - "metadata-ingestion/**"
       - "metadata-models/**"
--- a/.github/workflows/post-workflow-actions.yml
+++ b/.github/workflows/post-workflow-actions.yml
@@ -1,5 +1,5 @@
 name: Post Workflow Actions
-run-name: Post Workflow Actions for ${{ github.event.workflow_run.name }}
+run-name: Post Workflow Actions for ${{ github.event_name == 'workflow_dispatch' && format('run {0}', inputs.run_id) || github.event.workflow_run.name }}
 
 # Triggered when a monitored workflow completes.
 # NOTE: workflow_run.workflows does not support wildcards —
@@ -15,17 +15,42 @@ on:
     types:
       - completed
 
+  # Allows manual triggering for ad-hoc testing against any existing workflow run.
+  # publish defaults to false so artifacts are not forwarded to downstream by accident.
+  workflow_dispatch:
+    inputs:
+      run_id:
+        description: "Workflow run ID to collect metrics for"
+        required: true
+        type: string
+      attempt:
+        description: "Run attempt number"
+        required: false
+        type: string
+        default: "1"
+      publish:
+        description: "Publish metrics artifacts downstream"
+        required: false
+        type: boolean
+        default: false
+
 permissions:
   actions: read
   contents: read
 
 jobs:
   collect-metrics:
+    # For workflow_run: only collect metrics for PRs and pushes to the default/release/hotfix
+    # branches — skip feature branches, dependabot bumps, etc.
+    # workflow_dispatch is always allowed for ad-hoc testing.
+    if: >-
+      github.event_name == 'workflow_dispatch' ||
+      github.event.workflow_run.event == 'pull_request' ||
+      (github.event.workflow_run.event == 'push' &&
+        (github.event.workflow_run.head_branch == 'master' ||
+         startsWith(github.event.workflow_run.head_branch, 'release/') ||
+         startsWith(github.event.workflow_run.head_branch, 'hotfix/')))
     runs-on: ubuntu-latest
-
-    env:
-      ARTIFACT_NAME: workflow-metrics-${{ github.event.workflow_run.id }}-attempt-${{ github.event.workflow_run.run_attempt }}
-
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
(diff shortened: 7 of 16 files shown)


sgomezvillamor:
# Redundancy Concern

I’m a bit concerned about potential overlap between the new `push_down_metadata_patterns` flag and the existing `push_down_database_pattern_access_history`.

Please correct me if I’m wrong:

* `push_down_database_pattern_access_history` applies to queries against the `access_history` table (i.e., lineage/usage).
* `push_down_metadata_patterns` applies to queries against `information_schema` (i.e., metadata extraction).

Is that accurate?

If so:

* The user-facing documentation should clearly explain this distinction.
* From a maintenance and regression-resilience perspective 😄 — how can we make it explicit in the code that filters generated from `push_down_metadata_patterns` are only used for metadata queries? I’m concerned about accidentally reusing those filters in the wrong context and introducing subtle regressions.

---

# Risk of Over-Filtering

When `push_down_database_pattern_access_history` was introduced, there were concerns about overly aggressive filtering leading to missed lineage — for example, temporary tables created in databases outside the configured `database_name` patterns. That led to introducing `additional_database_names_allowlist` as a workaround.

If `push_down_metadata_patterns` is strictly limited to metadata extraction, this specific issue shouldn’t apply. However, we do have some complex handling around `discovered_datasets` and `discovered_tables`, and I’m not entirely sure whether additional filtering at the SQL level could introduce unintended side effects.

Could you double-check that?

---

# Practical Performance Impact

`push_down_database_pattern_access_history` made clear sense because `access_history` is typically a massive table, and processing everything in Python is very expensive.

Pushing filters down to SQL instead of applying them later in Python is conceptually sound for metadata extraction — but I’m wondering how meaningful the improvement is in practice.

Specifically:

* Can we identify from logs whether metadata extraction stage is a performance bottleneck in Snowflake? we should find metrics about metadata extraction stage
* In those cases where elapsed time is high for metadata extraction, are there in the recipe selective database or table patterns that would significantly reduce the scanned data?
* Or are most real-world cases ingesting all databases/tables anyway (in which case pushdown wouldn’t help much)?

In short, I’d like to understand whether we have concrete performance pain that this change meaningfully addresses.

Also, is fine-grained table-level pushdown worth the added complexity?

---

# `exclude_external_tables` Design

According to Snowflake docs, `TABLE_TYPE` can be:

> 1. `BASE TABLE`
> 2. `TEMPORARY TABLE`
> 3. `EXTERNAL TABLE`
> 4. `EVENT TABLE`
> 5. `VIEW`
> 6. `MATERIALIZED VIEW`

From what I can see, we currently consider only `BASE TABLE` and `EXTERNAL TABLE`, and this PR makes external tables optional.

To make this more scalable and future-proof, I’d suggest a more general approach:

* Introduce something like `table_types` or `metadata_table_types` as a `set[str]`
* Initially support `BASE TABLE` and `EXTERNAL TABLE` as only valid values
* Leave room to extend in the future if we decide to include other types

That would make the configuration more flexible and avoid adding more boolean flags later.

---

Overall, I’m supporting this feature enhancement — I just want to ensure we’re clear about scope, performance justification, and long-term maintainability.

rajatoss:
Hi @sgomezvillamor ,
I went through the suggestions:

Regarding 1:
Yes, your understanding is correct:

`push_down_database_pattern_access_history` applies to `account_usage.access_history queries (lineage/usage extraction)`

`push_down_metadata_patterns` applies to `information_schema queries (metadata extraction)`

I have added more clear docs, to avoid any confusion for the end user as well as for the developer.

Regarding 2:
I checked the `discovered_datasets / discovered_tables` flow:

`discovered_datasets` is populated  from `schema_extractor.databases`, which is the post-filtered result. Today (without pushdown), Python filtering already removes non-matching tables before they enter `discovered_datasets`. With pushdown, SQL filtering removes them earlier, but the final set is identical, if you are following the `RLIKE` syntax when the flag for push down is enabled and python `re` syntax when it is disabled.


Regarding 3:
There is a comment by Piotr that this was already analysed and concluded, on the original Ticket ING-1102

Regarding 4:

I have introduced a set for the table types for future extensibility, while for `exclude_dynamic_tables` I have kept the flag as its not a Snowflake table type.

skrydal:
As for the pushdown flags, I wonder whether we shouldn't logically group them under the pattern itself. We could create a new class - `AllowDenyPatternWithPushDown` (the name should be adjusted...), which would hold the flag for that particular pattern. It would be clearer. Now we have free-floating flags which change behavior of other pieces of configuration. This will make it also clear which patterns are supported and by what.

Existing flag for the queries history could be kept, however I can imagine marking it deprecated and follow general logic. Do we need that fine grained configuration (pushdown in one place but not the other) anyway? We have to consider maintainability of the configs, which is raising exponentially with the number of flags - this is very similar to the point made about external tables flag - **very good catch**!

As for the **performance impact** - the request for this feature has been actually preceded by a careful analysis of timings for certain usage patterns - those changes will be extremely useful, allowing to reduce time in some cases by factors of 10-30x.

sgomezvillamor:
@rajatoss 

> Regarding 1: Yes, your understanding is correct:
> 
> `push_down_database_pattern_access_history` applies to `account_usage.access_history queries (lineage/usage extraction)`
> 
> `push_down_metadata_patterns` applies to `information_schema queries (metadata extraction)`
> 
> I have added more clear docs, to avoid any confusion for the end user as well as for the developer.

👌 

> Regarding 2: I checked the `discovered_datasets / discovered_tables` flow:
> 
> `discovered_datasets` is populated from `schema_extractor.databases`, which is the post-filtered result. Today (without pushdown), Python filtering already removes non-matching tables before they enter `discovered_datasets`. With pushdown, SQL filtering removes them earlier, but the final set is identical, if you are following the `RLIKE` syntax when the flag for push down is enabled and python `re` syntax when it is disabled.

👌 
We may add some docstring for `discovered_datasets` and `discovered_tables` mentioning that these list must contain already-filtered assets (regardless of them being filtered in SQL engine or later in python)

> Regarding 3: There is a comment by Piotr that this was already analysed and concluded, on the original Ticket ING-1102

👌 

So we can monitor this enhancement, I would like to have a metric or log trace with the number of records returned by the metadata extraction sql query.

An increase in that metric when enabling the pushdown filter + a reduction on the elapsed time for the metadata extraction stage would confirm the enhancement.

> Regarding 4:
> 
> I have introduced a set for the table types for future extensibility, while for `exclude_dynamic_tables` I have kept the flag as its not a Snowflake table type.

👌 

@skrydal @rajatoss 

> As for the pushdown flags, I wonder whether we shouldn't logically group them under the pattern itself. We could create a new class - `AllowDenyPatternWithPushDown` (the name should be adjusted...), which would hold the flag for that particular pattern. 

_hold the flag for that particular pattern_ The problem is some patterns can be pushed down to some query while not to other query. So we cannot solve with one flag per pattern.

> Existing flag for the queries history could be kept, however I can imagine marking it deprecated and follow general logic. Do we need that fine grained configuration (pushdown in one place but not the other) anyway? 

Basically yes, we need it.

> We have to consider maintainability of the configs, which is raising exponentially with the number of flags - this is very similar to the point made about external tables flag - very good catch!

I fully agree. The more configs, the more complexity for the both support and maintenance of the code.

However, I find very difficult to generalize this since optimizations usually are coupled to the implementation itself  - so no generic.

rajatoss:
> @rajatoss
> 
> > Regarding 1: Yes, your understanding is correct:
> > `push_down_database_pattern_access_history` applies to `account_usage.access_history queries (lineage/usage extraction)`
> > `push_down_metadata_patterns` applies to `information_schema queries (metadata extraction)`
> > I have added more clear docs, to avoid any confusion for the end user as well as for the developer.
> 
> 👌
> 
> > Regarding 2: I checked the `discovered_datasets / discovered_tables` flow:
> > `discovered_datasets` is populated from `schema_extractor.databases`, which is the post-filtered result. Today (without pushdown), Python filtering already removes non-matching tables before they enter `discovered_datasets`. With pushdown, SQL filtering removes them earlier, but the final set is identical, if you are following the `RLIKE` syntax when the flag for push down is enabled and python `re` syntax when it is disabled.
> 
> 👌 We may add some docstring for `discovered_datasets` and `discovered_tables` mentioning that these list must contain already-filtered assets (regardless of them being filtered in SQL engine or later in python)
> 
> > Regarding 3: There is a comment by Piotr that this was already analysed and concluded, on the original Ticket ING-1102
> 
> 👌
> 
> So we can monitor this enhancement, I would like to have a metric or log trace with the number of records returned by the metadata extraction sql query.
> 
> An increase in that metric when enabling the pushdown filter + a reduction on the elapsed time for the metadata extraction stage would confirm the enhancement.
> 
> > Regarding 4:
> > I have introduced a set for the table types for future extensibility, while for `exclude_dynamic_tables` I have kept the flag as its not a Snowflake table type.
> 
> 👌
> 
> @skrydal @rajatoss
> 
> > As for the pushdown flags, I wonder whether we shouldn't logically group them under the pattern itself. We could create a new class - `AllowDenyPatternWithPushDown` (the name should be adjusted...), which would hold the flag for that particular pattern.
> 
> _hold the flag for that particular pattern_ The problem is some patterns can be pushed down to some query while not to other query. So we cannot solve with one flag per pattern.
> 
> > Existing flag for the queries history could be kept, however I can imagine marking it deprecated and follow general logic. Do we need that fine grained configuration (pushdown in one place but not the other) anyway?
> 
> Basically yes, we need it.
> 
> > We have to consider maintainability of the configs, which is raising exponentially with the number of flags - this is very similar to the point made about external tables flag - very good catch!
> 
> I fully agree. The more configs, the more complexity for the both support and maintenance of the code.
> 
> However, I find very difficult to generalize this since optimizations usually are coupled to the implementation itself - so no generic.

Thanks for the suggestion @skrydal. I did consider grouping pushdown into the pattern class, but the pushdown eligibility depends on the query path, not just the pattern as mentioned by @sgomezvillamor . For example, view_pattern is pushed down only when fetch_views_from_information_schema=True, with SHOW VIEWS (default), SQL filtering isn't possible so Python filtering is used regardless. I have keept the flag count minimal , one flag (push_down_metadata_patterns) covers all 4 information_schema patterns. If more pushdown use cases emerge, we should revisit the abstraction then.

rajatoss:
Please let me know any further feedback on this PR.

There is also a PR which add few tests in connector-tests repo to validate the results of pushdown pattern:
https://github.com/acryldata/connector-tests/pull/672

rajatoss:
Reminder for this @sgomezvillamor.

metadata-ingestion/docs/sources/snowflake/snowflake_pre.md
```diff
@@ -218,6 +218,14 @@ source:
         - ".*_BACKUP$" # Exclude tables ending with _BACKUP
 ```
 
+#### View Pattern Limitation
+
+By default, Snowflake views are fetched using `SHOW VIEWS`, which does **not** support SQL-level filtering. When `push_down_metadata_patterns: true`, the `view_pattern` is pushed down **only** if `fetch_views_from_information_schema: true` is also set. Otherwise, view filtering falls back to Python's `re.match()`, even with pushdown enabled.
+
+This means if you write patterns in Snowflake RLIKE syntax (e.g., `PROD.*` for prefix matching), they will still work correctly with Python filtering since `PROD.*` is valid in both. However, patterns that rely on RLIKE's full-string matching semantics (e.g., exact match `PROD_DB` without `.*`) will behave differently — Python `re.match()` treats it as a prefix match.
+
+**Recommendation**: If you enable `push_down_metadata_patterns`, also enable `fetch_views_from_information_schema: true` to ensure consistent behavior for view patterns.
```
sgomezvillamor, line 249: may this recommendation become a pydantic validation?
failing the validation may be excessive, but we could raise a warning

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_query.py
```diff
@@ -31,6 +31,117 @@ def create_deny_regex_sql_filter(
...
+                         "CONCAT(table_catalog, '.', table_schema, '.', table_name)")
+
+        Returns:
+            SQL WHERE clause string, or empty string if no filtering needed
+        """
+        if not pattern:
+            return ""
+
+        if pattern.allow == []:
+            return "FALSE"
```
sgomezvillamor, line 75: we may raise a warning here
otherwise, the ingestion would complete with zero entities and no explanation, right?

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema_gen.py
```diff
@@ -260,10 +260,17 @@ def get_workunits_internal(self) -> Iterable[MetadataWorkUnit]:
...
+            if self.config.push_down_metadata_patterns:
                 self.databases.append(database)
+            else:
+                # Filter in Python when pushdown is disabled
+                if not self.filters.filter_config.database_pattern.allowed(
+                    database.name
+                ):
+                    self.report.report_dropped(f"{database.name}.*")
+                else:
+                    self.databases.append(database)
```
sgomezvillamor, line 273: This `if/else` block seems duplicated several times in this file
wondering if this could be refactored in a helper-like method

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_query.py
```diff
@@ -92,38 +203,94 @@ def use_schema(schema_name: str) -> str:
...
-        order by schema_name"""
+        FROM {db_clause}information_schema.schemata
+        WHERE {where_clause}
+        ORDER BY schema_name"""
+
+    @staticmethod
+    def _build_table_type_conditions(
+        table_types: AbstractSet[str],
+        exclude_dynamic_tables: bool = False,
+    ) -> list:
```
sgomezvillamor:
```suggestion
    ) -> List[str]:
```

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_query.py
```diff
@@ -31,6 +31,117 @@ def create_deny_regex_sql_filter(
 
 
 class SnowflakeQuery:
+    @staticmethod
+    def _build_pattern_filter(
+        pattern: Optional[AllowDenyPattern],
+        column_expr: str,
+    ) -> str:
```
sgomezvillamor, line 43: can you double check if we can unify this method and the `create_deny_regex_sql_filter` one above?

```
def create_deny_regex_sql_filter(
    deny_pattern: List[str], filter_cols: List[str]
) -> str:
```

metadata-ingestion/docs/sources/snowflake/snowflake_pre.md
rajatoss: added.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema_gen.py
rajatoss: I looked into this, there are 4 instances of this pattern (databases, schemas, tables, views), but each differs enough that a unified helper would be more complex than the inline code and will decrease the readability.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_query.py
rajatoss: I've unified the two methods, `create_deny_regex_sql_filter` is now a thin wrapper that delegates to `_build_pattern_filter` for SQL generation. This gives us a single code path for all RLIKE-based filtering (both metadata pushdown and lineage/usage deny filters).

rajatoss:
> LGTM
> 
> I left some final comments for your consideration
> 
> Please, remember to add proper integration tests in `connector-tests` for the new feature

Pr with connector tests: https://github.com/acryldata/connector-tests/pull/672

rajatoss:
## Connector Tests Results

**Connector tests failed** for commit `0bd061e`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22487696475)

PR #16102, opened 2026-02-05
fix(cypress): fix cypress tests after updating feature flags

v-tarasevich-blitz-brain:
Linear: https://linear.app/acryl-data/issue/CAT-1329/fix-up-cypress-tests-for-new-homepages-asset-summary-pages

Follow up of https://github.com/datahub-project/datahub/pull/16088


chriscollins3456:
hmm for some reason cypress didn't run here. so let's see how it goes once you fix the conflict and we can make sure everything is green before merging

chriscollins3456 (approved):

PR #16105, opened 2026-02-05
feat(ingest/clickhouse): extract sharding_key for Distributed tables

### LINKED ISSUE #16103: Add sharding key property to Clickhouse tables.
### Summary

Add support for extracting and displaying the `sharding_key` property for ClickHouse Distributed tables. This metadata is critical for understanding data distribution strategy but is currently not captured by the ClickHouse ingestion source.

### Background

**ClickHouse Distributed Engine Syntax:**
```sql
Distributed(cluster, database, table [, sharding_key [, policy_name]])
```

The sharding key determines how data is distributed across shards:
- **Explicit key** (e.g., `cityHash64(user_id)`) → deterministic distribution, enables co-located JOINs
- **No key specified** → ClickHouse defaults to `rand()` (random distribution)

From the [ClickHouse Distributed Engine docs](https://clickhouse.com/docs/en/engines/table-engines/special/distributed):
> *"sharding_key — Sharding key (optional). If the sharding key is not specified, distribution is random."*

### Current Behavior

The ClickHouse source extracts `engine` (e.g., `Distributed`) but does not extract the sharding key from `engine_full`. Users cannot see how data is distributed across shards.

### Proposed Solution

1. **Add `engine_full` to the properties query** to capture the full engine definition including arguments

2. **Add `sharding_key` to customProperties** for Distributed tables

**Example output:**
```
# Explicit sharding key
engine: Distributed
sharding_key: cityHash64(user_id)

# Implicit sharding (no 4th argument)
engine: Distributed  
sharding_key: rand()
```

### References

- [ClickHouse Distributed Table Engine Documentation](https://clickhouse.com/docs/en/engines/table-engines/special/distributed)
- [DataHub ClickHouse Source](https://datahubproject.io/docs/generated/ingestion/sources/clickhouse)

alucaptej:
Closes datahub-project/datahub#16103

Add sharding key extraction from ClickHouse Distributed engine definitions. For tables with explicit sharding (4+ args), extracts the 4th argument. For implicit sharding (3 args), returns 'rand()' as the default as per [Clickhouse Docs](<https://clickhouse.com/docs/engines/table-engines/special/distributed>).

* Add engine_full to PROPERTIES_COLUMNS query
* Add helper functions for parsing Distributed() args
* Override get_table_properties() to extract and expose sharding_key
* Add unit tests for sharding key extraction


metadata-ingestion/src/datahub/ingestion/source/sql/clickhouse.py
```diff
@@ -819,6 +914,22 @@ def _parse_query_log_row(self, row: Dict) -> Optional[ObservedQuery]:
             )
             return None
 
+    def get_table_properties(
+        self, inspector: reflection.Inspector, schema: str, table: str
+    ) -> Tuple[Optional[str], Dict[str, str], Optional[str]]:
+        description, properties, location = super().get_table_properties(
+            inspector, schema, table
+        )
+        engine_full = properties.get("engine_full") if properties else None
```
sgomezvillamor: To avoid this additional override of `get_table_properties`, wouldn’t it make sense to enrich the `sharding_key` from `engine_full` directly at this point?

https://github.com/alucaptej/datahub/blob/ac16ac3cfc2c5c455872343de5d1b2f5bce33b2f/metadata-ingestion/src/datahub/ingestion/source/sql/clickhouse.py#L397-L427

That way, the logic is handled in a single place instead of being split across multiple overrides.

metadata-ingestion/src/datahub/ingestion/source/sql/clickhouse.py
```diff
@@ -819,6 +914,22 @@ def _parse_query_log_row(self, row: Dict) -> Optional[ObservedQuery]:
...
+        sharding_key = _extract_sharding_key(engine_full)
+        if sharding_key:
+            properties["sharding_key"] = sharding_key
+        elif engine_full and "Distributed" in engine_full:
+            logger.debug(f"Could not extract sharding_key from: {engine_full}")
```
sgomezvillamor: If this message needs to be logged, it should be done inside `_extract_sharding_key` itself.
That shouldn’t be the caller’s responsibility.

metadata-ingestion/src/datahub/ingestion/source/sql/clickhouse.py
```diff
@@ -291,11 +291,106 @@ def projects_backward_compatibility(cls, values: Dict) -> Dict:
...
+def _extract_sharding_key(
+    engine_full: Optional[str], default_sharding_key: str = "rand()"
+) -> Optional[str]:
+    """
+    Extract the sharding key from a ClickHouse Distributed engine definition.
+
+    For Distributed tables with 3 arguments (cluster, db, table), ClickHouse
+    implicitly uses rand() as the sharding key. For 4+ arguments, the 4th
+    argument is the explicit sharding key.
+
```
sgomezvillamor, line 364: let's add a reference to https://clickhouse.com/docs/engines/table-engines/special/distributed

metadata-ingestion/src/datahub/ingestion/source/sql/clickhouse.py
```diff
@@ -291,11 +291,106 @@ def projects_backward_compatibility(cls, values: Dict) -> Dict:
...
+    match = re.match(r"^Distributed\s*\((.*)\)\s*$", normalized)
+    if not match:
+        return None
+
+    args = _split_distributed_args(match.group(1))
+    if len(args) < 3:
+        return None
+    if len(args) == 3:
+        # 3 args = cluster, database, table - implicit rand() sharding
+        return default_sharding_key
```
sgomezvillamor, line 388: Instead of solving via regexp and then split, couldn't this be solved with a regexp only?

sgomezvillamor (commented):
Thanks for the contrib!
I left some comments and suggestions

metadata-ingestion/src/datahub/ingestion/source/sql/clickhouse.py
alucaptej: valid, fixed

metadata-ingestion/src/datahub/ingestion/source/sql/clickhouse.py
alucaptej: agreed, fixed

metadata-ingestion/src/datahub/ingestion/source/sql/clickhouse.py
alucaptej: added

metadata-ingestion/src/datahub/ingestion/source/sql/clickhouse.py
alucaptej: here- no, there can be 5th arg policy_name-  `Distributed(cluster, db, table[, sharding_key[, policy_name]])`.
A single regex can't distinguish the 4th from the 5th arg, so i have to have parenthesis aware splitting in addition. 
valid?

sgomezvillamor: makes sense

metadata-ingestion/tests/unit/test_clickhouse_source.py
sgomezvillamor, line 1: Parametrized tests (input, expected) would help here a lot

sgomezvillamor (approved):
LGTM thanks for the contrib

Just a comment on the new tests: parametrized tests may simplify a lot

sgomezvillamor:
Hi @alucaptej — just checking, is there anything blocking this from being merged?

PR #16108, opened 2026-02-05
fix(ingestion): Include s3 resources for file based ingestion source

esteban:
Add pre-requisites that are needed by the `file` ingestion source when importing metadata from `s3`.


metadata-ingestion/setup.py
```diff
@@ -667,6 +667,7 @@
     # S3 includes PySpark by default for profiling support (backward compatible)
     # Standard installation: pip install 'acryl-datahub[s3]' (with PySpark)
     # Lightweight installation: pip install 'acryl-datahub[s3-slim]' (no PySpark, no profiling)
+    "file": {*s3_base},
```
sgomezvillamor, line 670: This is the `file` sink, right? if so, can you move it to the top of the `plugins` dict, where we have all `sinks` listed first?

Also, if the comment above applies to all plugins with the `s3_base` dep, maybe we should add the comment to the `s3_base` itself?

PR #16118, opened 2026-02-06
fix(ingest/snowflake): map MERGE and COPY operations to semantic types

treff7es:
## Summary
- Maps Snowflake `MERGE` operations to `OperationTypeClass.UPDATE` (upsert semantics)
- Maps Snowflake `COPY` operations to `OperationTypeClass.INSERT` (bulk load semantics)
- Aligns Snowflake with BigQuery and Unity Catalog operation type mappings

## Background
MERGE and COPY operations were previously mapped to `CUSTOM`, causing them to be excluded from write operation counts in usage statistics. This fix ensures these common Snowflake operations are properly categorized.

**Cross-source consistency:**
| Source | MERGE | COPY |
|--------|-------|------|
| BigQuery | UPDATE | - |
| Unity Catalog | UPDATE | INSERT |
| Snowflake (before) | CUSTOM | CUSTOM |
| Snowflake (after) | UPDATE | INSERT |


sgomezvillamor:
> CUSTOM, causing them to be excluded from write operation counts in usage statistics

I didn't know about this; wondering what we can do so we don't make such a mistake again 🤔

sgomezvillamor (approved):

PR #16120, opened 2026-02-06
fix(sql-parser): upgrade sqlglot to v29.0.1 and fix semantic view parsing + Block statement handling

sgomezvillamor:
## Summary

Upgrades `sqlglot` from `v27.27.0` to `v29.0.1` to enable Snowflake semantic view parsing and fixes regression issues with Block statement handling introduced in sqlglot v29+.

### Initial Issue: Snowflake Semantic Views
- **Problem**: Infinite loop when parsing Snowflake semantic views
- **Root Cause**: sqlglot v27 didn't support semantic view syntax
- **Upstream Fix**: https://github.com/tobymao/sqlglot/issues/6287 (requires v29+)

### v29 Upgrade Regressions
During the upgrade to v29, two regressions were identified and reported upstream:
1. **ClickHouse**: Cannot parse `PRIMARY KEY tuple()` syntax → https://github.com/tobymao/sqlglot/issues/6989
2. **Semantic View Aliases**: Cannot parse dimensions with aliases (e.g., `DIMENSIONS col1, date_trunc('month', timestamp_col) AS col2`) → https://github.com/tobymao/sqlglot/issues/6993

Both issues were **fixed by sqlglot maintainers** and released in **v29.0.1**.

### v29 Block Statement Handling
sqlglot v29 introduced Block statement parsing for SQL with multiple statements or trailing semicolons. This required additional handling in DataHub's parsing layer to correctly process:
- SQL with double semicolons (e.g., `CREATE VIEW ...;\n;`) → Block([Create, None])
- Multiple statements (handled as error - should use `parse_statements_and_pick()`)

### 1. sqlglot Upgrade
- **Upgrade**: `sqlglot v27.27.0` → `v29.0.1`
- **Reason**: Required for Snowflake semantic view support + includes fixes for reported regressions

### 2. Block Statement Handling (New in v29)
- Added Block statement handling in `sqlglot_utils.py::_parse_statement()`
- Filters out None expressions (empty statements from double semicolons)
- Properly handles error cases:
  - Block with only None expressions → ParseError
  - Block with multiple non-None statements → ParseError (use `parse_statements_and_pick()`)
- Valid case: Block with 1 non-None statement + Nones → extracts the statement

### 3. Semantic View Support
- Added support for extracting table names from `SemanticView` expressions in both `_models.py` and `sqlglot_lineage.py`
- Enables column-level lineage for Snowflake semantic view queries

### 4. v29 Compatibility Updates
- Updated patches to handle v29 API changes:
  - `UPDATE` statement: `from` → `from_` parameter
  - MSSQL temp tables: `global` → `global_` flag
  - Athena: `HOUR()` now parsed as specific `Hour` type instead of `Anonymous`
- Updated golden files to reflect v29 behavior changes:
  - Type inference improvements
  - Better subquery scope tracking
  - JOIN type normalization

### 5. Test Coverage
Added comprehensive test coverage:
- **Semantic views**: Simple queries, qualified names, metrics/dimensions, WHERE clauses
- **Block statements**:
  - Valid case: 1 non-None + Nones (double semicolons)
  - Error case: All None expressions
  - Error case: Multiple non-None statements

## Future Work (Separate PR)
- Migrate from `sqlglot[rs]` to `sqlglot[c]` (Rust → C tokenizer) which is the new standard in v29+

## Test Results
- ✅ All 87 SQL parsing unit tests pass (84 passed, 3 skipped)
- ✅ ClickHouse integration tests now pass (regression fixed upstream)
- ✅ Semantic view parsing tests pass


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `2a4f221`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22391836493)

metadata-ingestion/tests/unit/sql_parsing/goldens/test_postgres_update_subselect.json
```diff
@@ -40,8 +40,7 @@
                 "urn:li:dataset:(urn:li:dataPlatform:postgres,my_db.my_schema.accounts,PROD)"
             ],
             "right_tables": [
-                "urn:li:dataset:(urn:li:dataPlatform:postgres,my_db.my_schema.employees,PROD)",
-                "urn:li:dataset:(urn:li:dataPlatform:postgres,my_db.my_schema.accounts,PROD)"
+                "urn:li:dataset:(urn:li:dataPlatform:postgres,my_db.my_schema.employees,PROD)"
```
treff7es, line 43: Why did we lose `accounts`?

metadata-ingestion/tests/unit/sql_parsing/goldens/test_select_max_with_schema.json
```diff
@@ -11,12 +11,8 @@
             "downstream": {
                 "table": null,
                 "column": "max_col",
-                "column_type": {
-                    "type": {
-                        "com.linkedin.pegasus2avro.schema.NumberType": {}
-                    }
-                },
-                "native_column_type": "DECIMAL"
+                "column_type": null,
```
treff7es, line 14: We lost column type and native column type

metadata-ingestion/tests/unit/sql_parsing/goldens/test_postgres_select_subquery.json
```diff
@@ -82,8 +82,7 @@
                 "urn:li:dataset:(urn:li:dataPlatform:postgres,my_db.my_schema.table1,PROD)"
             ],
             "right_tables": [
-                "urn:li:dataset:(urn:li:dataPlatform:postgres,my_db.my_schema.table2,PROD)",
-                "urn:li:dataset:(urn:li:dataPlatform:postgres,my_db.my_schema.table1,PROD)"
+                "urn:li:dataset:(urn:li:dataPlatform:postgres,my_db.my_schema.table2,PROD)"
```
treff7es, line 85: we lost `my_db.my_schema.table1`

sgomezvillamor: this is a join
before some tables where incorrectly tracked as both left and right tables
newer sqlglot versions do it better

metadata-ingestion/tests/unit/sql_parsing/goldens/test_postgres_update_subselect.json
sgomezvillamor: this is a join
before some tables where incorrectly tracked as both left and right tables
newer sqlglot versions do it better

metadata-ingestion/tests/unit/sql_parsing/goldens/test_select_max_with_schema.json
sgomezvillamor: We were relying on sort of unexpected feature in `sqlglot`

We reported the issue https://github.com/tobymao/sqlglot/issues/6983 and they confirmed that new behaviour is the expected one

This is also tracked in the unit tests:

```
def test_select_max_with_schema() -> None:
    # Note that `this_will_not_resolve` will be dropped from the result because it's not in the schema.

    # With sqlglot v27: When it sees MAX(DECIMAL, DECIMAL, UNKNOWN), it was "smart" enough to infer the result as DECIMAL
    # With sqlglot v28: Same input → returns UNKNOWN (more conservative behavior)
    # Sqlglot confirmed this is intentional - v28 is more conservative with type inference when any argument is UNKNOWN.
    # See https://github.com/tobymao/sqlglot/issues/6983
    # The impact is that we miss type information for the max_col column when queries reference unresolved columns.
    # That doesn't happen with the test_select_max_with_schema_all_resolved test because all columns are resolved.
```

PR #16124, opened 2026-02-06
feat(clickhouse): emit DatasetProfile stats without profiling  

kyungsoo-datahub:
- Emit `DatasetProfile` aspect with `rowCount` and `sizeInBytes` for ClickHouse tables                                  
  - Values come from pre-computed `total_rows` and `total_bytes` in `system.tables`
  - Enables Stats tab in DataHub UI without expensive COUNT(*) queries


sgomezvillamor (commented):
AFAIK we usually emit the `DatasetProfile` aspect only when profiling is enabled. Could you double check for other sources?

Having an optimized and inexpensive way to compute total row count and total bytes is great, but I’d expect it to be used only within the ClickHouse profiling module.

PR #16130, opened 2026-02-06
fix(ui/ingest/create): Remote AI chat placeholder

asikowitz:

### DIFF
--- a/datahub-web-react/src/app/ingestV2/runDetails/IngestionRunDetailsPage.tsx
+++ b/datahub-web-react/src/app/ingestV2/runDetails/IngestionRunDetailsPage.tsx
@@ -67,7 +67,7 @@ export default function IngestionRunDetailsPage() {
     );
 
     return (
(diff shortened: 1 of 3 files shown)


datahub-web-react/src/app/ingestV2/runDetails/IngestionRunDetailsPage.tsx
```diff
@@ -67,7 +67,7 @@ export default function IngestionRunDetailsPage() {
     );
 
     return (
-        <PageLayout title="Run Details" titlePill={titlePill} rightPanelContent={<AIChat />} topBreadcrumb={breadCrumb}>
+        <PageLayout title="Run Details" titlePill={titlePill} rightPanelContent={AIChat()} topBreadcrumb={breadCrumb}>
```
asikowitz, line 70: AIChat() instead of <AIChat /> properly passes null

PR #16138, opened 2026-02-08
fix(clickhouse): add direct SRC→TO lineage for materialized views with TO clause

kyungsoo-datahub:
Previously, ClickHouse materialized views with TO clause only generated:
- SRC → MV (table + CLL)
- MV → TO (table-level only)

This caused lineage to appear as a chain: SRC → MV → TO

Now generates:
- SRC → MV (table + CLL)
- SRC → TO (table + CLL)

Changes:
- Updated materialized_view_lineage_query to add direct source→target lineage
- Removed MV→TO lineage (not needed)
- Updated _add_view_to_aggregator to register CLL for both MV and TO table

### DIFF
--- a/metadata-ingestion/tests/integration/clickhouse/clickhouse_mces_golden.json
+++ b/metadata-ingestion/tests/integration/clickhouse/clickhouse_mces_golden.json
@@ -123,7 +123,7 @@
                         "time": 0,
                         "actor": "urn:li:corpuser:unknown"
                     },
-                    "dataset": "urn:li:dataset:(urn:li:dataPlatform:clickhouse,clickhousetestserver.db1.mv_with_target_table,PROD)",
+                    "dataset": "urn:li:dataset:(urn:li:dataPlatform:clickhouse,clickhousetestserver.db1.test_data_types,PROD)",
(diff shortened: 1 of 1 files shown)


metadata-ingestion/src/datahub/ingestion/source/sql/clickhouse.py
```diff
@@ -985,6 +989,9 @@ def _populate_lineage(self) -> None:
         )
 
         # get materialized view downstream and upstream
+        # This query generates 2 types of lineage for materialized views with TO clause:
+        # 1. SRC → MV: Source tables referenced in SELECT → materialized view
+        # 2. SRC → TO: Source tables → target table (direct lineage)
```
sgomezvillamor, line 994: should we update or mention this in the user-facing docs too?

kyungsoo-datahub: Thank you for the comment. Clickhouse doesn't have -pre.md file yet. Will address this in a separate PR for comprehensive user-facing doc for Clickhouse.

PR #16142, opened 2026-02-09
fix(ingestion/oracle): retrieve database name when using service_name with ALL mode

acrylJonny:
## Summary

Fixes an issue where Oracle database containers were created with no name (only an ID) when using `service_name` instead of `database` configuration with `data_dictionary_mode: ALL`.

## Changes

- Modified `OracleSource.get_db_name()` to query Oracle for the database name using `sys_context('USERENV','DB_NAME')` when the database name cannot be determined from the connection URL
- Added fallback query for regular SQLAlchemy inspector (ALL mode) while preserving existing behavior for `OracleInspectorObjectWrapper` (DBA mode)
- Added unit test `test_oracle_get_db_name_with_service_name()` to verify the fix

## Root Cause

When using `service_name` instead of `database` in the Oracle configuration:
- The SQLAlchemy connection URL contains no database name (`engine.url.database = None`)
- The existing fallback to query Oracle only worked with `data_dictionary_mode: DBA` (which uses `OracleInspectorObjectWrapper`)
- Users with `data_dictionary_mode: ALL` (the default) would get empty database names, resulting in containers with no name

## Test Plan

- [x] All existing Oracle unit tests pass (18 tests)
- [x] New test verifies database name retrieval with `service_name` and `data_dictionary_mode: ALL`
- [x] Code passes linting (`./gradlew :metadata-ingestion:lintFix`)

## Impact

Users with Oracle configurations using `service_name` and `data_dictionary_mode: ALL` will now see properly named database containers instead of containers with only IDs.

Example configuration that benefits from this fix:
```
source:
  type: oracle
  config:
    service_name: test_service_name
    data_dictionary_mode: ALL  # or omitted (default)
```

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
@@ -1108,6 +1108,7 @@ class OracleSource(SQLAlchemySource):
         In that case, it tries to retrieve the database name by sending a query to the DB.
 
         Note: This is used as a fallback if database is not specified in the config.
+        Returns a normalized (lowercased) database name for consistency with schema/table names.
         """
 
         # call default implementation first
@@ -1135,7 +1136,7 @@ class OracleSource(SQLAlchemySource):
             # Normalize database name to match Oracle dialect behavior
             # Oracle returns names in uppercase, but SQLAlchemy's normalize_name
             # converts them to lowercase for consistency with schema/table names
-            if db_name and hasattr(inspector, "dialect"):
+            if db_name:
                 normalized = inspector.dialect.normalize_name(db_name)
                 if normalized:
                     db_name = normalized
--- a/metadata-ingestion/tests/integration/oracle/golden_files/golden_mces_oracle_to_file_with_all_tables.json
+++ b/metadata-ingestion/tests/integration/oracle/golden_files/golden_mces_oracle_to_file_with_all_tables.json
@@ -1,7 +1,7 @@
 [
 {
     "entityType": "container",
-    "entityUrn": "urn:li:container:3a4688ecc84119d5c1b3004e4a9a63b0",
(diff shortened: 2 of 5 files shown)


metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
```diff
@@ -1011,8 +1015,32 @@ def get_db_name(self, inspector: Inspector) -> str:
...
+                except sqlalchemy.exc.DatabaseError as e:
+                    logger.warning(
+                        f"Error fetching database name using sys_context: {e}",
+                        exc_info=True,
+                    )
+
+            # Normalize database name to match Oracle dialect behavior
+            # Oracle returns names in uppercase, but SQLAlchemy's normalize_name
+            # converts them to lowercase for consistency with schema/table names
+            if db_name and hasattr(inspector, "dialect"):
```
kyungsoo-datahub: SQLAlchemy inspectors always have .dialect. I don't think wee need this guard. If it's missing, we'd want to fail loudly rather than silently skip normalization. WDYT?

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
kyungsoo-datahub, line 1103: nit: Would be nice have brief comment for this function to notify that it produces normalized db name.

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
acrylJonny: fair point - we will always have dialect here so you're correct. I've updated this

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
acrylJonny: extra line now added to the comment at the top of the method

PR #16145, opened 2026-02-09
fix(CLI): filter on datasets for platform instance migration

kevinkarchacryl:
When doing `datahub migrate dataplatform2instance --platform snowflake --instance <some-instance> --env <some-env>`
I am seeing this error 
```
line 160, in dataplatform2instance_func
    assert key
           ^^^
AssertionError
```
I believe this is because we are not filtering by datasets and when a container shows up, it throws an exception


metadata-ingestion/src/datahub/cli/migrate.py
```diff
@@ -155,7 +155,9 @@ def dataplatform2instance_func(
     urns_to_migrate: List[str] = []
 
     # we first calculate all the urns we will be migrating
-    for src_entity_urn in graph.get_urns_by_filter(platform=platform, env=env):
+    for src_entity_urn in graph.get_urns_by_filter(
+        platform=platform, env=env, entities_types=["dataset"]
```
sgomezvillamor: what about dashboards, dataflows, datajobs, ...?
or was this migration only expected to be used for datasets?

kevinkarchacryl: I believe it's only meant for datasets. I also found these comments in mce_builder.py, I assume it's still true:
```
# FIXME: dashboards don't currently include data platform urn prefixes.
 # FIXME: charts don't currently include data platform urn prefixes.
```

sgomezvillamor: @cursoragent could you double check if this migration feature was meant only for datasets?

sgomezvillamor (approved):
LGTM

it would be great if you could make more explicit in the help command that this migration tool is only for datasets and containers

PR #16147, opened 2026-02-09
feat(ingestion/oracle): implement standard stored procedure lineage pattern

acrylJonny:
## Summary

Implements proper stored procedure lineage for Oracle connector following DataHub's standard pattern. Adds comprehensive testing for query usage extraction from V$SQL and table-to-table lineage from DML operations.

### Core Implementation
- **Added `include_lineage` config**: New boolean field (default: `true`) for stored procedure lineage extraction
- **Added `include_lineage` property**: Combines `include_lineage` and `include_view_lineage` flags
- **Set schema context**: Added `default_db` and `default_schema` to `BaseProcedure` instances for SQL parsing
- **Fixed domain handling**: Replaced incorrect domain URN logic with `get_domain_wu` utility
- **Leverages parent class**: Uses `generate_procedure_lineage()` from `stored_procedures/base.py`

### Documentation
- **Updated prerequisites**: Added V$SQL permissions section for query usage feature

### Validated Features
- **Stored procedure lineage**: `PROCESS_ORDER` shows `2 inputs → 3 outputs` with cross-schema dependencies
- **Table-to-table lineage**: DML operations captured (e.g., `employees → employee_backup`, `orders → daily_revenue`)
- **View lineage**: Properly extracted from view definitions
- **Query usage**: 29 queries captured from V$SQL

### DIFF
--- a/metadata-ingestion/tests/integration/oracle/test_oracle.py
+++ b/metadata-ingestion/tests/integration/oracle/test_oracle.py
@@ -60,12 +60,19 @@ def filter_volatile_vsql_queries(metadata_json: List[dict]) -> List[dict]:
             ):
                 volatile_query_urns.add(entity.get("entityUrn", ""))
 
-    # Second pass: filter out volatile queries and their lineage
+    # Second pass: filter out volatile queries, staging datasets, and their lineage
     for entity in metadata_json:
         entity_urn = entity.get("entityUrn", "")
 
         # Skip query entities that reference volatile staging tables
         if entity.get("entityType") == "query" and entity_urn in volatile_query_urns:
+            continue
+
+        # Skip all aspects of staging table datasets (they only exist due to volatile queries)
+        if (
+            entity.get("entityType") == "dataset"
+            and entity_urn in VOLATILE_STAGING_TABLES
+        ):
             continue
 
         # Skip upstreamLineage that references volatile queries
--- a/metadata-ingestion/tests/integration/oracle/golden_files/golden_mces_oracle_to_file_with_query_usage.json
+++ b/metadata-ingestion/tests/integration/oracle/golden_files/golden_mces_oracle_to_file_with_query_usage.json
@@ -3114,7 +3114,7 @@
                             "actor": "urn:li:corpuser:_ingestion"
                         },
                         "created": {
-                            "time": 1770980715000,
+                            "time": 1770987565000,
                             "actor": "urn:li:corpuser:_ingestion"
                         },
                         "dataset": "urn:li:dataset:(urn:li:dataPlatform:oracle,xepdb1.sales_schema.orders,PROD)",
@@ -3172,11 +3172,11 @@
                 },
                 "source": "SYSTEM",
                 "created": {
-                    "time": 1770980715000,
+                    "time": 1770987565000,
                     "actor": "urn:li:corpuser:_ingestion"
                 },
                 "lastModified": {
-                    "time": 1770980715000,
+                    "time": 1770987565000,
                     "actor": "urn:li:corpuser:_ingestion"
                 }
             }
@@ -3252,7 +3252,7 @@
                             "actor": "urn:li:corpuser:_ingestion"
                         },
                         "created": {
-                            "time": 1770980715000,
+                            "time": 1770987565000,
                             "actor": "urn:li:corpuser:_ingestion"
                         },
                         "dataset": "urn:li:dataset:(urn:li:dataPlatform:oracle,xepdb1.hr_schema.employees,PROD)",
@@ -3427,11 +3427,11 @@
                 },
                 "source": "SYSTEM",
                 "created": {
-                    "time": 1770980715000,
+                    "time": 1770987565000,
                     "actor": "urn:li:corpuser:_ingestion"
                 },
                 "lastModified": {
-                    "time": 1770980715000,
+                    "time": 1770987565000,
                     "actor": "urn:li:corpuser:_ingestion"
                 }
             }


metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
```diff
@@ -1518,6 +1698,164 @@ def _get_materialized_view_definition_fallback(
...
+
+        with self.report.new_stage("Schema metadata extraction"):
+            yield from super().get_workunits_internal()
+            logger.info("Completed schema metadata extraction")
+
+        if self.config.include_query_usage or self.config.include_usage_stats:
+            self._populate_aggregator_from_queries()
+
+        with self.report.new_stage("Lineage and usage processing"):
+            yield from super()._generate_aggregator_workunits()
```
kyungsoo-datahub, line 2118: super().get_workunits_internal() already calls _generate_aggregator_workunits(). Calling it again duplicates all lineage MCPs. Override it as a no-op and call the aggregator once after adding V$SQL queries.

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
```diff
@@ -1518,6 +1698,164 @@ def _get_materialized_view_definition_fallback(
...
+                            if re.search(pattern, sql_text, re.IGNORECASE):
+                                logger.debug(
+                                    f"Excluding query matching pattern '{pattern}': {sql_text[:100]}..."
+                                )
+                                should_exclude = True
+                                break
+                        if should_exclude:
+                            continue
+
+                    db_name = self.get_db_name(inspector)
```
kyungsoo-datahub: Do we need to call this every row? DB name doesn't change per row.

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
```diff
@@ -201,6 +281,29 @@ class DataDictionaryMode(StrEnum):
...
+            sql_id,
+            sql_text,
+            parsing_schema_name,
+            executions,
+            elapsed_time/1000000 as elapsed_seconds,
+            first_load_time
+        FROM V$SQL
+        WHERE parsing_schema_name IS NOT NULL
+            AND parsing_schema_name NOT IN ({system_schemas})
+            AND command_type IN (2, 3, 6, 7, 189)
```
kyungsoo-datahub: Would be nice to define the numbers as constants to represent what they mean.

metadata-ingestion/tests/integration/oracle/test_oracle.py
```diff
@@ -20,6 +24,56 @@
 
 FROZEN_TIME = "2022-02-03 07:00:00"
 
+# V$SQL DML query URNs that are volatile (may or may not be in cache)
+# These are from test setup INSERT operations (employee_backup, daily_revenue)
+VOLATILE_VSQL_QUERY_URNS = {
+    "urn:li:query:532ceadbf6076617af1a7ac10b50c4e76cab762e2265566dacc64aa25c2160b3",  # employee_backup INSERT
```
kyungsoo-datahub: Do we want to keep these specific hashes? Could be burden to maintain whenever we want to change input data.

metadata-ingestion/tests/integration/oracle/startup/01_setup.sql
```diff
@@ -8,6 +8,9 @@
 -- Connect to the PDB first
 ALTER SESSION SET CONTAINER = XEPDB1;
 
+-- Grant DBMS_SHARED_POOL access to SYSTEM for query pinning in V$SQL tests
+GRANT EXECUTE ON SYS.DBMS_SHARED_POOL TO SYSTEM;
```
kyungsoo-datahub: Which test is using this DBMS_SHARED_POOL?

metadata-ingestion/tests/integration/oracle/source_files/oracle_to_file_materialized_view_lineage.yml
```diff
@@ -23,6 +23,14 @@ source:
         - "^ANALYTICS_SCHEMA$"  # Has materialized views
         - "^SYSTEM$"            # Has test_mview
     
+    # Exclude staging tables created in post_setup.sql for DML lineage testing
```
kyungsoo-datahub: Not a blocker since this follows existing patterns, but a separate schema for these tables would avoid updating 6 test configs.

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
```diff
@@ -1518,6 +1698,164 @@ def _get_materialized_view_definition_fallback(
...
+
+    def get_workunits_internal(self) -> Iterable[Union[MetadataWorkUnit, SqlWorkUnit]]:
+        """Override to add query extraction for usage statistics."""
+        logger.info("Starting Oracle metadata extraction")
+
+        with self.report.new_stage("Schema metadata extraction"):
+            yield from super().get_workunits_internal()
+            logger.info("Completed schema metadata extraction")
+
+        if self.config.include_query_usage or self.config.include_usage_stats:
```
kyungsoo-datahub: The if condition here is redundant in that _populate_aggregator_from_queries already guards on
  include_query_usage. Either remove the outer check or make them consistent.

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
acrylJonny: Fixed. Overrode `_generate_aggregator_workunits()` as a no-op and restructured `get_workunits_internal()` to control when aggregator workunits are generated, ensuring they're only emitted once after all queries (both table usage and `V$SQL` queries) have been added to the aggregator.
Changes:

- Overrode `_generate_aggregator_workunits()` to return empty generator
- Restructured `get_workunits_internal()` to explicitly call `super()._generate_aggregator_workunits()` only after `V$SQL` processing completes
- Removed redundant if condition that was guarding the aggregator call

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
acrylJonny: Moved `db_name = self.get_db_name(inspector)` outside the row processing loop. The database name is constant for the entire connection, so calling it once before the loop is sufficient.

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
acrylJonny: Created `OracleSQLCommandType` class with named constants for Oracle's `COMMAND_TYPE` values, and defined `VSQL_DML_COMMAND_TYPES` tuple using these constants.

metadata-ingestion/tests/integration/oracle/test_oracle.py
acrylJonny: Replaced hardcoded query hash URNs with a more maintainable approach

metadata-ingestion/tests/integration/oracle/startup/01_setup.sql
acrylJonny: Removed. After investigation, the `GRANT EXECUTE ON SYS.DBMS_SHARED_POOL TO SYSTEM;` statement was not being used by any test. It was originally added for potential query pinning in `V$SQL` tests but never implemented. Removed to simplify the setup script.

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
acrylJonny: Removed the redundant outer `if self.config.include_query_usage or self.config.include_usage_stats:` check since the logic within `_populate_aggregator_from_queries()` already handles these conditions.

metadata-ingestion/tests/integration/oracle/source_files/oracle_to_file_materialized_view_lineage.yml
acrylJonny: Created a dedicated `STAGING_SCHEMA` for all test staging tables, eliminating the need for table exclusion patterns in configuration files.

PR #16149, opened 2026-02-10
fix(ingestion/datahub): prevent partial entity sync in DataHub source using URN pattern filtering

aviraj-gour:
### Problem

Linear Issue: ING-1466
The DataHub source creates broken/invalid entities when copying metadata between instances due to aspect-level filtering.

The current implementation filters aspects independently using `DEFAULT_EXCLUDE_ASPECTS`. When sensitive aspects like `dataHubIngestionSourceInfo` are excluded, other aspects for the same entity (e.g., `ownership`) can still sync through, creating partial entities that lack their key configuration aspect.

### Solution

Implemented **URN pattern filtering** to exclude entire entity types that contain environment-specific or sensitive data:

- Added `DEFAULT_URN_DENY_PATTERNS` with regex patterns for entities that shouldn't be copied:
  - `urn:li:dataHubIngestionSource:.*` - Ingestion source configs
  - `urn:li:dataHubSecret:.*` - Secrets and credentials
  - `urn:li:globalSettings:.*` - Environment settings
  - `urn:li:dataHubExecutionRequest:.*` - Execution requests

- Added `urn_pattern` config field using DataHub's standard `AllowDenyPattern` class
- Applied filtering in all three readers (Database, Kafka, API)

This ensures that when an entity type is denied, **all its aspects** (including ownership) are filtered out, preventing creation of partial/broken entities.

### DIFF
--- a/metadata-ingestion/tests/unit/test_datahub_source.py
+++ b/metadata-ingestion/tests/unit/test_datahub_source.py
@@ -4,11 +4,14 @@ from unittest.mock import MagicMock, patch
 
 import pytest
 
+from datahub.ingestion.api.common import PipelineContext
+from datahub.ingestion.source.datahub.config import DataHubSourceConfig
 from datahub.ingestion.source.datahub.datahub_database_reader import (
     DATETIME_FORMAT,
     DataHubDatabaseReader,
     VersionOrderer,
 )
+from datahub.ingestion.source.datahub.datahub_source import DataHubSource
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


docs/how/updating-datahub.md
```diff
@@ -37,6 +37,9 @@ This file documents any backwards-incompatible changes in DataHub and assists pe
   - Spark integration upgraded from 3.0.3 to 3.3.4 (minimum version required for Java 17 support)
   - Hadoop upgraded from 2.7.2 to 3.3.6 (addresses Hadoop CVEs, bundled with Spark 3.3+)
   - Note: If you're a self-hosted user still running Java 11, you must upgrade to Java 17 before deploying this release. Spark lineage users must upgrade to Spark 3.3.0+.
+- #16149 (Ingestion) DataHub source: now uses URN pattern filtering to exclude environment-specific entities by default. This prevents copying credentials and creating invalid entities.
+  - **Default behavior change**: Previously `exclude_aspects` contained `dataHubIngestionSourceInfo`, `dataHubSecretValue`, `dataHubExecutionRequestInput`, `globalSettingsInfo`. Now `urn_pattern.deny` contains `urn:li:dataHubIngestionSource:.*`, `urn:li:dataHubSecret:.*`, `urn:li:globalSettings:.*`, `urn:li:dataHubExecutionRequest:.*` to exclude entire entity types instead of individual aspects.
+  - **Note**: If you override `urn_pattern` or `exclude_aspects` in recipe, carefully configure them based on your requirements to avoid syncing sensitive data or creating invalid entities. Recommended to keep new defaults.
```
skrydal, line 43: I would consider adding pydantic-validator based (or even better - do the check in the source code, so that warning will surface in UI, via report.warning method) **warning** to whoever sets urn_pattern, as previously they were overriding allow-all.

PR #16151, opened 2026-02-10
refactor(ingestion): save and run without browser state

v-tarasevich-blitz-brain:
Linear: https://linear.app/acryl-data/issue/CAT-1267/code-refactor-look-into-refactoring-save-and-run-in-new-ingestion

This PR removed passing state to history.push to handle "Save and Run"

### DIFF
--- a/datahub-web-react/src/app/ingestV2/source/IngestionSourceList.tsx
+++ b/datahub-web-react/src/app/ingestV2/source/IngestionSourceList.tsx
@@ -146,7 +146,12 @@ export const IngestionSourceList = ({
 }: Props) => {
     const location = useLocation();
 
-    const { createdOrUpdatedSource, shouldRunCreatedOrUpdatedSource } = useIngestionContext();
+    const {
+        createdOrUpdatedSource,
+        shouldRunCreatedOrUpdatedSource,
+        setCreatedOrUpdatedSource,
+        setShouldRunCreatedOrUpdatedSource,
+    } = useIngestionContext();
 
     // Query inputs after redirect to restore initial state of query and sorting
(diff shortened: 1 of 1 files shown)


datahub-web-react/src/app/ingestV2/source/IngestionSourceList.tsx
```diff
@@ -470,26 +465,21 @@ export const IngestionSourceList = ({
...
+            setIsCreatedOrUpdatedSourceHandled(true);
+            if (shouldRunCreatedOrUpdatedSource) {
+                executeIngestionSource(createdOrUpdatedSource);
             } else {
-                setSourcesToRefetch((prev) => new Set([...prev, createdOrUpdatedSourceUrnFromLocation]));
+                setSourcesToRefetch((prev) => new Set([...prev, createdOrUpdatedSource]));
             }
-            // clear browser state vars for shouldRun to prevent duplicate runs
-            history.replace(history.location.pathname, { ...location.state, shouldRun: undefined });
         }
```
chriscollins3456, line 482: so i'm not positive that this would not reintroduce the main bug we discovered here.

the way you can check is do the following:
- create a new source and click save & run
- go to view the run results
- click the back button
- this should retrigger a new ingestion execution request

the reason i think this would happen is because the state for `createdOrUpdatedSource` and `shouldRunCreatedOrUpdatedSource` would likely remain intact through the back button on the browser because it's in the context above the route but `isCreatedOrUpdatedSourceHandled` would get reset

we can certainly reset both variables in the context here though at the bottom where i was calling `history.replace` in the diff that was removed and that should hopefully fix this

chriscollins3456 (commented):
let's double check the bug we had to fix is not re-introduced!

PR #16152, opened 2026-02-10
fix(ingestion): prevent changing of source type for existing ingestion sources

v-tarasevich-blitz-brain:
Linear: https://linear.app/acryl-data/issue/CAT-1225/bug-we-can-change-source-type-via-yaml-editor-and-save-changes

This fix prevents changing of source type for existing ingestion source


chriscollins3456 (approved):
nice this makes sense to me

PR #16153, opened 2026-02-10
improvement(ingestion): update ordering of filter options

v-tarasevich-blitz-brain:
Linear: https://linear.app/acryl-data/issue/CAT-1038/ingestion-source-update-ordering-of-options-of-subtype-select-in

This PR updates ordering of filters for ingestion sources

### DIFF
--- a/datahub-web-react/src/app/ingestV2/source/builder/RecipeForm/confluence.ts
+++ b/datahub-web-react/src/app/ingestV2/source/builder/RecipeForm/confluence.ts
@@ -190,6 +190,7 @@ export const CONFLUENCE_SPACE_ALLOW: FilterRecipeField = {
     buttonLabel: 'Add space',
     rules: null,
     section: 'Spaces',
+    filteringResource: 'Space',
     rule: FilterRule.INCLUDE,
     setValueOnRecipeOverride: (recipe: any, values: string[]) =>
         setListValuesOnRecipe(recipe, values, spaceAllowFieldPath),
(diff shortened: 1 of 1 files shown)


datahub-web-react/src/app/ingestV2/source/multiStepBuilder/steps/step2ConnectionDetails/sections/recipeSection/sections/filtersSection/utils.ts
```diff
@@ -58,12 +58,10 @@ export function getOptionsForTypeSelect(): SelectOption[] {
 }
 
 export function getSubtypeOptions(fields: FilterRecipeField[]): SelectOption[] {
-    return [...new Set(fields.map((field) => field.section))]
-        .map((section) => ({
-            label: section,
-            value: section,
-        }))
-        .sort((optionA, optionB) => optionA.label.localeCompare(optionB.label));
```
chriscollins3456, line 66: so are we getting rid of `label` entirely then? can we remove it from the configs?

and do we want the `filteringResource` which appears like a singular noun instead of plural (ie. "Database" instead of "Databases") to be what the user sees in the UI? is that what this is doing?

v-tarasevich-blitz-brain: I think that we have to keep label until old version is available as sections are using there. New field  `filteringResource`  is used only as options for filters in the new version

PR #16154, opened 2026-02-10
tests(cypress): add lineage v3 tests

v-tarasevich-blitz-brain:
This PR adds lineage v3 cypress tests

### DIFF
--- a/smoke-test/tests/cypress/cypress/e2e/lineageV3/v3_impact_analysis.js
+++ b/smoke-test/tests/cypress/cypress/e2e/lineageV3/v3_impact_analysis.js
@@ -1,5 +1,5 @@
 import { getTimestampMillisNumDaysAgo } from "../../support/commands";
-import { setFeatureFlags } from "./utils";
+import { setLineageV3FeatureFlags } from "./utils";
 
 const JAN_1_2021_TIMESTAMP = 1609553357755;
 const JAN_1_2022_TIMESTAMP = 1641089357755;
@@ -21,7 +21,7 @@ const startAtDataSetLineage = () => {
 
 describe("impact analysis", () => {
   beforeEach(() => {
-    setFeatureFlags();
+    setLineageV3FeatureFlags();
   });
 
   it("can see 1 hop of lineage by default", () => {
--- a/smoke-test/tests/cypress/cypress/e2e/lineageV3/v3_download_lineage_results.js
+++ b/smoke-test/tests/cypress/cypress/e2e/lineageV3/v3_download_lineage_results.js
@@ -1,3 +1,5 @@
+import { setLineageV3FeatureFlags } from "./utils";
+
 const test_dataset =
   "urn:li:dataset:(urn:li:dataPlatform:kafka,SampleCypressKafkaDataset,PROD)";
 const first_degree = [
@@ -31,7 +33,7 @@ const downloadCsvFile = (filename) => {
 
 describe("download lineage results to .csv file", () => {
   beforeEach(() => {
-    cy.setIsThemeV2Enabled(true);
+    setLineageV3FeatureFlags();
     cy.on("uncaught:exception", (err, runnable) => false);
   });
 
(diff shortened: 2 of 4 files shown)


smoke-test/tests/cypress/cypress/e2e/lineageV2/utils.js
```diff
@@ -75,3 +75,71 @@ export function checkIfEdgeBetweenColumnsNotExist(
     `rf__edge-${node1Urn}::${column1Name}-${node2Urn}::${column2Name}`,
   ).should("not.exist");
 }
+
+function getFilteringNode(nodeUrn, direction) {
+  const directionSign = direction === "up" ? "u" : "d";
+  return cy.getWithTestId(`rf__node-lf:${directionSign}:${nodeUrn}`);
+}
```
chriscollins3456, line 82: can we just have one utils file we use here that works across lineage versions tests? or do they need to be separate?

maybe they need to be separate since the code is separate we may update one set of test ids and not another?

ideally i do not want to maintain all of these different lineage test suites. i suppose at some point we will be able to delete old ones once we remove feature flags and delete old code no one uses

smoke-test/tests/cypress/cypress/e2e/lineageV3/v3_impact_analysis.js
```diff
@@ -0,0 +1,201 @@
...
+  });
+
+  it("can filter lineage edges by time", () => {
+    cy.login();
+    cy.visit(
+      `/dataset/${DATASET_URN}/Lineage?filter_degree___false___EQUAL___0=1&is_lineage_mode=false&page=1&unionType=0&start_time_millis=${JAN_1_2021_TIMESTAMP}&end_time_millis=${JAN_1_2022_TIMESTAMP}`,
+    );
+
+    // impact analysis can take a beat- don't want to time out here
+    cy.wait(5000);
```
chriscollins3456: why do we need these waits? i suppose we're just copying the tests for v2 lineage?

i'd rather not have these waits if we can

chriscollins3456 (commented):
this is pretty tough to review since it's so huge, but i tried to walk through and make sure things made sense.

in general it looks like we're duplicating these tests from v2 lineage where we can and turning on the flag for v3.

my main ask is to make sure that for the tests we're duplicating from v2 that we have any of the fixes that have gone into those tests since you created this PR that we include those.

in the future i think it would make sense to break these tests up into smaller PRs. not as huge of a deal for adding cypress tests vs real feature work though.

smoke-test/tests/cypress/cypress/e2e/lineageV3/v3_impact_analysis.js
v-tarasevich-blitz-brain: yeah, it's the same test from v2 lineage, I'll try to get rid of this wait

smoke-test/tests/cypress/cypress/e2e/lineageV2/utils.js
v-tarasevich-blitz-brain: I think I can remove lineageV3/utils.js for now

datahub-web-react/src/app/lineageV3/LineageFilterNode/LineageFilterNodeBasic.tsx
```diff
@@ -25,17 +24,17 @@ import { EntityType } from '@types';
 export const LINEAGE_FILTER_NODE_NAME = 'lineage-filter';
 
 const NodeWrapper = styled.div`
-    background-color: white;
-    border: 1px solid ${LINEAGE_COLORS.NODE_BORDER};
+    background-color: ${(props) => props.theme.colors.bg};
+    border: 1px solid ${(props) => props.theme.colors.borderFocused};
```
annadoesdesign, line 28: this can just be border not border focused (border focused is only when a its selected (and maybe thats whats happening)

PR #16155, opened 2026-02-10
feat(ingestion): dynamodb configurable maxitems

davidwalker-tp:
Updates the DynamoDb ingestion functionality to allow user to specify how many items to retrieve from DynamoDb. 

Originally this value was hardcoded to `100` - this change allows a positive value to be specified in the ingestion recipe by adding a `schema_sampling_size` config setting. If not set, it defaults to `100`.

### DIFF
--- a/metadata-ingestion/tests/unit/dynamodb/test_dynamodb.py
+++ b/metadata-ingestion/tests/unit/dynamodb/test_dynamodb.py
@@ -215,14 +215,9 @@ class TestDynamoDBSchemaSampling:
         mock_page = {"Items": [{"field1": {"S": "value1"}}]}
         mock_paginator.paginate.return_value = [mock_page]
 
-        # Mock include_table_item_to_schema to do nothing
-        with (
-            patch.object(source, "include_table_item_to_schema"),
-            patch.object(source, "construct_schema_from_items"),
-        ):
-            source.construct_schema_from_dynamodb(
-                mock_dynamodb_client, "us-west-2", "test_table"
-            )
+        source.construct_schema_from_dynamodb(
+            mock_dynamodb_client, "us-west-2", "test_table"
+        )
 
         # Verify paginate was called with the correct MaxItems
         mock_paginator.paginate.assert_called_once()


metadata-ingestion/tests/unit/dynamodb/test_dynamodb.py
```diff
@@ -189,3 +190,71 @@ def test_tag_format_variations(
...
+    def test_custom_schema_sampling_size(self, mock_context):
+        """Test that custom schema_sampling_size is properly set."""
+        config = DynamoDBConfig(
+            aws_access_key_id="test",
+            aws_secret_access_key="test",
+            aws_region="us-west-2",
+            schema_sampling_size=500,
+        )
+        assert config.schema_sampling_size == 500
+
```
skrydal: Let's remove those tests, they don't test any business logic.
```suggestion
```

davidwalker-tp: Good point, will remove.

metadata-ingestion/src/datahub/ingestion/source/dynamodb/dynamodb.py
skrydal, line 364: Major concern related to this value was the cost of scan, how about we add info-level message if the value is higher than the default?
```suggestion
        self.include_table_item_to_schema(dynamodb_client, region, table_name, schema)
        if self.config.schema_sampling_size > PAGE_SIZE:
            self.report.info(
                title="Large schema sampling size configured",
                message="High schema_sampling_size increases DynamoDB read capacity consumption and ingestion time.",
                context=f"schema_sampling_size={self.config.schema_sampling_size}",
            )
```

metadata-ingestion/tests/unit/dynamodb/test_dynamodb.py
```diff
@@ -189,3 +190,71 @@ def test_tag_format_variations(
...
+    """Test suite for DynamoDB schema sampling configuration"""
+
+    @pytest.fixture
+    def mock_context(self):
+        """Fixture for mock pipeline context."""
+        mock_ctx = MagicMock(spec=PipelineContext)
+        mock_ctx.pipeline_name = "test_pipeline"
+        mock_ctx.run_id = "test_run"
+        mock_ctx.graph = None
+        return mock_ctx
```
skrydal: Such fixture is already defined for the previous test suite in this class, could you please extract it to the module level?

metadata-ingestion/tests/unit/dynamodb/test_dynamodb.py
```diff
@@ -189,3 +190,71 @@ def test_tag_format_variations(
...
+        mock_paginator.paginate.return_value = [mock_page]
+
+        # Mock include_table_item_to_schema to do nothing
+        with (
+            patch.object(source, "include_table_item_to_schema"),
+            patch.object(source, "construct_schema_from_items"),
+        ):
+            source.construct_schema_from_dynamodb(
+                mock_dynamodb_client, "us-west-2", "test_table"
+            )
```
skrydal: What's the reason for patching those functions? Test passes even without it.


```suggestion
        source.construct_schema_from_dynamodb(
            mock_dynamodb_client, "us-west-2", "test_table"
        )
```

metadata-ingestion/src/datahub/ingestion/source/dynamodb/dynamodb.py
davidwalker-tp: Sensible idea. Will add in.

metadata-ingestion/tests/unit/dynamodb/test_dynamodb.py
davidwalker-tp: Good spot. Updated in [commit](https://github.com/datahub-project/datahub/pull/16155/commits/7e821b5b28b82a37fe0b8814cc41c0275a41b8a0).

metadata-ingestion/tests/unit/dynamodb/test_dynamodb.py
davidwalker-tp: An oversight on my part - thanks for the suggestion. Committed the change.

davidwalker-tp:
> Thank you for the contribution, I agree with the change but I had couple of comments. Please review them.

Thanks @skrydal for the feedback. I've made changes and accepted your suggestions, if you could take another look when you get chance.

PR #16156, opened 2026-02-10
feat(ingestion): update the navigation buttons on create/edit ingestion source

purnimagarg1:
**Linear ticket:**
https://linear.app/acryl-data/issue/CAT-1314/ingestion-create-edit-move-back-next-buttons
https://linear.app/acryl-data/issue/CAT-1351/combine-sync-schedule-step-with-connection-details-step

**Description:**

This PR removes the navigation buttons in create/edit ingestion flow and adds them at the top besides breadcrumbs.
It also removes the back button. Users can go back using the breadcrumbs or the browser back button.

It also combines the sync schedule step into connection details step.

**Screenshots:**

image

image

### DIFF
--- a/datahub-web-react/src/app/sharedV2/layouts/PageLayout.tsx
+++ b/datahub-web-react/src/app/sharedV2/layouts/PageLayout.tsx
@@ -3,7 +3,7 @@ import React from 'react';
 import styled from 'styled-components';
 
 const Card = styled.div`
-    background-color: #ffffff;
+    background-color: ${(props) => props.theme.colors.bg};
     display: flex;
     flex-direction: column;
     overflow: hidden;


datahub-web-react/src/app/ingestV2/source/multiStepBuilder/steps/step2ConnectionDetails/sections/recipeSection/recipeForm/RecipeForm.tsx
```diff
@@ -193,6 +194,7 @@ function RecipeForm({
                 <FiltersSection fields={filterFields} updateRecipe={updateRecipe} recipe={displayRecipe} />
 
                 <SettingsSection settingsFields={advancedFields} updateFormValue={updateFormValue} />
+                <ScheduleSection />
```
chriscollins3456: sadly we can't put this only in here because this is not part of the form itself

if a source does not have a form then they will not be able to add a schedule. so i think we will need to somehow have this below the yaml editor when they are either viewing yaml or when there is no form to view at all.

datahub-web-react/src/app/ingestV2/source/multiStepBuilder/IngestionSourceNavigationButtons.tsx
```diff
@@ -0,0 +1,113 @@
...
+            </Button>,
+        );
+
+        if (canGoToNext() && (isEditing || currentStepIndex !== 0)) {
+            const isDisabled = !isCurrentStepCompleted();
+            const nextButton = (
+                <Button key="next" size="sm" disabled={isDisabled} onClick={goToNext} data-testid="next-button">
+                    Next
+                </Button>
+            );
```
chriscollins3456: will we ever have a next button now that we're combining the sync schedule step in here?

purnimagarg1: if we decide to show it on select platform step in future. For now, removing this logic of adding next button from here.

datahub-web-react/src/app/ingestV2/source/multiStepBuilder/steps/step2ConnectionDetails/sections/recipeSection/recipeForm/RecipeForm.tsx
purnimagarg1: added it to both form and yaml tabs

datahub-web-react/src/app/sharedV2/layouts/PageLayout.tsx
```diff
@@ -76,14 +79,21 @@ export function PageLayout({
...
         <VerticalContainer>
             <HorizontalContainer>
                 {leftPanelContent && <SidePannel>{leftPanelContent}</SidePannel>}
 
                 <PageWrapper $hasBottomPanel={!!bottomPanelContent}>
-                    {topBreadcrumb && <BreadcrumbContainer>{topBreadcrumb}</BreadcrumbContainer>}
+                    {(topBreadcrumb || topRightContent) && (
+                        <TopContainer>
+                            {topBreadcrumb && <>{topBreadcrumb}</>}
+                            {topRightContent && <>{topRightContent}</>}
```
chriscollins3456: there would be a bug here if there are no breadcrumbs then the top right content would then appear on the left i believe due to the way the CSS is set up

purnimagarg1:
> overall looks good! i think we can fix a future bug in PageLayout before merging though.
> 
> also curious how does this look on smaller pages with the AI chat placeholder side content at regular width?

It would look something like this on a regular-sized screen. On smaller-sized screens, the layout would start getting impacted with the breadcrumbs shifting to 2 lines.

image

image

v-tarasevich-blitz-brain:
Added truncation of breadcrumbs
image

PR #16160, opened 2026-02-10
CAT-1335: tree shake phosphor and chunk

benjiaming:
### Summary: Phase 1 of FE performance optimizations

The DataHub frontend currently produces around **20MB** of minified JavaScript across 5 chunks, compressing to approximately **5 MB** when gzipped. This is a substantial bundle size that impacts initial load time and performance, particularly on slower networks. Phase 1 addresses some of the issues:
* ported chunking from SaaS to `vite.config.ts`
* added `antd` to chunks
* tree-shaking Phosphor icons (we were bundling all 4,539 icons despite using < 200)

### Results

Metric | Before | After | Improvement
-- | -- | -- | --
Total Bundle | 20.45 MB | 12.9 MB | -36.5% (-7.55 MB)
Gzipped | 4.89 MB | 3.28 MB | -32.9% (-1.61 MB)
Phosphor Icons | 6 MB (4,539 icons) | 372 KB (133 icons) | -95.7%
Page Load (3G) | ~27s | ~18s | -9 seconds

### Before
* giant monolithic bundle: `index-B3Suu7YI.js` = 19,395 KB (~19 MB)
* no chunking  (everything bundled together)
* all 4,539 Phosphor icons included (a large block)
* no vendor separation
Screenshot 2026-02-10 at 12 12 24 PM

### After
1. vendor-yJ7kOkOV.js: 8.6 MB (8,999.72 KB) [KEY CHUNK]
   - React + ReactDOM (bundled together)
   - Ant Design + rc-* components (bundled together)
   - Material UI + @emotion (bundled together)
   - Apollo Client + GraphQL
   - Other dependencies (lodash, moment, etc.)
   - This is the FIX: React + Ant Design + MUI together = no load order issues

2. source-CbhsHxku.js: 3.9 MB (4,042.65 KB)
   - Application source code
   - Custom components, pages, features

3. phosphor-vendor-D8fS90Jx.js: 372 KB (408.51 KB) [OPTIMIZED]
   - Custom Phosphor icons bundle (133 icons only)
   - Down from 6 MB originally (95.7% reduction)

4. index-CxYYPpoy.js: 4 KB (0.74 KB)
   - Entry point

Total Size:
   - Minified: 12.9 MB
   - Gzipped: 3.28 MB (what users actually download)

Screenshot 2026-02-12 at 11 28 05 AM

### Bundle Analyzer
* `rollup-plugin-visualize` installed as a dev dependency
* run `yarn build:analyze` and `open dist/stats.html`

### Risks
* certain Phosphor icons not loading
 → Mitigation: `npm run validate-bundle`
* chunking fails
 → Mitigation: added `datahub-web-react/src/__tests__/bundle-optimization.test.ts`
* merge issues (acryl-main does chunking already but a bit differently)

### Reference
https://www.notion.so/acryldata/DataHub-Frontend-Performance-Tech-Plan-246fc6a64277803e9976fdde565602d8


datahub-web-react/.eslintrc.js
```diff
@@ -26,6 +27,19 @@ module.exports = {
         'import/no-extraneous-dependencies': 'off',
         'import/no-relative-packages': 'error',
         'import/prefer-default-export': 'off', // TODO: remove this lint rule
+        '@typescript-eslint/no-restricted-imports': [
```
AdrianMachado, line 30: Excellent idea, I was about to suggest this while reading the PR

datahub-web-react/src/alchemy-components/components/Icon/constants.ts
```diff
@@ -547,7 +547,6 @@ export const MATERIAL_UI_ICONS = [
 ];
 
 export const PHOSPHOR_ICONS = [
-    'Activity',
```
AdrianMachado, line 550: This was unused?

datahub-web-react/src/alchemy-components/components/Icon/phosphor-icons.ts
```diff
@@ -0,0 +1,150 @@
+/**
```
AdrianMachado, line 1: I don't quite understand this change. 
Phosphor says it supports [tree-shaking out of the box](https://github.com/phosphor-icons/react). 

[Claude mentions that Vite should also should support tree shaking with named imports](https://claude.ai/share/ab7c0bc2-2cb4-453d-91e7-dcbc100f98d1). So I am confused why we aren't getting the benefits of that without having to create this file?

This leads me to believe that either our build tools are out of date, or are misconfigured. 

If we do get that working, we can simply rewrite icon imports to the `dics/csr` subdirectory?

datahub-web-react/src/__tests__/bundle-optimization.test.ts
```diff
@@ -0,0 +1,270 @@
...
+
+            expect(phosphorVendorFile).toBeDefined();
+            expect(vendorFile).toBeDefined();
+            expect(sourceFile).toBeDefined();
+
+            const phosphorSize = fs.statSync(path.join(distPath, phosphorVendorFile!)).size;
+            const vendorSize = fs.statSync(path.join(distPath, vendorFile!)).size;
+
+            // Phosphor should be optimized (< 500 KB)
+            expect(phosphorSize).toBeLessThan(500 * 1024);
```
AdrianMachado, line 263: Will this break if we add more icons?

datahub-web-react/src/__tests__/bundle-optimization.test.ts
```diff
@@ -0,0 +1,270 @@
...
+
+            // Vendor (React + Ant Design + MUI together) should be ~8-9 MB
+            expect(vendorSize).toBeGreaterThan(7 * 1024 * 1024);
+            expect(vendorSize).toBeLessThan(10 * 1024 * 1024);
```
AdrianMachado, line 267: This is a moving target as we update these libraries, no?

AdrianMachado (changes requested):
Take a look at my comment about the phosphor icons. Everything else seems to work based on your tests, but I think either @chriscollins3456 or @asikowitz should review it too. I'm not an expert on build tools 😅

datahub-web-react/scripts/validate-chunks.js
```diff
@@ -0,0 +1,147 @@
...
+}
+
+const files = fs.readdirSync(distPath);
+const jsFiles = files.filter((f) => f.endsWith('.js'));
+
+// Step 1: Verify required chunks exist
+const requiredChunks = [
+    'framework', // React core
+    'antd-vendor', // Ant Design
+    'mui-vendor', // Material UI
```
ani-malgari, line 35: these aren't the chunks based on the new chunking strategy in `vite.config`

datahub-web-react/scripts/validate-chunks.js
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env node
```
ani-malgari, line 1: where and when do we plan to run this script?

datahub-web-react/scripts/validate-icon-imports.js
```diff
@@ -0,0 +1,184 @@
...
+console.log(`  Checked ${files.length} files\n`);
+
+// Step 2: Verify all imported icons are exported
+console.log('=== Verifying Icon Exports ===\n');
+
+const phosphorIconsPath = path.join(__dirname, '..', 'src/alchemy-components/components/Icon/phosphor-icons.ts');
+const phosphorIconsContent = fs.readFileSync(phosphorIconsPath, 'utf8');
+
+// Extract exported icons
+const exportedIcons = new Set();
```
ani-malgari, line 71: Step 1 and 2 are redundant imo, and should be caught by ESLint and typescript.

ani-malgari: also it depends on how frequently we run this script

ani-malgari (commented):
The performance gains are already amazing. Everything looks good to me except I just left some minor comments with regards to the validation scripts.

datahub-web-react/src/alchemy-components/components/Icon/phosphor-icons.ts
ani-malgari: Interesting that phosphor-icons does support tree shaking, but clearly the bundle size has reduced significantly based on the PR description. agreed that it's worth investigating

AdrianMachado: Yeah, his approach will decrease it for sure. Like Claude basically reinvented tree shaking, but more restrictive.

AdrianMachado: But obviously the native implementation in our build tool is nicer because it doesn't require us to alias our imports

datahub-web-react/scripts/validate-icon-imports.js
benjiaming: I think we could turn this into a unit test.

datahub-web-react/src/alchemy-components/components/Icon/phosphor-icons.ts
benjiaming: Yes, Phosphor does support tree-shaking but only when:
**1. You import from individual CSR files:** 
As you said.
**2. You don't use wildcard imports (`import *`):**
We have 3 files that do `import * as phosphorIcons from '@phosphor-icons/react';`. Easy to fix.
**3. You don't use dynamic icon resolution:**
We actually have quite a lot of those.
```typescript
// Icon name comes from props/API at runtime
<Icon icon="BookmarkSimple" source="phosphor" />

// Resolved dynamically from imported namespace
const IconComponent = phosphorIcons[iconName]; // Can't be tree-shaken
```

If we want to eliminate dynamic icon resolution, we'd need to refactor ~100 files using the `Icon` pattern to something like this:
```typescript
// Old: 
<Icon icon="MagnifyingGlass" source="phosphor" />
<Icon icon="Search" source="material" /> // hypothetical, nobody actually uses source "material"

// New: 
import { MagnifyingGlass } from '@phosphor-icons/react'; // but what if the icon name comes from API?
import { Search } from '@mui/icons-material';

<Icon component={MagnifyingGlass} />
<Icon component={Search} />  
```

Not impossible but it would be a breaking change and quite a lot of effort.

The current solution is much simpler in that it creates a curated whitelist of icons and allows us to continue the same dynamic pattern. The disadvantage is that when we add an icon, we must update the list (but the new `eslint` rule and potentially unit test should catch that).

BTW `mui-icons` are also currently not tree shaken (~5 MB minified / 1.27 MB gzipped) but that's a follow-up PR.

datahub-web-react/src/__tests__/bundle-optimization.test.ts
benjiaming: yes - it's 372kb now. If we add many more, this would need to be updated.

datahub-web-react/src/alchemy-components/components/Icon/phosphor-icons.ts
```diff
@@ -0,0 +1,150 @@
...
+export { UserPlus } from '@phosphor-icons/react/dist/csr/UserPlus';
+export { Users } from '@phosphor-icons/react/dist/csr/Users';
+export { UsersThree } from '@phosphor-icons/react/dist/csr/UsersThree';
+export { Warning } from '@phosphor-icons/react/dist/csr/Warning';
+export { WarningCircle } from '@phosphor-icons/react/dist/csr/WarningCircle';
+export { Wrench } from '@phosphor-icons/react/dist/csr/Wrench';
+export { X } from '@phosphor-icons/react/dist/csr/X';
+export { XCircle } from '@phosphor-icons/react/dist/csr/XCircle';
+
+// Text formatting icons (for Editor toolbar)
```
asikowitz, line 141: I see no reason to distinguish between the two

datahub-web-react/src/alchemy-components/components/Icon/phosphor-icons.ts
asikowitz: It seems like what we should disallow is an import * from phosphor-icons, not all imports. Then we wouldn't have to include most of the import changes in this PR.

That being said, I'm not opposed to changing the Icon component entirely. I always thought it was weird we specified icon via string. I guess the idea was to limit what icons people could use via the component -- now you're limited to the ones provided by material + phosphor? But tbh that's an annoying restriction too -- we should be able to support custom icons as well. So I would be fine changing the 100+ files to pass the icon component in directly. AI should be able to make that change pretty easily.

benjiaming: Disallowing `import * from phosphor-icons` is not sufficient. The core issue is dynamic icon resolution, not wildcard imports and we would still be bundling all 4k icons with named imports because: 

1. Dynamic resolution requires all icons in scope:
```typescript
// Icon component uses dynamic lookup:
const IconComponent = phosphorIcons[iconName]; // iconName from props
```
2. Vite can't tree-shake dynamic property access - it must keep all icons since it doesn't know which `iconName` values will be used at runtime.


We have two options:
1. Curated whitelist: Current PR. Ships faster with lower risk, no API changes needed but we would need to maintain the list (with automation).
2. Native tree-shaking: Cleaner but would need to refactor the following:
* ~85 simple static patterns and ~20 conditional patterns e.g. `<Icon icon={linkIcon ?? 'ArrowRight'} .../>` 
*  1 mapped Alchemy `Icon` pattern  
```typescript 
export const ASSERTION_TYPE_TO_ICON_MAP: Record<AssertionType, JSX.Element> = {
    [AssertionType.Freshness]: <Clock size={20} />,  // Direct Phosphor component
    [AssertionType.Volume]: <Database size={20} />,
};
```

* 13 maps `MODULE_TYPE_TO_ICON` and `DATA_TYPE_ICON_MAP`

Not impossible and AI will certainly help but it does add complexity and risk to the PR.

Both approaches should work and would deliver the same savings.

What is your preference?

AdrianMachado: I think I'm with Andrew in saying option 2 may make this PR a bit bloated, but long term its less annoying.

benjiaming:
Closed, this one got merged instead: https://github.com/datahub-project/datahub/pull/16338

PR #16165, opened 2026-02-11
feat(ingestion): add configurable report sample sizes and failure logging

rob-1019:
## Summary

Adds configurable sample sizes for failure and warning entries in ingestion reports via environment variables.

This simplified implementation follows @sgomezvillamor's feedback to initialize directly from environment variables at construction time, avoiding config classes or post-init methods that callers need to remember to invoke.

## Environment Variables

| Variable | Default | Description |
|----------|---------|-------------|
| `DATAHUB_REPORT_FAILURE_SAMPLE_SIZE` | `10` | Maximum number of failure entries to include in the report |
| `DATAHUB_REPORT_WARNING_SAMPLE_SIZE` | `10` | Maximum number of warning entries to include in the report |

## Usage

Set environment variables before running ingestion to capture more failure/warning samples when debugging:

```bash
export DATAHUB_REPORT_FAILURE_SAMPLE_SIZE=100
export DATAHUB_REPORT_WARNING_SAMPLE_SIZE=50
datahub ingest -c recipe.yaml
```

## Changes

- `env_vars.py`: Added `get_report_failure_sample_size()` and `get_report_warning_sample_size()`
- `sink.py`: `SinkReport` initializes `LossyList` sizes from env vars via `default_factory` lambda
- `source.py`: `StructuredLogs` initializes `LossyDict` sizes from env vars via `default_factory` lambda
- `docs/cli.md`: Documented new environment variables

## Test plan

- [x] 9 unit tests covering default values, custom values, and LossyList behavior
- [x] `ruff check` passes
- [x] `mypy` passes


sgomezvillamor:
if the use case is to get a list of failed mcps, have you checked this?
https://github.com/datahub-project/datahub/blob/b96146d16e4c31054cda08e445e7258a29eb9fb9/metadata-ingestion/src/datahub/ingestion/run/pipeline_config.py#L34-L40

metadata-ingestion/src/datahub/ingestion/api/sink.py
```diff
@@ -20,12 +20,41 @@
...
+
+    def __post_init__(self) -> None:
+        self._init_lossy_lists()
+
+    def _init_lossy_lists(self) -> None:
+        """Initialize or reinitialize the LossyList fields based on current config."""
+        self.warnings = LossyList(max_elements=self.warning_sample_size)
+        self.failures = LossyList(max_elements=self.failure_sample_size)
+
+    def configure_report(
```
sgomezvillamor: So far, we don’t have any cases where Report classes are instantiated from Config classes.

The solution proposed here works by introducing a kind of post-init method that the caller has to remember to invoke, whereas this feels like something that should be handled automatically in the constructor.

My proposal would be to initialize the warnings/failures “loose” lists directly from the environment variables, and skip initialization from Config for now, until there’s a clearer or stronger need for it.

rob-1019: Sounds reasonable - will do.

rajatoss:
## Connector Tests Results

**Connector tests failed** for commit `13ed477`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/23120365553)

metadata-ingestion/tests/unit/api/test_report_sample_size.py
sgomezvillamor, line 1: alternatively, instead of patching the get env methods, you can act directly on the env vars

eg

https://github.com/datahub-project/datahub/blob/e068f4e704bc6ec9ff8908ea0b3c83fea2383065/metadata-ingestion/tests/unit/test_unity_catalog_proxy.py#L1621-L1627

metadata-ingestion/tests/unit/api/test_report_sample_size.py
```diff
@@ -0,0 +1,124 @@
...
+        """Test SinkReport reads warning sample size from env var."""
+        with mock.patch(
+            "datahub.ingestion.api.sink.get_report_warning_sample_size",
+            return_value=25,
+        ):
+            report = SinkReport()
+            assert report.failures.max_elements == 10  # unchanged
+            assert report.warnings.max_elements == 25
+
+    def test_sink_report_both_custom_sample_sizes(self) -> None:
```
sgomezvillamor: previous two test validate that we can set them independently and then altogether; previous two test seem redundant with this one

metadata-ingestion/tests/unit/api/test_report_sample_size.py
```diff
@@ -0,0 +1,124 @@
...
+        with mock.patch(
+            "datahub.ingestion.api.source.get_report_warning_sample_size",
+            return_value=25,
+        ):
+            logs = StructuredLogs()
+
+            assert logs._entries[StructuredLogLevel.ERROR].max_elements == 10
+            assert logs._entries[StructuredLogLevel.WARN].max_elements == 25
+
+    def test_source_report_inherits_sample_sizes(self) -> None:
```
sgomezvillamor: previous two test validate that we can set them independently and then altogether; previous two test seem redundant with this one

metadata-ingestion/tests/unit/api/test_report_sample_size.py
```diff
@@ -0,0 +1,124 @@
...
+            assert (
+                report._structured_logs._entries[StructuredLogLevel.ERROR].max_elements
+                == 100
+            )
+            assert (
+                report._structured_logs._entries[StructuredLogLevel.WARN].max_elements
+                == 75
+            )
+
+    def test_lossy_list_respects_max_elements(self) -> None:
```
sgomezvillamor: this is a test that should exist in `LossyList` unit tests cases, and I assume there exists one already

PR #16167, opened 2026-02-11
fix(build): setuptools constraint to fix setup issue

anshbansal:


sgomezvillamor:
Aren't those inheriting the upper bound restriction from `-e ../metadata-ingestion` ?

What about `datahub-actions`? We don't have upper bound limit there, just `acryl-datahub` dep.

PR #16170, opened 2026-02-11
feat(glue): enhance partition key metadata extraction

kartikey-visa:
- Add comprehensive partition key information to dataset properties
- Support multiple partition keys with proper metadata extraction
- Add partition key names, types, and comments to custom properties
- Enhance backward compatibility for single partition key scenarios

Fixes: Missing partition key label in Glue table schema
image

Added partition key details in dataset properties
image

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
@@ -473,9 +473,7 @@ class GlueSource(StatefulIngestionSourceBase):
 
     def get_all_lf_tags(self) -> List:
         # 1. Get all LF-Tags in your account (metadata only)
-        response = self.lf_client.list_lf_tags(
-            MaxResults=50  # Adjust as needed
-        )
+        response = self.lf_client.list_lf_tags(MaxResults=50)  # Adjust as needed
         all_lf_tags = response["LFTags"]
         # Continue pagination if necessary
         while "NextToken" in response:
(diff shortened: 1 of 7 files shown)


kartikey-visa:
@chriscollins3456 
Can you please review this

gabe-lyons:
thanks for the contribution!

gabe-lyons:
@sgomezvillamor could you take a look?

metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
sgomezvillamor, line 1: The code currently processes `PartitionKeys` in three different places. Since we’re aiming to improve partition key metadata extraction, this could be a good opportunity to refactor that logic into a dedicated helper method.

```
  @dataclass
  class PartitionKeyInfo:
      """Structured partition key information from AWS Glue."""
      name: str
      type: str
      comment: str

  def _extract_partition_key_info(self, table: Dict) -> List[PartitionKeyInfo]:
```

That would also help to unit test different unexpected scenarios

metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
```diff
@@ -1510,6 +1522,20 @@ def _get_dataset_properties(self, table: Dict) -> DatasetPropertiesClass:
...
+            partition_key_names = [pk["Name"] for pk in partition_keys]
+            partition_key_types = [pk.get("Type", "unknown") for pk in partition_keys]
+            partition_key_comments = [pk.get("Comment", "") for pk in partition_keys]
+
+            custom_properties["partition_keys"] = ",".join(partition_key_names)
+            custom_properties["partition_key_types"] = ",".join(partition_key_types)
+            custom_properties["partition_key_comments"] = ",".join(
+                partition_key_comments
+            )
+            custom_properties["num_partition_keys"] = str(len(partition_keys))
```
sgomezvillamor, line 1537: I see three options for including partition info in the CustomProperties:

1. Use the raw format from glue, eg 

```
custom_properties["partition_keys_metadata"] = table.get("PartitionKeys", [])
````

2. Use a custom json format instead, eg

```
  custom_properties["partition_keys_metadata"] = json.dumps([
      {"name": "year", "type": "string", "comment": "Year"},
      {"name": "month", "type": "string", "comment": "Month"},
      {"name": "day", "type": "string", "comment": "Day"}
  ])
```

3. (your proposal) Split the info across different custom properties keys

1 is good for those familiar with the Glue format. And 2 is good for programatic processing. Not much sure of the use case for 3.

WDYT?

----

This case in particular is problematic because the chance of a `,` in the description is high

```
            custom_properties["num_partition_keys"] = str(len(partition_keys))
```

sgomezvillamor (commented):
Thanks for the contrib!

I left a couple of comments

sgomezvillamor:
@kartikey-visa — just checking, are you planning to pick this back up?

PR #16178, opened 2026-02-11
fix(redirect): fix open redirect filter

david-leifker:
**Fix open redirect in BasePathRedirectFilter**

`BasePathRedirectFilter` runs before routing and was building redirect URLs from the raw request path. Paths like `////google.com/` produced a scheme-relative `//google.com` Location, so the browser redirected off-site.

Redirect URLs are now built from a normalized path: leading slashes are stripped so `//google.com` becomes a same-origin path `/google.com` instead of a scheme-relative URL. Matches the sanitization already used in `Application.redirectTrailingSlash`, which never ran for these requests because the filter handles them first.

### DIFF
--- a/datahub-frontend/app/controllers/Application.java
+++ b/datahub-frontend/app/controllers/Application.java
@@ -136,26 +136,6 @@ public class Application extends Controller {
   }
 
   /**
-   * Moves permanently the get into version without trailing slash
-   *
-   * @param path String
-   * @return Result
-   */
-  @Nonnull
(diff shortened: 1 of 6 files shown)


mathiasror:
I have not been able to trigger the redirect on this branch. :)

Since the redirect is handled in this filter, perhaps `Application.redirectTrailingSlash` and its route in `/conf/routes` could be removed entirely or optionally the other way around (remove filter, keep route)?

craig-rueda (approved):
LGTM

PR #16183, opened 2026-02-11
fix(clickhouse): include dictGet() table references in SQL lineage

kyungsoo-datahub:
ClickHouse `dictGet(dict_name, attr, key)` functions reference dictionaries backed by tables. These were missing from upstream lineage because sqlglot parses the first argument as a `Column` or `Literal` node, not a `Table` node — so standard table extraction never finds them.                                                                                    
                                                                                                                
- Add `_clickhouse_extract_dictget_tables()` to extract dictionary references from dictGet-family function calls and include them in `in_tables`
- Remove dead `_is_in_clickhouse_dict_function()` filter (was a no-op, it looked for Table nodes that never existed in dictGet args)


metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
```diff
@@ -412,31 +410,59 @@ def _extract_table_names(
...
+) -> OrderedSet[_TableName]:
+    """Extract dictionary references from ClickHouse dictGet() function calls.
+
+    sqlglot parses dictGet(dict_name, ...) first arg as a Column node, not a
+    Table node. This extracts those references as _TableName for upstream lineage.
 
-    ClickHouse dictionary functions like DICTGET(dict_name, attr_name, key) take a
-    dictionary name (e.g., 'default.subscriptions') as the first argument. sqlglot
-    parses this as a Table node, but it's not a real table reference for lineage.
+    TODO: CLL is still wrong — dictionary name appears as a column on the FROM
```
sgomezvillamor: Let me check if I understand this TODO correctly.

With an expression like `DICTGET(dict_name, attr_name, key)`, we currently produce table-level lineage to `dict_name`, and column-level lineage (CCL) also points to `dict_name`. The table-level lineage is correct, but the column-level lineage does not. Is that what the TODO is referring to?

If so, should the column-level lineage instead point to `attr_name` to be accurate?

kyungsoo-datahub: Yes, your understanding is correct.

Current behavior: CLL points to dict_name (treated incorrectly as a column
name)
Correct behavior: CLL should point to dict_name.attr_name (table.column)

For dictGet(subscriptions, 'type', subscription_id):
- Table lineage: subscriptions (working after the fix)
- Column lineage should be: subscriptions.type
- Currently shows: subscriptions (as a column, which is wrong)

The fix needs to parse the second argument (attr_name) from dictGet() calls and create CLL to dict_name.attr_name instead of just dict_name.

ToDo updated. Thanks.

sgomezvillamor: Thanks for reply and udpate the TODO

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
```diff
@@ -412,31 +410,60 @@ def _extract_table_names(
...
+        ):
+            continue
+
+        first_arg = func.expressions[0]
+        if isinstance(first_arg, sqlglot.exp.Column):
+            # dictGet(schema.dict_name, ...) → Column(table='schema', this='dict_name')
+            # dictGet(db.schema.dict_name, ...) → Column(catalog='db', table='schema', ...)
+            table_name = first_arg.name
+            schema = first_arg.table if first_arg.table else None
+            database = first_arg.args.get("catalog")
```
sgomezvillamor: what's the expected type of `database`? is it safe to access `database.name`?

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
```diff
@@ -412,31 +410,60 @@ def _extract_table_names(
...
+                result.add(
+                    _TableName(database=None, db_schema=parts[0], table=parts[1])
+                )
+            elif len(parts) == 1:
+                result.add(_TableName(database=None, db_schema=None, table=parts[0]))
+            else:
+                logger.warning(
+                    f"Unexpected dictGet literal with {len(parts)} parts: {first_arg.this}"
+                )
+
```
sgomezvillamor, line 467:
```
if Column
elif Literal
```

we may add an `else` case to log a warning or debug message

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
```diff
@@ -412,31 +410,60 @@ def _extract_table_names(
...
+            table_name = first_arg.name
+            schema = first_arg.table if first_arg.table else None
+            database = first_arg.args.get("catalog")
+            db_name = database.name if database else None
+            result.add(_TableName(database=db_name, db_schema=schema, table=table_name))
+        elif isinstance(first_arg, sqlglot.exp.Literal) and first_arg.is_string:
+            # dictGet('schema.dict_name', ...) → Literal('schema.dict_name')
+            parts = first_arg.this.split(".")
+            parts = [p for p in parts if p]  # guard against empty parts
+            if len(parts) == 3:
```
sgomezvillamor: I haven't seen any test case with 3 parts.
Is that a valid scenario in Clickhouse?

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
```diff
@@ -412,31 +410,60 @@ def _extract_table_names(
...
+                    _TableName(database=None, db_schema=parts[0], table=parts[1])
+                )
+            elif len(parts) == 1:
+                result.add(_TableName(database=None, db_schema=None, table=parts[0]))
+            else:
+                logger.warning(
+                    f"Unexpected dictGet literal with {len(parts)} parts: {first_arg.this}"
+                )
+
+    return result
```
sgomezvillamor, line 470: For debugging and troubleshooting, I would log.debug the result.

metadata-ingestion/tests/unit/sql_parsing/test_sqlglot_lineage.py
sgomezvillamor, line 1: I miss some test for:
* `db.schema.dict` format, if that's a valid case
* `dict` (len(parts==1)

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
kyungsoo-datahub: Found that Click House doesn't support 3 and 4 levels. Remove this part. Thanks.

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
kyungsoo-datahub: Thank you for pointing this. I re-investigate if they have 3 or 4 levels. Clickhouse uses only 2 levels. Will remove 3, 4 levels part.

Reference: https://clickhouse.com/docs/en/sql-reference/statements/create/dictionary 
https://clickhouse.com/docs/en/sql-reference/statements/create/table

metadata-ingestion/tests/unit/sql_parsing/test_sqlglot_lineage.py
kyungsoo-datahub: This ```db.schema.dict``` case is invalid as I found that Click House doesn't support 3 and 4 levels. The test case for 1 part got added. Thanks.

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
kyungsoo-datahub: Added.

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
kyungsoo-datahub: Added.

PR #16202, opened 2026-02-13
fix(ci): remove labeled trigger

chakru-r:
Removing the labeled trigger for docker-unified.yml since it hasnt been used for a while now. This was used occasionally to make PRs from repo forks run on depot runners, but some recent cypress fixes seems to have removed this need. PRs from forks can still be run on depot via workflow_dispatch or pushes after applying depot label. The label trigger was causing some false "skipped" for the smoke tests even though the workflow was triggered twice -- once on push and once on label

### DIFF
--- a/.github/workflows/docker-unified.yml
+++ b/.github/workflows/docker-unified.yml
@@ -22,9 +22,7 @@ concurrency:
   # Using `github.run_id` (unique val) instead of `github.ref` here
   # because we don't want to cancel this workflow on master only for PRs
   #   as that makes reproducing issues easier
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-unified.yml
```diff
@@ -12,7 +12,7 @@ on:
       - master
       - releases/**
   pull_request:
-    types: [opened, synchronize, reopened, labeled]
```
devashish2203, line 15: Should we also clean up the other places 'labeled' is used and remove those conditions for that build path?

devashish2203 (approved):

PR #14715, opened 2025-09-09
fix(ingest/teradata):  Fix Teradata lineage URN generation to prevent extra database prefixes

treff7es:
Problem Fixed

  1. Teradata Table Lineage URN Issue

  Teradata table lineage was incorrectly showing dbc.database.table instead of database.table in URNs. This happened because the SQL parsing framework was adding an unwanted "dbc" database prefix to
  qualified table names.

  2. Teradata Workflow Architecture Issues

  - Audit log processing was happening after schema extraction, causing lineage data to bypass proper work unit processing chains
  - Method name inconsistencies in lineage processing
  - Missing incremental lineage integration

PR #16208, opened 2026-02-13
fix(ingest/teradata): set DATABASE context for view HELP commands

JohnRTurner:
## Summary

Teradata's `SHOW VIEW` (used via SQLAlchemy's `get_view_definition`) requires the current database context to be set when retrieving view definitions across schemas. Without it, HELP commands fail for views outside the default connection schema, resulting in empty view definitions and no lineage being extracted.

**Fix:** Add `DATABASE "{schema}"` context setting before processing views in each schema, allowing HELP commands to resolve view definitions correctly without requiring a database in the connection string.

**Background on the double-prefix bug:** The `wl_w.wl_w.table` double-prefix issue was already fixed in #14715 (changing `default_db=default_database` to `default_db=None` for observed DBQL queries). `TwoTierSQLAlchemySource.get_db_schema` has always returned `(None, schema)` correctly for 2-tier identifiers, so no view-path override was needed.

## Test Plan

- [x] Linting passed with `./gradlew :metadata-ingestion:lintFix`
- [x] Existing Teradata tests continue to pass
- [x] Manually tested against live Teradata Vantage instance with views across multiple schemas


metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -943,6 +943,72 @@ def get_db_name(self, inspector: Inspector) -> str:
         else:
             raise Exception("Unable to get database name from Sqlalchemy inspector")
 
+    def get_db_schema(self, dataset_identifier: str) -> Tuple[Optional[str], str]:
```
treff7es: Is this needed? Isn't the `TwoTierSQLAlchemySource.get_db_schema` already doing this?

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -943,6 +943,72 @@ def get_db_name(self, inspector: Inspector) -> str:
...
+            return None, ""
+        else:
+            # Shouldn't happen, but log a warning and fall back to parent
+            logger.warning(
+                f"Unexpected dataset_identifier format: {dataset_identifier} "
+                f"(expected 2 parts for Teradata, got {len(parts)})"
+            )
+            return super().get_db_schema(dataset_identifier)
+
+    def get_view_default_db_schema(
```
treff7es: Is this needed? Isn't the `TwoTierSQLAlchemySource.get_db_schema` already doing this?

PR #16218, opened 2026-02-15
feat(ingestion): add Apache Flink connector for metadata and lineage …

dinesh-verma-datahub:
Adds a new Flink connector that extracts metadata from Apache Flink clusters via the JobManager REST API, with optional SQL Gateway for multi-platform lineage resolution.

Features:

- **Job metadata**: DataFlow + DataJob entities from Flink execution plans
- **Lineage**: multi-platform lineage (Kafka, Postgres, Iceberg, JDBC, and more) via SQL Gateway catalog resolution; DataStream Kafka lineage works without SQL Gateway
- **Dataset materialization**: source/sink datasets discovered via lineage are emitted as Dataset entities
- **Run history**: DataProcessInstance entities for job execution tracking
- **Operator chaining**: handles merged source+sink plan nodes (verified on Flink 1.19)
- **Error isolation**: failures at any level don't block other entities
- **Stateful ingestion**: stale entity removal support
- **Platform registration**: added to integrations catalog (UI wizard) and Python plugin entrypoints


metadata-ingestion/tests/unit/flink/test_flink_config.py
```diff
@@ -0,0 +1,130 @@
...
+def test_minimal_config():
+    config = FlinkSourceConfig.model_validate(
+        {"rest_endpoint": "http://localhost:8081"}
+    )
+    assert config.rest_endpoint == "http://localhost:8081"
+    assert config.env == "PROD"
+    assert config.platform_instance is None
+    assert config.include_lineage is True
+    assert config.include_run_history is True
+    assert config.max_workers == 10
```
skrydal: Do we need to test defaults?

metadata-ingestion/tests/unit/flink/test_flink_config.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def test_trailing_slash_stripped():
+    config = FlinkSourceConfig.model_validate(
+        {"rest_endpoint": "http://localhost:8081/"}
+    )
+    assert config.rest_endpoint == "http://localhost:8081"
```
skrydal: What changes would need to happen to our business logic for such assertion to fail?

metadata-ingestion/tests/integration/flink/__init__.py
skrydal, line 1: I am missing content of this directory

rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `8a9a96e`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/23450448971)

metadata-ingestion/tests/unit/flink/test_flink_config.py
dinesh-verma-datahub: not needed, removed it.

metadata-ingestion/tests/unit/flink/test_flink_config.py
dinesh-verma-datahub: was added due to boilerplate code from LLMs and forgot to set the PR in draft mode, i have cleaned up these.

metadata-ingestion/tests/integration/flink/__init__.py
dinesh-verma-datahub: it was not added at that time, now i have added it.

skrydal:
Thank you for sharing this contribution. However, I have concerns about the fundamental modeling approach taken in this PR.

The connector uses two complementary APIs to retrieve metadata from Flink — the JobManager REST API and the SQL Gateway. In practice these are tightly coupled (the test setup itself uses the SQL Gateway to define jobs). Unfortunately, the connector treats them completely independently, generating entities that are never connected to each other.

This is already exemplified in the integration test, though the problem is obscured by the test setup. Kafka-backed tables are created in an ephemeral (in-memory) SQL client session, while Postgres-backed tables are registered in a persistent (file-based) SQL Gateway session. When the connector opens its own SQL Gateway session, it only sees the Postgres-backed tables — not the Kafka ones. Meanwhile, the Kafka topics are only visible through the JobManager as nodes of the running job.

If the Kafka-backed tables were also created in a persistent SQL Gateway session (as they would be in a real deployment), the connector would emit:
  - `urn:li:dataset:(urn:li:dataPlatform:flink,default_catalog.default_database.orders,PROD)` — from catalog extraction via SQL Gateway
  - `urn:li:dataset:(urn:li:dataPlatform:kafka,orders,PROD)` — from lineage extraction via JobManager

These represent the same physical table but appear as two completely disconnected entities in DataHub.

I would lean towards treating the JobManager as the primary source of truth and using the SQL Gateway to enrich what we learn from it — specifically, to resolve what platform an entity actually belongs to. In the generic case (`TableSourceScan(table=[[catalog, db, table]])`), it is impossible to determine the underlying platform without cross-referencing the catalog definition from the SQL Gateway. The current approach works for Kafka only because Flink happens to use "Kafka" in its operator naming convention (`KafkaSource-{topic}`), but this is not extensible beyond Kafka.

dinesh-verma-datahub:
@skrydal  Thanks for the thorough review.

1) Regarding modeling fix, should we take this approach:
    - Agreed with you - 'JobManager as primary source of truth, SQL Gateway as enrichment for platform resolution'
    - Will wire SQL Gateway INTO the lineage pipeline to resolve the actual connector type 
        (kafka, jdbc, hive, etc.) and emit a single correctly-platformed Dataset URN
    - If No SQL Gateway configured → skip lineage for SQL/Table API jobs entirely, 
        (DataFlow + DataJob + DPI remain unaffected — always emitted from JobManager regardless)

2) Specific fixes:
    - Remove hardcoded platform="kafka" for TableSourceScan/Sink(table=...) nodes
    - Include catalog + database in dataset naming (currently discarded — causes collisions)
    - Defer DataStream API lineage (vertex name regex is convention-dependent, not reliable, will check DataStream support in next iteration)
    - Will also fix integration tests to use persistent catalog sessions for
      both Kafka and non-Kafka tables so the dual-entity problem is visible and verifiable

Does this look okay for the v1(MVP) scope?

skrydal:
@dinesh-verma-datahub I appreciate you considering my comment.

**Ad 1.**
I am a bit puzzled by the 3rd bullet point. It seems to be in conflict with itself - first it says the lineage will be skipped, but then, only then to say dataflows/jobs are always emitted. Do you mean emitting them without input/output references?


**Ad 2.**
Let's reconsider the approach here. Your connector bases on parsing `description` field and currently supports cases where it is like `KafkaSource-{topic}` (Kafka source defined via DataStream API) or `TableSourceScan(table=[[cat, db, tbl]])` (Source defined through the SQL Gateway catalog mapping) - and similarly for sinks. From your description I understand you want to remove support for DataStream case, for the simplicity of the initial approach. In general that is fine but we need to consider what data we have available from Flink's REST API.

I have done some research and it seems that `description` field is the only thing we can base on. It implies we need to continue using regular expressions. To get some understanding what I compiled (via LLM) the table below:

  | Connector | API | Source `description` Pattern | Sink `description` Pattern |
  |---|---|---|---|
  | **Kafka** | DataStream | `KafkaSource-{topic} -> ...` | `KafkaSink-{topic} -> ...` |
  | **Kafka** | SQL/Table | `TableSourceScan(table=[[cat, db, tbl]], fields=[...])` | `Sink(table=[[cat, db, tbl]])` or `tbl[N]: Writer` |
  | **JDBC** | SQL/Table | `TableSourceScan(table=[[cat, db, tbl]], fields=[...])` | `Sink(table=[[cat, db, tbl]])` or `tbl[N]: Writer` |
  | **Filesystem** | SQL/Table | `TableSourceScan(table=[[cat, db, tbl]], fields=[...])` | `Sink(table=[[cat, db, tbl]])` or `tbl[N]: Writer` |
  | **Filesystem** | DataStream | `Source: {user-provided-name}` | `Sink: {user-provided-name}` |
  | **Iceberg** | SQL/Table | `TableSourceScan(table=[[cat, db, tbl]], fields=[...])` | `Sink(table=[[cat, db, tbl]])` or `tbl[N]: Writer` |
  | **Hudi** | SQL/Table | `TableSourceScan(table=[[cat, db, tbl]], fields=[...])` | `Sink(table=[[cat, db, tbl]])` or `tbl[N]: Writer` |
  | **Paimon** | SQL/Table | `TableSourceScan(table=[[cat, db, tbl]], fields=[...])` | `Sink(table=[[cat, db, tbl]])` or `tbl[N]: Writer` |
  | **HBase** | SQL/Table | `TableSourceScan(table=[[cat, db, tbl]], fields=[...])` | `Sink(table=[[cat, db, tbl]])` |
  | **MongoDB** | SQL/Table | `TableSourceScan(table=[[cat, db, tbl]], fields=[...])` | `Sink(table=[[cat, db, tbl]])` |
  | **Elasticsearch** | SQL/Table | N/A | `Sink(table=[[cat, db, tbl]])` |
  | **OpenSearch** | SQL/Table | N/A | `Sink(table=[[cat, db, tbl]])` |
  | **Doris** | SQL/Table | `TableSourceScan(table=[[cat, db, tbl]], fields=[...])` | `Sink(table=[[cat, db, tbl]])` |
  | **StarRocks** | SQL/Table | `TableSourceScan(table=[[cat, db, tbl]], fields=[...])` | `Sink(table=[[cat, db, tbl]])` |
  | **ClickHouse** | SQL/Table | `TableSourceScan(table=[[cat, db, tbl]], fields=[...])` | `Sink(table=[[cat, db, tbl]])` |
  | **DynamoDB** | SQL/Table | N/A | `Sink(table=[[cat, db, tbl]])` |
  | **Kinesis** | DataStream | `Source: {user-provided-name}` | `Sink: {user-provided-name}` |
  | **Kinesis** | SQL/Table | `TableSourceScan(table=[[cat, db, tbl]], fields=[...])` | `Sink(table=[[cat, db, tbl]])` |
  | **Pulsar** | DataStream | `Source: {user-provided-name}` | N/A |
  | **RabbitMQ** | DataStream | `Source: {user-provided-name}` | `Sink: {user-provided-name}` |
  | **Cassandra** | DataStream | `Source: {user-provided-name}` | `Sink: {user-provided-name}` |
  | **Google PubSub** | DataStream | `Source: {user-provided-name}` | `Sink: {user-provided-name}` |
  | **Flink CDC** (MySQL, PG, Oracle, etc.) | SQL/Table | `TableSourceScan(table=[[cat, db, tbl]], fields=[...])` | N/A (source-only) |
  | **DataGen** | SQL/Table | `TableSourceScan(table=[[cat, db, tbl]], fields=[...])` | N/A |
  | **Print** | SQL/Table | N/A | `Sink(table=[[cat, db, tbl]])` |
  | **Blackhole** | SQL/Table | N/A | `Sink(table=[[cat, db, tbl]])` |

Assuming we can trust this structure to be somehow stable, we really have 2 cases - SQL/Table and DataStream. SQL/Table seems to be substantially easier for us, because there is always information about the Flink catalog (`cat`) used. Using this value we can query catalog details. SQL Gateway is useful exactly here, however catalogs can be setup via other means too (programmatically and statically via a yaml file).

Our goal here is to build a proper URN of the input/output. For this we need:
* platform - this can be extracted, for most cases, quite easily
* platform_instance - this is problematic. We need a mapping mechanism for it, using information extracted from the catalog. Also static platform -> platform_instance mapping, can be a rudimentary default (you have it already implemented via `DatasetLineageProviderConfigBase` inheritance). For TableSourceScan/Sink nodes, we can have fine grained mapping based on catalog name -> platform_instance.
* entity name - while in TableSourceScan/Sink you have db and table name, which, combined with platform_instance, can be used to reliably build the name, **please note that it appears that `tbl` field will look like `schema.table_name` for 3-level databases (like Postgres). Please include such case.**


**My suggestion** is to have the extractor focus on SQL/Table and DataStream-Kafka-only cases. In the later it can assign `kafka` platform + platform_instance based on platform mapping (so simple platform_instance for Kafka). Then for SQL/Tables we try to extract catalog information from SQL Gateway. If it is there then we should have the platform. Basing on platform (from the catalog), catalog name and two mappings from the config (platform -> platform_instance as a fallback, catalog -> platform_instance as the primary) we should be able to build the URNs. This probably will not cover cases of catalog being defined outside of a permanent SQL Gateway session - please document that. In future we can consider allowing users to provide us catalog information via other means.
If we can't figure out the catalog (or at least a platform), we should raise a warning. Also we might consider adding this to the report. I guess ingesting just the job without input/output is fine.

In integration tests, please make sure we have case of >2 entities lineage (e.g. topic1 -> flink -> topic2 -> flink -> topic3), as well as SQL/Table sources with 2 and 3 levels in naming (example of 2 level naming is Iceberg, 3 level naming example could be Postgres - both can be created locally).

dinesh-verma-datahub:
@skrydal Clarification on - "if No SQL Gateway configured → skip lineage for SQL/Table API jobs entirely,
(DataFlow + DataJob + DPI remain unaffected — always emitted from JobManager regardless)":

Yes, i meant emitting them without input/output references.

Thanks for the detailed inputs, let me work on these.

askumar27:
Interest from OSS - https://datahubspace.slack.com/archives/CUMUWQU66/p1773751968368269

skrydal:
@dinesh-verma-datahub

I appreciate your efforts to improve the PR, especially in terms of tests robustness. I have just 3 comments, let's address them and we can merge the PR.

**Regarding integration tests**
Thank you for improving integration tests, `test_iceberg_flink_lineage_stitching` is spot-on as the proper integration test for this source, as the core functionality is indeed to connect to other sources.

However, currently we are not seeing the output dataset for that flow as `matched_events` is a temporary table instead of a proper one - could you adjust it so we can see it `iceberg -- flink --> kafka` lineage in that case (or alternatively - some other test case which would cover it).
 
I would also **suggest** to extend the test to confirm that, for example, `schemaMetadata` and similar aspects is coming indeed only from iceberg source and not `flink`. Also I would ensure that `flink` ingests only those datasets which are matched to the actual source (`iceberg`) ingestion.

What I am **missing** in those integration tests is stitching for `kafka` and `postgres` cases (they are there, but without stitching like for the iceberg).

Moreover - the tests should include `platform_instance` (and possibly `platform` as well) mapping cases.


------
**As for documentation**
I like how documentation explains complexities in a straight-forward way. Could you please add information about limitations when it comes to temporary tables as well (very minor change)?


----
There is a failing CI/CD case for UI part of your changes:
```
$ tsc --noEmit
src/app/ingest/source/conf/flink/flink.ts(1,30): error TS2307: Cannot find module '@app/ingest/source/conf/types' or its corresponding type declarations.
src/app/ingestV2/source/conf/flink/flink.ts(1,30): error TS2307: Cannot find module '@app/ingestV2/source/conf/types' or its corresponding type declarations.
```
It needs to be addressed before we merge.

dinesh-verma-datahub:
@skrydal Thank you for the review and inputs. I have addressed the comments, summary:

- **Docs**: added temporary table limitation (session-scoped, invisible to SQL Gateway, unresolvable).
- **TypeScript CI Failure**: removed invalid import
- **Iceberg stitching test**: uses persistent Iceberg tables for both input and output, giving full `iceberg → flink → iceberg` lineage with a resolved output dataset.
- **Iceberg → Flink → Kafka**: not achievable in SQL/Table API — Kafka Sink V2 operator chaining exposes no catalog path, so the sink is unresolvable. Added a test for the reverse (`kafka → flink → iceberg`), which is the more common lakehouse pattern and fully resolvable on both sides.
- **Kafka and Postgres stitching tests**: added, mirroring the same structure as the Iceberg test.
- **All stitching tests** now assert: 
          (1) both connectors emit matching URNs,
          (2) Flink does not emit schema metadata for datasets owned by the authoritative source,
          (3) Flink references no datasets the authoritative source did not also emit.
- **Platform instance mapping**: covered in all stitching tests — no instance (baseline), platform-level map, catalog-level override, and global Flink instance (affects job URNs only, not dataset URNs).

metadata-ingestion/tests/integration/flink/setup_flink_test_data.py
```diff
@@ -65,6 +65,19 @@
     event_time TIMESTAMP(6)
 )"""
 
+CREATE_ICEBERG_RESULTS_TABLE_SQL = """CREATE TABLE IF NOT EXISTS ice_catalog.lake.results (
```
skrydal, line 76: Do I understand correctly that Iceberg tables are created differently than postgres tables, because of Flink acting as the Iceberg Catalog itself? There is no "WITH" statement specifying the catalog for iceberg create statements. I understand this is because Flink allows you to create those in the Iceberg itself. I suggest adding a minor explanation how is it done and that we actually do have a separate IRC running.

dinesh-verma-datahub: Yeah mostly right — minor correction:  Flink is the client, not the catalog, it delegates DDL to a separate IRC container. Added a comment explaining this.

PR #16222, opened 2026-02-16
fix(tableau): extract table level lineage if column metadata is missing

alokr-dhub:
### Problem

The Tableau connector was completely skipping upstream tables when the Tableau Metadata API did not return column information (`columnsConnection.totalCount` is null or 0). This resulted in:

- **No table-level lineage** being created
- **No column-level lineage** being created  
- **External tables deleted** in stateful ingestion
- **Missing upstream relationships** in DataHub

This issue occurs in several scenarios:
- Insufficient permissions (roles below Site Administrator Explorer)
- Tableau's internal metadata collection limitations
- Specific database connector behaviors
- External/published data sources with metadata gaps

Three cascading bugs prevented proper ingestion:
1. **Line 1953:** Premature `continue` in `get_upstream_tables()` skipped tables entirely
2. **Lines 2985-3014:** Tables with IDs but no columns fell through the emission logic:
   - **Phase 1** (tables returned by Tableau API) - skipped because API filters them out
   - **Phase 2** (tables parsed from SQL) - skipped because they have Tableau IDs, assumed already handled
3. **Line 3062:** Early `return` in `emit_table()` when browse paths missing

---

### Solution

This PR modifies the Tableau connector to create **table-level lineage** for upstream tables even when column metadata is unavailable, while still skipping column-level lineage.

#### Key Changes:

**1. Added config flag** `extract_table_lineage_without_columns` (default: `true`)
   - Enables table-level lineage creation for tables without column metadata
   - Provides opt-out for users who want legacy behavior
   - Follows pattern of similar experimental flags

**2. Fixed `get_upstream_tables()` logic** (~lines 1957-1973)
   - Removed unconditional `continue` that was skipping tables
   - Added conditional logic based on config flag
   - Added `num_upstream_table_processed_without_columns` counter for observability

**3. Fixed `emit_table()` browse paths bug** (~lines 3076-3081)
   - Removed early `return` when tables lack browse paths
   - Browse paths are optional UI metadata, not required for emission
   - Allows external upstream tables (postgres, snowflake, etc.) to be emitted

**4. Ensure all upstream tables get emitted, even when filtered by Tableau** (~lines 2996-3052)
   - Tracks emitted URNs across all phases
   - Catches tables with IDs that weren't returned by Tableau's API re-query
   - Emits orphaned tables using metadata from worksheet/datasource context
   - Respects `ingest_tables_external` config flag

**5. Updated documentation**
   - Added "Tables Without Column Metadata" section in `tableau_pre.md`
   - Added example configuration in `tableau_recipe.yml`

---

#### Source Code
- `metadata-ingestion/src/datahub/ingestion/source/tableau/tableau.py`
  - Added `extract_table_lineage_without_columns` config field (~line 607)
  - Added `num_upstream_table_processed_without_columns` counter (~line 841)
  - Modified `get_upstream_tables()` method (~lines 1957-1973)
  - Fixed `emit_table()` browse paths handling (~lines 3076-3081)
  - Added Phase 3 with URN tracking (~lines 2996-3052)

#### Tests
- `metadata-ingestion/tests/unit/tableau/test_tableau_source.py`
  - Added parametrized test for config flag validation
- `metadata-ingestion/tests/integration/tableau/test_tableau_ingest.py`
  - Updated `test_tableau_stateful` assertions

#### Documentation  
- `metadata-ingestion/docs/sources/tableau/tableau_pre.md`
  - Added lineage behavior explanation
  - Updated authentication section
- `metadata-ingestion/docs/sources/tableau/tableau_recipe.yml`
  - Added example configuration

### Impact
**Before:** Tables without column metadata completely skipped (0 entities, 0 lineage)  
**After:** Tables emitted with full entity records (+4,296 lines across golden files)


metadata-ingestion/src/datahub/ingestion/source/tableau/tableau.py
```diff
@@ -604,6 +604,13 @@ class TableauConfig(
         description="[Experimental] Force extraction of lineage from Custom SQL queries using DataHub's SQL parser, even when the Tableau Catalog API returns lineage already.",
     )
 
+    extract_table_lineage_without_columns: bool = Field(
```
skrydal: Do we really need capability to not ingest such tables? This will be a flag which will not be really used I am afraid. Seems to me that having lineage without columns would be considered to be better state than not having lineage at all.

metadata-ingestion/src/datahub/ingestion/source/tableau/tableau.py
```diff
@@ -1940,18 +1948,31 @@ def get_upstream_tables(
...
+            # Check if we should skip tables without column metadata
             if not is_custom_sql and not num_tbl_cols:
-                self.report.num_upstream_table_skipped_no_columns += 1
-                logger.warning(
-                    f"Skipping upstream table with id {table[c.ID]}, no columns: {table}"
-                )
-                continue
-            elif table[c.NAME] is None:
+                if not self.config.extract_table_lineage_without_columns:
+                    # Legacy behavior: skip entirely
```
skrydal: comment like that ages super-quickly, I don't see value of having it in the code base

alokr-dhub: sure. let me remove it.

metadata-ingestion/src/datahub/ingestion/source/tableau/tableau.py
alokr-dhub: added the flag for backward compatibility to ensure no regression from legacy behavior. Let me remove the flag

metadata-ingestion/src/datahub/ingestion/source/tableau/tableau.py
```diff
@@ -1940,18 +1941,21 @@ def get_upstream_tables(
...
+            # Extract column count if available
             num_tbl_cols: Optional[int] = table.get(c.COLUMNS_CONNECTION) and table[
                 c.COLUMNS_CONNECTION
             ].get("totalCount")
+
+            # Tables without column metadata: create table-level lineage only
             if not is_custom_sql and not num_tbl_cols:
-                self.report.num_upstream_table_skipped_no_columns += 1
+                self.report.num_upstream_table_processed_without_columns += 1
                 logger.warning(
```
skrydal: imho we can reduce severity of this log to info or debug

metadata-ingestion/src/datahub/ingestion/source/tableau/tableau.py
```diff
@@ -1940,18 +1941,21 @@ def get_upstream_tables(
         # Same table urn can be used when setting fine grained lineage,
         table_id_to_urn: Dict[str, str] = {}
         for table in tables:
-            # skip upstream tables when there is no column info when retrieving datasource
-            # Lineage and Schema details for these will be taken care in self.emit_custom_sql_datasources()
+            # Extract column count if available
             num_tbl_cols: Optional[int] = table.get(c.COLUMNS_CONNECTION) and table[
                 c.COLUMNS_CONNECTION
             ].get("totalCount")
```
skrydal: There's already a constant for this in `tableau_constant.py`:

```suggestion
            num_tbl_cols: Optional[int] = table.get(c.COLUMNS_CONNECTION) and table[
                c.COLUMNS_CONNECTION
            ].get(c.TOTAL_COUNT)
```

metadata-ingestion/src/datahub/ingestion/source/tableau/tableau.py
```diff
@@ -3013,6 +3020,17 @@ def emit_upstream_tables(self) -> Iterable[MetadataWorkUnit]:
...
+
+        # Phase 3: Catch tables that weren't emitted in Phase 1 or 2
+        if self.config.ingest_tables_external:
+            for database_table in self.database_tables.values():
+                if database_table.urn not in emitted_urns:
+                    logger.debug(
+                        f"Table {database_table.urn} not returned by Tableau re-query (likely no columns). Emitting anyway."
+                    )
+                    yield from self.emit_table(database_table, None)
+                    emitted_urns.add(database_table.urn)
```
skrydal: Phase 2 and Phase 3 can be merged into a single pass. Both iterate `self.database_tables.values()`, both emit with `tableau_columns=None`, and both are gated by `ingest_tables_external`.

Phase 2 filters on "no ID" (SQL-parsed tables), Phase 3 filters on "not in `emitted_urns`". But since Phase 1 queries Tableau by ID, SQL-parsed tables (which have no ID) are never returned by Phase 1 and thus never in `emitted_urns`. The `emitted_urns` check already subsumes the `database_table.id` check.

This block together with the Phase 2 loop above (lines 3007-3023) can be replaced with:

```python
        # Emit remaining tables: SQL-parsed (no Tableau ID) and
        # tables not returned by Tableau re-query (no column metadata)
        if self.config.ingest_tables_external:
            for database_table in self.database_tables.values():
                if database_table.urn not in emitted_urns:
                    yield from self.emit_table(database_table, None)
                    emitted_urns.add(database_table.urn)
```

Removes one full iteration over `self.database_tables`, the `database_table.id` check, and a redundant `ingest_tables_external` gate. No behavior change.

skrydal:
**Minor cleanup needed:** The field `num_upstream_table_skipped_no_columns` in `TableauSourceReport` (line 833) is now dead code — it's still declared but never incremented anywhere after this PR's changes. The only call site was removed when the `continue` was replaced with the new table-level lineage logic.

Users will see `num_upstream_table_skipped_no_columns: 0` in reports and may incorrectly conclude all tables had column metadata, when in reality the counter is just unused. Its replacement is `num_upstream_table_processed_without_columns`.

Please remove the dead field to avoid confusion:

```python
# Remove this line from TableauSourceReport:
num_upstream_table_skipped_no_columns: int = 0
```

metadata-ingestion/src/datahub/ingestion/source/tableau/tableau.py
alokr-dhub: updated the code to use TOTAL_COUNT const instead of str

metadata-ingestion/src/datahub/ingestion/source/tableau/tableau.py
alokr-dhub: I have merged phase 2 and 3 iterations in a single loop as per your suggestions

alokr-dhub:
> **Minor cleanup needed:** The field `num_upstream_table_skipped_no_columns` in `TableauSourceReport` (line 833) is now dead code — it's still declared but never incremented anywhere after this PR's changes. The only call site was removed when the `continue` was replaced with the new table-level lineage logic.
> 
> Users will see `num_upstream_table_skipped_no_columns: 0` in reports and may incorrectly conclude all tables had column metadata, when in reality the counter is just unused. Its replacement is `num_upstream_table_processed_without_columns`.
> 
> Please remove the dead field to avoid confusion:
> 
> ```python
> # Remove this line from TableauSourceReport:
> num_upstream_table_skipped_no_columns: int = 0
> ```

removed the field from report.

PR #16224, opened 2026-02-16
ci: test connector-tests-trigger workflow 

rajatoss:
## Summary

Adds the `connector-tests-trigger.yml` workflow that automatically triggers connector tests on the private `acryldata/connector-tests` repo when a PR's upstream workflows pass.

## How it works

1. Fires on completion of either "Python Build" or "Metadata Ingestion" workflows
2. Checks that **both** workflows have passed for the same commit SHA
3. Skips fork PRs (`head_repository.full_name == github.repository` guard)
4. Finds the Cloudflare Pages deployment URL for the built wheel
5. Sets a `pending` commit status (`connector-tests` context) on the PR
6. Dispatches a `repository_dispatch` event to `acryldata/connector-tests` with the PR number, SHA, source repo, and wheel URL
7. On failure, sets an `error` commit status so the PR doesn't hang at "pending"

## Key design decisions

- **Concurrency group** uses `workflow name + upstream workflow name + branch` so both "Python Build" and "Metadata Ingestion" triggers run independently (each checks if both passed)
- **`cancel-in-progress: true`** ensures rapid pushes don't pile up trigger runs
- **Cloudflare URL** is the base deployment URL — the install script appends the wheel path itself
- **Fire-and-forget** — the dispatch is async; result reporting is handled by the receiving workflow in connector-tests

## Required secrets

- `GITHUB_TOKEN` — read workflow runs, deployments, PRs; write commit statuses (same repo)
- `CONNECTOR_TESTS_DISPATCH_TOKEN` — PAT with repo scope on `acryldata/connector-tests` to send the dispatch

## Companion PR

- acryldata/connector-tests#681 — the receiving workflow that runs tests and reports results back

## Test plan

- [ ] Verify "Python Build" completion triggers this workflow
- [ ] Verify "Metadata Ingestion" completion triggers this workflow
- [ ] Verify both-passed check works (only dispatches when both succeed)
- [ ] Verify fork PRs are skipped
- [ ] Verify Cloudflare wheel URL is resolved correctly
- [ ] Verify `pending` commit status appears on the PR
- [ ] Verify dispatch reaches connector-tests repo
- [ ] Verify rapid pushes cancel stale trigger runs


rajatoss:
To test these workflows, I need to merge them to master/main, I have tried to test each step individually wherever it was possible, the connector-tests results currently will be used for informational purposes, they are not blocking. Once I confirm that the complete flow is working as expected, I will add it as a blocking step.
These two workflow additions shouldn't have any side effects on the current state of our CI checks.

.github/workflows/connector-tests-trigger.yml
```diff
@@ -0,0 +1,132 @@
...
+            echo "::error::No Cloudflare deployment found for SHA ${SHA}. Re-run the Python Build workflow."
+            exit 1
+          fi
+
+          DEPLOY_URL=$(gh api "/repos/${REPO}/deployments/${DEPLOY_ID}/statuses" \
+            --jq '.[0].environment_url // empty')
+
+          if [ -z "$DEPLOY_URL" ]; then
+            echo "::error::Cloudflare deployment ${DEPLOY_ID} exists but has no URL."
+            exit 1
```
sgomezvillamor, line 66: this `exit 1` and the one a few lines above, will make the workflow to fail and so continue the execution flow here, right?

```
      - name: On failure — set error commit status
        if: failure() && steps.pr-info.outputs.has_pr == 'true'
```

.github/workflows/connector-tests-trigger.yml
```diff
@@ -0,0 +1,132 @@
...
+          MI=$(gh api "/repos/${REPO}/actions/workflows/metadata-ingestion.yml/runs?head_sha=${SHA}&per_page=1" \
+            --jq '.workflow_runs[0].conclusion // empty')
+
+          echo "Python Build: $WB, Metadata Ingestion: $MI"
+
+          if [ "$WB" = "success" ] && [ "$MI" = "success" ]; then
+            echo "both_passed=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "Not both passed yet (WB=$WB, MI=$MI) — other trigger will handle it."
+            echo "both_passed=false" >> "$GITHUB_OUTPUT"
```
sgomezvillamor, line 41: we may force the failure here too, so it goes to

```
      - name: On failure — set error commit status
        if: failure() && steps.pr-info.outputs.has_pr == 'true'
```

WDYT?

.github/workflows/connector-tests-trigger.yml
```diff
@@ -0,0 +1,132 @@
...
+        run: |
+          gh api "/repos/${{ github.repository }}/statuses/${{ github.event.workflow_run.head_sha }}" \
+            -f state=pending \
+            -f context="connector-tests" \
+            -f description="Connector tests triggered, waiting for results..."
+
+      - name: Dispatch to connector-tests
+        if: steps.check-both.outputs.both_passed == 'true' && steps.pr-info.outputs.has_pr == 'true'
+        env:
+          GH_TOKEN: ${{ secrets.CONNECTOR_TESTS_DISPATCH_TOKEN }}
```
sgomezvillamor, line 109: Are you using a PAT, or do we have a GitHub App set up for this?

It would be good to leave some kind of breadcrumb or comment — something that helps us trace this back in case we ever need to _eg_ rotate the token.

.github/workflows/connector-tests-trigger.yml
rajatoss: Yes, when exit 1 fires, the step fails, the job gets failure() status, and the "On failure" step  catches it and sets an error commit status.

.github/workflows/connector-tests-trigger.yml
rajatoss: the workflow fires twice per SHA, once per upstream workflow completion. The first trigger will see only 1/2 passed and skip, the second trigger sees both passed and dispatches. If we exit 1 here, every PR would briefly get a misleading "Failed to trigger connector tests" error status from the first trigger. If either upstream workflow failed, we simply don't dispatch,  the user needs to fix the upstream failure first.

.github/workflows/connector-tests-trigger.yml
rajatoss: yes, added a comment for PAT rotation

.github/workflows/connector-tests-trigger.yml
rajatoss: Improved the log a bit to avoid the confusion.

PR #16228, opened 2026-02-16
fix(Es8): use system properties for elastic hosts behin a proxy, and use async mode for reindex (SystemUpdate)

amidani-inovany:


david-leifker:
Sorry for the late response, and thanks for the ES-behind-a-proxy / async reindex report.

We're not going to merge this PR as written:

- **`waitForCompletion(false)` on ES8 `submitReindexTask`** is already on master. The unified OpenSearch shim (OS 2 and OS 3) already submits reindex asynchronously.
- **`HttpAsyncClientBuilder.useSystemProperties()`** is process-wide and also picks up non-proxy JVM settings (`http.maxConnections`, SSL trust stores, etc.). That fights our per-client connection manager and cannot diverge across named search clusters (`elasticsearch.clusters.primary` vs `secondary`).

**Follow-up on master:** per-cluster `elasticsearch.clusters.<name>.proxy.*` (explicit host/port/scheme/credentials). When `proxy.host` is blank and `useSystemProxyProperties` is true (default), we read **only** JVM proxy properties (`http.proxyHost` / `http.proxyPort`, `https.proxyHost` / `https.proxyPort`, `http.nonProxyHosts`). We still do **not** call Apache `useSystemProperties()`. A secondary cluster can set `useSystemProxyProperties: false` to stay direct while primary uses a JVM proxy.

That work is tracked on OSS-1359. Closing this PR in favor of that approach.

david-leifker:
Closing in favor of per-cluster proxy configuration (OSS-1359). See the previous review comment for the alternative.

PR #16235, opened 2026-02-17
feat(ingestion): Add multi-statement SQL parsing support for Airflow lineage extraction

aviraj-gour:
This PR implements multi-statement SQL parsing capabilities in the sqlglot-based lineage extraction, enabling DataHub to properly handle complex SQL workflows with multiple statements, temporary tables, and intermediate transformations commonly found in Airflow DAGs.

## Core SQL Parsing (sqlglot_lineage.py)

- Multi-statement parsing: Refactored SQL parsing to handle multiple statements in a single query string
- Statement boundary tracking: Properly tracks lineage relationships across statement boundaries
- Temporary table resolution: Enhanced logic to track temporary tables created and consumed across statements
- Improved column lineage: Better column-level lineage extraction for multi-statement scenarios

## Configuration (_config.py)

Added enable_multi_statement_parsing configuration flag
Allows users to opt-out of multi-statement parsing if needed for compatibility

### DIFF
--- a/metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
+++ b/metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
@@ -2235,7 +2235,7 @@ def _prepare_sql_query_list(queries: Union[str, List[str]]) -> List[str]:
     if isinstance(queries, str):
         return [stmt for stmt in split_statements(queries) if stmt.strip()]
     else:
-        result = []
+        result: List[str] = []
         for q in queries:
             if q and str(q).strip():
                 result.extend(stmt for stmt in split_statements(str(q)) if stmt.strip())


metadata-ingestion-modules/airflow-plugin/src/datahub_airflow_plugin/airflow2/_airflow2_sql_parser_patch.py
```diff
@@ -48,6 +53,56 @@
...
+    sql: Union[str, List[str]],
+    default_database: Optional[str],
+    platform: str,
+    env: str,
+    default_schema: Optional[str],
+    graph: Optional[DataHubGraph],
+    enable_multi_statement: bool,
+) -> SqlParsingResult:
+    if enable_multi_statement:
+        return create_lineage_from_sql_statements(
```
treff7es: Do we log these errors?

metadata-ingestion-modules/airflow-plugin/src/datahub_airflow_plugin/airflow2/_airflow2_sql_parser_patch.py
```diff
@@ -48,6 +53,56 @@
...
+            default_db=default_database,
+            platform=platform,
+            platform_instance=None,
+            env=env,
+            default_schema=default_schema,
+            graph=graph,
+        )
+
+
+def _format_sql_for_job_facet(sql: Any, enable_multi_statement: bool) -> str:
```
treff7es:
```suggestion
def _format_sql_for_job_facet(sql: Union[str, List[str]], enable_multi_statement: bool) -> str:
```

metadata-ingestion-modules/airflow-plugin/src/datahub_airflow_plugin/airflow3/_airflow3_sql_parser_patch.py
```diff
@@ -45,6 +50,56 @@
 _original_sql_parser_method: Optional[Callable[..., Any]] = None
 
 
+def _parse_sql_with_datahub(
```
treff7es: Can we add some unit tests for this new feature?

aviraj-gour: added

metadata-ingestion-modules/airflow-plugin/src/datahub_airflow_plugin/airflow2/_airflow2_sql_parser_patch.py
aviraj-gour: added warning log

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
```diff
@@ -2229,6 +2231,174 @@ def create_lineage_sql_parsed_result(
             schema_resolver.close()
 
 
+def _prepare_sql_query_list(queries: Union[str, List[str]]) -> List[str]:
+    if isinstance(queries, str):
+        return [stmt for stmt in split_statements(queries) if stmt.strip()]
```
treff7es, line 2236: I think this doesn't handle multiple statements which are separated by `;` character like:
```
["CREATE TEMP
  TABLE t AS SELECT 1; INSERT INTO target SELECT * FROM t"]
```

this can be a known issue but maybe worth to document it if we don't want to support it

treff7es: Would it make sense to support it?

aviraj-gour: This is a very specific case, but we can support it, so I’ve updated the code accordingly.

PR #16253, opened 2026-02-18
fix(redshift): use boundary-aware segment stitching for query reconstruction

kyungsoo-datahub:
Redshift stores queries in fixed-width segments (200 chars for provisioned, 4000 for serverless). When using LISTAGG with per-segment RTRIM, tokens at boundaries merge  without spaces (GROUP BY → GROUPBY).

Fix: Insert a space when trimmed segment length is less than segment size (indicates padding). Applied to all 6 LISTAGG expressions in both modes.

Also: Replace stl_query.querytxt (truncated to 4000 chars) with STL_QUERYTEXT CTE in provisioned scan-based lineage.

Added tests for boundary detection and segment size correctness.

### DIFF
--- a/datahub-web-react/src/app/lineageV3/components/LineageCard.tsx
+++ b/datahub-web-react/src/app/lineageV3/components/LineageCard.tsx
@@ -119,6 +119,7 @@ const ColumnButtonWrapper = styled.div`
 const StyledPlatformIcon = styled.img`
     height: 16px;
     width: 16px;
+    object-fit: contain;
 `;
 
 // Defaults to phosphor icon if the img fails to load for any reason
--- a/metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/index/entity/v2/V2MappingsBuilder.java
+++ b/metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/index/entity/v2/V2MappingsBuilder.java
@@ -166,7 +166,15 @@ public class V2MappingsBuilder implements MappingsBuilder {
               ((Map<String, Object>) mappings.get(PROPERTIES))
                   .computeIfAbsent(
                       STRUCTURED_PROPERTY_MAPPING_FIELD,
-                      (key) -> new HashMap<>(Map.of(PROPERTIES, new HashMap<>())));
+                      (key) ->
+                          new HashMap<>(
+                              Map.of(
+                                  TYPE,
+                                  ESUtils.OBJECT_FIELD_TYPE,
+                                  "dynamic",
+                                  true,
+                                  PROPERTIES,
+                                  new HashMap<>())));
 
       props.merge(
           PROPERTIES,
@@ -334,6 +342,7 @@ public class V2MappingsBuilder implements MappingsBuilder {
       mappingForField.put(TYPE, ESUtils.DATE_FIELD_TYPE);
     } else if (OBJECT_FIELD_TYPES.contains(fieldType)) {
       mappingForField.put(TYPE, ESUtils.OBJECT_FIELD_TYPE);
+      mappingForField.put("dynamic", true);
     } else if (fieldType == FieldType.DOUBLE) {
       mappingForField.put(TYPE, ESUtils.DOUBLE_FIELD_TYPE);
     } else {
--- a/metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/index/entity/v3/MultiEntityMappingsBuilder.java
+++ b/metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/index/entity/v3/MultiEntityMappingsBuilder.java
@@ -838,6 +838,9 @@ public class MultiEntityMappingsBuilder implements MappingsBuilder {
       // Create root field mapping - this is the target field that aspect fields will copy_to
       Map<String, Object> rootFieldMapping = new HashMap<>();
       rootFieldMapping.put(TYPE, resolvedElasticsearchType);
+      if (ESUtils.OBJECT_FIELD_TYPE.equals(resolvedElasticsearchType)) {
+        rootFieldMapping.put("dynamic", true);
+      }
 
       // Check if any of the conflicting fields have eagerGlobalOrdinals set to true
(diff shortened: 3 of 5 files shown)


rajatoss:
## Connector Tests Results

**Connector tests failed** for commit `e27b075`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22675539486)

sgomezvillamor:
My overall concern is that we’re adding quite a lot of complexity (and therefore risk of errors) to address what seems to be an anecdotal issue. How recurrent is this issue?

Beyond complexity, proposed solution is not _perfect_.

The keyword-adjacency heuristic is fragile. We’d need a comprehensive reserved word list, and edge cases will grow quickly. For example:

```sql
-- GROUP as a keyword — needs space
SELECT a, COUNT(*) FROM t GROUP<CHR1>BY a

-- GROUP as an identifier — space is harmless but detection logic
-- must correctly classify it, which requires parsing context
SELECT group<CHR1>_id FROM t   -- identifier, no space needed
```

This would require bidirectional lookahead around every marker, which increases complexity even more.

The reserved word list itself is also a maintenance burden. Redshift’s reserved words may change across versions, and include non-obvious entries. Keeping that list correct and up to date is operational overhead.

Additionally, there’s the risk of false positives from string literals. `CHR(1)` is the SOH (Start of Heading) ASCII control character. It’s exotic, but Redshift allows it inside string literals. Very unlikely, but still technically possible.

---

We currently read from `SQL_QUERYTEXT` (or `SQL_QUERY_TEXT` in serverless).

There is also `SVL_STATEMENTTEXT`: https://docs.aws.amazon.com/redshift/latest/dg/r_SVL_STATEMENTTEXT.html

That view is built on top of `SQL_QUERYTEXT`. I’m not sure if it’s available in serverless. The docs describe how to reconstruct statements there.

Does `SVL_STATEMENTTEXT` suffer from the same reconstruction issues as `SQL_QUERYTEXT`? If not, maybe we could switch to it.

If it’s not supported in serverless (given the 4K char limit there), is this issue still frequent enough that we must address it?

---

Another alternative would be audit logging to S3: https://docs.aws.amazon.com/redshift/latest/mgmt/db-auditing.html

From what I understand, the query statement is not split there. But this would be a completely different ingestion approach.

I agree this may not feasible in the short term.

---

My current view:

* The current proposal adds significant complexity and is still not a perfect solution.
* Can we consider a simpler, more naive concat? Not perfect either, but much less complex.
* And should we evaluate `SVL_STATEMENTTEXT` more seriously first?

IMO final decision depends on how recurrent is the issue.

metadata-ingestion/src/datahub/ingestion/source/redshift/query.py
```diff
@@ -507,14 +517,24 @@ def stl_scan_based_lineage_query(
         db_name: str, start_time: datetime, end_time: datetime
     ) -> str:
         return """
+                    WITH query_txt AS (
+                        SELECT
+                            query,
+                            userid,
+                            RTRIM(LISTAGG(RTRIM(text) || CASE WHEN LEN(RTRIM(text)) < {_PROVISIONED_SEGMENT_SIZE} THEN ' ' ELSE '' END, '')
```
sgomezvillamor, line 524: LGTM
Just wondering how is the implementation in PR is slightly different to the one suggested here

https://docs.aws.amazon.com/redshift/latest/dg/r_SVL_STATEMENTTEXT.html

```
select LISTAGG(CASE WHEN LEN(RTRIM(text)) = 0 THEN text ELSE RTRIM(text) END, '') within group (order by sequence) AS query_statement 
from SVL_STATEMENTTEXT where pid=pg_backend_pid();
```

metadata-ingestion/tests/unit/redshift/test_redshift_queries.py
sgomezvillamor, line 1: Basically, we are just doing _string contains_ validations

So I would remove the focus on "stitching" and make this just a generic test for redshift queries
* `test_redshift_queries.py`
* `class TestProvisionedQueries`
* `class TestServerlessQueries`
* ...

So we can add more validations in the future

metadata-ingestion/tests/unit/redshift/test_redshift_query_stitching.py
```diff
@@ -0,0 +1,182 @@
...
+            RedshiftServerlessQuery.temp_table_ddl_query(
+                start_time=START_TIME, end_time=END_TIME
+            ),
+        ]:
+            # Should not have bare LISTAGG without RTRIM wrapper
+            assert 'LISTAGG(qt."text")' not in sql
+            assert "LEN(RTRIM(querytxt)) = 0" not in sql
+
+
+class TestBoundaryAwareStitchingLogic:
```
sgomezvillamor: I don't see much value on these tests using `_simulate_boundary_aware_stitching`. We could just remove them.

If we really want to test this in an integration test, we could add a eg long enough select in https://github.com/acryldata/connector-tests/blob/main/smoke-test/integration/create_data/redshift_data.py

metadata-ingestion/src/datahub/ingestion/source/redshift/query.py
kyungsoo-datahub: AWS's approach doesn't fix the keyword merge problem. Their condition LEN(RTRIM(text)) = 0 only preserves completely empty segments. A segment like "GROUP   " becomes "GROUP" (length 5, not 0), which merges with the next "BY" segment -> "GROUPBY"
                                                                                                                        
Our approach uses segment size as a boundary detector: if LEN(RTRIM(text)) < segment_size, we add a space back because the segment likely reached a natural boundary (not just had trailing spaces stripped).

metadata-ingestion/tests/unit/redshift/test_redshift_queries.py
kyungsoo-datahub: Revised.

metadata-ingestion/tests/unit/redshift/test_redshift_query_stitching.py
kyungsoo-datahub: Removed.

PR #16281, opened 2026-02-19
fix(semanticsearchretry): ensure retry for semanticsearch

nwadams:
Fix OpenAI and Cohere provider `connection reset` errors in semantic search
add 1 retry (max 2 attempts)

https://bugs.openjdk.org/browse/JDK-8297558#:~:text=JDK-,Description,Specification

The issue happens due to HTTP2 holding the connection open for a long lived HTTP Client. Load balancer on our end, or OpenAI end is closing the connections before the 1200s keepalive timeout occurs causing these connection reset errors. 

This change prevents the errors by using HTTP1.1 as recommended to reduce connection reuse as well as adding a retry to ensure we make a new connection upon error. This should improve the stability of semantic search calls.

Changes not made for Bedrock Provider because we use the AWS Bedrock client which has its own internal connection pooling + retries.

exception:
```
026-02-19 18:10:47,907 [ForkJoinPool.commonPool-worker-605] ERROR c.l.m.s.e.OpenAIEmbeddingProvider:179 - Failed to generate embedding with model text-embedding-3-large: Connection reset
java.io.IOException: Connection reset
	at java.net.http/jdk.internal.net.http.HttpClientImpl.send(HttpClientImpl.java:917)
	at java.net.http/jdk.internal.net.http.HttpClientFacade.send(HttpClientFacade.java:133)
	at com.linkedin.metadata.search.embedding.OpenAIEmbeddingProvider.embed(OpenAIEmbeddingProvider.java:126)
	at com.linkedin.metadata.search.semantic.SemanticEntitySearchService.search(SemanticEntitySearchService.java:184)
	at com.linkedin.metadata.search.SemanticSearchService.performSemanticSearch(SemanticSearchService.java:166)
	at com.linkedin.metadata.search.SemanticSearchService.semanticSearchAcrossEntities(SemanticSearchService.java:103)
	at com.linkedin.metadata.search.SemanticSearchService.semanticSearch(SemanticSearchService.java:129)
	at com.linkedin.datahub.graphql.resolvers.semantic.SemanticSearchResolver.lambda$get$1(SemanticSearchResolver.java:100)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1768)
	at io.opentelemetry.context.Context.lambda$wrap$1(Context.java:241)
	at java.base/java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1395)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:373)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1182)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1655)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1622)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:165)
Caused by: java.net.SocketException: Connection reset
	at java.base/sun.nio.ch.SocketChannelImpl.throwConnectionReset(SocketChannelImpl.java:394)
	at java.base/sun.nio.ch.SocketChannelImpl.read(SocketChannelImpl.java:426)
	at java.net.http/jdk.internal.net.http.SocketTube.readAvailable(SocketTube.java:1178)
	at java.net.http/jdk.internal.net.http.SocketTube$InternalReadPublisher$InternalReadSubscription.read(SocketTube.java:841)
	at java.net.http/jdk.internal.net.http.SocketTube$SocketFlowTask.run(SocketTube.java:181)
	at java.net.http/jdk.internal.net.http.common.SequentialScheduler$SchedulableTask.run(SequentialScheduler.java:230)
	at java.net.http/jdk.internal.net.http.common.SequentialScheduler.runOrSchedule(SequentialScheduler.java:303)
	at java.net.http/jdk.internal.net.http.common.SequentialScheduler.runOrSchedule(SequentialScheduler.java:256)
	at java.net.http/jdk.internal.net.http.SocketTube$InternalReadPublisher$InternalReadSubscription.signalReadable(SocketTube.java:782)
	at java.net.http/jdk.internal.net.http.SocketTube$InternalReadPublisher$ReadEvent.signalEvent(SocketTube.java:965)
	at java.net.http/jdk.internal.net.http.SocketTube$SocketFlowEvent.handle(SocketTube.java:253)
	at java.net.http/jdk.internal.net.http.HttpClientImpl$SelectorManager.handleEvent(HttpClientImpl.java:1402)
	at java.net.http/jdk.internal.net.http.HttpClientImpl$SelectorManager.lambda$run$3(HttpClientImpl.java:1347)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.net.http/jdk.internal.net.http.HttpClientImpl$SelectorManager.run(HttpClientImpl.java:1347)

```

### DIFF
--- a/metadata-io/src/main/java/com/linkedin/metadata/search/embedding/CohereEmbeddingProvider.java
+++ b/metadata-io/src/main/java/com/linkedin/metadata/search/embedding/CohereEmbeddingProvider.java
@@ -42,8 +42,7 @@ public class CohereEmbeddingProvider implements EmbeddingProvider {
   private static final String INPUT_TYPE = "search_query";
   private static final String COHERE_VERSION = "2024-01-01";
   private static final Duration DEFAULT_TIMEOUT = Duration.ofSeconds(30);
-  private static final int MAX_ATTEMPTS = 3;
-  private static final long RETRY_DELAY_MS = 1000L;
+  private static final int MAX_ATTEMPTS = 2;
 
   private final String apiKey;
   private final String endpoint;
@@ -121,18 +120,11 @@ public class CohereEmbeddingProvider implements EmbeddingProvider {
         lastException = e;
         if (attempt < MAX_ATTEMPTS) {
           log.warn(
-              "Cohere embedding attempt {}/{} failed for model {}, retrying in {}s: {}",
+              "Cohere embedding attempt {}/{} failed for model {}, retrying: {}",
               attempt,
               MAX_ATTEMPTS,
-              attempt,
               modelToUse,
               e.getMessage());
-          try {
-            Thread.sleep(attempt * RETRY_DELAY_MS);
-          } catch (InterruptedException ie) {
-            Thread.currentThread().interrupt();
-            break;
-          }
         }
       }
     }
--- a/metadata-io/src/main/java/com/linkedin/metadata/search/embedding/OpenAIEmbeddingProvider.java
+++ b/metadata-io/src/main/java/com/linkedin/metadata/search/embedding/OpenAIEmbeddingProvider.java
@@ -38,8 +38,7 @@ public class OpenAIEmbeddingProvider implements EmbeddingProvider {
   private static final String DEFAULT_MODEL = "text-embedding-3-small";
   private static final String DEFAULT_ENDPOINT = "https://api.openai.com/v1/embeddings";
   private static final Duration DEFAULT_TIMEOUT = Duration.ofSeconds(30);
-  private static final int MAX_ATTEMPTS = 3;
-  private static final long RETRY_DELAY_MS = 1000L;
+  private static final int MAX_ATTEMPTS = 2;
 
   private final String apiKey;
   private final String endpoint;
@@ -117,18 +116,11 @@ public class OpenAIEmbeddingProvider implements EmbeddingProvider {
         lastException = e;
         if (attempt < MAX_ATTEMPTS) {
           log.warn(
-              "OpenAI embedding attempt {}/{} failed for model {}, retrying in {}s: {}",
+              "OpenAI embedding attempt {}/{} failed for model {}, retrying: {}",
               attempt,
               MAX_ATTEMPTS,
-              attempt,
               modelToUse,
               e.getMessage());
-          try {
-            Thread.sleep(attempt * RETRY_DELAY_MS);
-          } catch (InterruptedException ie) {
-            Thread.currentThread().interrupt();
-            break;
-          }
         }
       }
     }


metadata-io/src/main/java/com/linkedin/metadata/search/embedding/CohereEmbeddingProvider.java
```diff
@@ -102,104 +106,111 @@ public CohereEmbeddingProvider(
...
+
+    for (int attempt = 1; attempt <= MAX_ATTEMPTS; attempt++) {
+      try {
+        return embedInternal(text, modelToUse);
+      } catch (RuntimeException e) {
+        // Non-retryable: bad API key, malformed response, 4xx errors, etc.
+        lastException = e;
+        break;
+      } catch (Exception e) {
+        lastException = e;
```
alexsku, line 120: we need a sleep between the retries, if I correctly understand what you are trying to achieve

alexsku:
have you considered switching to openai sdk instead of manually written client? I would expect it would have proper retries implemented

nwadams:
@alexsku I added a second retry (3 total attempts) with a small delay between attempts.

alexsku:
After looking at this more carefully, I think the original approach (single retry, no sleep) was actually the better one. Here's my reasoning:

**The sleep is counterproductive for this failure mode.** The root cause is a stale connection being reset by a load balancer. When we retry, we get a fresh connection regardless of whether we wait 0ms or 3 seconds — the old connection is already dead, there's nothing to "recover." This isn't a rate-limited or overloaded server situation where backing off helps.

**Sleeping on `ForkJoinPool.commonPool` threads is problematic.** Looking at the stack trace, these embedding calls run on common pool workers. The common pool has limited parallelism (typically CPU count - 1), so blocking a thread for 1-3 seconds can starve other concurrent work — including other embedding requests. To do retries with delays properly here, we'd need async retry with `CompletableFuture.delayedExecutor`, which is a bigger change.

**The current code also has a couple of issues:**
- The log message has swapped parameters — `attempt` and `modelToUse` are in the wrong positions, so warnings would log something like `"attempt 1/3 failed for model 1, retrying in text-embedding-3-larges"`
- `MAX_ATTEMPTS = 3` but tests assert `times(2)` — these are inconsistent
- 429 (rate limit) is not retried since it's a 4xx, but that's arguably the most common transient error worth retrying

**My suggestion:** let's go back to the simpler approach — exactly one retry (`MAX_ATTEMPTS = 2`), no sleep. The HTTP/1.1 downgrade is the real fix. The single retry is just a safety net for the rare case a connection still gets reset, and it doesn't need a delay to work. This keeps the code simple and avoids the thread-blocking concern entirely.

alexsku:
LGTM

PR #16290, opened 2026-02-20
ci(pre-commit) Use action lint docker hook

devashish2203:
The github based install hook required people to install the go toolchain on their systems.
Instead moving to a docker based pre-commit hook to remove that dependency.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -90,11 +90,25 @@ repos:
         files: ^metadata-dao-impl/kafka-producer/.*\.java$
         pass_filenames: false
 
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -90,13 +90,27 @@ repos:
         files: ^metadata-dao-impl/kafka-producer/.*\.java$
         pass_filenames: false
 
+      - id: metadata-events-mxe-avro-spotless
```
devashish2203, line 93: Need to investigate why these get added or removed.
They were removed in my earlier PR automatically and now added back.

anshbansal (approved):

PR #16292, opened 2026-02-20
feat(trino): add column-level lineage on upstreamLineage to connector…

alfiyas-datahub:
**Summary**

Adds column-level lineage (CLL) on the upstreamLineage from Trino to connector sources (e.g. Iceberg, Hive) and emits upstreamLineage for views (previously only siblings were emitted). Addresses ING-1608.

**Solution**

New config include_column_lineage (default True). When set and schema is available, gen_lineage_workunit emits fineGrainedLineages with a 1:1 column mapping in addition to table-level upstreams.
Tables: build schema and pass it into gen_lineage_workunit when emitting lineage to a connector.
Views: now call gen_lineage_workunit as well (so they get upstreamLineage and optional CLL, not only siblings).
Unit tests in test_sql_common.py cover CLL on/off and schema present/absent.


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `b255323`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22755196862)

skrydal:
The Trino-specific unit tests added here (`test_trino_gen_lineage_workunit_*`, `test_trino_process_table_*`, `test_trino_process_view_*`) should live in a dedicated Trino test file (e.g., `tests/unit/test_trino.py`) rather than `test_sql_common.py`.

`test_sql_common.py` tests shared SQL infrastructure (foreign keys, SQLAlchemy URI mapping, fine-grained lineage in the base class, etc.). These new tests import and exercise `TrinoSource` and `TrinoConfig` directly — they're connector-specific, not shared infrastructure.

skrydal:
### Replace `Optional[Any]` with `Optional[SchemaMetadataClass]`

Three occurrences use `Optional[Any]` where the type is always `SchemaMetadataClass`:

1. `gen_lineage_workunit` parameter: `schema_metadata: Optional[Any] = None`
2. `_process_table` local: `schema_metadata: Optional[Any] = None`
3. `_process_view` local: `schema_metadata_view: Optional[Any] = None`

`get_schema_metadata()` already returns `SchemaMetadataClass`, so erasing this to `Any` disables all downstream type checking. The project style guide explicitly says: *"Avoid `Any` type — use specific types."*

**Fix** — change all three to `Optional[SchemaMetadataClass]` and add `SchemaMetadataClass` to the existing `datahub.metadata.schema_classes` import block in `trino.py` (it's already imported in the test file):

```python
def gen_lineage_workunit(
    self,
    dataset_urn: str,
    source_dataset_urn: str,
    schema_metadata: Optional[SchemaMetadataClass] = None,
```

```python
schema_metadata: Optional[SchemaMetadataClass] = None   # in _process_table
schema_metadata_view: Optional[SchemaMetadataClass] = None  # in _process_view
```

skrydal:
### Replace `getattr` duck typing with direct attribute access

Once `schema_metadata` is typed as `Optional[SchemaMetadataClass]` (see previous comment), the `getattr` fallback in `gen_lineage_workunit` (`trino.py`) becomes unnecessary:

```python
# Current:
and getattr(schema_metadata, "fields", None)

# With proper typing, simplify to:
and schema_metadata.fields
```

`SchemaMetadataClass.fields` is a known attribute (always present, defaults to `[]`), so after the `is not None` check, direct access is safe and mypy can verify it.

skrydal:
### Suggestion: Reuse schema from parent instead of re-fetching

The CLL here builds a 1:1 field mapping — the same field path is used for both the upstream (connector source) and downstream (Trino) dataset. This relies on the assumption that field paths are identical on both sides.

If that assumption holds, there's no need to re-fetch columns from the inspector. The parent `_process_table` (in `sql_common.py`) already calls `inspector.get_columns()`, `get_schema_fields()`, and `get_schema_metadata()` for the same table via the same Trino inspector. The Trino override then duplicates all of those calls just to get the field paths.

Instead, you could intercept the workunits from `super()` and extract the already-emitted `SchemaMetadataClass`:

```python
def _process_table(self, dataset_name, inspector, schema, table, sql_config, data_reader):
    schema_metadata = None
    for wu in super()._process_table(
        dataset_name, inspector, schema, table, sql_config, data_reader
    ):
        sm = wu.get_aspect_of_type(SchemaMetadataClass)
        if sm is not None:
            schema_metadata = sm
        yield wu

    if self.config.ingest_lineage_to_connectors:
        # ... use schema_metadata directly for CLL, no redundant inspector calls
```

Same approach would apply to `_process_view`.

Could you check whether this would be a valid approach? It would eliminate the duplicate inspector queries and the redundant `try/except` blocks for schema building.

metadata-ingestion/tests/unit/test_sql_common.py
```diff
@@ -200,3 +205,275 @@ def get_field_from_urn(urn):
...
+        "urn:li:dataset:(urn:li:dataPlatform:iceberg,contextad.accountcontact,PROD)"
+    )
+    schema_metadata = get_test_trino_schema_metadata(["accountid"])
+
+    workunits = list(
+        source.gen_lineage_workunit(dataset_urn, source_dataset_urn, schema_metadata)
+    )
+    assert len(workunits) == 1
+    upstream_lineage = workunits[0].get_aspect_of_type(UpstreamLineageClass)
+    assert upstream_lineage is not None
```
skrydal: to avoid calling getattr you could assert here `isinstance`.

alfiyas-datahub: Done!

alfiyas-datahub:
> ### Suggestion: Reuse schema from parent instead of re-fetching
> The CLL here builds a 1:1 field mapping — the same field path is used for both the upstream (connector source) and downstream (Trino) dataset. This relies on the assumption that field paths are identical on both sides.
> 
> If that assumption holds, there's no need to re-fetch columns from the inspector. The parent `_process_table` (in `sql_common.py`) already calls `inspector.get_columns()`, `get_schema_fields()`, and `get_schema_metadata()` for the same table via the same Trino inspector. The Trino override then duplicates all of those calls just to get the field paths.
> 
> Instead, you could intercept the workunits from `super()` and extract the already-emitted `SchemaMetadataClass`:
> 
> ```python
> def _process_table(self, dataset_name, inspector, schema, table, sql_config, data_reader):
>     schema_metadata = None
>     for wu in super()._process_table(
>         dataset_name, inspector, schema, table, sql_config, data_reader
>     ):
>         sm = wu.get_aspect_of_type(SchemaMetadataClass)
>         if sm is not None:
>             schema_metadata = sm
>         yield wu
> 
>     if self.config.ingest_lineage_to_connectors:
>         # ... use schema_metadata directly for CLL, no redundant inspector calls
> ```
> 
> Same approach would apply to `_process_view`.
> 
> Could you check whether this would be a valid approach? It would eliminate the duplicate inspector queries and the redundant `try/except` blocks for schema building.



Done — refactored both _process_table and _process_view to intercept workunits from super() and extract the already-emitted SchemaMetadataClass as suggested. This eliminates the duplicate inspector.get_columns(), get_pk_constraint(), get_schema_fields(), and get_schema_metadata() calls along with the try/except blocks. The get_schema_metadata import from sql_common is also removed since it's no longer needed.

skrydal:
This comment from the review appears to still be unaddressed:

> The Trino-specific unit tests added here (`test_trino_gen_lineage_workunit_*`, `test_trino_process_table_*`, `test_trino_process_view_*`) should live in a dedicated Trino test file (e.g., `tests/unit/test_trino.py`) rather than `test_sql_common.py`.
>
> `test_sql_common.py` tests shared SQL infrastructure (foreign keys, SQLAlchemy URI mapping, fine-grained lineage in the base class, etc.). These new tests import and exercise `TrinoSource` and `TrinoConfig` directly — they're connector-specific, not shared infrastructure.

— [Original comment](https://github.com/datahub-project/datahub/pull/16292#issuecomment-3972728288) by @skrydal (Feb 27)

The latest diff still has all Trino-specific tests in `tests/unit/test_sql_common.py` with no `test_trino.py` created. Could you move them to a dedicated file?

skrydal:
### Redundant `include_column_lineage` ternary at call sites

`gen_lineage_workunit` already checks `self.config.include_column_lineage` internally (line 386), so the ternary at both call sites is redundant — just pass `schema_metadata` directly.

**`_process_table` (line 448–452):**
```python
# Current:
                yield from self.gen_lineage_workunit(
                    dataset_urn,
                    source_dataset_urn,
                    schema_metadata if self.config.include_column_lineage else None,
                )

# Suggested:
                yield from self.gen_lineage_workunit(
                    dataset_urn,
                    source_dataset_urn,
                    schema_metadata,
                )
```

**`_process_view` (line 483–487):**
```python
# Current:
                yield from self.gen_lineage_workunit(
                    dataset_urn,
                    source_dataset_urn,
                    schema_metadata if self.config.include_column_lineage else None,
                )

# Suggested:
                yield from self.gen_lineage_workunit(
                    dataset_urn,
                    source_dataset_urn,
                    schema_metadata,
                )
```

skrydal:
### Minor: extract shared connector-lineage logic

The post-`super()` block in `_process_table` (lines 436–452) and `_process_view` (lines 472–487) is nearly identical:

```python
        if self.config.ingest_lineage_to_connectors:
            dataset_urn = make_dataset_urn_with_platform_instance(
                self.platform,
                dataset_name,
                self.config.platform_instance,
                self.config.env,
            )
            source_dataset_urn = self._get_source_dataset_urn(
                dataset_name, inspector, schema, table  # or `view`
            )
            if source_dataset_urn:
                yield from self.gen_siblings_workunit(dataset_urn, source_dataset_urn)
                yield from self.gen_lineage_workunit(
                    dataset_urn,
                    source_dataset_urn,
                    schema_metadata if self.config.include_column_lineage else None,
                )
```

The only difference is the `table`/`view` argument to `_get_source_dataset_urn`. Consider extracting a helper:

```python
def _emit_connector_lineage(
    self,
    dataset_name: str,
    inspector: Inspector,
    schema: str,
    entity: str,
    schema_metadata: Optional[SchemaMetadataClass],
    if not self.config.ingest_lineage_to_connectors:
        return
    dataset_urn = make_dataset_urn_with_platform_instance(
        self.platform, dataset_name, self.config.platform_instance, self.config.env,
    )
    source_dataset_urn = self._get_source_dataset_urn(
        dataset_name, inspector, schema, entity
    )
    if source_dataset_urn:
        yield from self.gen_siblings_workunit(dataset_urn, source_dataset_urn)
        yield from self.gen_lineage_workunit(dataset_urn, source_dataset_urn, schema_metadata)
```

Then both overrides reduce to the workunit interception loop + `yield from self._emit_connector_lineage(...)`. Not blocking — just a cleanup suggestion.

PR #16294, opened 2026-02-20
feat(lineage): add manual lineage editing for ML Models and Model Groups

acrylJonny:
## Summary

Adds support for manually editing lineage relationships between ML Models, ML Model Groups, and DataJobs through the UI. This complements the automatic lineage extraction from sources like Vertex AI.

### New feature: ML Model and ML Model Group manual lineage
Adds support for manually editing lineage for `MLModel` and `MLModelGroup` entities via the UI:
- Added in all three lineage versions (v1, v2, v3) to include `MLModel` and `MLModelGroup` as valid upstream/downstream types for `DataJob`, and adds new cases for `MLModel` (upstream: `DataJob`, `DataProcessInstance`) and `MLModelGroup` (upstream / downstream: `DataJob`)
- Backend: Added `updateMlModelLineage` and `updateMlModelGroupLineage` functions that write to the `trainingJobs` and `downstreamJobs` for `MLModelProperties` / `MLModelGroupProperties`
- GraphQL resolver: `mlModel` and `mlModelGroup` entity types are routed to the new service methods

### Fix: deprecated field handling for all lineage entity types
Improves how the `LineageService` interacts with the deprecated list-based lineage fields (`ChartInfo.inputs`, `DashboardInfo.charts/datasets`, `DataJobInputOutput.inputDatasets / inputDatajobs / outputDatasets`) that may have been populated through an ingestion:
- **Dedup**: check both the new edge fields and the deprecated list fields to so that duplicate entries are not created when adding edges
- **Removal**: clears it from the deprecated list field as well (guarded by `hasXxx()` to avoid modifying fields that were never set) when removing an edge
- **No new writes to deprecated fields**: the UI no longer writes to the deprecated fields on add operations; those fields stay as being owned by ingestion


datahub-web-react/src/app/lineage/utils/manageLineageUtils.ts
```diff
@@ -76,7 +76,11 @@ export function getValidEntityTypes(lineageDirection: Direction, entityType?: En
             case EntityType.Dashboard:
                 return [EntityType.Chart, EntityType.Dataset];
             case EntityType.DataJob:
-                return [EntityType.DataJob, EntityType.Dataset];
+                return [EntityType.DataJob, EntityType.Dataset, EntityType.Mlmodel, EntityType.MlmodelGroup];
```
asikowitz: Can you make these changes to lineageV2 and lineageV3 files as well? It should be in `manualLineage/utils.ts` for both directories. We will be removing both lineage (v1) and lineage v2 soon

datahub-web-react/src/app/lineage/utils/manageLineageUtils.ts
```diff
@@ -76,7 +76,11 @@ export function getValidEntityTypes(lineageDirection: Direction, entityType?: En
...
+            case EntityType.Mlmodel:
+                return [EntityType.DataJob];
```
asikowitz: I thought we also support data process instances attached to ml models for mlflow

EDIT: See image below -- data process instance is upstream of ml model
Image

metadata-service/services/src/main/java/com/linkedin/metadata/service/LineageService.java
```diff
@@ -141,7 +141,8 @@ public MetadataChangeProposal buildDatasetLineageProposal(
             opContext,
             Constants.DATASET_ENTITY_NAME,
             downstreamUrn,
-            ImmutableSet.of(Constants.UPSTREAM_LINEAGE_ASPECT_NAME));
+            ImmutableSet.of(Constants.UPSTREAM_LINEAGE_ASPECT_NAME),
+            false);
```
asikowitz, line 145: This is a functional change, as not providing this last parameter defaults it to `true`. Are you confident that this is a safe change?

metadata-service/services/src/main/java/com/linkedin/metadata/service/LineageService.java
```diff
@@ -210,6 +211,7 @@ public void updateChartLineage(
 
   /** Builds an MCP of ChartInfo for chart entities. */
   @Nonnull
+  @SuppressWarnings("deprecation")
```
asikowitz, line 214: Is this necessary? We just removed the deprecated call so idk why this is necessary

metadata-service/services/src/main/java/com/linkedin/metadata/service/LineageService.java
```diff
@@ -352,53 +357,51 @@ private void updateUpstreamCharts(
         upstreamUrnsToAdd.stream()
             .filter(urn -> urn.getEntityType().equals(Constants.CHART_ENTITY_NAME))
             .collect(Collectors.toList());
-    final ChartUrnArray charts = dashboardInfo.getCharts();
     final EdgeArray chartEdges = dashboardInfo.getChartEdges();
 
     final List<Urn> upstreamsChartsToAdd =
-        getUpstreamChartToAdd(upstreamChartUrnsToAdd, chartEdges, charts);
+        getUpstreamChartToAdd(upstreamChartUrnsToAdd, chartEdges, dashboardInfo);
```
asikowitz, line 367: It seems like you're making functional changes to this class to "avoid duplicates with ingestion data". This isn't described in the PR title or description. Can you give an overview of what these changes are trying to accomplish and why we need them? I'm not that familiar with this code so I'm hesitant to make broad changes on it without inspection from someone who understands the code

asikowitz (commented):
Overall the changes seem reasonable. I want us to also make the appropriate frontend changes for v2 and v3 lineage because v1 UI is now deprecated. I also have some questions around the backend changes -- just want to make sure they make sense

datahub-web-react/src/app/lineage/utils/manageLineageUtils.ts
acrylJonny: Done — both files now match v1, including DataProcessInstance as a valid upstream for Mlmodel.

metadata-service/services/src/main/java/com/linkedin/metadata/service/LineageService.java
acrylJonny: The 5-argument form with false bypasses the SystemEntityClient in-process cache and fetches up-to-date data. The 4-argument deprecated overload hardcodes alwaysIncludeKeyAspect=true, which forces inclusion of the key aspect (e.g. datasetKey) in the response even though we never use it in these methods.

metadata-service/services/src/main/java/com/linkedin/metadata/service/LineageService.java
acrylJonny: Yes, it is going to be needed. While the approach doesn't write to the deprecated fields anymore on the add operations, we still read from them (e.g. `chartInfo.getInputs()`, `dashboardInfo.getCharts()`) to avoid creating duplicate edges when ingestion has already populated those fields.

metadata-service/services/src/main/java/com/linkedin/metadata/service/LineageService.java
acrylJonny: Good call, should have been in the description. Essentially what we have changed is:

- No longer writing to deprecated fields on add — the UI only touches the new edge-based fields, leaving the deprecated list fields to ingestion
- Deduplication against deprecated fields — when adding an edge we check both field types to avoid duplicates. There was also a bug here where the old check compared ChartDataSourceType to Urn which would never match which I've fixed
- Removal from deprecated fields — when removing an edge we clear it from the deprecated list too (guarded by hasXxx()), so edges ingested via the old format can be properly removed from the UI

Ive' added tests covering all three behaviours for Chart, Dashboard, and DataJob entities. I'll also update the PR description.

datahub-web-react/src/app/lineage/utils/manageLineageUtils.ts
acrylJonny: Good catch — addressed. `Mlmodel` now accepts `DataJob` and `DataProcessInstance` as valid upstreams in all three lineage versions:

- `datahub-web-react/src/app/lineage/utils/manageLineageUtils.ts` (v1)
- `datahub-web-react/src/app/lineageV2/manualLineage/utils.ts`
- `datahub-web-react/src/app/lineageV3/manualLineage/utils.ts`

The backend `validateMlModelLineageUrns` in `LineageService.java` accepts both `DATA_JOB_ENTITY_NAME` and `DATA_PROCESS_INSTANCE_ENTITY_NAME`. Tests in the v2/v3 `manualLineage/_test_/utils.test.ts` files also exercise `[DATA_JOB, DATA_PROCESS_INSTANCE]` for `Mlmodel` upstream.

PR #16295, opened 2026-02-20
fix(ingest): handle null error_after criteria in dbt source freshness assertions

### LINKED ISSUE #16293: dbt source freshness assertions fail when error_after is not configured
### Description

After upgrading the DataHub CLI from 1.3.1 to 1.4.0, the dbt source ingestion fails due to invalid `assertionInfo` work units generated by the new source freshness assertion feature introduced in #15885.

When a dbt source defines `freshness.warn_after` but **not** `freshness.error_after`, the ingestion produces invalid metadata work units. The `error_after_count` is serialized as the string `'None'` and `error_after_period` as Python `None`, which fails validation.

This causes the source to report failures, which in turn **blocks stateful ingestion stale entity removal** (hard-coded safety check in `stale_entity_removal_handler.py`).

### Steps to Reproduce

1. Define a dbt source with `freshness.warn_after` but no `error_after`:
   ```yaml
   sources:
     - name: my_source
       loaded_at_field: _loaded_at
       freshness:
         warn_after:
           count: 1
           period: day
         # error_after intentionally omitted
   ```

2. Run `datahub ingest` with a dbt recipe that includes `sources_path` pointing to the `sources.json` output

3. Observe failures in the ingestion report

### Actual Behavior

357 assertion work units fail with:
```
'Source produced bad metadata'
ValueError: source produced an invalid metadata work unit
```

The assertion payload shows:
```python
'warn_after_count': '1', 'warn_after_period': 'day',
'error_after_count': 'None', 'error_after_period': None
```

The pipeline then reports:
```
'Skipping stateful ingestion / stale entity removal'
'The soft-deletion of stale entities will be skipped because the source reported a failure.'
```

### Expected Behavior

Sources without `error_after` configured should either:
- Omit the `error_after_count` / `error_after_period` custom properties entirely
- Set them to a valid empty/null representation that passes validation

The freshness assertion should not cause the entire source to report a failure.

### Prior Behavior (CLI 1.3.1)

With CLI 1.3.1 against the same dbt project and server (v1.4.0), the pipeline completed successfully with only warnings, given this feature wasn't there yet?
```
Pipeline finished with at least 3 warnings; produced 1991 events
```

### Environment

- DataHub CLI: 1.4.0
- DataHub Server (GMS): 1.4.0
- dbt: 1.9.4 (Snowflake adapter)
- dbt manifest schema: v12

### Workaround

Remove `sources_path` from the dbt ingestion recipe to disable freshness assertion ingestion entirely.

### Related

Introduced by #15885

danielford831:
## Summary

Fixes #16293

When a dbt source defines `freshness.warn_after` but not `freshness.error_after`, some dbt versions serialize the missing criteria as `{"count": null, "period": null}` rather than `null`. This caused `parse_freshness_criteria` to create a truthy `DBTFreshnessCriteria(count=None, period=None)` object, bypassing the guard in `make_assertion_from_freshness` and producing an invalid `AssertionInfo` work unit with `"None"` string values — failing validation and blocking stale entity removal.

**Fix:** `parse_freshness_criteria` now returns `None` when `count` or `period` is `None`, treating a dict with null fields the same as a missing/null criteria block.

## Changes

- `parse_freshness_criteria` in `dbt_tests.py`: return `None` when `count` or `period` is `None`
- Added `test_parse_freshness_criteria_with_null_fields`: covers all null/empty input edge cases
- Added `test_make_assertion_from_freshness_warn_only`: end-to-end test for a warn-only freshness assertion including serialization

## Test plan

- [x] `test_parse_freshness_criteria_with_null_fields` — verifies null count/period handling
- [x] `test_make_assertion_from_freshness_warn_only` — verifies warn-only assertion is valid and serializable
- [x] Existing freshness tests continue to pass

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_tests.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_tests.py
@@ -72,9 +72,13 @@ def parse_freshness_criteria(data: Optional[Dict]) -> Optional[DBTFreshnessCrite
     """Parse warn_after or error_after criteria dict into DBTFreshnessCriteria."""
     if not data:
         return None
+    count = data.get("count")
+    period = data.get("period")
+    if count is None or period is None:
+        return None
     return DBTFreshnessCriteria(
-        count=data.get("count", 0),
-        period=data.get("period", "hour"),
+        count=count,
+        period=period,
     )
 
 
--- a/metadata-ingestion/tests/unit/dbt/test_dbt_source.py
+++ b/metadata-ingestion/tests/unit/dbt/test_dbt_source.py
@@ -32,6 +32,7 @@ from datahub.ingestion.source.dbt.dbt_tests import (
     DBTFreshnessInfo,
     make_assertion_from_freshness,
     make_assertion_result_from_freshness,
+    parse_freshness_criteria,
 )
 from datahub.metadata.schema_classes import (
     AssertionInfoClass,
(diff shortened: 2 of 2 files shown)


metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_tests.py
AdrianMachado, line 71: Are there any links to dbt docs or their data model or something?

danielford831: Not mega familiar -- these are probably best bet outside of their user-facing docs.

https://github.com/dbt-labs/dbt-core/blob/e6da05fd0f267e238a879b52649363a78823d21b/core/dbt/artifacts/resources/v1/components.py#L174

https://github.com/dbt-labs/dbt-core/blob/e6da05fd0f267e238a879b52649363a78823d21b/schemas/dbt/sources/v3.json#L117

danielford831:
@AdrianMachado I had them all green but I keep getting out of date with base, can you merge when happy pls?

PR #16300, opened 2026-02-20
feat(ingestion/mode): Improve Mode connector performance with threading, SQL caching, and instrumentation

treff7es:
## Summary

Major performance, correctness, and resilience overhaul for the Mode Analytics ingestion connector. Addresses OOM crashes, excessive API calls, broken lineage, and slow ingestion for large Mode workspaces (1000+ reports).

### Performance

- **Multi-threaded report processing**: New `max_threads` config (default 1, max 50) parallelizes per-report API fetching via `ThreadedIteratorExecutor` with thread-safe counters, proper error isolation, and connection pool scaling (`pool_size = max_threads + 10`)
- **Eliminate redundant SQL deepcopy**: Pass `copy=False` to `sqlglot.lineage.lineage()` since the statement is already copied, and swap the selects list in-place in `_get_raw_col_upstreams_for_expression` instead of deep-copying the entire AST — eliminates ~90% of deepcopy calls that were the primary driver of OOM when query parsing timeout is disabled
- **Enable lineage result caching**: Use `create_and_cache_schema_resolver` + `sqlglot_lineage` instead of `create_lineage_sql_parsed_result` which created a new `SchemaResolver` per call, defeating the `_sqlglot_lineage_cached` LRU cache
- **Remove unbounded `lru_cache` on `_get_request_json`**: The `@lru_cache(maxsize=None)` cached every API response forever — with 75k+ unique URLs the cache accumulated hundreds of MBs with zero eviction. Manual caches (`_creator_cache`, `_data_sources_by_id_cache`, `_definitions_map_cache`) already handle repeated lookups
- **Clear SQL parsing caches between spaces**: Global LRU caches (`_parse_statement`, `_sqlglot_lineage_cached`, `try_format_query`) retain large sqlglot ASTs keyed by full SQL strings — for Mode's definition-expanded queries (64 KB+) these ASTs can be 10-100x larger than source SQL, causing memory to grow to ~9 GiB. Caches are now cleared after each space
- **Thread-safe creator cache with `serialized_lru_cache`**: Replaces plain dict with per-key locking so only one thread fetches per unique user href while others wait, eliminating the thundering-herd problem (40,754 duplicate API calls reduced to ~673 in production)
- **Rate limiter lock contention fix**: Release the rate limiter lock before sleeping so other threads aren't blocked during rate limit cooldown
- **Skip chart API calls**: When `explorations_count == 0`, skip unnecessary HTTP requests for queries with no charts
- **Server-side personal collection filtering**: New `exclude_personal_collections` config (default True) uses Mode's `?filter=custom` to avoid fetching personal collections from the API

### Bug Fixes

- **Fix data source ID type mismatch**: Mode API returns string IDs but lookup used int, causing all lineage to be silently skipped
- **Emit table-level lineage even when column-level lineage times out**: Previously `get_upstream_lineage_for_parsed_sql` checked `debug_info.error` (which combines table_error and column_error) and returned early on any error, discarding valid table-level lineage when only column-level lineage failed (e.g. `CooperativeTimeout`)
- **Fix false circular definition detection**: Restructure `_replace_definitions` to use per-definition recursion so sibling definitions don't pollute each other's ancestry chain
- **Handle definitions without aliases** (`{{ @name }}`): Omit the `as` clause to prevent invalid SQL that broke sqlglot parsing
- **Fix definition name regex**: Use `@\w+` instead of `@[a-zA-Z_]+` so definitions like `report_v2` with digits are parsed correctly
- **Add newline before closing paren in definition expansion**: When a definition body ends with a SQL line comment (`--`), the closing paren was swallowed by the comment, leaving unclosed parens that made sqlglot fail with "Expecting )"
- **Fix `logging.debug` -> `logger.debug`** for space skip messages (ING-1689)
- **Fix typo**: `Data Soutce` -> `Data Source` in error message
- **Fix `_is_http_404` helper**: Safely check `error.response` is not None before accessing `status_code`
- **Fix `HTTPError429`/`HTTPError504` re-raise**: Preserve the original response object and exception chain

### Resilience

- **Proactive rate limiter**: 180 req/min default (Mode's limit is ~240/min) to prevent 429 retry spirals
- **Increased retry budget**: `max_attempts: 5→10`, `max_retry_interval: 10→60` for more resilient backoff
- **Error isolation per report**: One failing report doesn't kill the entire ingestion — errors are caught, logged, and reported without aborting
- **Cache-on-success pattern**: Creator, data sources, and definitions caches only store successful results so transient API failures are retried instead of permanently cached as empty
- **Bound `_get_ownership` cache**: `maxsize=None` -> `maxsize=5000`
- **Add missing `cachetools` dependency** for `serialized_lru_cache`

### Code Quality

- **Unified processing loop**: Replace separate `emit_dashboard_mces`, `emit_chart_mces`, `emit_dataset_mces` with single `_emit_workunits_for_space` → `_process_report` flow that processes queries/charts/dashboard in one pass per report
- **O(1) data source lookup**: Replace linear scan of data sources list with dict indexed by ID
- **Thread-safe report counters**: All counter increments and `report_warning`/`report_failure` calls are protected by a lock
- **`_parse_timestamp_ms` helper**: Deduplicate timestamp parsing scattered across methods
- **Type cleanups**: `Optional[bool]` → `bool` for `owner_username_instead_of_email` and `tag_measures_and_dimensions`; `imported_datasets` default `{}` → `[]`
- **`ThreadedIteratorExecutor` improvements**: Proper error propagation from worker threads, `stop_event` for graceful shutdown, drain thread to prevent deadlock on `executor.shutdown()`

### Tests

- **28 unit tests** (`tests/unit/test_mode_source.py`): Definition expansion, query normalization, data source lookup, creator caching, timestamp parsing, chart skip optimization, rate limiting
- **12 integration tests** (`tests/integration/mode/test_mode_threading.py`): Thread-safe mock session, correctness of threaded vs sequential execution, connection pool sizing, wall-clock speedup verification
- **`ThreadedIteratorExecutor` tests** (`tests/unit/utilities/test_threaded_iterator_executor.py`): Error propagation, graceful shutdown
- Golden file comparison validates identical output (lineage, CLL, schemas, containers)

## Test plan

- [x] 28 unit tests pass (`tests/unit/test_mode_source.py`)
- [x] 12 integration tests pass (`tests/integration/mode/test_mode_threading.py`)
- [x] Golden file comparison validates identical output
- [x] `lintFix` passes (ruff + mypy, 0 issues)


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `cae8721`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22992015642)

metadata-ingestion/src/datahub/utilities/ratelimiter.py
```diff
@@ -29,16 +29,16 @@ def __init__(self, max_calls: int, period: float = 1.0) -> None:
         self._lock = threading.Lock()
 
     def __enter__(self) -> "RateLimiter":
+        # Compute sleep time under the lock, but sleep outside it so other
+        # threads aren't blocked while we wait.
+        sleeptime = 0.0
```
askumar27: [Bug] Race condition in rate limiter when used with multiple threads
Sleeping outside the lock allows N threads to all compute the same sleep time, release the lock, sleep concurrently, and then burst simultaneously, defeating rate limiting.

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
```diff
@@ -1364,28 +1368,60 @@ def _get_column_transformation(
...
-    )
-    return OrderedSet(col.table for col in columns_used)
+    elif isinstance(source, sqlglot.optimizer.Scope):
+        if source.scope_type == source.scope_type.UDTF:
+            # UDTF scopes (LATERAL, UNNEST) include correlated references as
+            # direct Table sources, which would pollute results. These are
+            # handled by the dedicated LATERAL section in _list_joins.
+            return OrderedSet()
+        return _collect_tables_from_scope(source, dialect)
+    return OrderedSet()
```
askumar27, line 1424: [Behavior Change] `_get_join_side_tables` now returns an empty set for unresolved sources instead of the previous fallback that used `_get_raw_col_upstreams_for_expression` with a Star column to resolve tables via column lineage tracing.

This affects all connectors using `sqlglot_lineage` (Snowflake, BigQuery, dbt, etc.), not just Mode. Could this cause missing join lineage for edge cases where scope-based resolution fails but the old column-tracing fallback would have succeeded?

[Test Required] CTEs where only some are referenced in the final SELECT

metadata-ingestion/src/datahub/ingestion/source/mode.py
```diff
@@ -1041,69 +1211,90 @@ def construct_query_or_dataset(
             ),
         ).as_workunit()
 
+        data_source_id = query_data.get("data_source_id")
+        if data_source_id is None:
```
askumar27, line 1215: Add logging when `data_source_id` is None

askumar27:
Let's also update `updating-datahub.md`, as this PR has some behavior changes

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
treff7es: The old fallback used `_get_raw_col_upstreams_for_expression` with a synthetic `Star` column to resolve tables via column-lineage tracing. The new
  `_collect_tables_from_scope` walks `scope.sources` recursively — this is actually more correct because it uses sqlglot's own scope resolution rather than an indirect column-tracing hack. The
  edge case where the source isn't in `scope.sources` at all would only happen with broken SQL or incomplete qualification, and the old fallback would likely produce wrong results there anyway.

PR #16310, opened 2026-02-23
feat(ingestion/kafka): add option to disable Avro schema name validation

Devarsh23:
## Summary
                                                                                                                                                                                             
  - Adds `validate_avro_schema_names` config option (default: `True`) to the Kafka source that controls whether Avro schema names/namespaces are validated against the Avro naming           
  specification during parsing                                                                                                                                                               
  - Passes this config through to all `avro.schema.parse()` call sites in the Kafka ingestion pipeline (ConfluentSchemaRegistry and KafkaSource)                                             
  - Also exposes the parameter on the `avro_schema_to_mce_fields()` utility function for callers that pass schema strings directly                                                           
                                                                                                                                                                                             
  ## Motivation                                                                                                                                                                              
                                                                                                                                                                                             
  Debezium and similar CDC connectors embed the Kafka topic name as the Avro schema namespace. When topic names contain hyphens (e.g., `my-debezium-topic`), the resulting namespace         
  (`my-debezium-topic.public.users`) violates the Avro naming spec (`[A-Za-z_][A-Za-z0-9_]*`), causing `avro.schema.parse()` to raise `avro.errors.InvalidName`.

  This prevents DataHub from ingesting schema metadata for these topics, even though downstream consumers like Flink and Kafka Connect handle these schemas without issues.

  ### Why Java consumers work but DataHub fails

  The root cause is a behavioral difference between the Java and Python Avro libraries:

  | Component | Library | Name Validation | Handles hyphens? |
  |---|---|---|---|
  | Confluent SR (reading existing schemas) | Java `org.apache.avro` | `NO_VALIDATION` (`isNew=false`) | Yes |
  | Flink / Kafka Connect | Java `org.apache.avro` | `NO_VALIDATION` (via SR client) | Yes |
  | confluent-kafka Python (ser/de) | `fastavro` | None | Yes |
  | **DataHub ingestion (schema parsing)** | **Python `avro`** | **Strict (`validate_names=True`)** | **No** |

  The Java Confluent Schema Registry client [explicitly disables name
  validation](https://github.com/confluentinc/schema-registry/blob/master/client/src/main/java/io/confluent/kafka/schemaregistry/avro/AvroSchema.java) when reading existing schemas
  (`isNew=false` → `NameValidator.NO_VALIDATION`). The Python `avro` library added the `validate_names` parameter in v1.11.2 ([AVRO-3680](https://issues.apache.org/jira/browse/AVRO-3680)),
  but defaults to `True`.

  DataHub uses the Python `avro` library (not `fastavro`) specifically for parsing schema JSON into structured field objects, which is where the strict validation triggers.

  ### Example error
  avro.errors.InvalidName: my-debezium-topic is not a valid Avro name because it
  does not match the pattern (?:^|.)[A-Za-z_][A-Za-z0-9_]*$

  ### Usage
  ```yaml
  source:
    type: kafka
    config:
      connection:
        bootstrap: "broker:9092"
        schema_registry_url: "http://schema-registry:8081"
      validate_avro_schema_names: false  # Allow hyphens in schema namespaces

  Test plan

  - Added unit test for ConfluentSchemaRegistry._get_schema_fields() with hyphenated namespace and validate_avro_schema_names=False
  - Added unit test verifying default validation still rejects non-compliant names
  - Added end-to-end test for KafkaSource ingestion with Debezium-style schemas (hyphenated namespaces)
  - Added unit test for avro_schema_to_mce_fields() with validate_names=False
  - Verified existing tests are unaffected (default behavior preserved)
  - All code passes ruff check and ruff format

  Checklist

  - The PR conforms to DataHub's Contributing Guideline (particularly PR Title Format)
  - Links to related issues (if applicable) — N/A, new feature
  - Tests for the changes have been added/updated (if applicable)
  - Docs related to the changes have been added/updated (if applicable). Config field includes a detailed description visible in auto-generated docs.
  - For any breaking change/potential downtime/deprecation/big changes an entry has been made in Updating DataHub — N/A, no breaking changes (default behavior unchanged)

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/kafka/kafka_config.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/kafka/kafka_config.py
@@ -72,15 +72,6 @@ class KafkaSourceConfig(
         default=False,
         description="Enables ingesting schemas from schema registry as separate entities, in addition to the topics",
     )
-    validate_avro_schema_names: bool = Field(
-        default=True,
-        description="Whether to enforce Avro schema name validation during parsing. "
-        "Set to False to allow schema names/namespaces that are not valid Avro names "
-        "(e.g., namespaces containing hyphens from Debezium CDC subjects like "
-        "'my-topic.public.users'). By default, the Avro specification requires names "
-        "to match [A-Za-z_][A-Za-z0-9_]*, which rejects hyphens. Disabling this "
-        "validation allows such schemas to be ingested successfully.",
-    )
     external_url_base: Optional[str] = Field(
         default=None,
         description="Base URL for external platform (e.g. Aiven) where topics can be viewed. The topic name will be appended to this base URL.",
--- a/metadata-ingestion/src/datahub/ingestion/source/confluent_schema_registry.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/confluent_schema_registry.py
@@ -318,7 +318,7 @@ class ConfluentSchemaRegistry(KafkaSchemaRegistryBase):
             cleaned_str: str = self.get_schema_str_replace_confluent_ref_avro(schema)
             avro_schema = avro.schema.parse(
                 cleaned_str,
-                validate_names=self.source_config.validate_avro_schema_names,
+                validate_names=False,
             )
 
             # "value.id" or "value.[type=string]id"
--- a/metadata-ingestion/src/datahub/ingestion/source/kafka/kafka.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/kafka/kafka.py
@@ -431,7 +431,7 @@ class KafkaSource(StatefulIngestionSourceBase, TestableSource):
             # DataHub Dataset "description" field is mapped to documentSchema's "doc" field.
             avro_schema = avro.schema.parse(
                 schema_metadata.platformSchema.documentSchema,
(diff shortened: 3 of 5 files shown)


gabe-lyons:
Neat -thanks for the contribution. I'll call on @treff7es to take a look

metadata-ingestion/src/datahub/ingestion/source/kafka/kafka_config.py
```diff
@@ -72,6 +72,15 @@ class KafkaSourceConfig(
         default=False,
         description="Enables ingesting schemas from schema registry as separate entities, in addition to the topics",
     )
+    validate_avro_schema_names: bool = Field(
```
treff7es: The config field is unnecessary. There's no use case for a user wanting strict Avro name validation during ingestion of existing schemas. The Java Schema Registry client already disables it unconditionally when reading. Adding a config option creates complexity for no practical benefit.

What do you think?

Devarsh23: I agree! Let me update the PR to reflect it

Devarsh23: Done!

PR #16319, opened 2026-02-23
feat(authorization): add VIEW_ENTITY_QUERIES privilege for Dataset, Chart, and Data Job

rob-1019:
## Summary

This PR adds two new privileges:

- **`VIEW_ENTITY_QUERIES`** (Dataset, Chart, Data Job): dataset-scoped privilege controlling
  whether a user can see queries and view definitions (SQL logic) for that entity.
- **`VIEW_ALL_QUERIES`**: platform-level privilege granting visibility into every query,
  including ones with no recorded subject dataset, which `VIEW_ENTITY_QUERIES` alone can
  never grant access to.

Before this PR, query visibility was only an implicit side effect of `VIEW_ENTITY_PAGE`, so
there was no way to grant or revoke it independently.

As of this PR, query-page visibility and query-SQL visibility are each governed by their own
privilege: `VIEW_ENTITY_QUERIES` can be granted or revoked independently of `VIEW_ENTITY_PAGE`.
In default mode, holding `VIEW_ENTITY_QUERIES` on a dataset is enough to see `topSqlQueries`
for that dataset too, but `REQUIRE_ALL_SUBJECTS` mode requires holding the privilege on every
dataset a query touches — and since a `topSqlQueries` string carries no such list, only
`VIEW_ALL_QUERIES` can satisfy that mode (see below).

**Existing permissions are unimpacted, for both privileges.** Two one-time upgrade steps run
during GMS's boot/system-update sequence, each scanning every existing policy — shipped
default or customer-edited — and adding its privilege to any policy that already grants
`VIEW_ENTITY_PAGE`:

- `BackfillViewEntityQueriesPrivilege` backfills `VIEW_ENTITY_QUERIES`, toggled off with
  `BOOTSTRAP_SYSTEM_UPDATE_VIEW_ENTITY_QUERIES_PRIVILEGE_ENABLED=false` (disable) /
  `REPROCESS_VIEW_ENTITY_QUERIES_PRIVILEGE=true` (force a re-run).
- `BackfillViewAllQueriesPrivilege` backfills `VIEW_ALL_QUERIES`, toggled off with
  `BOOTSTRAP_SYSTEM_UPDATE_VIEW_ALL_QUERIES_PRIVILEGE_ENABLED=false` (disable) /
  `REPROCESS_VIEW_ALL_QUERIES_PRIVILEGE=true` (force a re-run).

Both are independent, each protected against re-running by its own `DataHubUpgradeResult`
marker so it only runs
once, and both target the same broad set (every policy granting `VIEW_ENTITY_PAGE`) because
that matches actual pre-upgrade behavior: before either privilege existed, page visibility
implied visibility into all of an entity's queries, orphans included. So everyone who could
see queries — any queries, subjects or not — before upgrading can still see them after.

Fresh installs get both privileges via the default policies out of the box. By actor:

- Root: `VIEW_ALL_QUERIES` and `VIEW_ENTITY_QUERIES`
- Admin role: `VIEW_ALL_QUERIES` and `VIEW_ENTITY_QUERIES`
- Editor role: `VIEW_ALL_QUERIES` and `VIEW_ENTITY_QUERIES`
- Reader role: `VIEW_ALL_QUERIES` and `VIEW_ENTITY_QUERIES`
- All users (default, no role assigned): `VIEW_ALL_QUERIES` and `VIEW_ENTITY_QUERIES`
- Resource owners (of the asset): `VIEW_ENTITY_QUERIES` only

Administrators who want to be more granular can remove either privilege from the relevant
policies: removing `VIEW_ENTITY_QUERIES` hides SQL logic for datasets that privilege isn't
held on, while leaving the rest of a user's entity-page access intact; removing
`VIEW_ALL_QUERIES` narrows a user back down to only the queries `VIEW_ENTITY_QUERIES` can
grant per-dataset.

### `VIEW_ENTITY_QUERIES`

The `VIEW_ENTITY_QUERIES` privilege is exposed as `privileges.canViewQueries` in the GraphQL
schema and controls what's shown in the Queries tab and View Definition tab in the UI:
neither tab is ever hidden, but users without the privilege see an explanatory message in
place of the query content. It is enforced server-side on every query read path — GraphQL and REST (OpenAPI
v1/v2/v3, Rest.li) — via `EntityAuthorizationUtils.isAPIAuthorizedEntityUrns`.

A query is typically associated with more than one dataset (e.g. a multi-table `JOIN`), and
two modes govern how many of those datasets the privilege must be held on:

- **Any-subject (default).** A user can read a query if they hold the privilege on **at
  least one** of its associated datasets.
- **All-subjects**, via `QUERY_ENTITY_AUTHORIZATION_REQUIRE_ALL_SUBJECTS=true`. A user must
  hold the privilege on **every** associated dataset — a stricter posture, since a query's
  SQL reveals information about every dataset it touches, not just one.

A query with no associated datasets has nothing to grant access against, so it is hidden
rather than exposed in either mode (re-ingesting the query's dataset associations restores
visibility). Read enforcement can be disabled entirely with
`QUERY_ENTITY_AUTHORIZATION_ENABLED=false`.

The `VIEW_ENTITY_QUERIES` privilege also covers the SQL embedded in dataset usage
statistics: requests selecting `usageStats.buckets[].metrics.topSqlQueries` are denied with
an explicit authorization error when the user lacks the privilege on the dataset, so
consumers can tell "queries hidden" from "no queries ran" without a response-shape change.
Selections without that field still serve the numeric usage data normally, authorized by
`VIEW_DATASET_USAGE` alone as before. In the require-all mode such requests are always
denied — a documented limitation, since the stored statements don't record which other
datasets they reference, so the all-datasets rule can't be verified per statement.

### `VIEW_ALL_QUERIES`

DataHub already has a precedent for this shape of problem: Document entities (bridge to a
source entity that may not resolve) fall back to a platform-level `MANAGE_DOCUMENTS` check
before ever denying view access. `VIEW_ALL_QUERIES` is the equivalent for queries — checked
first in `EntityAspectAuthorizationUtils#filterViewableQueryEntities`, before the per-subject
logic runs, so it grants visibility into every query, including orphans (and even queries
with subjects, where the per-dataset check would otherwise deny). It is deliberately narrow:
never consulted by any query create/edit/delete check, so granting it — even to a read-only
role — cannot confer any write capability. It is not a superset of `VIEW_ENTITY_QUERIES` in
the sense of also granting ordinary per-dataset access; an actor without it still goes
through the normal subject-derived check.

**`VIEW_ALL_QUERIES` is honored everywhere query visibility is decided, not just for Query
entities themselves.** `topSqlQueries` needed it too, since it's a separate decision point
from Query entities — the SQL strings in usage statistics have no entity representation of
their own, so they can't carry subject datasets the way a Query entity does:

- **`topSqlQueries`.** An actor holding `VIEW_ALL_QUERIES` now sees the raw SQL in usage
  statistics unconditionally, including under `REQUIRE_ALL_SUBJECTS`, where every other actor
  — even one holding `VIEW_ENTITY_QUERIES` on the dataset — is denied (see above). This is
  consistent: `VIEW_ALL_QUERIES` grants visibility into every query platform-wide regardless
  of subject datasets, so per-statement dataset association is moot for an actor who already
  holds it.
- **View Definition / Data Transform Logic SQL.** `Dataset.viewProperties.logic`/
  `formattedLogic` and `DataJob.dataTransformLogic.transforms[].queryStatement` are now
  server-enforced the same way — previously these fields returned their SQL text
  unconditionally (gated only by ordinary entity-page visibility), so revoking
  `VIEW_ENTITY_QUERIES` hid the SQL in the View Definition tab's UI but not from a raw
  read of the same fields via GraphQL, OpenAPI (v1/v2/v3), or Rest.li. GraphQL's mappers
  (`DatasetMapper`, `VersionedDatasetMapper`, `DataTransformLogicMapper`) withhold just the
  SQL-bearing fields, keeping `materialized`/`language` visible; OpenAPI/Rest.li have no
  per-field concept in their generic aspect serialization, so those surfaces withhold the whole
  aspect instead via the new `EntityAuthorizationUtils#isQuerySqlAspectRestricted`. Both share
  the same underlying privilege decision — the new
  `EntityAspectAuthorizationUtils#canViewQueriesOnEntity` helper — which folds in the ordinary
  per-dataset `VIEW_ENTITY_QUERIES` check (via `allowedToViewQueriesOnEntity`), the
  `VIEW_ALL_QUERIES` bypass, the `QUERY_ENTITY_AUTHORIZATION_ENABLED` escape valve, and system
  auth, all in one call — before populating the SQL, withholding it — while still returning
  `materialized`/`language`, which aren't sensitive — when the actor lacks it.
- **The Queries/View Definition tab UI flag (`privileges.canViewQueries`).** Also delegates to
  `canViewQueriesOnEntity`, so the tab's client-side gating and the server-side field
  enforcement above can't drift out of sync.

### Known limitations: totals, facets, and browse groups on mixed pages

Every search/scroll/browse surface that can return Query entities now filters denied queries out
of the returned page rather than rejecting the whole request (GraphQL's `listQueries`, OpenAPI
v2/v3, and all relevant Rest.li `EntityResource` actions). Of those, only `listQueries` scrolls to
exhaustion and authorizes the full ordered stream before paginating, so only its `total` is exact.
The other surfaces below still return a count computed over the raw, unfiltered candidate set, or
an aggregate field with the same problem. These are unresolved limitations, not accepted
trade-offs — each should be fixed in a follow-up PR by adopting `listQueries`'s
scroll-to-exhaustion approach; fixing them all here would make an already large PR larger, so
they're listed instead of fixed:

- `GenericEntitiesController.getEntities` (v3 single-entity-type route): `totalCount` is the raw,
  unfiltered candidate count and will include queries the actor can't see once denied ones are
  filtered from the page.
- `EntityController.scrollEntities` (v3 multi-entity-type route): same `totalCount` problem as
  above.
- `EntityController.scrollEntities` (v3 multi-entity-type route): `facets` (from
  `result.getMetadata()`) is also computed over the raw, unfiltered candidate set. This one has no
  clean fix at all without further work — either backend support for authorization-aware
  aggregation, or reimplementing bucketing logic in `scrollEntities` itself.
- `EntityApiDelegateImpl.scroll` (OpenAPI v2 typed delegate): no total/count field exists here, but
  the returned page can come back with fewer entities than requested once denied queries are
  dropped, since nothing backfills the page from further scrolling.
- `EntityResource.search`, `.searchAcrossEntities`, `.scrollAcrossEntities`, `.list`, `.filter`
  (Rest.li): `total`/`numEntities` on the returned `SearchResult`/`ScrollResult` is the raw,
  unfiltered candidate count and will include queries the actor can't see.
- `EntityResource.searchAcrossLineage`, `.scrollAcrossLineage` (Rest.li): same problem as above,
  on `LineageSearchResult`/`LineageScrollResult`.
- `EntityResource.browse` (Rest.li): `groups`/`numGroups`/`numElements` are computed over the raw,
  unfiltered candidate set. Like `scrollEntities`'s facets, this has no clean fix without backend
  support for authorization-aware aggregation, or reimplementing bucketing logic in `browse`
  itself.

Rest.li `autocomplete` is unaffected — its result has no total/count field.

## Test plan

- [x] Unit tests for the UI components that check `canViewQueries`
- [x] Backend upgrade-step unit tests (`BackfillViewEntityQueriesPrivilegeStepTest`,
      `BackfillViewAllQueriesPrivilegeStepTest`)
- [x] Deny/allow authorization tests for each read path (`ListQueriesResolverTest`,
      `QueryTypeTest`, `EntityControllerTest`, `EntitiesControllerTest`,
      `EntityV2ResourceTest`)
- [x] `EntityAspectAuthorizationUtilsTest`: `VIEW_ALL_QUERIES` grants an orphan query in both
      any-subject and strict modes, is a true tier-1 short-circuit (grants even where the
      per-subject check would deny), and — without it — orphan denial is unchanged from before
      (61 tests passing, full class)
- [x] Smoke test exercising both a privileged and unprivileged role
      (`test_view_entity_queries_privilege.py`) — covers grant, revoke, and default-denial

All items below were manually verified against a fresh-nuked local `quickstartDebug` stack at
this PR's tip, using a privileged native test user and an unprivileged one
(`eve@example.com`, created via `/signUp` with no policies beyond the default), plus a
dataset-scoped test policy where noted:

- [ ] Manual: existing policies retain Queries tab access after the backfill — verified via a
      full disable-then-re-enable cycle of `BackfillViewEntityQueriesPrivilege` on a real
      system-update run: with the backfill disabled, a policy lacking the privilege stays
      denied (no upgrade marker written); re-enabling and re-running grants it and restores
      access, including on a policy an admin had deliberately revoked from
- [ ] Manual: a freshly created user gets Queries tab access via the default policy — verified
      (fresh unprivileged user saw both seeded queries with zero extra grants)
- [ ] Manual: removing the privilege from a policy switches the Queries tab to its
      no-permission content — verified via
      `updatePolicy` removing `VIEW_ENTITY_QUERIES` from the all-users policy, both via API
      (`listQueries` → 0) and a watched, human-observed UI pass (Queries tab: admin sees SQL,
      revoked user sees none)
- [ ] Manual: with default settings, a user holding the privilege on one of a multi-dataset
      query's datasets can read the query; a user holding it on none of them cannot — verified
      (dataset-scoped policy on one of two subject datasets → both the single- and
      two-dataset query visible; no grant at all → neither visible)
- [ ] Manual: with `QUERY_ENTITY_AUTHORIZATION_REQUIRE_ALL_SUBJECTS=true`, a user holding the
      privilege on only one of a multi-dataset query's datasets can no longer read the query —
      verified (same dataset-scoped policy; the two-dataset query dropped out of `listQueries`
      while the single-dataset query remained)
- [ ] Manual: with `QUERY_ENTITY_AUTHORIZATION_ENABLED=false`, query reads succeed regardless
      of the privilege (escape valve) — verified, including the orphan-subjects query
- [ ] Manual: without the privilege, a `usageStats` request selecting `topSqlQueries` gets an
      authorization error, while a request without that field returns the usage numbers —
      verified (explicit 403 with an actionable message; separately, a request omitting
      `topSqlQueries` returned `totalSqlQueries` normally)
- [ ] Manual: with `QUERY_ENTITY_AUTHORIZATION_REQUIRE_ALL_SUBJECTS=true`, a `usageStats`
      request selecting `topSqlQueries` is denied even for a user holding the privilege on
      the dataset — verified (denied despite the user holding the privilege specifically on
      that dataset)

The following are new with `VIEW_ALL_QUERIES` (see Summary) and are unit-tested only so far —
manual verification against a live stack is still pending:

- [ ] Manual: a fresh install's ordinary user (default all-users policy, no role assigned)
      can view a query with no recorded subject datasets (orphan query); after an admin
      revokes `VIEW_ALL_QUERIES` from the all-users policy, that same user cannot
- [ ] Manual: on an upgrade, `BackfillViewAllQueriesPrivilege` grants `VIEW_ALL_QUERIES` to
      existing policies that grant `VIEW_ENTITY_PAGE`, restoring orphan-query visibility to
      the same users who had it pre-upgrade
- [ ] Manual: a user holding only `VIEW_ALL_QUERIES` (no other query privilege) can view an
      orphan query but cannot create, edit, or delete it
- [x] Unit tests for the View Definition / Data Transform Logic SQL server-side enforcement:
      `DatasetMapperTest`, `VersionedDatasetMapperTest`, `DataTransformLogicMapperTest` (GraphQL,
      field-level); `EntityAuthorizationUtilsTest` (shared predicate);
      `EntitiesControllerTest`/`EntityControllerTest` (OpenAPI v1/v3, whole-aspect) — added after
      Cursor Bugbot flagged that `viewProperties.logic`/`formattedLogic` and
      `dataTransformLogic[].queryStatement` were unconditional on every one of these surfaces
- [ ] Manual: a raw GraphQL query for `dataset(urn: ...) { viewProperties { logic
      formattedLogic } }` returns `null` for both fields once `VIEW_ENTITY_QUERIES` is revoked
      from a user's policy, and the real SQL once it's restored — confirms the View Definition
      tab's server-side enforcement isn't UI-only
- [ ] Manual: same check for `dataJob(urn: ...) { dataTransformLogic { transforms {
      queryStatement { value } } } }` and `queryStatement`
- [ ] Manual: `GET /openapi/v3/entity/dataset/<urn>/viewProperties` returns 404 once
      `VIEW_ENTITY_QUERIES` is revoked and 200-with-SQL once restored; same check for
      `GET /openapi/v1/entities/latest?urns=...&aspectNames=viewProperties`. OpenAPI v2 and
      Rest.li `EntityV2Resource` share the identical `isQuerySqlAspectRestricted` call but have
      no dedicated automated test for it (v2 has no existing controller test file; Rest.li's
      only exercises the underlying privilege check by reflection) — worth a manual pass here
      specifically.

## Checklist

- [x] The PR conforms to DataHub's [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (especially [PR Title Format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#pr-title-format))


gabe-lyons:
This makes sense, but often permission changes introduce sneaky unexpected issues so we'll want to take a close look here. Also - will this introduce regressions for existing DataHub users who didn't grant this permission before? We definitely don't want to cause this tab to disappear for most installs. cc @chriscollins3456

rob-1019:
@chriscollins3456 Dusting this PR off a bit.

As per Gabe's comments:

> This makes sense, but often permission changes introduce sneaky unexpected issues so we'll want to take a close look here. Also - will this introduce regressions for existing DataHub users who didn't grant this permission before? We definitely don't want to cause this tab to disappear for most installs. cc @chriscollins3456

I think I see TWO issues.

1. On existing installs doing an upgrade to a new version that included this PR the policy this is being added to by default "View Dataset Sensitive Information" is flagged as user-editable so our addition to it in the code will be ignored by upgraders and they won't have VIEW_ENTITY_QUERIES in that policy. As such Readers Role would lose access to view these queries. Editor and Admin roles still have it "accidentally" now because the code also allows `EDIT_ENTITY` (aka `ALL_PRIVLIEGES_GROUP`?) in addition to `VIEW_ENTITY_QUERIES` to see the queries. This is probably fixable by adding another separate new policy for upgraders instead of or in-addition-to re-using the View Dataset Sensitive Information policy.
2. All existing users WITHOUT a role DO currently LOSE their permissions to view these queries. Since we are expectively splitting EDIT_ENTITY in two a migration for upgraders only that found users/groups who effectively held that permission and granted them same-scope access VIEW_ENTITY_QUERIES seems like a (rather complicated) way to do it. Have we done a permission split similar to this before in a way that went well? @chriscollins3456  @gabe-lyons

rob-1019:
Dusting this off again — addressing Gabe's concern above about this introducing a behavior change for existing users.

This now adds a one-time upgrade step (`BackfillViewEntityQueriesPrivilege`, modeled after the existing `BackfillPolicyFields` pattern) that adds `VIEW_ENTITY_QUERIES` to **every** policy that already grants `VIEW_ENTITY_PAGE` — not just the shipped default policies, so it also reaches policies that installs have customized. Anyone in the Editor/Reader/Admin roles, or on any custom policy that already grants `VIEW_ENTITY_PAGE`, keeps their existing access automatically.

To directly answer your question: is there a case where someone who could see a dataset's queries before this lands can no longer see them after? No — because we're adding `VIEW_ENTITY_QUERIES` to every policy that grants `VIEW_ENTITY_PAGE`, not replacing or narrowing anything, so the set of users with query visibility can only stay the same or grow, never shrink, on upgrade.

Customers who *do* want to restrict query visibility going forward can remove `VIEW_ENTITY_QUERIES` from the relevant policy after upgrading — it's independently revocable from `VIEW_ENTITY_PAGE` now, which is the actual point of this PR.

One more thing for internal deployments that already have an equivalent patch applied via some other path: you can avoid the backfill re-doing that work by either marking the upgrade step as already-run in advance (ingest a `dataHubUpgradeResult` aspect with `state: SUCCEEDED` on `urn:li:dataHubUpgrade:BackfillViewEntityQueriesPrivilegeStep` before upgrading), or disable it permanently by setting `BOOTSTRAP_SYSTEM_UPDATE_VIEW_ENTITY_QUERIES_PRIVILEGE_ENABLED=false` on that deployment.

chriscollins3456:
is there any way we can break this down into more manage-able PRs? 7,000+ lines is asking a ton out of one PR and one engineer.

rob-1019:
> is there any way we can break this down into more manage-able PRs? 7,000+ lines is asking a ton out of one PR and one engineer.

It's still 1900 lines when you rip our tests/docs - definately big.

Happy to break this into stacked PRs along the chunk boundaries above if there's firm agreement that it would help. Bottom-ilne is that ~2,000 lines is a lot of code-review.

The reason I'd rather not land pieces of this to master independently: the privilege is called `VIEW_ENTITY_QUERIES`. If the privilege definition and UI hiding land but the Rest.li or OpenAPI enforcement doesn't, the name promises something the system doesn't actually enforce. That could potentially draw a CVE, and while we could probably paper over it with docs as 'intended functionality' and a release note, I'm hesitant to recommend a misleadingly privilege name hit master if we can avoid it.

We could divvy up rougly into these 4 chunks:

- **Core authorization helpers (~410 lines) 
- **The backfill (~625 lines) is a third of the total** but lives in new files under `datahub-upgrade` with no callers in the serving path. It can be reviewed in isolation or by a different person.
- **The Rest.li, OpenAPI, and GraphQL enforcement chunks have quite a few idioms repeated
- ** UI, Config, docs, pro-forma

Glad to do stacked prs or rewrite commits to match above or other reasonable divisons.

benjiaming:
`COMPAT` in docs is not what the code does**

`docs/authorization/policies.md`, `application.yaml`, and `docs/how/updating-datahub.md` say `COMPAT` (the default) keeps **direct Query reads, `listQueries`, REST/OpenAPI, and `topSqlQueries` on any-subject**, and only tightens **search-result / related-entity masking** when `VIEW_AUTHORIZATION_ENABLED` is on.

`requireAllQuerySubjects()` returns `view.isEnabled()` in `COMPAT` mode. `ListQueriesResolver`, `QueryType`, and `filterAPIAuthorizedQueryUrns` all pass that into `filterViewableQueryEntities`. With VBAC on, those paths become **require-all**, not any-subject.

`topSqlQueries` is the real carve-out (`requireAllQuerySubjectsForTopSqlQueries` is true only for literal `TRUE`).

Either change the docs to match “VBAC on ⇒ every Query-entity read is require-all under `COMPAT`”, or change the call sites so only `canViewEntity` uses require-all. As written, an admin who enables VBAC and then revokes `VIEW_ALL_QUERIES` (required for per-dataset restriction to mean anything) will see join queries disappear from the Queries tab, not just from search masking.

benjiaming (approved):
The checklist mentions `test_view_entity_queries_privilege.py)` that is not part of the PR or codebase. Would be good to get this and the other comment fixed while rebasing.

rob-1019:
> The checklist mentions `test_view_entity_queries_privilege.py)` that is not part of the PR or codebase. Would be good to get this and the other comment fixed while rebasing.

Added

rob-1019:
> `COMPAT` in docs is not what the code does**
> 
> `docs/authorization/policies.md`, `application.yaml`, and `docs/how/updating-datahub.md` say `COMPAT` (the default) keeps **direct Query reads, `listQueries`, REST/OpenAPI, and `topSqlQueries` on any-subject**, and only tightens **search-result / related-entity masking** when `VIEW_AUTHORIZATION_ENABLED` is on.
> 
> `requireAllQuerySubjects()` returns `view.isEnabled()` in `COMPAT` mode. `ListQueriesResolver`, `QueryType`, and `filterAPIAuthorizedQueryUrns` all pass that into `filterViewableQueryEntities`. With VBAC on, those paths become **require-all**, not any-subject.
> 
> `topSqlQueries` is the real carve-out (`requireAllQuerySubjectsForTopSqlQueries` is true only for literal `TRUE`).
> 
> Either change the docs to match “VBAC on ⇒ every Query-entity read is require-all under `COMPAT`”, or change the call sites so only `canViewEntity` uses require-all. As written, an admin who enables VBAC and then revokes `VIEW_ALL_QUERIES` (required for per-dataset restriction to mean anything) will see join queries disappear from the Queries tab, not just from search masking.

docs were wrong, fixed

PR #16321, opened 2026-02-23
feat(doc rate limit): Add limits to document ingestion AI-380

nwadams:
AI-380

10k document limit per ingestion run. 
- Error when 10k hit
300 documents embedded per minute rate limit. 
- rate limit to throttle

These limits are put in place to prevent runaway ingestion runs from breaking other functionality on the Datahub Platform

Added some consolidation of logic and refactors to help simplify the code going forward. 

tested locally with wheel

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/confluence/confluence_source.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/confluence/confluence_source.py
@@ -1063,7 +1063,7 @@ class ConfluenceSource(StatefulIngestionSourceBase, TestableSource):
 
                 if (
                     not self.config.advanced.continue_on_failure
-                    and not self.chunking_source.report.num_documents_limit_reached
+                    or self.chunking_source.report.num_documents_limit_reached
                 ):
                     raise
 
--- a/metadata-ingestion/src/datahub/ingestion/source/datahub_documents/datahub_documents_source.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/datahub_documents/datahub_documents_source.py
@@ -13,6 +13,7 @@ Supports both batch (GraphQL) and event-driven (Kafka MCL) modes.
 import hashlib
 import json
 import logging
+from dataclasses import field
 from datetime import datetime
 from pathlib import Path
 from typing import Any, Dict, Iterable, Optional, cast
@@ -61,6 +62,8 @@ class DataHubDocumentsReport(StatefulIngestionReport):
     num_documents_skipped_empty: int = 0
     num_chunks_created: int = 0
     num_embeddings_generated: int = 0
+    num_embedding_failures: int = 0
+    embedding_failures: list[str] = field(default_factory=list)
     processing_errors: list[str] = []
     num_documents_limit_reached: bool = False
 
@@ -219,10 +222,14 @@ class DataHubDocumentsSource(StatefulIngestionSourceBase):
 
     def get_workunits_internal(self) -> Iterable[MetadataWorkUnit]:
         """Main entry point - route to batch or event mode."""
-        if self.config.event_mode.enabled:
-            yield from self._process_event_mode()
-        else:
-            yield from self._process_batch_mode()
+        try:
+            if self.config.event_mode.enabled:
+                yield from self._process_event_mode()
+            else:
+                yield from self._process_batch_mode()
+        except Exception as e:
+            logger.error(f"Failed to run Unstructured pipeline: {e}", exc_info=True)
+            self.report.report_failure(str(e))
 
         # Save state after processing
         if self.config.incremental.enabled:
@@ -960,20 +967,15 @@ class DataHubDocumentsSource(StatefulIngestionSourceBase):
                 document_urn=document_urn, elements=elements
             )
 
-        except RuntimeError as e:
+        except Exception as e:
             if self.chunking_source.report.num_documents_limit_reached:
                 self.report.num_documents_limit_reached = True
                 raise
             error_msg = f"Failed to process document {doc.get('urn', 'unknown')}: {e}"
-            logger.error(error_msg, exc_info=True)
-            self.report.report_error(error_msg)
-        except Exception as e:
-            error_msg = f"Failed to process document {doc.get('urn', 'unknown')}: {e}"
-            logger.error(error_msg, exc_info=True)
-            self.report.report_error(error_msg)
+            logger.warning(error_msg, exc_info=True)
 
     def get_report(self) -> SourceReport:
-        # Forward embedding stats from the chunking sub-component into our report
+        # Forward stats from the chunking sub-component into our report
         self.report.num_documents_processed = (
             self.chunking_source.report.num_documents_processed
         )
@@ -983,6 +985,15 @@ class DataHubDocumentsSource(StatefulIngestionSourceBase):
         )
         self.report.num_documents_limit_reached = (
             self.chunking_source.report.num_documents_limit_reached
+        )
+        self.report.num_embedding_failures = (
+            self.chunking_source.report.num_embedding_failures
+        )
+        self.report.embedding_failures = list(
+            self.chunking_source.report.embedding_failures
+        )
+        self.report.processing_errors = list(
+            self.chunking_source.report.processing_errors
         )
         return self.report
 
--- a/metadata-ingestion/src/datahub/ingestion/source/notion/notion_source.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/notion/notion_source.py
@@ -1629,7 +1629,7 @@ class NotionSource(StatefulIngestionSourceBase, TestableSource):
 
                 if (
                     not self.config.advanced.continue_on_failure
-                    and not self.chunking_source.report.num_documents_limit_reached
+                    or self.chunking_source.report.num_documents_limit_reached
                 ):
                     raise
 
--- a/metadata-ingestion/src/datahub/ingestion/source/confluence/confluence_config.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/confluence/confluence_config.py
@@ -242,10 +242,10 @@ class ConfluenceSourceConfig(
 
     max_documents: int = Field(
         default=10000,
-        gt=-2,
+        ge=-1,
         description="Maximum number of documents to process per ingestion run. "
         "The job will stop and fail with an error once this limit is reached. "
-        "Set to -1 to disable the limit.",
+        "Set to 0 or -1 to disable the limit.",
     )
 
     # Advanced options
--- a/metadata-ingestion/src/datahub/ingestion/source/datahub_documents/datahub_documents_config.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/datahub_documents/datahub_documents_config.py
@@ -128,10 +128,10 @@ class DataHubDocumentsSourceConfig(
 
     max_documents: int = Field(
         default=10000,
-        gt=-2,
+        ge=-1,
         description="Maximum number of documents to process per ingestion run. "
         "The job will stop and fail with an error once this limit is reached. "
-        "Set to -1 to disable the limit.",
+        "Set to 0 or -1 to disable the limit.",
     )
 
     # Partitioning configuration
--- a/metadata-ingestion/src/datahub/ingestion/source/notion/notion_config.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/notion/notion_config.py
@@ -101,10 +101,10 @@ class NotionSourceConfig(
 
     max_documents: int = Field(
         default=10000,
-        gt=-2,
+        ge=-1,
         description="Maximum number of documents to process per ingestion run. "
         "The job will stop and fail with an error once this limit is reached. "
-        "Set to -1 to disable the limit.",
+        "Set to 0 or -1 to disable the limit.",
     )
 
     advanced: AdvancedConfig = Field(
--- a/metadata-ingestion/src/datahub/ingestion/source/unstructured/chunking_config.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/unstructured/chunking_config.py
@@ -417,10 +417,10 @@ class DocumentChunkingSourceConfig(ConfigModel):
     # Processing options
     max_documents: int = Field(
         default=10000,
-        gt=-2,
+        ge=-1,
         description="Maximum number of documents to process per ingestion run. "
         "The job will stop and fail with an error once this limit is reached. "
-        "Set to -1 to disable the limit.",
+        "Set to 0 or -1 to disable the limit.",
     )
     batch_size: int = Field(
         default=10, description="Number of documents to process in parallel"
--- a/metadata-ingestion/src/datahub/ingestion/source/unstructured/chunking_source.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/unstructured/chunking_source.py
@@ -203,7 +203,7 @@ class DocumentChunkingSource(Source):
                 max_calls=config.embedding.documents_per_minute,
                 period=60.0,
             )
-            if config.embedding.rate_limit
+            if self.embedding_model and config.embedding.rate_limit
             else None
         )
 
@@ -248,11 +248,17 @@ class DocumentChunkingSource(Source):
         # Failures are raised directly so the caller can decide how to handle them.
         embeddings = []
         if self.embedding_model:
-            if self.rate_limiter:
-                with self.rate_limiter:
+            try:
+                if self.rate_limiter:
+                    with self.rate_limiter:
+                        embeddings = self._generate_embeddings(chunks)
+                else:
                     embeddings = self._generate_embeddings(chunks)
-            else:
-                embeddings = self._generate_embeddings(chunks)
+                self.report.report_embedding_success()
+            except Exception as e:
+                short_error = str(e).split("\n")[0][:200]
+                self.report.report_embedding_failure(document_urn, short_error)
+                raise
         else:
             logger.debug(
                 f"Skipping embedding generation for {document_urn} - no embedding provider configured"
@@ -265,7 +271,7 @@ class DocumentChunkingSource(Source):
         self.report.report_document_processed(len(chunks))
         self.report.report_embeddings_generated(len(embeddings))
 
-        # Check document limit (max_documents > 0 means the limit is active; -1 disables it)
+        # Check document limit (max_documents > 0 means the limit is active; 0 or -1 disables it)
         if (
             self.config.max_documents > 0
             and self.report.num_documents_processed >= self.config.max_documents
--- a/metadata-ingestion/tests/unit/ingestion/source/confluence/test_confluence_source.py
+++ b/metadata-ingestion/tests/unit/ingestion/source/confluence/test_confluence_source.py
@@ -824,25 +824,13 @@ def test_max_documents_limit_raises_error(
         ingested_ids = {"11111", "22222"}
         parent_ids: set = set()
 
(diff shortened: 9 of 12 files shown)


metadata-ingestion/src/datahub/ingestion/source/datahub_documents/datahub_documents_source.py
```diff
@@ -1148,14 +1169,25 @@ def _generate_embeddings(self, chunks: list[dict[str, Any]]) -> list[list[float]
...
-                response = litellm.embedding(
-                    model=self.embedding_model,
-                    input=batch,
-                    api_key=self.config.embedding.api_key.get_secret_value()
-                    if self.config.embedding.api_key
-                    else None,  # Only used for Cohere
-                    aws_region_name=self.config.embedding.aws_region,  # Only used for Bedrock
-                )
+                if self.rate_limiter:
+                    with self.rate_limiter:
```
jjoyce0510: Is this effectively per-document rate limiting? Is 1 batch == 1 document in this source?

nwadams: It is, I will leave a comment to clarify. We batch, but we never batch across documents.

alexsku:
## CLAUDE: PR Review — Issues, Limitations & Edge Cases

### 🔴 BUG 1 (Critical): Notion source is missing the `max_documents` limit check entirely

The PR adds `max_documents` config to `notion_config.py` and `num_documents_limit_reached` to `notion_report.py`, but **never actually checks the limit** in `notion_source.py`. The diff for `notion_source.py` only adds rate limiter logic — there is no `if report.num_documents_created >= config.max_documents` check anywhere. Compare:

- **Confluence** (`confluence_source.py`): checks `self.report.pages_processed >= self.config.max_documents` → raises `RuntimeError` ✅
- **DataHub Documents** (`datahub_documents_source.py`): checks `self.report.num_documents_processed >= self.config.max_documents` → raises `RuntimeError` ✅
- **Notion** (`notion_source.py`): **no limit check at all** — the config field is dead weight ❌

The Notion tests pass only because they mock `_create_document_entity` with a `side_effect` that manually simulates the limit behavior rather than testing the actual source code path.

---

### 🔴 BUG 2 (Critical): DataHub Documents `RuntimeError` is silently swallowed

In `datahub_documents_source.py`, the limit check raises `RuntimeError` inside `_process_single_document`. But this method wraps everything in a broad `except Exception` that **suppresses the error**:

```python
except Exception as e:
    error_msg = f"Failed to process document {doc.get('urn', 'unknown')}: {e}"
    logger.error(error_msg, exc_info=True)
    self.report.report_error(error_msg)
    # NO re-raise — exception is swallowed!
```

The `RuntimeError` will be caught, logged as a generic processing error, and ingestion will **continue processing the next document**. The flag `num_documents_limit_reached` gets set to `True`, but the run does not actually stop. This completely defeats the purpose of the limit for the DataHub Documents source.

---

### 🟠 BUG 3 (Major): RateLimiter wrapping `yield from` is semantically incorrect

In all three sources, the pattern is:

```python
with self.rate_limiter:
    yield from self.chunking_source.process_elements_inline(...)
```

`RateLimiter.__enter__` sleeps (if needed) then returns. The `yield from` lazily produces work units. `RateLimiter.__exit__` only fires when the **entire generator is exhausted**. This means:

- The "call" timestamp is recorded in `__exit__` after all yields complete, not when the embedding API is actually invoked inside `process_elements_inline`.
- If `process_elements_inline` takes a long time (many chunks → many API calls), the sliding window is inaccurate.
- The rate limiter treats the entire document as a single "call" even though a document can produce many embedding API calls internally.

The config says `documents_per_minute` which is arguably what's being limited, but the actual concern described in the PR ("300 documents embedded per minute rate limit") may not align with what actually happens when chunk counts vary significantly per document.

---

### 🟠 BUG 4 (Major): Confluence limit check + `continue_on_failure=True` doesn't actually stop the run

In `confluence_source.py`, the limit check raises `RuntimeError` from inside `_create_document_entity`. At the call site in `get_workunits_internal`, the error is caught:

```python
except Exception as e:
    ...
    if not self.config.advanced.continue_on_failure:
        raise
```

If `continue_on_failure=True`, the `RuntimeError` is caught, reported, and **ingestion continues to the next page**. The limit is effectively advisory, not enforced. Additionally, `report_text_extracted(len(text))` is never called for the document that hits the limit because the raise happens before it.

---

### 🟡 ISSUE 5 (Medium): No input validation on config fields

The new config fields lack Pydantic validators:

- `max_documents: int = Field(default=10000)` — accepts `0` or negative values. `max_documents=0` triggers the limit on the very first document. `max_documents=-1` never triggers (count ≥ 0 but never ≥ -1).
- `documents_per_minute: int = Field(default=300)` — `0` or negative crashes at `RateLimiter` instantiation with an unhelpful `ValueError`.
- `rate_limit: bool = Field(default=True)` — rate limiting is **on by default**, which changes behavior for existing users on upgrade. Existing fast-running ingestion will suddenly slow down.

Suggest adding `Field(default=10000, gt=0)` or Pydantic validators.

---

### 🟡 ISSUE 6 (Medium): Inconsistent document counter names across sources

Each source uses a different report field for the limit check:

| Source | Counter used |
|--------|-------------|
| Confluence | `self.report.pages_processed` |
| DataHub Docs | `self.report.num_documents_processed` |
| Notion | *(missing entirely)* |

This makes behavior inconsistent — e.g., does a skipped page increment the counter? Hard to reason about without a shared abstraction.

---

### 🔵 ISSUE 7 (Minor): Code duplication — identical config blocks ×3

The same three fields (`max_documents`, `rate_limit`, `documents_per_minute`) with identical defaults and descriptions are copy-pasted into three config classes. Consider extracting into a shared mixin/base class to avoid drift.

---

### 🔵 ISSUE 8 (Minor): Rate limiter initialized when embeddings are disabled

The rate limiter is created unconditionally when `config.rate_limit=True`, even if no embedding provider is configured. In that case, it wraps `process_elements_inline` which may not make any API calls, adding unnecessary sleeping overhead.

---

### 🔵 ISSUE 9 (Minor): Tests mock the behavior instead of testing it

`test_max_documents_limit_raises_error` in `test_datahub_documents_source.py` mocks `_process_single_document` with a `side_effect` that manually raises `RuntimeError` and sets the flag. It never exercises the **actual** code path where the limit check lives. Because of this, the test passes even though Bug #2 (the broad `except` swallowing the `RuntimeError`) means the real code would never stop.

nwadams:
> test_max_documents_limit_raises_error



> ## CLAUDE: PR Review — Issues, Limitations & Edge Cases
> ### 🔴 BUG 1 (Critical): Notion source is missing the `max_documents` limit check entirely
> The PR adds `max_documents` config to `notion_config.py` and `num_documents_limit_reached` to `notion_report.py`, but **never actually checks the limit** in `notion_source.py`. The diff for `notion_source.py` only adds rate limiter logic — there is no `if report.num_documents_created >= config.max_documents` check anywhere. Compare:
> 
> * **Confluence** (`confluence_source.py`): checks `self.report.pages_processed >= self.config.max_documents` → raises `RuntimeError` ✅
> * **DataHub Documents** (`datahub_documents_source.py`): checks `self.report.num_documents_processed >= self.config.max_documents` → raises `RuntimeError` ✅
> * **Notion** (`notion_source.py`): **no limit check at all** — the config field is dead weight ❌
> 
> The Notion tests pass only because they mock `_create_document_entity` with a `side_effect` that manually simulates the limit behavior rather than testing the actual source code path.
> 
This seems wrong. Notion has the check

```
# Check document limit after incrementing count
        if self.report.num_documents_created >= self.config.max_documents:
            self.report.num_documents_limit_reached = True
            raise RuntimeError(
                f"Document limit of {self.config.max_documents} reached. "
                "Increase max_documents in the source config to process more."
            )
```

> ### 🔴 BUG 2 (Critical): DataHub Documents `RuntimeError` is silently swallowed
> In `datahub_documents_source.py`, the limit check raises `RuntimeError` inside `_process_single_document`. But this method wraps everything in a broad `except Exception` that **suppresses the error**:
> 
> ```python
> except Exception as e:
>     error_msg = f"Failed to process document {doc.get('urn', 'unknown')}: {e}"
>     logger.error(error_msg, exc_info=True)
>     self.report.report_error(error_msg)
>     # NO re-raise — exception is swallowed!
> ```
> 
> The `RuntimeError` will be caught, logged as a generic processing error, and ingestion will **continue processing the next document**. The flag `num_documents_limit_reached` gets set to `True`, but the run does not actually stop. This completely defeats the purpose of the limit for the DataHub Documents source.
> 

Thanks. This is super important. Modified logic.

> ### 🟠 BUG 3 (Major): RateLimiter wrapping `yield from` is semantically incorrect
> In all three sources, the pattern is:
> 
> ```python
> with self.rate_limiter:
>     yield from self.chunking_source.process_elements_inline(...)
> ```
> 
> `RateLimiter.__enter__` sleeps (if needed) then returns. The `yield from` lazily produces work units. `RateLimiter.__exit__` only fires when the **entire generator is exhausted**. This means:
> 
> * The "call" timestamp is recorded in `__exit__` after all yields complete, not when the embedding API is actually invoked inside `process_elements_inline`.
> * If `process_elements_inline` takes a long time (many chunks → many API calls), the sliding window is inaccurate.
> * The rate limiter treats the entire document as a single "call" even though a document can produce many embedding API calls internally.
> 
> The config says `documents_per_minute` which is arguably what's being limited, but the actual concern described in the PR ("300 documents embedded per minute rate limit") may not align with what actually happens when chunk counts vary significantly per document.

I think this is ok for now. If we see widely variable document sizes we can adjust to be more fine grained.

> 
> ### 🟠 BUG 4 (Major): Confluence limit check + `continue_on_failure=True` doesn't actually stop the run
> In `confluence_source.py`, the limit check raises `RuntimeError` from inside `_create_document_entity`. At the call site in `get_workunits_internal`, the error is caught:
> 
> ```python
> except Exception as e:
>     ...
>     if not self.config.advanced.continue_on_failure:
>         raise
> ```
> 
> If `continue_on_failure=True`, the `RuntimeError` is caught, reported, and **ingestion continues to the next page**. The limit is effectively advisory, not enforced. Additionally, `report_text_extracted(len(text))` is never called for the document that hits the limit because the raise happens before it.
> 
👍 

> ### 🟡 ISSUE 5 (Medium): No input validation on config fields
> The new config fields lack Pydantic validators:
> 
> * `max_documents: int = Field(default=10000)` — accepts `0` or negative values. `max_documents=0` triggers the limit on the very first document. `max_documents=-1` never triggers (count ≥ 0 but never ≥ -1).
> * `documents_per_minute: int = Field(default=300)` — `0` or negative crashes at `RateLimiter` instantiation with an unhelpful `ValueError`.
> * `rate_limit: bool = Field(default=True)` — rate limiting is **on by default**, which changes behavior for existing users on upgrade. Existing fast-running ingestion will suddenly slow down.
> 
> Suggest adding `Field(default=10000, gt=0)` or Pydantic validators.
> 
👍 

> ### 🟡 ISSUE 6 (Medium): Inconsistent document counter names across sources
> Each source uses a different report field for the limit check:
> 
> Source	Counter used
> Confluence	`self.report.pages_processed`
> DataHub Docs	`self.report.num_documents_processed`
> Notion	_(missing entirely)_
> This makes behavior inconsistent — e.g., does a skipped page increment the counter? Hard to reason about without a shared abstraction.
> 

Agree that we should eventually make a better shared abstraction

> ### 🔵 ISSUE 7 (Minor): Code duplication — identical config blocks ×3
> The same three fields (`max_documents`, `rate_limit`, `documents_per_minute`) with identical defaults and descriptions are copy-pasted into three config classes. Consider extracting into a shared mixin/base class to avoid drift.
> 
> ### 🔵 ISSUE 8 (Minor): Rate limiter initialized when embeddings are disabled
> The rate limiter is created unconditionally when `config.rate_limit=True`, even if no embedding provider is configured. In that case, it wraps `process_elements_inline` which may not make any API calls, adding unnecessary sleeping overhead.

This is probably fine, in practice we should always embed. 

> 
> ### 🔵 ISSUE 9 (Minor): Tests mock the behavior instead of testing it
> `test_max_documents_limit_raises_error` in `test_datahub_documents_source.py` mocks `_process_single_document` with a `side_effect` that manually raises `RuntimeError` and sets the flag. It never exercises the **actual** code path where the limit check lives. Because of this, the test passes even though Bug #2 (the broad `except` swallowing the `RuntimeError`) means the real code would never stop.

nwadams:
@alexsku @jjoyce0510 can you take a look at this PR again. I added a bunch of refactors to consolidate logic from the chunking_source that was duplicated in the datahub_document_source. This also allowed us to avoid tripling the max_documents and rate_limit configs.

metadata-ingestion/src/datahub/ingestion/source/notion/notion_source.py
```diff
@@ -1626,7 +1626,10 @@ def _process_unstructured_output(
                 logger.error(error_msg, exc_info=True)
                 self.report.report_file_failed(error_msg)
 
-                if not self.config.advanced.continue_on_failure:
+                if (
+                    not self.config.advanced.continue_on_failure
+                    and not self.chunking_source.report.num_documents_limit_reached
```
jjoyce0510: Curious - why did we need to add this to the conditional?

metadata-ingestion/src/datahub/ingestion/source/datahub_documents/datahub_documents_source.py
```diff
@@ -1063,154 +953,36 @@ def _process_single_document(
...
-            logger.debug(
-                f"Created {len(chunk_dicts)} chunks using {self.config.chunking.strategy} strategy"
-            )
-            return chunk_dicts
-
         except Exception as e:
-            logger.error(f"Failed to chunk elements: {e}", exc_info=True)
-            return []
-
-    def _generate_embeddings(self, chunks: list[dict[str, Any]]) -> list[list[float]]:
```
jjoyce0510, line 1131: All of this now happens inside chunking source I guess?

rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `b319017`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22369320350)

metadata-ingestion/src/datahub/ingestion/source/confluence/confluence_source.py
```diff
@@ -939,6 +939,15 @@ def _create_document_entity(
...
             )
+        except RuntimeError as e:
+            if self.chunking_source.report.num_documents_limit_reached:
+                self.report.num_documents_limit_reached = True
+                raise
+            logger.warning(
+                f"Failed to generate embeddings for {document_urn}: {e}. "
+                f"Document will be ingested without embeddings.",
+                exc_info=True,
+            )
```
alexsku, line 951: this is a design decision, right? I think Shirshanka mentioned this some time ago

jjoyce0510: If it fails, we want to ingest with keyword support!

metadata-ingestion/src/datahub/ingestion/source/datahub_documents/datahub_documents_source.py
nwadams: that is correct.

metadata-ingestion/src/datahub/ingestion/source/notion/notion_source.py
nwadams: if we hit max_documents we want to stop processing, even if the `continue_on_failure` flag is set to true.

metadata-ingestion/src/datahub/ingestion/source/confluence/confluence_source.py
nwadams: verified that it continues to ingest with keyword support when it fails (except for hitting max documents which hard fails)

alexsku:
## CLAUDE: Follow-up Review (v2)

Good refactoring — consolidating limits, rate-limiting, and embedding logic into `DocumentChunkingSource` fixes the majority of the original issues (1–4, 6–8). Here are the remaining and new issues:

---

### 🟠 1. Confluence/Notion: `continue_on_failure` logic is inverted for the limit error

Both Confluence and Notion have this at the outer call site:
```python
if (
    not self.config.advanced.continue_on_failure
    and not self.chunking_source.report.num_documents_limit_reached
):
    raise
```

When `num_documents_limit_reached=True`, the condition becomes `not X and not True` → always `False` → **the exception is never re-raised**. The `RuntimeError` gets caught, logged as a page failure, and ingestion continues to the next page.

Should be `or` instead of `and`:
```python
if (
    not self.config.advanced.continue_on_failure
    or self.chunking_source.report.num_documents_limit_reached
):
    raise
```

---

### 🟠 2. `max_documents` validation allows `0` — ambiguous behavior

```python
max_documents: int = Field(default=10000, gt=-2, ...)
```

`gt=-2` allows `0`, but the limit check is `if self.config.max_documents > 0 ...` — so `0` silently disables the limit, same as `-1`. Confusing since `-1` is the documented "disable" value. Should use `ge=-1` or a validator that only allows `-1` or positive integers.

---

### 🟠 3. `DataHubDocumentsReport.get_report()` doesn't forward embedding failure stats

Confluence and Notion both copy these fields from `DocumentChunkingReport`:
- `num_documents_with_embeddings`
- `num_embedding_failures`
- `embedding_failures` (list)

DataHub Documents `get_report()` only copies `num_documents_processed`, `num_chunks_created`, `num_embeddings_generated`, and `num_documents_limit_reached`. The `DataHubDocumentsReport` class doesn't even *have* the embedding failure fields. Users will have no visibility into embedding failures for this source.

---

### 🟡 4. Embedding failures in inline mode now crash the caller (behavior change)

In v1, `process_elements_inline` caught embedding failures internally and logged warnings — documents were ingested without embeddings. In v2, embedding failures propagate as raw exceptions. Confluence and Notion handle this correctly (catch + warn), but in DataHub Documents, embedding failures hit the generic `except Exception` → reported as **errors** via `report.report_error()` rather than warnings. Previously these were silent successes (document ingested, embeddings skipped).

---

### 🟡 5. Tests still mock `process_elements_inline` with fake reimplementation of limit logic

All three test suites use a `fake_process_inline` that manually reimplements the limit check (counter increment → compare to max → raise RuntimeError) rather than exercising the real `process_elements_inline`. If the real implementation drifts, these tests would still pass. Consider at least one integration-style test that calls the real `process_elements_inline` (mocking only `_generate_embeddings`).

---

### 🔵 6. `processing_errors: list[str] = []` — mutable default in `DataHubDocumentsReport`

Pre-existing issue, but now visible next to `DocumentChunkingReport` which correctly uses `field(default_factory=list)`. Low impact since the class is single-instance per run, but inconsistent.

nwadams:
@alexsku @jjoyce0510 updated with latest fixes from PR comments.

metadata-ingestion/src/datahub/ingestion/source/datahub_documents/datahub_documents_source.py
```diff
@@ -1063,154 +961,40 @@ def _process_single_document(
...
+        self.report.num_documents_limit_reached = (
+            self.chunking_source.report.num_documents_limit_reached
         )
-
-        logger.debug(
-            f"Emitting SemanticContent for {document_urn} with {len(chunks)} chunks"
+        self.report.num_embedding_failures = (
+            self.chunking_source.report.num_embedding_failures
+        )
+        self.report.embedding_failures = list(
```
jjoyce0510, line 992: nice this is useful..

PR #16327, opened 2026-02-24
fix(ingestion/s3): disable Avro schema name validation in S3 connector

treff7es:
## Summary

- Adds `validate_names` parameter to `avro_schema_to_mce_fields()` utility function, allowing callers to control Avro schema name validation
- S3 `AvroInferrer` passes `validate_names=False` to prevent failures on Avro files with non-compliant names (e.g., hyphens in Debezium-style namespaces)
- Prevents silent schema corruption in multi-file batches where an Avro parse failure would cause subsequent files to get empty `fields: []`

## Motivation

The Python `avro` library (v1.12.0) defaults to `validate_names=True` in `avro.schema.parse()`, rejecting schema names that don't match `[A-Za-z_][A-Za-z0-9_]*`. The S3/data lake connector encounters arbitrary files and should be permissive — matching the behavior of the Java Confluent Schema Registry client which [explicitly disables name validation](https://github.com/confluentinc/schema-registry/blob/master/client/src/main/java/io/confluent/kafka/schemaregistry/avro/AvroSchema.java) when reading existing schemas.

## Why it is safe

  - DataHub's job is to catalog what exists, not enforce Avro naming conventions. If a schema is running in production, DataHub should ingest it.                                                                         
  - The Java ecosystem already does this — Confluent SR uses NO_VALIDATION for existing schemas. The Python avro library being strict is the outlier.
  - The current avro default is dangerous — with validate_names=True + swallow_exceptions=True, non-compliant schemas silently return fields: [], which is data loss. That's worse than accepting a hyphenated name.

Companion to #16310 which adds the same parameter for the Kafka connector path.

## Test plan

- [x] All 88 S3 integration tests pass (`pytest tests/integration/s3/test_s3.py`)
- [x] All 28 schema_util unit tests pass (`pytest tests/unit/test_schema_util.py`)
- [x] `ruff check` and `mypy` pass on changed files


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `a68bb2f`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22346166643)

PR #16334, opened 2026-02-24
fix(ui): prevent document editor toolbar from disappearing when click…

annadoesdesign:
https://github.com/user-attachments/assets/cac95165-52aa-415d-9bda-11d88926cf60

Clicking the font size or heading type dropdowns in the document editor toolbar caused the entire toolbar to vanish because the editor lost focus to a non-focusable div, triggering the blur handler that hides the toolbar.

- Add onMouseDown preventDefault on Toolbar container to keep editor focus
- Defer blur check with requestAnimationFrame to handle portal-rendered dropdowns
- Add commands.focus() in HeadingMenu after toggling to match FontSizeSelect

Fixes CAT-1392

### DIFF
--- a/datahub-web-react/src/app/entityV2/shared/EntityDropdown/AnnounceMenuAction.tsx
+++ b/datahub-web-react/src/app/entityV2/shared/EntityDropdown/AnnounceMenuAction.tsx
@@ -1,0 +1,34 @@
+import { Tooltip } from '@components';
+import { MegaphoneSimple } from '@phosphor-icons/react';
+import React, { useState } from 'react';
+
+import { useEntityData, useRefetch } from '@app/entity/shared/EntityContext';
+import { ActionMenuItem } from '@app/entityV2/shared/EntityDropdown/styledComponents';
+import CreateEntityAnnouncementModal from '@app/entityV2/shared/announce/CreateEntityAnnouncementModal';
+
+export default function AnnounceMenuAction() {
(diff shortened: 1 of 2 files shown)


datahub-web-react/src/app/entityV2/document/summary/EditableContent.tsx
```diff
@@ -306,11 +306,21 @@ export const EditableContent: React.FC<EditableContentProps> = ({
...
-                        handleBlur();
-                    }
+                    const currentTarget = e.currentTarget;
+                    // Delay the check to let focus settle — dropdown overlays
+                    // render in portals outside this DOM subtree, so relatedTarget
+                    // can be null or outside the editor section even when the user
+                    // is still interacting with toolbar dropdowns.
+                    requestAnimationFrame(() => {
+                        const stillInsideEditor = currentTarget.contains(document.activeElement);
+                        const hasOpenToolbarDropdown = !!document.querySelector(
```
chriscollins3456: hmm we don't typically use document selectors like this. 

this whole thing feels kind of hacky. 

i'm curious - why do we need `handleBlur` and `setIsEditorFocused(false)`? does `handleBlur` have other side effects?

i wonder if we can use `ClickOutside` here in any way with `ignoreSelector` or `ignoreWrapper` props to ignore clicking on the toolbar and its dropdowns?

PR #16335, opened 2026-02-24
fix(sql-parser): migrate from sqlglot[rs] to sqlglot[c] with runtime patches

sgomezvillamor:
## Summary

This PR migrates from sqlglot[rs] (Rust tokenizer, deprecated) to sqlglot[c] (C tokenizer) while maintaining compatibility with our patchy-based patches.

## Key Changes

**Patching approach:**
- Continue using `patchy` for most sqlglot patches (traverse, decorrelate, lineage, to_node)
- Use a simple wrapper function for `__deepcopy__` instead of patchy

**Why the wrapper for __deepcopy__?**
In sqlglot[c], `expression_core.py` is mypyc-compiled to a `.so` file. This makes `Expression.__deepcopy__` unpatchable with patchy because patchy operates on Python source code, which isn't available for mypyc-compiled modules.

The wrapper approach:
1. Store the original `__deepcopy__` method reference
2. Create a wrapper that calls cooperative timeout, then delegates to the original
3. Replace the method directly via assignment

This is cleaner and more maintainable than full function replacement.

## Testing

- ✅ All 84 SQL parsing lineage tests pass
- ✅ Linting and mypy checks pass
- ✅ Updated golden file for join ordering (semantically equivalent, different order)

## Migration Path

This PR is stacked on top of #16318 (fix-parsing-semantic-views branch) which already contains the v29 migration work. The changes here are minimal:
- Replace patchy patch for `__deepcopy__` with wrapper function
- Add documentation explaining the approach


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `b4c1eb4`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22393222784)

PR #16336, opened 2026-02-24
fix(ui): propagate validator plugin error messages to the UI

rob-1019:
## Summary
- Validator rejections (custom plugin or built-in) now reach the UI as their actual
  message instead of a generic "Something went wrong" toast
- They also get correct API classification: 400 BAD_REQUEST (403 UNAUTHORIZED for
  authorization-subtype failures) instead of a generic 500, with a clean message
  instead of a Java object dump
- Scope is deliberately narrow: the UI surfaces server messages ONLY for BadRequests
  that result from validation exceptions. All other paths (plain 400s, 403s, 5xx)
  keep today's generic toasts.

## Problem
When an AspectPayloadValidator rejects a write, the message died twice on the way out:
1. The structured-property resolvers caught the exception and re-wrapped it in
   `RuntimeException("Failed to perform update against input <entire input>")`
2. `DataHubDataFetcherExceptionHandler` had no case for metadata-io's
   `ValidationException`, so it fell through to the RuntimeException branch: a 500
   whose message was the Lombok dump of `ValidationExceptionCollection` (including
   the submitted payloads), with the validator's one useful sentence buried inside.
The frontend renders 500s as a canned toast, so users saw "An unexpected error
occurred" no matter what the validator said.

**Frontend**
- `handleGraphQLError`: on 400 with `errorSource === 'VALIDATION'`, show the server
  message (fallback: caller's badRequestMessage/defaultMessage). All other codes are
  unchanged - non-validation 400s, 403s, and 5xx still show only caller-supplied text
- `handleBatchError` (both entity/ and entityV2/ copies): same VALIDATION-gated rule
  for bulk-action 400s
- `EditColumn` structured-property remove path now routes through `handleGraphQLError`
  (it previously swallowed the error entirely and showed nothing actionable)

## What API clients see
- Validator rejections: 400 (or 403) + `extensions.errorSource=VALIDATION` + clean
  message, instead of 500 + object dump. No other response changes.
- Note: multi-item batch failures no longer carry per-item urn/aspect attribution in
  the message text (it was only ever inside the unparseable toString dump; it remains
  in server logs). A structured `extensions.validationErrors` array is a possible
  follow-up if attribution is needed programmatically.

## Test plan
- [x] `DataHubDataFetcherExceptionHandlerTest`: VALIDATION marker on 400s;
      AUTHORIZATION subtype -> 403 without the marker
- [x] `handleGraphQLError.test.ts`: server message shown only for VALIDATION 400s;
      generic behavior preserved for plain 400/403/5xx/unknown
- [x] `utils.test.ts` (entityV2): handleBatchError VALIDATION gate + bulk-403 string
- [x] `EditColumn.remove.test.tsx`: remove-path error surfaces via handleGraphQLError
- [x] Manual: mounted (not compiled-in) validator plugin reading a live message file;
      verified add/edit/remove of structured properties on a dataset shows the file's
      message verbatim in the toast, and succeeds when the file is absent/blank

## Checklist


datahub-web-react/src/app/entityV2/shared/utils.ts
```diff
@@ -144,6 +144,18 @@ function getGraphqlErrorCode(e) {
     return undefined;
 }
 
+/**
+ * Extracts the actual error message from GraphQL errors.
+ * This is where validator plugin messages are found.
+ */
+function getGraphqlErrorMessage(e): string | undefined {
```
gabe-lyons: will this affect any other errors?

gabe-lyons: im a bit confused here.. do we not already have some utility for this? how is this done elsewhere in the app?

datahub-web-react/src/app/shared/handleGraphQLError.ts
```diff
@@ -22,18 +22,33 @@ export default function handleGraphQLError({
     message.destroy();
     const { graphQLErrors } = error;
     if (graphQLErrors && graphQLErrors.length) {
-        const { extensions } = graphQLErrors[0];
+        const { extensions, message: serverMessage } = graphQLErrors[0];
```
gabe-lyons, line 26: im confused again here... why is this case different than other cases?

entity-registry/src/main/java/com/linkedin/metadata/aspect/plugins/validation/ValidationExceptionCollection.java
```diff
@@ -97,7 +97,11 @@ public String toString() {
             .map(
                 e ->
                     String.format(
-                        "EntityAspect:%s Exceptions: %s", e.getKey().toString(), e.getValue()))
+                        "EntityAspect:%s: %s",
+                        e.getKey().toString(),
+                        e.getValue().stream()
+                            .map(AspectValidationException::getMsg)
+                            .collect(Collectors.joining("; "))))
```
gabe-lyons: are there risks of NPEs here? this seems to assume a lot of non-nulls

gabe-lyons:
seems directionally correct, but im confused why some core code paths need to be impacted for this case

datahub-web-react/src/app/shared/handleGraphQLError.ts
rob-1019: Other errors make do with a "good-enough" general error message thats tied to the apollo client error type and ignore the serverMessage thats returned. The Apollo catches the validation error and throws it as a ErrorCodes.BadRequest an give a generic uhoh message that doesn't even hint that it was caused by a validation error AND it definately doesn't show the actual validation message text. Things like permission errors work just off hte HTTP 403 response code and use the same error message for every 403 but we specifically want to show a validator-specific error message.

datahub-web-react/src/app/entityV2/shared/utils.ts
rob-1019: I'd anticipate it does affect anything that throws a BadRequest, ServerError, or an unknown error code that provides a serverMessage to show the explicitly passed error message.

As initially proposed users would get lower-level more technical error messages from a variety of error messages. I'll update to target to only BadRequests that result in Validation Exceptions. There may be other areas where we might want to expose deeper error messages but that doesn't need to be part of this line of work.

entity-registry/src/main/java/com/linkedin/metadata/aspect/plugins/validation/ValidationExceptionCollection.java
rob-1019: yep, will address

PR #16339, opened 2026-02-25
feat(ingestion): migrate setup.py dependencies to pyproject.toml (PEP621)

kyungsoo-datahub:
- Generates `pyproject.toml` dependency sections from `setup.py` via
  `scripts/generate_pyproject_deps.py`
  - Flattens all Python set math (`sql_common | aws_common | ...`) into fully
  inlined arrays per plugin — no self-referencing extras leaked to PyPI
  - Deduplicates overlapping specifiers to narrowest constraint (e.g.
  `pandas<2.2.0` + `pandas<3.0.0` → `pandas<2.2.0`)
  - Adds `verifyPyprojectSync` Gradle task (120 equivalence checks) wired into
  `check` and `buildWheel`
  - Adds `generatePyprojectDeps` Gradle task for regeneration
  - Pins `setuptools<82` for runtime deps that import `pkg_resources` (mlflow,
  stopit)
  - `setup.py` remains the editing source for now; `pyproject.toml` is the
  generated output

### DIFF
--- a/metadata-ingestion/pyproject.toml
+++ b/metadata-ingestion/pyproject.toml
@@ -760,7 +760,7 @@ metabase = [
 
 mlflow = [
     "mlflow-skinny>=2.3.0,<2.21.0",
-    "setuptools",
+    "setuptools<82",
 ]
 
 mode = [
@@ -910,7 +910,7 @@ powerbi = [
     "more-itertools<11.0.0",
     "msal>=1.31.1,<2.0.0",
     "patchy==2.8.0",
-    "setuptools",
+    "setuptools<82",
     "sqlglot[c]==29.0.1",
     "sqlparse<1.0.0",
     "stopit==1.1.2",
@@ -1417,7 +1417,7 @@ all = [
     "requests_ntlm<2.0.0",
     "schwifty<2026.0.0",
     "scipy>=1.7.2,<2.0.0",
-    "setuptools",
+    "setuptools<82",
     "simple-salesforce<2.0.0",
     "slack-sdk==3.18.1",
     "smart-open[azure,gcs,s3]>=5.2.1,<8.0.0",
--- a/metadata-ingestion/setup.py
+++ b/metadata-ingestion/setup.py
@@ -393,7 +393,8 @@ threading_timeout_common = {
     "stopit==1.1.2",
     # stopit uses pkg_resources internally, which means there's an implied
     # dependency on setuptools.
-    "setuptools",
+    # setuptools 82 removed pkg_resources.
+    "setuptools<82",
 }
 
 abs_base = {
@@ -668,7 +669,8 @@ plugins: Dict[str, Set[str]] = {
         # Upper bound can be removed once the upstream issue is resolved,
         # or we have a reliable and backward-compatible way to handle prompt filtering.
         # It's technically wrong for packages to depend on setuptools. However, it seems mlflow does it anyways.
-        "setuptools",
+        # setuptools 82 removed pkg_resources, which mlflow uses at runtime.
+        "setuptools<82",
     },
     "datahub-debug": {"dnspython==2.7.0", "requests<3.0.0"},
     "datahub-documents": unstructured_lib,


rajatoss:
## Connector Tests Results

**Connector tests failed** for commit `dd80167`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22862214774)

sgomezvillamor:
I've been looking into this and basically found same options you found too:

> If your goal is to get rid of setup.py entirely, your real options are:
> - Accept duplication — inline all deps into `pyproject.toml`, no shared groups
> - Accept public internal extras — use the self-referencing trick in `pyproject.toml` and just prefix internal extras with an underscore or _internal- as a convention to signal they're not meant for end users
> - Use a code generation script — have a script that reads your private groups and writes the final `pyproject.toml`, which you commit and treat as generated. You maintain the script, not the `toml` directly.

So far, option 3 implemented in the proposal (`setup.py` as source of thruth and generate `pyproject.toml` automatically) seems to be best option.

However, I would like to understand what's exactly the requirement on having TOML file. Depending on that requirement, I also found that we can have a hybrid approach with `pyproject.toml` + `setup.cfg `to have the static metadata and keep `setup.py` for the dynamic extras logic:

> The key split is:
> 
> * `pyproject.toml` → build system declaration + tool configs (`ruff`, `mypy`, `pytest`, etc.)
>
> ```
> [build-system]
> requires = ["setuptools>=61"]
> build-backend = "setuptools.build_meta"
> 
> [tool.ruff]
> line-length = 88
> 
> [tool.mypy]
> python_version = "3.10"
> 
> [tool.pytest.ini_options]
> testpaths = ["tests"]
> ```
> 
> * `setup.cfg` → static metadata (`name`, `version`, `description`, base `install_requires`)
> 
> ```
> [metadata]
> name = your-package
> version = attr: your_package.__version__
> description = Your package description
> long_description = file: README.md
> long_description_content_type = text/markdown
> author = Your Name
> author_email = you@example.com
> license = Apache-2.0
> requires-python = >=3.8
> 
> [options]
> packages = find:
> python_requires = >=3.8
> install_requires =
>     some-core-dep>=1.0.0
> 
> [options.packages.find]
> where = src
> ```
> 
> * `setup.py` → only the dynamic extras logic that needs Python set operations
>
> ```
> from setuptools import setup
> 
> # --- Reusable private dependency groups ---
> path_spec_common = {
>     "pathspec>=0.9.0",
>     "parse>=1.19.0",
> }
> 
> bigquery_common = {
>     "google-cloud-logging<=3.5.0",
>     "google-cloud-bigquery<4.0.0",
>     "google-cloud-datacatalog>=1.5.0,<4.0.0",
>     "google-cloud-resource-manager<2.0.0",
>     "more-itertools>=8.12.0,<11.0.0",
>     "sqlalchemy-bigquery>=1.4.1,<2.0.0",
>     *path_spec_common,
> }
> 
> # --- Plugin definitions ---
> plugins = {
>     "bigquery": bigquery_common | {"bigquery-specific-dep>=1.0"},
>     "snowflake": {
>         "snowflake-connector-python>=3.0.0",
>         "snowflake-sqlalchemy>=1.4.0",
>     },
> }
> 
> all_exclude_plugins = {"some-heavy-plugin"}
> 
> dev_requirements = {
>     "pytest>=7.0",
>     "pytest-mock>=3.0",
> }
> 
> lint_requirements = {
>     "ruff>=0.1.0",
>     "mypy>=1.0",
> }
> 
> # --- Only extras_require here, everything else is in setup.cfg ---
> setup(
>     extras_require={
>         "bigquery": list(bigquery_common),
>         **{
>             plugin: list(dependencies)
>             for plugin, dependencies in plugins.items()
>         },
>         "all": list(
>             set().union(
>                 *[
>                     deps
>                     for plugin, deps in plugins.items()
>                     if plugin not in all_exclude_plugins
>                 ]
>             )
>         ),
>         "dev": list(dev_requirements),
>         "lint": list(lint_requirements),
>     }
> )
> ```
> 
> Setuptools merges all three automatically

It seems this hybrid approach is what many large projects (including `pip` itself) do when `extras` logic gets complex. Would that approach cover the TOML requirement?

sgomezvillamor:
I miss a new or existing gradle task that updates the pyproject.toml file
Also we need a new or existing gradlew task that verfyies that both file are in-sync, and that task should be included in the CI

PR #4793, opened 2022-04-29
feat(blame) - add schema history blame UI

aditya-radhakrishnan:
## Checklist
- [x] The PR conforms to DataHub's [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [Commit Message Format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#commit-message-format))
- [x] Links to related issues (if applicable)
- [x] Tests for the changes have been added/updated (if applicable)
- [x] Docs related to the changes have been added/updated (if applicable). If a new feature has been added a Usage Guide has been added for the same.
- [x] For any breaking change/potential downtime/deprecation/big changes an entry has been made in [Updating DataHub](https://github.com/datahub-project/datahub/blob/master/docs/how/updating-datahub.md)

PR #4687, opened 2022-04-18
feat(gms): Entity change events v1 (Platform Event)

jjoyce0510:
**Summary**

This PR introduces a new Platform Event called **entityChangeEvent** which is a high-level event emitted to Kafka when an entity on DataHub changes. Change events are emitted in the following scenarios:

1. Tag added / removed from an entity.
2. Glossary term added / removed from an entity.
3. Domain added / removed from an entity.
4. Schema fields added / removed from an entity. 
5. New entity created. 
6. Entity hard / soft deleted.
7. Entity deprecation status changes. 

To do so, we reuse a slightly modified version of the Differ interface provided as part of the timeline API. We've gone ahead and added a few new differs related to domain, soft deleted status, and more. 

**Changes**

- Adding Audit Stamp to MCL
- Minor refactoring (naming) of the Timeline abstractions. (see ChangeEvent.java)
- New Platform Event registered in the entity registry. See EntityChangeEvent.pdl! 
- Adding strongly-typed Change Events to Timeline API. 

**Status**
Ready

## Checklist
- [ ] The PR conforms to DataHub's [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [Commit Message Format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#commit-message-format))
- [ ] Links to related issues (if applicable)
- [ ] Tests for the changes have been added/updated (if applicable)
- [ ] Docs related to the changes have been added/updated (if applicable). If a new feature has been added a Usage Guide has been added for the same.
- [ ] For any breaking change/potential downtime/deprecation/big changes an entry has been made in [Updating DataHub](https://github.com/datahub-project/datahub/blob/master/docs/how/updating-datahub.md)

PR #16342, opened 2026-02-25
feat(timeline): add Change History sidebar for Dataset, Domain, Glossary Term, and Data Product entities

max-datahub:
## Summary

Extends the Change History sidebar to cover **9 change categories** across **4 entity types**: Dataset, Glossary Term, Domain, and **Data Product**. Each entity type's category filter dropdown only shows the event types that are actually supported by the backend. Data Products support **asset membership tracking** — when datasets or other assets are added to or removed from a Data Product.

### Breaking Changes

- **`ChangeCategory.OWNER` renamed to `OWNERSHIP`** in the GraphQL enum. Clients querying the timeline/change history API with `OWNER` must update to `OWNERSHIP`. The REST API (`/timeline`) and the Python CLI accept both values for backward compatibility. This fixes an accidental 2022 divergence between the backend enum and the GraphQL schema. Documented in `docs/how/updating-datahub.md`.

### Supported event types per entity

| Event Type | Dataset | Glossary Term | Domain | Data Product |
|---|:---:|:---:|:---:|:---:|
| Schema changes | Yes | N/A | N/A | N/A |
| Tags | Yes | N/A | N/A | Yes |
| Glossary terms | Yes | Yes (related terms) | N/A | Yes |
| Documentation | Yes | Yes | Yes | Yes |
| Ownership | Yes | Yes | Yes | Yes |
| Domains | Yes | Yes | N/A | Yes |
| Structured properties | Yes | Yes | Yes | Yes |
| Applications | Yes | Yes | N/A | Yes |
| Asset membership | N/A | N/A | N/A | Yes |

- **Yes** = supported with a change event generator registered in `TimelineServiceImpl`
- **N/A** = entity does not have the relevant aspect(s) in the entity registry — structurally not applicable

There are no remaining "Not supported" gaps. Every aspect that exists on a supported entity has a registered generator.

### Aspect-to-generator mapping

Each registered aspect produces change events via a dedicated generator:

| Aspect | Entity | Category | Generator |
|---|---|---|---|
| `schemaMetadata` | Dataset | TECHNICAL_SCHEMA, TAG, DOCUMENTATION, GLOSSARY_TERM | `SchemaMetadataChangeEventGenerator` |
| `editableSchemaMetadata` | Dataset | TAG, DOCUMENTATION, GLOSSARY_TERM | `EditableSchemaMetadataChangeEventGenerator` |
| `globalTags` | Dataset, DataProduct | TAG | `GlobalTagsChangeEventGenerator` |
| `glossaryTerms` | Dataset, DataProduct | GLOSSARY_TERM | `GlossaryTermsChangeEventGenerator` |
| `ownership` | Dataset, GlossaryTerm, Domain, DataProduct | OWNERSHIP | `OwnershipChangeEventGenerator` |
| `domains` | Dataset, GlossaryTerm, DataProduct | DOMAIN | `SingleDomainChangeEventGenerator` |
| `structuredProperties` | Dataset, GlossaryTerm, Domain, DataProduct | STRUCTURED_PROPERTY | `StructuredPropertyChangeEventGenerator` |
| `applications` | Dataset, GlossaryTerm, DataProduct | APPLICATION | `ApplicationsChangeEventGenerator` |
| `institutionalMemory` | Dataset, Domain | DOCUMENTATION | `InstitutionalMemoryChangeEventGenerator` |
| `datasetProperties` | Dataset | DOCUMENTATION | `DatasetPropertiesChangeEventGenerator` |
| `editableDatasetProperties` | Dataset | DOCUMENTATION | `EditableDatasetPropertiesChangeEventGenerator` |
| `glossaryTermInfo` | GlossaryTerm | DOCUMENTATION | `GlossaryTermInfoChangeEventGenerator` |
| `glossaryRelatedTerms` | GlossaryTerm | GLOSSARY_TERM | `GlossaryRelatedTermsChangeEventGenerator` |
| `domainProperties` | Domain | DOCUMENTATION | `DomainPropertiesChangeEventGenerator` |
| `dataProductProperties` | DataProduct | DOCUMENTATION, ASSET_MEMBERSHIP | `DataProductPropertiesChangeEventGenerator` |

#### Out of scope — inverse-relationship problem (architecturally hard)

The timeline service queries aspects belonging to a single entity's own URN via the SQL primary key index `(urn, aspect, version)`. The events below require **reverse lookups** — finding all *other* entities whose aspects reference this entity. The SQL table has no index on JSON metadata content, so this would require a full table scan across potentially millions of rows. The UI can show current members because it uses Elasticsearch (which maintains a real-time inverted index on the latest state), but ES does not store version history needed for timeline diffs.

| Entity | Event |
|---|---|
| **Domain** | Asset additions/removals — assets reference the domain via *their own* `domains` aspect; the domain entity has no aspect listing its members |
| **Glossary Term** | Asset additions/removals — same inverse relationship; assets reference the term via their `glossaryTerms` aspect |

### What changed

**Backend:**
- Register Dataset, GlossaryTerm, Domain, and **Data Product** entity types in the timeline service with change event generators for all supported aspects
- Register **Application** change tracking for Datasets (previously missing despite the aspect existing)
- Add `ASSET_MEMBERSHIP` change category to enum and GraphQL schema for tracking Data Product asset additions/removals
- Add `DataProductPropertiesChangeEventGenerator` — diffs name/description (DOCUMENTATION) and `assets` array (ASSET_MEMBERSHIP) using sorted two-pointer merge on `destinationUrn`
- Add `StructuredPropertyChangeEventGenerator` with sorted two-pointer merge diff algorithm
- Add semantic diff support to `SingleDomainChangeEventGenerator`
- Add `DomainPropertiesChangeEventGenerator` — diffs name and description on the domain properties aspect (DOCUMENTATION category)
- Add `GlossaryRelatedTermsChangeEventGenerator` — diffs Inherits, Contains, Has Value, and Related To arrays on the glossary related terms aspect (GLOSSARY_TERM category)
- Add `ApplicationsChangeEventGenerator` — diffs applications URN array (APPLICATION category)
- Add count-based timeline fetching that reuses the existing `AspectDao.getAspectsInRange` with a full time range (epoch to now). DataHub's retention policy caps each (urn, aspect) pair to ~20 versions by default, so the result set is inherently bounded without needing a custom SQL `ORDER BY + LIMIT` query. The `maxChangeTransactions` cap is enforced after the diff computation.
- Add `DOMAIN`, `STRUCTURED_PROPERTY`, `APPLICATION`, and `ASSET_MEMBERSHIP` to the `ChangeCategory` enum
- **Rename `ChangeCategory.OWNER` → `OWNERSHIP`** to fix an accidental 2022 divergence between the backend enum (`OWNER`) and the GraphQL enum (`OWNERSHIP`). This eliminates the bidirectional mapping tables that were previously needed. `OWNER` is accepted as a backward-compat alias in the REST API (`StringToChangeCategoryConverter`) and the Python CLI.
- Fix `combineTransactionsByTimestamp` crash on unmodifiable lists

**Frontend:**
- Add Change History button (clock icon) to Domain, Glossary Term, and **Data Product** entity page headers
- Add a category filter dropdown to the change history and make it entity-type-aware — only shows categories supported by the backend for each entity type (including "Assets" for Data Products)
- Expand `changeEventToString` from 2 to 9 categories (schema, documentation, tags, terms, owners, domains, structured properties, applications, asset membership)
- Centralize all change-history constants (parameter keys, relationship type labels, category values) in `HistorySidebar.utils.ts` — the single source of truth for the backend-frontend contract
- Map technical relationship types to human-readable labels (Is A → "inherited term", Has A → "contained term", Has Value → "value term", Is Related To → "related term") with safe fallback for unknown types
- Strip entity URN references from all change event descriptions (redundant since the user is viewing that entity's history)
- Add actor attribution ("by {actorName}") next to each timestamp
- Add long-text truncation with Show more/less toggle for verbose change descriptions
- Lazy-load the timeline GraphQL query (only fires when sidebar opens)
- Add transaction limit of 100 events with truncation indicator in UI

### Tests

- `HistorySidebar.utils.test.ts` — 7 tests: per-entity category filtering (Dataset, GlossaryTerm, Domain, DataProduct with 9 categories), fallback for unknown types, display order preservation
- `changeEventToString.test.ts` — 42 tests: all categories, all operations, field-level variants, relationship type mapping (all 4 types + unknown fallback), entity URN stripping
- `DataProductPropertiesChangeEventGeneratorTest.java` — 14 tests: name add/change/remove, description add/change/remove, both changed, non-DOCUMENTATION category ignored, asset added/removed/swapped/no-change/large-diff/null-to-assets
- Backend: 41 Java tests across `ChangeEventMapper`, `ChangeTransactionMapper`, `GetTimelineResolver`, `CombineTransactionsByTimestamp`, `SingleDomainChangeEventGenerator`, `DataProductPropertiesChangeEventGenerator`, `TimelineService`
- `timeline_change_history_test.py` — 29 smoke tests: end-to-end GraphQL verification of all change categories across all 4 entity types (including ASSET_MEMBERSHIP for Data Products), actor attribution, response structure validation

## Test plan
- [x] `./gradlew :metadata-io:compileJava :datahub-graphql-core:compileJava` — compilation succeeds
- [x] `./gradlew :datahub-graphql-core:test` — all GraphQL tests pass
- [x] `./gradlew :metadata-io:test` — all timeline tests pass (14 DataProductProperties tests + existing)
- [x] `yarn type-check` — no TypeScript errors
- [x] `yarn test` — 42 frontend tests pass (changeEventToString) + 7 (HistorySidebar.utils) = 49 total
- [x] Java Spotless formatting — passes
- [x] Frontend ESLint — passes on all changed files
- [x] Live tested against `quickstartDebug` — verified all four entity types produce correct change events via GraphQL API and render correctly in the UI (seed script: `out/seed-timeline-demo-data.py`)
- [ ] Smoke test suite (`timeline_change_history_test.py`) — 29 tests across Dataset, GlossaryTerm, Domain, DataProduct

## Checklist
- [x] Tests added/updated
- [x] Breaking changes documented in `docs/how/updating-datahub.md` (`ChangeCategory.OWNER` → `OWNERSHIP` GraphQL enum rename)


datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/timeline/GetTimelineResolver.java
```diff
@@ -28,6 +30,34 @@
 public class GetTimelineResolver implements DataFetcher<CompletableFuture<GetTimelineResult>> {
   private final TimelineService _timelineService;
 
+  // GraphQL ChangeCategoryType -> backend ChangeCategory mapping.
+  // Needed because GraphQL uses OWNERSHIP while backend uses OWNER.
```
gabe-lyons: is it possible to fix this? this seems like a reasonable breaking change to execute rather than maintain this mapping forever.

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/types/timeline/mappers/ChangeEventMapper.java
```diff
@@ -13,17 +17,60 @@
 @Slf4j
 public class ChangeEventMapper {
 
+  // Backend ChangeCategory -> GraphQL ChangeCategoryType mapping.
+  // Needed because backend uses OWNER while GraphQL uses OWNERSHIP.
```
gabe-lyons: same here - but either way there should not be duplicated

datahub-web-react/src/app/entityV2/shared/EntityDropdown/ChangeHistoryMenuAction.tsx
```diff
@@ -0,0 +1,30 @@
+import { Tooltip } from '@components';
+import { ClockCounterClockwise } from '@phosphor-icons/react';
+import React, { useState } from 'react';
+
+import { useEntityData } from '@app/entity/shared/EntityContext';
+import { ActionMenuItem } from '@app/entityV2/shared/EntityDropdown/styledComponents';
+import HistorySidebar from '@app/entityV2/shared/tabs/Dataset/Schema/history/HistorySidebar';
+
+export default function ChangeHistoryMenuAction() {
```
gabe-lyons, line 9: neat idea!

datahub-web-react/src/app/entityV2/shared/tabs/Dataset/Schema/history/changeEventToString.ts
```diff
@@ -12,6 +18,19 @@ const SET_ASSET_DOC = (description) => `Set asset documentation to ${description
...
     `Set field documentation for ${downgradeV2FieldPath(fieldPath)} to ${description}`;
 
+// Maps backend relationship type strings to human-readable labels.
+// Uses the same vocabulary as the Glossary Term "Related Terms" tab in the UI.
+const RELATIONSHIP_TYPE_LABELS: Record<string, string> = {
+    'Is A': 'inherited',
+    'Has A': 'contained',
+    'Has Value': 'value',
+    'Is Related To': 'related',
+};
```
gabe-lyons: why is this record defined twice if its already used somewhere

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/timeline/GetTimelineResolver.java
max-datahub: > Good catch — this mismatch is actually an accidental divergence. The original enum in #4252 (Feb 2022) used `OWNERSHIP`, then #4687 (Apr 2022) renamed it to `OWNER`, and #4793 (May 2022) created the GraphQL schema with `OWNERSHIP` without noticing the backend had changed.
> 
> I've renamed the backend enum back to `OWNERSHIP` and added `OWNER` as a backward-compat alias in `StringToChangeCategoryConverter` (REST API) and the CLI so existing callers aren't broken. This eliminates both mapping tables.

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/types/timeline/mappers/ChangeEventMapper.java
max-datahub: > Fixed — both mapping tables removed since the backend enum is now `OWNERSHIP` matching the GraphQL enum directly.

datahub-web-react/src/app/entityV2/shared/tabs/Dataset/Schema/history/changeEventToString.ts
max-datahub: > This isn't actually duplicated — the comment was misleading (fixed now). The backend `GlossaryRelatedTermsChangeEventGenerator` emits formal relationship type strings (`"Is A"`, `"Has A"`, `"Has Value"`, `"Is Related To"`), which are different from the UI tab labels (`"Contains"`, `"Inherits"`, `"Contained by"`, `"Inherited by"` from `RelatedTermTypes` in `GlossaryRelatedTermsResult.tsx`). The existing UI labels describe direction-from-entity, while these are the formal type names. Plus `"Has Value"` and `"Is Related To"` have no equivalent in the existing UI enum at all (it only has `HasA`/`IsA`).

max-datahub: See Claude's comment above. Let me know if I misunderstood you and you want to pull out some global relationship type constants

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/timeline/GetTimelineResolver.java
max-datahub: Let me know if the latest (fairly broad) set of changes addresses the breaking change appropriately

datahub-web-react/src/app/entityV2/shared/tabs/Dataset/Schema/history/changeEventToString.ts
gabe-lyons: We should pull out global constants rather than duplicate mappings.

metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectDao.java
```diff
@@ -1019,6 +1019,25 @@ public List<EntityAspect> getAspectsInRange(
     return ebeanAspects.stream().map(EbeanAspectV2::toEntityAspect).collect(Collectors.toList());
   }
 
+  @Override
+  @Nonnull
+  public List<EntityAspect> getLatestAspects(
```
gabe-lyons: why is a new method required of the Ebean Aspect Dao? This is very scary

gabe-lyons:
Let's make sure this breaking change is documented. @chriscollins3456 can you review the rest?

metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectDao.java
max-datahub: The previous implementation was fetching all timeseries records, whereas the new one only fetches "last n" records. This pushes down the filter to the database and should mean that the query is actually more performant that the full scan.
If you prefer fewer DAOs (e.g. for maintenance reasons) over having better DB performance (by fetching excess data for assets with a long history), we can remove the new DAO.
Claude proposed the following refactor (https://github.com/datahub-project/datahub/pull/16342/changes/664379dfca3c77bc10fedf599ef06ac6499ed9bd)

> >   The latest change converts it from an abstract method to a default method with a safe fallback implementation (delegates to
>   the existing getAspectsInRange + sort/truncate in Java). This means:                                                 
>                                                                         
>   - No implementation is forced to override it — the interface provides a working default
>   - EbeanAspectDao still overrides with the efficient SQL ORDER BY + LIMIT for production performance
>   - CassandraAspectDao's override was removed since it was identical to the default
> 
>   The Javadoc also now explicitly documents why this method exists — none of the existing DAO methods support "latest N
>    historical versions" (getAspectsInRange needs time bounds, batchGet needs exact version tuples, getLatestAspect
>   returns only version 0).
> 
>   So the interface surface area technically grows by one method, but it's a non-breaking addition with a default
>   implementation — similar to how getLatestAspect (singular) is already a default method on the same interface.

max-datahub: Removed the DAO to address review concerns. The overall, negative performance impact of fetching an unlimited number of versions from the backend db is limited by the max. version history of 20

datahub-web-react/src/app/__tests__/Routes.test.tsx
```diff
@@ -16,7 +16,7 @@ test('renders embed page properly', async () => {
     );
 
     await waitFor(() => expect(getByText('Yet Another Dataset')).toBeInTheDocument());
-});
+}, 15_000);
```
chriscollins3456, line 19: i think the reason this is failing is because we have some known issues with PRs put up from forks as opposed to the main repo (which is obviously a problem we're working on)

so for the best chance of this PR getting through CI I would create this PR off of the main repo instead of your personal fork. i'm happy to help with that you shouldn't need to redo any work

datahub-web-react/src/app/entityV2/shared/EntityDropdown/EntityMenuActions.tsx
```diff
@@ -107,6 +110,7 @@ function EntityMenuActions(props: Props) {
             ) : (
                 <MenuItems $shouldFillAllAvailableSpace={shouldFillAllAvailableSpace} data-testid="entity-menu-actions">
                     <ExternalUrlMenuAction shouldFillAllAvailableSpace={shouldFillAllAvailableSpace} />
+                    {menuItems.has(EntityMenuItems.CHANGE_HISTORY) && <ChangeHistoryMenuAction />}
```
chriscollins3456, line 113: let's double check we want this always pulled out of the 3 dot menu as opposed to being in the menu when the entity sidebar is open. ie. this:

Image

datahub-web-react/src/app/entityV2/glossaryTerm/GlossaryTermEntity.tsx
```diff
@@ -34,6 +34,7 @@ import { GetGlossaryTermQuery, useGetGlossaryTermQuery } from '@graphql/glossary
 import { EntityType, GlossaryTerm, SearchResult } from '@types';
 
 const headerDropdownItems = new Set([
+    EntityMenuItems.CHANGE_HISTORY,
```
chriscollins3456, line 37: let's add this to the menu of datasets and any of the other entities that support it

i think we should remove this from the Schema tab of datasets since now it's much more high level than just schema changes

datahub-web-react/src/app/entityV2/shared/tabs/Dataset/Schema/history/HistorySidebar.tsx
```diff
@@ -109,16 +172,40 @@ const HistorySidebar = ({ open, onClose, urn, siblingUrn, versionList, hideSeman
...
+                        placeholder="Filter"
+                        selectLabelProps={{ variant: 'labeled', label: 'Types' }}
+                        options={categoryOptions}
+                        values={selectedCategories}
+                        onUpdate={(values) => setSelectedCategories(values)}
+                        width="fit-content"
+                        showClear={false}
+                        isMultiSelect
+                        showSelectAll
+                    />
```
chriscollins3456, line 185: maybe we can move this out of the header of this component and place it at the top above the change history timeline? it just looks a little awkward in the header right now

Image

datahub-web-react/src/app/entityV2/shared/tabs/Dataset/Schema/history/useResolveEntityNames.ts
```diff
@@ -0,0 +1,64 @@
...
+const URN_PREFIX = 'urn:li:';
+
+/**
+ * Collects all entity URNs referenced in timeline change events (from modifier
+ * and parameters) and batch-resolves them to display names via a single GraphQL
+ * query using the entityDisplayNameFields fragment.
+ *
+ * Returns a Map<string, string> from URN → human-readable display name.
+ */
+export function useResolveEntityNames(transactions: ChangeTransaction[] | undefined): Map<string, string> {
```
chriscollins3456, line 17: nice

datahub-web-react/src/app/entityV2/glossaryTerm/GlossaryTermEntity.tsx
chriscollins3456: also for some reason i'm not seeing name change events appearing for glossary terms. i do for domains and data products though

metadata-io/src/main/java/com/linkedin/metadata/entity/cassandra/CassandraAspectDao.java
```diff
@@ -737,9 +737,9 @@ public List<EntityAspect> getAspectsInRange(
             .whereColumn(CassandraAspect.ASPECT_COLUMN)
             .in(aspectNamesToLiterals(aspectNames))
             .whereColumn(CassandraAspect.CREATED_ON_COLUMN)
-            .isLessThanOrEqualTo(literal(startTimeMillis))
+            .isGreaterThanOrEqualTo(literal(startTimeMillis))
             .whereColumn(CassandraAspect.CREATED_ON_COLUMN)
-            .isGreaterThan(literal(endTimeMillis))
+            .isLessThan(literal(endTimeMillis))
```
chriscollins3456, line 742: woah nice fix

metadata-io/src/main/java/com/linkedin/metadata/timeline/eventgenerator/DataProductPropertiesChangeEventGenerator.java
```diff
@@ -0,0 +1,220 @@
...
+import com.linkedin.metadata.timeline.data.ChangeTransaction;
+import com.linkedin.metadata.timeline.data.SemanticChangeType;
+import jakarta.json.JsonPatch;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class DataProductPropertiesChangeEventGenerator
```
chriscollins3456, line 23: just double checking we're not expecting change events to show up on dataset pages for adding/removing a data product right?

it doesn't seem like that was included in the PR description.

obviously this change event generator is for changes on the data product itself and that looks good

metadata-io/src/main/java/com/linkedin/metadata/timeline/TimelineServiceImpl.java
```diff
@@ -296,16 +517,49 @@ public List<ChangeTransaction> getTimeline(
...
+  @Nonnull
+  @Override
+  public List<ChangeTransaction> getTimeline(
+      @Nonnull final Urn urn,
+      @Nonnull final Set<ChangeCategory> elementNames,
+      int maxChangeTransactions,
+      boolean rawDiffRequested) {
+
+    List<ChangeTransaction> allTransactions =
+        getTimeline(urn, elementNames, 0, 0, null, null, rawDiffRequested);
```
chriscollins3456, line 540: since people could increase their retention to more than 20 aspects do you think it makes sense to place a limit on the number of transactions we should be getting?

just in case this ends up getting thousands of events... it might be nice to play a little defence and set a limit of like 1000 or something

max-datahub:
Moved to #16929 (branch on main repo instead of fork) per reviewer feedback about CI issues with forked PRs.

metadata-io/src/main/java/com/linkedin/metadata/timeline/eventgenerator/DataProductPropertiesChangeEventGenerator.java
max-datahub: Looking up data products would be really expensive:
> This is the same inverse-relationship problem documented in the PR description. The dataset doesn't own the dataProductProperties aspect; the data product does. Fetching this would require scanning all data product URNs.

datahub-web-react/src/app/entityV2/shared/tabs/Dataset/Schema/history/HistorySidebar.tsx
annadoesdesign: agree here: if you want to add a search bar (idk hwo easy, but an example in design) it may make it feel balance 
image

datahub-web-react/src/app/entityV2/shared/EntityDropdown/EntityMenuActions.tsx
annadoesdesign: Agree here as well - I think were trying to keep things as little overwhelming as possible and evveryting should go in the 3 dot menu

metadata-io/src/main/java/com/linkedin/metadata/timeline/eventgenerator/DataProductPropertiesChangeEventGenerator.java
max-datahub: but @chriscollins3456 you are right, this was not documented explicitly! Will add it in the PR message

metadata-io/src/main/java/com/linkedin/metadata/timeline/TimelineServiceImpl.java
max-datahub: I actually had introduced a DAO to limit the fetch size on the DB side, but @gabe-lyons voiced concerns in his initial PR review, so I removed that functionality in https://github.com/datahub-project/datahub/pull/16342/changes/7f5dde1ccd3fb52887faec9d6acc824acfe68236 
I would thus suggest to either re-introduce the "scary" DAO (which includes ORDER BY and LIMIT to safeguard against excessive data fetching) or depend on the retention policy as our safeguard

metadata-io/src/main/java/com/linkedin/metadata/entity/cassandra/CassandraAspectDao.java
max-datahub: Credit to my buddy Claude :)

datahub-web-react/src/app/entityV2/shared/EntityDropdown/EntityMenuActions.tsx
max-datahub: Done. Moved change history into the 3 dot menu for all entity types

datahub-web-react/src/app/__tests__/Routes.test.tsx
max-datahub: Moved to PR directly on the OSS repo at https://github.com/datahub-project/datahub/pull/16929

datahub-web-react/src/app/entityV2/glossaryTerm/GlossaryTermEntity.tsx
max-datahub: > also for some reason i'm not seeing name change events appearing for glossary terms. i do for domains and data products though
Thanks for pointing this out. Should be fixed now.

Moved all change history menu items into the 3 dot menu as well

datahub-web-react/src/app/entityV2/shared/tabs/Dataset/Schema/history/HistorySidebar.tsx
max-datahub: @annadoesdesign I added a search bar, let me know if you have any special requirements on what it should and should not filter for (I re-used the pre-existing search bar implementation).
image

PR #16355, opened 2026-02-25
feat(metadata-io): reindex retry, createIndex retry, and ES7 test fixes

david-leifker:
## Summary
Adds retries for index-build operations (count and createIndex), fixes ES7 test setup so retry integration tests run against the correct client, and fixes an NPE in search query building when validation config is missing.

## Changes
- **ESIndexBuilder**: Count operations use resilience4j retry; createIndex retries on `IOException` when enabled.
- **BuildIndicesConfiguration**: New options `countRetryMaxAttempts`, `countRetryWaitSeconds`, `createIndexRetryEnabled` (and application.yaml defaults).
- **ElasticSearchSuite**: Waits for cluster health before returning the container to avoid "connection refused" in tests.
- **SearchClientShimUtil**: Detection order is ES7 → OpenSearch 2 → ES8; on failure logs a warning with probe outcomes and throws an `IOException` with the same details.
- **SearchQueryBuilder**: Uses a default `SearchValidationConfiguration` when `SearchConfiguration.getValidation()` is null (avoids NPE in tests).
- **Integration tests**: `IndexBuilderRetryElasticSearchIntegrationTest` and `IndexBuilderRetryOpenSearchIntegrationTest` using `FaultInjectingSearchClientShim` and `FaultSpec`.


metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/indexbuilder/ESIndexBuilder.java
```diff
@@ -136,6 +138,19 @@ public ESIndexBuilder(
     this.indexSettingOverrides = indexSettingOverrides;
     this.gitVersion = gitVersion;
 
+    var buildIndices =
```
RyanHolstien: nit: I'm generally against var usage for cases like this. It can make sense sometimes, but this mainly just masks the type information unless I pull up this PR in an IDE. I know this one's just a config class, but with AI drawing from usage patterns in code I'd rather not have more of these instances build up.

RyanHolstien (approved):

PR #16358, opened 2026-02-26
feat(dbt): add convert_urns_to_lowercase support for dbt ingestion

alfiyas-datahub:
## Summary

- Adds `convert_urns_to_lowercase` config option to dbt ingestion (both dbt-core and dbt-cloud), resolving duplicate entities caused by schema name casing differences (e.g., `app_sales` vs `APP_SALES` producing two separate dbt URNs for the same table)
- Reuses the existing `LowerCaseDatasetUrnConfigMixin` already used by Snowflake, BigQuery, SQL, Kafka, S3, and other sources
- Defaults to `true` for Snowflake target platform (matching the existing `convert_column_urns_to_lowercase` behavior)


rajatoss:
## Connector Tests Results

**Connector tests failed** for commit `458c19c`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22707020196)

sgomezvillamor:
First of all, thanks for raising the issue and proposing a solution. A few comments/questions before diving into the code review.

In the past, we introduced `convert_urns_to_lowercase` and `convert_column_urns_to_lowercase` to ensure linkable lineage across different sources. We’ve seen issues in setups like dbt + Snowflake or dbt + BigQuery. In those cases, we solved them by setting `convert_urns_to_lowercase=true` in the Snowflake or BigQuery connector itself.

In this proposal, you suggest setting the flag in the dbt source instead. Could you elaborate on why that’s needed? Why wouldn’t setting `convert_urns_to_lowercase=true` in the Snowflake or BigQuery connector be enough to solve the issue?

Regarding the breaking change: it’s true we currently have some inconsistencies in the default values of `convert_urns_to_lowercase` across different sources. While we do want to clean that up eventually, I’d prefer to avoid introducing a breaking change at this point.

alfiyas-datahub:
> First of all, thanks for raising the issue and proposing a solution. A few comments/questions before diving into the code review.
> 
> In the past, we introduced `convert_urns_to_lowercase` and `convert_column_urns_to_lowercase` to ensure linkable lineage across different sources. We’ve seen issues in setups like dbt + Snowflake or dbt + BigQuery. In those cases, we solved them by setting `convert_urns_to_lowercase=true` in the Snowflake or BigQuery connector itself.
> 
> In this proposal, you suggest setting the flag in the dbt source instead. Could you elaborate on why that’s needed? Why wouldn’t setting `convert_urns_to_lowercase=true` in the Snowflake or BigQuery connector be enough to solve the issue?
> 
> Regarding the breaking change: it’s true we currently have some inconsistencies in the default values of `convert_urns_to_lowercase` across different sources. While we do want to clean that up eventually, I’d prefer to avoid introducing a breaking change at this point.

Hi Sergio, thanks for the thoughtful review. Let me clarify the specific scenario.
You're right that setting convert_urns_to_lowercase=true on the Snowflake/BigQuery connector solves lineage linkage, the target platform URNs already match because the dbt connector lowercases them when target_platform != "dbt" (line 937 in dbt_common.py). That part works fine today.

The issue this PR addresses is different: duplicate entities on the dbt platform itself.
When the dbt connector runs, it creates two sets of entities per node:
Target platform URN (e.g., urn:li:dataPlatform:snowflake) — already lowercased by existing code, no problem here
dbt platform URN (e.g., urn:li:dataPlatform:dbt) — preserves original casing from the manifest

With Snowflake, identifiers in the dbt manifest often appear in uppercase (MY_DB.APP_SALES.dim_industry). If the casing in the manifest changes between runs or if different nodes reference the same schema with different casing, we will end up with duplicate dbt entities in DataHub:
      **urn:li:dataset:(urn:li:dataPlatform:dbt, MY_DB.APP_SALES.dim_industry, PROD)
      urn:li:dataset:(urn:li:dataPlatform:dbt, my_db.app_sales.dim_industry, PROD)**
    
Setting convert_urns_to_lowercase on the Snowflake connector can't fix this, because it only affects dataPlatform:snowflake URNs — it has no control over the dataPlatform:dbt URNs that the dbt connector generates.

On the breaking change concern — completely agree. I'll remove the Snowflake auto-default (convert_urns_to_lowercase=True when target_platform=snowflake) and keep it as a purely opt-in flag. Users who hit this issue can explicitly enable it in their dbt recipe.

Want me to also go ahead and remove the Snowflake auto-default from the code so it's ready for the next push?

sgomezvillamor:
I'm not super happy with the accumulation of lowercasing config flags across sources and while we should strategically revisit this at some moment, I guess there is no immediate pragmatic solution.

Other options for dbt case would be:

* Always lowercase dbt platform urns (no config flag). But this would be a breaking change and would require lowercasing in the target platform to have linkable lineage
* Inherit casing behavior from target platform (Implicit). In case insensitive platforms such as snowflake, bigquery ,etc we shouldn't be concerned about casing at all and lowercase anyway. But again that would be a breaking change. This approach may be considered for the future strategic solution.

Opt-in config flag proposed here is conservative and safe solution.

> On the breaking change concern — completely agree. I'll remove the Snowflake auto-default (convert_urns_to_lowercase=True when target_platform=snowflake) and keep it as a purely opt-in flag. Users who hit this issue can explicitly enable it in their dbt recipe.

Yes, please

> Want me to also go ahead and remove the Snowflake auto-default from the code so it's ready for the next push?

WDYM?

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
sgomezvillamor, line 1: What about having `convert_urns_to_lowercase` on `DBTNode` at construction time instead of passing it through every function signature. This eliminates the parameter from all many call sites and prevents silent omissions if new call forgets the parameter.

sgomezvillamor: However, `DBTNode` is a model class... so not sure if adding a behaviour flag there is the best option.

Can you give this a thought, please?

metadata-ingestion/tests/integration/dbt/test_dbt.py
```diff
@@ -618,3 +619,53 @@ def test_dbt_only_test_definitions_and_results(
...
+        output = json.load(f)
+
+    dbt_urns = {
+        item["entityUrn"]
+        for item in output
+        if "entityUrn" in item and "dataPlatform:dbt" in item["entityUrn"]
+    }
+    assert len(dbt_urns) > 0, "Expected at least one dbt platform URN in output"
+    for urn in dbt_urns:
+        dataset_name = urn.split(",")[1]
```
sgomezvillamor: we have `DatasetUrn.from_string().get_dataset_name()`

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
alfiyas-datahub: DBTNode already has behavior methods like get_urn() and is_ephemeral_model(), so it's not a pure data class. The flag naturally fits there since it controls how the URN is built.
The alternative is threading it through 15+ call sites, which works but risks silent omissions if someone adds a new get_urn() call without the parameter.
Happy to go either way, let me know your preference.

sgomezvillamor: `DBTNode` seems like the safer approach

PR #16360, opened 2026-02-26
test(ingestion): fix (or mitigate) flakiness of sql parser aggregator perf tests

sgomezvillamor:
Adding retries for sql parsing aggregator performance tests to mitigate flakiness in CI env

### DIFF
--- a/metadata-ingestion/docs/sources/snowflake/snowflake_pre.md
+++ b/metadata-ingestion/docs/sources/snowflake/snowflake_pre.md
@@ -81,6 +81,28 @@ If you plan to enable extraction of table lineage via the `include_table_lineage
 grant imported privileges on database snowflake to role datahub_role;
 ```
 
+Note that `imported privileges` grants access to all schemas and views in the shared `SNOWFLAKE` database, primarily:
+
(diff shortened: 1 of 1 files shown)


metadata-ingestion/setup.py
```diff
@@ -846,6 +846,7 @@
     "pytest-asyncio>=0.16.0,<2.0.0",
     "pytest-cov>=2.8.1,<8.0.0",
     "pytest-random-order~=1.1.0,<2.0.0",
+    "pytest-rerunfailures",
```
anshbansal: @sgomezvillamor we need to add upper bounds here to avoid breaking CI

rajatoss:
## Connector Tests Results

**Connector tests failed** for commit `399a706`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22439172715)

metadata-ingestion/setup.py
sgomezvillamor: I’ve just pushed the upper bound in commit c2c0fb91c95626fc0e3a3aef243884cb9ba7ea37.

I’m not a big fan of this requirement to always add upper bounds, as it often leads to somewhat artificial constraints like this one. But that’s a broader discussion and probably out of scope for this PR.

PR #16361, opened 2026-02-26
fix(ingestion/oracle): fix geospatial type recognition for SDO_GEOMETRY columns

acrylJonny:
## Problem

Oracle geospatial columns (`SDO_GEOMETRY`, `SDO_POINT_TYPE`, etc.) were being ingested with a `null` native data type and emitting a SQLAlchemy warning:

```
SAWarning: Did not recognize type 'SDO_GEOMETRY' of column 'contour_polygon_geom'
```

## Root cause

`OracleSource.get_workunits()` registered the geospatial types into SQLAlchemy's `ischema_names` using `patch.dict`, but used `return` to hand back the parent generator:

```
with patch.dict("...OracleDialect.ischema_names", ..., clear=False):
    return super().get_workunits()   # bug: exits the context immediately
```

Because super().get_workunits() is a generator, return creates the generator object and exits the with block before any iteration happens. By the time the pipeline iterates the workunits and column types are resolved, ischema_names no longer contains the geospatial types.

## Fix
Change return to yield from, keeping the patch.dict context alive for the full duration of iteration:

```
with patch.dict("...OracleDialect.ischema_names", ..., clear=False):
    yield from super().get_workunits()   # fix: context stays open during iteration
```

Tests

- `test_get_columns_sdo_geometry` — directly calls `OracleInspectorObjectWrapper.get_columns` with an `SDO_GEOMETRY` column inside the patch and asserts the result is not `NullType`
- `test_sdo_geometry_types_registered_during_workunits_iteration` — regression test verifying that `ischema_names` still contains the geospatial types at the moment the workunit generator body executes

### DIFF
--- a/metadata-ingestion/setup.py
+++ b/metadata-ingestion/setup.py
@@ -846,6 +846,7 @@ base_dev_requirements = {
     "pytest-asyncio>=0.16.0,<2.0.0",
     "pytest-cov>=2.8.1,<8.0.0",
     "pytest-random-order~=1.1.0,<2.0.0",
+    "pytest-rerunfailures<17.0",
     "requests-mock<2.0.0",
     "freezegun<2.0.0",  # TODO: fully remove and use time-machine
     "time-machine<4.0.0",  # better Pydantic v2 compatibility
--- a/metadata-ingestion/tests/performance/sql_parsing/test_sql_aggregator.py
+++ b/metadata-ingestion/tests/performance/sql_parsing/test_sql_aggregator.py
@@ -151,6 +151,8 @@ def run_configuration(
 
 
 @pytest.mark.integration
+# adding retries to mitigate flakiness in CI env
+@pytest.mark.flaky(reruns=5)
 def test_benchmark(pytestconfig: pytest.Config) -> None:
     """Run benchmark test across a matrix of configurations.
 


rajatoss:
## Connector Tests Results

**Connector tests failed** for commit `1715150`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22443840379)

metadata-ingestion/tests/unit/test_oracle_source.py
```diff
@@ -1111,3 +1144,39 @@ def test_get_procedure_dependencies_returns_upstream_tables(
...
+    before any iteration; `yield from` keeps it open for the full duration.
+    """
+    sdo_registered_during_iteration: list = []
+
+    def mock_parent_workunits():
+        for type_cls in extra_oracle_types:
+            sdo_registered_during_iteration.append(type_cls.__name__ in ischema_names)
+        return
+        yield  # make it a generator
+
```
kyungsoo-datahub, line 1162: Someone unfamiliar with this trick may delete the unreachable yield as dead code, silently breaking the test (it becomes a regular function, the body runs eagerly at return_value=mock_parent_workunits() time, outside the patch.dict context, and the test passes vacuously regardless of return vs yield from).
```
  def mock_parent_workunits():
      for type_cls in extra_oracle_types:
          sdo_registered_during_iteration.append(type_cls.__name__ in ischema_names)
      yield from ()  # empty generator
```

metadata-ingestion/tests/unit/test_oracle_source.py
```diff
@@ -304,6 +307,36 @@ def test_get_materialized_view_definition(self):
         call_args = self.mock_inspector.bind.execute.call_args
         assert "dba_mviews" in str(call_args[0][0]).lower()
 
+    def test_get_columns_sdo_geometry(self):
```
kyungsoo-datahub, line 310: This test is redundant given the `test_sdo_geometry_types_registered_during_workunits_iteration` test already covers the meaningful behavior.

PR #16363, opened 2026-02-26
fix(snowflake): handle Snowflake 2026_01 BCR property format change in DDL lineage

max-datahub:
## Summary

Snowflake's `2026_01` BCR bundle (which includes [BCR-2178](https://docs.snowflake.com/en/release-notes/bcr-bundles/2026_01/bcr-2178)) changed the format of `object_modified_by_ddl` properties in `access_history` from nested `{"key": {"value": "..."}}` to flat `{"key": "..."}`.

This causes a `TypeError: string indices must be integers` crash in `parse_ddl_query` when it does `object_modified_by_ddl["properties"]["objectName"]["value"]` on a plain string instead of a dict. The crash breaks DDL lineage extraction for **RENAME TABLE** and **SWAP TABLE** operations on any Snowflake account that has enabled the `2026_01` BCR bundle.

### The breaking change

**Pre-BCR (current code handles this):**
```json
{
  "objectName": "DB.SCHEMA.OLD_TABLE",
  "operationType": "ALTER",
  "properties": {"objectName": {"value": "DB.SCHEMA.NEW_TABLE"}}
}
```

**Post-BCR (crashes current code):**
```json
{
  "objectName": "DB.SCHEMA.OLD_TABLE",
  "operationType": "ALTER",
  "properties": {"objectName": "DB.SCHEMA.NEW_TABLE"}
}
```

The crash: `["properties"]["objectName"]["value"]` → `"some_string"["value"]` → `TypeError`

### Fix

- Add a `_get_ddl_property(properties, key)` helper that handles both pre-BCR (wrapped in `{"value": ...}`) and post-BCR (flat string) formats
- Add an `isinstance(dict)` guard in `parse_ddl_query` for non-dict `object_modified_by_ddl` values (e.g., strings from session-level DDL)
- Improve type annotation from `Any` to `object` for `object_modified_by_ddl` parameter, forcing explicit type narrowing

### Testing

- 6 unit tests using real Snowflake `access_history` payloads observed before and after enabling the `2026_01` BCR:
  - `test_rename_table_pre_2026_01_bcr` — backward compatibility
  - `test_rename_table_post_2026_01_bcr` — the fix (was `TypeError`)
  - `test_swap_table_pre_2026_01_bcr` — backward compatibility
  - `test_swap_table_post_2026_01_bcr` — the fix (was `TypeError`)
  - `test_non_dict_ddl_returns_none` — graceful handling of non-dict values
  - `test_non_alter_ddl_dropped` — non-ALTER DDL is dropped correctly
- Validated end-to-end against a live Snowflake instance with `2026_01` BCR enabled: 2 table renames and 1 table swap parsed successfully, zero ingestion failures

## Checklist
- [x] The PR conforms to DataHub's [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (particularly PR title format)
- [x] Links to related issue (if applicable) — N/A, discovered via testing
- [x] Tests added/updated
- [ ] Docs added/updated — N/A, no user-facing config changes

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_queries.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_queries.py
@@ -544,7 +544,22 @@ class SnowflakeQueriesExtractor(SnowflakeStructuredReportMixin, Closeable):
         objects_modified = res["objects_modified"]
         object_modified_by_ddl = res["object_modified_by_ddl"]
 
-        if object_modified_by_ddl and not objects_modified:
+        if (
+            object_modified_by_ddl
+            and not isinstance(object_modified_by_ddl, dict)
+            and not objects_modified
+        ):
+            self.report.num_ddl_queries_dropped += 1
+            if direct_objects_accessed:
+                pass  # Fall through to normal query processing for usage tracking
+            else:
+                return None
+
+        if (
+            isinstance(object_modified_by_ddl, dict)
+            and object_modified_by_ddl
+            and not objects_modified
+        ):
             known_ddl_entry: Optional[Union[TableRename, TableSwap]] = None
             with self.structured_reporter.report_exc(
                 "Error fetching ddl lineage from Snowflake"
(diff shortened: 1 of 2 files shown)


rajatoss:
## Connector Tests Results

**Connector tests failed** for commit `9c2f41f`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22449379380)

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_queries.py
```diff
@@ -755,7 +785,9 @@ def parse_ddl_query(
 
             new_urn = self.identifiers.gen_dataset_urn(
                 self.identifiers.get_dataset_identifier_from_qualified_name(
-                    object_modified_by_ddl["properties"]["objectName"]["value"]
+                    self._get_ddl_property(
```
sgomezvillamor, line 788: sort of cosmetic concern: given `_get_dd_property` is static method, you could use class name instead of `self` for referencing the method

PR #16366, opened 2026-02-26
feat(ingestion/oracle): recognize XMLTYPE and geospatial columns instead of falling back to NullType

acrylJonny:
## Summary
Oracle's XMLTYPE and geospatial types (SDO_GEOMETRY, SDO_POINT_TYPE, SDO_ELEM_INFO_ARRAY, SDO_ORDINATE_ARRAY) are not natively registered in SQLAlchemy's Oracle dialect. When SQLAlchemy encounters an unrecognized type it falls back to NullType, which DataHub surfaces as a null native data type and emits a SAWarning: Did not recognize type '...'.
DataHub's Oracle source already had a patch.dict mechanism to register these types into ischema_names at runtime, and extra_oracle_types already covered the SDO family. Two bugs prevented it from working:
Generator context bug — get_workunits used return super().get_workunits() inside the patch.dict block. This returns the generator object and immediately exits the with block, undoing the patch before any iteration occurs. Changed to yield from super().get_workunits() so the patch stays active for the full duration of workunit generation.
Missing XMLTYPE — XMLTYPE was not in extra_oracle_types, so XML columns were still unrecognized. Added make_sqlalchemy_type("XMLTYPE") to the set.

### DIFF
--- a/metadata-ingestion/tests/unit/test_oracle_source.py
+++ b/metadata-ingestion/tests/unit/test_oracle_source.py
@@ -307,6 +307,36 @@ class TestOracleInspectorObjectWrapper:
         call_args = self.mock_inspector.bind.execute.call_args
         assert "dba_mviews" in str(call_args[0][0]).lower()
 
+    def test_get_columns_sdo_geometry(self):
+        """SDO_GEOMETRY columns must resolve to a non-null type when ischema_names is patched."""
+        self.mock_inspector.dialect.server_version_info = (19,)
+        self.mock_inspector.bind.execute.return_value = [
+            [
+                "GEOM_COLUMN",
+                "SDO_GEOMETRY",
+                0,
+                None,
+                None,
+                "Y",
+                None,
+                None,
+                "NO",
+                None,
+                None,
+            ]
+        ]
+
+        with patch.dict(
+            "sqlalchemy.dialects.oracle.base.OracleDialect.ischema_names",
+            {klass.__name__: klass for klass in extra_oracle_types},
+            clear=False,
+        ):
+            columns = self.wrapper.get_columns("TEST_TABLE", "TEST_SCHEMA")
+
+        assert len(columns) == 1
+        assert columns[0]["name"] == "geom_column"
+        assert not isinstance(columns[0]["type"], sqltypes.NullType)
+
     def test_get_columns_xmltype(self):
         """XMLTYPE columns must resolve to a non-null type when ischema_names is patched."""
         self.mock_inspector.dialect.server_version_info = (19,)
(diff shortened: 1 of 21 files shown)


metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
```diff
@@ -2127,7 +2128,7 @@ def get_workunits(self):
             {klass.__name__: klass for klass in extra_oracle_types},
             clear=False,
         ):
-            return super().get_workunits()
+            yield from super().get_workunits()
```
kyungsoo-datahub, line 2131: This one is duplicated from #16361. Will be merged from #16361.

metadata-ingestion/tests/unit/test_oracle_source.py
```diff
@@ -304,6 +307,24 @@ def test_get_materialized_view_definition(self):
         call_args = self.mock_inspector.bind.execute.call_args
         assert "dba_mviews" in str(call_args[0][0]).lower()
 
+    def test_get_columns_xmltype(self):
```
kyungsoo-datahub, line 340: I don't think we need either of these tests. Adding XMLTYPE to extra_oracle_types is the only change in this PR. The patch mechanism and yield from fix are already covered by #16361.
                                                                                                                           
`test_get_columns_xmltype` manually patches ischema_names and verifies a dict lookup succeeds — it doesn't exercise the actual bug path. `test_extra_oracle_types_registered_during_workunits_iteration` is a rename of #16361's regression test and would be a duplicate after rebase.

Both tests validate the workflow, not the change in this PR.

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
acrylJonny: Yeah, this needed some of the same resolution - I will merge in the changes from the other change

metadata-ingestion/tests/unit/test_oracle_source.py
acrylJonny: In the other change it was allowing for geospatial not xml - XML types are newly registered SQLalchemy types in the oracle.py

kyungsoo-datahub: Right, XMLTYPE is new. `test_get_columns_xmltype` tests `XMLTYPE` the same way #16361 tests `SDO_GEOMETRY`. I'd prefer one test that validates the mechanism works for all types in extra_oracle_types, rather than individual per-type tests. `test_extra_oracle_types_registered_during_workunits_iteration` test already iterates over every element in the set.

rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `f198edf`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22461185364)

PR #16368, opened 2026-02-26
feat(ingestion): store raw chunks in SemanticContent for Notion/Confl…

shirshanka:
Implements a two-phase chunking approach for semantic search ingestion:

**Phase 1 (Notion/Confluence):** compute text chunks and store them as `rawChunks` in `SemanticContent` even when no embedding provider is configured (or when provider credentials are unavailable). This preserves chunking work for later use.

**Phase 2 (datahub-documents):** detect documents with `rawChunks`, generate embedding vectors for them, and write real `SemanticContent.embeddings` — preserving `rawChunks` alongside so they remain available for future re-embedding.

**Phase 1 re-ingestion guard:** Notion/Confluence use content-hash state to skip unchanged documents on subsequent runs. This prevents a no-credential Notion run from overwriting the embeddings that Phase 2 computed.

---

### Schema changes
- Add `TextChunk.pdl`: standalone text chunk record (position, offsets, text)
- Add `RawChunks.pdl`: container for pre-computed chunks awaiting embeddings
- Refactor `EmbeddingChunk.pdl` to use `includes TextChunk` + vector field
- Add optional `rawChunks` field to `SemanticContent.pdl`

### Python changes
- Add `emit_chunks_without_embeddings` flag to `DocumentChunkingSourceConfig`
- Add `_emit_raw_chunks()` to `DocumentChunkingSource`
- `process_elements_inline()` emits rawChunks when flag=True and no embedding model
- Notion/Confluence sources set `emit_chunks_without_embeddings=True`
- Notion `_get_processing_config_fingerprint` now always includes chunking params (triggers one-time re-ingestion of existing documents to emit rawChunks)
- `DataHubDocumentsSource`: reuses rawChunks instead of re-partitioning text; preserves rawChunks when writing embeddings; handles `semanticContent` MCL events (with anti-loop guard)
- **Startup embedding probe:** `DocumentChunkingSource.__init__` now makes a test embedding call after resolving the provider. Failure behaviour is source-type-aware:
  - Notion/Confluence (`emit_chunks_without_embeddings=True`): logs a warning and degrades to raw-chunks mode — documents are never lost and can be embedded later
  - DataHub documents source (`emit_chunks_without_embeddings=False`): re-raises — embedding is the entire purpose of this source, so silent failure would skip every document without any indication

### Tests
- `tests/unit/ingestion/source/notion/test_notion_docs_pipeline.py`: end-to-end unit test covering the full round-trip with no live services:
  1. Notion ingests a page with broken/absent credentials → probe fails → rawChunks emitted, no embeddings
  2. DataHub docs source reads rawChunks, generates embeddings → both embeddings and rawChunks preserved
  3. Notion re-ingests the same unchanged page → hash matches → document skipped, embeddings not overwritten
  4. DataHub docs source with broken credentials raises at construction (not silently skips documents)

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/datahub_documents/datahub_documents_source.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/datahub_documents/datahub_documents_source.py
@@ -13,7 +13,7 @@ Supports both batch (GraphQL) and event-driven (Kafka MCL) modes.
 import hashlib
 import json
 import logging
-from dataclasses import field
+from dataclasses import dataclass, field
 from datetime import datetime
 from pathlib import Path
 from typing import Any, Dict, Iterable, Optional, cast
@@ -53,6 +53,7 @@ from datahub.metadata.schema_classes import RawChunksClass, SemanticContentClass
 logger = logging.getLogger(__name__)
 
 
+@dataclass
 class DataHubDocumentsReport(StatefulIngestionReport):
     """Report for DataHub documents source."""
 
@@ -65,7 +66,7 @@ class DataHubDocumentsReport(StatefulIngestionReport):
     num_embeddings_generated: int = 0
     num_embedding_failures: int = 0
     embedding_failures: list[str] = field(default_factory=list)
-    processing_errors: list[str] = []
+    processing_errors: list[str] = field(default_factory=list)
     num_documents_limit_reached: bool = False
 
     def report_document_fetched(self) -> None:


rajatoss:
## Connector Tests Results

**Connector tests failed** for commit `46cc089`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22526523293)

alexsku:
## Code Review — Actionable Items

Treating the two-phase rawChunks approach as decided, here are the implementation issues to address:

---

### High Priority (should fix before merge)

**1. Anti-loop guard is broken when `semanticContent` events are forwarded**

`_process_document_for_raw_chunks` passes `raw_chunks=raw` to `_emit_semantic_content`, preserving rawChunks in the output. If the resulting MCL is ever routed back through `_process_single_event`, `_aspect_has_raw_chunks` returns `True` again → infinite re-embedding loop.

Currently masked because the event consumer only forwards `documentInfo` events, but the code and tests claim this path works.

**Fix:** Either check `rawChunks present AND embeddings empty` (not just rawChunks present), or clear rawChunks in the embedding output.

**2. `semanticContent` event handler is unreachable in production**

`event_consumer.py:319` hard-filters `aspect_name != "documentInfo"`, so `semanticContent` MCL events never reach `_process_single_event`. The new handler is dead code through the normal streaming path.

**Fix:** Either update the event consumer to also forward `semanticContent` events, or add a comment/TODO documenting this as scaffolding for a follow-up PR.

**3. `autouse=True` conftest fixture scope is too broad**

`tests/unit/conftest.py` patches `DocumentChunkingSource._generate_embeddings` for **every** unit test under `tests/unit/`. This could silently change behavior of existing tests that construct a `DocumentChunkingSource` with embedding config.

**Fix:** Scope the fixture to the specific test modules that need it (module-level conftest or explicit marker), rather than directory-wide autouse.

---

### Medium Priority (should fix soon, could merge with follow-up)

- **No `skip_if_embeddings_exist` check in rawChunks batch path** — `_process_single_document`'s rawChunks shortcut re-embeds on every `force_reprocess` run even if embeddings already exist.
- **Private method coupling** — `_process_document_for_raw_chunks` calls `self.chunking_source._generate_embeddings()` and `_emit_semantic_content()` directly. Consider adding a public method on `DocumentChunkingSource`.

---

### Missing Tests

- **Anti-loop regression test**: Document with *both* embeddings and rawChunks arrives as a `semanticContent` event → should be a no-op (not re-embedded).
- **Batch dedup test**: Batch mode with `force_reprocess=True` on a document that already has embeddings alongside rawChunks → verify it doesn't re-embed unnecessarily.

---

### Questions

1. Is the `semanticContent` MCL handler intended to be active now, or is it scaffolding for a future event consumer update?
2. Should rawChunks be cleared (set to null) after embedding to break the loop, or is preservation intentional for future re-embedding scenarios?
3. The Notion `_get_processing_config_fingerprint` still sets `chunking_enabled = embedding_enabled`. Is it intentional that chunking config changes *without* a provider won't trigger reprocessing in raw-chunks mode?

alexsku:
## High-Level Architectural Concerns

Before diving into implementation details (see my other comment), I want to raise some questions about whether this approach is the right one.

**1. Does a server-side embedding API make this unnecessary?**

If DataHub exposes an API that computes embeddings, Notion/Confluence sources can send text directly to the server — no need for rawChunks, two-phase coordination, or credential distribution to remote executors. If this is on the roadmap, shipping new schema types (`TextChunk`, `RawChunks`, `SemanticContent.rawChunks`) that we plan to obsolete adds tech debt.

**2. Is a locally configured embedding provider a viable fallback?**

For self-hosted deployments, adding `embedding.provider` + `api_key` to a recipe is straightforward. For managed/SaaS, the credential distribution problem is real — but a server-side API solves it more cleanly at the architecture level.

**3. The datahub-documents source can already re-chunk from raw text**

`_process_single_document` partitions text from scratch when no rawChunks exist. Chunking is cheap compared to embedding. So the rawChunks mechanism saves a fast operation at the cost of new schema types, two-phase coordination, a startup probe with silent degradation, and an anti-loop guard that doesn't fully work yet (see implementation review comment).

**4. Fail-fast > silent degradation**

If we do go forward with 1 or 2, this PR is counterproductive. It turns embedding credential failures into warnings — Notion silently emits rawChunks, and if Phase 2 is misconfigured too, the user discovers semantic search is broken weeks later. I'd rather the pipeline fail immediately so the operator knows.

PR #16371, opened 2026-02-26
ci(metrics): Add scripts to collect workflow metrics

devashish2203:
- Trigger a workflow at the end of all relevant workflows - post-workflow-actions.yml
- This calls a script .github/scripts/collect_workflow_metrics.py
- Add utilities to for above script. Potentially reusable in other future scripts.

Test Plan:
- [ ] Merge PR
- [ ] Verify Post Workflow Action runs
- [ ] Verify metrics are collected and uploaded.

Planned Changes in subsequent PRs
1. Post the collected Metrics to Posthog
2. Combine Workflow Monitor and this Post Workflow Actions into a single workflow.


devashish2203:
> Just a note that the current workflow-monitor is unaffected by rate-limits. So the retry logic would not get delayed on account of rate limits. Also, the ratelimit will keep the runner alive for the duration of the rate-limited wait time -- and this is would be for all workflow runs during that period. May be this is OK, but something to watch for.

Yeah. I don't expect the rate limits to be hit in this workflow as we are only making 2 API calls, but have kept it in for safety since I don't know how parallel workflow runs will affect things. 

Also trying to create the GithubClient as a reusable piece for future requirements as well.

PR #16372, opened 2026-02-26
fix(ingest/vertexai): fix unthrottled pagination in ML metadata and model list calls

acrylJonny:
## Summary

This PR fixes unthrottled API pagination in the Vertex AI connector that caused `429 ResourceExhausted` errors even when `rate_limit: true` was set, and improves the robustness of rate limiting across all SDK list calls.

### Root cause

The Vertex AI Python SDK returns lazy pager objects from its `list()` methods — the HTTP requests for each page don't fire until the pager is iterated. In two places, a pager was created inside a `with self.rate_limiter:` block but iterated outside it, meaning all page-fetch HTTP calls fired unthrottled:

- **`_load_execution_cache`** (`vertexai_ml_metadata_helper.py`) — `list(itertools.islice(paged_response, 500))` was outside the `with` block, so all bulk execution page fetches bypassed rate limiting entirely.
- **`_list_models`** (`vertexai_model_extractor.py`) — the raw pager was cached rather than a materialised list, so all model page fetches happened later during iteration, again outside the rate limiter context.

### Fix

Rather than patching individual call sites, this PR introduces a `rate_limited_paged_call` utility in `vertexai_utils.py` that wraps the pager's internal `_method` to acquire a rate-limit token per page fetch. A higher-level `rate_limited_gapic_list` helper builds on this to list any SDK resource class with per-page throttling via the underlying GAPIC client, bypassing the lazy pager issue entirely.

All SDK `list()` call sites are migrated to these helpers:

| Extractor | Previously | Now |
|---|---|---|
| `vertexai_ml_metadata_helper.py` | `with rate_limiter: pager = ...; list(islice(pager, n))` outside block | `_list_executions()` via `rate_limited_paged_call` |
| `vertexai_model_extractor.py` | `with rate_limiter: self._models_cache = Model.list(...)` (lazy) | `rate_limited_gapic_list(Model, ...)` |
| `vertexai_model_extractor.py` | `with rate_limiter: list(model.versioning_registry.list_versions())` | `_list_versions()` via `rate_limited_paged_call` on GAPIC |
| `vertexai_model_extractor.py` | `with rate_limiter: list(model.list_model_evaluations())` | `rate_limited_gapic_list(ModelEvaluation, ...)` |
| `vertexai_model_extractor.py` | `with rate_limiter: list(Endpoint.list())` | `rate_limited_gapic_list(Endpoint, ...)` |
| `vertexai_training_extractor.py` | `with rate_limiter: list(dataset_class.list())` | `rate_limited_gapic_list(dataset_class, ...)` |
| `vertexai_experiment_extractor.py` | `aiplatform.Experiment.list()` / `experiment.get_runs()` (unthrottled) | `_list_experiments_rate_limited()` / `_list_experiment_runs_rate_limited()` via `rate_limited_gapic_list` |

### Retry policy change

The original `_METADATA_RETRY` in `vertexai_ml_metadata_helper.py` retried only on `ResourceExhausted` (429). This was inverted: when rate limiting isn't working, retrying on 429 just burns more quota. The new `create_vertex_retry_without_429()` (also used to patch SDK `list()` methods globally via `patch_vertex_sdk_retry`) retries on all transient errors *except* 429, so quota errors surface immediately rather than being hidden behind a 120-second retry deadline.

### `requests_per_min` default change

The default was raised from `60` to `600` to match Google's actual default quota of 600 resource-management requests per minute per region. The previous default of 60 was unnecessarily conservative and made rate-limited ingestion 10× slower than needed. The description is updated to reflect that users should lower this only if sharing quota with other workloads.

### Config: `region` made optional

`region` is changed from a required `str` to `Optional[str]` with `default=None`. A new `require_region_source` model validator enforces that at least one of `region`, `regions`, or `discover_regions=true` is provided — giving a clear error message rather than a pydantic "field required" error, and enabling configurations that use only `regions` or `discover_regions` without providing the deprecated `region` field at all.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/fabric/data_factory/lineage.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/fabric/data_factory/lineage.py
@@ -228,11 +228,16 @@ class CopyActivityLineageExtractor:
         conn_type_props = (dataset_settings.get("connectionSettings") or {}).get(
             "properties", {}
         ).get("typeProperties") or {}
+        ls_type_props = (dataset_settings.get("linkedService") or {}).get(
+            "properties", {}
+        ).get("typeProperties") or {}
         ds_type_props = dataset_settings.get("typeProperties") or {}
 
-        artifact_id: Optional[str] = conn_type_props.get(
-            "artifactId"
-        ) or ds_type_props.get("artifactId")
+        artifact_id: Optional[str] = (
+            conn_type_props.get("artifactId")
+            or ls_type_props.get("artifactId")
+            or ds_type_props.get("artifactId")
+        )
         if not artifact_id:
             logger.debug(
                 "No artifactId found in OneLake datasetSettings for activity '%s'",
@@ -242,6 +247,7 @@ class CopyActivityLineageExtractor:
 
         resolved_workspace_id: str = (
             conn_type_props.get("workspaceId")
+            or ls_type_props.get("workspaceId")
             or ds_type_props.get("workspaceId")
             or pipeline_workspace_id
         )
--- a/metadata-ingestion/tests/unit/fabric_data_factory/test_lineage.py
+++ b/metadata-ingestion/tests/unit/fabric_data_factory/test_lineage.py
@@ -312,6 +312,29 @@ class TestResolveOnelakeUrn:
         assert result is not None
         assert pipeline_ws in result
 
+    def test_linked_service_type_properties(self) -> None:
+        """linkedService.properties.typeProperties provides artifactId and workspaceId."""
+        ds = {
+            "linkedService": {
+                "name": "ConnectorTestLakehouse",
+                "properties": {
+                    "type": "Lakehouse",
+                    "typeProperties": {
+                        "workspaceId": WS_ID,
+                        "artifactId": ARTIFACT_ID,
+                        "rootFolder": "Tables",
+                    },
+                },
+            },
+            "typeProperties": {"schema": "dbo", "table": "Employee"},
+        }
+        result = self.extractor._resolve_onelake_urn(
+            ds, _make_activity(), "fallback-ws"
+        )
+        assert result is not None
+        assert "fabric-onelake" in result
+        assert f"{WS_ID}.{ARTIFACT_ID}.dbo.Employee" in result
+
 
 class TestCopyExtractLineage:
     """End-to-end tests for CopyActivityLineageExtractor.extract_lineage."""


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `1041d35`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/23649646266)

metadata-ingestion/src/datahub/ingestion/source/vertexai/vertexai_config.py
```diff
@@ -174,20 +173,18 @@ class VertexAIConfig(
...
         "Enable if you see '429 Quota Exceeded' errors during ingestion.",
     )
     requests_per_min: int = Field(
-        default=60,
-        description="How many Vertex AI API calls to allow per minute when rate_limit is enabled. "
-        "Start low (30–60) and increase only if ingestion is too slow — some calls fetch multiple "
-        "pages of results internally, so the real quota usage is higher than this number suggests.",
+        default=600,
+        description="Max API requests per minute when rate_limit is enabled. "
+        "Matches Google's default quota of 600 requests/min/region. "
```
alokr-dhub: does google have per user/account quota rate limits? i think using that will be better

metadata-ingestion/src/datahub/ingestion/source/vertexai/vertexai_experiment_extractor.py
```diff
@@ -95,6 +103,68 @@ def __init__(
...
+        )
+        contexts = rate_limited_gapic_list(
+            MetadataContext,
+            self.rate_limiter,
+            parent=self._metadata_store_parent(),
+            filter_str=filter_str,
+        )
+        experiments = []
+        for ctx in contexts:
+            if not Experiment._is_tensorboard_experiment(ctx):
```
alokr-dhub: `Experiment._is_tensorboard_experiment()` is a private method. i think we should avoid accessing private internal methods.

metadata-ingestion/src/datahub/ingestion/source/vertexai/vertexai_config.py
acrylJonny: 600 is actually the hard quota ceiling, not just a default. Per the [Vertex AI quotas and limits docs](https://cloud.google.com/vertex-ai/docs/quotas), the limit for **Resource management (CRUD) requests** is 600 per minute, applied per project per region independently (so a single project scanning two regions effectively gets 600 × 2 = 1,200 requests per minute total). There's no per-user or per-account tier — it's purely project × region.

Since 600 is the ceiling rather than a starting point you'd typically lower, the description is slightly misleading. Will update it to make that clearer:

```python
requests_per_min: int = Field(
    default=600,
    description="Max API requests per minute when rate_limit is enabled. "
    "600 is Google's quota ceiling for resource management requests per project per region "
    "(see https://cloud.google.com/vertex-ai/docs/quotas). Lower this only if you share "
    "quota with other workloads running in the same project and region.",
)

metadata-ingestion/src/datahub/ingestion/source/vertexai/vertexai_experiment_extractor.py
acrylJonny: `_is_tensorboard_experiment` is a single-line check: `return constants.TENSORBOARD_CUSTOM_JOB_EXPERIMENT_FIELD in context.metadata`. Replaced it with that inline check using the public `metadata_constants.TENSORBOARD_CUSTOM_JOB_EXPERIMENT_FIELD` constant directly, so no private method access needed.

alokr-dhub:
@acrylJonny LGTM. Please fix the breaking test.

acrylJonny:
> @acrylJonny LGTM. Please fix the breaking test.

@alokr-dhub This should now be fixed

PR #16176, opened 2026-02-11
refactor(ingest/vertexai): Major code quality improvements and comment cleanup

acrylJonny:
## Summary
Comprehensive refactoring of the Vertex AI connector to improve code quality, maintainability, and consistency with DataHub best practices. This PR addresses multiple code quality issues including DRY violations, magic strings, redundant comments, type safety, and test maintainability.

### **Code Organization & DRY Principle**
- **Extracted 5 new utility modules** to eliminate code duplication:
  - `vertexai_builder.py`: URN generation, name formatting, URL building, URI parsing (250 lines)
  - `vertexai_constants.py`: Centralized magic strings, patterns, defaults (348 lines)
  - `vertexai_models.py`: Pydantic models and dataclasses for type-safe configuration (288 lines)
  - `ml_metadata_helper.py`: ML Metadata API operations (225 lines)
  - `protobuf_utils.py`: Protobuf value extraction utilities (67 lines)
- **Introduced helper method** `_yield_common_aspects()` to eliminate repetitive aspect yielding for `ContainerClass`, `DataPlatformInstanceClass`, and `SubTypesClass`
- Reduced `vertexai.py` complexity while adding functionality

### **Constants & Magic String Elimination**
Created comprehensive constants in `vertexai_constants.py`:
- **`DurationUnit`**: Millisecond/second unit strings
- **`DateTimeFormat`**: ISO 8601 timestamp formats  
- **`LabelFormat`**: Actor label separators
- **`MLModelType`**: Hardcoded model type strings
- **`DatasetTypes`**: All Vertex AI dataset class names
- **`TrainingJobTypes`**: All training job class names
- **`ExternalURLs`**: URL templates for Vertex AI console
- **Pattern matchers**: `HyperparameterPatterns`, `MetricPatterns`, `URIPatterns` with regex-based identification
- **`MLMetadataDefaults`**: ML Metadata configuration defaults

### **Type Safety & Pydantic Models**
- **Replaced manual dictionary parsing** with type-safe Pydantic models:
  - `AutoMLJobConfig`: Parse AutoML job configuration with validation
  - `MLMetadataConfig`: ML Metadata client configuration
  - `LineageMetadata`, `ExecutionMetadata`: Structured lineage data
  - `*CustomProperties` classes: Type-safe custom property builders
- **Improved numeric type checking**: Replaced `isinstance(value, (int, float))` with `numbers.Real` check that handles numpy types, Decimal, and explicitly excludes booleans
[...]

PR #16375, opened 2026-02-27
feat: parallel processing of resources

alokr-dhub:
# Summary
Extending on this PR: https://github.com/datahub-project/datahub/pull/16176  to implement the following functionality:
 - GCP project pattern regex filtering to reject invalid labels and patterns in recipe
 - Parallel processing to process project level resources in parallel

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/vertexai/vertexai.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/vertexai/vertexai.py
@@ -1,5 +1,4 @@
 import logging
-import os
 from contextlib import AbstractContextManager, nullcontext
 from typing import Dict, Iterable, List, Literal, Optional, Union
 
@@ -17,6 +16,7 @@ from google.cloud.aiplatform.models import Model
 from google.cloud.aiplatform_v1 import MetadataServiceClient
 from google.oauth2 import service_account
 
+from datahub.configuration.env_vars import get_vertexai_disable_parallelism
 from datahub.emitter.mcp import MetadataChangeProposalWrapper
 from datahub.emitter.mcp_builder import ProjectIdKey, gen_containers
 from datahub.ingestion.api.common import PipelineContext
@@ -411,11 +411,7 @@ class VertexAISource(StatefulIngestionSourceBase):
                 if self.config.include_pipelines:
                     phase1_resources.append(("pipelines",))
 
-                disable_parallelism = os.environ.get(
-                    "DATAHUB_VERTEXAI_DISABLE_PARALLELISM", ""
-                ).lower() in ("1", "true")
-
-                if not disable_parallelism and len(phase1_resources) > 1:
+                if not get_vertexai_disable_parallelism() and len(phase1_resources) > 1:
                     max_workers = len(phase1_resources)
                     logger.info(
                         "Fetching resources for project=%s region=%s with %d threads",
--- a/metadata-ingestion/src/datahub/configuration/env_vars.py
+++ b/metadata-ingestion/src/datahub/configuration/env_vars.py
@@ -234,6 +234,14 @@ def get_spark_version() -> Optional[str]:
     return os.getenv("SPARK_VERSION")
 
 
+def get_vertexai_disable_parallelism() -> bool:
+    """Disable parallel resource fetching in Vertex AI connector."""
+    return os.getenv("DATAHUB_VERTEXAI_DISABLE_PARALLELISM", "").lower() in (
+        "1",
+        "true",
+    )
+
+
 def get_bigquery_schema_parallelism() -> int:
     """Parallelism level for BigQuery schema extraction."""
     return int(os.getenv("DATAHUB_BIGQUERY_SCHEMA_PARALLELISM", "20"))
--- a/metadata-ingestion/tests/unit/vertexai/test_vertexai_source.py
+++ b/metadata-ingestion/tests/unit/vertexai/test_vertexai_source.py
@@ -263,6 +263,10 @@ def test_parallelism_enabled_automatically(
     assert kwargs["max_workers"] == 3
 
 
+@patch(
+    "datahub.ingestion.source.vertexai.vertexai.get_vertexai_disable_parallelism",
+    return_value=True,
+)
 @patch("datahub.ingestion.source.vertexai.vertexai.MetadataServiceClient")
 @patch("datahub.ingestion.source.vertexai.vertexai.aiplatform.init")
 @patch("datahub.ingestion.source.vertexai.vertexai.ThreadedIteratorExecutor")
@@ -274,10 +278,9 @@ def test_parallelism_disabled_via_env_var(
     mock_executor: MagicMock,
     _mock_init: MagicMock,
     _mock_metadata_client: MagicMock,
-    monkeypatch: pytest.MonkeyPatch,
+    _mock_disable_parallelism: MagicMock,
 ) -> None:
     """DATAHUB_VERTEXAI_DISABLE_PARALLELISM=1 should skip ThreadedIteratorExecutor."""
-    monkeypatch.setenv("DATAHUB_VERTEXAI_DISABLE_PARALLELISM", "1")
     mock_projects.return_value = [PROJECT_ID]
 
     config = VertexAIConfig.model_validate(


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `9afc819`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/23484250296)

gabe-lyons:
routing to @sgomezvillamor

metadata-ingestion/tests/unit/sql_parsing/test_sqlglot_patch.py
```diff
@@ -30,7 +30,7 @@ def test_cooperative_timeout_sql() -> None:
             time.sleep(0.0001)
 
     # To avoid flakiness, this range is quite generous.
-    assert 0.6 <= timer.elapsed_seconds() <= 1.2
+    assert 0.6 <= timer.elapsed_seconds() <= 1.5
```
sgomezvillamor: Instead of extending the range, which was quite generous already, we may add retry.
I did that recently for a performant-dependant test here
https://github.com/datahub-project/datahub/blob/146dff056cd7dadad0bf30646de83aa841ad60e3/metadata-ingestion/tests/performance/sql_parsing/test_sql_aggregator.py#L152-L162

metadata-ingestion/src/datahub/ingestion/source/common/gcp_project_filter.py
sgomezvillamor, line 1: Since `gcp_project_filter` is a common utility, are there other connectors using it?
Maybe `dataplex`? Not very likely, but we should probably check `bigquery` as well.

At the very least, could we update the configs for any connectors that rely on `gcp_project_filter`?

metadata-ingestion/src/datahub/ingestion/source/common/gcp_project_filter.py
```diff
@@ -11,6 +12,83 @@
...
+# GCP label format: key:value, both lowercase letters/digits/underscores/hyphens
+GCP_LABEL_PATTERN = re.compile(r"^[a-z][a-z0-9_-]*:[a-z0-9_-]+$")
+
+
+class GCPValidationError(ValueError):
+    """Raised when GCP configuration values fail format validation."""
+
+    pass
+
+
```
sgomezvillamor, line 26: how much of these validations would be automatically handled by pydantic itself?

does `set`/`frozenset` handle the duplication validation? https://docs.pydantic.dev/2.0/usage/types/set_types/

does `pattern` solve our use case? 
https://docs.pydantic.dev/2.0/usage/types/string_types/#constrained-types

metadata-ingestion/src/datahub/ingestion/source/vertexai/vertexai_config.py
```diff
@@ -72,7 +86,15 @@ class VertexAIConfig(
     credential: Optional[GCPCredential] = Field(
         default=None, description="GCP credential information"
     )
-    project_id: str = Field(description=("Project ID in Google Cloud Platform"))
+
+    project_id: Optional[str] = Field(
```
sgomezvillamor: we have `pydantic_removed_field` for removing a field in a backwards compatible fashion https://github.com/datahub-project/datahub/blob/146dff056cd7dadad0bf30646de83aa841ad60e3/metadata-ingestion/src/datahub/configuration/validate_field_removal.py#L12-L16

maybe we could just handle removal and migration at the same time?

metadata-ingestion/src/datahub/ingestion/source/vertexai/vertexai.py
```diff
@@ -404,22 +383,37 @@ def get_workunits_internal(self) -> Iterable[MetadataWorkUnit]:
...
+
+                if (
+                    self.config.max_threads_resource_parallelism > 1
+                    and len(phase1_resources) > 1
+                ):
+                    logger.info(
+                        "Fetching resources for project=%s region=%s with %d threads",
+                        project_id,
+                        region,
+                        self.config.max_threads_resource_parallelism,
```
sgomezvillamor: Assuming N threads, do I understand correctly that we have at most 3 fixed and known resource types? If so, the expectation would be that the user sets N accordingly to get the right level of parallelism?

If that’s the case, I think we’re exposing unnecessary complexity to the user. Overall, I’m concerned about pushing this complexity to the user when we likely have enough information on our side to make a better decision automatically.

My suggestion would be:

* Set max workers in the Thread Pool internally based on the number of resource types to be ingested, so we effectively have one thread per resource type.
* Optionally allow disabling this parallelism via an env var (not even exposed in the config).

WDYT?

metadata-ingestion/tests/unit/sql_parsing/test_sqlglot_patch.py
alokr-dhub: ACK. fixed this

metadata-ingestion/src/datahub/ingestion/source/common/gcp_project_filter.py
alokr-dhub: Yes, BigQuery also uses gcp_project_filter. We verified that BigQuery already handles its own project_id deprecation/migration independently in
  BigQueryFilterConfig. Dataplex has its own project_ids field but doesn't use the shared utility. No config changes are needed for other connectors as part of this PR.The shared
  GcpProjectFilterConfig and its functions remain unchanged and backward-compatible.

metadata-ingestion/src/datahub/ingestion/source/common/gcp_project_filter.py
alokr-dhub: Pydantic's Set handles dedup but loses ordering and is a breaking change for the shared GcpProjectFilterConfig (used by BigQuery).
  constr(pattern=...) works but produces less helpful error messages. We simplified VertexAI's validators to use the regex constants directly with dict.fromkeys() for ordered dedup,
  which gets most of the benefit without the downsides.

metadata-ingestion/src/datahub/ingestion/source/vertexai/vertexai_config.py
alokr-dhub: I considered pydantic_removed_field but it only discards the field. It can't migrate the value to project_ids. Since we need backward-compatible migration (not just removal), we use a mode="before" model validator that pops project_id, moves it into project_ids, and logs a deprecation warning. The pattern is the same as pydantic_removed_field under the hood (both are mode="before" validators that pop the field), just with the added migration step. Once we're ready to fully remove project_id support in a future major release, we can swap this validator for pydantic_removed_field directly.

metadata-ingestion/src/datahub/ingestion/source/vertexai/vertexai.py
alokr-dhub: Already had checks to cap parallelism to the number or resources in each phase.I have made it implicit as suggested with a flag to enable and disable parallelism

metadata-ingestion/docs/sources/vertexai/vertexai_pre.md
```diff
@@ -97,3 +97,165 @@ Grant the following permissions to the service account on all target projects.
     client_email: "test@suppproject-id-1234567.iam.gserviceaccount.com"
     client_id: "123456678890"
   ```
+
+#### Integration Details
```
sgomezvillamor: The `Integration Details` (H4) section inside `Prerequisites` (H3) doesn't look nice and it's mixing different things: prerequisites, capabilities, limitations, concept mapping, etc.

Could you take a look at https://github.com/datahub-project/datahub/blob/master/metadata-ingestion/docs/sources/AGENTS.md and reorganize the docs accordingly?

Also, I'd avoid language like "New Features" — it won't age well 😅

metadata-ingestion/docs/sources/vertexai/vertexai_pre.md
```diff
@@ -97,3 +97,165 @@ Grant the following permissions to the service account on all target projects.
...
+**Performance**: Resources are fetched ordered by update time (most recently updated first). Limits like `max_training_jobs_per_type` cap how many resources are processed per run — for example, `max_training_jobs_per_type: 1000` will process only the 1000 most recently updated training jobs of each type.
+
+Enabling `stateful_ingestion` has two effects: (1) resources not updated since the previous run are skipped, reducing redundant API calls on subsequent runs; and (2) entities deleted from Vertex AI are automatically soft-deleted in DataHub. Use `stateful_ingestion.ignore_old_state: true` to get soft-deletion only without the incremental skip behaviour.
+
+For improved organization in the DataHub UI:
+
+- Model versions are organized under their respective model group folders
+- Pipeline tasks and task runs are nested under their parent pipeline folders
+
+##### New Features for CustomJob Lineage
```
sgomezvillamor: eg, this should go inside "Capabilities" as eg "CustomJob Lineage"

metadata-ingestion/docs/sources/vertexai/vertexai_pre.md
```diff
@@ -97,3 +97,165 @@ Grant the following permissions to the service account on all target projects.
...
+
+The connector now supports extracting lineage and metrics from CustomJob training jobs using the **Vertex AI ML Metadata API**. This enables:
```
sgomezvillamor: The connector supports ....

metadata-ingestion/docs/sources/vertexai/vertexai_pre.md
```diff
@@ -97,3 +97,165 @@ Grant the following permissions to the service account on all target projects.
...
+- **Hyperparameters and metrics extraction** from training jobs that log to ML Metadata Executions
+- **Model evaluation** ingestion with evaluation metrics and lineage to models
+
+These features are controlled by new configuration options:
+
+- `use_ml_metadata_for_lineage` (default: `true`) - Extracts lineage from ML Metadata for CustomJob and other non-AutoML training jobs
+- `extract_execution_metrics` (default: `true`) - Extracts hyperparameters and metrics from ML Metadata Executions
+- `include_evaluations` (default: `true`) - Ingests model evaluations and evaluation metrics
+
+**Note**: For CustomJob lineage to work, your training jobs must log to Vertex AI ML Metadata. This happens automatically when using Vertex AI Experiments SDK or manually logging artifacts/executions to ML Metadata.
```
sgomezvillamor:
```suggestion
:::tip "Log to Vertex AI ML Metadata"

For CustomJob lineage to work, your training jobs must log to Vertex AI ML Metadata. This happens automatically when using Vertex AI Experiments SDK or manually logging artifacts/executions to ML Metadata.
:::
```

metadata-ingestion/docs/sources/vertexai/vertexai_pre.md
```diff
@@ -97,3 +97,165 @@ Grant the following permissions to the service account on all target projects.
...
+
+##### Concept Mapping
```
sgomezvillamor: Concept Mapping is expected to be a H2 section in the platform doc, that is metadata-ingestion/docs/sources/vertexai/README.md
Add it there or update it if there is one already

metadata-ingestion/docs/sources/vertexai/vertexai_pre.md
```diff
@@ -97,3 +97,165 @@ Grant the following permissions to the service account on all target projects.
...
+|| [`PipelineJob Task`](https://cloud.google.com/vertex-ai/docs/pipelines/build-pipeline#understanding-pipelines) | [`DataJob`](https://docs.datahub.com/docs/generated/metamodel/entities/datajob/) | Each task within a Vertex AI pipeline is modeled as a DataJob in DataHub, nested under its parent pipeline DataFlow. Tasks represent individual steps in the pipeline workflow. |
+|| [`PipelineJob Task Run`](https://cloud.google.com/vertex-ai/docs/pipelines/build-pipeline#understanding-pipelines) | [`DataProcessInstance`](https://docs.datahub.com/docs/generated/metamodel/entities/dataprocessinstance/) | Each execution of a pipeline task is modeled as a DataProcessInstance, linked to its DataJob (task definition). This captures runtime metadata, inputs/outputs, and lineage for each task execution. |
+
+Vertex AI Concept Diagram:
+
+<p align="center">
+  <img width="70%"  src="https://raw.githubusercontent.com/datahub-project/static-assets/main/imgs/metadata-ingestion/vertexai/concept-mapping.png"/>
+</p>
+
+##### Lineage
```
sgomezvillamor: --> to be moved somewhere inside Capabilities

metadata-ingestion/src/datahub/ingestion/source/vertexai/vertexai.py
```diff
@@ -404,22 +400,39 @@ def get_workunits_internal(self) -> Iterable[MetadataWorkUnit]:
...
-                    )
-
+                    phase1_resources.append(("experiments",))
                 if self.config.include_pipelines:
-                    yield from auto_workunit(self.pipeline_extractor.get_workunits())
+                    phase1_resources.append(("pipelines",))
+
+                disable_parallelism = os.environ.get(
+                    "DATAHUB_VERTEXAI_DISABLE_PARALLELISM", ""
+                ).lower() in ("1", "true")
```
sgomezvillamor: `metadata-ingestion/src/datahub/configuration/env_vars.py` centralizes all env vars

alokr-dhub: moved to env vars

PR #16382, opened 2026-02-27
feat(search): surface matched chunks in semantic search results for API callers

shirshanka:
This PR:
Adds chunk-level highlighting to semantic search: the nested kNN query now includes inner_hits to surface the specific passage that drove each document's ranking.

It does NOT:
Add any changes to the UI to render such highlights since semantic search is not wired to the UI yet.

Key changes:
- New SemanticMatchedChunk PDL record (position, text, characterOffset, characterLength, score) and SemanticMatchedChunkArray typeref added to SearchEntity
- SemanticEntitySearchService adds inner_hits to the nested kNN query and extracts matched chunks via extractMatchedChunks(); _source uses excludes filter to drop the dense vector while returning all other chunk fields
- matchedChunksPerResult (default 1) wired from SemanticSearchConfiguration through SemanticEntitySearchServiceFactory into the service as an instance field
- GraphQL: extend type SearchResult with matchedChunks: [SemanticMatchedChunk!]! and new SemanticMatchedChunk type in semantic-search.graphql
- MapperUtils maps SearchEntity.semanticMatchedChunks to GraphQL SemanticMatchedChunk
- Smoke tests: timing instrumentation, mae_only=True (correct consumer group for quickstartDebug), removed hardcoded 20s sleep in favour of wait_for_writes_to_sync

### DIFF
--- a/smoke-test/tests/semantic/test_semantic_search.py
+++ b/smoke-test/tests/semantic/test_semantic_search.py
@@ -664,6 +664,25 @@ class TestSemanticSearchWithIngestion:
                         assert len(chunk["text"]) > 0, (
                             f"chunk[{i}].text is empty for {entity_urn}"
                         )
+                        # Verify the chunk text contains words from the document we
+                        # ingested. We don't do an exact substring match because the
+                        # chunking library normalizes whitespace (e.g. list-item newlines
+                        # become spaces). Instead we check for a few distinctive phrases
+                        # that must survive any reasonable normalization.
+                        doc_title = focused_docs[0]["title"]
+                        doc_text = focused_docs[0]["text"]
+                        distinctive_phrases = [
+                            w
+                            for w in doc_text.split()
+                            if len(w) > 6  # skip short/common words
+                        ]
+                        chunk_words = set(chunk["text"].split())
+                        overlap = [p for p in distinctive_phrases if p in chunk_words]
+                        assert len(overlap) >= 3, (
+                            f"chunk[{i}].text for {entity_urn} shares fewer than 3 words "
+                            f"with the ingested document '{doc_title}' — likely wrong content.\n"
+                            f"  chunk text (first 300): {chunk['text'][:300]!r}"
+                        )
 
                     if has_offset:
                         assert chunk.get("characterLength") is not None, (


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `f2d1420`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22520648080)

alexsku:
## Review — Issues & Questions

### Medium Issues

#### 1. `NullNode` handling in `extractMatchedChunks()` can surface literal string `"null"`

**Category**: correctness | **Confidence**: medium
**Location**: `SemanticEntitySearchService.java` — `extractMatchedChunks()`

If OpenSearch stores an explicit JSON `null` for the `text`, `characterOffset`, or `characterLength` fields in the nested chunk source, Jackson represents it as a `NullNode`. `NullNode.isMissingNode()` returns `false`, so the guard passes. Then `NullNode.asText()` returns the literal string `"null"`, and `NullNode.asInt()` returns `0`.

This means an explicit-null `text` field would surface as `text = "null"` to API consumers, and explicit-null offset/length would silently become `0` instead of remaining unset.

**Suggested fix**: Add `&& !src.path("text").isNull()` to each guard:
```java
if (!src.path("text").isMissingNode() && !src.path("text").isNull()) {
    chunk.setText(src.path("text").asText());
}
```
Same for `characterOffset` and `characterLength`.

**Suggested test**: A unit test where the chunk `_source` contains `"text": null` and asserts `chunk.hasText()` is `false`.

---

#### 2. No null-guard on `getMatchedChunks()` input

**Category**: correctness | **Confidence**: low
**Location**: `MapperUtils.java` — `getMatchedChunks()`

The method calls `.stream()` on the input list directly. The PDL default `= []` makes this safe under normal operation, but if a `SearchEntity` is constructed manually (e.g., in tests or a future code path) without setting `semanticMatchedChunks`, a defensive null check is cheap insurance.

**Suggested fix**:
```java
if (pdlChunks == null || pdlChunks.isEmpty()) {
    return Collections.emptyList();
}
```

---

#### 3. No unit test asserting the `_source.excludes` pattern in the query body

**Category**: tests | **Confidence**: high
**Location**: `SemanticEntitySearchServiceTest.java`

The `_source.excludes` filter on inner_hits is critical — without it, every chunk returns ~12KB of dense vector data. There's a test for `inner_hits.size` (`testFiveArgConstructorSetsMatchedChunksPerResult`), but no test asserting the excludes pattern is present. If someone refactors the query builder and accidentally drops the excludes, it would silently return large payloads with no test failure.

**Suggested fix**: Add an assertion in the existing constructor test that captures the request body and checks:
```java
assertTrue(requestBody.contains("\"excludes\""), "inner_hits should exclude vector field from _source");
assertTrue(requestBody.contains(".vector"), "excludes should target the vector field");
```

---

#### 4. `matchedChunksPerResult` not validated at construction time

**Category**: edge-case | **Confidence**: medium
**Location**: `SemanticEntitySearchService.java` — constructor

The `matchedChunksPerResult` parameter is passed directly to OpenSearch's `inner_hits.size` without validation. A value of `0` would silently disable the feature. A negative value would likely cause an OpenSearch error at query time.

**Suggested fix**: Add a precondition in the constructor:
```java
if (matchedChunksPerResult < 1) {
    throw new IllegalArgumentException(
        "matchedChunksPerResult must be >= 1, got " + matchedChunksPerResult);
}
```

---

#### 5. Snapshot files contain unrelated changes

**Category**: dx | **Confidence**: high
**Location**: `metadata-service/restli-api/src/main/snapshot/*.snapshot.json`

The five snapshot files add ~400 lines of diff unrelated to semantic search (`sanitizeRichText` additions, `fieldName: "id"`, `SchemaFieldPath` doc updates, `lastModifiedAt` Searchable annotations). While harmless, they make the diff harder to review and obscure the actual semantic search changes.

---

### Questions for Author

1. The restli snapshot diffs contain many changes unrelated to semantic search (`sanitizeRichText`, `fieldName:"id"`, `SchemaFieldPath` doc updates). Were these picked up from a rebase/snapshot regeneration, or are they intentionally part of this branch?

2. The `_source.excludes` pattern uses the full nested path (`embeddings.text_embedding_3_large.chunks.vector`). The inline comment explains why `excludes` is used instead of `includes` — was this behavior verified against the specific OpenSearch version DataHub targets (2.17+)? Any risk of behavioral differences across ES/OS versions?

3. Is there a plan to add a unit test asserting the `_source.excludes` pattern in the generated query body, similar to how `testFiveArgConstructorSetsMatchedChunksPerResult` already verifies the `inner_hits.size`?

4. The `extractMatchedChunks()` method always sets `score` (even when `_score` might be 0.0 from a missing node), while the PDL declares score as `optional float`. Was the intent to always populate score for inner_hits (where `_score` is guaranteed), or should there be a missing-node guard matching the pattern used for `text`/`characterOffset`/`characterLength`?

alexsku:
## API Proposal: Unify Semantic Chunk Highlighting with `matchedFields`

### Problem

The current PR introduces a parallel API surface for semantic search highlighting:

| | Keyword search | Semantic search (this PR) |
|---|---|---|
| **PDL type** | `MatchedField {name, value}` | `SemanticMatchedChunk {position, text, characterOffset, characterLength, score}` |
| **PDL field** | `SearchEntity.matchedFields` | `SearchEntity.semanticMatchedChunks` |
| **GraphQL type** | `MatchedField {name, value, entity}` | `SemanticMatchedChunk {position, text, characterOffset, characterLength, score}` |
| **GraphQL field** | `SearchResult.matchedFields` | `SearchResult.matchedChunks` |
| **UI rendering** | Full pipeline: field config registry, grouping, prioritization, description slicing | Nothing yet (explicitly out of scope) |

Both answer the same user question — *"why did this result match?"* — but through divergent types, fields, and code paths. When hybrid search arrives (keyword + semantic), consumers would need to read from two different places and reconcile them.

### Proposed Alternative

Extend the existing `MatchedField` with two optional fields and funnel **both** keyword highlights and semantic chunks through the single `matchedFields` array.

#### PDL Changes

**`MatchedField`** (in `SearchEntity.pdl`) — add two optional fields:

```pdl
matchedFields: array[record MatchedField {
  name: string
  value: string

  /** Relevance or similarity score for this match. Present for semantic matches. */
  score: optional float

  /** Zero-based position of this match within the document. Present for semantic chunk matches. */
  position: optional int
}] = []
```

**`SearchEntity.pdl`** — no new `semanticMatchedChunks` field needed.

**`SemanticMatchedChunk.pdl`** — not needed, delete.

#### GraphQL Changes

**`MatchedField`** (in `search.graphql`) — add two optional fields:

```graphql
type MatchedField {
  name: String!
  value: String!
  entity: Entity

  """
  Relevance or similarity score for this match.
  Present for semantic search matches (cosine similarity).
  Null for keyword search matches.
  """
  score: Float

  """
  Zero-based position of this match within the document.
  Present for semantic chunk matches. Null for keyword matches.
  """
  position: Int
}
```

**`semantic-search.graphql`** — no `extend type SearchResult` with `matchedChunks`, no `SemanticMatchedChunk` type needed.

#### What Gets Dropped

- `SemanticMatchedChunk` PDL record
- `SemanticMatchedChunk` GraphQL type
- `SearchEntity.semanticMatchedChunks` field
- `SearchResult.matchedChunks` GraphQL field
- `MapperUtils.getMatchedChunks()` method
- `characterOffset` and `characterLength` — these exist for the case "when chunk text is absent," but inner_hits always returns text via `_source`. If a future need arises, these can be added at that point (backward-compatible since they'd be optional).

#### Population Logic

**Keyword search** (unchanged):
```
name = "description"          value = "...highlighted snippet..."
name = "fieldPaths"           value = "customer_id"
```

**Semantic search** (new — in `SemanticEntitySearchService.extractMatchedChunks()`):
```
name = "semanticContent"      value = "The actual matched passage text..."
                              score = 0.92
                              position = 0
```

The `SemanticEntitySearchService` would populate the existing `matchedFields` array on `SearchEntity` instead of a separate `semanticMatchedChunks` array. Each inner_hit chunk becomes a `MatchedField` with `name = "semanticContent"`, `value` = chunk text, plus the new optional fields.

**Hybrid search** (future): both keyword and semantic entries coexist naturally in the same array. Consumers iterate one list.

#### UI Integration

The existing `constants.ts` field config registry just needs one new entry:

```typescript
{
    name: 'semanticContent',
    label: 'matched passage',
    showInMatchedFieldList: true,
},
```

The existing rendering pipeline (grouping, prioritization, description slicing) mostly works as-is. A small helper distinguishes semantic fields:

```typescript
export const isSemanticField = (field: MatchedField) => field.name === 'semanticContent';
```

The `score` field could optionally be displayed as a confidence indicator — the data is available whenever the UI is ready for it.

### Tradeoffs vs. Current PR Approach

| Dimension | Current PR (`matchedChunks`) | This proposal (extend `matchedFields`) |
|---|---|---|
| **API surface** | +1 PDL record, +1 GraphQL type, +2 new fields | +2 optional fields on existing types |
| **Client integration** | Clients must read a new field for semantic results | Clients already reading `matchedFields` get semantic highlights for free |
| **Hybrid search readiness** | Consumers reconcile two arrays | One array, naturally mixed |
| **Type safety of metadata** | All chunk metadata is strongly typed | `score` and `position` are typed; `characterOffset`/`characterLength` deferred |
| **UI effort** | Entirely new rendering path | One config entry + one small helper |
| **Loss** | None | `characterOffset`/`characterLength` deferred (not needed while inner_hits returns text) |
| **Extensibility** | Easy to add fields to `SemanticMatchedChunk` | Easy to add optional fields to `MatchedField`; or can add a dedicated type later if the models truly diverge |

### Recommendation

The unified approach is simpler for consumers today and positions the API better for hybrid search. The only real cost is deferring `characterOffset`/`characterLength`, which are speculative — they serve a "text not stored" scenario that doesn't currently exist. If that scenario materializes, those fields can be added to `MatchedField` at that point (backward-compatible, since they'd be optional).

PR #16383, opened 2026-02-27
 test(frontend): use MockLink catch-all handler to fix Routes.test.tsx timeout

daha:
## Summary

The `renders embed page properly` test in `Routes.test.tsx` times out on fork CI runners (2-CPU). The root cause is the cumulative cost of the full `<Routes />` component tree rendering, compounded by error-handling and retry logic in components that receive errors from unmatched GraphQL queries. Apollo's `MockLink` calls `observer.error()` for any query without a matching mock, and adding mocks one-by-one proved to be a losing game as the component tree fires many queries.

**Fix:** Create a custom `MockLink` with a `setOnError` catch-all handler that returns empty data (`{ data: {} }`) for any unmatched query instead of erroring. This eliminates the error-handling overhead and makes the test resilient to
future component changes that introduce new queries. Also increases the test timeout to 15s as a safety net for slow runners.

Additionally keeps the `manageServiceAccounts: true` fix added to the inline `GetMe` mock in a prior commit, and removes the individual mocks that were added in previous attempts (`GetGlobalViewsSettings`, `GetHomePageSettings`,
`BatchGetStepStates`, `GetSearchAcrossLineageCounts`) since the catch-all makes them unnecessary.

## Checklist


gabe-lyons (approved):

gabe-lyons:
thanks for the contribution @daha

PR #16386, opened 2026-02-27
fix(dagster): preserve DataJob lineage on failed/canceled runs

### LINKED ISSUE #16024: [dagster-plugin] Datajob op lineage in DataHub removed after a failed materialization
**Describe the bug**
After succesfully running a DataJob with appropriate `output.datahub` metadata set the lineage is pushed to DataHub. On a subsequent run, if that asset materialization fails, the datahub sensor will remove the lineage from the DataJob which is registered in DataHub.

The problem is that the lineage built is using the Dagster output_events which are not emitted from a failed materialization. Even though I explicitly set the `output.datahub` metadata on the Dagster asset the lineage is not taken into account:

```python
@dg.asset(
    name="material2_dim_def_def",
    metadata={
        "datahub.output": [
            DatasetUrn(
                platform="databricks",
                name="hdp_cma_prd.shared_material_curated.material2_dim_def_def",
                env="PROD",
            ).urn()
        ]
    },
)
def material2_dim_def_def(context: dg.AssetExecutionContext):    
    # raise ValueError("Simulated failure for testing purposes.")
    
    return {
        "table": "material2_dim_def_def",
    }
```

**To Reproduce**

1. Materialize an Asset which succeeds.
2. Observe the lineage correctly build
3. Make the Asset materialization fail on a subsequent run
4. Lineage will be removed

**Expected behavior**
I expect the lineage to stay even though a DataJob does not correctly run.

**Screenshots**

Image

treff7es:
## Summary
- Fixes a bug where the dagster-plugin sensor would overwrite existing DataJob lineage with empty values when a Dagster run fails or is canceled
- When no lineage events are collected for an op in a failed run, the `DataJobInputOutput` aspect is now skipped to preserve existing lineage from the last successful run
- If an op has static lineage from `datahub.inputs`/`datahub.outputs` metadata on its definition, that lineage is still emitted even on failure

Closes #16024

## Root Cause
The plugin registers sensors for SUCCESS, FAILURE, and CANCELED run statuses — all calling the same `_emit_metadata()` function. For failed runs, no `ASSET_MATERIALIZATION` events exist, so `DataJob.emit()` would emit a `DataJobInputOutputClass` aspect with empty lists, overwriting existing lineage. Additionally, `DataProcessInstance.emit_process_start()` re-emits the parent DataJob template (including lineage) by default.

## Changes
- **`datahub_sensors.py`**: Detect failed/canceled runs and skip lineage emission for ops without lineage data
- **`dagster_generator.py`**: Add `emit_template` parameter to `emit_op_run()` to control parent DataJob re-emission from `DataProcessInstance`
- **`test_dagster.py`**: Add 5 new tests covering failure scenarios, partial failures, static metadata preservation, run status tracking, and success regression

## Test plan
- [x] Single-op failure: no `DataJobInputOutput` emitted
- [x] Failed run still emits `DataProcessInstance` (run tracking works)
- [x] Failed op with `datahub.inputs` metadata: lineage IS still emitted
- [x] Partial failure (multi-op): no empty lineage overwrites
- [x] Successful run regression: lineage still emitted normally
- [x] Existing golden-file test still passes


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `442a7a0`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22625403956)

metadata-ingestion-modules/dagster-plugin/src/datahub_dagster_plugin/sensors/datahub_sensors.py
```diff
@@ -845,7 +852,15 @@ def _emit_metadata(self, context: RunStatusSensorContext) -> SensorReturnTypesUn
                     )
                     datajob.name = datajob.name.split("__")[-1]
 
-                datajob.emit(self.graph)
+                op_has_lineage = bool(datajob.inlets) or bool(datajob.outlets)
+                skip_lineage = run_is_not_successful and not op_has_lineage
```
sgomezvillamor, line 856: Just wondering: should we skip regardless of `op_has_lineage`?
I mean, what's the point of emitting lineage when: `run_is_not_successful and op_has_lineage`?

sgomezvillamor (approved):
LGTM
Just left a question :-)

metadata-ingestion-modules/dagster-plugin/src/datahub_dagster_plugin/sensors/datahub_sensors.py
treff7es: Well, in case we get lineage but the job fails, we might be ok

PR #16387, opened 2026-02-27
feat(gms) Add backend resolvers to support multiple data products per asset

chriscollins3456:
This PR is part 1 of 2 in order to support the option of having multiple data products associated with a single asset.

This PR handles the new backend endpoints that support adding multiple products to a list of assets, instead of just a single data product like our previous mutation. This is supported behind a feature flag.

The other major change is to disable the side effect to remove assets from additional data products if the feature flag to allow multiple data products is enabled.

The next PR will include the frontend changes as well hidden behind a feature flag to enable this for folks.

### DIFF
--- a/metadata-service/services/src/test/java/com/linkedin/metadata/service/DataProductServiceTest.java
+++ b/metadata-service/services/src/test/java/com/linkedin/metadata/service/DataProductServiceTest.java
@@ -23,7 +23,6 @@ import com.linkedin.entity.client.EntityClient;
 import com.linkedin.metadata.Constants;
 import com.linkedin.metadata.aspect.AspectRetriever;
 import com.linkedin.metadata.graph.GraphClient;
-import com.linkedin.metadata.resource.ResourceReference;
 import com.linkedin.metadata.utils.GenericRecordUtils;
 import com.linkedin.mxe.MetadataChangeProposal;
 import io.datahubproject.metadata.context.OperationContext;
@@ -80,12 +79,9 @@ public class DataProductServiceTest {
             any(OperationContext.class), any(MetadataChangeProposal.class), anyBoolean()))
         .thenReturn(TEST_DATA_PRODUCT_URN.toString());
 
-    final List<ResourceReference> resources =
-        ImmutableList.of(
-            new ResourceReference(TEST_RESOURCE_URN_1, null, null),
-            new ResourceReference(TEST_RESOURCE_URN_2, null, null));
+    final List<Urn> resources = ImmutableList.of(TEST_RESOURCE_URN_1, TEST_RESOURCE_URN_2);
 
-    service.batchAddToDataProduct(opContext, TEST_DATA_PRODUCT_URN, resources, null);
+    service.batchAddToDataProduct(opContext, TEST_DATA_PRODUCT_URN, resources);
 
     final ArgumentCaptor<MetadataChangeProposal> mcpCaptor =
         ArgumentCaptor.forClass(MetadataChangeProposal.class);
@@ -151,12 +147,9 @@ public class DataProductServiceTest {
             any(OperationContext.class), any(MetadataChangeProposal.class), anyBoolean()))
         .thenReturn(TEST_DATA_PRODUCT_URN.toString());
 
-    final List<ResourceReference> resources =
-        ImmutableList.of(
-            new ResourceReference(TEST_RESOURCE_URN_1, null, null),
-            new ResourceReference(TEST_RESOURCE_URN_2, null, null));
+    final List<Urn> resources = ImmutableList.of(TEST_RESOURCE_URN_1, TEST_RESOURCE_URN_2);
 
-    service.batchAddToDataProduct(opContext, TEST_DATA_PRODUCT_URN, resources, null);
+    service.batchAddToDataProduct(opContext, TEST_DATA_PRODUCT_URN, resources);
 
     final ArgumentCaptor<MetadataChangeProposal> mcpCaptor =
         ArgumentCaptor.forClass(MetadataChangeProposal.class);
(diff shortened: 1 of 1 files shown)


metadata-service/services/src/main/java/com/linkedin/metadata/service/DataProductService.java
```diff
@@ -309,15 +311,13 @@ public void deleteDataProduct(@Nonnull OperationContext opContext, @Nonnull Urn
...
-   * @param dataProductUrn the urn of the Data Product to set - null if removing Data Product
-   * @param resourceUrns the urns of the entities to add the Data Product to
-   * @param authentication the current authentication
+   * @param dataProductUrn the urn of the Data Product to set
+   * @param resources references to the resources to change
    */
   public void batchSetDataProduct(
       @Nonnull OperationContext opContext,
       @Nonnull Urn dataProductUrn,
-      @Nonnull List<Urn> resourceUrns,
```
ani-malgari, line 319: any reason why we changed to passing resources? all we need is still the urns, and we do not need to extract the urns here. Not sure if this abstraction is necessary here.

metadata-service/services/src/main/java/com/linkedin/metadata/service/DataProductService.java
```diff
@@ -399,6 +475,35 @@ public void unsetDataProduct(
...
+  /**
+   * Batch unsets Data Products for a given list of entities. Remove these entities from their data
+   * product(s).
+   *
+   * <p>Note that this method does not do authorization validation. It is assumed that users of this
+   * class have already authorized the operation
+   *
+   * @param resources references to the resources to change
+   */
+  public void batchUnsetDataProduct(
```
ani-malgari, line 484: unrelated to this PR: any reason why we have inconsistencies between how it's handled for setting and unsetting data products? 
batchSetDataProduct -> sets a data product for a list of resources
batchUnsetDataProduct -> looks up data products for the list of resources, and unsets all.

Is there an opportunity to fix this?

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/dataproduct/DataProductResolverUtils.java
```diff
@@ -0,0 +1,121 @@
...
+   * them.
+   *
+   * @param resources list of resource URN strings to verify
+   * @param context the GraphQL query context
+   * @param dataProductService service to check entity existence
+   * @throws RuntimeException if any resource does not exist
+   * @throws AuthorizationException if user is not authorized to update data products for any
+   *     resource
+   */
+  public static void verifyResources(
```
ani-malgari, line 29: why don't we just move this validation/authorization logic to the service itself? it would make the resolver clean without needing this class at all.

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/dataproduct/DataProductResolverUtils.java
```diff
@@ -0,0 +1,121 @@
...
+      @Nonnull final List<String> dataProductUrns,
+      @Nonnull final QueryContext context,
+      @Nonnull final DataProductService dataProductService) {
+    for (String dataProductUrn : dataProductUrns) {
+      if (!dataProductService.verifyEntityExists(
+          context.getOperationContext(), UrnUtils.getUrn(dataProductUrn))) {
+        throw new RuntimeException(
+            String.format(
+                "Failed to update data products, data product %s does not exist", dataProductUrn));
+      }
```
ani-malgari: nit: we could just call `verifyDataProduct`

metadata-service/services/src/main/java/com/linkedin/metadata/service/DataProductService.java
chriscollins3456: tbh no not really anything specific. i think my main thinking was that if we eventually needed the other pieces like `subResource` then we'd be good to go. but not sure if that will need to be the case so keeping at urns only could make more sense

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/dataproduct/DataProductResolverUtils.java
chriscollins3456: i think mostly because the clients of these different lower level services handle authentication differently... but tbh not positive there's a great reason and would want to chat with some others. but as of now this is the pattern we've established..

metadata-service/services/src/main/java/com/linkedin/metadata/service/DataProductService.java
chriscollins3456: ah you mean you would expect `batchUnsetDataProduct` to unset a specific data product for a list of resources? basically what my new `batchRemoveFromDataProduct` does?

i think this weirder naming just comes from the original implementation of only one data product can exist for any given asset, so setting and unsetting was the only options kind of

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/dataproduct/DataProductResolverUtils.java
chriscollins3456: makes sense!

metadata-service/services/src/main/java/com/linkedin/metadata/service/DataProductService.java
chriscollins3456: made the change to revert this change

PR #16388, opened 2026-02-27
feat(ui) Add support for multiple data products per asset - frontend

chriscollins3456:
This PR is built on top of https://github.com/datahub-project/datahub/pull/16387

Hidden behind a feature flag this PR enables the ability to support multiple data products on a single asset, as opposed to the legacy behavior of a single one.

This includes updating the entity profile page where you can add data products, as well as the ability to bulk set data products through the search page where we support bulk actions.

I also added a few other minor UX fixes like a better sized create DP modal.

### DIFF
--- a/datahub-web-react/src/app/entityV2/shared/containers/profile/sidebar/DataProduct/SetDataProductModal.tsx
+++ b/datahub-web-react/src/app/entityV2/shared/containers/profile/sidebar/DataProduct/SetDataProductModal.tsx
@@ -5,6 +5,7 @@ import React, { useMemo, useRef, useState } from 'react';
 import styled from 'styled-components';
 
 import analytics, { EntityActionType, EventType } from '@app/analytics';
+import { useEntityContext } from '@app/entity/shared/EntityContext';
 import { getParentEntities } from '@app/entityV2/shared/containers/profile/header/getParentEntities';
 import { handleBatchError } from '@app/entityV2/shared/utils';
 import ContextPath from '@app/previewV2/ContextPath';
@@ -22,7 +23,6 @@ import { getModalDomContainer } from '@src/utils/focus';
 import { useBatchAddToDataProductsMutation, useBatchSetDataProductMutation } from '@graphql/dataProduct.generated';
 import { useGetAutoCompleteMultipleResultsLazyQuery } from '@graphql/search.generated';
 import { DataHubPageModuleType, DataProduct, Entity, EntityType } from '@types';
-import { useEntityContext } from '@app/entity/shared/EntityContext';
 
 const LoadingWrapper = styled.div`
     display: flex;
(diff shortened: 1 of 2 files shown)


datahub-web-react/src/app/entityV2/shared/components/styled/search/action/DataProductsDropdown.tsx
```diff
@@ -52,16 +54,16 @@ export default function DataProductsDropdown({ urns, disabled = false, refetch }
...
                     {
-                        title: 'Set Data Product',
+                        title: isMultipleDataProductsEnabled ? 'Add to Data Products' : 'Set Data Product',
                         onClick: () => {
                             setIsEditModalVisible(true);
                         },
                     },
                     {
-                        title: 'Unset Data Product',
+                        title: isMultipleDataProductsEnabled ? 'Unset All Data Products' : 'Unset Data Product',
```
ani-malgari, line 66: Isn't this a limitation where we only allow users to unset all data products and not give them capability to unset a specific data product?

datahub-web-react/src/app/entityV2/shared/containers/profile/sidebar/DataProduct/DataProductSection.tsx
```diff
@@ -45,62 +51,105 @@ export default function DataProductSection({ readOnly }: Props) {
...
+                })
+                .catch((e: unknown) => {
+                    message.destroy();
+                    if (e instanceof Error) {
+                        message.error({
+                            content: `Failed to remove from data product. An unknown error occurred.`,
+                            duration: 3,
+                        });
+                    }
+                });
```
ani-malgari, line 95: nit: we could just use async/await here and it looks a lot cleaner without this nesting.

datahub-web-react/src/app/entityV2/shared/components/styled/search/action/DataProductsDropdown.tsx
chriscollins3456: so this is specifically from the bulk update option in search results. since you're selecting multiple assets that could have different data products on them it makes less sense to ask them to unset a specific one... i could potentially see it though.

previous behavior was just to remove the data product from all assets, so this keeps that same behavior at least. but could use a review from product i suppose

datahub-web-react/src/app/entityV2/shared/containers/profile/sidebar/DataProduct/DataProductSection.tsx
chriscollins3456: that's fair - so you're thinking because it would have less nesting then?

i may opt to leave it since .then and .catch are our typical patterns for mutation after-effects, unless i may be missing something

ani-malgari: yes, less nesting and more readable. yeah i noticed we use then-catch everywhere mostly, but we should start to adopt async/await. Functionally they're the same, but makes it more readable and debuggable.

datahub-web-react/src/app/entityV2/shared/components/styled/search/action/DataProductsDropdown.tsx
ani-malgari: yeah functionally, i think the granularity is more helpful with the option to remove all data products.

PR #16393, opened 2026-02-28
fix(ingest/snowflake): Snowflake quoting fix

treff7es:
## Summary

Two fixes for Snowflake ingestion resilience, plus a SQL parsing performance improvement:

### Fix 1: Escape embedded double-quotes in Snowflake identifiers

Table names from `information_schema.tables` can contain embedded double-quotes and dots (e.g. staging tables, or tables created with quoted identifiers). The `get_quoted_identifier_for_table` function did not escape these, producing malformed SQL like:

```sql
SELECT * FROM table("DB"."SCHEMA".""SANDBOX"."ETL"..."", 'table')
```

Fix: double any embedded `"` characters (`""`) before wrapping in outer quotes, per Snowflake's identifier escaping rules. Applied consistently to `get_quoted_identifier_for_database`, `get_quoted_identifier_for_schema`, and `get_quoted_identifier_for_table`.

### Fix 2: Wrap tag extraction in try/except

The `get_tags_on_object` call in `_process_table` and `_process_view` was not wrapped in try/except, unlike the column-fetching code directly above it. A single malformed table name causing a tag query failure.

Fix: wrap both call sites in try/except with a warning log, matching the existing pattern used for column fetching.

### Perf: Eliminate redundant deepcopy in column-level lineage

`_get_raw_col_upstreams_for_expression` was calling `.select(copy=True)`, which deep-copies the entire AST for every column expression. Instead, temporarily swap the selects list in-place (restored in a finally block), since `to_node()` only reads from the tree. Also pass `copy=False` to `sqlglot.lineage.lineage()`.

## Test Plan

- Added 8 unit tests for identifier escaping (`TestSnowflakeIdentifierQuoting`): normal names, embedded quotes, multiple quotes, dots-only, database-level, schema-level
- All 286 existing Snowflake unit tests pass
- Doctest on `split_qualified_name` passes

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema_gen.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema_gen.py
@@ -1011,12 +1011,17 @@ class SnowflakeSchemaGenerator(SnowflakeStructuredReportMixin):
             )
 
         if self.config.extract_tags != TagOption.skip:
-            table.tags = self.tag_extractor.get_tags_on_object(
-                table_name=table.name,
-                schema_name=schema_name,
-                db_name=db_name,
-                domain="table",
-            )
+            try:
+                table.tags = self.tag_extractor.get_tags_on_object(
+                    table_name=table.name,
+                    schema_name=schema_name,
+                    db_name=db_name,
+                    domain="table",
+                )
+            except Exception as e:
+                self.structured_reporter.warning(
+                    "Failed to get tags for table", table_identifier, exc=e
+                )
 
         if self.config.include_technical_schema:
             if self.config.include_primary_keys:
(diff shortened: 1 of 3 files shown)


treff7es:
> Awesome. Would the unit test fail if you revert your changes in utils/sqlglot?

no and that is expected as the SQLGLot change is to push down the memory usage

rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `8315bd6`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22571968993)

skrydal:
> > Awesome. Would the unit test fail if you revert your changes in utils/sqlglot?
> 
> no and that is expected as the SQLGLot change is to push down the memory usage

I meant the changes you introduced to improve quoting.

PR #16394, opened 2026-03-01
feat(dev-tooling): improve dev ergonomics for running datahub

shirshanka:
## Summary

This PR introduces a developer-iteration CLI tool (`scripts/datahub-dev.sh`) designed for fast, agent-driven development against a local DataHub stack, along with the server-side and build infrastructure it depends on.

### `datahub-dev` CLI (`scripts/datahub_dev.py`)

A stdlib-only Python script (no venv needed) exposing agent-friendly subcommands:

| Command | What it does |
|---|---|
| `start` | `quickstartDebug` + wait for readiness |
| `status` | JSON stack health summary |
| `wait` | Block until GMS is healthy |
| `rebuild --wait` | Module-aware incremental build + container restart |
| `flag list/get` | Read-only inspection of live feature flag values |
| `env set/restart/clean` | Flag/config changes (writes env file, restarts containers) |
| `sync-flags` | Regenerate flag classification manifest |
| `reset` / `nuke` | Recovery escalation |

All flag changes go through `env set` + `restart` — there is no hot-toggle path. Some `FeatureFlags.java` fields are consumed at bean construction time and can never be changed without a restart; others read at request time would only be transient anyway (lost on the next rebuild). The safe, uniform rule is: restart to apply.

Branch-isolated env files (`docker/datahub-dev-<branch>.env`) are injected automatically into every subprocess via `DATAHUB_LOCAL_COMMON_ENV` — no manual shell setup needed.

### Flag classification manifest (`FlagClassificationGenerator.java`)

The CLI and agents need a way to discover what flags exist and what env var name controls each one. This PR introduces a Gradle-generated manifest at `scripts/generated/flag-classification.json` (gitignored):

- **FeatureFlags.java booleans** — discovered via reflection on `FeatureFlags.class`
- **Infrastructure flags** — regex scan of `application.yaml` for `${ENV_VAR:default}` bindings
- Gradle task in `metadata-service/configuration/build.gradle` with `inputs`/`outputs` for UP-TO-DATE caching; re-runs only when `FeatureFlags.java` or `application.yaml` changes
- `datahub_dev.py` warns with the exact Gradle command when the manifest is missing (fresh clone or after adding fields to `FeatureFlags.java`)

### Server-side dev tooling (`DevToolingController`)

New read-only `/dev/featureFlags` GET endpoints for flag inspection:

- Reflects on the live `FeatureFlags` singleton to return current runtime values
- Gated by `DEV_TOOLING_ENABLED=true` (on by default in all `quickstartDebug` profiles via `docker/build.gradle`)
- `/health/detailed` on `HealthCheckController` returns bootstrap status + ES health in one JSON response for `datahub-dev status`
- Both endpoints added to `application.yaml` auth `excludedPaths`

### `scripts/` Gradle subproject

`scripts/build.gradle` registered as a proper Gradle subproject with `lint` and `lintFix` tasks running ruff in an isolated `scripts/.venv` — no dependency on `metadata-ingestion`. Hooked into the root `format` and `formatChanged` tasks.

### Structured test output

`smoke-test/tests/utilities/agent_reporter.py` — pytest plugin that emits a machine-readable JSON report to `smoke-test/build/test-report.json` when `AGENT_MODE=1`.

### Agent workflow documentation (multi-agent)

The workflow reference previously lived only in `CLAUDE.md` (Claude Code only). Extracted into a shared canonical doc and created thin per-agent config files:

- `docs/dev-guides/agent-workflow.md` — canonical reference (CLI tool, module mapping, flag lifecycle, recovery escalation, structured test output)
- `AGENTS.md` — root-level, self-contained; picked up automatically by OpenAI Codex CLI, Devin, and similar agents
- `.cursor/rules/datahub-dev.mdc` — `alwaysApply: true` Cursor rule with full content inline
- `CLAUDE.md` — agent section trimmed to a 5-step summary + pointer to the canonical doc

## Review feedback addressed (latest commit)

- **`datahub_dev.py`**: removed stale `flag set` entry from module docstring; added path traversal guard in `cmd_test` to reject paths escaping `SMOKE_TEST_DIR`; replaced bare `subprocess.run` with `_run()` in `cmd_test` for consistent timeout and env injection; added newline rejection in `cmd_env_set` to prevent malformed env files
- **`FlagClassificationGenerator`**: null-check `outputPath.getParent()` before `Files.createDirectories`; nested `BufferedReader` inside its own try-with-resources block; added `StandardCharsets.UTF_8` for explicit encoding
- **`DevToolingController`**: removed unused `@Slf4j`; unknown flag now returns 404 instead of 400; test updated to match

## Test plan

- [x] `./gradlew :scripts:lintFix` — lints `datahub_dev.py` cleanly
- [x] `./gradlew :metadata-service:configuration:generateFlagClassification` — produces `scripts/generated/flag-classification.json` with all `FeatureFlags.java` booleans; UP-TO-DATE on second run
- [x] `scripts/datahub-dev.sh start` — starts stack and waits for readiness
- [x] `scripts/datahub-dev.sh flag list` — returns live flag values from `/dev/featureFlags`
- [x] `scripts/datahub-dev.sh env set SHOW_BROWSE_V2=true && env restart` — flag change confirmed applied after restart
- [x] `scripts/datahub-dev.sh rebuild --wait` — incremental rebuild detects changed modules via `git diff`
- [x] `./gradlew :metadata-service:configuration:test :metadata-service:openapi-servlet:test` — all tests pass
- [x] `docs/dev-guides/agent-workflow.md` renders correctly — tables and code blocks intact
- [x] `AGENTS.md` is self-contained — reading it alone gives a complete workflow picture
- [x] `.cursor/rules/datahub-dev.mdc` exists with `alwaysApply: true`

### DIFF
--- a/docker/profiles/docker-compose.gms.yml
+++ b/docker/profiles/docker-compose.gms.yml
@@ -495,6 +495,7 @@ services:
       <<: *datahub-gms-env
       MAE_CONSUMER_ENABLED: false
       MCE_CONSUMER_ENABLED: false
+      DEV_TOOLING_ENABLED: ${DEV_TOOLING_ENABLED:-false}
     depends_on:
       system-update-quickstart:
         condition: service_completed_successfully


shirshanka:
@cursoragent review

docker/build.gradle
```diff
@@ -338,6 +338,7 @@ dockerCompose {
             environment.put 'DATAHUB_APP_VERSION', System.getenv("DATAHUB_APP_VERSION") ?: "v${version}"
             environment.put 'DATAHUB_TELEMETRY_ENABLED', 'false'
             environment.put "METADATA_TESTS_ENABLED", "true"
+            environment.put "DEV_TOOLING_ENABLED", "true"
```
devashish2203: THis will add the variable to all quickStart profiles. Not just Debug.

docker/build.gradle
```diff
@@ -725,6 +726,7 @@ tasks.register("reloadEnv", Exec) {
         environment.put 'DATAHUB_APP_VERSION', System.getenv("DATAHUB_APP_VERSION") ?: "v${version}"
         environment.put 'DATAHUB_TELEMETRY_ENABLED', 'false'
         environment.put "METADATA_TESTS_ENABLED", "true"
+        environment.put "DEV_TOOLING_ENABLED", "true"
```
devashish2203: Same as above.

scripts/dev/datahub-dev.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+# Thin wrapper around datahub_dev.py so agents and humans can run:
+#   scripts/datahub-dev.sh status
+#   scripts/datahub-dev.sh rebuild --wait
+# etc.
+exec python3 "$(dirname "$0")/datahub_dev.py" "$@"
```
devashish2203: nit: We can consider updating mise.toml with the dev tooling tasks as well to directly call the python scripts. It allows having autocomplete and descriptions for tasks. I already added mise.toml to manage dependencies for those who might prefer that.
https://mise.jdx.dev/tasks/

scripts/dev/datahub_dev.py
```diff
@@ -0,0 +1,1038 @@
...
+GMS_URL = os.environ.get("DATAHUB_GMS_URL", "http://localhost:8080")
+FRONTEND_URL = os.environ.get("DATAHUB_FRONTEND_URL", "http://localhost:9002")
+COMPOSE_PROJECT = os.environ.get("DOCKER_COMPOSE_PROJECT_NAME", "datahub")
+
+DEFAULT_TIMEOUT = 300  # seconds
+POLL_INTERVAL = 3  # seconds
+
+# Module path -> (gradle task, container service name)
+MODULE_TO_CONTAINER = {
+    "metadata-service/": (":metadata-service:war:bootJar", "datahub-gms"),
```
devashish2203: nit: We should standardize the task names in gradle for the build tasks to make maintenance easier.
Not required in this PR.

smoke-test/tests/utilities/agent_reporter.py
```diff
@@ -0,0 +1,149 @@
...
+def pytest_addoption(parser):
+    parser.addoption(
+        "--agent-report",
+        action="store",
+        default=None,
+        help="Path for agent-friendly JSON test report",
+    )
+
+
+class AgentReporter:
```
devashish2203, line 35: Instead of handrolling something for this we should use the Common Test Reporting Standard to get JSON reports in a industry wide known standard.
https://github.com/infopulse/pytest-common-test-report-json
Installing and configuring the plugin via pytest options should give us similar output.

AGENTS.md
```diff
@@ -0,0 +1,75 @@
+# DataHub — Agent Workflow Guide
+
+## `datahub-dev` CLI Tool
+
+A stdlib-only Python CLI for agent-driven development. No venv needed — runs with system `python3`.
```
devashish2203, line 5: Can we point to docs/dev-guides/agent-workflow.md in this file too instead of duplicating the content?

.cursor/rules/datahub-dev.mdc
```diff
@@ -0,0 +1,81 @@
...
+description: DataHub dev workflow — use datahub-dev.sh for all build/test/flag operations
+globs:
+alwaysApply: true
+---
+
+# DataHub Agent Workflow
+
+## `datahub-dev` CLI Tool
+
+A stdlib-only Python CLI for agent-driven development. No venv needed — runs with system `python3`.
```
devashish2203, line 11: Can we point to docs/dev-guides/agent-workflow.md instead of duplicating content.

scripts/build.gradle
```diff
@@ -0,0 +1,26 @@
...
+  inputs.file('datahub_dev.py')
+  commandLine 'bash', '-c',
+    "${venv_ruff} check datahub_dev.py && ${venv_ruff} format --check datahub_dev.py"
+}
+
+task lintFix(type: Exec, dependsOn: installRuff) {
+  description = 'Auto-fix formatting of datahub_dev.py'
+  inputs.file('datahub_dev.py')
+  commandLine 'bash', '-c',
+    "${venv_ruff} check --fix datahub_dev.py && ${venv_ruff} format datahub_dev.py"
```
devashish2203: nit: Maybe use *.py to allow splitting datahub_dev.py into multiple modules later.

scripts/dev/build.gradle
devashish2203, line 1: Please move this under scripts/dev-only to be clear the purpose of the scripts. 
Top level scripts folder can then be used to add more scripts in the future.

shirshanka: Fixed — moved to `scripts/dev/` (cleaner than `dev-only/`).

docker/build.gradle
shirshanka: Fixed — scoped to `if (config.isDebug)` so non-debug profiles like `quickstart` and `quickstartCLI` no longer get `DEV_TOOLING_ENABLED`.

docker/build.gradle
shirshanka: Fixed — same `if (config.isDebug)` guard applied here too.

scripts/build.gradle
shirshanka: Fixed — switched to `inputs.files(fileTree('.').include('*.py'))` and `*.py` glob in both lint and lintFix tasks.

scripts/dev/datahub-dev.sh
shirshanka: Good call — will add mise task definitions in a follow-up PR.

scripts/dev/datahub_dev.py
shirshanka: Acknowledged — out of scope for this PR.

smoke-test/tests/utilities/agent_reporter.py
shirshanka: The custom reporter provides DataHub-specific value that a generic CTRF library doesn't: it pattern-matches GMS error output to emit actionable hints like "check GMS logs" and "is the DataHub stack running? Try `datahub-dev status`". A standard CTRF reporter gives pass/fail counts but loses those agent-specific hints. Happy to adopt CTRF as the output *format* in a follow-up PR while keeping the DataHub error-categorisation layer on top.

AGENTS.md
shirshanka: Intentional — `AGENTS.md` needs to be self-contained because agents scanning the repo root don't follow relative links; they treat the file they find as their complete context. The canonical doc is already linked at the bottom for human readers. The duplication is deliberate.

.cursor/rules/datahub-dev.mdc
shirshanka: Same constraint as `AGENTS.md`, plus a Cursor-specific one: `alwaysApply: true` injects only the `.mdc` file's own content — Cursor doesn't follow links when injecting context. Inline content is the only way to make the rule actually work.

scripts/dev/datahub_dev.py
anshbansal, line 1: Need to add `pyproject.toml` to add ruff, pyright and complexity checks for this folder. Otherwise easy to get a giant piece of code that is hard for anyone other than original author to make changes in.

As we use this to enable more agents to work with the project these static checks for code quality and consistency will become important

scripts/dev/datahub-dev.sh
anshbansal: Strongly suggest using `uv python` commands here so we can force a specific version of python

scripts/dev/datahub_dev.py
shirshanka: Actually, pushing back on this a bit. The whole point of this script is that it's stdlib-only with no deps and no venv required — that constraint is itself the complexity guardrail. If someone needs a third-party library they simply can't import it, which naturally keeps the code lean. Adding a `pyproject.toml` + pyright setup would introduce tooling overhead that contradicts the 'just run it with python3' philosophy. Ruff already runs on this file today via the `metadata-ingestion:lintFix` task (`ruff check --fix scripts/`), so code style is already enforced.

shirshanka: Correction: ruff runs via `:scripts:dev:lintFix`, not `metadata-ingestion:lintFix` — there's a dedicated `build.gradle` in `scripts/dev/` that installs ruff into a local venv and runs it over the `*.py` files in that directory.

scripts/dev/datahub-dev.sh
shirshanka: Done — switched to `uv run --python 3.11 --no-project`. The `--no-project` flag prevents uv from walking up the tree and discovering a `pyproject.toml`, which matters here since this is a stdlib-only script with no deps.

PR #16396, opened 2026-03-02
fix(ingestion/oracle): fix profiling crashes and silent table exclusions

acrylJonny:
## Summary

Fixes three latent bugs that crash or silently break Oracle ingestion when profiling is enabled (all integration test configs have profiling disabled, so none of these were caught). Also fixes a class of URN construction bugs that cause view lineage and usage lineage to produce non-matching entity references when connecting via `service_name` in Oracle Multitenant (PDB) environments.

### Bug 1 — Invalid SQL for sample value queries

`AND ROWNUM <= N` was appended to a bare `SELECT * FROM table` with no `WHERE` clause, producing invalid SQL and crashing all column sample value collection for Oracle.

### Bug 2 — `AttributeError: self.report` in DBA mode

`OracleInspectorObjectWrapper` called `self.report.failure()` / `self.report.warning()` in `get_db_name`, `get_pk_constraint`, and `get_foreign_keys`, but the class had no `report` attribute. Fixed by adding `report: SQLSourceReport` to the constructor and passing `self.report` at the single instantiation site. Only affects `data_dictionary_mode: DBA`.

### Bug 3 — Profiling silently excludes all tables when limits are `null`

In SQL, `NULL < NULL` evaluates to `NULL` not `TRUE`, so setting `profile_table_row_limit` or `profile_table_size_limit` to `null` caused the entire `WHERE` clause to evaluate to `NULL`, silently excluding every table from profiling. Fixed with `IS NULL` short-circuit checks.

### View lineage and usage URN construction for `service_name` connections

When using `service_name` (Oracle Multitenant / PDB) with `add_database_name_to_urn: true`, several issues caused entity URNs, view lineage URNs, and V$SQL usage lineage URNs to diverge:

- **`DB_NAME_QUERY` now prefers `CON_NAME` over `DB_NAME`**: In multitenant Oracle, `sys_context('USERENV','DB_NAME')` returns the CDB name (often `NONE`), while `sys_context('USERENV','CON_NAME')` returns the PDB name. The query now uses `CON_NAME` unless connected to the CDB root, eliminating the `none.edw.table` URN problem in most setups without any config changes.

- **New `urn_db_name` config field**: An explicit override for the rare cases where auto-detection still returns the wrong value (e.g. service name does not route directly to the target PDB).

- **`get_identifier` now uses `urn_db_name`**: Entity URNs now include the DB prefix from `urn_db_name` when `add_database_name_to_urn: true`, so all three URN construction paths (entities, view lineage, V$SQL lineage) produce the same result.

- **Casing normalisation across all paths**: `urn_db_name` values that are ALL_UPPERCASE are lowercased (matching Oracle's `normalize_name` convention) so entity URNs and lineage URNs share the same DB component regardless of whether `convert_urns_to_lowercase` is set.

- **`get_db_schema` `"None"` string bug**: The fallback for 2-part identifiers was calling `str(self.config.database)` which produced the literal string `"None"` when `database` was unset, causing the SQL parser to construct `None.EDW.TABLE` URNs for view definitions.

### `eager_graph_load` for restricted ingestion runs

Oracle now enables `eager_graph_load` when not ingesting both tables and views (e.g. when running with a restricted `table_pattern`). This pre-fetches all known schemas from DataHub before processing view definitions and V$SQL queries, so column-level lineage can resolve references to tables outside the current scope. Matches the pattern used by Snowflake.

### Ingestion stage reporting

Oracle now reports `METADATA_EXTRACTION`, `QUERIES_EXTRACTION`, and `LINEAGE_EXTRACTION` using the shared constants from `ingestion_stage.py`, consistent with Snowflake and Redshift.

## Testing

- Added regression test for the `WHERE ROWNUM` fix (no integration coverage since profiling is disabled everywhere in integration configs)
- Added test that `report.failure()` is called in the DBA mode error path
- Added tests for `urn_db_name` in `get_db_name`, `get_identifier`, and `get_db_schema`
- Added tests for the `get_db_schema` `"None"` regression (returns `None` not `"None"`)
- Added tests for `eager_graph_load` being `True`/`False` based on ingestion config
- Updated `DB_NAME_QUERY` test to assert presence of both `CON_NAME` and `DB_NAME`
- Fixed broken `OracleInspectorObjectWrapper` test setup (missing `report` constructor arg)
- Fixed mypy errors in integration tests (`tuple[str]` message, `.impact` attribute, missing constructor arg)


rajatoss:
## Connector Tests Results

**Connector tests failed** for commit `1f2cc00`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22678704273)

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
```diff
@@ -560,20 +584,29 @@ def get_sql_alchemy_url(
...
         regular = f"{schema}.{table}"
         if self.add_database_name_to_urn:
-            if self.database:
-                return f"{self.database}.{regular}"
-            return regular
-        else:
-            return regular
+            db = self.database
+            if not db and self.urn_db_name:
+                # Replicate Oracle's normalize_name: ALL_UPPERCASE → lowercase.
```
kyungsoo-datahub: Would be nice to have common function to nomalize urn_db_name for this function and get_db_schema function.

metadata-ingestion/tests/unit/test_oracle_source.py
```diff
@@ -1164,6 +1200,176 @@ def test_get_procedure_dependencies_returns_upstream_tables(
         assert dependencies.upstream_tables[1].type == OracleObjectType.VIEW
 
 
+def test_oracle_sample_query_uses_where_rownum():
+    from datahub.ingestion.source.sql.sqlalchemy_data_reader import (
```
kyungsoo-datahub: Move all imports on the top of the file.

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
```diff
@@ -1247,23 +1280,23 @@ def __init__(self, config, ctx):
...
+        self.aggregator = SqlParsingAggregator(
+            platform=self.platform,
+            platform_instance=self.config.platform_instance,
+            env=self.config.env,
+            graph=self.ctx.graph,
+            generate_lineage=self.include_lineage,
+            generate_usage_statistics=self.config.include_usage_stats,
+            generate_operations=self.config.include_operational_stats,
+            usage_config=self.config if self.config.include_usage_stats else None,
+            eager_graph_load=not (
```
kyungsoo-datahub: Users with restricted table_pattern now get eager_graph_load=true automatically. On large DataHub instances, this could cause memory/performance problems. Consider making this configurable.

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
```diff
@@ -2093,31 +2115,25 @@ def _populate_aggregator_from_queries(self) -> None:
...
-            logger.info("Completed schema metadata extraction")
+        logger.info("Schema metadata extraction complete")
 
-        # Step 2: Query extraction after schema extraction
-        # This allows lineage processing to have access to all discovered schema information
-        self._populate_aggregator_from_queries()
+        # V$SQL second: observed queries are added to the aggregator after the
+        # schema resolver is populated. _generate_aggregator_workunits is a no-op
+        # in the parent call above (overridden below) so lineage is not emitted yet.
+        with self.report.new_stage(QUERIES_EXTRACTION):
```
kyungsoo-datahub: Remove `self.report.new_stage(QUERIES_EXTRACTION): `. The inner  _populate_aggregator_from_queries() already has the same stage.

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
```diff
@@ -350,7 +354,16 @@ class ProcedureDependencies(BaseModel):
 )
 
 DB_NAME_QUERY = """
```
kyungsoo-datahub, line 367: Multitenant Oracle users with service_name will get new URNs (PDB name instead of CDB). Add entry to docs/how/updating-datahub.md with migration path and mention the urn_db_name workaround.

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
```diff
@@ -409,6 +422,17 @@ class OracleConfig(BasicSQLAlchemyConfig, BaseUsageConfig):
         default=None,
         description="If using, omit `service_name`.",
     )
+    urn_db_name: Optional[str] = Field(
```
kyungsoo-datahub, line 436: Clarify when urn_db_name should be used. Currently if both database and urn_db_name are set, entity and lineage URNs diverge.

metadata-ingestion/tests/unit/test_oracle_source.py
acrylJonny: done

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
acrylJonny: Done — extracted `_normalize_db_name()` at module level. Both `get_identifier` and `get_db_schema` now call it instead of duplicating the inline expression.

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
acrylJonny: Added `lazy_schema_resolver: bool = Field(default=False)` to `OracleConfig`, mirroring Snowflake's implementation exactly. When set, it suppresses the automatic eager load even when `table_pattern` is restricted.

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
acrylJonny: Done — `get_workunits_internal` now calls `_populate_aggregator_from_queries()` directly. That method owns the stage itself, so there's no duplicate.

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
acrylJonny: Added an entry to `docs/how/updating-datahub.md` explaining the CDB/PDB distinction, why `service_name` connections now correctly use the PDB name, and how to use `urn_db_name` to pin the old value if needed to avoid re-creating existing entities.

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
acrylJonny: Updated the field description to make clear it only applies when `service_name` is used (i.e. `database` is not set), and added an explicit warning: "Do not set this alongside `database`; only one should be used."

kyungsoo-datahub: Could you use model validator?
If both database and urn_db_name are set, this PR takes urn_db_name and database will be silently ignored. Instead, it's good to raise error so that a customer recognizes which value would be used when both values are set.

Example: https://github.com/datahub-project/datahub/blob/755f0bf3821fecbc51ced48a22e407f184c6c823/metadata-ingestion/src/datahub/ingestion/source/snowplow/snowplow_config.py#L478-L495

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
```diff
@@ -353,6 +353,17 @@ class ProcedureDependencies(BaseModel):
 """
 )
 
+
+def _normalize_db_name(name: str) -> str:
```
kyungsoo-datahub: Please, remove "_" in front of the function name for a public function.

docs/how/updating-datahub.md
```diff
@@ -33,6 +33,7 @@ This file documents any backwards-incompatible changes in DataHub and assists pe
 
 ### Breaking Changes
 
+- #16396: Oracle connector: When connecting via `service_name` to a multitenant Oracle database with `add_database_name_to_urn: true`, the database component of URNs will now reflect the Pluggable Database (PDB) name instead of the Container Database (CDB) name. In Oracle Multitenant architecture, a CDB is the top-level container (e.g. `cdb`) and a PDB is an individual tenant database within it (e.g. `mypdb`); `service_name` typically routes to the PDB, so the PDB name is the correct identifier for your datasets. If your existing metadata was ingested with the old CDB-based URNs, re-ingesting will create new dataset entities under the corrected URNs. To preserve the old URN shape and avoid re-creating entities, set `urn_db_name` explicitly in your recipe to match your previous CDB name.
```
kyungsoo-datahub: This entry should mention that container URNs are also affected, not just dataset URNs. Container creation uses get_db_name() regardless of add_database_name_to_urn.

metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
acrylJonny: done

docs/how/updating-datahub.md
acrylJonny: Sure, that's added now

PR #16397, opened 2026-03-02
fix(ingest/iceberg): inject S3 credentials alongside Glue during cross-account role assumption

sergey-pozdnyakov-epam:
The `_custom_glue_catalog_handling()` method performs STS AssumeRole when `glue.role-arn` is configured, but only injected credentials for the Glue client. The S3 FileIO layer still used the executor's default credentials, causing cross-account Iceberg ingestion to fail when reading S3 data/metadata files.

This change:
- Injects assumed-role credentials for S3 FileIO (`s3.access-key-id`, `s3.secret-access-key`, `s3.session-token`) alongside Glue credentials
- Respects user-provided S3 credentials or a separate `s3.role-arn`
- Recognizes `s3.role-arn` as a trigger for role assumption
- Cleans up role-arn keys after assumption to prevent pyiceberg from attempting its own (broken) role assumption

Ref: https://github.com/apache/iceberg-python/issues/2747

🤖 AI Assisted with Claude Code

Co-Authored-By: ai-tools@stepstone.com

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/iceberg/iceberg_common.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/iceberg/iceberg_common.py
@@ -50,6 +50,14 @@ DEFAULT_REST_RETRY_POLICY = {"total": 3, "backoff_factor": 0.1}
 
 GLUE_ROLE_ARN = "glue.role-arn"
 
+# S3 FileIO credential property keys.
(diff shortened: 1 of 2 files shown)


skrydal:
We concluded that there is a way to achieve this functionality now (by specifying both `s3.` and `glue.` parameter `role-arn`). Closing the PR as agreed.

PR #16398, opened 2026-03-02
fix(frontend): add missing applications field to glossaryTerm GraphQL query

max-datahub:
## Summary
- Add the `applications { ...entityApplication }` field to the `getGlossaryTerm` GraphQL query in `glossaryTerm.graphql`
- This fixes the glossary term sidebar "Applications" section always rendering empty, even when applications are assigned to the term
- The frontend was already fully wired up (`SidebarApplicationSection` imported and used, `APPLICATIONS` capability declared) — only the query field was missing

## Root Cause
`glossaryTerm.graphql` never requested the `applications` field, so `entityData.applications` was always `undefined` and `SidebarApplicationSection` rendered nothing.

## Test plan
- [x] Added unit test (`glossaryApplicationsQuery.test.ts`) verifying the generated query document includes `applications` and `entityApplication`
- [x] Manual: assign an application to a glossary term, open the term detail page, confirm the application appears in the sidebar


kevinkarchacryl:
Tested locally, fix works for me, thanks.

PR #16399, opened 2026-03-02
fix(ingestion/dbt): always emit sibling relationships to fix duplicate search results for semantic views

alfiyas-datahub:
## Summary
- The dbt source now always emits sibling relationships for entities that exist
  in the target platform, regardless of the `dbt_is_primary_sibling` setting.
  Previously it only emitted them when `dbt_is_primary_sibling=False`, relying
  on the `SiblingAssociationHook` for the default case. The hook's
  `handleDbtDatasetEvent` only handles the `"source"` subtype, so semantic views
  ended up without sibling links — causing both the dbt entity and the Snowflake
  entity to appear as separate results in search.
- The hook safely skips when siblings already exist, so there is no conflict
  with the new behavior.

## Test plan
- [x] Unit tests: updated `test_dbt_sibling_aspects_creation` and added
  `test_dbt_sibling_not_created_for_ephemeral_or_test_nodes`,
  `test_dbt_sibling_created_for_semantic_views`, and
  `test_dbt_sibling_created_for_all_materializations` (10 parametrized cases)
- [x] Manual verification: ingested a dbt manifest with a semantic view into
  local quickstart and confirmed search shows a single combined entry instead
  of duplicates


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `c8a548d`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22893752267)

PR #16400, opened 2026-03-02
ingest(snowflake): Emulate tag inheritance in-memory to eliminate N+1 queries

treff7es:
## Summary

- **Eliminates N+1 tag queries** when using `extract_tags: with_lineage`. Previously, each database/schema/table object triggered a separate Snowflake query for inherited tags. Now all tags for a database are fetched in a single query and inheritance is computed client-side via `_SnowflakeTagCache`.
- **Adds `_SnowflakeTagCache`** with hierarchical inheritance: database → schema → table.
- **Deduplication logic** ensures direct tags override inherited ones when the same tag name appears at multiple levels.
- **Null safety**: Guards against null `TAG_DATABASE`, `TAG_SCHEMA`, `TAG_NAME`, `DOMAIN`, and object fields from Snowflake audit query results.
- **Error handling hardening**: Replaces `assert` with `ValueError` for production safety, adds `report.warning()` for user-visible error reporting, separates column tag extraction from column fetch.
- **Removes dead code**: `get_tags_on_object_with_propagation` query and related report fields that are no longer needed.

## Test plan

- [x] 22 unit tests in `test_snowflake_tag_inheritance.py` covering:
  - `SnowflakeTag.as_inherited()` copy semantics
  - `_SnowflakeTagCache` deduplication (direct beats inherited, first-inherited wins)
  - Schema/table inheritance from parent levels
  - Column tags are always direct-only
  - `SnowflakeTagExtractor` routing for `with_lineage`, `without_lineage`, `skip`
  - Cache-load-once-per-database behavior
  - Cache load failure graceful degradation
  - Tag pattern deny filtering on inherited tags
- [x] ruff lint + format clean
- [x] mypy clean (0 issues)
- [x] All pre-commit hooks pass

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema.py
@@ -492,9 +492,19 @@ class _SnowflakeTagCache:
         return self._deduplicate_tags(direct + schema_inherited + db_inherited)
 
     def get_column_tags_for_table_with_inheritance(
-        self, table_name: str, schema_name: str, db_name: str
+        self,
+        table_name: str,
+        schema_name: str,
+        db_name: str,
+        column_names: Optional[List[str]] = None,
     ) -> Dict[str, List[SnowflakeTag]]:
-        """Return column tags with inheritance from table, schema, and database levels."""
+        """Return column tags with inheritance from table, schema, and database levels.
+
+        Args:
+            column_names: All column names in the table. When provided,
+                inherited parent tags are applied to every column, not just
+                those with direct column tags.
+        """
         direct_column_tags = self.get_column_tags_for_table(
             table_name, schema_name, db_name
         )
@@ -518,9 +528,16 @@ class _SnowflakeTagCache:
         if not parent_tags:
             return dict(direct_column_tags)
 
+        # Apply parent tags to all known columns, merging with direct tags
+        all_columns = (
+            column_names
+            if column_names is not None
+            else list(direct_column_tags.keys())
+        )
         result: Dict[str, List[SnowflakeTag]] = {}
-        for col_name, col_tags in direct_column_tags.items():
-            result[col_name] = self._deduplicate_tags(list(col_tags) + parent_tags)
+        for col_name in all_columns:
+            col_tags = list(direct_column_tags.get(col_name, []))
+            result[col_name] = self._deduplicate_tags(col_tags + parent_tags)
         return result
 
 
--- a/metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tag.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tag.py
@@ -187,6 +187,7 @@ class SnowflakeTagExtractor(SnowflakeCommonMixin):
         table_name: str,
         schema_name: str,
         db_name: str,
+        column_names: Optional[List[str]] = None,
     ) -> Dict[str, List[SnowflakeTag]]:
         if self.config.extract_tags == TagOption.skip:
             return {}
@@ -196,7 +197,7 @@ class SnowflakeTagExtractor(SnowflakeCommonMixin):
 
         if self.config.extract_tags == TagOption.with_lineage:
             temp_column_tags = cache.get_column_tags_for_table_with_inheritance(
-                table_name, schema_name, db_name
+                table_name, schema_name, db_name, column_names=column_names
             )
         else:
             temp_column_tags = cache.get_column_tags_for_table(
--- a/metadata-ingestion/tests/unit/snowflake/test_snowflake_tag_inheritance.py
+++ b/metadata-ingestion/tests/unit/snowflake/test_snowflake_tag_inheritance.py
@@ -205,7 +205,9 @@ class TestColumnTagsWithInheritance:
 
     def test_inherits_from_all_levels(self):
         cache = _build_cache()
-        col_tags = cache.get_column_tags_for_table_with_inheritance("TBL", "SCH", "DB")
+        col_tags = cache.get_column_tags_for_table_with_inheritance(
+            "TBL", "SCH", "DB", column_names=["COL", "OTHER_COL"]
+        )
         assert "COL" in col_tags
         tag_map = {t.name: t for t in col_tags["COL"]}
         # Direct column tag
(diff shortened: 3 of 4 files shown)


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `2fa7884`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22621099832)

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_report.py
```diff
@@ -134,11 +134,6 @@ class SnowflakeV2Report(
     num_get_tables_for_schema_queries: int = 0
     num_get_views_for_schema_queries: int = 0
 
-    # these will be non-zero if the user choses to enable the extract_tags = "with_lineage" option, which requires
```
sgomezvillamor, line 137: if not these ones, may we still have some metrics about number of tags processed successfully or unsuccessfully?

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema.py
sgomezvillamor, line 387: We may add some docstring about the responsibilities of this class.

Eg the responsibilities now include: "all tags for a database are fetched in a single query and inheritance is computed client-side"

Wondering if it still make sense to keep it as "private" with the`_` prefix? I see it largely used in other files.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema.py
```diff
@@ -436,6 +447,99 @@ def get_column_tags_for_table(
...
+        self, schema_name: str, db_name: str
+    ) -> List[SnowflakeTag]:
+        direct = self.get_schema_tags(schema_name, db_name)
+        inherited = self._mark_inherited(
+            self.get_database_tags(db_name), SnowflakeObjectDomain.DATABASE
+        )
+        return self._deduplicate_tags(direct + inherited)
+
+    def get_table_tags_with_inheritance(
+        self, table_name: str, schema_name: str, db_name: str
```
sgomezvillamor, line 482: why the reverse order, wouldn't be more natural and so less error prone: db, schema, table?

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema.py
```diff
@@ -436,6 +447,99 @@ def get_column_tags_for_table(
...
+    ) -> List[SnowflakeTag]:
+        direct = self.get_table_tags(table_name, schema_name, db_name)
+        schema_inherited = self._mark_inherited(
+            self.get_schema_tags(schema_name, db_name),
+            SnowflakeObjectDomain.SCHEMA,
+        )
```
sgomezvillamor, line 488: with so many stringified args we cannot rely on types so I would use named args when calling the methods; more resilient

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema.py
```diff
@@ -436,6 +447,99 @@ def get_column_tags_for_table(
...
+        db_inherited = self._mark_inherited(
+            self.get_database_tags(db_name), SnowflakeObjectDomain.DATABASE
+        )
+        return self._deduplicate_tags(direct + schema_inherited + db_inherited)
+
+    def get_column_tags_for_table_with_inheritance(
```
sgomezvillamor, line 494: same comments than in `get_table_tags_with_inheritance` apply here

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema.py
```diff
@@ -1527,56 +1673,13 @@ def get_tags_for_database_without_propagation(
...
-        )
-
-        for tag in cur:
-            tags.append(
-                SnowflakeTag(
-                    database=tag["TAG_DATABASE"],
-                    schema=tag["TAG_SCHEMA"],
-                    name=tag["TAG_NAME"],
-                    value=tag["TAG_VALUE"],
+                self.report.warning(
```
sgomezvillamor, line 1676: `logger.warning` + `report.warning`
is that redundant? isn't `report.warning` logging too?

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tag.py
```diff
@@ -50,37 +50,83 @@ def __init__(
...
+                    self.data_dictionary.get_tags_for_database_without_propagation(
+                        db_name
+                    )
+                )
+            except Exception as e:
+                logger.warning(
+                    f"Failed to load tag cache for database {db_name}: {e}",
+                    exc_info=True,
+                )
+                self.report.warning(
```
sgomezvillamor, line 66: `logger.warning` + `report.warning` redundancy?

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tag.py
```diff
@@ -50,37 +50,83 @@ def __init__(
...
         self.tag_cache: Dict[str, _SnowflakeTagCache] = {}
 
-    def _get_tags_on_object_without_propagation(
+    def _ensure_cache_loaded(self, db_name: str) -> None:
+        if db_name not in self.tag_cache:
+            try:
+                self.tag_cache[db_name] = (
+                    self.data_dictionary.get_tags_for_database_without_propagation(
+                        db_name
+                    )
```
sgomezvillamor, line 59: I understand we are fetching here tags for a database in a single query.
Do we track directly or indirectly metrics here about number of tags retrieved, count failures of these queries, etc...?

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tag.py
```diff
@@ -50,37 +50,83 @@ def __init__(
...
-        ):  # Views belong to this domain as well.
-            assert schema_name is not None
-            assert table_name is not None
-            tags = self.tag_cache[db_name].get_table_tags(
-                table_name, schema_name, db_name
-            )
+            if schema_name is None:
+                raise ValueError(
+                    f"schema_name is required for domain {domain} (db_name={db_name})"
+                )
```
sgomezvillamor, line 94: I see many `assert` changed by `ValueError`

is there anything preventing defining the params as required instead? and so we avoid the never-ending discussion `assert` vs `ValueError` 😅 

```
        schema_name: str,
        table_name: str,
```

maybe, instead of methods like `_get_tags_on_object` we should have `_get_tags_on_database`, `_get_tags_on_schema`, `_get_tags_on_table`, ...

No strong opinion, just raising ideas.

metadata-ingestion/tests/unit/snowflake/test_snowflake_tag_inheritance.py
sgomezvillamor, line 1: I would name this with a more generic name: `test_snowflake_tags.py`

metadata-ingestion/tests/unit/snowflake/test_snowflake_tag_inheritance.py
```diff
@@ -0,0 +1,435 @@
...
+from datahub.ingestion.source.snowflake.snowflake_utils import (
+    SnowflakeIdentifierBuilder,
+)
+
+
+def _tag(db: str, schema: str, name: str, value: str) -> SnowflakeTag:
+    return SnowflakeTag(database=db, schema=schema, name=name, value=value)
+
+
+def _build_cache() -> _SnowflakeTagCache:
```
sgomezvillamor, line 24: fixtures?

metadata-ingestion/tests/unit/snowflake/test_snowflake_tag_inheritance.py
sgomezvillamor, line 1: As soon as we have good coverage here is fine.
Just wondering if the use of fixtures and parametrized tests may help.

treff7es:
> LGTM Just a bunch of very minor suggestions.
> 
> I see connector tests passing, which is great Just wondering if we include tags in the refresh data for snowflake. If not, we may take this as the opportunity to improve that part in connector-tests.

We have a connector test for tags, and this is how I tested as well. As now connector tests pass this mean we generate the same tags but with imprvoved performance

PR #16407, opened 2026-03-03
fix(ui): alchemy-components page title cleanups

jayacryl:
- Generalize PageTitle actionButton type to accept full ButtonProps
- Spread actionButton props onto Button instead of manually passing onClick/icon
- Change SimpleSelect dropdown placement from bottomRight to bottomLeft

### DIFF
--- a/datahub-web-react/src/alchemy-components/components/Select/SimpleSelect.tsx
+++ b/datahub-web-react/src/alchemy-components/components/Select/SimpleSelect.tsx
@@ -82,6 +82,7 @@ export const SimpleSelect = ({
     isLoading = false,
     dataTestId,
     visibilityDeps,
+    placement = 'bottomRight',
     ...props
 }: SelectProps) => {
     const [searchQuery, setSearchQuery] = useState('');
@@ -181,7 +182,7 @@ export const SimpleSelect = ({
                 <Dropdown
                     open={isOpen}
                     disabled={isDisabled}
-                    placement="bottomLeft"
+                    placement={placement}
                     dropdownRender={() => (
                         <DropdownContainer
                             ref={dropdownRef}
--- a/datahub-web-react/src/alchemy-components/components/Select/types.ts
+++ b/datahub-web-react/src/alchemy-components/components/Select/types.ts
@@ -61,6 +61,7 @@ export interface SelectProps<OptionType extends SelectOption = SelectOption> {
     descriptionMaxWidth?: number;
     dataTestId?: string;
     visibilityDeps?: React.DependencyList;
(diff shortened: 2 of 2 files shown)


datahub-web-react/src/alchemy-components/components/Select/SimpleSelect.tsx
```diff
@@ -181,7 +181,7 @@ export const SimpleSelect = ({
                 <Dropdown
                     open={isOpen}
                     disabled={isDisabled}
-                    placement="bottomRight"
+                    placement="bottomLeft"
```
AdrianMachado: Shouldn't this be configurable?

datahub-web-react/src/alchemy-components/components/PageTitle/types.ts
```diff
@@ -1,14 +1,12 @@
...
     title: string;
     subTitle?: string | React.ReactNode;
     pillLabel?: string;
     variant?: 'pageHeader' | 'sectionHeader';
-    actionButton?: {
-        label: string;
-        icon?: React.ReactNode;
-        onClick: () => void;
-    };
+    actionButton?: ButtonProps & { label: string };
```
AdrianMachado, line 10: Buttons support icons but not a label? Is it normally handled by children?

datahub-web-react/src/alchemy-components/components/PageTitle/PageTitle.tsx
```diff
@@ -34,8 +34,7 @@ export const PageTitle = ({
                 {subTitle ? <SubTitle variant={variant}>{subTitle}</SubTitle> : null}
             </Container>
             {actionButton ? (
-                <Button onClick={actionButton.onClick} size="md">
-                    {actionButton.icon && actionButton.icon}
+                <Button size="md" {...actionButton}>
```
AdrianMachado, line 37: Note that ordering change would mean size is overridable. I'll let you decide if that's desirable

datahub-web-react/src/alchemy-components/components/PageTitle/types.ts
jayacryl: Yeah lol

datahub-web-react/src/alchemy-components/components/PageTitle/PageTitle.tsx
jayacryl: Yeah exactly

datahub-web-react/src/alchemy-components/components/Select/SimpleSelect.tsx
jayacryl: ah good catch

PR #16408, opened 2026-03-03
perf(ingestion/mode): Parallelize report processing across all spaces

askumar27:
## Summary

Major performance overhaul of the Mode ingestion connector: cross-space parallelized report processing, multi-process SQL parsing, rate limiting, and numerous bug fixes for definition expansion, lineage extraction, and memory usage.

## Motivation

The Mode connector had several performance and correctness issues for large workspaces:

1. **Straggler problem**: Spaces were processed sequentially — each space's reports were parallelized, but threads sat idle waiting for slow reports before moving to the next space. With 80 spaces and 2,756 reports, effective parallelism was ~3.5x out of 15 configured threads (~23% utilization).
2. **Unbounded memory**: `lru_cache(maxsize=None)` on `_get_request_json` cached every API response forever, causing OOM on large workspaces.
3. **CPU-bound SQL parsing blocking I/O threads**: `sqlglot` lineage computation held the GIL, preventing concurrent API calls.
4. **No rate limiting**: Concurrent threads could trigger Mode's 429 rate limits with no backoff coordination.
5. **Broken cooperative timeout**: The `copy=False` optimization (commit `d5a1311a50`) eliminated all `deepcopy` calls, but `cooperate()` — which enforces the 10-second SQL lineage timeout — only ran inside the `__deepcopy__` wrapper. SQL lineage could run indefinitely.
6. **Definition expansion bugs**: False circular detection between sibling definitions, broken no-alias and digit-named definitions, and trailing `--` comments swallowing closing parens.

### Architecture: Cross-space parallelization

**Three-phase processing model:**

1. **Collection phase** (main thread, fast): Iterate all spaces, emit space containers, collect all `(space_token, report/dataset)` tuples into flat lists.
2. **Report processing** (global thread pool): All reports from all spaces processed in one `ThreadedIteratorExecutor` pool — no more per-space thread pools.
3. **Dataset processing** (global thread pool): Same pattern for datasets.

This eliminates the straggler problem — threads are always busy processing the next available report regardless of which space it belongs to.

### Multi-process SQL parsing

- CPU-bound `sqlglot` lineage parsing offloaded to a `ProcessPoolExecutor` (`spawn` context) to bypass the GIL
- Worker processes get their own `SchemaResolver` and `DataHubGraph` connections via `_init_sql_parse_worker`
- Auto-sized to `cpu_count - 1`, capped at `max_threads`
- Falls back to in-thread parsing if the process pool breaks or times out

### Rate limiting

- All Mode API calls go through a configurable `RateLimiter` (`api_options.requests_per_minute`, default 180)
- Mode's limit is ~240 req/min; default of 180 leaves headroom
- `RateLimiter` sleep moved outside the lock to reduce thread contention

### SQL parsing performance (sqlglot_lineage.py)

- **Scope-based table resolution** in `_get_join_side_tables`: Direct scope source walking instead of expensive `to_node()` column-lineage approach
- **Scope-aware join traversal**: `find_all_in_scope` instead of `find_all` to avoid descending into subqueries
- **Eliminated redundant deepcopy**: In-place select list swapping in `_get_raw_col_upstreams_for_expression`
- **`copy=False` in `to_node()`**: Avoids full AST deep copy during column-level lineage
- **Restored cooperative timeout**: Added explicit `cooperate()` calls at 4 hot-loop points (column loop, tree walk, `to_node()` call, scope traversal)

### Bug fixes

- **Definition expansion**: Fixed false circular detection (siblings no longer block each other), added support for nested/no-alias/digit-named definitions, added newline before closing paren to prevent trailing `--` comments from swallowing it
- **Table-level lineage on CLL timeout**: Now emits table-level lineage even when column-level lineage fails (previously returned nothing)
- **OOM from unbounded cache**: Removed `@lru_cache(maxsize=None)` on `_get_request_json`; replaced with bounded caches and manual caching
- **Data source lookup**: O(1) dict lookup by ID instead of O(n) scan, with improved error messages
- **Thread-safe report**: Added `threading.Lock` to `ModeSourceReport` for counter increments and `report_warning`/`report_failure`
- **Safe HTTP error checks**: `_is_http_404` helper checks `error.response` for None before accessing `.status_code`
- **HTTPError subclass construction**: `HTTPError429`/`HTTPError504` now pass `response=` kwarg

### Other improvements

- **Chart API call skipping**: When `explorations_count == 0`, chart API calls are skipped entirely
- **`exclude_personal_collections` config** (default `True`): Server-side filtering via Mode's `?filter=custom`
- **`items_per_page` validation**: Range expanded from `1-30` to `1-1000` (Mode API supports up to 1000)
- **Removed unused `_get_last_query_run`** method
- **Unified processing**: Merged `emit_dashboard_mces`/`emit_chart_mces`/`emit_dataset_mces` into `_process_report` and `_process_dataset`

### ThreadedIteratorExecutor improvements

- Proper exception propagation from worker threads (was silently swallowed before)
- `stop_event` for cooperative shutdown
- Drain thread to prevent deadlock when worker exceptions trigger cancellation

### Dependencies

- Added `cachetools` to Mode connector's dependency set

## Impact Assessment

| Metric | Before | After |
|--------|--------|-------|
| Wall clock (projected) | ~5h 46m | ~1h 20m |
| Effective parallelism | 3.5x | ~14.5x |
| Thread utilization | ~23% | ~97% |
| SQL lineage timeout | broken (dead code) | restored |
| Memory | unbounded (`lru_cache` on all API responses) | bounded |

**Affected Components:** Mode ingestion source, `sqlglot_lineage.py`, `ThreadedIteratorExecutor`, `RateLimiter`
**Breaking Changes:** None — `max_threads=1` (default) preserves sequential behavior. `exclude_personal_collections=True` (new default) changes the API filter from `?filter=all` to `?filter=custom`; set to `False` to restore old behavior.
**Risk Level:** Medium — substantial refactoring of processing loop, but guarded by fallback paths (single-thread mode, in-process SQL parsing) and comprehensive test coverage.

## Test plan

- [x] `./gradlew :metadata-ingestion:lintFix` — ruff + mypy pass
- [x] SQL parsing tests: 334+ passed (`tests/unit/sql_parsing/`), including new CTE chain join tables test
- [x] Mode unit tests: 36+ passed (`tests/unit/test_mode_source.py`) — definition expansion, `_is_http_404`, thread-safe report, timestamp parsing
- [x] Mode threading tests: 880 lines (`tests/integration/mode/test_mode_threading.py`) — cross-space parallelization, single-threaded fallback, dataset error isolation, chart skip optimization
- [x] ThreadedIteratorExecutor tests: exception propagation and stop-event behavior
- [x] Updated golden file for new chart URN ordering and lineage changes
- [ ] Manual validation with Mode ingestion on production-scale workspace


rajatoss:
## Connector Tests Results

**Connector tests failed** for commit `1b32a7b`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22687772447)

PR #16413, opened 2026-03-03
fix(ingest/kafka-connect): use canonical 'mssql' platform for Debezium SQL Server SchemaResolver

treff7es:
## Summary

- **Fixes Debezium SQL Server column-level lineage** when using `use_schema_resolver: true`. The `SchemaResolver` was initialized with `platform="sqlserver"` but tables ingested via `mssql-odbc` are stored as `platform="mssql"`, causing 0 schemas to be found and no column-level lineage.
- Changes `DebeziumSourceConnector.get_platform()` to return `"mssql"` (DataHub's canonical name) instead of `"sqlserver"` (Debezium's name), fixing the mismatch at its source.
- Cleans up the now-redundant `parser_platform` conversion in `get_parser()`.

Closes ING-1705

## Breaking Change

The Debezium SQL Server connector now reports its platform as `mssql` instead of `sqlserver`. If you have `platform_instance_map` or `connect_to_platform_map` entries keyed on `"sqlserver"` for Debezium SQL Server connectors, update them to use `"mssql"`.

## Test plan

- [x] Existing `test_debezium_sqlserver_with_database_and_schema` passes (already asserted `source_platform == "mssql"`)
- [x] New `test_schema_resolver_uses_mssql_for_sqlserver_connector` verifies SchemaResolver is initialized with `platform="mssql"`
- [x] All 97 kafka-connect tests pass (connector registry, schema resolver, table discovery, main)
- [x] `lintFix` passes cleanly


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `aab545b`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22633039622)

metadata-ingestion/tests/unit/test_kafka_connect_connector_registry.py
```diff
@@ -473,6 +473,44 @@ def test_schema_resolver_creation_fails_gracefully(self) -> None:
         # But schema resolver should be None
         assert connector.schema_resolver is None
 
+    def test_schema_resolver_uses_mssql_for_sqlserver_connector(self) -> None:
+        """Test that Debezium SQL Server connector creates SchemaResolver with platform='mssql'.
```
sgomezvillamor, line 477: Given the fix in `DebeziumParser.get_platform`, I was expecting a unit test for that specific method to be updated.

Instead, this adds what is effectively an integration test around the schema resolver. That’s fine, but maybe a bit excessive for this change.

I’m not saying we should remove the new integration test, but we should at least have a focused unit test covering `get_platform`.

PR #16432, opened 2026-03-04
fix(web): revert default select placement from to 'bottomLeft'

jayacryl:


jayacryl:
@cursor Noticing @datahub-web-react/src/alchemy-components/components/Select/types.ts:65 enumerates all options, can we reuse antd `Placement` type instead?

PR #16440, opened 2026-03-05
feat(pr-to-linear): notify authors when reviewer is assigned in linear

gabe-lyons:
This action runs once every 15 minutes, checks open PRs without reviewers assigned, and checks to see if the ticket in linear has a reviewer assigned - if so, it will comment to that effect.

### DIFF
--- a/.github/workflows/linear-assignment-notify.yml
+++ b/.github/workflows/linear-assignment-notify.yml
@@ -1,0 +1,147 @@
+name: Notify PR Author of Linear Assignment
+
+# Polls every 15 minutes. Uses a single paginated GraphQL query to fetch all
+# open PRs with their comments (~8 API calls total regardless of PR count),
+# then only calls Linear for PRs that have a ticket but no notification yet.
+#
+# Designed to fail silently — errors are logged but never propagate.
+# continue-on-error: true ensures this job never blocks other workflows.
+
+on:
+  schedule:
+    - cron: "*/15 * * * *"
+  workflow_dispatch:
+
+jobs:
+  notify:
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    timeout-minutes: 10
+    steps:
+      - name: Notify PR authors when Linear ticket is assigned
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/linear-assignment-notify.yml
```diff
@@ -0,0 +1,147 @@
+name: Notify PR Author of Linear Assignment
+
+# Polls every 15 minutes. Uses a single paginated GraphQL query to fetch all
```
sgomezvillamor, line 3: 15 mins might be more frequent than we need. We don’t have strict low-latency requirements, so we could consider running it hourly and reduce the cost (even if small) by ~4x.

.github/workflows/linear-assignment-notify.yml
```diff
@@ -0,0 +1,147 @@
...
+    - cron: "*/15 * * * *"
+  workflow_dispatch:
+
+jobs:
+  notify:
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    timeout-minutes: 10
+    steps:
+      - name: Notify PR authors when Linear ticket is assigned
```
sgomezvillamor, line 21: > when Linear ticket is assigned

I haven't checked the code but those Linear tickets are by default assigned to you.
Maybe the action (the graphql) should check that the assignment is == gabe so action only appiles to unassigned (non triaged yet) tickets?

.github/workflows/linear-assignment-notify.yml
gabe-lyons: i want to do 15 mins while debugging it for faster test cycles, but ill move to an hour once it stabilizes

.github/workflows/linear-assignment-notify.yml
gabe-lyons: the tickets come in by default to triage - once the ticket is moved out of triage, this action comments. while on triage, the action ignores the ticket. so there won't be gabe spam.

gabe-lyons: (hopefully!)

PR #16442, opened 2026-03-05
fix(ingestion/snowplow): add missing cachetools dependency to snowplow connector

treff7es:
## Summary
- The snowplow connector's `CacheManager` imports `cachetools` but it was not listed in the `snowplow` extras in `setup.py`
- This caused `datahub check plugins --source snowplow` to fail with: `snowplow is disabled due to a missing dependency: cachetools`
- Added `cachetools_lib` to the snowplow dependency set

## Test plan
- [x] Verified `datahub check plugins --source snowplow` returns "Source snowplow is enabled" after the fix
- [x] Linting passes

### DIFF
--- a/metadata-ingestion/setup.py
+++ b/metadata-ingestion/setup.py
@@ -431,9 +431,9 @@ slack = {
     "tenacity>=8.0.1,<9.0.0",
 }
 
-# Snowplow uses base requirements only (requests, pydantic)
-# No additional dependencies needed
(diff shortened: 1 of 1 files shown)


rajatoss:
## Connector Tests Results

**Connector tests failed** for commit `ce793aa`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22715044820)

PR #16445, opened 2026-03-05
feat(kafka-connect): bundle JVM via jdk4py to remove system Java dependency

StanDmitrievAiven:
## Summary

The `kafka-connect` ingestion source requires a JVM for Single Message Transform (SMT) parsing via JPype, but many execution environments (minimal Docker containers) don't have Java pre-installed, causing `JVMNotFoundException` at runtime.

This PR bundles a pip-installable JVM (`jdk4py`) as a dependency of the `kafka-connect` plugin and sets `JAVA_HOME` automatically at module load time, making the plugin fully self-contained.

## Changes

- **`metadata-ingestion/setup.py`**: Added `jdk4py` to the `kafka-connect` plugin's extras dependencies.
- **`metadata-ingestion/src/datahub/ingestion/source/kafka_connect/transform_plugins.py`**: Added module-level code that uses `os.environ.setdefault("JAVA_HOME", ...)` to point JPype at the `jdk4py`-bundled JVM when no system Java is available. If `JAVA_HOME` is already set (system JDK exists), it is not overridden. If `jdk4py` is not installed for some reason, the `ImportError` is silently caught and the existing behavior is preserved.

## Motivation

When running the `kafka-connect` source in minimal containers, there is no system-level JDK available. The `JPype1` library fails with:

```
jpype._jvmfinder.JVMNotFoundException: No JVM shared library file (libjvm.so) found.
```

This makes the kafka-connect connector unusable in those environments without manual intervention (installing a JDK, setting `JAVA_HOME`, etc.).

By bundling `jdk4py`, the plugin "just works" out of the box in any Python environment.

## Test Plan

- [x] Tested on a Linux host with `JAVA_HOME` unset and no system Java — `jdk4py` JVM is used automatically and kafka-connect ingestion succeeds.
- [x] Verified that when `JAVA_HOME` is already set (system JDK), the system JDK is used (no override via `setdefault`).
- [x] Verified that if `jdk4py` is not installed, the `ImportError` is caught and existing behavior is unchanged.

## Checklist

- [x] Links to any related GitHub issues or PRs

### DIFF
--- a/metadata-ingestion/setup.py
+++ b/metadata-ingestion/setup.py
@@ -648,7 +648,7 @@ plugins: Dict[str, Set[str]] = {
     "iceberg-catalog": aws_common,
     "json-schema": {"requests<3.0.0"},
     "kafka": kafka_common | kafka_protobuf,
-    "kafka-connect": sql_common | {"requests<3.0.0", "JPype1<2.0.0", "jdk4py"},
+    "kafka-connect": sql_common | {"requests<3.0.0", "JPype1<2.0.0", "jdk4py>=21.0,<22.0"},
     "ldap": {"python-ldap>=2.4,<4.0.0"},
     "looker": looker_common,
     "lookml": looker_common,


treff7es (approved):
lgtm, thanks for the contribution!

PR #16450, opened 2026-03-05
[ai-378] feat(doc-embed): automatically embed datahub-documents every hour

nwadams:
Add support for automatically embedding `datahub-documents` every hour. 
- No-op if no documents exist or semantic search is disabled
- only embeds new/changed documents so no re-embedding every hour. 

may have to set `DATAHUB_DOCUMENTS_BOOTSTRAP_VALUES` to a newer version of the cli for datahub-documents. min `v1.4.0.x` is required

Screenshot 2026-03-05 at 2 32 40 PM

### DIFF
--- a/metadata-service/configuration/src/main/resources/bootstrap_mcps/ingestion-datahub-documents.yaml
+++ b/metadata-service/configuration/src/main/resources/bootstrap_mcps/ingestion-datahub-documents.yaml
@@ -10,7 +10,7 @@
     name: '{{ingestion.name}}{{^ingestion.name}}datahub-documents{{/ingestion.name}}'
     schedule:
       timezone: '{{schedule.timezone}}{{^schedule.timezone}}UTC{{/schedule.timezone}}'
-      interval: '{{schedule.interval}}{{^schedule.interval}}0 * * * *{{/schedule.interval}}'
+      interval: '{{schedule.interval}}{{^schedule.interval}}*/15 * * * *{{/schedule.interval}}'
     config:
       version: '{{&ingestion.version}}{{^ingestion.version}}{{/ingestion.version}}'
       recipe:


metadata-service/configuration/src/main/resources/bootstrap_mcps/ingestion-datahub-documents.yaml
```diff
@@ -0,0 +1,26 @@
...
+  aspectName: dataHubIngestionSourceInfo
+  changeType: UPSERT
+  aspect:
+    type: 'datahub-documents'
+    name: '{{ingestion.name}}{{^ingestion.name}}datahub-documents{{/ingestion.name}}'
+    schedule:
+      timezone: '{{schedule.timezone}}{{^schedule.timezone}}UTC{{/schedule.timezone}}'
+      interval: '{{schedule.interval}}{{^schedule.interval}}0 * * * *{{/schedule.interval}}'
+    config:
+      version: '{{&ingestion.version}}{{^ingestion.version}}{{/ingestion.version}}'
```
jjoyce0510, line 15: Only minor thing: Need to confirm that this will have the correct veresion (v1.4.0.4+) when we release it.. I kinda forget how this version gets populated

metadata-service/configuration/src/main/resources/bootstrap_mcps.yaml
```diff
@@ -55,3 +55,10 @@ bootstrap:
       mcps_location: "bootstrap_mcps/ingestion-datahub-gc.yaml"
       values_env: "DATAHUB_GC_BOOTSTRAP_VALUES"
       revision_env: "DATAHUB_GC_BOOTSTRAP_REVISION"
+
+    - name: ingestion-datahub-documents
+      version: v1
+      optional: true
+      mcps_location: "bootstrap_mcps/ingestion-datahub-documents.yaml"
+      values_env: "DATAHUB_DOCUMENTS_BOOTSTRAP_VALUES"
```
jjoyce0510, line 63: Where does this come from? Just environment variable I guess??

alexsku:
## PR Summary

**PR**: #16450 — [ai-378] feat(doc-embed): automatically embed datahub-documents every hour
**Author**: nwadams (Nick Adams)
**Base**: master ← na--ai-378-auto-doc-embed
**Type**: config change (new bootstrap ingestion source)
**Size**: +33 / -0 across 2 files

This PR adds a new bootstrap ingestion source `datahub-documents` that runs every hour to automatically embed DataHub Document entities for semantic search. It follows the existing bootstrap MCP pattern (modeled after `ingestion-datahub-gc`). The bootstrap is marked `optional: true`, so failures won't block system startup. The source is designed to be incremental (only re-embeds changed documents) and a no-op when no documents exist.

**Key invariants:**
1. The bootstrap step must not block system startup if the source fails to register or run (requires `optional: true`)
2. The ingestion source must only embed new/changed documents (handled by the source's `incremental.enabled: True` default)
3. The YAML template must render correctly with both empty and populated env vars
4. The ingestion source must be a no-op when no documents exist or semantic search is not configured

---

## Risk Assessment

| | |
|---|---|
| **Risk** | Low |
| **Blast radius** | Ingestion scheduler — a new hourly job will appear for all deployments |
| **Rollback** | Safe — revert removes the config; existing bootstrap upgrade-result entity prevents re-application on downgrade unless version is bumped again |
| **Rollout** | Ship it — `optional: true` protects startup |
| **CI** | Passing (Aikido, CodeRabbit, Mergeable, Meticulous all green; Vercel pending but docs-only) |

---

## Blocking Issues

None found.

## High-Priority Issues

None found.

## Other Issues

- **[MEDIUM / operability]** `bootstrap_mcps.yaml:64` — `revision_env: "DATAHUB_DOCUMENTS_REVISION"` doesn't follow the naming convention of the existing GC entry (`DATAHUB_GC_BOOTSTRAP_REVISION`). The values_env correctly uses `DATAHUB_DOCUMENTS_BOOTSTRAP_VALUES` but the revision_env drops the `_BOOTSTRAP_` infix. Fix: rename to `DATAHUB_DOCUMENTS_BOOTSTRAP_REVISION` for consistency, or document the intentional deviation.

- **[MEDIUM / operability]** `ingestion-datahub-documents.yaml` — In OSS deployments without semantic search configured, this ingestion source will be created and scheduled to run hourly. The PR description claims "No-op if semantic search is disabled," but the recipe has an empty `config: { }` with no visible mechanism to detect whether semantic search/embeddings are available. If the source throws errors every hour in OSS, this creates log noise. Confidence: medium — the source code may handle this gracefully internally via `get_semantic_search_config()`, but it's worth confirming.

- **[LOW / dx]** `ingestion-datahub-documents.yaml:13` — Default cron `0 * * * *` (every hour) is more aggressive than the GC source's `0 1 * * *` (daily). This is likely intentional for near-real-time doc embedding, but worth a brief comment in the YAML or PR description explaining the frequency choice vs. resource cost.

---

## What's Missing

- **No test coverage**, though this follows an established pattern (config-only bootstrap YAML) and existing tests for `BootstrapMCPUtil` cover the rendering/loading mechanics. Acceptable for this change.
- **No documentation update** — the semantic search configuration docs (`docs/how-to/semantic-search-configuration.md`) could mention that document embedding now runs automatically if the bootstrap is applied. Minor gap.

## Test Plan

| Invariant | Covered? | Recommendation |
|---|---|---|
| Bootstrap doesn't block startup on failure | Yes — `optional: true` + existing `BootstrapMCPStep.isOptional()` tests | No new test needed |
| Template renders with empty env vars | Yes — existing `BootstrapMCPUtilTest` covers mustache rendering | No new test needed |
| Source is no-op with no documents | Not covered by this PR | Verify manually or via smoke test that empty Document set produces zero work units |
| Source is no-op when semantic search disabled | Not covered | **Recommend verifying**: run the source in an environment without embedding config and confirm it exits gracefully rather than throwing |

## Questions for Author

1. How does the `datahub-documents` source detect that semantic search is disabled and become a no-op? With an empty `config: { }`, do the `EmbeddingConfig` defaults gracefully handle a missing model endpoint, or does it require the `get_semantic_search_config()` call to short-circuit?

2. Is the `revision_env` naming (`DATAHUB_DOCUMENTS_REVISION` vs `DATAHUB_GC_BOOTSTRAP_REVISION`) intentional or an oversight?

3. The PR description mentions "may have to set `DATAHUB_DOCUMENTS_BOOTSTRAP_VALUES` to a newer version of the cli for datahub-documents" — should there be a minimum CLI version documented, or will the source simply be absent in older CLI versions and fail gracefully?

metadata-service/configuration/src/main/resources/bootstrap_mcps/ingestion-datahub-documents.yaml
nwadams: We will set the correct version in helm as needed. By default we should use v1.4.0.4+

alexsku (approved):

metadata-service/configuration/src/main/resources/bootstrap_mcps.yaml
nwadams: It is templated from
https://github.com/acryldata/datahub-helm/blob/master/charts/datahub/templates/datahub-upgrade/datahub-system-update-job.yml#L204-L224

jjoyce0510 (approved):

PR #16451, opened 2026-03-05
feat(upgrade): K8 scale-down with conditional evaluation and tests

david-leifker:
## Summary

Adds optional Java-based Kubernetes resource management to the system-update blocking upgrade. When running in Kubernetes with the feature enabled, the job can set GMS environment variables (to disable embedded consumers, etc.) and scale down MAE/MCE consumer deployments before blocking upgrades (e.g. reindex), then restore them after. Scale-down is **conditional**: it runs only when a blocking upgrade (e.g. BuildIndices when reindex is needed) requires it. GMS is not scaled down—only its env vars are updated. All cluster interaction uses the Fabric8 Kubernetes Java client. Not a breaking change: the Java implementation is disabled by default.

## Motivation

**Problem today:** During an upgrade that triggers a reindex (e.g. BuildIndices), GMS and MAE/MCE consumers can continue writing to Elasticsearch. Those writes can be lost or create inconsistency while indices are read-only, cloned, or reindexed. There is no coordinated quiescing of writers before the reindex runs.

- **Prevent lost ES writes during reindex**: By setting GMS env vars (disabling embedded consumers) and scaling down MAE/MCE before blocking steps like reindex, we avoid writes to indices that are being rebuilt and ensure a consistent upgrade.
- **Single place for upgrade logic**: Scale-down and restore live in the same codebase as other blocking upgrade steps.
- **Conditional scale-down**: Avoid scaling when no blocking upgrade needs it (e.g. when BuildIndices does not require reindex).
- **Config-driven**: Behavior via `application.yaml` and env vars; no `kubectl` or init-container scripts.
- **Graceful when not in K8**: Step is a no-op when not in Kubernetes or when scale-down / Java implementation is disabled.
- **Failure recovery**: Retry count and last-known state stored in a ConfigMap; after max retries the step restores and fails.

## Key design decisions

- **GMS**: Only env vars are updated (e.g. `MCE_CONSUMER_ENABLED=false`); GMS deployment is not scaled to zero. MAE/MCE deployments are scaled to zero when in standalone mode.
- **ConfigMap**: Holds only retry count and last-known values for restore. Actual cluster state is always read from the Kubernetes API.
- **Cleanup**: State ConfigMap is deleted only when upgrade result is `SUCCEEDED`.
- **Dual enable**: Legacy Helm flag for “scale-down on/off”; separate `useJavaImplementation` (default false) to opt in to the Java step.
- **Evaluation step**: `ScaleDownEvaluationStep` runs first and asks each blocking upgrade (except KubernetesScaleDown) whether scale-down is required; only then does `KubernetesScaleDownStep` run if required.

## Configuration

- **application.yaml**: `systemUpdate.kubernetesScaleDown.enabled`, `useJavaImplementation` (default false), `maxRetries`, `gmsEnvVarsToSet`, KEDA/rollout, etc.


metadata-service/configuration/src/main/java/com/linkedin/metadata/config/SystemUpdateConfiguration.java
```diff
@@ -14,4 +14,8 @@ public class SystemUpdateConfiguration {
 
   /** Entity consistency checking configuration */
   private EntityConsistencyConfiguration entityConsistency;
+
+  /** Kubernetes scale-down during system-update (GMS/MAE/MCE) */
+  private com.linkedin.metadata.config.kubernetes.KubernetesScaleDownConfiguration
```
RyanHolstien: nit: full path

metadata-service/configuration/src/main/java/com/linkedin/metadata/config/kubernetes/KubernetesJacksonConfig.java
```diff
@@ -0,0 +1,63 @@
...
+    mapper.setFilterProvider(filterProvider);
+
+    return mapper;
+  }
+
+  /**
+   * Returns the ObjectMapper configured for Kubernetes API responses. This mapper strips verbose
+   * metadata fields that add noise without value to API consumers.
+   */
+  public static ObjectMapper getObjectMapper() {
```
RyanHolstien: This class is designed weirdly. The mix of static and object references and then how it's passed into Spring config as well are strange.

metadata-service/openapi-servlet/src/main/java/io/datahubproject/openapi/operations/k8/KubernetesController.java
```diff
@@ -0,0 +1,826 @@
...
+            + "⚠️ WARNING: These APIs can modify cluster resources and should be used with caution "
+            + "in production environments. Requires MANAGE_SYSTEM_OPERATIONS privilege.")
+public class KubernetesController {
+
+  private final OperationContext systemOperationContext;
+  private final AuthorizerChain authorizerChain;
+  private final KubernetesClient kubernetesClient;
+  private final ObjectMapper k8sObjectMapper;
+
+  @Autowired
```
RyanHolstien, line 78: I have mixed feeling around exposing this over REST. It makes it so people can elevate themselves within an infrastructure layer via app level permissions if the app itself is given the necessary permissions to run. I think this likely should not use the app level creds and instead should operate with user level creds to the backing infra, but we'd need to manage that as well.

david-leifker: You're suggesting that the user of the API should provide k8 credentials when using the API? 

I think that is an unnecessary step since the intent of the SYSTEM_OPERATIONS privilege is designed for system operators who are responsible for running the platform. We call this out as a highly sensitive privilege for this reason. It already allows for low level operations like ES truncation/bulk deletion, kafka consumer offset management, and to expose low level raw configuration details.

I'll add a configuration option which allows disabling this controller. The API is not strictly required but included as a nice to have option for restricted environments where application control of infrastructure is the only allowed channel.

PR #16453, opened 2026-03-06
fix(ingestion/dbt): remove hard generate_docs requirement from dbt cloud auto-discovery

alfiyas-datahub:
## Summary

- dbt Cloud auto-discovery previously skipped all jobs without "Generate docs on run" enabled,
  silently dropping their metadata even though it was available in the Discovery API
- The `generate_docs` filter is now opt-in via `require_generate_docs` (default: `False`),
  so all production jobs are ingested by default
- Users who want the old behavior can set `require_generate_docs: true` in their recipe

## Test plan

- [x] Unit tests updated and passing (59 tests)
- [x] Integration tests updated and passing
- [x] Full dbt test suite passing (126 tests)
- [x] Lint clean (ruff check + format)
- [x] Coverage at 92% on dbt_cloud.py
- [x] Live verified against dbt Cloud instance — job with `generate_docs=false`
      now ingested by default, correctly skipped when `require_generate_docs: true`
- [x] Backward compatible: `require_generate_docs: true` restores old behavior

### DIFF
--- a/metadata-ingestion/tests/integration/dbt/test_dbt_cloud_autodiscovery_integration.py
+++ b/metadata-ingestion/tests/integration/dbt/test_dbt_cloud_autodiscovery_integration.py
@@ -331,7 +331,9 @@ class TestAutoDiscoveryEndToEnd:
         # Execute
         nodes, additional_metadata = source.load_nodes()
 
-        assert len(nodes) == 1
+        # mock_graphql_response fixture returns 1 model + 1 semantic model per job
+        assert len(nodes) == 2
+        assert {n.node_type for n in nodes} == {"model", "semantic_model"}
         assert additional_metadata["account_id"] == "123456"
 
         job_ids_queried = {


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `2340dd0`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/24437787821)

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_cloud.py
```diff
@@ -640,22 +620,35 @@ def _auto_discover_projects_and_jobs(self) -> List[int]:
...
             else:
                 self.report.total_jobs_processed_skipped += 1
+                skip_reasons = []
+                if not passes_pattern_check:
+                    skip_reasons.append("did not match job_id_pattern")
+                if not passes_docs_check:
+                    skip_reasons.append("generate_docs is not enabled")
+                reason_str = " and ".join(skip_reasons)
                 logger.debug(
                     f"Skipping job {job.id}: generate_docs={job.generate_docs}, "
```
kyungsoo-datahub: For this log, do we have always job.generate_docs? It can be hit when `require_generate_docs=False` as well.

PR #16457, opened 2026-03-06
feature/harmonize-the-environment-filter

mihai103:
Currently there are 2 filters named Environment, which is very confusing for users.

Both filters return the same result sets, but show different counts.

This PR tries to fix this, by merging the counts in the backend to reflect the actual returned result with that filter. For backwards compatibility the backend still returns both env and origin. If this is not needed we can simplify it further.

As the backend is still returning both, we filter one of the duplicate filters in the frontend.

Depending on your feedback of the roadmap for the 2 different environment paths, I can refactor the code accordingly.


gabe-lyons:
Thanks for the contribution! Can you post some screenshots of before/after?

mihai103:
Hello Gabe,

Before, 2 Environment filters can be seen, with different counts (but both return the same result set)
image
image

After, aggregated:
image

datahub-web-react/src/app/search/utils/filterUtils.ts
```diff
@@ -194,3 +194,13 @@ export function getAutoCompleteInputFromQuickFilter(selectedQuickFilter: QuickFi
 
     return { filters, types };
 }
+
+/**
+ * This handles the case where only "env" is present,
+ * renaming it to "origin" so it is not hidden by FILTERS_TO_REMOVE.
+ */
+export function mergeEnvIntoOriginFacets(facets: FacetMetadata[]): FacetMetadata[] {
```
chriscollins3456: so this function would still let duplicate facets exist right if there is an origin filter? what exactly is it doing? i also don't see anything about env or origin in `FILTERS_TO_REMOVE`

chriscollins3456: and if we rename the `field` wouldn't that break filtering? because then we'd try to filter on the wrong field in the backend and get the wrong results. if something is indexed as `env` in elastic, we can't pass in `origin` as the filter field name

metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/query/request/AggregationQueryBuilder.java
```diff
@@ -619,6 +620,41 @@ private Map<String, String> getFacetToDisplayNames() {
...
+        .getAggregations()
+        .forEach((value, count) -> mergedCounts.merge(value, count, Long::sum));
+
+    final FilterValueArray mergedFilterValues =
+        new FilterValueArray(SearchUtil.convertToFilters(mergedCounts, Collections.emptySet()));
+    originFacet.setAggregations(new LongMap(mergedCounts));
+    originFacet.setFilterValues(mergedFilterValues);
+    envFacet.setAggregations(new LongMap(mergedCounts));
+    envFacet.setFilterValues(
+        new FilterValueArray(SearchUtil.convertToFilters(mergedCounts, Collections.emptySet())));
```
chriscollins3456: same concept as my previous comment - wouldn't combining these filters result in inaccurate filtering from the frontend?

if we did this and combined them, if we apply a filter from the frontend it would not return the same number of results as advertised in the filter dropdown since the fields in elastic are different.

chriscollins3456 (commented):
thank you for calling this issue out and working on this fix!

i have a few questions for you.

also some ideas here for this issue:

since we have two fields with different names in elastic but both show "Environment" in the filters section, if we want to combine them in the UI then i think we will need to be smart about generating a filter where we have an OR with `env` and `origin`, which can certainly get complicated if there are other filters applied as well. we do something similar with how we handle entity subtype filters (when you click on one of the sub types underneath a parent type in the entity type filter) where this OR gets applied to all additional filters.

another option we could pursue is to have a different name other than "Environment" for one of the two filters and render that in the UI. that would be a change to `filterNameOverride` in one of the PDL model files

there may be some other options but they would likely require large reindexes or breaking changes that i would rather us not do.

datahub-web-react/src/app/search/utils/filterUtils.ts
mihai103: The idea was to try and hide the fact that we have 2 different properties for the same concept (environment) for the UI clients. 
From what I understand the ESUtils.buildFilterQuery is the entrypoint for searches. Here we have FIELDS_TO_EXPANDED_FIELDS_LIST which is used to expand filters. And both origin and env are present, so filtering by origin should be translated to origin OR env and vice-versa.
Did I got this wrong?

If we agree env and origin are implementation details, then the clean solution would also be to hide this on API level, but this would be more of a breaking change and a lot more complicated to do.

mihai103: FILTERS_TO_REMOVE contains ENV_FILTER_NAME ("env")

metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/query/request/AggregationQueryBuilder.java
mihai103: Considering we expand the env / origin filters to each other anyway, would this still an issue?

datahub-web-react/src/app/search/utils/filterUtils.ts
chriscollins3456: ahah! i forgot about `FIELDS_TO_EXPANDED_FIELDS_LIST` - this makes life so much easier and your implementation reasonable. sorry i totally forgot about that one

definitely agree we don't want the users to have to worry about this implementation detail

metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/query/request/AggregationQueryBuilder.java
chriscollins3456: nope! i just forgot about that expanding of filters on the backend side - nice work tracking that down

chriscollins3456: nit: this could also just be passed `mergedFilterValues`

metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/query/request/AggregationQueryBuilder.java
```diff
@@ -619,6 +620,41 @@ private Map<String, String> getFacetToDisplayNames() {
     return filtersToDisplayName;
   }
 
+  /**
+   * Merges "env" counts with "origin" counts since both represent the same "Environment" concept
+   * using the same FabricType enum values and return the same set of results. Both facets are kept
+   * in the response with identical merged counts;
+   */
+  private void mergeAliasedFacets(@Nonnull List<AggregationMetadata> aggregationMetadataList) {
```
chriscollins3456, line 628: would you mind adding some unit tests for this method?

datahub-web-react/src/app/search/utils/filterUtils.ts
chriscollins3456: i still am confused about this function `mergeEnvIntoOriginFacets` - the implementation just doesn't seem to match what the comment says.

the way i'm reading it is we take in some facets, if there is an `origin` facet, we don't do anything and return our facets. if there is no `origin` facet we rename the `env` facet field to `origin`

but the backend could still return both `env` and `origin` right?

i'm looking at laster `master` and not seeing `ENV_FILTER_NAME` at all, and no `env` or `origin` in `FILTERS_TO_REMOVE` here on the frontend.

chriscollins3456: also ideally we just have one of these functions defined instead of two for v1 and v2 files - we should just have the searchV2 utils file updated and just import that into the v1 file

we ultimately will be removing these v1 files soon anyways.

chriscollins3456: finally once we clear up the confusion on this function some unit tests would be swell :)

chriscollins3456 (commented):
thank you for clarifying! sorry i totally forgot about expanding fields on the backend - your approach makes sense to me!

i think there's still just a few things to clear up on the frontend, then there's some unit tests and getting CI green.

thank you so much for this!

datahub-web-react/src/app/search/utils/filterUtils.ts
mihai103: @chriscollins3456  The function handles two cases: (1) if both origin and env facets are returned by the backend, we do nothing — env is already hidden by FILTERS_TO_REMOVE so only origin is shown; (2) if only env is returned, we rename it to origin so the UI shows a consistent label. The backend never returns duplicate counts because mergeAliasedFacets in AggregationQueryBuilder merges them before they reach the frontend.

I modified the comment so it reflects what the function actually does.

mihai103: Also added units tests

metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/query/request/AggregationQueryBuilder.java
mihai103: done

metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/query/request/AggregationQueryBuilder.java
mihai103: sure, added a few tests

datahub-web-react/src/app/search/utils/filterUtils.ts
mihai103: I just removed the function from V1 as in latest it would no longer be used anyway

datahub-web-react/src/app/search/SearchPage.tsx
chriscollins3456, line 1: i think this file may have since been deleted

metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/query/request/AggregationQueryBuilder.java
```diff
@@ -619,6 +620,40 @@ private Map<String, String> getFacetToDisplayNames() {
...
+    AggregationMetadata originFacet =
+        aggregationMetadataList.stream()
+            .filter(a -> "origin".equals(a.getName()))
+            .findFirst()
+            .orElse(null);
+    AggregationMetadata envFacet =
+        aggregationMetadataList.stream()
+            .filter(a -> "env".equals(a.getName()))
```
chriscollins3456, line 636: there should be constants for these two facet names somewhere otherwise we can add them

chriscollins3456 (approved):
thank you for this! i just want to make sure we handle the search page file added - whatever's going on with that - before we merge

also have some conflicts.

PR #16461, opened 2026-03-06
fix(ingestion/snowflake): map COPY query type to INSERT instead of UNKNOWN

treff7es:
## Summary
- Snowflake `COPY INTO` statements were being classified as `UNKNOWN` operation type in the `snowflake_queries.py` code path
- Changed the mapping from `QueryType.UNKNOWN` to `QueryType.INSERT`, which resolves to `OperationTypeClass.INSERT`
- This aligns with the existing mapping in `snowflake_usage_v2.py` where `COPY` is already mapped to `OperationTypeClass.INSERT` with the comment "Bulk load semantics"

## Test plan
- [ ] Verify existing Snowflake ingestion tests pass
- [ ] Confirm COPY INTO operations now show as INSERT in the operations aspect


rajatoss:
## Connector Tests Results

**Connector tests failed** for commit `3ae1558`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22765423524)

sgomezvillamor:
> LGTM, however this change impacts emitted `OperationClass` aspects in case this SQL appears. I expect an integration test should fail now because of the aspect content change. If not, it means we do not cover this case, could you add a simple test for handling snowflake's COPY INTO?

`connector-tests` failed with the golden file diff

so once merged, this requires: (i) create RC release and (ii) update golden files

and until we have better solution, this requires to merge with `connector-tests` CI check failing 😅

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_queries.py
```diff
@@ -1219,6 +1219,6 @@ def build_enriched_query_log_query(self) -> str:
     "CREATE_VIEW": QueryType.CREATE_VIEW,
     "CREATE_TABLE_AS_SELECT": QueryType.CREATE_TABLE_AS_SELECT,
     "MERGE": QueryType.MERGE,
-    "COPY": QueryType.UNKNOWN,
+    "COPY": QueryType.INSERT,
```
sgomezvillamor, line 1222: is this just a cosmetic fix or was this actually causing some issue downstream?

treff7es: It is not cosmetic; it is a requested fix as we didn't consider COPY as a write operation, which caused metrics to be off on the UI.
All the other sources we consider these as INSERT

PR #16462, opened 2026-03-06
feat(kafka_connect): add topics2TableMap support to BigQuerySinkConnector

hasb-trst:
Relevant Kafka Connect Docs: https://docs.confluent.io/kafka-connectors/bigquery/current/kafka_connect_bigquery_config.html

### DIFF
--- a/docs/how/updating-datahub.md
+++ b/docs/how/updating-datahub.md
@@ -33,6 +33,22 @@ This file documents any backwards-incompatible changes in DataHub and assists pe
 
 ### Breaking Changes
 
+### Known Issues
+
+### Potential Downtime
+
+### Deprecations
+
+### Other Notable Changes
+
+## v1.5.0
+
+Requirements:
+
+- Helm Chart: 0.9.0
+
+### Breaking Changes
+
 - #16685:(Ingestion) PowerBI M-Query lineage extraction has been rewritten using Microsoft's official `@microsoft/powerquery-parser`. As part of this change, the `native_query_parsing: false` configuration flag now suppresses only expressions containing `Value.NativeQuery`. Previously it suppressed all M-Query lineage extraction. Users who set this flag to block all lineage extraction will now see lineage produced for non-NativeQuery sources (Snowflake, PostgreSQL, MSSQL, etc.). To restore the suppress-all behaviour, add a `table_pattern.deny` rule in your recipe.
 - #16341:(Ingestion) SQL parsing: View query IDs are now generated using a SHA-256 hash instead of URL-encoding the view URN. This affects all connectors that use view lineage tracking (Snowflake, Oracle, BigQuery, Postgres, MySQL, Hive, Trino, ClickHouse, DB2, Dremio, SQL Server, and others). Previously, query entities had URNs like `urn:li:query:view_urn%3Ali%3Adataset%3A%28...%29`; they now use `urn:li:query:view_<sha256hash>`. After upgrading, the old URL-encoded query entities will become stale and orphaned. To clean them up, enable stateful ingestion with stale entity removal in your recipe and re-run ingestion.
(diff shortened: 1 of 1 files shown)


metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
```diff
@@ -356,6 +357,23 @@ def get_parser(
...
+            for mapping in mappings:
+                if ":" not in mapping:
+                    logger.warning(
+                        f"Invalid topic:table mapping format: '{mapping}'. Expected 'topic:table'."
+                    )
+                    continue
+                topic_key, table_name = mapping.split(":", 1)
+                topic_key = topic_key.strip()
+                table_name = table_name.strip()
+                topics2TableMap[topic_key] = table_name
```
sgomezvillamor: This could be wrapped up in a method, and so it could be unit tested easily

sgomezvillamor: actually similar logic exists already here
https://github.com/datahub-project/datahub/blob/146dff056cd7dadad0bf30646de83aa841ad60e3/metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py#L201-L211

and not sure if somewhere else, could you check?

so we may have a common function

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
```diff
@@ -356,6 +357,23 @@ def get_parser(
     ) -> BQParser:
         project: str = connector_manifest.config["project"]
         sanitizeTopics: str = connector_manifest.config.get("sanitizeTopics") or "false"
+        topic2table_map_str: Optional[str] = connector_manifest.config.get(
+            "topic2TableMap"
```
sgomezvillamor: I see the config is documented as `topic2Table.map`
is that transformed to camel case somewhere ? https://docs.confluent.io/cloud/current/connectors/cc-gcp-bigquery-storage-sink.html#insertion-and-ddl-support

hasb-trst: Its actually supposed to be [this page](https://docs.confluent.io/kafka-connectors/bigquery/current/kafka_connect_bigquery_config.html). Wrong link in the description, let me update

hasb-trst: I see its a legacy property, maybe we can add option to check for both `topic2TableMap` and `topic2table.map` property to add support for both versions.

sgomezvillamor: so `topic2table.map` is the new one and `topic2TableMap` the legacy one, right?

I see both have the same datatype (string) and format (`topic-1>:<table-1>,<topic-2>:<table-2>,…`) so you could do just a fallback.

```
        topic2table_map_str: Optional[str] = connector_manifest.config.get("topic2table.map") or connector_manifest.config.get("topic2TableMap")
```

please, leave those links as comments in the code

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/config_constants.py
```diff
@@ -111,3 +111,21 @@ def parse_comma_separated_list(value: str) -> List[str]:
             items.append(cleaned_item)
 
     return items
+
+
+def parse_topic_to_table_map(mappings: List[str]) -> Dict[str, str]:
```
sgomezvillamor, line 116: Let's add some unit testing for this method

metadata-ingestion/tests/unit/test_kafka_connect_config_constants.py
```diff
@@ -98,6 +99,52 @@ def test_very_long_list(self) -> None:
         assert parse_comma_separated_list(input_str) == items
 
 
+class TestParseTopicToTableMap:
```
sgomezvillamor, line 104: this is ideal case for parametrized tests

hasb-trst: makes sense

hasb-trst: Updated

PR #16464, opened 2026-03-06
chore(ingestion): urllib3 v2

sgomezvillamor:
We can relax the `urllib3` upper bound to include v2, since we’re no longer supporting Python 3.9, and `botocore` on Python 3.9 was the main blocker.

That said, installing dev dependencies — or any dependency that still pulls in Great Expectations — will downgrade `urllib3` back to v1.


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `b1907fd`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22779746259)

PR #16471, opened 2026-03-07
feat(cli): add search command with semantic search, projection, and agent context

shirshanka:
## Summary

- Adds `datahub search` CLI command with keyword and semantic search support
- Supports flexible filtering via three mutually exclusive options:
  - `--filter key=value` — simple key=value (repeatable, comma for OR)
  - `--where 'SQL expr'` — SQL-like WHERE clause (recommended for agents): `platform = snowflake AND env = PROD`, `platform IN (snowflake, bigquery)`, `glossary_term IS NOT NULL`, etc.
  - `--filters '{json}'` — full JSON DSL for complex AND/OR/NOT logic
- `--projection` flag for custom GraphQL field selection to reduce response size and token cost
- `--dry-run` flag to inspect compiled query without executing
- `--agent-context` flag and automatic non-TTY help injection so AI agents get best practices without prior knowledge of the CLI
- Deduplicates `search_queries.gql` — agent-context now loads it at runtime from `datahub.cli.gql` via `importlib.resources` instead of keeping its own 786-line copy
- Moves `search_filter_parser.py` from `datahub-agent-context` into `metadata-ingestion` to avoid circular dependency; re-exports from original location for backward compatibility
- Agent eval (eval 11) validates agent discovery and usage of search best practices

## Test plan

- [x] Unit tests: 115 CLI tests passing (`test_search_cli.py`)
- [x] Linting passes (`./gradlew :metadata-ingestion:lintFix`, `./gradlew :datahub-agent-context:lintFix`)
- [x] Live smoke tests against running DataHub: `--where` filter correctness (platform eq, AND, IN list), `--projection` payload reduction
- [x] Agent eval 11 passes at 80%+ (validated via `python scripts/dev/tests/test_agent_evals.py --eval 11`)
- [x] Manual smoke test: `datahub search --agent-context` prints skill file
- [x] Non-TTY `--help` auto-appends agent context
- [x] `--where` mutual exclusion with `--filter`/`--filters` returns exit code 2

### DIFF
--- a/datahub-agent-context/src/datahub_agent_context/mcp_tools/search.py
+++ b/datahub-agent-context/src/datahub_agent_context/mcp_tools/search.py
@@ -20,10 +20,11 @@ from datahub_agent_context.mcp_tools.search_filter_parser import (
 logger = logging.getLogger(__name__)
 
 # Load GraphQL queries from the canonical copy in datahub.cli.gql
+_gql = pkg_resources.files("datahub.cli.gql")
 search_gql = (
-    pkg_resources.files("datahub.cli.gql")
-    .joinpath("search_queries.gql")
-    .read_text(encoding="utf-8")
+    _gql.joinpath("fragments.gql").read_text(encoding="utf-8")
+    + "\n"
+    + _gql.joinpath("search.gql").read_text(encoding="utf-8")
 )
 
 


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `ec76d7c`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22912397420)

metadata-ingestion/src/datahub/cli/search_cli.py
```diff
@@ -0,0 +1,1333 @@
...
+from datahub.ingestion.graph.config import ClientMode
+from datahub.sdk.search_client import compile_filters
+from datahub.sdk.search_filters import Filter, load_filters
+from datahub.upgrade import upgrade
+
+logger = logging.getLogger(__name__)
+
+
+def parse_simple_filters(filters_list: List[str]) -> Dict[str, List[str]]:
+    """Parse --filter flags into dict of filter_name -> [values].
```
alexsku, line 63: do we want to support sql like filters, we added support for this here https://github.com/datahub-project/datahub/blob/master/datahub-agent-context/src/datahub_agent_context/mcp_tools/search_filter_parser.py

shirshanka: done -- added a `--where` which supports the SQL style filters

shirshanka:
@alexsku thanks for the thorough review! Addressing your feedback in commit 151e45dbe3:

**[HIGH] Fragile GQL operation stripping — fixed**

Replaced the `.index()`-based string surgery entirely. `search_queries.gql` is now deleted and split into three dedicated files:
- `gql/fragments.gql` — all three fragments (`PlatformFields`, `SearchEntityInfo`, `FacetEntityInfo`)
- `gql/search.gql` — only `query search(...)`
- `gql/semantic_search.gql` — only `query semanticSearch(...)`

`_build_search_query` now just reads and concatenates the right pair of files — no string parsing, no `ValueError` risk. Also added unit tests that verify `_build_search_query(semantic=False)` contains `query search` but NOT `query semanticSearch`, and vice versa.

**[HIGH] Projection injection via `fragment` keyword — fixed**

Added `re.compile(r"\bfragment\b", re.IGNORECASE)` to `_PROJECTION_BLOCKED_PATTERNS`. `fragment Foo on Bar { urn }` is now rejected with a clear error suggesting inline spreads instead. Unit test added to cover this case.

**[MEDIUM] Stale comment on `search_queries.gql`**

Moot — the file is deleted.

---

Leaving the remaining items for a follow-up:
- **[MEDIUM] `searchFlags` gap in semantic projection path** — noted, will address separately
- **[MEDIUM] `search_cli.py` size** — agree it's getting large, will split in a follow-up PR
- **[LOW] `--list-filters` / `--agent-context` as subcommands** — good call, will consider for the follow-up refactor

shirshanka:
## Refactor: `--list-filters` / `--describe-filter` → subcommands

Just pushed a follow-up commit that promotes the two discovery meta-flags off `query` into proper first-class subcommands, following the same pattern as `diagnose`:

| Old | New |
|---|---|
| `datahub search --list-filters` | `datahub search list-filters` |
| `datahub search --describe-filter platform` | `datahub search describe-filter platform` |
| `datahub search --agent-context` | `datahub search --help` (already works via `_AgentAwareGroup`) |

**Why:**
- `--list-filters` and `--describe-filter` were never search operations — they print discovery info and exit early. Subcommands are the right abstraction for that.
- `--agent-context` was redundant: `_AgentAwareGroup.format_help()` already appends `SEARCH_AGENT_CONTEXT.md` to `--help` when stdout is not a TTY (i.e. in any agent/pipe context).
- `query` drops from 18 → 15 parameters as a result.

No backward-compat cost since this PR isn't merged yet.

gabe-lyons:
Linear: ING-1855

PR #16475, opened 2026-03-08
feat(ingest/doc-embed): two-phase pending embeddings via __pending__ sentinel key

shirshanka:
## Summary

- **No metadata model changes** — reuses the existing `embeddings: map[string, EmbeddingModelData]` field with a reserved `"__pending__"` sentinel key and empty `vector=[]` arrays (valid Avro for required array fields)
- **Phase 1** (any source without an embedding provider, e.g. Notion/Confluence): `chunking_source.process_elements_inline()` emits `SemanticContent` with `embeddings={"__pending__": EmbeddingModelData(chunks=[EmbeddingChunkClass(vector=[], text=...)])}` — preserving chunk text for later embedding
- **Phase 2** (`DataHubDocumentsSource` in event mode): detects `semanticContent` MCL events where `embeddings` contains `"__pending__"`, fetches the stored chunks, generates real embeddings using the configured model, and writes back `SemanticContent` with the actual model key — naturally clearing the sentinel and breaking the re-trigger loop

### Key design decisions

- `vector=[]` (empty list) is valid Avro for a required `array[float]` field — clean sentinel with no schema changes
- `"__pending__"` is a reserved key; real model keys are always lowercase identifiers like `text_embedding_3_large` so there's no collision risk
- Phase 2 auto-loads embedding config from the DataHub server via \`AppConfig\` API — no local credentials required in the recipe

### Files changed

| File | Change |
|---|---|
| \`chunking_source.py\` | Add \`PENDING_CHUNKS_KEY\`, \`_emit_pending_chunks()\`; wire into \`process_elements_inline()\` else-branch |
| \`datahub_documents_source.py\` | Add \`_aspect_has_pending_chunks\`, \`_get_pending_chunks\`, \`_pending_chunks_to_dicts\`, \`_process_document_for_pending_chunks\`; extend \`_process_single_event\` to handle \`semanticContent\` events; fix Avro union unwrapping for \`entityUrn\`/\`aspectName\` |
| \`event_consumer.py\` | Extend filter to forward \`semanticContent\` MCL events alongside \`documentInfo\` |
| \`test_notion_docs_pipeline.py\` (new) | 4 unit tests for Phase 1/2 behavior via \`DocumentChunkingSource\` |
| \`test_process_document_for_pending_chunks.py\` (new) | 15 unit tests for \`DataHubDocumentsSource\` pending-chunk methods |
| \`smoke-test/tests/semantic/test_pending_embeddings.py\` (new) | 2 smoke tests: Phase 1 sentinel storage + Phase 2 end-to-end resolution via event mode |
| \`smoke-test/tests/semantic/test_semantic_search.py\` | Add explicit \`__pending__\` guard to \`verify_semantic_content()\` |

## Test plan

- [x] Unit tests: \`test_notion_docs_pipeline.py\` (4 tests), \`test_process_document_for_pending_chunks.py\` (15 tests), \`test_chunking_source.py\` (updated) — all pass
- [x] Smoke tests: \`test_pending_embeddings.py\` — both pass end-to-end against a running DataHub with OpenAI embeddings
  - Phase 1: \`__pending__\` sentinel stored correctly with 2 chunks and empty vectors
  - Phase 2: \`DataHubDocumentsSource\` in event mode detects MCL, generates real \`text-embedding-3-large\` embeddings (dim=3072), replaces \`__pending__\` with \`text_embedding_3_large\` key
- [x] Lint: \`ruff\` + \`mypy\` clean on both \`metadata-ingestion\` and \`smoke-test\`

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/datahub_documents/datahub_documents_source.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/datahub_documents/datahub_documents_source.py
@@ -13,7 +13,7 @@ Supports both batch (GraphQL) and event-driven (Kafka MCL) modes.
 import hashlib
 import json
 import logging
-from dataclasses import field
+from dataclasses import dataclass, field
 from datetime import datetime
 from pathlib import Path
 from typing import Any, Dict, Iterable, Optional, cast
@@ -55,6 +55,7 @@ from datahub.ingestion.source.unstructured.event_consumer import DocumentEventCo
 logger = logging.getLogger(__name__)
 
 
+@dataclass
 class DataHubDocumentsReport(StatefulIngestionReport):
     """Report for DataHub documents source."""
 
@@ -67,7 +68,7 @@ class DataHubDocumentsReport(StatefulIngestionReport):
     num_embeddings_generated: int = 0
     num_embedding_failures: int = 0
     embedding_failures: list[str] = field(default_factory=list)
-    processing_errors: list[str] = []
+    processing_errors: list[str] = field(default_factory=list)
     num_documents_limit_reached: bool = False
 
     def report_document_fetched(self) -> None:


alexsku (commented):
## PR Summary

**PR**: #16475 — feat(ingest/doc-embed): two-phase pending embeddings via \_\_pending\_\_ sentinel key
**Author**: shirshanka (Shirshanka Das)
**Base**: master ← worktree-ingest/semantic-search-pending-embeddings
**Type**: behavior change
**Size**: +1263 / -141 across 10 files

This PR introduces a two-phase embedding pipeline for document semantic search. **Phase 1**: Sources like Notion/Confluence that lack an embedding provider now emit `SemanticContent` with a `__pending__` sentinel key and empty vectors (instead of silently skipping). **Phase 2**: `DataHubDocumentsSource` in event mode detects `semanticContent` MCL events with `__pending__`, fetches the stored chunks, generates real embeddings, and writes back `SemanticContent` with the actual model key — breaking the re-trigger loop. Additionally, a startup probe is added to `DocumentChunkingSource.__init__` to verify embedding API connectivity upfront, and per-document embedding failures now fall back to `__pending__` instead of raising.

**Key invariants:**
1. `__pending__` sentinel key must never collide with real model keys (real keys are lowercase identifiers like `text_embedding_3_large`)
2. Phase 2 re-processing must NOT create an infinite loop — writing real embeddings back clears `__pending__`, so subsequent `semanticContent` MCLs are ignored
3. `vector=[]` is valid Avro for a required `array[float]` field — no schema changes needed
4. `DataHubDocumentsSource` must hard-fail at init if no embedding model is available (it exists solely to resolve pending chunks)

---

## Risk Assessment

| | |
|---|---|
| **Risk** | Medium |
| **Blast radius** | Document semantic search pipeline; sources using `DocumentChunkingSource` inline (Notion, Confluence); `DataHubDocumentsSource` event mode consumers |
| **Rollback** | Safe — no schema/model changes. Documents with `__pending__` would remain unresolved until re-deployed, but existing real embeddings are unaffected |
| **Rollout** | Ship it — behavioral change is backwards-compatible. Existing sources that had embedding configured continue to work. Sources without embedding now emit pending instead of nothing |
| **CI** | 1 failure: `testIntegrationBatch1` (likely unrelated — all semantic/ingestion tests pass). `Mergeable` check also failing |

---

## Blocking Issues

None found.

## High-Priority Issues

### [HIGH] Document limit not enforced on embedding failure fallback path
**Location**: `metadata-ingestion/src/datahub/ingestion/source/unstructured/chunking_source.py` — `process_elements_inline()`, the `except Exception` branch

When embedding generation fails per-document, the code falls back to `_emit_pending_chunks` but then falls through to the post-`if embeddings:` code which calls `report_document_processed()` but does NOT check `max_documents`. The `else` (no embedding model) branch has a proper document limit check with `RuntimeError`, but the `except` (embedding failure) branch skips it.

This means if you have `max_documents=10` and the embedding API is intermittently failing, the pipeline will process unlimited documents through the failure path without ever hitting the limit.

**Suggested fix**: Extract the document limit check into a shared helper, or add the same limit check after the `except` branch's pending chunk emission.

**Suggested test**: Add a test where `max_documents=2`, the startup probe succeeds, but per-document embedding calls fail. Verify the limit is still enforced.

---

### [HIGH] Potential redundant embedding generation when both documentInfo and semanticContent MCLs arrive
**Category**: edge-case | **Confidence**: medium
**Location**: `metadata-ingestion/src/datahub/ingestion/source/datahub_documents/datahub_documents_source.py` — `_process_single_event()`

When a Phase 1 source (e.g., Notion) ingests a document, it writes both `documentInfo` (with `unstructured_elements`) and `semanticContent` (with `__pending__`). Both MCL events are now forwarded by the event consumer. If `DataHubDocumentsSource` processes the `documentInfo` MCL first, it does full chunk+embed from scratch. Then the `semanticContent` MCL arrives — the pending check against the event payload returns True, but `_get_pending_chunks` re-fetches from the graph (which now has real embeddings) and returns None, so it gracefully skips.

However, if the `semanticContent` MCL arrives first, Phase 2 generates embeddings from pending chunks, then the `documentInfo` MCL triggers a second full chunk+embed cycle, overwriting the first result. This is redundant work (two embedding API calls) but not incorrect.

**Suggested fix**: Consider adding a check in the `documentInfo` path: if the entity already has non-pending `semanticContent`, skip re-processing. Alternatively, document this as acceptable behavior.

**How to validate**: Check MCL event ordering in a real Kafka deployment — do `documentInfo` and `semanticContent` events interleave?

---

## Other Issues

- **[MEDIUM / correctness]** `chunking_source.py:~845` — `_emit_pending_chunks` uses `datetime.utcnow()` which is deprecated since Python 3.12. Should use `datetime.now(datetime.timezone.utc)` instead. Fix: `from datetime import datetime, timezone` and `int(datetime.now(timezone.utc).timestamp() * 1000)`.

- **[MEDIUM / architecture]** `datahub_documents_source.py:~556` — `_process_document_for_pending_chunks` calls `self.chunking_source._generate_embeddings()` and `self.chunking_source._emit_semantic_content()` — both private methods. This creates tight coupling. Fix: consider making these public or adding a dedicated public method on `DocumentChunkingSource` for resolving pending chunks.

- **[LOW / operability]** `datahub_documents_source.py` — No dedicated counter for "pending chunks resolved" in the report. `num_embeddings_generated` covers both fresh and resolved-from-pending embeddings. Adding a `num_pending_resolved` counter would improve Phase 2 observability.

---

## What's Missing

- **Document limit enforcement in the embedding-failure fallback path** (see high-priority issue above).
- **No test for the redundant-processing scenario** where both `documentInfo` and `semanticContent` MCLs arrive for the same document in the same event-mode run.
- **No test for the startup probe adding latency** — the probe makes a real API call (`_generate_embeddings`) during `__init__`. In a hot-path integration test, this adds measurable latency. The existing tests mock it correctly, but there's no integration test validating probe behavior under real conditions (the smoke test covers end-to-end but not probe-specific scenarios).

## Test Plan

| Invariant | Covered? | Gap |
|---|---|---|
| `__pending__` sentinel emitted when no embedding model | ✅ — `test_phase1_no_embedding_emits_pending_sentinel` | |
| Phase 2 resolves pending to real embeddings | ✅ — `test_generates_and_emits_embeddings` | |
| Re-trigger loop is broken (non-pending events ignored) | ✅ — `test_ignores_semanticContent_event_without_pending` | |
| Startup probe failure disables embedding gracefully | ✅ — `test_embedding_failure_falls_back_to_pending` | |
| Hard-fail when DataHubDocumentsSource has no embedding model | ✅ — `TestEmbeddingHardFail` | |
| Document limit enforced in pending path | ⚠️ — `test_phase1_document_limit_respected` covers `else` branch | Missing: limit check in `except` (failure fallback) branch |
| Avro union unwrapping for entityUrn/aspectName | ✅ — covered in `_process_single_event` tests | |
| Smoke test: end-to-end Phase 1 → Phase 2 | ✅ — `test_pending_embeddings.py` (disabled by default) | |

## Questions for Author

1. Is the missing `max_documents` check in the embedding-failure fallback path intentional? The `else` branch (no model) enforces it, but the `except` branch (model present but call failed) does not.
2. For the startup probe in `DocumentChunkingSource.__init__`, is there a concern about latency or cost from making a real embedding API call on every pipeline start? Would caching or a simpler health check be preferable?
3. When both `documentInfo` and `semanticContent` MCL events arrive in the same event-mode run, is the potential for double embedding generation acceptable, or should we add deduplication logic?

shirshanka:
> **[HIGH] Document limit not enforced on embedding failure fallback path**

Thanks for the sharp eye, but this one is actually covered — the `except` branch has no early `return`, so control falls through to the shared `report_document_processed()` + limit check at lines 331–343, same as the success path. The `else` (no model) branch is the only one that needs its own copy of the check because it does `return` early.

Added `test_max_documents_limit_enforced_on_embedding_failure` (b5cf3b3) to make this explicit: sets `max_documents=2`, keeps `embedding_model` set, patches `_generate_embeddings` to raise on every call, and asserts the `RuntimeError` still fires on the 2nd document with `num_documents_limit_reached=True`. Test passes.

shirshanka:
> **[HIGH] Potential redundant embedding generation when both documentInfo and semanticContent MCLs arrive**

This one is a non-issue, but the reasoning is worth spelling out properly.

**`DataHubDocumentsSource` never processes `documentInfo` for EXTERNAL sources**

For sources like Notion and Confluence (`sourceType=EXTERNAL`), the `documentInfo` path in `_process_single_event` is filtered out at lines 401–405 — an empty `platform_filter` means "NATIVE only." So for a Notion document, only the `semanticContent` → `__pending__` path ever fires. There's no `documentInfo`-triggered chunk+embed to be redundant with.

**Kafka ordering makes the other scenario impossible anyway**

MCL events are keyed by entity URN, so both MCLs for the same document land in the same Kafka partition and are delivered in production order. Phase 1 always writes `documentInfo` before `semanticContent`, so Phase 2 always sees them in that order. The "semanticContent arrives first" scenario can't happen in a standard deployment.

**Why `__pending__` is the right design (not re-partitioning from `contents.text`)**

The source extractor is the right owner of chunking — it has domain knowledge the `DataHubDocumentsSource` doesn't. Notion uses unstructured-ingest to extract real typed elements (`Title`, `NarrativeText`, `Table`, `ListItem`, etc.) from Notion's block hierarchy. `chunk_by_title` uses those type boundaries to produce semantically coherent section-level chunks. By the time content reaches `contents.text` it's been flattened to plain text — that structural information is gone. `DataHubDocumentsSource` re-partitioning from `contents.text` would produce different, lower-quality chunks.

The `__pending__` interface is intentional: source extractor owns "understand the source and produce semantic chunks," Phase 2 owns "take those chunks and generate vectors." That separation of concerns is why the sentinel key carries the actual chunk text rather than just being a flag.

(Confluence currently synthesises a single `NarrativeText` from its full page text rather than doing real structured extraction — that's a gap in the Confluence source, not a reason to change the Phase 2 design.)

gabe-lyons:
Linear: ING-1858

rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `e7b90f7`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22936730848)

PR #16478, opened 2026-03-09
feat: Privilege constraints for structured property queries

DiogoVala:
Adds the privilegeConstraint feature to structured property policies.
Allows specifying which structured property a user can add, modify or remove from assets.

### DIFF
--- a/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/authorization/AuthorizationUtils.java
+++ b/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/authorization/AuthorizationUtils.java
@@ -167,6 +167,13 @@ public class AuthorizationUtils {
   }
 
   public static boolean canEditProperties(@Nonnull Urn targetUrn, @Nonnull QueryContext context) {
+    return canEditProperties(targetUrn, context, java.util.Collections.emptyList());
+  }
+
+  public static boolean canEditProperties(
+      @Nonnull Urn targetUrn,
+      @Nonnull QueryContext context,
+      @Nonnull Collection<Urn> structuredPropertyUrns) {
     // If you either have all entity privileges, or have the specific privileges required, you are
     // authorized.
     final DisjunctivePrivilegeGroup orPrivilegeGroups =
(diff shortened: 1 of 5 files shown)


gabe-lyons:
Linear: CAT-1504

RyanHolstien (changes requested):
This is not sufficient for additional support for the privilege constraints feature on structured properties. There are validators and other flows that need to be updated along with this.

This PR also adds no tests nor documentation. 

Please see the original PR here for reference on flows that should also be covered and tested: https://github.com/datahub-project/datahub/pull/13819

DiogoVala:
> This is not sufficient for additional support for the privilege constraints feature on structured properties. There are validators and other flows that need to be updated along with this.
> 
> This PR also adds no tests nor documentation.
> 
> Please see the original PR here for reference on flows that should also be covered and tested: #13819

Hi Ryan, thank you for the feedback. I have refactored the authorization flow to more closely match what's done for Tags.
Also added a few unit tests for basic functionality.
From the PR you referenced, I couldn't see anything else I would need to modify, but please check again if I missed something.
Thank you!

max-datahub:
Thanks @DiogoVala for starting this — I've pushed onto your branch to carry it the rest of the way (your original commits are preserved). Summary of the rework:

**Addressing the earlier review** (the privilege-constraints pattern needs enforcement across *all* write paths, plus tests and docs):

- **Cross-API enforcement:** added `StructuredPropertyPrivilegeConstraintsValidator`, a commit-layer `AspectPayloadValidator` mirroring `TagPrivilegeConstraintsValidator`, registered on the `structuredProperties` aspect. The constraint now applies to OpenAPI/REST/PATCH writes, not just the GraphQL resolvers (the resolver-only check was bypassable).
- **PATCH handling:** routes on `ChangeType.PATCH` and rebuilds the aspect via `ProposedItem`/`applyPatch`, so patch writes are enforced under alternate MCP validation.
- **Diff semantics:** a property counts as modified when it is added, removed, or its value changes.
- **DELETE:** deleting the whole aspect authorizes every currently-assigned property URN (fail-closed).
- **GraphQL pre-check:** kept your resolver checks as an early, clear error + UI parity.
- **Docs:** updated `docs/authorization/policies.md` and added a `docs/how/updating-datahub.md` note (enforcement is on by default).

**Tests:** validator unit tests (UPSERT / PATCH / DELETE, asserting the exact set of property URNs authorized) plus strengthened GraphQL tests that verify the structured-property URNs are actually enforced as policy sub-resources.

Also merged the latest `master` to resolve the staleness — the branch is now mergeable.

**Automated review:** the DELETE gap and test-coverage findings flagged by cubic are fixed. The i18n suggestion doesn't apply here — backend `AspectValidationException` messages aren't translated, consistent with `TagPrivilegeConstraintsValidator`.

cc @RyanHolstien for re-review.

> - Claude (on behalf of Max)

PR #16480, opened 2026-03-09
feat(openapi): Dynamic topics available for polling

skrydal:
This PR introduces definitions of topics allowed to be polled through `/poll` API through an environmental variable (with default being the 3 topics we have now). The explicit allow list can be defined via direct matches and/or light-weight prefixes (using `startsWith` and not full regular expressions).

Due to the above I had to also adjust policy engine. The old approach with platform-level privileges governing access to the 3 existing topics is retained, however I have created a generic, metadata-level privilege, allowing for specifying topics (or giving a wildcard) to be accessed. We should consider deprecating platform-level ones.

This is not a breaking change as, without defining the new env variable (`EVENTS_API_POLL_ALLOWED_TOPICS`) the internal logic stays the same.

Note that, once this PR is merged, when creating new policy using Get Topic Events **via the UI**, it is impossible to specify topic names (as they are not resources tracked by DataHub, so they do not appear in search). Therefor, via the UI, it is only possible to create policy matching all topics. Policy which would further filter topics is possible to be created only via custom graphQL call (or by sending a proper policy aspect). This is a limitation of this PR, but one which can be addressed with a follow-up PR, should we decide this is the approach we want.

Another point to note - having blank policy to get topic events will allow for reading from all topics but the 3 ones defined by platform-level policies (`PlatformEvent_v1`, `MetadataChangeLog_Versioned_v1`, `MetadataChangeLog_Timeseries_v1`) - to access those, other privileges are required (`GET_PLATFORM_EVENTS_PRIVILEGE`, `GET_METADATA_CHANGE_LOG_EVENTS`).

The PR also introduces new variables:
* KAFKA_TOPIC_PREFIX - to store kafka topic prefix
* KAFKA_TOPIC_PREFIX_ENABLED - to guard the logic with prefix enforcement (by default true)
Since we didn't have the concept of kafka topic prefixes before, and set directly entire topics names, we need to have a way to define the prefix which will be checked against the request. The "_enabled" flag is there to support cases where we are not using prefixes.

### DIFF
--- a/metadata-service/events-service/src/main/java/io/datahubproject/event/ExternalEventsService.java
+++ b/metadata-service/events-service/src/main/java/io/datahubproject/event/ExternalEventsService.java
@@ -347,10 +347,11 @@ public class ExternalEventsService {
       throw new UnsupportedTopicException(topicName);
     }
     // Custom topic already has prefix: use as-is
-    if (topicName.startsWith(this.topicPrefix)) {
+    String prefixWithSeparator = this.topicPrefix + "_";
+    if (topicName.startsWith(prefixWithSeparator)) {
       return topicName;
     }
     // Custom topic without prefix: prepend it
-    return this.topicPrefix + topicName;
+    return prefixWithSeparator + topicName;
   }
 }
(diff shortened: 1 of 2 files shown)


gabe-lyons:
Linear: PFP-2911

metadata-service/configuration/src/main/resources/application.yaml
```diff
@@ -548,6 +548,8 @@ elasticsearch:
 
 # TODO: Kafka topic convention
 kafka:
+  topicPrefixEnabled: ${KAFKA_TOPIC_PREFIX_ENABLED:true}
```
david-leifker, line 551: Investigating further into how topics are controlled. It seems that unlike Elasticsearch indices which have explicit prefixes, the same is not generally enforced for kafka topics. The prefix convention is used internally but not enforced in the code. This was my mistake on my original comment.

@chakru-r I would appreciate your review on this one given the plans or changes around kafka configuration management. Does this make sense?

skrydal:
Closing in favor of https://github.com/datahub-project/datahub/pull/16622/changes

PR #16483, opened 2026-03-09
feat(ingestion/kafka-connect): add support for Debezium and Confluent JDBC sink connectors

acrylJonny:
### Summary

- Registers `io.debezium.connector.jdbc.JdbcSinkConnector` and `io.confluent.connect.jdbc.JdbcSinkConnector` in the connector registry so they are handled by the existing `JdbcSinkConnector` implementation instead of being silently skipped with "Lineage for Sink Connector not supported"
- `JdbcSinkConnector` now auto-detects the target platform (`postgres`, `mysql`, `oracle`, `mssql`, etc.) from `connection.url` when the platform cannot be inferred from the connector class name alone
- Extends `_platform_alchemy_uri_tester_gen` in the SQLAlchemy URI mapper to accept a list of URL prefixes, and adds `sqlserver://` → `mssql` mapping to cover the JDBC URL scheme used by SQL Server connectors
- Updates the supported connector table in docs and adds unit + golden file tests

### Motivation

Customers running self-hosted Kafka Connect with Debezium JDBC sink connectors had no lineage extracted. All connectors of class `io.debezium.connector.jdbc.JdbcSinkConnector` were logged as unsupported and produced empty `customProperties` with no `dataJobInputOutput` edges — pipelines appeared in DataHub as dead ends. The `generic_connectors` workaround was not viable for sink connectors since it only models upstream (source) lineage.

### Test plan

- [x] Unit tests added in `test_kafka_connect_connector_registry.py`: registry routing for Debezium JDBC sink (postgres, oracle) and Confluent JDBC sink (mysql)
- [x] Parametrized unit tests added in `test_kafka_connect.py`: platform auto-detection from `connection.url` for postgres, mysql, oracle, mssql, and the no-URL fallback
- [x] Integration test golden file (`kafka_connect_mces_golden.json`) updated to reflect that `mysql_sink` (class `io.confluent.connect.jdbc.JdbcSinkConnector`) now emits lineage
- [x] All existing `test_sql_common.py` URI mapper tests continue to pass

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
@@ -6,6 +6,7 @@ from typing import Dict, Final, Iterable, List, Optional, Tuple
 from sqlalchemy.engine.url import URL, make_url
 
 from datahub.ingestion.source.kafka_connect.common import (
+    JDBC_PREFIX,
     KAFKA,
     BaseConnector,
     ConnectorManifest,
@@ -664,7 +665,7 @@ class JdbcSinkParserFactory:
             JdbcSinkParser with parsed configuration
         """
         # Parse JDBC URL using SQLAlchemy
-        jdbc_url = remove_prefix(connection_url, "jdbc:")
+        jdbc_url = remove_prefix(connection_url, JDBC_PREFIX)
         url_instance = make_url(jdbc_url)
 
         # Extract database name
@@ -854,7 +855,7 @@ class JdbcSinkConnector(BaseConnector):
                     )
                     self.platform = "unknown"
                 else:
-                    jdbc_url = remove_prefix(connection_url, "jdbc:")
+                    jdbc_url = remove_prefix(connection_url, JDBC_PREFIX)
                     self.platform = get_platform_from_sqlalchemy_uri(jdbc_url)
                     if self.platform == "external":
                         report.warning(
--- a/metadata-ingestion/src/datahub/ingestion/source/kafka_connect/common.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/kafka_connect/common.py
@@ -41,6 +41,7 @@ KAFKA: Final[str] = "kafka"
 SOURCE: Final[str] = "source"
 SINK: Final[str] = "sink"
 CONNECTOR_CLASS: Final[str] = "connector.class"
+JDBC_PREFIX: Final[str] = "jdbc:"
 
 # Default connection settings
 DEFAULT_CONNECT_URI: Final[str] = "http://localhost:8083/"
@@ -529,8 +530,7 @@ def validate_jdbc_url(url: str) -> bool:
     if not url or not isinstance(url, str):
         return False
 
-    # Basic JDBC URL validation
-    if not url.startswith("jdbc:"):
+    if not url.startswith(JDBC_PREFIX):
         return False
 
     parts = url.split(":", 3)
--- a/metadata-ingestion/src/datahub/ingestion/source/kafka_connect/source_connectors.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/kafka_connect/source_connectors.py
@@ -9,6 +9,7 @@ from datahub.ingestion.source.kafka_connect.common import (
     CLOUD_JDBC_SOURCE_CLASSES,
     CONNECTOR_CLASS,
     DEBEZIUM_CONNECTORS_WITH_2_LEVEL_CONTAINER,
+    JDBC_PREFIX,
     KAFKA,
     KNOWN_TOPIC_ROUTING_TRANSFORMS,
     MYSQL_CDC_SOURCE_V2_CLOUD,
@@ -37,9 +38,6 @@ from datahub.ingestion.source.sql.sqlalchemy_uri_mapper import (
 )
 
 logger = logging.getLogger(__name__)
-
-# Constants for JDBC URL parsing
-JDBC_PREFIX_LENGTH: Final[int] = 5  # Length of "jdbc:" prefix
 
 
 @dataclass(frozen=True)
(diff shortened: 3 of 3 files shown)


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `5b9c0ca`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22904671612)

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
```diff
@@ -816,28 +816,41 @@ def _extract_schema_from_url(
...
+        if platform:
+            self.platform = platform
+        else:
+            # Auto-detect from connection.url for self-hosted connectors
+            # (e.g. io.debezium.connector.jdbc.JdbcSinkConnector,
+            #  io.confluent.connect.jdbc.JdbcSinkConnector)
+            connection_url = manifest.config.get("connection.url", "")
+            if connection_url:
+                jdbc_url = remove_prefix(connection_url, "jdbc:")
+                self.platform = get_platform_from_sqlalchemy_uri(jdbc_url)
```
treff7es: if the parser fails to map to any known platform then it will return `external`. Should we map that as well to `unknown` or is it ok?

Maybe some check can help to identify these:
```
if self.platform == "external":
      self.report.warning(
          f"Could not detect target platform from connection.url '{connection_url}' "
          f"in connector '{manifest.name}'. JDBC scheme not in known platforms.",
          context=manifest.name,
      )
```

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
```diff
@@ -816,28 +816,41 @@ def _extract_schema_from_url(
...
+            else:
+                self.platform = "unknown"
```
treff7es, line 870: Maybe we should emit some `self.report.warning` about this to make sure user is aware we couldn't detect platform

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
```diff
@@ -816,28 +816,41 @@ def _extract_schema_from_url(
...
-        self.platform = platform
+        if platform:
+            self.platform = platform
+        else:
+            # Auto-detect from connection.url for self-hosted connectors
+            # (e.g. io.debezium.connector.jdbc.JdbcSinkConnector,
+            #  io.confluent.connect.jdbc.JdbcSinkConnector)
+            connection_url = manifest.config.get("connection.url", "")
+            if connection_url:
+                jdbc_url = remove_prefix(connection_url, "jdbc:")
```
treff7es: Should we do some validation if the connection string starts with `jdbc:` otherwise it is invalid?
here we try to strip it and if the url doesn't start with jdbc we return the string as it is.
We have method to check jdbc url -> https://github.com/datahub-project/datahub/blob/33aa413a2ad4c432b89ecac41aaad00264368e46/metadata-ingestion/src/datahub/ingestion/source/kafka_connect/common.py#L527

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
```diff
@@ -816,28 +816,41 @@ def _extract_schema_from_url(
...
-    platform: str = "postgres"  # Default platform, overridden in __init__
+    platform: str = "unknown"
 
     def __init__(
         self,
         manifest: ConnectorManifest,
         config: KafkaConnectSourceConfig,
         report: KafkaConnectSourceReport,
-        platform: str = "postgres",
+        platform: str = "",
```
treff7es: Is it ok to have it empty string?
Shouldn't it better to be Optional[str] = None

metadata-ingestion/docs/sources/kafka-connect/README.md
```diff
@@ -81,12 +81,17 @@ DataHub supports different connector types with varying levels of lineage extrac
 | **Cloud PostgreSQL Sink**<br/>`PostgresSink`                                   | ✅ Full             | ✅ Full                 | Runtime API / Config-based | Topic → Table mapping     |
 | **Cloud MySQL Sink**<br/>`MySqlSink`                                           | ✅ Full             | ✅ Full                 | Runtime API / Config-based | Topic → Table mapping     |
 | **Cloud Snowflake Sink**<br/>`SnowflakeSink`                                   | ✅ Full             | ✅ Full                 | Runtime API / Config-based | Topic → Table mapping     |
+| **Debezium JDBC Sink**<br/>`io.debezium.connector.jdbc.JdbcSinkConnector`      | ✅ Full             | ❌ Not supported        | Runtime API / Config-based | Topic → Table mapping     |
```
treff7es: Is this true these are not supported on confluent cloud?

acrylJonny: Neither `io.debezium.connector.jdbc.JdbcSinkConnector` nor `io.confluent.connect.jdbc.JdbcSinkConnector` are Confluent Cloud managed connectors, but both can appear as custom (self-managed) connectors deployed against a Confluent Cloud Kafka cluster — and the DataHub code already handles them without any Cloud-specific exclusion. I've updated the README from ❌ Not supported to ✅ Partial, and expanded the note below the table to explain the specific limitation: platform detection depends on `connection.url` being present with a recognised JDBC scheme. For true Confluent Cloud managed JDBC sinks, the dedicated `PostgresSink` / `MySqlSink` classes (which have explicit platform support) should be used instead.

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
acrylJonny: Changed to `Optional[str] = None` with the guard updated to `if platform is not None:`.

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
acrylJonny: Changed to `Optional[str] = None` with the guard updated to `if platform is not None:`.

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
acrylJonny: Done — the `else` branch now emits a `report.warning` before setting `self.platform = "unknown"`.

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
acrylJonny: Now I've added a `report.warning` when `get_platform_from_sqlalchemy_uri` returns `"external"` so users are notified the JDBC scheme wasn't recognised. The platform remains `"external"` (rather than mapping it to `"unknown"`) to preserve the original return value from the mapper.

PR #16499, opened 2026-03-09
perf(ingest): fix schema resolver bulk-fetch caching

treff7es:
## Summary

Fixes a critical performance regression where `initialize_schema_resolver_from_datahub` ran a full bulk GMS fetch (2k schemas, ~28 min each) once per connector in the same process. With multiple Kafka Connect connectors enabled, this compounded into multi-hour ingestion runs.

- **Adds `provide_schema_resolver`** — module-level `@lru_cache`'d factory that deduplicates bulk schema fetches across all callers in the same process, keyed on `(graph, platform, platform_instance, env)`
- **Extracts `SchemaResolverProvider`** — standalone class owning the `SchemaResolver` lifecycle, decoupled from `DataHubGraph`
- **Surfaces failures to ingestion report** — schema load failures now appear in `report_warning()` instead of being silently logged
- **Uses `_SchemaFetcher` Protocol** — narrow interface requiring only `_bulk_fetch_schema_info_by_filter`, breaking the circular dependency between `DataHubGraph` ↔ `SchemaResolver`

### Root cause

Each connector (BigQuery, Teradata, KafkaConnect, etc.) independently calls `initialize_schema_resolver_from_datahub()` on the graph. The `@lru_cache` on `_make_schema_resolver` only cached the *resolver object* — it did not prevent the expensive bulk fetch loop from running again.

### Fix

```
Caller → provide_schema_resolver(graph, platform, ...) → SchemaResolver
              ↑ module-level lru_cache — one fetch per (graph, platform, env) per process
```

### Architecture

**Before:** `DataHubGraph` owned both data fetching and `SchemaResolver` lifecycle. Every call to `initialize_schema_resolver_from_datahub()` triggered a new bulk GMS fetch.

**After:** `provide_schema_resolver()` is the single entry point. All callers (BigQuery, Teradata, SqlParsingAggregator) share the same cached resolver for the same platform/env combination.

### Changes by file

| File | Change |
|------|--------|
| `schema_resolver_provider.py` | **NEW** — `_SchemaFetcher` Protocol + `SchemaResolverProvider` + `provide_schema_resolver` module-level cache |
| `client.py` | Removed `initialize_schema_resolver_from_datahub()` (moved to provider) |
| `schema_resolver.py` | `DataHubGraph` import moved under `TYPE_CHECKING` to break circular import |
| `bigquery.py` | Uses `provide_schema_resolver` + `report_warning` on failure |
| `teradata.py` | Uses `provide_schema_resolver` + `report_warning` on failure |
| `sql_parsing_aggregator.py` | Uses `provide_schema_resolver` |
| `kafka_connect.py` | Creates shared provider in `__init__`, passes to registry, cleanup in `close()` |
| `connector_registry.py` | Accepts `SchemaResolverProvider` instead of `PipelineContext` |

## Test plan

- [x] Unit tests proving `provide_schema_resolver` deduplicates across provider instances (`test_schema_resolver_cache.py`)
- [x] Same-key dedup, distinct-key separation, object identity assertions
- [x] Updated connector registry tests: provider mock, no-provider, graceful failure, platform instance mapping
- [x] Updated `test_sql_queries.py` to remove deleted method mock
- [x] mypy + ruff pass on all changed files
- [x] 60 unit tests pass

### DIFF
--- a/metadata-ingestion/tests/unit/sdk/test_schema_resolver_cache.py
+++ b/metadata-ingestion/tests/unit/sdk/test_schema_resolver_cache.py
@@ -1,14 +1,28 @@
 from unittest.mock import patch
 
+import pytest
+
 from datahub.ingestion.graph.client import DatahubClientConfig, DataHubGraph
 from datahub.sql_parsing.schema_resolver import GraphQLSchemaMetadata
-from datahub.sql_parsing.schema_resolver_provider import SchemaResolverProvider
+from datahub.sql_parsing.schema_resolver_provider import (
+    SchemaResolverProvider,
+    provide_schema_resolver,
+)
 
 # A minimal GraphQL schema payload yielded by _bulk_fetch_schema_info_by_filter.
 _FAKE_URN = "urn:li:dataset:(urn:li:dataPlatform:bigquery,project.dataset.table,PROD)"
 _FAKE_SCHEMA: GraphQLSchemaMetadata = {
     "fields": [{"fieldPath": "id", "nativeDataType": "INT64"}]
 }
+
+
+@pytest.fixture(autouse=True)
+def clear_module_level_cache():
+    # provide_schema_resolver is module-level lru_cache — must be cleared between
+    # tests to prevent cross-test contamination.
+    provide_schema_resolver.cache_clear()
+    yield
+    provide_schema_resolver.cache_clear()
 
 
 @patch("datahub.emitter.rest_emitter.DataHubRestEmitter.test_connection")
@@ -66,4 +80,31 @@ def test_returned_resolver_is_same_object(mock_test_connection):
             platform="bigquery", platform_instance=None, env="PROD"
         )
 
+    assert resolver1 is resolver2
+
+
+@patch("datahub.emitter.rest_emitter.DataHubRestEmitter.test_connection")
+def test_provide_schema_resolver_deduplicates_across_instances(mock_test_connection):
+    """provide_schema_resolver must bulk-fetch only once even when called from
+    different SchemaResolverProvider instances with the same graph/platform/env.
+    This is the cross-instance deduplication guarantee that the per-instance
+    lru_cache on SchemaResolverProvider.get() cannot provide alone."""
+    mock_test_connection.return_value = {}
+    graph = DataHubGraph(DatahubClientConfig(server="http://fake-domain.local"))
+
+    with patch.object(
+        graph,
+        "_bulk_fetch_schema_info_by_filter",
+        return_value=iter([(_FAKE_URN, _FAKE_SCHEMA)]),
+    ) as mock_fetch:
+        resolver1 = provide_schema_resolver(
+            graph=graph, platform="bigquery", platform_instance=None, env="PROD"
+        )
+        # Simulate a second call site (e.g. BigQuery and sql_parsing_aggregator
+        # both resolving the same platform in the same process).
+        resolver2 = provide_schema_resolver(
+            graph=graph, platform="bigquery", platform_instance=None, env="PROD"
+        )
+
+    assert mock_fetch.call_count == 1
     assert resolver1 is resolver2


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `9135d13`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/23001195586)

metadata-ingestion/src/datahub/ingestion/graph/client.py
```diff
@@ -1565,6 +1568,13 @@ def initialize_schema_resolver_from_datahub(
...
     ) -> "SchemaResolver":
         logger.info("Initializing schema resolver")
+        cache_key = (platform, platform_instance, env)
+        if cache_key in self._initialized_schema_resolvers:
+            logger.debug(
+                f"Schema resolver for {platform}/{env} already initialized, reusing cached resolver"
+            )
+            return self._initialized_schema_resolvers[cache_key]
+
         schema_resolver = self._make_schema_resolver(
```
sgomezvillamor: Overall, I don't like the coupling between `DataHubGraph` and `SchemaResolver`
In my opinion, there should be a separated provider different from `DataHubGraph` and responsible for: instantiation and initialization (ie pre-fetching)

Anyway and regardless of my opinion, we already have a `lru_cache` in the `_make_schema_resolver`. Isn't that enough to prevent repeated calls to `_make_schema_resolver`?

```
    @functools.lru_cache
    def _make_schema_resolver(
        self,
        platform: str,
        platform_instance: Optional[str],
        env: str,
        include_graph: bool = True,
        report: Optional["SchemaResolverReport"] = None,
    ) -> "SchemaResolver":
```

gabe-lyons:
Linear: ING-1865

metadata-ingestion/src/datahub/ingestion/graph/client.py
treff7es: The lru_cache on _make_schema_resolver isn't enough because it only prevents duplicate resolver objects from being created, but it doesn't prevent _bulk_fetch_schema_info_by_filter from running multiple times. Without the cache guard, a second call to initialize_schema_resolver_from_datahub with the same args would get the same (already-populated) resolver back from lru_cache, but then run the bulk fetch again, adding
  all schemas a second time to the same resolver, which was the issue reported.

treff7es: I would defer bigger refactors as a followup if possible

treff7es: In the meantime I did the refactor

metadata-ingestion/src/datahub/sql_parsing/schema_resolver_provider.py
```diff
@@ -0,0 +1,83 @@
...
+        self._batch_size = batch_size
+        self._report = report
+
+    @functools.lru_cache
+    def get(
+        self,
+        platform: str,
+        platform_instance: Optional[str],
+        env: str,
+    ) -> SchemaResolver:
```
sgomezvillamor, line 60: Not super familiar with `functools.lru_cache`. I guess it works at object level, right?

I mean, if we have two calls:

```
new SchemaResolverProvider(...).get(platform="P", platform_instance="PI", env="E")
new SchemaResolverProvider(...).get(platform="P", platform_instance="PI", env="E")
```

this will do urn pre-fetching twice, right?

Is this a pattern we want to prevent?

Maybe we can just make provider module method to prevent that?

``` 
@functools.lru_cache
def provide_schema_resolver(
        graph: "DataHubGraph",
        batch_size: int = 100,
        platform: str,
        platform_instance: Optional[str],
        env: str,
        report: Optional[SchemaResolverReport] = None,
) -> SchemaResolver
```

metadata-ingestion/src/datahub/ingestion/graph/client.py
```diff
@@ -1556,45 +1555,6 @@ def _make_schema_resolver(
             report=report,
         )
 
-    def initialize_schema_resolver_from_datahub(
```
sgomezvillamor, line 1559: ❤️

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery.py
```diff
@@ -243,12 +244,21 @@ def _init_schema_resolver(self) -> SchemaResolver:
...
+                    return SchemaResolverProvider(
+                        graph=self.ctx.graph,
+                        batch_size=self.config.schema_resolution_batch_size,
+                    ).get(
+                        platform=self.platform,
+                        platform_instance=self.config.platform_instance,
+                        env=self.config.env,
+                    )
+                except Exception:
+                    logger.warning(
```
sgomezvillamor: beyond logging it, should we fail and keep raising exception?

metadata-ingestion/src/datahub/sql_parsing/schema_resolver_provider.py
treff7es: We can't put `@functools.lru_cache` on this method unfortunately because graph is not hashable

treff7es: anyway I solved it with a module level method

PR #16506, opened 2026-03-10
fix(ci): Update filter paths for github-actions workflow

devashish2203:
The issue is that GitHub Actions does not support brace expansion {} in path filters. The [filter pattern cheat sheet](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet) only supports *, **, ?, +, [], and !. So .github/**/*.{yml,yaml} is treated as a literal string pattern that never matches any file.

The fix is to split it into two separate path entries:


devashish2203:
Closing as the changes are being merged with https://github.com/datahub-project/datahub/pull/16502/changes to fix the CI failures.

PR #16509, opened 2026-03-10
fix(assertions): Prevent phantom assertion entities after deletion

javabrett:
## Problem

When assertion monitors with cron schedules are deleted (via `deleteAssertion` GraphQL mutation or Terraform destroy), the scheduled evaluator continues to fire on those URNs and writes run result events. This implicitly recreates the deleted assertions as **phantom entities** in the search index with:
- No `info` aspect (assertion definition)
- No dataset association
- No platform
- No monitor
- Only `runEvents` with ERROR results

These phantoms accumulate and appear in the UI under **Data Health -> Assertions** as:
- Name: "No description found"
- Type: Unknown
- Dataset: Unknown

## Root Cause

**Validated in OSS code (CAT-1500):**

1. `DeleteAssertionResolver.java` (lines 52-59): Only calls `deleteEntity()` and `deleteEntityReferences()` - **does NOT deregister from cron scheduler**
2. `AssertionService.addAssertionRunEvent()` (lines 153-179): Writes `assertionRunEvent` aspect **without validating** assertion still exists
3. `ingestProposal()` **implicitly creates entities** if they don't exist - deleted assertion reappears as empty shell

On-demand evaluation (`runAssertion` mutation) doesn't create phantoms because it validates the assertion exists first via `getEntityUrnForAssertion()`.

## Solution

**This PR implements Fix B: Defense-in-Depth Validation**

Add validation to `AssertionService.addAssertionRunEvent()` that:
1. Calls `getEntityUrnForAssertion()` before writing run events
2. Returns `null` if assertion has no "Asserts" relationship (doesn't exist)
3. If assertion doesn't exist: logs warning and returns early (does NOT write run event)
4. Prevents phantom entity recreation from **ANY caller** (cron scheduler, future schedulers, etc.)

**Note:** Fix A (scheduler deregistration in `DeleteAssertionResolver`) should also be implemented for complete cleanup.

## Changes

**Modified Files:**
- `metadata-service/services/.../AssertionService.java` (+12 lines)
  - Added validation logic in `addAssertionRunEvent()`
- `metadata-io/src/test/.../AssertionServiceTest.java` (+40 lines)
  - Added `testAddAssertionRunEventForDeletedAssertionDoesNotRecreateEntity()`

## Testing

**Unit Test Coverage:**
- New test verifies that `ingestProposal()` is **NEVER** called for deleted assertions
- Mocks `GraphClient` to return empty relationships (simulating deleted assertion)
- Prevents regression of this bug

**Manual Testing (recommended):**
1. Create assertion with cron schedule (`*/1 * * * *`)
2. Delete assertion via `deleteAssertion` mutation
3. Wait for next cron cycle (~1 minute)
4. Verify assertion does NOT reappear in search
5. Check logs for warning: "Skipping assertion run event for assertion..."

## Impact

- Prevents phantom entities from being created by ANY code path (not just cron)
- Defense-in-depth: protects against future callers of `addAssertionRunEvent()`
- Minimal performance impact: single graph query per evaluation (already cached in hot path)

## Related

- Linear: [CAT-1500](https://linear.app/acryl-data/issue/CAT-1500/fix-phantom-assertion-entities-created-by-cron-evaluator-after)
- Reproduction: See CAT-1500 for full reproduction steps and evidence
- Future Work: Implement Fix A (scheduler deregistration)

### DIFF
--- a/metadata-io/src/test/java/com/linkedin/metadata/service/AssertionServiceTest.java
+++ b/metadata-io/src/test/java/com/linkedin/metadata/service/AssertionServiceTest.java
@@ -187,8 +187,22 @@ public class AssertionServiceTest {
   @Test
   public void testAddAssertionRunEventRequiredFields() throws Exception, RemoteInvocationException {
     SystemEntityClient mockedEntityClient = mock(SystemEntityClient.class);
-    AssertionService assertionService =
-        new AssertionService(mockedEntityClient, mock(GraphClient.class));
+    GraphClient mockGraphClient = mock(GraphClient.class);
+    Mockito.when(
+            mockGraphClient.getRelatedEntities(
+                Mockito.eq(TEST_ASSERTION_URN.toString()),
+                Mockito.eq(ImmutableSet.of("Asserts")),
+                Mockito.eq(RelationshipDirection.OUTGOING),
+                Mockito.eq(0),
+                Mockito.eq(1),
+                Mockito.anyString()))
+        .thenReturn(
+            new EntityRelationships()
+                .setTotal(1)
+                .setRelationships(
+                    new EntityRelationshipArray(
+                        ImmutableList.of(new EntityRelationship().setEntity(TEST_DATASET_URN)))));
+    AssertionService assertionService = new AssertionService(mockedEntityClient, mockGraphClient);
     Long eventtime = 1718619000000L;
 
     Mockito.doAnswer(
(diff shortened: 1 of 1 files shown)


gabe-lyons:
Linear: OBS-1695

metadata-service/services/src/main/java/com/linkedin/metadata/service/AssertionService.java
```diff
@@ -156,6 +156,18 @@ public void addAssertionRunEvent(
       @Nonnull Urn asserteeUrn,
       @Nonnull Long timestampMillis,
       @Nonnull AssertionResult assertionResult) {
+    // Validate that the assertion still exists and is associated with an entity
+    // This prevents recreating deleted assertions as phantom entities when background
+    // evaluators (e.g., cron schedulers) attempt to write run events
+    final Urn validatedAsserteeUrn = getEntityUrnForAssertion(opContext, assertionUrn);
```
AdrianMachado, line 162: The only thing that calls into this code is the `ReportAssertionRunEventResolver` which already has the same guard.

https://github.com/datahub-project/datahub/blob/f9529bb32aec17578f640166cae96c155f382bd8/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/assertion/ReportAssertionResultResolver.java#L59-L67 

Additionally, this is actually the wrong codepath for what you are trying to fix. Will discuss over slack

PR #16515, opened 2026-03-10
feat(ingestion/kafka-connect): add column-level lineage for sink connectors and field-level transform support

acrylJonny:
## Summary

- **Sink column-level lineage**: Kafka Connect sink connectors (JDBC, Snowflake, BigQuery) now emit fine-grained lineage mapping Kafka topic fields to destination table columns, complementing the existing source-direction lineage (DB table → Kafka topic).
- **Field-level transform support**: `ReplaceField$Value` (include/exclude/rename), `ExtractField$Value` (struct unwrapping, e.g. Debezium `after` field), `HoistField$Value` (field nesting), and `Flatten$Value` (delimiter substitution) are all applied when building the column mapping, so SMT chains are reflected accurately in lineage.
- **Kafka field path normalisation**: Fixed a latent bug where Kafka schema field paths in v2 format (e.g. `[version=2.0].[type=Order].id`) were not being normalised to dot-notation before matching against database column names, causing no column matches in real deployments.
- **Unit tests**: New tests cover all four transform types, including edge cases (no-match extract, chained transforms, dropped fields passing through flatten) and schema-resolver-based sink lineage extraction.
- **Integration tests**: Two new golden-file tests — one for Snowflake sink column lineage and one for an end-to-end Postgres → Kafka → Snowflake pipeline covering both source and sink directions.
- **Docs**: Updated the Kafka Connect README with a Column-Level Lineage section documenting prerequisites, configuration, case-insensitive column matching, and transform behaviour.

## Test plan

- [x] Unit tests: `pytest tests/unit/kafka_connect/test_kafka_connect_replace_field_transform.py tests/unit/kafka_connect/test_kafka_connect_schema_resolver.py`
- [x] Integration tests: `pytest tests/integration/kafka-connect/test_kafka_connect.py -k "col_lineage or snowflake_sink"`


rajatoss:
## Connector Tests Results

**Connector tests failed** for commit `5e0cc98`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/23145472849)

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/common.py
```diff
@@ -983,25 +1013,126 @@ def _extract_fine_grained_lineage(
...
+        if not (
+            self.config.use_schema_resolver
+            and self.config.schema_resolver_finegrained_lineage
+            and self.schema_resolver
+            and self.schema_resolver.graph
+        ):
+            return None
+
+        try:
+            kafka_platform_instance = (
```
treff7es: Users with per-connector Kafka platform instances configured will get wrong upstream URNs in sink CLL — and the per-field target_field not found warning is at debug level, so the failure is invisible. Fix #16526 (the PR that added get_platform_instance) is effectively reverted in the sink direction.

Fix:
```
kafka_platform_instance = get_platform_instance(
      self.config, self.connector_manifest.name, KAFKA
  )
```

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/common.py
```diff
@@ -983,25 +1013,126 @@ def _extract_fine_grained_lineage(
...
         except Exception as e:
             logger.debug(
                 f"Failed to extract fine-grained lineage for "
-                f"{source_dataset} → {target_dataset}: {e}"
+                f"{source_dataset} -> {target_dataset}: {e}"
+            )
+
+        return None
+
+    def _extract_sink_fine_grained_lineage(
```
treff7es, line 1044: This swallows all errors at debug level.

The headline feature of this PR will silently produce empty CLL with no signal in the ingestion report. Schema network errors, malformed schemas, and `pydantic.ValidationErrors` all become invisible.

Fix (suggested for both source-side and sink-side catches):
```

  except Exception as e:
      self.report.warning(
          title="Failed to extract sink column-level lineage",
          message=(
              "An error occurred while resolving schemas for sink column-level "
              "lineage. Asset-level lineage is unaffected."
          ),
          context=(
              f"connector={self.connector_manifest.name}, "
              f"source_topic={source_topic}, target_platform={target_platform}, "
              f"target={target_dataset}, error_type={type(e).__name__}"
          ),
          exc=e,
      )
```

metadata-ingestion/tests/integration/kafka-connect/test_kafka_connect.py
```diff
@@ -821,225 +821,376 @@ def test_kafka_connect_snowflake_sink_ingest(
 
 
 @freeze_time(FROZEN_TIME)
```
treff7es: You should use `time-machine`

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/common.py
acrylJonny: I've updated the logs to be as per your suggestion

metadata-ingestion/tests/integration/kafka-connect/test_kafka_connect.py
acrylJonny: updated to use time-machine

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/common.py
acrylJonny: Updated as per your suggestion

treff7es (approved):

PR #16522, opened 2026-03-11
feat(ingestion): add GraphQL query projection system for schema compa…

shirshanka:
## Summary

AST-based GraphQL query projection that automatically adapts client queries to match the server's schema capabilities. When the CLI or MCP tools use newer schema features (e.g., `Document` entity added in v0.13.0), the projector strips unsupported fields/types rather than failing, enabling forward-compatible queries across server versions.

### Key capabilities

- **Fragment inlining**: Resolves named fragment spreads into inline fragments before projection, handling multi-level nesting (up to 10 levels deep)
- **Type-aware field removal**: `UnsupportedFieldRemover` AST visitor prunes inline fragments for missing types and fields not present on parent types
- **Production-grade caching**:
  - **Tier 1 (inline cache)**: `query text → inlined DocumentNode` — pure function, never invalidated
  - **Tier 2 (projection cache)**: `(query, schema_generation) → adapted result` — dict lookup, sub-microsecond
  - **On-disk cache**: `~/.datahub/schema_cache/` — avoids cold-start introspection cost across process restarts
  - Schema invalidation via commit-hash + 10-min TTL backstop
- **Required fields**: Declarative `required_fields` parameter raises `RequiredFieldUnsupportedError` if critical fields would be stripped
- **Wired into `execute_graphql`**: Activated via `strip_unsupported_fields=True` parameter — CLI search and MCP tools opt in

### Performance (benchmark results from local + CI)

**Per-stage breakdown** (N=100, median ms):

| Workload | parse | inline | visit | print_ast | cold total |
|---|---|---|---|---|---|
| CLI search (735 lines, 14 spreads) | 3.8 | 1.6 | 2.7 | 4.1 | 14.0 |
| entity_details (1735 lines, 93 spreads) | 10.1 | 43.1 | 16.0 | 13.9 | 55.3 |
| doc_search (114 lines, 2 spreads) | 0.65 | 0.26 | 0.43 | 0.62 | 3.4 |
| simple (15 lines, 0 spreads) | 0.06 | 0.00 | 0.05 | 0.07 | 1.4 |

**Cache effectiveness:**
- Tier 2 hit (dict lookup): **<0.001ms**
- Tier 1 hit + visitor re-run: **~7ms**

**Live-server results** (against local DataHub):

| Measurement | Median | P95 |
|---|---|---|
| Raw introspection query | 150ms | 472ms |
| Cold adapt_query (CLI search) | 36ms | 263ms |
| Warm adapt_query (tier 2 hit) | <0.01ms | <0.01ms |
| Disk cache hit (no introspection) | wall time drops from ~1.08s to ~0.85s |

### Files changed

| File | Lines | Purpose |
|---|---|---|
| `graphql_query_adapter.py` | 605 | QueryProjector, UnsupportedFieldRemover, fragment inlining, caching |
| `test_graphql_query_adapter.py` | 2315 | Comprehensive unit tests (real .gql files, cache tiers, edge cases) |
| `test_graphql_projection_perf.py` | 874 | Performance benchmark (mock + optional live-server mode) |
| `client.py` | +57 | Wire projector into `execute_graphql` |
| `test_query_projection_smoke.py` | 223 | Smoke tests against running DataHub instance |
| `setup.py` / `pyproject.toml` | +11 | `graphql-core>=3.2.0` dependency |
| `setup.cfg` | +2 | Register `perf` marker, add test path |

## Test plan

- [x] 2315 lines of unit tests covering inline fragments, named fragments, nested fragments, cache tiers, required_fields, edge cases
- [x] Performance benchmark validates sub-ms cache hits and quantifies per-stage costs
- [x] Live-server benchmark confirms disk caching eliminates introspection on subsequent runs
- [x] Smoke tests verify end-to-end against running DataHub
- [x] CI green (all 11 Metadata Ingestion jobs passing)

## Checklist

- [x] PR conforms to [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md)
- [x] Tests added/updated
- [ ] No breaking changes

### DIFF
--- a/metadata-ingestion/constraints.txt
+++ b/metadata-ingestion/constraints.txt
@@ -534,7 +534,9 @@ googleapis-common-protos==1.73.0
 gql==4.0.0
     # via acryl-datahub
 graphql-core==3.2.8
-    # via gql
+    # via
+    #   acryl-datahub
+    #   gql
 greenlet==3.3.2
     # via
     #   acryl-datahub


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `9f54438`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/23029290884)

metadata-ingestion/src/datahub/utilities/graphql_query_adapter.py
```diff
@@ -0,0 +1,605 @@
...
+        if server_url and commit_hash:
+            disk_schema = self._load_from_disk(server_url, commit_hash)
+            if disk_schema is not None:
+                return disk_schema
+
+        introspection_query = get_introspection_query()
+
+        try:
+            # Execute introspection query with strip_unsupported_fields=False
+            # to avoid recursion (the projector itself calls execute_graphql)
```
nwadams: can we detect potential recursion here and escape?

metadata-ingestion/src/datahub/utilities/graphql_query_adapter.py
```diff
@@ -0,0 +1,605 @@
...
+if TYPE_CHECKING:
+    from datahub.ingestion.graph.client import DataHubGraph
+
+logger = logging.getLogger(__name__)
+
+SCHEMA_TTL_SECONDS: float = 600.0  # 10 min backstop TTL
+SCHEMA_FAILURE_TTL_SECONDS: float = 60.0  # 1 min negative cache
+QUERY_CACHE_MAX_SIZE: int = 128
+INLINE_CACHE_MAX_SIZE: int = 128
+DISK_CACHE_DIR: str = os.path.join(os.path.expanduser("~"), ".datahub", "schema_cache")
```
nwadams, line 45: should we get a /tmp directory `tempfile.gettempdir()` and store there to avoid polluting?

metadata-ingestion/src/datahub/utilities/graphql_query_adapter.py
```diff
@@ -0,0 +1,605 @@
...
+                self._schema_fetched_at = time.monotonic()
+                raise
+
+    def _disk_cache_path(self, server_url: str, commit_hash: str) -> Path:
+        """Get the disk cache file path for a given server + commit hash."""
+        cache_dir = Path(DISK_CACHE_DIR)
+        key = hashlib.sha256(f"{server_url}:{commit_hash}".encode()).hexdigest()[:32]
+        return cache_dir / f"{key}.json"
+
+    def _load_from_disk(
```
nwadams, line 496: how big is the schema? Is it worth just storing in memory by default to avoid potential issues with disk permissions?

shirshanka:
Addressing @nwadams's review comments:

**Line 587 — Recursion detection**: Good call. The current guard is `strip_unsupported_fields=False` on the introspection call, which prevents `execute_graphql` from calling back into the projector. But that's implicit and fragile — if someone accidentally changes it to `True`, we'd loop. I'll add an explicit reentrant guard (`_fetching_schema` flag) that raises immediately on re-entry.

**Line 45 — Disk cache location**: `~/.datahub/` is the established DataHub config directory (same as `~/.datahubenv`, CLI config, etc.), so `schema_cache/` lives alongside related state. Using `/tmp` would lose the cache across reboots, which defeats the warm-start optimization. Happy to discuss if you feel strongly — we could also make the path configurable.

**Line 488 — Schema size / memory vs disk**: The schema is ~1.5MB JSON (introspection result). The primary cache is already in-memory (`_cached_schema`), so repeated calls within a process are instant. Disk is purely a warm-start optimization for CLI invocations — it saves ~150-200ms of introspection latency on the first call after process restart. If disk permissions are a concern, the disk cache already gracefully degrades: any read/write failure is caught and logged as a warning, falling back to a fresh introspection. So it's best-effort — never blocks the happy path.

shirshanka:
Addressed the recursion guard feedback — added `_fetching_schema` reentrant guard in aa1cbf4. Flag is set before the introspection call and cleared in a `finally` block. `adapt_query` checks it on entry and raises `RuntimeError` immediately if re-entered. Tests in `TestReentrantGuard` cover: raises on re-entry, cleared after success, cleared after failure.

PR #16526, opened 2026-03-11
fix(ingest): kafka platform instance helper

kevinkarchacryl:
When building column-level lineage URNs for Kafka topics, the code was looking up the Kafka platform instance using only platform_instance_map, completely bypassing connect_to_platform_map. The helper get_platform_instance() already existed in the same file to check both maps so I switched it to that. Added a bunch of tests that are passing


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `926b82b`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22952595538)

PR #16527, opened 2026-03-11
fix(ingestion/snowflake-sql-parsing): strip governance DDL before SQL parsing to preserve lineage

acrylJonny:
## Summary

Snowflake governance DDL constructs (`WITH TAG`, `WITH MASKING POLICY`,
`WITH PROJECTION POLICY`, `WITH ROW ACCESS POLICY`) are not supported by
sqlglot. When encountered, sqlglot silently falls back to treating the entire
statement as an opaque `Command` node, which causes DataHub to raise
`UnsupportedStatementTypeError` and lose all lineage for that statement.

This fix sanitizes Snowflake SQL before parsing by stripping these constructs
via regex. They carry no table-reference information so their removal has no
effect on lineage correctness.

- Adds `_SNOWFLAKE_UNSUPPORTED_DDL_PATTERNS` and `_sanitize_snowflake_ddl` in
  `sqlglot_utils.py`
- Applies sanitization in `parse_statement` for Snowflake dialect only
- Updates snapshot goldens for the two existing `WITH TAG` tests to reflect
  the now-clean normalised SQL (lineage and column lineage are unchanged)
- Adds unit tests covering all four pattern types and false-positive cases
  (CTEs, `JOIN USING`, string literals) to guard against over-stripping

### DIFF
--- a/metadata-ingestion/src/datahub/sql_parsing/sqlglot_utils.py
+++ b/metadata-ingestion/src/datahub/sql_parsing/sqlglot_utils.py
@@ -129,7 +129,10 @@ def parse_statement(
     # to modify the parsed expression that sits in the cache. We keep
     # the cached versions pristine by returning a copy on each call.
     if isinstance(sql, str) and is_dialect_instance(dialect, "snowflake"):
-        sql = _sanitize_snowflake_ddl(sql)
+        sanitized = _sanitize_snowflake_ddl(sql)
+        if sanitized != sql:
+            logger.debug("Sanitized Snowflake DDL: %s -> %s", sql, sanitized)
+        sql = sanitized
     return _parse_statement(sql, dialect).copy()
 
 


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `df87a4c`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/23587359849)

metadata-ingestion/src/datahub/sql_parsing/sqlglot_utils.py
```diff
@@ -87,10 +124,15 @@ def parse_statement(
     # Because the expressions are mutable, we don't want to allow the caller
     # to modify the parsed expression that sits in the cache. We keep
     # the cached versions pristine by returning a copy on each call.
+    if isinstance(sql, str) and is_dialect_instance(dialect, "snowflake"):
+        sql = _sanitize_snowflake_ddl(sql)
```
alokr-dhub: adding a debug log will be helpful here

alokr-dhub:
@acrylJonny  LGTM. Please resolve the merge conflicts and logs as suggested.

acrylJonny:
> @acrylJonny LGTM. Please resolve the merge conflicts and logs as suggested.

@alokr-dhub - this should be resolved now. Let me know if this is now good to approve.
Thanks in advance

metadata-ingestion/src/datahub/sql_parsing/sqlglot_utils.py
acrylJonny: Debug added for when the SQL is updated

PR #16531, opened 2026-03-11
Optimise plugins

Dutt23:
Currently while loading the PDL files , we need to scan the entire class graph because of the packages it generates the java files in (com.linked.* , com.datahub.*). This is resource intensive and during startup slows the loading down significantly. 
Optimising this during build time by generating .idx files from different directories which have all the fqcn we need to load. The pdl files are converted to pdsc (json) files and then to their respective formats (java, avro). We read the pdsc files and generate this fqcn mapping. 

The plugin factory was creating classgraphs each time wasting resources. Since we had the classloaders , className, and baseclass. We try to load is directly saving resources and time. 

**Testing**
1. Modified plugin factory tests to use both old and new paths. 
2. Added test to DataSchemaFactoryTest to validate keys and classes in new and old paths. 
3. Tested locally by starting everything , ran basic flows. 
4. Deployed on staging env and tested for gms if service starts.

**Improvments**
Around 6-9 seconds improvements during startup.


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `13992f8`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/22965848068)

PR #16540, opened 2026-03-11
feat(ingestion/snowflake): extract dynamic table lineage from DYNAMIC_TABLE_GRAPH_HISTORY

acrylJonny:
## Summary

- Populates table-level lineage for dynamic tables from `DYNAMIC_TABLE_GRAPH_HISTORY().INPUTS`, which Snowflake pre-computes and makes available independently of DDL. This is used alongside SQL parsing when DDL is present, and as the sole lineage source when it is not.
- When `SHOW DYNAMIC TABLES` returns a `NULL` `text` column (i.e. the `MONITOR` privilege is not granted), a structured pipeline warning is now emitted per affected table and `num_dynamic_tables_missing_definition` is incremented in the source report, replacing a previously silent failure.
- Documentation updated to clarify that without `MONITOR`, dynamic table entities still appear in DataHub but lineage is not extracted.


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `a0dc367`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/26243438159)

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema_gen.py
```diff
@@ -596,21 +596,44 @@ def _process_tables(
...
+                            message=(
+                                "The DDL for this dynamic table could not be retrieved. "
+                                "Grant MONITOR on the dynamic table to enable lineage extraction."
+                            ),
+                            context=f"{db_name}.{schema_name}.{table.name}",
+                        )
+
+                    # Table-level lineage from DYNAMIC_TABLE_GRAPH_HISTORY().INPUTS —
+                    # supplements SQL parsing when DDL is available, sole source when not.
+                    for upstream_qualified_name in table.upstream_tables:
```
kyungsoo-datahub: When DDL is present, this also runs alongside add_view_definition. Since  add_known_lineage_mapping uses COPY lineage type (higher precedence than VIEW), its identity column-level lineage wins in the aggregator merge. Identity CLL assumes each downstream column maps to the same-named upstream column, which is wrong when the dynamic table's SQL aliases, aggregates, or computes columns (e.g. SUM(amount) AS total,  identity CLL maps total -> total instead of amount -> total). Should this loop be inside the else branch only?

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema_gen.py
```diff
@@ -596,21 +596,44 @@ def _process_tables(
...
+                        upstream_identifier = (
+                            self.identifiers.get_dataset_identifier_from_qualified_name(
+                                upstream_qualified_name
+                            )
+                        )
+                        self.aggregator.add_known_lineage_mapping(
```
kyungsoo-datahub: Dynamic tables are SQL-defined (like materialized views), not byte-for-byte copies. The default lineage_type=COPY is semantically incorrect here. Consider passing lineage_type=DatasetLineageTypeClass.VIEW.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema_gen.py
kyungsoo-datahub: Also, upstream table names from INPUTS aren't checked against configured database_pattern / schema_pattern / table_pattern filters. If an upstream is in an excluded database or schema, this creates a lineage edge pointing to a dataset that was never ingested — resulting in a phantom entity in DataHub. Should these be filtered before creating lineage edges?

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema.py
```diff
@@ -1865,18 +1867,27 @@ def get_dynamic_tables_with_definitions(
...
+                        if not target_lag:
+                            if graph_info.get("target_lag_type") and graph_info.get(
+                                "target_lag_sec"
+                            ):
+                                target_lag = f"{graph_info['target_lag_sec']} {graph_info['target_lag_type']}"
+                        raw_inputs = graph_info.get("inputs")
+                        if raw_inputs:
+                            # INPUTS is ARRAY of OBJECTs [{name, kind}, ...]; may be a JSON string.
+                            if isinstance(raw_inputs, str):
+                                raw_inputs = json.loads(raw_inputs)
```
kyungsoo-datahub: If this string is malformed JSON, the JSONDecodeError propagates up and skips all remaining dynamic tables in this database. Consider wrapping this in a try/except with a warning for the specific table, so one bad INPUTS value doesn't block the rest.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema_gen.py
acrylJonny: I've moved the loop into the `else` branch. When DDL is present, `add_view_definition` and SQL parsing handles lineage with accurate column-level lineage. INPUTS is now only used as a fallback when DDL is unavailable.
I've also fixed this based upon your second comment. `self.filters.is_dataset_pattern_allowed(upstream_identifier, SnowflakeObjectDomain.TABLE)` is now called before each `add_known_lineage_mapping` call, making it more consistent with how other lineage paths (e.g. semantic view column lineage) handle the mapping. Filtered upstreams are now also skipped with a debug log.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema_gen.py
acrylJonny: Updated to `lineage_type=DatasetLineageTypeClass.VIEW` now

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema.py
acrylJonny: Now wrapped in a try/catch block logging a warning if anything goes wrong. Unit test also added for this where we have one string malformed.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema_gen.py
```diff
@@ -596,21 +596,52 @@ def _process_tables(
...
+                        )
+                        # Fall back to table-level lineage from DYNAMIC_TABLE_GRAPH_HISTORY().INPUTS
+                        # when DDL is unavailable. Skipped when DDL is present because SQL parsing
+                        # produces accurate column-level lineage; identity CLL from INPUTS would be
+                        # wrong for aliased/aggregated columns (e.g. SUM(amount) AS total).
+                        for upstream_qualified_name in table.upstream_tables:
+                            upstream_identifier = self.identifiers.get_dataset_identifier_from_qualified_name(
+                                upstream_qualified_name
+                            )
+                            if not self.filters.is_dataset_pattern_allowed(
```
kyungsoo-datahub, line 647: INPUTS can contain views/dynamic tables, not just tables. kind field is available in each INPUTS entry but discarded — all upstreams are hardcoded as SnowflakeObjectDomain.TABLE. A view upstream gets checked against table_pattern instead of view_pattern, so it can either leak through as a dangling lineage edge or get  silently dropped. Propagate kind and map to the correct SnowflakeObjectDomain.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema_gen.py
```diff
@@ -596,21 +596,52 @@ def _process_tables(
...
+                    if table.definition:
+                        self.aggregator.add_view_definition(
+                            view_urn=downstream_urn,
+                            view_definition=table.definition,
+                            default_db=db_name,
+                            default_schema=schema_name,
+                        )
+                    else:
+                        self.report.num_dynamic_tables_missing_definition += 1
+                        self.structured_reporter.warning(
```
kyungsoo-datahub: Warning message says "lineage skipped" but the code right after adds table-level lineage from INPUTS. Also, missing MONITOR is a deliberate privilege choice — info fits better than warning for expected degraded behavior. Suggest updating the message to reflect that table-level lineage is still produced but column-level lineage requires MONITOR.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema.py
```diff
@@ -1927,8 +1939,9 @@ def populate_dynamic_table_definitions(
                             if show_dt.name == table.name:
                                 table.definition = show_dt.definition
                                 table.target_lag = show_dt.target_lag
+                                table.upstream_tables = show_dt.upstream_tables
                                 break
         except Exception as e:
-            logger.debug(
+            logger.warning(
```
kyungsoo-datahub: This is a catch-all around get_dynamic_tables_with_definitions, which already reports failures internally via structured reporter. Bumping to warning risks duplicate noise for the same root cause. Suggest keeping debug.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema_gen.py
acrylJonny: I've added a SnowflakeDynamicTableInput NamedTuple (name, kind) and changed upstream_tables from List[str] to List[SnowflakeDynamicTableInput]. INPUTS parsing now captures kind alongside the name, and the fallback lineage loop resolves it to the appropriate SnowflakeObjectDomain (defaulting to TABLE for anything unrecognised).

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema.py
acrylJonny: Downgraded to debug

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema_gen.py
acrylJonny: Downgraded to info and updated the message to say table-level lineage is still produced from INPUTS, column-level is what requires MONITOR.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema_gen.py
```diff
@@ -596,21 +596,61 @@ def _process_tables(
...
+                            context=f"{db_name}.{schema_name}.{table.name}",
+                        )
+                        # Fall back to table-level lineage from DYNAMIC_TABLE_GRAPH_HISTORY().INPUTS
+                        # when DDL is unavailable. Skipped when DDL is present because SQL parsing
+                        # produces accurate column-level lineage; identity CLL from INPUTS would be
+                        # wrong for aliased/aggregated columns (e.g. SUM(amount) AS total).
+                        _known_domains = {d.value for d in SnowflakeObjectDomain}
+                        for upstream_input in table.upstream_tables:
+                            upstream_qualified_name = upstream_input.name
+                            kind_lower = upstream_input.kind.lower()
```
kyungsoo-datahub: This silently misclassifies non-TABLE upstreams. Snowflake's INPUTS.kind returns "MATERIALIZED_VIEW", "EXTERNAL_TABLE", "DYNAMIC_TABLE" (underscored uppercase). SnowflakeObjectDomain values use spaces ("materialized view",  etc.). After .lower() the underscored values don't match -> falls back to TABLE, so filters get applied to the wrong *_pattern.

Fix: kind_lower = upstream_input.kind.lower().replace("_", " ")

And add a test with kind="MATERIALIZED_VIEW". None of the current tests exercise a non-TABLE kind, which is why this slipped.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema.py
```diff
@@ -1925,18 +1933,36 @@ def get_dynamic_tables_with_definitions(
...
+                                    raw_inputs = json.loads(raw_inputs)
+                                upstream_tables = [
+                                    SnowflakeDynamicTableInput(
+                                        name=inp["name"],
+                                        kind=inp.get("kind", "Table"),
+                                    )
+                                    for inp in raw_inputs
+                                    if isinstance(inp, dict) and inp.get("name")
+                                ]
+                            except json.JSONDecodeError as e:
```
kyungsoo-datahub, line 1966: Only catches JSONDecodeError, but the comprehension above assumes raw_inputs is a list of dicts. If json.loads succeeds and returns a non-list — e.g. "null" -> None, or a single object -> dict. The iteration raises TypeError/KeyError uncaught and crashes ingestion for the whole database
(the very thing the earlier try/except was added to prevent).

Either broaden the catch:
    except (json.JSONDecodeError, TypeError, KeyError) as e:

Or guard the shape explicitly (cleaner — makes the assumption visible):
    if isinstance(raw_inputs, str):
        raw_inputs = json.loads(raw_inputs)
    if not isinstance(raw_inputs, list):
        raw_inputs = []

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema_gen.py
acrylJonny: Fixed — extracted a static `_resolve_input_kind` helper that normalizes via `kind.lower().replace("_", " ")` and falls back to TABLE on unknown kinds. Added `test_resolve_input_kind_normalization` parametrized over TABLE, MATERIALIZED_VIEW, EXTERNAL_TABLE, DYNAMIC_TABLE, and an unknown kind.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema.py
acrylJonny: Fixed — went with the explicit shape guard (`if not isinstance(raw_inputs, list): raw_inputs = []`) after the `json.loads` call, since it keeps the assumption visible and won't silently swallow real `TypeError`/`KeyError` bugs in the comprehension. Added `test_get_dynamic_tables_with_definitions_inputs_non_list_json` parametrized over `null` and a single-object payload.

kyungsoo-datahub (approved):

PR #16541, opened 2026-03-11
refactor(ui): migrate Groups tab and CreateGroupModal to Alchemy components CAT-1546

annadoesdesign:
Before:
Screenshot 2026-03-11 at 4 52 44 PM
Screenshot 2026-03-11 at 4 52 51 PM

After:
Screenshot 2026-03-12 at 1 21 20 PM
Screenshot 2026-03-12 at 1 21 23 PM

## Summary

- **Migrate Groups tab** (`GroupList.tsx`) from antd/custom components to Alchemy equivalents: `Table`, `SearchBar`, `Pagination`, `Avatar`, `Pill`, `Menu`, `SimpleSelectRole`, and `Button` — preserving all existing functionality (CRUD, role assignment, search, onboarding tour)
- **Migrate CreateGroupModal** from antd `Form`/`Input`/`Collapse` to Alchemy `Input` (with built-in labels), `Label`, and `Button` toggle for the Advanced section
- **Fix flex height chain** through antd Tabs internals (`Content` → `RoutedTabsStyle` → `StyledTabsPrimary` → tab pane → `RouteContainer`) so the table fills available space with pagination pinned at the bottom instead of overflowing off-screen
- **Replace hardcoded colors** with semantic theme tokens (`theme.colors.textSecondary`, `theme.colors.border`, `theme.colors.bg`) across all touched files
- **Remove extra padding** from ServiceAccountList containers for consistency

## Checklist

- [x] PR conforms to DataHub's [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (PR Title Format)
- [x] No breaking changes — all existing functionality preserved
- [x] No new dependencies added
- [x] `yarn lint` (eslint + prettier + type-check) passes
- [ ] Tests for the changes have been added/updated (no existing tests affected; CSS-only layout changes)
- [ ] Docs related to the changes have been added/updated (not applicable — internal UI refactor)

## Test plan

- [ ] Verify Groups tab renders correctly: table, search, pagination, avatars, pills, role selector, actions menu
- [ ] Verify Create Group modal opens, validates inputs, and creates groups successfully
- [ ] Verify group deletion via actions menu works
- [ ] Verify role assignment via inline select works with confirmation modal
- [ ] Verify table scrolls internally with pagination pinned at the bottom (not pushed off-screen)
- [ ] Verify Users and Service Accounts tabs are unaffected
- [ ] Verify light and dark theme rendering (semantic color tokens)

### DIFF
--- a/datahub-web-react/src/app/identity/ManageUsersAndGroups.components.tsx
+++ b/datahub-web-react/src/app/identity/ManageUsersAndGroups.components.tsx
@@ -79,34 +79,32 @@ export const ManageUsersAndGroupsHeader = ({
     onCreateServiceAccount,
     onCreateGroup,
 }: ManageUsersAndGroupsHeaderProps) => {
-    const isServiceAccountsTab = activeTab === 'service-accounts';
-    const isGroupsTab = activeTab === 'groups';
-
     const renderActionButton = () => {
-        if (isServiceAccountsTab) {
-            return (
-                <Button
-                    variant="filled"
-                    disabled={!canManageServiceAccounts}
-                    onClick={onCreateServiceAccount}
-                    data-testid="create-service-account-button"
-                >
-                    Create Service Account
-                </Button>
-            );
+        switch (activeTab) {
+            case 'service-accounts':
+                return (
+                    <Button
+                        variant="filled"
+                        disabled={!canManageServiceAccounts}
+                        onClick={onCreateServiceAccount}
+                        data-testid="create-service-account-button"
+                    >
+                        Create Service Account
+                    </Button>
+                );
+            case 'groups':
+                return (
+                    <Button variant="filled" onClick={onCreateGroup} data-testid="create-group-button">
+                        Create Group
+                    </Button>
+                );
+            default:
+                return (
+                    <Button variant="filled" disabled={!canManageUsers} onClick={onInviteUsers}>
+                        Invite Users
+                    </Button>
+                );
         }
-        if (isGroupsTab) {
-            return (
-                <Button variant="filled" onClick={onCreateGroup} data-testid="create-group-button">
-                    Create Group
-                </Button>
-            );
-        }
-        return (
-            <Button variant="filled" disabled={!canManageUsers} onClick={onInviteUsers}>
-                Invite Users
-            </Button>
-        );
     };
 
     return (
--- a/datahub-web-react/src/app/identity/group/CreateGroupModal.tsx
+++ b/datahub-web-react/src/app/identity/group/CreateGroupModal.tsx
@@ -2,15 +2,15 @@ import { CaretDown } from '@phosphor-icons/react/dist/csr/CaretDown';
 import { CaretRight } from '@phosphor-icons/react/dist/csr/CaretRight';
 import { message } from 'antd';
 import React, { useState } from 'react';
-import styled from 'styled-components/macro';
 
 import analytics, { EventType } from '@app/analytics';
 import { useUserContext } from '@app/context/useUserContext';
 import { ActorsSearchSelect } from '@app/entityV2/shared/EntitySearchSelect/ActorsSearchSelect';
 import { ActorEntity } from '@app/entityV2/shared/utils/actorUtils';
+import { AdvancedButton, AdvancedContent, FormSection } from '@app/identity/group/CreateGroupModal.components';
 import { validateCustomUrnId } from '@app/shared/textUtil';
 import { useEnterKeyListener } from '@app/shared/useEnterKeyListener';
-import { Button, Input, Modal, TextArea } from '@src/alchemy-components';
+import { Input, Modal, TextArea } from '@src/alchemy-components';
 
 import { useAddGroupMembersMutation, useCreateGroupMutation } from '@graphql/group.generated';
 import { useAddOwnerMutation } from '@graphql/mutations.generated';
@@ -20,23 +20,6 @@ type Props = {
     onClose: () => void;
     onCreate: (group: CorpGroup) => void;
 };
-
-const FormSection = styled.div`
-    display: flex;
-    flex-direction: column;
-    gap: 4px;
-    margin-bottom: 16px;
-`;
-
-const AdvancedContent = styled.div`
-    margin-top: 12px;
-`;
-
-const AdvancedButton = styled(Button)`
-    padding-left: 0;
-    padding-right: 0;
-    color: ${(props) => props.theme.colors.textSecondary};
-`;
 
 export default function CreateGroupModal({ onClose, onCreate }: Props) {
     const { urn: currentUserUrn } = useUserContext();
--- a/datahub-web-react/src/app/identity/group/GroupList.tsx
+++ b/datahub-web-react/src/app/identity/group/GroupList.tsx
@@ -16,6 +16,7 @@ import {
     GroupRoleCell,
     ModalFooter,
     PageContainer,
+    PaginationContainer,
     SearchContainer,
     TableContainer,
 } from '@app/identity/group/GroupList.components';
@@ -200,14 +201,14 @@ export const GroupList = ({
                 {groups.length > 0 ? (
                     <>
                         <Table columns={columns} data={groups} isLoading={loading} isScrollable />
-                        <div style={{ paddingTop: '8px', display: 'flex', justifyContent: 'center' }}>
+                        <PaginationContainer>
                             <Pagination
                                 currentPage={page}
                                 itemsPerPage={pageSize}
                                 total={totalGroups}
                                 onPageChange={onChangePage}
                             />
-                        </div>
+                        </PaginationContainer>
                     </>
                 ) : (
                     <EmptyState
--- a/datahub-web-react/src/app/identity/useRoleAssignment.ts
+++ b/datahub-web-react/src/app/identity/useRoleAssignment.ts
@@ -67,40 +67,41 @@ export function useRoleAssignment({
         const roleToAssign = selectRoleOptions.find((role) => role.urn === roleAssignmentState.currentRoleUrn);
         const newRoleUrn = roleToAssign?.urn || NO_ROLE_URN;
 
+        const isRemoval = newRoleUrn === NO_ROLE_URN;
+        const successMsg = isRemoval
+            ? `Removed role from ${entityLabel} ${roleAssignmentState.actorName}!`
+            : `Assigned role ${roleToAssign?.name} to ${entityLabel} ${roleAssignmentState.actorName}!`;
+
         batchAssignRoleMutation({
             variables: {
                 input: {
-                    roleUrn: newRoleUrn === NO_ROLE_URN ? null : newRoleUrn,
+                    roleUrn: isRemoval ? null : newRoleUrn,
                     actors: [roleAssignmentState.actorUrn],
                 },
             },
         })
             .then(({ errors }) => {
                 if (!errors) {
-                    toast.success(
-                        newRoleUrn === NO_ROLE_URN
-                            ? `Removed role from ${entityLabel} ${roleAssignmentState.actorName}!`
-                            : `Assigned role ${roleToAssign?.name} to ${entityLabel} ${roleAssignmentState.actorName}!`,
-                    );
+                    toast.success(successMsg);
 
                     onSuccess?.(roleAssignmentState.actorUrn, newRoleUrn);
                     const { actorUrn } = roleAssignmentState;
                     setRoleAssignmentState(null);
 
-                    setTimeout(async () => {
+                    setTimeout(() => {
                         clearRoleListCache(client);
-                        await refetch();
-                        onPostRefetch?.(actorUrn);
+                        Promise.resolve(refetch())
+                            .then(() => onPostRefetch?.(actorUrn))
+                            .catch((err) => console.error('Failed to refetch after role assignment:', err));
                     }, 3000);
                 }
             })
             .catch((e) => {
                 onError?.(roleAssignmentState.actorUrn, roleAssignmentState.originalRoleUrn);
-                toast.error(
-                    newRoleUrn === NO_ROLE_URN
-                        ? `Failed to remove role from ${entityLabel} ${roleAssignmentState.actorName}: ${e.message || ''}`
-                        : `Failed to assign role ${roleToAssign?.name} to ${entityLabel} ${roleAssignmentState.actorName}: ${e.message || ''}`,
-                );
+                const failureMsg = isRemoval
+                    ? `Failed to remove role from ${entityLabel} ${roleAssignmentState.actorName}: ${e.message || ''}`
+                    : `Failed to assign role ${roleToAssign?.name} to ${entityLabel} ${roleAssignmentState.actorName}: ${e.message || ''}`;
+                toast.error(failureMsg);
             });
     }, [
         roleAssignmentState,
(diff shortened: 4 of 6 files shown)


datahub-web-react/src/app/entity/user/UserGroups.tsx
```diff
@@ -91,8 +91,13 @@ export default function UserGroups({ urn, initialRelationships, pageSize }: Prop
     };
 
     const relationships = groupsData ? groupsData.corpUser?.relationships?.relationships : initialRelationships;
-    const total = relationships?.length || 0;
-    const userGroups = relationships?.map((rel) => rel.entity as CorpGroup) || [];
+    const userGroups = [...(relationships || [])]
+        .filter((rel) => {
+            const group = rel?.entity as CorpGroup;
+            return group?.info || group?.editableProperties;
+        })
```
chriscollins3456, line 99: not against this filter but i assume there was some bug we're fixing here?

datahub-web-react/src/app/entity/user/UserProfile.tsx
```diff
@@ -132,7 +138,14 @@ export default function UserProfile() {
...
                         <UserInfoSideBar sideBarData={sideBarData} refetch={refetch} />
                     </Col>
-                    <Col xl={19} lg={19} md={19} sm={24} xs={24} style={{ borderLeft: '1px solid #E9E9E9' }}>
+                    <Col
+                        xl={19}
+                        lg={19}
+                        md={19}
+                        sm={24}
+                        xs={24}
+                        style={{ borderLeft: `1px solid ${theme.colors.border}` }}
```
chriscollins3456: minor: could use a styled component here instead

datahub-web-react/src/app/entityV2/shared/sidebarSection/UserGroupSidebarSection.tsx
```diff
@@ -28,15 +21,27 @@ type Props = {
...
+
+    const groupLinkStyle = {
+        backgroundColor: theme.colors.bgHover,
+        '& svg': {
+            color: theme.colors.icon,
+        },
+    };
+    const validGroups = groupsDetails.filter((detail) => {
+        const group = detail?.entity;
+        return group && ((group as any).info || (group as any).editableProperties);
```
chriscollins3456: i always try to avoid any type castling like this but especially `as any`

i can see why in this situation that would be tough though after looking that `groupsDetails` is `EntityWithRelationship[]` so we have no typing on the `entity`

so maybe this slides for now... but ideally we fix the typing upstream so we don't have to cast like this

in fact i have some suggestions on how we can improve these sorts of type casting and wrote them in this notion doc: https://www.notion.so/acryldata/Typescript-Best-Practices-231fc6a6427780b5889bd09ea40b55ac?source=copy_link#231fc6a64277808eb016e6f1a89b5ae1

datahub-web-react/src/app/identity/serviceAccount/ServiceAccountList.components.tsx
```diff
@@ -15,30 +15,14 @@ export const ServiceAccountContainer = styled.div`
...
-        z-index: 1;
-        background: ${(props) => props.theme.colors.bg};
-    }
-
-    /* Ensure header cells have proper background */
-    .ant-table-thead > tr > th {
-        background: ${(props) => props.theme.colors.bg} !important;
-        border-bottom: 1px solid ${(props) => props.theme.colors.border};
-    }
+    overflow: hidden;
```
chriscollins3456, line 31: cool so we're updating the service accounts tab as well?

datahub-web-react/src/app/identity/group/CreateGroupModal.tsx
```diff
@@ -28,105 +42,117 @@ export default function CreateGroupModal({ onClose, onCreate }: Props) {
...
-        if (isEditorNewlineKeypress) {
-            createGroupMutation({
-                variables: {
-                    input: {
-                        id: stagedId,
-                        name: stagedName,
-                        description: stagedDescription,
-                    },
+        if (document.activeElement instanceof HTMLTextAreaElement) return;
+        if (!validateName(stagedName) || !validateId(stagedId)) return;
```
chriscollins3456, line 64: okay cool so we're moving to US being the one to fully manage validation instead of passing in validation rules to the `form` component?

datahub-web-react/src/app/identity/group/GroupList.tsx
```diff
@@ -1,64 +1,314 @@
...
+type GroupRoleCellProps = {
+    group: CorpGroup;
+    selectRoleOptions: DataHubRole[];
+    optimisticRoleUrn?: string;
+    onRoleChange: (groupUrn: string, groupName: string, newRoleUrn: string, originalRoleUrn: string) => void;
+};
+
+const GroupRoleCell = ({ group, selectRoleOptions, optimisticRoleUrn, onRoleChange }: GroupRoleCellProps) => {
+    const entityRegistry = useEntityRegistry();
+    const castedGroup = group as any;
```
chriscollins3456: yeah not a fan of the type casting `as any`

datahub-web-react/src/app/identity/group/GroupList.tsx
```diff
@@ -1,64 +1,314 @@
...
+};
+
+type GroupRoleCellProps = {
+    group: CorpGroup;
+    selectRoleOptions: DataHubRole[];
+    optimisticRoleUrn?: string;
+    onRoleChange: (groupUrn: string, groupName: string, newRoleUrn: string, originalRoleUrn: string) => void;
+};
+
+const GroupRoleCell = ({ group, selectRoleOptions, optimisticRoleUrn, onRoleChange }: GroupRoleCellProps) => {
```
chriscollins3456: in general i'm a fan of moving individual components into their own files. you can see how extremely bloated this one `GroupList` file is - it's so hard to navigate and reason about anything if i'm looking for something specific

overall this file could use some breaking out and separation of concerns

datahub-web-react/src/app/identity/group/GroupList.tsx
```diff
@@ -1,64 +1,314 @@
...
+                if (newRoleUrn !== currentRoleUrn) {
+                    onRoleChange(group.urn, displayName, newRoleUrn, serverRoleUrn);
+                }
+            }}
+            placeholder="No Role"
+            size="md"
+            width="fit-content"
+        />
+    );
+};
```
chriscollins3456: could we reuse the same role selector that's used in the users list? and then reuse the same mutation so we don't have to duplicate that mutation logic you have lower in this file where you call `onConfirmRoleAssignment`

these look like the same components, the only difference is which urn you pass into the mutation. so ideally they're just the same component

smoke-test/tests/cypress/cypress/e2e/mutations/dataset_ownership.js
```diff
@@ -63,7 +63,9 @@ describe("add, remove ownership for dataset", () => {
   it("create test user and test group, add user to a group", () => {
     cy.loginWithCredentials();
     cy.createUser(username, password, email);
+    cy.setIsThemeV2Enabled(true);
     cy.createGroup(group_name, "Test group description", test_id);
+    cy.setIsThemeV2Enabled(false);
```
chriscollins3456: what's going on here? for a single test you should not have to change the feature flags in the middle of it

smoke-test/tests/cypress/cypress/e2e/settingsV2/v2_managing_groups.js
```diff
@@ -52,18 +52,19 @@ describe("create and manage group", () => {
...
+    cy.visitWithLogin("/settings/identities/groups");
     cy.get('[data-testid="manage-users-groups-v2"]', { timeout: 10000 }).should(
       "be.visible",
     );
-    cy.get(".ant-tabs-tab-btn").contains("Groups").click();
-    // Wait for groups tab to load and group to be visible
-    cy.contains(group_name, { timeout: 10000 }).should("be.visible");
-    cy.clickOptionWithText(group_name);
-    cy.clickTextOptionWithClass(".ant-typography", group_name);
-    // cy.get(".ant-typography").contains(group_name).should("be.visible");
```
chriscollins3456, line 64: seems like we're getting rid of the functionality to click on the groups tab to take you there? why?

datahub-web-react/src/app/identity/group/GroupList.tsx
jjoyce0510: +1

datahub-web-react/src/app/entity/user/UserGroups.tsx
annadoesdesign: Good call — this filter removes soft-deleted or orphaned groups that have lost both their `info` and `editableProperties`. Without it, the UI would render empty/broken group cards. Added a comment explaining the intent.

datahub-web-react/src/app/entity/user/UserProfile.tsx
annadoesdesign: Done — converted to a `ContentColumn` styled component.

datahub-web-react/src/app/entityV2/shared/sidebarSection/UserGroupSidebarSection.tsx
annadoesdesign: Agreed — updated to cast as `CorpGroup` instead of `any`. That gives us proper type narrowing for `info` and `editableProperties`. Will check out the Notion doc for longer-term improvements to how we handle these relationship entity types upstream.

datahub-web-react/src/app/identity/serviceAccount/ServiceAccountList.components.tsx
annadoesdesign: Yes — we're migrating the service accounts tab to use the same Alchemy table/components pattern as groups and users for consistency across all identity management tabs.

datahub-web-react/src/app/identity/group/CreateGroupModal.tsx
annadoesdesign: Correct — we dropped the Antd `Form` component and handle validation directly. The modal fields are simple enough (name required, optional description/ID) that inline validation keeps it straightforward without the overhead of Antd's form validation framework. The submit button is disabled until the name field is filled, and we show inline error states.

datahub-web-react/src/app/identity/group/GroupList.tsx
annadoesdesign: Fixed — created a `ListGroupsGroup` type derived from the generated `ListGroupsQuery`, which properly types `memberCount` and `roles` without any `as any` casts.

datahub-web-react/src/app/identity/group/GroupList.tsx
annadoesdesign: Agreed — split GroupList into `GroupList.tsx` (main logic) and `GroupList.components.tsx` (styled components, cell components, actions menu). Follows the same pattern as `ServiceAccountList` + `ServiceAccountList.components.tsx`.

smoke-test/tests/cypress/cypress/e2e/mutations/dataset_ownership.js
annadoesdesign: Good catch — moved the V2 toggle into the `createGroup` Cypress command itself, since the "Create Group" button only exists in V2 mode. The test no longer toggles flags mid-test; `createGroup` handles it internally and the test just restores V1 mode afterward for its remaining assertions.

smoke-test/tests/cypress/cypress/e2e/settingsV2/v2_managing_groups.js
annadoesdesign: We're not removing that functionality — the Groups tab navigation still works the same way. The test was updated because in V2, the groups page loads directly at `/settings/identities/groups` without needing an explicit tab click. The "Groups" tab click was only needed in the V1 settings layout where you had to navigate between tabs manually.

datahub-web-react/src/app/identity/group/CreateGroupModal.tsx
```diff
@@ -18,8 +21,21 @@ type Props = {
...
+`;
+
+const AdvancedContent = styled.div`
+    margin-top: 12px;
+`;
+
+const AdvancedButton = styled(Button)`
+    padding-left: 0;
+    padding-right: 0;
+    color: ${(props) => props.theme.colors.textSecondary};
```
benjiaming: Ideally we should refactor styles into .components. file (especially for larger files).

datahub-web-react/src/app/identity/group/GroupList.tsx
```diff
@@ -91,87 +126,129 @@ export const GroupList = () => {
...
-                        }}
+                    </SearchContainer>
+                </FiltersHeader>
+            </GroupContainer>
+
+            <TableContainer>
+                {groups.length > 0 ? (
+                    <>
+                        <Table columns={columns} data={groups} isLoading={loading} isScrollable />
+                        <div style={{ paddingTop: '8px', display: 'flex', justifyContent: 'center' }}>
```
benjiaming: Nit: we should not be embedding styles inline (not until we switch to Tailwind!)

datahub-web-react/src/app/identity/ManageUsersAndGroups.components.tsx
```diff
@@ -76,8 +77,37 @@ export const ManageUsersAndGroupsHeader = ({
     activeTab,
     onInviteUsers,
     onCreateServiceAccount,
+    onCreateGroup,
 }: ManageUsersAndGroupsHeaderProps) => {
     const isServiceAccountsTab = activeTab === 'service-accounts';
+    const isGroupsTab = activeTab === 'groups';
+
+    const renderActionButton = () => {
+        if (isServiceAccountsTab) {
```
benjiaming: Nit: missed opportunity for a `switch`!

datahub-web-react/src/app/identity/useRoleAssignment.ts
```diff
@@ -0,0 +1,132 @@
...
+                            : `Assigned role ${roleToAssign?.name} to ${entityLabel} ${roleAssignmentState.actorName}!`,
+                    );
+
+                    onSuccess?.(roleAssignmentState.actorUrn, newRoleUrn);
+                    const { actorUrn } = roleAssignmentState;
+                    setRoleAssignmentState(null);
+
+                    setTimeout(async () => {
+                        clearRoleListCache(client);
+                        await refetch();
```
benjiaming: Technically `refetch` can fail too. We should probably try/catch (use .then() without async/await).

datahub-web-react/src/app/identity/useRoleAssignment.ts
```diff
@@ -0,0 +1,132 @@
...
+                    }, 3000);
+                }
+            })
+            .catch((e) => {
+                onError?.(roleAssignmentState.actorUrn, roleAssignmentState.originalRoleUrn);
+                toast.error(
+                    newRoleUrn === NO_ROLE_URN
+                        ? `Failed to remove role from ${entityLabel} ${roleAssignmentState.actorName}: ${e.message || ''}`
+                        : `Failed to assign role ${roleToAssign?.name} to ${entityLabel} ${roleAssignmentState.actorName}: ${e.message || ''}`,
+                );
```
benjiaming: Instead of repeating ternary logic in toasts we could precalculate the message e.g.
```
const isRemoval = newRoleUrn === NO_ROLE_URN;
const successMsg = isRemoval ? `Removed role from ...` : `Assigned role ...`;
const failureMsg = isRemoval ? `Failed to remove role from ...` : `Failed to assign role ...`;
```

benjiaming:
Thanks for migrating Groups!

PR #16544, opened 2026-03-12
feat(ingest/test): presort output data to match golden data

ligfx:
In compare_metadata_json, presort output data to match the same sort as the existing golden file to give better output when MCEs are present.

### DIFF
--- a/metadata-ingestion/src/datahub/testing/compare_metadata_json.py
+++ b/metadata-ingestion/src/datahub/testing/compare_metadata_json.py
@@ -155,6 +155,10 @@ def sort_output_in_same_order_as_golden(
     # iterate through golden and for each entry, find the matching one from output if it exists
     # then append any extras from output that weren't found in golden
 
+    if not isinstance(output, list) or not isinstance(golden, list):
+        # avoid pathological inputs, as seen in tests/unit/sql_parsing/test_sql_aggregator.py::test_aggregator_dump
+        return output
+
     # 1. build matching index for output
     mcps_index: dict[tuple[str, str], MetadataJson] = {}
     mces_index: dict[str, MetadataJson] = {}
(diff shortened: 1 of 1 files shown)


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `c5fee0d`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/23171817557)

sgomezvillamor:
**Minor bug in `other_counts`**

`other_counts: dict[dict[str, Any], int]` uses a `dict` as a dictionary key, but Python dicts are not hashable — any entry that is neither an MCP nor an MCE will raise `TypeError: unhashable type: 'dict'` at runtime. In practice this path rarely triggers since most output is MCPs/MCEs, but it is a latent crash. The simplest fix is replacing the counter-dict with a plain list.

---

**Background**

I've thought about tackling this at the source — normalizing entry order when golden files are written — rather than sorting at comparison time. This feels more principled but requires migrating all existing golden files, which is significant churn.

---

**Three alternatives worth considering before committing to this approach:**

**Alt 1: Canonical sort on both sides during comparison**
Sort both output and golden by `(entityUrn, aspectName)` before diffing, instead of sorting output to match golden's existing (possibly arbitrary) order.
- ✅ Symmetric, fixes the hashable-key bug, less code, works before a golden file exists
- ❌ Diffs appear in canonical order rather than the order a developer sees in their golden file

**Alt 2: Sort when writing golden files**
Make the golden update path always write entries in a canonical sort order. Run a one-time migration to re-sort all existing goldens.
- ✅ No sorting at comparison time; golden files stay consistently ordered forever; solves the problem permanently
- ❌ Large one-time migration PR touching all golden files; need to guard non-test uses of `write_metadata_file`

**Alt 3: Improve MCE diff display instead of sorting**
`MCPDiff` already groups diffs by URN and is readable. The actual pain point is the fallback to `DeepDiff` for MCEs. Improving the MCE diff path to group by URN would fix readability without any reordering.
- ✅ Fixes the root cause rather than working around it; no golden file churn
- ❌ More invasive change to the diff logic itself

---

Wondering if you considered any of these? — would appreciate feedback on them before deciding on this PR.

---

PR #16545, opened 2026-03-12
fix(ingest/glue): in Glue Jobs don't modify S3 URNs or override datasetProperties

ligfx:
fix(ingest/glue): don't add format extension to Glue Job S3 URNs
- Remove logic that would add the format extension to the end of S3 URLs in Glue Job nodes if they appeared multiple times. This makes no sense as a legitimate S3 path and doesn't match with how the S3 connector creates URNs.

fix(ingest/glue): don't override S3 entity datasetProperties when seen in Glue Jobs
- Remove logic that would ingest an S3 entity and set its datasetProperties aspect (overwriting datasetProperties from the actual S3 ingestion source) if it appeared in a Glue Job node.

fix(ingest/glue): fix S3 URI not found message
- small fix while we're here

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
@@ -880,7 +880,7 @@ class GlueSource(StatefulIngestionSourceBase):
                 if s3_uri is None:
                     self.report_warning(
                         flow_urn,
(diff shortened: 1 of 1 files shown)


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `35150fd`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/24505821173)

ligfx:

- #16547
    - #16545 ◀
        - #16546
            - #16609
                - #16636
                    - #16653

PR #16554, opened 2026-03-12
feat(ui/search): bring back support other languages in search bar PR and other relevant changes

purnimagarg1:
**Linear ticket:**
https://linear.app/acryl-data/issue/CAT-1497/bring-back-support-other-languages-in-search-bar-pr-to-oss

**Description:**

This PR brings back the following PRs to OSS:

https://github.com/acryldata/datahub-fork/pull/6025
https://github.com/acryldata/datahub-fork/pull/6380
https://github.com/acryldata/datahub-fork/pull/7318
https://github.com/acryldata/datahub-fork/pull/8386

It also removes some conditional styling considering that `isShowNavBarRedesign` flag is by default on for everyone. This was done to prevent lint issues due to hardcoded color values in relevant files.
We can remove the logic of using this flag throughout in a separate ticket.


datahub-web-react/src/app/searchV2/searchBarV2/hooks/useSelectOption.ts
```diff
@@ -13,7 +13,6 @@ import { FacetFilterInput } from '@src/types.generated';
 
 export default function useSelectOption(
     onSearch: (query: string, filters?: FacetFilterInput[]) => void,
-    onClear: () => void,
```
chriscollins3456, line 16: hmm i'm not seeing this change in saas anywhere? is this needed?

datahub-web-react/src/app/searchV2/searchBarV2/SearchBarV2.tsx
```diff
@@ -103,30 +107,31 @@ export const SearchBarV2 = ({
...
     );
 
     const clearQueryAndFilters = useCallback(() => {
+        setInputValue('');
         setSearchQuery('');
         clear();
     }, [clear]);
 
-    const selectOption = useSelectOption(onSearch, clearQueryAndFilters, flatAppliedFilters);
+    const selectOption = useSelectOption(onSearch, flatAppliedFilters);
```
chriscollins3456, line 120: yeah just not positive about this change but will wait to hear from you

datahub-web-react/src/app/searchV2/searchBarV2/components/SearchBarInput.tsx
```diff
@@ -10,29 +10,9 @@ import { BOX_SHADOW } from '@app/searchV2/searchBarV2/constants';
 import { Icon, SearchBar, colors, radius, transition } from '@src/alchemy-components';
 import { useShowNavBarRedesign } from '@src/app/useShowNavBarRedesign';
 
-const PRE_NAV_BAR_REDESIGN_SEARCHBAR_BACKGROUND = '#343444';
-
-const StyledSearchBar = styled(SearchBar)<{ $isShowNavBarRedesign?: boolean }>`
```
chriscollins3456: i know i told you i don't think we don't need to support this anymore, but in OSS i think we should delay a little longer i think some folks do use the old nav bar redesign still, so that was my bad

chriscollins3456 (commented):
thank you! just a comment and a question

datahub-web-react/src/app/searchV2/searchBarV2/hooks/useSelectOption.ts
purnimagarg1: In SaaS, this is still being used for an additional condition with option type "Ask Datahub". This option and condition is not present in OSS, hence the prop `onClear` is not used in the hook anymore. That's why removed it in just OSS.

datahub-web-react/src/app/searchV2/searchBarV2/SearchBarV2.tsx
purnimagarg1: this is for the same reason as above

datahub-web-react/src/app/searchV2/searchBarV2/components/SearchBarInput.tsx
purnimagarg1: oh, we'll then have to add these hardcoded colors in the theme files. Or find some new ones after discussing with Anna maybe.

datahub-web-react/src/app/searchV2/searchBarV2/hooks/useSelectOption.ts
chriscollins3456: nice

datahub-web-react/src/app/searchV2/searchBarV2/components/SearchBarInput.tsx
chriscollins3456: yeah let's just find the closes matches here in the new theme files even if they're not perfect I think. but we can double check with anna. maybe i'd share a screenshot of whatever you land on color-wise just to be sure! thank you

chriscollins3456 (approved):

PR #16557, opened 2026-03-12
fix(ingestion/kafka-connect): produce datajobs for JDBC sink when runtime topics API is empty

acrylJonny:
## Summary

- `JdbcSinkConnector.extract_lineages()` intersects the runtime topic list (from Kafka Connect's `/connectors/{name}/topics` API) with the topics from the connector's config. When the runtime API returns an empty list — which happens when a connector hasn't yet processed any messages or after a topic reset — this intersection is always empty, so no lineage edges and no DataJob entities are emitted, even though the DataFlow (pipeline) is.
- Fix: when the runtime topic set is empty but the config topic set is non-empty, fall back to the config topics directly instead of performing an empty intersection.


rajatoss:
## Connector Tests Results

**Connector tests failed** for commit `be7bf66`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/23141449100)

PR #15324, opened 2025-11-18
feat(masking): support masking secrets inside quotations

kyungsoo-datahub:
Register quoted versions of secrets (single and double quotes) to mask them in error messages that format values with quotes, such as "Role '8080' not found". Includes comprehensive test coverage for the new functionality.

PR #16567, opened 2026-03-12
fix(test): remove flaky test_pattern_rebuild_no_lock_contention 

kyungsoo-datahub:
Removes `test_pattern_rebuild_no_lock_contention` from `test_masking_filter.py`.

The test asserted wall-clock thresholds (0.1s rebuild, 2.0s concurrent log) to verify that `re.compile()` happens outside the pattern lock. These thresholds vary with CI runner load — the test failed on 3.10 testQuick with 0.94s against a 0.2s threshold (PR #15324).

The property it guards — no lock contention during pattern rebuild — can't actually deadlock. The code uses a single `RLock` with no circular wait. The lock discipline is 6 lines of straightforward code, better caught in review than a timing test.

### DIFF
--- a/metadata-ingestion/tests/unit/test_masking_filter.py
+++ b/metadata-ingestion/tests/unit/test_masking_filter.py
@@ -724,69 +724,6 @@ class TestEdgeCases:
 class TestP1Fixes:
     """Test P1 critical fixes from production hardening review."""
 
-    def test_pattern_rebuild_no_lock_contention(self):
-        """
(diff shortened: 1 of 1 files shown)


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `1851e02`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/23038150551)

sgomezvillamor (approved):

PR #16573, opened 2026-03-13
feat: Entity Change Event Generator for structured properties

DiogoVala:
Modifying structured properties wasn't producing change events, which are crucial for audit purposes.
This feature implements change events so that every Upsert or Removal of structured properties from assets generates ADD and/or REMOVE events according to the diff of the change.

### DIFF
--- a/metadata-io/src/main/java/com/linkedin/metadata/timeline/data/entity/StructuredPropertyChangeEvent.java
+++ b/metadata-io/src/main/java/com/linkedin/metadata/timeline/data/entity/StructuredPropertyChangeEvent.java
@@ -1,0 +1,41 @@
+package com.linkedin.metadata.timeline.data.entity;
+
+import com.google.common.collect.ImmutableMap;
+import com.linkedin.common.AuditStamp;
+import com.linkedin.common.urn.Urn;
+import com.linkedin.metadata.timeline.data.ChangeCategory;
+import com.linkedin.metadata.timeline.data.ChangeEvent;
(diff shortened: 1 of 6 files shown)


david-leifker:
@DiogoVala - please run ./gradlew spotlessApply and push the changes. Thanks!

DiogoVala:
> @DiogoVala - please run ./gradlew spotlessApply and push the changes. Thanks!

Thank you for the quick review! Linting changes have been pushed.

asikowitz:
Hey @DiogoVala, thanks for your contribution! We've implemented this on our own here: https://github.com/datahub-project/datahub/pull/16701. The upsert implementation is a bit different than yours -- we emit a MODIFY event rather than a REMOVE and ADD when a structured property's value is changed. Does this satisfy your audit use case?

DiogoVala:
> Hey @DiogoVala, thanks for your contribution! We've implemented this on our own here: #16701. The upsert implementation is a bit different than yours -- we emit a MODIFY event rather than a REMOVE and ADD when a structured property's value is changed. Does this satisfy your audit use case?

This isn't ideal because if you have a structured property with values "1", "2" and "3" and you modify "3" to "4", you get an event saying the structured property was modified to "1", "2", "4", but you don't know the previous state.
For audit purposes it's important to know exactly what got changed into what.

From an auditor's perspective, it would be very difficult go into the logs to find where a specific value was removed. In this example, they couldn't look for where "3" got removed.

PR #16574, opened 2026-03-13
fix(ingest): merge generatePyprojectDeps into updateLockFile

kyungsoo-datahub:
This PR merge generatePyprojectDeps into updateLockFile so that we have

- one for validate all these files are in-sync (pyproject.toml, uv.lock and constraints.txt)
- one for udpating all these files (pyproject.toml, uv.lock and constraints.txt)


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `b6f8794`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/23051246910)

metadata-ingestion/build.gradle
```diff
@@ -113,15 +113,10 @@ task checkLockFile(type: Exec, dependsOn: installPackageOnly) {
...
 }
 
-task generatePyprojectDeps(type: Exec, dependsOn: installPackageOnly) {
-  commandLine 'bash', '-c',
-    venv_activate_command +
-    "python scripts/generate_pyproject_deps.py"
-}
-
-task updateLockFile(type: Exec, dependsOn: generatePyprojectDeps) {
+task updateLockFile(type: Exec, dependsOn: installPackageOnly) {
```
sgomezvillamor, line 116: we may also need some input and output, no?

kyungsoo-datahub: `updateLockFile` intentionally has no inputs/outputs. It's a developer command that should always re-run to regenerate the full chain. Adding Gradle caching would cause it to be skipped when you actually want it to execute. 
`checkLockFile` has inputs/outputs because it's a CI gate. Gradle caching avoids redundant re-runs when nothing changed.

PR #16578, opened 2026-03-13
fix(logging): reduce log volume phase 2 — demote per-event/per-request INFO logging

max-datahub:
## Summary

Follow-up to #16577. Demotes 21 additional noisy loggers across GMS, MAE, and MCE consumers, adds a slow-request threshold to the REST access log, fixes auth logging architecture, and aligns code-level log levels with existing OTel pipeline drop filters.

**Note:** Three files modified in the fork (`EntityChangeEventSinkManager`, `AwsEventBridgeChangeEventSink`, `EntityChangeNotificationGenerator`) don't exist in OSS and were dropped during cherry-pick.

---

### MCE: BatchMetadataChangeProposalsProcessor

4 INFO lines per Kafka consumer batch cycle. Batch failures remain at ERROR. Throughput tracked via OTel spans and Dropwizard counters.

### MAE: UpdateIndicesHook + UpdateIndicesV2Strategy

**UpdateIndicesHook** logs per-batch entry. **UpdateIndicesV2Strategy** `shouldWriteToSemanticIndex()` logs at INFO for every entity in every MCL event, even when semantic search is disabled — producing millions of identical "SKIP" lines/day. One-time startup config log kept at INFO.

### MAE: SiblingAssociationHook

Per-event entry log demoted. Mutation log `"Associating {} and {} as siblings"` **kept at INFO**.

### MAE: DataHubUsageEventsProcessor + DataHubUsageEventTransformer

Per-event entry log and "Invalid event type" for non-matching events demoted.

### GMS: AspectResource + EntityResource

Per-request CRUD entry logs demoted. `RestliLoggingFilter` access log provides request visibility. Deletion audit logs **kept at INFO**.

### GMS: StructuredPropertiesValidator

Full structured property payload dump at INFO on every validation call — demoted to DEBUG.

### GMS: UsageServiceUtil

Two per-query INFO lines demoted.

### GMS: EbeanRetentionService

Per-deletion confirmation demoted. Batch-level summary logs **kept at INFO**.

### GMS: RestliLoggingFilter — slow-request threshold

Same pattern as `GraphQLController` in #16577:
- **INFO** only for slow requests (≥1000ms): `"Slow request: POST /entities?action=search - 200 - 1523ms"`
- **DEBUG** for all other requests
- Errors still logged at ERROR via `onError()`

### Auth: Move WARN to the enforcement point

**Root cause:** `AuthenticatorChain` logged WARN on every unauthenticated request — the vast majority being K8s health probes with no auth header.

**Fix:**
- `AuthenticatorChain`: WARN → **DEBUG** (utility that doesn't know if auth is required)
- `AuthenticationEnforcementFilter`: DEBUG → **WARN** at the actual 401 rejection point with path + source IP for forensic value. Only fires for actually rejected requests.

---

## Commit 2: OTel pipeline alignment

Cross-referenced the OTel agent config which already drops these log patterns at the pipeline level. Demoting at the source eliminates lines before they reach the OTel collector, reducing both ingestion volume and collector CPU.

### EntityServiceImpl

Selectively demote only per-batch/per-event noise:
- `"Ingesting aspects batch to database"` → DEBUG (per-batch with content dump)
- `"Generated N MCP SideEffects"` → DEBUG (per-batch count)
- `"Skipped producing MCL for no-op"` → DEBUG (per-event no-op)
- `"Ingesting entity urn"` → DEBUG (per-entity entry)
- **KEPT at INFO:** `"Producing MCL for ingested aspect"` — mutation audit trail

### PlatformEventProcessor

Same pattern as `AbstractKafkaListener` hooks demoted in #16577:
- Per-event entry, hook invocation, and completion logs → DEBUG
- **KEPT at INFO:** startup logs

### BulkListener + Es8BulkListener

- `"Successfully fed bulk request"` → DEBUG
- **KEPT at ERROR:** `"Failed to feed bulk request"`

---

## Risks

| Change | Risk | Mitigation |
|--------|------|------------|
| RestliLoggingFilter threshold | Lose per-request access log at INFO for fast requests | Slow requests (≥1s) still at INFO; errors still at ERROR; metrics capture all |
| EntityServiceImpl selective demotion | Lose per-batch ingestion detail | MCL production log kept at INFO — the mutation audit trail |
| PlatformEventProcessor | Lose per-event PE hook visibility | Same pattern as #16577 hooks; errors at ERROR; metrics track counts |
| AuthenticatorChain WARN→DEBUG | Auth chain failures invisible at WARN | WARN moved to enforcement filter with path+IP — only fires for actually rejected requests |

All other changes are pure log-level demotions with zero behavioral impact.

## Test plan

- [x] `./gradlew :metadata-jobs:mce-consumer-job:test` — all passing
- [x] `./gradlew :metadata-jobs:mae-consumer-job:test` — all passing
- [x] `./gradlew :metadata-service:restli-servlet-impl:test` — all passing
- [x] `./gradlew :metadata-service:auth-impl:test` — all passing
- [x] `./gradlew :metadata-service:auth-filter:test` — all passing
- [x] `./gradlew :metadata-jobs:pe-consumer:compileJava` — clean
- [x] Spotless — clean across all modules

### DIFF
--- a/metadata-service/factories/src/test/java/com/linkedin/gms/factory/s3/S3UtilFactoryTest.java
+++ b/metadata-service/factories/src/test/java/com/linkedin/gms/factory/s3/S3UtilFactoryTest.java
@@ -56,13 +56,9 @@ class TestStsClientConfiguration {
  * outside AWS: aws.region + dummy AWS_ENDPOINT_URL (system property) avoid real AWS calls.
  */
 @SpringBootTest(classes = {S3UtilFactory.class, TestStsClientConfiguration.class})
+@ContextConfiguration(initializers = SetAwsRegionAndEndpointInitializer.class)
 @TestPropertySource(properties = {"datahub.s3.roleArn="})
 public class S3UtilFactoryTest extends AbstractTestNGSpringContextTests {
-
-  static {
-    System.setProperty("aws.region", "us-east-1");
-    System.setProperty("AWS_ENDPOINT_URL", "http://localhost:9999");
-  }
 
   @Autowired
   @Qualifier("s3Util")
@@ -86,12 +82,9 @@ public class S3UtilFactoryTest extends AbstractTestNGSpringContextTests {
  * or outside AWS.
  */
 @SpringBootTest(classes = {S3UtilFactory.class, TestStsClientConfiguration.class})
+@ContextConfiguration(initializers = SetAwsRegionInitializer.class)
 @TestPropertySource(properties = {"datahub.s3.roleArn=arn:aws:iam::123456789012:role/test-role"})
 class S3UtilFactoryWithRoleArnTest extends AbstractTestNGSpringContextTests {
-
-  static {
-    System.setProperty("aws.region", "us-east-1");
-  }
 
   @Autowired
   @Qualifier("s3Util")
@@ -119,6 +112,27 @@ class S3UtilFactoryEndpointOnlyTest extends AbstractTestNGSpringContextTests {
     assertNotNull(
         s3Util,
         "S3Util bean should be created when only AWS_ENDPOINT_URL set (region defaulted to US_EAST_1)");
+  }
+}
+
+/**
+ * Sets aws.region before context loads, immune to test ordering with ClearAwsPropertiesInitializer.
+ */
+final class SetAwsRegionInitializer
+    implements ApplicationContextInitializer<ConfigurableApplicationContext> {
+  @Override
+  public void initialize(ConfigurableApplicationContext applicationContext) {
+    System.setProperty("aws.region", "us-east-1");
+  }
+}
+
+/** Sets aws.region + AWS_ENDPOINT_URL before context loads, immune to test ordering. */
+final class SetAwsRegionAndEndpointInitializer
+    implements ApplicationContextInitializer<ConfigurableApplicationContext> {
+  @Override
+  public void initialize(ConfigurableApplicationContext applicationContext) {
+    System.setProperty("aws.region", "us-east-1");
+    System.setProperty("AWS_ENDPOINT_URL", "http://localhost:9999");
   }
 }
 


metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/client/shim/impl/v8/Es8BulkListener.java
```diff
@@ -47,7 +47,7 @@ public void afterBulk(
               + " Message: "
               + response);
     } else {
-      log.info(
+      log.debug(
```
RyanHolstien: I'd say this is one of our rather useful messages from my perspective. Debug demotion means we will effectively never see it.

metadata-jobs/mae-consumer/src/main/java/com/linkedin/metadata/kafka/hook/UpdateIndicesHook.java
```diff
@@ -85,7 +85,7 @@ public void invokeBatch(@Nonnull final Collection<MetadataChangeLog> events) {
         events.stream().filter(this::shouldProcessEvent).collect(Collectors.toList());
 
     if (!eventsToProcess.isEmpty()) {
-      log.info(
+      log.debug(
           "Processing batch of {} MCL events with UpdateIndicesService", eventsToProcess.size());
```
RyanHolstien: Same here, this gives us perspective on how large the batch sizes coming through are at a particular time.

metadata-jobs/mae-consumer/src/main/java/com/linkedin/metadata/kafka/transformer/DataHubUsageEventTransformer.java
```diff
@@ -70,7 +70,7 @@ public Optional<TransformedDocument> transformDataHubUsageEvent(String dataHubUs
     }
     DataHubUsageEventType eventType = DataHubUsageEventType.getType(usageEvent.get(TYPE).asText());
     if (eventType == null) {
-      log.info("Invalid event type: {}", usageEvent.get(TYPE).asText());
+      log.debug("Invalid event type: {}", usageEvent.get(TYPE).asText());
```
RyanHolstien: This is basically an error, not sure it should be demoted

metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/client/shim/impl/v8/Es8BulkListener.java
max-datahub: Reverted

metadata-jobs/mae-consumer/src/main/java/com/linkedin/metadata/kafka/hook/UpdateIndicesHook.java
max-datahub: Does https://github.com/datahub-project/datahub/pull/16578/changes/1d74b4317566cfb53fae788b25715b77179635a8 address your concerns?

metadata-jobs/mae-consumer/src/main/java/com/linkedin/metadata/kafka/transformer/DataHubUsageEventTransformer.java
max-datahub: Rephrased the log messages to highlight that it appears to be a normal filtering, rather than an error condition

RyanHolstien: The conversion above this is just to a generic json doc, this is where it's actually validating that the event type is supported. Does this generate a ton of logs in practice? I would view that as a problematic state if it does, i.e. we have people generating lots of invalid events that are silently dropped.

metadata-jobs/mae-consumer/src/main/java/com/linkedin/metadata/kafka/hook/UpdateIndicesHook.java
max-datahub: the log impact of this message is in the millions/day

metadata-jobs/mae-consumer/src/main/java/com/linkedin/metadata/kafka/transformer/DataHubUsageEventTransformer.java
max-datahub: the log impact of this message is in the 10s of thousands per day, so I think we can keep it in the original state. I will revert the changes 👍

PR #16579, opened 2026-03-13
fix(bigquery): pass explicit credentials to GCP clients for thread safety

jayacryl:
## Summary
- **Thread-safe credential handling**: Build `google.oauth2.service_account.Credentials` from the credential config and pass explicitly to all GCP client constructors (`bigquery.Client`, `ProjectsClient`, `PolicyTagManagerClient`, `GCPLoggingClient`), instead of relying on the process-global `GOOGLE_APPLICATION_CREDENTIALS` env var which races under concurrent BigQuery connections in `ThreadPoolExecutor`.
- **Fail-fast on invalid credentials**: Remove silent fallback to env-based auth when `from_service_account_info()` fails, since `None` credentials cause GCP clients to call `google.auth.default()` which reads the racy env var.
- **Test coverage**: Add test for explicit credentials construction path using mocked `from_service_account_info`.

## Motivation
When multiple BigQuery connections with different service accounts are used concurrently, threads race to overwrite `GOOGLE_APPLICATION_CREDENTIALS`, causing clients to authenticate with the wrong credentials and get intermittent 403 Forbidden errors.

## Test plan
- [x] Unit test for explicit credentials construction
- [x] Existing BigQuery tests updated to mock `from_service_account_info`

WRT other platforms carrying the same risk, we did a scan and found no impact:
```
Platform scan results — BigQuery is the only one at risk:

Databricks: Uses explicit access_token passed directly to sql.connect(). No env vars or global state. Thread-safe.
Redshift: Uses explicit user/password passed to redshift_connector.connect(). No env vars or global state. Thread-safe.
Snowflake: Uses explicit connection parameters (password, token, account, etc.). No env vars or global state. Thread-safe.
BigQuery: The only platform that sets os.environ["GOOGLE_APPLICATION_CREDENTIALS"] and relied on implicit google.auth.default() resolution. This PR fixes it.
```

### DIFF
--- a/metadata-ingestion/tests/integration/bigquery_v2/test_bigquery.py
+++ b/metadata-ingestion/tests/integration/bigquery_v2/test_bigquery.py
@@ -42,6 +42,15 @@ from tests.test_helpers.state_helpers import run_and_get_pipeline
 FROZEN_TIME = "2022-02-03 07:00:00"
 
 
+@pytest.fixture(autouse=True)
+def mock_service_account_credentials():
+    """Mock from_service_account_info so tests with dummy private keys don't fail PEM validation."""
+    with patch(
+        "datahub.ingestion.source.bigquery_v2.bigquery_connection.service_account.Credentials.from_service_account_info"
+    ):
+        yield
+
+
 def random_email():
     return (
         "".join(
--- a/metadata-ingestion/tests/integration/fivetran/test_fivetran.py
+++ b/metadata-ingestion/tests/integration/fivetran/test_fivetran.py
@@ -25,6 +25,16 @@ from datahub.testing import mce_helpers
 
 FROZEN_TIME = "2022-06-07 17:00:00"
 
+
+@pytest.fixture(autouse=True)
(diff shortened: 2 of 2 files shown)


jayacryl:
@cursor fix CI

jayacryl:
Our commit 9e9f1d1132 removed the try/except around from_service_account_info() in BigQueryConnectionConfig.__init__ so invalid credentials fail fast instead of silently
  falling back to env-based auth (which reintroduced the thread-safety bug). The unit tests were updated to mock this call, but the integration tests and fivetran tests were not — they still
  pass dummy "private_key": "private_key" values that fail PEM parsing.

  Fix: Added an autouse pytest fixture in both test_bigquery.py and test_fivetran.py that mocks from_service_account_info, matching the pattern already used in the unit tests.

rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `a072f46`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/23070618370)

PR #16591, opened 2026-03-15
Feature/health downstream services

relaxedboi:


relaxedboi:
> Overall I agree with considering backend systems when determining GMS health in some cases; however, there are cases where this is not the desired configuration.
> 
> Consider the scenario where we are able to accept async ingestion traffic because Kafka is online but other systems are not. If this approach is enabled, it would reject traffic that would otherwise succeed. We may also want GMS to host or display system status information at some point, so the GMS pod should still receive traffic while being aware of its capabilities.
> 
> For those reasons we need this to be flexible and configurable so users can decide how to handle GMS's degraded state.
> 
> There is existing logic in the HealthCheckController that already partially considers Elasticsearch. I suggest extending that existing framework.
> 
> The general concept there is that each subsystem can be checked individually and one or more subsystems can be used for the overall health. Importantly, it uses a cache (`memoizeWithExpiration`) that avoids calling the storage health too frequently, especially with high numbers of GMS instances.
> 
> All backend systems should be optional; in this implementation it seems that Neo4j is not optional.
> 
> Health check endpoints are expected to bypass authentication. That behavior is already present in the existing framework but is missing in this implementation.
> 
> In addition to backend storage, GMS itself must be reporting healthy; this implementation does not wait for GMS to reach the expected state after bootstrapping.

My intention with this API is slightly different from the existing health check framework in HealthCheckController. This endpoint is primarily designed for external alerting and monitoring outside the DataHub architecture, rather than controlling traffic routing or readiness within GMS itself.

PR #16599, opened 2026-03-16
fix(bigquery): use SchemaResolverProvider directly to avoid module-level caching in _init_schema_resolver

jayacryl:
The previous change in #16499 introduced a module-level lru_cache for
provide_schema_resolver, which could cause subtle issues in connector test
environments. Using SchemaResolverProvider directly ensures a fresh bulk
fetch per source instance, matching the original behavior of
initialize_schema_resolver_from_datahub.

Additionally, improve the error fallback to use lazy-loading
(graph=self.ctx.graph) instead of an empty resolver without any
schema source, and include platform_instance in the fallback resolver
for correctness.

https://claude.ai/code/session_01LbGYSUcdihjouXmV74wNmC


rajatoss:
## Connector Tests Results

**Connector tests failed** for commit `964a371`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/23125775125)

PR #16601, opened 2026-03-16
fix(delete-refs): fix cascade stall when deleting entities referenced by >5000 entities

max-datahub:
## Summary

`DeleteEntityService.deleteReferencesTo()` silently stops cleaning up references after the first batch when a deleted entity has more than ~5,000 incoming graph relationships. This leaves entities with permanent orphaned references (ghost tags, ghost domains, etc.) with zero errors logged.

**Discovered by**: `datahub-stress` S1 tag deletion scenario at 10K scale
**Reproduction**: Delete a tag applied to 8,002 datasets → only 5,016 cleaned → **2,986 orphaned (37%)**

### Three bugs fixed

| Bug | Severity | What | Why it was never caught |
|-----|----------|------|------------------------|
| Pagination loop exits early | Critical | Loop uses total-based termination against eventually-consistent graph index — exits after first batch | Invisible below 5,000 entities (the default `GRAPH_SERVICE_LIMIT_RESULTS_API_DEFAULT`) |
| NPE on concurrent entity deletion | High | `getAspectsReferringTo()` doesn't null-check `getEntityV2()` return | Only triggers when entities are deleted during the cascade (race condition) |
| `handleError()` is a NO-OP | High | All 5 callers pass structured error context that is silently discarded | Private method with zero observable side effects; callers already `log.error()` before calling it, so individual errors appear in logs but the centralized context (reason enum, entity URNs) is lost |

### Blast radius

**19 callers** trigger `deleteReferencesTo()` — all affected when the deleted entity has >5,000 incoming relationships:
- 11 GraphQL resolvers (tag, domain, glossaryTerm, form, structuredProperty, businessAttribute, corpuser, corpgroup, assertion, monitor, application)
- 7 service-layer callers (ownershipType, view, dataProduct, incident, assertion, pageTemplate, pageModule)
- 1 connection service caller

### The fix

**Pagination**: Replaced the broken `for` loop with `while (!entities.isEmpty())`. Offset stays at 0 (correct — deleted edges shift the result window between iterations). Added max-iterations circuit breaker to prevent infinite loops if graph index updates stall.

**NPE**: Added null check on `entityResponse` in `getAspectsReferringTo()`.

**handleError**: Replaced `// NO-OP for now.` with `log.warn("deleteReference error: reason={}, context={}", ...)`.

## Changes

| File | Change |
|------|--------|
| `DeleteEntityService.java` | Fix pagination loop, add null check, implement handleError logging |
| `DeleteEntityServiceTest.java` | Add test for multi-batch pagination (simulates 6 entities in 2 batches) |

## Test plan

- [x] New test: `testDeleteReferencesToProcessesAllBatchesWhenTotalExceedsBatchSize` — verifies all batches are processed when total exceeds batch size
- [x] All 9 existing `DeleteEntityServiceTest` tests pass
- [x] Verified at 10K scale on EKS cluster using `datahub-stress` S1 scenario (before fix: 37% orphaned; to be re-verified after deployment)

## Full analysis

See `datahub-stress/phases/analyze/results/2026-03-15-finding-delete-references-pagination-bug.md` for the complete bug report including cascade timeline, Kafka topic analysis, MAE consumer logs, and code trace.

### DIFF
--- a/metadata-service/services/src/main/java/com/linkedin/metadata/entity/DeleteEntityService.java
+++ b/metadata-service/services/src/main/java/com/linkedin/metadata/entity/DeleteEntityService.java
@@ -20,9 +20,10 @@ import com.linkedin.file.BucketStorageLocation;
 import com.linkedin.file.DataHubFileInfo;
 import com.linkedin.form.FormInfo;
 import com.linkedin.metadata.Constants;
+import com.linkedin.metadata.aspect.models.graph.Edge;
+import com.linkedin.metadata.aspect.models.graph.RelatedEntitiesScrollResult;
 import com.linkedin.metadata.aspect.models.graph.RelatedEntity;
 import com.linkedin.metadata.graph.GraphService;
-import com.linkedin.metadata.graph.RelatedEntitiesResult;
 import com.linkedin.metadata.models.AspectSpec;
 import com.linkedin.metadata.models.EntitySpec;
 import com.linkedin.metadata.models.RelationshipFieldSpec;
@@ -46,7 +47,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -66,8 +66,8 @@ public class DeleteEntityService {
   private final EntitySearchService _searchService;
   private final S3Util _s3Util;
 
-  private static final Integer ELASTIC_BATCH_DELETE_SLEEP_SEC = 5;
   private static final Integer BATCH_SIZE = 1000;
+  private static final String SCROLL_KEEP_ALIVE = "5m";
 
   /**
    * Public endpoint that deletes references to a given urn across DataHub's metadata graph. This is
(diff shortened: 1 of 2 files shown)


metadata-service/services/src/main/java/com/linkedin/metadata/entity/DeleteEntityService.java
```diff
@@ -123,26 +123,42 @@ public DeleteReferencesResponse deleteReferencesTo(
       return result;
     }
 
-    for (int processedEntities = 0;
-        processedEntities < relatedEntities.getTotal();
-        processedEntities += relatedEntities.getCount()) {
-      log.info("Processing batch {} of {} aspects", processedEntities, relatedEntities.getTotal());
+    // Re-fetch from offset 0 until no related entities remain. Offset stays at 0 because
```
david-leifker: This should be using a scroll with either point in time (PIT) or search_after for reliable pagination through results. Offset based pagination doesn't work as seen here.

max-datahub: Pushed https://github.com/datahub-project/datahub/pull/16601/changes/9a354b260c4375144bb5b54fdf67f695f4b6dd34

Summary:
> Replaced offset-based `findRelatedEntities` with scrollRelatedEntities (PIT + search_after), matching the pattern used in `BusinessAttributeUpdateHookService` and `DataProductUnsetSideEffect`. Also removed the sleep between batches since the PIT snapshot eliminates the need to wait for eventual consistency.

david-leifker (approved):

PR #16612, opened 2026-03-16
fix(docs): don't include hidden files in docs

ligfx:
Fixes broken gh-pages workflow which is including `.agent-skills` and `.claude` directories introduced by https://github.com/datahub-project/datahub/pull/16604, which then don't get copied from docs to genDocs and make rmdir fail.


ligfx:
Fixed by #16614

PR #16613, opened 2026-03-16
refactor(ui): migrate Policies page from antd to Alchemy components CAT-1547

annadoesdesign:
Before:
Screenshot 2026-03-16 at 12 47 12 PM

After:
Screenshot 2026-03-16 at 12 46 15 PM
Screenshot 2026-03-16 at 12 46 28 PM
Screenshot 2026-03-16 at 12 46 25 PM
Screenshot 2026-03-16 at 12 46 20 PM

## Summary
- Replace all Ant Design components on the Manage Permissions > Policies page with Alchemy equivalents (Table, Button, Pagination, SearchBar, SimpleSelect, Pill, Menu, Avatar, AvatarStackWithHover, Tooltip, Text, PageTitle, toast)
- Swap antd icons for Phosphor icons and consolidate row actions (Edit, Activate/Deactivate, Delete) into a three-dot overflow Menu
- Move "Create new policy" button to the page header to match Users & Groups layout, use semantic theme color tokens throughout, and replace antd `message` API with Alchemy `toast`
- Update all related Cypress e2e tests (v1, v2, ingestion privilege) to match new Alchemy component selectors and menu interaction patterns

## Checklist
- [x] The PR conforms to DataHub's [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (specifically the [PR title format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#pr-title-format))
- [x] Links to related issue (if applicable)
- [x] Tests updated for Cypress e2e (v1 `manage_policies.js`, v2 `v2_manage_policies.js`, `manage_ingestion_secret_privilege.js`)
- [x] No breaking changes to public APIs

## Test plan
- [x] `yarn prettier --write src` — passes
- [x] `yarn lint` (ESLint + Prettier check + TypeScript type check) — passes
- [x] Cypress test selectors verified against Alchemy component source (`SearchBar`, `SimpleSelect`, `Menu`, `Table`)
- [ ] Cypress e2e tests pass against running DataHub instance
- [ ] Visual verification of Policies page in browser

### DIFF
--- a/smoke-test/tests/cypress/cypress/e2e/settingsV2/v2_manage_policies.js
+++ b/smoke-test/tests/cypress/cypress/e2e/settingsV2/v2_manage_policies.js
@@ -5,28 +5,40 @@ const metadata_policy_name = `Metadata test policy ${test_id}`;
 const metadata_policy_edited = `Metadata test policy ${test_id} EDITED`;
 
 function searchForPolicy(policyName) {
-  cy.get('[data-testid="search-bar-input"]').should("be.visible");
-  cy.get('[data-testid="search-bar-input"]').clear().type(policyName);
+  cy.get('[data-testid="search-bar-input"]').should("be.visible").clear();
+  cy.get('[data-testid="search-bar-input"]').type(policyName);
   cy.wait(500);
 }
 
 function openRowMenu(policyName) {
-  cy.contains("tr", policyName).find("button").last().click({ force: true });
-  cy.wait(300);
+  cy.contains("tr", policyName)
+    .find("button")
+    .last()
+    .should("not.be.disabled")
+    .click();
+  cy.get('[data-testid^="menu-item-"]').should("be.visible");
 }
 
 function clickMenuAction(actionText) {
-  cy.get('[data-testid^="menu-item-"]').contains(actionText).click();
+  cy.get('[data-testid^="menu-item-"]')
+    .contains(actionText)
+    .should("be.visible")
+    .click();
 }
 
 function clickFocusAndType(Id, text) {
-  cy.get(`[data-testid="${Id}"]`).should("be.visible").clear().type(text);
+  cy.get(`[data-testid="${Id}"]`).should("be.visible").click();
+  cy.get(`[data-testid="${Id}"]`).clear().type(text);
 }
 
 function updateAndSave(Id, groupName, text) {
   cy.clickOptionWithTestId(Id).type(groupName);
   cy.get(`[title='${text}']`).click();
-  cy.focused().blur();
+  cy.document().then((doc) => {
+    if (doc.activeElement && doc.activeElement !== doc.body) {
+      doc.activeElement.blur();
+    }
+  });
 }
 
 function clickOnButton(saveButton) {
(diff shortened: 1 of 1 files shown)


jjoyce0510 (commented):
## Issues Found

### 🔴 Critical: `toast` is not exported from `@components`

The PR imports `toast` from `@components`:

```ts
import { ..., toast } from '@components';
```

and uses `toast.success(...)` / `toast.error(...)` in both `ManagePolicies.tsx` and `usePolicy.ts`. However, **there is no `toast` export in the Alchemy component library** (`datahub-web-react/src/alchemy-components/index.ts`). This will cause a build failure.

The existing codebase pattern uses either:
- `message.success(...)` / `message.error(...)` from `antd` (what this PR is trying to replace)
- `showToastMessage(ToastType.SUCCESS, 'message', 3)` from `@app/sharedV2/toastMessageUtils`

**Recommendation:** Replace `toast.success(...)` / `toast.error(...)` with the existing `showToastMessage` utility, or add a proper `toast` export to the Alchemy library first.

---

### 🟡 Medium: `useEffect` with no dependency array runs every render

```ts
useEffect(() => {
    onRegisterCreatePolicy?.(onClickNewPolicy);
}); // intentionally no deps — keeps registration in sync
```

While the comment says "intentionally no deps," this means `onRegisterCreatePolicy` is called on **every single render** of `ManagePolicies`. Since `onClickNewPolicy` closes over state (`focusPolicyUrn`, `focusPolicy`), it changes every render anyway, so this technically works — but it's unusual and creates unnecessary function calls. A more idiomatic approach would be to use `useCallback` for `onClickNewPolicy` and include it in the deps array, or use a stable ref pattern.

---

### 🟡 Medium: `antd Modal` still remains in `usePolicy.ts`

The PR replaces `message` with `toast` in `usePolicy.ts` but leaves `Modal.confirm()` from `antd` unchanged (line 2 still imports `Modal` from `antd`). This is fine functionally but means the migration is partial — the delete confirmation dialog is still antd.

---

### 🟢 Low: `entityRegistry as any` type cast

```ts
entityRegistry={entityRegistry as any}
```

`AvatarStackWithHover` expects `EntityRegistry` which is what `useEntityRegistry()` returns. The `as any` cast is unnecessary and should be removed since the types should already match.

---

### 🟢 Low: `alignment: 'right' as const` in column definition

The PR uses `alignment: 'right' as const`. The Alchemy `Column` type accepts `AlignmentOptions` which is `'left' | 'right' | 'center' | 'justify'`. Using `as const` works but `as AlignmentOptions` would be more idiomatic (and matches the existing pattern in `Table.stories.tsx`).

---

### 🟢 Low: Error toasts called during render

```ts
{policiesError && toast.error('Failed to load policies! An unexpected error occurred.')}
{updateError && toast.error('Failed to update policies. An unexpected error occurred.')}
```

Calling `toast.error()` (or any side effect) directly in JSX during render is an anti-pattern in React. If `policiesError` persists across re-renders, this will fire on every render. This should be moved into a `useEffect` that watches `policiesError` and `updateError`.

annadoesdesign:
@jjoyce0510 

1. @components resolves to src/alchemy-components which does export toast. The import is already correct.
2. fixed
3. out of scope
4.  entityRegistry as any — actually necessary. useEntityRegistry() returns EntityRegistryV1 but AvatarStackWithHover expects EntityRegistryV2. ManageRoles.tsx uses the exact same as any cast on line 217. Can't remove it without a type error.
5. fixed 
6. fixed

PR #16615, opened 2026-03-16
perf: remove momentjs

benjiaming:
Replace moment.js with day.js.

Scope: 30 files use moment, 11 use moment-timezone. API is nearly identical.

Net saving: ~318 KB gz. Risk: Low (API compatible, dayjs already in bundle, timezone plugin already used in 3 places).

### DIFF
--- a/datahub-web-react/src/alchemy-components/components/Editor/toolbar/CodeBlockToolbar.tsx
+++ b/datahub-web-react/src/alchemy-components/components/Editor/toolbar/CodeBlockToolbar.tsx
@@ -1,7 +1,8 @@
 import { findParentNodeOfType, isElementDomNode } from '@remirror/core';
 import { FloatingWrapper, useAttrs, useCommands } from '@remirror/react';
 import { Select } from 'antd';
-import React, { useEffect, useState } from 'react';
+import React from 'react';
+import { listLanguages } from 'refractor';
 import { Positioner, defaultAbsolutePosition, hasStateChanged, isPositionVisible } from 'remirror/extensions';
 import styled from 'styled-components';
 
@@ -57,12 +58,7 @@ export const CodeBlockMenu = () => {
     const commands = useCommands();
     const value = (useAttrs(true).codeBlock()?.language as string) ?? 'markup';
 
-    const [options, setOptions] = useState<{ value: string }[]>([]);
-    useEffect(() => {
-        import('refractor').then(({ listLanguages }) => {
-            setOptions(listLanguages().map((lang) => ({ value: lang })));
-        });
-    }, []);
+    const options = listLanguages().map((language) => ({ value: language }));
 
     const onChange = (language) => {
         commands.updateCodeBlock({ language });
--- a/datahub-web-react/src/app/entity/shared/tabs/Documentation/components/editor/toolbar/CodeBlockToolbar.tsx
+++ b/datahub-web-react/src/app/entity/shared/tabs/Documentation/components/editor/toolbar/CodeBlockToolbar.tsx
@@ -1,7 +1,8 @@
 import { findParentNodeOfType, isElementDomNode } from '@remirror/core';
 import { FloatingWrapper, useAttrs, useCommands } from '@remirror/react';
 import { Select } from 'antd';
-import React, { useEffect, useState } from 'react';
+import React from 'react';
+import { listLanguages } from 'refractor';
 import { Positioner, defaultAbsolutePosition, hasStateChanged, isPositionVisible } from 'remirror/extensions';
 import styled from 'styled-components';
 
@@ -57,12 +58,7 @@ const CodeBlockMenu = () => {
     const commands = useCommands();
     const value = (useAttrs(true).codeBlock()?.language as string) ?? 'markup';
 
-    const [options, setOptions] = useState<{ value: string }[]>([]);
-    useEffect(() => {
-        import('refractor').then(({ listLanguages }) => {
-            setOptions(listLanguages().map((lang) => ({ value: lang })));
-        });
-    }, []);
+    const options = listLanguages().map((language) => ({ value: language }));
 
     const onChange = (language) => {
         commands.updateCodeBlock({ language });
--- a/datahub-web-react/src/app/entityV2/shared/versioning/VersionsDrawer.tsx
+++ b/datahub-web-react/src/app/entityV2/shared/versioning/VersionsDrawer.tsx
@@ -185,7 +185,7 @@ export default function VersionsDrawer({ versionSetUrn, open }: Props) {
             ),
             menu: (
                 <StyledDropdown
-                    menu={{ items, style: { borderRadius: '12px' } }}
+                    menu={{ items, style: { borderRadius: '12px', boxShadow: '0px 0px 14px 0px rgba(0, 0, 0, 0.15)' } }}
                     trigger={['click']}
                     overlayStyle={{ borderRadius: '100px' }}
                 >


datahub-web-react/src/alchemy-components/components/CalendarChart/utils.ts
```diff
@@ -197,13 +190,16 @@ export function getMockedProps(
         {
             insertsAndUpdates: {
                 valueAccessor: (datum) => datum.inserts + datum.updates,
+                // eslint-disable-next-line rulesdir/no-hardcoded-colors
```
ani-malgari: i understand this is out of scope, but can we just change these to corresponding theme colors? If we disable next line, it'll never be caught again

datahub-web-react/.eslintrc.js
```diff
@@ -105,6 +105,26 @@ module.exports = {
         tsconfigRootDir: __dirname,
     },
     rules: {
+        'no-restricted-imports': [
```
ani-malgari, line 105: do we also want to add a rule that restricts date.js directly?

benjiaming: yes, will do, thank you!

benjiaming:
> Looks like we're also importing dayjs directly in some places and when we call it's functions, it silently depends on it's plugins having to be imported in @utils/dayjs and might throw an error if dayjs hasn't been mutated already. The patterns are also inconsistent in some places.
> 
> The ideal case is to use @utils/dayjs in all imports, and enforce a ESLint rule to import from utils. It can increase the scope of the PR, and it's okay if we want to do a separate immediate follow-up PR.

I agree! Thanks for flagging this.

datahub-web-react/src/alchemy-components/components/CalendarChart/utils.ts
AdrianMachado: Alternatively, stack a PR on top of this that removes ignores and fixes them. I also don't having these, but I don't think we should penalize Ben for trying to do a sprawling improvement

datahub-web-react/src/utils/dayjs.ts
```diff
@@ -0,0 +1,33 @@
...
+dayjs.extend(relativeTime);
+dayjs.extend(timezone);
+dayjs.extend(utc);
+dayjs.extend(weekOfYear);
+dayjs.extend(weekYear);
+dayjs.extend(weekday);
+
+// Default timezone for date filter parsing (structured properties, search filters).
+// Previously set inside render functions — moved here to avoid mutating global state on every render.
+dayjs.tz.setDefault('GMT');
```
AdrianMachado: Huh, weird choice over UTC but if it's what we had, then we can't really go back now

AdrianMachado: What's the default? Are we now opting surfaces that previously were using a different timezone into GMT now?

datahub-web-react/yarn.lock
```diff
@@ -10220,14 +10220,7 @@ mkdirp@^0.5.1:
...
 
-moment-timezone@^0.5.35:
-  version "0.5.35"
-  resolved "https://registry.npmjs.org/moment-timezone/-/moment-timezone-0.5.35.tgz"
-  integrity sha512-cY/pBOEXepQvlgli06ttCTKcIf8cD1nmNwOKQQAdHBqYApQSpAqotBMX0RJZNgMp6i0PlZuf1mFtnlyEkwyvFw==
-  dependencies:
-    moment ">= 2.9.0"
-
-"moment@>= 2.9.0", moment@^2.24.0, moment@^2.29.2, moment@^2.29.4:
+moment@^2.24.0, moment@^2.29.2:
```
ani-malgari, line 10223: do you know what this is for?

datahub-web-react/src/app/search/filters/utils.tsx
```diff
@@ -59,14 +58,15 @@ function getDataPlatformInstanceIconAndLabel(
     filterEntity: Entity | null,
     entityRegistry: EntityRegistry,
     size?: number,
+    iconColor?: string,
```
ani-malgari: why are we creating a new prop for iconColor, which was earlier a constant?

ani-malgari: and we haven't updated iconColor to the caller of this function. this i'm afraid will break or return an undesired default color at the `entityRegistry.getIcon` call.

datahub-web-react/src/alchemy-components/components/Editor/toolbar/CodeBlockToolbar.tsx
```diff
@@ -58,7 +57,12 @@ export const CodeBlockMenu = () => {
     const commands = useCommands();
     const value = (useAttrs(true).codeBlock()?.language as string) ?? 'markup';
 
-    const options = listLanguages().map((language) => ({ value: language }));
+    const [options, setOptions] = useState<{ value: string }[]>([]);
+    useEffect(() => {
+        import('refractor').then(({ listLanguages }) => {
+            setOptions(listLanguages().map((lang) => ({ value: lang })));
+        });
+    }, []);
```
AdrianMachado: What does this changer have to do with `dayjs`?

datahub-web-react/src/app/entity/shared/tabs/Documentation/components/editor/toolbar/CodeBlockToolbar.tsx
```diff
@@ -58,7 +57,12 @@ const CodeBlockMenu = () => {
     const commands = useCommands();
     const value = (useAttrs(true).codeBlock()?.language as string) ?? 'markup';
 
-    const options = listLanguages().map((language) => ({ value: language }));
+    const [options, setOptions] = useState<{ value: string }[]>([]);
+    useEffect(() => {
```
AdrianMachado: Same comment as prev: what's this got to do with dayjs?

datahub-web-react/src/app/entityV2/shared/versioning/VersionsDrawer.tsx
```diff
@@ -172,20 +172,20 @@ export default function VersionsDrawer({ versionSetUrn, open }: Props) {
...
                     ]}
                 >
-                    {moment(versionProperties?.created?.time).fromNow()}
+                    {dayjs(versionProperties?.created?.time).fromNow()}
                 </StructuredPopover>
             ),
             menu: (
                 <StyledDropdown
-                    menu={{ items, style: { borderRadius: '12px', boxShadow: '0px 0px 14px 0px rgba(0, 0, 0, 0.15)' } }}
+                    menu={{ items, style: { borderRadius: '12px' } }}
```
AdrianMachado: I assume this was just a styling tweak you through in?

datahub-web-react/src/app/ingest/source/builder/TimezoneSelect.tsx
```diff
@@ -13,7 +12,7 @@ type Props = {
 };
 
 export const TimezoneSelect = ({ value, onChange }: Props) => {
-    const timezones = moment.tz.names();
+    const timezones = (Intl as any).supportedValuesOf('timeZone') as string[];
```
AdrianMachado, line 15: why do we need to do a cast here?

datahub-web-react/src/app/ingestV2/source/builder/TimezoneSelect.tsx
```diff
@@ -13,7 +12,7 @@ type Props = {
 };
 
 export const TimezoneSelect = ({ value, onChange }: Props) => {
-    const timezones = moment.tz.names();
+    const timezones = (Intl as any).supportedValuesOf('timeZone') as string[];
```
AdrianMachado, line 15: Does `dayjs` not provide a list of timezones?

datahub-web-react/src/app/shared/time/timeUtils.tsx
```diff
@@ -206,10 +200,10 @@ export function getTimeRangeDescription(startDate: moment.Moment | null, endDate
 }
 
 export function formatDuration(durationMs: number): string {
-    const duration = moment.duration(durationMs);
-    const hours = Math.floor(duration.asHours());
-    const minutes = duration.minutes();
-    const seconds = duration.seconds();
+    const dur = dayjs.duration(durationMs);
```
AdrianMachado: nit: I don't think there's a strong reason to shorten the name here

datahub-web-react/src/utils/dayjs.ts
```diff
@@ -0,0 +1,29 @@
...
+dayjs.extend(isoWeek);
+dayjs.extend(localeData);
+dayjs.extend(localizedFormat);
+dayjs.extend(relativeTime);
+dayjs.extend(timezone);
+dayjs.extend(utc);
+dayjs.extend(weekOfYear);
+dayjs.extend(weekYear);
+dayjs.extend(weekday);
+
```
AdrianMachado, line 27: So previously we had 

```
dayjs.extend(customParseFormat);
dayjs.extend(advancedFormat);
dayjs.extend(weekday);
dayjs.extend(localeData);
dayjs.extend(weekOfYear);
dayjs.extend(weekYear);
```

And now we have a slightly longer list that features `localizedFormat` and `utc` which were used in many places in the codebase but not consistently. Can you help me understand what these extensions do? Will this change how date information is presented on pages that were already using `dayjs`?

datahub-web-react/yarn.lock
benjiaming: `moment` is still pulled in by Antd. The 360kB reduction in bundle is from dropping `moment-timezone`.

datahub-web-react/src/app/ingestV2/source/builder/TimezoneSelect.tsx
benjiaming: I don't think it does: https://day.js.org/docs/en/timezone/timezone

datahub-web-react/src/app/search/filters/utils.tsx
benjiaming: You're right. The proper fix would be to `useTheme` and get color from there in all locations but that's outside of the scope of this PR. Since the eslint rule is now disabled, I will revert to original code.

datahub-web-react/src/utils/dayjs.ts
benjiaming: > Can you help me understand what these extensions do? 

`localizedFormat`: enables locale-aware format tokens like 'LL' (e.g. "March 23, 2026"), 'LT' (time), 'LLL' etc. Without this plugin, dayjs().format('LL') would literally output the string "LL". There's one usage: `TimelineHeader.tsx`:30 uses 'LL'. This would break without the plugin.

`utc`: adds .utc() and .utc(true) methods. .utc() converts to UTC mode (shifts the time). .utc(true) keeps the same clock time but switches to UTC mode (no shift — useful for treating a timestamp as already being in UTC). Used in ~10 places across stats graphs, DeprecationIcon, timeUtils, CalendarChart. All of these would throw at runtime without the plugin.

`timezone`: adds .tz() for timezone conversion and dayjs.tz.guess() to detect the user's local timezone. Used in stats graphs and ingestV2/source/utils.ts. Would break without it.

> Will this change how dates are presented on pages already using dayjs?

No — these plugins were already being used in those files before this PR (they imported dayjs directly without going through the wrapper). This PR just centralizes them in the shared dayjs.ts wrapper, ensuring the plugins are always available consistently. The behavior is unchanged.

datahub-web-react/src/app/ingest/source/builder/TimezoneSelect.tsx
benjiaming: We use Typescript 4.8.4 and according to https://github.com/microsoft/TypeScript/issues/49231 `supportedValuesOf` landed somewhere in 5.x. I suppose we could extend Intl namespace manually, as suggested in the issue but I think `any` might be ok here.

datahub-web-react/src/utils/dayjs.ts
AdrianMachado: OK, so these are basically all additive. sgtm

PR #16618, opened 2026-03-16
feat(sqlglot): upgrade sqlglot to v30.0.3 and fix v30 API compatibility

askumar27:
## Summary

Upgrades `sqlglot[c]` from `29.0.1` → `30.0.3` to fix an import-time crash on fresh installs, removes the explicit `sqlglotc` workaround pin, and updates DataHub's sqlglot monkey-patches for full v30 API compatibility.

## Motivation / Context

`sqlglot[c]==29.0.1` had a packaging metadata bug where the `c` extra declared `sqlglotc` **without a version pin**, so pip would resolve the latest available `sqlglotc` (30.x at the time). This created a mismatched install:

- `sqlglot==29.0.1`
- `sqlglotc==30.x` ← mismatched compiled extension

The mismatch caused an import-time crash on any fresh install:
```
AttributeError: module 'sqlglot.expressions' has no attribute 'Expr'
```

Upstream packaging issue: https://github.com/tobymao/sqlglot/issues/7304

We target **v30.0.3** specifically (rather than 30.0.1) because it also fixes a Snowflake `AS :name` placeholder alias bug ([tobymao/sqlglot#7310](https://github.com/tobymao/sqlglot/pull/7310)) that previously required a dedicated DataHub workaround patch (`_patch_alias_placeholder`), which is now removed.

## Changes

**Dependencies (`setup.py` / `pyproject.toml` / `constraints.txt` / `uv.lock`)**
- Upgrade `sqlglot[c]` pin: `29.0.1` → `30.0.3`
- Remove the now-redundant explicit `sqlglotc==29.0.1` workaround pin — `sqlglotc` is properly version-coupled via the `[c]` extra in v30+
- Regenerated lock files via `./gradlew :metadata-ingestion:updateLockFile`

**`sql_parsing/_sqlglot_patch.py`**

| Patch | Change |
|-------|--------|
| `Scope.traverse` | mypyc-compiled in `sqlglot[c]` v30 — `patchy` can no longer diff-patch it. Replaced with a pure Python reimplementation monkeypatched directly onto the class (same approach as `__deepcopy__`), preserving identical cycle-detection behavior |
| `to_node` Patch 1 | Updated line offsets for v30's renumbered source; removed stale `source = scope.expression` line (now assigned upstream); updated `derived_tables` context line which gained a type annotation in v30 |
| `to_node` Patch 2 | Updated line offsets and context to match v30's renamed loop variables (`source` → `col_source` / `col_expr`) |
| `_deepcopy_wrapper` return type | `ExpressionCore` removed in v30, replaced by `Expr` as the new base class |
| `sqlglot.Expression` compat shim | `sqlglot.Expression` was removed as a top-level re-export in v30; restored via a one-line shim to avoid cascading annotation updates across the codebase |
| `_patch_alias_placeholder` | **Removed** — fixed upstream in v30.0.3 ([tobymao/sqlglot#7310](https://github.com/tobymao/sqlglot/pull/7310)) |

**`sql_parsing/sqlglot_lineage.py` / `sqlglot_utils.py` / `query_types.py` / `athena_properties_extractor.py` / `sdk/dataset.py`**
- Updated type annotations from removed `sqlglot.Expression` / `ExpressionCore` to `sqlglot.exp.Expression` / `sqlglot.exp.Expr`
- Added explicit `isinstance` guards before assignments where v30 optimizer/qualify APIs widened their return types

## Impact

**Affected components:** All ingestion sources that use sqlglot for SQL parsing / column-level lineage (BigQuery, Snowflake, Redshift, dbt, Fivetran, Looker, etc.)

**Breaking changes:** None — runtime behavior is identical

**Risk:** Low — patches preserve identical semantics; the pure Python `traverse` wrapper adds negligible overhead


metadata-ingestion/src/datahub/sql_parsing/_sqlglot_patch.py
```diff
@@ -77,27 +77,41 @@ def _patch_scope_traverse() -> None:
...
-+        seen_scopes.add(id(scope))
-+
-         result.append(scope)
-         stack.extend(
-             itertools.chain(
-""",
-    )
+    #
+    # In sqlglot[c] v30+, Scope.traverse is mypyc-compiled and cannot be patched
+    # with patchy. We reimplement it entirely in pure Python with cycle detection.
```
sgomezvillamor, line 82: I was about to mention performance regression risk

however, it seems that was considered already
> Risk: Low — the patches preserve identical runtime behavior; the pure Python traverse wrapper adds negligible overhead (~microseconds per query)

so 👌

rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `13a31fd`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/23464091091)

treff7es (approved):

PR #16619, opened 2026-03-16
feat(upgrade): add Cleanup upgrade for Helm pre-delete infrastructure teardown

pedro93:
## Summary
  - Adds a new `Cleanup` upgrade type to `datahub-upgrade` that tears down all DataHub infrastructure (ES indices, Kafka topics, SQL database)
  - Invoked via `-u Cleanup` CLI arg, with per-subsystem enable flags (`CLEANUP_KAFKA_ENABLED`, `CLEANUP_ELASTICSEARCH_ENABLED`, `CLEANUP_SQL_ENABLED`)
  - Companion Helm chart PR: https://github.com/acryldata/datahub-helm/pull/675

  ## Test plan
  - [x] `./gradlew :datahub-upgrade:compileJava` passes
  - [ ] Run `datahub-upgrade -u Cleanup` against a local docker-compose instance and verify topics, indices, and database are removed
  - [ ] Unit tests for each step with mocked clients

### DIFF
--- a/docs/how/updating-datahub.md
+++ b/docs/how/updating-datahub.md
@@ -48,6 +48,8 @@ This file documents any backwards-incompatible changes in DataHub and assists pe
 
 ### Other Notable Changes
 
+- #16619 **(Operations / Helm)** Added a `Cleanup` upgrade that tears down all DataHub-owned infrastructure resources (Elasticsearch indices, Kafka topics, SQL database and users). It is designed to run as a Helm **pre-delete hook** so that `helm uninstall` leaves no DataHub-specific state on shared infrastructure. Each component (ES, Kafka, SQL) can be disabled independently via environment variables (`CLEANUP_ELASTICSEARCH_ENABLED`, `CLEANUP_KAFKA_ENABLED`, `CLEANUP_SQL_ENABLED`; all default to `true`). Elasticsearch cleanup uses scoped `IndexConvention` patterns to enumerate only DataHub-owned indices — it does **not** issue a wildcard `DELETE /*` that would be dangerous on shared clusters without an index prefix configured.
+
 ## v1.5.0
 
 Requirements:


datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/cleanup/DeleteElasticsearchIndicesStep.java
```diff
@@ -0,0 +1,151 @@
...
+      } catch (Exception e) {
+        log.error("DeleteElasticsearchIndicesStep failed.", e);
+        return new DefaultUpgradeStepResult(id(), DataHubUpgradeState.FAILED);
+      }
+    };
+  }
+
+  /** Delete all indices matching {prefix}* */
+  private void deleteAllIndices(String prefix) {
+    String pattern = prefix + "*";
```
david-leifker: This is very dangerous. Consider the case where there is no prefix, this is true in many cases on a shared cluster this will remove ALL indices. This is a blocker and each ES service should already support clean-up so they target only their own indices.

pedro93: > each ES service should already support clean-up

Do you mean DataHub internal java services like: SearchService and LineageSearchService?

datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/cleanup/DeleteElasticsearchIndicesStep.java
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  /** Deletes usage event data streams, index templates, and ILM/ISM policies. */
+  private void deleteUsageEventResources(
+      SearchClientShim<?> client, IndexConvention convention, boolean isOpenSearch) {
+    String dataStreamName = convention.getIndexName(DATAHUB_USAGE_EVENT_INDEX);
+    String templateName = convention.getIndexName("datahub_usage_event_index_template");
+    String policyName = convention.getIndexName("datahub_usage_event_policy");
+
+    if (isOpenSearch) {
```
david-leifker, line 144: This should be implemented in a shim, the goal should be to not have to branch on opensearch/elasticsearch at this level

datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/cleanup/DeleteElasticsearchIndicesStep.java
```diff
@@ -0,0 +1,211 @@
...
+    }
+  }
+
+  /** Deletes the security role and user created by the ES setup job. */
+  private void deleteSecurityResources(
+      SearchClientShim<?> client, IndexConvention convention, boolean isOpenSearch) {
+    String roleName = convention.getIndexName("access");
+    String username = EnvironmentUtils.getString("CREATE_USER_ES_USERNAME");
+
+    if (isOpenSearch) {
```
david-leifker, line 163: This also better implemented in the shim vs having to maintain branching logic throughout the code base

PR #16620, opened 2026-03-16
refactor(ingest/powerbi): clean up MCP and workunit handling

ligfx:


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `6a42f3a`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/23442868880)

metadata-ingestion/tests/unit/test_powerbi_user_creation.py
```diff
@@ -380,90 +380,3 @@ def test_create_corp_user_should_default_to_false(self) -> None:
 
         config = OwnershipMapping()
         assert config.create_corp_user is False
-
-
-class TestStatefulIngestionBehavior:
```
alokr-dhub, line 385: Why these tests are deleted?

metadata-ingestion/src/datahub/ingestion/source/powerbi/powerbi.py
```diff
@@ -1522,13 +1468,13 @@ def report_to_datahub_work_units(
...
         for mcp in mcps:
-            yield self._to_work_unit(mcp)
+            yield mcp.as_workunit()
 
         # Handle user MCPs separately with is_primary_source=False
         # This prevents stateful ingestion from tracking/soft-deleting users
         if self.__config.ownership.create_corp_user:
             for mcp in user_mcps:
-                yield self._to_user_work_unit(mcp)
+                yield mcp.as_workunit(is_primary_source=False)
```
alokr-dhub, line 1478: I see the tests deleted are corresponding to this flag but there seems to be unit test gap for this flag. It would be better if we can add unit test for this else covering this in integration test would also work

metadata-ingestion/src/datahub/utilities/dedup_list.py
```diff
@@ -1,18 +1,21 @@
-from typing import Iterable, List, Set, TypeVar
+from typing import Any, Callable, Iterable, List, TypeVar
 
 _T = TypeVar("_T")
 
 
-def deduplicate_list(iterable: Iterable[_T]) -> List[_T]:
+def deduplicate_list(
+    iterable: Iterable[_T],
+    key: Callable[[_T], Any] = lambda x: x,
```
alokr-dhub: i think we can make the key param optional which will be more consistent with python built-ins.just a minor performance improvement
```suggestion
def deduplicate_list(
    iterable: Iterable[_T],
    key: Optional[Callable[[_T], Any]] = None,
) -> List[_T]:
    seen = set()
    result: List[_T] = []
    for item in iterable:
        k = key(item) if key is not None else item
        if k not in seen:
            seen.add(k)
            result.append(item)
    return result

```

metadata-ingestion/src/datahub/ingestion/source/powerbi/powerbi.py
ligfx: Good call, I added an integration test that tests that (1) CorpUsers are created when `create_corp_user=True` and (2) CorpUsers are _not_ deleted due to stateful ingestion.

metadata-ingestion/tests/unit/test_powerbi_user_creation.py
ligfx: Addressed in other comment.

metadata-ingestion/src/datahub/utilities/dedup_list.py
ligfx: Sure, updated.

alokr-dhub:
Approved. Let's add a PR summary for context and reference.

ligfx:
@alokr-dhub what do you want for a summary? The summary is: "Clean up MCP and workunit handling"

alokr-dhub:
> @alokr-dhub what do you want for a summary? The summary is: "Clean up MCP and workunit handling"

We can add the same in description which makes it easier for bots to refer changelogs.

PR #16622, opened 2026-03-17
feat(openapi): Dynamic config for /poll, based on application.yaml

skrydal:
Topics should be (for now) controlled via `application.yaml`. A new field is added to the list of topics pollEnabled, which controls whether topic can be polled through the OpenAPI. This contrasts old state where those topics were hardcoded in the service code. The PR also introduces a new topic for work events, as well as, a generic privilege to allow polling any topic.

### DIFF
--- a/metadata-service/events-service/src/test/java/io/datahubproject/event/ExternalEventsServiceTest.java
+++ b/metadata-service/events-service/src/test/java/io/datahubproject/event/ExternalEventsServiceTest.java
@@ -191,6 +191,27 @@ public class ExternalEventsServiceTest {
     verify(kafkaConsumer).partitionsFor("acme_myTopic");
   }
 
+  @Test
+  public void testWorkEventWithPrefixedKafkaTopicName() throws Exception {
+    // Simulates real-world scenario: displayName=WorkEvent_v1, actual Kafka
+    // topic=2391478-acme_WorkEvent_v1
+    Map<String, String> names = new HashMap<>(topicNames);
+    names.put("WorkEvent_v1", "2391478-acme_WorkEvent_v1");
+    Set<String> allowedTopics = Set.of("PlatformEvent_v1", "WorkEvent_v1");
+    ExternalEventsService svc =
+        new ExternalEventsService(allowedTopics, consumerPool, objectMapper, names, 10, 100);
+
+    when(kafkaConsumer.partitionsFor("2391478-acme_WorkEvent_v1"))
+        .thenReturn(Collections.emptyList());
+    when(objectMapper.writeValueAsString(any())).thenReturn("encodedString");
+
+    // API call uses displayName (WorkEvent_v1)
+    svc.poll("WorkEvent_v1", null, 10, 5, null);
+
+    // Kafka consumer should be called with actual topic name (2391478-acme_WorkEvent_v1)
+    verify(kafkaConsumer).partitionsFor("2391478-acme_WorkEvent_v1");
+  }
+
   @Test(expectedExceptions = UnsupportedTopicException.class)
   public void testUnmappedTopicThrows() throws Exception {
     Set<String> allowedTopics = Set.of("PlatformEvent_v1", "customTopic");


chakru-r:
A doc update may be helpful since the display name can now be modified and the events poll api needs to use the configured name  (which earlier was hardcoded).Though the defaults ensure things remain exactly same as before.

metadata-service/factories/src/main/java/com/linkedin/gms/factory/event/ExternalEventsServiceFactory.java
```diff
@@ -35,18 +33,27 @@ public class ExternalEventsServiceFactory {
...
-        topicConvention.getMetadataChangeLogVersionedTopicName());
-    topicNames.put(
-        METADATA_CHANGE_LOG_TIMESERIES_TOPIC_NAME,
-        topicConvention.getMetadataChangeLogTimeseriesTopicName());
-    return topicNames;
+    return configurationProvider.getKafka().getTopics().getTopics().values().stream()
+        .filter(t -> t.getPollEnabled())
+        .collect(
+            Collectors.toMap(
+                TopicsConfiguration.TopicConfiguration::getName,
```
chakru-r: This will make the name that the client doing the polling use the actual internal topic name and not a static name (the hardcoded constant earlier). This will break the clients + requires publishing internal topic names to users  and will be different for each deployment. We should have a `displayName` or `apiName` on how this topic is exposed to pollers - different from actual name. This can default to the earlier hardcoded name.

metadata-service/configuration/src/main/resources/application.yaml
```diff
@@ -583,18 +583,21 @@ kafka:
 
     metadataChangeLogVersioned:
       name: ${METADATA_CHANGE_LOG_VERSIONED_TOPIC_NAME:MetadataChangeLog_Versioned_v1}
+      pollEnabled: true
       partitions: ${METADATA_CHANGE_LOG_VERSIONED_TOPIC_PARTITIONS:${PARTITIONS:1}}
       replicationFactor: ${METADATA_CHANGE_LOG_VERSIONED_TOPIC_REPLICATION_FACTOR:${REPLICATION_FACTOR:1}}
 
     metadataChangeLogTimeseries:
       name: ${METADATA_CHANGE_LOG_TIMESERIES_TOPIC_NAME:MetadataChangeLog_Timeseries_v1}
```
chakru-r, line 594: Need a `displayName` here so that the API user can use a static name without having to use the actual topic name that has a prefix. This replaces the hardcoded constants we had. Without this, all clients polling need to use a deployment specific name (instead of a static MetadataChangeLog_Timeseries_v1 name that can work across deployments

chakru-r: i am wondering if `externalName` is a better name for this property -- the name that clients or API polling users reference this topic

skrydal: I have made a change to use `displayName`, now with settings:
```
WORK_EVENT_TOPIC_NAME=2391478-acme_WorkEvent_v1
```

When I try to poll topics (via the API) I get results:
* PlatformEvent_v1 -> HTTP 200
* WorkEvent_v1 -> HTTP 200
* 2391478-acme_WorkEvent_v1 -> HTTP 500 (Unsupported events topic provided)
* MetadataChangeLog_Versioned_v1 -> HTTP 200
* MetadataChangeLog_Timeseries_v1 -> HTTP 200

With topics:
```
❯ kcl topic list | awk '{print $1}'
NAME
2391478-acme_WorkEvent_v1
DataHubUpgradeHistory_v1
DataHubUsageEvent_v1
FailedMetadataChangeProposal_v1
MetadataChangeEvent_v4
MetadataChangeLog_Timeseries_v1
MetadataChangeLog_Versioned_v1
MetadataChangeProposal_v1
PlatformEvent_v1
```

PR #16623, opened 2026-03-17
docs(release): Announce V1 UI deprecation

AdrianMachado:
## Summary

Announcing the deprecation and eventual sunset of the V1 UI Theme. The goal is to reduce bloat in the codebase, and remove dead code. We have had V2 be the default UI since #12692 which is over a year ago.

- This will help AI agents more easily navigate and understand our codebase and avoid unnecessary extra work to update V1 clients (I've inspected the thinking logs and have seen wasted cycles on this). 
- Additionally, removing the v1 UI code can reduce our react bundle size significantly. 
- Lastly, this sunsetting also ensures folks are using the latest and greatest version of our UI.

### DIFF
--- a/.github/path-filters/lint-jobs-filters.yml
+++ b/.github/path-filters/lint-jobs-filters.yml
@@ -1,0 +1,31 @@
+lint-job:
+  - modified: ".github/workflows/lint-jobs.yml"
+  - ".github/path-filters/lint-job-filter.yml"
+markdown:
+  - added|modified: "**/*.md"
+github-actions:
+  - added|modified: ".github/**/*.yml"
+  - added|modified: ".github/**/*.yaml"
+workflow-files:
+  - added|modified: ".github/workflows/**"
+workflow-validation:
+  - ".github/workflows/**"
+  - ".github/scripts/validate_post_workflow_list.py"
+code-check-sources:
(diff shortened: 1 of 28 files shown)


AdrianMachado:
> Might be nice to discuss upgrading via helm as well as env variables. Otherwise LGTM

Good idea

docs/managed-datahub/release-notes/v_0_3_17.md
```diff
@@ -54,6 +54,27 @@ authentication.tokenService.salt must be set and not be empty
...
+or via helm chart in your `values.yaml`
+
+```yaml
+datahub-gms:
+  theme_v2:
+    enabled: true
+    default: true
+```
+
+Starting from our next release (v0.3.18), the V1 theme will be sunset, and associated code will be actively removed from that release onwards. This will help improve UI performance and simplify the codebase for contributors.
```
chriscollins3456: so this means that there will be no v1 UI on v0.3.18 right? meaning we can start deleting the experience now that v0.3.17 has been code freezed

AdrianMachado: Updated to reflect we are sunset from 3.17 onwards

PR #16626, opened 2026-03-17
fix(frontend): replace deprecated ANTD_GRAY with semantic theme token in sources.tsx

acrylJonny:
Uses useTheme() via a small wrapper component to access theme.colors.icon, fixing a pre-existing no-restricted-imports lint violation.


acrylJonny:
Fixed by #16680

PR #16629, opened 2026-03-17
fix(ci): add Cypress HTML report artifact for OSS PR debugging

priyabratadas-dh:
## Summary

- Adds `cypress-multi-reporters` + `mochawesome` alongside the existing `cypress-junit-reporter` so two reporters run in parallel per Cypress batch
- `mochawesome` writes **one JSON file per spec** as each spec completes (`overwrite: false`), so partial results are preserved even when the Electron process crashes mid-run
- After each Cypress batch, the workflow merges the per-spec JSONs with `mochawesome-merge` and generates a single **self-contained HTML report** using `mochawesome-report-generator`
- The HTML report is uploaded as a GitHub Actions artifact (`cypress-html-report-<strategy>-<batch>`) that **always** uploads (not just on failure), giving OSS contributors an inspectable report without Cypress Cloud access

### Problem this fixes

OSS contribution PRs don't have access to Cypress Cloud, making test failures hard to diagnose from logs alone. When the Electron process crashes, Cypress doesn't write the JUnit XML at all — so the GitHub Test Summary is incorrect (shows 0 failures or no results). Having a browsable HTML report makes it much easier to spot issues like undismissed modals or the last-passing spec before a crash.

## Test plan

- [x] Verify `cypress-html-report-cypress-<batch>` artifact appears in the Actions run (even on success)
- [x] Verify the HTML report opens correctly and shows per-spec results with pass/fail counts
- [x] Verify the JUnit XML files are still generated correctly (retry logic depends on them)
- [x] Simulate an Electron crash: confirm partial JSON files are still merged into a partial HTML report

### DIFF
--- a/smoke-test/tests/cypress/cypress/support/e2e.js
+++ b/smoke-test/tests/cypress/cypress/support/e2e.js
@@ -20,6 +20,10 @@ import "./commands";
 import "@testing-library/cypress/add-commands";
 import "cypress-real-events/support";
 
+// Attaches screenshots of failed tests to the mochawesome JSON via addContext,
+// so they get embedded when the unified HTML report is generated post-matrix.
+import "cypress-mochawesome-reporter/register";
+
 // Alternatively you can use CommonJS syntax:
 // require('./commands')
 


.github/workflows/docker-unified.yml
```diff
@@ -791,6 +791,30 @@ jobs:
...
+      - name: Generate Cypress HTML report
+        # Runs even on failure/crash: merges per-spec mochawesome JSONs written during the run
+        # and produces a single self-contained HTML file. Skipped for non-cypress batches.
+        if: always() && matrix.test_strategy == 'cypress'
+        run: |
+          REPORT_DIR="smoke-test/tests/cypress/build/mochawesome-report"
+          if compgen -G "$REPORT_DIR/*.json" > /dev/null 2>&1; then
+            npx --prefix smoke-test/tests/cypress mochawesome-merge "$REPORT_DIR/*.json" \
+              --output "$REPORT_DIR/merged.json"
+            npx --prefix smoke-test/tests/cypress marge "$REPORT_DIR/merged.json" \
```
devashish2203: Typo? Is this supposed to be marge or merge?

devashish2203: Also if merge is running inside the matrix then it's not really merging anything right? It's just creating an HTML report of the single jobs reports.

The Cypress HTML report and Upload of HTML report should probably be in a separate Job outside the matrix, if you are merging them.

.github/workflows/docker-unified.yml
```diff
@@ -864,6 +874,38 @@ jobs:
...
           key: ${{ needs.setup.outputs.yarn_cache_key }}
 
+  cypress_html_report:
+    name: Generate unified Cypress HTML report
+    runs-on: ubuntu-latest
+    needs: [smoke_test]
+    if: ${{ !cancelled() && needs.smoke_test.result != 'skipped' }}
+    steps:
+      - name: Download all Cypress mochawesome JSON artifacts
+        uses: actions/download-artifact@v4
```
devashish2203: Latest version is @v8. Let's use that.
https://github.com/actions/download-artifact/releases

priyabratadas-dh: Done

.github/workflows/docker-unified.yml
priyabratadas-dh: Fixed

devashish2203 (approved):

PR #16630, opened 2026-03-17
fix(ci): add Electron launch options to prevent OOM crashes in Cypress

priyabratadas-dh:
## Summary

- Adds `before:browser:launch` hook in `cypress/plugins/index.js` to push memory/GPU flags when running under Electron
- Flags added: `--disable-dev-shm-usage`, `--disable-gpu`, `--no-sandbox`, `--js-flags=--max-old-space-size=8192`
- Addresses intermittent Electron renderer-process crashes seen in CI (e.g. [run 22675206549](https://github.com/datahub-project/datahub/actions/runs/22675206549/job/65729928957?pr=16344#step:25:1324))

## Test plan

- [ ] Verify Cypress smoke tests complete without Electron OOM crashes in CI
- [ ] Confirm no regression in local Cypress open-mode runs

### DIFF
--- a/.github/workflows/docker-unified.yml
+++ b/.github/workflows/docker-unified.yml
@@ -721,6 +721,17 @@ jobs:
             echo "RETRY MODE: Running all tests (fallback)"
           fi
 
+          # Log free memory every 30 s in the background so we can correlate
+          # renderer crashes with the exact moment available RAM drops.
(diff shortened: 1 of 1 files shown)


benjiaming:
Have you looked into the current batch of errors yet?

PR #16637, opened 2026-03-17
fix(cli): Add gql files to the wheel build

skrydal:
Commands like:
```
datahub search "*" 
```
Will fail without those files (if `acryl-datahub` is installed from PyPi).


rajatoss:
## Connector Tests Results

**All connector tests passed** for commit `cd9930b`

[View full test logs →](https://github.com/acryldata/connector-tests/actions/runs/23211609429)

PR #16642, opened 2026-03-18
refactor(ui): migrate Manage Ownership page from antd to Alchemy components CAT-1563

annadoesdesign:
Before:
Screenshot 2026-03-18 at 11 45 53 AM
Screenshot 2026-03-18 at 11 45 55 AM
Screenshot 2026-03-18 at 11 47 02 AM
Screenshot 2026-03-18 at 11 47 10 AM
Screenshot 2026-03-18 at 11 47 05 AM

After: 
Screenshot 2026-03-17 at 5 47 19 PM
Screenshot 2026-03-17 at 5 47 17 PM
Screenshot 2026-03-17 at 5 47 12 PM
Screenshot 2026-03-17 at 5 46 44 PM

## Summary
- Migrates the Manage Ownership settings page from antd components to Alchemy design system components
- Moves the Create button to the page header (right-aligned) to match the Users & Groups page pattern
- Replaces antd Table, Dropdown/Menu, Form/Input, Pagination, notification/message with Alchemy equivalents (Table, Menu, Input, TextArea, Pagination, toast)
- Adds Alchemy EmptyState component when no ownership types exist
- Removes all styled-component wrappers on top of Alchemy components
- Deletes NameColumn.tsx and DescriptionColumn.tsx (inlined directly in table columns)
- Net reduction of ~176 lines of code

## Test plan
- [x] Cypress test updated to match new Alchemy component selectors (`v2_manage_ownership.js`)
- [ ] Verify create ownership type modal opens and saves correctly
- [ ] Verify edit ownership type modal pre-fills and updates correctly
- [ ] Verify delete ownership type works from three-dot menu
- [ ] Verify search filters ownership types
- [ ] Verify pagination works when >10 ownership types exist
- [ ] Verify empty state shows when no ownership types match search

### DIFF
--- a/datahub-web-react/src/app/entityV2/ownership/table/OwnershipTable.tsx
+++ b/datahub-web-react/src/app/entityV2/ownership/table/OwnershipTable.tsx
@@ -1,11 +1,16 @@
 import { Table } from '@components';
 import React from 'react';
+import styled from 'styled-components';
 
 import { Column } from '@components/components/Table/types';
 
 import { ActionsColumn } from '@app/entityV2/ownership/table/ActionsColumn';
 
 import { OwnershipTypeEntity } from '@types';
+
+const NameText = styled.span`
+    font-weight: 700;
+`;
 
 type Props = {
     ownershipTypes: OwnershipTypeEntity[];
@@ -21,7 +26,7 @@ export const OwnershipTable = ({ ownershipTypes, setIsOpen, setOwnershipType, re
             key: 'name',
             width: '25%',
             sorter: (a, b) => (a?.info?.name || '').localeCompare(b?.info?.name || ''),
-            render: (record) => <b>{record?.info?.name || record?.urn}</b>,
+            render: (record) => <NameText>{record?.info?.name || record?.urn}</NameText>,
         },
         {
             title: 'Description',


datahub-web-react/src/app/entityV2/ownership/table/OwnershipTable.tsx
```diff
@@ -16,47 +15,45 @@ type Props = {
...
+    const columns: Column<OwnershipTypeEntity>[] = [
         {
             title: 'Name',
-            dataIndex: 'name',
-            sorter: (a: any, b: any) => a?.info?.name?.localeCompare(b?.info?.name),
             key: 'name',
-            render: (_, record: any) => <NameColumn ownershipType={record} />,
+            width: '25%',
+            sorter: (a, b) => (a?.info?.name || '').localeCompare(b?.info?.name || ''),
+            render: (record) => <b>{record?.info?.name || record?.urn}</b>,
```
benjiaming: It's a good practice to use a class with `<b>` to add semantic meaning. Or, if this just meant to make it boldface, we should use `font-weight` instead. If we want to to specify importance, we should use `<strong>`.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/b

PR #16643, opened 2026-03-18
fix(): Remove constraint for pydantic

david-leifker:


david-leifker:
Closing for https://github.com/datahub-project/datahub/pull/16645

PR #16647, opened 2026-03-18
fix(ingestion/kafka-connect): fix JDBC sink lineage for Oracle, SQL Server, and non-standard JDBC URLs

acrylJonny:
## Overview

Fixes several issues that caused JDBC sink connector lineage to silently fail
or produce dataset URNs that couldn't resolve against ingested source assets.
Also extends `generic_connectors` to support explicit sink-direction lineage
mapping.

### JDBC URL normalization
Adds `normalize_jdbc_url` to convert non-standard JDBC URL formats that
SQLAlchemy cannot parse:
- Oracle thin/OCI: `oracle:thin:@host:port/service` → `oracle://host:port/service`
- SQL Server JDBC: `sqlserver://host;databaseName=db` → `mssql://host:port/db`
- jTDS: `jtds:sqlserver://host/db` → `mssql://host:port/db`

SQL Server parameter values may contain `@` (passwords, Azure `user@tenant`).
Oracle TNS descriptor (`@(DESCRIPTION=...)`) and LDAP (`@ldap://...`) URLs are
not converted — those connectors warn and skip lineage; use `generic_connectors`.

Previously these connectors logged `Failed to extract lineage for JDBC sink
connector` and produced no lineage at all.

### Oracle URN format
- The Oracle service name was being used as the dataset identifier, producing
  URNs that could never match ingested Oracle assets
- Oracle defaults to `schema.table` format (no database prefix); `database_name`
  is now `None` for Oracle
- Schema is taken from `schema.name` / `db.schema` if set, otherwise from
  `connection.user` / `connection.username` (lowercased to match the Oracle
  source's `normalize_name` convention, e.g. `Mps` / `MPS` → `mps`)
- If no schema can be resolved, lineage is skipped and a report warning is
  emitted instead of a bare `table` URN

### SQL Server URN format
- `mssql` was missing from `has_three_level_hierarchy`, so the schema was never
  included in target URNs, producing `database.table` instead of
  `database.dbo.table`
- Adds `mssql` to `has_three_level_hierarchy` and defaults schema to `dbo`
- **Source connectors too:** this helper is shared. MSSQL sources now take the
  three-level branch; if schema cannot be resolved the dataset is dropped with
  a warning instead of emitting a two-level URN. See `docs/how/updating-datahub.md`.

### Generic connectors: sink direction support
- `generic_connectors` previously only supported source direction
  (database → Kafka); `target_dataset` and `target_platform` were silently
  rejected by Pydantic
- Adds optional `target_dataset` and `target_platform` to
  `GenericConnectorConfig`; when set, lineage is emitted as
  `kafka topic → target dataset` instead of `source dataset → kafka topic`
- `_get_sink_connector` checks `generic_connectors` **before** class-name
  matches, so an explicit mapping wins over auto-detection (including
  `io.confluent.connect.jdbc.JdbcSinkConnector`)

This makes the following config valid for a real JDBC sink named `my-oracle-sink`:

```yaml
generic_connectors:
  - connector_name: my-oracle-sink
    source_dataset: my-topic
    source_platform: kafka
    target_dataset: mps.my_table
    target_platform: oracle
```


alokr-dhub (commented):
Thanks for tackling this — silent lineage loss is the worst kind of bug, and normalizing the URL at parse time is the right shape for the fix.

I went through the diff plus the surrounding call sites. Summary: the approach is sound, but three of the four claimed fixes have holes in the exact configurations they target, and the Oracle change introduces a *new* silent failure of the same class the PR is fixing. Details inline; the three I'd consider blocking:

1. **The `generic_connectors` sink override never fires for real JDBC sinks.** The fallback loop is placed *after* the class-name matches, so `io.confluent.connect.jdbc.JdbcSinkConnector` returns `JdbcSinkConnector` and the generic config is ignored. The example in the PR description doesn't work as written.
2. **`_SQLSERVER_RE` fails on any `@` in the parameters** — including `password=p@ss` and the standard Azure SQL `user=svc@tenant` form.
3. **Oracle sinks without `connection.user` now emit a bare `table` URN with no warning**, where the old `ValueError` path at least surfaced a report warning.

Plus some contained-but-real URN changes for users who aren't the reporter of this bug (points 4–7 inline), and a test-coverage gap: the added tests cover the two URL helpers but none of the three headline behaviors.

Verification note: I reviewed statically and checked the regexes against the real patterns with the exact expressions from the diff; I did not run the test suite. Given the changes in `has_three_level_hierarchy` and the lineage-building branch, worth confirming the existing kafka-connect golden files still pass.

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/connector_registry.py
```diff
@@ -255,6 +256,11 @@ def _get_sink_connector(
         ):
             return JdbcSinkConnector(manifest, config, report)
 
+        # Fall back to generic connector config for explicitly mapped sink connectors
```
alokr-dhub: **Blocking: this override never fires for the connectors it's meant for.**

The loop sits after the class-name matches, so for a self-hosted JDBC sink we return at the `DEBEZIUM_JDBC_SINK_CONNECTOR_CLASS` / `CONFLUENT_JDBC_SINK_CONNECTOR_CLASS` branch above and never reach here.

The PR description says "any sink connector can be explicitly overridden regardless of class name", and the headline example is `my-oracle-sink` with `target_platform: oracle`. But a real Oracle sink runs `io.confluent.connect.jdbc.JdbcSinkConnector` — so it matches above and the `generic_connectors` entry is silently ignored. The example as written in the description doesn't work.

The entire point of this escape hatch is "auto-detection produced the wrong URN, let me force it", so it has to win over auto-detection. Moving the loop to the top of `_get_sink_connector` gets the semantics the description claims.

Note `_get_source_connector` has the same trailing placement — moving both keeps them consistent, though the source side has never claimed override semantics, so that one is arguably fine as-is.

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/common.py
```diff
@@ -56,6 +56,24 @@
...
+_ORACLE_THIN_RE: Final = re.compile(
+    r"^(?:jdbc:)?oracle:(?:thin|oci):@/?"
+    r"/?(?P<host>[^:/@\s]+)"
+    r":(?P<port>\d+)"
+    r"(?:[:/](?P<db>[^@?\s]+))?"
+)
+_SQLSERVER_RE: Final = re.compile(
+    r"^(?:jdbc:)?sqlserver://(?P<host>[^:;/\s]+)"
+    r"(?::(?P<port>\d+))?"
+    r"(?:;(?P<params>[^@]*))?$"
```
alokr-dhub: **Blocking: `[^@]` breaks on realistic SQL Server URLs.**

Running the exact pattern from this diff:

```
jdbc:sqlserver://host:1433;databaseName=db;password=p@ss     -> NO MATCH
jdbc:sqlserver://host:1433;databaseName=db;user=svc@tenant   -> NO MATCH
```

Both fall through to `remove_prefix`, which hands `sqlserver://host:1433;databaseName=db;password=p@ss` to `make_url` — that cannot yield the intended `mssql://host:1433/db` under any reading. And `validate_jdbc_url` still returns `True` (the string starts with `jdbc:`), so it takes the parse path rather than emitting the warning.

`@` in passwords is common, and `user@server` is the *standard* Azure SQL form — so this is likely to hit real deployments.

The `[^@]` exclusion doesn't buy anything: the branch is already anchored on `sqlserver://` at the front and `$` at the end, so there's no ambiguity with the Oracle branch to guard against.

```suggestion
    r"(?:;(?P<params>.*))?$"
```

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
```diff
@@ -927,28 +926,37 @@ def _create_parser_from_url(
...
-        if not database_name:
-            raise ValueError(
-                f"Missing database name in JDBC URL: {jdbc_url}. "
-                f"JDBC URLs must include a database name, e.g., 'jdbc:postgresql://host:port/database_name'"
-            )
-
         # Get target platform from SQLAlchemy URL
         target_platform = get_platform_from_sqlalchemy_uri(str(url_instance))
 
+        # Oracle service names (e.g. myservice.corp.example.com) are not database
```
alokr-dhub, line 936: **Blocking: this replaces a reported error with a silent wrong URN.**

Setting `database_name = None` here also skips the `ValueError` that the `else` branch raises. If `_extract_schema_from_url` then finds nothing — no `schema.name`, and no `connection.user`, which is common when auth comes from a secret provider or an Oracle wallet — line 1374 evaluates to:

```python
target_dataset = get_dataset_name(None, table_name)   # -> "my_table"
```

A bare `my_table` URN that resolves against nothing. Previously the missing-database `ValueError` was caught by the handler at the bottom of `extract_lineages` and surfaced as a report warning, so the user at least learned something was wrong.

Could we add a `report.warning` when `target_platform == "oracle"` and no schema could be resolved? Silently emitting an unresolvable URN is the exact failure mode this PR set out to fix.

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/common.py
```diff
@@ -535,14 +555,55 @@ def unquote(
...
+    return remove_prefix(url, JDBC_PREFIX)
+
+
 def validate_jdbc_url(url: str) -> bool:
-    """Validate JDBC URL format and return whether it's well-formed."""
+    """Return whether url is a JDBC URL format we can parse."""
     if not url or not isinstance(url, str):
         return False
-
+    if _ORACLE_THIN_RE.match(url) or _SQLSERVER_RE.match(url) or _JTDS_RE.match(url):
```
alokr-dhub, line 615: **The validator was widened, but only one of its two consumers was.**

`validate_jdbc_url` now accepts URLs with no `jdbc:` prefix. `sink_connectors.py` was updated to call `normalize_jdbc_url`, but `source_connectors.py::_extract_platform_from_jdbc_url` (~line 1457) still does `remove_prefix` plus a naive split on the first `:`.

Net effect: `sqlserver://host;databaseName=db` now passes validation and yields platform `"sqlserver"`, where it previously returned `"unknown"`. `"sqlserver"` isn't a DataHub platform — `"mssql"` is.

Suggest having that function call `normalize_jdbc_url` too, or just delegate to `get_platform_from_sqlalchemy_uri`, so `normalize_jdbc_url` is genuinely the single entry point rather than a second path running alongside `remove_prefix`. Widening a validator without widening its consumers is how the next silent-lineage bug gets filed.

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
```diff
@@ -1345,17 +1360,17 @@ def extract_lineages(self) -> List[KafkaConnectLineage]:
                     parser.schema_name,
                 )
 
-                # Build fully qualified dataset name using helper function
-                if parser.schema_name and has_three_level_hierarchy(
-                    parser.target_platform
-                ):
-                    # Platform supports schema hierarchy: database.schema.table
+                if parser.schema_name:
```
alokr-dhub: **This changes URNs for two-level platforms too.**

The guard went from `if parser.schema_name and has_three_level_hierarchy(...)` to just `if parser.schema_name:`.

`_extract_schema_from_url` reads `schema.name` / `db.schema` / `currentSchema` *before* any platform check, so a MySQL sink that sets `schema.name` now produces `schema.table` where it previously produced `database.table` — a silent URN change for existing users who aren't affected by the Oracle/mssql bugs this PR is fixing.

Gating the `get_dataset_name(None, ...)` branch on `parser.target_platform == "oracle"` targets the case you actually want, instead of "any schema happens to be present".

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
```diff
@@ -1070,9 +1078,16 @@ def _extract_schema_from_url(
...
+                schema = "public"
+            elif platform == "mssql":
+                schema = "dbo"
+
+        # Oracle: connecting user = schema owner. Lowercase ALL_CAPS names to match
+        # SQLAlchemy's normalize_name convention used by the Oracle source.
+        if not schema and platform == "oracle":
+            raw = config.get("connection.user") or config.get("connection.username")
+            if raw:
+                schema = raw.lower() if raw.isupper() else raw
```
alokr-dhub: **Duplicates an existing helper, and the casing rule is wrong for this input.**

This is a verbatim copy of `normalize_db_name` in `metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py` (~line 360). Two separate issues:

**Duplication** — two copies of a convention that must stay in lockstep with the Oracle source, or lineage silently stops matching. Worth lifting into a shared util and importing in both places. (I realize importing `sql/oracle.py` directly from `kafka_connect` would drag in the Oracle extras, which is presumably why it was copied — a shared location neither module owns would solve both.)

**Correctness** — `normalize_db_name` operates on a name read *back from* Oracle, which is already uppercase; that's why the `isupper()` guard is right there. Here the input is `connection.user`, a credential a human typed into a config file. So:

- `connection.user=Mps` -> Oracle folds the unquoted identifier to owner `MPS` -> the Oracle source normalizes it to `mps` -> this returns `Mps` -> the URN never matches.

For this input a plain `.lower()` is the correct rule.

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/common.py
```diff
@@ -693,7 +754,7 @@ def transform_connector_config(
 
 # TODO: Find a more automated way to discover new platforms with 3 level naming hierarchy.
 def has_three_level_hierarchy(platform: str) -> bool:
-    return platform in ["postgres", "trino", "redshift", "snowflake"]
+    return platform in ["postgres", "trino", "redshift", "snowflake", "mssql"]
```
alokr-dhub, line 769: **Adding `mssql` here also changes the source connector path — worth calling out explicitly.**

This helper is shared. `source_connectors.py::_resolve_source_table_schema` (~line 316) now takes the three-level branch for mssql *sources*, and when the schema can't be resolved it warns and **drops the dataset** (`include_dataset=False`) where it previously emitted a working two-level URN.

SQL Server genuinely is three-level, so I think the change is correct — but it's a behavior change for existing mssql source users that isn't mentioned in the PR description. Could we get a note in the description and an entry in `docs/how/updating-datahub.md`?

Also `test_has_three_level_hierarchy` in `tests/unit/kafka_connect/test_kafka_connect.py` (~line 3590) still asserts only the original four platforms — worth extending it with `mssql`.

metadata-ingestion/tests/unit/kafka_connect/test_kafka_connect_jdbc_parser_factories.py
```diff
@@ -11,6 +15,49 @@
...
+            "jdbc:sqlserver://host:1433;databaseName=mydb;integratedSecurity=false",
+            "mssql://host:1433/mydb",
+        ),
+        ("jdbc:sqlserver://host;databaseName=mydb", "mssql://host:1433/mydb"),
+        ("jdbc:sqlserver://host:1433", "mssql://host:1433/"),
+        ("jdbc:jtds:sqlserver://host:1433/mydb", "mssql://host:1433/mydb"),
+        ("jdbc:jtds:sqlserver://host:1433", "mssql://host:1433/"),
+    ],
+)
+def test_normalize_jdbc_url(url: str, expected: str) -> None:
```
alokr-dhub, line 54: **Coverage gap: the added tests don't reach any of the headline fixes.**

These two parametrized tests cover `normalize_jdbc_url` and `validate_jdbc_url` thoroughly, which is great. But nothing exercises the three behaviors the PR is actually for:

- Oracle producing `schema.table` with `database_name=None`
- mssql producing `database.dbo.table` via the new `dbo` default
- `_GenericConnector` emitting sink-direction lineage when `target_dataset`/`target_platform` are set

Those are also the three most likely to regress, since they're branch changes in URN construction rather than pure string transforms. One parametrized test over `JdbcSinkConnector.extract_lineages` asserting the final `target_dataset` would cover the first two, and a small one over `_GenericConnector.extract_lineages` the third.

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/common.py
```diff
@@ -56,6 +56,24 @@
 CONNECTOR_CLASS: Final[str] = "connector.class"
 JDBC_PREFIX: Final[str] = "jdbc:"
 
+# Regex patterns for non-standard JDBC URL formats (see normalize_jdbc_url)
+_ORACLE_THIN_RE: Final = re.compile(
```
alokr-dhub, line 60: **Two nits on the Oracle/jTDS patterns.**

**Anchoring is inconsistent.** `_SQLSERVER_RE` ends with `$`; these two don't. The Oracle `db` group `[^@?\s]+` is greedy, so `jdbc:oracle:thin:@host:1521/svc;foo=bar` yields `db="svc;foo=bar"`. Worth anchoring all three the same way.

**Two common Oracle forms still fall through.** Confirmed non-matching:

```
jdbc:oracle:thin:@(DESCRIPTION=(ADDRESS=(PROTOCOL=TCP)(HOST=h)(PORT=1521))(CONNECT_DATA=(SERVICE_NAME=svc)))
jdbc:oracle:thin:@ldap://ldap.example.com:389/svc,cn=OracleContext
```

Both reach `make_url` and fail. Entirely reasonable to leave out of scope — but the new docs section presents Oracle as handled, so a sentence noting the descriptor and LDAP forms aren't covered would set expectations. A `report.warning` would also be friendlier than an exception here.

metadata-ingestion/docs/sources/kafka-connect/kafka-connect_post.md
```diff
@@ -155,9 +155,40 @@ transforms.RegexRouter.replacement: "events.$1"
...
+#### JDBC Sink: Matching Target Dataset URNs
+
+For lineage to resolve correctly, the dataset URNs generated by Kafka Connect ingestion must match the URNs produced when the target database was ingested. DataHub constructs URNs from a `connection.url` in the connector config; the format depends on the platform:
+
+| Platform   | Default URN format      | Schema used                                            |
+| ---------- | ----------------------- | ------------------------------------------------------ |
+| PostgreSQL | `database.schema.table` | `public` unless `currentSchema` is set in the URL      |
+| SQL Server | `database.schema.table` | `dbo` unless `schema.name` is set in connector config  |
+| MySQL      | `database.table`        | N/A (MySQL has no separate schema level)               |
+| Oracle     | `schema.table`          | `connection.user` (the connecting Oracle user = owner) |
```
alokr-dhub: Small gap in this row: `schema.name` also overrides `connection.user` for Oracle — `_extract_schema_from_url` checks `schema.name` / `db.schema` first, before it ever looks at `connection.user`.

The SQL Server row above already documents this ("`dbo` unless `schema.name` is set"), so it's just an inconsistency between the two rows. Since this table is the thing users will act on when their URNs don't match, worth making the Oracle row say the same.

acrylJonny:
Addressed the review on `e615b09`:

**Blocking**
1. `generic_connectors` is now checked at the top of `_get_sink_connector`, so it wins over `JdbcSinkConnector` class-name matching. Added a registry test for the Oracle sink example.
2. `_SQLSERVER_RE` params are `.*` so `password=p@ss` and Azure `user=svc@tenant` parse. Tests added.
3. Oracle with no resolvable schema raises and is reported as a warning; no bare `table` URN. `extract_lineages` test covers this.

**Other**
- Source JDBC platform extract now goes through `normalize_jdbc_url` + `get_platform_from_sqlalchemy_uri` (`sqlserver` → `mssql`).
- `schema.table` without database is gated on `target_platform == "oracle"`; MySQL with `schema.name` still emits `database.table`.
- `connection.user` is always `.lower()` (typed credential, not a name read back from Oracle). Not sharing `normalize_db_name` — the rules differ.
- `mssql` in `has_three_level_hierarchy` called out in the PR description and `docs/how/updating-datahub.md`.
- Oracle/jTDS regexes are `$`-anchored; TNS descriptor and LDAP forms documented as out of scope.
- Added `extract_lineages` tests for Oracle `schema.table`, mssql `database.dbo.table`, and generic sink-direction lineage.

acrylJonny:
Follow-up for the latest review notes and the two CI failures:

- **markdown_format_check**: ran `mdPrettierWrite` on `kafka-connect_post.md`.
- **testQuick** (`test_infer_mappings_handles_unknown_platform`): JDBC platform extract now falls back to the URL protocol when the mapper returns `external`, so `jdbc:unknown://…` stays `unknown` and `jdbc:db2://…` stays `db2`.
- Sink-direction `generic_connectors` lineage now always uses `kafka` as the source platform.
- Oracle `schema.name` / `db.schema` are lowercased before the URN is built.
- `generic_connectors` only overrides a sink handler when both `target_dataset` and `target_platform` are set; source-only entries fall through. Those two fields are also required together on the config model.
- Docs: Oracle `schema.table` vs `database.schema.table` note no longer points at `connect_to_platform_map` for the dataset name, and `updating-datahub.md` covers the Oracle sink URN change.

acrylJonny:
Follow-up for the latest cubic notes (docs first):

- Oracle URN table again lists `schema.name` / `db.schema`, and the schema override snippet mentions both. `connect_to_platform_map` is documented as instance-only (it does not change `schema.table` vs `database.schema.table`).
- Schema defaults live in one helper, so the managed-connector path also gets `dbo` for SQL Server.
- Explicit Oracle `schema.name` only folds ALL-UPPERCASE values; mixed-case quoted schemas are preserved. `connection.user` still lowercases.
- Empty `target_dataset` / `target_platform` are rejected. Sink-direction `generic_connectors` entries no longer attach to source connectors.
- Unrecognized JDBC subprotocols stay `unknown` except an allowlist (`db2`, `teradata`, `cassandra`).

acrylJonny:
Follow-up for the latest cubic notes and the markdown lint failure:

- **markdown_format_check**: prettier-aligned the Oracle URN table in `kafka-connect_post.md` (column padding). Local `--check` and the pre-push `mdPrettierWriteChanged` hook both pass on this commit.
- Source connectors that match a sink-direction `generic_connectors` entry now emit a report warning instead of being dropped as a silent "no handler" miss.
- Added an outcome-level registry test: generic JDBC sink mapping emits kafka topic → `target_dataset` lineage.

alokr-dhub (approved):
Approved. Please check the failing lint checks

PR #16655, opened 2026-03-18
feat(cli): add caller context to User-Agent header for operator auditability

shirshanka:
## Summary
- Wires `identify_caller()` into the REST emitter's User-Agent header so GMS can see which tool invoked the CLI (e.g. coding agents, CI systems, human terminals)
- User-Agent format changes from `DataHub-Client/1.0 (cli; datahub; 1.0.1)` to `DataHub-Client/1.0 (cli; datahub/<caller>; 1.0.1)`
- Zero Java backend changes — the existing `agentName` extraction at position 2 just starts carrying useful data
- Adds `DATAHUB_CALLER` env var as an operator override for custom tools or environment-specific labels

## Test plan
- [x] 15 unit tests passing (caller detection + User-Agent integration)
- [x] Verified caller detection works from multiple environments
- [ ] Verify backward compat: old user-agent strings without `/caller` still parse on the backend

### DIFF
--- a/.github/workflows/metadata-ingestion.yml
+++ b/.github/workflows/metadata-ingestion.yml
@@ -31,7 +31,7 @@ env:
 jobs:
   ci:
     runs-on: ${{ vars.DEPOT_PROJECT_ID != '' && 'depot-ubuntu-latest-2' || 'ubuntu-latest' }}
-    timeout-minutes: 75
+    timeout-minutes: 60
     env:
       DATAHUB_TELEMETRY_ENABLED: false
       # TODO: Enable this once the test is fixed.


metadata-ingestion/src/datahub/utilities/caller_context.py
```diff
@@ -0,0 +1,281 @@
...
+                return label
+
+    # Also check CI generically
+    ci_val = os.environ.get("CI", "").lower()
+    if ci_val in ("true", "1", "yes"):
+        return "ci"
+
+    # Tier 3: parent process environment (catches vars not inherited)
+    ppid = os.getppid()
+    if platform.system() == "Linux":
```
nwadams, line 227: do we need to handle windows case?

shirshanka:
> ## PR Review Summary
> **Risk**: Low | **Verdict**: Approve
> 
> Clean, well-structured addition. The 4-tier caller detection is thoughtful, the `lru_cache` ensures no perf penalty after first call, and the try/except wrapper guarantees `identify_caller()` never crashes the CLI. Verified that `agentName` on the backend is only used in debug logs (not metrics or conditional logic), so the UA format change is backward-compatible.
> 
> ### Issues to consider (non-blocking)
> 1. **File handle leak** — `caller_context.py:49`: `open()` without `with` in `_read_parent_env_linux`. Harmless in CPython but should be `with open(...) as f:` for correctness.

addressed

> 2. **DATAHUB_CALLER unsanitized** — The env var value flows directly into the User-Agent header. A basic `strip()` + length cap would prevent confusing errors from malformed values.

addressed

> 3. **Backend YAUAA test case** — Consider adding a test entry to `datahub_user_agents.yaml` for the new `datahub/<caller>` format (e.g. `datahub/claude-code`) to close the backward-compat item in the test plan.

addressed

> 
> ### Questions
> * Any concern about the `/` in the new AgentName value being misinterpreted by dashboards or analytics pipelines that consume `agentName`?

No concern - at least currently there is nothing that looks at it as anything other than a string.

metadata-ingestion/src/datahub/utilities/caller_context.py
shirshanka: theoretically yes - but essentially I'm relying on env var fingerprints as the default mechanism for windows. Skipping the process-hierarchy based analysis for windows because ps isn't always present.

PR #16621, opened 2026-03-17
feat(ingest/powerbi): emit browsepathsv2, put tables, tiles, and pages under respective parent

ligfx:

PR #16616, opened 2026-03-16
feat(ingest/powerbi): support 'Sql.Databases' M-Query data access function

ligfx:

PR #16656, opened 2026-03-18
fix(ingest/powerbi): remove optional whitespace in Power Query M parser rules which is already ignored (10x speedup)

ligfx:
On my machine, running `time pytest tests/integration/powerbi/test_m_parser.py -k 'not test_m_query_timeout'`:

Before the change: ~1m30s
After the change: ~9s (10x speedup)


ligfx:

- #16656 ◀
    - #16684
    - #16616
        - #16621

PR #16663, opened 2026-03-19
fix(ingestion): Prevent duplicate [CLI] source creation during executor-managed ingestion

javabrett:
## Summary

- **Fix:** Prevent `DatahubIngestionRunSummaryProvider` from creating duplicate `[CLI]` ingestion source entities when running under the embedded executor (triggered via `createIngestionExecutionRequest` GraphQL mutation)
- **Root cause:** Two parallel result-reporting paths both fire - the executor path (Path A) and the CLI run summary provider (Path B). Path B unconditionally creates a new `[CLI] {type}` source entity, causing duplicates where the managed source stays "Pending" while the CLI source shows results
- **Fix approach:** Detect executor context by checking if `ctx.run_id` starts with `urn:li:dataHubExecutionRequest:`. When detected, skip CLI source creation in `__init__` and `ExecutionRequestInput` emission in `on_start`. `on_completion` is unchanged - it correctly reports results to the execution request URN regardless

## Checklist

- [x] Links to the relevant Linear issue (ING-1973)
- [ ] Tests, if applicable
  - [x] 5 unit tests added covering executor detection, CLI source creation, and `on_start` behavior for both executor and standalone CLI paths
- [ ] Documentation, if applicable

## Test plan

- [x] Unit tests: `test_executor_detection_with_execution_request_urn`, `test_cli_source_creation_with_normal_run_id`, `test_on_start_skips_execution_request_input_under_executor`, `test_on_start_creates_execution_request_input_for_cli`
- [ ] Manual verification: trigger ingestion via GraphQL `createIngestionExecutionRequest` and confirm no `[CLI]` duplicate sources are created
- [ ] Verify standalone `datahub ingest` CLI runs still create CLI source entities as before

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/reporting/datahub_ingestion_run_summary_provider.py
+++ b/metadata-ingestion/src/datahub/ingestion/reporting/datahub_ingestion_run_summary_provider.py
@@ -19,7 +19,6 @@ from datahub.ingestion.api.pipeline_run_listener import PipelineRunListener
 from datahub.ingestion.api.sink import NoopWriteCallback, Sink
 from datahub.ingestion.run.pipeline_config import PipelineConfig
 from datahub.ingestion.sink.sink_registry import sink_registry
-from datahub.metadata._urns.urn_defs import DataHubExecutionRequestUrn
 from datahub.metadata.schema_classes import (
     DataHubIngestionSourceConfigClass,
     DataHubIngestionSourceInfoClass,
@@ -29,6 +28,7 @@ from datahub.metadata.schema_classes import (
     StructuredExecutionReportClass,
     _Aspect,
 )
+from datahub.metadata.urns import DataHubExecutionRequestUrn
 from datahub.utilities.logging_manager import get_log_buffer
 from datahub.utilities.urns.error import InvalidUrnError
 from datahub.utilities.urns.urn import Urn


metadata-ingestion/src/datahub/ingestion/reporting/datahub_ingestion_run_summary_provider.py
```diff
@@ -118,6 +118,20 @@ def __init__(self, sink: Sink, report_recipe: bool, ctx: PipelineContext) -> Non
         ingestion_source_key = self.generate_unique_key(ctx.pipeline_config)
         self.entity_name: str = self.generate_entity_name(ingestion_source_key)
 
+        # Check if we're running under an embedded executor (not standalone CLI)
+        # If run_id is already an execution request URN, we're being invoked by
+        # the executor (via GraphQL createIngestionExecutionRequest mutation).
+        # In this case, skip creating a duplicate CLI source - let the executor handle it.
+        self._is_running_under_executor = ctx.run_id.startswith(
+            "urn:li:dataHubExecutionRequest:"
+        )
```
gabe-lyons: this seems fragile. what if we change our run id pattern? this will silently stop working and will be extremely difficult to diagnose.

metadata-ingestion/src/datahub/ingestion/reporting/datahub_ingestion_run_summary_provider.py
```diff
@@ -128,25 +142,27 @@ def __init__(self, sink: Sink, report_recipe: bool, ctx: PipelineContext) -> Non
...
+                type=ctx.pipeline_config.source.type,
+                platform=make_data_platform_urn(
+                    getattr(ctx.pipeline_config.source, "platform", "unknown")
+                ),
+                config=DataHubIngestionSourceConfigClass(
+                    recipe=self._get_recipe_to_report(ctx),
+                    version=nice_version_name(),
+                    executorId=self._EXECUTOR_ID,
+                ),
+            )
```
gabe-lyons, line 173: can we just do a "set if not exists" here rather than baking in all these assumptions about run id formats?

treff7es: "set if not exists" won't always work. We need a graph to check whether it exists; it may not.  Kafka sink users have write-only access; there's no read path.

PR #16668, opened 2026-03-19
perf(graphql): use batch queries instead of single ones in resolvers

Dutt23:
Implemented N+1 query optimizations for DataHub GraphQL resolvers, achieving faster responses for EntityExistsResolver, with additional optimizations for batch tag/owner/term operations. And fixed a small bug in the way we handle batch mcp's.

## Bug fixes. 
There was a bug in how we handle batch ops in our resolver incase we try to update multiple tags for multiple fields in one call. The bug went un-notificed as the Ui design does not allow batch ops. For each field one mcp is generated. So we generate two mcp's, one in which field A's tags have been modified and one in which field B's tags have been modified. So whichever gets inserted last wins. 
eg query to test. `  mutation {
    batchAddTags(input: {
      tagUrns: [
        "urn:li:tag:gold"
        "urn:li:tag:low_queries"
        "urn:li:tag:high_queries"
      ]
      resources: [
        {
          resourceUrn: "urn:li:dataset:(urn:li:dataPlatform:hive,SampleHiveDataset,PROD)"
          subResource: "field_bar"
          subResourceType: DATASET_FIELD
        }
        {
          resourceUrn: "urn:li:dataset:(urn:li:dataPlatform:hive,SampleHiveDataset,PROD)"
          subResource: "field_foo"
          subResourceType: DATASET_FIELD
        }
        {
          resourceUrn: "urn:li:dataset:(urn:li:dataPlatform:hdfs,SampleHdfsDataset,PROD)"
          subResource: "shipment_info"
          subResourceType: DATASET_FIELD
        }
        {
          resourceUrn: "urn:li:dataset:(urn:li:dataPlatform:hdfs,SampleHdfsDataset,PROD)"
          subResource: "shipment_info.date"
          subResourceType: DATASET_FIELD
        }
        {
          resourceUrn: "urn:li:dataset:(urn:li:dataPlatform:hdfs,SampleHdfsDataset,PROD)"
          subResource: "shipment_info.target"
          subResourceType: DATASET_FIELD
        }
      ]
    })
  }`

**Key Results:**
- EntityExistsResolver: (337ms → 44ms average)
- Batch Operations: **60.7% faster** (14.6s → 5.75s average)
- Query Reduction: **99%** fewer database queries

## Test Configuration BatchAddResolver
  - **Load Profile:** 1000 concurrent users
  - **Ramp Rate:** 19 users/second
  - **Duration:** 60 seconds
  - **Query:** BatchAddTagsResolver (batch operations)

  ## Results Comparison

  ### BEFORE Optimization (Baseline - 3 Runs Average)
  | Metric | Value |
  |--------|-------|
  | **Average Response Time** | 6,253.10 ms |
  | **Run 1** | 6,337.30 ms |
  | **Run 2** | 6,208.79 ms |
  | **Run 3** | 6,213.22 ms |

  ### AFTER Optimization (Optimized - 3 Runs Average)
  | Metric | Value |
  |--------|-------|
  | **Average Response Time** | 4,470.21 ms |
  | **Run 1** | 4,449.05 ms |
  | **Run 2** | 4,320.42  ms |
  | **Run 3** | 4,641.59 ms |
  
  
  ## Test Configuration EntityExistsResolver N+1 Optimization
  - **Load Profile:** 1000 concurrent users
  - **Ramp Rate:** 19 users/second
  - **Duration:** 60 seconds
  - **Query:** EntityExistsResolver (entity existence checks)

  ## Performance Results

  | Metric | BEFORE | AFTER | Improvement |
  |--------|--------|-------|-------------|
  | **Average Response Time** | 2,078.43 ms | 44.21 ms | 87% faster  |
  | **95th Percentile** | 140 ms | 79 ms | 44% faster |
  | **Max Response Time** | 8,425.75 ms | 286 ms | 97% reduction  |
  | **Database Queries** | 14,000 | 140 | 99% reduction |
  | **Request Rate** | 186.32 req/sec | 240+ req/sec | 35%+ increase |

### DIFF
--- a/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/mutate/util/LabelUtils.java
+++ b/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/mutate/util/LabelUtils.java
@@ -269,10 +269,8 @@ public class LabelUtils {
         }
       }
     } catch (Exception e) {
-      log.error("Failed to batch-read {}, falling back to empty map", aspectName, e);
-      resources.stream()
-          .map(r -> UrnUtils.getUrn(r.getResourceUrn()))
-          .forEach(urn -> result.put(urn, defaultSupplier.get()));
+      log.error("Failed to batch-read {}, throwing error", aspectName, e);
+      throw new RuntimeException(e);
     }
     return result;
   }


datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/mutate/util/LabelUtils.java
```diff
@@ -124,6 +134,163 @@ public static void addTagsToResources(
     EntityUtils.ingestChangeProposals(opContext, changes, entityService, actor, false);
   }
 
+  /**
+   * Optimized batch version of addTagsToResources that reduces N+1 queries. Batch-reads all aspects
+   * instead of reading per-resource.
+   */
+  public static void addTagsToResourcesOptimized(
```
RyanHolstien: We should condense, rather than having a new method named "optimized" where it's unclear why we're not always using it and leaving the old method it should either:

1. Just always be used
2. If it's not truly optimized in the case of single add, then split pathways coming in through a single focal point i.e. addTagsToResources -> split path for single and batch

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/loaders/EntityExistsBatchDataLoader.java
```diff
@@ -0,0 +1,84 @@
...
+import org.dataloader.DataLoader;
+import org.dataloader.DataLoaderOptions;
+
+/**
+ * DataLoader for batching multiple entity existence checks into a single database query.
+ *
+ * <p>Converts N+1 queries (one exists() call per entity) into a single batched call. Example: 100
+ * entities requesting exists field → 1 batched query instead of 100 individual queries.
+ */
+public class EntityExistsBatchDataLoader {
```
RyanHolstien: I think we should instead be following something more similar to what we've done with LoadableType where we set up cross-entity type DataLoaders. Having this class be a factory for itself and designing its own pattern for batchLoad feels off to me.

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/mutate/util/LabelUtils.java
```diff
@@ -124,6 +134,163 @@ public static void addTagsToResources(
...
+                Collectors.partitioningBy(
+                    r -> r.getSubResource() == null || r.getSubResource().isEmpty()));
+
+    List<ResourceRefInput> entityResources = partitioned.get(true);
+    List<ResourceRefInput> subResourcesList = partitioned.get(false);
+
+    final List<MetadataChangeProposal> changes = new ArrayList<>();
+
+    Map<Urn, GlobalTags> globalTagsByUrn =
+        batchReadAspect(
```
RyanHolstien, line 146: We should instead shift to patch MCPs for this rather than continue to do read-modify-write.

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/mutate/util/LabelUtils.java
Dutt23: My bad, both methods have the same signature. Removed old one.

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/loaders/EntityExistsBatchDataLoader.java
Dutt23: Could you clarify what you mean by following the LoadableType pattern? LoadableType is for loading entity objects, but exists returns Boolean. Should I:
 1. Keep the current logic but change the registration pattern?
 2. Something else I'm missing?"

RyanHolstien: LoadableType is an interface for generating per entity DataLoaders and contained methods. What I'm proposing is that we wrap this exists functionality under a new cross-entity type interface similar in structure to LoadableType where we have utility methods that apply to multiple entities at once. We have a few data fetchers where this would be useful and setting up this pattern makes it easier to migrate others rather than setting up a bunch of different patterns for generating DataLoaders.

RyanHolstien: It could also even maybe just be a LoadableType that has a type of ALL or something like that with a new exists method (default function exists that calls the batch exist service layer). If you look at the implementing classes of LoadableType like DatasetType, there are already other utility methods included like browse etc., they are designed as a focal point for running the batch queries down to the DAO layer and less about specifically "loading" particular entities. They are intended to roll-up all of the queries into large batches, right now this happens at a per entity point to be able to target the specific aspects in each query, but for cross-entity functionality like pure existence checks it makes sense to make it coarser grained.

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/mutate/util/LabelUtils.java
Dutt23: So I did consider and had done an initial analysis using claude. Seemed like a decent amount of refactor and testing as it touches a lot of common methods. For this PR I just just considering to batch queries to gain some performance and reduce queries made to DB.
I can take this up in a follow up PR if required ?

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/mutate/util/LabelUtils.java
```diff
@@ -110,20 +120,163 @@ public static void removeTagsFromResources(
...
+            result.put(entry.getKey(), aspectClass.cast(aspectRecord));
+          } else {
+            result.put(entry.getKey(), defaultSupplier.get());
+          }
+        } else {
+          result.put(entry.getKey(), defaultSupplier.get());
+        }
+      }
+    } catch (Exception e) {
+      log.error("Failed to batch-read {}, falling back to empty map", aspectName, e);
```
RyanHolstien: This is incorrect behavior that can wind up wiping out data unintentionally. If you don't grab existing tags it will treat the add as an upsert on an empty set of tags wiping out any existing tags that are not included in the add. Most exceptions on the read should result in an API failure.

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/mutate/util/LabelUtils.java
RyanHolstien: If the goal here is to improve performance I think it makes sense to take the time to do patch updates. It fully eliminates the need for batching the entity gets since you don't need to do a read-modify-write.

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/mutate/util/LabelUtils.java
Dutt23: Fixed it.

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/mutate/util/LabelUtils.java
Dutt23: I tried using the patch update objects , `GlobalTagsPatchBuilder`, `EditableSchemaMetadataPatchBuilder`. But it became slower for some reason. On checking the mysql slows for patch I can see single statements being executed not able to trace from where though. 
eg : 
```
select urn, aspect, version, metadata, systemMetadata, createdOn, createdBy, createdFor FROM metadata_aspect_v2 WHERE (urn, aspect, version) IN (('urn:li:dataset:(urn:li:dataPlatform:hdfs,SampleHdfsDataset,PROD)', 'globalTags', 0))
select urn, aspect, version, metadata, systemMetadata, createdOn, createdBy, createdFor FROM metadata_aspect_v2 WHERE (urn, aspect, version) IN (('urn:li:dataset:(urn:li:dataPlatform:hive,SampleHiveDataset,PROD)', 'globalTags', 0))
select urn, aspect, version, metadata, systemMetadata, createdOn, createdBy, createdFor FROM metadata_aspect_v2 WHERE (urn, aspect, version) IN (('urn:li:dataset:(urn:li:dataPlatform:hive,fct_users_created,PROD)', 'globalTags', 0))
select urn, aspect, version, metadata, systemMetadata, createdOn, createdBy, createdFor FROM metadata_aspect_v2 WHERE (urn, aspect, version) IN (('urn:li:dataset:(urn:li:dataPlatform:hive,fct_users_deleted,PROD)', 'globalTags', 0))
select urn, aspect, version, metadata, systemMetadata, createdOn, createdBy, createdFor FROM metadata_aspect_v2 WHERE (urn, aspect, version) IN (('urn:li:dataset:(urn:li:dataPlatform:hive,logging_events,PROD)', 'globalTags', 0))
```

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/loaders/EntityExistsBatchDataLoader.java
Dutt23: Used the LoadableType interface, let me know me if it's fine now.

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/mutate/BatchAddOwnersResolver.java
```diff
@@ -68,27 +68,23 @@ private void validateOwners(@Nonnull OperationContext opContext, List<OwnerInput
...
-      @Nonnull OperationContext opContext, ResourceRefInput resource, QueryContext context) {
-    final Urn resourceUrn = UrnUtils.getUrn(resource.getResourceUrn());
-
-    if (resource.getSubResource() != null) {
-      throw new IllegalArgumentException(
-          "Malformed input provided: owners cannot be applied to subresources.");
+      @Nonnull OperationContext opContext, List<ResourceRefInput> resources, QueryContext context) {
+    for (ResourceRefInput resource : resources) {
+      if (resource.getSubResource() != null) {
+        throw new IllegalArgumentException(
```
gdatahub, line 78: if one of the ResourceRefInput items (resource.getSubResource() != null) fails, it appears that the processing of the remaining resources stops. Could you please confirm if this is the intended behavior?

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/mutate/util/LabelUtils.java
```diff
@@ -110,20 +120,161 @@ public static void removeTagsFromResources(
     EntityUtils.ingestChangeProposals(opContext, changes, entityService, actor, false);
   }
 
+  /**
+   * Optimized batch version of addTagsToResources that reduces N+1 queries. Batch-reads all aspects
+   * instead of reading per-resource.
+   */
   public static void addTagsToResources(
```
gdatahub, line 127: Batch reads are implemented only for addTagsToResources. Should we consider applying them to removeTagsFromResources and other entities as well, or is that not required?

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/mutate/util/LabelUtils.java
RyanHolstien: When it's doing the patch it will pull the current aspect data within the transaction to apply the patch update. It should still be faster than the original state, but the batch gets will probably be faster while risking significantly more correctness problems since it's outside of a transaction.

Dutt23: I understand your concern about race conditions, but that issue already exists in the current code that does sequential read-modify-write. This PR doesn't introduce it.
The Patch MCPS brings performance brings the performance back to the baseline , because of a lot of overhead. The purpose of this PR is to bound the queries which are made during bulk updates. 

another issue here is that Patch always assumes entries for the that aspect are in the DB, would fail on first time creates. And not sure would it work for soft hard deletes. 
Issue with Patch and array items :
EditableSchemaMetadataPatchBuilder.addTag() generates patch paths like /editableSchemaFieldInfo/field_foo/globalTags/tags/-
This assumes EditableSchemaFieldInfo for field_foo already exists with fieldPath set
On first-time inserts, the patch fails: "fieldPath field is required but not found"

If we want to address the race condition concern, that would be a separate architectural change that affects all read-modify-write operations in DataHub, not just this optimization. We could look into into in a separate issue if required.

Dutt23:
I'm closing this PR after doing some research on the transaction timing concerns.

*What I found:*
- 36+ files using this exact pattern: `OwnerUtils`, `DomainUtils`, `LinkUtils`, `DescriptionUtils`, `GlossaryTermService`, etc.

Fixing this properly requires migrating ALL mutation operations to transaction-scoped batching, which is beyond the scope of this performance improvement.

I'll document these findings separately for the team to consider as a future initiative.

PR #16683, opened 2026-03-19
feat(ingest/mssql): skip inaccessible databases when configured database=null

ligfx:


ligfx:

- #16683 ◀
    - #16735
        - #16736

metadata-ingestion/src/datahub/ingestion/source/sql/mssql/source.py
```diff
@@ -1320,7 +1320,18 @@ def get_inspectors(self) -> Iterable[Inspector]:
                     url = self.config.get_sql_alchemy_url(
                         current_db=db["name"], is_odbc=self._is_odbc
                     )
-                    engine = create_engine(url, **self.config.options)
+                    try:
+                        engine = create_engine(url, **self.config.options)
```
kyungsoo-datahub, line 1324: Based on the reference(https://docs.sqlalchemy.org/en/14/core/engines.html#sqlalchemy.create_engine), ```The [create_engine()](https://docs.sqlalchemy.org/en/14/core/engines.html#sqlalchemy.create_engine) call itself does not establish any actual DBAPI connections directly.```

And found that Inspect is doing some probe for the connection (https://github.com/sqlalchemy/sqlalchemy/blob/rel_1_4_54/lib/sqlalchemy/engine/reflection.py#L128)

Based on that, it looks like we need to do try-except for `inspector = inspect(engine)`

ligfx: Classic SQLAlchemy 😂 Good catch, fixed.

PR #16695, opened 2026-03-20
fix(ingest/tableau): prevent embedded datasource IDs from leaking into published datasource filter

treff7es:
## Summary

- **Bug**: `emit_datasource()` in the Tableau connector unconditionally added all datasource IDs to `datasource_ids_being_used` (the published datasource tracking list), even when processing embedded datasources. This caused embedded datasource IDs to pollute the filter used to query Tableau's `publishedDatasourcesConnection` API, resulting in phantom/unwanted datasources appearing during ingestion.
- **Fix**: Guard the tracking list addition with the existing `is_embedded_ds` flag so only published datasource IDs are tracked.
- **Test**: Added regression test `test_emit_datasource_does_not_add_embedded_ids_to_published_list` verifying embedded IDs don't leak into the published list while published IDs are still tracked correctly.

Fixes: CUS-8144

## Test plan

- [x] All 166 Tableau unit tests pass
- [x] Lint passes (`lintFix` clean)
- [ ] Verify with customer's Tableau ingestion setup that phantom embedded datasources no longer appear

### DIFF
--- a/metadata-ingestion/tests/unit/tableau/test_tableau_source.py
+++ b/metadata-ingestion/tests/unit/tableau/test_tableau_source.py
@@ -17,6 +17,7 @@ from datahub.ingestion.source.tableau.tableau import (
     SiteIdContentUrl,
     TableauConfig,
     TableauPageSizeConfig,
+    TableauProject,
     TableauSiteSource,
     TableauSource,
     TableauSourceReport,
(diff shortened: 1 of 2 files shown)


metadata-ingestion/tests/unit/tableau/test_tableau_source.py
```diff
@@ -1013,3 +1014,117 @@ def test_table_lineage_without_columns():
             assert len(upstream_tables) == 1
             assert len(table_id_to_urn) == 1
             assert report.num_upstream_table_processed_without_columns == 1
+
+
+@freeze_time(FROZEN_TIME)
+def test_emit_datasource_does_not_add_embedded_ids_to_published_list():
+    """Embedded datasource IDs should not leak into datasource_ids_being_used,
+    which is the tracking list for published datasources.
```
sgomezvillamor, line 1022: if `datasource_ids_being_used` is the tracking list for published datasources,
may it be renamed `published_datasource_ids_being_used`? 

looking at current lists, it would make sense

https://github.com/datahub-project/datahub/blob/0c91ecda0af9ec7e220cce5696ec06abe491d373/metadata-ingestion/src/datahub/ingestion/source/tableau/tableau.py#L1111-L1119

PR #16699, opened 2026-03-20
fix(ingest/teradata): apply convert_urns_to_lowercase in get_identifier()

JohnRTurner:
## Summary

- Teradata returns all object names in UPPERCASE from its system catalog. The `convert_urns_to_lowercase: true` config option was only applied at the final URN emission step, **after** the schema resolver had already been populated with uppercase keys.
- During view SQL parsing (Phase 2), sqlglot normalizes all identifiers to lowercase. The resulting case mismatch caused every schema resolver lookup to miss, silently preventing column-level lineage from being generated.
- Adds a `get_identifier()` override to `TeradataSource` that lowercases the `database.table` identifier when `convert_urns_to_lowercase` is enabled — consistent with how MSSQL and PostgreSQL implement the same behaviour.

## Root Cause

```
Phase 1 - Schema extraction:   get_identifier() → "C3_T.MRCH_PRFL"  → schema resolver key (UPPERCASE)
Phase 2 - View SQL parsing:    sqlglot normalizes → "c3_t.mrch_prfl" → schema resolver lookup (lowercase)
                                                                         → MISS → no column lineage
```

The fix ensures both phases use the same lowercase key when `convert_urns_to_lowercase: true`.

## Test plan

- [ ] Added `TestTeradataGetIdentifier` unit tests covering: lowercase when flag enabled, case preserved when flag disabled, and key matching sqlglot output
- [ ] All 66 existing Teradata unit tests pass
- [ ] `./gradlew :metadata-ingestion:lintFix` passes clean


JohnRTurner:
The CI failure in `test_dataset_sync_bidirectional` is unrelated to this PR. The test references `urn:li:structuredProperty:non_existent_property` and `urn:li:structuredProperty:non_existent_field_property` which don't exist in the test environment — this code already exists in master and predates this branch.

PR #16703, opened 2026-03-20
feat(ingestion): add configurable log levels for sink errors and warnings

rob-1019:
## Summary

Adds configurable log levels for ingestion errors and warnings, allowing users to see failures in real-time without enabling full debug logging.

### Environment Variables

**Source errors/warnings** (from metadata extraction):
- `DATAHUB_SOURCE_ERROR_LOG_LEVEL` (default: `DEBUG`) - Controls log level for source errors
- `DATAHUB_SOURCE_WARNING_LOG_LEVEL` (default: `DEBUG`) - Controls log level for source warnings

**Sink errors/warnings** (from writing to DataHub):
- `DATAHUB_SINK_ERROR_LOG_LEVEL` (default: `DEBUG`) - Controls log level for sink errors
- `DATAHUB_SINK_WARNING_LOG_LEVEL` (default: `DEBUG`) - Controls log level for sink warnings

Set to `ERROR` or `WARNING` to surface these messages at standard logging levels.

### Log Format

Logs use a consistent, greppable format:

```
[INGEST SOURCE ERROR] title=API Fail error=timeout
[INGEST SOURCE WARNING] title=Schema Issue warning=bad column
[INGEST SINK ERROR] urn=urn:li:dataset:(urn:li:dataPlatform:kafka,test,PROD) aspect=datasetProperties error=Unable to emit
[INGEST SINK WARNING] warning=rate limited
```

### Benefits

- Real-time visibility into ingestion issues without waiting for final report
- Separate control over source vs sink logging
- Separate control over error vs warning visibility
- Includes contextual information (URN, aspect, workunit, title)
- Default DEBUG level means no noise unless explicitly enabled
- No changes to report structure

## Test plan

- [x] Unit tests for configurable log levels
- [x] End-to-end testing verified logs hidden by default and visible when env vars set
- [x] Linting passed


metadata-ingestion/src/datahub/ingestion/api/source.py
```diff
@@ -258,6 +264,12 @@ def report_warning(
         """
         See docs of StructuredLogs.report_log for details of args
         """
+        # Log at configured level (this method is for silent reporting, not immediate logging)
+        log_content = f"{message} => {context}" if context else message
+        if title:
+            log_content = f"{title}: {log_content}"
+        logger.log(_SINK_WARNING_LOG_LEVEL, log_content)
+
```
treff7es, line 271: The feature is already served by _log_sink_failure_or_warning() in datahub_rest.py. SourceReport is a generic type that should not encode sink-specific logging concerns.

``` suggestion
```

metadata-ingestion/src/datahub/ingestion/api/source.py
```diff
@@ -302,6 +314,13 @@ def report_failure(
         """
         See docs of StructuredLogs.report_log for details of args
         """
+        # Log at configured level if not already logging at ERROR level
+        if not log:
+            log_content = f"{message} => {context}" if context else message
+            if title:
+                log_content = f"{title}: {log_content}"
+            logger.log(_SINK_ERROR_LOG_LEVEL, log_content)
+
```
treff7es, line 325: The feature is already served by _log_sink_failure_or_warning() in datahub_rest.py. SourceReport is a generic type that should not encode sink-specific logging concerns.

```suggestion
```

metadata-ingestion/src/datahub/ingestion/sink/datahub_rest.py
```diff
@@ -130,6 +130,47 @@ def _get_partition_key(record_envelope: RecordEnvelope) -> str:
...
+
+    Args:
+        record_envelope: The record envelope containing the failed record
+        message: The error or warning message
+        is_error: True for errors, False for warnings
+    """
+    from datahub.configuration.env_vars import (
+        get_sink_error_log_level,
+        get_sink_warning_log_level,
+    )
```
treff7es: Import should be at the top

metadata-ingestion/tests/unit/api/test_report_sample_size.py
```diff
@@ -31,3 +33,85 @@ def test_source_report_sample_sizes(self, monkeypatch: pytest.MonkeyPatch) -> No
...
+        Note: Need to reload the module to pick up the new env var value since
+        log levels are computed at module load time.
+        """
+        import importlib
+        import sys
+
+        monkeypatch.setenv("DATAHUB_SINK_WARNING_LOG_LEVEL", "WARNING")
+
+        # Reload modules to pick up new env var
+        if "datahub.configuration.env_vars" in sys.modules:
```
treff7es: Why is this not needed in other tests?
If you were to move the module-level constants and you would call get_sink_error_log_level() / get_sink_warning_log_level() directly at the call site in both methods (the cost is negligible — os.getenv is a dict lookup). This also aligns with the existing datahub_rest.py pattern.

metadata-ingestion/src/datahub/ingestion/api/source.py
```diff
@@ -69,6 +71,10 @@
 
 _MAX_CONTEXT_STRING_LENGTH = 1000
 
+# Compute log levels once at module load for better performance
+_SINK_ERROR_LOG_LEVEL = get_sink_error_log_level()
```
treff7es: This just makes testing complex and this is a cheap operation. You can get this where you use the variable

metadata-ingestion/src/datahub/ingestion/sink/datahub_rest.py
```diff
@@ -278,12 +322,22 @@ def _write_done_callback(
                 if workunit_id := record_envelope.metadata.get("workunit_id"):
                     e.info["workunit_id"] = workunit_id
 
+                # Log detailed context before reporting
                 if not self.treat_errors_as_warnings:
+                    _log_sink_failure_or_warning(
+                        record_envelope, e.message, is_error=True
+                    )
                     self.report.report_failure({"error": e.message, "info": e.info})
```
treff7es, line 328: I think it would be easier to add one extra line adn remove all the other changes which just adds additional complexity even though the only thing you want is to log the sink errors in the live logs:

```suggestion
logger.debug("Sink failure: %s | %s", e.message, e.info)
self.report.report_failure({"error": e.message, "info": e.info})
```

metadata-ingestion/src/datahub/ingestion/sink/datahub_rest.py
```diff
@@ -278,12 +322,22 @@ def _write_done_callback(
...
                 else:
+                    _log_sink_failure_or_warning(
+                        record_envelope, e.message, is_error=False
+                    )
                     self.report.report_warning({"warning": e.message, "info": e.info})
                 write_callback.on_failure(record_envelope, e, e.info)
```
treff7es, line 334:
```suggestion
				   self.report.report_warning({"warning": e.message, "info": e.info})
write_callback.on_failure(record_envelope, e, e.info)
```

treff7es (changes requested):

treff7es:
I think the pr could be simplified by:

1. Remove all source.py changes (the _SINK_* constants, the new log blocks in report_warning / report_failure, and the new imports).
  2. Remove `TestDetailedSinkLogging` test class.
  3. In `_write_done_callback, add logger.warning()` for the `OperationalError` failure path and `logger.info()` for the warning path (or use `logger.debug()` if consistent with existing codebase norms).
  4. Optionally keep `_log_sink_failure_or_warning()` as a small helper if you want structured context in the log line — but it can be inlined trivially since the context (`e.info`) already has the URN and workunit_id.
  5. Keep the two env var getters only if the env-var-configurable level is a deliberate design decision — but given Python's existing logging infrastructure, it's likely unnecessary.

metadata-ingestion/src/datahub/ingestion/api/source.py
treff7es: Why did you resolve this comment? This should be removed; this PR will touch parts it shouldn't. This is a generic component that shouldn't interact with the sink.

metadata-ingestion/src/datahub/ingestion/api/source.py
treff7es: Same here, it shouldn't know anything about sinks. This should be removed.

metadata-ingestion/src/datahub/configuration/env_vars.py
```diff
@@ -159,6 +160,60 @@ def get_report_warning_sample_size() -> int:
     return int(os.getenv("DATAHUB_REPORT_WARNING_SAMPLE_SIZE", "10"))
 
 
+def get_sink_error_log_level() -> int:
```
treff7es, line 163: I don't think these env vars needed

metadata-ingestion/src/datahub/ingestion/sink/datahub_rest.py
```diff
@@ -130,6 +132,43 @@ def _get_partition_key(record_envelope: RecordEnvelope) -> str:
     return str(uuid.uuid4())
 
 
+def _log_sink_failure_or_warning(
```
treff7es, line 135: I don't think this is needed

metadata-ingestion/src/datahub/ingestion/sink/datahub_rest.py
treff7es: Why is the whole `_log_sink_failure_or_warning` needed and not just logging a debug or warning level log `logger.debug("Sink failure: %s | %s", e.message, e.info)`

PR #16705, opened 2026-03-20
fix(ci): yarn lint failure

benjiaming:
### Summary
 `yarn lint` ran 3 heavy tasks sequentially in a single process: ESLint (type-aware, 3,664 files), Prettier, and `tsc --noEmit`. ESLint with airbnb-typescript creates a full TypeScript program in memory (~4-6 GB). Combined with tsc and Gradle JVM overhead, this easily exceeds 8 GB.

### Fix
`yarn lint` is split into 3 jobs.

### Important

Branch protection rules requiring the old check name `datahub-web-react-lint` will need to be updated to new check names instead.


devashish2203:
Oss runners are already 16GB machines so the issue does not happen in OSS, will need to test in fork with a smaller runner. Cost wise I'm not sure if this helps as running 3 smaller runners vs one bigger runner would balance out roughly. I was hoping the fix would be to reduce memory consumption so that the current job can run faster and with lesser resources.

Also if the tasks are sequential why would running them in one job increase memory usage, won't running the heavy task alone still use similar amount of memory?

Mergeable will automatically take care of branch protection rules, so no update needed there.

PR #16709, opened 2026-03-21
perf(ui): Drop react-phosphor v1 and migrate to v2

benjiaming:
### Summary
Migrating left over phosphor v1 icons to v2. Decreasing bundle size by 50kB.

Meticulous is reporting a regression (health check not showing up). Confirmed that this still works:

Screenshot 2026-03-20 at 7 03 06 PM

### DIFF
--- a/datahub-web-react/package.json
+++ b/datahub-web-react/package.json
@@ -77,7 +77,6 @@
         "moment": "^2.29.4",
         "moment-timezone": "^0.5.35",
         "monaco-editor": "^0.28.1",
-        "phosphor-react": "^1.4.1",
         "prosemirror-autocomplete": "^0.4.3",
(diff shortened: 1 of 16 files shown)


datahub-web-react/package.json
```diff
@@ -77,7 +77,6 @@
         "moment": "^2.29.4",
         "moment-timezone": "^0.5.35",
         "monaco-editor": "^0.28.1",
-        "phosphor-react": "^1.4.1",
```
AdrianMachado, line 80: Maybe I missed something, but why can we remove this now?

benjiaming: This is v1. We use `phosphor-icons' which is v2. https://github.com/phosphor-icons/react

PR #16711, opened 2026-03-21
fix(test): mock WorkspaceClient in tests to prevent CI hangs

shirshanka:
## Summary

- Mock `create_workspace_client` in unity catalog and fivetran tests to prevent real network calls to fake Databricks URLs on CI
- On GitHub Actions, the Databricks SDK's `WorkspaceClient` attempts real connections to `https://test.databricks.com` which time out at ~5 min per test
- **testQuick**: 9 unity catalog tests × 5 min = ~47 min wasted
- **integrationBatch0**: 16 fivetran parametrized tests × 5 min = ~75 min wasted (caused timeout)

The issue doesn't reproduce locally because the Databricks SDK behaves differently when `GITHUB_ACTIONS=true` is in the environment.

## Test plan

- [x] All 19 unity catalog tests pass locally in 1.9s (was 2.6s, no behavior change)
- [x] All 16 fivetran transpilation tests pass locally in 2.3s
- [x] `./gradlew :metadata-ingestion:lintFix` passes
- [ ] CI testQuick completes in ~8 min (down from 60+ min)
- [ ] CI integrationBatch0 completes without timeout (down from 75+ min)

### DIFF
--- a/metadata-ingestion/tests/integration/fivetran/test_fivetran.py
+++ b/metadata-ingestion/tests/integration/fivetran/test_fivetran.py
@@ -208,9 +208,14 @@ def test_quoted_query_transpilation(schema):
     fivetran_log_query = FivetranLogQuery()
     fivetran_log_query.use_database("test_database")
 
-    with mock.patch(
-        "datahub.ingestion.source.fivetran.fivetran_log_api.create_engine"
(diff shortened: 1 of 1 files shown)


shirshanka:
Merging because test failures are unrelated

PR #16720, opened 2026-03-23
ci: skip connector tests when PR only changes tests or docs

rajatoss:
## Summary

- Adds path-based filtering to the connector-tests-trigger workflow to skip connector test dispatch when a PR's `metadata-ingestion/` changes are limited to **tests** (`metadata-ingestion/tests/`) or **docs** (`*.md` files)
- Uses the GitHub API (`pulls/{pr}/files`) to detect changed files 
- When skipped, creates a check run with conclusion `skipped` and a clear reason message
- All source code, config (`pyproject.toml`, `setup.py`, `constraints.txt`), and dependency changes still trigger the full connector test suite

## How it works

After the existing skip-label check, a new step fetches the PR's changed files via `gh api` and filters:
1. Keep only `metadata-ingestion/` files
2. Strip out `metadata-ingestion/tests/**` and `**/*.md`
3. If nothing remains → skip connector tests
4. If any file remains → run connector tests as normal

## Test plan

- [ ] PR with only `metadata-ingestion/tests/` changes → connector tests skipped with reason "changes are limited to tests or docs"
- [ ] PR with only `*.md` changes in metadata-ingestion → connector tests skipped
- [ ] PR with source code changes → connector tests dispatched as normal
- [ ] PR with `skip-connector-tests` label → still shows label-specific skip reason
- [ ] PR with mixed test + source changes → connector tests run

### DIFF
--- a/.github/workflows/connector-tests-trigger.yml
+++ b/.github/workflows/connector-tests-trigger.yml
@@ -78,18 +78,11 @@ jobs:
       - name: Get PR info
         if: steps.check-both.outputs.both_passed == 'true'
         id: pr-info
-        env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          BRANCH="${{ github.event.workflow_run.head_branch }}"
-          REPO="${{ github.repository }}"
-          HEAD_OWNER="${{ github.event.workflow_run.head_repository.owner.login }}"
-
-          PR_NUMBER=$(gh api "/repos/${REPO}/pulls?head=${HEAD_OWNER}:${BRANCH}&state=open" \
-            --jq '.[0].number // empty')
+          PR_NUMBER="${{ github.event.workflow_run.pull_requests[0].number }}"
 
           if [ -z "$PR_NUMBER" ]; then
-            echo "No open PR found for branch ${BRANCH} — skipping"
+            echo "No open PR found — skipping"
             echo "has_pr=false" >> "$GITHUB_OUTPUT"
           else
             echo "has_pr=true" >> "$GITHUB_OUTPUT"


.github/workflows/connector-tests-trigger.yml
```diff
@@ -137,22 +137,70 @@ jobs:
...
+
+          if [ -z "$mi_files" ]; then
+            echo "No metadata-ingestion changes — skipping connector tests"
+            echo "has_connector_changes=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          # Strip out files that cannot affect connectors
+          connector_relevant_files="$mi_files"
+          connector_relevant_files=$(echo "$connector_relevant_files" | grep -v '^metadata-ingestion/tests/' || true)
```
devashish2203: You can combine this into a single statement I think.
$(echo "$connector_relevant_files" | grep -v '^metadata-ingestion/tests/' | grep -v '\.md$' || 'true)

.github/workflows/connector-tests-trigger.yml
```diff
@@ -137,22 +137,70 @@ jobs:
             echo "skip=false" >> "$GITHUB_OUTPUT"
           fi
 
+      - name: Check if connector-relevant files changed
```
devashish2203, line 133: This should be one of the initial steps to run. If no relevant files are changed or the label exists, all above steps are not required to run.

You can refactor this workflow into two sequential jobs like below.

- prepare-trigger: Runs the following in order, and sets an output trigger-tests: true / false
  - get-pr-info
  - check-label. If label found no further steps run
  - paths-check. If paths don't match filters no further steps run.
  - Adds comment/skip check if either of the above requires that the tests don't run.
  - check-both-workflows-passed. If not, sets trigger-tests to false.
  
 

- Trigger tests: needs: prepare-trigger and checks if the outputs.prepare-trigger.trigger-tests="true"
  - Runs app-token, get-cloudflare-url, and the following steps to actually trigger the connector tests.



Later prepare-trigger can be enhanced to check and return a list of connectors to run as output and used in trigger-tests.

.github/workflows/connector-tests-trigger.yml
devashish2203: Do you need the API to fetch this information?

Doesn't github.event.workflow_run.pull_requests work?
https://docs.github.com/en/webhooks/webhook-events-and-payloads#workflow_run:~:text=pull_requests%20array%20of%20object%2Cnulls%20Required

.github/workflows/connector-tests-trigger.yml
rajatoss: done

.github/workflows/connector-tests-trigger.yml
rajatoss: yes, initially we were thinking of running them on the community fork PR's also, we can use this as we just run them for internal users on the same repo.

.github/workflows/connector-tests-trigger.yml
rajatoss: i will try to take this in next PR

PR #16726, opened 2026-03-23
fix(ingest/bigquery): replace deprecated __TABLES__ with INFORMATION_SCHEMA.TABLE_STORAGE

treff7es:
## Summary

- Replaces all usages of the deprecated `__TABLES__` BigQuery pseudo-table with `INFORMATION_SCHEMA.TABLE_STORAGE` in the BigQuery v2 ingestion source
- Affected queries: `tables_for_dataset`, `views_for_dataset`, `snapshots_for_dataset`
- Column mapping: `row_count` → `total_rows`, `size_bytes` → `total_logical_bytes` (aliased to preserve downstream field names)
- Join type changed from `INNER JOIN` to `LEFT JOIN` to handle empty/recently-created tables that may not yet have a `TABLE_STORAGE` entry

## Test plan

- [ ] All 193 existing BigQuery unit tests pass
- [ ] Run integration test against a live BigQuery project to verify row counts and storage sizes are populated correctly

Relates to ACR-6838


metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_schema.py
```diff
@@ -510,7 +522,7 @@ def _make_bigquery_table(
             name=table.table_name,
             created=table.created,
             table_type=table.table_type,
-            last_altered=parse_ts_millis(table.get("last_altered")),
+            last_altered=table.get("last_altered"),
```
sgomezvillamor, line 525: how is not parsing anymore?

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/queries.py
```diff
@@ -41,12 +41,12 @@ class BigqueryQuery:
   t.table_name as table_name,
   t.table_type as table_type,
   t.creation_time as created,
-  ts.last_modified_time as last_altered,
+  ts.storage_last_modified_time as last_altered,
```
sgomezvillamor, line 44: storage_last_modified_time | TIMESTAMP

I guess that's why we are not parsing anymore 😅

pdruley:
> AFAIK this was requested to get rid of the _get data_ permission on `__TABLES__` table
> 
> So we may need to update required permissions on the user facing docs

@treff7es I actually agree, we should update the user docs to remove the "getData" permission if it's not needed anymore or if it's truly optional. We make a note saying it's required if profiling is enabled or if use_tables_list_query_v2 is enabled.

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/queries.py
treff7es: exactly

PR #16735, opened 2026-03-23
chore(tests/mssql): update formatting of golden files

ligfx:

PR #16736, opened 2026-03-23
fix(ingest/mssql): lowercase field paths when convert_urns_to_lowercase=True

ligfx:

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/sql/mssql/source.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/sql/mssql/source.py
@@ -1272,10 +1272,13 @@ class SQLServerSource(SQLAlchemySource):
         fk_dict: Dict[str, Any],
         inspector: Inspector,
     ) -> ForeignKeyConstraintClass:
-        fk_dict["constrained_columns"] = [
-            f.lower() for f in fk_dict["constrained_columns"]
-        ]
-        fk_dict["referred_columns"] = [f.lower() for f in fk_dict["referred_columns"]]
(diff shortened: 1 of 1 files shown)


ligfx:

- #16683
    - #16735
        - #16736 ◀

metadata-ingestion/src/datahub/ingestion/source/sql/mssql/source.py
```diff
@@ -1244,6 +1265,19 @@ def construct_flow_workunits(
...
 
+    def get_foreign_key_metadata(
+        self,
+        dataset_urn: str,
+        schema: str,
+        fk_dict: Dict[str, Any],
+        inspector: Inspector,
+    ) -> ForeignKeyConstraintClass:
+        fk_dict["constrained_columns"] = [
+            f.lower() for f in fk_dict["constrained_columns"]
```
askumar27: We are missing the `self.config.convert_urns_to_lowercase` guard

ligfx: Yes indeed. I guess we don't actually test when `convert_urns_to_lowercase == False`. Fixed.

askumar27 (approved):
LGTM! 

Can you add test coverage for `convert_urns_to_lowercase=False` case?

ligfx:
> Can you add test coverage for convert_urns_to_lowercase=False case?

Not here, someone else can pick that up. I don't believe we actually care about that case — the ingestion process throws a big warning if it's not set, and cross-platform lineage will break.

PR #16748, opened 2026-03-24
feat(ingest/snowflake): cross account lineage through snowflake shares

alokr-dhub:
## Summary

Adds zero-config cross-account lineage between Snowflake accounts that share data. When a producer account ingests, a separate consumer-side recipe automatically discovers the share, resolves the producer, and emits Siblings + COPY lineage — **no manual `shares` config required, no elevated privileges**.

---

### Auto-discovery on consumers

For each shared database (detected via `SHOW DATABASES.origin`), the connector resolves the producer in two independent steps:

1. **Producer identity** — `account_mapping` config, then `account_locator_fallback` convention, then skip with structured warning. (No graph lookup here today — see Known Limitations.)
2. **Producer database name** — graph lookup of the producer's published `share_database_mapping`, falling back to consumer-side `share_database_mapping` config, then skip with structured warning.

Then it emits `Siblings` + `UpstreamLineage` to the producer URN. Skipped when manual `shares` config already covers the database, so existing recipes are unchanged.

### Producer-side share→DB mapping publication

Mines `SNOWFLAKE.ACCOUNT_USAGE.QUERY_HISTORY` for `GRANT USAGE ON DATABASE … TO SHARE …` DDL and publishes the resulting `share_database_mapping` on the producer's `DataPlatformInstance.customProperties`. Consumers read this from the graph to resolve producer database names.

Uses `IMPORTED PRIVILEGES on SNOWFLAKE` (already granted for lineage). Subject to `QUERY_HISTORY`'s 365-day retention; falls back to consumer-side `share_database_mapping` config for older shares.

### Org / account metadata on `DataPlatformInstance`

`organization_name`, `account_locator`, and `account_identifier` (org-qualified) are emitted as custom properties whenever `platform_instance` is set. Captured via `CURRENT_ORGANIZATION_NAME()` and `CURRENT_ACCOUNT()`.

### Enhanced database metadata

`origin`, `kind`, `is_transient`, `retention_time`, `owner` merged from `SHOW DATABASES` into `information_schema` results and exposed as database-container `customProperties`.

---

## Configuration

| Option                            | Default | Purpose                                                              |
| --------------------------------- | ------- | -------------------------------------------------------------------- |
| `include_organization_metadata`   | `true`  | Capture org name on `DataPlatformInstance`                           |
| `include_show_databases_metadata` | `true`  | Merge `SHOW DATABASES` fields (required for auto-discovery)          |
| `auto_discover_inbound_shares`    | `true`  | Master toggle for consumer-side auto-discovery                       |
| `publish_share_database_mapping`  | `true`  | Master toggle for producer-side `QUERY_HISTORY` mining               |
| `account_mapping`                 | `{}`    | `account_identifier → platform_instance` (Option B)                  |
| `account_locator_fallback`        | `false` | Use `account_locator` as `platform_instance` when no mapping (Option A) |
| `share_database_mapping`          | `{}`    | Manual `share_name → producer_db` fallback for shares > 365 days     |
| `validate_producer_urns_in_graph` | `false` | Opt-in `graph.exists()` check before emitting siblings (fail-open on graph errors) |

A model validator enforces the dependency: setting `include_show_databases_metadata: false` while `auto_discover_inbound_shares` is `true` would silently no-op auto-discovery (no `origin` field), so the validator auto-disables auto-discovery and warns.

---

## Privilege Model

All features work with the standard `datahub_role` plus existing `IMPORTED PRIVILEGES on SNOWFLAKE`. See the [Snowflake Shares → Requirements](metadata-ingestion/docs/sources/snowflake/snowflake_post.md) section in the connector docs for a per-flag breakdown.

The OWNERSHIP-based enumeration was explicitly rejected:

- Share OWNERSHIP is single-owner and would strip rights from the customer's existing share-management role
- MANAGE SHARES is account-scoped and rarely granted

---

## New Report Counters

- `num_siblings_emitted`
- `num_auto_shares_discovered`
- `num_auto_shares_skipped_unresolved_producer`
- `num_auto_shares_skipped_unknown_share_db`
- `num_auto_shares_skipped_producer_urn_missing`
- `num_share_database_mappings_published`
- `num_share_grant_queries_skipped_oversized`

---

## Observability

All failure paths use structured `report.warning(title=…, message=…, context=…, exc=…)` with `LiteralString` titles/messages and dynamic values in `context` so the report deduplicates correctly. Notable surfaces:

- **Producer URN existence check fails** — fail-open: emit lineage, surface a warning so the user knows validation was skipped.
- **`CURRENT_REGION` query fails when `include_external_url: true`** — warn so the user knows Snowsight URLs will be omitted (previously silent debug).
- **`publish_share_database_mapping: true` but `organization_name` is unknown** — warn that cross-account consumers can't resolve the producer by graph lookup; they need `account_mapping` with the bare locator instead.
- **Share-grant queries skipped due to >1MB body** — counted on `report.num_share_grant_queries_skipped_oversized` (no longer silently dropped).
- **`account_mapping` / `share_database_mapping` keys collide after upper-casing** — config validator warns instead of silent last-write-wins.

---

## Type Design

- `ShareOrigin` (frozen dataclass) — `parse()` rejects malformed and empty-segment origins; `is_shared_database()` and `share_origin` derive from a single parse so they can never disagree.
- `ResolvedProducer` (frozen dataclass) — replaces an untyped `Optional[tuple]` from `_resolve_producer` with explicit `platform_instance` / `database` fields.
- `ShareGrantDiscovery` (frozen dataclass) — replaces `parse_share_grants`'s `Dict[str, str]` empty-dict overload with explicit `mapping` + `skipped_oversized_queries` fields.
- `SnowflakeDatabase.is_transient`: `Optional[bool] = None` for symmetry with the other SHOW DATABASES fields ("not transient" vs "we didn't query SHOW DATABASES" are distinguishable).

---

## Backward Compatibility

- Existing `shares` config continues to take precedence; auto-discovery skips databases already covered manually
- All new config flags default to safe values; setting `auto_discover_inbound_shares: false` and `publish_share_database_mapping: false` restores prior behavior exactly
- `inbounds()` / `outbounds()` now return upper-cased keys (Snowflake's storage form) — callers no longer need per-call `.upper()` plumbing
- A `## Next > Other Notable Changes` entry is added to `docs/how/updating-datahub.md`

---

## Test Plan

- **499 snowflake unit tests passing** (was 477 before this branch; +22 net since the original review)
- **33 snowflake integration tests passing**
- `snowflake_privatelink_golden.json` regenerated for the new `DataPlatformInstance` MCP
- Lint and markdown-prettier clean (`./gradlew :metadata-ingestion:lintFix :datahub-web-react:mdPrettierCheck`)

### New unit coverage

- `parse_share_grants` — DDL with comments, large query texts (>1MB), quoted identifiers, oversized-skip count
- `discover_share_database_mapping` — success, query failure, truncation at over-fetch boundary, no-warn at exact cap, oversized skip count
- Graph-backed resolution — mapping precedence over local config, fallback when graph misses, corrupt JSON, non-dict JSON, null customProperties
- `validate_producer_urns_in_graph` — fail-open on `graph.exists()` exception, skip when not exists
- `ShareOrigin` — 3-part / 2-part / malformed / empty-segment rejection; `is_shared_database()` invariant
- Account resolver — Option A and Option B fallback ordering, case-insensitive matching, locator fallback, mapping precedence
- SHOW DATABASES merge — happy path, uppercase keying, missing rows, opt-out flag, `getattr` fallback for `SnowflakeSummaryConfig`
- `show_databases()` parsing — `is_transient` YES/NO/missing → True/False/None, `retention_time` int coercion, `origin` empty-string normalization
- `_gen_platform_instance_workunits` — `publish_share_database_mapping` happy path, empty mapping, disabled flag, org-name-missing warning
- `inspect_session_metadata` — region failure warning gated on `include_external_url`
- Validator — `auto_discover_inbound_shares` auto-disabled when `include_show_databases_metadata: false`

---

## Docs

- New `##### Requirements` h5 under "Snowflake Shares" in `metadata-ingestion/docs/sources/snowflake/snowflake_post.md` covering per-flag privileges, Snowflake 6.0 caveat for org name, and the 365-day `QUERY_HISTORY` retention caveat
- "Automatic Share Discovery" subsection covering the consumer-side resolution order, producer-side mining, manual fallback, and backward-compat note
- `## Next > Other Notable Changes` entry in `docs/how/updating-datahub.md`

---

## Known Limitations

- **`account_mapping` (or `account_locator_fallback`) is required, not automatic.** Producer *identity* resolution has no graph-lookup path — only `account_mapping`/`account_locator_fallback` config. Only producer *database name* resolution is genuinely automatic via the graph. So this isn't fully zero-config: the consumer recipe must declare `account_mapping` (or opt into `account_locator_fallback`) for auto-discovery to resolve anything.
- **Outbound/producer-side ghost nodes are unaddressed** (pre-existing, not introduced by this PR). `get_shares_workunits` with `is_outbound=True` emits siblings for every table/view in the producer's local database, not just the ones actually granted to the share — Snowflake shares are object-level grants, not whole-database. A fix would need either `SHARE OWNERSHIP`/`MANAGE SHARES` (rejected — see the OWNERSHIP-based enumeration note above) or mining `GRANT ... ON TABLE/VIEW ... TO SHARE` from `QUERY_HISTORY` alongside the existing `GRANT USAGE ON DATABASE` mining, within the same privilege footprint already required.
- **`platform_instance` ↔ Snowflake account is user-asserted, not enforced.** If one account is ingested under multiple `platform_instance` values (e.g. split by filter across recipes), a stale or wrong `account_mapping` entry can construct a producer URN that was never actually ingested — a dangling sibling link. `validate_producer_urns_in_graph` (off by default) is the existing safety net; recommended, not just optional, for anyone running multiple recipes per account.

Follow-up idea (not implemented here): since `account_identifier` is already published as a `customProperties` entry, a `DataHubGraph.get_urns_by_filter` search on that property could resolve the producer's `platform_instance` without static `account_mapping`, and self-disambiguate the multi-recipe-per-account case by checking which candidate's published `share_database_mapping` actually contains the resolved share name.


puneetagarwal-datahub:
Closing this since we need a generic solution across conectors for this

PR #16752, opened 2026-03-24
feat(ingestion/hex): Allow category filters

alisa-aylward-toast:
Allowing for filtering on categories in Hex. In the ingestion, users can decide if they want to include/exclude certain Hex categories.


metadata-ingestion/src/datahub/ingestion/source/hex/hex.py
```diff
@@ -256,6 +260,19 @@ def get_report(self) -> HexReport:
     def get_workunits_internal(self) -> Iterable[MetadataWorkUnit]:
         with self.report.new_stage("Fetch Hex assets from Hex API"):
             for project_or_component in self.hex_api.fetch_projects():
+                skip_item = False
+                if project_or_component.categories:
+                    for category in project_or_component.categories:
+                        if not self.source_config.category_pattern.allowed(
+                            category.name
+                        ):
+                            skip_item = True
```
alokr-dhub: from the code it seems the requirement is to skip the `project_or_component` entirely if any of the categories does not match the allowed pattern. Please add more info in the description explaining the expected behavior. Even considering the expected behavior based on code logic, the code looks more complex than needed. this can be simplified as  
```
if project_or_component.categories and any(
    not self.source_config.category_pattern.allowed(c.name)
    for c in project_or_component.categories               
):                                          
    continue  
```

metadata-ingestion/tests/unit/hex/test_data/hex_projects_page3.json
```diff
@@ -0,0 +1,147 @@
...
+        "email": "user4a@example.com"
+      },
+      "owner": {
+        "email": "user4a@example.com"
+      },
+      "status": null,
+      "categories": [{
+          "name": "Keep_Scratchpad",
+          "description": "Intended for broad consumption"
+        }],
```
alokr-dhub, line 165: Please add more test cases with mix of allowed and denied categories for more rigorous testing.

metadata-ingestion/src/datahub/ingestion/source/hex/hex.py
alisa-aylward-toast: question on how patterns work -- if the no pattern is added for this in the ingestion, will self.source_config.category_pattern.allowed always be true?

alokr-dhub: yes. the default is set to `allow_all()`. `.allowed` always returns True unless an allow list is specified.

alokr-dhub: Also, i would suggest moving the filtering logic to a separate method will be cleaner.

metadata-ingestion/src/datahub/ingestion/source/hex/hex.py
```diff
@@ -256,6 +260,12 @@ def get_report(self) -> HexReport:
     def get_workunits_internal(self) -> Iterable[MetadataWorkUnit]:
         with self.report.new_stage("Fetch Hex assets from Hex API"):
             for project_or_component in self.hex_api.fetch_projects():
+                if project_or_component.categories and any(
```
alisa-aylward-toast: @alokr-dhub  where would you suggest putting this logic in terms of a new function? I ask because having it here pattern matches lines 270 and 278, so I didn't find a natural place to put the function.
i could add it to the HexAPI class if you think that's a good spot for it

alokr-dhub: ok. we can keep it here now. ideally all the checks should be in a separate internal method in the hex source class

alokr-dhub:
@alisa-aylward-toast LGTM. Please add the test cases.

alisa-aylward-toast:
@alokr-dhub since your first check I did add two more test cases. I added one with two categories -- one that is excluded -- https://github.com/datahub-project/datahub/pull/16752/changes#diff-5c43981e8e278bbb4aec0157f6dc7bce9da1425d7d6e7485ca807a223a499035R101
and I added tests
https://github.com/alisa-aylward-toast/datahub/commit/b72b29594dd8cdd6a598a3c7faa7fbb61d18806e

PR #16764, opened 2026-03-24
feat(cli): add --context / -C flag for passing context in event properties

shirshanka:
## Summary

- Add `--context` / `-C` option on the root `datahub` command that accepts `key=value` pairs
- Can be repeated: `datahub -C skill=audit -C caller=claude search "..."`
- Values are included in event properties prefixed with `ctx_`

## Test plan

- [x] Unit tests for context parsing and propagation
- [ ] CI passes

### DIFF
--- a/metadata-ingestion/src/datahub/telemetry/telemetry.py
+++ b/metadata-ingestion/src/datahub/telemetry/telemetry.py
@@ -417,6 +417,19 @@ _T = TypeVar("_T")
 _P = ParamSpec("_P")
 
 
+def _get_cli_context() -> Dict[str, str]:
+    """Read --context key=value pairs from the Click context, if available."""
+    try:
+        import click
+
+        ctx = click.get_current_context(silent=True)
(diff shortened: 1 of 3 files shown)


metadata-ingestion/src/datahub/telemetry/telemetry.py
```diff
@@ -110,14 +110,48 @@
...
+    "DSN",
+    "URI",
+    "KEY",
+    "CERT",
+    "AUTH",
+}
+_DATAHUB_ENV_MAX_VALUE_LEN = 128
+
+
+def _collect_datahub_env_context() -> Dict[str, str]:
```
nwadams: would it be better to allowlist specific collection? Also do we know if envvar or `~/.datahubenv` is more prevalent. If the latter, we should capture as well.

shirshanka: changed approach to just have `--context / -C` on the command line

nwadams (approved):

PR #16769, opened 2026-03-25
fix(ingest/mssql): set convert_urns_to_lowercase=True in default and example recipes

ligfx:


ligfx:
@askumar27 yep, exactly!

PR #16770, opened 2026-03-25
fix: Revert "fix(ingest/mssql): lowercase field paths when convert_urns_to_lowercase=True (#16736)"

ligfx:
Reverts #16736, which breaks pre-existing column documentation and tags/terms/structured properties.


askumar27 (approved):
LGTM!

PR #16779, opened 2026-03-25
feat: Update csv enricher to add support for structured properties

hasb-trst:


metadata-ingestion/src/datahub/ingestion/source/csv_enricher.py
```diff
@@ -335,6 +338,106 @@ def get_resource_description_work_unit(
...
+                column[len("sp__") :]
+                if column.startswith("sp__")
+                else column[len("sp.") :]
+            ).strip()
+            if len(raw_property_name) <= 0:
+                resource_urn = row.get("resource", "unknown")
+                self.report.report_warning(
+                    f"{resource_urn}-invalid-structured-property-column-{column}",
+                    f"Ignoring structured property column '{column}' because the property name is empty.",
+                )
```
treff7es:
```suggestion
				self.report.warning(
				      title="Invalid Structured Property Column",
				      message=f"Ignoring column '{column}' because the property name is empty.",
				      context=f"resource={resource_urn}",
  )```

**Review by treff7es — COMMENTED** [review]
Please, can you add some doc in `csv-enricher_pre.md` about the new feature? 
Without that I don't think anybody will be able to find it.

**Comment by hasb-trst** [comment]
> Please, can you add some doc in `csv-enricher_pre.md` about the new feature? Without that I don't think anybody will be able to find it.

Updated the docs, let me know if something else needs to be added

File: metadata-ingestion/src/datahub/ingestion/source/csv_enricher.py [hunk 2]
```diff
@@ -336,6 +339,107 @@ def get_resource_description_work_unit(
...
+                if column.startswith("sp__")
+                else column[len("sp.") :]
+            ).strip()
+            if len(raw_property_name) <= 0:
+                resource_urn = row.get("resource", "unknown")
+                self.report.warning(
+                    title="Invalid Structured Property Column",
+                    message=f"Ignoring column '{column}' because the property name is empty.",
+                    context=f"resource={resource_urn}",
+                )
```
sgomezvillamor, line 381:
```suggestion
                self.report.warning(
                    title="Invalid Structured Property Column",
                    message=f"Ignoring column because the property name is empty.",
                    context=f"resource={resource_urn}, column={column}",
                )
```

sgomezvillamor:
Thanks for this contribution! Adding structured properties support to the CSV enricher is a great enhancement that makes bulk metadata management much more flexible. A few notes below:

---

**⚠️ Column prefix design — worth a discussion**

The current approach uses two interchangeable prefixes (`sp__` and `sp.`) to identify structured property columns. A more generic and configurable approach could make this feature more robust and consistent with how similar patterns are handled elsewhere in DataHub. Would you be open to taking it?

Two options worth considering:

1. **Configurable prefix** — add a `structured_property_column_prefix` field to `CSVEnricherConfig` (defaulting to `sp__`). This mirrors the `key_value_property_prefix` pattern already used by `TagsToStructuredPropertiesTransformer` and gives users an escape hatch if the prefix conflicts with their CSV schema.

2. **Explicit column mapping in config** *(preferred — most flexible)* — a `structured_properties` dict in config mapping CSV column names directly to property URNs or names:
   ```yaml
   structured_properties:
     tier: "io.acryl.metadata.tier"
     classification: "urn:li:structuredProperty:io.acryl.privacy.classification"
   ```
   This avoids any magic prefix convention entirely, keeps the CSV schema clean, works with arbitrary column names, and is completely unambiguous. Existing CSVs wouldn't need column renames to start using structured properties.

Personally I'd go with option 2 — more configurable and less restrictive. Option 1 works too though.

---

**💡 OVERRIDE mode behavior — consider a docs/runtime notice**

This is expected behavior consistent with how other fields (description, glossary terms) work in OVERRIDE mode, so no code change is strictly required. That said, structured properties are a multi-valued map and a user might not realize that columns omitted from the CSV will remove existing properties on the entity. A small improvement could be:

- Adding a warning note to the docs (similar to the existing OVERRIDE caveat in the config description).
- Optionally, emitting an `info`-level log at runtime when OVERRIDE is active and structured properties are being written.

---

**💡 `StructuredPropertyUrn` helper available for URN construction**

Rather than building URNs via string formatting, `make_structured_property_urn` from `datahub.utilities.urns.structured_properties_urn` is already available and used in other connectors (Snowflake, Snowplow). It provides free format validation and keeps the code consistent:

```python
from datahub.utilities.urns.structured_properties_urn import make_structured_property_urn
property_urn = make_structured_property_urn(raw_property_name)
```

---

Happy to discuss any of the above — especially the prefix design, as that decision shapes the UX for anyone adopting this feature going forward.

---

sgomezvillamor (commented):
Overall LGTM
Just left a couple of enhancements
Thanks for the contrib!

hasb-trst:
> StructuredPropertyUrn helper available for URN construction

> Rather than building URNs via string formatting, make_structured_property_urn from datahub.utilities.urns.structured_properties_urn is already available and used in other connectors (Snowflake, Snowplow). It provides free format validation and keeps the code consistent:

> ```python 
> from datahub.utilities.urns.structured_properties_urn import make_structured_property_urn
> property_urn = make_structured_property_urn(raw_property_name)
> ```
@sgomezvillamor This function just accepts the string urn, doesn't work with ids or names. 
it fails on [this check](https://github.com/datahub-project/datahub/blob/master/metadata-ingestion/src/datahub/utilities/urns/_urn_base.py#L129) when I try to give string id

sgomezvillamor:
@hasb-trst thanks for addressing the comments

About `make_structured_property_urn`, it uses `from_string()` internally, which only accepts full URNs. But `StructuredPropertyUrn(id)` (the constructor) already accepts plain IDs — exactly how the Snowflake connector uses it.

The ideal fix would be a one-liner in `datahub/utilities/urns/structured_properties_urn.py`:

```python
def make_structured_property_urn(structured_property_id: str) -> str:
    if structured_property_id.startswith("urn:li:"):
        return str(StructuredPropertyUrn.from_string(structured_property_id))
    return StructuredPropertyUrn(structured_property_id).urn()
```

With that, the `csv-enricher` code simplifies to just `make_structured_property_urn(raw_property_name)` — no manual prefix construction needed. This also makes the helper more useful for other callers.

Approving anyway

Thanks for the contrib!

sgomezvillamor (approved):

hasb-trst:
@sgomezvillamor thanks for the approval, the CI seems to be stuck for quite a long time. Its been 3 hours and its still in progress.

sgomezvillamor:
@hasb-trst merged, thanks for the contrib!

PR #16787, opened 2026-03-25
feat(ingest/graph): Add openapi endpoints; refactor into separate file

asikowitz:
Add client for new endpoint in https://github.com/datahub-project/datahub/pull/16785 and refactored to separate out openapi endpoints into separate mixin for code cleanliness, as already done with `EntityVersioningAPI`.

### DIFF
--- a/smoke-test/tests/openapi/v3/test_scroll.py
+++ b/smoke-test/tests/openapi/v3/test_scroll.py
@@ -80,8 +80,8 @@ def test_scroll_entities_by_entity_type(graph_client: DataHubGraph) -> None:
     assert set(datasets.entities.keys()) == ALL_DATASET_URNS
     for urn, aspects in datasets.entities.items():
         assert urn.startswith("urn:li:dataset:")
-        dataset_key_aspect, _ = aspects["datasetKey"]
-        dataset_key_aspect = cast(DatasetKeyClass, dataset_key_aspect)
+        raw_aspect, _ = aspects["datasetKey"]
+        dataset_key_aspect = cast(DatasetKeyClass, raw_aspect)
         assert dataset_key_aspect.platform == PLATFORM
         assert dataset_key_aspect.name
 


metadata-ingestion/src/datahub/ingestion/graph/openapi.py
```diff
@@ -0,0 +1,525 @@
...
+        if entity_names is not None:
+            body["entities"] = entity_names
+        if aspects is not None:
+            body["aspects"] = aspects
+        if filter is not None:
+            body["filter"] = _raw_filter_to_v3_body(filter)
+        if sort_criteria is not None:
+            body["sortCriteria"] = sort_criteria
+
+        response = self._session.post(
```
askumar27, line 294: Use `_post_generic` similar to scroll_relationships instead of `self._session.post()` ?

smoke-test/conftest.py
```diff
@@ -107,17 +107,16 @@ def ingest_cleanup_data(auth_session, graph_client):
                 "tags_and_terms"
             )
     """
-    print(f"deleting {test_name} test data for idempotency")
+    logger.info(f"deleting {test_name} test data for idempotency")
     delete_urns_from_file(graph_client, data_file)
-    print(f"ingesting {test_name} test data")
+    logger.info(f"ingesting {test_name} test data")
     ingest_file_via_rest(auth_session, data_file)
-    wait_for_writes_to_sync()
```
askumar27, line 114: `wait_for_writes_to_sync` is needed here right? Would make the tests flaky.

metadata-ingestion/src/datahub/ingestion/graph/__init__.py
```diff
@@ -0,0 +1,6 @@
+from datahub.ingestion.graph.client import (
+    DatahubClientConfig,
+    DataHubGraph,
+    get_default_graph,
+)
+from datahub.ingestion.graph.openapi import RelatedEntity, RelationshipDirection
```
askumar27: We have this guideline here https://github.com/datahub-project/datahub/blob/master/AGENTS.md?plain=1#L238 -  "Code Quality: Avoid global state, use named arguments, don't re-export in __init__.py, refactor repetitive code"

smoke-test/conftest.py
asikowitz: `delete_urns_from_file` and `ingest_file_via_rest` already call it so I figured this was redundant

metadata-ingestion/src/datahub/ingestion/graph/__init__.py
asikowitz: Ok, I can get rid of these imports

askumar27: heads-up: We will need to re-import `RelatedEntity, RelationshipDirection`  it in the `client.py` to keep it backward compatible.

smoke-test/conftest.py
askumar27: It hard to track is the methods call it, is it better to call it outside for clarity ?

Because we call it outside here:
```
    # L117
    if to_delete_urns:
        delete_urns(graph_client, to_delete_urns)
        wait_for_writes_to_sync()
```

metadata-ingestion/src/datahub/ingestion/graph/openapi.py
asikowitz: Forgot about this, will handle in follow up

smoke-test/conftest.py
asikowitz: I don't feel strongly here

PR #16793, opened 2026-03-25
feat(ingest/mssql): add convert_column_urns_to_lowercase config

ligfx:


metadata-ingestion/src/datahub/ingestion/source/sql/mssql/source.py
```diff
@@ -408,6 +416,14 @@ def __init__(
                 title="Potential issue with lineage",
                 message="Lineage may not resolve accurately because 'convert_urns_to_lowercase' is False. To ensure lineage correct, set 'convert_urns_to_lowercase' to True.",
             )
+        if (
+            self.config.include_lineage
+            and not self.config.convert_column_urns_to_lowercase
+        ):
+            self.report.warning(
+                title="Potential issue with lineage",
```
askumar27: Can we make this explicit eg: "Potential issue with column-level lineage"

metadata-ingestion/src/datahub/ingestion/source/sql/mssql/source.py
```diff
@@ -408,6 +416,14 @@ def __init__(
...
+                message="Lineage may not resolve accurately because 'convert_column_urns_to_lowercase' is False. To ensure lineage correct, set 'convert_column_urns_to_lowercase' to True.",
```
askumar27: [nit] "To ensure lineage correct" -> "To ensure correct lineage"

metadata-ingestion/src/datahub/ingestion/source/sql/mssql/source.py
ligfx: Done

metadata-ingestion/src/datahub/ingestion/source/sql/mssql/source.py
ligfx: Good catch!

PR #16799, opened 2026-03-26
feat(ingestion/json-schema): add dataset_name_strategy for display name control

rospe:
## Summary

Add configurable display name strategies for JSON schema datasets. Previously, the display name was always the last path segment of the schema identifier (basename), which produced unhelpful names like `1.0` when schemas use versioned `$id` URIs.

## Motivation

Organizations using JSON Schema registries with structured `$id` URIs like `https://dev.schema.example.com/data-lake/enums/domain/1.0` get display names of just `1.0` in the DataHub UI. When many schemas share version-based names, this makes them indistinguishable. The current workaround requires a separate post-ingestion process to patch `DatasetProperties.name` via the API, which introduces timing dependencies and operational fragility.

## Changes

New config options on `JsonSchemaSourceConfig`:

- `dataset_name_strategy`: Controls how the display name is computed
  - `BASENAME` (default): Last path segment — current behavior, no breaking change
  - `SCHEMA_ID`: Full `$id` URI as the display name. Falls back to BASENAME if no `$id`
  - `TITLE`: Schema's `title` field. Falls back to BASENAME if no title
- `dataset_name_replace_pattern`: Simple string match-replace on the computed display name, applied after the strategy

These only affect the display name (`DatasetProperties.name`), not the dataset URN.

**Example:**
```yaml
source:
  type: json-schema
  config:
    path: /schemas
    platform: schemaregistry
    dataset_name_strategy: SCHEMA_ID
    dataset_name_replace_pattern:
      match: "https://dev.schema.example.com/"
      replace: ""

### DIFF
--- a/metadata-ingestion/docs/sources/dataplex/dataplex_post.md
+++ b/metadata-ingestion/docs/sources/dataplex/dataplex_post.md
@@ -93,9 +93,45 @@ For more details, see [Dataplex Lineage Documentation](https://docs.cloud.google
 - **`include_schema`** (default: `true`): Extract column metadata and types
 - **`include_lineage`** (default: `true`): Extract table-level lineage (automatically retries transient errors)
 
-**Performance Tuning:**
+#### Parallel Processing
 
-- **`batch_size`** (default: `1000`): Entries per batch for memory optimization. Set to `None` to disable batching (small deployments only)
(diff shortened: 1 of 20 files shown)


rospe:
HI @maggiehays , can you say when this will get picked up for review?

treff7es (approved):
lgtm, thanks for the contribution

Issue #11276, opened 2024-08-30
datahub-upgrade RestoreIndices process performance is poor on large sets of data


nmbryant:
**Describe the bug**
The datahub-upgrade RestoreIndices process significantly degrades as it runs when running against a large set of data. In our lower environment we have 9 million aspects and in our other environments we have 50+ million. With urnBasedPagination set to false, threads set to 10 and running in dev on all data, it was able to process 1 million in 20 minutes which was reasonably. However, as it runs the estimate to complete starts getting longer and longer with each batch taking significantly more time. After running for over 12 hours, it had only processed 4 million aspects before I stopped it.

Because of this, back in December we implemented urnBasedPagination which had linear performance regardless of the number of aspects. When we implemented it, RestoreIndices was able to process all 9M aspects in the dev environment in less than an hour. However, after upgrading to v0.12.1 from v0.10.5 the process is much slower with urnBasedPagination although still linear execution time, taking 1 hour to process 600K aspects.

**To Reproduce**
1. Run the RestoreIndices process of datahub-upgrade on all aspects, need at least a few million aspects to reproduce
2. With urnBasedPagination set to false and threads at 10, should see the performance of each batch degrade over time
3. With urnBasedPagination set to true, should see linear performance but slow at 600K aspects per hour

**Expected behavior**
Expect the process when threading and urnBasedPagination to false to be able to run against a large set of data in a reasonable amount of time. As it is now, it is not usable for large amounts of data.
With urnBasedPagination set to true, would like the performance restored to what it was when we first implemented it.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser: Chrome
 - Version v0.13.3

**Additional context**

nmbryant:
After some further analysis I was able to make some progress. I was able to restore the performance by adding the [original restoreIndices](https://github.com/nmbryant/datahub/blob/3ed6102d875b40452354d4044d7b89724c702ab6/metadata-io/src/main/java/com/linkedin/metadata/entity/EntityServiceImpl.java#L1136) function to EntityServiceImpl and adding [getPagedAspects](https://github.com/nmbryant/datahub/blob/3ed6102d875b40452354d4044d7b89724c702ab6/metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectDao.java#L464) to EbeanAspectDao.

Comparing the results between these changes and what's currently in main, there appear to be 2 parts that are significantly slower with the newer code. 

[...]

PR #16800, opened 2026-03-26
fix(database): Add index on version column for RestoreIndices performance

vejeta:
Fixes [#11276](https://github.com/datahub-project/datahub/issues/11276)

The RestoreIndices process runs COUNT(*) queries with WHERE version = ? which requires a full table scan without an index. On large datasets (80M+ aspects), this causes queries to take hours, making RestoreIndices  to timeout on this stage.

Adding an index on the version column reduces query time to minutes and the reindex process can go past the rowcount.
that is performed in https://github.com/datahub-project/datahub/blob/master/datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/restoreindices/SendMAEStep.java#L155

The error that we saw in our restore index process after the process being stuck for hours is:
```
[EbeanHook] WARN  io.ebean.datasource:116 - DataSource closing busy connection? name[gmsEbeanDatabaseConfig1] startTime[1774311959728] busySeconds[180] stackTrace[] stmt[select count(*) from metadata_aspect_v2 t0 where 
 t0.version = ?]
[EbeanHook] INFO  io.ebean.datasource:122 - Busy Connection - name[gmsEbeanDatabaseConfig1] startTime[1774311959728] busySeconds[180] stackTrace[] stmt[select count(*) from metadata_aspect_v2 t0 where t0.version = ?]   
[SpringApplicationShutdownHook] INFO  c.l.r.t.h.c.c.ChannelPoolManagerImpl:152 - Shutting down 0 connection pools                                                                                                          
[SpringApplicationShutdownHook] INFO  c.l.r.t.h.c.c.AbstractNettyClient:252 - Shutting down                                                                                                                                
[SpringApplicationShutdownHook] INFO  c.l.r.t.h.c.c.AbstractNettyClient:249 - Shutdown requested                                                                                                                           
[EbeanHook] INFO  io.ebean.datasource:122 - Dumping [1] busy connections: (Use datasource.xxx.capturestacktrace=true  ... to get stackTraces)                                                                              
[EbeanHook] WARN  io.ebean.datasource:116 - Closing busy connections on shutdown size: 1                                                                                                                                   
[main] INFO  c.l.d.u.impl.DefaultUpgradeReport:15 - Sending MAE from local DB .                 
```

Changes:
- Added index to PostgreSQL init.sql for new installations
- Added index to PostgreSQL test setup.sql
- Created migration guide for existing installations

Tested on production environment with 80M+ aspects.

Out of the Scope of this PR:

This could be improved for Ebean Auto-Migration:

[metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectV2.java](https://github.com/datahub-project/datahub/blob/7048e43d1df1865ed65e3b347ee5275aa59b9712/metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectV2.java#L130)

Current code (line 130-131):
```
  @Column(name = VERSION_COLUMN, nullable = false)
  private long version;
```

Proposed change:
```
  @Index  // Add this annotation
  @Column(name = VERSION_COLUMN, nullable = false)
  private long version;
```

- Why this might be needed:

The @Index annotation on line 67 is inside the PrimaryKey embedded class and the Ebean may not create a separate index for embedded fields. 
So Adding @Index to the non-embedded field (line 130) tells Ebean to create the index would help MySQL users get the index automatically

- Why it's NOT in this PR:

It requires testing Ebean DDL generation and verification on MySQL

### DIFF
--- a/docker/postgres/init.sql
+++ b/docker/postgres/init.sql
@@ -12,6 +12,7 @@ create table metadata_aspect_v2 (
 );
 
 create index if not exists timeIndex ON metadata_aspect_v2 (createdon);
+create index if not exists idx_version ON metadata_aspect_v2 (version);
 
 insert into metadata_aspect_v2 (urn, aspect, version, metadata, createdon, createdby) values(
   'urn:li:corpuser:datahub',
--- a/docs/how/restore-indices-performance.md
+++ b/docs/how/restore-indices-performance.md
@@ -1,0 +1,195 @@
+# Improving RestoreIndices Performance
+
+If you're experiencing slow RestoreIndices performance on large datasets (>10M aspects), you may need to add a database index.
+
+---
+
+## Symptoms
+
+- RestoreIndices job takes many hours (>10 hours)
+- Database CPU usage is high during RestoreIndices
+- Slow queries on `metadata_aspect_v2` table
+- Logs show busy connections with queries like:
+  ```
+  stmt[select count(*) from metadata_aspect_v2 t0 where t0.version = ?]
+  busySeconds[180]
+  ```
+
+---
+
+## Root Cause
+
+The RestoreIndices process runs this query for each batch:
+
+```sql
+SELECT COUNT(*) FROM metadata_aspect_v2 WHERE version = ?
+```
+
+Without an index on the `version` column, this requires a full table scan, making it extremely slow on large datasets.
+
+**Evidence from production:**
+
+- Query time: 180+ seconds (3 minutes) per batch
+- On 80M+ aspects: 21+ hours total time
+- High database CPU usage
+
+---
+
+## Solution
+
+Add an index on the `version` column to the `metadata_aspect_v2` table.
+
+### For New Installations
+
+**PostgreSQL:**
+
+If you're using the Docker setup, the index is automatically created from the schema file (`docker/postgres/init.sql`).
+
+**MySQL:**
+
+If you're using MySQL, the schema is managed by Ebean. For now, you'll need to add the index manually (see below). Future versions may include automatic index creation via Ebean migrations.
+
+### For Existing Installations
+
+If you already have DataHub running and want to add the index:
+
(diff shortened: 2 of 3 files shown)


max-datahub:
Hi @vejeta,
I believe the creation of an index on the version column is already being handled by the new system-update job during table creation:
`CREATE INDEX IF NOT EXISTS versionIndex ON metadata_aspect_v2 (version);`
https://github.com/datahub-project/datahub/blob/master/datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/sqlsetup/PostgresDatabaseOperations.java#L147

In your case, I would suggest you simply create those indices manually.

vejeta:
> Hi @vejeta, I believe the creation of an index on the version column is already being handled by the new system-update job during table creation: `CREATE INDEX IF NOT EXISTS versionIndex ON metadata_aspect_v2 (version);` https://github.com/datahub-project/datahub/blob/master/datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/sqlsetup/PostgresDatabaseOperations.java#L147
> 
> In your case, I would suggest you simply create those indices manually.

Thanks @max-datahub, since I discovered this on MySQL, is there a similar solution for MySQL, I don't see them in:
https://github.com/datahub-project/datahub/blob/master/datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/sqlsetup/MySqlDatabaseOperations.java

max-datahub:
@vejeta MySQL syntax can be found here: 
`INDEX versionIndex (version)`
https://github.com/datahub-project/datahub/blob/master/datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/sqlsetup/MySqlDatabaseOperations.java#L112

vejeta:
Ah, makes sense @max-datahub , closing this PR then. Thanks for the quick review!

PR #16801, opened 2026-03-26
feat(ingestion/snowflake): Mark cluster keys are partitioning keys

alisa-aylward-toast:
Added "isPartitioningKey": true to all fields that have clustering keys in Snowflake. To do this, we have to parse the function LINEAR()

Thank you for contributing to DataHub!

Allowed Types in PR Title: _feat_, _fix_, _refactor_, _docs_, _test_, _perf_, _style_, _build_, _ci_, _chore_

-->


metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema_gen.py
```diff
@@ -1634,6 +1661,7 @@ def gen_schema_metadata(
                             else None
                         )
                     ),
+                    isPartitioningKey=col.name in clustering_columns,
```
askumar27: issue (blocking): `clustering_columns` contains uppercased names, but `col.name` is raw from Snowflake's `COLUMN_NAME`. 

The existing `isPartOfKey` comparison at L1632 handles this with `col.name.upper()` for semantic views — same pattern needed here.

  ```suggestion
                      isPartitioningKey=col.name.upper() in clustering_columns,

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema_gen.py
```diff
@@ -1516,6 +1516,29 @@ def gen_column_tags_as_structured_properties(
                 ),
             )
 
+    def parse_clustering_key(self, clustering_key: str) -> List[str]:
+        """
+        Parse clustering key string and return list of Snowflake column nmes objects.
```
askumar27: suggestion (non-blocking): Two small things in the docstring:
1. "column nmes objects" → "column names"
2. The examples show lowercase output ('yyyymmdd') but the code uppercases everything, so the actual return would be 'YYYYMMDD'

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema_gen.py
```diff
@@ -1588,6 +1611,10 @@ def gen_schema_metadata(
         )
         dataset_urn = self.identifiers.gen_dataset_urn(dataset_name)
 
+        clustering_columns = set()
+        if isinstance(table, SnowflakeTable) and table.clustering_key:
+            clustering_columns = set(self.parse_clustering_key(table.clustering_key))
```
askumar27, line 1616: suggestion (minor): A debug log here would make it much easier to troubleshoot when users report clustering keys not showing up.

askumar27 (changes requested):
Thanks @alisa-aylward-toast for tagging clustering keys as partitioning keys is a great addition and something users have been asking for. The implementation is clean and well-scoped. Left some minor comments.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema_gen.py
```diff
@@ -1516,6 +1516,29 @@ def gen_column_tags_as_structured_properties(
                 ),
             )
 
+    def parse_clustering_key(self, clustering_key: str) -> List[str]:
```
sgomezvillamor, line 1519: Are the formats listed in the examples all possible cases, or just the ones currently covered? Either way, more may be added in the future.

For that reason, this parsing should be wrapped in a `try/except`. So a parsing failure doesn't blow up the ingestion of the asset: catch the exception, log it, and return an empty list.

Also, could you also add unit tests for this new method?

sgomezvillamor (commented):
LGTM
Just left a couple of minor comments.
Once addressed we can merge

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema_gen.py
treff7es, line 1519: Heads-up that this file already has a robust sqlglot-based helper at L1912 — `_extract_columns_from_expression` — that we use for view/lineage column extraction. It handles the trickier Snowflake clustering expressions that a naive `str.split(',')` misses, e.g.:

- `LINEAR(SUBSTR(col, 1, 4))` → split gives `["SUBSTR(COL", "1", "4)"]` ❌ vs sqlglot → `["COL"]` ✅
- `LINEAR(to_date(c1), c2)` → split gives `["TO_DATE(C1)", "C2"]` ❌ vs sqlglot → `["C1", "C2"]` ✅
- `LINEAR(v:click_info::int)` (VARIANT path) — sqlglot handles, split doesn't

Snowflake's `INFORMATION_SCHEMA.TABLES.CLUSTERING_KEY` can contain any of these in production, so reusing the existing helper would make this more robust. It also addresses @sgomezvillamor's try/except concern — the helper already wraps the parse in try/except + `logger.warning`. If you go that route, this whole method goes away (see suggestion at the call site below).

One small bug worth flagging if you do keep this method: the case-insensitive prefix check (`clustering_key.upper().startswith("LINEAR(")`) combined with the case-sensitive slice (`clustering_key[7:-1]`) happens to work for `LINEAR(` and `linear(` because both are 7 chars, but for mixed-case input like `Linear(C1)` the slice yields `ear(C1` — silently wrong. A regex like `re.match(r"^LINEAR\((.*)\)$", s, re.IGNORECASE)` would be safer.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema_gen.py
```diff
@@ -1588,6 +1611,15 @@ def gen_schema_metadata(
         )
```
treff7es, line 1611: Following on from the helper suggestion above — if you reuse `_extract_columns_from_expression`, this whole block can collapse to roughly:

```python
clustering_columns: Set[str] = set()
if isinstance(table, SnowflakeTable) and table.clustering_key:
    clustering_columns = {
        col for _, col in self._extract_columns_from_expression(table.clustering_key)
    }
```

The helper already returns uppercase column names, handles parse failures gracefully (try/except + `logger.warning`), and is battle-tested for Snowflake SQL in this same file (used at L2040 and L2287). That removes the magic `7`, the `set(list(...))` double-wrap, and the slice fragility in one go.

treff7es (commented):
Hi @alisa-aylward-toast, thanks so much for tackling this! 🙏 The intent is spot on — `isPartitioningKey` is exactly the right DataHub field for surfacing Snowflake clustering semantics. BigQuery uses the same field for partition columns, so this nicely unifies how partitioning shows up across SQL platforms in DataHub.

A few suggestions inline, plus two things that don't fit neatly on a single line:

**1. The golden-file diff is dominated by whitespace, not real changes.** Most of the +41k/-40k diff across the seven `*_golden.json` files is whitespace re-indentation rather than actual metadata additions. DataHub's `write_metadata_file` (in `metadata-ingestion/src/datahub/ingestion/sink/file.py`) writes each record at column 0 inside the outer `[ ]`, but the new files have records indented by 4 spaces — that's what `json.dumps(list, indent=4)` produces, which suggests the files were regenerated outside the standard pytest flow. The easiest fix:

```bash
git checkout master -- metadata-ingestion/tests/integration/snowflake/*_golden.json
cd metadata-ingestion
pytest tests/integration/snowflake/ --update-golden-files
```

This should shrink the PR from +41k/-40k to maybe a few hundred lines of real `"isPartitioningKey": true` additions — makes review much easier and avoids future merge conflicts with master.

**2. A unit test or two would lock this down nicely.** The integration fixture in `tests/integration/snowflake/common.py:288` only sets `LINEAR(COL_1)` (single column), so the multi-column path and edge cases (empty clustering key, nested expression like `LINEAR(SUBSTR(...))`, etc.) aren't actually exercised today. A handful of focused tests in `tests/unit/snowflake/` covering single-column, multi-column, empty/None, and a nested-expression case would be great.

Bonus idea: bumping one of the fixture tables in `common.py` to use `LINEAR(COL_1, COL_2)` would also surface the multi-column behavior in a golden file, which is nice for regression coverage.

Happy to help walk through any of these if useful! Once these are addressed, this is a really nice quality-of-life improvement for Snowflake users. 🎉

PR #16802, opened 2026-03-26
feat(ingestion/dbt): add GCS support for dbt core artifact file access

alfiyas-datahub:
**Summary**

Adds Google Cloud Storage (gs://) support to the dbt core source connector, allowing users to read dbt artifacts directly from GCS buckets. This mirrors the existing S3 (s3://) support and uses GCS's S3-compatible XML API with HMAC key authentication, so no new dependencies are required.
Introduce GCSConnectionConfig with HMAC credentials that adapts to an AwsConnectionConfig for boto3 S3-compatible access
Add gcs_connection config field to DBTCoreConfig with validation that gs:// URIs require a gcs_connection
Extend load_file_as_json and _expand_run_results_paths with GCS branches for file loading and glob expansion
Refactor _expand_s3_glob into generic _expand_object_store_glob shared by both S3 and GCS
Add optional endpoint_url override on GCSConnectionConfig for testing with S3-compatible servers


metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_core.py
```diff
@@ -48,11 +49,38 @@
...
+from datahub.ingestion.source.gcs.gcs_utils import is_gcs_uri
 
 logger = logging.getLogger(__name__)
 
 _GLOB_CHARACTERS = frozenset("*?[]")
 
+GCS_ENDPOINT_URL = "https://storage.googleapis.com"
+
+
+class GCSConnectionConfig(ConfigModel):
```
alokr-dhub: wouldn't it be better if we move this to `datahub/ingestion/source/common` like other Auth configs?

treff7es: 🟡 **Architecture — reuse the existing GCS auth building blocks instead of a parallel config.**

This repo already solves "GCS over the S3-compatible API" in `src/datahub/ingestion/source/gcs/gcs_source.py`, and `GCSConnectionConfig` re-creates parts of it:

| Here (`dbt_core.py`) | Already exists in `gcs_source.py` |
|---|---|
| `GCS_ENDPOINT_URL = "https://storage.googleapis.com"` | **identical** constant (line 55) |
| `hmac_access_id: str` / `hmac_access_secret: SecretStr` | `HMACKey` ConfigModel — **identical fields** (line 130) |
| HMAC-only | `GCSAuthType` enum: HMAC + Workload Identity Federation + Workload Identity / ADC (line 124) |
| `get_s3_compatible_connection()` | the HMAC branch of `create_equivalent_s3_config()` (line 308) |

The established pattern there is: *GCS-specific auth → build an `AwsConnectionConfig` pointed at the GCS endpoint → hand it to the shared S3 code* — exactly what this method does for the HMAC case. Two asks:

1. **Reuse `HMACKey` and `GCS_ENDPOINT_URL`** rather than redefining them, for a single source of truth. Caveat: `gcs_source.py` does a top-level `from google.auth.credentials import Credentials`, so importing directly from it would drag `google-auth` into the dbt import path (not guaranteed installed for dbt-only setups). The clean fix is to relocate the lightweight pieces (`HMACKey`, the endpoint constant, `GCSAuthType`) into a dependency-free module like `gcs/gcs_utils.py` and import from there in both places — that's the only change that touches `gcs_source.py`, and it's a mechanical move of definitions.

2. **Shape `gcs_connection` like `GCSSourceConfig`** (`auth_type` + `credential`) even if this PR stays HMAC-only, so WIF / Workload Identity (the keyless, GKE-recommended options) can be added later without a breaking config change. Extracting a shared `to_aws_connection()` translator would let dbt get all three auth types for free; that can be a follow-up and does **not** need to touch `gcs_source.py`.

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_core.py
```diff
@@ -48,11 +49,38 @@
...
+    """GCS connection using HMAC keys, accessed via the S3-compatible XML API."""
+
+    hmac_access_id: str = Field(
+        description="GCS HMAC access ID. See https://cloud.google.com/storage/docs/authentication/hmackeys",
+    )
+    hmac_access_secret: SecretStr = Field(
+        description="GCS HMAC access secret.",
+    )
+    endpoint_url: Optional[str] = Field(
```
treff7es: ℹ️ **Suggestion (optional):** consider `endpoint_url: str = Field(default=GCS_ENDPOINT_URL, ...)` and use `self.endpoint_url` directly in `get_s3_compatible_connection()`. This makes "endpoint is always non-empty" a type-level guarantee and removes the `or GCS_ENDPOINT_URL` runtime branch. Trade-off: callers can no longer pass `None` to mean "default" — arguably clearer. Low priority.

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_core.py
```diff
@@ -48,11 +49,38 @@
...
+        default=None,
+        description="Override the GCS S3-compatible endpoint URL. "
+        "Defaults to https://storage.googleapis.com. Useful for testing with local S3-compatible servers.",
+    )
+
+    def get_s3_compatible_connection(self) -> AwsConnectionConfig:
```
treff7es: 🟡 **Two things on this method (see also the architecture comment on the class above):**

**Perf — rebuilds a fresh `AwsConnectionConfig` (and downstream a fresh boto client) on every call.** It's invoked separately for manifest, catalog, sources, each run-results file, and each glob — so N artifacts → N+ clients. The S3 path doesn't have this because it shares one `self.config.aws_connection` instance. Consider memoizing, e.g.:
```python
@functools.cached_property
def s3_compatible_connection(self) -> AwsConnectionConfig:
    return AwsConnectionConfig(...)
```
and have call sites use `gcs_connection.s3_compatible_connection`. (Or resolve once into a local and pass the `AwsConnectionConfig` down.)

**Tests — this method's behavior is never asserted.** It's the correctness core of the feature (`aws_region="auto"`, `endpoint_url or GCS_ENDPOINT_URL`, HMAC mapping) but is only mocked-around in `test_load_file_as_json_gcs`. Add a direct test: construct the connection config with and without `endpoint_url`, then assert `aws_endpoint_url` (default + override) and `aws_region == "auto"`.

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_core.py
```diff
@@ -720,28 +763,42 @@ def test_connection(config_dict: dict) -> TestConnectionReport:
 
     @staticmethod
     def load_file_as_json(
-        uri: str, aws_connection: Optional[AwsConnectionConfig]
+        uri: str,
+        aws_connection: Optional[AwsConnectionConfig],
+        gcs_connection: Optional[GCSConnectionConfig] = None,
     ) -> Dict:
         if re.match("^https?://", uri):
             return json.loads(requests.get(uri).text)
```
treff7es: 🟡 **`requests.get(uri)` has no `.raise_for_status()` or `timeout`.** A 4xx/5xx body (an HTML error page, an S3/GCS `AccessDenied` XML body, a login redirect) flows straight into `json.loads`, surfacing as a cryptic `JSONDecodeError` instead of the real HTTP error; and a hung endpoint blocks ingestion indefinitely. Pre-existing, but this PR is actively editing this method so it's cheap to fix here:
```python
resp = requests.get(uri, timeout=30)
resp.raise_for_status()
return resp.json()
```

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_core.py
```diff
@@ -720,28 +763,42 @@ def test_connection(config_dict: dict) -> TestConnectionReport:
...
             response = aws_connection.get_s3_client().get_object(
                 Bucket=u.netloc, Key=u.path.lstrip("/")
             )
             return json.loads(response["Body"].read().decode("utf-8"))
+        elif is_gcs_uri(uri):
+            u = urlparse(uri)
+            if not gcs_connection:
+                raise ValueError(f"GCS connection required for GCS URI: {uri}")
+            s3_compat = gcs_connection.get_s3_compatible_connection()
+            response = s3_compat.get_s3_client().get_object(
```
treff7es: ℹ️ **IO errors lose URI context.** `get_object(...)` can raise `ClientError` (NoSuchKey, AccessDenied, SignatureDoesNotMatch — common with GCS HMAC setups) and propagates raw. `load_file_as_json` is called directly for `manifest_path`/`catalog_path`/`sources_path` with no surrounding try/except, so an HMAC typo aborts the run with a bare botocore traceback. Consider wrapping (symmetric for S3 and GCS) to re-raise with the URI:
```python
raise ValueError(f"Failed to read {uri} from object store: {e}") from e
```
Not a GCS-vs-S3 regression — the two branches are consistent.

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_core.py
```diff
@@ -807,6 +866,35 @@ def _expand_run_results_paths(self) -> List[str]:
...
+                if not gcs_paths:
+                    self.report.warning(
+                        title="GCS glob pattern matched no objects",
+                        message=f"Pattern '{path}' did not match any GCS objects",
+                    )
+                else:
+                    logger.info(
+                        f"GCS glob pattern '{path}' expanded to {len(gcs_paths)} file(s)"
+                    )
+                expanded_paths.extend(gcs_paths)
```
treff7es: 🟡 **DRY + test gap on this GCS block.**

**DRY:** this block is a near-verbatim copy of the S3 block above (~25 lines), differing only by scheme string, predicate (`is_gcs_uri`), connection accessor, and the word "GCS"/"S3" in messages. Consider driving both from a small table and extracting the shared body, e.g.:
```python
_OBJECT_STORE_GLOBS = [
    ("s3", is_s3_uri, lambda cfg: cfg.aws_connection, lambda cfg: cfg.aws_connection),
    ("gs", is_gcs_uri, lambda cfg: cfg.gcs_connection, lambda cfg: cfg.gcs_connection.get_s3_compatible_connection()),
]
```
then one dispatch loop calls a shared helper. Collapses ~50 lines to ~20 and makes a future `az://` a one-line entry. Keep the exact `title`/`message` strings to stay golden-stable.

**Tests:** this is the largest new code block and has no end-to-end coverage. `test_expand_object_store_glob_gcs` only tests the static helper with a `MagicMock` — it never goes through `_expand_run_results_paths`, so scheme dispatch, the `get_s3_compatible_connection()` wiring, and `expanded_paths.extend` are untested. S3 has `test_expand_run_results_paths_s3_glob` + an error-reports-failure test + a missing-connection test. Please mirror those three for GCS: success returning `gs://...` paths; paginator raises → `report.failure` contains `"GCS glob expansion failed"`; no match → `report.warning`.

metadata-ingestion/tests/unit/dbt/test_dbt_source.py
```diff
@@ -3117,6 +3121,173 @@ def test_load_file_as_json_s3():
...
+        "manifest_path": "dummy_path",
+        "catalog_path": "dummy_path",
+        "target_platform": "bigquery",
+        "run_results_paths": ["gs://bucket/results/*/run_results.json"],
+    }
+    with pytest.raises(ValidationError, match="provide gcs_connection"):
+        DBTCoreConfig.model_validate(config_dict)
+
+
+def test_run_results_gcs_glob_valid_config():
```
treff7es: ℹ️ **Trivial test — consider removing.** This asserts `config.run_results_paths == [...]`, i.e. that pydantic stored the list it was given — it tests pydantic, not PR logic. `test_dbt_gcs_config` already proves a valid `gcs_connection` is accepted. (The S3 side has the same redundant pair, so this matches an existing pattern — but it adds no value.)

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_core.py
alfiyas-datahub: Done

metadata-ingestion/tests/integration/s3/test_s3.py
```diff
@@ -253,7 +253,7 @@ def test_data_lake_gcs_ingest(
 
     config_dict["run_id"] = source_file
 
-    with patch("datahub.ingestion.source.gcs.gcs_source.GCS_ENDPOINT_URL", None):
+    with patch("datahub.ingestion.source.gcs.gcs_utils.GCS_ENDPOINT_URL", None):
```
treff7es, line 256: 🟡 **Regression: this patch no longer affects the code under test.**

When `GCS_ENDPOINT_URL` moved into `gcs_utils.py`, `gcs_source.py` started doing `from ...gcs_utils import GCS_ENDPOINT_URL` and references the bare name `GCS_ENDPOINT_URL` directly (lines 278 & 310). That `from ... import` binds the value into the **`gcs_source` namespace** at import time, so patching `gcs_utils.GCS_ENDPOINT_URL` here rebinds the name in `gcs_utils` only — `gcs_source` still reads its own binding (`https://storage.googleapis.com`). The `None` override is silently ineffective.

Patch where the name is *looked up*:
```python
with patch("datahub.ingestion.source.gcs.gcs_source.GCS_ENDPOINT_URL", None):
```
Before the move this test patched `gcs_source.GCS_ENDPOINT_URL` and worked correctly. Worth confirming this GCS integration test actually exercises the local/moto endpoint in CI — if it's green today it may be passing for the wrong reason.

metadata-ingestion/tests/unit/dbt/test_dbt_source.py
```diff
@@ -3121,6 +3125,208 @@ def test_load_file_as_json_s3():
...
+        "catalog_path": "dummy_path",
+        "target_platform": "bigquery",
+        "run_results_paths": ["gs://bucket/results/*/run_results.json"],
+    }
+    with pytest.raises(ValidationError, match="provide gcs_connection"):
+        DBTCoreConfig.model_validate(config_dict)
+
+
+def test_load_file_as_json_gcs():
+    from datahub.ingestion.source.common.gcs_connection_config import (
```
treff7es, line 3162: 🟡 **Late/inline imports.** `GCSConnectionConfig` / `HMACKey` (and `GCS_ENDPOINT_URL` at 3295, `ClientError` at 3254) are imported inside the test bodies. Per the testing standard, function-scoped imports should be hoisted to the top of the file (or carry a comment justifying why they must be late). Please move at least the `gcs_connection_config` / `gcs_utils` imports up top. (The `botocore.ClientError` ones match the existing S3 tests at 2825/2868, so those are lower priority.)

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_core.py
```diff
@@ -769,10 +809,48 @@ def _expand_s3_glob(uri: str, aws_connection: AwsConnectionConfig) -> List[str]:
...
+        path: str,
+        connection: Optional[AwsConnectionConfig],
+        scheme: str,
+        *,
+        store_label: str,
+        connection_field: str,
+    ) -> List[str]:
+        # store_label is the user-facing storage name ("S3"/"GCS"); connection_field
+        # is the recipe key that supplies credentials ("aws_connection"/"gcs_connection").
+        connection_label = connection_field.split("_")[0].upper()
```
treff7es, line 828: ℹ️ **Nit (optional):** `connection_label` is reconstructed from `connection_field.split("_")[0].upper()` to produce `"AWS"`/`"GCS"`, but the function already receives `store_label`. Having two near-identical labels for the same message set is a little confusing — consider using `store_label` throughout, or passing the connection label explicitly. Cosmetic.

treff7es (commented):
Re-review (round 2). Thanks for the thorough rework — all 8 comments from the previous round are addressed: config moved to `common/`, `HMACKey`/`GCS_ENDPOINT_URL` hoisted to `gcs_utils.py` and reused by `gcs_source.py`, connection memoized via `cached_property`, `requests.get` now has `timeout`+`raise_for_status`, IO errors wrapped with URI context, glob logic deduplicated into `_expand_cloud_glob`, and the trivial config-equality test replaced with real validator coverage.

One regression slipped in via the refactor (the integration-test patch target), plus two minor items below. Only the first needs a fix before merge.

metadata-ingestion/tests/integration/s3/test_s3.py
alfiyas-datahub: Done

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_core.py
alfiyas-datahub: I removed the logic that derived the connection type from the field name and now pass it explicitly from each caller (for example, "AWS" for S3 and "GCS" for GCS). This makes the code clearer and less error-prone, while keeping the user-facing error messages unchanged, so the existing tests still pass.

treff7es (approved):

PR #16807, opened 2026-03-26
feat(ingest/powerbi): support NativeQuery queries on BigQuery

ligfx:
powerbi: reorganize NativeQueryLineage.get_db_name

feat(ingest/powerbi): support NativeQuery queries on BigQuery

### DIFF
--- a/metadata-ingestion/tests/integration/powerbi/test_m_parser.py
+++ b/metadata-ingestion/tests/integration/powerbi/test_m_parser.py
@@ -684,6 +684,39 @@ def test_redshift_native_query():
     )
 
 
+def test_bigquery_native_query():
+    table = powerbi_data_classes.Table(
+        name="mytable",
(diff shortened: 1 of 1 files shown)


metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/pattern_handler.py
kyungsoo-datahub, line 1337: I'm wondering if it's true for BigQuery as well.

ligfx: What do you mean?

kyungsoo-datahub: I misunderstood some here. Sorry for the confusion.

PR #16808, opened 2026-03-27
fix(graph): drain slice futures before PIT/scroll cleanup

david-leifker:
- Add sliceFutureDrainTimeoutSeconds config and cancelAndDrainSliceFutures helper
- Fix per-slice deadline (wall-clock) for PIT and scroll pagination
- Hoist slice futures list in PIT DAO so finally can cancel before PIT delete
- Add regression tests and update test fixtures for required config

Made-with: Cursor

### DIFF
--- a/metadata-io/src/main/java/com/linkedin/metadata/graph/elastic/GraphQueryBaseDAO.java
+++ b/metadata-io/src/main/java/com/linkedin/metadata/graph/elastic/GraphQueryBaseDAO.java
@@ -49,12 +49,14 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import lombok.Value;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.tuple.Pair;
@@ -93,7 +95,7 @@ public abstract class GraphQueryBaseDAO implements GraphQueryDAO {
 
   protected abstract SearchClientShim<?> getClient();
 
-  protected abstract List<LineageRelationship> searchWithSlices(
+  protected abstract LineageSliceFetchResult searchWithSlices(
       @Nonnull OperationContext opContext,
       @Nonnull QueryBuilder query,
       LineageGraphFilters lineageGraphFilters,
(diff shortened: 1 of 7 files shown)


alexsku (commented):
## PR Summary

**PR**: #16808 — fix(graph): drain slice futures before PIT/scroll cleanup
**Author**: david-leifker
**Base**: master ← fix/graph-slice-pit-drain
**Type**: behavior change (bug fix)
**Size**: +424 / -38 across 10 files

This PR fixes two interacting bugs in PIT-based lineage graph search. **Bug 1**: A shared PIT was deleted in the `finally` block while slice worker threads were still running, because `sliceFutures` was scoped inside the `try` block and invisible to `finally`. **Bug 2**: The per-page `remainingTime` update was a no-op (`t1 - (t2 - R) ≈ R`), so slices never timed out and paginated indefinitely — triggering Bug 1 reliably on non-trivial graphs. The fix moves `sliceFutures` to outer scope, adds a `cancelAndDrainSliceFutures` helper in the `finally` block, and replaces the no-op timeout with an absolute deadline comparison. The same fixes are applied consistently to `GraphQueryElasticsearch7DAO`.

**Key invariants:**
1. The shared PIT must not be deleted while any slice worker is still executing `client.search()`.
2. Each slice must respect its time budget and stop paginating when the deadline passes.
3. PIT cleanup must always execute (via `finally`) regardless of how `processSliceFutures` exits.
4. The `sliceFutureDrainTimeoutSeconds` config must always be set (enforced via `Objects.requireNonNull`).

---

## Risk Assessment

| | |
|---|---|
| **Risk** | Low-Medium |
| **Blast radius** | Impact/lineage search queries (getSearchAcrossLineageCounts, impact analysis) |
| **Rollback** | Safe — no migrations, no data changes, purely runtime behavior |
| **Rollout** | Ship it — fixes a production bug, configurable drain timeout provides a knob |

---

## Blocking Issues

None found.

## High-Priority Issues

None found.

## Other Issues

- **[MEDIUM / correctness]** `GraphQueryBaseDAO.java:1589-1597` — `cancelAndDrainSliceFutures` — The drain mechanism is likely ineffective at its stated goal. `CompletableFuture.cancel(true)` marks futures as immediately "done" (with `CancellationException`) without actually interrupting the underlying thread. This means `CompletableFuture.allOf(pending).get(drainTimeout)` will throw `CancellationException` immediately (caught by the catch block) rather than waiting for worker threads to finish their in-flight `client.search()` calls. The threads continue executing against a PIT that may be deleted moments later. **Mitigated** by the deadline fix (Bug 2) which dramatically shrinks the race window, and by OpenSearch's own PIT keepalive. Fix: Consider waiting with `allOf().get(drainTimeout)` *before* calling `cancel(true)`, or use `orTimeout`/`completeOnTimeout` to give futures a chance to finish naturally. Alternatively, swap `CompletableFuture.supplyAsync` for `ExecutorService.submit` which produces `FutureTask` where `cancel(true)` actually interrupts the thread.

- **[MEDIUM / edge-case]** `GraphQueryBaseDAO.java:1626` — `processSliceFutures` — `futureTimeout = Math.max(1, remainingTime / 1000)` uses integer division which truncates. If `remainingTime` is 1999ms, `futureTimeout` becomes 1 second rather than ~2. This was pre-existing (not introduced by this PR) but interacts with the new deadline: a tight `remainingTime` could cause the outer `future.get()` to timeout before the inner deadline check fires, triggering Path A (hard timeout + cancellation) instead of the gentler deadline-based exit. Not a correctness bug, but could cause unnecessary `TimeoutException` noise in logs.

- **[LOW / dx]** The test `testSliceDeadlineTerminatesPaginationAfterRemainingTimeElapsed` relies on a 900ms sleep to push past an 800ms deadline while staying under a 1s outer timeout. This leaves only 100ms of margin, which could be tight on slow CI machines. The `@Test(timeOut = 5000)` provides test-level protection, but flaky failures are possible. Consider widening the timing margin.

---

## What's Missing

Nothing significant. The PR covers both the PIT DAO and ES7 DAO, adds three well-targeted regression tests, adds the `sliceFutureDrainTimeoutSeconds` config with a sensible default, and updates the security properties classification test.

## Test Plan

| Invariant | Covered? | Test |
|---|---|---|
| Slices respect deadline | ✅ | `testSliceDeadlineTerminatesPaginationAfterRemainingTimeElapsed` |
| PIT cleaned up after timeout (Path A) | ✅ | `testPitCleanedUpAfterSliceTimeoutWithConcurrentBlockedSlice` |
| PIT cleaned up after budget exhaustion (Path C) | ✅ | `testPitCleanedUpWhenRemainingBudgetExhaustedBySuccessfulSlice` |
| Config must be set | Implicit | `Objects.requireNonNull` throws at runtime; all test configs set the value |
| ES7 DAO has same fixes | Not directly | No new ES7-specific regression tests, but the shared `cancelAndDrainSliceFutures` is tested via PIT DAO tests |

## Questions for Author

1. On the drain mechanism: `CompletableFuture.cancel(true)` doesn't actually interrupt threads — it just marks the future as "done." This means `allOf().get(drainTimeout)` resolves immediately rather than giving threads time to finish. Was the intent to just have a best-effort attempt, or should the drain actually block until threads finish? If the latter, waiting *before* calling cancel would work (or using `ExecutorService.submit` instead of `CompletableFuture.supplyAsync`).

2. Have you considered adding a metric for how often the drain timeout is actually hit (vs. futures completing before the drain)? This would help monitor the effectiveness of the fix in production.

---

PR #16809, opened 2026-03-27
feat(dagster): Emit StatusClass aspect during asset ingestion to handle soft deleted assets

bryanprosser-acryl:
Emit StatusClass aspect during asset ingestion to handle soft deleted assets and additional unit test

### DIFF
--- a/metadata-ingestion-modules/dagster-plugin/tests/unit/test_dagster.py
+++ b/metadata-ingestion-modules/dagster-plugin/tests/unit/test_dagster.py
@@ -587,4 +587,6 @@ def test_emit_asset_emits_status_aspect() -> None:
         and isinstance(m.aspect, StatusClass)
     ]
     assert len(status_mcps) == 1
-    assert status_mcps[0].aspect.removed is False
+    status_aspect = status_mcps[0].aspect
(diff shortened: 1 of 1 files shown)


metadata-ingestion-modules/dagster-plugin/src/datahub_dagster_plugin/client/dagster_generator.py
```diff
@@ -809,6 +810,13 @@ def emit_asset(
         for mcp in dataset.generate_mcp():
```
sgomezvillamor, line 810: is there any reason for not adding the `Status` aspect in the `generate_mcp` method?

https://github.com/datahub-project/datahub/blob/master/metadata-ingestion/src/datahub/api/entities/dataset/dataset.py#L493

I'm looking at other entities in `datahub.api.entities` module and I see some of them include the `Status` aspect in the `generate_mcp` while others don't.

@treff7es do you know any reason for this inconsistency?

treff7es: Unfortunately I was not involed in the api 2.0 so I don't know :(

PR #16819, opened 2026-03-27
fix(ingest/powerbi): extend SQL cleanup to strip T-SQL control statements

kyungsoo-datahub:
M-Queries containing USE, SET, GO, and SELECT INTO #temp cause sqlglot to fail with an invalid expression parse error, silently losing all lineage for those tables.

The existing remove_drop_statement() only handled DROP TABLE. Rename it to remove_tsql_control_statements() and extend it to also strip USE <db>, SET <option> ON|OFF, GO batch separators, and SELECT INTO #temp (only the INTO clause is stripped so FROM/WHERE lineage is preserved). All patterns are line-anchored to avoid false matches inside valid SQL identifiers.

remove_drop_statement() is kept as a backward-compatible shim. 

Builds on top of #mm--powerbi-nativequery-bigquery.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/native_sql_parser.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/native_sql_parser.py
@@ -85,16 +85,23 @@ def remove_tsql_control_statements(query: str) -> str:
         r"^\s*SET\s+\w+\s+(?:ON|OFF)\s*;?\s*$",
         # GO — T-SQL batch separator
         r"^\s*GO\s*$",
-        # INTO #<temp_table> within SELECT … INTO — redirects output to temp table;
-        # strip only the INTO clause so FROM/WHERE lineage remains parseable.
-        # ##name = global temp table, #name = local temp table.
-        r"\s+INTO\s+##?\w+",
     ]
 
     new_query = query
 
     for pattern in patterns:
         new_query = re.sub(pattern, "", new_query, flags=re.IGNORECASE | re.MULTILINE)
+
+    # SELECT … INTO #<temp> — strip only the INTO clause so FROM/WHERE lineage remains
+    # parseable. Anchored to SELECT so INSERT INTO and MERGE INTO are never matched.
+    # [^;\n] stops at semicolons and uses \n(?!\s*\n) to not cross blank lines,
+    # preventing a SELECT in one statement from reaching an INSERT INTO in the next.
+    new_query = re.sub(
+        r"(SELECT\b(?:[^;\n]|\n(?!\s*\n))*?)\s+INTO\s+##?\w+",
+        r"\1",
+        new_query,
+        flags=re.IGNORECASE,
+    )
 
     # Leading semicolon before WITH — T-SQL defensive pattern (;WITH ...) used to ensure
     # the previous statement is terminated before a CTE. Strip only the semicolon,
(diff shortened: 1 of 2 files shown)


metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/native_sql_parser.py
```diff
@@ -108,15 +135,32 @@ def parse_custom_sql(
...
-        platform_instance=platform_instance,
-        env=env,
-        graph=ctx.graph,
-    )
+    # Blank-line-separated SELECTs appear after DROP TABLE stripping removes the DDL
+    # between statements, leaving no semicolons. Insert them so the multi-statement
+    # check below treats them correctly.
+    normalized = re.sub(r"\n{2,}(\s*SELECT\b)", r";\n\n\1", query, flags=re.IGNORECASE)
+
+    if ";" in normalized:
```
treff7es, line 152: The heuristic ";" in normalized will false-positive on semicolons inside string literals (e.g., WHERE name = 'foo;bar'), routing a single-statement query to create_lineage_from_sql_statements. This could silently produce empty or partial lineage.

metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/native_sql_parser.py
```diff
@@ -69,35 +70,61 @@ def get_tables(native_query: str) -> List[str]:
...
+        # identifiers (USE Reports) and bracketed ones (USE [Reports])
+        r"^\s*USE\s+\S+\s*$",
+        # SET <option> ON|OFF — T-SQL session-level options (NOCOUNT, QUOTED_IDENTIFIER, etc.)
+        r"^\s*SET\s+\w+\s+(?:ON|OFF)\s*;?\s*$",
+        # GO — T-SQL batch separator
+        r"^\s*GO\s*$",
+        # INTO #<temp_table> within SELECT … INTO — redirects output to temp table;
+        # strip only the INTO clause so FROM/WHERE lineage remains parseable.
+        # ##name = global temp table, #name = local temp table.
+        r"\s+INTO\s+##?\w+",
```
treff7es: This pattern matches anywhere in the query, not just after SELECT. It would incorrectly strip INSERT INTO ##metrics or MERGE INTO ##staging, losing the INTO clause from valid DML statements.

Either anchor the pattern to require line-start context, or add a preceding-keyword check. At minimum, add tests confirming INSERT INTO dbo.Table is NOT stripped.

metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/native_sql_parser.py
kyungsoo-datahub: Thank you for the comment. `split_statements` uses a state machine that transitions to ParserState.STRING on ' and consumes all characters verbatim until the closing quote. semicolons inside string literals never trigger a split.(Ref: https://github.com/datahub-project/datahub/blob/master/metadata-ingestion/src/datahub/sql_parsing/split_statements.py#L156-L159)

metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/native_sql_parser.py
kyungsoo-datahub: Thank you for pointing out. I covered all SELECT..INTO cases and avoid false positives with INSERT..INTO and MERGE..INTO case.

PR #16834, opened 2026-03-28
feat(ingestion/powerbi): extend Oracle lineage to support TNS connections and native SQL queries

acrylJonny:
## Summary

Extends the PowerBI connector's Oracle lineage extraction with three improvements:

- **TNS alias support**: `_get_server_and_db_name` previously only handled JDBC-style `host:port/servicename` strings and returned `(None, None)` for anything else. It now handles bare hostnames and TNS aliases, returning `(hostname, None)` so the caller omits the database segment from the qualified name — consistent with how `platform_instance` is applied.
- **Native SQL via `Query=`**: `OracleLineage.create_lineage` now detects an inline `Query=` record field (e.g. `Oracle.Database("host", [Query="SELECT * FROM schema.table"])`) and routes it through the existing `parse_custom_sql` path, matching the behaviour of `MSSqlLineage`.
- **Case-insensitive server matching**: Oracle connection strings in M-Queries are case-insensitive (`ORACLE-HOST.EXAMPLE.COM` and `oracle-host.example.com` are the same server). `_get_server_and_db_name` now lowercases the input so `server_to_platform_instance` keys match regardless of capitalisation in the M-Query.
- **Column lineage case remapping**: sqlglot derives downstream column names from the upstream schema (Oracle columns stored lowercase in DataHub), but PowerBI field names come back from the API in uppercase. A new `_remap_column_lineage_to_pbi_fields` helper remaps downstream column names using a case-insensitive lookup so the generated `schemaField` URNs resolve to actual fields in the PowerBI dataset.

## Test plan

- [x] New unit tests for `_get_server_and_db_name` (TNS alias parsing), `OracleLineage.create_lineage` (hierarchical + native SQL paths), and `_remap_column_lineage_to_pbi_fields` (case remapping, no-op cases, unknown columns)
- [x] New integration tests for TNS hierarchical navigation and `Query=` native SQL through the full M-Query parser
- [x] Existing PowerBI parser test suite passes (`test_powerbi_parser.py`, `test_m_parser.py`)


sgomezvillamor:
Is there any way we could know in which scenarios missing the database will be valid and which scenarios will be invalid? We have `default_schema` in `server_to_platform_instance`, would a `default_database` make sense in some case?

In general, the docs we have about `server_to_platform_instance ` is quite vague and limited to the text below:

```
    # Optional -- This mapping is optional and only required to configure platform-instance for upstream tables
    # A mapping of PowerBI datasource's server i.e host[:port] to data platform instance.
    # :port is optional and only needed if your datasource server is running on non-standard port.
    # For Google BigQuery the datasource's server is google bigquery project name.
    # For Fabric OneLake (DirectLake lineage), use PowerBI workspace ID as the key.
    # The platform_instance typically represents the Fabric tenant identifier, matching how
    # the OneLake connector uses platform_instance to group workspaces by tenant.
```

Also, not sure if this is up to date https://docs.datahub.com/docs/generated/ingestion/sources/powerbi#m-query-pattern-supported-for-lineage-extraction

Even, the description in this PR looks interesting content for the user-facing docs.

Overall, could you take this as the opportunity to better document this?

acrylJonny:
Good call — pushed changes to address this.

On when the database segment is valid vs not: it comes down to what shape the Oracle ingestion emits. Oracle ingestion defaults to 2-part `schema.table` URNs and only produces 3-part `database.schema.table` when run with `add_database_name_to_urn: true`. On the PowerBI side, an EZ-Connect `host:port/service` connection carries the service (so we get a database segment), but a bare TNS alias / descriptor carries no database, so we'd emit 2-part. That's correct for the default Oracle setup and wrong only if the Oracle ingestion is on the 3-part shape.

So yes, a `default_database` does make sense for exactly that case, and I've added it — symmetric with `default_schema` on the Oracle entry in `server_to_platform_instance`. Set it to your Oracle ingestion's `database`/`urn_db_name` when it uses 3-part URNs; leave it unset otherwise. It's wired into both the hierarchical-navigation and inline-`Query=` paths. At least one of `default_schema`/`default_database` must be set on an Oracle entry, otherwise it's just a plain platform-instance mapping.

On the docs: agreed they were thin. I've expanded the `server_to_platform_instance` field description with the per-platform key semantics, and added a section to the PowerBI connector docs ("Matching the Oracle URN shape") that spells out the 2-part vs 3-part rule with a small table, plus the `default_schema`/`default_database` guidance. I kept the starter recipe minimal on purpose and put the detail in the prose docs.

sgomezvillamor (approved):
Thanks for the thorough follow-up — this resolves my comment.

`default_database` symmetric with `default_schema` and wired into both the hierarchical and inline-`Query=` paths is exactly what I was after, and the expanded docs (the 2-part vs 3-part URN rule with the table, plus when to set each knob) make `server_to_platform_instance` much clearer than what we had.

Also nice touches beyond the ask: surfacing the missing schema/table case as a report warning instead of skipping silently, and the info nudge when a bare TNS alias produces a 2-part URN so operators on 3-part know to set `default_database`. Tests cover the new paths well.

LGTM.

---

PR #16840, opened 2026-03-30
fix(ingest/redshift): fix orphaned containers and missing dataset linking when include_tables=False with profiling

treff7es:
## Summary

Two independent Redshift ingestion bugs:

1. **Profiled datasets appeared outside the container hierarchy.** With `include_tables=false` + `profiling.enabled=true`, only `DatasetProfile` aspects were emitted for tables. The `container` and `dataPlatformInstance` aspects (normally emitted by `gen_dataset_workunits`, which is gated on `include_tables=true`) were missing, so profiled datasets appeared at the wrong place in the UI — orphaned from their schema, and outside the configured platform instance when one was set.
2. **Lineage URNs contained the literal string `"None"` for unqualified table refs.** `RedshiftSqlLineage._get_sources_from_query` wrapped `self.config.default_schema` with `str(...)`, which turned a runtime `None` into `"None"` and produced URNs like `urn:li:dataset:(...,db.none.tbl,PROD)`.

### `RedshiftProfiler.get_workunits` (`profile.py`)

When `include_tables=false` and a profile request is produced for a table, also emit:
- `container` aspect linking the dataset to its schema container (`add_table_to_schema_container`)
- `dataPlatformInstance` aspect via `get_dataplatform_instance_aspect`

Both are emitted **after** `get_profile_request(...)` returns a non-`None` request, so excluded tables (`table_pattern` denial, `profile_pattern` denial, zero columns, size/row limits, recently profiled via stateful ingestion) do **not** get phantom dataset entities in DataHub. The dataset URN is derived from `profile_request.pretty_name` so it matches the URN used by the profile MCP itself. `gen_schema_key(...)` is computed once per schema (hoisted out of the inner per-table loop) since it depends only on db/schema/platform/instance/env.

### `RedshiftSqlLineage._get_sources_from_query` (`lineage.py`) + `RedshiftConfig.default_schema` type (`config.py`)

Drop the `str(...)` wrapper around `self.config.default_schema` so a runtime `None` is passed through unchanged. `SchemaResolver.get_urn_for_table` then drops the schema qualifier from the URN entirely (via `filter(None, [database, db_schema, table])`) instead of injecting `"None"`.

The root-cause fix is also applied at the type level: `RedshiftConfig.default_schema` is now `Optional[str]` (default `"public"`) — matching the rest of the codebase (`Sigma`, `create_lineage_sql_parsed_result`, `_TableName.qualified`, etc.) and removing the need for `# type: ignore[assignment]` in tests that exercise the `None` path. As a follow-on, `parse_alter_table_rename` accepts `Optional[str]` and raises `ValueError` if the query has no schema qualifier and `default_schema` is `None`, so callers (already wrapped in try/except) cleanly skip rather than silently emit `db.None.tbl` URNs.

## Test Plan

New file `test_redshift_profiling_containers.py` (6 tests):

| Test | Guards |
|---|---|
| `test_profiler_emits_container_aspect_when_include_tables_false` | container MCP exists on the expected dataset URN, points to expected schema container URN |
| `test_profiler_emits_dataplatforminstance_when_include_tables_false` | DPI MCP attached to expected dataset URN, exact `platform` + `instance` values |
| `test_profiler_does_not_emit_container_when_include_tables_true` | no container/DPI MCPs from the profiler when the regular path handles them |
| `test_profiler_container_links_to_correct_schema` | container URN equals `gen_schema_key("seed", ...).as_urn()` (uses non-default `"seed"` schema so a wrong-schema regression would fail) |
| `test_profiler_emits_container_but_no_dpi_without_platform_instance` | container emitted, DPI omitted when `platform_instance` is unset |
| `test_profiler_does_not_emit_container_for_excluded_table` | regression guard for the ghost-entity bug — denies via `profile_pattern`, pins `column_count=5` so the test routes through the right code path |

New tests in `test_redshift_lineage.py`:

| Test | Guards |
|---|---|
| `test_get_sources_from_query_with_none_default_schema` | exact URN comparison (`urn:li:dataset:(urn:li:dataPlatform:redshift,test.source,PROD)`); defensive `.none.` and `.None.` substring checks |
| `test_parse_alter_table_rename_none_default_schema_with_qualifier` | when the query already has a schema qualifier, `default_schema=None` is accepted and the parsed schema is used |
| `test_parse_alter_table_rename_raises_on_none_schema` | unqualified ALTER + `default_schema=None` raises `ValueError` instead of silently producing `db.None.tbl` URNs in the caller |

### Verification

- `ruff format --check` + `ruff check` + `mypy` pass on all 5 changed files
- 74 redshift unit tests pass
- Mutation-tested the new tests against deliberate code regressions (MCP emission outside the eligibility check, hardcoded wrong schema in `gen_schema_key`, removing the `ValueError` guard in `parse_alter_table_rename`) — each mutation was caught by the corresponding test

## Backward compatibility

- `include_tables=true` (default): unchanged behavior — the new emission path is gated on `not self.config.include_tables`
- `include_tables=false` + profiling enabled: profiled datasets now appear correctly under their schema containers (this is the fix)

Fixes: ING-2018


metadata-ingestion/tests/unit/redshift/test_redshift_lineage.py
```diff
@@ -117,6 +117,39 @@ def test_get_sources_from_query_with_only_table():
...
+    #
+    # Note: pydantic types `default_schema` as `str`, so this `None`
+    # assignment is intentionally type-bypassing (`# type: ignore` below).
+    # The defensive coverage matters because the bug surfaced via subclasses
+    # / runtime mutation that fed `None` into the field in production.
+    test_query = """
+        insert into target select * from source
+    """
+    lineage_extractor = get_lineage_extractor()
+    lineage_extractor.config.default_schema = None  # type: ignore[assignment]
```
treff7es: We should make sure not to silence type issues unless there is a strong reason.

Please, can you declare the type for `lineage_extractor.config.default_schema` to `Optional` ?

metadata-ingestion/src/datahub/ingestion/source/redshift/profile.py
```diff
@@ -68,9 +73,45 @@ def get_workunits(
                             )
                             # Continue, since we were unable to retrieve cheap profiling stats from svv_table_info.
                             continue
+
                     # Emit the profile work unit
                     profile_request = self.get_profile_request(table, schema, db)
                     if profile_request is not None:
```
treff7es, line 85: This should be above `for table in tables[db].get(schema, {}):` line to make sure we don't regenerate it for every table when container only depend on scheam/db

sgomezvillamor:
## On "Empty schema containers were generated"

Thanks for the thorough PR description and the effort put into the fixes!

One concern about the `process_schema` change in `redshift.py`: the stated problem — _"empty schema containers appear in the UI when nothing populates them"_ — does not actually occur in DataHub. Two independent mechanisms already prevent empty containers from being visible:

1. **Elasticsearch aggregations** ([`ESBrowseDAO.java:752–766`](https://github.com/datahub-project/datahub/blob/master/metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/query/ESBrowseDAO.java)): browse results are built from `browsePathV2` prefix-match aggregations. A container entity only produces an aggregation bucket if at least one dataset/chart/dashboard has that container in its indexed `browsePathsV2`. An empty container — one that no entity references — simply never enters the result set.

2. **Frontend filter** ([`useSidebarEntities.ts:29`](https://github.com/datahub-project/datahub/blob/master/datahub-web-react/src/app/searchV2/sidebar/useSidebarEntities.ts)): as a second layer, the sidebar explicitly drops any aggregation entry where `!!base.count` is falsy before rendering.

This means the same "empty container" situation exists in many other connectors today and users never see it. The `redshift.py` hunk is addressing a non-issue and introduces a behaviour difference from every other connector without a user-visible benefit. It would be cleaner to drop that change and keep this PR focused on the two genuine bugs (profiling orphan fix and the `"None"` lineage fix).

If there is a real symptom that motivated this (e.g. showing up in a direct container entity search, or causing churn in stateful ingestion), it would be worth documenting — but as written the PR description and the debug log message overstate the impact.

---

sgomezvillamor (commented):
## Review — remaining bugs (excluding `process_schema` empty-container hunk)

> See separate comment for why the `redshift.py` schema-container suppression change appears to address a non-issue — empty containers are already invisible in the UI via Elasticsearch aggregations and the frontend `!!base.count` filter.

---

### Bug 1 — `lineage.py`: `str(None)` → `"None"` in URNs ✅

One-line fix, correct. Dropping `str(...)` from `self.config.default_schema` lets `create_lineage_sql_parsed_result` receive `None` and properly skip the schema qualifier via its internal `filter(None, ...)` call. The regression test is solid and meaningful.

**🟡 Warning — unresolved reviewer request:** `lineage_extractor.config.default_schema = None  # type: ignore[assignment]` — `@treff7es` asked to fix the root cause by declaring `default_schema: Optional[str]` in `RedshiftConfig`. The thread was resolved without the config type being changed. Either make `default_schema` properly `Optional[str]` (preferred) or add a comment explaining why the `type: ignore` is the intentional resolution.

---

### Bug 2 — `profile.py`: orphaned profiled datasets when `include_tables=False` ✅

Correct approach: emit `container` and `dataPlatformInstance` aspects for profiled datasets only after `get_profile_request` returns non-`None`, ensuring no ghost entities are created for filtered/excluded tables. Test suite is high quality — the ghost-entity regression test in particular is well-designed.

**🔴 Blocker — unresolved reviewer request:** `gen_schema_key(...)` is still computed inside the inner `for table in ...` loop (`profile.py:83–91`). This is the change `@treff7es` requested in thread `r3137674556`, which was resolved without the code being updated. Move it above the `for table` loop — it depends only on `db`/`schema`, not on the individual table.

**🟡 Warning:** The `expected_dataset_urn` test helper constructs URNs as manual f-strings instead of calling `make_dataset_urn_with_platform_instance`. If the URN format changes the helper silently diverges from production.

**ℹ️ Suggestion:** The 9-line block comment in `profile.py` explaining the guard can be condensed to 2 lines.

---

**Verdict: two reviewer-requested changes (`gen_schema_key` hoist, `Optional[str]` type fix) were closed without being applied and should be addressed before merge.**

---

sgomezvillamor (commented):
Left a couple of comments
The most concerning is the one about preventing empty schemas to be emitted, AFAIK that shouldn't be an issue

dinesh-verma-datahub:
@sgomezvillamor I have addressed all the concerns in the latest commits (https://github.com/datahub-project/datahub/pull/16840/changes/be46d27edee5d50c85dada03388535e38187a043) and (https://github.com/datahub-project/datahub/pull/16840/changes/7b9b6012cd29102cc9f54c656ab531b8d351a483)

treff7es:
> ## Review — remaining bugs (excluding `process_schema` empty-container hunk)
> 
> > See separate comment for why the `redshift.py` schema-container suppression change appears to address a non-issue — empty containers are already invisible in the UI via Elasticsearch aggregations and the frontend `!!base.count` filter.
> 
> ---
> 
> ### Bug 1 — `lineage.py`: `str(None)` → `"None"` in URNs ✅
> 
> One-line fix, correct. Dropping `str(...)` from `self.config.default_schema` lets `create_lineage_sql_parsed_result` receive `None` and properly skip the schema qualifier via its internal `filter(None, ...)` call. The regression test is solid and meaningful.
> 
> **🟡 Warning — unresolved reviewer request:** `lineage_extractor.config.default_schema = None  # type: ignore[assignment]` — `@treff7es` asked to fix the root cause by declaring `default_schema: Optional[str]` in `RedshiftConfig`. The thread was resolved without the config type being changed. Either make `default_schema` properly `Optional[str]` (preferred) or add a comment explaining why the `type: ignore` is the intentional resolution.
> 
> ---
> 
> ### Bug 2 — `profile.py`: orphaned profiled datasets when `include_tables=False` ✅
> 
> Correct approach: emit `container` and `dataPlatformInstance` aspects for profiled datasets only after `get_profile_request` returns non-`None`, ensuring no ghost entities are created for filtered/excluded tables. Test suite is high quality — the ghost-entity regression test in particular is well-designed.
> 
> **🔴 Blocker — unresolved reviewer request:** `gen_schema_key(...)` is still computed inside the inner `for table in ...` loop (`profile.py:83–91`). This is the change `@treff7es` requested in thread `r3137674556`, which was resolved without the code being updated. Move it above the `for table` loop — it depends only on `db`/`schema`, not on the individual table.
> 
> **🟡 Warning:** The `expected_dataset_urn` test helper constructs URNs as manual f-strings instead of calling `make_dataset_urn_with_platform_instance`. If the URN format changes the helper silently diverges from production.
> 
> **ℹ️ Suggestion:** The 9-line block comment in `profile.py` explaining the guard can be condensed to 2 lines.
> 
> ---
> 
> **Verdict: two reviewer-requested changes (`gen_schema_key` hoist, `Optional[str]` type fix) were closed without being applied and should be addressed before merge.**
> 
> ---

These are actually non empty container.
The problem was if schema ingestion was disabled but profiling wasn't then it caused to have datasets without container

sgomezvillamor (approved):
All feedback addressed — approving.

- `gen_schema_key` correctly hoisted outside the per-table loop
- `default_schema: Optional[str]` in config; `# type: ignore` removed from test
- Empty-container gate in `redshift.py` removed (not a real problem — UI already hides empty containers)
- `expected_dataset_urn` test helper uses `make_dataset_urn_with_platform_instance`
- `parse_alter_table_rename` raises `ValueError` cleanly when schema can't be resolved

---

PR #16841, opened 2026-03-30
chore(spot-termination): Add spot termination configs for gms and datahub-frontend

Dutt23:
Summary
Implements graceful shutdown handling for spot instance terminations in Kubernetes environments, allowing pods to drain connections and complete in-flight requests before being killed.

Changes

1. Graceful Shutdown Handler
Added a handler that flips a readiness flag when SIGTERM is received, signaling to Kubernetes that the pod should stop receiving new requests while existing connections drain.

2. Spring Component Shutdown Timeouts (GMS Backend)
Configured Spring Boot graceful shutdown with a 65-second timeout per shutdown phase (spring.lifecycle.timeout-per-shutdown-phase:65s).

Why 65 seconds?
- AWS spot instance termination sends SIGTERM 2 minutes before SIGKILL
-65 seconds per Spring component provides sufficient time for:
  - In-flight requests to complete
  - Connection pools to drain  
  - Background tasks to finish
- Still leaves buffer time within the 2-minute termination window

3. Frontend Pod Shutdown Configuration (Play Framework with Akka)
Configured shutdown phase timeouts for the frontend pod using Akka coordinated shutdown:
- before-service-unbind: 10s (default) - Time to signal clients before connections close
- service-requests-done: 65s (default) - Time for in-flight HTTP requests to drain (equivalent to Spring's timeout-per-shutdown-phase)
- service-stop: 15s (default) - Time for cleanup tasks (WebSocket clients, caches, etc.)
- Total shutdown time: 70 seconds (10s +65 + 15s)

4. Feature Flags for Shutdown Behavior
Added feature flags to control shutdown hook behavior per component:
- set SERVER_SHUTDOWN_MODE to graceful (defaults to immediate)
- FRONTEND_GRACEFUL_SHUTDOWN_ENABLED (defaults to false)

This allows quick rollback if issues arise in production.

### DIFF
--- a/datahub-frontend/app/config/GracefulShutdownModule.java
+++ b/datahub-frontend/app/config/GracefulShutdownModule.java
@@ -14,9 +14,42 @@ import org.apache.http.impl.client.CloseableHttpClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * Guice module managing graceful shutdown for the DataHub Frontend pod.
+ *
+ * <p>When enabled via FRONTEND_GRACEFUL_SHUTDOWN_ENABLED, this module coordinates with Akka's
+ * CoordinatedShutdown to gracefully handle Kubernetes pod termination:
+ *
+ * <ol>
+ *   <li>AWS Spot/K8s sends SIGTERM to the pod
+ *   <li>Akka CoordinatedShutdown triggers, calling registered phase tasks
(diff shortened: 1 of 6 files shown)


kewats:
GMS API calls have a timeout of 60 sec, should we tweak the config based on that, instead of using 45?

Dutt23:
> GMS API calls have a timeout of 60 sec, should we tweak the config based on that, instead of using 45?

@kewats Fair point bumped it up to 65s , 5 second buffer just in case.

PR #16842, opened 2026-03-30
feat(ingest/athena): add upstream lineage for Iceberg/Glue

alokr-dhub:
## Summary

Improves upstream lineage for the Athena connector by routing lineage to the most semantically accurate upstream entity based on catalog type, instead of always pointing to the raw S3 location.

**Lineage routing logic:**
- **Glue catalog** (`AwsDataCatalog` or any catalog confirmed as type `GLUE` via `ListDataCatalogs`): emits upstream lineage to the corresponding Glue dataset entity (`glue://<schema>.<table>`)
- **Non-Glue catalog, Iceberg table**: emits upstream lineage to the corresponding Iceberg dataset entity (`iceberg://<schema>.<table>`)
- **Non-Glue catalog, non-Iceberg table**: falls back to the S3 location as before

**Catalog type detection:**
- Queries the Athena `ListDataCatalogs` API (paginated) to determine the catalog type (`GLUE`, `LAMBDA`, `HIVE`, `FEDERATED`)
- Falls back gracefully: `AwsDataCatalog` is always assumed Glue (it is always backed by Glue); any other catalog name that isn't found or fails the API call is assumed non-Glue
- Result is cached via a `_is_glue_catalog: Optional[bool]` field (initialized to `None`, populated on first access via a `@property`) to avoid redundant API calls

**Other changes:**
- Updated `LINEAGE_COARSE` / `LINEAGE_FINE` capability descriptions to reflect new routing logic
- Updated autogenerated connector registry description
- Removed `@pytest.mark.integration` from `test_athena_get_table_properties`, making it a proper unit test

## Test plan

- [x] `test_get_catalog_type_*` — covers successful lookup, case-insensitive match, catalog-not-found, and API exception paths
- [x] `test_is_glue_catalog_*` — covers GLUE/non-GLUE catalog types, `AwsDataCatalog` fallback (true), non-default catalog fallback (false), API exception fallback for both default and non-default catalogs, and caching behavior
- [x] `test_athena_get_table_properties_glue_iceberg_returns_glue_urn` — Glue catalog always routes to Glue URN even for Iceberg tables
- [x] `test_athena_get_table_properties_iceberg_location` — non-Glue Iceberg table routes to Iceberg URN
- [x] `test_athena_get_table_properties_non_glue_non_iceberg_location` — non-Glue non-Iceberg table falls back to S3 URN


metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
```diff
@@ -440,17 +440,17 @@ def get_table_properties(
...
-        location: Optional[str] = custom_properties.get("location")
-        if location is not None:
-            if location.startswith("s3://"):
-                location = make_s3_urn(location, self.config.env)
-            else:
-                logging.debug(
-                    f"Only s3 url supported for location. Skipping {location}"
-                )
-                location = None
+        # All Athena tables are backed by the Glue Data Catalog.
```
ligfx: > All Athena tables are backed by the Glue Data Catalog.

This is only true when `catalog.lower() == 'awsdatacatalog'`

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
```diff
@@ -440,17 +440,17 @@ def get_table_properties(
...
+        # Iceberg tables → upstream lineage to the Iceberg entity
+        # Non-Iceberg tables → upstream lineage to the Glue entity
+        if metadata.parameters.get("table_type") == "ICEBERG":
```
ligfx: Why Iceberg and not just Glue? They are Glue tables, the Glue tables themselves will have the lineage back to Iceberg.

alokr-dhub: For Iceberg tables, I'm pointing directly to the Iceberg entity rather than Glue because:
1. The Iceberg entity already carries the granular metadata (full schema with evolution, partition specs,
   snapshots). Glue is a thin catalog pointer on top of it.
2. An Iceberg table on AWS implicitly means a Glue catalog entry exists, so Athena→Iceberg already       
  implies the Glue relationship. 
3. Going Athena→Glue→Iceberg adds an extra hop through a less informative entity without adding value.
4. Glue doesn't automatically track storage level lineage by default. So Glue tables will have lineage to S3 as was the case for Athena initialy
Also, this was suggested in the ticket.
 Let me know your thoughts. I will change it to Glue only

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
alokr-dhub: ACK. Filtering on catalog name isn't ideal since awsdatacatalog is just the default name and could vary. 
  Updated to query the GetDataCatalog API and filter on the catalog Type field instead (GLUE vs 
  LAMBDA/HIVE/FEDERATED)

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
```diff
@@ -401,6 +402,29 @@ def get_inspectors(self) -> Iterable[Inspector]:
             inspector = inspect(conn)
             yield inspector
 
+    def _get_catalog_type(self) -> Optional[str]:
+        """Query the Athena API to get the catalog type (GLUE, LAMBDA, HIVE, FEDERATED)."""
+        assert self.cursor
+        try:
+            client = self.cursor._connection.client  # type: ignore[union-attr]
+            response = client.get_data_catalog(Name=self.config.catalog_name)  # type: ignore[attr-defined]
```
ligfx: Note that this always returns `InvalidRequestException: DataCatalog AwsDataCatalog was not found` when used for `AwsDataCatalog`. You can also get the catalog type from the ListDataCatalogs API call.

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
```diff
@@ -440,15 +464,11 @@ def get_table_properties(
...
-            else:
-                logging.debug(
-                    f"Only s3 url supported for location. Skipping {location}"
-                )
-                location = None
+        # Emit upstream lineage to Glue when the catalog is Glue-backed.
+        # Iceberg lineage is handled by the Glue connector.
+        location: Optional[str] = None
+        if self.is_glue_catalog:
+            location = make_dataset_urn("glue", f"{schema}.{table}", self.config.env)
```
ligfx: If it's not a Glue catalog, let's keep the prior behavior?

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
ligfx: You can also use the fact that `AwsDataCatalog` is _always_ a Glue catalog.

alokr-dhub: ACK

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
alokr-dhub: ACK.

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
```diff
@@ -440,15 +474,20 @@ def get_table_properties(
...
+        else:
+            # Fall back to S3 lineage for non-Glue catalogs.
+            s3_location = custom_properties.get("location")
+            if s3_location is not None:
+                if s3_location.startswith("s3://"):
+                    location = make_s3_urn(s3_location, self.config.env)
+                else:
+                    logging.debug(
+                        f"Only s3 url supported for location. Skipping {s3_location}"
+                    )
```
ligfx: In the case where this is an Iceberg table, it would be more correct to generate lineage to the Iceberg entity instead of S3.

alokr-dhub: fixed this.

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
```diff
@@ -401,16 +404,77 @@ def get_inspectors(self) -> Iterable[Inspector]:
...
+
+    @property
+    def is_glue_catalog(self) -> bool:
+        if self._is_glue_catalog is None:
+            catalog_type = self._get_catalog_type()
+            # Default to True (Glue) when catalog type cannot be determined,
+            # since Glue is the most common Athena catalog type.
+            self._is_glue_catalog = (
+                catalog_type == "GLUE" if catalog_type is not None else True
+            )
```
ligfx, line 443: This isn't correct. Glue is the most common catalog type because all Athena configs have `AwsDataCatalog` which is _always_ Glue. For an Athena config with more than one catalog, the higher likelihood is that all non-`AwsDataCatalog` catalogs are _not_ Glue.

alokr-dhub: fixed it. 
The fallback now keys off catalog_name:                                                                                                                                                               
  - AwsDataCatalog → assume Glue (it's always backed by Glue)                                                                                                  
  - any other catalog name → assume non-Glue, since an explicitly-configured catalog is much more likely Lambda / Hive / Federated
 
This way a failed ListDataCatalogs call (or a catalog missing from the response) no longer silently mis-labels non-Glue catalogs as Glue.                    
Also cleaned up the implementation while I was in there , `is_glue_catalog` is now a `functools.cached_property` and `_get_catalog_type` has a single return None fallthrough. Tests updated accordingly.

puneetagarwal-datahub (dismissed):
.

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
```diff
@@ -401,16 +404,75 @@ def get_inspectors(self) -> Iterable[Inspector]:
...
+        except Exception as e:
+            self.report.warning(
+                message="Failed to determine catalog type",
+                context=f"catalog={catalog_name}, error={e}",
+            )
+        return None
+
+    @functools.cached_property
+    def is_glue_catalog(self) -> bool:
+        catalog_type = self._get_catalog_type()
```
treff7es: When _get_catalog_type() fails (transient throttle, missing athena:ListDataCatalogs IAM permission, network blip, PyAthena private-attrdrift) it returns None and the property silently falls back to:
```
return self.config.catalog_name.lower() == "awsdatacatalog"
```

Then the result is memoized for the entire ingestion. One transient ThrottlingException on table #1 produces wrong Glue/non-Glue classification for every subsequent table, with only one warning ("Failed to determine catalog type") whose user-facing consequence (heuristic applied → maybe-wrong URNs everywhere) is never reported.

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
```diff
@@ -401,16 +404,75 @@ def get_inspectors(self) -> Iterable[Inspector]:
...
+
+        Returns None if the catalog isn't found or the API call fails.
+        """
+        if self.cursor is None:
+            raise RuntimeError(
+                "Cursor not initialized; call get_table_properties first"
+            )
+        catalog_name = self.config.catalog_name
+        try:
+            client = self.cursor._connection.client  # type: ignore[union-attr]  # PyAthena internal: accessing boto3 Athena client
```
treff7es: Pyathena's BaseCursor exposes a public connection property (verified at pyathena/common.py:128-130). The reason string is factually wrong, and the code is more brittle than it needs to be (a future PyAthena release can rename _connection and quietly break this at runtime while mypy stays happy)

Use self.cursor.connection.client and update the reason. 

Even cleaner: build a dedicated boto3 Athena client from AthenaConfig credentials, matching the pattern used elsewhere in metadata-ingestion/src/datahub/ingestion/source/aws/.

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
```diff
@@ -401,16 +404,75 @@ def get_inspectors(self) -> Iterable[Inspector]:
...
+            return catalog_type == "GLUE"
+        # AwsDataCatalog is always backed by Glue. Any other catalog name was
+        # explicitly configured and is more likely non-Glue (Lambda/Hive/Federated).
+        return self.config.catalog_name.lower() == "awsdatacatalog"
+
     def get_db_schema(self, dataset_identifier: str) -> Tuple[Optional[str], str]:
         schema, _view = dataset_identifier.split(".", 1)
         return None, schema
 
+    def _get_upstream_location(
```
treff7es, line 475: Isn't this a breaking change we should mention in `docs/how/updating-datahub.md` ? 

This change silently rewrites lineage URNs for every existing Athena user — a Glue-catalog user who previously saw
  `urn:li:dataset:(urn:li:dataPlatform:s3,…)` upstream will now see `urn:li:dataset:(urn:li:dataPlatform:glue,…)`. Downstream dashboards, lineage queries, and impact-analysis workflows that referenced the S3 URN will break.

treff7es:
@alokr-dhub I left a few comment and please, can you update the pr description to be align with the implementation?

treff7es (dismissed):
oops, this comment got here accidentally

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
alokr-dhub: Switched to the public self.cursor.connection.client property chain.                                                           Considered building a dedicated boto3 client from AthenaConfig but decided against it. AthenaConfig doesn't extend AwsConnectionConfig, so we'd have to manually replicate PyAthena's credential logic from a partial set of fields. That risks the two clients diverging when users rely on ambient credentials. Reusing PyAthena's existing client is simpler and guaranteed to match.

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
alokr-dhub: removed caching and added a report warning in case we fallback to catalog name to resolve `is_glue_catalog` property.

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
alokr-dhub: updated the docs

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
```diff
@@ -401,16 +404,84 @@ def get_inspectors(self) -> Iterable[Inspector]:
...
+        self,
+        schema: str,
+        table: str,
+        metadata: AthenaTableMetadata,
+        custom_properties: Dict[str, str],
+    ) -> Optional[str]:
+        """Determine the upstream lineage URN based on catalog type and table properties."""
+        if self.is_glue_catalog:
+            return make_dataset_urn("glue", f"{schema}.{table}", self.config.env)
```
treff7es: I don't know if we need that, but it's worth noting that this way we won't be able to specify platform_instance/env for the glue/iceberg platform.
You assume it from the Athena config, which might not be correct.

treff7es: We can workaround it using some explicit config like:
```
glue_platform_instance: Optional[str] = pydantic.Field(
          default=None,
          description="Platform instance for the upstream Glue datasets that Athena "
                      "lineage points to. Set this if your Glue connector is configured "
                      "with a non-default platform_instance. Leave unset if Glue uses the "
                      "default (no platform_instance).",
      )
      iceberg_platform_instance: Optional[str] = pydantic.Field(
          default=None,
          description="Platform instance for the upstream Iceberg datasets. Set this if "
                      "your Iceberg connector is configured with a non-default platform_instance.",
      )
```

or a mapping

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
```diff
@@ -401,16 +404,84 @@ def get_inspectors(self) -> Iterable[Inspector]:
             inspector = inspect(conn)
             yield inspector
 
+    def _get_catalog_type(self) -> Optional[str]:
```
treff7es:
```suggestion
    def _get_catalog_type(self) -> Optional[Literal["GLUE", "LAMBDA", "HIVE", "FEDERATED"]]:
```

This forces catalog_type == "GLUE" to be type-checked.

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
```diff
@@ -401,16 +404,84 @@ def get_inspectors(self) -> Iterable[Inspector]:
...
+
+        Returns None if the catalog isn't found or the API call fails.
+        """
+        if self.cursor is None:
+            raise RuntimeError(
+                "Cursor not initialized; call get_table_properties first"
+            )
+        catalog_name = self.config.catalog_name
+        try:
+            client = self.cursor.connection.client  # type: ignore[union-attr]
```
treff7es: Can you add some doc to justify the mypy silencing?

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
```diff
@@ -401,16 +404,84 @@ def get_inspectors(self) -> Iterable[Inspector]:
...
+            )
+        except Exception as e:
+            self.report.warning(
+                message="Failed to determine catalog type",
+                context=f"catalog={catalog_name}, error={e}",
+            )
+        return None
+
+    @property
+    def is_glue_catalog(self) -> bool:
```
treff7es: Currently bool collapses "confirmed-Glue / confirmed-non-Glue / unknown" → 2 states. Returning Optional[bool] would let _get_upstream_location skip emitting lineage on unknown rather than guessing. Trade-off vs. complexity.

treff7es (commented):
Accidentakly

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
treff7es: When _get_catalog_type() returns None (API failure or catalog not found), _is_glue_catalog stays None, so:
  1. Every per-table call to _get_upstream_location re-invokes _get_catalog_type → N× ListDataCatalogs API calls (one per table)
  2. Every call emits report.warning again — and _get_catalog_type itself emits a warning too → 2N+ duplicate warnings
  3. The warning message says "may be incorrect for all tables in this run" — implying it's expected to fire once

  For a 10k-table run with a transient API blip, this means 10k × 2 = 20k warnings and 10k API calls. Test gap: test_is_glue_catalog_caches_result (line 403) only covers the success
  path.

  Fix: Always assign self._is_glue_catalog before returning (cache the fallback too).

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
```diff
@@ -401,16 +404,84 @@ def get_inspectors(self) -> Iterable[Inspector]:
...
+        self,
+        schema: str,
+        table: str,
+        metadata: AthenaTableMetadata,
+        custom_properties: Dict[str, str],
+    ) -> Optional[str]:
+        """Determine the upstream lineage URN based on catalog type and table properties."""
+        if self.is_glue_catalog:
+            return make_dataset_urn("glue", f"{schema}.{table}", self.config.env)
+        if metadata.parameters.get("table_type", "").upper() == "ICEBERG":
```
treff7es: PyAthena's AthenaTableMetadata.parameters can be empty dict — but if Glue ever returns no Parameters map, this is None.get(...) → AttributeError that propagates out of
  get_table_properties. Pre-existing code at line 501 has the same pattern but the PR adds another usage. Defensive fix:

```
  params = metadata.parameters or {}
  if params.get("table_type", "").upper() == "ICEBERG":
```

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
alokr-dhub: I had removed caching due to the concern that in case of API failure it would set the cache incorrectly, leading to incorrect flag for the remaining tables. Let me split the fallback and cache the API response separately.

alokr-dhub: ACK.

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
alokr-dhub: ACK

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
alokr-dhub: Thanks for pointing out. missed this in my implementation.

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
alokr-dhub: ACK.

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
alokr-dhub: fixed it. it's a stale type ignore due to previously accessed _connection attribute and no longer required.

alokr-dhub:
@treff7es addressed all comments and fixed the issues. Please review.

treff7es (approved):
lgtm

PR #16846, opened 2026-03-30
perf(metadata-io): prevent MySQL hard parsing via IN-statement padding

raudzis:
This pull request introduces a targeted performance optimization for batch fetching logic within `EbeanAspectDao` and refactors the `EbeanAspectV2` entity to improve data integrity for composite keys.

#### 1. Conditional IN-Clause Padding (MySQL Specific)
The `padToNextBatchSize` logic is now restricted to **MySQL** environments. This addresses a specific requirement of the MySQL query optimizer to maximize performance.

* **The "Why": Hard vs. Soft Parsing**
    * **Hard Parsing (Suboptimal):** Without consistent query strings, MySQL treats `IN (?, ?, ?)` and `IN (?, ?, ?, ?)` as different statements, forcing a CPU-intensive generation of a new execution plan for every variation in list size.
    * **Soft Parsing (Optimal):** By padding the number of parameters to the next **power of two** ($2, 4, 8, 16, \dots$), the query string remains stable. This increases the cache hit rate in the MySQL statement cache, allowing for immediate plan reuse (Soft Parsing).
* **The Optimization:** Using Ebean’s `platform().base()`, the implementation now detects the platform. If the database is not MySQL (e.g., PostgreSQL, H2), padding is bypassed to avoid unnecessary memory allocation and CPU cycles on the application layer.

#### 2. EbeanAspectV2 Entity Refactoring
~~The `EbeanAspectV2` class has been updated to handle its `EmbeddedId` more efficiently, specifically concerning the `urn`, `aspect`, and `version` fields.~~

~~* **Redundancy Removal:** Direct fields for `urn`, `aspect`, and `version` were removed to prevent data duplication, as these values are already contained within the composite key object.~~
~~* **Proxy Accessors:** Getter methods for `urn`, `aspect`, and `version` now act as **proxies**, retrieving the values directly from the `EmbeddedId`. This ensures a single source of truth and prevents state synchronization~~ ~~issues within the entity.~~

### 📈 Impact Analysis

| Feature | Change | Technical Benefit |
| :--- | :--- | :--- |
| **Power-of-Two Padding** | Stabilized `IN` clauses | Reduced MySQL CPU load via improved plan caching. |
| **Platform Check** | Conditional execution | Zero overhead for non-MySQL deployments (H2/Postgres). |
| ~~**Proxy Getters**~~ | ~~Refactored `EbeanAspectV2`~~ | ~~Guaranteed consistency for composite key components.~~ |

---

### 🧪 Validation

* **Logic Isolation:** The visibility of `padToNextBatchSize` was adjusted to `package-private` to facilitate direct unit testing without reflection.
* **Comprehensive Testing:** `EbeanAspectDaoTest` was expanded to verify:
    * No padding occurs on the default H2 test platform.
    * Correct padding behavior is triggered when a MySQL platform is detected via a mocked database instance.
~~* **Entity Integrity:** Verified that proxy getters correctly surface values from the underlying `EmbeddedId` without side effects during Ebean bean enhancement.~~

---

### 🔗 Implementation Details & 💬 Review & Feedback

Please review the changes and provide feedback accordingly. If any further information or clarification is required, please do not hesitate to reach out—I am happy to provide additional details.

* **Platform Detection:** Performed via `server.platform().base() == Platform.MYSQL`, ensuring high-performance, constant-time checks.
* **Branch:** `perf/io/ebean/ebean-in-tuples-padding`


metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectV2.java
```diff
@@ -119,17 +127,6 @@ public int compareTo(EbeanAspectV2.PrimaryKey other) {
 
   @Nonnull @EmbeddedId @Index protected PrimaryKey key;
 
-  @Nonnull
-  @Column(name = URN_COLUMN, length = 500, nullable = false)
```
Dutt23, line 123: Most of the sqls we run rely on direct access to these fields urn, aspect, version because of the java orm. 
eg : [deleteAspects](https://github.com/datahub-project/datahub/blob/master/metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectDao.java#L262). This could lead to failures. We probably need to change everything to use rawSql and column mappings like this [batchGetIN](https://github.com/datahub-project/datahub/blob/master/metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectDao.java#L502C1-L507C23) incase we remove the fields. 

However we could argue that this could be separate PR , given that this PR's main focus is optimising mysql in queries ?

metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectDao.java
```diff
@@ -905,7 +883,8 @@ public Map<String, Map<String, Long>> getNextVersions(
     if (canWrite) {
       // Sorting is required to ensure consistent lock ordering and avoid deadlocks
       Collections.sort(forUpdateKeys);
-      server.find(EbeanAspectV2.class).where().idIn(forUpdateKeys).forUpdate().findList();
+      final List<EbeanAspectV2.PrimaryKey> paddedBatch = padToNextBatchSize(forUpdateKeys);
```
Dutt23: Instead of doing `padToNextBatchSize(forUpdateKeys)` everywhere should we have a common method called `findByPK` where we internally do the padding and then fire the query ? 
Otherwise everywhere we would always have to call padding first and then fire the query , would be easy to forget at times noh ?

Dutt23:
From what I understand the progress for the IN query would go something like this. 
    1 → 1
    2 → 2
    4 → 4
    8 → 8
    16 → 16
    32 → 32
    64 → 64
    128 → 128
    256 → 256
    512 → 512
    
Would it be a good idea to change `queryKeysCount` variable to 256 ? The current value is 375. Looking at the progression it would create a list of  512 , which is a waste. Resulting in a larger request size and buffer.

metadata-io/src/test/java/com/linkedin/metadata/entity/ebean/EbeanAspectDaoTest.java
```diff
@@ -532,4 +538,109 @@ public void testWritabilityDuringMigration() {
         testDao.getAspect("urn:li:corpuser:postMigration", "status", ASPECT_LATEST_VERSION);
     assertTrue(aspect != null, "Writes work after migration");
   }
+
+  // region padToNextBatchSize tests
+  @Test
+  public void testPadToNextBatchSizeForH2() {
```
Dutt23: Would it be a good idea to add parameterised tests here ? so we test easily for edge case sizes eg : 1, 2, 9, 15, 16, 255, 256

metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectV2.java
raudzis: I noticed that ```the EbeanAspectV2``` entity is already using a workaround with a composite PK, which is why ```urn```, ```aspect```, and ```version``` are being double-generated in the SQL. I decided to roll back my changes since messing with the base logic would’ve been a massive effort. A cleaner fix would be nice, but yeah, the Ebean implementation just isn't perfect.

metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectDao.java
raudzis: Good catch, fixed!

metadata-io/src/test/java/com/linkedin/metadata/entity/ebean/EbeanAspectDaoTest.java
raudzis: Good catch, fixed!

raudzis:
**regrading**: _Would it be a good idea to change queryKeysCount variable to 256 ? The current value is 375. Looking at the progression it would create a list of 512 , which is a waste. Resulting in a larger request size and buffer._

While 512 involves some padding, it is intentional. Splitting ~375 keys into two 256-batches would double our network roundtrips and transaction overhead. The database handles the 512-element 'waste' extremely efficiently via plan reuse and its internal range optimizer, making it significantly faster than executing multiple smaller queries. 

I would leave the value at 375 for now to maintain current behavior and ensure we don't regress into multiple roundtrips for medium-sized batches. In a follow-up step, we could consider either reducing the value or explicitly increasing the batch size to 512

metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectDao.java
```diff
@@ -1068,4 +1022,108 @@ private Map<String, Map<String, SystemAspect>> toUrnAspectMap(
...
+    // SQL (one for the PK object and one for the property itself).
+
+    final Query<EbeanAspectV2> query =
+        server
+            .find(EbeanAspectV2.class)
+            .select(
+                String.join(
+                    ", ",
+                    EbeanAspectV2.METADATA_COLUMN,
+                    "systemMetadata",
```
Dutt23: Shouldn't this be 
```
.select(
      String.join(
          ", ",
          EbeanAspectV2.URN_COLUMN,
          EbeanAspectV2.ASPECT_COLUMN,
          EbeanAspectV2.VERSION_COLUMN,
          EbeanAspectV2.METADATA_COLUMN,
          "systemMetadata",
          EbeanAspectV2.CREATED_ON_COLUMN,
          EbeanAspectV2.CREATED_BY_COLUMN,
          EbeanAspectV2.CREATED_FOR_COLUMN))
```. or remove this select and let Ebean use default projections ?

Comment (raudzis) [reply metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectDao.java]: If I remove the SELECT for urn, aspect, and version, the entries are listed twice due to the composite key. I tried to prevent this with the SELECT statement, but I can remove it if needed.

```
SELECT
  `t0` . `urn`, 
  `t0` . `aspect`,
  `t0` . `version`,
  `t0` . `urn`,
  `t0` . `aspect`,
  `t0` . `version`,
  `t0` . `createdOn`,
  `t0` . `createdBy`,
  `t0` . `createdFor`,
  `t0` . `systemmetadata`
FROM
  `metadata_aspect_v2` `t0`
WHERE
  ( `t0` . `urn`,
    `t0` . `aspect`,
    `t0` . `version` ) IN ( (...) ) FOR
UPDATE
```

Comment (raudzis) [reply metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectDao.java]: I’ve implemented the explicit select as suggested, but with a specific focus on performance. In Ebean, omitting the .select() clause causes LOB columns to be excluded and fetched later via lazy loading.

Since this query uses a FOR UPDATE lock, triggering lazy loading later is inefficient. It is better to fetch all columns—including LOBs—in the initial statement to avoid the overhead of additional database round-trips while the row is locked.

This change also simplifies our execution plan: tests now show a lower count of generated SQL statements, which will significantly reduce the database load during heavy data ingestion.

**Comment by raudzis** [comment]
PR Summary: Optimization of Ebean Query Projection for FOR UPDATE Locks
🎯 Problem Description

In the current implementation, using Ebean's default finders without an explicit .select() clause leads to Lazy Loading for @Lob columns (such as metadata and systemMetadata). Ebean excludes these columns from the initial SQL statement to keep the payload small.

However, when a record is fetched with a FOR UPDATE lock, the intention is almost always to read or modify the full state of the record. If the code later accesses the metadata field, Ebean triggers a second SQL query (Lazy Loading) to fetch the missing LOB data.

🛠 Proposed Change

I have updated the Ebean query to use .select("*") (or an explicit column list) when performing locks:

Java
server
    .find(EbeanAspectV2.class)
    .select("*") // Ensures LOB columns are included in the initial fetch
    // ...
    .forUpdate();
🚀 Benefits & Advantages

Avoids "N+1" Queries within Locks: By fetching everything in a single trip, we eliminate the secondary "Lazy Loading" SQL statement that occurs when accessing the metadata or systemMetadata fields.

Reduced Database Overhead: Executing one comprehensive SQL statement is significantly more efficient than two separate statements, especially under high load.

Optimized for Data Ingestion: During heavy ingestion (e.g., metadata updates), this change reduces the overall number of generated SQL statements, lowering the pressure on the DB connection pool.

Consistency under Lock: It is safer and more performant to read the entire row state while the FOR UPDATE lock is active, ensuring that all data (including LOBs) corresponds to the locked version.

Improved Test Performance: Since no additional queries are triggered, the execution plan in tests is more predictable and faster.

📊 Technical Context (EbeanAspectV2)

The metadata_aspect_v2 table heavily relies on "Lob" columns:

metadata (String/Lob)

systemMetadata (String/Lob)

By using .select("*"), these are eagerly fetched, preventing the performance bottleneck of delayed LOB loading.

File: metadata-io/src/test/java/com/linkedin/metadata/entity/EbeanEntityServiceOptimizationTest.java [hunk 5]
```diff
@@ -69,6 +69,9 @@ public class EbeanEntityServiceOptimizationTest {
 
   @BeforeMethod
   public void setupTest() {
+    System.setProperty("logging.level.root", "DEBUG");
```
Dutt23: Needs to be removed

metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectDao.java
```diff
@@ -1068,4 +1022,95 @@ private Map<String, Map<String, SystemAspect>> toUrnAspectMap(
...
+    // We manually select all columns here.
+    // Reason: The EbeanAspectV2 entity uses a composite primary key. Without explicit
+    // selection, Ebean's default behavior would generate duplicate columns in the
+    // SQL (one for the PK object and one for the property itself).
+
+    final Query<EbeanAspectV2> query =
+        server.find(EbeanAspectV2.class).setReadOnly(!forUpdate).where().idIn(paddedBatch).query();
+
+    if (forUpdate) {
+      query.select(EbeanAspectV2.ALL_COLUMNS).forUpdate();
```
Dutt23, line 1126: Do we need this only for this branch `forUpdate` ? Should it be applied to all ?

raudzis: I would reserve FOR UPDATE strictly operations, as it requires Metadata and SystemMetadata which are currently defined as lazy-loaded. By omitting FOR UPDATE, we can leverage lazy loading; if only specific attributes like the key are needed, the database avoids the overhead of fetching the full metadata objects.

metadata-io/src/test/java/com/linkedin/metadata/entity/EbeanEntityServiceOptimizationTest.java
raudzis: Applied the suggested changes. Thanks!

metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectDao.java
```diff
@@ -1142,4 +1096,95 @@ private Map<String, Map<String, SystemAspect>> toUrnAspectMap(
...
+   * Pads the input list to the next power of two. This is a workaround for MySQL to ensure a stable
+   * query plan by providing a fixed number of placeholders in the IN clause. This helps MySQL to
+   * reuse the execution plan (Soft Parsing). Other databases like PostgreSQL or H2 do not require
+   * this padding.
+   *
+   * @param list The original sublist of keys.
+   * @return A list padded to the next power of two, or the original list if not using MySQL.
+   */
+  // Package-private for testing purposes
+  List<EbeanAspectV2.PrimaryKey> padToNextBatchSize(List<EbeanAspectV2.PrimaryKey> list) {
```
Dutt23, line 1141: If list is null, this returns null, and the caller at line ~1123 does idIn(paddedBatch) which would NPE. The immediate call site does check keys.isEmpty() first, but the null-returning contract here is still fragile. Consider either adding @Nonnull to the parameter and failing fast, or returning Collections.emptyList() instead of propagating null

PR #16862, opened 2026-03-31
feat(ci): selective integration test routing for metadata-ingestion

treff7es:
## Summary

Instead of running the full integration test suite for every metadata-ingestion PR, analyze changed files and run only the relevant connector tests.

## How it decides what to run

For every changed file in a PR, the script classifies it:

**Skip entirely** (no integration tests needed):
- Files outside `metadata-ingestion/` and `metadata-models/`
- Unit tests, performance tests, scripts, docs, root conftest

**Run the full suite:**
- Any `metadata-ingestion/` file NOT under `source/` or `tests/integration/` (e.g. `setup.py`, `api/`, `emitter/`)
- Any `metadata-models/` file
- Unknown source files not covered by any known connector
- Files directly in `tests/integration/` (shared test infra like `conftest.py`)
- Safety net: changed source dirs that produce zero test paths

**Run specific connector tests** — files under `source/{connector}/` or `tests/integration/{connector}/`, determined by:

### Step 1: Discover connectors
- **Directory connectors**: `source/{name}/` matched to `tests/integration/{name}/` by convention
- **Single-file connectors**: `source/{name}.py` matched to `tests/integration/{name}/`
- **connector.yaml overrides** for exceptions (currently only `powerbi/connector.yaml`)

### Step 2: Build dependency maps
- **Import graph**: parse `.py` files with `ast`, map imports back to known connector source dirs
- **Entry-point map**: parse `setup.py` entry points to map test dirs to source modules (e.g. `clickhouse → datahub.ingestion.source.sql.clickhouse → source/sql/`)

### Step 3: Entry-point narrowing (for shared source dirs like `source/sql/`)
- `sql/clickhouse.py` → matches clickhouse EP → only `tests/integration/clickhouse/`, does NOT trigger other SQL tests
- `sql/sql_common.py` → no EP match (shared utility) → import graph propagates to all SQL dependents (mysql, postgres, clickhouse, etc.)

### Step 4: Propagate via import graph
If a connector's source dir was directly changed, or any of its import dependencies were changed → add its test paths

### Step 5: Collect integration test changes
If a file under `tests/integration/{connector}/` changed (test code or golden file) → add that test dir directly

### Example flows

| Changed file | What runs |
|---|---|
| `source/powerbi/powerbi.py` | Only PowerBI tests |
| `source/sql/clickhouse.py` | Only clickhouse tests (EP narrowing) |
| `source/sql/sql_common.py` | All SQL-family connectors (import graph) |
| `source/powerbi/config.yaml` | Only PowerBI tests |
| `source/powerbi/README.md` | Nothing (`.md` ignored) |
| `tests/integration/kafka/golden.json` | Only kafka tests |
| `tests/integration/conftest.py` | Full suite (shared test infra) |
| `setup.py` | Full suite (core change) |
| `source/brand_new_connector/foo.py` | Full suite (unknown connector → safety net) |

## Files

- `scripts/selective_ci_checks.py` — main detection script (~700 lines)
- `tests/unit/test_selective_ci_checks.py` — 72 unit tests
- `src/datahub/ingestion/source/powerbi/connector.yaml` — 1 exception file (extra_source_paths)
- `.github/workflows/metadata-ingestion.yml` — detect job + gate job + validate step
- `scripts/test_ci_simulation.py` — replay script for testing against git history
- `docs/dev_guides/selective_ci.md` — developer documentation
- Test directory renames to match source dir names (mssql, looker/lookml, fabric/*, kafka_connect/*)

## Test plan

- [x] `python -m pytest tests/unit/test_selective_ci_checks.py` — 72 tests pass
- [x] `python scripts/selective_ci_checks.py --validate` — all mappings valid
- [x] `python scripts/test_ci_simulation.py --count 30` — simulation against real PRs
- [x] ruff lint + format clean
- [x] 3 review cycles with 4 automated agents — all clear
- [ ] Verify `detect` job runs correctly on this PR
- [ ] Verify `integration-tests-gate` passes
- [ ] Test `run-all-tests` label forces full suite

### DIFF
--- a/.github/workflows/metadata-ingestion.yml
+++ b/.github/workflows/metadata-ingestion.yml
@@ -193,6 +193,21 @@ jobs:
           TEST_PATH=$(echo "$MATRIX_JSON" | jq -r '.test_path // empty')
           CONNECTOR=$(echo "$MATRIX_JSON" | jq -r '.connector // empty')
 
+          # Log what this job is running to the step summary
+          {
+            echo "### Test Job"
+            echo "| Key | Value |"
+            echo "|-----|-------|"
+            echo "| Python | \`${{ matrix.python-version }}\` |"
+            if [ -n "$COMMAND" ]; then
+              echo "| Command | \`$COMMAND\` |"
+            fi
+            if [ -n "$CONNECTOR" ]; then
+              echo "| Connector | \`$CONNECTOR\` |"
+              echo "| Test Path | \`$TEST_PATH\` |"
+            fi
+          } >> "$GITHUB_STEP_SUMMARY"
+
           if [ -n "$COMMAND" ]; then
             ./gradlew :metadata-ingestion:"$COMMAND"
           elif [ -n "$TEST_PATH" ]; then


askumar27 (approved):
This is excellent work! the safety nets are neat.

metadata-ingestion/scripts/selective_ci_checks.py
```diff
@@ -0,0 +1,699 @@
...
+            print("All connector mappings valid.")
+            sys.exit(0)
+
+    if args.force_all:
+        decisions = CIDecisions.run_all()
+    else:
+        changed = get_changed_files(args.base_ref)
+        decisions = classify(changed, repo_root)
+
+    if args.dry_run or args.output == "json":
```
devashish2203, line 682: We should always write the json file from this script and then use that to
1. Write to GITHUB_OUTPUT
2. Save a artifact for debugging
3. Write to GITHUB_SUMMARY to easily see the output of this step.

devashish2203:
> I'm concerned with all the complexity introduced here Have you checked if there are existing solutions out there for this use case?

While the solution here is impressive, I agree with Sergio that the maintenance overhead here incase something changes might be too much. I'm okay with merging this if Ingestion team is okay to maintain changes to these custom scripts going forward.


Main request before merging is to ensure we have GITHUB_SUMMARY and output of the json matrix generated available to easily see which jobs have ran.

treff7es:
> I'm concerned with all the complexity introduced here Have you checked if there are existing solutions out there for this use case?

The tools you suggested are great tools for unit tests but for our integration tests won't work and here is what this pr does what those tools can't:
1. Connector discovery — Convention-based (source/{name}/ + tests/integration/{name}/), entry-point-based (parsing setup.py), and connector.yaml overrides. Snob just sees Python files; it
  doesn’t know about DataHub’s connector registry.
  2. Tiered classification — Your three-tier system (non-connector = full suite, shared-base = graph analysis, direct connector = selective) is domain-specific CI policy. Snob would tell you
  “these tests are affected” but wouldn’t know that changes to emitter/ or setup.py should force the full suite.
  3. GitHub Actions matrix generation — The JSON matrix output (test_matrix) that drives the CI workflow’s strategy.matrix. Snob outputs a list of test files, not a CI matrix.
  4. Entry-point narrowing — Your _narrow_ep_tests() logic that maps thin-wrapper connectors (e.g., sql/clickhouse.py) to their specific test directories via setup.py entry points. This is a
  DataHub-specific mapping layer.
  5. Integration test batching — The run_all_integration flag and the 7-batch static split for full suite runs.

.github/workflows/metadata-ingestion.yml
```diff
@@ -27,32 +27,104 @@ env:
...
+  # ── Detect which tests need to run and build the full CI matrix ────────────
+  detect:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.build-matrix.outputs.matrix }}
+    steps:
+      - uses: acryldata/sane-checkout-action@186e92cc5948a9c3e1cc7a96eaff9f776f3fc8e3 # v7
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6
```
sgomezvillamor, line 39: why not using v6/v7 respectively?

.github/workflows/metadata-ingestion.yml
```diff
@@ -27,32 +27,104 @@ env:
...
+        with:
+          python-version: "3.11"
+      - name: Install PyYAML
+        run: pip install pyyaml
+      - name: Check for force-all label
+        id: label-check
+        env:
+          LABELS: ${{ toJson(github.event.pull_request.labels.*.name) }}
+        run: |
+          if echo "$LABELS" | jq -e 'any(. == "run-all-tests")' > /dev/null 2>&1; then
```
sgomezvillamor: should we narrow down the label? `run-all-ingestion-tests`

.github/workflows/metadata-ingestion.yml
```diff
@@ -27,32 +27,104 @@ env:
...
+          EVENT_NAME: ${{ github.event_name }}
+          FORCE_ALL: ${{ steps.label-check.outputs.force_all }}
+        run: |
+          # Force full suite for: push to master, releases, manual dispatch, or label override
+          if [ "$EVENT_NAME" != "pull_request" ] || [ "$FORCE_ALL" = "true" ]; then
+            python metadata-ingestion/scripts/selective_ci_checks.py --force-all
+          else
+            python metadata-ingestion/scripts/selective_ci_checks.py \
+              --base-ref "$BASE_REF"
+          fi
```
sgomezvillamor, line 67: is this step missing $GITHUB_OUTPUT for `steps.ci.outputs.run_all_integration` and `steps.ci.outputs.test_matrix`?

.github/workflows/metadata-ingestion.yml
```diff
@@ -100,9 +175,32 @@ jobs:
               git diff "$json_file"
               exit 1
             fi
-          done
+          done < <(find metadata-ingestion/src/datahub/ingestion/autogenerated -name "*.json")
+      - name: Validate selective CI connector mappings
```
sgomezvillamor, line 179: what are we doing here in this validations?

.github/workflows/metadata-ingestion.yml
```diff
@@ -100,9 +175,32 @@ jobs:
...
+          python metadata-ingestion/scripts/selective_ci_checks.py --validate
       - name: Run metadata-ingestion tests
-        run: ./gradlew :metadata-ingestion:${{ matrix.command }}
+        run: |
+          # matrix.command is set for Gradle batch jobs (testQuick, testIntegrationBatch*)
+          # For selective connector tests, test_path and connector come from the dynamic matrix
+          MATRIX_JSON='${{ toJSON(matrix) }}'
+          COMMAND=$(echo "$MATRIX_JSON" | jq -r '.command // empty')
+          TEST_PATH=$(echo "$MATRIX_JSON" | jq -r '.test_path // empty')
+          CONNECTOR=$(echo "$MATRIX_JSON" | jq -r '.connector // empty')
```
sgomezvillamor, line 194: you may dump these in the step summary to understand/troubleshoot what's going on

.github/workflows/metadata-ingestion.yml
treff7es: This is the established pattern across the entire repo — every single action in metadata-ingestion.yml and build-and-test.yml is SHA-pinned with a # vN comment. SHA pinning is a GitHub-recommended security best practice to prevent supply chain attacks.

.github/workflows/metadata-ingestion.yml
treff7es: The Python script itself writes to GITHUB_OUTPUT — it reads os.environ.get("GITHUB_OUTPUT") at selective_ci_checks.py:686 and writes key=value pairs to it (lines 700-704). The $GITHUB_OUTPUT env var is automatically set by GitHub Actions for every step. So the step doesn't need to explicitly reference it in the workflow YAML; the script handles it internally. The step has id: ci, so steps.ci.outputs.* will pick up whatever the script writes.

metadata-ingestion/docs/dev_guides/selective_ci.md
sgomezvillamor, line 1: `# Selective CI for Integration Tests`, that's too generic while it only applies to `metadata-ingestion`.

Also, I understand these dev_guides are user-facing docs. Docs there should be to explain contributors how to develop on datahub. But there is no action to be done by them here.

Instead this explains how the selective ci works. I would place this docs closer to the `metadata-ingestion.yml` github workflow, maybe even within itself. No need to be user-facing doc in the portal.

IMO

metadata-ingestion/tests/integration/clickhouse/test_clickhouse.py
```diff
@@ -1,19 +1,21 @@
+from datetime import datetime, timezone
 from typing import List
 
 import pytest
-from freezegun import freeze_time
```
sgomezvillamor, line 4: any reason why mixing these refactor updates here?

treff7es: yes, without it the selective integration tests would fail because of the pydantic 2.x upgrade in the past. 
This would be an issue locally as well if you try to run those integration tests selectively.

metadata-ingestion/src/datahub/ingestion/source/powerbi/connector.yaml
```diff
@@ -0,0 +1,3 @@
+# Exception: powerbi_report_server source is also covered by powerbi tests.
```
sgomezvillamor, line 1: to understand how critical is this, what'd be the impact if we miss this?

if critical, there should be a mechanism to check that these files are not missed when necessary?

sgomezvillamor:
I understand file movements are not cosmetic but functional for the selective ci to work.
If we have some test directory structure requirement, should we have a CI validation for both:
* making explicit what the structure requirement is
* and ensuring the structure requirement is preserved

PR #16863, opened 2026-03-31
feat(agent-context-kit): Add Cloud tools for interacting with Ask DataHub

jjoyce0510:
# Summary

In this PR, we are adding cloud-only tools for using Ask DataHub (starting a chat, checking a chat). 

# Status

Ready for review.

### DIFF
--- a/datahub-agent-context/examples/google_adk/basic_agent.py
+++ b/datahub-agent-context/examples/google_adk/basic_agent.py
@@ -7,23 +7,32 @@ This example demonstrates how to create a simple Google ADK agent that can:
 - Get detailed entity information
 - Query lineage relationships
 - Update metadata like tags and descriptions
+- Query BigQuery directly (list datasets/tables, run SQL, get insights)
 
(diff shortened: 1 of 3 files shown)


datahub-agent-context/src/datahub_agent_context/mcp_tools/ask_datahub.py
```diff
@@ -0,0 +1,331 @@
...
+        result = ask_datahub_chat(message="What are the most queried datasets?")
+        print(result["response"])
+
+        # Follow-up in the same conversation
+        result = ask_datahub_chat(
+            message="Show me the schema of the top one",
+            conversation_urn=result["conversation_urn"],
+        )
+    """
+    _require_cloud()
```
nwadams, line 173: should we be more explicit about the error message here? Returning an error message instead of raising allows the agent to more gracefully recover.

jjoyce0510:
I will address the concerns here

PR #16871, opened 2026-04-01
fix(ingest/postgres): enable stored procedure lineage extraction

### LINKED ISSUE #13541: feat(ingest/postgres): support stored procedures for Postgres
Greetings!

It is very common thing in the world of Data Warehouses, built on a top of relational databases (Oracle / Postgres / GreenPlum / MSSQL / MySQL / you name it), that the data processing is happening inside of a database by stored procedures means. There you have tables to store original data from various source systems and a number of stored procedures that process that data to transform and load it into other tables. So you're moving data across dwh layers (like raw -> ods -> dds -> dm) by calling procedures or running some DML with the assistance of tools like DBT. 

I have checked the [Postgres connector](https://github.com/datahub-project/datahub/blob/master/metadata-ingestion/src/datahub/ingestion/source/sql/postgres.py) (the one I'm currently most interested in) and can see it  only extracts tables and views. However in the DataHub documentation there're entity types like [DataProcess](https://docs.datahub.com/docs/generated/metamodel/entities/dataprocess), [DataJob](https://docs.datahub.com/docs/generated/metamodel/entities/datajob), [DataFlow](https://docs.datahub.com/docs/generated/metamodel/entities/dataflow) that can be used as a container to extract stored procedures and functions, and add proper lineage among stored code and tables. Some work in that direction was already done in #12929  for Snowflake, in #12356 for MySQL and in #11287/#12483 for MSSQL

Are there any plans to add stored procedures ingestion support for Postgres?

jinhyukchang:
## Problem

Users running Postgres ingestion with stored procedures saw no lineage edges emitted — procedures appeared as data jobs with empty inputs/outputs. Two bugs caused this:

1. **Wrong function for procedure body** — `pg_get_functiondef()` returns the full `CREATE OR REPLACE PROCEDURE ... AS $$ ... $$` wrapper. The lineage parser skips `CREATE` statements, so the entire procedure body was silently discarded before parsing.
2. **Case mismatch on language** — Postgres stores language names in lowercase (`"sql"`) in `pg_proc.lanname`, but the lineage check compared against `"SQL"`. This meant even after fix #1, the lineage extraction branch was never entered.

Fixes #13541

## Changes

**Bug fixes:**
- `source.py`: Replace `pg_get_functiondef()` with `p.prosrc` (body only, no `CREATE` wrapper)
- `source.py`: Normalize `language` to uppercase at `BaseProcedure` construction time

**Test infrastructure:**
- `compare_metadata_json.py`: Add `auditStamp.time` to `default_exclude_paths` — it's a wall-clock timestamp (same as `lastObserved`), and 6+ test suites had redundant per-test exclusions for it

**Tests:**
- Add ETL tables and `etl_process_orders` procedure to integration test setup
- Update integration golden files with table-level and column-level lineage
- Add unit test for `get_procedures_for_schema` and a Postgres lineage golden file

## Test plan

**Unit test** (`test_postgres_source.py`): Verifies field mapping, uppercase language normalization, and `prosrc` body-only behavior. Runs without a database.

**Lineage golden file** (`etl_procedure.sql`): Verifies end-to-end SQL parsing of a Postgres procedure body into correct `dataJobInputOutput` aspects with table and column lineage.

**Integration golden files**: Verified against a live Postgres 15.2 container to confirm the full pipeline produces expected output including `etl_process_orders` lineage from `raw_orders` to `processed_orders`.


metadata-ingestion/src/datahub/ingestion/source/sql/postgres/source.py
```diff
@@ -645,7 +645,7 @@ def get_procedures_for_schema(
                         p.proname AS name,
                         l.lanname AS language,
                         pg_get_function_arguments(p.oid) AS arguments,
-                        pg_get_functiondef(p.oid) AS definition,
+                        p.prosrc AS definition,
```
kyungsoo-datahub, line 648: Please consider updating "Other Notable Changes" in docs/how/updating-datahub.md (or the connector changelog pattern you use) so anyone with custom tooling that assumed queryStatement.value was always pg_get_functiondef-shaped DDL can adjust. Not a merge blocker unless we know of an internal consumer that depends on full DDL in that field.

kyungsoo-datahub (approved):
Thanks for fixing the CREATE-wrapper skip and the sql vs SQL mismatch; the approach (prosrc + normalizing language at BaseProcedure construction) matches how lineage gating works in generate_procedure_lineage. Tests and goldens look thorough.

LGTM with one minor comment.

metadata-ingestion/src/datahub/ingestion/source/sql/postgres/source.py
jinhyukchang: Sure and added.

jinhyukchang:
@kyungsoo-datahub It seems mergeable is stuck after hours of waiting. Tried couple times but the same. Is this something you could help?

kyungsoo-datahub:
> @kyungsoo-datahub It seems mergeable is stuck after hours of waiting. Tried couple times but the same. Is this something you could help?
 
The process is retriggered. I'll check when it's done. Thank you.

PR #16872, opened 2026-04-01
test(aspect migration mutator): Automated tests for implementations o…

gdatahub:
…f Abstract AspectMigrationMutators

**Automated tests for implementations of Abstract AspectMigrationMutators**
  
  Two approaches: 
   1. Reflection-based (fully automatic): discovers all AspectMigrationMutator implementations and runs tests automatically.
   2. Template pattern-based (semi-automatic): uses a base test class as a template; each implementation must create a test class extending it.


entity-registry/src/test/java/com/linkedin/metadata/aspect/hooks/AspectMigrationMutatorBaseTest.java
```diff
@@ -0,0 +1,234 @@
...
+ *     Owner migrated = (Owner) result;
+ *     assertEquals(migrated.getNewField(), "expected value");
+ *     assertFalse(migrated.hasOldField());
+ *   }
+ * }
+ * }</pre>
+ */
+public abstract class AspectMigrationMutatorBaseTest {
+
+  private static final Urn DATASET_URN =
```
chakru-r: This wont make sense in the base class -- isnt it? Should be supplied by the subclass

entity-registry/src/test/java/com/linkedin/metadata/aspect/hooks/AspectMigrationMutatorBaseTest.java
```diff
@@ -0,0 +1,234 @@
...
+
+  // Shared
+
+  protected EntityRegistry entityRegistry;
+  protected RetrieverContext retrieverContext;
+
+  @BeforeMethod
+  public void setUp() {
+    entityRegistry = new TestEntityRegistry();
+    AspectRetriever mockRetriever = mock(AspectRetriever.class);
```
chakru-r: Will need this retriver mocks behaviour be customized by each subclass. Some of them may need to use the retriever to compute the transform.

entity-registry/src/test/java/com/linkedin/metadata/aspect/hooks/AspectMigrationMutatorBaseTest.java
```diff
@@ -0,0 +1,234 @@
...
+
+    List<Pair<ChangeMCP, Boolean>> results =
+        mutator().writeMutation(List.of(item), retrieverContext).collect(Collectors.toList());
+
+    assertTrue(results.get(0).getSecond(), "Expected mutated=true for null schemaVersion (=1)");
+    assertEquals((long) item.getSystemMetadata().getSchemaVersion(), mutator().getTargetVersion());
+  }
+
+  @Test
+  public void writeMutation_alreadyAtTargetVersion_isNoOp() {
```
chakru-r: this isnt really testing the mutator implementation -- this is pretty much the base class test, so doesnt need to run on each subclass. Same with a few other such methods. All we need to test here - is use sourceAspect and assertTransform with suitable retriever mocking (which is very mutator implementation specific). The retriever may be used by the mutator to fetch some other value in order to compute what the transform should be. Rest are really base class tests

entity-registry/src/test/java/com/linkedin/metadata/aspect/hooks/AspectMigrationMutatorContractTest.java
```diff
@@ -0,0 +1,232 @@
...
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
+import org.reflections.Reflections;
+import org.reflections.scanners.SubTypesScanner;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+/**
+ * Reflection-based test.
```
chakru-r: Lets drop this. This isnt really testing anything meaingul of the subclass. Its all  base class tests and not really testing the transform in anyway.

chakru-r (changes requested):

entity-registry/src/test/java/com/linkedin/metadata/aspect/hooks/AspectMigrationMutatorContractTest.java
gdatahub: done

entity-registry/src/test/java/com/linkedin/metadata/aspect/hooks/AspectMigrationMutatorBaseTest.java
gdatahub: Good Point. Added a method `configureRetrieverMock(@Nonnull AspectRetriever retrieverMock) `that is invoked at the end of setUp(), this hook handles subclasses that need common retriever setup across all tests .

 A subclass whose mutator calls
  retrieverContext.getAspectRetriever().getLatestAspect(...) can override this and add stubs:
```
  @Override
  protected void configureRetrieverMock(AspectRetriever retrieverMock) {
      when(retrieverMock.getLatestAspect(any(), eq("ownership")))
          .thenReturn(someAspect);
  }

```

entity-registry/src/test/java/com/linkedin/metadata/aspect/hooks/AspectMigrationMutatorBaseTest.java
gdatahub: Removed the static field. Added **abstract method Urn entityUrn()**, each test mutator need to implement this.

entity-registry/src/test/java/com/linkedin/metadata/aspect/hooks/AspectMigrationMutatorBaseTest.java
gdatahub: Made the following changes:

- Removed **writeMutation_alreadyAtTargetVersion_isNoOp**, **readMutation_alreadyAtTargetVersion_isNoOp**, and **writeMutation_futureVersion_isNoOp**, as these base-class version-gating tests are already covered by AspectMigrationMutatorTest.
- Added configureRetrieverMock(AspectRetriever) hook so subclasses can stub retriever behavior specific to their **transform** logic. mockRetriever is also exposed as a field for per-test verify() calls.
- Kept contract_targetVersionIsSourceVersionPlusOne and contract_sourceVersionAtLeastDefault as lightweight checks to ensure each concrete implementation correctly declares its version numbers.

PR #16873, opened 2026-04-01
fix(ingest/unity): emit ownership and datasetProperties as standard MCPWs with incremental config

max-datahub:
## Summary

The Unity (Databricks) connector emitted ownership and datasetProperties as raw patch MCPs
(`MetadataChangeProposalClass` via `mcp_raw`), which bypassed the transformer pipeline. All
`SingleAspectTransformer`-based transformers — including `pattern_cleanup_ownership` — received
`aspect=None` for Unity-sourced datasets and silently did nothing.

This PR moves both aspects into standard `MetadataChangeProposalWrapper` emission (UPSERT),
restoring transformer compatibility. For users who need the previous additive/merge semantics,
new `incremental_ownership` and `incremental_properties` config flags convert MCPWs to patch
MCPs via workunit processors — following the same pattern as Snowflake's `auto_incremental_lineage`
and `auto_incremental_properties`.

### Changes

- **Ownership and datasetProperties** now emit via `MetadataChangeProposalWrapper.construct_many()`
  instead of raw patch MCPs, making them visible to the transformer pipeline.
- **New `incremental_ownership_helper.py`**: `auto_incremental_ownership` workunit processor and
  `IncrementalOwnershipConfigMixin`, mirroring the existing `incremental_lineage_helper.py` and
  `incremental_properties_helper.py` patterns.
- **Config**: Added `IncrementalOwnershipConfigMixin` and `IncrementalPropertiesConfigMixin` to
  `UnityCatalogSourceConfig`. Both default to `False` (UPSERT), matching the standard behavior of
  every other connector.
- **Tests**: 3 unit tests for `auto_incremental_ownership` covering UPSERT passthrough, PATCH
  conversion with payload verification, and container entity-type exclusion.

### Breaking Change

Ownership and datasetProperties now use UPSERT (replace) semantics by default, consistent with
all other connectors. Users who need additive/merge behavior can set `incremental_ownership: true`
and/or `incremental_properties: true` in their recipe config.

## Checklist
- [x] The PR conforms to DataHub's [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (specifically the [PR Title Format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#pr-title-format))
- [x] Links to related issues: Fixes transformer incompatibility with Unity ownership aspects
- [x] Tests updated: Golden files regenerated, unit tests added for incremental ownership processor
- [ ] Docs added/updated (if applicable)

## Test plan
- [x] Lint passes (`./gradlew :metadata-ingestion:lintFix`)
- [x] Unity integration tests pass (2/2) with updated golden files
- [x] Unity unit tests pass (46/46)
- [x] New `test_incremental_ownership_helper.py` tests pass (3/3)
- [x] No stale references to removed imports


metadata-ingestion/src/datahub/ingestion/source/unity/source.py
```diff
@@ -683,31 +686,17 @@ def process_table(self, table: Table, schema: Schema) -> Iterable[MetadataWorkUn
...
-                yield MetadataWorkUnit(
-                    id=f"{dataset_urn}-{patch_mcp.aspectName}", mcp_raw=patch_mcp
-                )
-
         yield from [
             mcp.as_workunit()
             for mcp in MetadataChangeProposalWrapper.construct_many(
                 entityUrn=dataset_urn,
                 aspects=[
+                    table_props,
```
treff7es, line 704: This will be a breaking change because earlier we emitted owners and table props as patch while now we are overwriting existing/user assigned ones.

treff7es (changes requested):

metadata-ingestion/src/datahub/ingestion/source/unity/source.py
max-datahub: @treff7es, if we want to address the issue at hand, transformers not working correctly, how do you propose we address it in this case without breaking existing behavior?
The way I see it, there are a couple of options:
>   Option A: auto_incremental_ownership workunit processor (recommended)
> 
>   Follow the exact pattern established by auto_incremental_lineage and auto_incremental_properties:
> 
>   1. Source emits ownership as standard MCPW (transformer-compatible)
>   2. Transformers run and can modify ownership normally
>   3. A new auto_incremental_ownership workunit processor converts the MCPW to a patch MCP after
>   transformers run
>   4. Config flag incremental_ownership: bool = True (default True to preserve current additive
>   behavior)
> 
>   The processing pipeline would be:
>   Source (MCPW) → Transformers → auto_incremental_ownership → Emitter (patch MCP)
> 
>   This is exactly how Snowflake does it for properties (snowflake_v2.py:519-521) and lineage
>   (snowflake_v2.py:516-518). The helper would live in a new incremental_ownership_helper.py
>   alongside the existing incremental_lineage_helper.py and incremental_properties_helper.py.
> 
>   Pros: Clean, follows established patterns, fixes transformers AND preserves patch semantics,
>   reusable by other connectors
>   Cons: New helper file + config mixin, slightly more work
> 
>   ---
>   Option B: Fix the transformer framework to handle raw MCPs
> 
>   Add a branch in BaseTransformer.transform() (~line 566) that deserializes raw
>   MetadataChangeProposalClass records into their aspect class, calls transform_aspect(), then
>   re-serializes.
> 
>   Pros: Fixes the problem for ALL sources and ALL transformers at once
>   Cons: Larger blast radius, JSON Patch format is harder to deserialize cleanly (it's operations,
>   not a full aspect), risk of subtle breakage across the framework

sgomezvillamor: My understanding is that the standard connector behavior is to upsert aspects, ownership and table props included, with patching as an opt-in via config. So while this is a breaking change, I think the previous behavior was sort of buggy (or at least non-standard). We should move toward the correct standard: upsert by default, patch via optional config. IMO even at the cost of a breaking change, as long as there's a way to preserve the old behavior, of course.

On processor vs. transformer, no strong opinion. They're different abstractions solving the same problem: avoiding reimplementing the same logic in every connector. Unless I'm missing something, patching complexity should be the same either way.

One minor correction though:

> Source (MCPW) → Transformers → auto_incremental_ownership → Emitter (patch MCP)

This looks wrong to me. AFAIK auto processors run before transformers, not after.

max-datahub: > This looks wrong to me. AFAIK auto processors run before transformers, not after.

@sgomezvillamor you are correct. Apologies for not fact-checking.

  
I pushed an updated commit:

> What changed:
> Instead of simply switching ownership/properties from PATCH to UPSERT (which breaks existing users), this introduces an auto_incremental_ownership workunit processor — the same pattern used by auto_incremental_lineage and auto_incremental_properties in the Snowflake connector.
> 
> How it works:
> 1. Source now emits ownership and datasetProperties as standard MCPWs (transformer-compatible)
> 2. auto_incremental_ownership and auto_incremental_properties processors convert MCPWs to patch MCPs when the respective config flags are enabled
> 3. Default is UPSERT (standard behavior, matching every other connector)
> 4. Users who need the old additive semantics set incremental_ownership: true and/or incremental_properties: true in their recipe
> 
> Addressing the reviewer concerns:
> - @treff7es: The breaking change is now opt-in in the other direction — UPSERT by default (standard), PATCH via config. Users who relied on additive ownership have a clear migration path.
> - @sgomezvillamor: Agreed that UPSERT should be the standard default. Also corrected the processor ordering — as you noted, processors run before transformers (not after), which is the same trade-off that exists for Snowflake's incremental_lineage. When incremental mode is off (default), transformers work correctly on ownership.

docs/how/updating-datahub.md
```diff
@@ -39,6 +39,7 @@ This file documents any backwards-incompatible changes in DataHub and assists pe
...
   - **Labels:** When multiple MBeans map to the same metric name, series may include an **`_objectname`** label; adjust queries or aggregations if you previously assumed a single series per name.
 - #16723 (Ingestion) Dataplex source configuration cleanup: `filter_config.entries.dataset_pattern` was removed, use `filter_config.entries.pattern` instead; `entries_location` was removed, use `entries_locations` (list) instead.
+- #16873 (Ingestion) Unity Catalog (Databricks) connector now emits ownership and datasetProperties as UPSERT (full replace) by default, instead of incremental PATCH. This means each ingestion run will re-state the complete ownership and properties for each dataset. If you rely on the previous additive/merge behavior (e.g., manually added owners surviving ingestion runs), set `incremental_ownership: true` and/or `incremental_properties: true` in your recipe config.
```
sgomezvillamor, line 42: We are fixing consistency across connectors at the cost of introducing a breaking change.
That sounds good to me.
Are you ok with this @treff7es ?

metadata-ingestion/src/datahub/ingestion/api/incremental_ownership_helper.py
```diff
@@ -0,0 +1,55 @@
...
+        yield from stream
+        return  # early exit
+
+    for wu in stream:
+        urn = wu.get_urn()
+
+        if (
+            isinstance(wu.metadata, MetadataChangeProposalWrapper)
+            and isinstance(wu.metadata.aspect, OwnershipClass)
+            and wu.metadata.entityType == "dataset"
```
sgomezvillamor, line 39: how other _incremental_ auto processors handle this entity type scope support?

At the very least, we should document that the processor only works for "dataset" entities.

metadata-ingestion/src/datahub/ingestion/api/incremental_ownership_helper.py
```diff
@@ -0,0 +1,55 @@
...
+        ):
+            ownership_aspect = wu.metadata.aspect
+            if ownership_aspect.owners:
```
sgomezvillamor, line 42: When owners is empty, the workunit is silently dropped. In terms of patching (incremental), that sounds good.
Please double check how other _incremental_ auto processors handle similar scenario.

metadata-ingestion/src/datahub/ingestion/api/incremental_ownership_helper.py
```diff
@@ -0,0 +1,55 @@
...
+) -> Iterable[MetadataWorkUnit]:
+    if not incremental_ownership:
+        yield from stream
+        return  # early exit
+
+    for wu in stream:
+        urn = wu.get_urn()
+
+        if (
+            isinstance(wu.metadata, MetadataChangeProposalWrapper)
```
sgomezvillamor, line 37: Unlike `auto_incremental_properties` and `auto_incremental_lineage` (both handle `MetadataChangeEventClass`), `auto_incremental_ownership` only handles MCPW. Sources emitting ownership via MCE (e.g. Glue appends OwnershipClass to MCE snapshots) would silently bypass conversion and emit as UPSERT when `incremental_ownership=True`. 

Not an issue for Unity today, but limits reusability of the helper without a clear comment explaining the scope.

Can you add support for MCE and the corresponding test?

metadata-ingestion/tests/unit/test_incremental_ownership_helper.py
sgomezvillamor, line 1: Few missing tests — add to `test_incremental_ownership_helper.py`:

* `test_enabled_with_empty_owners` — documents the silent-drop behavior for owners=[] - this one depends on comment above
* `test_enabled_preserves_system_metadata` — verifies `system_metadata` flows through to the PATCH MCP
* `test_enabled_passes_through_non_ownership_aspects` — confirms the else: yield wu branch works

sgomezvillamor (commented):
The breaking change is acceptable IMO, let's wait on @treff7es concerns, if any.

Overall the implementation LGTM, I just left some final comments/suggestions.

sgomezvillamor:
@max-datahub are you planning to take this over? or should this just be closed?

max-datahub:
@sgomezvillamor I can clean up the branch and address your remaining comments next week. I was waiting for feedback from @treff7es to do so, but since there is none, I assume that we are okay otherwise. I will ping you once I am done

metadata-ingestion/src/datahub/ingestion/api/incremental_ownership_helper.py
max-datahub: > _[Claude (Claude Code) responding on Max's behalf — pushed in 897e4e1.]_

Good catch. The reason this differs from `auto_incremental_properties` is that `DatasetPropertiesClass` is entity-bound at the aspect level (only attached to datasets), while `OwnershipClass` is attached to ~28 entity types per the entity registry — datasets, containers, charts, dashboards, glossary terms, ML models, etc. Filtering by aspect type alone wouldn't be enough here: without the explicit dataset filter, a container `OwnershipClass` workunit would be routed through `create_dataset_owners_patch_builder` → `DatasetPatchBuilder` and produce a malformed patch.

`auto_incremental_lineage` has the same shape of problem and solves it by dispatching across entity-specific patch builders (`DatasetPatchBuilder` / `ChartPatchBuilder` / `DashboardPatchBuilder`); same extension path is open here when a future consumer needs non-dataset ownership patches.

Documented this in a module docstring in the latest commit. Non-dataset ownership passes through as UPSERT.

metadata-ingestion/src/datahub/ingestion/api/incremental_ownership_helper.py
max-datahub: > _[Claude (Claude Code) responding on Max's behalf — pushed in 897e4e1.]_

Verified, and the silent drop turns out to be load-bearing rather than stylistic. `create_dataset_owners_patch_builder` emits one patch op per owner — empty `owners` yields zero ops, and `MetadataPatchProposal.build()` returns `[]` (one MCP per `aspect_name` in `self.patches`). Without the guard, `next(iter(patch_builder.build()))` would raise `StopIteration`.

`auto_incremental_properties` doesn't hit this because `create_dataset_props_patch_builder` unconditionally calls seven `set_*`/`add_*` methods regardless of input, so `build()` always returns ≥1 MCP. So the two helpers diverge for a structural reason, not a behavioral inconsistency.

Documented in the module docstring and pinned with `test_enabled_with_empty_owners_drops_workunit` in the latest commit.

metadata-ingestion/src/datahub/ingestion/api/incremental_ownership_helper.py
max-datahub: > _[Claude (Claude Code) responding on Max's behalf — pushed in 897e4e1.]_

Added in the latest commit. The MCE branch mirrors `auto_incremental_properties` exactly: pull `OwnershipClass` off the snapshot, yield the trimmed MCE if other aspects remain, emit the ownership separately as a PATCH MCP. Same dataset-only filter applies (URN prefix check `urn:li:dataset:`), so a Glue-style MCE on a non-dataset entity passes through unchanged.

Test: `test_enabled_converts_dataset_ownership_in_mce_to_patch`.

metadata-ingestion/tests/unit/test_incremental_ownership_helper.py
max-datahub: > _[Claude (Claude Code) responding on Max's behalf — pushed in 897e4e1.]_

All three added in the latest commit:

- `test_enabled_with_empty_owners_drops_workunit` — pins the silent-drop behavior (see the empty-owners thread for the reasoning)
- `test_enabled_preserves_system_metadata` — verifies `systemMetadata` flows through to the PATCH MCP
- `test_enabled_passes_through_non_ownership_aspects` — covers the `else: yield wu` branch via a Status workunit

Plus `test_enabled_converts_dataset_ownership_in_mce_to_patch` for the new MCE branch.

max-datahub:
@sgomezvillamor this PR is ready for re-review

sgomezvillamor (approved):
Solid fix — architecture is correct, all substantive concerns addressed, tests are thorough. Two minor nits, neither blocks merge:

1. `incremental_properties_helper.py:58`: `# TODO: Use this in SQLCommonConfig. Currently only used in snowflake` — Unity now also uses this mixin, so the comment is stale.

2. In `auto_incremental_ownership`, prefer `guess_entity_type(urn) == DatasetUrn.ENTITY_TYPE` over `urn.startswith("urn:li:dataset:")` (MCE branch) and `DatasetUrn.ENTITY_TYPE` over the `"dataset"` string literal (MCPW branch).

---

treff7es (approved):

PR #16879, opened 2026-04-01
fix(ingest/powerbi): apply lowercase normalization to all upstream lineage URN code paths

aviraj-gour:
### Summary

- Fix inconsistent URN casing for PowerBI upstream lineage to BigQuery (and other platforms) when convert_lineage_urns_to_lowercase=True (the default)
- The standard M-Query accessor path correctly lowercased URNs via make_urn(), but the native SQL query path (Value.NativeQuery(...)) bypassed lowercasing because the shared SQL parser returns original-casing URNs for BigQuery (marked as case-sensitive in PLATFORMS_WITH_CASE_SENSITIVE_TABLES)
- Apply lineage_urn_to_lowercase() at the consumption point in extract_lineage() and make_fine_grained_lineage_class(), catching URNs from all code paths in one place

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/powerbi/powerbi.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/powerbi/powerbi.py
@@ -203,7 +203,6 @@ class Mapper:
                     else column_ref.column,
                 )
                 for column_ref in cll_info.upstreams
-                if column_ref.column
             ]
 
             fine_grained_lineages.append(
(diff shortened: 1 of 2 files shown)


metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/pattern_handler.py
```diff
@@ -362,8 +362,10 @@ def parse_custom_sql(
             )
         )
 
-        query = native_sql_parser.remove_drop_statement(query)
+        # remove_special_characters must run first to expand #(lf) → \n before
+        # remove_drop_statement applies line-anchored patterns (USE, GO, SET, etc.)
         query = native_sql_parser.remove_special_characters(query)
+        query = native_sql_parser.remove_drop_statement(query)
```
askumar27, line 368: Good catch on the ordering!

askumar27 (approved):
Good bug fix!

metadata-ingestion/src/datahub/ingestion/source/powerbi/powerbi.py
```diff
@@ -195,8 +195,14 @@ def make_fine_grained_lineage_class(
...
             upstreams = [
-                builder.make_schema_field_urn(column_ref.table, column_ref.column)
+                builder.make_schema_field_urn(
+                    self.lineage_urn_to_lowercase(column_ref.table),
+                    column_ref.column.lower()
+                    if self.__config.convert_lineage_urns_to_lowercase
+                    else column_ref.column,
+                )
                 for column_ref in cll_info.upstreams
+                if column_ref.column
```
askumar27: `ColumnRef.column` is typed as str (non-optional), so if empty values occur, it indicates an upstream parsing bug that should be surfaced, not masked.

PR #16889, opened 2026-04-02
fix(ingestion/sql-parsing): fix multi-part table name truncation for 5+ part paths

acrylJonny:
## Summary
Fixes a bug in the SQL parser where table paths with 5 or more components (e.g. `"src"."ns"."sub"."schema"."table"`) silently dropped middle path components, preventing schema resolution and column-level lineage.

**Root cause:** `Dot` expressions in sqlglot are left-associative — `a.b.c` parses as `Dot(Dot(a, b), c)`. The traversal loop in both `_table_name_from_sqlglot_table` (`sqlglot_lineage.py`) and `_TableName.from_sqlglot_table` (`_models.py`) iterated right with `exp = exp.expression` but collected names with `exp.this.name`. For a nested inner `Dot`, `Dot.name` returns only the *leaf* identifier — so a 5-part path like `catalog.ns.sub.schema.table` resolved as `catalog.ns.schema.table`, with `sub` silently lost.

**Fix:** Collect `exp.expression` (right side) at each level and reverse. This correctly handles any path depth and preserves MSSQL temp table prefix logic on the rightmost identifier.

**Impact:** Any connector ingesting datasets with 5+ part paths (e.g. Dremio with deep folder hierarchies) would produce wrong URNs from SQL-parsed table references, causing all schema lookups to fail and producing no column-level lineage.

## Changes
- `sqlglot_lineage.py`, `_models.py`: fix the Dot traversal in both copies of the table-name extraction logic
- `test_table_name.py`: corrected existing `TestDotExpressionTableNames` tests — they were manually constructing right-nested `Dot(a, Dot(b, c))` structures, which sqlglot never actually produces (it always produces left-nested). Updated to left-nested to match real parser output.
- Added `test_5part_path_preserves_all_components` — regression test using real SQL parsing confirming no middle component is dropped
- Added `test_deep_path_up_to_10_parts` — confirms correctness at arbitrary depth

## Test plan
- [x] `pytest tests/unit/sql_parsing/test_table_name.py` — all 53 tests pass
- [x] `pytest tests/unit/sql_parsing/` — all 439 tests pass
- [x] `pytest tests/unit/dremio/` — all 152 tests pass


metadata-ingestion/src/datahub/sql_parsing/_models.py
```diff
@@ -112,13 +112,16 @@ def from_sqlglot_table(
                 )
 
         if isinstance(table.this, sqlglot.exp.Dot):
-            # Multi-part tables (>3 parts) have extra parts in a Dot expression
+            # Multi-part tables (>3 parts) have extra parts in a Dot expression.
+            # Dot is left-associative (a.b.c = Dot(Dot(a,b),c)), so collect right-side
+            # identifiers while walking left, then reverse.
             parts = []
             exp = table.this
             while isinstance(exp, sqlglot.exp.Dot):
```
kyungsoo-datahub, line 120: Mirror of this traversal lives in sqlglot_lineage.py — that's why the bug existed in two places. Out of scope here, but worth a follow-up to deduplicate (e.g., let from_sqlglot_table take an optional leaf-name transform so sqlglot_lineage.py becomes a thin wrapper).

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
```diff
@@ -173,26 +173,36 @@ def _table_name_from_sqlglot_table(
...
             )
 
-    # Handle Dot expressions (more than 3-part names)
+    # Handle Dot expressions (more than 3-part names).
+    # Dot is left-associative (a.b.c = Dot(Dot(a,b),c)), so collect right-side
+    # identifiers while walking left, then reverse.
     if isinstance(table.this, sqlglot.exp.Dot):
-        parts = []
-        exp = table.this
+        all_parts_exp: List[sqlglot.exp.Expression] = []
```
kyungsoo-datahub, line 181: Why insert(0, ...) here when _models.py uses append + reverse? Same algorithm, two patterns; the _models.py version is cleaner and O(n) instead of O(n²). Worth aligning before merge.

kyungsoo-datahub (commented):
This changes the URN format for datasets with 5+ part paths. Existing entities in DataHub will have the old (truncated) URN; re-ingestion after this lands will create new entities at the correct URN, leaving the originals as orphans with all their tags, owners, lineage, etc. stuck on the wrong entity.

Could you add an entry to `docs/how/updating-datahub.md` calling this out? Users with Dremio (or any deep-path setup) need to know to expect duplicates post-upgrade and clean up the old URNs manually.

The PR template checklist for "Breaking changes documented in updating-datahub.md" is currently unchecked.

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
acrylJonny: updated

metadata-ingestion/src/datahub/sql_parsing/_models.py
acrylJonny: Addressed as a follow-up in #17513. That PR adds an optional `leaf_name_transform: Callable[[sqlglot.exp.Expression], str]` to `_TableName.from_sqlglot_table` and reduces `_table_name_from_sqlglot_table` to an 8-line wrapper that injects `_restore_mssql_temp_table_prefix` as the leaf transform. SemanticView+Dot propagation is covered by a new `TestLeafNameTransform` class. Kept this PR focused on the bug fix; the dedup goes in once #16889 is merged.

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
acrylJonny: Aligned to `append + reverse` in 1c67cb8 — both files now use the same O(n) pattern. The duplication itself is removed in the follow-up #17513, which collapses this branch into a leaf-transform wrapper around `_TableName.from_sqlglot_table`.

kyungsoo-datahub (approved):
LGTM.

One nit, not blocking: the formatting-only changes in scripts/dev/tests/* are unrelated to the SQL parsing fix. Would be cleaner as a separate PR so the diff stays focused on the bug fix. Up to you whether to strip them here or land as-is.

PR #16902, opened 2026-04-03
fix(actions): route sub-ERROR logs to stdout for observability

askumar27:
## Summary

Route DataHub Actions console logging so DEBUG / INFO / WARNING go to **stdout** and ERROR / CRITICAL go to **stderr**, instead of sending every level to stderr by default.

## Motivation

`logging.StreamHandler()` with no stream writes to **stderr**. Many log pipelines treat **stderr as error severity**, so operational messages (INFO, etc.) can appear as errors even though they are not. Splitting streams matches common Unix / container logging expectations and makes severity alignment with aggregators more reliable.

## Changes overview

**Bug fixes**

- **datahub-actions:** Replace the single default `StreamHandler` on the `datahub_actions` logger with two handlers: stdout (levels below ERROR, via a small `logging.Filter`) and stderr (ERROR and above).

**New tests**

- **`datahub-actions/tests/unit/test_entrypoints_logging.py`:** Unit coverage for handler targets, the below-ERROR filter, and end-to-end routing with patched `sys.stdout` / `sys.stderr`.

**Refactoring**

- Handler setup is **inline** in the `datahub_actions` Click callback (no separate attach helper).

**Dependencies**

- None.

## Architecture / design notes

- **Mutually exclusive routing:** Each record is intended to go to **one** stream (no duplicate ERROR lines on both stdout and stderr).
- **`datahub_actions` logger** behavior is unchanged besides handlers: still isolated from the root logger (`propagate = False`) for the same reasons as before.
- **`logging.basicConfig`** at import time is unchanged.

## Testing

- New unit tests invoke the real group **`callback`** under **`click.Context.scope()`** so `@pass_context` behaves as in production.
- Scenarios: two handlers on the logger; filter accepts INFO and rejects ERROR; WARNING vs CRITICAL on the filter class; INFO vs ERROR landing in patched stdout/stderr buffers.
- Run via `./gradlew :datahub-actions:lint` and `pytest tests/unit/test_entrypoints_logging.py` (or full `:datahub-actions:testFull` in CI).

## Impact assessment

| Area | Notes |
|------|--------|
| **Affected components** | `datahub-actions` CLI entrypoint logging only (`entrypoints.py`). |
| **Breaking changes** | None for APIs or configs; **operational** change only (where log lines are emitted). |
| **Performance** | Negligible (one extra handler + filter per record on the stdout path). |
| **Risk** | **Low:** behavior change is limited to stream selection; semantics and format string are unchanged. |

## Deployment notes

- No migrations, flags, or new env vars.
- Operators or dashboards that **assumed all Actions logs on stderr** may need to ingest **stdout** for non-error logs (typical k8s setups already collect both).
- Rollback: revert to prior single-stderr handler behavior.

### DIFF
--- a/datahub-actions/src/datahub_actions/entrypoints.py
+++ b/datahub-actions/src/datahub_actions/entrypoints.py
@@ -35,6 +35,27 @@ logging.basicConfig(format=BASE_LOGGING_FORMAT)
 MAX_CONTENT_WIDTH = 120
 
 
+class _BelowErrorFilter(logging.Filter):
+    """Restrict a handler to DEBUG / INFO / WARNING so ERROR+ use a separate stream."""
+
+    def filter(self, record: logging.LogRecord) -> bool:
+        return record.levelno < logging.ERROR
+
+
+def _attach_datahub_actions_console_handlers(datahub_logger: logging.Logger) -> None:
+    """Route sub-error logs to stdout and ERROR+ to stderr (avoids aggregators marking INFO as errors)."""
+    formatter = logging.Formatter(BASE_LOGGING_FORMAT)
+    stdout_handler = logging.StreamHandler(sys.stdout)
+    stdout_handler.setLevel(logging.DEBUG)
+    stdout_handler.addFilter(_BelowErrorFilter())
+    stdout_handler.setFormatter(formatter)
+    stderr_handler = logging.StreamHandler(sys.stderr)
+    stderr_handler.setLevel(logging.ERROR)
+    stderr_handler.setFormatter(formatter)
+    datahub_logger.addHandler(stdout_handler)
+    datahub_logger.addHandler(stderr_handler)
+
+
 @click.group(
     context_settings=dict(
         # Avoid truncation of help text.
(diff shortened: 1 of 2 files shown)


datahub-actions/src/datahub_actions/entrypoints.py
```diff
@@ -87,11 +94,17 @@ def datahub_actions(
...
-    datahub_logger.addHandler(stream_handler)
+    stdout_handler = logging.StreamHandler(sys.stdout)
+    stdout_handler.setLevel(logging.DEBUG)
+    stdout_handler.addFilter(_BelowErrorFilter())
+    stdout_handler.setFormatter(formatter)
+    stderr_handler = logging.StreamHandler(sys.stderr)
+    stderr_handler.setLevel(logging.ERROR)
+    stderr_handler.setFormatter(formatter)
+    datahub_logger.addHandler(stdout_handler)
+    datahub_logger.addHandler(stderr_handler)
```
skrydal, line 107: I agree with this change. Have you thought about doing this at root level rather than just for actions?

datahub-actions/tests/unit/test_entrypoints_logging.py
skrydal, line 1: Those tests are very complex compared with:
1. Complexity of the solution
2. The nature of the change (logging...)

In my opinion this is over-testing. It would be enough to have a single test in non-dedicated flag emitting 2 levels of log and checking the output stream. There is no need for so many asserts on handlers and filters being present.

skrydal (changes requested):
LGTM but please consider reducing tests.

datahub-actions/tests/unit/test_entrypoints_logging.py
askumar27: Removed redundant tests

datahub-actions/src/datahub_actions/entrypoints.py
askumar27: I scoped it to `datahub_actions` because this logger already has propagate=False specifically to insulate from third-party root logger changes. Applying at root level would route stdout/stderr for every library too, which could be noisy. Happy to broaden in a follow-up PR if you think it's worth it though.

skrydal (approved):

PR #16903, opened 2026-04-03
feat(): Support context document import from local filesystem and GitHub, Notion, Confluence

jjoyce0510:
## Summary

Adding a PR to introduce support for importing context inside the Documents section from local filesystem (one-time), GitHub (ingestion), Notion (ingestion), and Confluence (ingestion).

Also, extend the Notion + Confluence sources to support configuring the document source type, so you can ingest as a native document if you want to import into the editable experience under documents.

See the video below for the full experience. 

## Video Walkthrough

https://github.com/user-attachments/assets/bd197d97-9a21-4572-bcd4-9121c4a2a518

## Screenshots

Screenshot 2026-05-28 at 3 46 51 PM
Screenshot 2026-05-28 at 3 14 05 PM
Screenshot 2026-05-28 at 3 12 30 PM
Screenshot 2026-05-28 at 3 06 42 PM
Screenshot 2026-05-28 at 3 06 35 PM
Screenshot 2026-05-28 at 3 06 28 PM

## Status

Ready for review.


alexsku (commented):
## PR Review

**Type**: behavior change (new feature)
**Size**: +4432 / -60 across 49 files

This PR adds the ability to import context documents into DataHub from two sources: local file upload (via browser-side text extraction) and GitHub repositories (via the GitHub REST API). It introduces a new `DocumentImportService` backend, three new GraphQL resolvers (import from files, import from GitHub, preview from GitHub), a React modal UI with source selection, and parent document hierarchy support. It also adds `ignore_above` to keyword mappings in ES to prevent indexing failures on long text values.

**Key invariants:**
1. Authorization: All import operations require `MANAGE_DOCUMENTS` platform privilege
2. Idempotency: Documents are created with deterministic IDs derived from source IDs, so re-imports update rather than duplicate
3. Hierarchy: GitHub folder structure is preserved as parent-child document relationships; candidates must be ordered parents-first
4. The GitHub token is passed per-request and not stored server-side

---

## Risk Assessment

| | |
|---|---|
| **Risk** | Medium |
| **Blast radius** | Document entities only; ES mapping change affects all entities with TEXT/TEXT_PARTIAL fields |
| **Rollback** | Safe for import feature (new code, no migrations). ES mapping change (`ignore_above`) is additive and backward-compatible |
| **Rollout** | Ship it — import resolvers are null-guarded if factory bean is absent |
| **CI** | Failing: `build-and-test (except_metadata_ingestion, UTC)` |

---

## Blocking Issues

None found.

## High-Priority Issues

### [HIGH] GitHub token passed through GraphQL — potential for logging/persistence
**Category**: security | **Confidence**: medium
**Location**: `datahub-graphql-core/src/main/resources/documents.graphql:871` — `githubToken: String!`

The GitHub PAT is passed as a plain `String!` in the GraphQL input. While it's used only server-side and not stored, GraphQL query logging, APM tracing, or error serialization could inadvertently capture the token in server logs. The same `ImportDocumentsFromGitHubInput` type is reused for both the `previewDocumentsFromGitHub` query and the `importDocumentsFromGitHub` mutation, meaning the token is sent even for read-only preview operations.

**Suggested fix**: Ensure GraphQL request logging is configured to redact `githubToken`. Consider whether the preview query truly needs the token in the same input type (it does for private repos, so this is likely acceptable). Document the logging risk.
**How to validate**: Check if the DataHub GMS logging configuration logs full GraphQL variables. Search for query logging middleware.

### [HIGH] No limit on number of files fetched from GitHub — unbounded server-side work
**Category**: performance | **Confidence**: high
**Location**: `metadata-service/services/src/main/java/com/linkedin/metadata/service/docimport/GitHubDocumentSource.java:166-187` — `fetchDocuments()`

The `fetchDocuments` method iterates over all matching files from the GitHub tree API and fetches each file's content one-by-one via individual HTTP requests (line 167: `fetchFileContent`). For a large repo with thousands of matching files, this creates an unbounded number of sequential HTTP requests from the GMS server, tying up a GraphQL worker thread for potentially minutes. There's no configurable limit or pagination.

**Suggested fix**: Add a `maxFiles` parameter (defaulting to something like 500) and stop fetching after the limit. Log a warning when truncated.
**How to validate**: Point the import at a large repo (e.g., a docs monorepo with 1000+ .md files) and observe GMS behavior.

---

## Other Issues

- **[MEDIUM / edge-case]** `GitHubDocumentSource.java:391` — `matchesFilters` uses `startsWith(pathPrefix)` without a `/` boundary. Prefix `"doc"` would match `"document/README.md"`. Fix: append `/` to non-empty prefix before the startsWith check, or use `filePath.startsWith(pathPrefix + "/")`.

- **[MEDIUM / edge-case]** `GitHubDocumentSource.java:90-92` — The GitHub tree API returns a `truncated: true` field when the tree is too large (>100K entries). This is not checked. If truncated, the user silently gets an incomplete file list. Fix: check `tree.path("truncated").asBoolean()` and warn the user.

- **[MEDIUM / performance]** `DocumentImportService.java:190` — `entityClient.exists()` is called for every candidate inside a loop. For a batch import of N documents, this makes N individual existence checks. Fix: consider batching existence checks or accepting that the first import will always be "create".

- **[MEDIUM / correctness]** `DocumentImportService.java:275-283` — `makeDocumentId` can produce collisions. Two different source IDs that differ only in characters replaced by `-` (e.g., `"docs/a b"` and `"docs/a-b"`) will produce the same document ID. File-upload source IDs like `upload.my file` and `upload.my-file` would collide.

- **[MEDIUM / dx]** `ImportDocumentsModal.tsx:210-212` — `<Text color="gray" colorLevel={1700}>` passes color directly to alchemy `<Text>` components. Per the project's theming guidance, colors should come from theme tokens rather than being passed as props to alchemy components.

- **[LOW / operability]** `GitHubDocumentSource.java:362-386` — `executeGitHubGet` returns `null` on non-200 responses but doesn't distinguish between 401 (bad token), 403 (rate limit), and 404 (repo not found). A more specific error message would help users troubleshoot.

---

## What's Missing

- **No limit on imported file count**: The GitHub import has no cap on the number of files processed. Both the frontend and backend should enforce a maximum.
- **No cancellation support**: The GitHub import is synchronous on the GraphQL thread. Consider async execution with progress tracking for large imports.
- **Bundle size increase**: Codecov reports +528KB (2.33%) from the `mammoth` DOCX library. Consider lazy-loading it only when the import modal is opened.

## Test Plan

| Invariant | Covered? | Suggested test |
|---|---|---|
| Auth check on all import endpoints | Yes — resolver tests verify `AuthorizationException` | — |
| Idempotent re-import (same URN) | Yes — smoke test `test_import_file_upload_idempotent` | — |
| Parent-child hierarchy | Yes — smoke test `test_import_file_upload_with_parent` | — |
| `makeDocumentId` collision | No | Unit test: verify distinct source IDs don't collide |
| Large repo truncation handling | No | Integration test: mock truncated tree response |
| ES `ignore_above` behavior | Partial — `FieldTypeMapperTest` covers mapping generation | Could add integration test indexing a >8191 char keyword |

## Questions for Author

1. Is the CI build failure (`build-and-test`) unrelated to this PR, or does it need to be fixed before merge?
2. Was a file count limit for GitHub imports intentionally omitted, or is it planned for a follow-up?
3. Should the `mammoth` dependency be lazy-loaded to reduce the bundle impact for users who don't use the import feature?
4. The ES `ignore_above` / keyword mapping changes seem orthogonal to the document import feature — was this bundled intentionally due to document text fields hitting the limit, or could it be split into a separate PR for cleaner rollback?

datahub-web-react/src/app/ingestV2/source/multiStepBuilder/IngestionSourceCreatePage.tsx
```diff
@@ -50,6 +51,7 @@ const STEPS: IngestionSourceFormStep[] = [
 
 export function IngestionSourceCreatePage() {
     const history = useHistory();
+    const location = useLocation<GitHubDocumentsIngestionLocationState>();
```
jjoyce0510: this should not use this type.

datahub-web-react/src/app/ingestV2/source/builder/sources.json
```diff
@@ -579,6 +579,17 @@
         "isPopular": false,
         "isNew": true
     },
+    {
+        "urn": "urn:li:dataPlatform:github",
+        "name": "github-documents",
+        "displayName": "GitHub Documents",
+        "description": "Import markdown and text files from a GitHub repository as Documents.",
+        "docsUrl": "https://docs.datahub.com/docs/generated/ingestion/sources/github-documents/",
+        "recipe": "source:\n  type: github-documents\n  config:\n    github_token: null\n    repository: org/repo\n    branch: main\n    path_prefix: \"\"\n    file_extensions:\n      - .md\n      - .txt\n    parent_document_urn: null\n    document_import_mode: NATIVE\n    show_in_global_context: true",
```
jjoyce0510: We should also have a form for the user to complete here, not just the reference Yaml.

metadata-integration/java/datahub-client/src/main/resources/MetadataChangeProposal.avsc
```diff
@@ -172,6 +172,11 @@
         "type" : [ "null", "string" ],
         "doc" : "Aspect version\n   Initial implementation will use the aspect version's number, however stored as\n   a string in the case where a different aspect versioning scheme is later adopted.",
         "default" : null
+      }, {
```
jjoyce0510, line 175: Can we remove this diff from the PR?

metadata-integration/java/datahub-event/src/main/resources/MetadataChangeProposal.avsc
```diff
@@ -172,6 +172,11 @@
         "type" : [ "null", "string" ],
         "doc" : "Aspect version\n   Initial implementation will use the aspect version's number, however stored as\n   a string in the case where a different aspect versioning scheme is later adopted.",
         "default" : null
+      }, {
```
jjoyce0510, line 175: can we remove this?

metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/index/entity/v2/V2MappingsBuilder.java
```diff
@@ -383,6 +383,7 @@ private Map<String, Object> getMappingsForSearchText(FieldType fieldType) {
     Map<String, Object> mappingForField = new HashMap<>();
     mappingForField.put(TYPE, ESUtils.KEYWORD_FIELD_TYPE);
     mappingForField.put(NORMALIZER, KEYWORD_NORMALIZER);
+    mappingForField.put("ignore_above", ESUtils.KEYWORD_MAXLENGTH);
```
jjoyce0510, line 386: This is required to ensure we can index large documents and avoid hitting errors when trying to index long text fields. This simply truncates above the configured limit for keyword fields.

metadata-io/src/main/java/com/linkedin/metadata/search/utils/ESUtils.java
```diff
@@ -134,8 +134,16 @@ public static LifecycleStageTypeService getLifecycleStageTypeService() {
...
   public static final String TOKEN_COUNT_FIELD_TYPE = "token_count";
+
   // End of field types
 
+  /**
+   * Maximum character length for keyword fields before OpenSearch silently skips indexing the
+   * value. Lucene keyword terms have a hard 32,766-byte limit; using 32766/4 = 8191 ensures safety
+   * even for all-multibyte UTF-8 content. Tokenized text sub-fields are unaffected.
+   */
+  public static final int KEYWORD_MAXLENGTH = 8191;
```
jjoyce0510: What is the default limit we are using now? Can we make this configurable via application.yaml?

metadata-service/war/build.gradle
```diff
@@ -29,6 +29,9 @@ dependencies {
   runtimeOnly project(':metadata-jobs:mae-consumer')
   runtimeOnly project(':metadata-jobs:pe-consumer')
 
+  // Optional PDF text extraction for document import (loaded reflectively)
```
jjoyce0510: Removethis..

metadata-service/services/build.gradle
```diff
@@ -29,6 +29,7 @@ dependencies {
 
     implementation externalDependency.slf4jApi
     implementation externalDependency.swaggerAnnotations
+    implementation externalDependency.jacksonDataBind
```
jjoyce0510: Do we actually need this?

metadata-service/war/gradle.lockfile
jjoyce0510, line 1: remove this lockfile change. why would it be necessary

datahub-web-react/package.json
```diff
@@ -77,6 +77,7 @@
         "i18next-resources-to-backend": "^1",
         "js-cookie": "^2.2.1",
         "js-yaml": "^4.1.0",
+        "mammoth": "^1.12.0",
```
jjoyce0510, line 80: do we really need this? why?

datahub-web-react/src/app/document/hooks/useLoadDocumentTree.ts
```diff
@@ -83,6 +83,7 @@ export function useLoadDocumentTree() {
                         input: {
                             query: '*',
                             parentDocuments: parentUrn ? [parentUrn] : undefined,
+                            rootOnly: parentUrn === null ? true : undefined,
```
jjoyce0510, line 149: is this a bug fix? i forget it's been so long..

datahub-web-react/src/app/context/ContextSidebar.tsx
```diff
@@ -254,6 +264,31 @@ export default function ContextSidebar({
         [entityRegistry, history],
     );
 
+    const handleImportSuccess = useCallback(
+        (parentUrn: string | null) => {
+            const delays = [1500, 3000, 6000];
+
+            if (parentUrn) {
+                // Expand the ancestor chain immediately so the path is visible
+                let current: string | null = parentUrn;
```
jjoyce0510: extract biz logic out of here ideally.

datahub-web-react/src/app/context/ContextSidebar.tsx
```diff
@@ -254,6 +264,31 @@ export default function ContextSidebar({
         [entityRegistry, history],
     );
 
+    const handleImportSuccess = useCallback(
```
jjoyce0510: add comments describing why we need this hook.

datahub-web-react/src/app/context/ContextSidebar.tsx
jjoyce0510: maybe a separate well-tested hook file?

datahub-web-react/src/app/context/import/ImportParentSelector.tsx
```diff
@@ -0,0 +1,169 @@
...
+    flex-shrink: 0;
+`;
+
+interface ImportParentSelectorProps {
+    selectedParentUrn: string | null;
+    onSelectParent: (urn: string | null) => void;
+    getNode?: (urn: string) => DocumentTreeNode | undefined;
+}
+
+export default function ImportParentSelector({
```
jjoyce0510: Is this being leveraged for the GitHub import flow as well? I think it does actually have relevancy, since we can import into specific sub-tree with the GitHub flow.

datahub-web-react/src/app/context/import/ImportDocumentsModal.tsx
```diff
@@ -0,0 +1,327 @@
...
+
+    return (
+        <Modal
+            title="Import Documents"
+            onCancel={handleClose}
+            buttons={buttons}
+            width={560}
+            dataTestId="import-documents-modal"
+        >
+            <DocumentTreeContext.Provider value={documentTreeContext!}>
```
jjoyce0510: This component is too long. Refactor into smaller, well named sub-components and extract biz logic into well-tested util and hook files.

datahub-web-react/src/app/context/import/hooks/useLaunchGitHubDocumentsIngestion.ts
```diff
@@ -0,0 +1,33 @@
...
+type LaunchGitHubDocumentsIngestionParams = {
+    parentDocumentUrn?: string | null;
+};
+
+export function useLaunchGitHubDocumentsIngestion() {
+    const history = useHistory();
+
+    return useCallback(
+        ({ parentDocumentUrn }: LaunchGitHubDocumentsIngestionParams) => {
+            const locationState: GitHubDocumentsIngestionLocationState = {
```
jjoyce0510: this is nice but let's generalize this type so that other areas of the product can also redirect here.

datahub-web-react/src/app/context/import/hooks/useImportFromFiles.ts
```diff
@@ -0,0 +1,114 @@
...
+        return result.value || null;
+    }
+    return null;
+}
+
+function mapUseCase(useCase: ImportUseCase): DocumentImportUseCase {
+    return useCase === ImportUseCase.SKILL ? DocumentImportUseCase.Skill : DocumentImportUseCase.ContextDocument;
+}
+
+export function useImportFromFiles() {
```
jjoyce0510, line 21: make sure these utilities are well unit tested!

datahub-web-react/src/app/ingestV2/source/multiStepBuilder/IngestionSourceCreatePage.tsx
jjoyce0510: use a more general type.

jjoyce0510 (commented):
One general comment. I do like the pattern of allowing users to configure whether docs are ingested as NATIVE (writable) vs EXTERNAL (read only). We should extend the Notion + Confluence ingestion sources to support this exact thing with consistent paramter names as part of this PR, and then include options to import from Notion and Confluence here as well! 

Also, by default, if you Don't select a ROOT document to import into for GitHub, we should create a parent document with the name of the repository!! E.g. repository_name/file/etc.

jjoyce0510:
Actually, we can also introduce 

Import from Notion
Import from Confluence

as part of this PR if we do this, following the exact same convention of redirecting to the ingestion flow! 

Only show this to the user if the user can manage ingestion (has the right platform privilege) since these things will be scheduled jobs running on a schedule. 

This will form a quite consistent, coherent story as part of this PR. 

Also, make sure the GitHub source has a nice Ui form just like the Notion + Confluence sources.

alexsku (dismissed):
## Summary

Nice work pulling this together — the file-upload flow is solid (good Playwright coverage, idempotency via deterministic IDs, proper `MANAGE_DOCUMENTS` gating, the truncation/text-limit behavior is reasonable). The split into a separate `github-documents` ingestion source feels right.

I have two issues I'd like fixed before merge, plus a handful of medium / cleanup items that can land here or in follow-ups. Both blockers are small (single-digit lines each) and both prevent silent failure modes that would be hard for an operator to diagnose.

---

## High-priority (please fix)

### 1. UTF-8 decode error can abort the entire GitHub ingestion run
[`metadata-ingestion/src/datahub/ingestion/source/github_documents/github_api.py:161`](https://github.com/datahub-project/datahub/blob/jj--support-context-document-import-oss/metadata-ingestion/src/datahub/ingestion/source/github_documents/github_api.py#L161)

```python
return base64.b64decode(encoded).decode("utf-8")
```

A single `.md`/`.txt` file that's not UTF-8 (e.g. Latin-1 README, accidentally committed binary) raises `UnicodeDecodeError`. `_emit_file_document` doesn't catch it, so the generator stops mid-iteration and the rest of the repo is never imported. Even with the default extension allowlist, this isn't unusual in long-lived repos.

**Fix**: `decode("utf-8", errors="replace")`, or wrap in try/except and call `self.report.warning(...)` so the operator sees the skip.

**Suggested test**: feed `b"\xff\xfe"` as base64 content, assert `fetch_file_content` returns gracefully and surfaces a warning.

### 2. Truncated GitHub tree silently drops files
[`metadata-ingestion/src/datahub/ingestion/source/github_documents/github_api.py:114-120`](https://github.com/datahub-project/datahub/blob/jj--support-context-document-import-oss/metadata-ingestion/src/datahub/ingestion/source/github_documents/github_api.py#L114-L120)

When `git/trees?recursive=true` returns `"truncated": true` (>100k entries or >7MB), the code emits a Python `logger.warning(...)` and continues with a partial subset. There's no entry in `GitHubDocumentsSourceReport`, no exception. From DataHub's ingestion dashboard the run looks successful — the operator can't tell hundreds of files were dropped.

**Fix**: call `self.report.warning("github-tree-truncated", f"{owner_repo}@{branch} tree was truncated; narrow path_prefix or split sources")` (or equivalent). Optionally add a `tree_truncated: bool` field to `GitHubDocumentsSourceReport`.

**Suggested test**: stub `_get_json` to return `{"tree": [...], "truncated": True}` and assert the report contains a warning.

---

## Medium (would be nice, OK as follow-ups)

- **Dead server-side `TextExtractors` code** — [`TextExtractors.java:30-105`](https://github.com/datahub-project/datahub/blob/jj--support-context-document-import-oss/metadata-service/services/src/main/java/com/linkedin/metadata/service/docimport/TextExtractors.java#L30-L105): the `extract` / `extractHtml` / `extractPlainText` / `isSupported` / `EXTRACTORS` machinery is unreferenced — the resolver only uses `getExtension` and `titleFromFilename`. ~100 lines of unused Java plus its test. Either delete it or surface it through the mutation (so the client can send raw bytes for server-side parsing). Related: `HTML_ENTITIES` regex is declared but never used, and `decodeHtmlEntities` only handles 6 named entities (numeric `&#65;` / hex `&#x41;` slip through).

- **Parent URN not validated** — [`ImportDocumentsFromFilesResolver.java:66-69`](https://github.com/datahub-project/datahub/blob/jj--support-context-document-import-oss/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/knowledge/ImportDocumentsFromFilesResolver.java#L66-L69): `parentDocumentUrn` is parsed and passed straight into `ParentDocument.setDocument(...)` with no entity-type check and no existence check. The PDL `@Relationship` only constrains relationship indexing, not aspect storage — a `MANAGE_DOCUMENTS` holder could pass `urn:li:dataset:...` and create a dangling parent. Asserting `entityType == "document"` + `entityClient.exists(parentUrn)` here would tighten this up.

- **GitHub rate-limit responses indistinguishable from 404** — [`github_api.py:176-188`](https://github.com/datahub-project/datahub/blob/jj--support-context-document-import-oss/metadata-ingestion/src/datahub/ingestion/source/github_documents/github_api.py#L176-L188): all non-200 responses are treated the same (warn + return None). HTTP 403 with `X-RateLimit-Remaining: 0` deserves a clearer report warning, and ideally a `sleep_until(X-RateLimit-Reset)` retry.

- **32-bit hash suffix on document IDs** — [`DocumentImportService.java:194-207`](https://github.com/datahub-project/datahub/blob/jj--support-context-document-import-oss/metadata-service/services/src/main/java/com/linkedin/metadata/service/docimport/DocumentImportService.java#L194-L207) and Python counterpart. Collision requires both the sanitized-base AND the 4-byte hash to match for two distinct source IDs — small chance in practice, but when it happens the new doc silently overwrites the old. Bumping `ID_HASH_SUFFIX_LENGTH` to 12-16 is cheap and forecloses the failure mode.

- **No shared-corpus test for Java `makeDocumentId` ↔ Python `normalize_document_id` parity.** The two are documented to agree (the Python helper's docstring says so explicitly), but nothing in CI enforces it. Worth a small fixture file checked into both sides.

---

## Smaller things

- [`github_documents_source.py:3`](https://github.com/datahub-project/datahub/blob/jj--support-context-document-import-oss/metadata-ingestion/src/datahub/ingestion/source/github_documents/github_documents_source.py#L3) and [`github_api.py:3`](https://github.com/datahub-project/datahub/blob/jj--support-context-document-import-oss/metadata-ingestion/src/datahub/ingestion/source/github_documents/github_api.py#L3): `from __future__ import annotations` isn't needed on 3.11+.
- [`DocumentImportServiceTest.java:298`](https://github.com/datahub-project/datahub/blob/jj--support-context-document-import-oss/metadata-service/services/src/test/java/com/linkedin/metadata/service/docimport/DocumentImportServiceTest.java#L298) test comment says "Source type should always be NATIVE" — accurate for the file-upload service, but worth a one-liner noting the Python source can emit EXTERNAL via the SDK.
- The PR is over 4.7k lines / 86 files and bundles search-mapping hardening with the document-import feature. Not a blocker, but splitting the ES `ignore_above` / V2MappingsBuilder constructor changes into their own PR would make both easier to review and revert independently.

---

## Questions

1. The server-side `TextExtractors` extraction methods are written but never called — placeholder for a future server-side path, or leftover from an earlier design? Determines whether the right move is "delete" vs "wire up".
2. For the truncated-tree case, do you want it to be a fail-the-run or warn-and-continue? I'd default to warn-and-continue (with a visible report warning) since splitting by `path_prefix` is the documented escape hatch, but happy to defer.
3. `DocumentSourceType` is hardcoded to `NATIVE` in the import service. Worth supporting an EXTERNAL mode in the file-upload mutation too (e.g. for "imported from <X>" provenance), or is that out of scope for OSS v1?
4. `lastIngested` was added to the `getDocument` query — is it set automatically by the entity service for imported docs, or is it intentionally unset for native uploads?

Once the two HIGH items are addressed I'm happy to re-review and approve. Thanks again for the work here — the feature itself is solid and the testing breadth is good.

jjoyce0510:
Thanks - will try to close the gaps here. Nice feedbak

shirshanka (approved):
Approving - please check CI!

PR #16906, opened 2026-04-06
feat(patch): Add support for * in path with arrayPrimaryKeys; add documentation patch

asikowitz:
Makes the following changes:
- Updates tags, terms, ownership, and structured property templates to include attribution in key fields
- Updates Python patch client to support attribution + bulk deletion
- Fixes ownership patching so that we support multiple entries with the same owner and type enum, but different type urn
- Fixes a few bugs in core patching code
- Add documentation patch template and corresponding Python / Java client code. Adds documentation aspect to most asset types -- already the case in DataHub Cloud
- Add significant testing
- Some cleaning up of unused code

### DIFF
--- a/entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/template/TemplateUtil.java
+++ b/entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/template/TemplateUtil.java
@@ -73,8 +73,8 @@ public class TemplateUtil {
     JsonNode transformedNodeClone = transformedNode.deepCopy();
     List<Pair<PatchOperationType, String>> paths = getPaths(jsonPatch);
     for (Pair<PatchOperationType, String> operationPath : paths) {
-      // Negative limit preserves trailing empty strings for paths ending with "/" (unattributed
-      // ops)
+      // Negative limit preserves trailing empty strings for paths ending with "/"
+      // e.g. unattributed operations
       String[] keys = operationPath.getSecond().split("/", -1);
       JsonNode parent = transformedNodeClone;
 
--- a/entity-registry/src/test/java/com/linkedin/metadata/aspect/patch/builder/GlobalTagsPatchBuilderTest.java
+++ b/entity-registry/src/test/java/com/linkedin/metadata/aspect/patch/builder/GlobalTagsPatchBuilderTest.java
@@ -1,17 +1,14 @@
 package com.linkedin.metadata.aspect.patch.builder;
 
 import static org.testng.Assert.assertEquals;
-import static org.testng.Assert.assertFalse;
 import static org.testng.Assert.assertNotNull;
 import static org.testng.Assert.assertNull;
 import static org.testng.Assert.assertThrows;
 import static org.testng.Assert.assertTrue;
 
 import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.linkedin.common.urn.TagUrn;
 import com.linkedin.common.urn.Urn;
-import com.linkedin.mxe.MetadataChangeProposal;
 import java.net.URISyntaxException;
 import java.util.List;
 import org.apache.commons.lang3.tuple.ImmutableTriple;
@@ -74,10 +71,7 @@ public class GlobalTagsPatchBuilderTest {
 
     ImmutableTriple<String, String, JsonNode> operation = pathValues.get(0);
     assertEquals(operation.getLeft(), "add");
-    // Tag first, trailing slash for empty attribution source: /tags/<tagUrn>/
-    assertTrue(
-        operation.getMiddle().startsWith("/tags/") && !operation.getMiddle().startsWith("/tags//"));
-    assertTrue(operation.getMiddle().endsWith("/"));
+    assertTrue(operation.getMiddle().startsWith("/tags/"));
     assertTrue(operation.getRight().isObject());
     assertEquals(operation.getRight().get("tag").asText(), tagUrn.toString());
     assertEquals(operation.getRight().get("context").asText(), context);
@@ -95,10 +89,7 @@ public class GlobalTagsPatchBuilderTest {
 
     ImmutableTriple<String, String, JsonNode> operation = pathValues.get(0);
     assertEquals(operation.getLeft(), "add");
-    // Tag first, trailing slash for empty attribution source: /tags/<tagUrn>/
-    assertTrue(
-        operation.getMiddle().startsWith("/tags/") && !operation.getMiddle().startsWith("/tags//"));
-    assertTrue(operation.getMiddle().endsWith("/"));
+    assertTrue(operation.getMiddle().startsWith("/tags/"));
     assertTrue(operation.getRight().isObject());
     assertEquals(operation.getRight().get("tag").asText(), tagUrn.toString());
     assertNull(operation.getRight().get("context"));
@@ -116,9 +107,7 @@ public class GlobalTagsPatchBuilderTest {
 
     ImmutableTriple<String, String, JsonNode> operation = pathValues.get(0);
     assertEquals(operation.getLeft(), "remove");
-    // Remove at tag level: /tags/<tagUrn> (no wildcards)
     assertTrue(operation.getMiddle().startsWith("/tags/"));
-    assertFalse(operation.getMiddle().contains("*"));
     assertNull(operation.getRight());
   }
 
@@ -148,47 +137,5 @@ public class GlobalTagsPatchBuilderTest {
     } catch (URISyntaxException e) {
       throw new RuntimeException(e);
     }
-  }
-
-  @Test
-  public void testBuildPlainAdd() throws Exception {
-    TagUrn tagUrn = TagUrn.createFromString(TEST_TAG_URN);
-    MetadataChangeProposal mcp = builder.addTag(tagUrn, null).build();
-
-    assertEquals(mcp.getAspect().getContentType(), "application/json-patch+json");
-
-    byte[] bytes = mcp.getAspect().getValue().copyBytes();
-    JsonNode payload = new ObjectMapper().readTree(bytes);
-
-    // Plain patch: a JSON array, no arrayPrimaryKeys envelope
-    assertTrue(payload.isArray(), "add must produce a plain JSON array (no envelope)");
-    assertEquals(payload.size(), 1);
-    JsonNode op = payload.get(0);
-    assertEquals(op.get("op").asText(), "add");
-    // Tag first, trailing slash for empty attribution source: /tags/<tagUrn>/
-    assertTrue(
-        op.get("path").asText().startsWith("/tags/")
-            && !op.get("path").asText().startsWith("/tags//"));
-    assertTrue(op.get("path").asText().endsWith("/"));
-  }
-
-  @Test
-  public void testBuildPlainRemove() throws Exception {
-    TagUrn tagUrn = TagUrn.createFromString(TEST_TAG_URN);
-    MetadataChangeProposal mcp = builder.removeTag(tagUrn).build();
-
-    assertEquals(mcp.getAspect().getContentType(), "application/json-patch+json");
-
-    byte[] bytes = mcp.getAspect().getValue().copyBytes();
-    JsonNode payload = new ObjectMapper().readTree(bytes);
-
-    // Plain patch: a JSON array, no arrayPrimaryKeys envelope
-    assertTrue(payload.isArray(), "remove must produce a plain JSON array (no envelope)");
-    assertEquals(payload.size(), 1);
-    JsonNode op = payload.get(0);
-    assertEquals(op.get("op").asText(), "remove");
-    // Remove at tag level: /tags/<tagUrn> (no wildcards)
-    assertTrue(op.get("path").asText().startsWith("/tags/"));
-    assertFalse(op.get("path").asText().contains("*"));
   }
 }
--- a/entity-registry/src/test/java/com/linkedin/metadata/aspect/patch/builder/DocumentationPatchBuilderTest.java
+++ b/entity-registry/src/test/java/com/linkedin/metadata/aspect/patch/builder/DocumentationPatchBuilderTest.java
@@ -6,9 +6,7 @@ import static org.testng.Assert.assertNull;
 import static org.testng.Assert.assertTrue;
 
 import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.linkedin.common.urn.Urn;
-import com.linkedin.mxe.MetadataChangeProposal;
 import java.net.URISyntaxException;
 import java.util.List;
 import org.apache.commons.lang3.tuple.ImmutableTriple;
@@ -46,11 +44,10 @@ public class DocumentationPatchBuilderTest {
 
     ImmutableTriple<String, String, JsonNode> operation = pathValues.get(0);
     assertEquals(operation.getLeft(), "add");
-    // Unattributed: BASE_PATH + "" = "/documentations/" (trailing slash only)
+    // Trailing slash indicates unattributed documentation
     assertEquals(operation.getMiddle(), "/documentations/");
     assertTrue(operation.getRight().isObject());
     assertEquals(operation.getRight().get("documentation").asText(), "Some documentation text");
-    // No attribution node for unattributed docs
     assertNull(operation.getRight().get("attribution"));
   }
 
@@ -67,8 +64,6 @@ public class DocumentationPatchBuilderTest {
     ImmutableTriple<String, String, JsonNode> operation = pathValues.get(0);
     assertEquals(operation.getLeft(), "remove");
     assertTrue(operation.getMiddle().startsWith("/documentations/"));
-    // Path should contain the encoded source (not wildcard)
-    assertTrue(!operation.getMiddle().endsWith("/*"));
     assertTrue(operation.getMiddle().length() > "/documentations/".length());
     assertNull(operation.getRight());
   }
@@ -84,25 +79,8 @@ public class DocumentationPatchBuilderTest {
 
     ImmutableTriple<String, String, JsonNode> operation = pathValues.get(0);
     assertEquals(operation.getLeft(), "remove");
-    // Remove all: plain path at field level (no wildcard)
+    // Remove all: no trailing slash
     assertEquals(operation.getMiddle(), "/documentations");
     assertNull(operation.getRight());
-  }
-
-  @Test
-  public void testBuildPlainAdd() throws Exception {
-    MetadataChangeProposal mcp = builder.addDocumentation("text").build();
-
-    assertEquals(mcp.getAspect().getContentType(), "application/json-patch+json");
-
-    byte[] bytes = mcp.getAspect().getValue().copyBytes();
-    JsonNode payload = new ObjectMapper().readTree(bytes);
-
-    // Plain patch: a JSON array, no arrayPrimaryKeys envelope
-    assertTrue(payload.isArray(), "add must produce a plain JSON array (no envelope)");
-    assertEquals(payload.size(), 1);
-    JsonNode op = payload.get(0);
-    assertEquals(op.get("op").asText(), "add");
-    assertEquals(op.get("path").asText(), "/documentations/");
   }
 }
--- a/entity-registry/src/test/java/com/linkedin/metadata/aspect/patch/builder/GlossaryTermsPatchBuilderTest.java
+++ b/entity-registry/src/test/java/com/linkedin/metadata/aspect/patch/builder/GlossaryTermsPatchBuilderTest.java
@@ -1,17 +1,14 @@
 package com.linkedin.metadata.aspect.patch.builder;
 
 import static org.testng.Assert.assertEquals;
-import static org.testng.Assert.assertFalse;
 import static org.testng.Assert.assertNotNull;
 import static org.testng.Assert.assertNull;
 import static org.testng.Assert.assertThrows;
 import static org.testng.Assert.assertTrue;
 
 import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.linkedin.common.urn.GlossaryTermUrn;
 import com.linkedin.common.urn.Urn;
-import com.linkedin.mxe.MetadataChangeProposal;
 import java.net.URISyntaxException;
 import java.util.List;
 import org.apache.commons.lang3.tuple.ImmutableTriple;
@@ -74,11 +71,7 @@ public class GlossaryTermsPatchBuilderTest {
 
     ImmutableTriple<String, String, JsonNode> operation = pathValues.get(0);
     assertEquals(operation.getLeft(), "add");
-    // Term first, trailing slash for empty attribution source: /terms/<termUrn>/
-    assertTrue(
-        operation.getMiddle().startsWith("/terms/")
-            && !operation.getMiddle().startsWith("/terms//"));
-    assertTrue(operation.getMiddle().endsWith("/"));
+    assertTrue(operation.getMiddle().startsWith("/terms/"));
     assertTrue(operation.getRight().isObject());
     assertEquals(operation.getRight().get("urn").asText(), termUrn.toString());
     assertEquals(operation.getRight().get("context").asText(), context);
@@ -96,11 +89,7 @@ public class GlossaryTermsPatchBuilderTest {
 
     ImmutableTriple<String, String, JsonNode> operation = pathValues.get(0);
     assertEquals(operation.getLeft(), "add");
-    // Term first, trailing slash for empty attribution source: /terms/<termUrn>/
-    assertTrue(
-        operation.getMiddle().startsWith("/terms/")
-            && !operation.getMiddle().startsWith("/terms//"));
-    assertTrue(operation.getMiddle().endsWith("/"));
+    assertTrue(operation.getMiddle().startsWith("/terms/"));
     assertTrue(operation.getRight().isObject());
     assertEquals(operation.getRight().get("urn").asText(), termUrn.toString());
     assertNull(operation.getRight().get("context"));
@@ -118,9 +107,7 @@ public class GlossaryTermsPatchBuilderTest {
 
     ImmutableTriple<String, String, JsonNode> operation = pathValues.get(0);
     assertEquals(operation.getLeft(), "remove");
-    // Remove at term level: /terms/<termUrn> (no wildcards)
     assertTrue(operation.getMiddle().startsWith("/terms/"));
-    assertFalse(operation.getMiddle().contains("*"));
     assertNull(operation.getRight());
   }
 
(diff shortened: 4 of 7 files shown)


entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/template/AspectTemplateEngine.java
```diff
@@ -19,33 +16,6 @@
  * templates
  */
 public class AspectTemplateEngine {
-
-  public static final Set<String> SUPPORTED_TEMPLATES =
```
asikowitz, line 23: Unused and confusing imo

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/template/CompoundKeyTemplate.java
```diff
@@ -1,29 +1,6 @@
...
-import jakarta.json.Json;
-import jakarta.json.JsonObject;
-import jakarta.json.JsonPatch;
-import java.io.StringReader;
 
 public abstract class CompoundKeyTemplate<T extends RecordTemplate>
-    implements ArrayMergingTemplate<T> {
-
-  @Override
-  public T applyPatch(RecordTemplate recordTemplate, JsonPatch jsonPatch)
```
asikowitz, line 19: This override was identical to the default in `Template.java` except it (i) didn't call `TemplateUtil.validatePatch(jsonPatch);` and (ii) didn't catch the exception. I'm guessing it's not purposeful to have these similar but slightly different implementations, because nothing about `CompoundKeyTemplate` necessitates them.

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/template/Template.java
```diff
@@ -40,19 +40,8 @@ default T getSubtype(RecordTemplate recordTemplate) throws ClassCastException {
...
-   * @param jsonPatch patch to apply
-   * @return patched value
-   * @throws JsonProcessingException if there is an issue converting the input to JSON
-   */
-  default T applyPatch(RecordTemplate recordTemplate, JsonPatch jsonPatch)
-      throws JsonProcessingException {
-    TemplateUtil.validatePatch(jsonPatch);
-
-    JsonNode transformed = populateTopLevelKeys(preprocessTemplate(recordTemplate), jsonPatch);
+  default T applyPatchToJson(JsonNode templateAsJson, JsonPatch jsonPatch) {
```
asikowitz: Helper gets called by GenericPatchTemplate

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/builder/AbstractArrayPrimaryKeysPatchBuilder.java
```diff
@@ -0,0 +1,71 @@
...
+   * GenericPatchTemplate}, which supports {@code *} wildcard paths.
+   */
+  @Override
+  protected GenericAspect buildPatch() {
+    if (pathValues.isEmpty()) {
+      throw new IllegalArgumentException("No patches specified.");
+    }
+
+    ArrayNode patchOps = instance.arrayNode();
+    for (ImmutableTriple<String, String, com.fasterxml.jackson.databind.JsonNode> triple :
```
RyanHolstien: nit: No full classpaths, add this to your base prompt 😄

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/builder/GlobalTagsPatchBuilder.java
```diff
@@ -31,19 +43,35 @@ public GlobalTagsPatchBuilder addTag(@Nonnull TagUrn urn, @Nullable String conte
       value.put(CONTEXT_KEY, context);
     }
 
-    pathValues.add(
-        ImmutableTriple.of(
-            PatchOperationType.ADD.getValue(), BASE_PATH + encodeValueUrn(urn), value));
+    // Empty source segment produces a double-slash: /tags//<tagUrn>
+    pathValues.add(ImmutableTriple.of("add", BASE_PATH + "/" + encodeValueUrn(urn), value));
```
RyanHolstien: Huh? Why is this adding a slash, it already includes the slash in the BASE_PATH static var. It isn't changing the BASE_PATH so this will always have a double slash...

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/builder/GlobalTagsPatchBuilder.java
```diff
@@ -5,23 +5,35 @@
...
 import com.linkedin.common.urn.TagUrn;
-import com.linkedin.metadata.aspect.patch.PatchOperationType;
+import java.util.List;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import org.apache.commons.lang3.tuple.ImmutableTriple;
 
-public class GlobalTagsPatchBuilder extends AbstractMultiFieldPatchBuilder<GlobalTagsPatchBuilder> {
+public class GlobalTagsPatchBuilder
+    extends AbstractArrayPrimaryKeysPatchBuilder<GlobalTagsPatchBuilder> {
```
RyanHolstien: Why are we converting these to using the generic typing with the arrayPrimaryKeys specified?

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/builder/AbstractArrayPrimaryKeysPatchBuilder.java
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Base class for patch builders that emit a {@code GenericJsonPatch} payload ({@code
+ * {"arrayPrimaryKeys": {...}, "patch": [...]}}) enabling compound-key semantics and {@code *}
+ * wildcard removals on the backend.
+ *
+ * <p>Subclasses declare which array field and primary-key list to use via {@link
+ * #getArrayFieldName()} and {@link #getArrayPrimaryKeys()}.
+ */
+public abstract class AbstractArrayPrimaryKeysPatchBuilder<
```
RyanHolstien: I'm confused why we need this class, Generic Patches are already handled differently, we don't want to use this as an interface for other path builders

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/builder/GlossaryTermsPatchBuilder.java
```diff
@@ -32,19 +43,35 @@ public GlossaryTermsPatchBuilder addTerm(@Nonnull GlossaryTermUrn urn, @Nullable
...
+   * regardless of which source attributed it.
+   *
+   * @param urn required
+   * @return this builder
+   */
   public GlossaryTermsPatchBuilder removeTerm(@Nonnull GlossaryTermUrn urn) {
-    pathValues.add(
-        ImmutableTriple.of(
-            PatchOperationType.REMOVE.getValue(), BASE_PATH + encodeValueUrn(urn), null));
+    pathValues.add(ImmutableTriple.of("remove", BASE_PATH + "*/" + encodeValueUrn(urn), null));
```
RyanHolstien: Injecting * syntax into the base removal should not be done, we should be specifying the proper pathing to remove the term.

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/builder/GlossaryTermsPatchBuilder.java
```diff
@@ -32,19 +43,35 @@ public GlossaryTermsPatchBuilder addTerm(@Nonnull GlossaryTermUrn urn, @Nullable
       value.put(CONTEXT_KEY, context);
     }
 
-    pathValues.add(
-        ImmutableTriple.of(
-            PatchOperationType.ADD.getValue(), BASE_PATH + encodeValueUrn(urn), value));
+    // Empty source segment produces a double-slash: /terms//<termUrn>
+    pathValues.add(ImmutableTriple.of("add", BASE_PATH + "/" + encodeValueUrn(urn), value));
```
RyanHolstien: Same comment, this doesn't make sense.

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/builder/OwnershipPatchBuilder.java
```diff
@@ -6,62 +6,149 @@
...
+  private static final String TYPE_URN_KEY = "typeUrn";
 
-  public OwnershipPatchBuilder addOwner(@Nonnull Urn owner, @Nonnull OwnershipType type) {
+  /**
+   * Compound key that mirrors the Python SDK: (attribution.source, owner, type, typeUrn). Using all
+   * four lets multiple entries for the same owner with the same type enum but different typeUrns
+   * coexist, and allows {@code *} wildcards for flexible removal.
+   */
+  private static final List<String> ARRAY_PRIMARY_KEYS =
+      List.of("attribution" + UNIT_SEPARATOR + "source", OWNER_KEY, TYPE_KEY, TYPE_URN_KEY);
```
RyanHolstien: We don't need the array primary keys logic injected in here, it's already baked in to how the patch builder is constructed. That is the path where it's doing the addOwner at.

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/template/Template.java
RyanHolstien: I don't think this is a valid change, it basically is creating a helper pathway so it doesn't have to validate the patch

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/template/TemplateUtil.java
```diff
@@ -73,7 +78,9 @@ public static JsonNode populateTopLevelKeys(JsonNode transformedNode, JsonPatch
     JsonNode transformedNodeClone = transformedNode.deepCopy();
     List<Pair<PatchOperationType, String>> paths = getPaths(jsonPatch);
     for (Pair<PatchOperationType, String> operationPath : paths) {
-      String[] keys = operationPath.getSecond().split("/");
+      // Need negative limit in split to preserve appropriate length of path,
+      // when trailing path entries are empty string
+      String[] keys = operationPath.getSecond().split("/", -1);
```
RyanHolstien, line 78: This makes me very suspicious 🤔 what is the purpose of this change? Why do we need the empty string pulled now when we didn't previously?

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/template/CompoundKeyTemplate.java
RyanHolstien: Yeah it's an outdated distinction. Originally the Template class didn't do populateTopLevelKeys, but it got moved into both so now it's an unnecessary abstraction layer. We could just move to using the ArrayMergingTemplate where this class is being used.

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/template/AspectTemplateEngine.java
RyanHolstien: Yep, this is removable now. It gets pulled from the registry instead. Originally a way to pull from the template engine itself to specify what was supported, but now it's outdated.

entity-registry/src/test/java/com/linkedin/metadata/aspect/patch/builder/GlobalTagsPatchBuilderTest.java
```diff
@@ -89,7 +92,8 @@ public void testAddTagWithoutContext() throws URISyntaxException {
 
     ImmutableTriple<String, String, JsonNode> operation = pathValues.get(0);
     assertEquals(operation.getLeft(), "add");
-    assertTrue(operation.getMiddle().startsWith("/tags/"));
+    // New format: double-slash for empty source segment
+    assertTrue(operation.getMiddle().startsWith("/tags//"));
```
RyanHolstien: Yeah okay this confirms the suspicions above, this is doing weird stuff

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/builder/AbstractArrayPrimaryKeysPatchBuilder.java
```diff
@@ -0,0 +1,71 @@
...
+        T extends AbstractArrayPrimaryKeysPatchBuilder<T>>
+    extends AbstractMultiFieldPatchBuilder<T> {
+
+  /** Unit separator (U+241F) used in arrayPrimaryKeys for nested field paths. */
+  public static final String UNIT_SEPARATOR = "\u241f";
```
RyanHolstien: Let's keep the usage of the UNIT_SEPARATOR constant to be consistent from a single class as much as possible. I'd rather it be easily be traceable where it's getting used and not have to change it in a bunch of places if we change what we use.

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/template/TemplateUtil.java
```diff
@@ -94,6 +101,141 @@ public static JsonNode populateTopLevelKeys(JsonNode transformedNode, JsonPatch
...
+   * attribution. Previously, if two entries had the same key, at patch for a different tag would
+   * still merge both entries together and cause one to be lost.
+   *
+   * <p>Supports nested key paths using the {@link ArrayMergingTemplate#UNIT_SEPARATOR_DELIMITER}
+   * separator (e.g. {@code "attribution␟source"} navigates to {@code entry.attribution.source}).
+   *
+   * @param keyField the field (or nested path) used as the map key (e.g. {@code "tag"}, {@code
+   *     "urn"}, or {@code "attribution␟source"})
+   */
+  public static JsonNode arrayFieldToListMap(
```
RyanHolstien: Why are we implementing these as static util functions instead of keeping them together with the other functions? Also this only supports a single keyField instead of multiple?

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/template/TemplateUtil.java
asikowitz: Previously we didn't have optional fields in the path. Once you have that, you can pass empty string in the path and then this breaks.

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/template/CompoundKeyTemplate.java
asikowitz: Done

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/template/Template.java
asikowitz: Reverted

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/builder/GlobalTagsPatchBuilder.java
asikowitz: Reverted most of the changes to the patch builders as well -- now they just support adding / removing with attribution

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/builder/AbstractArrayPrimaryKeysPatchBuilder.java
asikowitz: Removed

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/template/TemplateUtil.java
RyanHolstien: We did, I had used "NONE" to signify it was missing for FineGrainedLineage where some keys were optional fields.

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/builder/AbstractMultiFieldPatchBuilder.java
```diff
@@ -95,12 +101,32 @@ protected GenericAspect buildPatch() {
                     .set(VALUE_KEY, triple.right)));
 
     GenericAspect genericAspect = new GenericAspect();
-    genericAspect.setContentType("application/json");
+    genericAspect.setContentType("application/json-patch+json");
```
asikowitz, line 104: This just seems wrong

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/builder/GlobalTagsPatchBuilder.java
```diff
@@ -33,7 +34,7 @@ public GlobalTagsPatchBuilder addTag(@Nonnull TagUrn urn, @Nullable String conte
 
     pathValues.add(
         ImmutableTriple.of(
-            PatchOperationType.ADD.getValue(), BASE_PATH + encodeValueUrn(urn), value));
+            PatchOperationType.ADD.getValue(), BASE_PATH + encodeValueUrn(urn) + "/", value));
```
asikowitz, line 37: Last field is now attribution. This PR is already massive so I didn't add support for adding metadata with attribution in the Java client, but that is available in the Python client

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/template/ArrayMergingTemplate.java
```diff
@@ -103,6 +103,11 @@ default JsonNode transformedMapToArray(
       return transformedNode;
     }
     ObjectNode rebasedNode = transformedNode.deepCopy();
+    // fieldNode is null when the entire array field was removed by the patch.
+    // Return an empty array rather than NPE.
+    if (fieldNode == null || fieldNode.isNull()) {
+      return rebasedNode.set(arrayFieldName, instance.arrayNode());
+    }
```
asikowitz, line 110: We previously didn't support clearing the entire top-level array. Now we do with the "remove all documentations" patch

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/template/ArrayMergingTemplate.java
```diff
@@ -118,6 +123,11 @@ default JsonNode transformedMapToArray(
 
   default ArrayNode mergeToArray(JsonNode mapNode, List<String> keyFields) {
     if (keyFields.isEmpty()) {
+      // When a plain add sets an array at an entity-key level (e.g. /tags/<urn> with
+      // value [{...}]), the leaf ArrayNode's elements must be expanded into the result.
+      if (mapNode instanceof ArrayNode) {
+        return instance.arrayNode().addAll((ArrayNode) mapNode);
+      }
```
asikowitz, line 130: This fixes a pre-existing bug. Without it, the following test fails on master:

```
package com.linkedin.metadata.aspect.patch.template;

import com.linkedin.common.OwnerArray;
import com.linkedin.common.Ownership;
import com.linkedin.metadata.aspect.patch.template.common.OwnershipTemplate;
import jakarta.json.Json;
import jakarta.json.JsonPatch;
import java.util.List;
import java.util.stream.Collectors;
import org.testng.Assert;
import org.testng.annotations.Test;

public class OwnershipTemplateTest {
  OwnershipTemplate TEMPLATE = new OwnershipTemplate();

  @Test
  public void testAddTwoDistinctEntries() throws Exception {
    Ownership initial = new Ownership();
    initial.setOwners(new OwnerArray());

    JsonPatch patchA =
        Json.createPatch(
            Json.createArrayBuilder()
                .add(
                    Json.createObjectBuilder()
                        .add("op", "add")
                        .add("path", "/owners/urn:li:corpuser:userA/DATAOWNER")
                        .add(
                            "value",
                            Json.createArrayBuilder()
                                .add(
                                    Json.createObjectBuilder()
                                        .add("owner", "urn:li:corpuser:userA")
                                        .add("type", "DATAOWNER"))))
                .build());

    JsonPatch patchB =
        Json.createPatch(
            Json.createArrayBuilder()
                .add(
                    Json.createObjectBuilder()
                        .add("op", "add")
                        .add("path", "/owners/urn:li:corpuser:userB/DATAOWNER")
                        .add(
                            "value",
                            Json.createArrayBuilder()
                                .add(
                                    Json.createObjectBuilder()
                                        .add("owner", "urn:li:corpuser:userB")
                                        .add("type", "DATAOWNER"))))
                .build());

    Ownership afterA = TEMPLATE.applyPatch(initial, patchA);
    Ownership result = TEMPLATE.applyPatch(afterA, patchB);

    Assert.assertNotNull(result.getOwners());
    Assert.assertEquals(result.getOwners().size(), 2);
    List<String> ownerUrns =
        result.getOwners().stream().map(o -> o.getOwner().toString()).collect(Collectors.toList());
    Assert.assertTrue(ownerUrns.contains("urn:li:corpuser:userA"), "userA should be present");
    Assert.assertTrue(ownerUrns.contains("urn:li:corpuser:userB"), "userB should be present");
  }
}
```

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/template/TemplateUtil.java
asikowitz: What is the benefit of that over empty string? It is not intuitive to me to pass a dummy value (also, what happens if there is no such dummy value?) and I don't think that's documented anywhere

metadata-integration/java/datahub-client/src/main/java/datahub/client/v2/entity/Entity.java
```diff
@@ -58,10 +59,7 @@ public abstract class Entity {
   @Nonnull protected final List<MetadataChangeProposalWrapper> pendingMCPs;
 
   /** Accumulated patch builders by aspect name (for combining operations). */
-  @Nonnull
-  protected final Map<
-          String, com.linkedin.metadata.aspect.patch.builder.AbstractMultiFieldPatchBuilder<?>>
-      patchBuilders;
+  @Nonnull protected final Map<String, AbstractMultiFieldPatchBuilder<?>> patchBuilders;
```
asikowitz, line 62: Just cleaning up -- as you said, no full classpaths

metadata-io/metadata-io-api/src/main/java/com/linkedin/metadata/entity/ebean/batch/PatchItemImpl.java
```diff
@@ -127,12 +127,20 @@ public ChangeItemImpl applyPatch(RecordTemplate recordTemplate, AspectRetriever
   private ChangeItemImpl applyGenericPatch(
       RecordTemplate recordTemplate, AspectRetriever aspectRetriever) {
     try {
+      // Prefer the registered template's default (e.g. GlossaryTermsTemplate.getDefault() which
+      // injects the required auditStamp) over a bare constructor call which yields an incomplete
+      // object that fails validation.
+      AspectTemplateEngine templateEngine =
+          aspectRetriever.getEntityRegistry().getAspectTemplateEngine();
+      RecordTemplate templateDefault = templateEngine.getDefaultTemplate(getAspectName());
```
asikowitz, line 135: Not using `getDefaultTemplate` here was a bug. You can see we do that for `applyTemplatePatch` -- no reason to not do it for generic patch as well

metadata-io/src/test/java/com/linkedin/metadata/entity/EntityServiceTest.java
```diff
@@ -2713,12 +2714,7 @@ public void testBatchPatchWithTrailingNoOp() throws Exception {
                 _testEntityRegistry
                     .getEntitySpec(DATASET_ENTITY_NAME)
                     .getAspectSpec(GLOBAL_TAGS_ASPECT_NAME))
-            .patch(
-                GenericJsonPatch.builder()
-                    .arrayPrimaryKeys(Map.of("tags", List.of("tag")))
-                    .patch(List.of(tagPatchOp(PatchOperationType.ADD, tag2)))
-                    .build()
-                    .getJsonPatch())
+            .patch(new GlobalTagsPatchBuilder().addTag(tag2, "test-context").getJsonPatch())
```
asikowitz, line 2717: The old generic patches don't make sense to me... we're setting `arrayPrimaryKeys` but that doesn't get read because we're building with `.patch()` and not `.genericJsonPatch()`. In any case, I think it's more accurate to use the patch builder's generated patch, so I added the method `getJsonPatch` to `AbstractMultiFieldPatchBuilder` to facilitate this testing

metadata-ingestion/src/datahub/specific/aspect_helpers/structured_properties.py
```diff
@@ -25,75 +35,93 @@ def set_structured_property(
...
     def add_structured_property(
-        self, key: str, value: Union[str, float, List[Union[str, float]]]
+        self,
+        key: str,
+        value: Union[str, float, List[Union[str, float]]],
+        attribution_source: Optional[Union[str, Urn]] = None,
     ) -> Self:
         """Add a structured property.
 
+        Currently equivalent to set_structured_property: overwrites all values for the given property.
```
asikowitz, line 90: Note: old description was inaccurate. We never supported appending a value to an existing structured property. I think this could be done without too much effort, but this PR is already very large

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/builder/AbstractMultiFieldPatchBuilder.java
RyanHolstien: Yeah we didn't previously use the patch+json for anything logically and this just never got updated I think

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/template/TemplateUtil.java
RyanHolstien: Because it's easier to key on and doesn't require this change which will also pull empty string from trailing slashes mainly. Effectively they're the same. It's like the difference between using a unit separator vs using some other token. It's probably not well documented, but the logic for sure exists in current code for fineGrainedLineage. I'm not strongly opposed to this assuming it doesn't do weird things when pulling the empty strings and can properly key them in the map.

For Strings there is a (usually trivial) semantic difference between empty string and null so having an intentional "this does not exist" behavior for optional key values rather than allowing empty strings feels better to me, but like I said I won't die on that hill 😄

entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/builder/AbstractMultiFieldPatchBuilder.java
AdrianMachado: It's actually correct: https://datatracker.ietf.org/doc/html/rfc6902#section-3

asikowitz: I meant the old code was wrong... unless you're saying I should revert?

PR #16910, opened 2026-04-06
feat(stateful-ingestion): add per-entity-type deletion summary to ingestion report

aviraj-gour:
## Summary

- Adds entity_type_deletion_summary to StaleEntityRemovalSourceReport — a per-entity-type breakdown showing checkpoint counts and actual soft-deletion counts
- Introduces EntityTypeDeletionSummary dataclass with previous_checkpoint_count, current_checkpoint_count, and soft_deleted_count
- soft_deleted_count only tracks URNs that were actually soft-deleted (passed all checks), not the raw checkpoint diff — non-deletable entities are excluded
- Summary is always computed when both checkpoints exist, including when deletions are skipped.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/state/stale_entity_removal_handler.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/state/stale_entity_removal_handler.py
@@ -357,6 +357,9 @@ stateful_ingestion:\
                 self.add_entity_to_state("", urn)
             return
 
+        report = self.source.get_report()
+        assert isinstance(report, StaleEntityRemovalSourceReport)
+
         # Everything looks good, emit the soft-deletion workunits
         for urn in last_checkpoint_state.get_urns_not_in(
             type="*", other_checkpoint_state=cur_checkpoint_state
--- a/metadata-ingestion/src/datahub/ingestion/source/state/stale_entity_removal_handler.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/state/stale_entity_removal_handler.py
@@ -1,4 +1,5 @@
 import logging
+from collections import Counter
 from dataclasses import dataclass, field
 from functools import partial
 from typing import Dict, Iterable, Optional, Set, Type, cast
@@ -49,9 +50,19 @@ class StatefulStaleMetadataRemovalConfig(StatefulIngestionConfig):
 
 
 @dataclass
+class EntityTypeDeletionSummary:
+    previous_checkpoint_count: int = 0
+    current_checkpoint_count: int = 0
+    soft_deleted_count: int = 0
+
+
+@dataclass
 class StaleEntityRemovalSourceReport(StatefulIngestionReport):
     soft_deleted_stale_entities: LossyList[str] = field(default_factory=LossyList)
     last_state_non_deletable_entities: LossyList[str] = field(default_factory=LossyList)
+    entity_type_deletion_summary: Dict[str, EntityTypeDeletionSummary] = field(
+        default_factory=dict
+    )
 
     def report_stale_entity_soft_deleted(self, urn: str) -> None:
         self.soft_deleted_stale_entities.append(urn)
@@ -237,6 +248,32 @@ class StaleEntityRemovalHandler(
 
         self._urns_to_skip.add(urn)
 
+    @staticmethod
+    def _build_checkpoint_comparison(
+        last_checkpoint_state: GenericCheckpointState,
+        cur_checkpoint_state: GenericCheckpointState,
+    ) -> Dict[str, "EntityTypeDeletionSummary"]:
+        previous_counts = Counter(
+            entity_type
+            for urn in last_checkpoint_state._urns_set
+            if (entity_type := guess_entity_type(urn))
+        )
+        current_counts = Counter(
+            entity_type
+            for urn in cur_checkpoint_state._urns_set
+            if (entity_type := guess_entity_type(urn))
+        )
+
+        return {
+            entity_type: EntityTypeDeletionSummary(
+                previous_checkpoint_count=previous_counts[entity_type],
+                current_checkpoint_count=current_counts[entity_type],
+            )
+            for entity_type in sorted(
+                set(previous_counts.keys()) | set(current_counts.keys())
+            )
+        }
+
     def gen_removed_entity_workunits(self) -> Iterable[MetadataWorkUnit]:
         if not self.is_checkpointing_enabled() or self._ignore_old_state():
             return
@@ -254,6 +291,13 @@ class StaleEntityRemovalHandler(
 
         assert self.stateful_ingestion_config
 
+        # Compute per-entity-type summary before any early-exit checks
+        # so the summary is always available in the report.
+        report = self.source.get_report()
+        assert isinstance(report, StaleEntityRemovalSourceReport)
+        report.entity_type_deletion_summary = self._build_checkpoint_comparison(
+            last_checkpoint_state, cur_checkpoint_state
+        )
         copy_previous_state_and_exit = False
 
         # If the source already had a failure, skip soft-deletion.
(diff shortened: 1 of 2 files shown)


sgomezvillamor:
Thanks for the PR! A few points before merging:

**❓ Scope question — is this actionable or deferred?**
The existing `fail_safe_threshold` is a global check across all entity types. A scenario where all `corpUser` entities disappear while datasets stay stable would not trigger it — the bulk deletion would proceed silently. This summary gives visibility after the fact, but there's no per-entity-type guard. Is a follow-up planned to use this data for per-entity-type thresholds?

---

**⚠️ `assert isinstance(report, StaleEntityRemovalSourceReport)` risk**
This assert now fires earlier in the flow — before all early-exit paths (`copy_previous_state_and_exit`) — so it triggers even when no deletions happen (e.g. on source failure runs). Any source using `StaleEntityRemovalHandler` with a report that doesn't inherit from `StaleEntityRemovalSourceReport` would now crash in more scenarios. Consider turning this into an `if isinstance` with a graceful fallback (skip summary population), so misconfigured sources degrade rather than hard-fail. Also, after this PR the same `assert isinstance` + `report = self.source.get_report()` pair appears twice in `gen_removed_entity_workunits` — the second one is now redundant.

---

**⚠️ `if entity_type in report.entity_type_deletion_summary` guard is undocumented**
```python
if entity_type in report.entity_type_deletion_summary:
    report.entity_type_deletion_summary[entity_type].soft_deleted_count += 1
```
At this point in the code, every URN came from `last_checkpoint_state`, so its entity type should always be in the summary. It's not clear what scenario this guard defends against. Either remove it (letting a `KeyError` surface any real bug) or add a comment explaining the concrete edge case. As written it obscures a potential bug rather than documenting intent.

---

**⚠️ `_build_checkpoint_comparison` accesses private attribute `_urns_set`**
```python
for urn in last_checkpoint_state._urns_set
```
`_urns_set` is a `PrivateAttr` internal to `GenericCheckpointState`. The public `last_checkpoint_state.urns` list should be used instead — consistent with how `get_urns_not_in` works and avoiding silent breakage if internals change.

---

**💡 Misleading truthiness check on `guess_entity_type`**
```python
if (entity_type := guess_entity_type(urn))
```
`guess_entity_type` returns `str`, not `Optional[str]`, and raises `AssertionError` on invalid URNs rather than returning `None`. The truthiness check only filters an empty string — it does not protect against malformed URNs. A comment clarifying this (or an explicit `if entity_type != ""`) would make the intent clearer.

---

**💡 Missing tests for `guess_entity_type` coverage**
`test_build_checkpoint_comparison` covers a few entity types by hand. It would be stronger to also assert that `guess_entity_type` works correctly for all registered entity types — the model registry (e.g. `entity-registry.yml` or the generated entity type constants) provides a canonical list. This would prevent future entity type additions from silently being miscounted.


---

aviraj-gour:
@sgomezvillamor Thanks for review!!

> Thanks for the PR! A few points before merging:
> 
> ❓ Scope question — is this actionable or deferred?
> The existing fail_safe_threshold is a global check across all entity types. A scenario where all corpUser entities disappear while datasets stay stable would not trigger it — the bulk deletion would proceed silently. This summary gives visibility after the fact, but there's no per-entity-type guard. Is a follow-up planned to use this data for per-entity-type thresholds?

This was discussed in [ING-2106] — per-entity-type **thresholds/warnings are intentionally paused**, not just deferred. The reasoning (per @maggiehays on the ticket): we don't yet have intuition for what a reasonable per-type threshold would be, and surfacing warnings without that grounding risks alert fatigue / customer confusion.

This PR is the deliberate first step: expose the per-entity-type breakdown in the report so we can hook it into internal observability. Once we've established realistic baselines from real data, we can revisit whether thresholded warnings or deletion-prevention rules make sense — but as a separate, data-informed decision, not a planned immediate follow-up.

So: actionable today (visibility + internal monitoring), threshold logic explicitly out of scope until we have baselines.

> 
> ⚠️ assert isinstance(report, StaleEntityRemovalSourceReport) risk
> This assert now fires earlier in the flow — before all early-exit paths (copy_previous_state_and_exit) — so it triggers even when no deletions happen (e.g. on source failure runs). Any source using StaleEntityRemovalHandler with a report that doesn't inherit from StaleEntityRemovalSourceReport would now crash in more scenarios. Consider turning this into an if isinstance with a graceful fallback (skip summary population), so misconfigured sources degrade rather than hard-fail. Also, after this PR the same assert isinstance + report = self.source.get_report() pair appears twice in gen_removed_entity_workunits — the second one is now redundant.

**1. Duplicate `report = ...; assert isinstance(...)` pair** — Good catch, fixed in the latest push. The earlier assignment at line 292 already validates and `report` stays in scope through the rest of the function.

**2. "Expanded crash surface" concern** — I'd push back on softening the assert. The contract that any source using `StaleEntityRemovalHandler` must report via `StaleEntityRemovalSourceReport` (or a subclass) is already enforced in [`_create_soft_delete_workunit`](https://github.com/datahub-project/datahub/blob/master/metadata-ingestion/src/datahub/ingestion/source/state/stale_entity_removal_handler.py#L231) — any source violating it would have crashed the moment a real deletion needed to happen. This PR just surfaces that latent misconfiguration earlier instead of letting it lurk until the first deletion.

A few points worth noting:

- The assert is reachable only when stateful ingestion is **enabled** AND a prior checkpoint **exists** — sources without stateful ingestion or on first run are unaffected (early `return` hits before the assert).
- `isinstance` is polymorphic, so any custom subclass (e.g. `MyCustomReport(StaleEntityRemovalSourceReport)`) passes fine — the contract is on hierarchy, not exact class.
- A graceful `if isinstance` fallback would silently mask the misconfiguration permanently.

If there's a specific source you're worried about, happy to verify and fix it. Otherwise I'd lean toward keeping the fail-fast behavior.


> ⚠️ if entity_type in report.entity_type_deletion_summary guard is undocumented
> if entity_type in report.entity_type_deletion_summary:
>     report.entity_type_deletion_summary[entity_type].soft_deleted_count += 1
> At this point in the code, every URN came from last_checkpoint_state, so its entity type should always be in the summary. It's not clear what scenario this guard defends against. Either remove it (letting a KeyError surface any real bug) or add a comment explaining the concrete edge case. As written it obscures a potential bug rather than documenting intent.

Yes, guard is dead code that masks bugs. removed it.

> ⚠️ _build_checkpoint_comparison accesses private attribute _urns_set
> 
> for urn in last_checkpoint_state._urns_set
> _urns_set is a PrivateAttr internal to GenericCheckpointState. The public last_checkpoint_state.urns list should be used instead — consistent with how get_urns_not_in works and avoiding silent breakage if internals change.

Using public urns instead of private _urns_set.

> 💡 Misleading truthiness check on guess_entity_type
> if (entity_type := guess_entity_type(urn))
> guess_entity_type returns str, not Optional[str], and raises AssertionError on invalid URNs rather than returning None. The truthiness check only filters an empty string — it does not protect against malformed URNs. A comment clarifying this (or an explicit if entity_type != "") would make the intent clearer.

Dropped if condition and calling guess_entity_type(urn) directly in the generator.

> 💡 Missing tests for guess_entity_type coverage
> test_build_checkpoint_comparison covers a few entity types by hand. It would be stronger to also assert that guess_entity_type works correctly for all registered entity types — the model registry (e.g. entity-registry.yml or the generated entity type constants) provides a canonical list. This would prevent future entity type additions from silently being miscounted.
> 

Thanks for the suggestion. I considered this but landed on declining for these reasons:

- `guess_entity_type` is a 2-line structural function (`urn.split(":")[2]`) — it has no hard-coded list of entity types, so a new type added in the future cannot be silently miscounted. As long as URNs follow the standard `urn:li:<type>:...` format, it works by construction.
- The PR uses `guess_entity_type` as an existing utility; comprehensive coverage of that utility belongs in `tests/unit/utilities/urns/`, not in this feature's tests.
- The current test exercises 3 entity types as a representative sample.

but it'd be unrelated to this change.

sgomezvillamor (approved):
LGTM
Thanks for addressing all concerns

PR #16915, opened 2026-04-06
feat(upgrade): remove legacy upgrade jobs

RyanHolstien:
Remove no longer needed upgrade jobs

### DIFF
--- a/docs/managed-datahub/release-notes/next.md
+++ b/docs/managed-datahub/release-notes/next.md
@@ -25,7 +25,6 @@ TBD
 
 - Upgrades to DataHub should be done in sequence by minor version for proper data migration to occur. Schema updates can occur in minor version changes and to migrate underlying data we are using a mutator mechanism with a one schema version look-back window to facilitate zero-downtime upgrades.
 
-
 New Features:
 
 - TODO


docs/how/updating-datahub.md
```diff
@@ -1,5 +1,7 @@
 # Updating DataHub
 
+- Upgrades to DataHub should be done in sequence by minor version for proper data migration to occur. Schema updates can occur in minor version changes and to migrate underlying data we are using a mutator mechanism with a one schema version look-back window to facilitate zero-downtime upgrades.
```
anshbansal, line 3: Would be good to add the version number where this is going to land with zero downtime upgrade. So this requirement becomes applicable from version X and not for all older versions

RyanHolstien: It won't show up in the old version documents, will only show up in future ones since those are a snapshot of a particular point in time. We don't have a version specified for the next release at the moment so I'm not sure what I would put here. v2.0?

RyanHolstien:
> There still are some backfill steps present. What scenarios would those be still required? like - BackfillBrowsePathsV2, BackfillDataProcessInstances, BackfillDatasetLineageIndexFields

I had removed just the ones called out in the tech spec as ones to fully remove, there were some that were mentioned as being kept so wasn't sure if it was intentional omissions or accidental. If we're good to remove everything I can.

RyanHolstien:
Parking until next release for backwards incompatible change

PR #16916, opened 2026-04-06
fix(ci): split frontend build into shards to reduce flakiness

RyanHolstien:
Changes targeted at making the frontend build much less flakey for Open Source contributors. My experience has been that basically the frontend build will hard-block PRs unless I run on the depot images and even then is extremely flakey. This shards out the frontend tests so that:
1. They take less time per split so it's easier to re-run flakes
2. Flake less often since work is distributed out to other runners
3. The logs are parseable since it's only running part of the frontend tests per shard and the Java frontend build is run on the main artifact build action

There are also some increased timeout relaxations to reduce failure rate on some very flakey tests and some clean up in the test logic that was previously manually constructing the Entity list in a second spot which increased maintenance load. This also does some refactoring to make it so the sharded builds don't do repeated work of creating the artifacts.

Not a fix for Cypress flakiness as it has not been a hard blocker in general for me while the frontend build has.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -218,7 +218,7 @@ jobs:
           override_branch: ${{ github.head_ref || github.ref_name }}
 
   frontend-build:
-    runs-on: ${{ vars.DEPOT_PROJECT_ID != '' && 'depot-ubuntu-latest-4' || 'ubuntu-latest' }}
+    runs-on: ${{ needs.setup.outputs.runner_type }}
     timeout-minutes: 30
     needs: setup
     if: ${{ needs.setup.outputs.frontend_change == 'true' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -228,12 +202,160 @@ jobs:
...
+          token: ${{ secrets.CODECOV_TOKEN }}
+      - name: Upload test results to Codecov on release
+        if: ${{ !cancelled() && github.event_name == 'release' }}
+        uses: codecov/test-results-action@0fa95f0e1eeaafde2c782583b36b28ad0d8c77d3 # v1
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          override_branch: ${{ github.head_ref || github.ref_name }}
+
+  frontend-build:
+    runs-on: ${{ vars.DEPOT_PROJECT_ID != '' && 'depot-ubuntu-latest-4' || 'ubuntu-latest' }}
```
devashish2203: This undos the changes @chriscollins3456 did in https://github.com/datahub-project/datahub/pull/16956 to allow running the fork PRs on Depot runners if a depot label is added.

If this split makes the test pass on ubuntu-latest then maybe Chris's changes are not needed at all.

PR #16929, opened 2026-04-07
feat(timeline): add Change History sidebar for Dataset, Domain, Glossary Term, and Data Product entities

max-datahub:
## Summary

Extends the Change History sidebar to cover **9 change categories** across **4 entity types**: Dataset, Glossary Term, Domain, and **Data Product**. Each entity type's category filter dropdown only shows the event types that are actually supported by the backend. Data Products support **asset membership tracking** — when datasets or other assets are added to or removed from a Data Product.

### Breaking Changes

- **`ChangeCategory.OWNER` renamed to `OWNERSHIP`** in the GraphQL enum. Clients querying the timeline/change history API with `OWNER` must update to `OWNERSHIP`. The REST API (`/timeline`) and the Python CLI accept both values for backward compatibility. This fixes an accidental 2022 divergence between the backend enum and the GraphQL schema. Documented in `docs/how/updating-datahub.md`.

### Supported event types per entity

| Event Type | Dataset | Glossary Term | Domain | Data Product |
|---|:---:|:---:|:---:|:---:|
| Schema changes | Yes | N/A | N/A | N/A |
| Tags | Yes | N/A | N/A | Yes |
| Glossary terms | Yes | Yes (related terms) | N/A | Yes |
| Documentation | Yes | Yes | Yes | Yes |
| Ownership | Yes | Yes | Yes | Yes |
| Domains | Yes | Yes | N/A | Yes |
| Structured properties | Yes | Yes | Yes | Yes |
| Applications | Yes | Yes | N/A | Yes |
| Asset membership | N/A | N/A | N/A | Yes |

- **Yes** = supported with a change event generator registered in `TimelineServiceImpl`
- **N/A** = entity does not have the relevant aspect(s) in the entity registry — structurally not applicable

There are no remaining "Not supported" gaps. Every aspect that exists on a supported entity has a registered generator.

### Aspect-to-generator mapping

Each registered aspect produces change events via a dedicated generator:

| Aspect | Entity | Category | Generator |
|---|---|---|---|
| `schemaMetadata` | Dataset | TECHNICAL_SCHEMA, TAG, DOCUMENTATION, GLOSSARY_TERM | `SchemaMetadataChangeEventGenerator` |
| `editableSchemaMetadata` | Dataset | TAG, DOCUMENTATION, GLOSSARY_TERM | `EditableSchemaMetadataChangeEventGenerator` |
| `globalTags` | Dataset, DataProduct | TAG | `GlobalTagsChangeEventGenerator` |
| `glossaryTerms` | Dataset, DataProduct | GLOSSARY_TERM | `GlossaryTermsChangeEventGenerator` |
| `ownership` | Dataset, GlossaryTerm, Domain, DataProduct | OWNERSHIP | `OwnershipChangeEventGenerator` |
| `domains` | Dataset, GlossaryTerm, DataProduct | DOMAIN | `SingleDomainChangeEventGenerator` |
| `structuredProperties` | Dataset, GlossaryTerm, Domain, DataProduct | STRUCTURED_PROPERTY | `StructuredPropertyChangeEventGenerator` |
| `applications` | Dataset, GlossaryTerm, DataProduct | APPLICATION | `ApplicationsChangeEventGenerator` |
| `institutionalMemory` | Dataset, Domain | DOCUMENTATION | `InstitutionalMemoryChangeEventGenerator` |
| `datasetProperties` | Dataset | DOCUMENTATION | `DatasetPropertiesChangeEventGenerator` |
| `editableDatasetProperties` | Dataset | DOCUMENTATION | `EditableDatasetPropertiesChangeEventGenerator` |
| `glossaryTermInfo` | GlossaryTerm | DOCUMENTATION | `GlossaryTermInfoChangeEventGenerator` |
| `glossaryRelatedTerms` | GlossaryTerm | GLOSSARY_TERM | `GlossaryRelatedTermsChangeEventGenerator` |
| `domainProperties` | Domain | DOCUMENTATION | `DomainPropertiesChangeEventGenerator` |
| `dataProductProperties` | DataProduct | DOCUMENTATION, ASSET_MEMBERSHIP | `DataProductPropertiesChangeEventGenerator` |

#### Out of scope — inverse-relationship problem (architecturally hard)

The timeline service queries aspects belonging to a single entity's own URN via the SQL primary key index `(urn, aspect, version)`. The events below require **reverse lookups** — finding all *other* entities whose aspects reference this entity. The SQL table has no index on JSON metadata content, so this would require a full table scan across potentially millions of rows. The UI can show current members because it uses Elasticsearch (which maintains a real-time inverted index on the latest state), but ES does not store version history needed for timeline diffs.

| Entity | Event |
|---|---|
| **Domain** | Asset additions/removals — assets reference the domain via *their own* `domains` aspect; the domain entity has no aspect listing its members |
| **Glossary Term** | Asset additions/removals — same inverse relationship; assets reference the term via their `glossaryTerms` aspect |
| **Dataset** | Data Product additions/removals — the data product owns the `dataProductProperties` aspect (which contains the `assets` array); the dataset has no aspect recording which data product it belongs to |

### What changed

**Backend:**
- Register Dataset, GlossaryTerm, Domain, and **Data Product** entity types in the timeline service with change event generators for all supported aspects
- Register **Application** change tracking for Datasets (previously missing despite the aspect existing)
- Add `ASSET_MEMBERSHIP` change category to enum and GraphQL schema for tracking Data Product asset additions/removals
- Add `DataProductPropertiesChangeEventGenerator` — diffs name/description (DOCUMENTATION) and `assets` array (ASSET_MEMBERSHIP) using sorted two-pointer merge on `destinationUrn`
- Add `StructuredPropertyChangeEventGenerator` with sorted two-pointer merge diff algorithm
- Add semantic diff support to `SingleDomainChangeEventGenerator`
- Add `DomainPropertiesChangeEventGenerator` — diffs name and description on the domain properties aspect (DOCUMENTATION category)
- Add `GlossaryRelatedTermsChangeEventGenerator` — diffs Inherits, Contains, Has Value, and Related To arrays on the glossary related terms aspect (GLOSSARY_TERM category)
- Add `ApplicationsChangeEventGenerator` — diffs applications URN array (APPLICATION category)
- Add count-based timeline fetching that reuses the existing `AspectDao.getAspectsInRange` with a full time range (epoch to now). DataHub's retention policy caps each (urn, aspect) pair to ~20 versions by default, so the result set is inherently bounded without needing a custom SQL `ORDER BY + LIMIT` query. The `maxChangeTransactions` cap is enforced after the diff computation.
- Add `DOMAIN`, `STRUCTURED_PROPERTY`, `APPLICATION`, and `ASSET_MEMBERSHIP` to the `ChangeCategory` enum
- **Rename `ChangeCategory.OWNER` → `OWNERSHIP`** to fix an accidental 2022 divergence between the backend enum (`OWNER`) and the GraphQL enum (`OWNERSHIP`). This eliminates the bidirectional mapping tables that were previously needed. `OWNER` is accepted as a backward-compat alias in the REST API (`StringToChangeCategoryConverter`) and the Python CLI.
- Fix `combineTransactionsByTimestamp` crash on unmodifiable lists

**Frontend:**
- Add Change History button (clock icon) to Domain, Glossary Term, and **Data Product** entity page headers
- Add a category filter dropdown to the change history and make it entity-type-aware — only shows categories supported by the backend for each entity type (including "Assets" for Data Products)
- Expand `changeEventToString` from 2 to 9 categories (schema, documentation, tags, terms, owners, domains, structured properties, applications, asset membership)
- Centralize all change-history constants (parameter keys, relationship type labels, category values) in `HistorySidebar.utils.ts` — the single source of truth for the backend-frontend contract
- Map technical relationship types to human-readable labels (Is A → "inherited term", Has A → "contained term", Has Value → "value term", Is Related To → "related term") with safe fallback for unknown types
- Strip entity URN references from all change event descriptions (redundant since the user is viewing that entity's history)
- Add actor attribution ("by {actorName}") next to each timestamp
- Add long-text truncation with Show more/less toggle for verbose change descriptions
- Lazy-load the timeline GraphQL query (only fires when sidebar opens)
- Add transaction limit of 100 events with truncation indicator in UI

### Tests

- `HistorySidebar.utils.test.ts` — 7 tests: per-entity category filtering (Dataset, GlossaryTerm, Domain, DataProduct with 9 categories), fallback for unknown types, display order preservation
- `changeEventToString.test.ts` — 42 tests: all categories, all operations, field-level variants, relationship type mapping (all 4 types + unknown fallback), entity URN stripping
- `DataProductPropertiesChangeEventGeneratorTest.java` — 14 tests: name add/change/remove, description add/change/remove, both changed, non-DOCUMENTATION category ignored, asset added/removed/swapped/no-change/large-diff/null-to-assets
- Backend: 41 Java tests across `ChangeEventMapper`, `ChangeTransactionMapper`, `GetTimelineResolver`, `CombineTransactionsByTimestamp`, `SingleDomainChangeEventGenerator`, `DataProductPropertiesChangeEventGenerator`, `TimelineService`
- `timeline_change_history_test.py` — 29 smoke tests: end-to-end GraphQL verification of all change categories across all 4 entity types (including ASSET_MEMBERSHIP for Data Products), actor attribution, response structure validation

## Test plan
- [x] `./gradlew :metadata-io:compileJava :datahub-graphql-core:compileJava` — compilation succeeds
- [x] `./gradlew :datahub-graphql-core:test` — all GraphQL tests pass
- [x] `./gradlew :metadata-io:test` — all timeline tests pass (14 DataProductProperties tests + existing)
- [x] `yarn type-check` — no TypeScript errors
- [x] `yarn test` — 42 frontend tests pass (changeEventToString) + 7 (HistorySidebar.utils) = 49 total
- [x] Java Spotless formatting — passes
- [x] Frontend ESLint — passes on all changed files
- [x] Live tested against `quickstartDebug` — verified all four entity types produce correct change events via GraphQL API and render correctly in the UI (seed script: `out/seed-timeline-demo-data.py`)
- [ ] Smoke test suite (`timeline_change_history_test.py`) — 29 tests across Dataset, GlossaryTerm, Domain, DataProduct

## Checklist
- [x] Tests added/updated
- [x] Breaking changes documented in `docs/how/updating-datahub.md` (`ChangeCategory.OWNER` → `OWNERSHIP` GraphQL enum rename)


datahub-web-react/src/app/entityV2/shared/EntityDropdown/EntityMenuActions.tsx
```diff
@@ -83,34 +82,38 @@ function EntityMenuActions(props: Props) {
...
-                                        ? new Set([EntityMenuItems.LINK_VERSION])
-                                        : new Set()
+                                    new Set([
+                                        ...(menuItems.has(EntityMenuItems.LINK_VERSION)
+                                            ? [EntityMenuItems.LINK_VERSION]
+                                            : []),
+                                        ...(menuItems.has(EntityMenuItems.CHANGE_HISTORY)
+                                            ? [EntityMenuItems.CHANGE_HISTORY]
+                                            : []),
+                                    ])
```
chriscollins3456: think we've got a minor bug here where if the menuItems has change history we will always show link version even if entity versioning isn't enabled

also i think what we can do is still show the history menu action component on the entity header when the entity side bar is closed. so basically i would revert the change to delete ChangeHistoryMenuAction.tsx and add it back up top in this file like you had it when `isClosed`. then you can remove this change i highlighted here altogether

datahub-web-react/src/app/entityV2/shared/EntityDropdown/EntityMenuActions.tsx
```diff
@@ -83,34 +82,38 @@ function EntityMenuActions(props: Props) {
                 <MenuItems $shouldFillAllAvailableSpace={shouldFillAllAvailableSpace} data-testid="entity-menu-actions">
                     <ExternalUrlMenuAction shouldFillAllAvailableSpace={shouldFillAllAvailableSpace} />
                     {menuItems.has(EntityMenuItems.MOVE) && <MoveEntityMenuAction />}
-                    {menuItems.has(EntityMenuItems.CHANGE_HISTORY) && <ChangeHistoryMenuAction />}
```
chriscollins3456: so yeah we could add this back to show it when the entity sidebar is closed still

datahub-web-react/src/app/entityV2/shared/tabs/Dataset/Schema/history/HistorySidebar.tsx
```diff
@@ -172,6 +184,18 @@ const HistorySidebar = ({ open, onClose, urn, siblingUrn, versionList, hideSeman
...
+                        <CloseOutlinedIcon />
+                    </CloseIcon>
+                </FieldHeaderWrapper>
+                <FilterBar>
+                    <SearchBar
+                        placeholder="Search changes..."
+                        value={searchText}
+                        onChange={(val) => setSearchText(val)}
+                        width="100%"
+                        height="32px"
```
chriscollins3456: let's make this 36px instead to match the same height as the filter on its right

datahub-web-react/src/app/entityV2/shared/tabs/Dataset/Schema/history/HistorySidebar.utils.ts
```diff
@@ -105,3 +105,26 @@ export function getCategoryOptions(entityType?: EntityType): CategoryOption[] {
...
+    const filterBySearch = lowerSearch.length > 0;
+
+    if (!filterByCategory && !filterBySearch) return entries;
+
+    const selected = new Set(selectedCategories);
+    return entries.filter((entry) => {
+        const changes = entry.transaction.changes ?? [];
+        const matchesCategory = !filterByCategory || changes.some((c) => c?.category && selected.has(c.category));
+        const matchesSearch =
+            !filterBySearch || changes.some((c) => c?.description?.toLowerCase().includes(lowerSearch));
```
chriscollins3456: not positive if there's a reasonably easy way to fix this but resolved names are not able to be searched for

so if i am looking for a change where the entity was added to a domain, unless that domain name is in the `urn` of the domain, you can't search for it.

datahub-web-react/src/app/entityV2/shared/EntityDropdown/EntityMenuActions.tsx
chriscollins3456: i don't feel super strong about pulling it out when the sidebar is closed, but do want to make sure we fix the bug

datahub-web-react/src/app/entityV2/shared/tabs/Dataset/Schema/SchemaTab.tsx
```diff
@@ -212,14 +209,6 @@ export const SchemaTab = ({ renderType, properties }: { renderType: TabRenderTyp
...
     return (
         <SchemaContext.Provider value={{ refetch }}>
-            <HistorySidebar
-                urn={urn}
-                siblingUrn={siblingUrn}
-                versionList={versionList}
-                hideSemanticVersions={hideSemanticVersions}
-                open={showSchemaTimelineView}
-                onClose={() => setShowSchemaTimelineView(false)}
-            />
```
chriscollins3456, line 222: okay i'm so sorry but now i think i'm regretting my earlier comment

i could easily see folks getting confused when this history sidebar button gets moved from the schema tab.

what do you think about keeping it there, but we pass in a default list of filters to just show schema changes if they open the history sidebar from the schema tab?  that way it's an easy way to just view schema related changes from the schema tab

then if they open it from the entity header the default is to show everything

chriscollins3456: and in fact i think removing the history button caused this cypress test to fail: `v2_schema_blame`

so if we decide not to go through with my suggestion we will need to update or remove this cypress test. but if we go through with my suggestion then i think this test will pass again

datahub-web-react/src/app/entityV2/shared/EntityDropdown/EntityMenuActions.tsx
max-datahub: Thanks for spotting the bug! Fixed.
Reverted the other change as well

datahub-web-react/src/app/entityV2/shared/tabs/Dataset/Schema/history/HistorySidebar.utils.ts
max-datahub: I adapted the search behavior, so that it now searches the displayed text properly:
>  Fixed — search now matches against the rendered display strings (with resolved entity names) instead of the raw backend descriptions. We pre-compute the display text for each transaction using getChangeEventString + the resolved nameMap, so searching for "Engineering" will find a domain change even if the URN is urn:li:domain:engineering.

datahub-web-react/src/app/entityV2/shared/tabs/Dataset/Schema/SchemaTab.tsx
max-datahub: Fixed and verified in the UI that different filters get passed in depending on which button is used:

> Brought the clock button back to the Schema tab. It now passes `defaultCategories={[ChangeCategoryType.TechnicalSchema]}` to the HistorySidebar, so opening from the Schema tab pre-filters to schema changes only. Opening from the entity header still shows all categories.

PR #16930, opened 2026-04-07
feat(zdu): background aspect schema version migration sweep infrastructure

chakru-r:
## Summary

Adds the infrastructure for a `NonBlockingSystemUpgrade` that migrates all aspect rows registered with the `AspectMigrationMutatorChain`. 

### DAO layer (`metadata-io`)

**`AspectMigrationsDao`** — new interface method:
- `hasAspectsNeedingMigration(aspectName, sourceVersions)`: LIMIT-1 pre-flight check; returns true if any v0 row for the aspect is at a source version. 

**`AspectDao`** — new method:
- `streamAspectBatches(aspectTargetVersions, afterCreatedOnMs, batchSize, limit)`: streams v0 rows ordered by `createdon ASC` across all registered aspects, pre-filtering to rows not yet at their target version. `afterCreatedOnMs=0` starts from the beginning; a non-zero value resumes after the last processed batch (cursor-based).

**`EbeanAspectDao`** implements both. All `schemaVersion` comparisons use LIKE-based `raw()` rather than Ebean's `jsonEqualTo()`. `jsonEqualTo()` on Postgres generates `(systemmetadata ->> 'schemaVersion')::bigint = ?` where the bind parameter is untyped, causing `operator does not exist: bigint = unknown` at runtime. LIKE on the serialised JSON text is DB-agnostic (Postgres + MySQL); `schemaVersion` values are small integers that cannot appear as substrings in other JSON field names or string values, so there is no false-positive/negative risk. Rows where the `schemaVersion` key is absent are treated as `DEFAULT_SCHEMA_VERSION` via a separate `IS NULL / NOT LIKE` clause.

**`CassandraAspectDao`**: stub implementations (unsupported; throws / returns null).

**`PartitionedStream`**: added `filter(Predicate<T>)` to allow in-memory post-filtering of streamed batches.

> **Note on `createdon` semantics**: on `version=0` rows, `createdon` reflects the last-write timestamp (updated on every upsert), not the original entity creation time. The true creation time is preserved in `systemMetadata.aspectCreated.time` across all writes. The sweep therefore processes least-recently-written aspects first.

### Upgrade step (`datahub-upgrade`)

**`MigrateAspects`** (`NonBlockingSystemUpgrade`):
- Discovers all aspects registered in the `AspectMigrationMutatorChain` and builds a per-aspect target version map (max `targetVersion` across mutators).
- Short-circuits to an empty step list when the flag is off or the chain has no registered mutators.
- Gated on `systemUpdate.migrateAspects.enabled` (default: `false`, env: `SYSTEM_UPDATE_MIGRATE_ASPECTS_ENABLED`).

**`MigrateAspectsStep`** (`UpgradeStep`):
- Step ID: `"migrate-aspects-<gitVersion>-<revision>"` — matches the `IncrementalReindexConfig` convention; encodes the deployment version so the step is re-run on each new deployment and skipped once `SUCCEEDED`.
- `skip()`: checks for prior `SUCCEEDED`/`ABORTED` result; then runs `hasAspectsNeedingMigration` pre-flight if the DAO supports it.
- `executable()`: loads `lastCreatedOnMs` from any prior `IN_PROGRESS` result, streams rows via `streamAspectBatches`, ingests each batch through `EntityService.ingestProposal` (triggering the mutation chain), persists `IN_PROGRESS` + cursor after each batch, writes `SUCCEEDED` on completion.
- `openStream` / `cursorState` / `loadResumeState` are `protected` to mirror a planned future base-class abstraction for resumable sweep steps.

### Configuration

`application.yaml` — new `systemUpdate.migrateAspects` block:
| Property | Env var | Default |
|----------|---------|---------|
| `enabled` | `SYSTEM_UPDATE_MIGRATE_ASPECTS_ENABLED` | `false` |
| `batchSize` | `SYSTEM_UPDATE_MIGRATE_ASPECTS_BATCH_SIZE` | `500` |
| `delayMs` | `SYSTEM_UPDATE_MIGRATE_ASPECTS_DELAY_MS` | `1000` |
| `limit` | `SYSTEM_UPDATE_MIGRATE_ASPECTS_LIMIT` | `0` (unlimited) |

## Test plan

- [ ] `./gradlew :datahub-upgrade:test` — `MigrateAspectsTest`, `MigrateAspectsStepTest`
- [ ] `./gradlew :metadata-io:test` — `PartitionedStreamTest`
- [ ] Manual: set `SYSTEM_UPDATE_MIGRATE_ASPECTS_ENABLED=true`, register a concrete `AspectMigrationMutator`, run `datahub-upgrade`, verify sweep log lines and `SUCCEEDED` upgrade record


metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectDao.java
```diff
@@ -234,6 +235,48 @@ public boolean checkIfAspectExists(@Nonnull String aspectName) {
...
+    // schemaVersion values are small integers that cannot appear as substrings of other
+    // JSON field names or string values (the key name "schemaVersion" is controlled by us).
+    //
+    // For DEFAULT_SCHEMA_VERSION we also need to catch rows where the schemaVersion key is
+    // absent entirely (AspectMigrationMutator treats absent == DEFAULT_SCHEMA_VERSION).
+    final io.ebean.Junction<EbeanAspectV2> or =
+        server
+            .find(EbeanAspectV2.class)
+            .where()
+            .eq(EbeanAspectV2.ASPECT_COLUMN, aspectName)
```
RyanHolstien: Aspect keyed searches on large instances are pretty slow from my experience, what's the biggest DB you've tested this query against and what's the performance look like? Mostly just curious if the exists is enough to bring it down to a good performance level and maybe it's because I'm typically querying for data.

metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectDao.java
```diff
@@ -679,6 +722,75 @@ private ExpressionList<EbeanAspectV2> buildExpressionList(
     return exp;
   }
 
+  @Override
+  @Nonnull
+  public PartitionedStream<EbeanAspectV2> streamAspectBatches(
```
RyanHolstien: We already have a function called streamAspectBatches, I think that likely some of this should be shared code and it should then have a helper function for any System Metadata based filtering needed on top with explicit naming rather than copying the same name. If we're using this method sometimes and the other sometimes based on flagging, it would make more sense to have the subset of operations within the same method. When combining please avoid further telescoping of the method from what already exists and prefer parameter objects to avoid method changes for new optional arguments.

metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectDao.java
```diff
@@ -679,6 +722,75 @@ private ExpressionList<EbeanAspectV2> buildExpressionList(
...
+    Map<String, Long> versionedAspects =
+        aspectTargetVersions.entrySet().stream()
+            .filter(e -> e.getValue() > DEFAULT_SCHEMA_VERSION)
+            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+
+    if (versionedAspects.isEmpty()) {
+      return PartitionedStream.<EbeanAspectV2>builder().delegateStream(Stream.empty()).build();
+    }
+
+    final String smCol = EbeanAspectV2.SYSTEM_METADATA_COLUMN;
```
RyanHolstien: nit: this adds an unnecessary abstraction around a constant making it distanced from how the constant is used when needed for code auditing.

metadata-io/src/main/java/com/linkedin/metadata/entity/cassandra/CassandraAspectDao.java
```diff
@@ -182,6 +182,13 @@ public boolean checkIfAspectExists(@Nonnull String aspectName) {
     return rs.one() != null;
   }
 
+  @Override
+  public boolean hasAspectsNeedingMigration(
+      @Nonnull String aspectName, @Nonnull Set<Long> sourceVersions) {
+    throw new UnsupportedOperationException(
```
RyanHolstien: Kind of off topic, but I feel like we just stop officially supporting Cassandra as a backend if we're just going to ignore implementing new features for it.

RyanHolstien (approved):
Looks mostly good, but would like to see the streamAspectBatches not create a new pathway for that method in the AspectDao

metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectDao.java
chakru-r: My testing so far has been functional than scale -- did not use a large test data. There there is any data to be migrated, this would stop at the first occurrence (equivalent of limit 1). That said, if there is _no_ data to be migrated, this does mean a full table scan. I am not sure how this can be mitigated. On rethinking on this, I will simply drop this pre-check and just hit the streamAspectBatches -- that may turn out to be a no-op after a full scan. Having the two step check is choosing between full scan + skip (when there are no updates) or skip quick + full scan when there are updates required. I was originally looking for a quick way to skip if no updates are needed, but that needs a full scan.

metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectDao.java
chakru-r: I do want to take another pass at this -- and possibly considering not using RestoreIndicesArgs  as the parameter but a filtering parameter object (since this function isnt just for restoreIndices though those args do work for some cases). With the current methods, there really just very few lines that are truly duplicated due to the filter criteria and ordering differences in the two approaches. Had left a comment of eventually

PR #16933, opened 2026-04-07
fix(ingest/powerbi): scan _all_ workspaces before looking up cross-workspace references

ligfx:
## Summary

PowerBI ingestion now scans every workspace before processing any of them, so cross-workspace dataset references resolve correctly. Previously a tile or report could reference a dataset that lived in another workspace, and if that workspace had not been scanned yet the reference was reported as missing — even though the dataset would be ingested moments later.

The scan now runs as a first pass over all workspaces, populating a global registry of datasets, endorsements, app metadata, and Fabric artifacts (Lakehouses, Warehouses, SQLAnalytics endpoints). The per-workspace processing pass then runs against that fully-populated view of the world.

## What you'll notice

- **Cross-workspace lineage is no longer dropped at random.** A tile in workspace A pointing at a dataset in workspace B will resolve regardless of the order workspaces are processed in.
- **DirectLake / Fabric lineage works again.** Lakehouse, Warehouse, and SQLAnalytics endpoint references that the source needs to resolve come from the scan response, which is the only place those keys exist.
- **Inactive or filtered-out workspaces no longer trigger redundant per-workspace API calls.** When the scan reports that a workspace was skipped (wrong type, deleted, etc.), the second pass leaves it alone.
- **Partial scan responses surface as warnings** instead of silently producing incomplete metadata. If the admin API returns successfully but omits a workspace you asked about, the source report now flags it as `Incomplete Scan Metadata` with the workspace name and id, so you can investigate (usually a permissions issue).
- **Fewer surprise crashes on malformed scan responses.** Truthy-but-empty payloads, active entries missing an id, missing `workspaces` keys, and unnamed workspaces are all handled defensively rather than raising.
- **Better performance on large tenants.** The per-workspace second pass uses a simple iteration instead of repeatedly shifting the front of a list, which matters when you have hundreds of workspaces.

## Behavior unchanged

- Stateful ingestion, checkpointing per workspace, and the `modified_since` flow all work identically — they're just driven from the new two-pass loop.
- All existing config knobs (`scan_batch_size`, `workspace_type_filter`, `extract_endorsements_to_tags`, etc.) keep their meaning. `scan_batch_size` still controls how many workspaces are scanned per admin-API call; the difference is that the registry is now built from all batches before the second pass begins.

## Test plan

- [x] 15 focused unit tests covering the two-pass flow, registry accumulation across batches, partial / malformed scan responses, the missing-workspace warning, the filter-skip path, and the `modified_since` branch


sgomezvillamor (approved):
Good fix and well-written tests. The two-pass architecture is the right solution — scanning all workspaces before processing any of them is the correct way to guarantee the dataset registry is fully populated for cross-workspace reference resolution. The 15 unit tests are high quality: no trivial tests, all imports at the top, precise `type: ignore` specifiers, and the batch-failure-isolation test covers a realistic production failure mode.

A few things worth addressing, ordered by importance.

---

**⚠️ HTTP scan errors silently bypass the Phase 1 reporter warning loop**

`_get_scan_result` converts HTTP errors (401, 403, 429, 5xx) into a `None` return rather than raising. The Phase 1 `try/except Exception` in `get_workunits_internal` therefore never fires for the most common production failure mode. `fill_metadata_from_scan_result` hits `if not scan_result: return filtered_ids` and returns silently — no per-workspace `reporter.warning` is filed, and the scan-derived metadata gap is invisible in the structured ingestion report (only a raw `log_http_error` log line appears).

The fix is to issue warnings inside `fill_metadata_from_scan_result` on the `None` path, mirroring what the `except` handler does for raised exceptions:

```python
if not scan_result:
    for ws in workspaces:
        self.__reporter.warning(
            title="Incomplete Scan Metadata",
            message="Scan API call failed; this workspace will be ingested without scan-derived metadata. Check logs for details.",
            context=f"workspace={ws.name} id={ws.id}",
        )
    return filtered_ids
```

---

**⚠️ Phase 1 exception handler double-logs**

The `except` block fires both `logger.warning(..., exc_info=True)` (once for the batch) and `reporter.warning(exc=e)` (once per workspace). Since `reporter.warning` already logs with the exception by default, this produces N+1 overlapping log entries for a single failure. Remove the standalone `logger.warning` call — the per-workspace `reporter.warning` loop is sufficient.

---

**⚠️ Verify the mock data removal doesn't break the cross-workspace integration test**

Removing the "Global Workspace" entry from `cross_workspace_mock_response.json` is a signal worth double-checking. The existing `test_powerbi_cross_workspace_reference_info_message` test asserts that a `"Missing Dataset Lineage For Tile"` info message **is present** — which was the broken behavior this PR fixes. If the fix now correctly resolves cross-workspace tiles, that assertion may no longer hold and the golden file may be stale. Please confirm whether `golden_test_cross_workspace_dataset.json` needs regenerating and whether the test should be updated to assert the fixed behavior (tile dataset resolved, warning absent for the cross-workspace tile).

---

**Suggestions (non-blocking)**

- `fill_metadata_from_scan_result` — the name implies reading from a scan result, but the method also mutates the workspace objects in place and updates `self.dataset_registry`. Something like `apply_scan_result_to_workspaces` or `populate_from_scan` would better capture both effects. Minor, but the current name surprised readers who expected a pure read.

- `get_workunits_internal` does two conceptually distinct things with different error-handling strategies. Extracting a `_run_phase1_scan(workspaces)` helper would make the two-phase structure explicit in the call tree and allow Phase 1 to be tested more directly.

- `scan_filtered_workspace_ids` — "scan filtered" is ambiguous (filtered *by* the scan, or filtered *for* the scan?). `excluded_workspace_ids` is clearer.

- The `if scan_filtered_workspace_ids:` guard before the list comprehension is a no-op when the set is empty. Dropping it simplifies the code.

- The three existing `if scan_result is None` guards in the same file and the new `if not scan_result` guard are inconsistent. Unifying them (preferably to `is None`, or to `not scan_result` with a `Workspace.scan_result: Optional[dict]` type annotation) would remove a future source of confusion.


---

dinesh-verma-datahub:
@sgomezvillamor  

Applied the following fixes:
  - Per-workspace reporter.warning on the not scan_result path
  - Removed standalone logger.warning from Phase 1 except block (double-logging)
  - Renamed scan_filtered_workspace_ids to excluded_workspace_ids
  - Dropped no-op if scan_filtered_workspace_ids: guard
  - Unified is None not scan_result across all guards (workspace.scan_result is always dict, never None)

 Skipped:
  - Mock data removal concern: the removed entry was a duplicate of "Global Workspace" inside the wrong scan result batch and the workspace still exists in its own scan result (81B02907), golden file is unchanged, all 112 tests pass
  - Rename fill_metadata_from_scan_result: fill_* is the established mutation convention in this file
  - Extract _run_phase1_scan helper:  Phase 1 is ~20 lines, already directly testable via batch-size test

sgomezvillamor:
Thanks for addressing all the feedback — the silent-failure fix, double-logging cleanup, variable renaming, and mock data clarification all look good. Keeping my approval.

---

PR #16934, opened 2026-04-07
feat(ingest/powerbi): generate external URLs for Workspaces

ligfx:


metadata-ingestion/src/datahub/ingestion/source/powerbi/config.py
```diff
@@ -369,6 +369,19 @@ def web_app_base_url(self) -> str:
...
+    def workspace_url(self, workspace_id: str, workspace_type: str) -> Optional[str]:
+        """Build a clickable PowerBI UI URL for a workspace.
+
+        Personal workspaces (``PersonalGroup``, ``Personal``) are not
+        addressable by id in the PowerBI UI - they are reachable only via
+        the special ``/groups/me`` alias and only by their owner. Surfacing
+        ``/groups/{guid}`` for those types would produce a dead link
+        (``GroupNotAccessible``), so we return ``None`` instead.
+        """
+        if workspace_type in ("PersonalGroup", "Personal"):
```
treff7es: These would be nicer as constants

treff7es (approved):

metadata-ingestion/src/datahub/ingestion/source/powerbi/config.py
dinesh-verma-datahub: moved to constants

PR #16935, opened 2026-04-07
fix(ingest/powerbi): remove redundant requests for dataset information

ligfx:
When the Admin API is being used, the connector gets a workspace scan result, which contains a list of datasets and their info, and then redundantly re-requests information for each individual dataset.

This PR removes the redundant request:
- `webUrl` now has to be imputed, as it doesn't appear in the scan result
- Confirmed manually that `description` does indeed get populated in the workspace scan results, updated test mocks.


sgomezvillamor (approved):
**Warning:** `connector-tests` CI check is failing — this will require updating golden files once the base branch (`mm--powerbi-feat-workspace-external-url`) is merged.

---

PR #16937, opened 2026-04-08
fix(ingest/powerbi): remove redundant report users calls

ligfx:
## Problem

When the PowerBI connector ingests a workspace, the admin scan response already contains, for every report, the metadata DataHub uses (id, name, type, webUrl, embedUrl, description, dataset id) and the list of users with access. Despite having that data in hand, the connector still issued, per workspace whose scan succeeded:

- a per-workspace `/reports` call to refetch the same reports, on every ingestion regardless of `extract_ownership`, and
- a per-report `/admin/reports/{id}/users` call for every report — only when `extract_ownership` was enabled (the helper had an internal early-return when it was off, so the ownership-disabled case never produced users traffic).

For workspaces with many reports, the per-report users calls multiplied API load against the PowerBI admin API and increased throttling exposure.

Three related issues were uncovered while removing those calls:

- A report with a missing or unknown `reportType`, or a missing `id`, raised `KeyError` from inside the report list comprehension; the surrounding `try/except` swallowed the error but dropped **every** report for the workspace — not just the malformed record. Missing `name` silently produced report entities with `None` in fields typed `str`.
- `User.emailAddress` and `Report.embedUrl` were typed as `str`, but PowerBI legitimately omits them (App principals have no email; scan entries may omit `embedUrl`). Python did not crash on the mismatch but materialised entities with values that did not match their declared types.
- The duplicate report records PowerBI emits when a report is published to an App (the duplicate carries `appId`; the original does not) were filtered out only on the per-workspace `/reports` path — the scan-result path did not run the filter because it did not consume scan reports at all.

### Use the scan-result data

Report ingestion now consults `workspace.scan_result` first. When the scan result is present, both the report list and the users on each report are built directly from it; no HTTP request is sent to the per-workspace `/reports` endpoint or to the per-report `/users` endpoint. Whether per-report users are populated from the scan response is controlled by the existing `extract_ownership` flag inside the new `new_powerbi_reports` factory.

When the scan result is empty (the admin scan failed for that workspace or the principal lacks the permission to run it), report ingestion falls back to the previous behaviour: the per-workspace `/reports` endpoint is called and, only when `extract_ownership` is enabled, the per-report `/users` endpoint is called for each report. The ownership-disabled case therefore issues no users traffic on either path. Tags continue to come from the workspace endorsements already collected.

Page (chart) information is still fetched from PowerBI for every report on both paths, but now one report at a time with per-report error handling: a failure surfaces a `Report Pages Not Fetched` warning on the source report and leaves `report.pages = []`, so the affected report still ingests without its chart children rather than the whole workspace failing.

### Tolerate the API's actual response shapes

Field-level fixes in the new `new_powerbi_reports` / `new_powerbi_user` factories:

- `User.emailAddress` and `Report.embedUrl` are widened to `Optional[str]`. Service principals (`principalType=App`) without email are accepted; reports without `embedUrl` are accepted.
- A report whose `reportType` is missing or not in the known enum is skipped individually with a warning that lists the known types, instead of raising `KeyError` from inside the comprehension and dropping every report for the workspace.
- A report missing `id`, `name`, or `reportType`, or a user record missing `identifier`, `displayName`, `graphId`, or `principalType`, is skipped individually with a warning instead of silently materialising an entity with `None` in fields typed `str`.
- The App-duplicate filter (drop the entry carrying `appId`, keep the one that does not) is now applied inside the factory, so both the scan path and the `/reports` fallback path de-duplicate consistently.

Each skip path emits a log line — `DEBUG` for the expected app-duplicate filter, `WARNING` for the genuinely surprising shapes — so operators can see in the log what was dropped and why.

### Synthesise paginated report URLs with the right prefix

When the factory needs to synthesise a report's external URL because the scan entry omitted `webUrl`, it now picks the prefix based on report type: `/rdlreports/{id}` for `PaginatedReport`, `/reports/{id}` for the rest. URLs PowerBI returned directly are still used verbatim; if neither a raw URL nor `workspace.webUrl` is available (e.g. `PersonalGroup`), the URL is left as `None`.

### Operator visibility

Three reporter entries are emitted against the source report so the degradation modes are visible without reading raw logs:

- `Report Scan Fallback Active` — info, emitted once per workspace when the fallback path engages.
- `Reports Not Fetched` — warning, emitted when the fallback's `/reports` call raises.
- `Report Pages Not Fetched` — warning, emitted per report when the pages fetch raises.

A new `#### Admin scan vs. workspace listing` section in `powerbi_post.md` documents which path runs when and points operators at the `Report Scan Fallback Active` entry for correlation.

## What changes for users

- **Admin-mode ingestion is faster against PowerBI.** When the workspace scan succeeds, the per-workspace `/reports` call and (when ownership extraction is on) the per-report users call are no longer issued.
- **Workspaces no longer disappear from DataHub because of one bad PowerBI record.** A report with a missing `id`, a missing/unknown `reportType`, or a pages-endpoint failure used to abort all reports for the workspace; each is now contained to the offending record.
- **Paginated report links generated by the connector now resolve.** Synthesised URLs use `/rdlreports/{id}` for `PaginatedReport`. (No effect when PowerBI returned a `webUrl` directly, which is the common case; the synthesis is only used when the scan entry omits `webUrl`.)
- **Pages-fetch failures degrade gracefully.** The affected report ingests with `pages=[]` and a warning; the workspace continues.
- **Operator visibility** for fallback and degradation, via the three reporter entries above.

There are no configuration changes required; behaviour is controlled entirely by existing flags (`extract_ownership`, `ownership.create_corp_user`, `extract_endorsements_to_tags`). For every combination of those flags, the set of corp-user MCPs emitted is identical to the set emitted before this PR. Internally, `Mapper.to_datahub_users` now gates on `extract_ownership AND create_corp_user`; that is a defensive duplicate of the upstream user-fetch gating that was already in place (when `extract_ownership` is off, upstream code never populates `Report.users`, so `to_datahub_users` was already running against an empty list), not a new opt-out.

## Tests

- Existing PowerBI integration tests have been updated so the mock fixtures match the new traffic pattern: the redundant per-workspace `/reports` and per-report `/users` endpoints are no longer mocked, and the report metadata and users that the connector previously fetched separately are now embedded inline in the mock scan response. Golden files reflect (a) corrected workspace IDs in synthesised report URLs and (b) the additional dataset/container/dashboard aspects emitted now that the second mock report (a paginated report in the second workspace) is consistently visible to the test ingestion (it was previously dropped by a swallowed `/reports` HTTP error).
- New unit tests cover each hardened branch independently: app-duplicate skipping, missing required fields on reports, unknown report types, the paginated vs regular URL prefix choice, `description=null` coercion to `""`, the `PersonalGroup` case where no URL can be synthesised at all, the case where PowerBI returned a URL directly (used verbatim), `embedUrl` Optional (present + absent), the `extract_ownership` flag's effect on user construction, and the App-principal-without-email factory case.
- The user factory has a dedicated test that verifies each required field being absent or null returns `None`.
- A `TestToDatahubUsersGate` class exercises `Mapper.to_datahub_users` end-to-end for all four `(extract_ownership, create_corp_user)` combinations and the App-principal filter.


sgomezvillamor (commented):
Great cleanup — the forensic story is clear and the factory approach is a solid improvement. A few things worth addressing:

---

**🟡 Warning — Fallback path: `extract_ownership` flag silently ignored + `get_report_users` unguarded**

In the `else` branch of the new `get_reports()`, the `extract_ownership` gate from the old `fill_ownership()` is gone. Users who set `extract_ownership: false` on the fallback path silently trigger ownership API calls anyway. The call is also outside any try/except, so any HTTP failure propagates and can abort ingestion for all remaining workspaces — unlike the scan path, which handles user-parsing failures per-record.

Related: the hardcoded `extract_ownership=False` in `data_resolver.py`'s refactored `get_reports()` is correct only because the caller populates ownership separately — but it reads like a configuration-ignoring bug without a comment.

Suggested fix for the fallback:
```python
if self.__config.extract_ownership:
    for report in reports.values():
        try:
            report.users = self.get_report_users(workspace.id, report.id)
        except Exception:
            self.log_http_error(...)
            self.reporter.warning(title="Report Users Not Fetched", message="...", context=f"workspace={workspace.name}")
```

---

**📝 Strongly Recommended — Document ingestion modes**

This PR makes admin-mode meaningfully faster: one workspace scan vs O(workspaces + reports) per-entity calls. Worth capturing in `powerbi_pre.md`: a short table comparing admin vs regular mode (performance, pages, profiling, ownership) and a note that when the scan returns empty results the connector gracefully falls back to per-workspace calls. Users should know to watch for that in the ingestion report.

---

**ℹ️ Minor — Reporter visibility + LiteralString**

Two small adjacent fixes: (1) `reporter.warning(message=f"Pages for report '{report.name}'...")` — `message` is typed `LiteralString` so the f-string is a type error; move `report.name` to `context`. (2) The fallback `except Exception` only calls `log_http_error()`, which goes to the Python logger but not the ingestion report — a `reporter.warning()` alongside it would surface workspace-level failures in the DataHub UI.

---

**ℹ️ Minor — No signal when fallback path activates**

When `workspace.scan_result` is empty and `get_reports()` silently takes the fallback path (e.g. after a permissions change), nothing in the ingestion report indicates this. A `reporter.info("Report Scan Fallback Active", ...)` at the top of the `else` branch would let operators catch it without digging into logs.

---

**ℹ️ Minor — Test coverage gaps to consider**

- `to_datahub_users()` gate change (`extract_ownership=False, create_corp_user=True → []`) has no direct test; current tests simulate the logic inline rather than calling the method.
- Fallback path with `extract_ownership=False` not asserted to skip users API calls.
- Pages fetch failure on the non-scan (fallback) path untested — the existing pages-failure test only exercises the scan path.

---

dinesh-verma-datahub:
@sgomezvillamor  Thanks, all five addressed in commit 239a940313:

  1. Added the extract_ownership outer gate on the fallback. Didn't wrap get_report_users in a per-call try/except — _get_entity_users already catches Exception and returns [], so a wrapper would just duplicate
  that. Comment added on the hardcoded extract_ownership=False in data_resolver.py.
  2. New Admin scan vs. workspace listing section. Put it in powerbi_post.md rather than _pre.md — Capabilities lives in _post.md per the connector-docs convention.
  3. Moved report.name into context for the pages warning, and added a Reports Not Fetched reporter warning alongside log_http_error in the fallback except.
  4. Report Scan Fallback Active info at the top of the else branch.
  5. TestToDatahubUsersGate exercises Mapper.to_datahub_users end-to-end across the three meaningful flag combinations ((False, True), (True, False), (True, True)) plus an App-principal case that pins the upstream
   principalType == "User" filter. New tests also assert the fallback path skips the users API when ownership is off, and that pages-failure recovery works on the fallback path.

  Picked up along the way: User.emailAddress and Report.embedUrl widened to Optional[str] (App principals; scan entries can omit embedUrl), per-record skip-and-continue on malformed factory inputs with log lines,
  App-duplicate filter consolidated so it runs on both paths, and PaginatedReport synthesised URLs use /rdlreports/{id}.

  Heads-up — also visible in the diff but intentionally not changed in this PR:

  - Per-record skips in new_powerbi_reports / new_powerbi_user emit logger.warning only, not a reporter.warning. Same operator-visibility gap you flagged for the fallback path, just one layer deeper. Plumbing a
  reporter into the data-class factories changes their contract — happy to do it in a follow-up if you'd like, but felt out of scope here.
  - AdminAPIResolver.get_pages_by_report still silently returns [] for admin-only ingestion, so reports under that mode never get a Report Pages Not Fetched signal. Pre-existing; worth a separate one-shot
  reporter.info at API init.
  - Container golden grew ~1240 lines. This is a previously-silently-dropped report from the second mock workspace that now ingests correctly — the old /reports HTTP call for that workspace was raising and the
  exception was being swallowed. Latent bug fix exposed by the refactor.
  - Some golden URLs shifted GUIDs (f089354e-... → 64ED5CAD-...). This is a mock-fixture correction — the old /reports mock returned a workspace GUID that didn't match the workspace's actual ID, and the connector
  trusted that response. The new code synthesises URLs from workspace.webUrl, which uses the real workspace ID. No production-customer impact (PowerBI's real responses don't carry that mismatch).

sgomezvillamor (approved):
Thanks for the thorough follow-up — all concerns addressed, LGTM.

Agreed on the reporter-in-factories point: threading a reporter into `new_powerbi_reports` / `new_powerbi_user` would change their contract in a way that doesn't feel right. The `logger.warning` at that layer is the correct boundary; the follow-up idea of surfacing dropped-record counts at the caller level is worth considering when the time comes.

---

PR #16947, opened 2026-04-08
fix(ingest/powerbi): remove redundant calls to dashboard users and tiles

ligfx:
## Summary

The PowerBI connector was making redundant per-dashboard API calls to fetch users and tiles for every dashboard during ingestion, even when that data was already available in the admin scan result. This caused unnecessary API load and rate-limit pressure on the PowerBI admin API.

## Root cause

Dashboard users and tiles were always fetched via dedicated per-resource API calls, regardless of whether the scan result (retrieved during the earlier workspace scan phase) already contained that data.

## Fix

- When a scan result is present, dashboards (with their tiles and users), reports (with their users), and datasets are now constructed directly from the scan result data rather than issuing additional API requests. For dashboards and reports, per-resource API calls are retained only as a fallback when no scan result is available; datasets require the scan result.
- The two-phase ingestion flow (scan all batches first, then process each workspace) is now made explicit: the scan phase populates datasets into a shared registry so cross-workspace references resolve correctly before any workspace is processed.
- Workspace URLs are now derived from the environment base URL and workspace ID (the admin API does not return them), and are used to impute missing `webUrl` fields on datasets, reports, and dashboards rather than leaving them null.
- `emailAddress` on users is made optional to handle service principals (principalType=App), which do not have one. `embedUrl` on reports, dashboards, and tiles and `title` on tiles are also made optional to match actual API response variability.
- Scan failures in the first phase now emit per-workspace warnings to the source report and allow ingestion to continue with empty scan metadata, rather than silently losing all workspaces in an affected batch.
- Skipped entities (dashboards, reports, tiles with missing required fields) and missing scan-result coverage are surfaced as reporter warnings so gaps are discoverable without reading raw logs.


sgomezvillamor (approved):
LGTM — clean fix. Left one comment with observability suggestions for follow-up.

---

sgomezvillamor:
Nice work removing the redundant API calls. A few observability gaps worth addressing in a follow-up:

**1. No signal for which code path is taken**
`fill_dashboards()` and `get_reports()` silently switch between the scan path and the per-resource fallback with no log entry. A simple `logger.info("Using scan result for dashboards in workspace %s", workspace.name)` / `logger.info("Falling back to per-resource API for workspace %s", workspace.name)` at each branch entry would make this diagnosable.

**2. No scan-path vs. fallback counters in `PowerBiDashboardSourceReport`**
There is no way for an operator to confirm the optimisation is working in production. Adding counters like `workspaces_via_scan_path` / `workspaces_via_fallback_api_path` (and equivalents for reports/dashboards) would surface silent fallbacks without reading logs.

**3. No timing for the two-phase flow**
Phase 1 (scan batches) and Phase 2 (per-workspace processing) in `get_workunits_internal` have zero timing instrumentation. `PerfTimer` is already used for M-query parsing in the same report class — applying the same pattern here would make slow workspaces or unexpected fallbacks immediately visible.

---

PR #16974, opened 2026-04-10
fix(docs): Remove wrong indentation in glue docs

pedro93:


lakshay-nasa (approved):

AdrianMachado (approved):
I think we usually tag title these as `docs()` rather than `fix()` but doesn't really matter

anshbansal (approved):

PR #16980, opened 2026-04-10
fix(ui/glossary): fix copy name on glossary term returns uuid instead of term name

purnimagarg1:
**Linear ticket:**
https://linear.app/acryl-data/issue/CAT-1706/copy-name-on-glossary-term-returns-uuid-instead-of-term-name

**Description:**

This fixes the bug where Copy Name on a glossary term was returning UUID instead of the term name.


chriscollins3456 (approved):
nice

PR #16982, opened 2026-04-10
fix(auth): ignore deprecated corpUserInfo.active for session eligibility

david-leifker:
Remove LoginDenialReason.INACTIVE and stop using corpUserInfo.active=false in ActorContext.isActive and UserSessionEligibilityChecker. Update docs and tests accordingly.

Made-with: Cursor


AdrianMachado (approved):
Accepting to unblock

RyanHolstien (approved):

PR #16991, opened 2026-04-11
fix(ingestion): stream JSON schema inference to prevent OOM on large files

kyungsoo-datahub:
Replace ujson.load() with ijson streaming parser capped at max_rows.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/schema_inference/json.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/schema_inference/json.py
@@ -2,6 +2,7 @@ import itertools
 import logging
 from typing import IO, Dict, List, Type, Union
 
+import ijson
 import jsonlines as jsl
 import ujson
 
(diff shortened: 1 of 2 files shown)


metadata-ingestion/src/datahub/ingestion/source/schema_inference/json.py
```diff
@@ -50,12 +51,26 @@ def infer_schema(self, file: IO[bytes]) -> List[SchemaField]:
...
+                # ijson.items(file, 'item') lazily yields elements of a top-level JSON array.
+                file.seek(0)
+                datastore = list(
+                    itertools.islice(ijson.items(file, "item"), self.max_rows)
+                )
+                if not datastore:
+                    # Not a top-level array — likely a single JSON object.
+                    file.seek(0)
+                    datastore = [ujson.load(file)]
+            except (ijson.common.JSONError, UnicodeDecodeError) as e:
```
treff7es:
```suggestion
            except (ujson.JSONDecodeError, ijson.common.JSONError, UnicodeDecodeError) as e:
```

treff7es: `ujson.JSONDecodeError` needs to be added as well because you use ujson as a fallback

metadata-ingestion/src/datahub/ingestion/source/schema_inference/json.py
```diff
@@ -50,12 +51,26 @@ def infer_schema(self, file: IO[bytes]) -> List[SchemaField]:
...
+                # Stream-parse to avoid loading the entire file into memory.
+                # ijson.items(file, 'item') lazily yields elements of a top-level JSON array.
+                file.seek(0)
+                datastore = list(
+                    itertools.islice(ijson.items(file, "item"), self.max_rows)
+                )
+                if not datastore:
+                    # Not a top-level array — likely a single JSON object.
+                    file.seek(0)
+                    datastore = [ujson.load(file)]
```
treff7es, line 63: Can `ujson.load(file)` be a list? If it can be, there is a chance we double wrap it

treff7es (approved):

metadata-ingestion/src/datahub/ingestion/source/schema_inference/json.py
kyungsoo-datahub: If the file were a top-level array, ijson.items would have already yielded elements. We only reach ujson.load() when it's not an array, so double-wrapping won't happen in practice.

PR #16975, opened 2026-04-10
docs(rfc): RFC for TLS/mTLS support in Kafka and OIDC

manuschillerdev:
## Summary

We self-host DataHub Core on AKS and experienced several issues around TLS / mTLS - especially in the Python Codebase. 

It is currently possible to configure DataHub Core with TLS, but it is tedious, because the Helm-Chart interface is not ideal, and we need to manual run patches in side-car containers for the python workloads to make it work.
Being able to use Client Certificates additionally to client-credentials for OIDC/SSO would be a great addition as well (https://github.com/datahub-project/datahub/pull/15479)

RFC proposing X.509 client-certificate authentication (TLS / mTLS) as a first-class citizen for Kafka, Schema-Registry and OIDC. The goal woud be a single, consistent configuration surface for keystores and truststores honored identically by GMS, `mae-consumer`, `mce-consumer`, `datahub-actions`, `datahub-frontend`, `acryldata/datahub-helm`, and `acryldata/mcp-server-datahub`.

## Motivation
enterprise (banking, insurance, public sector) deployments move toward credential-less service auth (cert-manager, Vault PKI, SPIFFE). Short-lived certs rotate without config churn and unify the PKI story for Kafka + OIDC into one identity.

## Related issues

Pain points this RFC consolidates:

- datahub-project/datahub#4287 — Actions container not configurable with TLS
- datahub-project/datahub#14354 — MAE/MCE consumers can't reach Schema Registry with TLS + basic auth
- datahub-project/datahub#14568 — `datahub-actions` missing connection parameters
- datahub-project/datahub#14576 — `SchemaRegistryClient` doesn't pass `ssl.ca.location` to SSL context
- datahub-project/datahub#5786 — actions pod can't reach MSK TLS-only brokers
- datahub-project/datahub#13223 — Kafka OAUTHBEARER misconfigured
- acryldata/datahub-helm#601 — chart has no unified SSL story (java→JKS vs python→PEM)
- acryldata/datahub-helm#687 — SCRAM auth against AWS MSK broken
- acryldata/datahub-helm#79 — `kafka-setup` unconditionally emits `ssl.keystore.*`
- acryldata/mcp-server-datahub#52 — MCP server has no custom-CA path for GMS

Adjacent (not superseded, explicitly out of scope): datahub-project/datahub#15755 (server-side mTLS for GMS Jetty listener).

## Related PRs

- datahub-project/datahub#16997 — `fix(gms): forward ssl.keystore.type/truststore.type to schema registry client`
[...]

Issue #14354, opened 2025-08-06
Unable to Connect to Kafka Schema Registry using TLS and basic-auth from datahub-mae-consumer and datahub-mce-consumer Despite Matching Configuration


NunoTrigo1986:
**Describe the bug**

I am unable to connect to our Kafka Schema Registry (which uses TLS and Basic Auth) from both `datahub-mae-consumer` and `datahub-mce-consumer` containers. The connection fails even though I use the exact same configuration and environment variables as other DataHub Java containers (`gms`, `datahubSystemUpdate`) where the connection works successfully.

---

**To Reproduce**

1. Deploy `datahub-mae-consumer` and `datahub-mce-consumer` with the provided keystore/truststore configuration and `JDK_JAVA_OPTIONS`.
2. Attempt to connect to a secured (TLS + Basic Auth) Kafka Schema Registry endpoint.
3. Only these consumers fail to connect—other DataHub Java containers with the identical setup connect successfully.
4. See error in the pod logs.

---

**Expected behavior**

Both `datahub-mae-consumer` and `datahub-mce-consumer` should connect to the Kafka Schema Registry with the provided configuration, just like other DataHub Java containers do.

---

**Screenshots / Logs**

```
2025/08/06 19:35:49 Problem with request: Get "https://schemaregistry-bootstrap-lb.kafka.xx.test.xx.xxx:443/": tls: failed to verify certificate: x509: certificate signed by unknown authority. Sleeping 1s
```

---

**Environment**

- **OS:** Kubernetes containers (DataHub base image)
- **DataHub Helm chart:** `datahub/datahub` 0.6.15 (App version: v1.2.0)
- **DataHub Prerequisites Helm chart:** `datahub/datahub-prerequisites` 0.1.15

---

**Additional context**

Configuration (shared by all DataHub components, both working and failing):
```yaml
extraVolumes:
  - name: cert-vol
    secret:
      secretName: selvbetjent-analyse-kafka-sertifikat-keystore-jks 
extraVolumeMounts:
  - name: cert-vol
    mountPath: /mnt/datahub/certs

env:
  - name: JDK_JAVA_OPTIONS
    value: "-Djavax.net.ssl.trustStore=/mnt/datahub/certs/truststore.jks \
        -Djavax.net.ssl.trustStoreType=JKS \
        -Djavax.net.ssl.trustStorePassword=password \
        -Dspring.kafka.properties.ssl.keystore.location=/mnt/datahub/certs/keystore.jks \
[...]

PR #16997, opened 2026-04-13
fix(gms): forward ssl.keystore.type/truststore.type to schema registr…

manuschillerdev:
Adds `ssl.keystore.type` / `ssl.truststore.type` to the `@Value` allowlist in `KafkaSchemaRegistryFactory`. Without this, the factory silently drops those two keys from the schema-registry props map, and the consumer/producer factories then `putAll()` the factory's output **after** Spring Boot's auto-binding, so any `SPRING_KAFKA_PROPERTIES_SCHEMA_REGISTRY_SSL_KEYSTORE_TYPE=PEM` the operator set
gets overwritten. This blocks PEM TLS for the Confluent schema-registry REST hop from GMS/MAE/MCE consumers, even though the client itself fully supports PEM via its own `SslFactory` (https://github.com/confluentinc/schema-registry/blob/v8.0.0/client/src/main/java/io/confluent/kafka/schemaregistry/client/security/SslFactory.java).

The two new fields default to empty string and are **only forwarded to the props map when actually set** — existing JKS deployments that rely on the client's default `type` behavior are unaffected.

This is one of the code fixes from RFC [datahub-project/datahub#16975](https://github.com/datahub-project/datahub/pull/16975) (PEM-first TLS for DataHub outbound connections). Adjacent to [datahub-project/datahub#14354](datahub-project/datahub/#14354) (MAE/MCE consumers can't reach Schema Registry with TLS) and [acryldata/datahub-helm#601](https://github.com/acryldata/datahub-helm/issues/601) (chart has no unified SSL story)

### DIFF
--- a/metadata-service/factories/src/main/java/com/linkedin/gms/factory/kafka/schemaregistry/KafkaSchemaRegistryFactory.java
+++ b/metadata-service/factories/src/main/java/com/linkedin/gms/factory/kafka/schemaregistry/KafkaSchemaRegistryFactory.java
@@ -57,10 +57,18 @@ public class KafkaSchemaRegistryFactory {
     props.put(AbstractKafkaSchemaSerDeConfig.SCHEMA_REGISTRY_URL_CONFIG, kafkaSchemaRegistryUrl);
     props.put(withNamespace(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG), sslTruststoreLocation);
     props.put(withNamespace(SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG), sslTruststorePassword);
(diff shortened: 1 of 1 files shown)


david-leifker (approved):
Happy to approve this, is it possible to generate some test coverage for the unit test requirements? I know this is just configuration boilerplate, however the check is currently blocking merge.

david-leifker (approved):

david-leifker:
@manuschillerdev Thanks for the contribution, the blocker to merging this PR is unit test code coverage. Please ping me when updated.

manuschillerdev:
> @manuschillerdev Thanks for the contribution, the blocker to merging this PR is unit test code coverage. Please ping me when updated.

hey @david-leifker - thank you so much for looking into this. I will have time to work on this and properly test it this week. Will gladly add tests as well.

OliverLeighC:
Any update on this?

PR #16998, opened 2026-04-13
fix(ui): fix duplication of owners in applications

v-tarasevich-blitz-brain:
Linear: https://linear.app/acryl-data/issue/CAT-1708/duplicate-owner-added-when-creating-application-due-to-default-owners

This PR fixes duplication of owners when you're adding yourself as owner too:
- added new optional field to the graphql endpoint to add creator as owner (it's true by default)
- added skipping of adding creator as owner (that optional field is false) when we create an application in UI
- the input is filled with the current user by default


chriscollins3456 (approved):
nice!

PR #16999, opened 2026-04-13
feat(ingestion/athena): support s3 tables (iceberg) in athena ingestion

acrylJonny:
## Summary

Bring the Athena connector into shape for S3 Tables (Iceberg) catalogs and fix three independent column-handling bugs surfaced while ingesting them.

### Athena S3 Tables (`s3tablescatalog/<bucket>` catalog)

- `CustomAthenaRestDialect.get_table_names` overridden to include `ICEBERG` in the allowed table types — PyAthena's base method drops it, so every S3 Tables table was silently invisible (`tables_scanned: 0`, no errors).
- `_list_tables_via_boto3` fallback added for `s3tablescatalog/*` catalogs: if PyAthena's `list_table_metadata` returns empty (S3 Tables catalogs are not visible via `ListDataCatalogs`, which PyAthena depends on), call `glue/athena boto3` directly using the connection's existing credentials. Errors in the fallback are reported and downgraded to an empty list rather than raised.
- `_get_upstream_location` short-circuits S3 Tables tables to an Iceberg URN at the top: every table in an `s3tablescatalog/` is Iceberg, the reported `s3://` location is internal storage, and the Iceberg URN matches what the Iceberg connector emits.
- `AthenaConfig` derives `platform_instance` from the catalog name when it starts with `s3tablescatalog/` and the user has not set one explicitly. Avoids accidental URN collisions between two S3 Tables buckets ingested under the default platform_instance.

### Column-handling fixes (also affect non-S3-Tables setups)

- `HiveColumnToAvroConverter`: treat `list<T>` as a synonym for `array<T>`. Iceberg and some Athena/Glue clients emit `list<...>`; without this, the type falls through to the unknown-type path and the whole column collapses to `NullType`, which downstream presents as a missing or empty column name (especially visible for nested `struct<...,items:list<struct<...>>>`).
- `SqlAlchemyColumnToAvroConverter`: append the column name to the field path for `types.NullType` columns. Two unrelated columns that both end up as `NullType` previously produced identical `fieldPath` values (`[version=2.0].[type=null]`), causing one to overwrite the other in the emitted schema.
- `CustomAthenaRestDialect._get_column_type`: map Pandas nullable dtype names (`Int64Dtype`, `BooleanDtype`, …) back to SqlAlchemy types. They leak in from `great_expectations` profiling on Athena result sets; without the mapping they fall through to `NullType`.

### Tests

Unit tests cover: `ICEBERG` table type inclusion, the boto3 fallback being invoked only for `s3tablescatalog/` catalogs and silently returning `[]` on error, the `platform_instance` auto-derive rules, the S3 Tables → Iceberg URN routing (with and without `iceberg_platform_instance`), `list<>` parsed identically to `array<>` (including the `struct<...,list<struct<...>>>` case that previously collapsed), `NullType` columns getting unique field paths, and Pandas dtype mapping.

### IAM

Adds `athena:ListDataCatalogs` to the documented IAM policy. The catalog-type lookup used for upstream URN routing (Glue vs Iceberg) needs it.


ligfx:
Possible dupe of https://github.com/datahub-project/datahub/pull/16842 ?

acrylJonny:
> Possible dupe of #16842 ?

I'll do a quick audit in the next 24 hours to check if they are entirely the same in terms of the functionality added and will close this if it is the case. If there are things to add to #16842 then want me to add a PR to your branch?

ligfx:
> I'll do a quick audit in the next 24 hours to check if they are entirely the same in terms of the functionality added and will close this if it is the case. If there are things to add to #16842 then want me to add a PR to your branch?

Not my PR, good question for @alokr-dhub

acrylJonny:
> > I'll do a quick audit in the next 24 hours to check if they are entirely the same in terms of the functionality added and will close this if it is the case. If there are things to add to #16842 then want me to add a PR to your branch?
> 
> Not my PR, good question for @alokr-dhub
 
They do not overlap - below is a comparison table of what has been implemented in each:

Feature | athena-s3Tables | feat/add-upstream-lineage-for-iceberg
-- | -- | --
S3 Tables catalog discovery (boto3 fallback) | Yes | No
s3tablescatalog/ → Iceberg URN lineage w/ platform_instance_map | Yes | No
Auto-derive platform_instance from catalog name | Yes | No
list<> type as alias for array<> in Hive schema converter | Yes | No
Pandas dtype handling in Athena dialect | Yes | No
NullType fieldPath fix in SQLAlchemy converter | Yes | No
Glue catalog detection (via list_data_catalogs API) | No | Yes
Glue catalog → Glue URN upstream lineage | No | Yes

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
```diff
@@ -108,6 +109,56 @@ class CustomAthenaRestDialect(AthenaRestDialect):
...
+            boto3_client = raw_connection.connection.cursor().connection.client
+            paginator = boto3_client.get_paginator("list_table_metadata")
+            return [
+                t["Name"]
+                for page in paginator.paginate(
+                    CatalogName=catalog, DatabaseName=schema_name
+                )
+                for t in page.get("TableMetadataList", [])
+            ]
+        except Exception as e:
```
treff7es: 🔴 **Required:** `except Exception` swallows every distinct failure category (missing `athena:ListTableMetadata` IAM permission, expired credentials, PyAthena internal API changes, network errors, etc.) and returns an empty list — indistinguishable to the user from a legitimately empty catalog.

Two issues:
1. `logger.warning` bypasses `report.warning`, so this failure is invisible in the DataHub ingestion UI run report. The analogous failure in `_catalog_type` (line ~526) correctly uses `self.report.warning(message=..., exc=e)` — follow the same pattern. Since `_list_tables_via_boto3` is on the dialect and has no `report`, the simplest fix is to let exceptions propagate from here and catch them in `get_table_names` (where the caller has access to `AthenaSource.report`).
2. At minimum, pass `exc_info=True` so the stack trace appears in logs instead of just `str(e)`.

Also: the access chain `raw_connection.connection.cursor().connection.client` opens a cursor that is never closed. `pyathena.connection.Connection` exposes `.client` directly — `raw_connection.client` achieves the same result without the intermediate cursor allocation.

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
```diff
@@ -478,6 +559,16 @@ def _get_upstream_location(
...
     ) -> Optional[str]:
+        # Every table in an S3 Tables catalog is Iceberg. The `s3://` location reported
+        # by Athena is internal storage, not a meaningful upstream — emit an Iceberg URN
+        # directly so lineage stitches to whatever Iceberg-side ingestion is configured.
+        if self.config.catalog_name.startswith("s3tablescatalog/"):
+            return make_dataset_urn_with_platform_instance(
+                platform="iceberg",
+                name=f"{schema}.{table}",
+                platform_instance=self.config.iceberg_platform_instance,
+                env=self.config.env,
```
treff7es: 🔴 **Required:** Silent lineage breakage.

`_derive_platform_instance_from_s3_catalog` correctly sets `self.platform_instance = self.catalog_name`, so the emitted dataset URN includes the catalog as `platform_instance`. But this lineage URN is built with `platform_instance=self.config.iceberg_platform_instance`, which defaults to `None`. The two sides will never stitch in DataHub's lineage graph — silently, with no warning.

Options:
- Auto-default `iceberg_platform_instance` to the catalog name in the validator (alongside `platform_instance`), OR
- Add `if self.config.iceberg_platform_instance is None: self.report.warning(message="iceberg_platform_instance not set for S3 Tables catalog; upstream lineage may not stitch", context=f"catalog={self.config.catalog_name}")` before this return.

metadata-ingestion/src/datahub/utilities/sqlalchemy_type_converter.py
```diff
@@ -227,11 +227,14 @@ def get_schema_fields_for_sqlalchemy_column(
...
 
     # for all non-nested data types an additional modification of the `fieldPath` property is required
+    # NullType (fallback for unrecognised types) is included so that each column gets a unique path;
+    # without it all unrecognised columns would share "[version=2.0].[type=null]".
     if type(column_type) in (
         *SqlAlchemyColumnToAvroConverter.PRIMITIVE_SQL_ALCHEMY_TYPE_TO_AVRO_TYPE.keys(),
         types.TIMESTAMP,
         types.DATE,
         types.DECIMAL,
+        types.NullType,
```
treff7es, line 237: 🔴 **Required:** The NullType fieldPath uniqueness fix has no regression test. Without one, this change could be silently reverted and DataHub would go back to emitting duplicate-keyed schema fields for tables with multiple unrecognised column types.

Please add a test to `tests/unit/utilities/test_sqlalchemy_type_converter.py` like:
```python
def test_null_type_columns_get_unique_field_paths():
    fields_a = get_schema_fields_for_sqlalchemy_column("col_a", types.NullType(), ...)
    fields_b = get_schema_fields_for_sqlalchemy_column("col_b", types.NullType(), ...)
    assert fields_a[0].fieldPath != fields_b[0].fieldPath
```

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
```diff
@@ -108,6 +109,56 @@ class CustomAthenaRestDialect(AthenaRestDialect):
...
+                table_names = self._list_tables_via_boto3(
+                    raw_connection=raw_connection,
+                    catalog=catalog,
+                    schema_name=schema_name,
+                )
+
+        return table_names
+
+    @typing.no_type_check
+    def _list_tables_via_boto3(
```
treff7es, line 174: 🟡 **Optional:** The string `"s3tablescatalog/"` drives behaviour in three places: here, in `_derive_platform_instance_from_s3_catalog` (line 385), and in `_get_upstream_location` (line 565). A class-level constant on `AthenaConfig` (or a module-level constant with a small `_is_s3_tables_catalog(name)` helper) would centralise future changes.

metadata-ingestion/tests/unit/test_athena_source.py
```diff
@@ -1673,3 +1673,219 @@ def test_sanitize_identifier_error_handling_in_generate_partition_profiler_query
...
+
+def _mock_table(name: str, table_type: str) -> mock.MagicMock:
+    t = mock.MagicMock()
+    t.name = name
+    t.table_type = table_type
+    return t
+
+
+def _base_athena_kwargs() -> dict:
+    return {
```
treff7es, line 1687: 🟡 **Optional:** `StringDtype` and `ObjectDtype` are real pandas nullable dtypes that map to `types.String` in the new branch, but neither is exercised here. Common for Athena VARCHAR/STRING columns — worth adding to this test.

treff7es (changes requested):
Good feature addition — the S3 Tables support is well thought out and the inline comments are excellent. Three required changes before merge:

1. **Silent lineage breakage** (`iceberg_platform_instance=None` produces mismatched URNs between datasets and their upstream lineage, so nothing stitches)
2. **Silent failures** (bare `except Exception` in `_list_tables_via_boto3` hides credentials/permission errors as an empty table list; also `logger.warning` bypasses `report.warning` so failures don't surface in the DataHub UI)
3. **Missing regression test** for the NullType fieldPath uniqueness fix in `sqlalchemy_type_converter.py`

Optional improvements noted inline.

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
acrylJonny: Fixed in d6ff0ab07b. Dropped the bare `except` from `_list_tables_via_boto3`, switched to `raw_connection.client` (no cursor leak), and gave the dialect an optional back-reference to the source report so failures surface via `report.warning` in the UI. `logger.exception` captures the trace.

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
acrylJonny: Fixed in d6ff0ab07b. The validator (now `_derive_defaults_from_s3_catalog`) auto-defaults `iceberg_platform_instance` to the catalog name alongside `platform_instance` for S3 Tables catalogs, so the URNs stitch by default. Explicit overrides still win.

metadata-ingestion/src/datahub/utilities/sqlalchemy_type_converter.py
acrylJonny: Added `test_null_type_columns_get_unique_field_paths` in `tests/unit/utilities/test_sqlalchemy_type_converter.py` (d6ff0ab07b).

metadata-ingestion/src/datahub/ingestion/source/sql/athena.py
acrylJonny: Done in d6ff0ab07b — added module-level `_S3_TABLES_CATALOG_PREFIX` + `_is_s3_tables_catalog(name)` helper. All three call sites use it.

metadata-ingestion/tests/unit/test_athena_source.py
acrylJonny: Added to `test_get_column_type_pandas_nullable_dtypes` in d6ff0ab07b.

treff7es (approved):

PR #17002, opened 2026-04-13
fix(ui): Make updating data products on asset profile page consistent with other metadata

chriscollins3456:
We were experiencing some issues and confusion where a user had permission to edit the data product of one sibling but not the other, but no matter what "page" they were on, when they were on a combined siblings page, they got an error.

This is because on siblings pages wee were always trying to add a sibling urns when adding to a data product. Now, I make adding data products consistent with other ways we add tags, terms, owners, domains etc. - when you are on a combined siblings page we get the urn of the "primary" sibling (using `useMutationUrn`) so the user needs permission on that entity. And then if we are on an individual sibling page we use that urn as expected (in the past we were still adding it to all siblings mistakenly).

### DIFF
--- a/datahub-web-react/src/app/entityV2/shared/containers/profile/sidebar/DataProduct/DataProductSection.tsx
+++ b/datahub-web-react/src/app/entityV2/shared/containers/profile/sidebar/DataProduct/DataProductSection.tsx
@@ -1,15 +1,16 @@
 import { toast } from '@components';
-import { PencilSimple } from '@phosphor-icons/react/dist/csr/PencilSimple';
 import { Plus } from '@phosphor-icons/react/dist/csr/Plus';
 import React, { useEffect, useState } from 'react';
 import styled from 'styled-components';
 
-import { useEntityData } from '@app/entity/shared/EntityContext';
+import { useDataNotCombinedWithSiblings, useEntityData, useMutationUrn } from '@app/entity/shared/EntityContext';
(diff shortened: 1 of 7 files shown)


datahub-web-react/src/app/entityV2/shared/containers/profile/sidebar/DataProduct/DataProductSection.tsx
```diff
@@ -38,33 +39,71 @@ export default function DataProductSection({ readOnly }: Props) {
...
+    function getAssociatedUrns(dataProductUrn: string): string[] {
+        if (isSeparateSiblingsMode) {
+            return [urn];
+        }
+
+        const assocUrns: string[] = [];
+
+        // Check the root entity using the non-combined query data so we don't pick up
+        // relationships that were contributed only by a sibling.
+        const entityKey = dataNotCombined ? Object.keys(dataNotCombined)[0] : null;
```
v-tarasevich-blitz-brain: nit: can we extract getting of root entity to the separated function or hook? (as this peace of code looks a little low-leveled)

v-tarasevich-blitz-brain (approved):
Looks good to me. One suggestion

datahub-web-react/src/app/entityV2/shared/containers/profile/sidebar/DataProduct/DataProductSection.tsx
chriscollins3456: yup i see that for sure

PR #15261, opened 2025-11-10
refactor(pydantic): remove Pydantic v1 legacy code and migrate fully to v2

sgomezvillamor:
## Context

This PR removes all Pydantic v1 compatibility code and completes the migration to Pydantic v2. This is now safe because:

- **All Python projects in the repo have pydantic>=2.0 as a dependency**
  - `metadata-ingestion`: `pydantic>=2.4.0,<3.0.0`
  - `datahub-actions`: `pydantic>=2.0.0,<3.0.0`
  - `smoke-test`: uses metadata-ingestion (Pydantic v2)
  - **Note**: The codebase maintains compatibility with Pydantic 2.0+ (some CI environments use older constraint files with Pydantic 2.0-2.3)
- **The codebase no longer needs to support Pydantic v1 users**
- **Pydantic v2 has been stable since June 2023** (over a year and a half ago)
- **This simplifies the codebase** by removing 254 lines of compatibility/branching code

### Deleted compatibility layer files:
- `metadata-ingestion/src/datahub/configuration/pydantic_migration_helpers.py` (provided v1_* aliases)
- `metadata-ingestion/src/datahub/pydantic/compat.py` (version detection utilities)

### Migrated to native Pydantic v2 APIs (38 files):
- `v1_ConfigModel` → `BaseModel`
- `v1_Field` → `Field`
- `v1_validator` → `@field_validator`
- `v1_root_validator` → `@model_validator`
- `Extra.forbid` → `model_config = {"extra": "forbid"}`
- `__fields__` → `model_fields`
- `parse_obj()` → `model_validate()`
- `_init_private_attributes` → `model_post_init`

### Removed all version checks:
- Deleted conditional code for `PYDANTIC_VERSION`, `PYDANTIC_VERSION_2`, `PYDANTIC_SUPPORTS_CALLABLE_DISCRIMINATOR`
- Kept only Pydantic v2 code paths

### Updated discriminated unions:
- Leverages Pydantic v2's smart union matching for assertion types and search filters
- Avoids `Discriminator`/`Tag` (Pydantic 2.4+) to maintain compatibility with Pydantic 2.0-2.3 used in some Airflow constraint files
- Added comments explaining where smart union matching is used

### Updated configurations:
- Removed `pydantic.v1.mypy` from mypy plugins
- Removed Pydantic v1 deprecation warnings from pytest config

## Breaking Changes

**None for users** - all projects already require Pydantic v2. This only affects:
[...]

PR #17007, opened 2026-04-13
fix(ingestion): fix parse_obj_allow_extras failing on models with deep inheritance

jayacryl:
## Summary

Fixes `SourceConnectionErrorException` in datahub-executor when parsing `BigQueryV2Config` (and similar deeply-inherited config models) from ingestion recipes via `parse_obj_allow_extras`.

The old implementation used `unittest.mock.patch.dict` + `model_rebuild(force=True)` to temporarily allow extra fields. This had two problems:
1. **Race condition** — `model_rebuild` mutates shared class state, so concurrent Celery tasks could corrupt each other's validation
2. **MRO cascade** — `model_rebuild(force=True)` cascades to parent classes, which retain `extra="forbid"` and reject fields from sibling parents

Replaced with a dynamic subclass approach: `type(cls.__name__, (cls,), {"model_config": ConfigDict(extra="ignore")})`. Pydantic's metaclass compiles a fresh validator with `extra="ignore"` baked in, merging the full MRO into one flat schema. This:
- Has no shared state mutation (inherently thread-safe)
- Handles all alias types natively (`AliasChoices`, `AliasPath`, string aliases) — the intermediate key-stripping approach silently dropped `AliasChoices` fields
- Requires no `unittest.mock` import

Cherry-picked from acryldata/datahub-fork#9005 and then improved.

## Test plan
- [x] New unit tests for multi-inheritance, diamond inheritance, aliases, `AliasChoices`, no class corruption
- [x] Existing `test_extras_allowed`, `test_extras_not_allowed`, `test_state_provider_configs` pass
- [x] Empirically verified with `BigQueryV2Config` (11 parent classes, 82 fields)
- [x] Ruff lint passes

### DIFF
--- a/metadata-ingestion/src/datahub/configuration/common.py
+++ b/metadata-ingestion/src/datahub/configuration/common.py
@@ -1,7 +1,6 @@
 import contextvars
 import dataclasses
 import re
-import unittest.mock
 from abc import ABC, abstractmethod
 from enum import auto
 from functools import cached_property as functools_cached_property
@@ -277,22 +276,23 @@ class ConfigModel(BaseModel):
 
     @classmethod
     def parse_obj_allow_extras(cls, obj: Any) -> Self:
-        """Parse an object while allowing extra fields.
+        """Parse an object, silently ignoring any fields not defined on the model.
 
-        This method temporarily modifies the model's configuration to allow extra fields.
-
-        TODO: Do we really need to support this behaviour? Consider removing this method in future.
+        This is useful when parsing a config dict that may contain fields for a
+        broader config class (e.g. a full ingestion recipe) into a narrower one.
         """
-        try:
-            with unittest.mock.patch.dict(
-                cls.model_config,  # type: ignore
-                {"extra": "allow"},
-                clear=False,
-            ):
-                cls.model_rebuild(force=True)  # type: ignore
-                return cls.model_validate(obj)
-        finally:
-            cls.model_rebuild(force=True)  # type: ignore
+        if isinstance(obj, dict):
+            known_keys: set = set()
+            for field_name, field_info in cls.model_fields.items():
+                known_keys.add(field_name)
+                if field_info.alias and isinstance(field_info.alias, str):
+                    known_keys.add(field_info.alias)
+                if field_info.validation_alias and isinstance(
+                    field_info.validation_alias, str
+                ):
+                    known_keys.add(field_info.validation_alias)
+            obj = {k: v for k, v in obj.items() if k in known_keys}
+        return cls.model_validate(obj)
 
 
 class PermissiveConfigModel(ConfigModel):
--- a/metadata-ingestion/src/datahub/configuration/common.py
+++ b/metadata-ingestion/src/datahub/configuration/common.py
@@ -281,18 +281,12 @@ class ConfigModel(BaseModel):
         This is useful when parsing a config dict that may contain fields for a
         broader config class (e.g. a full ingestion recipe) into a narrower one.
         """
-        if isinstance(obj, dict):
-            known_keys: set = set()
-            for field_name, field_info in cls.model_fields.items():
-                known_keys.add(field_name)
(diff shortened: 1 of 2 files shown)


treff7es (dismissed):

sgomezvillamor (changes requested):
Oh man, every time we patch, it comes back to bite us. 😅 Great root cause analysis — moving away from `model_rebuild` and patching is the right call. 

## One correctness concern before this lands

The key-filtering loop silently drops `AliasChoices` / `AliasPath` fields

```
if field_info.validation_alias and isinstance(field_info.validation_alias, str):
    known_keys.add(field_info.validation_alias)
```

In Pydantic v2, `validation_alias` can be `AliasChoices` or `AliasPath`, not just `str`. When it is, the guard silently skips it — the user-supplied key gets stripped before `model_validate` sees it, and the field silently falls back to its default. This is already live in the codebase: `dataplex_config.py:30` uses `AliasChoices("pattern", "dataset_pattern")`, so any caller passing `dataset_pattern` would have it silently discarded.

## Simpler alternative: dynamic subclass

```
@classmethod
def parse_obj_allow_extras(cls, obj: Any) -> Self:
    """Parse an object, silently ignoring any fields not defined on the model."""
    relaxed_cls = type(cls.__name__, (cls,), {"model_config": ConfigDict(extra="ignore")})
    return relaxed_cls.model_validate(obj)
```

`type(...)` triggers Pydantic's metaclass, which compiles a fresh `__pydantic_validator__` for `relaxed_cls` with `extra="ignore"` baked in. Pydantic merges the full MRO into one flat schema during that compilation, so all alias types are handled correctly by Pydantic itself — no manual key enumeration needed. No shared state mutation, no `model_rebuild`, no `unittest.mock` import.

The only caveat: `type(result)` is a dynamic subclass of `cls`, not `cls` itself — but `isinstance` checks and `model_dump()` work identically, and all existing call sites are unaffected.

treff7es (commented):
See @sgomezvillamor 's comment. I prefer that one as well if it works

jayacryl:
## Initial analysis & intermediate approach (key-stripping)

The first commit used a key-stripping approach: enumerate `cls.model_fields` + aliases, filter the input dict, then call `model_validate()`. This worked for most cases but had a subtle bug: fields using non-string validation aliases (`AliasChoices`, `AliasPath`) were silently dropped because the `isinstance(alias, str)` check returned `False` for those types.

### Root Cause Analysis

The failure chain has two distinct errors, and the first one causes the second.

**Error 1: Pydantic `extra_forbidden` validation failure**

At `bigquery.py:24`, the executor calls:
```python
bigquery_config = BigQueryV2Config.parse_obj_allow_extras(source_config)
```

The old implementation used `unittest.mock.patch.dict` to temporarily set `extra="allow"` on the model config, then called `model_rebuild(force=True)`. In pydantic v2, `model_rebuild` cascades to parent classes in the MRO, but those parents (e.g. `GcsDatasetLineageProviderConfigBase`) retain their own `extra="forbid"`, causing `ValidationError: extra_forbidden` for fields inherited from sibling parent classes.

**Error 2: `SourceConnectionErrorException` (the symptom users see)**

Because the `BigQueryV2Config` parsing throws, `extract_connection_from_bigquery_recipe` catches the exception, logs it, and returns `None`. The caller then raises `SourceConnectionErrorException`.

**Why intermittent:** The `datahub-executor` runs as a Celery worker processing tasks concurrently. The old code's `patch.dict` + `model_rebuild` operates on shared class-level state with zero synchronization — if two tasks call `parse_obj_allow_extras()` concurrently, one task's `finally` cleanup can reset the model config while the other is mid-validation.

**Timeline:** The bug was introduced on November 12, 2025 in PR #15261 (pydantic v2 migration). Before that, pydantic v1 used a simpler `patch.object(cls.Config, "extra", Extra.allow)` which didn't have cascade issues.

jayacryl:
@cursor fix the ingestion lint CI issues

sgomezvillamor (approved):

PR #17009, opened 2026-04-14
feat(search): add V3 entity type aggregation and filter compatibility layer

loustler:
## Summary

DataHub's V3 entity index uses a shared multi-entity architecture where all entities are stored in a single index with `_entityType` field discrimination, unlike V2 which has separate per-entity indices (e.g., `datasetindex_v2`, `chartindex_v2`). When search queries span both V2 and V3 indices, several compatibility issues arise:

1. **Entity type aggregation**: V2 uses `_index` field to determine entity type; V3 documents share one index, so `_index` returns the shared index name instead of the entity type.
2. **Filter transformation**: Entity type filters must translate to `_index = <entityName>index_v2` for V2 and `_entityType = <camelCaseName>` for V3.
3. **Result deduplication**: The same entity can appear in both V2 and V3 hits.
4. **Aggregation deduplication**: V2 and V3 return separate buckets for the same entity type that must be merged.
5. **Keyword field compatibility**: V3 alias fields cannot have `.keyword` subfields, so filter/aggregation queries must use the base field name.

This PR builds the compatibility layer to handle all of these issues, preparing the search pipeline for dual V2/V3 querying.

### Existing Behavior — Modified

**Entity type aggregation** previously used `.field("_index")` for the entity type facet, which returns index names like `datasetindex_v2`. This works for V2 but returns the shared V3 index name (e.g., `primaryindex_v3`) for V3 documents, causing all V3 entities to appear as a single "OTHER" bucket.

→ Replaced with a **Painless script** that returns `_entityType` for V3 documents (which have this field) and falls back to `_index` for V2 documents (which don't).

**Filter transformation** previously converted `_entityType = DATASET` to `_index = datasetindex_v2` only.

→ Now generates **dual filter variants**: a V2 branch (`_index = datasetindex_v2`) and a V3 branch (`_entityType = dataset`), combined as OR alternatives. Uses `EntityRegistry` for canonical camelCase lookup (e.g., `CORP_USER` → `corpUser`, not `corpuser`).

**Aggregation normalization** previously passed through the entity name token as-is when `getEntityName()` returned empty.

→ Now applies `.toLowerCase()` as fallback, normalizing V3 camelCase values (`glossaryNode`) to match V2 convention (`glossarynode`).

### New Behavior — Added

[...]

PR #17013, opened 2026-04-14
feat(search): enable V3 entity index as dual search target alongside V2

loustler:
## Summary

DataHub's V3 entity index is currently write-only — data is indexed into V3 during ingestion and metadata changes, but search queries only target V2 per-entity indices. This means V3-only features (unified scoring, tiered search fields, custom analyzers) are not available to end users.

This PR enables the V3 shared entity index as a search target **alongside** existing V2 indices, making V3 data discoverable while maintaining full backward compatibility with V2. Every search query now targets both `<entity>index_v2` and `*index_v3` patterns simultaneously.

> **Depends on**: PR #17009  (V3 entity type aggregation and filter compatibility layer) which handles result merging, deduplication, and aggregation normalization.

### Existing Behavior — Modified

**Search queries** previously targeted only V2 per-entity indices (e.g., `datasetindex_v2`, `chartindex_v2`).

→ Now target both V2 and V3 indices via `Stream.concat(v2Patterns, indexConvention.getV3EntityIndexPatterns().stream())`. This applies to all five search paths:

| Method | Purpose |
|--------|---------|
| `buildSearchRequest()` | Main search (search bar, API) |
| `filter()` | Filter-based entity listing (admin pages, glossary) |
| `buildAutocompleteRequest()` | Typeahead autocomplete suggestions |
| `buildAggregateByValue()` | Facet aggregation for sidebar filters |
| `buildScrollRequest()` | Paginated scroll through large result sets |

### New Behavior — Added

**Entity type filter** (`applyEntityTypeFilter()`): Since V3 is a shared multi-entity index, queries must be wrapped with an entity type filter to prevent cross-entity contamination. The filter uses a `bool` query with two `should` branches:
- `must_not exists _entityType` — lets V2 documents (which lack this field) pass through
- `terms _entityType [entityNames]` — restricts V3 documents to requested types

Entity names **must** use camelCase from `EntitySpec.getName()` (e.g., `glossaryNode`, `corpUser`) to match V3's stored `_entityType` values. Lowercase names produce 0 matches on V3.

**V3 keyword subfield** (`MultiEntityMappingsBuilder`): Added `.keyword` subfield for keyword-type fields in V3 mappings so aggregation queries like `owners.keyword` work consistently across V2 and V3 indices during the transition period.

**Filesystem config loading** (`BaseConfigurationLoader`): Added filesystem path fallback before classpath resource lookup. This enables loading external analyzer configurations (e.g., nori, kuromoji) from Kubernetes ConfigMap mounts at paths like `/etc/datahub/analyzer-config.yaml` without requiring them to be on the Java classpath.

### Key Files Modified

| File | Change |
|------|--------|
| `metadata-io/.../ESSearchDAO.java` | V3 index pattern concatenation in all 5 search methods + `applyEntityTypeFilter()` |
| `metadata-io/.../MultiEntityMappingsBuilder.java` | `.keyword` subfield for V3 keyword fields |
| `metadata-io/.../BaseConfigurationLoader.java` | Filesystem-first config loading fallback |
| `metadata-io/.../BaseConfigurationLoaderTest.java` | Test for filesystem path loading |

### Configuration

No new configuration required. V3 index patterns are derived from the existing `IndexConvention.getV3EntityIndexPatterns()` which is controlled by the `ELASTICSEARCH_ENTITY_INDEX_V3_ENABLED` environment variable. If V3 is not enabled, the V3 pattern matches no indices and queries behave identically to before.

### Migration Notes

- **Backward compatible**: Deployments without V3 indices see no behavior change (wildcard pattern matches nothing).
- **No reindexing required**: V3 indices are populated by existing write paths; this PR only adds them to read paths.
- **Gradual rollout**: V3 can be enabled/disabled without data loss since V2 indices remain the primary source of truth.

## Checklist

- [x] Links to related issue: N/A
- [x] Tests added/updated: `BaseConfigurationLoaderTest` for filesystem path loading


alexjst:
Thanks @loustler, solid work, you clearly got deep into the V3 internals.

I'm going to close it for direction, not quality. V3's main goal is cutting shards/cost, and that only lands once V2 is gone, so we're planning to migrate V2→V3 and drop V2 rather than run both indices and dedup. The dual-read approach here wouldn't fit that.

Appreciate the effort, and sorry it sat so long.

loustler:
@alexjst 
Thanks for the kind response. I completely understand the decision and welcome the transition to V3. When are the V3 migration and V2 drop planned? I'm really looking forward to it!

PR #17014, opened 2026-04-14
fix(agent): return message instead of raising ItemNotFoundError in get_lineage_paths_between

### LINKED ISSUE #16882: get_lineage_between return the ItemNotFoundError, Should be a message instead of error from adk module
**Describe the bug**
get_lineage_between return the ItemNotFoundError, for ADK module response it should rather be a text response

**To Reproduce**
LLM agent call for get_lineage_between between any two assets which do not have a lineage among those.

It Raises a ItemNotFoundError

ItemNotFoundError(datahub.errors.ItemNotFoundError: No lineage path found between urn:li:dataset:(urn:li:dataPlatform:bigquery,project.dataset.table_1,PROD) and urn:li:dataset:(urn:li:dataPlatform:bigquery,wf-project.troas.table_2,PROD) in either upstream or downstream direction)

**Expected behavior**
Should have a wrapper function on top of core datahub function translating the ItemNotFoundError into a simple message. 

 v1.4.0.6 and v1.5.0.2

psaikaushik:
## Summary

When no lineage path exists between two assets, `get_lineage_paths_between` raises `ItemNotFoundError`. In an agent/ADK context this surfaces as an unhandled exception instead of a useful response.

Closes #16882

### `datahub-agent-context/src/datahub_agent_context/mcp_tools/lineage.py`
- `get_lineage_paths_between()` now catches `ItemNotFoundError` and returns a friendly message dict instead of raising
- Added `_build_no_path_found_message()` helper that returns a consistent response with `message`, `source`, `target`, `pathCount=0`, and empty `paths`
- Both code paths are covered: auto-discover (tries downstream then upstream) and explicit direction
- Updated docstring to document the new return shape for the no-path case
- Removed `ItemNotFoundError` from the `Raises` section of the docstring

### `datahub-agent-context/tests/unit/mcp_tools/test_lineage.py`
- Replaced `test_path_not_found_raises_error` (which expected `ItemNotFoundError`) with `test_path_not_found_returns_message` that checks the message dict
- Added `test_path_not_found_auto_discover_returns_message` for the auto-discover code path

## Before vs After

**Before (raises, agent sees unhandled error):**
```
ItemNotFoundError: No lineage path found between urn:li:dataset:A and urn:li:dataset:B
in either upstream or downstream direction
```

**After (returns dict, agent gets a usable response):**
```json
{
  "message": "No lineage path found between urn:li:dataset:A and urn:li:dataset:B in either upstream or downstream direction.",
  "source": {"urn": "urn:li:dataset:A"},
  "target": {"urn": "urn:li:dataset:B"},
  "pathCount": 0,
  "paths": []
}
```

## Notes
- Internal helpers (`_find_upstream_lineage_path`, `_find_lineage_path`) still raise `ItemNotFoundError` for control flow. Only the public-facing `get_lineage_paths_between` catches and converts them.
- No changes to `get_lineage()` which has different behavior and is not affected by this issue.

### DIFF
--- a/datahub-agent-context/src/datahub_agent_context/mcp_tools/lineage.py
+++ b/datahub-agent-context/src/datahub_agent_context/mcp_tools/lineage.py
@@ -564,6 +564,46 @@ def _find_lineage_path(
         )
 
 
+def _build_no_path_found_message(
+    source_urn: str,
+    target_urn: str,
+    source_column: Optional[str] = None,
+    target_column: Optional[str] = None,
+) -> dict:
+    """Build a friendly response when no lineage path is found.
+
+    Returns a dict instead of raising an exception so that agent/ADK
+    consumers get a usable message rather than an unhandled error.
+
+    Args:
+        source_urn: Source entity URN
+        target_urn: Target entity URN
+        source_column: Optional source column name
+        target_column: Optional target column name
+
+    Returns:
+        Dictionary with a message and metadata about the failed lookup
+    """
+    source_display = source_urn + (f".{source_column}" if source_column else "")
+    target_display = target_urn + (f".{target_column}" if target_column else "")
+    return {
+        "message": (
+            f"No lineage path found between {source_display} and {target_display} "
+            f"in either upstream or downstream direction."
+        ),
+        "source": {
+            "urn": source_urn,
+            **({"column": source_column} if source_column else {}),
+        },
+        "target": {
+            "urn": target_urn,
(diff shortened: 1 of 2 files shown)


sgomezvillamor:
ICYMI https://github.com/datahub-project/datahub/pull/17073 is about to be merged, addressing the issue with a global mechanism rather than patching tools individually. Jump in if you think anything's missing

Issue #16839, opened 2026-03-30
Flaky: build-and-test (frontend, UTC) fails intermittently on MfeConfigController


shirshanka:
## Summary

The `build-and-test (frontend, UTC)` CI check is intermittently failing while `build-and-test (frontend, America/New_York)` passes on the same commit. This has been observed across multiple re-runs on PR #16718.

## Error

```
java.io.IOException: Network error
controllers.MfeConfigController - MfeConfigController error! Failed to read YAML config file at: mfe.config.local.yaml
```

## Observations

- The `America/New_York` variant of the same job passes consistently
- The `UTC` variant has failed on every run in PR #16718 (4+ attempts)
- The error appears to be a race condition or environment issue in the frontend test suite, not a code problem
- Previous runs also showed npm registry 502 errors, but even after npm recovered, the UTC variant still fails with the MfeConfigController error

## Reproduction

Any PR that triggers the `build-and-test (frontend, UTC)` workflow may hit this. See PR #16718 for examples.

## Impact

This blocks the `Mergeable` check, preventing PRs from merging even when all code-related checks pass.

PR #17015, opened 2026-04-14
fix(openlineage): accept custom producer URIs instead of returning 500

psaikaushik:
## Summary

The OpenLineage converter rejects events with custom producer URIs that don't match the hardcoded OpenLineage, Airflow, or Trino patterns, returning a 500 error. According to the OpenLineage spec, the producer field is a URI that can be any value.

Closes #16961

### `OpenLineageToDataHub.java`
- Added `extractOrchestratorFromProducerUri()` helper that extracts the last non-empty path segment from a producer URI to use as the orchestrator name
- In `getOrchestrator()`, added an `else` branch for unknown producers that calls the new helper instead of falling through to the `RuntimeException`
- Existing behavior for OpenLineage, Airflow, and Trino producers is unchanged

### `OpenLineageOrchestratorTest.java` (new)
- 7 tests covering:
  - Custom producer URIs (the bug scenario from the issue)
  - Simple paths, trailing slashes
  - Known producers still work (OpenLineage, Airflow, Trino)
  - processingEngine takes precedence over producer

## Before vs After

**Before:**
```
POST /openapi/openlineage/api/v1/lineage
{"producer": "https://github.com/myorg/myproducer/blob/v1/client", ...}

=> 500 Internal Server Error (RuntimeException: Unable to determine orchestrator)
```

**After:**
```
POST /openapi/openlineage/api/v1/lineage
{"producer": "https://github.com/myorg/myproducer/blob/v1/client", ...}

=> 200 OK (orchestrator = "client", extracted from last path segment)
```

### DIFF
--- a/metadata-integration/java/openlineage-converter/src/main/java/io/datahubproject/openlineage/converter/OpenLineageToDataHub.java
+++ b/metadata-integration/java/openlineage-converter/src/main/java/io/datahubproject/openlineage/converter/OpenLineageToDataHub.java
@@ -1292,12 +1292,48 @@ public class OpenLineageToDataHub {
         orchestrator = "airflow";
       } else if (producer.startsWith("https://github.com/trinodb/trino/")) {
         orchestrator = "trino";
+      } else {
+        // Fallback for custom producers: extract the last non-empty path segment
+        // from the producer URI. Per the OpenLineage spec, the producer field is
+        // a URI that can be any value, so we should not reject unknown producers.
+        // See: https://github.com/datahub-project/datahub/issues/16961
+        orchestrator = extractOrchestratorFromProducerUri(producer);
       }
     }
     if (orchestrator == null) {
       throw new RuntimeException("Unable to determine orchestrator");
     }
     return orchestrator;
+  }
+
+  /**
+   * Extract an orchestrator name from a custom producer URI by using the last non-empty path
+   * segment. For example, "https://github.com/myorg/myproducer/blob/v1/client" returns "client".
+   *
+   * @param producer the producer URI string
+   * @return the last path segment, or null if no usable segment is found
+   */
+  private static String extractOrchestratorFromProducerUri(String producer) {
+    try {
+      URI uri = new URI(producer);
+      String path = uri.getPath();
+      if (path != null && !path.isEmpty()) {
+        // Remove trailing slashes and split
+        String trimmedPath = path.replaceAll("/+$", "");
+        if (!trimmedPath.isEmpty()) {
+          String[] segments = trimmedPath.split("/");
+          // Return the last non-empty segment
+          for (int i = segments.length - 1; i >= 0; i--) {
+            if (!segments[i].isEmpty()) {
+              return segments[i].toLowerCase();
+            }
+          }
+        }
+      }
+    } catch (URISyntaxException e) {
+      log.warn("Unable to parse producer URI '{}': {}", producer, e.getMessage());
+    }
+    return null;
   }
 
   public static SchemaFieldDataType.Type convertOlFieldTypeToDHFieldType(
--- a/metadata-integration/java/openlineage-converter/src/test/java/io/datahubproject/openlineage/OpenLineageOrchestratorTest.java
+++ b/metadata-integration/java/openlineage-converter/src/test/java/io/datahubproject/openlineage/OpenLineageOrchestratorTest.java
@@ -1,105 +1,105 @@
-package io.datahubproject.openlineage.converter;
+package io.datahubproject.openlineage;
 
 import static org.testng.Assert.assertEquals;
 
+import com.linkedin.common.FabricType;
+import com.linkedin.common.urn.DataFlowUrn;
 import io.datahubproject.openlineage.config.DatahubOpenlineageConfig;
+import io.datahubproject.openlineage.converter.OpenLineageToDataHub;
+import java.net.URI;
 import org.testng.annotations.Test;
 
+/**
+ * Tests for custom OpenLineage producer support.
+ *
+ * <p>Regression tests for https://github.com/datahub-project/datahub/issues/16961
+ */
 public class OpenLineageOrchestratorTest {
 
   private static DatahubOpenlineageConfig defaultConfig() {
-    return DatahubOpenlineageConfig.builder().build();
+    return DatahubOpenlineageConfig.builder().fabricType(FabricType.PROD).build();
   }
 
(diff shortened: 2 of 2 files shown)


psaikaushik:
Looks like the failed cypress batch tests are already flaky as per https://github.com/datahub-project/datahub/issues/16839 thanks! this PR should not be triggering Cypress tests. thanks!

cedricp24:
any updates on this? would be cool to get this issue fixed

metadata-integration/java/openlineage-converter/src/main/java/io/datahubproject/openlineage/converter/OpenLineageToDataHub.java
```diff
@@ -1292,6 +1292,12 @@ private static String getOrchestrator(
         orchestrator = "airflow";
       } else if (producer.startsWith("https://github.com/trinodb/trino/")) {
         orchestrator = "trino";
+      } else {
+        // Fallback for custom producers: extract the last non-empty path segment
+        // from the producer URI. Per the OpenLineage spec, the producer field is
+        // a URI that can be any value, so we should not reject unknown producers.
+        // See: https://github.com/datahub-project/datahub/issues/16961
+        orchestrator = extractOrchestratorFromProducerUri(producer);
```
treff7es, line 1300: `orchestrator` ends up inside `urn:li:dataFlow:(<orchestrator>,<flowId>,<cluster>)` and can never be changed afterwards, so a scraped path segment is risky in three specific ways:

- **It churns.** This repo's own fixture `ol_events/debezium_event.json` has `producer: https://github.com/debezium/debezium/v3.2.0.CR1` → orchestrator `v3.2.0.cr1`. The next release mints a new DataFlow + DataJob + dataPlatform and orphans the previous lineage, ownership and run history.
- **It can break the URN.** `URI.getPath()` percent-*decodes*, and `TupleKey.create` validates only null/empty — no delimiter check. So `.../a%2Cb` → `urn:li:dataFlow:(a,b,my_job,my_ns)`, which then fails to re-parse (`Invalid number of keys`). Worth noting `STRICT_URN_VALIDATION_ENABLED` defaults to `false` in `ValidationApiUtils`, so a non-docker deploy would store that.
- **It collides.** Both `testCustomProducerExtractsOrchestratorFromPath` and `testKnownOpenLineageProducerStillWorks` below expect `client`, so two unrelated tools with a same-named job end up on one URN and overwrite each other.

This also closes the gap that's still open on this branch: `https://example.com` and `urn:my:producer` have an empty/null `getPath()`, so they never reach the helper and still return 500 — the original symptom from #16961.

```suggestion
      } else {
        // Per the OpenLineage spec, producer is a free-form URI, so an unrecognized one must not
        // be rejected. But don't derive a name from it either: orchestrator is a DataFlowKey key
        // field, so it can never change once the DataFlow exists, and a producer URI usually
        // carries the emitter's version. Use the platform already seeded in data-platforms.yaml;
        // set the `orchestrator` config to record a specific name instead.
        // See: https://github.com/datahub-project/datahub/issues/16961
        orchestrator = "unknown";
```

treff7es (commented):
Good catch on the root cause — the 500 is real, and `datahub.openlineage.orchestrator` genuinely doesn't help since it's server-wide.

One thing I'd want changed before this lands: don't derive the name *from* the producer URI. `orchestrator` is a `DataFlowKey` key field, so it's baked into the DataFlow URN and can't be changed once the entity exists. Suggested a one-liner below — fall back to the `unknown` platform that's already seeded in `data-platforms.yaml`. That's also *more* complete than the current fallback, since `https://example.com` and `urn:my:producer` still 500 on this branch.

I ran a property suite (totality / stability / URN round-trip) against the branch: 24 of 37 cases fail. With the two suggestions below, 6. The remaining 6 are all pre-existing and fine as follow-ups — locale-sensitive `toLowerCase()`, the OpenLineage regex capturing an empty group on a trailing slash, and a blank `processing_engine.name`. Happy to send those separately.

metadata-integration/java/openlineage-converter/src/main/java/io/datahubproject/openlineage/converter/OpenLineageToDataHub.java
```diff
@@ -1300,6 +1306,36 @@ private static String getOrchestrator(
...
+            }
+          }
+        }
+      }
+    } catch (URISyntaxException e) {
+      log.warn("Unable to parse producer URI '{}': {}", producer, e.getMessage());
+    }
+    return null;
+  }
+
```
treff7es, line 1338: Unused after the change above.

For future reference, most of this was already unreachable: `split("/")` discards *all* trailing empty strings (`"/a/b//".split("/")` → `["", "a", "b"]`), so the `replaceAll`, the `isEmpty()` guard and the reverse loop can never do anything. And `new URI(producer)` re-parses a `URI` the caller already holds, so the `catch (URISyntaxException)` is unreachable from every in-tree call site. (Minor: `$` in Java matches before a final line terminator, so `"/+$"` would strip an *interior* slash — that'd need `\z`.)

```suggestion
```

metadata-integration/java/openlineage-converter/src/test/java/io/datahubproject/openlineage/OpenLineageOrchestratorTest.java
```diff
@@ -0,0 +1,106 @@
...
+  @Test
+  public void testCustomProducerExtractsOrchestratorFromPath() {
+    DataFlowUrn flowUrn =
+        OpenLineageToDataHub.getFlowUrn(
+            "my_namespace",
+            "my_job",
+            null,
+            URI.create("https://github.com/myorganization/myproducer/blob/v1-0-0/client"),
+            defaultConfig());
+    assertEquals(flowUrn.getOrchestratorEntity(), "client");
```
treff7es, line 32: These two both expect `client` — one via the new fallback, one via the existing OpenLineage regex — which pins the collision as the contract. `client` also isn't a platform; it's a directory in the OpenLineage repo.

Also worth knowing: 4 of the 7 tests here pass unchanged on `master` (`testKnownOpenLineageProducerStillWorks`, `testAirflow…`, `testTrino…`, `testProcessingEngineOverridesProducer`), and the airflow/override cases duplicate the sibling `OpenLineageConfigTest`. The three that would actually catch a regression are stability across producer versions, totality (no 500), and a URN round-trip — none are covered yet. Glad to send that suite.

PR #17016, opened 2026-04-14
feat(secret): add AWS Secrets Manager and GCP Secret Manager secret stores

rajatoss:
## Summary

Add native `SecretStore` implementations for AWS Secrets Manager and GCP Secret Manager, enabling the Remote Executor to fetch credentials dynamically at ingestion/assertion time via IAM-based pod authentication (IRSA on EKS, Workload Identity on GKE).

### What's included

- **`AwsSecretsManagerStore`** — fetches secrets from AWS Secrets Manager using the `BatchGetSecretValue` API (up to 20 secrets per call)
- **`GcpSecretManagerStore`** — fetches secrets from GCP Secret Manager with retry handling for gRPC cold-channel issues
- **Configurable prefix** — secrets are namespaced with a prefix (default `datahub-`). Recipe `${SNOWFLAKE_TOKEN}` fetches `datahub-SNOWFLAKE_TOKEN` from the provider. Prefix is validated against `[a-zA-Z0-9_-]` to ensure compatibility with both AWS and GCP naming rules.
- **Thread-safe TTL cache** — uses `cachetools.TTLCache` with `threading.Lock` to avoid redundant API calls across repeated assertion/monitor runs. TTL is configurable.
- **`SecretStore` base class** now properly inherits from `ABC` for abstract method enforcement
- **Optional extras** — dependencies are not pulled by default. Install via `pip install acryl-datahub[aws-secret-manager]` or `pip install acryl-datahub[gcp-secret-manager]`
- **Shared prefix validation** — extracted to `secret_utils.py` to avoid duplication

### Design decisions

- **Prefix validated for both providers** — GCP secret IDs only allow `[a-zA-Z0-9_-]` (no slashes), so the prefix uses hyphens (e.g., `datahub-`) 
- **AWS batch API** — `BatchGetSecretValue` fetches up to 20 secrets in one call. For >20, requests are chunked automatically.
- **GCP sequential fetch** — GCP has no batch API. Sequential calls are efficient after the first one due to gRPC channel reuse.
- **TTL cache per store instance** — the store instance is created once at executor startup and shared across all assertion/monitor threads. Cache is protected by a per-instance lock. Lock is released during network calls so threads don't block each other on API latency.
- **Optional extras not framework deps** — `boto3` and `google-cloud-secret-manager` are optional plugin extras, not in `framework_common`, to avoid conflicts with Airflow's constraint files.

### Store config

| Field | AWS | GCP | Description |
|-------|-----|-----|-------------|
| `region` | Required | N/A | AWS region (e.g., `us-east-1`) |
| `project_id` | N/A | Required | GCP project ID |
| `prefix` | Required | Required | Prepended to secret names (e.g., `datahub-`) |
| `cache_ttl` | Required | Required | Cache TTL in seconds |

### Files

| File | Description |
|------|-------------|
| `src/datahub/secret/aws_secret_store.py` | AWS Secrets Manager store with batch API + cache |
| `src/datahub/secret/gcp_secret_store.py` | GCP Secret Manager store with retry + cache |
| `src/datahub/secret/secret_utils.py` | Shared prefix validation |
| `src/datahub/secret/secret_store.py` | Added `ABC` to base class |
| `setup.py` | Optional extras for `aws-secret-manager` and `gcp-secret-manager` |
| `tests/unit/secret/test_aws_secret_store.py` | 6 behavioral tests using moto |
| `tests/unit/secret/test_gcp_secret_store.py` | 6 behavioral tests using unittest.mock |

###  Unrelated Change 
aerospike update from 19.1.0 to 19.2.0

## Test plan

- [x] 6 AWS unit tests with `moto` (prefix mapping, batch fetch, not-found, mixed, single, empty)
- [x] 6 GCP unit tests with `unittest.mock` (prefix mapping, custom prefix, not-found, mixed, single, empty)
- [x] All 51 existing secret store tests pass (no regressions from ABC change)
- [x] Manual verification: created real secrets with multiple prefixes (`datahub-`, `temp-prod-`, `staging_`) in AWS Secrets Manager (sandbox) and GCP Secret Manager (acryl-staging), read them back through the stores


metadata-ingestion/src/datahub/secret/gcp_secret_store.py
```diff
@@ -0,0 +1,61 @@
...
+        if self._client is None:
+            self._client = secretmanager.SecretManagerServiceClient()
+        return self._client
+
+    def get_secret_values(self, secret_names: List[str]) -> Dict[str, Optional[str]]:
+        client = self._get_client()
+        results: Dict[str, Optional[str]] = {}
+        for name in secret_names:
+            resource = (
+                f"projects/{self.config.project_id}/secrets/{name}/versions/latest"
```
skrydal: Don't we need to validate chars in `{name}` here?

metadata-ingestion/tests/unit/secret/test_aws_secret_store.py
```diff
@@ -0,0 +1,88 @@
...
+from datahub.secret.aws_secret_store import (
+    AwsSecretsManagerStore,
+    AwsSecretsManagerStoreConfig,
+)
+
+
+class TestAwsSecretsManagerStore:
+    def test_create(self):
+        store = AwsSecretsManagerStore.create({"region": "us-east-1"})
+        assert isinstance(store, AwsSecretsManagerStore)
```
skrydal: under what conditions this line would not be `True`?

metadata-ingestion/tests/unit/secret/test_aws_secret_store.py
```diff
@@ -0,0 +1,88 @@
...
+        assert store.config.region == "us-east-1"
+
+    def test_get_id(self):
+        store = AwsSecretsManagerStore.create({"region": "us-east-1"})
+        assert store.get_id() == "aws-sm"
```
skrydal, line 20: This is quite trivial, maybe merge it with the above test.

metadata-ingestion/tests/unit/secret/test_aws_secret_store.py
```diff
@@ -0,0 +1,88 @@
...
+        assert result == {}
+
+    def test_close(self):
+        store = AwsSecretsManagerStore.create({"region": "us-east-1"})
+        store.close()
+        assert store._client is None
+
+    def test_config_defaults(self):
+        config = AwsSecretsManagerStoreConfig(region="eu-west-1")
+        assert config.region == "eu-west-1"
```
skrydal: this is testing:
* something already tested in "test_create"
* not testing defaults (which would be also wrong on its own)

metadata-ingestion/src/datahub/secret/gcp_secret_store.py
```diff
@@ -0,0 +1,61 @@
...
+                    f"Failed to fetch secret '{name}' from GCP Secret Manager"
+                )
+                results[name] = None
+        return results
+
+    def get_id(self) -> str:
+        return "gcp-sm"
+
+    def close(self) -> None:
+        self._client = None
```
skrydal: what's the intended use of this function?

metadata-ingestion/setup.py
```diff
@@ -42,7 +42,16 @@
...
+cloud_secret_common = {
+    # AWS Secrets Manager support
+    "boto3>=1.35.0,<2.0.0",
+    "botocore!=1.23.0,<2.0.0",
+    # GCP Secret Manager support
+    "google-cloud-secret-manager>=2.0.0,<3.0.0",
+}
+
 framework_common = {
+    *cloud_secret_common,
```
skrydal: boto and gcp dependencies are problematic. Let's not pull them by default.
The client using the new stores should install gcp/aws extras if they want to use those stores.

metadata-ingestion/src/datahub/secret/gcp_secret_store.py
rajatoss: when we extract secrets from recipe we already have regex which ensure that they are valid chars

metadata-ingestion/tests/unit/secret/test_aws_secret_store.py
rajatoss: updated

metadata-ingestion/tests/unit/secret/test_aws_secret_store.py
rajatoss: updated, i should have been more careful with claude while writing the tests, will keep in mind

metadata-ingestion/src/datahub/secret/gcp_secret_store.py
rajatoss: it's required by the abstract base class, but it's never called  in practice. I could just pass like the other stores do

metadata-ingestion/tests/unit/secret/test_aws_secret_store.py
rajatoss: removed

metadata-ingestion/setup.py
rajatoss: updated

rajatoss:
for gcp we dont have a batch api, but for gcp we use grpc so once the connection is established it is very fast.

metadata-ingestion/src/datahub/secret/secret_utils.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def validate_prefix(prefix: str) -> str:
+    """Validate prefix contains only characters safe for both AWS and GCP secret names."""
+    if not prefix:
+        return ""
+    if not _VALID_PREFIX_PATTERN.match(prefix):
+        raise ValueError(
+            f"Invalid secret prefix '{prefix}'. "
+            f"Only letters, digits, hyphens, and underscores are allowed (must work for both AWS and GCP)."
```
skrydal, line 14:
```suggestion
            f"Only letters, digits, hyphens, and underscores are allowed."
```

metadata-ingestion/src/datahub/secret/gcp_secret_store.py
```diff
@@ -0,0 +1,120 @@
...
+                    except ImportError as e:
+                        raise ImportError(
+                            "google-cloud-secret-manager is required for GCP Secret Manager support. "
+                            "Install with: pip install 'acryl-datahub[gcp-secret-manager]'"
+                        ) from e
+                    self._client = secretmanager.SecretManagerServiceClient()
+        return self._client
+
+    def _fetch_from_gcp(self, secret_names: List[str]) -> Dict[str, Optional[str]]:
+        from google.api_core.retry import Retry
```
skrydal: why importing in a function?

metadata-ingestion/src/datahub/secret/gcp_secret_store.py
```diff
@@ -0,0 +1,120 @@
...
+
+        # Retry config for GCP Secret Manager.
+        # The first gRPC call to establish the channel can take 10-30s ("cold channel" problem).
+        default_retry = Retry(initial=1.0, maximum=10.0, multiplier=2.0, deadline=30.0)
```
skrydal: I would move it to the __init__, to avoid object recreation with each call

metadata-ingestion/src/datahub/secret/gcp_secret_store.py
```diff
@@ -0,0 +1,120 @@
...
+                    cached[name] = self._cache[name]
+                else:
+                    to_fetch.append(name)
+
+        if to_fetch:
+            fresh = self._fetch_from_gcp(to_fetch)
+
+            with self._cache_lock:
+                for k, v in fresh.items():
+                    if v is not None:
```
skrydal: We should cache the empty result from the store as well, to avoid extensive calls for assertions/monitors.

metadata-ingestion/src/datahub/secret/gcp_secret_store.py
rajatoss: google.api_core is an optional dependency, it's only guaranteed to be installed when the gcp-secret-manager extra is installed.

metadata-ingestion/src/datahub/secret/gcp_secret_store.py
skrydal: This applies to both implementations.

metadata-ingestion/src/datahub/secret/gcp_secret_store.py
```diff
@@ -0,0 +1,120 @@
...
+            try:
+                resp = client.access_secret_version(
+                    request={"name": resource}, retry=default_retry
+                )
+                results[name] = resp.payload.data.decode("UTF-8")
+            except Exception:
+                logger.exception(
+                    f"Failed to fetch secret '{name}' from GCP Secret Manager"
+                )
+                results[name] = None
```
skrydal, line 93: We should have a distinction between an actual error and secret simply not being there (if possible).
I wonder whether we need to use exception level logging, actually it might be considered a normal course of operation to not have a secret in secret manager and have it provided via different means (i.e. env variables).
Even for other type of errors (timeouts, auth) I would not print on exception level but rather error, with one-line indication of the exception. We do not need entire stack.

metadata-ingestion/src/datahub/secret/gcp_secret_store.py
```diff
@@ -0,0 +1,120 @@
...
+
+            cached.update(fresh)
+
+        return cached
+
+    def get_id(self) -> str:
+        return "gcp-sm"
+
+    def close(self) -> None:
+        pass
```
skrydal: Actually it is possible to close gRPC connection.

metadata-ingestion/src/datahub/secret/aws_secret_store.py
```diff
@@ -0,0 +1,132 @@
...
+            raise ValueError("cache_ttl must be a positive integer")
+        return v
+
+
+class AwsSecretsManagerStore(SecretStore):
+    """SecretStore implementation that fetches secrets from AWS Secrets Manager."""
+
+    def __init__(self, config: AwsSecretsManagerStoreConfig):
+        self.config = config
+        self._client: Optional[Any] = None
```
skrydal: Please specify a type here.

metadata-ingestion/src/datahub/secret/aws_secret_store.py
```diff
@@ -0,0 +1,132 @@
...
+        self._cache: TTLCache = TTLCache(maxsize=1000, ttl=config.cache_ttl)
+        self._cache_lock = threading.Lock()
```
skrydal, line 46: I would consider having 2 locks (one for cache and one for init)

metadata-ingestion/src/datahub/secret/gcp_secret_store.py
```diff
@@ -0,0 +1,120 @@
...
+            raise ValueError("cache_ttl must be a positive integer")
+        return v
+
+
+class GcpSecretManagerStore(SecretStore):
+    """SecretStore implementation that fetches secrets from GCP Secret Manager."""
+
+    def __init__(self, config: GcpSecretManagerStoreConfig):
+        self.config = config
+        self._client: Optional[Any] = None
```
skrydal: Please specify the type here

metadata-ingestion/src/datahub/secret/gcp_secret_store.py
skrydal: oh ok to be able to statically import it in the client code even without extras, is that right?

metadata-ingestion/src/datahub/secret/gcp_secret_store.py
rajatoss: this is cool, agree.

metadata-ingestion/constraints.txt
```diff
@@ -12,7 +12,7 @@ adlfs==2026.2.0
     # via pyiceberg
 aenum==3.1.16
     # via okta
-aerospike==19.1.0
+aerospike==19.2.0
```
skrydal, line 15: I would consider calling out this line in the PR description as unrelated (but keep it).

skrydal (approved):
LGTM, please address the remaining comments.

metadata-ingestion/src/datahub/secret/gcp_secret_store.py
rajatoss: added

metadata-ingestion/src/datahub/secret/gcp_secret_store.py
rajatoss: added

metadata-ingestion/src/datahub/secret/gcp_secret_store.py
rajatoss: added

metadata-ingestion/src/datahub/secret/aws_secret_store.py
rajatoss: added

metadata-ingestion/src/datahub/secret/gcp_secret_store.py
rajatoss: added

metadata-ingestion/src/datahub/secret/aws_secret_store.py
rajatoss: added

rajatoss:
merged.

PR #17018, opened 2026-04-14
perf(dataplex): parallelize entry detail fetching and lineage lookups

sgomezvillamor:
## Summary

The Dataplex connector made one sequential blocking RPC per entry (`get_entry`) and one per entry × location pair for lineage (`search_links`). On large deployments this made ingestion wall-clock time proportional to entry count.

This PR parallelises both stages with `ThreadPoolExecutor`, adds thread-safety (locks on shared report objects and the project-container deduplication set), and removes a leftover `batch_size` config field along with a stale `dataset_pattern` backwards-compatibility alias.

**Key changes:**
- `process_entries`: three-phase — sequential list stubs → parallel `get_entry` (pool of `max_workers_entries`) → sequential Spanner search
- `get_lineage_workunits`: one thread per entry for `search_links` lookups (pool of `max_workers_lineage`)
- Both stages submit work in batches of `WORKERS_BATCH_SIZE=200` to cap in-flight futures (naive backpressure; proper implementation is a future TODO)
- Default workers: `max_workers_entries=10`, `max_workers_lineage=10`
- Removed: `DataplexConfig.batch_size` (unused leftover), `dataset_pattern` alias (breaking change already declared)

## Performance results (10 workers vs serial)

Tested against two real deployments. Project names anonymised.

**Deployment A — ~5 800 events, moderate lineage (~1 400 edges)**

| | Serial | 10 workers |
|---|---|---|
| Entries stage | 149 s | 48 s (3.1×) |
| Lineage stage | 16 036 s | 7 729 s (2.1×) |
| **Total** | **4 h 30 m** | **2 h 10 m** |

The lineage improvement was limited (~2×) rather than ~10× because the `search_links` API was responding ~137× slower than in Deployment B — consistent with per-project quota throttling under load.

**Deployment B — ~208 000 events, larger lineage (~8 400 edges)**

| | Serial | 10 workers |
|---|---|---|
| Entries stage | 3 822 s | 1 171 s (3.3×) |
| Lineage stage | 11 801 s | 1 052 s (11.2×) |
| **Total** | **4 h 21 m** | **37 m** |

Deployment B shows near-linear lineage scaling (11.2× with 10 workers) where API latency was low. Entry fetching improves consistently at ~3× in both environments regardless of API conditions.

## Thread safety

- `DataplexEntriesReport` and `DataplexLineageReport` protect all mutable fields with an internal `threading.Lock` (initialised in `__post_init__`).
- `DataplexEntriesProcessor` uses `_container_lock` for atomic check+add on `_emitted_project_containers` and `_entry_data_lock` for appends to the shared entry list.
- GCP gRPC clients (`CatalogServiceClient`, `LineageClient`) are thread-safe and shared across all workers.

## Test plan

- [x] All existing unit tests pass (`224 passed`)
- [x] `ruff check` clean
- [x] Manual end-to-end runs on two production-like deployments (results above)

https://claude.ai/code/session_01ELd1nW2NYxUbvMzt4xETpB


metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_entries.py
```diff
@@ -139,103 +162,168 @@ def __init__(
...
+        * Phase 1c (sequential): run the Spanner ``search_entries`` workaround
+          for each project × location pair.  Spanner entries come back already
+          fully-fetched from the search API and require no separate
+          ``get_entry`` call, so there is nothing to parallelise here.
+        """
+        # Phase 1a: accumulate entry stubs across all projects and locations.
+        entry_stubs: List[Tuple[str, str]] = []  # (entry_name, location)
+        for project_id in project_ids:
+            for location in self.config.entries_locations:
+                for name in self._list_entry_stubs(project_id, location):
```
treff7es: The old code had a per-project try/except/continue so one failing project (e.g., PermissionDenied) would not prevent processing of other projects. The new code calls
  _list_entry_stubs in a bare loop during Phase 1a — if ANY (project_id, location) pair fails, the entire entries stage aborts and all previously collected stubs are lost.

metadata-ingestion/docs/sources/dataplex/dataplex_post.md
```diff
@@ -97,6 +97,46 @@ For more details, see [Dataplex Lineage Documentation](https://docs.cloud.google
...
+**Lineage stage** dispatches one worker per entry to fetch `search_links` results across all
+configured `lineage_locations`, so total API call time scales with
+`max(entries / max_workers_lineage)` rather than `entries × lineage_locations`.
+
+Two config fields control the thread pool sizes:
+
+| Field                 | Default | Description                                      |
+| --------------------- | ------- | ------------------------------------------------ |
+| `max_workers_entries` | `10`    | Workers for `get_entry` calls (entries stage)    |
+| `max_workers_lineage` | `20`    | Workers for `search_links` calls (lineage stage) |
```
treff7es: `max_workers_lineage` default is `20`, but the config declares default=`10`.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_entries.py
```diff
@@ -139,103 +162,168 @@ def __init__(
...
+        if project_container is not None:
+            with self._container_lock:
+                project_container_urn = project_container.urn.urn()
+                if project_container_urn not in self._emitted_project_containers:
+                    self._emitted_project_containers.add(project_container_urn)
+                    results.append(project_container)
 
-        yield from self.collect_spanner_entries(project_id, location)
+        entity = self.build_entity_for_entry(entry)
+        self._track_entry_for_lineage(dataplex_location=location, entry=entry)
```
treff7es: _track_entry_for_lineage is called unconditionally (before checking if entity is not None). This creates phantom lineage tracking entries — lineage API calls will be wasted for
  entries that produced no entity.

metadata-ingestion/docs/sources/dataplex/dataplex_post.md
```diff
@@ -97,6 +97,46 @@ For more details, see [Dataplex Lineage Documentation](https://docs.cloud.google
 
 - **`batch_size`** (default: `1000`): Entries per batch for memory optimization. Set to `None` to disable batching (small deployments only)
```
treff7es: batch_size still documented in dataplex_post.md after removal

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_config.py
```diff
@@ -73,7 +73,6 @@ class EntriesFilterConfig(ConfigModel):
 
     pattern: AllowDenyPattern = Field(
         default=AllowDenyPattern.allow_all(),
-        validation_alias=AliasChoices("pattern", "dataset_pattern"),
```
treff7es, line 76: Do we need to document this in breaking change?

treff7es (approved):

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_config.py
sgomezvillamor: It was documented in a previous PR while the alias was added too 🤦 
That's why I'm removing now.

metadata-ingestion/docs/sources/dataplex/dataplex_post.md
sgomezvillamor: addressed in 1408c7e7185ca3a1c66c903ea2f879d1a38e3b6a

metadata-ingestion/docs/sources/dataplex/dataplex_post.md
sgomezvillamor: addressed in 1408c7e7185ca3a1c66c903ea2f879d1a38e3b6a

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_entries.py
sgomezvillamor: good catch! addressed in ce33a816962a55c3bdd8a187222644b9cc6b51ce

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_entries.py
sgomezvillamor: In practice that's not the case because _track_entry_for_lineage already has early-return guards for all the same conditions that cause `build_entity_for_entry` to return `None`. However, the concern is technically fair as a code smell.
Addressing in commit 520b043ae008eec34c7e0371338f6f5d3198203d to make the intent obvious.

PR #17020, opened 2026-04-14
fix(ingest): database urn with special characters

alokr-dhub:
# Summary                                                                                                    
                                                                                                             
  - URN encoding bug: Database names containing (, ), or / (e.g. data_(read/write)_warehouse) were passed raw
   to make_dataset_urn_with_platform_instance. The parentheses break URN tuple parsing and the forward slash
  conflicts with URL path separators in REST API calls. Fixed by encoding each name component with           
  UrnEncoder.encode_string() + explicit %2F for / before joining, while keeping the unencoded name for
  pattern filtering and display.
  - Regex filtering bug: AllowDenyPattern.allowed() / denied() compile user-supplied patterns as regexes, so
  a database name like sales_(jan/feb)_data listed literally in an allow/deny rule fails to match itself     
  because ( and ) are interpreted as regex group delimiters. Fixed by adding a cached literal-string fallback
   in both methods: if no compiled pattern matches, the input is compared directly against the pattern       
  strings (case-insensitive when ignoreCase is set).


metadata-ingestion/src/datahub/configuration/common.py
```diff
@@ -449,10 +462,20 @@ def allowed(self, string: str) -> bool:
...
             return False
 
-        return any(pattern.match(string) for pattern in self._compiled_allow)
+        if any(pattern.match(string) for pattern in self._compiled_allow):
+            return True
+
+        # Literal fallback: names with regex metacharacters (e.g. 'sales_(jan/feb)_data')
+        # fail pattern.match() even when listed exactly. Try plain string comparison.
+        compare = string.lower() if self.ignoreCase else string
+        return compare in self._allow_literal_set
```
sgomezvillamor: I don't understand this fallback.

Isn't user responsibility to escape those characters if the intent is matching character instead of regexp group: `sales_\(jan/feb\)_data` ?

metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
```diff
@@ -1614,9 +1615,15 @@ def _gen_table_wu(self, table: Dict) -> Iterable[MetadataWorkUnit]:
             self.report.report_table_dropped(full_table_name)
             return
 
+        # Encode components for the URN; '/' must also be encoded to avoid conflicts
+        # with URL path separators when URNs appear in REST API calls.
+        urn_dataset_name = (
+            f"{UrnEncoder.encode_string(database_name).replace('/', '%2F')}"
+            f".{UrnEncoder.encode_string(table_name).replace('/', '%2F')}"
+        )
         dataset_urn = make_dataset_urn_with_platform_instance(
```
sgomezvillamor, line 1617: is there any reason for not pushing down this fix instead of solving this exclusively for glue? eg in the `make_dataset_urn**` methods
or also, why not part of the `encode_string` in `UrnEncoder`?

alokr-dhub: added it here as implementing it encode_string would affect other connectors. Let me move this to UrnEncoder.

metadata-ingestion/src/datahub/configuration/common.py
alokr-dhub: added it as precautionary measure in case the escape sequences are missed. Let me revert this.

metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
alokr-dhub: @sgomezvillamor Adding / to RESERVED_CHARS globally encodes every / in every URN name across the entire codebase. That breaks S3, GCS, HDFS, and Azure Blob URNs where / is the legitimate path separator (e.g.,urn:li:dataset:(urn:li:dataPlatform:s3,bucket/folder/file,PROD))

sgomezvillamor (changes requested):
Blocking this until we discuss internally the impact of the breaking change

PR #17021, opened 2026-04-14
feat(observability): add ingestion metrics emission in GMS request layer

kyungsoo-datahub:
Adds ingestion run metrics emission from GMS via a new MCPObserver plugin type.                                                                      

When an ExecutionRequestResult aspect is written (CLI, UI, or executor-based ingestion runs), the IngestionMetricsEmitter observer extracts metrics from the structured report and emits:                     

- Micrometer counters/summaries: com.datahub.ingest.runs, duration_ms, events_produced, records_written, warnings, failures, sink_failures — tagged by connector, status, cli_version                                   
- Structured log: JSON payload with per-run detail (execution_id, connector, stage durations, failure messages, etc.), identified by MDC field messageType=ingestionRunReport


metadata-io/src/main/java/com/linkedin/metadata/entity/EntityServiceImpl.java
```diff
@@ -2699,6 +2708,12 @@ public void setRetentionService(RetentionService<ChangeItemImpl> retentionServic
     this.retentionService = retentionService;
   }
 
+  @Autowired(required = false)
```
RyanHolstien: We don't use setter injection, please change this to have it set up via the constructor the same pattern we do for all of the other injected objects for this class.

metadata-io/src/main/java/com/linkedin/metadata/ingestion/IngestionMetricsEmitter.java
```diff
@@ -0,0 +1,490 @@
...
+ *       high-cardinality fields are emitted in the {@code [INGESTION_RUN_EVENT]} structured log
+ *       only.
+ *   <li>warnings/failures counters reflect the JSON array size, which the ingestion framework caps
+ *       at 10 via LossyList. A run with 200 failures shows failures_count=10. No total count field
+ *       exists in the structured report.
+ *   <li>No entity client RPCs are made — all data comes from the proposal itself.
+ * </ul>
+ */
+@Slf4j
+@Component
```
RyanHolstien: Please follow the standard pattern we use rather than having a conditional component scan.

metadata-io/src/main/java/com/linkedin/metadata/ingestion/IngestionMetricsEmitter.java
```diff
@@ -0,0 +1,490 @@
...
+@ConditionalOnProperty(name = "ingestionMetrics.enabled", havingValue = "true")
+public class IngestionMetricsEmitter {
+
+  private static final String ASPECT_NAME = "dataHubExecutionRequestResult";
+  // JMX-style prefix intentionally chosen: Micrometer converts dots to underscores in Prometheus,
+  // producing com_datahub_ingest_* which matches the Observe Agent allowlist filter regex.
+  private static final String METRIC_PREFIX = "com.datahub.ingest.";
+  private static final Set<String> SUPPORTED_REPORT_TYPES =
+      ImmutableSet.of("CLI_INGEST", "RUN_INGEST");
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
RyanHolstien: This should use the injected ObjectMapper not constructing a new one.

metadata-io/src/main/java/com/linkedin/metadata/ingestion/IngestionMetricsEmitter.java
```diff
@@ -0,0 +1,490 @@
...
+ *       exists in the structured report.
+ *   <li>No entity client RPCs are made — all data comes from the proposal itself.
+ * </ul>
+ */
+@Slf4j
+@Component
+@ConditionalOnProperty(name = "ingestionMetrics.enabled", havingValue = "true")
+public class IngestionMetricsEmitter {
+
+  private static final String ASPECT_NAME = "dataHubExecutionRequestResult";
```
RyanHolstien: This should use the constant for this aspect name in Constants.java

metadata-io/src/main/java/com/linkedin/metadata/entity/EntityServiceImpl.java
```diff
@@ -1678,6 +1682,11 @@ public IngestResult ingestProposal(
   @Override
   public List<IngestResult> ingestProposal(
       @Nonnull OperationContext opContext, AspectsBatch aspectsBatch, final boolean async) {
+    // Emit ingestion metrics on the actual processing path only (async=false).
```
RyanHolstien: Why are we injecting this as a new pattern instead of using the hook patterns we have? This should be written as a side effect not some new pattern.

metadata-io/src/main/java/com/linkedin/metadata/ingestion/IngestionMetricsEmitter.java
```diff
@@ -0,0 +1,490 @@
...
+    }
+  }
+
+  @Nullable
+  private JsonNode parseStructuredReport(@Nonnull ExecutionRequestResult result) {
+    try {
+      if (!result.hasStructuredReport()) {
+        return null;
+      }
+      String serializedValue = result.getStructuredReport().getSerializedValue();
```
RyanHolstien: Structured report has a content type, this seems invalid if it's not json so I would expect a check for that.

metadata-io/src/main/java/com/linkedin/metadata/ingestion/IngestionMetricsEmitter.java
```diff
@@ -0,0 +1,490 @@
...
+    event.put("execution_id", r.executionRequestUrn().getId());
+    event.put("connector", r.connector());
+    event.put("status", r.status());
+    if (r.durationMs() != null) {
+      event.put("duration_ms", r.durationMs());
+    }
+
+    JsonNode cli = r.reportJson().path("cli");
+    if (!cli.isMissingNode()) {
+      putIfPresent(event, "cli_version", cli, "cli_version");
```
RyanHolstien: I think it would be much more preferable to create actual POJOs with the expected fields to convert the Json into rather than doing a ton of hard coded strings and JsonNode logic. You have the record created so you should be able to define JsonProperties on it to convert to the serialized map.

metadata-io/src/main/java/com/linkedin/metadata/ingestion/IngestionMetricsEmitter.java
```diff
@@ -0,0 +1,490 @@
...
+      JsonNode reportJson = parseStructuredReport(result);
+      if (reportJson == null) {
+        log.warn("Could not parse structured report for {}", executionRequestUrn);
+        return;
+      }
+
+      JsonNode sourceReport = findSourceReport(reportJson);
+      JsonNode sinkReport = findSinkReport(reportJson);
+
+      String connector = extractConnector(reportJson, sourceReport);
```
RyanHolstien: If you define models then a lot of these extractions become unnecessary, let Jackson's mapping do the work for you.

metadata-io/src/main/java/com/linkedin/metadata/ingestion/IngestionMetricsEmitter.java
```diff
@@ -0,0 +1,444 @@
...
+      @Nullable Long durationMs,
+      long eventsProduced,
+      long recordsWritten,
+      int warningsCount,
+      int failuresCount,
+      int sinkFailuresCount,
+      @Nonnull IngestionRunReport.Report report) {
+    Map<String, Object> event = new LinkedHashMap<>();
+
+    event.put("execution_id", executionRequestUrn.getId());
```
RyanHolstien: nit: Can these be in Constants so we have a shared place for sourcing the Strings and they're traceable for usage in code? Should also use the static constant for the Json Prop annotations

RyanHolstien (approved):
Looking good! Would appreciate the raw Strings getting extracted to reusable constants, but going ahead and pre-approving.

david-leifker (approved):

metadata-io/src/main/java/com/linkedin/metadata/ingestion/IngestionMetricsEmitter.java
kyungsoo-datahub: Thank you for the suggestion. Common Constants.java is created for sharing common constants.

PR #17024, opened 2026-04-14
fix(ingest): split statement support end/END and IF not exists

kevinkarchacryl:
The sql statement splitter contains two bugs: 
1. it incorrectly treats the END keyword closing a CASE expression as a statement boundary due to a case-sensitive string comparison
2. it incorrectly treats IF in CREATE TABLE IF NOT EXISTS / DROP TABLE IF EXISTS DDL syntax as a control flow keyword. 

Both of these cause multi-statement SQL blocks containing these keywords to be split into broken fragments, resulting in failed lineage extraction parsing issues in the airflow plugin.


skrydal (approved):
Excellent catch

PR #17026, opened 2026-04-14
feat(assertions): Assertion source rollout

AdrianMachado:
## Summary

Adds a `Source` to every assertion created outside of the UI through areas we control (SDK, ingestion, etc.) The `create` time is generally useful metadata for auditing, and is necessary for properly configuring backfill. The `external` type is nice to have as a distinguishing characteristic for these assertions


jayacryl (approved):

fZhang-j:
Nice work!

PR #17027, opened 2026-04-14
feat(ingest): configurable report sample sizes for progress and final reports

rob-1019:
## Summary

Add configurable caps for ingestion report sizes, separately for interim (60-second progress) and final reports. Both source and sink reports are affected.

### Configuration

All settings default to 10. Recipe flags take precedence over env vars.

**Interim progress report display caps:**
```yaml
flags:
  progress_report_max_failures: 10
  progress_report_max_warnings: 10
  progress_report_max_infos: 10
```

**Final report sample sizes:**
```yaml
flags:
  report_failure_sample_size: 10
  report_warning_sample_size: 10
  report_info_sample_size: 10
```

Equivalent env vars: `DATAHUB_PROGRESS_REPORT_MAX_FAILURES`, `DATAHUB_PROGRESS_REPORT_MAX_WARNINGS`, `DATAHUB_PROGRESS_REPORT_MAX_INFOS`, `DATAHUB_REPORT_FAILURE_SAMPLE_SIZE`, `DATAHUB_REPORT_WARNING_SAMPLE_SIZE`, `DATAHUB_REPORT_INFO_SAMPLE_SIZE`.

The underlying `LossyDict` retains `max(report_size, progress_max)` entries so both reports can draw from a sufficient pool. Each report then displays up to its own configured cap.

Info sample size (`DATAHUB_REPORT_INFO_SAMPLE_SIZE`) is now configurable via env var and recipe flag, matching the existing behavior of failures and warnings.

## Test plan

- [x] 24 unit tests covering capping logic, sentinel handling, FlagsConfig defaults/env/recipe precedence, retention combos, and sink sizing
- [x] End-to-end: interim + final reports with various flag combinations (suppress interim, cap interim, override per severity, size > generated count)

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/api/report.py
+++ b/metadata-ingestion/src/datahub/ingestion/api/report.py
@@ -106,8 +106,12 @@ class Report(SupportsAsObj):
             if value is not None and not str(key).startswith("_")
         }
 
-    def as_string(self) -> str:
+    def as_string(
+        self, progress_sample_caps: Optional[Dict[str, int]] = None
+    ) -> str:
         self_obj = self.as_obj()
+        if progress_sample_caps is not None:
+            self_obj = _cap_report_samples(self_obj, progress_sample_caps)
         _aspects_by_subtypes = self_obj.pop("aspects_by_subtypes", None)
 
         # Format the main report data
@@ -529,3 +533,17 @@ class EntityFilterReport(ReportAttribute):
         return dataclasses.field(
             default_factory=lambda: EntityFilterReport(type=type, severity=severity)
         )
+
+
+def _cap_report_samples(obj: dict, caps: Dict[str, int]) -> dict:
+    """Cap the failures/warnings/infos lists in an as_obj() output."""
+    result = {}
+    for k, v in obj.items():
+        cap = caps.get(k)
+        if cap is not None and isinstance(v, list) and len(v) > cap:
+            result[k] = v[:cap] + [
+                f"... showing {cap} of {len(v)} (interim report)"
+            ]
+        else:
+            result[k] = v
+    return result
--- a/metadata-ingestion/src/datahub/ingestion/api/report.py
+++ b/metadata-ingestion/src/datahub/ingestion/api/report.py
@@ -106,10 +106,16 @@ class Report(SupportsAsObj):
             if value is not None and not str(key).startswith("_")
         }
 
-    def as_string(self, progress_sample_caps: Optional[Dict[str, int]] = None) -> str:
+    def as_string(
+        self,
+        progress_sample_caps: Optional[Dict[str, int]] = None,
+        cap_label: str = "interim report",
+    ) -> str:
         self_obj = self.as_obj()
         if progress_sample_caps is not None:
-            self_obj = _cap_report_samples(self_obj, progress_sample_caps)
+            self_obj = _cap_report_samples(
+                self_obj, progress_sample_caps, label=cap_label
+            )
         _aspects_by_subtypes = self_obj.pop("aspects_by_subtypes", None)
 
         # Format the main report data
@@ -533,13 +539,15 @@ class EntityFilterReport(ReportAttribute):
         )
 
 
-def _cap_report_samples(obj: dict, caps: Dict[str, int]) -> dict:
+def _cap_report_samples(
+    obj: dict, caps: Dict[str, int], label: str = "interim report"
+) -> dict:
     """Cap the failures/warnings/infos lists in an as_obj() output."""
     result = {}
     for k, v in obj.items():
         cap = caps.get(k)
         if cap is not None and isinstance(v, list) and len(v) > cap:
-            result[k] = v[:cap] + [f"... showing {cap} of {len(v)} (interim report)"]
+            result[k] = v[:cap] + [f"... showing {cap} of {len(v)} ({label})"]
         else:
             result[k] = v
     return result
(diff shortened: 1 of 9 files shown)


metadata-ingestion/src/datahub/ingestion/run/pipeline.py
```diff
@@ -253,6 +253,24 @@ def __init__(
                     )
                     logger.info("Source configured successfully.")
 
+                    # Retain enough entries for whichever is larger: the
+                    # final report size or the interim display cap.
+                    flags = self.config.flags
+                    self.source.get_report()._structured_logs.set_sample_sizes(
```
askumar27: suggestion (non-blocking): Add delegating `set_sample_sizes` on `SourceReport`. More interms of code hygiene here, we are accessing and modifying `._structured_logs`

metadata-ingestion/src/datahub/ingestion/run/pipeline_config.py
```diff
@@ -66,6 +74,56 @@ class FlagsConfig(ConfigModel):
...
+    )
+    report_info_sample_size: int = Field(
+        default_factory=get_report_info_sample_size,
+        description=(
+            "How many info entries to retain. Controls the final report size "
+            "and the pool that interim reports draw from. "
+            "Also settable via DATAHUB_REPORT_INFO_SAMPLE_SIZE env var."
+        ),
+    )
+
```
askumar27, line 132: suggestion (if-minor): Lets add a check for lower bounds `ge=0`

metadata-ingestion/src/datahub/ingestion/api/source.py
```diff
@@ -196,6 +197,24 @@ def report_log(
...
+    def set_sample_sizes(
+        self,
+        failure_size: Optional[int] = None,
+        warning_size: Optional[int] = None,
+        info_size: Optional[int] = None,
+    ) -> None:
+        """Override the max_elements on the underlying LossyDicts.
+
+        Must be called before any entries are added (i.e. right after source
+        construction) — resizing a non-empty LossyDict is not supported.
```
askumar27: suggestion: We can add a run time check here, but doc string works too - can go as a follow-up.

metadata-ingestion/src/datahub/ingestion/api/report.py
```diff
@@ -529,3 +534,15 @@ def field(type: str, severity: LogLevel = "DEBUG") -> "EntityFilterReport":
...
         )
+
+
+def _cap_report_samples(obj: dict, caps: Dict[str, int]) -> dict:
+    """Cap the failures/warnings/infos lists in an as_obj() output."""
+    result = {}
+    for k, v in obj.items():
+        cap = caps.get(k)
+        if cap is not None and isinstance(v, list) and len(v) > cap:
+            result[k] = v[:cap] + [f"... and {len(v) - cap} more"]
```
askumar27: issue (blocking): `_cap_report_samples` doesn't account for the LossyList sampling sentinel string. With default settings and >10 failures, the final report shows "... and 1 more" even when all 10 retained entries are already displayed — that "1 more" is the sentinel, not a real failure.

askumar27 (changes requested):
_Thought:_ May be we can combine the fine-grained env flags/configs to accept one json config, instead of setting many env flags/configs

metadata-ingestion/src/datahub/ingestion/run/pipeline.py
```diff
@@ -253,6 +253,24 @@ def __init__(
                     )
                     logger.info("Source configured successfully.")
 
+                    # Retain enough entries for whichever is larger: the
+                    # final report size or the interim display cap.
+                    flags = self.config.flags
+                    self.source.get_report().set_sample_sizes(
```
kyungsoo-datahub, line 271: SinkReport’s LossyList fields are still env-sized only; consider applying the same report_* / max(report, progress) sizing here so flags behave like the source.

metadata-ingestion/src/datahub/ingestion/api/report.py
```diff
@@ -106,8 +106,13 @@ def as_obj(self) -> dict:
             if value is not None and not str(key).startswith("_")
         }
 
-    def as_string(self) -> str:
+    def as_string(
+        self,
+        progress_sample_caps: Optional[Dict[str, int]] = None,
```
kyungsoo-datahub: progress_sample_caps is also used for the final print when currently_running=False; consider renaming (e.g. sample_caps) so it’s not read as interim-only.

metadata-ingestion/src/datahub/ingestion/api/report.py
rob-1019: Fixed - added a test and explicitly typed the sentinel to avoid similar bugs in the future

metadata-ingestion/src/datahub/ingestion/api/source.py
rob-1019: done

metadata-ingestion/src/datahub/ingestion/run/pipeline.py
rob-1019: Fixed and test added

metadata-ingestion/src/datahub/ingestion/api/report.py
rob-1019: renamed

askumar27 (approved):
LGTM! Thanks @rob-1019 for quick turnaround with review comments.

PR #17032, opened 2026-04-14
fix(oidc): fix redirect cookie creation

RyanHolstien:
Fixes issues with redirect url manipulation by an validly authenticated user that modifies their own redirect url cookie.


craig-rueda (approved):

chakru-r:
test

PR #17033, opened 2026-04-15
fix(mongodb): exclude system collections by default, correct required permissions

javabrett:
## Summary

MongoDB `system.*` collections are internal MongoDB collections not intended for user metadata. This PR fixes two silent problems:

- **`system.profile`** only exists when MongoDB profiling is enabled (non-default). When present, the connector attempted to run schema inference against it — but `system.profile` requires `dbAdmin`, not `read` or `readWrite`. This caused ingestion failures for users who had profiling enabled. A previous support incident was caused by this. The docs were then incorrectly updated to require `readWrite`, which was based on incorrect analysis and does not actually solve the `system.profile` permission issue.

- **`system.views`** is readable with the `read` role, so no auth error. But ingesting it creates noisy, incorrect metadata — the collection contains view definition records like `{"_id": "mydb.revenue_summary", "viewOn": "raw_transactions", "pipeline": [...]}`, not actual user data.

The connector had no intentional use of system collection data — all access was accidental.

**Fix:** New `excludeSystemCollections` boolean config field (default `true`). When enabled, system collections are skipped before any I/O, avoiding the auth error entirely. The `read` role (or `readAnyDatabase`) is now sufficient for all standard ingestion — `readWrite` is not required and was never needed.

## Changes

- `mongodb.py`: New `excludeSystemCollections` config field (default `true`); filtering applied before the collection pattern check
- `mongodb_pre.md`: Prerequisites section rewritten — corrects permission requirement to `read`, documents the system collections behavior and upgrade path for users on older versions
- `next.md`: Release note entry
- `test_mongodb_source.py`: Two new unit tests — default exclusion behavior and opt-in inclusion

## Test plan

- [x] `pytest tests/unit/test_mongodb_source.py` — all 9 tests pass (7 existing + 2 new)
- [ ] Manual: verify `excludeSystemCollections: false` still ingests `system.views` as expected


docs/managed-datahub/release-notes/next.md
```diff
@@ -29,7 +29,14 @@ New Features:
 
 Fixes:
 
-- TODO
+- **MongoDB**: System collections (`system.*`) are now excluded from ingestion by default.
```
sgomezvillamor: Ingestion connectors are OSS.
So AFAIK we should not mention this in the datahub cloud release notes.

Instead, we should mention the breaking change (even if a fix) in the `updating-datahub.md`

metadata-ingestion/docs/sources/mongodb/mongodb_pre.md
```diff
@@ -4,10 +4,43 @@ The `mongodb` module ingests metadata from Mongodb into DataHub. It is intended
...
+db.grantRolesToUser("datahub_ingest", [
+  { role: "readAnyDatabase", db: "admin" },
+]);
+```
+
+> **Note on system collections:** MongoDB `system.*` collections (such as `system.profile` and
+> `system.views`) are excluded from ingestion by default via the `excludeSystemCollections`
+> option (enabled by default). This means the `read` role is sufficient for all standard
+> ingestion use cases. If you disable `excludeSystemCollections`, ingestion of `system.profile`
+> requires the `dbAdmin` role — see [Config Details](#config-details) for more information.
```
sgomezvillamor:
```suggestion
:::tip "Note on system collections"

 MongoDB `system.*` collections (such as `system.profile` and
`system.views`) are excluded from ingestion by default via the `excludeSystemCollections`
option (enabled by default). This means the `read` role is sufficient for all standard
ingestion use cases. If you disable `excludeSystemCollections`, ingestion of `system.profile`
requires the `dbAdmin` role — see [Config Details](#config-details) for more information.
:::

```

metadata-ingestion/docs/sources/mongodb/mongodb_pre.md
```diff
@@ -4,10 +4,43 @@ The `mongodb` module ingests metadata from Mongodb into DataHub. It is intended
...
+The workaround is to explicitly deny system collections in your recipe until you can upgrade:
+
+```yaml
+source:
+  type: mongodb
+  config:
+    collection_pattern:
+      deny:
+        - ".*\\.system\\..*"
+```
```
sgomezvillamor: I would just put this "upgrading form earlier version" in the "note on system collections above"

sgomezvillamor: Also, I would avoid terms such as "earlier" version
Since tracking version numbers is tricky, you can refer to "if upgrading from a version that ingested system tables by default..." or similar

metadata-ingestion/src/datahub/ingestion/source/mongodb.py
```diff
@@ -141,6 +141,17 @@ class MongoDBConfig(
         default=AllowDenyPattern.allow_all(),
         description="regex patterns for collections to filter in ingestion.",
     )
+    excludeSystemCollections: bool = Field(
```
sgomezvillamor, line 144: Definitely, the naming style is out of control in this connector, not sure why
We usually follow snake case style, however I see lots of cases here with camel case style
😿

sgomezvillamor (changes requested):
Overall LGTM

Left some minor comments, but the one on datahub cloud release notes, which should be addressed

sgomezvillamor (approved):

PR #17039, opened 2026-04-15
fix(kafka): add bounded auth exception retry to consumer factories for MSK IAM resilience

max-datahub:
## Summary

- Adds configurable `authExceptionRetryInterval` to all Kafka consumer container factories (`KafkaEventConsumerFactory`, `SimpleKafkaConsumerFactory`, `CDCConsumerFactory`)
- Adds `KafkaConsumerAuthExceptionGuard` to bound retries: after max consecutive auth failures, the consumer container is stopped so `KafkaHealthIndicator` reports DOWN and Kubernetes can act
- Both settings are configuration-driven via `application.yaml` / env vars, matching the pattern used by `kafka.producer` (retryCount, deliveryTimeout, etc.)
- Prevents MAE/MCE/PE consumer pods from crashing on `SaslAuthenticationException` during MSK IAM credential rotation
- Gives consumers the same self-healing behavior that the GMS producer already has, with equivalent bounding

## Configuration

| Property | Env var | Default | Description |
|---|---|---|---|
| `kafka.consumer.authExceptionRetryIntervalSeconds` | `KAFKA_CONSUMER_AUTH_EXCEPTION_RETRY_INTERVAL_SECONDS` | `10` | Seconds between retries after auth exception. Set to `0` to disable (fatal on first auth exception). |
| `kafka.consumer.maxAuthExceptionRetries` | `KAFKA_CONSUMER_MAX_AUTH_EXCEPTION_RETRIES` | `3` | Max consecutive auth retries before stopping the container. Set to `0` to disable bounding (unbounded retries). |

## Problem

On MSK IAM deployments using EKS Pod Identity, credential rotation (~1h cycle) changes the STS session name. When MSK triggers re-authentication on an existing connection, it rejects the new principal: `Cannot change principals during re-authentication`.

Spring Kafka's default behavior treats `SaslAuthenticationException` as fatal and stops the consumer container immediately. This causes pod restarts on MAE, MCE, and PE consumers.

GMS (producer) already survives this because the Kafka client's `NetworkClient` automatically reconnects with fresh credentials on the next `send()` call. Consumer containers use the same `NetworkClient`, but Spring Kafka kills the container before reconnection can happen.

### Auth exception retry interval

`authExceptionRetryInterval` tells Spring Kafka to sleep and retry `poll()` instead of stopping the container. The retry forces a new connection with fresh authentication (not re-authentication), recovering the same way the producer already does.

This is the mechanism Spring Kafka explicitly designed for temporary auth failures. Zero infrastructure changes required.

### Bounded retries (aligned with GMS producer)

Spring Kafka's `authExceptionRetryInterval` retries indefinitely — the container stays "running" and the health check never detects a permanent auth failure. This is unlike the GMS producer, which bounds retries via `delivery.timeout.ms` and `retries`.

`KafkaConsumerAuthExceptionGuard` mirrors the producer's bounding behavior:
- Listens for `ConsumerRetryAuthEvent` and counts consecutive failures per container
- After exceeding the configured max retries, stops the container — `KafkaHealthIndicator` reports DOWN, K8s liveness probes restart the pod
- Resets the counter on `ConsumerRetryAuthSuccessfulEvent` (transient credential rotation recovered)

Transient MSK IAM rotation recovers on the first retry. Reaching the max retry threshold signals a permanent failure requiring operator intervention.

## Checklist
- [x] The PR conforms to DataHub's [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (specifically the [PR title format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#pr-title-format))
- [x] Tests added/updated
- [ ] Docs added/updated

### DIFF
--- a/metadata-service/configuration/src/main/java/com/linkedin/metadata/config/kafka/ConsumerConfiguration.java
+++ b/metadata-service/configuration/src/main/java/com/linkedin/metadata/config/kafka/ConsumerConfiguration.java
@@ -9,6 +9,8 @@ public class ConsumerConfiguration {
   private int maxPartitionFetchBytes;
   private boolean stopOnDeserializationError;
   private boolean healthCheckEnabled;
+  private int authExceptionRetryIntervalSeconds;
+  private int maxAuthExceptionRetries;
 
   private ConsumerOptions mcp;
   private ConsumerOptions mcl;
--- a/metadata-io/src/test/java/com/linkedin/metadata/system_info/collectors/PropertiesCollectorConfigurationTest.java
+++ b/metadata-io/src/test/java/com/linkedin/metadata/system_info/collectors/PropertiesCollectorConfigurationTest.java
@@ -277,6 +277,8 @@ public class PropertiesCollectorConfigurationTest extends AbstractTestNGSpringCo
           "kafka.consumer.metrics.slo",
           "kafka.consumer.pe.autoOffsetReset",
           "kafka.consumer.stopOnDeserializationError",
+          "kafka.consumer.authExceptionRetryIntervalSeconds",
+          "kafka.consumer.maxAuthExceptionRetries",
           "kafka.consumerPool.initialSize",
           "kafka.consumerPool.maxSize",
           "kafka.consumerPool.validationTimeoutSeconds",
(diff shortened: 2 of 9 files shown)


metadata-service/configuration/src/main/java/com/linkedin/metadata/config/kafka/ConsumerConfiguration.java
```diff
@@ -9,6 +9,8 @@ public class ConsumerConfiguration {
   private int maxPartitionFetchBytes;
   private boolean stopOnDeserializationError;
   private boolean healthCheckEnabled;
+  private int authExceptionRetryIntervalSeconds = 10;
```
david-leifker: Avoid defaults outside of the application.yaml config

david-leifker (approved):

metadata-service/configuration/src/main/java/com/linkedin/metadata/config/kafka/ConsumerConfiguration.java
max-datahub: fixed

PR #17047, opened 2026-04-16
perf(bigquery): Optimize policy tag extraction with batch API calls

sgomezvillamor:
## Summary

Refactored BigQuery policy tag extraction to use batch Data Catalog API calls instead of per-tag lookups, significantly improving performance when extracting policy tags from large datasets.

### Core Optimization
- **Replaced per-tag API calls with taxonomy-level batch calls**: Instead of calling `get_policy_tag()` for each individual tag, the new implementation calls `list_policy_tags(parent=taxonomy_id)` once per unique taxonomy, resolving all tags in that taxonomy in a single API call.
- **Added taxonomy-level caching**: Introduced `_policy_tag_mapping_cache` to cache policy tag display names across all datasets in a single ingestion run, eliminating redundant API calls when the same taxonomy is referenced across multiple datasets.
- **Integrated policy tags into INFORMATION_SCHEMA query**: Modified the `COLUMNS_QUERY` to fetch policy tags directly from `INFORMATION_SCHEMA.COLUMN_FIELD_PATHS`, eliminating the need for separate table schema lookups.

### Implementation Details
- Added `_parse_taxonomy_id()` helper function to extract taxonomy resource names from policy tag resource names using regex.
- Replaced `get_policy_tags_for_column()` with `build_policy_tag_display_name_mapping()` that accepts a set of policy tag resource names and returns a mapping to display names.
- Refactored `get_columns_for_dataset()` to:
  1. Collect all raw rows from INFORMATION_SCHEMA first
  2. Extract all unique policy tag resource names from the dataset
  3. Batch-resolve display names in a single call to `build_policy_tag_display_name_mapping()`
  4. Construct BigqueryColumn objects with resolved display names

### Graceful Degradation
- Malformed policy tag resource names are stored as-is (used as their own display name)
- API failures fall back to using resource names as display names with a warning logged
- Deleted tags (not returned by `list_policy_tags`) gracefully fall back to resource names

## Testing

- Added comprehensive unit test suite (`test_bigquery_policy_tags.py`) with 16 tests covering:
  - Taxonomy ID parsing (valid, different regions, malformed cases)
  - Policy tag display name mapping (empty input, deduplication, multiple taxonomies, caching)
  - Integration with `get_columns_for_dataset()` (with/without flag, empty tags, deduplication, cache reuse)
  - Error handling (API failures, deleted tags)

## Documentation

- Updated `bigquery_config.py` to clarify the new extraction mechanism
- Added entry to `docs/how/updating-datahub.md` documenting the breaking change and migration path

## Performance Impact

For a dataset with N columns across M unique taxonomies:
- **Before**: O(N) Data Catalog API calls (one per column with tags)
- **After**: O(M) Data Catalog API calls (one per unique taxonomy)

For typical datasets with many columns sharing the same taxonomy, this represents a significant reduction in API calls and latency.

## Related Issues

Addresses performance concerns with policy tag extraction in large BigQuery datasets.

https://claude.ai/code/session_01PWiX9Uy9TMyx5QD7xyX9KM


metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_schema.py
```diff
@@ -742,33 +792,66 @@ def get_columns_for_dataset(
...
+            # small number of Data Catalog API calls (one per unique taxonomy).
+            policy_tag_display_name_map: Dict[str, str] = {}
+            if extract_policy_tags_from_catalog:
+                all_resource_names: Set[str] = set()
+                for column in raw_rows:
+                    policy_tag_column = getattr(column, "policy_tags", None)
+                    if policy_tag_column:
+                        try:
+                            all_resource_names.update(policy_tag_column)
+                        except Exception:
```
treff7es: It would be nice to capture the exception as well in the log

treff7es (approved):

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_schema.py
sgomezvillamor: addressed in commit b096f101d601e1db735eab47766bdb3c766b712b

PR #17051, opened 2026-04-16
feat(ingest/informatica): add Informatica Cloud (IDMC) connector

alfiyas-datahub:
## Summary

Adds a new ingestion source for **Informatica Cloud (IDMC)** — the cloud ETL/data integration platform. This is Phase 1 (scaffold + core extraction), covering:

- **Authentication:** v2 login with automatic session re-auth on 401 and exponential backoff on 429/5xx
- **Container hierarchy:** IDMC Projects and Folders as DataHub Containers
- **Taskflows → DataFlow:** Orchestration pipelines extracted via v3 Objects API
- **Mappings & Mapping Tasks → DataJob:** Transformation definitions with v2↔v3 cross-referencing
- **Table-level lineage:** Source DB → Informatica Mapping → Target DB via v3 Export API (async ZIP download, nested `@3.bin` JSON parsing)
- **Filtering:** Three layers — project/folder patterns, mapping/taskflow patterns, tag-based filtering
- **Connection resolution:** 14-entry `CONNECTION_TYPE_MAP` mapping IDMC connection types to DataHub platforms (Snowflake, Oracle, Postgres, etc.) with manual override support
- **Stateful ingestion:** Stale entity removal across incremental runs
Screenshot 2026-04-23 at 12 26 32 PM
Screenshot 2026-04-23 at 12 25 22 PM


metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
```diff
@@ -0,0 +1,513 @@
...
+            )
+
+        logger.info(
+            "IDMC login successful. Base URL: %s, Org: %s",
+            self._base_url,
+            data.get("orgId", "unknown"),
+        )
+
+    def _ensure_authenticated(self) -> None:
+        if not self._session_id:
```
alokr-dhub: This only checks is session_id exists. sessions can get stale. I think we can use Endpoint: POST /api/v2/user/validSessionId to check if the session is valid.

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
```diff
@@ -0,0 +1,513 @@
...
+        *,
+        is_v3: bool = False,
+        params: Optional[Dict[str, Any]] = None,
+        json_body: Optional[Any] = None,
+        stream: bool = False,
+        _retry_count: int = 0,
+    ) -> requests.Response:
+        """Execute an authenticated HTTP request with retry and re-auth logic."""
+        self._ensure_authenticated()
+        assert self._session_id is not None
```
alokr-dhub: avoid asserts in code. this can be handled in `_ensure_authenticated`. log auth error if session_id is still none after login

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
```diff
@@ -0,0 +1,1004 @@
...
+        elif conn.db_schema:
+            parts.append(conn.db_schema)
+        parts.append(table.table_name)
+        return str(
+            DatasetUrn.create_from_ids(
+                platform_id=platform,
+                table_name=".".join(parts),
+                env=self.config.env,
+            )
+        )
```
treff7es, line 1311: 🔴 **BLOCKER — cross-connector lineage footgun.** The emitted URN uses IDMC's exact case (e.g. `snowflake,DWH.ANALYTICS.ORDERS,PROD`), but the DataHub Snowflake connector emits URNs in lowercase by default (`snowflake,dwh.analytics.orders,PROD`). Every lineage edge to Snowflake will dangle unless the user explicitly sets `convert_urns_to_lowercase: false` on their Snowflake source — which almost nobody will know to do.

**Suggested fix:** add a `convert_urns_to_lowercase: bool = True` option (to match Snowflake/Postgres/BigQuery defaults), or better, a per-platform normalization map so Oracle stays uppercase while Snowflake is lowercased.

This is probably the most important thing to fix — the entire point of the connector is cross-source lineage and it's broken for the common case.

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
```diff
@@ -0,0 +1,1004 @@
...
+            parts.append(table.schema_name)
+        elif conn.db_schema:
+            parts.append(conn.db_schema)
+        parts.append(table.table_name)
+        return str(
+            DatasetUrn.create_from_ids(
+                platform_id=platform,
+                table_name=".".join(parts),
+                env=self.config.env,
+            )
```
treff7es, line 1310: 🔴 **BLOCKER — no `platform_instance` threading.** `DatasetUrn.create_from_ids` is called without `platform_instance`. If a user has ingested Snowflake with `platform_instance: prod_sf`, the Snowflake URN is `snowflake,prod_sf.dwh.analytics.orders,PROD` — this connector will emit `snowflake,DWH.ANALYTICS.ORDERS,PROD` and the edge dangles.

**Suggested fix:** add config:
```python
connection_to_platform_instance: Dict[str, str] = Field(
    default={},
    description="Map IDMC connection ID → DataHub platform_instance for upstream URNs.",
)
```
See Fivetran's `destination_platform_instance_map` for precedent.

metadata-ingestion/docs/sources/informatica/informatica_pre.md
```diff
@@ -0,0 +1,65 @@
...
+#### Concept Mapping
+
+| IDMC concept  | DataHub entity                        | Subtype        |
+| ------------- | ------------------------------------- | -------------- |
+| Project       | Container                             | `Project`      |
+| Folder        | Container                             | `Folder`       |
+| Taskflow      | DataFlow                              | `Taskflow`     |
+| Mapping (v3)  | DataJob                               | `Mapping`      |
+| Mapping Task  | DataJob                               | `Mapping Task` |
+| Source/target | Dataset (upstream/downstream lineage) | —              |
```
treff7es: 🔴 **BLOCKER — concept mapping table is wrong.** The code emits:
- **Mapping** → `DataFlow` + inner `DataJob` pair (not just a DataJob)
- **Mapping Task** → `DataFlow` only (no DataJob, despite what this table says)

Please update this table to reflect the actual output, or refactor the code to match the table. Users reading the docs will be confused by what they see in DataHub.

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
```diff
@@ -0,0 +1,1004 @@
...
+            subtype="Mapping Task",
+        )
+        parent_path = mt.path or self._v2_id_to_path.get(mt.mapping_id, "")
+        parent_key = self._container_parent_key(parent_path)
+        if parent_key is not None:
+            flow._set_container(parent_key)
+        browse_entries = self._mapping_task_browse_entries(mt)
+        if browse_entries:
+            flow._set_aspect(BrowsePathsV2Class(path=browse_entries))
+        return flow
```
treff7es: 🔴 **BLOCKER — Mapping Task emitted as DataFlow-only cannot carry lineage.** In DataHub, `dataJobInputOutput` can only be attached to a `DataJob`. A Mapping Task runs a specific Mapping (`mt.mapping_id`) but the relationship is expressed only as a string in `customProperties` — there's no typed lineage edge from the MTT to the mapping it runs, and if this connector ever adds execution-based lineage, there's no DataJob to attach it to.

**Suggested fix:** emit the Mapping Task as a `DataFlow + DataJob` pair like `_emit_mapping_entities`, and wire `dataJobInputOutput.inputDatajobs = [mapping_datajob_urn]` so the UI shows the task-to-mapping relationship.

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
```diff
@@ -0,0 +1,1004 @@
...
+        # parent mapping's browse path when their own is absent.
+        self._v2_id_to_path: Dict[str, str] = {}
+        # Track emitted containers so filtered segments fall back to plain-name
+        # browse-path entries instead of dangling URN references.
+        self._emitted_project_names: Set[str] = set()
+        self._emitted_folder_keys: Set[Tuple[str, str]] = set()
+
+    @classmethod
+    def create(cls, config_dict: Dict, ctx: PipelineContext) -> "InformaticaSource":
+        config = InformaticaSourceConfig.parse_obj(config_dict)
```
treff7es: 🔴 **BLOCKER — Pydantic v1 `parse_obj` is deprecated.** DataHub standard requires `model_validate`. `test_config.py` already uses the correct form, so this looks like a missed migration.

```suggestion
        config = InformaticaSourceConfig.model_validate(config_dict)
```

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
```diff
@@ -0,0 +1,702 @@
...
+                json_body,
+                stream,
+            )
+            if self._is_auth_failure(resp):
+                self.report.failure(
+                    title="IDMC authentication rejected after retry",
+                    message="The server refused the request even after re-authentication. Check credentials and token permissions.",
+                    context=f"{method} {url}",
+                )
+        return resp
```
treff7es, line 281: 🔴 **Auth retry logic double-reports and misclassifies.** After a second auth rejection this calls `self.report.failure(...)` and then returns the 4xx response. Callers call `resp.raise_for_status()`, which raises an HTTPError. In `list_objects` that exception is then caught by `_safe_list` and downgraded to a `warning`. Net effect: a systemic auth failure becomes one `failure` + many per-endpoint `warnings`, making the root cause hard to find.

**Suggested fix:** raise `InformaticaLoginError` on the second failure instead of calling `failure()` + returning. Let `get_workunits_internal` abort cleanly at the top.

metadata-ingestion/tests/unit/informatica/test_client.py
```diff
@@ -0,0 +1,633 @@
...
+        assert InformaticaClient._is_auth_failure(_build_response(200)) is False
+
+
+class TestDecodeJson:
+    def test_raises_informatica_api_error_on_invalid_json(self):
+        # 200 with a non-JSON body (e.g., HTML error page, empty response)
+        # must be converted to InformaticaApiError so the pipeline reports
+        # a clean warning instead of crashing with JSONDecodeError.
+        resp = MagicMock()
+        resp.json.side_effect = __import__("json").JSONDecodeError("x", "", 0)
```
treff7es: 🔴 **BLOCKER — late/inline import** (`__import__("json")`). `json` is already imported at top of file. Per `testing.md` §'Late/Inline Imports', late imports are an absolute blocker.

```suggestion
                        raise json.JSONDecodeError("bad", "{", 0)
```

metadata-ingestion/tests/unit/informatica/test_client.py
```diff
@@ -0,0 +1,633 @@
...
+                    ],
+                )
+            return _build_response(404)
+
+        # Override raise_for_status to actually raise on 404 so the caller
+        # exception path runs.
+        def request_wrapper(method, url, *args, **kwargs):
+            resp = fake_request(method, url, *args, **kwargs)
+            if resp.status_code >= 400:
+                import requests
```
treff7es: 🔴 **BLOCKER — late/inline `import requests` inside a test function.** Move to the top of the file. Per `testing.md` §'Late/Inline Imports', late imports are an absolute blocker.

metadata-ingestion/tests/unit/informatica/test_client.py
```diff
@@ -0,0 +1,633 @@
...
+        assert mapplets[0].path == "Default/develop_test/Mapplet1"
+
+    def test_returns_empty_when_all_probes_fail(self, client):
+        client._session_id = "s"
+        client._base_url = "https://pod/saas"
+        client._session_last_validated_at = 1e12
+
+        def always_404(method, url, *args, **kwargs):
+            resp = _build_response(404)
+            import requests
```
treff7es: 🔴 **BLOCKER — another late/inline `import requests`.** Move to the top of the file.

metadata-ingestion/tests/unit/informatica/test_source.py
```diff
@@ -0,0 +1,983 @@
...
+            m.assert_not_called()
+
+
+class TestSourceLifecycle:
+    def test_stale_entity_removal_handler_registered(self):
+        # Stateful ingestion hinges on StaleEntityRemovalHandler being in the
+        # workunit processor chain. Verify the returned workunit_processor is
+        # actually registered — a refactor that calls ``create`` but drops the
+        # result would silently leave deleted IDMC entities live in DataHub.
+        from datahub.ingestion.source.state import stale_entity_removal_handler
```
treff7es: 🔴 **BLOCKER — late/inline import of `stale_entity_removal_handler`.** Move to the top of the file.

metadata-ingestion/src/datahub/ingestion/source/informatica/config.py
```diff
@@ -0,0 +1,346 @@
...
+        default=True,
+        description=(
+            "Whether to extract ownership from IDMC object createdBy/updatedBy fields."
+        ),
+    )
+
+    extract_tags: bool = Field(
+        default=True,
+        description="Whether to extract tags from IDMC objects.",
+    )
```
treff7es, line 249: 🟡 **`extract_tags: True` default does nothing.** `informatica_post.md` says 'Tag extraction not yet wired — the `extract_tags` flag is accepted for forward compatibility but tags are not emitted in this release.' A `True` default that silently does nothing is misleading.

Options:
1. Default to `False` until implemented, or
2. Wire it up now — `IdmcObject.tags` is already parsed; just emit a `GlobalTagsClass` aspect per entity.
3. Make it a `HiddenFromDocs` field until the feature ships.

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
```diff
@@ -0,0 +1,1004 @@
...
+        mapping_path = self._v2_id_to_path.get(mt.mapping_id)
+        if mapping_path:
+            return self._browse_path_entries(mapping_path)
+        return []
+
+    def _extract_lineage(self) -> Iterable[Union[MetadataWorkUnit, Entity]]:
+        self._load_connections()
+        if not self._mapping_ids:
+            logger.info("No mappings to extract lineage for.")
+            return
```
treff7es, line 972: 🟡 **Lineage phase should abort when connections fail to load.** `_load_connections` calls `report.failure(...)` on error but control returns here and the export-batch loop runs anyway. Every table lookup returns `None` → every mapping emits empty lineage → the connector spends minutes submitting export jobs that produce nothing.

**Suggested fix:** after `self._load_connections()`, add:
```python
if not self._connections_by_id:
    logger.warning("No connections loaded; skipping lineage extraction.")
    return
```

metadata-ingestion/src/datahub/ingestion/source/informatica/models.py
```diff
@@ -0,0 +1,402 @@
...
+
+
+class IdmcMapping(BaseModel):
+    """v2 mapping metadata (``/api/v2/mapping``)."""
+
+    model_config = _IDMC_MODEL_CONFIG
+
+    v2_id: str
+    name: str
+    asset_frs_guid: str  # v3 GUID used for cross-referencing
```
treff7es: 🟡 **Type ambiguity: `v2_id` vs `asset_frs_guid`.** Both are `str`; `v2_id` is the v2 REST ID, `asset_frs_guid` is the v3 GUID used as the canonical join key (see `source.py:482,615`). A transposed lookup silently returns `None`.

**Suggested fix:** `NewType("V3Guid", str)` + `NewType("V2MappingId", str)` catches mix-ups at type-check time with zero runtime cost. Alternative: rename `v2_id` → `v2_id_legacy` and document the canonical key in the docstring.

metadata-ingestion/src/datahub/ingestion/source/informatica/models.py
```diff
@@ -0,0 +1,402 @@
...
+
+class IdmcConnection(BaseModel):
+    """v2 connection metadata (``/api/v2/connection``)."""
+
+    model_config = _IDMC_MODEL_CONFIG
+
+    id: str
+    name: str
+    conn_type: str  # connParams["Connection Type"] — the platform signal
+    base_type: str = ""
```
treff7es, line 257: 🟡 **Same type-ambiguity issue as above: `id` vs `federated_id`.** Both `str`, both keyed into separate dicts in `source.py` (`_connections_by_id` vs `_connections_by_fed_id`). Also, `federated_id: str = ""` makes 'missing' and 'empty' indistinguishable and could collide with another empty key in the lookup dict. Prefer `Optional[str] = None` and enforce non-empty in the code that uses it as a lookup key.

metadata-ingestion/tests/unit/informatica/test_models.py
```diff
@@ -0,0 +1,241 @@
...
+
+class TestExportJobState:
+    def test_from_api_value_valid(self):
+        assert ExportJobState.from_api_value("SUCCESSFUL") == ExportJobState.SUCCESSFUL
+        assert ExportJobState.from_api_value("FAILED") == ExportJobState.FAILED
+
+    def test_from_api_value_unknown(self):
+        assert ExportJobState.from_api_value(None) == ExportJobState.UNKNOWN
+        assert ExportJobState.from_api_value("") == ExportJobState.UNKNOWN
+        assert ExportJobState.from_api_value("BOGUS") == ExportJobState.UNKNOWN
```
treff7es: 🟡 **Trivial test — should be deleted.** `test_from_api_value_valid` / `test_from_api_value_unknown` only exercise a one-line enum lookup that's already covered indirectly by `test_client.py:458` (`wait_for_export` returns UNKNOWN for bogus state) and the integration test. Per `testing.md` §'Tests That Must NOT Be Written', trivial enum-construction tests are an absolute blocker.

metadata-ingestion/tests/unit/informatica/test_models.py
```diff
@@ -0,0 +1,241 @@
...
+
+    def test_report_connection_unresolved(self):
+        report = InformaticaSourceReport()
+        report.report_connection_unresolved("cid", "cname", "no platform")
+        assert any("cname" in e and "cid" in e for e in report.connections_unresolved)
+
+    def test_report_export_failed(self):
+        report = InformaticaSourceReport()
+        report.report_export_failed("job-1", "timed out")
+        assert any("job-1" in e for e in report.export_jobs_failed)
```
treff7es: 🟡 **Entire `TestInformaticaSourceReport` class should be deleted.** These tests verify that `counter += 1` increments and `list.append(x)` appends. The behaviors are already asserted implicitly by `test_client.py:189,423,456,472`. Per `testing.md`, tests of trivial getters/counter-increments are absolute blockers.

metadata-ingestion/tests/unit/informatica/test_config.py
```diff
@@ -0,0 +1,70 @@
...
+        assert config.connection_type_overrides == {"01B": "snowflake"}
+
+    def test_tag_and_pattern_filters(self):
+        config = InformaticaSourceConfig.model_validate(
+            _base_config(
+                tag_filter_names=["pii", "critical"],
+                project_pattern={"allow": ["Prod.*"]},
+            )
+        )
+        assert config.tag_filter_names == ["pii", "critical"]
```
treff7es: 🟡 **Trivial round-trip tests.** `test_connection_type_overrides_accepts_known_platform` and `test_tag_and_pattern_filters` only assert that a dict round-trips through Pydantic — no validator logic. The `rejects_empty` and `warns_on_unknown` tests already cover the validator. Consider deleting both, or replacing `test_tag_and_pattern_filters` with a behavior test that actually exercises tag filtering in `_iter_with_tags`.

metadata-ingestion/tests/unit/informatica/test_source.py
```diff
@@ -0,0 +1,983 @@
...
+        )
+        assert urn is not None
+        assert "PUBLIC.T" in urn
+
+    def test_unknown_connection_reports_and_returns_none(self):
+        source = _make_source()
+        urn = source._resolve_table_to_dataset_urn(
+            LineageTable(table_name="T", connection_federated_id="missing")
+        )
+        assert urn is None
```
treff7es: 🟡 **Missing high-value test: cross-connector URN format.** The single most likely production failure (URN case mismatch with Snowflake, missing `platform_instance` — see the two blocker comments on `source.py`) has no test. Please add:

1. A test that sets `platform_instance` on the source and verifies it is threaded into the upstream dataset URN.
2. A test asserting the emitted URN matches what the Snowflake connector produces by default (lowercase) — or documents the configuration needed to make them agree.

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
```diff
@@ -0,0 +1,1004 @@
...
+                aspect=UpstreamLineageClass(
+                    upstreams=[
+                        UpstreamClass(
+                            dataset=src_urn,
+                            type=DatasetLineageTypeClass.TRANSFORMED,
+                        )
+                        for src_urn in input_urns
+                    ]
+                ),
+            ).as_workunit()
```
treff7es, line 1258: 🟡 **Duplicate lineage emission.** Emits both `DataJobInputOutputClass` on the inner `transform` DataJob AND `UpstreamLineageClass` on the downstream dataset. DataHub normally derives dataset-level upstream lineage from DataJob I/O automatically. Either pick one, or add a comment explaining why both are needed (e.g., if derivation isn't reliable for cross-platform downstreams here).

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
```diff
@@ -0,0 +1,1004 @@
...
+
+@platform_name("Informatica")
+@config_class(InformaticaSourceConfig)
+@support_status(SupportStatus.INCUBATING)
+@capability(SourceCapability.PLATFORM_INSTANCE, "Enabled by default")
+@capability(SourceCapability.CONTAINERS, "Projects and folders as containers")
+@capability(SourceCapability.LINEAGE_COARSE, "Table-level lineage via v3 Export API")
+@capability(SourceCapability.DELETION_DETECTION, "Via stateful ingestion")
+@capability(SourceCapability.OWNERSHIP, "From IDMC object createdBy/updatedBy")
+class InformaticaSource(StatefulIngestionSourceBase):
```
treff7es: 🟠 **Missing `TestableSource` mix-in.** Best practice for new API-based connectors is to implement `test_connection` so users can validate credentials in the UI before running full ingestion. See `api.md` §Step 3 — implementing `login()` already gets you 90% of the way there; just wrap it in a `TestConnectionReport`.

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
```diff
@@ -0,0 +1,702 @@
...
+                schema_name=obj.get("dbSchema") or "",
+                connection_federated_id=fed_id,
+                transformation_name=tx.get("name", ""),
+            )
+            tx_class = tx.get("$$class")
+            tx_name = tx.get("name", "").lower()
+            if tx_class == 9 or "source" in tx_name:
+                sources.append(lt)
+            elif tx_class == 8 or "target" in tx_name:
+                targets.append(lt)
```
treff7es, line 870: 🟠 **Magic numbers for transformation class.** `tx_class == 9` means 'source' and `tx_class == 8` means 'target'. A new reader has no way to verify that; a typo silently breaks lineage.

**Suggested fix:**
```python
class TxClass(IntEnum):
    TARGET = 8
    SOURCE = 9
```
Then `tx_class == TxClass.SOURCE` etc.

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
```diff
@@ -0,0 +1,702 @@
...
+                    inner_zip.close()
+                except Exception as e:
+                    self.report.warning(
+                        title="Failed to parse IDMC mapping export entry",
+                        message="Lineage for this mapping will be missing.",
+                        context=f"job_id={job_id}, entry={entry_name}",
+                        exc=e,
+                    )
+                    self.report.report_object_failed(entry_name, str(e))
+            outer_zip.close()
```
treff7es: 🟠 **ZIP file handle leak on exception.** `outer_zip` is opened manually and `.close()` is only called at the end. If an exception escapes the per-entry loop (e.g. an unexpected error during `self._resolve_entry_mapping_id`), the outer ZIP file handle leaks.

**Suggested fix:**
```python
with tempfile.NamedTemporaryFile(suffix=".zip") as tmp:
    ...
    with zipfile.ZipFile(tmp) as outer_zip:
        for idx, entry_name in enumerate(dtemplate_entries):
            with zipfile.ZipFile(io.BytesIO(outer_zip.read(entry_name))) as inner_zip:
                ...
```
Eliminates manual `.close()` and the leak on exception.

metadata-ingestion/docs/sources/informatica/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+| Source Concept  | DataHub Concept                                           | Notes                       |
+| --------------- | --------------------------------------------------------- | --------------------------- |
+| `"informatica"` | [Data Platform](../../metamodel/entities/dataPlatform.md) |                             |
+| Project         | [Container](../../metamodel/entities/container.md)        | SubType `"Project"`         |
+| Folder          | [Container](../../metamodel/entities/container.md)        | SubType `"Folder"`          |
+| Taskflow        | [DataFlow](../../metamodel/entities/dataFlow.md)          | SubType `"Taskflow"`        |
+| Mapping         | [DataJob](../../metamodel/entities/dataJob.md)            | SubType `"Mapping"`         |
+| Mapping Task    | [DataJob](../../metamodel/entities/dataJob.md)            | SubType `"Mapping Task"`    |
+| Source/Target   | [Dataset](../../metamodel/entities/dataset.md)            | Upstream/downstream lineage |
```
treff7es: 🔴 **Second copy of the wrong concept-mapping table.** First review's C3 flagged the same issue in `informatica_pre.md`. This file (`README.md`) is what gets published on the docs site and has the same problem: it says **Mapping → DataJob** and **Mapping Task → DataJob**, but the code emits Mappings as a `DataFlow + DataJob` pair and Mapping Tasks as `DataFlow` only.

Please keep `informatica_pre.md` and this file in sync when you update the table.

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
treff7es, line 1311: 🔴 **No guard against empty `table.table_name`.** `LineageTable.table_name` is populated from `obj.get("name") or obj.get("objectName") or obj.get("label") or ""` in `client.py:663-666`, so malformed IDMC transformations silently produce `table_name=""`. This code then builds a URN like `urn:li:dataset:(urn:li:dataPlatform:snowflake,DWH.SALES.,PROD)` — with a trailing dot and an empty table segment — and the lineage edge points at nothing.

**Suggested fix:**
```python
if not table.table_name:
    self.report.report_connection_unresolved(
        connection_id=conn.id,
        connection_name=conn.name,
        reason="Empty table name in IDMC transformation",
    )
    return None
```
...before the `parts.append(table.table_name)` line.

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
```diff
@@ -0,0 +1,1004 @@
...
+            parent_project_name or ORPHAN_PROJECT_SENTINEL, folder.name
+        )
+        return Container(
+            container_key=key,
+            display_name=folder.name,
+            description=folder.description,
+            subtype="Folder",
+            owners=self._owner_list(folder.updated_by or folder.created_by),
+            parent_container=parent_key,
+        )
```
treff7es, line 433: 🔴 **`project_pattern` doesn't cascade to folder's `parent_container` reference.** `parent_key` is built unconditionally from the resolved project name — it does NOT check `_emitted_project_names`. So if the user sets `project_pattern: allow=["Sales"]` and the org also has a `Marketing` project, the folder `Marketing/SubFolder` is still emitted with `parent_container = self._project_key("Marketing")` — a URN that was never emitted. The folder shows up in DataHub referencing a ghost container.

`_container_parent_key` (line 380-397) has the right check for flows/jobs; the same check should apply here.

**Suggested fix:**
```python
parent_key: Optional[ContainerKey] = (
    self._project_key(parent_project_name)
    if parent_project_name
    and parent_project_name in self._emitted_project_names
    else None
)
```

Also consider skipping folders whose parent project was filtered out — right now `_extract_containers:189-223` emits them with no parent, creating orphans. `test_project_pattern_filters_out_non_matching` doesn't cover this path (it only emits Projects, not Folders).

metadata-ingestion/tests/unit/informatica/test_client.py
```diff
@@ -0,0 +1,633 @@
...
+            200, {"icSessionId": "new", "serverUrl": "https://pod/saas"}
+        )
+
+        with (
+            patch.object(client._session, "request", side_effect=[fail1, fail2]),
+            patch.object(client._session, "post", return_value=login_resp),
+        ):
+            client._request("GET", "https://pod/saas/api/v2/thing")
+
+        assert len(client.report.failures) > 0
```
treff7es: 🟡 **This test locks in the anti-pattern from the first review's C7.** `_request` calling `report.failure(...)` and returning the 4xx response is the bug: callers then hit `raise_for_status()`, `_safe_list` downgrades to a warning, and the systemic auth failure is misclassified.

When you fix C7 to raise `InformaticaLoginError` on the second failure, this test will start failing. Please update it then to:

```python
with (
    patch.object(client._session, "request", side_effect=[fail1, fail2]),
    patch.object(client._session, "post", return_value=login_resp),
    pytest.raises(InformaticaLoginError),
):
    client._request("GET", "https://pod/saas/api/v2/thing")
```

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
```diff
@@ -0,0 +1,1004 @@
...
+        if obj.name == "Add-On Bundles":
+            return True
+        return any(seg == "Add-On Bundles" for seg in obj.path.split("/") if seg)
+
+    def _owner_list(
+        self, user_identifier: Optional[str]
+    ) -> Optional[List[CorpUserUrn]]:
+        if not user_identifier or not self.config.extract_ownership:
+            return None
+        return [CorpUserUrn(user_identifier)]
```
treff7es: 🟠 **Only one owner ever extracted.** Call sites pass `updated_by OR created_by` (e.g. `source.py:261,277,451,647,667,706`), so if an entity has both a creator and a later updater, only the updater is kept. When an entity has never been updated, only the creator is kept. Historical ownership info is dropped.

Other DataHub connectors (e.g. BigQuery, Tableau) emit both with different `ownerType`s (`DATAOWNER` for creator, `TECHNICAL_OWNER` for last updater, or similar).

**Suggested fix:** change `_owner_list` signature to accept multiple identifiers with types, e.g.:
```python
def _owner_list(
    self,
    *identifiers: Tuple[Optional[str], OwnershipTypeClass],
) -> Optional[List[OwnerClass]]:
```
...and pass `(created_by, DATAOWNER)` and `(updated_by, TECHNICAL_OWNER)` from the callers.

treff7es: ℹ️ **Nit:** `CorpUserUrn("alice@acme.com")` produces `urn:li:corpuser:alice@acme.com`, with the `@acme.com` in the username part of the URN. Some DataHub sources (e.g. Snowflake, Okta) strip the domain to get `urn:li:corpuser:alice`. Worth confirming this matches how user entities are ingested elsewhere in your deployment — otherwise ownership edges point at non-existent users.

Consider either using `make_user_urn(user_identifier)` (which handles both forms) or documenting that downstream Okta/AzureAD sources must be configured to preserve full emails.

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
```diff
@@ -0,0 +1,702 @@
...
+            method="GET",
+            url=f"{self._base_url}/api/v2/mapping",
+            status=200,
+            item_count=len(mappings),
+            extra="v2-mappings",
+        )
+        return mappings
+
+    def get_mapping(self, v2_id: str) -> IdmcMapping:
+        return IdmcMapping.from_api_response(self._get_v2(f"/api/v2/mapping/{v2_id}"))
```
treff7es: 🟠 **Dead code.** `get_mapping` (single-mapping fetch) has zero call sites in `source.py` or `tests/`. Same for `get_connection` (line 456-459). Either wire them in if intended for future use, or delete to keep the API surface minimal.

metadata-ingestion/src/datahub/ingestion/source/informatica/models.py
```diff
@@ -0,0 +1,402 @@
...
+            name=data.get("name", ""),
+            description=data.get("description") or None,
+            mapping_id=data.get("mappingId", ""),
+            mapping_name=data.get("mappingName", ""),
+            connection_id=data.get("connectionId", ""),
+            created_by=data.get("createdBy") or None,
+            updated_by=data.get("updatedBy") or None,
+            create_time=data.get("createTime") or None,
+            update_time=data.get("updateTime") or None,
+        )
```
treff7es: 🟠 **Dead code / orphan tests.** `IdmcMappingTask.from_api_response` and `from_v3_object` are not called from `source.py` or `client.py` (`list_mapping_tasks` uses `IdmcMappingTask.from_idmc_object(obj)` instead). But they have tests in `test_models.py:168-196` — so we're testing code that's not used in production.

Either wire these into `client.py` (there was presumably a reason they were written) or delete both the methods and their tests.

metadata-ingestion/tests/unit/informatica/test_source.py
```diff
@@ -0,0 +1,983 @@
...
+    def test_make_folder_container_has_parent_key(self):
+        source = _make_source()
+        folder = IdmcObject(
+            id="f1",
+            name="Mappings",
+            path="/Explore/Sales/Mappings",
+            object_type="Folder",
+        )
+        container = source._make_folder_container(folder)
+        assert isinstance(container, Container)
```
treff7es: 🟠 **Trivial tests — should be deleted.** `test_make_project_container` only asserts `isinstance(container, Container)` + URN starts with `urn:li:container:`. `test_make_folder_container_has_parent_key` only asserts `isinstance(container, Container)` — doesn't even verify the parent key is set, despite what the test name promises.

Per `testing.md` §"Tests That Must NOT Be Written", tests that verify a constructor returns a properly-typed object are testing the framework. If you want to keep the folder test, make it assert the actual behavior:
```python
assert container.parent_container == source._project_key("Sales").as_urn()
```

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
```diff
@@ -0,0 +1,702 @@
...
+                return
+            dtemplate_entries = [
+                n for n in outer_zip.namelist() if n.endswith(".DTEMPLATE.zip")
+            ]
+            for idx, entry_name in enumerate(dtemplate_entries):
+                mapping_id = self._resolve_entry_mapping_id(
+                    entry_name, submitted_ids, idx
+                )
+                try:
+                    inner_zip = zipfile.ZipFile(io.BytesIO(outer_zip.read(entry_name)))
```
treff7es: 🟠 **No zip-bomb protection.** Each export entry is fully read into memory (`outer_zip.read(entry_name)`) then opened as an in-memory ZIP. A malicious (or malformed) IDMC response could claim a small compressed size but decompress to gigabytes.

Lower priority because IDMC is trusted infrastructure, but for robustness consider:
- Check `ZipInfo.file_size` before reading (cap at e.g. 100MB)
- Use `zipfile.ZipFile.infolist()` to inspect sizes first

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
```diff
@@ -0,0 +1,702 @@
...
+
+        Prefer a substring match against the entry name (more robust against
+        server-side ordering changes); fall back to positional alignment.
+        """
+        for sid in submitted_ids:
+            if sid and sid in entry_name:
+                return sid
+        if 0 <= index < len(submitted_ids):
+            return submitted_ids[index]
+        return ""
```
treff7es, line 801: 🟠 **Positional fallback can assign the same id twice.** If the export contains 3 entries with `submitted_ids=[A, B, C]`:
- Entry 0 has "A" in name → matches A ✅
- Entry 1 has no substring match → positional fallback returns   `submitted_ids[1]` = **B**
- Entry 2 also has "A" in name (e.g. duplicate GUID in   filename) → matches A again 🚩

Result: two lineage records for mapping A, zero for C. Unlikely in practice but a real bug.

**Suggested fix:** track assigned ids in a set and skip them on substring match, so no id is used twice; if the fallback would reuse an already-assigned id, return `""` instead.

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
```diff
@@ -0,0 +1,702 @@
...
+    ):
+        self.config = config
+        self.report = report
+        self._session = requests.Session()
+        adapter = HTTPAdapter(max_retries=_RETRY_POLICY)
+        self._session.mount("https://", adapter)
+        self._session.mount("http://", adapter)
+        self._session_id: Optional[str] = None
+        self._base_url: Optional[str] = None
+        self._session_last_validated_at: Optional[float] = None
```
treff7es, line 103: 🟠 **Missing `close()` override leaks `requests.Session`.** `InformaticaClient` creates `self._session = requests.Session()` here but never closes it, and `InformaticaSource` doesn't override `close()` either, so the session stays open even after ingestion completes. In long-running DataHub pipelines (or tests that instantiate many sources), this leaks sockets / file descriptors.

**Suggested fix:** add to `InformaticaClient`:
```python
def close(self) -> None:
    self._session.close()
```
and to `InformaticaSource`:
```python
def close(self) -> None:
    self.client.close()
    super().close()
```

metadata-ingestion/tests/unit/informatica/test_models.py
```diff
@@ -0,0 +1,241 @@
...
+        )
+        assert conn.host == "legacy-host"
+        assert conn.database == "DB"
+        assert conn.db_schema == "S"
+        assert conn.conn_type == ""
+
+    def test_from_api_response_handles_missing_fields(self):
+        conn = IdmcConnection.from_api_response({})
+        assert conn.id == ""
+        assert conn.name == ""
```
treff7es: 🟠 **Another trivial test — delete.** Tests Pydantic's default behavior when given an empty dict. Same anti-pattern as the other trivial tests flagged in earlier rounds. Doesn't cover any business logic (`from_api_response` with a populated dict is already tested in `test_from_api_response_reads_conn_params`).

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
```diff
@@ -0,0 +1,1004 @@
...
+            for k, v in (
+                ("idmcId", tf.id),
+                ("path", tf.path),
+                ("createdBy", tf.created_by or ""),
+                ("updatedBy", tf.updated_by or ""),
+                ("createTime", tf.create_time or ""),
+                ("updateTime", tf.update_time or ""),
+            )
+            if v
+        }
```
treff7es, line 692: ℹ️ **Inconsistent custom-property keys across entity types.** Mappings emit `v3Id`, `objectType`, `path` (`source.py:628-632`). Taskflows emit `idmcId`, `path` but NOT `v3Id`. Mapping tasks emit both `idmcId` AND `v2Id` (same value). Users filtering or searching custom properties across entity types will be surprised by these differences. Consider standardizing on a common set: `v3Id`, `v2Id`, `objectType`, `path`, `createdBy`, `updatedBy`, `createTime`, `updateTime` — emit N/A when not applicable.

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
```diff
@@ -0,0 +1,702 @@
...
+            IdmcMapping.from_api_response(m) for m in items if isinstance(m, dict)
+        ]
+        self.report.report_api_response(
+            method="GET",
+            url=f"{self._base_url}/api/v2/mapping",
+            status=200,
+            item_count=len(mappings),
+            extra="v2-mappings",
+        )
+        return mappings
```
treff7es, line 432: 🔴 **Performance: `list_mappings` fetches ALL mappings in a single unpaginated call.** For an org with 50k+ mappings this likely:
- exceeds the hardcoded 60s request timeout (`_do_send` line 237)
- generates a multi-hundred-MB response
- blocks the entire run on one slow call that can't resume

Please verify whether IDMC's `/api/v2/mapping` supports `?offset=` / `?limit=` query params (most v2 list endpoints do). If so, paginate. If not, document the hard scale ceiling in `informatica_post.md` and consider falling back to the v3 DTEMPLATE query for the cache warm-up (which IS paginated).

The same issue applies to `list_connections` at line 441 — single unpaginated call. Smaller magnitude for most orgs but same failure mode on the largest deployments.

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
```diff
@@ -0,0 +1,702 @@
...
+            IdmcConnection.from_api_response(c) for c in items if isinstance(c, dict)
+        ]
+        self.report.report_api_response(
+            method="GET",
+            url=f"{self._base_url}/api/v2/connection",
+            status=200,
+            item_count=len(connections),
+            extra="v2-connections",
+        )
+        return connections
```
treff7es, line 447: 🔴 **Performance: `list_connections` is also unpaginated.** Same pattern as `list_mappings` — single v2 call, no pagination. Typically smaller scale (most orgs have <1k connections) but same failure mode on the largest deployments.

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
```diff
@@ -0,0 +1,1004 @@
...
+
+    def _tag_filters_or_none(self, object_type: IdmcObjectType) -> List[Optional[str]]:
+        # IDMC v3 rejects ``tag==`` on Project/Folder with HTTP 400;
+        # tags only apply to assets like TASKFLOW and DTEMPLATE.
+        if object_type in ("Project", "Folder"):
+            return [None]
+        if self.config.tag_filter_names:
+            tags: List[Optional[str]] = list(self.config.tag_filter_names)
+            return tags
+        return [None]
```
treff7es: 🔴 **Performance: tag filter causes N-fold API amplification.** `_iter_with_tags` runs `list_objects(type, tag=X)` once per configured tag. Inside `list_objects` (`client.py:297-346`), IDMC rejects server-side `tag==` filters, so the paginated call fetches ALL objects of that type and drops non-matching ones client-side.

Concrete impact: for 3 configured tags × 3 v3 mapping families (DTEMPLATE / MAPPLET / DMAPPLET) × `ceil(50000/200)` pages = **2,250 paginated API calls vs 750 without tag filter**. Plus 95% of the objects returned on each pass are thrown away.

**Suggested fix:** do ONE pass per type, and on the client side check whether ANY of the configured tags match:
```python
def _iter_with_tags(self, object_type: IdmcObjectType) -> Iterable[IdmcObject]:
    tags = self.config.tag_filter_names if object_type not in ("Project", "Folder") else None
    for obj in self.client.list_objects(object_type):
        if self._is_bundle_object(obj):
            ...
        if tags and not any(t in obj.tags for t in tags):
            self.report.report_filtered("tag", object_type, obj.path)
            continue
        yield obj
```
Changes complexity from `K × N` to `N`.

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
```diff
@@ -0,0 +1,1004 @@
...
+                # Per-entry parse/zip failures are reported deeper in the client;
+                # only expected IDMC/network errors are warned here so unexpected
+                # ones still propagate with a stack trace.
+                self.report.warning(
+                    title="IDMC export batch failed",
+                    message="Lineage for this batch will be missing.",
+                    context=f"batch_start={batch_start}, size={len(batch)}",
+                    exc=e,
+                )
+                self.report.export_jobs_failed.append(f"batch@{batch_start}: {e}")
```
treff7es: 🟡 **Performance: export batches run strictly sequentially.** `for batch_start in range(...)` processes each batch fully (submit → wait_for_export → download → parse → emit) before starting the next. For a 50k-mapping org with batch_size=1000, that's 50 batches × O(1-5 min) each = **1-4 hours wall-clock** for lineage alone.

IDMC's export API allows multiple concurrent jobs per account. Two practical improvements:

**Cheap:** use `ThreadPoolExecutor` to submit all batches in parallel, then poll-all-jobs in a round-robin loop, then download as each completes. Cuts wall-clock to roughly `max_job_duration + download_serialization`.

**Incremental:** just submit the NEXT batch while the current one is being downloaded. At minimum, that hides download latency behind the next submit-wait.

Add a config knob `max_concurrent_export_jobs: int = 4` so users can tune for their IDMC rate limits.

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
```diff
@@ -0,0 +1,702 @@
...
+        if json_body is not None:
+            headers["Content-Type"] = "application/json"
+        resp = self._session.request(
+            method,
+            url,
+            headers=headers,
+            params=params,
+            json=json_body,
+            stream=stream,
+            timeout=60,
```
treff7es: 🟠 **Performance: 60s request timeout is hardcoded and applies to export-ZIP downloads.** A large lineage export (hundreds of MB) may exceed 60s between chunks on a slow link, failing the batch entirely (and lineage for it is lost). The poll interval config maxes at 600s but the actual request timeout is fixed.

Suggested: add `request_timeout_secs: int = 60` to config and use a larger timeout (or `None` for read-timeout) specifically for `download_and_parse_export`'s streaming GET.

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
```diff
@@ -0,0 +1,702 @@
...
+        """Poll until complete or timeout."""
+        deadline = time.time() + self.config.export_poll_timeout_secs
+        while time.time() < deadline:
+            status = self.poll_export_job(job_id)
+            if status.state == ExportJobState.SUCCESSFUL:
+                return status
+            if status.state not in (ExportJobState.IN_PROGRESS, ExportJobState.QUEUED):
+                self.report.report_export_failed(job_id, status.message)
+                return status
+            time.sleep(self.config.export_poll_interval_secs)
```
treff7es: 🟠 **Performance: no jitter on export poll.** `time.sleep(config.export_poll_interval_secs)` is deterministic. If multiple DataHub pipelines ingest the same IDMC pod on the same cadence (common in scheduled runs), their polling calls stack up at the same moments. Adding `± 1-2 seconds` of jitter is cheap and reduces risk of hitting IDMC's rate limits:

```python
import random
time.sleep(self.config.export_poll_interval_secs + random.uniform(-1, 1))
```

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
```diff
@@ -0,0 +1,1004 @@
...
+
+    def __init__(self, config: InformaticaSourceConfig, ctx: PipelineContext):
+        super().__init__(config, ctx)
+        self.config = config
+        self.report = InformaticaSourceReport()
+        self.client = InformaticaClient(config, self.report)
+        # Lineage looks up by federated_id; config overrides look up by id.
+        self._connections_by_fed_id: Dict[str, IdmcConnection] = {}
+        self._connections_by_id: Dict[str, IdmcConnection] = {}
+        self._v2_mappings_by_guid: Dict[str, IdmcMapping] = {}
```
treff7es: 🟠 **Performance: `_v2_mappings_by_guid` held for the entire run.** The cache is populated in `_extract_mappings_and_tasks` (phase 3) and used only there — to look up `v2_id` + `valid` flag for each v3 mapping. But it's never cleared, so it stays resident through phase 4 (lineage) and entity teardown.

For a 50k-mapping org that's ~25MB of Pydantic objects held unnecessarily. Add `self._v2_mappings_by_guid.clear()` at the end of `_extract_mappings_and_tasks` (and same for `_v2_id_to_path` after mapping-task emission completes).

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
alokr-dhub: ACK. fixed at rest of the places. missed it here

metadata-ingestion/tests/unit/informatica/test_source.py
alokr-dhub: ACK

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
alokr-dhub: ACK

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
alokr-dhub: ACK. fixed this. Mapping Tasks are now modeled as DataFlow + Datajob . Also did code refactor

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
alokr-dhub: I have updated in the docstring under `_emit_downstream_upstream_lineage`. Summary: 
DataHub's UI lineage traversal has asymmetric edge-type filters: a Dataset→Dataset Lineage tab doesn't reliably traverse Consumes/Produces edges through an intermediate DataJob

alokr-dhub:
@treff7es addressed all comments. Please review it again.

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
```diff
@@ -0,0 +1,1244 @@
...
+            return {}
+        return _parse_taskflow_export_package(resp.iter_content, report=self.report)
+
+    def submit_export_job(self, object_ids: List[str]) -> str:
+        body = {
+            "objects": [
+                {"id": obj_id, "includeDependencies": False} for obj_id in object_ids
+            ]
+        }
+        data = self._post_v3("/public/core/v3/export", body)
```
puneetagarwal-datahub, line 622: The v3 /export submit is a POST; a 5xx retry after the server accepted the job may create duplicate export jobs. Check here

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
```diff
@@ -0,0 +1,1244 @@
...
+        job_id = data.get("id", "")
+        if not job_id:
+            raise InformaticaApiError(f"Export job submission returned no ID: {data}")
+        self.report.export_jobs_submitted += 1
+        logger.info("Export job submitted: %s (%d objects)", job_id, len(object_ids))
+        return job_id
+
+    def poll_export_job(self, job_id: str) -> ExportJobStatus:
+        data = self._get_v3(f"/public/core/v3/export/{job_id}")
```
puneetagarwal-datahub, line 631: handle api response gracefully . an intermittent error here might stop polling  . also check at other places . may be use try/catch

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
```diff
@@ -0,0 +1,1244 @@
...
+    def login(self) -> str:
+        """Authenticate via the v2 login endpoint. Returns the session id."""
+        resp = self._session.post(
+            f"{self.config.login_url}/ma/api/v2/user/login",
+            json={
+                "username": self.config.username,
+                "password": self.config.password.get_secret_value(),
+            },
+            headers={"Content-Type": "application/json"},
+            timeout=30,
```
puneetagarwal-datahub: timeout should be config driven ? Also do we know max timeout for each api ? Also timeouts are different in different apis here

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
alokr-dhub: if the job is accepted by the server a `200` response is expected. Even in the case job was accepted with a `500` response the retry request will be triggered until a `200` response with a valid job_id is returned. The job_id ensures  idempotency of reports here.

metadata-ingestion/src/datahub/ingestion/source/informatica/models.py
```diff
@@ -0,0 +1,476 @@
...
+
+    model_config = _IDMC_MODEL_CONFIG
+
+    v2_id: V2Id
+    name: str
+    asset_frs_guid: V3Guid
+    description: Optional[str] = None
+    created_by: Optional[str] = None
+    updated_by: Optional[str] = None
+    create_time: Optional[str] = None
```
puneetagarwal-datahub: can we validate this field to be datetime ? currently it can take any string value

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
```diff
@@ -0,0 +1,1244 @@
...
+        """
+        resp = self._request(
+            "GET",
+            f"{self._base_url}/public/core/v3/export/{job_id}/package",
+            is_v3=True,
+            stream=True,
+        )
+        resp.raise_for_status()
+        with tempfile.NamedTemporaryFile(suffix=".zip") as tmp:
+            for chunk in resp.iter_content(chunk_size=8192):
```
puneetagarwal-datahub, line 690: check max file size . The code streams unbounded bytes to /tmp until the server stops sending which might crash the pod

metadata-ingestion/src/datahub/ingestion/source/informatica/models.py
```diff
@@ -0,0 +1,476 @@
...
+    update_time: Optional[str] = None
+    document_type: Optional[str] = None
+    valid: bool = True
+    parameters: List[Dict[str, Any]] = Field(default_factory=list)
```
puneetagarwal-datahub: AGENTS.md (project root) is explicit: "  - **Type Safety**: Everything must have type annotations, avoid `Any` type, use specific types (`Dict[str, int]`, `TypedDict`)"

Can we use particular type here ?

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
alokr-dhub: ACK

metadata-ingestion/src/datahub/ingestion/source/informatica/models.py
alokr-dhub: ACK

metadata-ingestion/src/datahub/ingestion/source/informatica/models.py
alokr-dhub: updated to object

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
alokr-dhub: capped the max filesize to 500 mb

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
```diff
@@ -0,0 +1,1364 @@
...
+        self._connections_by_id: Dict[str, IdmcConnection] = {}
+        # Two indexes into v2 /api/v2/mapping — MT→Mapping refs use v2 ids,
+        # export batches use v3 GUIDs. State dict keys are plain str for
+        # test ergonomics; model-level V2Id/V3Guid NewTypes enforce the
+        # invariant on the boundary.
+        self._v2_mappings_by_guid: Dict[str, IdmcMapping] = {}
+        self._v2_mappings_by_v2_id: Dict[str, IdmcMapping] = {}
+        self._mapping_ids: List[str] = []
+        # v3 GUID → MT DataJob URNs that run that mapping (fan-out).
+        self._mapping_v3_to_mt_job_urns: Dict[str, List[str]] = {}
```
treff7es:
```suggestion
        self._v2_mappings_by_guid: Dict[V3Guid, IdmcMapping] = {}
        self._v2_mappings_by_v2_id: Dict[V2Id, IdmcMapping] = {}
        self._mapping_ids: List[V3Guid] = []
        # v3 GUID → MT DataJob URNs that run that mapping (fan-out).
        self._mapping_v3_to_mt_job_urns: Dict[V3Gui, List[str]] = {}
```

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
```diff
@@ -0,0 +1,1364 @@
...
+        # MT lookups for Taskflow step resolution — steps reference MTs by
+        # either v2 id or name, so we index both.
+        self._mt_v2_id_to_job_urn: Dict[str, str] = {}
```
treff7es:
```suggestion
        self._mt_v2_id_to_job_urn: Dict[V3Gui, V2Id] = {}
```

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
alokr-dhub: ACK. v2 APIs for connections and mappings do not support pagination. Let me document the case in docs with a suggestion to use filters in case of large connections and mappings.

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
alokr-dhub: I have verified and confirmed `list_mappings` and `list_connections` do not support pagination. I will add note in the doc to nudge for tag filtering

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
alokr-dhub: ACK

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
alokr-dhub: ack

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
alokr-dhub: I had deferred ThreadPool parallel execution for v2 . Let me get back to this

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
alokr-dhub: Ack. thanks for pointing out . missed this optimisation earlier.

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
alokr-dhub: fixed this

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
alokr-dhub: ACK

puneetagarwal-datahub (changes requested):

metadata-ingestion/docs/sources/informatica/informatica_pre.md
alokr-dhub: Updated the docs to be in sync. `_emit_mapping_task` emits DataFlow + transform DataJob for every Mapping Task, matching the table exactly. Mappings (the reusable logic definitions) are never emitted, only Mapping Tasks.

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
alokr-dhub: changed the method signature to accept created_by and updated_by

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
```diff
@@ -0,0 +1,1384 @@
...
+        if mapping_v3_guid:
+            self._mapping_v3_to_mt_job_urns.setdefault(mapping_v3_guid, []).append(
+                job_urn
+            )
+            if mapping_v3_guid not in self._mapping_ids:
+                self._mapping_ids.append(mapping_v3_guid)
+        if mt.v2_id:
+            self._mt_v2_id_to_job_urn[mt.v2_id] = job_urn
+        if mt.name:
+            self._mt_name_to_job_urn[mt.name] = job_urn
```
puneetagarwal-datahub: Two MTs with the same name in different folders (very common in IDMC — template copy-paste across project folders) collide.
 Last write wins; Taskflow step resolution at
source.py:803-804 will route to whichever MT was indexed last. 

Same problem applies to mapping_pattern.allowed(mapping_name) — patterns operate on bare name without project qualification, so two same-named mappings in different folders cannot be filtered separately.

Fix: key by (project_path, name) not name. If you must keep a name index for resolution, also keep a v2_id index and prefer it.

Check this

puneetagarwal-datahub (changes requested):

metadata-ingestion/src/datahub/ingestion/source/informatica/source.py
alokr-dhub: @puneetagarwal-datahub I have removed the mt resolution fallback from task ref step as it was redundant, since we will always have id and it's unique across projects. I have updated pattern filters to include path to solve for this edge case.

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
```diff
@@ -0,0 +1,1270 @@
...
+                dst.predecessor_step_ids.append(src_id)
+
+    return TaskflowDefinition(taskflow_id=tf_guid, taskflow_name=tf_name, steps=steps)
+
+
+def _parse_taskflow_export_package(
+    iter_content: Callable[..., Iterator[bytes]],
+    report: Optional[InformaticaSourceReport] = None,
+) -> Dict[str, TaskflowDefinition]:
+    """Stream an export package ZIP to a temp file and parse every
```
puneetagarwal-datahub, line 1221: Add size guard here do that it does not fill up pod disk

treff7es (approved):
lgtm, thanks for all the changes.

puneetagarwal-datahub (changes requested):
LGTM

metadata-ingestion/src/datahub/ingestion/source/informatica/client.py
alokr-dhub: ACK. added zip filesize cap for this report.

puneetagarwal-datahub (approved):

PR #17056, opened 2026-04-16
feat(ingestion/snowplow): add environment filtering, event spec status filtering, and UI source registration

treff7es:
## Summary

Addresses customer feedback on Snowplow ingestion:

- **Environment filtering**: New `deployment_environment` config filters schemas by Snowplow deployment env (PROD/DEV) and pipelines by label. Decoupled from DataHub's `env` field to avoid value-space conflicts.
- **Event spec status filtering**: New `event_spec_statuses` config filters event specs by status (draft/published/deprecated/archived). Validated against `EventSpecStatus` enum.
- **Organization display name**: Container now shows the org name from the BDP API instead of the raw UUID. Falls back gracefully on API errors.
- **UI source registration**: Snowplow is now a selectable source in the DataHub ingestion UI with a sample recipe template.

### New config fields

| Field | Type | Default | Description |
|---|---|---|---|
| `deployment_environment` | `Optional[str]` | `None` | Filter by Snowplow deployment env (e.g., `PROD`, `DEV`). Case-insensitive. |
| `event_spec_statuses` | `Optional[List[str]]` | `None` | Filter event specs by status. Valid: `draft`, `published`, `deprecated`, `archived`. |

### Example recipe
```yaml
source:
  type: snowplow
  config:
    bdp_connection:
      organization_id: "YOUR_ORG_UUID"
      api_key_id: "${SNOWPLOW_API_KEY_ID}"
      api_key: "${SNOWPLOW_API_KEY}"
    deployment_environment: PROD
    event_spec_statuses:
      - published
```

## Test plan

- [x] Existing unit tests pass (659 tests, all green)
- [x] Lint passes (ruff + mypy via `lintFix`)
- [x] Tested live against Snowplow BDP API with PROD and DEV environments
- [x] Verified ingestion into local DataHub instance
- [x] Verified pipeline label filtering (prod pipeline label correctly filtered DEV ingestion)
- [x] Verified event spec status filtering (12 draft specs correctly filtered)
- [x] Verified org display name resolved from API ("Acryl Datahub" instead of UUID)

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/snowplow/processors/event_spec_processor.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/snowplow/processors/event_spec_processor.py
@@ -99,10 +99,32 @@ class EventSpecProcessor(EntityProcessor):
             )
             return
 
+        # Pre-compute allowed statuses set (already normalized to lowercase by config validator)
+        allowed_statuses: Optional[frozenset] = None
+        if self.config.event_spec_statuses is not None:
+            allowed_statuses = frozenset(self.config.event_spec_statuses)
+
         for event_spec in event_specs:
             self.report.report_event_spec_found()
 
-            # Apply filtering
+            # Apply status filtering
+            if allowed_statuses is not None:
+                if event_spec.status is None:
+                    logger.info(
+                        f"Filtering out event spec '{event_spec.name}' "
+                        f"(no status set, filter requires one of {self.config.event_spec_statuses})"
+                    )
+                    self.report.report_event_spec_filtered(event_spec.name)
+                    continue
+                if event_spec.status.lower() not in allowed_statuses:
+                    logger.info(
+                        f"Filtering out event spec '{event_spec.name}' "
+                        f"(status='{event_spec.status}' not in {self.config.event_spec_statuses})"
+                    )
+                    self.report.report_event_spec_filtered(event_spec.name)
+                    continue
+
+            # Apply name pattern filtering
             if not self.config.event_spec_pattern.allowed(event_spec.name):
                 self.report.report_event_spec_filtered(event_spec.name)
                 continue
--- a/metadata-ingestion/src/datahub/ingestion/source/snowplow/processors/pipeline_processor.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/snowplow/processors/pipeline_processor.py
@@ -120,7 +120,38 @@ class PipelineProcessor(EntityProcessor):
                 )
                 return
 
-            # Use first pipeline (most orgs have one pipeline)
+            # Filter pipelines by label matching deployment_environment (if set)
+            # deployment_environment is already normalized to uppercase by config validator
+            if self.config.deployment_environment:
+                target_env = self.config.deployment_environment
+                matching_pipelines = []
+                for p in physical_pipelines:
+                    if not p.label:
+                        # Pipeline has no environment label — include by default
+                        logger.info(
+                            f"Pipeline '{p.name}' has no label, including it "
+                            f"(cannot determine if it matches deployment_environment='{target_env}')"
+                        )
+                        matching_pipelines.append(p)
+                    elif p.label.upper() != target_env:
+                        logger.info(
+                            f"Filtering out pipeline '{p.name}' "
+                            f"(label='{p.label}' does not match deployment_environment='{target_env}')"
+                        )
+                        self.report.report_pipeline_filtered_by_env(p.name)
+                    else:
+                        matching_pipelines.append(p)
+
+                if not matching_pipelines:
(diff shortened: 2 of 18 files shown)


metadata-ingestion/src/datahub/ingestion/source/snowplow/processors/event_spec_processor.py
```diff
@@ -99,10 +99,32 @@ def extract(self) -> Iterable[MetadataWorkUnit]:
...
+        if self.config.event_spec_statuses is not None:
+            allowed_statuses = frozenset(self.config.event_spec_statuses)
+
         for event_spec in event_specs:
             self.report.report_event_spec_found()
 
-            # Apply filtering
+            # Apply status filtering
+            if allowed_statuses is not None:
+                if event_spec.status is None:
```
sgomezvillamor, line 112: Is the `event_spec.status` a closed enum?
If not ie, if free text, we may consider defining `event_spec_status` fiter as `AllowDenyPattern`.

metadata-ingestion/src/datahub/ingestion/source/snowplow/processors/pipeline_processor.py
```diff
@@ -120,7 +120,38 @@ def _extract_pipelines(self) -> Iterable[MetadataWorkUnit]:
...
+                matching_pipelines = []
+                for p in physical_pipelines:
+                    if not p.label:
+                        # Pipeline has no environment label — include by default
+                        logger.info(
+                            f"Pipeline '{p.name}' has no label, including it "
+                            f"(cannot determine if it matches deployment_environment='{target_env}')"
+                        )
+                        matching_pipelines.append(p)
+                    elif p.label.upper() != target_env:
```
sgomezvillamor, line 136: How is we refer to this in our config as `deployment_environment`, however in the snowplow sdk this is just a "label"?

sgomezvillamor (commented):
Overall LGTM

My only blocking comment is about `p.label.upper() != target_env` which I don't understand why we are doing that 😅

metadata-ingestion/src/datahub/ingestion/source/snowplow/processors/event_spec_processor.py
treff7es: It's effectively a closed enum from the user's perspective — `EventSpecStatus` (StrEnum with draft/published/deprecated/archived) is defined in `constants.py`, and `validate_event_spec_statuses` in `snowplow_config.py` rejects any value not in that set with a clear error listing valid values.

Switching to `AllowDenyPattern` would actually regress UX here: typos would silently fail to match instead of raising a validation error, and we'd trade strict enum validation for regex complexity on a 4-value set. The on-wire type is `Optional[str]` for API defensiveness (API could return new values), but users can only pass known values via config.

metadata-ingestion/src/datahub/ingestion/source/snowplow/processors/pipeline_processor.py
treff7es: Good catch — the `.upper()` on the pipeline side looked asymmetric because the config value was already uppercased by the `validate_deployment_environment` validator, but that context wasn't obvious at the comparison site.

I've made two changes:
1. Applied `.upper()` symmetrically to both sides so the case-insensitive intent is self-documenting (and the extra `.upper()` on `target_env` is a no-op since the validator already uppercased it).
2. Updated the inline comment to explain the normalization (Snowplow's `pipeline.label` is free-form operator text, config value is already normalized).

On the broader naming question (`deployment_environment` in config vs `label` in the Snowplow SDK): I kept the config name as-is for UX (`label` alone is too generic) but updated the `deployment_environment` field description in `snowplow_config.py` to explicitly mention it maps to the Snowplow pipeline `label` field. Happy to rename to `pipeline_label` if you'd prefer stricter adherence to source-system naming.

sgomezvillamor: For me, this looks inconsistent or mixing different stuff; we are using the same config for filtering labels 

```
                    elif p.label.upper() != target_env:
```

and deployment envs

```
                deployed_envs = {dep.env for dep in ds.deployments if dep.env}
```

while config name matches for the second, it doesn't for the first.

Is Snowplow label just free text, or is it restricted to deployment env values? It's also odd that there's only one label. 
* If it's free text, it probably warrants a separate config like `filter_label: Optional[str]`.
* If it's restricted to deployment envs, at least a code comment would help future readers.

Approving to unblock.

sgomezvillamor (approved):
Approving to unblock

However, using deployment env new config to filter by label still looks wrong to me https://github.com/datahub-project/datahub/pull/17056/changes#r3117791080

PR #7506, opened 2023-03-07
feat(ingest): enable joins across FileBackedDicts + add FileBackedList

hsheth2:
## Checklist

- [ ] The PR conforms to DataHub's [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [Commit Message Format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#commit-message-format))
- [ ] Links to related issues (if applicable)
- [ ] Tests for the changes have been added/updated (if applicable)
- [ ] Docs related to the changes have been added/updated (if applicable). If a new feature has been added a Usage Guide has been added for the same.
- [ ] For any breaking change/potential downtime/deprecation/big changes an entry has been made in [Updating DataHub](https://github.com/datahub-project/datahub/blob/master/docs/how/updating-datahub.md)

PR #17057, opened 2026-04-16
fix(ingest): switch to pure Python sqlglot to avoid memory leak in sqlglot[c]

sgomezvillamor:
## Summary

Fixes critical memory leak in BigQuery ingestion that caused large memory consumption (GBs higher than usual) when processing views.

**Root Cause**: Reference count leak in `sqlglot[c]==30.0.3` C extension's `Table.name` accessor (see [sqlglot issue #7506](https://github.com/tobymao/sqlglot/issues/7506)). Each table name access leaks one reference, leading to massive memory accumulation during view lineage extraction.

**Temporary Fix**: Switch from `sqlglot[c]==30.0.3` to pure Python `sqlglot==30.0.3`. Performance impact is minimal, in small test at least (3x slower but only 0.004s difference per 20K table name accesses).

**Test Infrastructure Improvements**:
- Added `test_sqlglot_memory_leak.py` with reference count leak detection
- Created `testPerformance` Gradle task to run performance tests in CI
- Integrated into GitHub Actions workflow (Python 3.11 matrix)
- Cleaned up performance test markers for proper isolation

## Test Plan

- [x] Created test that reproduces the leak (fails with sqlglot[c], passes with pure Python)
- [x] Verified memory usage drops from 16GB to <1GB in production BigQuery ingestion
- [x] Confirmed performance impact is negligible (0.004s per 20K operations)
- [x] Added CI integration to catch future regressions

## Breaking Changes

None. This is a transparent fix with no API changes.

## Notes

This should be reverted once sqlglot upstream fixes the C extension leak. Monitor [sqlglot issue #7506](https://github.com/tobymao/sqlglot/issues/7506) for resolution.


metadata-ingestion/tests/performance/test_sqlglot_memory_leak.py
```diff
@@ -0,0 +1,321 @@
...
+        print(f"\nApproximate size per cached result: {result_size} bytes")
+        print(
+            f"Total cache size estimate: {result_size * cache_info_after.currsize / 1024 / 1024:.2f} MB"
+        )
+        print(
+            f"Projected for 1000 cache entries: {result_size * 1000 / 1024 / 1024:.2f} MB"
+        )
+
+
+if __name__ == "__main__":
```
skrydal, line 299: is this something we follow normally?

skrydal (approved):
LGTM, I would keep pytest as our runner for tests though

PR #17069, opened 2026-04-17
fix(timeout): Handle Report timeout properly In Mode

puneetagarwal-datahub:
**Problem:** _process_report caught all exceptions (including TimeoutError / requests.exceptions.Timeout) and called report_failure. This caused has_failures() to return true, marking the entire run as FAILURE and skipping stale entity removal — even if only 1 of 8,675 reports timed out.

**Fix:** Added a timeout detection check before deciding which report method to call:
  - TimeoutError or requests.exceptions.Timeout → report_warning (run continues, stale removal proceeds)
  - All other exceptions → report_failure (unchanged)
  
  Tests are added for same

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/mode.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/mode.py
@@ -1856,21 +1856,30 @@ class ModeSource(StatefulIngestionSourceBase):
             )
         except Exception as e:
             try:
+                is_timeout = isinstance(e, (TimeoutError, requests.exceptions.Timeout))
                 logger.warning(
-                    f"Failed to process report {report_name} ({report_token}) "
-                    f"in space {space_token}",
+                    f"{'Timed out' if is_timeout else 'Failed'} processing report "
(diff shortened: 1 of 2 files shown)


treff7es (approved):

metadata-ingestion/src/datahub/ingestion/source/mode.py
```diff
@@ -1856,21 +1856,30 @@ def _process_report(
             )
         except Exception as e:
             try:
+                is_timeout = isinstance(e, (TimeoutError, requests.exceptions.Timeout))
                 logger.warning(
```
treff7es, line 1860: Does this make sense if we log another below?

puneetagarwal-datahub: I added the explicit logger.warning because report_warning is hardcoded silent (log=False). But you're right that report_failure already logs internally, so the failure branch double-logs. One way is switching to self.report.warning(...) / self.report.failure(...) so both branches log exactly once and remove logger.warning ."

PR #17073, opened 2026-04-17
fix(adk): convert ItemNotFoundError to message dict in create_context_wrapper

### LINKED ISSUE #16882: get_lineage_between return the ItemNotFoundError, Should be a message instead of error from adk module

creazyfrog:
## Summary

Fixes #16882

Per reviewer feedback from @sgomezvillamor and @nwadams: instead of handling `ItemNotFoundError` inside individual tools, centralise the exception-to-message conversion in `create_context_wrapper` in `utils.py`. This wrapper is applied to every tool for both the LangChain and Google ADK adapters, so the policy is enforced automatically across all current and future tools.

## Problem

Agent/LLM callers cannot handle raw exceptions — they need a dict response they can relay to users. `get_lineage_paths_between` previously raised `ItemNotFoundError` when no lineage path existed, which broke agent callers. The original PR caught this inside the function itself, but that was a tactical one-off fix.

## Approach

Add a single `except ItemNotFoundError` clause to `create_context_wrapper`:

```python
@functools.wraps(func)
def wrapper(*args, **kwargs):
    token = set_client(client)
    try:
        return func(*args, **kwargs)
    except ItemNotFoundError as e:
        return {"message": str(e)}   # ← agent-friendly response
    finally:
        reset_client(token)          # ← always reset context
```

- `ItemNotFoundError` → returned as `{"message": "..."}` to the caller
- All other exceptions (e.g. `ValueError`, `RuntimeError`) still propagate normally — programming errors are not silently swallowed
- Context is always reset via `finally`, even when the error is caught

## Changes

| File | Change |
|---|---|
| `utils.py` | Catch `ItemNotFoundError` in `create_context_wrapper`; return `{"message": str(e)}` |
| `lineage.py` | Remove `_no_path_found_response()` helper; revert `try/except` blocks; `get_lineage_paths_between` raises `ItemNotFoundError` as before |
| `test_utils.py` | Add 3 tests: error→dict conversion, context reset after caught error, non-`ItemNotFoundError` propagation |
| `test_lineage.py` | Revert to asserting `ItemNotFoundError` is raised from the function directly; wrapper behaviour tested in `test_utils.py` |

## Test plan

- [ ] `pytest datahub-agent-context/tests/unit/test_utils.py` — covers wrapper conversion
- [ ] `pytest datahub-agent-context/tests/unit/mcp_tools/test_lineage.py` — covers function-level raise
- [ ] Call `get_lineage_paths_between` through an agent adapter with two unrelated datasets and verify a `{"message": "..."}` dict is returned (not an exception)


sgomezvillamor (commented):
Thanks for the contribution! Returning a structured dict instead of raising `ItemNotFoundError` when no lineage path exists is the right semantic move — "no path found" is a valid query result, not an execution error, and modeling it that way gives LLM callers cleaner signal.

That said, I have a broader concern: this is a tactical fix to one tool, but empty/not-found handling and error handling are inconsistent across the `mcp_tools/` module. A few examples:

- **`entities.py`**: the list call returns `{"error": "Entity not found", "urn": urn}` per item (dict pattern ✓), but single-entity helpers still raise `ItemNotFoundError`
- **`assertions.py`**: returns `{"message": "No assertions found..."}` for empty results (good!), but raises `RuntimeError` for actual fetch failures
- **`get_me.py`**, **`ask_datahub.py`**, and all mutation tools (`domains.py`, `owners.py`, `tags.py`, etc.) raise `ValueError` / `RuntimeError` freely with no dict fallback
- **`save_document.py`** is the only tool that consistently returns `{"success": bool, "message": "..."}` for all outcomes and never raises

A global mechanism would be more durable than patching tools individually. There's a natural interception point already in the codebase: `create_context_wrapper` in `utils.py`, which wraps every tool for both the LangChain and Google ADK adapters. Adding exception-to-dict conversion there would enforce the policy automatically across all current and future tools. To avoid silently swallowing unexpected programming errors, I'd suggest pairing this with a `DataHubToolError` base class — tools raise that for expected failures (not-found, invalid input, operation failed), and `create_context_wrapper` converts only those to structured dicts while letting genuine exceptions propagate.

I'm not deeply familiar with this component's architecture, so tagging @nwadams for a more authoritative review on whether this direction fits the roadmap.

---

creazyfrog:
@sgomezvillamor thanks for the review, @nwadams could also have a look. So that based on the feedback I will update the PR.

nwadams:
Thank you for the contribution @creazyfrog!

I think it makes sense to standardize in `create_context_wrapper` and wrap the tools to ensure a message is returned instead of raising an exception. 

Separately internally we should work to standardize the patterns.

creazyfrog:
@nwadams  thank you for the feedback — updated as suggested.

The exception-to-message conversion has been moved out of get_lineage_paths_between and into create_context_wrapper in utils.py, so it applies automatically to every tool wrapped for LangChain and Google ADK adapters.

Key decisions made:

- Only ItemNotFoundError is caught — other exceptions (ValueError, RuntimeError, etc.) still propagate so programming errors aren't silently swallowed
- The finally block still resets the client context regardless of whether the error was caught
- get_lineage_paths_between reverts to raising ItemNotFoundError as before — the wrapper handles the adapter boundary
- test_utils.py has new tests for the wrapper behaviour; test_lineage.py reverts to asserting the function raises directly

Happy to also catch ValueError in the wrapper if that fits the broader standardization direction you mentioned — just wanted to keep the scope minimal for now and let you guide that.

sgomezvillamor (commented):
LGTM 👌 
Leaving the final call to @nwadams.

creazyfrog:
@sgomezvillamor thanks for the confirmation, 
@nwadams could you also please have a look.

creazyfrog:
@nwadams could you also please have a look.

nwadams (approved):

PR #17077, opened 2026-04-17
fix(patch): Revert template attribution changes; use generic json patch on clients

asikowitz:
Stopgap to make CLI backwards compatible with older server versions again. #16906 was not backwards compatible as if the client sends a longer path than the backend is expecting (i.e. based on how it converts arrays to maps), the backend is unable to apply the patch. This reverts the template changes, because I'm entirely reverting schema field patch (see comment below) which thus depends on them being in their old state. After a tech spec on the best way to update patch semantics in a backwards compatible way, I can bring these changes back.

Instead, I use generic json patch to handle attributed patching, always specifying `arrayPrimaryKeys` because I want to avoid using the default templates which are now "broken" re: attribution. Generic json patch can't replace the more complicated logic of EditableSchemaMetadataTemplate so I just reverted the ability to patch that aspect with attribution. Because this is technically a backwards incompatible change of an already-released CLI, I kept the param but added a warning if users try to use it.

Also adds matching functionality to the Java sdk and adds a test to check the Python and Java clients behave similarly (the Java client functionality is a subset of that of the Python client).


david-leifker (approved):

RyanHolstien (approved):

jayacryl (approved):

jayacryl:
@cursor merge master, review each conflict carefully as you resolve it. Be self critical, understand the reason for changes thoroughly, and trace the surrounding code rigorously before making your decisions.

asikowitz:
I'll just do the merge

asikowitz:
@claude run `./gradlew :metadata-ingestion:testIntegrationBatch1` and fix the failing tests. You should only have to change a couple lines in golden files. You should be able to generate the changes using `pytest --update-golden-files` on the failing tests, but if you do that, verify after that the changes are small. Commit and push when you are confident about the changes.

PR #17086, opened 2026-04-17
feat(ingest/sigma): add intra-workbook element-to-element lineage

kyungsoo-datahub:
Sigma's /workbooks/{id}/lineage already returns element-to-element (type: sheet) edges; the connector was discarding them in sigma_api.py by filtering to type == "dataset" only. This PR keeps both and emits element-to-element lineage. (Every Sigma element — chart, table, pivot, input-table — maps to a DataHub Chart URN, so the edges are Chart→Chart at the DataHub level.)

Routing: Dataset upstreams stay on ChartInfo.inputs (the "Data Sources" UI reads that field); Chart upstreams go on inputEdges, the only field accepting Chart URNs per the PDL. The elementId → chart_urn map is built at workbook scope so cross-page refs resolve.


metadata-ingestion/src/datahub/ingestion/source/sigma/data_classes.py
```diff
@@ -50,6 +50,15 @@ def get_urn_part(self):
         return self.url.split("/")[-1]
 
 
+class ElementUpstream(BaseModel):
+    node_id: str
+    # "dataset" = Sigma Dataset node; "sheet" = Sigma element (any element.type).
+    # "table" (warehouse) and "join" (graph-only pass-through) are filtered at the API layer.
+    type: Literal["dataset", "sheet"]
+    name: Optional[str] = None
+    element_id: Optional[str] = None  # populated only when type == "sheet"
```
treff7es: Consider making this a discriminated union so the invariant in the comment on L59 becomes typed rather than documented:

```python
class DatasetUpstream(BaseModel):
    type: Literal["dataset"] = "dataset"
    node_id: str
    name: str                       # required — drives SQL-parse fuzzy match in sigma.py

class SheetUpstream(BaseModel):
    type: Literal["sheet"] = "sheet"
    node_id: str
    name: Optional[str] = None
    element_id: str                  # required — no more runtime None-check

ElementUpstream = Annotated[
    Union[DatasetUpstream, SheetUpstream], Field(discriminator="type")
]
```

Two wins from this:
1. **Eliminates the `element_id is None` runtime branch** in `sigma.py:437-441` — a sheet without an elementId would fail validation at the API-layer instead of silently producing an `ElementUpstream` that later gets skipped.
2. **Makes `name` non-Optional on `DatasetUpstream`**. Pre-PR `source_name` was `str`; the change to `Optional[str]` is a type-precision regression — for the `"dataset"` branch, `name` is essential (it's how `sigma.py:425-429` correlates Sigma datasets with SQL-parsed warehouse tables). If a dataset node ever comes back without a name, the element silently loses its warehouse-table attribution.

Minor: the `node_id` field duplicates the dict key in `Element.upstream_sources`. Either OK, but if you drop it the consumer in `sigma.py:424` (`node_id.split("-")[-1]`) already uses the dict key, so nothing downstream needs the model field.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
```diff
@@ -414,21 +419,35 @@ def _get_element_input_details(
...
+                            inputs[dataset_urn] = [in_table_urn]
+                        else:
+                            inputs[dataset_urn].append(in_table_urn)
+                        sql_parser_in_tables.remove(in_table_urn)
+            elif upstream.type == "sheet":
+                if upstream.element_id is None:
+                    logger.debug(
+                        f"Sheet upstream missing elementId for element {element.name}, skipping"
+                    )
+                    continue
```
treff7es: A `"sheet"` edge missing `elementId` is an API contract violation, not an expected case — silently dropping it at `logger.debug` means a user with incomplete intra-workbook lineage has no signal why. Per the connector error-handling standards, this should surface via `self.report.warning`:

```python
if upstream.element_id is None:
    self.reporter.warning(
        title="Sigma sheet upstream missing elementId",
        message=f"element={element.name} workbook={workbook.name}",
    )
    continue
```

Separately, this branch has **no test coverage**. Add a fixture dependency where `type: "sheet"` has no `elementId` key and assert the element still ingests without a spurious `inputEdges` entry.

(If you adopt the discriminated-union suggestion on `data_classes.py`, this whole branch goes away — the API-layer construction site becomes the point of enforcement.)

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
```diff
@@ -414,21 +419,35 @@ def _get_element_input_details(
...
+                chart_urn = elementId_to_chart_urn.get(upstream.element_id)
+                if chart_urn is None:
+                    # Target element type not in our allow-list (e.g. pivot-table).
+                    logger.debug(
+                        f"Upstream elementId {upstream.element_id} not in element map "
+                        f"for element {element.name} (possibly a filtered element type)"
+                    )
+                    continue
```
treff7es, line 452: This is the documented workaround for filtered element types (pivot-table, etc. — filtered at `sigma_api.py:399` to `["table", "visualization"]`), and it's the branch **most likely to regress** because the filter list and the chart-map construction are in separate files. It has zero test coverage right now — the new integration test only creates elements that are already in the allowlist.

Add a fixture: a downstream sheet whose `elementId` references a pivot-table (or other non-allowlisted type), and assert the downstream chart is still produced with no `inputEdges` entry for the missing target.

Minor: the debug message is a bit vague for users — consider including a pointer, e.g. `"...possibly filtered at get_page_elements; enable DEBUG logs for details"`.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma_api.py
```diff
@@ -323,14 +327,37 @@ def _get_element_upstream_sources(
...
+                elif source_type in ("table", "join"):
+                    # "table" = warehouse table, handled by SQL-parse path.
+                    # "join" = graph-only pass-through: its upstreams appear as
+                    # separate edges in the same response and are processed there.
+                    pass
+                else:
+                    logger.debug(
+                        f"Skipping unknown lineage node type '{source_type}' "
+                        f"for element {element.name} of workbook '{workbook.name}'"
+                    )
```
treff7es: Logging an unknown API node type at **debug** level is the exact pattern this PR exists to fix — the pre-PR code silently dropped `"sheet"` edges by filtering to `"dataset"` only, and nobody noticed because there was no warning surface. If Sigma adds another node type later (`"formula"`, `"input-table"`, `"model"`, etc.), the same regression recurs invisibly.

Promote to a report warning so API-schema drift is visible:

```python
else:
    self.report.warning(
        title="Unknown Sigma lineage node type",
        message=f"type={source_type!r} element={element.name} workbook={workbook.name}",
    )
```

Also — this branch has no test coverage. Add a fixture with an unrecognized `type` in `dependencies` and assert the element still ingests cleanly (no `inputEdges` for the unknown upstream).

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma_api.py
```diff
@@ -323,14 +327,37 @@ def _get_element_upstream_sources(
...
+
             return upstream_sources
         except Exception as e:
```
treff7es: Pre-existing, but the PR **amplifies** its blast radius. Before this change, the bare `except Exception: return {}` at the bottom of `_get_element_upstream_sources` dropped warehouse-table lineage on failure. Now it drops *both* dataset upstreams *and* intra-workbook sheet upstreams on a single malformed edge.

The new code paths introduce additional failure modes that also land here silently:
- `pydantic.ValidationError` from `ElementUpstream` (e.g. if a `"sheet"` node is constructed with an unexpected `type` after a future refactor — see defense-in-depth around the `Literal["dataset", "sheet"]`)
- `KeyError` on `dependencies[source_node_id]`, `DEPENDENCIES`, `EDGES`, `TYPE`, `SOURCE`

`_log_http_error` only surfaces `HTTPError` at warning level (L72); everything else goes to `logger.debug` on L73 and `SigmaSourceReport` never sees it. Recommend narrowing:

```python
try:
    response = self._get_api_call(...)
    # status-code checks + raise_for_status
    response_dict = response.json()
except requests.exceptions.RequestException as e:
    self._log_http_error(...)
    return {}

try:
    for edge in response_dict[Constant.EDGES]:
        ...
except (KeyError, ValidationError) as e:
    self.report.warning(
        title="Failed to parse Sigma element lineage response",
        message=f"element={element.name} workbook={workbook.name}",
        exc=e,
    )
    return {}
```

At minimum, upgrade the existing `_log_http_error` path to `self.report.warning` since the connector now loses more lineage per failure.

metadata-ingestion/tests/integration/sigma/test_sigma.py
```diff
@@ -510,6 +510,212 @@ def test_platform_instance_ingest(pytestconfig, tmp_path, requests_mock):
...
+    override_data: Dict[str, Dict] = {
+        # Add a third page containing the intra-workbook elements.
+        "https://aws-api.sigmacomputing.com/v2/workbooks/9bbbe3b0-c0c8-4fac-b6f1-8dfebfe74f8b/pages": {
+            "method": "GET",
+            "status_code": 200,
+            "json": {
+                "entries": [
+                    {"pageId": "OSnGLBzL1i", "name": "Page 1"},
+                    {"pageId": "DFSieiAcgo", "name": "Page 2"},
+                    {"pageId": "IntraWorkbookPage", "name": "Page 3"},
```
treff7es, line 531: The PR description says:

> The `elementId → chart_urn` map is built at workbook scope so cross-page refs resolve.

But the three elements on `IntraWorkbookPage` (`upstreamElem01`, `downstreamElem01`, `joinDownstreamElem`) all reference each other within the same page. A refactor that reverted the map to **page-scope** construction would still pass this test — the PR's own design justification isn't verified by CI.

Recommend adding a sheet upstream on Page 3 whose `elementId` belongs to an element defined on Page 1 or Page 2 (e.g. `kH0MeihtGs` or `Ml9C5ezT5W`), and asserting that the downstream chart on Page 3 emits an `inputEdges` edge pointing to the corresponding chart URN from the other page. This also exercises the real code path the workbook-scoped construction was introduced to support.

(Nice touch on the `src_node_upstream` fixture btw — having `nodeId != elementId` is a sharp test that would have caught the pre-PR `source_id.split("-")[-1]` bug.)

treff7es (approved):

metadata-ingestion/src/datahub/ingestion/source/sigma/data_classes.py
kyungsoo-datahub: Done. Also dropped node_id from both classes per your minor note: it's never accessed via upstream.node_id; all consumers use the dict key from the loop.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
kyungsoo-datahub: Thank you for the comment. Revised.

metadata-ingestion/tests/integration/sigma/test_sigma.py
kyungsoo-datahub: Revised. Thanks!

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
kyungsoo-datahub: Revised. Thanks.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma_api.py
kyungsoo-datahub: Done.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma_api.py
kyungsoo-datahub: Revised. Thanks.

PR #17087, opened 2026-04-17
fix(CLI): ingestion rollback privileges for cli command

kevinkarchacryl:
Currently you cannot do datahub ingest rollback --run-id urn:li:dataHubExecutionRequest:xxxx unless you are an admin. Even after granting a user every available platform and metadata privilege, the CLI rollback would still return 403 while the GraphQL rollback (UI) worked fine for the same user.

The root cause is in BatchIngestionRunResource.rollback(). The auth check AuthUtil.isAPIAuthorized(opContext, ENTITY, MANAGE) passes a null entity type, which causes MANAGE to resolve to a conjunction of EDIT_ENTITY + DELETE_ENTITY from the generic entity privilege map. Since no EntitySpec is constructed when entity type is null, the policy engine rejects all metadata policies (which require resource context to match). EDIT_ENTITY and DELETE_ENTITY only exist in metadata policies, so no policy can ever match — resulting in a permanent DENY for all non-admin users regardless of what privileges they have been granted.

The fix changes the call to AuthUtil.isAPIAuthorizedEntityType(opContext, MANAGE, INGESTION_SOURCE_ENTITY_NAME), which passes dataHubIngestionSource as the entity type. This causes the auth check to look up privileges from API_ENTITY_PRIVILEGE_MAP for that entity type, where MANAGE_INGESTION_PRIVILEGE satisfies both the UPDATE and DELETE legs of the MANAGE conjunction. This also aligns the REST path with the GraphQL rollback path (RollbackIngestionResolver), which already used the correct entity type context via canManageIngestion().

Changes were tested locally.


chriscollins3456 (approved):
nice thanks!

PR #17095, opened 2026-04-18
chore(ingestion): replace freezegun with time_machine for time mocking

sgomezvillamor:
## Description

This PR replaces the `freezegun` library with `time_machine` for time mocking across the codebase. The change involves:

1. **Dependency Update**: Removed `freezegun` from `pyproject.toml`, `setup.py`, and `constraints.txt`, replacing it with `time_machine`
2. **Import Changes**: Updated all test files to import `time_machine` instead of `freezegun`
3. **API Migration**: Converted all `@freeze_time()` decorators to `@time_machine.travel(..., tick=False)` to maintain equivalent behavior
4. **Import Organization**: Reorganized imports in multiple test files to follow consistent ordering patterns

## Motivation

`time_machine` is a more modern and actively maintained alternative to `freezegun` for time mocking in Python tests. It provides better compatibility with newer Python versions and has improved performance characteristics.

## Changes Made

- Replaced 80+ occurrences of `@freeze_time()` with `@time_machine.travel(..., tick=False)` across test files
- Updated imports in test files throughout the codebase
- Removed `freezegun` dependency from dependency management files
- Maintained backward compatibility by using `tick=False` parameter to preserve existing test behavior

## Post-migration CI fixes

Several integration tests needed additional fixes due to a behavioral difference between `freezegun` and `time_machine`:

**Root cause:** Tests that had *both* a `@time_machine.travel(FROZEN_TIME)` decorator *and* `mock_time` in their function signature behaved differently under the two libraries:
- With `freezegun` + the old monkeypatch-based `mock_time`: the Python-level monkeypatch overrode `@freeze_time`, so `time.time()` always returned `_MOCK_TIME` — golden files were generated with this value.
- With `time_machine`: the inner decorator wins, so `time.time()` returns `FROZEN_TIME`. Additionally, decorators missing `tick=False` (i.e. using the default `tick=True`) let time advance from `FROZEN_TIME` in real-time during test execution, producing non-deterministic millisecond offsets.

**Fixes applied per failing batch:**

- **Batch 0 (postgres):** Added `tick=False` to all 3 postgres decorators + updated golden files (`1615443388097` → `1646575200000`). Removed `mock_time` from function signatures (redundant — decorator already provides the frozen context).
- **Batch 2 (trino):** Updated golden files (`1615443388097` → `1632398400000`). Removed `mock_time` from function signatures.
- **Batch 3 (doris + powerbi):** Added `tick=False` to doris decorators + updated golden files (`1615443388097` → `1586847600000`). Removed `mock_time` from doris function signature. Removed `mock_time` from the two PowerBI MSAL token expiry tests — `time_machine`'s C-level time freeze prevented MSAL from ever detecting token expiry, while the old monkeypatch only patched Python-level `time.time` and left MSAL's internal C-level timers unaffected.

## Test Plan

All existing tests should pass without modification since the behavior is equivalent. The `tick=False` parameter ensures that time doesn&#39;t advance during test execution, matching the original `freezegun` behavior.

https://claude.ai/code/session_01EA2NhcpPbRjFohNUTeYztF


sgomezvillamor:
The `connector-tests` failure appears unrelated to this PR's changes. This PR only modifies test files (`tests/`) and golden JSON files — no connector source code (`src/`) was changed:

```
git diff origin/master --name-only | grep "src/"
# (no output)
```

Could an org member please add the `skip-connector-tests` label to unblock merging? The failure is in the external `acryldata/connector-tests` repo and cannot be caused by changes to test fixtures and golden files.

---

metadata-ingestion/tests/conftest.py
```diff
@@ -35,27 +35,13 @@
...
-    import pandas  # noqa: F401
-except ImportError:
-    pass
-
-import freezegun  # noqa: E402
-
-# The freezegun library has incomplete type annotations.
-# See https://github.com/spulec/freezegun/issues/469
-freezegun.configure(extend_ignore_list=["datahub.utilities.cooperative_timeout"])  # type: ignore[attr-defined]
+_MOCK_TIME = 1615443388.0975091  # 2021-03-11 06:16:28.097509+00:00
```
treff7es, line 38: This looks a bit odd. Why this timestamp?

treff7es (approved):

metadata-ingestion/tests/conftest.py
sgomezvillamor: This value predates this PR — it was the fixed timestamp used by the original `mock_time` fixture (previously implemented as a `monkeypatch` on `time.time()`). It's an arbitrary frozen point in time ("2021-03-11 06:16:28 UTC", as the comment shows) chosen when the fixture was first created.

We kept the exact value here so that golden files for the ~60 tests that use only `mock_time` (with no `@time_machine.travel` decorator) remain valid — those files contain the corresponding epoch-ms value `1615443388097` and don't need to change.

---

PR #17129, opened 2026-04-21
feat(sql-setup): align metadata_aspect_v2 indexes for Postgres and MySQL

david-leifker:
## Problem

- legacy single-column indexes (`urnIndex`, `aspectIndex`, `versionIndex`) from older incorrect DDL are redundant.
- **PostgreSQL:** Secondary indexes on large `metadata_aspect_v2` tables should avoid long exclusive locks during SqlSetup; 
- **MySQL:** `CREATE TABLE IF NOT EXISTS` does **not** add new inline indexes to an **existing** table, so upgraded databases could miss `timeIndex` and `idx_version_urn_aspect` even when DDL was updated.

## Improvement

The new metadata_aspect_v2 index speeds up the aspect reads that restore indices and metadata-tests exercise, so those paths spend less time scanning large tables.

### `CreateTablesStep`

After running table DDL, the step now:

1. `dropLegacyAspectTableIndexes(connection)`
2. `ensureAspectIndexes(connection)`
3. `postSetup(connection)` (unchanged role, e.g. optional `schemaVersionIndex` on Postgres)

### PostgreSQL (`PostgresDatabaseOperations`)

- **Table DDL:** `CREATE TABLE` only (no heavy secondary indexes in the same transactional `sqlUpdate` path).
- **`ensureAspectIndexes`:** `CREATE INDEX CONCURRENTLY IF NOT EXISTS` for:
  - `timeIndex` on `(createdon)`
  - Partial `idx_v0_urn_aspect` on `(urn, aspect) WHERE version = 0`
  - Partial corp user / corp group helper indexes on `(urn, aspect)` with `urn LIKE` filters and `version = 0`
- **`dropLegacyAspectTableIndexes`:** `DROP INDEX CONCURRENTLY IF EXISTS` for legacy `urnIndex`, `aspectIndex`, `versionIndex` when present.
- Connection uses autocommit where required for `CONCURRENTLY`.

### MySQL (`MySqlDatabaseOperations`)

- **Table DDL:** `timeIndex (createdon)` and composite `idx_version_urn_aspect (version, urn, aspect)` for **new** installs.
- **`dropLegacyAspectTableIndexes`:** If `urnIndex` / `aspectIndex` / `versionIndex` exist (via `information_schema.statistics`), `ALTER TABLE ... DROP INDEX`.
- **`ensureAspectIndexes`:** If `timeIndex` or `idx_version_urn_aspect` is missing, `ALTER TABLE metadata_aspect_v2 ADD INDEX ...` so **existing** databases converge without manual SQL.

### DIFF
--- a/datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/sqlsetup/PostgresDatabaseOperations.java
+++ b/datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/sqlsetup/PostgresDatabaseOperations.java
@@ -127,7 +127,9 @@ public class PostgresDatabaseOperations implements DatabaseOperations {
 
   @Override
   public java.util.List<String> createTableSqlStatements(boolean createSchemaVersionIndex) {
-    return java.util.Arrays.asList(
+    // Secondary indexes are created in ensureAspectIndexes() via CREATE INDEX CONCURRENTLY so
+    // existing populated tables are not blocked by long exclusive locks during index builds.
+    return java.util.List.of(
         """
         CREATE TABLE IF NOT EXISTS metadata_aspect_v2 (
           urn                           varchar(500) not null,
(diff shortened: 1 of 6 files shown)


datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/sqlsetup/PostgresDatabaseOperations.java
```diff
@@ -140,11 +142,40 @@ createdby                     varchar(255) not null,
...
+  @Override
+  public void ensureAspectIndexes(Connection connection) throws SQLException {
+    // CREATE INDEX CONCURRENTLY cannot run inside a transaction block.
+    boolean prevAutoCommit = connection.getAutoCommit();
+    connection.setAutoCommit(true);
+    try (java.sql.Statement stmt = connection.createStatement()) {
+      stmt.execute(
+          "CREATE INDEX CONCURRENTLY IF NOT EXISTS timeIndex ON metadata_aspect_v2 (createdon);");
+      stmt.execute(
+          "CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_v0_urn_aspect ON metadata_aspect_v2 (urn, aspect) WHERE version = 0;");
```
chakru-r, line 171: I should have done this `where version = 0` for the schemaVersion index from system metadata as well. will do that in another PR

chakru-r (approved):

PR #7397, opened 2023-02-21
fix(ingest): Do not require platform_instance for stateful ingestion

asikowitz:
Following the changes from https://github.com/datahub-project/datahub/pull/6795, I believe that `get_platform_instance_id` can now return `None`. See its only usage:

In `StatefulIngestionSourceBase._get_last_checkpoint`
```
                try:
                    platform_instance_id = self.get_platform_instance_id()
                except NotImplementedError:
                    pass
                else:
                    last_checkpoint_aspect = self.ingestion_checkpointing_state_provider.get_latest_checkpoint(  # type: ignore
                        pipeline_name=self.ctx.pipeline_name,
                        job_name=job_id,
                        platform_instance_id=platform_instance_id,
                    )
```
Note that in that PR, `get_latest_checkpoint` was configured to accept an optional `platform_instance_id` (and also note that in the not implemented case, `None` is implicitly passed?)

Since we still want to keep implementations of `get_platform_instance_id` for backwards compatibility, I allow it to return `None` for when you sometimes have a platform_instance, and other times don't, like with Kafka.
Also removes any validation around null `platform_instance` that I could find.

EDIT: Tested the following recipe
```
source:
  type: kafka
  config:
    connection: 
      bootstrap: localhost:9092
      schema_registry_url: http://localhost:8081/
```

## Checklist

- [ ] The PR conforms to DataHub's [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [Commit Message Format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#commit-message-format))
- [ ] Links to related issues (if applicable)
- [ ] Tests for the changes have been added/updated (if applicable)
- [ ] Docs related to the changes have been added/updated (if applicable). If a new feature has been added a Usage Guide has been added for the same.
- [ ] For any breaking change/potential downtime/deprecation/big changes an entry has been made in [Updating DataHub](https://github.com/datahub-project/datahub/blob/master/docs/how/updating-datahub.md)

PR #17130, opened 2026-04-21
fix(ingest/dbt): stop unconditional lowercasing of target-platform URNs

alfiyas-datahub:
## Summary

Fixes a regression where dbt ingestion unconditionally lowercased all non-dbt target-platform URNs (BigQuery, Snowflake, etc.), regardless of the `convert_urns_to_lowercase` config flag. This caused lineage breaks with case-sensitive platforms like BigQuery when table names contained uppercase characters (e.g. `AM100` → `am100`).

**Root cause:** The conditional `if target_platform != DBT_PLATFORM or self.convert_urns_to_lowercase` always evaluated to `True` for non-dbt platforms, bypassing the flag entirely. Changed to `if self.convert_urns_to_lowercase` so lowercasing is purely opt-in.

**Note:** Snowflake dbt users who previously relied on the implicit lowercasing to match Snowflake ingestion URNs should add `convert_urns_to_lowercase: true` to their dbt recipe.

Fixes Zendesk #7397 (Autotrader — lineage break between dbt and BigQuery entities).


docs/how/updating-datahub.md
```diff
@@ -58,6 +58,7 @@ This file documents any backwards-incompatible changes in DataHub and assists pe
 
 ### Other Notable Changes
 
+- #16358 (Ingestion / dbt) **dbt no longer unconditionally lowercases target-platform URNs.** Previously, URNs for non-dbt target platforms (BigQuery, Snowflake, etc.) were always lowercased, which caused lineage breaks with case-sensitive platforms like BigQuery when table names contained uppercase characters. URN lowercasing is now fully opt-in via `convert_urns_to_lowercase: true` in the dbt recipe. **Snowflake dbt users** who rely on lowercased URNs to match their Snowflake ingestion should add `convert_urns_to_lowercase: true` to their dbt recipe to preserve existing behavior.
```
treff7es: Shouldn't lowercase true be the default if the earlier default was to always lowercase?

treff7es: Mixed casing always causes issues with lineage and I wonder why we decided to have this default value?

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
```diff
@@ -1061,7 +1061,7 @@ def get_urn(
         data_platform_instance: Optional[str],
     ) -> str:
         db_fqn = self.get_db_fqn()
-        if target_platform != DBT_PLATFORM or self.convert_urns_to_lowercase:
```
treff7es, line 1064: I think we should keep this if as well with the DBT_PLATFORM

treff7es (approved):
lgtm

PR #17154, opened 2026-04-22
ci(playwright): add Playwright E2E CI pipeline with sharding and fix test failures

priyabratadas-dh:
## Summary

- Adds a new GitHub Actions workflow (`playwright-e2e-tests.yml`) that runs Playwright E2E tests against DataHub head images with configurable sharding (default 4 shards)
- Integrates the new workflow into `post-workflow-actions.yml` so it participates in the standard PR status checks
- Fixes several test failures uncovered while validating the pipeline end-to-end

### CI pipeline (`playwright-e2e-tests.yml`)
- Runs on pushes to `e2e/**` branches and via `workflow_dispatch` (configurable shard count and quickstart profile)
- Each shard: starts DataHub via `run-quickstart.sh` (head images), installs Playwright + Chromium, runs its shard, uploads blob/JUnit/trace artifacts
- Final `playwright_report` job merges all shard blob reports into a single HTML report
- All action pins are full commit SHAs

### Test fixes
| File | Fix |
|------|-----|
| `playwright.config.ts` |  honour `PLAYWRIGHT_REUSE_SERVER=1` so CI can reuse the externally started DataHub instance instead of launching a conflicting one |
| `fixtures/login.fixture.ts` | Also persist `skipOnboardingTour` in saved auth state — without it the ReactTour overlay blocked clicks in tests that reuse the cached session |
| `tests/auth/login.spec.ts` | Replace inaccessible logo `img` assertion with greeting text — the V2 sidebar collapses by default so the logo link is CSS-hidden |
| `tests/login-v2/v2-login.spec.ts` | Add `exact: true` to `getByRole('button', { name: 'Discover' })` — without it Playwright strict mode fails because both the nav button and a CTA button match |
| `tests/onboarding/welcome-modal.spec.ts` | Skip "auto-advance slides" test in CI — react-slick autoplay is paused in headless Chromium (window focus / `visibilityState` unreliable); test body is preserved for headed local runs |
| `utils/logger.ts` | Guard file transport setup against empty `logDir` — infra fixtures (e.g. seeding) don't have a log directory |

## Test plan
- [x] CI pipeline ran successfully on `e2e/playwright-ci-pipeline` branch with all shards passing
- [x] Login, login-v2, welcome-modal, and incidents tests verified passing in CI
- [x] Skipped carousel autoplay test can be run locally with `--headed` flag


e2e-test/ui/playwright/tests/search/search.spec.ts
```diff
@@ -29,6 +29,7 @@ test.describe('Search Functionality', () => {
   });
 
   test('should search, find a result, and visit the dataset page', async ({ page }) => {
+    test.setTimeout(45000);
```
benjiaming, line 32: Should this have a comment for why we need to explicitly wait longer?

e2e-test/ui/playwright/tests/auth/login.spec.ts
```diff
@@ -59,6 +59,6 @@ test.describe('Login', () => {
     await loginPage.login(username, password);
     await page.waitForLoadState('networkidle');
     await expect(page).toHaveURL(/\/$|\/home/);
-    await expect(page.getByRole('img', { name: /logo/i })).toBeVisible();
+    await expect(page.getByText(/Good (morning|afternoon|evening)/)).toBeVisible({ timeout: 10000 });
```
benjiaming: Explicit wait should have a comment.

benjiaming (approved):
It would be good to look into the reasons for why we must increase the explicit waits in the followup.

e2e-test/ui/playwright/tests/search/search.spec.ts
priyabratadas-dh: This isn't explicitly waiting for anything. This is a timeout I have given to the test so if it takes beyond the given time i.e. 45s to finish the test it would mark the test as failed

e2e-test/ui/playwright/tests/auth/login.spec.ts
priyabratadas-dh: Home page renders a greeting once the authenticated session and React hydration are complete and timeout on a Playwright expect call is not an explicit wait. It tells the auto-retrying assertion how long to keep retrying before failing. The assertion polls the DOM on its own schedule and resolves the moment the element appears — it doesn't sleep for the full 10 s. it the element loads in 30ms it'll complete the same in 30 ms but max it will wait for 20s post which it will mark the same as failed.

.github/workflows/playwright-e2e-tests.yml
```diff
@@ -0,0 +1,270 @@
...
+        default: "quickstart-consumers"
+        type: string
+      shard_count:
+        description: "Number of parallel shards for Playwright"
+        required: false
+        default: "4"
+        type: string
+  push:
+    branches:
+      - e2e/**
```
devashish2203: Remove this before merging.

.github/workflows/playwright-e2e-tests.yml
```diff
@@ -0,0 +1,270 @@
...
+    outputs:
+      test_runner_type: ${{ steps.set-runner.outputs.test_runner_type }}
+      yarn_cache_key: ${{ steps.yarn-cache-key.outputs.yarn_cache_key }}
+      yarn_cache_key_prefix: ${{ steps.yarn-cache-key.outputs.yarn_cache_key_prefix }}
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - name: Check out the repo
+        uses: acryldata/sane-checkout-action@186e92cc5948a9c3e1cc7a96eaff9f776f3fc8e3 # v7
+        with:
+          checkout-head-only: true
```
devashish2203: nit: You can remove checkout-head-only from here. That's only needed when calculating tags.

.github/workflows/playwright-e2e-tests.yml
devashish2203, line 1: Seeing this in the logs during [yarn install](https://github.com/datahub-project/datahub/actions/runs/24801734357/job/72585918612?pr=17154#:~:text=warning%20package%2Dlock.json%20found.%20Your%20project%20contains%20lock%20files%20generated%20by%20tools%20other%20than%20Yarn.%20It%20is%20advised%20not%20to%20mix%20package%20managers%20in%20order%20to%20avoid%20resolution%20inconsistencies%20caused%20by%20unsynchronized%20lock%20files.%20To%20clear%20this%20warning%2C%20remove%20package%2Dlock.json.) I think the package-lock.json file is leftover from the npm to yarn shift.

> warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.

.github/workflows/playwright-e2e-tests.yml
```diff
@@ -0,0 +1,270 @@
...
+                  }
+                }
+              }
+            }'
+
+      - name: Install Playwright dependencies
+        working-directory: e2e-test/ui/playwright
+        run: |
+          yarn install --frozen-lockfile
+          npx playwright install --with-deps chromium
```
devashish2203, line 149: To discuss for later. 
What is our browser testing strategy? Which browsers should we be testing on?
Should we download the latest version of the browser always or use a fixed version? Using latest version has potential to break things when chrome is updated upstream and our playwright version is not updated.

e2e-test/ui/playwright/tests/auth/login.spec.ts
devashish2203: Use previously created constants for timeout values.

e2e-test/ui/playwright/tests/search/search.spec.ts
devashish2203: Yes. I also think we should just increase the default test timeout to maybe 2 or 5 mins if this becomes a frequent requirement. Default currently is 30 secs only.

https://playwright.dev/docs/test-timeouts

.github/workflows/playwright-e2e-tests.yml
```diff
@@ -0,0 +1,270 @@
...
+    if: ${{ !cancelled() && needs.playwright_test.result != 'skipped' }}
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
+        with:
+          node-version: "20"
+
+      - name: Download all Playwright blob reports
```
devashish2203, line 237: Will this add trace information and screenshots to the report if the tests fail?

.github/workflows/playwright-e2e-tests.yml
```diff
@@ -0,0 +1,270 @@
...
+        run: |
+          # download-artifact v4+ places each artifact in its own subfolder under
+          # playwright-blob-reports/. Flatten all .zip files into a single dir so
+          # `playwright merge-reports` can find them.
+          mkdir -p playwright-blobs-flat
+          find playwright-blob-reports -name "*.zip" -exec cp {} playwright-blobs-flat/ \;
+          if compgen -G "playwright-blobs-flat/*.zip" > /dev/null 2>&1; then
+            npx playwright merge-reports --reporter html ./playwright-blobs-flat
+          else
+            echo "No blob reports found — all shards may have been skipped or crashed"
```
devashish2203, line 256: add and exit 1 here, to fail this step as well when not reports are found.

.github/workflows/playwright-e2e-tests.yml
devashish2203: Image
On my local HTML report I don't see the traces.

Screenshots are available though.

e2e-test/ui/playwright/tests/auth/login.spec.ts
devashish2203: We should also consider increasing default expectTimeout if we see us using timeouts in expect statements more often.

devashish2203 (approved):
Mostly minor comments. Main thing to fix is the traces being unavailable in the final merged report.

.github/workflows/playwright-e2e-tests.yml
priyabratadas-dh: You need to run it using the npx playwwright show-report to see traces

.github/workflows/playwright-e2e-tests.yml
priyabratadas-dh: Done

PR #17159, opened 2026-04-22
ci(security): Trivy/Grype scan workflow, Linear sync, and registry profiles

david-leifker:
Runs security scans, makes tickets

### DIFF
--- a/.github/scripts/security_scan_linear_sync.py
+++ b/.github/scripts/security_scan_linear_sync.py
@@ -1,0 +1,1282 @@
+#!/usr/bin/env python3
+"""Sync container security scan reports into Linear issues.
+
+Feature summary:
+- Scanner input modes:
+  - ``trivy``: parse Trivy JSON reports.
+  - ``trivy_grype``: parse mixed Trivy + Grype JSON, normalize and merge by image+CVE.
+- Grouping model:
+  - Create one issue per ``(vulnerability id, image scope)`` where scope is repo basename plus
+    optional variant suffix inferred from configured tag suffixes.
+- Issue title/description generation:
+  - Build deterministic Linear titles from CVE + package + scoped image.
+  - Build rich markdown issue bodies with vulnerability details, affected images, scanner metadata,
+    commit SHA, and workflow run URL.
+- Raw scan report attachments on create:
+  - For newly created issues, attach matching raw scanner JSON files (Trivy/Grype) when
+    ``--raw-report-paths`` is provided.
+  - Upload files via Linear signed upload URLs, then create issue attachments that reference the
+    uploaded assets.
+- Existing issue reconciliation:
+  - Find issues by exact title.
+  - Update refs comment and merge labels on existing issues without removing current labels.
+- Labeling behavior:
+  - Apply optional static labels from ``LINEAR_LABEL_IDS`` / ``TRIVY_LINEAR_LABEL_IDS``.
+  - Apply component labels mapped from image repositories.
+  - Apply a dynamic ref label using ``SCAN_REF_NAME`` (exact branch/tag text), team-scoped in
+    Linear, reusing an existing label when present or creating one with a random color when absent.
+- Refs comment tracking:
+  - Maintain a single marker comment per issue with deduped branch/tag history for where the
+    finding was observed.
+- Severity-based prioritization:
+  - For Trivy-derived findings, map worst severity to Linear priority and due date.
+- Issue relation graph (new issues created in this run):
+  - Build connected components where issues share CVE or package name, then create pairwise
+    ``related`` links across each component (best-effort; duplicate-link errors are ignored).
+- Team/state resolution:
+  - Resolve team via ``LINEAR_TEAM_ID`` (or legacy fallback).
+  - Resolve issue create state from explicit ``LINEAR_ISSUE_STATE_ID`` or team triage state.
+"""
+
+from __future__ import annotations
+
+import argparse
+from collections import defaultdict
+import json
+import os
+import re
+import sys
+from dataclasses import dataclass
+from datetime import datetime, timedelta, timezone
+from pathlib import Path
+from typing import Any, Callable
+
+import requests
+from utils.linear_sync_utils import (
+    attach_file_to_issue as _attach_file_to_issue_util,
+    create_comment as _create_comment_util,
+    create_issue as _create_issue_util,
+    dedupe_preserve_order as _dedupe_preserve_order,
+    find_issue_by_title as _find_issue_by_title_util,
+    get_or_create_team_label_id as _get_or_create_team_label_id_util,
+    get_marker_comment_id as _get_marker_comment_id_util,
+    issue_graphql_label_ids as _issue_graphql_label_ids_util,
+    issue_update_label_ids as _issue_update_label_ids_util,
+    link_issue_related_best_effort as _link_issue_related_best_effort_util,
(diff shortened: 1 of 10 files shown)


.github/scripts/security_scan_linear_sync.py
```diff
@@ -0,0 +1,1550 @@
...
+        LINEAR_GRAPHQL,
+        data=data,
+        headers={
+            "Authorization": api_key,
+            "Content-Type": "application/json",
+        },
+        method="POST",
+    )
+    try:
+        with urllib.request.urlopen(req, timeout=120) as resp:
```
devashish2203: Could we not have used requests library instead of using urllib which is much more lower level?

.github/scripts/security_scan_linear_sync.py
```diff
@@ -0,0 +1,1550 @@
...
+        b = (base or "").strip()
+        if b and b not in seen:
+            seen.add(b)
+            basenames.append(b)
+    return basenames
+
+
+# Baked-in: image repo basename (e.g. datahub-gms) → Acryl Linear label id. Edit in code to add/change.
+_DEFAULT_LINEAR_REPO_LABEL_MAP: dict[str, str] = {
+    "datahub-actions": "75489fcb-ab53-4087-a764-cd699db9c32a",
```
devashish2203, line 149: How does one get these ids if new ones are to be added? Can we add the human readable names as comments here too

.github/workflows/security-scan-linear-sync.yml
```diff
@@ -0,0 +1,958 @@
...
+      - name: Upload image list
+        if: success() && steps.resolve-images.outputs.need_build == 'false'
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
+        with:
+          name: security-scan-image-list
+          path: image-list.txt
+          if-no-files-found: error
+          retention-days: 1
+
+      - name: Select runner for main job
```
devashish2203, line 370: Please use the new determine-runners composable action instead to define the runners to use.

.github/workflows/security-scan-linear-sync.yml
```diff
@@ -0,0 +1,958 @@
...
+            .github/scripts/security_scan_linear_sync.py
+            trivy.yaml
+          sparse-checkout-cone-mode: false
+
+      - &resolve_security_profile_config
+        name: Resolve security scan profile config
+        run: |
+          set -euo pipefail
+          mkdir -p .github
+          git show "HEAD:.github/security-scan-profiles.saas.json" 2>/dev/null > .github/security-scan-profiles.saas.json || echo '{}' > .github/security-scan-profiles.saas.json
```
devashish2203: .github/security-scan-profiles.saas.json is not checked out above.

.github/workflows/security-scan-linear-sync.yml
```diff
@@ -0,0 +1,958 @@
...
+
+      - *apply_scan_profile
+
+      - *resolve_git_ref_linear
+
+      - *ecr_login
+
+      - *resolve_images_from_registry
+
+      - name: Checkout (full, for local build)
```
devashish2203: What is the use case for building the code in this workflow? 
Can we not use the images created via docker-unified.yml and simply pull them for any commit needed?

.github/workflows/security-scan-linear-sync.yml
```diff
@@ -0,0 +1,958 @@
...
+      - name: Materialize full tree for Gradle (undo sparse-checkout)
+        if: steps.resolve-images.outputs.need_build == 'true'
+        run: |
+          set -euo pipefail
+          git sparse-checkout disable || true
+          git checkout HEAD -- .
+          test -f ./gradlew
+
+      # Borrow from docker-unified.yml (same keys/paths): read-only — restore only, never save uv/yarn/~/.gradle buckets here.
+      - uses: actions/cache/restore@0057852bfaa89a56745cba8c7296529d2fc39830 # v4
```
devashish2203: We should create a reusable action for these steps and use it in all workflows that need this docker-unified.yml, docker-unified-nightly.yml and now this.

Although I would still argue that this workflow should not need to build images at all, but maybe I'm missing context.

docker/build.gradle
```diff
@@ -238,6 +239,12 @@ ext {
                         DATAHUB_LOCAL_ACTIONS_ENV: "${rootProject.project(':smoke-test').projectDir}/test_resources/actions/actions.env"
                 ]
             ],
+            'actionsImages': [
```
devashish2203: For a single image using bake doesn't really help right? 
We can run `gw :datahub-actions:docker` instead?

devashish2203 (approved):
A 1500+ line python file and a 1000+ line workflow is going to be hard to maintain/debug in the long term.
For the python file I would suggest breaking it down into modules and place under .github/scripts/utils. Some of the things around interacting with Linear can be reusable modules that other workflows can also benefit from instead of each workflow creating their own custom versiosn.


Okay to merge for the short term but we really need to rethink our whole pipeline and the tasks we want to do from raising the PR to deploying a release version and create a more maintainable pipeline.

RyanHolstien (approved):

.github/scripts/security_scan_linear_sync.py
david-leifker: The key is the plaintext

.github/workflows/security-scan-linear-sync.yml
david-leifker: I'll take a look

.github/workflows/security-scan-linear-sync.yml
david-leifker: The primary goal is to use existing docker tags for sure. There are a few times where a select build is useful when on a PR or branch:

* When we don't want to publish images
* When we want the variants, but don't want to cut a release which generates them

docker/build.gradle
david-leifker: No because we've decided to only build variants in bake

david-leifker: Open to changing this, but the initial intent was to only build variants in bake.

devashish2203: Got it. let's keep it as is for now.

.github/workflows/security-scan-linear-sync.yml
devashish2203: Hmm. I would still like to keep the security scan workflow simple.
We should be able to have a separate workflow to build images for any given commit and push to depot's temporary registry and pull from there in a this security workflow.

That way the security scan workflow remains simple
1. Pull given variants of images for give tag/commit 
2. Run security scans on them
3. Create Linear if required.

devashish2203: I have a few items that I am working on related to this refactor. I can try to incorporate the above in my changes.

.github/workflows/security-scan-linear-sync.yml
devashish2203: Sorry I misunderstood this. We are optionally checking it out if it exists so should be fine.

devashish2203 (approved):

.github/workflows/security-scan-linear-sync.yml
david-leifker: I will remove the build flow for now. Let me know your thoughts on how we can get to the point where a build workflow can run this workflow with a targeted scan on the required images.

PR #17165, opened 2026-04-23
feat(dataplex): add Dataplex Business Glossary ingestion support

sgomezvillamor:
## Overview
This PR adds comprehensive support for ingesting Dataplex Business Glossaries into DataHub as GlossaryNode and GlossaryTerm entities, with optional term-to-asset association resolution.

### New Features
- **Glossary Entity Ingestion**: Ingest Dataplex Business Glossaries, categories, and terms as DataHub GlossaryNode and GlossaryTerm entities, preserving the full Glossary → Category → Term hierarchy
- **Term-to-Asset Associations** (opt-in): Resolve term-to-asset links via the Dataplex `lookupEntryLinks` REST API, emitting `glossaryTerms` aspect updates on linked Dataset entities
- **Configurable Glossary Locations**: Support scanning multiple GCP locations for glossaries (default: `global`)
- **Parallel Processing**: Glossary ingestion and term association resolution use configurable worker threads (`max_workers_glossary`, default: 10)

### New Files
- `dataplex_glossary.py`: Core glossary ingestion processor with two-stage pipeline:
  1. `process_glossaries()` — lists and emits glossary/category/term entities
  2. `process_term_associations()` — resolves term-to-asset links via REST API
- `dataplex_context.py`: Shared context object threaded through all processors, centralizing entry data and project number mappings
- `test_dataplex_glossary.py`: Comprehensive unit tests covering URN helpers, parent parsing, glossary ingestion, and term association resolution

### Modified Files
- `dataplex.py`: 
  - Integrated glossary processor into main source pipeline
  - Added project number resolution at startup (required for `lookupEntryLinks` API)
  - Refactored entry data tracking into shared `DataplexContext`
  - Added `GLOSSARY_TERMS` capability declaration
  
- `dataplex_config.py`: Added configuration options:
  - `include_glossaries` (default: True)
  - `include_glossary_term_associations` (default: True)
  - `glossary_locations` (default: ["global"])
  - `max_workers_glossary` (default: 10, range: 1-100)

- `dataplex_entries.py`: Updated to use shared `DataplexContext` for entry data storage

- `dataplex_report.py`: Integrated `DataplexGlossaryReport` into main report aggregation

- `dataplex_source.py` tests: Updated mocks to reflect new context structure

- `dataplex_entries.py` tests: Updated to use `DataplexContext`

- `setup.py`: Added `google-cloud-resource-manager<2.0.0` dependency for project number resolution

- `source.py`: Added `GLOSSARY_TERMS` to `SourceCapability` enum

## Design Highlights
- **Thread-safe reporting**: `DataplexGlossaryReport` uses internal locks for concurrent worker updates
- **Efficient batching**: Glossary jobs and term associations processed in batches of 200 to balance parallelism and resource usage
- **Graceful degradation**: Failures in individual glossaries or term lookups are logged as warnings and don't block other processing
- **REST API integration**: Uses authenticated session for `lookupEntryLinks` calls with performance timing
- **Deduplication**: Term associations de-duplicated across multiple location scans to avoid duplicate workunits

## Testing
- Added 20+ unit tests covering:
  - URN ID generation and parsing
  - Parent URN resolution for glossaries and categories
  - Glossary/category/term emission
  - Term-asset association resolution
  - Edge cases (unknown entries, non-definition links, 404 responses)
- Existing tests updated to reflect new context structure
- All tests use mocks to avoid external API calls

## Configuration Example
```yaml
source:
  type: dataplex
  config:
    project_ids: ["my-project"]
    include_glossaries: true
    include_glossary_term_associations: true
    glossary_locations: ["global

https://claude.ai/code/session_01CkkM9C668tP7KTg7wncmww

### CODE REVIEW

File: metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_config.py [hunk 1]
```diff
@@ -206,6 +206,44 @@ class DataplexConfig(
         description="Stateful ingestion configuration for stale metadata removal.",
     )
 
+    include_glossaries: bool = Field(
```
treff7es, line 209: Maybe it would be better to group these configs into group and not having individual items:

```
class GlossaryConfig(ConfigModel):
      enabled: bool = True
      include_term_associations: bool = False  # safer default, see below
      locations: List[str] = Field(default_factory=lambda: ["global"])
      max_workers: int = Field(default=10, ge=1, le=100)
```

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_config.py
```diff
@@ -206,6 +206,44 @@ class DataplexConfig(
...
+        default=True,
+        description=(
+            "Whether to ingest Dataplex Business Glossary entities as DataHub GlossaryNodes "
+            "and GlossaryTerms. Glossaries, categories, and terms are emitted with correct "
+            "parent hierarchy. Default: True."
+        ),
+    )
+
+    include_glossary_term_associations: bool = Field(
+        default=True,
```
treff7es: Maybe this should be False by default

Currently default=True. This feature:
  - Requires an extra IAM role (roles/resourcemanager.projectViewer)
  - Issues O(terms × entries_locations × project_ids) REST calls — potentially very expensive
  - Silently produces zero results if the role is missing (warnings, not failures)

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_glossary.py
```diff
@@ -0,0 +1,576 @@
...
+        report: DataplexGlossaryReport,
+        source_report: SourceReport,
+    ) -> None:
+        self._ctx = ctx
+        self._glossary_client = glossary_client
+        self._report = report
+        self._source_report = source_report
+        # (project_id, location, glossary_id, term_id) — populated by process_glossaries,
+        # consumed by process_term_associations.
+        self._emitted_terms: List[Tuple[str, str, str, str]] = []
```
treff7es: This Tuple with 4 items is quite ugly, can we use a dataclass for it?

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_config.py
sgomezvillamor: Yeah, probably... likely even.
But that would introduce a new pattern, making the config feel inconsistent with the rest of configs in the connector itself... unless we refactor all the existing connector configs too.
I will left as it is.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_config.py
sgomezvillamor: Good catch — flipped to `false` in the latest commit.

A bit more context on what was happening: `_resolve_project_numbers` (which requires `roles/resourcemanager.projectViewer`) was being called whenever `include_glossaries=True`, not just for term associations. That was because project numbers were used in two places: the Dataplex console `source_url` for terms, and the `lookupEntryLinks` internal entry path. The source_url usage turned out to be unnecessary — the Dataplex console accepts both project ID and project number in the URL, so we can build it with the project ID directly.

After that cleanup, `_resolve_project_numbers` is now only called when `include_glossary_term_associations=True` (the `lookupEntryLinks` entry path genuinely requires the numeric project number). The default for `include_glossary_term_associations` is now `false`, the config description mentions the IAM requirement, and the docs are updated accordingly.

---

sgomezvillamor: On the two remaining concerns:

**O(terms × entries_locations × project_ids) REST calls**: addressed by the default flip to `false` — this is now opt-in and the description calls out the cost explicitly.

**Silently produces zero results if the role is missing**: the concern isn't quite accurate for either failure path:
- `_resolve_project_numbers` (Resource Manager API) is called at startup in `__init__`. A `PermissionDenied` there propagates immediately and hard-fails the pipeline — it was never silent. That said, it was failing with a raw exception traceback rather than a structured message in the ingestion report. Fixed that in the latest commit: the exception is now caught, recorded via `report.failure(...)` with an explicit message about `roles/resourcemanager.projectViewer`, and then re-raised to preserve the hard-fail behaviour.
- `lookupEntryLinks` failures (e.g. a 403 on individual API calls): each failure is already caught per-location and emitted as `report.warning(...)`, so they do appear in the report — not truly silent.

---

sgomezvillamor: One more note on the O(terms × entries_locations × project_ids) concern: term lookups are already parallelised. `process_term_associations` uses a `ThreadPoolExecutor(max_workers=max_workers_glossary)` — one task is submitted per term, and within each task the location pairs are iterated sequentially. So the wall-clock cost scales with `ceil(terms / max_workers_glossary) × entries_locations × project_ids`, not the full Cartesian product. Updated the config description and docs to reflect this rather than describing it as potentially slow.

---

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_glossary.py
sgomezvillamor: Done — replaced with a `GlossaryTermRef` dataclass. All append and destructuring sites updated to use named field access.

---

treff7es (approved):

PR #17171, opened 2026-04-24
fix(monitoring): limit jaeger-all-in-one in-memory trace storage to prevent OOM

rohansun:
## Problem
The `jaeger-all-in-one` container runs with no limit on the number of traces stored in its in-memory storage, which can lead to OOM and crash the host machine. 
According to [Jaeger’s official documentation](https://www.jaegertracing.io/docs/2.17/storage/memory/), the in-memory storage has no default limit on the number of traces stored.

## Fix
Added the `MEMORY_MAX_TRACES=100000` environment variable to cap the number of traces stored in memory. 
This value is based on the `max-traces` parameter example from [Jaeger’s official documentation.](https://www.jaegertracing.io/docs/2.17/storage/memory/)

The `MEMORY_MAX_TRACES` environment variable is derived from the `--memory.max-traces` CLI flag by changing all letters to upper-case and replacing all punctuation characters with the underscore, per [Jaeger's environment variable naming convention.](https://www.jaegertracing.io/docs/1.76/deployment/cli/)


rohansun:
Hi @david-leifker, following up on this PR. Happy to make any changes if needed. Thank you for your time!

david-leifker (approved):
lgtm

PR #17176, opened 2026-04-24
ci(ingestion): add selective connector test dispatch based on ci-decisions

sgomezvillamor:
## Description

This PR enhances the connector tests trigger workflow to intelligently dispatch tests only for affected connectors, rather than always running the full test suite.

### Changes

- **Extract metadata-ingestion run ID**: Modified the workflow to capture the metadata-ingestion workflow run ID so we can download its artifacts in subsequent steps.

- **Add "Determine target connectors" step**: New step that:
  - Downloads the `ci-decisions` artifact from the metadata-ingestion workflow run
  - Reads the `ci-decisions.json` file to determine which connectors are affected by the PR changes
  - Maps DataHub integration test connector names to connector-tests suite names (e.g., `bigquery_v2` → `bigquery`)
  - Falls back to running all connectors if the artifact is unavailable or if shared/non-connector files changed
  - Outputs the list of target connectors and a skip flag

- **Update dispatch logic**: 
  - Modified the repository dispatch to include the `connectors` parameter when selective testing is needed
  - Updated conditional checks to skip dispatch when no affected connectors are found
  - Enhanced logging to indicate whether all or selective connectors are being tested

- **Update skip conditions**: Added checks to skip the connector tests workflow when no affected connectors are detected by the ci-decisions analysis.

### Benefits

- Reduces CI time by only running tests for connectors affected by PR changes
- Maintains fallback behavior to run all tests when ci-decisions artifact is unavailable
- Provides clear visibility into which connectors are being tested via workflow summary

### Test Plan

The changes are workflow configuration updates. Verification will occur through:
- Existing CI pipeline execution on PRs
- Monitoring that selective dispatch works correctly when ci-decisions artifact is available
- Confirming fallback to full test suite when artifact is unavailable

https://claude.ai/code/session_01X7ptd35qGcGqUHvkGbenVy


treff7es (commented):
We could get the connector name in ci-decisions.json as well because the selective ci script should have this info and then you don't need that extra logic in your script

treff7es (approved):

sgomezvillamor:
@treff7es 

> We could get the connector name in ci-decisions.json as well because the selective ci script should have this info and then you don't need that extra logic in your script

Even in that case, connector name doesn't match names in connector-tests. So mapping would be require in any case, unless we rename dirs in connector-tests too.

treff7es:
> @treff7es
> 
> > We could get the connector name in ci-decisions.json as well because the selective ci script should have this info and then you don't need that extra logic in your script
> 
> Even in that case, connector name doesn't match names in connector-tests. So mapping would be require in any case, unless we rename dirs in connector-tests too.

I would prefer to rename and remove these type of logics. I did a bunch of renaming of connectors when I did selective ci changes.

rajatoss:
LGTM. 
only one suggestion I had was when we add a new connector and its test is added to connector-tests, the mapping in this workflow needs to be updated as well. Otherwise, changes to that connector won't be detected and its tests won't run on the pr.

rajatoss (approved):

sgomezvillamor:
The concern about the static mapping is completely valid — it's a maintenance burden that will need to be kept in sync manually as `acryldata/connector-tests` adds or removes connectors.

That said, eliminating it properly requires two things to align first:
1. `ci-decisions.json` surfacing canonical connector names (not test-dir names like `bigquery_v2`)
2. `acryldata/connector-tests` organising its test directories using those same canonical names

Neither is true today, so the mapping is the pragmatic bridge. I've added a TODO comment in the workflow marking exactly these two prerequisites so it's easy to find and remove in a future iteration.

Happy to merge this as-is to validate the selective dispatch is working end-to-end, and we can clean up the mapping in a follow-up once the underlying naming is aligned.

---

PR #17178, opened 2026-04-24
fix(actions): remove unused platformEvent_V1 topic from executor config

Itsdarkhere:
This topic is not used by the executor. Keeping it configured causes unnecessary Kafka consumption work and CPU overhead. Removed.


asikowitz (approved):
Makes sense to me

PR #17187, opened 2026-04-24
feat(): Adding views for service accounts, ensuring user default view honored 

jjoyce0510:
# Summary

In this PR, we add support for defining default views for Service Accounts. In addition, we align agent context kit with DataHub Cloud with respect to handling default view injection. Now we inject the user's PERSONAL default view first if one exists, otherwise fall back to the global org default (if the view resolution env variable is true) 

Screenshot 2026-04-24 at 11 31 46 AM

# Status

Ready for review.


shirshanka (approved):
Test improvement comments from Otto seem relevant. Take a look and see if its a quick improvement. 

Stamping to unblock.

PR #17191, opened 2026-04-25
fix(ingest/airflow-plugin): sanitize 'None' segments in OpenLineage dataset names

dinesh-verma-datahub:
## Summary

OpenLineage producers sometimes build dataset names with f-strings where one of the interpolated fields is Python `None`, which renders as the four-character string `"None"` and gets baked into the DataHub URN's entity-key. The resulting URNs (e.g. `mydb.None.public.events`) **orphan from the canonical lineage graph** — they never merge with the URN that DataHub's native Redshift / Snowflake / etc. ingestion source emits for the same physical table, so users see disconnected lineage in the UI.

An example we can see in upstream source is Apache Airflow's `S3ToRedshiftOperator.get_openlineage_facets_on_complete`, which constructs the name as [`f"{database}.{self.schema}.{self.table}"`](https://github.com/apache/airflow/blob/67b71d376454cc95cf2f5bb17e0f4edb0e05f480/providers/amazon/src/airflow/providers/amazon/aws/transfers/s3_to_redshift.py#L260) with no `None` guard. The same pattern can appear in any OL producer (other Airflow operators, custom integrations, dbt adapters, Spark listeners), so this PR adds a defensive sanitiser at the OL → DataHub URN boundary in the Airflow plugin rather than fixing any single operator. Customers get the fix on plugin upgrade, regardless of when upstream ships theirs.

## What changed

**In** — `metadata-ingestion-modules/airflow-plugin/src/datahub_airflow_plugin/_datahub_ol_adapter.py`

- New pure helper `_sanitize_ol_dataset_name` strips literal `"None"` and empty segments from dotted names. Returns input by `is`-identity when no rewriting is needed.
- `translate_ol_to_datahub_urn` calls it before constructing the URN — **the only line of existing code that changed**.
- WARNING is emitted via two `lru_cache(512)` helpers, deduplicated per `(original, sanitized)` pair per worker process. A 5-min schedule emitting the same buggy name yields one log line, not 1,440/day.
- All-junk inputs (e.g. `"None.None"`) keep the original to avoid emitting an empty-name URN (which is unsearchable, strictly worse than the orphan).

**Tests** — 33 new tests in `metadata-ingestion-modules/airflow-plugin/tests/unit/test_datahub_ol_adapter.py` covering happy paths (with `is`-identity), every sanitisation position, defensive non-string inputs, case-exact matching, dedupe semantics, and an end-to-end guard that pathological inputs never produce empty URNs.

**Docs**

- `docs/lineage/airflow.md` — new "Orphan dataset URNs containing `.None.`" debug section: explains the producer bug, both warning texts verbatim, per-process dedupe, the all-None edge case, and a `--dry-run`-first cleanup workflow with `undo-by-filter` rollback.
- `docs/how/updating-datahub.md` — release-note bullet under "Next" cross-linking the new section.

## Backward compatibility & risk

- Clean inputs are unchanged (sanitiser short-circuits and returns the input by identity). No URN produced before this PR can change as a result.
- No new dependencies (`functools.lru_cache` is stdlib).
- Customers already affected: future runs emit the corrected URN; historical orphans persist until cleaned up via the documented commands.
- Java/REST OpenLineage endpoint (GMS) is out of scope — separate code path, follow-up tracked.

## Test plan

- [x] `pytest tests/unit/test_datahub_ol_adapter.py` — 33/33 pass
- [x] `ruff check` + `ruff format --check` clean
- [x] `mypy` passes against the airflow-plugin venv (initial CI failure on the tests was fixed by typing against `openlineage.client.run.Dataset` instead of `SimpleNamespace`)
- [x] `./gradlew :datahub-web-react:mdPrettierWriteChanged` — both doc files report unchanged
- [x] End-to-end repro on a local DataHub stack: confirmed orphan URN before, correct URN merge with the native Redshift source's URN after, WARNING fires once with the documented text


metadata-ingestion-modules/airflow-plugin/src/datahub_airflow_plugin/_datahub_ol_adapter.py
```diff
@@ -25,6 +26,68 @@
...
+    A well-known case is Airflow's ``S3ToRedshiftOperator`` when ``schema``
+    is unset (ING-2018), but the bug class is generic.
+
+    Returns the input unchanged on every path that doesn't strictly need
+    rewriting (no-dot, dotted-but-clean, non-string). When every segment is
+    junk we keep the original so the orphan stays *findable* rather than
+    becoming an empty-name URN. Sanitising paths log a deduplicated WARNING.
+    """
+    if not isinstance(name, str):
+        return name  # type: ignore[unreachable]
```
treff7es: This looks weird; it seems the method signature is incorrect, or isinstance isn't needed.
Instead of ignoring it, can we update the signature?

something like this
```
def _sanitize_ol_dataset_name(name: object) -> object:
      if not isinstance(name, str):
          return name
      if "." not in name:
          return name
```

treff7es: Are you sure name can be anything else than str?

metadata-ingestion-modules/airflow-plugin/src/datahub_airflow_plugin/_datahub_ol_adapter.py
```diff
@@ -25,6 +26,68 @@
...
+    logger.warning(
+        "OpenLineage Dataset name %r contained 'None'/empty segments; "
+        "sanitized to %r before producing DataHub URN. Likely upstream bug "
+        "in the producer (unset field interpolated into an f-string).",
+        original,
+        sanitized,
+    )
+
+
+@lru_cache(maxsize=_WARN_DEDUPE_MAXSIZE)
```
treff7es: This is a weird way of deduplication

treff7es: Can we do something simpler and more explicit?

Replace lines 29-53 of _datahub_ol_adapter.py with:

```
  # A single per-process flag is enough: the warning's purpose is to alert
  # operators that some upstream OpenLineage producer is buggy. Once that
  # alarm has fired they go investigate; further occurrences add no signal.
  # Per-distinct-name dedupe would just enumerate the symptom; operators
  # can already do that by searching DataHub for ".None." or empty segments.
  _warning_logged: bool = False


  def _warn_once(message: str, *args: object) -> None:
      global _warning_logged
      if _warning_logged:
          return
      _warning_logged = True
      logger.warning(message, *args)
```

  And in _sanitize_ol_dataset_name, replace the two warning calls:

  # Before:
```
  _warn_all_none_once(name)
```
  # After:
```
  _warn_once(
      "OpenLineage Dataset name %r had only 'None'/empty segments; kept "
      "original to avoid emitting an empty URN. The resulting DataHub URN "
      "will literally contain %r in its name field — search DataHub for "
      "that substring to find orphans and report the upstream producer.",
      name, name,
  )
```

  # Before:
```
  _warn_sanitized_once(name, sanitized)
```
  # After:
```
  _warn_once(
      "OpenLineage Dataset name %r contained 'None'/empty segments; "
      "sanitized to %r before producing DataHub URN. Likely upstream bug "
      "in the producer (unset field interpolated into an f-string).",
      name, sanitized,
  )
```

dinesh-verma-datahub:
@treff7es Thank you for flagging both, i have made changes to address these.

treff7es (approved):
lgtm

PR #17192, opened 2026-04-25
fix(ingestion): patch litellm for GHSA-r75f-5x8p-qvmc

david-leifker:
Upgrade litellm to 1.83.7 and refresh generated lock artifacts to remediate the proxy API key SQL injection issue. Relax the stale iceberg pydantic cap to <3.0.0 so the patched litellm dependency graph resolves.

Made-with: Cursor


treff7es (approved):

david-leifker:
This is somewhat problematic litellm has hard pins on specific dependencies. These hard pins force another critical vulnerability in aiohttp

alexsku (approved):

skrydal (approved):

metadata-ingestion/pyproject.toml
```diff
@@ -2251,6 +2251,18 @@ datahub = ["py.typed", "constraints.txt"]
...
+# (notably vs litellm 1.83.7). Keep minimums aligned with a healthy lockfile, not litellm's pins.
+override-dependencies = [
+    "aiohttp>=3.13.5,<4",
+    "openai>=2.30.0,<3",
+    "click>=8.3.1,<9",
+    "importlib-metadata>=8.7.1,<9",
+    "jsonschema>=4.24.1,<5",
+    "typer>=0.24.1,<0.25",
+    "typer-slim>=0.24.0,<0.25",
+]
```
skrydal, line 2265: This is awful. `litellm` has its own dependencies set via `==` to exact versions.
If we had another package like that, with exact pins, they would conflict, even if they could actually work together. This seems to me like a poison pill for us. This breaks our dependencies forever and we will always need to use overrides, with no guarantee via standard dependencies systems whether our dependencies will work together.

I see no way of maintaining dependency like that, can't we use a different package?

skrydal (changes requested):
If my understanding of consequences is correct, we should stop using `litellm` immediately, due to hard-pinning of dependencies.

sgomezvillamor (commented):
Two long-term maintenance concerns worth tracking here.

**1. litellm violates Python packaging guidelines by exact-pinning its own dependencies**

The [Python Packaging Guide](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/) is clear: a library published to PyPI should declare version ranges, not exact pins. Exact pins belong in application lockfiles (`uv lock`, `pip-compile`, etc.) — environments that are self-contained and never composed with other packages. When a library uses `==`, it fights every other package in the installer's environment for the same slot.

https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/

> It is not considered best practice to use install_requires to pin dependencies to specific versions, or to specify sub-dependencies (i.e. dependencies of your dependencies). This is overly-restrictive, and prevents the user from gaining the benefit of dependency upgrades.

litellm is distributed as a library but acts like an application: its `pyproject.toml` hard-pins everything — `openai==2.24.0`, `httpx==0.28.1`, `tiktoken==0.12.0`, and dozens more. This is a known anti-pattern that has already caused downstream conflicts in langchain, crewAI, mlflow, and google/adk-python. It also means that other packages pinning litellm can lead to unresolvable conflicts for our users.

Worth opening an issue upstream in [BerriAI/litellm](https://github.com/BerriAI/litellm) to flag this. If we get a credible commitment to address it, the tactical workaround introduced here becomes acceptable short-lived tech debt — debt with a known exit. Without that, every litellm bump will repeat this work indefinitely.

**2. The `override-dependencies` block needs a version anchor comment**

The new `override-dependencies` stanza (7 packages) must be manually re-validated on every future litellm bump. Right now there's no record of which litellm version these overrides were tested against, so the next person bumping litellm won't know the scope of the re-validation work.

Minimal fix — add a comment directly above the block:

```toml
# Validated for litellm==1.83.14. Re-check these overrides on any litellm bump —
# litellm pins exact versions of these transitive deps in its wheel metadata.
```

---

PR #17195, opened 2026-04-26
feat(ingest/sigma): add formula bracket-ref extractor (state-machine impl)

kyungsoo-datahub:
## Summary
   Adds a small, standalone utility module that extracts `[Source/Column]`
   bracket references from Sigma formula strings — the foundational piece         
   needed by the Sigma column-level-lineage program. 
                                                                                  
   Implemented as a single-pass, char-by-char state machine. The module
   docstring carries the explicit state transition table + degenerate-input       
   decision matrix as the implementation contract.


treff7es (approved):
lgtm

PR #17206, opened 2026-04-27
feat(ebean): make batch query chunk size configurable

Dutt23:
Prevents full table scans on large IN clauses by chunking
  getLatestAspects() queries into configurable batches (default 375).

  Large batch entity updates (~1000 entities) caused single IN clause
  to exceed MySQL range_optimizer_max_mem_size (8MB default), forcing
  full table scan on metadata_aspect_v2 with row-level locks for hours.

  Chunking respects optimizer memory budget, uses indices normally,
  avoids locked row accumulation.

  Set EBEAN_QUERY_KEYS_COUNT_FOR_BATCH env var to tune (1-5000,
  default 375). Validation at startup logs warnings for invalid values.
  
  
Change optimised tests, by reducing number of query runs to one less as earlier since `systemMetadata` was beign fetched lazily , as it was required in the next steps. The current BatchGet fetches everything so overall query reduction. 
  
More details here : [PFP-3532](https://linear.app/acryl-data/issue/PFP-3532/getlatestaspect-batching)


metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectDao.java
```diff
@@ -205,13 +211,15 @@ public Map<String, Map<String, SystemAspect>> getLatestAspects(
...
-    if (forUpdate && canWrite) {
-      results = server.find(EbeanAspectV2.class).where().idIn(keys).forUpdate().findList();
-    } else {
-      results = server.find(EbeanAspectV2.class).where().idIn(keys).findList();
+    // Use batchGet to chunk large IN clauses and avoid optimizer memory exhaustion
+    // (range_optimizer_max_mem_size)
+    final List<EbeanAspectV2> results =
+        batchGet(new HashSet<>(keys), queryKeysCount, forUpdate && canWrite);
+    for (EbeanAspectV2 result : results) {
+      result.setUrn(result.getKey().getUrn());
```
RyanHolstien: Why is this needed?

metadata-service/configuration/src/main/java/com/linkedin/metadata/config/EbeanConfiguration.java
```diff
@@ -23,6 +27,14 @@ public class EbeanConfiguration {
   private boolean autoCreateDdl;
   private boolean postgresUseIamAuth;
   private String batchGetMethod;
+  private Integer queryKeysCountForBatch = DEFAULT_QUERY_KEYS_COUNT;
 
   public static final EbeanConfiguration testDefault = EbeanConfiguration.builder().build();
+
+  public Integer getQueryKeysCountForBatch() {
```
RyanHolstien: Don't think we want to have this pattern. Defaults usually just live in application.yaml

david-leifker: +1

metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectDao.java
david-leifker: +1

metadata-service/configuration/src/main/java/com/linkedin/metadata/config/EbeanConfiguration.java
Dutt23: Removed

metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectDao.java
Dutt23: The columnMapping("urn", "key.urn") tells Ebean to route the urn SQL column → key.urn. That means the SQL result's urn column is "consumed" for the embedded key. The top-level urn, aspect, version fields on the entity won't be set — they'll be null/0 after the batchGetIn query runs.So callers that access aspectV2.getUrn() / aspectV2.getAspect() directly (the top-level fields) instead of aspectV2.getKey().getUrn() will get null values. The composite key itself is fine; it's the duplicate flat fields that are left unhydrated.

I tried changing the code here, [ebean code](https://github.com/acryldata/datahub-fork/blob/acryl-main/metadata-io/src/main/java/com/linkedin/metadata/entity/ebean/EbeanAspectDao.java#L503) for dual mapping it started failing other tests because the Primary key wasn't being hydrated.

david-leifker (approved):

PR #7602, opened 2023-03-15
fix(search): tags with colons exercises search with urns, must follow…

david-leifker:
… fulltext

## Checklist

- [ ] The PR conforms to DataHub's [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [Commit Message Format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#commit-message-format))
- [ ] Links to related issues (if applicable)
- [ ] Tests for the changes have been added/updated (if applicable)
- [ ] Docs related to the changes have been added/updated (if applicable). If a new feature has been added a Usage Guide has been added for the same.
- [ ] For any breaking change/potential downtime/deprecation/big changes an entry has been made in [Updating DataHub](https://github.com/datahub-project/datahub/blob/master/docs/how/updating-datahub.md)

PR #17210, opened 2026-04-27
fix(ingestion/teradata): clear class-level caches on close to prevent OOM

JohnRTurner:
## Summary

- `TeradataSource.close()` now clears `_tables_cache` and `_table_creator_cache` (class-level, shared across all instances in the same process) so memory is fully released between sequential recipe runs
- `close()` also calls `.cache_clear()` on the three module-level `@lru_cache` functions (`get_schema_columns`, `get_schema_pk_constraints`, `get_schema_foreign_keys`) for the same reason
- Extended `test_close_cleanup` to assert both caches are empty and all LRU caches are cleared after `close()` is called

## Root cause

When running multiple Teradata recipes in the same process (e.g. 10 recipes against ~500k datasets via the Python SDK), class-level caches accumulated `TeradataTable` objects — including full view `request_text` — across every recipe run because `close()` only disposed the SQLAlchemy engine and never cleared these caches. Combined with module-level LRU caches that held per-schema column/PK/FK metadata indefinitely, memory grew monotonically until the Kubernetes pod was OOM-killed.

Fixes Zendesk ticket #7602 (JPMC: Teradata ingestion pod OOM with 10 recipes / 500k datasets).

## Test plan

- [ ] `test_close_cleanup` now pre-populates both class-level caches and asserts they are empty post-`close()`, and asserts all three LRU `cache_info().currsize == 0`
- [ ] All 122 Teradata unit tests pass (`test_teradata_source.py`, `test_teradata_performance.py`, `test_teradata_integration.py`)
- [ ] Pre-commit hooks (lint) pass

## Checklist

- [x] Tests added/updated
- [ ] Docs added/updated (N/A — internal fix)
- [ ] Breaking changes documented (N/A)

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
@@ -1991,18 +1991,21 @@ ORDER by DataBaseName, TableName;
                 self._pooled_engine.dispose()
                 self._pooled_engine = None
 
-        # Clear class-level caches so memory is released between recipe runs in the
-        # same process. Without this, sequential recipes accumulate all TeradataTable
-        # objects (including view request_text) and creator metadata indefinitely.
-        with self._tables_cache_lock:
-            self._tables_cache.clear()
-            self._table_creator_cache.clear()
+        try:
+            # Clear class-level caches so memory is released between recipe runs in the
+            # same process. Without this, sequential recipes accumulate all TeradataTable
+            # objects (including view request_text) and creator metadata indefinitely.
+            with self._tables_cache_lock:
+                self._tables_cache.clear()
+                self._table_creator_cache.clear()
 
-        # Clear module-level LRU caches for the same reason — schema column/PK/FK
-        # data is per-connection and must not carry over to the next recipe run.
-        get_schema_columns.cache_clear()
-        get_schema_pk_constraints.cache_clear()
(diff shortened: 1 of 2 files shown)


metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -1991,5 +1991,18 @@ def close(self) -> None:
...
+        # objects (including view request_text) and creator metadata indefinitely.
+        with self._tables_cache_lock:
+            self._tables_cache.clear()
+            self._table_creator_cache.clear()
+
+        # Clear module-level LRU caches for the same reason — schema column/PK/FK
+        # data is per-connection and must not carry over to the next recipe run.
+        get_schema_columns.cache_clear()
+        get_schema_pk_constraints.cache_clear()
+        get_schema_foreign_keys.cache_clear()
```
treff7es: It would be cleaner to put this into a try catch

treff7es (approved):

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
JohnRTurner: Good catch, thanks! Wrapped both cache-clearing blocks in a try/except with a warning log so super().close() always runs regardless. Updated in the latest commit.

PR #17211, opened 2026-04-27
fix(sdk): preserve URN cluster in DataFlow/DataJob and normalize env for aspects

sgomezvillamor:
## Summary

Fixes a bug where `client.entities.get(urn)` returned a `DataFlow` or `DataJob` with a **different URN** than the one requested — silently renaming entities on any subsequent `upsert()`.

### Root cause: two-layer mismatch

| Layer | Field | Type |
|-------|-------|------|
| URN (`DataFlowKey.cluster`) | free-form string | `"prod"` is valid |
| Aspect (`DataFlowInfo.env` / `DataJobInfo.env`) | `FabricType` enum | only `"PROD"`, `"DEV"`, … are valid |

The Airflow plugin defaults to `DEFAULT_FLOW_CLUSTER = "prod"` (lowercase), which is valid as a URN cluster value but not as a `FabricType` enum. The `DataFlowInfo.env` field carries `addToFilters=true` and drives the **Environment search facet**, so it must be a valid uppercase `FabricType`.

`datahub.api` already handled this correctly via `_get_env()`. `datahub.sdk` did not.

### Changes

**Bug fixes (`datahub.sdk`)**
- `DataFlow._new_from_graph()` was not passing `env=urn.cluster`, silently falling back to `DEFAULT_ENV = "PROD"` and reconstructing the entity under a different URN.
- `DataJob._new_from_graph()` had the same issue.
- `DataFlow.__init__` was writing `env` to the aspect without uppercasing, which would cause `AvroTypeException` on serialize for lowercase values.

**Refactor**
- Extracted `_normalize_env()` as a shared module-level helper in `dataflow.py` (used by both `DataFlow` and `DataJob`), replacing duplicated inline logic. Emits a `logger.debug` when the cluster value is not a valid `FabricType`.

**Documentation**
- Added a detailed comment on `DEFAULT_FLOW_CLUSTER` in `mce_builder.py` explaining the mismatch and the two paths to a proper fix: (A) tighten `DataFlowKey.cluster` to `FabricType` in the model, or (B) decouple cluster from env entirely — both requiring a URN migration.

**Tests** (`tests/unit/sdk_v2/`)
- `DataFlow`: uppercase env, lowercase cluster normalized in aspect, invalid cluster gives `None`, `_new_from_graph` preserves cluster.
- `DataJob`: lowercase cluster normalized in aspect, `_new_from_graph` preserves cluster end-to-end.

## Test plan

- [ ] `tests/unit/sdk_v2/test_dataflow.py` — new env/cluster tests pass
- [ ] `tests/unit/sdk_v2/test_datajob.py` — new env/cluster tests pass
- [ ] Existing sdk_v2 golden-file tests unchanged

---
_Generated by [Claude Code](https://claude.ai/code/session_01MhTp7R5o6HAaJwBWivx257)_


metadata-ingestion/tests/unit/sdk_v2/test_datajob.py
```diff
@@ -435,3 +435,43 @@ def test_datajob_with_non_prod_env() -> None:
...
+    assert "prod" in str(job.urn)
+    assert "PROD" not in str(job.urn)
+    # Aspect env must be uppercase for the Environment search filter.
+    assert job.env == "PROD"
+
+
+def test_datajob_new_from_graph_preserves_cluster() -> None:
+    """_new_from_graph must reconstruct the DataJob (and its embedded DataFlow URN) with
+    the exact cluster from the URN — not silently replace it with DEFAULT_ENV."""
+    import datahub.metadata.schema_classes as models
```
treff7es: Do we need this late import here or can we move it to the top?

metadata-ingestion/tests/unit/sdk_v2/test_dataflow.py
```diff
@@ -190,6 +190,56 @@ def test_client_get_dataflow() -> None:
...
+    d = DataFlow(platform="airflow", name="dag", env="custom-cluster")
+    assert d.urn.cluster == "custom-cluster"
+    assert d.env is None
+
+
+def test_dataflow_new_from_graph_preserves_cluster() -> None:
+    """_new_from_graph must reconstruct the entity with the exact URN cluster,
+    not silently replace it with DEFAULT_ENV.  This is the core of ING-2363:
+    client.entities.get(urn) was returning a DataFlow with a different URN."""
+    import datahub.metadata.schema_classes as models
```
treff7es: Can we move late imports to the top?

metadata-ingestion/src/datahub/sdk/datajob.py
```diff
@@ -65,7 +64,7 @@ def get_urn_type(cls) -> Type[DataJobUrn]:
         """Get the URN type for data jobs."""
         return DataJobUrn
 
-    def __init__(  # noqa: C901
```
treff7es, line 68: nice

treff7es (approved):

metadata-ingestion/tests/unit/sdk_v2/test_datajob.py
sgomezvillamor: Good catch, agreed. Moved `import datahub.metadata.schema_classes as models` to the top of the file and removed the inline import.

---

metadata-ingestion/tests/unit/sdk_v2/test_dataflow.py
sgomezvillamor: Good catch, agreed. Moved `import datahub.metadata.schema_classes as models` to the top of the file and removed the inline import.

---

PR #17212, opened 2026-04-27
refactor(ingest/looker): inject Looker creds via custom ApiSettings

askumar27:
## Summary

The Looker source no longer writes `client_id`, `client_secret`, and `base_url` to `os.environ`. Instead, it constructs a custom `ApiSettings` subclass that returns credentials from in-memory state and passes it to `looker_sdk.init40(config_settings=...)`. Credentials never touch the process environment.

## Motivation

The previous implementation set `LOOKERSDK_CLIENT_ID`, `LOOKERSDK_CLIENT_SECRET`, and `LOOKERSDK_BASE_URL` in `os.environ` immediately before `looker_sdk.init40()`:

```python
os.environ[\"LOOKERSDK_CLIENT_ID\"] = config.client_id
os.environ[\"LOOKERSDK_CLIENT_SECRET\"] = config.client_secret.get_secret_value()
os.environ[\"LOOKERSDK_BASE_URL\"] = config.base_url
self.client = looker_sdk.init40()
```

This is problematic for two reasons:

1. **Security** — `LOOKERSDK_CLIENT_SECRET` is stored in plaintext in the process environment, accessible via `/proc/<pid>/environ` on Linux and to any code running in the same process.
2. **Unnecessary** — the Looker SDK provides a supported mechanism for programmatic credential injection via `init40(config_settings=...)`. The SDK's own README recommends this over env vars for production use.

A previous attempt (release `v0.3.17`) tried to clear the env vars _after_ init, but that doesn't work: `auth_session._login()` is lazy — it calls `self.settings.read_config()` on the first authenticated request, well after any cleanup would run, and the secret is still observable in the meantime.

## Changes Overview

**Refactoring:**
- New private class `_DataHubLookerApiSettings` (subclass of `looker_sdk.rtl.api_settings.ApiSettings`) that overrides `read_config()` to return credentials from instance state.
- `LookerAPI.__init__` now constructs this settings object and passes it via `looker_sdk.init40(config_settings=settings)`. The three `os.environ[...] = ...` writes and the now-unused `import os` are removed.

No changes to `LookerAPIConfig`, `transport_options`, retry logic, or any other source files. The configuration schema is unchanged.

## Architecture / Design Notes

The Looker SDK forces this approach. `ApiSettings` deliberately does **not** store `client_id` / `client_secret` as instance attributes — they only live inside the dict returned by `read_config()`, which `auth_session._login` calls lazily on every login. So:

- Setting `settings.client_id = ...` would be ignored.
- Setting then clearing `LOOKERSDK_*` env vars doesn't work because login is lazy.
- Subclassing and overriding `read_config()` is the SDK's documented extension point.

## Testing

**New unit tests** (`metadata-ingestion/tests/unit/looker/test_looker_lib_wrapper.py`):
- `test_credentials_not_written_to_environ` — asserts no `LOOKERSDK_*` keys appear in `os.environ` after `LookerAPI` construction.
- `test_init40_called_with_in_memory_config_settings` — asserts `looker_sdk.init40` receives a `_DataHubLookerApiSettings` whose `read_config()` returns the supplied credentials.
- `test_settings_read_config_ignores_lookersdk_env_vars` — asserts `read_config()` returns the in-memory creds even when `LOOKERSDK_*` env vars are set on the process.

**Existing tests:** all 74 looker/lookml unit tests pass. The looker integration tests that mock `looker_sdk.init40` are unaffected; the one that constructs a real `LookerAPI` (`test_independent_soft_deleted_looks`) passes.

**Live verification:** ran both `looker` and `lookml` ingestion against a real Looker instance.

Both runs successfully authenticated, proving the new `read_config()` path works end-to-end (login flow could not have completed otherwise).

**Lint:** `./gradlew :metadata-ingestion:lint` clean (ruff + ruff format + mypy on 2,285 files).

## Impact Assessment

- **Affected components:** `metadata-ingestion` Looker / LookML connectors only.
- **Breaking changes:** None. `LookerAPIConfig` schema, recipe format, and external behavior are unchanged. Anyone relying on the side effect of `LOOKERSDK_*` env vars being set after a Looker ingest run will find them missing — this is a private SDK detail and was not a documented contract.
- **Performance impact:** None.
- **Risk level:** Low. Tightly scoped (one source file, ~36 added / 7 removed lines), validated by both unit tests and a live ingestion run. No config or schema migration.

## Deployment Notes

None required.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -248,6 +248,7 @@ Forgetting step 2 means the release note is published but never appears in the s
   - **Data Structures**: Prefer dataclasses/pydantic for internal data, return dataclasses over tuples
   - **Code Quality**: Avoid global state, use named arguments, don't re-export in `__init__.py`, refactor repetitive code
   - **Error Handling**: Robust error handling with layers of protection for known failure points
+  - **Security**: Never pass credentials to third-party SDKs via `os.environ`. Use the SDK's programmatic injection mechanism (a settings object, client constructor argument, or credential provider). Writing secrets to the process environment exposes them via `/proc/<pid>/environ` and to any code in the same process. See [`looker_lib_wrapper.py`](metadata-ingestion/src/datahub/ingestion/source/looker/looker_lib_wrapper.py) (`_DataHubLookerApiSettings`) for the canonical pattern.
 - **TypeScript**: Use Prettier formatting, strict types (no `any`), React Testing Library
 
 ### Frontend Theming (Colors)
--- a/metadata-ingestion/src/datahub/ingestion/source/looker/looker_lib_wrapper.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/looker/looker_lib_wrapper.py
@@ -87,14 +87,16 @@ class _DataHubLookerApiSettings(looker_api_settings.ApiSettings):
 
     The default ApiSettings reads credentials from a ``looker.ini`` file or
     ``LOOKERSDK_*`` env vars. We override ``read_config`` so secrets never touch
-    the process environment (CWE-526).
+    the process environment.
     """
 
     def __init__(self, client_id: str, client_secret: str, base_url: str) -> None:
         self._client_id = client_id
         self._client_secret = client_secret
         self._base_url = base_url
-        # filename="" skips the looker.ini lookup; env_prefix=None disables env overrides.
+        # Our read_config() override is the actual guard against env/file lookup.
+        # filename="" and env_prefix=None just keep the parent __init__'s helpers
+        # (which it would otherwise call) from touching the filesystem or environment.
         super().__init__(
             filename="",
             section=None,
--- a/metadata-ingestion/src/datahub/ingestion/source/looker/looker_lib_wrapper.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/looker/looker_lib_wrapper.py
@@ -1,7 +1,6 @@
 # Looker SDK is imported here and higher level wrapper functions/classes are provided to interact with Looker Server
 import json
 import logging
-import os
 from enum import Enum
 from functools import lru_cache
 from typing import Dict, List, MutableMapping, Optional, Sequence, Set, Union, cast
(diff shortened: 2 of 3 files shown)


sgomezvillamor (approved):
Clean fix for a real security issue. The approach is correct — using the SDK's documented extension point avoids both the plaintext env var exposure (CWE-526) and the lazy-auth race window. `env_prefix=None` is verified to work: the SDK guards with `if self.env_prefix:`, so `None` is falsy and `_override_from_env()` is never called. The attribute ordering before `super().__init__()` is also correct and necessary (parent calls `self.read_config()` immediately via polymorphic dispatch).

A few non-blocking suggestions:

**`_client_secret` stored as plain `str`**
`self._client_secret` is accessible via `vars(settings)` / `__dict__`, so any debug dump or structured log of the settings object leaks the secret in plaintext. Worth overriding `__repr__`:
```python
def __repr__(self) -> str:
    return f"_DataHubLookerApiSettings(base_url={self._base_url!r}, client_id={self._client_id!r}, client_secret=***)"
```

**`env_prefix=None` comment could be more precise**
The comment says "disables env overrides" but `None` actually sets the prefix to empty, so the SDK looks for `_CLIENT_ID` etc. (no prefix) — which effectively disables `LOOKERSDK_*` reading. Since `read_config()` is the real guard anyway, a comment like `# read_config() is the actual guard; env_prefix=None is belt-and-suspenders` is less misleading.

**`import os` inside test function body** (`test_credentials_not_written_to_environ`)
Should be at the top of the file with the other imports.

**`_clean_input()` silent-drop on empty credentials is untested**
The SDK's `_clean_input()` silently strips keys whose value is `""`. If `client_id=""` is passed, the SDK fails at first API call with a generic auth error rather than a clear config error. A test documenting this behaviour (or asserting on the message) would prevent silent regression.

**`clean_lookersdk_env` fixture is inert on `test_init40_called_with_in_memory_config_settings`**
`looker_sdk.init40` is fully mocked in that test, so env vars have no effect on the outcome. The fixture provides false assurance; either drop it from that test or add a comment.

**Follow-up: prevent this pattern from being reintroduced**
Would be worth adding to `AGENTS.md` (Python → Code Quality) and `standards/patterns.md` a short rule: _never use `os.environ` to pass credentials to third-party SDKs — use the SDK's programmatic injection mechanism_. This PR is the canonical example of both the anti-pattern and the fix.

---

sgomezvillamor (approved):

PR #17213, opened 2026-04-27
feat(cli): Add capability to pass recipe + secrets via stdin

skrydal:
Backward compatible capability to pass recipe + secrets via stdin, to avoid using environmental variables for that.


rajatoss:
looks good to me.

rajatoss (approved):

PR #17215, opened 2026-04-28
feat(ingest/fabric-onelake): add view ingestion with column-level lineage

aviraj-gour:
### View Extraction

Discovers views via `INFORMATION_SCHEMA.VIEWS` on the SQL Analytics Endpoint (same connection as table schema extraction) and emits them as Dataset entities with `viewProperties.viewLogic`, `subTypes=["View"]`, and column metadata. New `extract_views` flag (default `true`) and `view_pattern` allow-deny filter.

### View → Table Lineage

A `SqlParsingAggregator` parses each view DDL with sqlglot (T-SQL dialect) and emits both table-level and column-level upstream lineage. Aggregator drain runs once after all workspaces are processed so cross-item view → table references can resolve.

### `fabric-onelake` Wired into `sql_parsing_common`

Registered as `tsql` dialect and added to `DIALECTS_WITH_CASE_INSENSITIVE_COLS` so column lineage resolves against the Fabric/SQL Server case-insensitive collation.


sgomezvillamor (commented):
Good implementation overall — correct use of `SqlParsingAggregator`, proper dialect registration (`tsql`), `parse_view_lineage=False` to avoid duplicate lineage, and a comprehensive golden file. A few items to address:

**Warnings**

- **HTTP session leak:** `self.client.close()` is called in `get_workunits_internal()`'s `finally` block, but **not** in the new `close()` override. If the pipeline aborts before the generator is exhausted, the HTTP session leaks. Fix: move `self.client.close()` into `close()` alongside `self.aggregator.close()`.

- **Silent lineage skip for `NULL` `view_definition`:** When a user lacks `VIEW DEFINITION` permission, `view_definition=None` and `add_view_definition()` is silently skipped — no warning, no counter. Users will see views without lineage with no indication why. Add a `views_skipped_no_definition` counter to the report and emit a `report_warning` in the else branch.

- **Missing `exc=e` in `_create_schema_client_and_map`:** The `report_warning` call omits `exc=e` (no stack trace captured) and the `logger.warning()` above it lacks `exc_info=True`. For ODBC/auth failures `str(e)` alone is rarely enough to diagnose the root cause.

- **`extract_views` / `extract_schema` coupling:** The validator requires `extract_schema.enabled=True` as a prerequisite for `extract_views=True`, but these are independent features — both happen to need `sql_endpoint`, not each other. A user may want views without column schema (or vice versa). The prerequisite should be: if either feature is enabled, `sql_endpoint` must be configured. This also suggests `SqlAnalyticsEndpointClient` may belong in its own file rather than `schema_client.py` (which implies schema extraction specifically).

- **`_process_item_views` outer `except Exception` is too broad:** View discovery failures, per-view dataset creation failures, and aggregator rejections all produce the same "Unable to discover views" message. Consider separate try/except blocks per phase.

- **No dedicated warning when views are skipped due to `schema_client is None`:** The "Schema Extraction Failed" message buries the view-skip note in its body text. Add a distinct `title="View Extraction Skipped"` warning so it's easy to spot in the report.

- **`fabric-onelake_post.md` mixes features and prerequisites:** The prerequisites section (SQL Analytics Endpoint requirements, `VIEW DEFINITION` permission) should live in `fabric-onelake_pre.md`, not `_post.md`.

- **Missing view test in connector-tests:** A test case for view ingestion should be added to the connector-tests suite alongside the existing table tests.

**Suggestions**

- **`_FABRIC_SYSTEM_SCHEMAS` adds `queryinsights`:** This silently changes existing table extraction — any tables in a `queryinsights` schema will now be excluded. Worth a one-liner in `docs/how/updating-datahub.md`. Also: are there other Fabric system schemas we should be filtering out?

- **Aggregator drain `context=""`:** If `gen_metadata()` fails mid-iteration, partial lineage is already in DataHub. At minimum track the MCP count before the failure: `context=f"mcps_emitted_before_failure={mcp_count}"`.


---

sgomezvillamor (commented):
Thanks for the updates — significant progress here. Going through each concern:

**✅ Addressed**

- **HTTP session leak** — `close()` now correctly calls `self.client.close()` before `super().close()`. Generator `finally` removed.
- **Silent lineage skip for `NULL` view definition** — `views_missing_definition: LossyList[str]` added to the report with `report_view_missing_definition()`, called in the else-branch of `_create_view_dataset`. 
- **`exc=e` / `exc_info=True`** — both present in the new `_create_schema_client` exception handler.
- **`extract_views` / `extract_schema` coupling** — `validate_sql_endpoint_dependencies` correctly checks each feature independently: `needs_sql_endpoint = self.extract_views or (self.extract_schema.enabled and ...)`. 
- **Doc structure (`_pre` / `_post`)** — SQL Analytics Endpoint setup and `VIEW DEFINITION` permission guidance moved to `_pre.md`. 
- **Connector test for warehouse views** — `test_fabric_onelake_warehouse_with_views` added with golden file. 

---

**❌ Not yet addressed**

- **`_FABRIC_SYSTEM_SCHEMAS` / `updating-datahub.md`** — no entry added for the `queryinsights` schema exclusion change.
- **Aggregator drain `context=""`** — still empty string; a count of MCPs emitted before failure would help diagnose partial lineage.

---

**🆕 Stale doc comments introduced by the coupling fix**

The decoupling of `extract_views` from `extract_schema.enabled` is correct in `config.py`, but several doc comments weren't updated to match:

1. `_pre.md` (View Extraction section): *"views are skipped unless both `extract_schema.enabled` and `sql_endpoint.enabled` are `true`"* — should now say only `sql_endpoint.enabled` is required.

2. `_post.md` advanced config YAML comment: `extract_views: true # Requires extract_schema.enabled and sql_endpoint.enabled` — incorrect.

3. `_post.md` View Extraction config block:
   ```yaml
   # View extraction depends on these — both default to enabled.
   extract_schema:
     enabled: true
   ```
   This implies `extract_views` still depends on `extract_schema`, which is no longer true.

4. `_post.md` Limitations: *"If `extract_schema.enabled` or `sql_endpoint.enabled` is `false` … views … will not be ingested"* — `extract_schema.enabled` should be removed from that condition.


---

aviraj-gour:
> _FABRIC_SYSTEM_SCHEMAS / updating-datahub.md — no entry added for the queryinsights schema exclusion change.

Updated `updating-datahub.md`.

> Aggregator drain context="" — still empty string; a count of MCPs emitted before failure would help diagnose partial lineage.

emitted counter added at [source.py:262-273](vscode-webview://0hrot1oiova3h17e63b3d48dumjft8se4sfilknf7vjragm676ki/metadata-ingestion/src/datahub/ingestion/source/fabric/onelake/source.py#L262-L273); now passes context=f"mcps_emitted_before_failure={emitted}".

> Stale doc comments introduced by the coupling fix.

All spots has been fixed.

sgomezvillamor (approved):
LGTM
Thanks for addressing all the comments


Please, keep this in mind:

> Missing view test in connector-tests: A test case for view ingestion should be added to the connector-tests suite alongside the existing table tests.

PR #17217, opened 2026-04-28
feat(ingest/fivetran): REST API mode and Managed Data Lake destination support

treff7es:
## Summary

The Fivetran connector previously required customers to set up a **Fivetran Platform Connector** that writes log data to a Snowflake / BigQuery / Databricks warehouse. That Platform Connector is a higher-tier Fivetran feature (Business / Enterprise plans) — customers on lower plans, or customers whose data lives only in a Managed Data Lake (no warehouse), couldn't ingest Fivetran metadata into DataHub at all.

This PR adds a new **REST-API ingestion mode** that reads connectors / schemas / users / sync history / destination routing directly via Fivetran's REST API (broadly available across Fivetran plans). No Platform Connector required, no warehouse credentials needed. It also extends destination handling to Fivetran's **Managed Data Lake Service (MDL)** across AWS S3, Google Cloud Storage, and Azure Data Lake Storage Gen2.

### REST API mode (`log_source: rest_api`) — the headline feature

- Works without any Fivetran Platform Connector setup.
- Accessible to customers who don't have access to (or haven't enabled) the higher-tier Fivetran log-warehouse feature.
- Reads everything via the public REST API: connectors, full table+column lineage, users, destination routing, Google Sheets connection details.
- Per-table column-lineage fetch from `/v1/connections/{id}/schemas/{schema}/tables/{table}/columns` for full column coverage.
- Parallel per-connector REST work with bounded worker pool (`rest_api_max_workers`, `rest_api_per_connector_timeout_sec`).
- Per-destination cache shared between source and log reader; negative cache for failed destination discovery.
- Per-destination warning dedup — one warning per destination (not per lineage edge) for URN-construction failures.

### DB mode (`log_source: log_database`) — existing, unchanged

The existing path that reads the Fivetran Platform Connector log from Snowflake / BigQuery / Databricks warehouses. Continues to work as before; nothing in this PR breaks existing recipes.

### Hybrid mode (both blocks supplied)

When both credential blocks are configured, the connector runs in hybrid:

- Default: **DB-primary** — DB owns lineage / users / sync history; REST owns destination routing + Google Sheets details.
- Set `log_source: rest_api` explicitly for **REST-primary hybrid** — REST owns connectors / schemas / users / routing; the DB log fills in per-run `DataProcessInstance` events (REST has no sync-history endpoint) and higher-fidelity column lineage.

Log source is auto-inferred from which credential blocks are present; an explicit `log_source` only needs setting for the REST-primary-hybrid case.

### Managed Data Lake destination support

Fivetran's REST API reports `service: managed_data_lake` for every MDL destination regardless of underlying cloud (AWS / GCS / ADLS) or catalog (Polaris / Iceberg REST / Glue). The connector discovers the destination via `/v1/destinations/{id}` and routes URNs based on (in precedence order):

1. User override on `destination_to_platform_instance.<id>.platform`.
2. `glue` if the user supplied `database` (only Glue uses `database` among MDL platforms — protects against silent drops).
3. Auto-detect from MDL config toggles — `glue` when `should_maintain_tables_in_glue: true` is set on the destination.
4. Fallback to `iceberg` (Polaris / Iceberg REST default).

| Pin in `destination_to_platform_instance.<id>.platform` | URN shape |
| --- | --- |
| `iceberg` (default — no pin needed) | `iceberg.<schema>.<table>` |
| `glue` (or auto-detected) | `glue.<database>.<schema>.<table>` (user supplies actual Glue DB name) |
| `s3` | `s3.<bucket>/<prefix>/<schema>/<table>` (auto-discovered) |
| `gcs` | `gcs.<bucket>/<prefix>/<schema>/<table>` (auto-discovered) |
| `abs` | `abs.<storage_account>/<container>/<prefix>/<schema>/<table>` (auto-discovered) |

The Databricks-backed MDL variant (`service: databricks_via_managed_data_lake`) routes via the existing relational service map to `databricks`.

### Other improvements

- Google Sheets connector workaround extracted to a dedicated handler (`google_sheets_handler.py`).
- DB-log fallback for column lineage in REST-primary hybrid mode (DB log carries explicit `source_column_name` / `destination_column_name`).
- `fivetran_log_config` becomes optional when `log_source: rest_api`.

## Recipe — minimal REST-only setup (no warehouse credentials)

The smallest viable recipe for a customer on a lower Fivetran tier with a Managed Data Lake destination:

```yaml
source:
  type: fivetran
  config:
    log_source: rest_api
    api_config:
      api_key: "${FIVETRAN_API_KEY}"
      api_secret: "${FIVETRAN_API_SECRET}"
    # For Glue-backed MDL: pin the actual Glue database name (visible
    # in the AWS Glue console). `platform: glue` is auto-detected from
    # `should_maintain_tables_in_glue: true` on the destination response.
    destination_to_platform_instance:
      <destination_id>:
        database: <actual Glue database name>
        platform_instance: <glue_platform_instance>
        env: PROD
```

## Test plan

- 209 fivetran tests pass (unit + integration), including new integration tests for hybrid discovery, REST-only mode, and DB-vs-REST equivalence.
- mypy clean for files modified by this branch (5 errors remain on the branch — all pre-existing on master from unrelated commits, not in scope).
- Live-verified against a real customer Fivetran account in `rest_api` mode and DB-primary hybrid mode: 23 of 25 connectors with column lineage; 5,333 column-level lineage entries.
- Comprehensive capability matrix and credential-coverage tables in `fivetran_pre.md`; per-platform URN-shape table for MDL routing; storage-source URN-alignment caveat documented.

## Checklist

- [x] PR conforms to PR Title Format
- [x] Links related issue (CUS-7715)
- [x] Tests added/updated (209 passing)
- [x] Docs updated (`metadata-ingestion/docs/sources/fivetran/fivetran_pre.md`)
- [ ] No breaking changes — all existing recipes run unchanged


treff7es:
## Connector-tests CI status — explanation

The `connector-tests / fivetran` job is failing on this PR. After investigation, this is **a deliberate, correctness-driven behaviour change**, not a regression. The connector-tests golden file codifies a latent bug that this PR fixes.

### Root cause

**Master's behaviour** (relevant code: `fivetran/fivetran.py` on master):

```python
destination_details = self.config.destination_to_platform_instance.get(
    connector.destination_id, PlatformDetail()
)
if destination_details.platform is None:
    destination_details.platform = self.config.fivetran_log_config.destination_platform  # <-- always uses the LOG destination
if destination_details.database is None:
    destination_details.database = self.audit_log.fivetran_log_database
```

In hybrid mode (when both `fivetran_log_config` and `api_config` are configured), master never queries Fivetran's REST API to discover each destination's actual platform. It defaults every destination to the **log warehouse's** platform/database, regardless of where the connector's data physically goes.

**This branch's behaviour:** REST destination discovery is always-on when `api_config` is configured. The connector calls `GET /v1/destinations/{id}` per destination and uses the actual `service` returned by Fivetran. URNs match where the data actually lives.

### Evidence from the connector-tests Fivetran account

Extracted from the [failing CI run](https://github.com/acryldata/connector-tests/actions/runs/25152377032/job/73725810969)'s diff output, the test Fivetran account has **4 distinct destinations**:

| Platform | Database | Notes |
|---|---|---|
| `bigquery` | `harshal-playground-306419` | The log destination (matches `fivetran_log_config.bigquery_destination_config`) |
| `bigquery` | `acryl-prod` | A second BigQuery project — different from the log |
| `snowflake` | `datahub_community` | A Snowflake destination |
| `snowflake` | `fivetran_smoke_test_db` | Another Snowflake destination |

**Master's `fivetran_with_bigquery_dest_golden.json`:** 42,463 destination URN mentions, ALL `bigquery / harshal-playground-306419`. All four destinations conflated into one.

**Master's `fivetran_with_snowflake_dest_golden.json`:** 42,115 destination URN mentions, ALL `snowflake / fivetran_smoke_test_db`. Same Fivetran account, but because the log was switched to Snowflake, all destinations got mirrored to that Snowflake URL instead.

The two goldens have **mutually inconsistent URNs for the same logical connectors**, even though the underlying Fivetran account is identical. That's the smoking gun: master's URN routing was driven by the log destination, not the actual destination. Lineage to URNs like `bigquery.harshal-playground-306419.<snowflake-only-table>` goes nowhere — the table doesn't exist in that BigQuery project.

### This branch's output

REST discovery resolves each destination independently → all 4 platform/database pairs appear correctly in the emitted URNs. As a side effect, **`fivetran_with_bigquery_dest` and `fivetran_with_snowflake_dest` should now produce identical lineage outputs**, because the actual destinations don't depend on which log warehouse the connector reads from.

### Impact + migration

Documented in [`docs/how/updating-datahub.md`](docs/how/updating-datahub.md) under **Next → Breaking Changes**:

> **(Ingestion / Fivetran — Lineage URN shift for multi-destination accounts).** Per-destination URN routing now reflects each destination's actual platform instead of defaulting every destination to `fivetran_log_config.destination_platform`. … Single-destination accounts and recipes that already pin every destination explicitly are unaffected. To opt out and preserve the old behavior, pin each non-default destination explicitly via `destination_to_platform_instance.<destination_id>.platform`.

### Action for the connector-tests repo

The test goldens need regeneration to capture the now-correct URNs:

```bash
pytest --update-golden-files smoke-test/integration/fivetran/test_fivetran_with_bigquery_dest.py
pytest --update-golden-files smoke-test/integration/fivetran/test_fivetran_with_snowflake_dest.py
pytest --update-golden-files smoke-test/integration/fivetran/test_fivetran_with_databricks_dest.py
```

The two goldens should converge to (nearly) the same content modulo the log-source-specific DPI metadata, since the actual destination set is identical between runs.

Happy to do the regeneration as a follow-up PR against `acryldata/connector-tests` if helpful, but I don't have write access to that repo from this branch.

metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran.py
```diff
@@ -252,6 +381,347 @@ def _extend_lineage(self, connector: Connector, datajob: DataJob) -> Dict[str, s
...
+        # customer's Glue catalog uses a different table-name convention,
+        # the URN won't align — they must override per-destination.
+        if destination_details.database is None:
+            raise ValueError(
+                "destination_details.database must be set before constructing "
+                "the destination URN."
+            )
+        return DatasetUrn.create_from_ids(
+            platform_id=destination_details.platform,
+            table_name=f"{destination_details.database.lower()}.{table_name}",
```
askumar27: issue (blocking): Do we need to wrap this in a config to covert to lower case?

metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran.py
askumar27: issue(blocking): May be we a config to control this - source_table URN construction

metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran_rest_api.py
```diff
@@ -104,3 +115,174 @@ def get_connection_details_by_id(
...
+        payload = response.json()
+
+        code = payload.get("code")
+        if code != "Success":
+            raise ValueError(
+                f"Fivetran API returned non-success code {code!r} for "
+                f"destination {destination_id!r}: {payload.get('message')}"
+            )
+
+        details = FivetranDestinationDetails.model_validate(payload["data"])
```
askumar27: Direct key access

askumar27: Similar pattern in the file - needs to be reviewed

metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran_log_db_reader.py
```diff
@@ -0,0 +1,517 @@
...
+                connector_name = connector[Constant.CONNECTOR_NAME]
+                # Match either `connector_name` (the long-standing DB-mode
+                # behaviour) or `connector_id` (matches REST mode and the
+                # recipe documentation: "Fivetran connector IDs are visible
+                # in the UI under Connectors → Setup"). Allowing either
+                # avoids breaking existing recipes while letting users
+                # write portable patterns that work in both modes.
+                if not (
+                    connector_patterns.allowed(connector_name)
+                    or connector_patterns.allowed(connector_id)
```
askumar27: issue (blocking): Previously DB mode matched connector_name only.

For an existing recipe with a deny pattern like `deny: ["abc123"]`, this could now also deny connectors whose ID happens to contain that substring — silently dropping connectors that previously passed

metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran.py
```diff
@@ -252,6 +381,347 @@ def _extend_lineage(self, connector: Connector, datajob: DataJob) -> Dict[str, s
...
+        mapping = FivetranSource._RELATIONAL_SERVICE_MAP.get(discovered.service)
+        if mapping is None:
+            # Best-effort fallback: use the service name verbatim as the
+            # URN platform and the `database` field if present. Better than
+            # crashing for an unknown service; the user can override via
+            # `destination_to_platform_instance` if the platform name needs
+            # correction (e.g., to align with another DataHub source).
+            return base.model_copy(
+                update={
+                    "platform": base.platform or discovered.service,
```
askumar27: issue (blocking): If discovered.service is not in _RELATIONAL_SERVICE_MAP and the user hasn't pinned a platform, line 647 uses the raw service string as the URN's dataPlatform. This silently produces e.g. urn:li:dataPlatform:aurora_postgres_warehouse_v2  — junk that doesn't align with any DataHub source.

metadata-ingestion/src/datahub/ingestion/source/fivetran/response_models.py
```diff
@@ -82,3 +82,170 @@ class FivetranConnectionDetails(BaseModel):
...
+    completed_at: Optional[datetime.datetime] = None
+    status: str
+    message: Optional[str] = None
+
+
+class FivetranSyncHistoryResponse(BaseModel):
+    model_config = ConfigDict(extra="ignore")
+
+    items: List[FivetranSyncHistoryItem]
+    next_cursor: Optional[str] = None
```
askumar27: Not used in code.

askumar27 (changes requested):
issue (blocking): `report_connectors_scanned()` is called in two places for REST mode:
1. `fivetran_log_rest_reader.py:263` — inside `get_allowed_connectors_list` after `f.result()`
2. `fivetran.py:817` — inside `_get_connector_workunits` (the sole location in DB mode)

Result: the ingest report shows 2x the actual connector count for REST mode. On a 200-connector account, the report shows 400. This erodes trust in report metrics and could trigger false-positive threshold alerts.

issue (blocking): The two implementations handle the empty-string sentinel differently:
fivetran_log_rest_reader.py:134   if user_id is None:   ← passes "" through
fivetran_log_db_reader.py:365     if not user_id:       ← short-circuits on ""

Connectors with no connected_by user get "" assigned as a sentinel earlier in the REST reader. REST mode then traverses all groups via list_users for every such connector trying to look up "" in the cache — unnecessary API calls on every cache miss.

---

Adding a REST API ingestion path to a mature connector while maintaining backward compatibility is non-trivial, and the implementation shows real care! There are some blocking issues that need to be addressed. None require architectural rethinking.

treff7es:
Thanks for the careful pass — both blockers were real. All 7 inline comments + both review-level blockers addressed in `8729960fa2` (squashed). Per-comment replies inline.

Quick summary of what changed in `8729960fa2`:
- `report_connectors_scanned()` now counted only at the canonical site in `fivetran.py:817`; the duplicate REST-side increment was removed.
- REST `get_user_email` now uses `if not user_id:` to match the DB reader, so the empty-string sentinel short-circuits as expected (no more N-group `list_users` walk per anonymous connector).
- `connector_patterns` matching restored to name-only in DB mode to preserve master's contract; REST keeps the OR-match (no compat baseline there). Comment in REST reader spells out the asymmetry.
- Unknown-service fallback no longer synthesizes a URN platform from the raw `service` string. `platform=None` flows through; the existing per-destination dedup'd warning fires once. Regression test added.
- All `payload["data"]` direct accesses on the new REST paths route through a new `_extract_data(payload, *, context)` helper.
- Per-`PlatformDetail` `database_lowercase: bool = True` toggle (default preserves existing behaviour) addresses the lowercase concern; `database_for_urn` `@property` centralises the rule for both source-URN and destination-URN sites. Plain `@property` (not `@computed_field`) so it stays out of `model_dump()` and doesn't leak into `_compose_custom_properties`.
- Dropped unused `FivetranSyncHistoryResponse` / `FivetranSyncHistoryItem`.

174 unit + 43 integration tests passing, mypy clean.

metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran.py
treff7es: Done in `8729960fa2`. Added `database_lowercase: bool = True` to `PlatformDetail` plus a `database_for_urn` `@property` that centralises the rule. Default keeps the existing lower-case URN behaviour (master had `database.lower()` here too); users who need to align with another DataHub source whose URN preserves database casing can opt out per-destination via `destination_to_platform_instance.<id>.database_lowercase: false`. Same toggle covers the source-URN path (your other comment). Plain `@property` rather than `@computed_field` so it doesn't appear in `model_dump()` and pollute the `customProperties` aspect via `_compose_custom_properties`. Tests in `TestDatabaseForUrn` (incl. `test_property_does_not_appear_in_model_dump` regression) and `test_glue_database_case_preserved_when_opted_out`.

metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran.py
treff7es: Same toggle covers the source side — see reply on the destination URN comment. The new `PlatformDetail.database_lowercase` field is honoured by the `database_for_urn` `@property` at both call sites, so a recipe can flip it once on `sources_to_platform_instance.<connector_id>` (or `destination_to_platform_instance.<destination_id>`) without affecting other entries.

metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran_rest_api.py
treff7es: Good catch — fixed in `8729960fa2`. Added a private `_extract_data(payload, *, context)` helper that validates `code == "Success"` and returns `payload.get("data")` (raising a clear `ValueError` instead of `KeyError` on a malformed-but-200 envelope). All five new direct-access sites in this file route through it now (`get_destination_details_by_id`, `list_groups`, `list_connections`, `get_connection_schemas`, `get_table_columns`, `list_users`). `get_connection_details` was left alone — it predates this PR and was already defensive.

metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran_log_db_reader.py
treff7es: You're right, that was a silent backwards-compat regression. Reverted to master's name-only match in DB mode in `8729960fa2`. REST mode still ORs `connector_id` with `connector_name` (which in REST is `listed.schema_`), but REST is new — no compat baseline to preserve there. Added a comment in `fivetran_log_rest_reader.py` pointing at the DB reader and explaining the asymmetry, so the next person doesn't try to "unify" them.

metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran.py
treff7es: Fixed in `8729960fa2`. The unknown-service fallback no longer synthesizes a URN platform from `discovered.service`; we leave `base.platform` untouched (so user overrides still win) and let `build_destination_urn` raise. The existing per-destination dedup'd warning at `_extend_lineage` already tells the user to set `destination_to_platform_instance.<id>.platform` — same pattern used for missing-database-on-glue. We still mirror `discovered.config.database` so once the user pins a platform they don't have to also pin the database. Regression test in `test_unknown_service_without_override_leaves_platform_none`.

metadata-ingestion/src/datahub/ingestion/source/fivetran/response_models.py
treff7es: Correct — `FivetranSyncHistoryResponse` and `FivetranSyncHistoryItem` were dead. REST has no sync-history endpoint (call out in the PR description). Both classes deleted in `8729960fa2`.

treff7es:
**Re: Issue A — `report_connectors_scanned()` double-counted in REST mode**

Confirmed and fixed in `8729960fa2`. The increment at `fivetran_log_rest_reader.py:263` (per-worker on successful build) and the one at `fivetran.py:817` (start of `_get_connector_workunits`) were both running for every REST-mode connector. Kept the `fivetran.py` site as the single source of truth — it's the one DB mode uses too — and removed the REST-side call. The `_report_lock` is still used by the seven other report-write sites in that file so the lock itself stays.

treff7es:
**Re: Issue B — empty-string sentinel divergence in REST `get_user_email`**

Fixed in `8729960fa2`. `fivetran_log_rest_reader.py:134` is now `if not user_id:`, matching the DB reader's check. The empty-string sentinel set by `_build_connector` (`user_id=listed.connected_by or ""`) now short-circuits, so anonymous-connected connectors no longer trigger an N-group `list_users` walk on every cache miss. Added a comment explaining the sentinel for the next reader.

metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran_log_rest_reader.py
```diff
@@ -0,0 +1,579 @@
...
+        # writes its log to the configured warehouse — explicit source /
+        # destination column names from sync events). When the DB log
+        # has NO entry for this connector_id (typical for accounts where
+        # different connectors log to different warehouses, or for newly-
+        # added connectors that haven't synced yet) we fall through to
+        # REST. An empty result is not a failure — REST may still have
+        # full lineage for that connector via the schemas+columns path.
+        if self._db_lineage_reader is not None:
+            try:
+                lineage_by_id = self._db_lineage_reader.fetch_lineage_for_connectors(
```
askumar27, line 392: In hybrid mode, `self._db_lineage_reader.fetch_lineage_for_connectors([connector_id])` is called from inside a rest_api_max_workers-dispatched worker. 
The DB method is designed to issue 2 SQL queries regardless of list size. Result: 2×N concurrent warehouse queries instead of 2 sequential ones, and concurrent connections may not inherit the `USE DATABASE` session state set in `_setup_snowflake_engine`

askumar27 (approved):
LGTM! Left a minor comment. This change will require the connector-test golden files update.

PR #17229, opened 2026-04-28
fix(ingest/teradata): enable lazy graph schema loading for cross-recipe lineage

JohnRTurner:
## Summary

- Teradata deployments commonly split ingestion across multiple recipes (one per database group). Views in recipe A frequently reference upstream tables from recipe B's scope.
- `_init_schema_resolver` was returning a `SchemaResolver` without a `graph` reference, so cross-recipe upstream lookups always failed silently. The SQL parser fell back to low-confidence (0.2) positional matching, generating lowercase field URNs. Teradata's `schemaMetadata` stores column names in uppercase (native catalog casing), so the UI could not match them — column lineage never rendered.
- Fix: pass `graph=self.ctx.graph` to the fallback `SchemaResolver`. When a table is not in the local in-scope cache, the resolver now fetches its schema from DataHub on demand — exactly as every other SQL connector (Snowflake, MySQL, PostgreSQL, etc.) already does via the base-class aggregator initialisation.

**Root cause of divergence:** `sql_common.py`'s `SQLAlchemySource.__init__` initialises `SqlParsingAggregator` without an explicit `schema_resolver`, so the aggregator creates a lazy-loading `SchemaResolver(graph=graph)` automatically. Teradata overrides this by constructing its own `schema_resolver` and passing it to the aggregator, which locked it into "pre-provided resolver" mode and inadvertently disabled lazy loading.

## Test plan

- [ ] Existing Teradata unit tests pass
- [ ] Run a Teradata ingestion recipe that covers database A (views) but not database B (upstream tables already ingested separately) — verify column lineage resolves with confidence ≥ 0.9 and correct field URN casing
- [ ] Run with `include_tables: false` or `include_views: false` — verify the bulk `provide_schema_resolver` path is still taken (no regression)


sgomezvillamor (approved):
Root cause analysis is accurate and the fix is minimal and correct. The fallback `SchemaResolver` now has a `graph` reference, restoring lazy on-demand lookups for cross-recipe column lineage. Minor nit: the warning message at line 1034 still says "empty schema resolver" — worth updating since the resolver is no longer empty after this fix.

---

sgomezvillamor:
Thanks for the clean fix! Two pre-existing issues in the same area worth addressing as a follow-up:

1. **`self.schema_resolver` is never closed** — `SchemaResolver` is `Closeable` and wraps a `FileBackedDict` (SQLite connection). Since it's passed externally to the aggregator, the aggregator won't close it on `aggregator.close()`. `Teradata.close()` should call `self.schema_resolver.close()` to avoid file handle leaks in sequential recipe runs.

2. **`close()` doesn't call `super().close()`** — unlike the `postgres` and `mssql` sources, Teradata skips parent teardown. If the base class ever adds cleanup logic, it will silently be skipped.

Neither is introduced by this PR. Both would be naturally resolved by the strategic refactor (removing `_init_schema_resolver` / `self.schema_resolver` and relying on the aggregator's own initialization paths) mentioned in the description.

---

PR #17230, opened 2026-04-29
feat(search): support Elasticsearch 8.18+ semantic search [AI-548]

alexjst:
## Summary

- Adds Elasticsearch 8.18+ support for semantic search, complementing the existing OpenSearch 2.17+ implementation
- Adds `searchKnn()`, `createSemanticIndex()`, and `indexEmbeddings()` to `SearchClientShim`, replacing direct `performLowLevelRequest()` calls in `SemanticEntitySearchService`
- Extracts engine-specific query building and index mapping into dedicated ES8 and OpenSearch builder classes
- Adds ES 8.18+ version enforcement, semantic search gate rejecting ES7, and bidirectional space type translation (`cosinesimil` ↔ `cosine`)

## Changes

**Shared DTOs** (`metadata-utils`):
- `KnnSearchRequest`, `KnnSearchResponse`, `EmbeddingBatch`, `SemanticIndexSpec` — used by both ES8 and OpenSearch shim implementations

**ES8 builders** (`metadata-io`):
- `Es8KnnQueryBuilder` — nested kNN with post-filtering via `bool.filter`
- `Es8SemanticIndexMapper` — `dense_vector` field type with HNSW `index_options`
- `Es8SemanticIndexSettingsBuilder` — no `index.knn` setting (ES8 doesn't use it)

**OpenSearch builders** (`metadata-io`):
- `OpenSearch2KnnQueryBuilder` — nested kNN with pre-filtering inside kNN block
- `OpenSearch2SemanticIndexMapper` — `knn_vector` field type with HNSW method config
- `OpenSearch2SemanticIndexSettingsBuilder` — `index.knn=true` (required by OpenSearch)

**Shim implementations**:
- `Es8SearchClientShim` — semantic methods using ES8 native client, enforces 8.18+ minimum
- `OpenSearch2SearchClientShim` — semantic methods using low-level REST client

**Refactored**:
- `SemanticEntitySearchService` — calls `shim.searchKnn()` instead of building raw JSON and using `performLowLevelRequest()`
- `V2SemanticSearchMappingsBuilder` — dispatches to ES8 or OpenSearch mapper based on engine type
- `V2SemanticSearchSettingsBuilder` — only adds `index.knn=true` for OpenSearch

**Factory + gate**:
- `SearchClientShimFactory` — rejects ES7 when semantic search is enabled
- `MappingsBuilderFactory` / `SettingsBuilderFactory` — wire `SearchClientShim` into V2 builders

**Tests**:
- Testcontainers integration tests for ES 8.18 (create-index → index-embeddings → search-knn round-trip)
- Unit tests for all builders, version enforcement, and semantic search gate
- 134+ tests across all suites

### DIFF
--- a/docker/build.gradle
+++ b/docker/build.gradle
@@ -141,6 +141,14 @@ ext {
                             DATAHUB_LOCAL_ACTIONS_ENV: "${rootProject.project(':smoke-test').projectDir}/test_resources/actions/actions.env"
                     ]
             ],
+            'quickstartDebugElasticsearch': [
+                    profile: 'debug-elasticsearch',
+                    modules: python_services_modules + backend_profile_modules + [':datahub-frontend', ':datahub-actions'],
+                    isDebug: true,
+                    additionalEnv: [
+                            DATAHUB_LOCAL_ACTIONS_ENV: "${rootProject.project(':smoke-test').projectDir}/test_resources/actions/actions.env"
+                    ]
+            ],
             'quickstartDebugAws': [
                     profile: 'debug-aws',
                     modules: python_services_modules + backend_profile_modules + [':datahub-frontend', ':datahub-actions'],
--- a/metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/client/shim/builder/opensearch2/OpenSearch2SemanticIndexSettingsBuilder.java
+++ b/metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/client/shim/builder/opensearch2/OpenSearch2SemanticIndexSettingsBuilder.java
@@ -11,6 +11,9 @@ import javax.annotation.Nonnull;
  * <p>OpenSearch requires the top-level {@code "knn": true} index setting to enable the k-NN plugin
  * for an index. This is distinct from Elasticsearch 8, which uses the {@code dense_vector} field
  * mapping and does not require this separate index-level setting.
+ *
+ * <p>Shard and replica counts are left to engine defaults so that operators can override them via
+ * cluster-level index templates.
  */
 public final class OpenSearch2SemanticIndexSettingsBuilder {
 
@@ -20,8 +23,6 @@ public final class OpenSearch2SemanticIndexSettingsBuilder {
   public static Map<String, Object> build(@Nonnull SemanticIndexSpec spec) {
     Map<String, Object> settings = new LinkedHashMap<>();
     settings.put("knn", true);
-    settings.put("number_of_shards", 1);
-    settings.put("number_of_replicas", 1);
     return settings;
   }
 }
--- a/metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/client/shim/builder/opensearch2/OpenSearch2SemanticIndexSettingsBuilder.java
+++ b/metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/client/shim/builder/opensearch2/OpenSearch2SemanticIndexSettingsBuilder.java
@@ -1,0 +1,27 @@
+package com.linkedin.metadata.search.elasticsearch.client.shim.builder.opensearch2;
+
+import com.linkedin.metadata.utils.elasticsearch.shim.SemanticIndexSpec;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import javax.annotation.Nonnull;
+
+/**
+ * Builds the OpenSearch 2.x index settings for a semantic search index.
+ *
+ * <p>OpenSearch requires the top-level {@code "knn": true} index setting to enable the k-NN plugin
+ * for an index. This is distinct from Elasticsearch 8, which uses the {@code dense_vector} field
+ * mapping and does not require this separate index-level setting.
+ */
+public final class OpenSearch2SemanticIndexSettingsBuilder {
+
+  private OpenSearch2SemanticIndexSettingsBuilder() {}
+
+  @Nonnull
+  public static Map<String, Object> build(@Nonnull SemanticIndexSpec spec) {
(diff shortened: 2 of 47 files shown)


nwadams:
@alexjst can we modify quickstart to have an option to start datahub using elasticsearch vs opensearch?

metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/client/shim/builder/opensearch2/OpenSearch2SemanticIndexSettingsBuilder.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+public final class OpenSearch2SemanticIndexSettingsBuilder {
+
+  private OpenSearch2SemanticIndexSettingsBuilder() {}
+
+  @Nonnull
+  public static Map<String, Object> build(@Nonnull SemanticIndexSpec spec) {
+    Map<String, Object> settings = new LinkedHashMap<>();
+    settings.put("knn", true);
+    settings.put("number_of_shards", 1);
```
nwadams: should these be hardcoded?

metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/client/shim/impl/OpenSearch2SearchClientShim.java
```diff
@@ -674,6 +701,123 @@ protected void closeProcessor(BulkProcessor processor) {
...
+    JsonNode responseJson = objectMapper.readTree(responseBody);
+
+    return parseSearchKnnResponse(responseJson);
+  }
+
+  /**
+   * Shared mapper for response parsing; avoids per-call allocation since ObjectMapper is
+   * thread-safe after construction.
+   */
+  private static final ObjectMapper PARSE_MAPPER = new ObjectMapper();
```
nwadams: do we need global mapper as well as class specific mapper?

metadata-io/src/main/java/com/linkedin/metadata/search/semantic/SemanticEntitySearchService.java
```diff
@@ -297,54 +318,47 @@ private SearchResult emptyResult(int from, int size) {
...
    * @param vector query embedding vector
-   * @param k number of nearest neighbors to retrieve from OpenSearch
+   * @param k number of nearest neighbors to retrieve
    * @param docLevelFilterMap optional document-level filter map to apply within the kNN filter
    * @param fieldsToFetch set of fields to fetch in the _source (supports fetchExtraFields)
-   * @return list of {@link SearchEntity} constructed from OpenSearch hits
+   * @return list of {@link SearchEntity} constructed from kNN hits
    */
   private List<SearchEntity> executeKnn(
-      @Nonnull ObjectMapper objectMapper,
```
nwadams, line 311: why remove this objectmapper for global?

metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/client/shim/builder/opensearch2/OpenSearch2SemanticIndexSettingsBuilder.java
alexjst: Removed them from both the ES8 and OS2 settings builders. Engine defaults are 1 and 1 anyway, so behavior is the same, but operators can now override via cluster level index templates.

metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/client/shim/impl/OpenSearch2SearchClientShim.java
alexjst: Dropped PARSE_MAPPER. `parseSearchKnnResponse` now takes the mapper as a parameter so the class uses just the instance field.

metadata-io/src/main/java/com/linkedin/metadata/search/semantic/SemanticEntitySearchService.java
alexjst: Was a leftover from threading the shim through. Reverted to use opContext.getObjectMapper().

alexjst:
> @alexjst can we modify quickstart to have an option to start datahub using elasticsearch vs opensearch?

Added a `quickstartDebugElasticsearch` Gradle task that uses the `debug-elasticsearch` profile, and bumped the default ES image from 7.10.1 to 8.18.0. This mirrors other tasks like `quickstartDebugAws`, `quickstartDebugMin`, etc.

Run it with `./gradlew :docker:quickstartDebugElasticsearch`.

docker/build.gradle
```diff
@@ -141,6 +141,14 @@ ext {
                             DATAHUB_LOCAL_ACTIONS_ENV: "${rootProject.project(':smoke-test').projectDir}/test_resources/actions/actions.env"
                     ]
             ],
+            'quickstartDebugElasticsearch': [
```
nwadams, line 144: how does this make it run elasticsearch?

nwadams (approved):

docker/build.gradle
alexjst: Through the docker compose `debug-elasticsearch` profile. The `elasticsearch` container is gated on that profile (and `opensearch` isn't), so only ES starts. The matching `datahub-gms-debug-elasticsearch` service sets `ELASTICSEARCH_IMPLEMENTATION: elasticsearch` so GMS talks to ES. The profile wiring already existed; this PR just adds the Gradle entry point. Verified locally: ES 8.18.0 + healthy GMS, no opensearch container.

PR #17233, opened 2026-04-29
feat(airflow-plugin): add per-DAG lineage filter for DataJob lineage

alfiyas-datahub:
## Summary

Adds a new `datajob_lineage_dag_filter_str` config option (an `AllowDenyPattern`) to the Airflow plugin that lets users selectively disable `DataJobInputOutput` lineage for specific DAGs while keeping it enabled for others.

The existing `enable_datajob_lineage` is a global ON/OFF switch — too coarse for users who run dbt models through Airflow (e.g. via Cosmos), where the Airflow-task-level lineage adds noise on top of the cleaner dbt lineage. With the v2 plugin's OpenLineage extraction, every Cosmos task now emits its own dataset lineage, which combined with the UI's `ignoreAsHops` traversal causes the lineage graph to explode (in one customer's case, from 19 to 270 entities one hop upstream).


treff7es:
I wonder if we need this, as we already have a way to disable Dags.

From an implementation perspective, I wonder if something like this would be easier to use by the users:
```
[datahub]
  # Comma-separated list of dag_ids (or dag_id glob patterns) for which DataJob
  # lineage should be suppressed. Other DAGs keep lineage as configured by
  # enable_datajob_lineage. Glob: '*' wildcard, '?' single char.
  datajob_lineage_disabled_dags = dbt_orders, dbt_users, cosmos_*
  ```

PR #3298, opened 2021-09-24
feat(pk/fk): show primary keys & foreign keys in the schema

gabe-lyons:
This PR adds visuals for primary keys and foreign keys in the schema view.

At the schema level:
image

When an foreign key is clicked, a module pops open in-line displaying the foreign key parameters:
image

## Checklist
- [ ] The PR conforms to DataHub's [Contributing Guideline](https://github.com/linkedin/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [Commit Message Format](https://github.com/linkedin/datahub/blob/master/docs/CONTRIBUTING.md#commit-message-format))
- [ ] Links to related issues (if applicable)
- [ ] Tests for the changes have been added/updated (if applicable)
- [ ] Docs related to the changes have been added/updated (if applicable)

PR #17236, opened 2026-04-29
feat(unity-catalog): extract primary key, foreign key, and partition key constraints

Meehau:
# feat(unity-catalog): extract primary key, foreign key, and partition key constraints

## Summary

Databricks Unity Catalog supports informational (non-enforced) PK/FK constraints and column
partition indexes. The DataHub metadata model (`SchemaMetadata`) and the UI (merged via PR #3298)
already support these fields, but the Unity Catalog ingestion connector never populated them.

This PR closes that gap by patching the connector to:
1. Fetch `table_constraints` (PK/FK) via a per-table `tables.get()` call — the `tables.list()` API
   does not return constraints, which is documented Databricks API behaviour.
2. Map `partition_index` from `ColumnInfo` (already present in `tables.list()` responses) to
   `SchemaField.isPartitioningKey` — no extra API calls needed.
3. Populate `SchemaMetadata.primaryKeys`, `SchemaMetadata.foreignKeys`, and
   `SchemaField.isPartOfKey` accordingly.

Both new behaviours are **opt-in** (disabled by default) to avoid surprising API load increases
on existing pipelines.

## What changed

| File | Change |
|------|--------|
| `proxy_types.py` | `Column` gets a new `partition_index: Optional[int] = None` field |
| `proxy.py` | `_create_column()` propagates `partition_index`; new `get_table_constraints()` method calls `tables.get(full_name=…)` once per non-Hive table and returns the constraint list; errors are surfaced in the pipeline report |
| `config.py` | New `include_table_constraints: bool = False` flag (opt-in) and `include_partition_keys: bool = False` flag (opt-in, no extra API cost) |
| `source.py` | `_create_schema_metadata_aspect()` calls `get_table_constraints()`, builds `ForeignKeyConstraintClass` objects, and passes `primary_key_columns` to `_create_schema_field()`; `_create_schema_field()` now sets `isPartOfKey` and `isPartitioningKey` |

## Usage

Enable in your recipe:

```yaml
source:
  type: unity-catalog
  config:
    include_table_constraints: true   # adds one tables.get() call per table for PK/FK
    include_partition_keys: true      # no extra API calls — data already in tables.list()
```

## Testing

```
pytest tests/integration/unity/test_unity_catalog_ingest.py -v
# 4 passed — test_ingestion, test_ml_model_with_signature_and_run_details,
#             test_ingestion_with_constraints, test_constraints_disabled_does_not_call_tables_get
```

New tests:
- `test_ingestion_with_constraints` — mocks `tables.list()` returning a column with
  `partition_index: 0`, mocks `tables.get()` returning a `PrimaryKeyConstraint` on `columnA` and a
  `ForeignKeyConstraint` from `columnB` → `quickstart_catalog.quickstart_schema.quickstart_table_ref.columnA`;
  compares against a new golden file `unity_catalog_constraints_mces_golden.json`
- `test_constraints_disabled_does_not_call_tables_get` — asserts that `get_table_constraints()` is
  never called when `include_table_constraints: false`

## Screenshots / Verification

After running ingestion against a live Databricks workspace with tables that have PK/FK constraints
defined, the DataHub schema view shows:
- **PK icon** on primary key columns
- **FK badge** on foreign key columns (clickable to show the referenced dataset and column)
- **Partition key badge** on partition columns

**Unrelated fixes - but still kept in the PR**
The `QueryFilterWithStatementTypes.from_dict` coercion and ML model signature JSON-parsing fixes
were discovered while running the full test suite with the constraints feature enabled. Both are
required for the existing tests to pass in combination with the new code paths. They don't affect
existing production behavior - they only harden edge cases in test mocks.


shirshanka (commented):
Thanks for this PR — sorry for the slow turnaround on the review! The overall design is solid: opt-in flags, graceful error handling in `get_table_constraints()`, the Hive Metastore skip, and the `assert_not_called()` guard on the disabled path are all exactly right. A few things to address before merging:

## Must-fix: `isPartOfKey=False` written for all users even with flags disabled

In `source.py`, `_create_schema_field` now always sets:

```python
isPartOfKey=column.name in primary_key_columns,
```

When `include_table_constraints=False` (the default), `primary_key_columns` is an empty `frozenset()`, so this evaluates to `False` for every column. Before this PR the field was `None` (not set). `None` means "unknown" and `False` means "confirmed not a key" — DataHub treats these differently in the schema view, and downstream consumers may too. This will silently change the emitted metadata for all existing Unity Catalog users on their next run, without them opting in to anything.

Compare to how `isPartitioningKey` is handled — correctly:

```python
isPartitioningKey=(
    True if include_partition_keys and column.partition_index is not None else None
)
```

The fix is the same pattern. One clean option: change `primary_key_columns` from `AbstractSet[str]` to `Optional[AbstractSet[str]]` where `None` = flag disabled vs. empty set = enabled but no PKs found:

```python
isPartOfKey=(
    column.name in primary_key_columns
    if primary_key_columns is not None
    else None
),
```

## Suggestion: split out the two `proxy.py` refactors

The `QueryFilterWithStatementTypes.from_dict` coercion fix and the ML model signature JSON-parsing improvement are both reasonable fixes, but they're unrelated to PK/FK extraction. My guess is you hit these while running the full test suite with your new test — which exercises the whole pipeline including ML models. If either causes a production regression it'll be hard to attribute.

Would you be willing to either (a) move them into a separate commit with a short description of what was broken, or (b) keep them here but add a line in the PR description explaining what triggered them? Option (b) is fine with me — just want the context preserved.

## Minor: incorrect comment

In `_create_schema_metadata_aspect`:

```python
# Partition index values (used for `isPartitioningKey`) are fetched separately
# when `include_partition_keys` is enabled via an additional `tables.get()` call.
```

Partition keys come from `partition_index` in the `tables.list()` response — no extra API call needed. This is actually one of the selling points of the feature! Worth fixing the comment to reflect that.

## Nice-to-have: a couple of test gaps

Not blocking, but worth noting:
- No test for the `get_table_constraints()` exception path (rate limit, permission denied). That's the most likely real-world failure mode.
- No test for `include_partition_keys=True` with `include_table_constraints=False` — verifying that `isPartitioningKey` is set while `isPartOfKey` stays `None`. This is probably the most common usage pattern.

The `isPartOfKey` issue is the only hard blocker. Everything else is addressable here or in a follow-up. Happy to re-review quickly once that's fixed.

sgomezvillamor (commented):
Nothing to add on top of @shirshanka's review — the `isPartOfKey=False` vs `None` issue is the right blocker to fix first.

One additional ask before merging: the two new config flags are missing from the **user-facing connector docs** (no `.md` files were changed in this PR).

**`unity-catalog_post.md`** — the existing troubleshooting note says:
> "Unity Catalog REST API requires one call per table (table lineage) and one call per column (column lineage). To improve performance, disable column lineage with `include_column_lineage: false`."

Please extend it to mention that `include_table_constraints: true` also adds one `tables.get()` call per table.

**`unity-catalog_pre.md`** — the permissions section has per-flag entries for `include_usage_statistics`, profiling, etc. Please add a bullet for `include_table_constraints` confirming it requires no permissions beyond the `SELECT` privilege already listed as a base requirement (so users auditing a service principal's grants know they don't need to add anything).

`include_partition_keys` needs no doc entry — it uses data already returned by `tables.list()` with no extra calls or permissions.

Once Shirshanka's blocker and these doc additions are addressed, this looks ready to merge.

---

Meehau:
Thanks @shirshanka and @sgomezvillamor for the thorough review! All items have been addressed:

**Blocker fix — `isPartOfKey=False` regression:**

`primary_key_columns` is now `Optional[Set[str]]` initialized to `None`. It is only set to `set()`
inside the `if self.config.include_table_constraints:` block. `isPartOfKey` now mirrors the
`isPartitioningKey` pattern — only `True` when the column is explicitly known to be part of a
primary key, `None` otherwise:

```python
isPartOfKey=(
    True
    if primary_key_columns is not None and column.name in primary_key_columns
    else None
),
```

`isPartOfKey` is a non-nullable boolean in the Avro schema, so `None` and `False` produce
identical serialized output — pre-PR behavior is fully preserved for users who do not enable
`include_table_constraints`. A clarifying comment was added to the code noting this.

**Documentation:**

- `unity-catalog_post.md`: added performance note about `include_table_constraints`
- `unity-catalog_pre.md`: added permissions note (no extra privileges needed)

**Comment fix:**

Corrected the misleading comment about partition keys requiring `tables.get()`.

**Test gaps:**

- Added `test_constraints_exception_path` — makes `tables.get()` raise a 403, verifies the
  pipeline completes successfully (graceful degradation)

- Added `test_partition_keys_without_constraints` — verifies `isPartitioningKey=True` is set
  while `isPartOfKey` stays at the schema default when only `include_partition_keys=True`

**Unrelated fixes - kept in PR with explanation:**

The `QueryFilterWithStatementTypes.from_dict` coercion and ML model signature JSON-parsing fixes
were discovered while running the full test suite with the constraints feature enabled. Both are
required for the existing tests to pass in combination with the new code paths. They don't affect
existing production behavior — they only harden edge cases in test mocks.

**Additional fix (found during internal review):**

FK parent URN construction now uses `self.gen_dataset_urn(parent_ref)` instead of manual
`make_dataset_urn_with_platform_instance(...)` - this ensures correct URNs when
`include_metastore=True` (deprecated but still supported).

sgomezvillamor (approved):
All concerns addressed — approving. Thanks for the contribution and for being thorough in the follow-up!

---

PR #17237, opened 2026-04-29
fix(ingestV2): secret deletion always removes last secret instead of …

Itsdarkhere:
…selected

ConfirmationModal was rendered inside the per-row table column renderer, sharing a single boolean `showConfirmDelete` state. When any delete button was clicked all rows' modals opened simultaneously, and confirming always triggered the last row's handleConfirm closure, deleting the wrong secret. Fix tracks the pending-deletion URN in `deletingSecretUrn: string | null` and renders a single ConfirmationModal at the component root.


sgomezvillamor:
indeed, deleting wrong secrets is concerning 😅 
@chriscollins3456 would you be available to review this one?

sakethvarma397 (commented):
This is already fixed in the cloud version, and it seems like it's not added here. To avoid conflicts, it's better that we back port those changes instead of merging this commit.

sakethvarma397:
@Itsdarkhere I'll create a PR back-porting the changes. Will share it here once the PR is up.

PR #17238, opened 2026-04-29
fix(actions/kafka): re-raise from with_retry after exhausting attempts

max-datahub:
## Summary

`with_retry` was returning `None` after exhausting all attempts, so callers couldn't distinguish "operation returned None" from "operation failed permanently". The single caller in `KafkaEventSource.ack()` uses `with_retry` to commit offsets via `_commit_offsets`. Silent commit failures meant `Pipeline._ack_event`'s `try/except` never triggered, so `failed_ack_count` was never incremented and a broken broker would not show up in the pipeline's stats counters.

Re-raise the last exception after exhausting attempts. `Pipeline._ack_event` already wraps `source.ack` in `try/except` and increments `failed_ack_count`, so the caller side handles this correctly.

## Test plan

- [x] Failing test added first; verified RED then GREEN
- [x] All 11 tests in `tests/unit/plugin/source/kafka/` pass
- [x] `ruff check` and `mypy` clean

### DIFF
--- a/datahub-actions/src/datahub_actions/plugin/source/kafka/utils.py
+++ b/datahub-actions/src/datahub_actions/plugin/source/kafka/utils.py
@@ -19,8 +19,9 @@ def with_retry(
 
             curr_attempt = curr_attempt + 1
             if curr_attempt >= max_attempts:
+                # Relay failure to caller.
                 logger.warning("kafka event source: exhausted all attempts.")
-                return
+                raise
 
             backoff = backoff * 2
             if backoff > max_backoff:
--- a/datahub-actions/tests/unit/plugin/source/kafka/test_kafka_utils.py
+++ b/datahub-actions/tests/unit/plugin/source/kafka/test_kafka_utils.py
@@ -1,0 +1,53 @@
+# Copyright 2021 Acryl Data, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+
+from unittest.mock import patch
+
+import pytest
+
+from datahub_actions.plugin.source.kafka.utils import with_retry
+
+
+def test_with_retry_reraises_after_exhaustion() -> None:
+    # with_retry was returning None after exhausting attempts, so callers had
+    # no way to distinguish "operation succeeded with None result" from
+    # "operation failed permanently". The single caller in kafka_event_source
+    # uses with_retry to commit offsets — silent commit failures meant the
+    # pipeline incremented success_count instead of failed_ack_count, hiding
+    # broken brokers from operator monitoring.
+    def always_fails() -> None:
+        raise ValueError("permanent failure")
+
+    with patch("datahub_actions.plugin.source.kafka.utils.time.sleep"):
+        with pytest.raises(ValueError, match="permanent failure"):
+            with_retry(max_attempts=3, max_backoff=1.0, func=always_fails)
+
+
+def test_with_retry_returns_value_on_success() -> None:
+    # Confirm the success path still works: returns the function's return value.
+    def succeeds() -> str:
+        return "ok"
+
+    result = with_retry(max_attempts=3, max_backoff=1.0, func=succeeds)
+    assert result == "ok"
+
+
+def test_with_retry_succeeds_after_transient_failure() -> None:
+    # Confirm intermediate failures don't break the success path: function
+    # raises once, then returns.
+    attempts = {"n": 0}
+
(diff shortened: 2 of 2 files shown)


datahub-actions/src/datahub_actions/plugin/source/kafka/kafka_event_source.py
```diff
@@ -279,6 +279,20 @@ def close(self) -> None:
         # Then close consumer
         if self.consumer:
             self.running = False
+            # In async-commit mode (the default), acked offsets are only stored
+            # locally via store_offsets and committed asynchronously every
+            # auto.commit.interval.ms. consumer.close() does not reliably flush
+            # those staged offsets in librdkafka, so without an explicit final
+            # commit, up to auto.commit.interval.ms worth of acked events are
+            # re-processed on next worker boot. For non-idempotent actions this
```
asikowitz: We shouldn't have non idempotent actions imo, we are not set up for it

datahub-actions/src/datahub_actions/plugin/source/kafka/kafka_event_source.py
```diff
@@ -279,6 +279,20 @@ def close(self) -> None:
         # Then close consumer
         if self.consumer:
             self.running = False
+            # In async-commit mode (the default), acked offsets are only stored
```
asikowitz: I don't think this is necessary actually. This is the comment I see for `consumer.close()`:

```
    def close(self, *args, **kwargs): # real signature unknown
        """
        Close down and terminate the Kafka Consumer.
        
          Actions performed:
        
          - Stops consuming.
          - Commits offsets, unless the consumer property 'enable.auto.commit' is set to False.
          - Leaves the consumer group.
        
          .. note: Registered callbacks may be called from this method, see :py:func::`poll()` for more info.
        
          :rtype: None
        """
        pass
```

and note we do set `enable.auto.commit`:
```
        if self.source_config.async_commit_enabled:
            # See for details: https://github.com/confluentinc/librdkafka/blob/master/INTRODUCTION.md#auto-offset-commit
            async_commit_config["enable.auto.offset.store"] = False
            async_commit_config["enable.auto.commit"] = True
            async_commit_config["auto.commit.interval.ms"] = (
                self.source_config.async_commit_interval
            )
```

datahub-actions/src/datahub_actions/plugin/source/kafka/utils.py
```diff
@@ -19,8 +19,13 @@ def with_retry(
...
             curr_attempt = curr_attempt + 1
             if curr_attempt >= max_attempts:
+                # Re-raise the last exception so callers can distinguish
+                # "operation returned None" from "operation failed permanently".
+                # The single caller in kafka_event_source uses with_retry to
+                # commit offsets; silent failures hid broken brokers from the
+                # pipeline's failed_ack_count counter.
                 logger.warning("kafka event source: exhausted all attempts.")
-                return
+                raise
```
asikowitz: I'm fine with this change but don't think we need the comment here, or the comment can just be "Relay failure to caller". Note in practice this will just affect stats counting -- the exception is just absorbed by the caller:
```
    def _ack_event(self, enveloped_event: EventEnvelope, processed: bool) -> None:
        try:
            self.source.ack(enveloped_event, processed)
            self._stats.increment_success_count()
        except Exception:
            self._stats.increment_failed_ack_count()
            logger.exception(
                f"Caught exception while attempting to ack successfully processed event. event type: {enveloped_event.event_type}, pipeline name: {self.name}",
            )
            logger.debug(f"Failed to ack event: {enveloped_event}")
```

datahub-actions/src/datahub_actions/plugin/source/kafka/kafka_event_source.py
max-datahub: Confirmed and you are correct! Dropping this

datahub-actions/src/datahub_actions/plugin/source/kafka/kafka_event_source.py
max-datahub: I hadn't considered the implications on idempotency. Good and valid point, I agree

datahub-actions/src/datahub_actions/plugin/source/kafka/utils.py
max-datahub: Updated to be less verbose

sgomezvillamor (approved):

PR #17245, opened 2026-04-29
fix(ingest/preset): defensive position_json parsing and dashboard-to-chart lineage recovery

dinesh-verma-datahub:
## Summary

The Preset/Superset connector had a regression where dashboards using `position_json` for chart-to-dashboard lineage could silently produce zero chart edges — either because `position_json` was absent/unparseable, or because `meta.chartId` values were `False`/`""` which Python's `isinstance(False, int)` would have accepted. This PR fixes the root cause, adds a `/charts` fallback for empty layouts, hardens all API fetch paths, and replaces ad-hoc counters with a structured `SupersetSourceReport`.

### 1. `position_json` parsing and chartId extraction

- **`_parse_position_json()`** (new method): parses `position_json` defensively — accepts an already-deserialized `dict` (some clients pre-parse), catches `(json.JSONDecodeError, TypeError)` for non-string inputs, validates the result is a `dict` (not a JSON array/scalar), emits `Invalid position_json` warnings and increments `num_dashboards_invalid_position_json` on failure, returns `{}` so the dashboard still ingests.
- **`_extract_chart_id_with_status()`** (new module-level function): returns `(chart_id, was_uncoercible)` for a `CHART-*` entry. Explicitly rejects `bool` before `int` (Python footgun: `isinstance(False, int) is True`), rejects empty and whitespace-only strings (would mint `urn:li:chart:(preset,   )`), accepts `int` and non-empty `str`. `was_uncoercible=True` only when `meta.chartId` was present but invalid — callers increment `num_chart_ids_uncoercible`.
- **`_extract_chart_id()`** (new module-level function): thin wrapper discarding the status flag, used where only the ID is needed.
- **`construct_dashboard_from_api_data()`**: now accepts `position_data` as a pre-parsed parameter; uses `_extract_chart_id_with_status` to build chart URNs.
- **`_process_dashboard()`**: calls `_parse_position_json` once, passes the result into `construct_dashboard_from_api_data`.

### 2. `/api/v1/dashboard/{id}/charts` fallback

When `position_json` yields zero valid chart IDs (`position_yielded_charts=False`), `_process_dashboard` now calls `get_dashboard_charts(dashboard_id)` and injects synthetic `CHART-fallback-{id}` entries into `position_data`. This covers Preset Cloud dashboards where `position_json` is legitimately empty but the `/charts` endpoint still returns the chart list. The fallback is **only triggered when `position_json` yields no charts** — always-on has no impact on existing customers that have a populated `position_json`.

- **`get_dashboard_charts()`** (new method): fetches `GET /api/v1/dashboard/{id}/charts`, validates the response, rejects bool/non-int IDs (same rules as `_extract_chart_id_with_status`), aggregates malformed entries into `num_dashboards_charts_malformed_entries`, caches results in `_dashboard_charts_cache`.
- **`_warn_dangling_fallback_charts()`** (new method): for each fallback-recovered chart ID, warns when the chart was never ingested as a chart entity (lineage edge will dangle), increments `num_dashboards_charts_unknown`.
- **`num_dashboards_recovered_via_charts_endpoint`**: counter incremented when the fallback fires and returns at least one chart ID.

### 3. `_fetch_api_response()` shared HTTP helper

Replaces three copies of identical fetch-and-error-handle logic (`get_dashboard_info`, `get_dashboard_charts`, `get_dataset_info`).

- Handles `requests.RequestException` (network), non-200 HTTP, and `ValueError` (JSON decode) separately, emitting distinct warning titles: `"{endpoint} network error"`, `"{endpoint} failed"`, `"{endpoint} decode error"`.
- Uses `**{**base_context, "key": val}` for warning context merging — safe against key collisions that `**base_context, key=val` syntax would raise on.
- `failure_counter_attr` is validated with `hasattr` up front; a typo raises `AttributeError` immediately rather than being swallowed silently by the outer `except`.
- `_safe_body_preview()` (new static method): returns a redacted ≤200-char body preview — sniffs for HTML content (by `Content-Type` header and content inspection), redacts emails via `_EMAIL_REDACT_RE`, catches `UnicodeDecodeError` from binary bodies, avoids leaking cookies or session tokens into warning context.
- `_warning_context()` (new static method): serializes fields to JSON (`json.dumps(..., default=str)`) so warning context is parseable by log shippers.

### 4. Explicit dict caches replacing `@lru_cache`

`@lru_cache(maxsize=None)` is replaced with three explicit instance dicts: `_dashboard_info_cache`, `_dashboard_charts_cache`, `_dataset_info_cache`. Failed responses are **not cached** (no cache poisoning); only structurally valid responses with a dict `result` are stored.

### 5. Structured `SupersetSourceReport`

Added 26 new typed counters to `SupersetSourceReport` (grouped by entity/endpoint):

| Counter | Meaning |
|---|---|
| `num_dashboards_detail_api_failures` | Dashboard detail API failures (network, non-200, decode) |
| `num_dashboards_charts_api_failures` | Dashboard charts API failures |
| `num_datasets_detail_api_failures` | Dataset detail API failures |
| `num_dashboards_invalid_position_json` | Unparseable or non-dict `position_json` |
| `num_dashboards_missing_position_json` | Detail returned `null`/`""` for `position_json` |
| `num_dashboards_missing_id` | Dashboard entries with no `id` |
| `num_dashboards_with_no_charts` | Dashboards with no `CHART-*` keys after parsing |
| `num_dashboards_recovered_via_charts_endpoint` | Dashboards rescued by `/charts` fallback |
| `num_dashboards_charts_unknown` | Fallback charts not found in ingested chart set |
| `num_dashboards_charts_malformed_entries` | Malformed entries in `/charts` response |
| `num_dashboards_dropped_unexpected_error` | Dashboards dropped by the outer `except` |
| `num_datasets_missing_id` | Dataset entries with no `id` |
| `num_datasets_columns_payload_malformed` | Dataset `columns` was not a list |
| `num_datasets_metrics_payload_malformed` | Dataset `metrics` was not a list |
| `num_datasets_columns_dropped_malformed` | Columns dropped for missing name/type or non-dict shape |
| `num_datasets_metrics_dropped_malformed` | Metrics dropped for missing name/type or non-dict shape |
| `num_chart_ids_uncoercible` | `CHART-*` entries with invalid `meta.chartId` |
| `num_charts_missing_datasource_id_for_cll` | Charts with no `datasource_id` at CLL time |
| `num_charts_missing_dataset_detail_for_cll` | Charts where dataset detail was unavailable for CLL |
| `num_charts_without_datasource_for_urn` | Charts with no `datasource_id` at URN-build time |
| `num_charts_dropped_datasource_urn_failed` | Charts where URN construction raised `ValueError` |
| `num_charts_invalid_params_json` | Charts with unparseable `params` JSON |
| `num_charts_invalid_params_metrics_shape` | Charts where `params.metrics` was not a list |
| `num_charts_database_pattern_indeterminate` | Charts where `database_pattern` could not be evaluated |
| `num_metrics_sql_parse_errors` | Metric SQL expressions that sqlglot could not parse |

### 6. Chart `params` hardening

In `construct_chart_from_chart_data`:
- `raw_params = chart_data.get("params") or "{}"` — `None` and `""` both collapse to `"{}"` before `json.loads`.
- `parsed_params` guarded: non-dict JSON (array, scalar) treated as `{}` rather than crashing on `.get(...)`.
- `raw_metrics`: non-list shape increments `num_charts_invalid_params_metrics_shape` and skips metrics.
- `raw_filters = params.get("adhoc_filters") or []` + isinstance-list guard — non-list `adhoc_filters` silently dropped rather than crashing.
- Metrics and filters list comprehensions filter out empty strings: `[m for m in ... if m]`.

### 7. `get_metric_name` / `get_filter_name` guards

- `get_metric_name`: added `if not isinstance(metric, dict): return ""` before `.get("label")`.
- `get_filter_name`: added `if not isinstance(filter_obj, dict): return ""` guard; added `if clause is None or column is None or operator is None: return ""` to prevent `"None None = None"` strings entering chart custom properties.

### 8. `_fetch_dataset_columns` hardening

- `datasource_id is None or datasource_id == 0` replaces bare `not datasource_id` — 0 is not a valid Superset auto-increment ID; this avoids the truthiness footgun.
- Added `chart_id` keyword parameter for all warning/info context.
- Empty `dataset_info` check (when `get_dataset_info` returned `{}`) increments `num_charts_missing_dataset_detail_for_cll` and emits a chart-level locator info entry.
- `columns` and `metrics` payloads validated as `list` before iteration.
- Per-item non-dict guards for column and metric entries.

### 9. `_extract_and_map_tags` defensive typing

- Signature changed from `raw_tags: List[Dict[str, Any]]` to `raw_tags: Any`.
- Added `if not isinstance(raw_tags, list): return None` — Superset API observed to return `tags: null`.
- Added `isinstance(tag, dict)` guard in the comprehension for non-dict items in the list.

### 10. `_extract_columns_from_sql` narrowed exception

- Changed `except Exception` to `except sqlglot.errors.SqlglotError` so unrelated exceptions surface rather than being silently swallowed as a parse error.
- Emits structured warning with `num_metrics_sql_parse_errors` and `sql_preview` context.

### 11. `_process_dashboard` and `_process_chart` pre-initialisation

- `dashboard_id: Optional[int] = None` and `chart_id: Optional[int] = None` initialized before the `try` block so the outer `except` can include them in warning context even if early extraction raises.
- `_process_dashboard` gains explicit missing-ID early return with `num_dashboards_missing_id`.
- `construct_dataset_snapshot_from_api_data`: `dataset_response["result"]` → `dataset_response.get("result")` with an explicit `isinstance(result, dict)` check and a typed `ValueError` raise.

### 12. Warning context format

All `context=str(dict(...))` calls replaced with `context=self._warning_context(...)` which emits JSON — parseable by downstream log shippers and validation tooling.


metadata-ingestion/src/datahub/ingestion/source/superset.py
```diff
@@ -151,9 +151,141 @@
...
+        normalised = int(chart_id)
+    except (TypeError, ValueError):
+        return None
+    return processed_charts.get(normalised)
+
+
 @dataclass
 class SupersetSourceReport(StaleEntityRemovalSourceReport):
     filtered: LossyList[str] = field(default_factory=LossyList)
+    num_dashboards_with_no_charts: int = 0
```
kyungsoo-datahub, line 219: nit: Do we need this many new report counters? I like the added visibility, but several counters split the same endpoint failure into very fine-grained subtypes, e.g

```
num_dashboards_detail_api_failures
num_dashboards_detail_api_network_errors
num_dashboards_detail_api_decode_errors
```
They can be merged into one API failure counter.

 I think we can keep the report summary smaller: one counter per endpoint/failure category, with the specific subtype in warning context. That would reduce report noise, test burden, and the chance of counters drifting out of sync.

metadata-ingestion/src/datahub/ingestion/source/superset.py
```diff
@@ -429,27 +565,288 @@ def build_owner_urn(self, data: Dict[str, Any]) -> List[str]:
...
+                title="Dashboard charts API malformed result",
+            )
+            return []
+        chart_ids: List[int] = []
+        malformed = 0
+        for chart in results:
+            if not isinstance(chart, dict) or chart.get("id") is None:
+                malformed += 1
+                continue
+            chart_ids.append(chart["id"])
```
kyungsoo-datahub: This fallback path should validate chart IDs with the same rules as _extract_chart_id_with_status. Right now any non-None id from /charts is accepted, so values like false, "", {}, or [] can be stringified below and become synthetic chart IDs / URNs. Could we reject bool, empty strings, and uncoercible shapes before appending to chart_ids?

metadata-ingestion/src/datahub/ingestion/source/superset.py
```diff
@@ -429,27 +565,288 @@ def build_owner_urn(self, data: Dict[str, Any]) -> List[str]:
...
+        return chart_ids
+
+    def _parse_position_json(self, dashboard_data: Dict[str, Any]) -> Dict[str, Any]:
+        """Parse ``position_json`` defensively. Returns ``{}`` on any
+        non-JSON-object input — the dashboard still ingests, the failure
+        is visible in the report."""
+        raw_position_data = dashboard_data.get("position_json") or "{}"
+        try:
+            parsed = json.loads(raw_position_data)
+        except json.JSONDecodeError as e:
```
kyungsoo-datahub: json.loads can raise TypeError if position_json is already a non-string value, for example a dict/list from an API client or changed response shape. Since this helper is meant to defensively parse external payloads, could we either accept dicts directly or catch TypeError here too? Otherwise a truthy non-string position_json still escapes and drops the dashboard.

metadata-ingestion/src/datahub/ingestion/source/superset.py
```diff
@@ -909,22 +1528,83 @@ def construct_chart_from_chart_data(
...
+            self.report.info(
+                message="Chart params.metrics had unexpected shape; skipping derived metrics.",
+                context=self._warning_context(
+                    chart_id=chart_data.get("id"),
+                    metrics_type=type(raw_metrics).__name__,
+                ),
+                title="Chart params.metrics unexpected shape",
+            )
+            metrics_iter = []
+        metrics = [get_metric_name(metric) for metric in metrics_iter]
```
kyungsoo-datahub: Nit: we filtered out invalid params JSON. But valid JSON with malformed inner values can still crash later.

For example, {"metrics": [123]} or {"adhoc_filters": ["bad"]} parses successfully, but get_metric_name / get_filter_name expect string/dict-like values. Could skip unsupported metric/filter entries instead of dropping the chart.

metadata-ingestion/src/datahub/ingestion/source/superset.py
```diff
@@ -429,27 +565,288 @@ def build_owner_urn(self, data: Dict[str, Any]) -> List[str]:
...
+        body_smells_html = (
+            body.lstrip().startswith("<")
+            or "<html" in body.lower()
+            or "Set-Cookie" in body
         )
+        if "json" not in content_type or body_smells_html:
+            return f"<non-json body: content-type={content_type or 'unknown'}, len={len(body)}>"
+        return SupersetSource._EMAIL_REDACT_RE.sub("<email>", body[:200])
+
+    def get_dashboard_info(self, dashboard_id: int) -> Dict[str, Any]:
```
kyungsoo-datahub: These three functions have same code structure:
- get_dashboard_info
- get_dashboard_charts
- get_dataset_info

```
cache lookup
session.get
network error handling
non-200 handling
response.json()
result shape validation
cache success only
return empty fallback on failure
```

Could we factor the common “cache only successful valid responses” logic into a small helper that takes URL, cache, expected result type, and warning labels? That would reduce duplication and make it harder for future changes to fix one fetch path but miss the others.

metadata-ingestion/src/datahub/ingestion/source/superset.py
```diff
@@ -586,71 +997,183 @@ def construct_dashboard_from_api_data(
 
         return dashboard_snapshot
 
+    def _warn_database_filtered_charts(
```
kyungsoo-datahub: Nit: _warn_database_filtered_charts is a little ambiguous. The method doesn’t filter charts; it reports dashboard references to charts that were filtered because of database_pattern. Could we rename it to _report_dashboard_filtered_chart_references? I think that better matches the behavior and makes the call site easier to read.

kyungsoo-datahub (commented):
The PR description says the /charts fallback is opt-in via aggressive_chart_recovery, but the implementation seems to call it unconditionally whenever position_json yields no usable chart IDs. Is the fallback intended to be always enabled? If yes, can we update the PR description/docs to match? If not, should this be gated behind the config?

dinesh-verma-datahub:
@kyungsoo-datahub thanks for the thorough feedback, i have addressed all the concerns.

kyungsoo-datahub (approved):

PR #17250, opened 2026-04-30
docs(dataplex): update Dataplex prerequisites and IAM role requirements

sgomezvillamor:
## Description

Updates the Dataplex source documentation to clarify API and IAM role requirements for term-asset associations functionality.

### Changes

- Added **Cloud Resource Manager API** (`cloudresourcemanager.googleapis.com`) to the list of required APIs when using `include_glossary_term_associations: true`
- Updated the IAM role requirement for term-asset associations from `roles/resourcemanager.projectViewer` to `roles/resourcemanager.folderViewer`, which provides the necessary `resourcemanager.projects.get` permission for resolving GCP project numbers
- Reformatted the IAM roles table for improved readability

### Motivation

These documentation updates ensure users have accurate information about the APIs and permissions required to successfully configure and use the Dataplex source connector, particularly for the glossary term-asset associations feature.

### Test Plan

N/A - Documentation only change

https://claude.ai/code/session_018831CNLKuLDU5oS4mE4UGL

PR #17251, opened 2026-04-30
feat(ingest/dataplex): adds project_id_pattern and project_labels.

davidwalker-tp:
Adds `project_id_pattern` and `project_lables` to the Dataplex ingestion recipe - allowing for GCP projects to be dynamically ingested, rather than having to provide a set list of project names.

This functionality already exists in the BigQuery and VertexAI recipes and that approach was used as inspiration for Claude.


sgomezvillamor (changes requested):
Thanks for the contribution! The feature direction is correct and the shared `GcpProjectFilterConfig`/`resolve_gcp_projects` helper is a good reuse point. A few things need to be addressed before this is ready to merge.

---

## Breaking Change Assessment ✅

This is **not a breaking change**. Existing configs that set `project_ids` explicitly continue to work identically. The new fields (`project_id_pattern`, `project_labels`) are purely additive. The only soft-break is that the `ValidationError` message changed ("At least one project must be specified" → "At least one project selector must be specified"), which is acceptable.

---

## Issues

### 🔴 BLOCKER 1 — Role name in docs may be wrong

**Files:** `dataplex_pre.md`, `dataplex_recipe.yml`

The docs and recipe say auto-discovery requires `roles/resourcemanager.projectViewer` on the parent organization or folder. This is almost certainly incorrect.

`roles/resourcemanager.projectViewer` grants `resourcemanager.projects.get` on a **specific project** — it does **not** grant the ability to list or search projects across an org/folder. For `search_projects()` and `list_projects()` (the APIs called by `_search_projects_by_labels` and `_list_all_projects` in `gcp_project_filter.py`), you need `resourcemanager.projects.search` / `resourcemanager.projects.list`, which are provided by roles like `roles/resourcemanager.folderViewer` or `roles/resourcemanager.organizationViewer` at the org/folder scope.

This is the same class of error that was fixed in #17250 (which changed `roles/resourcemanager.projectViewer` → `roles/resourcemanager.folderViewer` for the term-association case). Please verify what the `SearchProjects` and `ListProjects` APIs actually require against the [GCP IAM permissions reference](https://cloud.google.com/resource-manager/docs/access-control-proj) and update the docs accordingly.

---

### 🔴 BLOCKER 2 — API call inside `__init__`

**File:** `dataplex.py`, line ~245

```python
self._project_ids: List[str] = self._resolve_target_projects(credentials)
```

Calling the Resource Manager API during object construction is a well-known anti-pattern in this codebase:
- Constructor failures are hard to distinguish from configuration errors
- Makes unit testing painful — all tests that construct `DataplexSource` via `object.__new__()` must manually stub `_project_ids` (visible in the 9 test patches added in `test_dataplex_source.py`)
- Breaks the contract of `test_connection()`, which instantiates the config but should not trigger full API calls

**Recommendation:** Resolve projects lazily — either in `get_workunits_internal()` before iteration begins, or via a `@cached_property`. This is how BigQuery handles its project resolution.

---

### 🔴 BLOCKER 3 — `test_connection` does not validate credentials or Resource Manager access in pattern-only mode

**File:** `dataplex.py`, lines ~349–370

When `project_ids` is empty (pattern/label mode), the test connection path does:
```python
catalog_client = dataplex_v1.CatalogServiceClient(credentials=credentials)
if config.project_ids:
    ...  # only branch that makes a real network call
# else: falls through — no network call made at all
```

Constructing a `CatalogServiceClient` **does not make a network call** and does not validate credentials. The `test_test_connection_pattern_only_skips_dataplex_listing` test confirms this — it patches the client constructor and asserts `capable=True` without any real I/O.

For pattern/label mode, `test_connection` should:
1. Attempt to resolve at least one project via Resource Manager (to validate those credentials/permissions)
2. If a project is found, attempt to list entry groups on it (to validate Dataplex API access)

Without this, users who misconfigure Resource Manager permissions will only discover the problem at full ingestion time, not at connection test time.

---

### 🟡 WARNING — `GcpProjectFilterConfig` already exists; don't duplicate its fields

**File:** `dataplex_config.py`

`GcpProjectFilterConfig` in `gcp_project_filter.py` already defines exactly these three fields:
```python
project_ids: List[str]
project_labels: List[str]
project_id_pattern: AllowDenyPattern
```

`DataplexConfig` re-declares all three, and `_resolve_target_projects` then copies them into a fresh `GcpProjectFilterConfig` instance just to call `resolve_gcp_projects`. This is a bridging anti-pattern.

**Recommendation:** Either have `DataplexConfig` inherit from `GcpProjectFilterConfig` (or a thin mixin extracted from it), so the fields are defined once. You'd then pass `self.config` (or `self`) directly to `resolve_gcp_projects` since it already satisfies `GcpProjectFilterConfig`. Note that `VertexAIConfig` and `BigQueryFilterConfig` have the same duplication problem — this PR is a good opportunity to introduce consistency, even if the other connectors are a separate follow-up.

---

### 🟡 WARNING — `_validate_pattern_does_not_filter_all` is logically inconsistent

**File:** `dataplex_config.py`, lines ~345–357

The docs and `resolve_gcp_projects` both state: "`project_id_pattern` is **ignored** when `project_ids` is set explicitly." Yet this validator raises an error when `project_id_pattern` would filter out all explicit `project_ids` — a combination that can never affect runtime behavior.

This validator creates user confusion (why does this combination fail at config time if `project_id_pattern` is ignored at runtime?) and should either be removed or the docs/runtime behavior should be made consistent. The equivalent validator in `VertexAIConfig` has the same issue.

---

### ℹ️ SUGGESTION — Redundant regex validation in `dataplex_config.py`

**File:** `dataplex_config.py`, `_validate_project_id_pattern_syntax`

`AllowDenyPattern` compiles its patterns at construction time (via `re.compile`). A separate validator that re-validates the same patterns is redundant and creates drift risk if the pattern type changes. Verify whether `AllowDenyPattern` already raises on invalid regex; if so, remove the custom validator.

---

### ℹ️ SUGGESTION — No test for empty project resolution at runtime

When `resolve_gcp_projects` returns `[]` (e.g. pattern matches nothing, or Resource Manager error), `get_workunits_internal` will silently produce zero workunits. `report.failure()` is called but the run completes without raising. A test case covering this path (and asserting the failure is surfaced) would be valuable.

---

## Summary

| # | Severity | Issue |
|---|----------|-------|
| 1 | 🔴 BLOCKER | Role name `roles/resourcemanager.projectViewer` for auto-discovery is likely wrong — please verify against GCP IAM docs |
| 2 | 🔴 BLOCKER | Network API call in `__init__` — move to lazy resolution |
| 3 | 🔴 BLOCKER | `test_connection` in pattern-only mode doesn't test anything meaningful |
| 4 | 🟡 WARNING | `DataplexConfig` duplicates `GcpProjectFilterConfig` fields instead of reusing the shared type |
| 5 | 🟡 WARNING | `_validate_pattern_does_not_filter_all` validates a combination the runtime explicitly ignores |
| 6 | ℹ️ | Redundant regex validator |
| 7 | ℹ️ | Missing test for empty-project runtime path |

---

sgomezvillamor:
@davidwalker-tp Overall the PR looks good, Claude was a bit picky with the review 😅
But the feedback is straightforward to address. Claude can handle it very easily. So it should be a quick fix to get this merged!

davidwalker-tp:
Thanks for the thorough review @sgomezvillamor — all blockers and warnings addressed in `9ecd71e26d` (with a follow-up merge `f7dee2a141`):

**🔴 BLOCKER 1 — IAM role name** — auto-discovery row in `dataplex_pre.md` and the comment in `dataplex_recipe.yml` now use `roles/resourcemanager.folderViewer` (matches #17250). The supporting prose was updated too.

**🔴 BLOCKER 2 — API call in `__init__`** — `_project_ids` is now a `@cached_property`. `__init__` only stores credentials on `self`; the Resource Manager call is deferred until first read. Tests using `object.__new__()` now seed `source._project_ids = [...]` to skip the property, which is straightforward.

**🔴 BLOCKER 3 — `test_connection` in pattern-only mode** — when `project_ids` is empty, `test_connection` now (a) constructs a `ProjectsClient` and resolves at least one project via `resolve_gcp_projects` (validating Resource Manager creds + role), then (b) lists entry groups on the resolved project (validating Dataplex API access). New tests cover both the success and the no-projects-resolved failure paths.

**🟡 WARNING — Field duplication** — `DataplexConfig` now inherits `GcpProjectFilterConfig`. The three duplicated field declarations are gone; `_resolve_target_projects` no longer constructs a fresh `GcpProjectFilterConfig` and instead passes `self.config` straight into `resolve_gcp_projects`.

**🟡 WARNING — Inconsistent validator** — `_validate_pattern_does_not_filter_all` removed. As you noted, the runtime ignores `project_id_pattern` when `project_ids` is set, so the validator was failing on a combination the runtime never honoured.

**ℹ️ Suggestion — Redundant regex validator** — `_validate_project_id_pattern_syntax` removed. `AllowDenyPattern` surfaces invalid regex on first compile; the early validator added timing rather than coverage.

**ℹ️ Suggestion — Empty-resolution test** — added `test_get_workunits_internal_with_empty_resolved_projects`, which exercises the path where `resolve_gcp_projects` returns `[]` (the resolver itself surfaces the failure on the report).

All 263 Dataplex unit tests pass; pre-commit is clean.

sgomezvillamor (approved):
All blockers and warnings addressed — role name corrected, lazy `cached_property` for project resolution, `test_connection` now validates Resource Manager access in pattern-only mode, and `GcpProjectFilterConfig` properly inherited. Great contribution, thanks for the quick turnaround on the feedback!

---

PR #17253, opened 2026-04-30
feat(bigquery): add wif based auth in bigquery

alokr-dhub:
# Summary                                                                                  
Adds Workload Identity Federation (WIF) as an alternative to long-lived service account keys for the BigQuery ingestion source. Enterprise customers have asked for this since their security teams disallow downloadable service account JSON keys.
WIF lets workloads running outside Google Cloud (AWS, Azure, on-prem Kubernetes, self-hosted Docker) impersonate a Google service account using short-lived tokens minted from an external identity provider.                                                       
 
## Implementation
    - New config field auth_type on BigQueryConnectionConfig (defaults to service_account, the existing behavior). Set to workload_identity_federation to enable WIF.
    - Reuses the existing GCPWIFConfig mixin added for GCS in #16002. Same three configuration options:                                                                   
    - gcp_wif_configuration — path to a WIF JSON file
    - gcp_wif_configuration_json — inline dict or JSON string
    - gcp_wif_configuration_json_string — JSON string (for secret managers)
    - Mixin enhancements (also benefits GCS): added to_wif_dict() and wif_config_source() methods, removed duplicated resolution logic from load_wif_credentials, eliminated three # type: ignore directives.
    - Validators enforce mutual exclusion: credential and the gcp_wif_configuration* options 
  can't be combined; WIF requires exactly one of the three options.                        
    - SQLAlchemy compat: writes the WIF JSON to a temp file and sets GOOGLE_APPLICATION_CREDENTIALS so the SQLAlchemy BigQuery dialect (used by the Fivetran source) continues to work.                                                                                

## Backward compatibility
    - Default auth_type=service_account preserves the existing behavior — no recipe changes  
  required.                                                 
    - credential remains Optional to keep the existing Application Default Credentials (ADC)
  path working for users on GCE/GKE.                                                       
    - Refactored credential setup from __init__ to a model_validator(mode="after") with an
  idempotency guard — pre-existing service-account paths are covered by new regression     
  tests.                                                    

## Files changed
    - metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_connection.py auth_type enum, WIF mixin, validators, credential setup
    - metadata-ingestion/src/datahub/ingestion/source/common/gcp_wif_config.py added to_wif_dict() + wif_config_source(), deduplicated load_wif_credentials
  - metadata-ingestion/tests/unit/bigquery/test_bigquery_connection.py new (10 tests covering WIF, SA regression, ADC fallback, SQLAlchemy URL)
  - metadata-ingestion/tests/unit/test_gcp_wif_config.py — added direct tests for to_wif_dict() + wif_config_source()
  - metadata-ingestion/docs/sources/bigquery/bigquery_pre.md — WIF section + role requirements
  - metadata-ingestion/docs/sources/bigquery/bigquery_recipe.yml — three WIF recipe examples
## Test plan                                                    
  - Unit tests added: 10 new tests in test_bigquery_connection.py covering WIF auth, service-account regression, ADC fallback, SQLAlchemy URL behavior
  - Mixin tests added: 9 new tests in test_gcp_wif_config.py covering to_wif_dict() and  wif_config_source()
  - All existing BigQuery unit tests pass (201/201, excluding 1 pre-existing unrelated failure on master)
  - All existing GCS WIF tests pass (16/16) — confirms the mixin refactor is non-breaking
  - ./gradlew :metadata-ingestion:lintFix clean
  - mypy clean on changed files
  - Markdown prettier check passes
  - Manual end-to-end test against a real WIF-configured BigQuery instance (recommended before merge)

- [ ] The PR conforms to DataHub's [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [PR Title Format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#pr-title-format))


askumar27 (commented):
Strong PR overall. A few things I particularly liked:

The security model is right -- credentials stay in memory via `PrivateAttr`, no temp files, no env var leaks. The tests explicitly verify this, which is great.

Moving from `__init__` to `model_validator(mode="after")` with the idempotency guard is the correct Pydantic v2 pattern, and the reentrancy tests documenting `model_copy()` vs `model_validate()` behavior are unusually thorough. Most PRs skip that kind of edge case.

The `default={}` to `default_factory=dict` fix on `extra_client_options` is a subtle but real bug fix. Good catch.

The `make_gcp_logging_client` refactor collapsing the duplicate if/else into one return is a nice cleanup -- small but it removes a whole class of future copy-paste bugs.

Fivetran inheritance tests are a smart addition. Easy to forget that `BigQueryDestinationConfig` inherits from `BigQueryConnectionConfig`, and these tests pin that the validator fires correctly through the subclass.

The 5 inline comments below are all minor. The only one I'd want addressed before merge is the breaking change to `gcp_wif_configuration_json`'s type -- existing GCS recipes that pass a string will break.

metadata-ingestion/src/datahub/ingestion/source/common/gcp_wif_config.py
```diff
@@ -33,10 +32,12 @@ class GCPWIFConfig(ConfigModel):
         ),
     )
 
-    gcp_wif_configuration_json: Optional[Union[str, Dict[str, Any]]] = Field(
+    gcp_wif_configuration_json: Optional[Dict[str, Any]] = Field(
         default=None,
```
askumar27, line 37: This is a breaking change for existing GCS users. The type went from `Optional[Union[str, Dict[str, Any]]]` to `Optional[Dict[str, Any]]`, which means anyone currently passing a JSON string to `gcp_wif_configuration_json` in their GCS recipe will get a validation error after upgrading.

Two options: add a `mode="before"` validator that auto-parses strings into dicts (backward compat), or document this as a breaking change in `docs/how/updating-datahub.md` with migration instructions.

metadata-ingestion/src/datahub/ingestion/source/common/gcp_wif_config.py
```diff
@@ -106,17 +105,99 @@ def _validate_wif_mutual_exclusion(self) -> "GCPWIFConfig":
...
+                ) from e
+            if not isinstance(loaded, dict):
+                raise ValueError(
+                    f"WIF configuration must be a JSON object, not a {type(loaded).__name__}"
+                )
+            return loaded
+        elif self.gcp_wif_configuration_json is not None:
+            # Return a copy to protect the model's internal dict from mutation by callers.
+            return dict(self.gcp_wif_configuration_json)
+        elif self.gcp_wif_configuration_json_string is not None:
```
askumar27, line 157: `dict(self.gcp_wif_configuration_json)` is a shallow copy. WIF configs always have nested dicts (`credential_source`, etc.), so a caller mutating a nested value would still change the model's internal state. The existing test only checks top-level mutation.

`json.loads(json.dumps(...))` or `copy.deepcopy()` would fix this.

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_connection.py
```diff
@@ -2,14 +2,19 @@
...
-from pydantic import Field, PrivateAttr
+from pydantic import Field, PrivateAttr, model_validator
 
 from datahub._version import __version__
-from datahub.configuration.common import ConfigModel
 from datahub.ingestion.source.common.gcp_credentials_config import GCPCredential
+from datahub.ingestion.source.common.gcp_wif_config import (
+    GCPWIFConfig,
+    _build_credentials_from_wif_dict,
+)
```
askumar27, line 16: `_build_credentials_from_wif_dict` has a leading underscore (private), but it's now imported across modules. Since this is part of the cross-module API, consider dropping the underscore -- or at least adding it to `__all__` in `gcp_wif_config.py`.

metadata-ingestion/src/datahub/ingestion/source/common/gcp_wif_config.py
```diff
@@ -106,17 +105,99 @@ def _validate_wif_mutual_exclusion(self) -> "GCPWIFConfig":
...
+        logger.info(
+            "Loading Workload Identity Federation credentials (source=%s)",
+            source_label,
+        )
+        credentials, project_id = load_credentials_from_dict(wif_dict)
+        # For WIF → SA impersonation, scopes must be set; otherwise IAM returns
+        # 400 "Scope required."  Guard with requires_scopes so credential types
+        # that manage access via audience (not scopes) are not broken.
+        if getattr(credentials, "requires_scopes", False):
+            credentials = credentials.with_scopes(
```
askumar27, line 188: Nitpick: `getattr(credentials, "requires_scopes", False)` works, but `google.auth.credentials.Scoped` exists for exactly this check. `isinstance(credentials, Scoped)` would be more explicit about intent.

That said, `getattr` is what google-auth uses internally too, so this is more of a style preference than a bug.

metadata-ingestion/tests/unit/test_gcp_wif_config.py
```diff
@@ -118,17 +111,88 @@ def test_no_config_raises(self) -> None:
...
+        config = GCPWIFConfig(gcp_wif_configuration=str(f))
+        assert config.to_wif_dict() == payload
+
+    def test_from_dict(self) -> None:
+        payload = {"type": "external_account"}
+        config = GCPWIFConfig(gcp_wif_configuration_json=payload)
+        assert config.to_wif_dict() == payload
+
+    def test_from_dict_returns_copy_not_alias(self) -> None:
+        # to_wif_dict() must return a copy so callers cannot mutate the model's
```
askumar27: The copy-not-alias test only checks top-level key insertion. It misses the nested case:

```python
result["credential_source"]["injected"] = True  # would mutate internal state
```

Worth adding since WIF configs always have nested dicts.

askumar27 (approved):
Approving to unblock, left minor comments

metadata-ingestion/src/datahub/ingestion/source/common/gcp_wif_config.py
alokr-dhub: Fixed. Added a `mode='before'` validator that auto-coerces JSON strings to dicts, so existing recipes passing a JSON string to `gcp_wif_configuration_json` continue working without changes. Invalid JSON strings raise a clear error. Field description updated to reflect that both dict and JSON string are accepted.

metadata-ingestion/src/datahub/ingestion/source/common/gcp_wif_config.py
alokr-dhub: Fixed. Changed `dict(self.gcp_wif_configuration_json)` to `copy.deepcopy(self.gcp_wif_configuration_json)`. The test is also updated to verify nested mutation is blocked — it now uses a payload with a `credential_source` sub-dict and asserts that mutating both the top-level and nested keys on the returned dict does not affect subsequent calls.

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/bigquery_connection.py
alokr-dhub: Fixed. Renamed to `build_credentials_from_wif_dict` (dropped the underscore) across all callers — `gcp_wif_config.py`, `bigquery_connection.py`, and all 14 test patch strings.

metadata-ingestion/src/datahub/ingestion/source/common/gcp_wif_config.py
alokr-dhub: Kept as `getattr(credentials, 'requires_scopes', False)`. Agree `isinstance(credentials, Scoped)` is more explicit, but `getattr` is easier to mock in tests without needing the mock to pass an `isinstance` check against an ABC. Since google-auth uses the same pattern internally, happy to leave it as-is unless you'd prefer the `isinstance` form.

metadata-ingestion/tests/unit/test_gcp_wif_config.py
alokr-dhub: Fixed. The test now uses a payload with a nested `credential_source` dict and asserts that mutating both `result['injected']` and `result['credential_source']['injected']` on the returned copy does not affect the model's internal state on the next call.

PR #17256, opened 2026-04-30
fix(ingestion): MCPObserver failures must not fail ingest

kyungsoo-datahub:
Three defensive layers, innermost to outermost:

MCPObserver.apply() — catch any Throwable from the shouldApply() filter or observeMCPs() implementation; log at WARN with batch size.
AspectsBatch.applyMCPObservers() dispatch loop — per-observer catch so one bad observer doesn't kill healthy ones.
EntityServiceImpl.ingestProposal() call site — outermost guard for anything that leaks through.
All logs at WARN — ERROR-level visibility is broken on the affected branch.

IngestionMetricsEmitter catch blocks also changed from ERROR to WARN and now log urn/aspect/changeType instead of e.getMessage() (which was literally "null", giving nothing to grep on).

### DIFF
--- a/entity-registry/src/main/java/com/linkedin/metadata/aspect/batch/AspectsBatch.java
+++ b/entity-registry/src/main/java/com/linkedin/metadata/aspect/batch/AspectsBatch.java
@@ -25,6 +25,8 @@ import java.util.stream.Stream;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * A batch of aspects in the context of either an MCP or MCL write path to a data store. The item is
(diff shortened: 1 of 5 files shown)


entity-registry/src/main/java/com/linkedin/metadata/aspect/batch/AspectsBatch.java
```diff
@@ -172,11 +173,21 @@ default void applyMCPObservers(Collection<? extends BatchItem> items) {
...
-        .forEach(observer -> observer.apply(items, retrieverContext));
+    for (MCPObserver observer :
+        retrieverContext.getAspectRetriever().getEntityRegistry().getAllMCPObservers()) {
+      try {
+        observer.apply(items, retrieverContext);
+      } catch (Throwable t) {
+        // Belt-and-suspenders. observer.apply() is final + try/catch, but this guarantees that an
+        // exception during dispatch (registry lookup, plugin construction, etc.) cannot fail the
+        // ingest batch.
+        org.slf4j.LoggerFactory.getLogger(AspectsBatch.class)
```
craig-rueda: This is a weird pattern. Why not just get the logger once?

kyungsoo-datahub: Thank you for your comment. Revised.

david-leifker (approved):

PR #17259, opened 2026-04-30
feat(ingest/sigma): emit cross-DM FineGrainedLineage via global bridge index

kyungsoo-datahub:
Extends intra-DM resolver to consult the globally-populatedcross-DM bridge index when a bracket ref doesn't match any sibling element in the current DM. Validates the column on the resolved upstream and emits FineGrainedLineage

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/sigma/config.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/sigma/config.py
@@ -272,14 +272,14 @@ class SigmaSourceReport(StaleEntityRemovalSourceReport):
     # How many FGL entries were emitted across all elements.
     data_model_element_fgl_emitted: int = 0
     # Refs where multiple sibling candidates passed the /lineage filter;
-    # sorted-first URN was chosen (matches T2 PR1's collision precedent).
+    # sorted-first URN was chosen (matches collision precedent).
     data_model_element_fgl_collision_pick_first: int = 0
     # Refs whose source element is outside this DM; deferred to cross-DM resolution.
     data_model_element_fgl_cross_dm_deferred: int = 0
     # Refs where element-name matches the element's own warehouse-table name
     # (e.g., element "data.csv" with formula "[data.csv/col]"). These are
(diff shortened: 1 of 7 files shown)


sgomezvillamor (approved):
Logic is sound and the unit/integration test coverage on the new FGL path is solid. A few things worth cleaning up before merge — see inline comment.

---

sgomezvillamor:
**Naming clarity**

A few related naming issues worth addressing together:

- `dm_url_id_to_dm_id` — `dm` appears on both sides with nothing to distinguish them. Following the existing `_by_url_id` suffix convention already used in this file, `data_model_id_by_url_id` would be unambiguous.
- `dm_element_urn_to_cols: Dict[str, Dict[str, str]]` — the inner dict structure (lowercase → canonical column name) is invisible from the type. A short inline comment `# {lowercase_col: canonical_col}` on the same line would help.
- `consumed_dm_element_names` on `SigmaDataModel` initially reads as "element names consumed from myself." Consider `source_dm_element_names`.
- `data_model_element_fgl_cross_dm_*` mixes full words and abbreviation for the same concept in a single identifier. Adding a comment block header clarifying that `dm` = data model throughout would reduce confusion for readers of the report.

**Tests and coverage**

Connector tests are failing on the latest commit — needs investigation before merge.

`sigma_api.py` patch coverage is 16.7% (5 lines): the `elif entry_type == "data-model":` branch edge cases (non-string `src_dm_id`, whitespace-only `src_name`) appear untested.

**Entity-level as soft tiebreaker, not a gate**

The inline comment in `_build_dm_element_fine_grained_lineages` explains this well. Worth echoing a one-liner near the `fgl_cross_dm_resolved` counter definition so operators reading the report know a resolved cross-DM FGL does **not** imply entity-level confirmation.

**Total FGL needs two counters**

`fgl_emitted` is intra-DM only; total FGL = `fgl_emitted + fgl_cross_dm_resolved`. A note near the counter definitions would save operators a source-dive.


---

PR #17267, opened 2026-04-30
build(docker): stable Oracle Instant Client path via symlink

david-leifker:
Point /opt/oracle/instantclient at the versioned unzip directory and register that path in ld.so.conf so Python thick mode and tooling keep working when IC_UNZIP_DIR changes.

Made-with: Cursor


askumar27 (approved):
LGTM!

RyanHolstien (approved):

PR #17269, opened 2026-05-01
feat(ingest/sigma): introduce Sigma Connection registry

kyungsoo-datahub:
Build connection_id -> {platform, host, account} registry from /v2/connections at SigmaSource init. 
Foundation for Sigma Warehouse Connection works; zero behavior change to existing CLL surfaces.


sgomezvillamor (approved):
Nice clean foundation — architecture, error handling, and test coverage are solid. A few suggestions before the follow-up PR lands:

**1. No API calls in `__init__`**
`_build_connection_registry()` is called inside `SigmaSource.__init__`, so every instantiation (including `test_connection()`) pays the `/v2/connections` round-trip. The DataHub pattern is to defer data-fetching to `get_workunits_internal`. Moving the registry build there — lazily on first use or as an explicit first step — keeps the constructor side-effect-free and consistent with the rest of the source.

**2. Registry isn't consumed yet — implications for the breaking-change note**
The registry is built but nothing reads it in this PR. Given that, the permission note in `sigma_pre.md` ("existing credentials may need regeneration") is slightly premature — no user-visible feature depends on `/v2/connections` scope yet. I'd suggest deferring the docs update to the PR that *uses* the registry. At that point: if the feature is opt-in, no `updating-datahub.md` entry is needed; if it's on by default, a short entry there will be the right place to surface the new permission requirement.

**3. Observability — surface unmappable types by name**
`connections_unmappable_type: int` tells operators *how many* types are unrecognised but not *which*, making it unactionable without debug logs. Suggest adding:
```python
connections_unmappable_types: List[str] = field(default_factory=list)  # distinct Sigma type strings
```
and populating it in `build()`.

**4. Observability — expose the resolved registry in the report**
For operators who want to audit what was discovered, a:
```python
connections_registry: List[SigmaConnectionRecord] = field(default_factory=list)
```
field on `SigmaSourceReport` would make the full resolved registry visible in the ingestion report without extra tooling. Given expected tens-of-entries size, this is inexpensive.

---

PR #17271, opened 2026-05-01
feat(ingest/clickhouse): add opt-in SQLAlchemy profiler backend

dinesh-verma-datahub:
## Summary

Adds an opt-in SQLAlchemy-based profiling backend for ClickHouse to work around a 100% column-profiling failure on the default Great Expectations profiler.

**Root cause.** Great Expectations' `SqlAlchemyDataset` materialises query batches via an internal `CREATE TEMPORARY TABLE ... AS SELECT` (triggered when `custom_sql`, `limit`, or `offset` is set). ClickHouse temporary tables are **session-scoped**, so when the follow-up read lands on a different pooled connection it raises `Code: 60. DB::Exception: Table … does not exist. (UNKNOWN_TABLE)`. The temp-table approach is incompatible with ClickHouse's connection-pool semantics — there is no fix on the GE side without a major upstream change.

**Approach.** This PR is fully **opt-in** — defaults are unchanged. Existing recipes continue to use the GE profiler with no behaviour change. To use the new path, users add a single line:

```yaml
profiling:
  method: sqlalchemy
```

When that flag is set, the ingestion routes through the SQLAlchemy profiler with a new `ClickHouseAdapter` that:

- Uses ClickHouse-native aggregates: `uniq()` (HyperLogLog distinct count), `quantile(0.5)(col)` two-call syntax, `quantiles(l1,l2,…)(col)` for batched quantiles, `stddevSamp()` for sample stddev (since ClickHouse's `stddev` is an alias for `stddevPop`).
- Never creates temporary tables, so the cross-connection-pool failure goes away.
- Reports failures through `SQLSourceReport.warning(...)` with `exc=` so they are visible in the ingestion report and Sentry, not just stdout.
- Explicitly warns (and falls back to full-table profiling) if `custom_sql`, `limit`, or `offset` are set — these options require temp tables and are not supported on this path.

## Why opt-in

The GE profiler and the SQLAlchemy profiler can produce slightly different numerical values on the same data (e.g. `uniq()` HyperLogLog vs GE's exact `COUNT(DISTINCT)`). To avoid any observable change for existing customers, the default stays at `method: ge`. Customers hitting ING-2086 (or who want the new aggregate path) opt in explicitly.

## Unsupported options on the SQLAlchemy path

`custom_sql`, `limit`, and `offset` are not supported when `method: sqlalchemy` is set. These options work by materialising a subset of the data into a temp table before profiling — exactly the mechanism that breaks on ClickHouse. If any of these are present in the recipe, the adapter emits a structured warning (`SQLSourceReport.warning`) and profiles the full table instead. The GE path (`method: ge`) still supports them.

## Test plan

- [x] Unit tests for `ClickHouseAdapter`: 31 tests covering happy paths, every error branch, the batched → per-quantile fallback chain, validation, label semantics, the `_format_context` helper, `setup_profiling` warning for each unsupported option (`custom_sql`, `limit`, `offset`), and the no-warning path for normal full-table profiling.
- [x] Existing tests pass — 82 unit tests in `tests/unit/sqlalchemy_profiler/test_adapters.py` and `tests/unit/test_clickhouse_source.py` green.
- [x] `./gradlew :metadata-ingestion:lintFix` clean (ruff + format + mypy).
- [x] Validated `method: ge` default still works (default behaviour preserved).

## Regression analysis

| Customer profile | Before | After |
|---|---|---|
| Profiling disabled | No profiling code runs | No profiling code runs |
| `enabled: true`, no `method` | GE profiler runs (the broken path) | **Identical** — GE profiler runs (unchanged default) |
| `enabled: true`, explicit `method: ge` | GE profiler runs | **Identical** |
| `enabled: true`, explicit `method: sqlalchemy` | SQLAlchemyProfiler with `GenericAdapter` — fails on `stddev()` / `PERCENTILE_CONT` | **Now works** with `ClickHouseAdapter` |
| `enabled: true`, `method: sqlalchemy`, `custom_sql` / `limit` / `offset` set | SQLAlchemyProfiler with `GenericAdapter` | Warning emitted; full-table profiling used (same as if options were absent) |
| Programmatic `ClickHouseConfig(profiling=GEProfilingConfig(...))` | Works | **Identical** |
| Other source connectors | n/a | Zero impact — only the `clickhouse` factory branch was added |


metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/clickhouse.py
```diff
@@ -0,0 +1,216 @@
...
+from sqlalchemy.exc import SQLAlchemyError
+from sqlalchemy.sql.elements import ColumnElement
+
+from datahub.ingestion.source.sqlalchemy_profiler.base_adapter import (
+    DEFAULT_QUANTILES,
+    PlatformAdapter,
+)
+
+
+class ClickHouseAdapter(PlatformAdapter):
```
kyungsoo-datahub, line 19: Is the omission of setup_profiling() intentional because ClickHouse SQLAlchemy profiling only supports full-table profiling? If so, can we explicitly reject or warn on custom_sql, limit, and offset, and update the docs? Right now those options appear to be silently ignored, which could produce misleading profiles.

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/clickhouse.py
```diff
@@ -0,0 +1,216 @@
...
+            self.report.warning(
+                title="Profiling: failed to disambiguate stdev null result",
+                message="Non-null count query failed after stddevSamp() returned NULL",
+                context=_format_context(table, column),
+                exc=e,
+            )
+            return None
+        return None if non_null_count <= 1 else 0.0
+
+    def supports_row_count_estimation(self) -> bool:
```
kyungsoo-datahub: If the intent is to use ClickHouse’s cheap system.tables.total_rows row count, I don’t think this path is currently called. In _profile_row_count(), use_estimation is only set for Postgres/MySQL:

```
use_estimation = (
    self.config.profile_table_row_count_estimate_only
    and platform in ("postgresql", "mysql")
)
```

So ClickHouse will still use the normal COUNT(*) row count path. Is that intentional? If yes, maybe we don’t need supports_row_count_estimation() / get_estimated_row_count() here, or the docs should avoid saying row counts come from system.tables.

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/clickhouse.py
```diff
@@ -0,0 +1,216 @@
...
+                results.append(float(result) if result is not None else None)
+            except (ValueError, TypeError) as e:
+                if first_exc is None:
+                    first_exc = e
+                failures.append(q)
+                results.append(None)
+
+        if failures:
+            self.report.warning(
+                title="Profiling: some ClickHouse quantiles unavailable",
```
kyungsoo-datahub, line 184: nit: this warning message includes dynamic values. In SourceReport.warning() / report_warning(), message is intended to be a stable LiteralString, while dynamic details should go in context for grouping/aggregation.

dinesh-verma-datahub:
@kyungsoo-datahub thanks for the review, addressed the comments.

kyungsoo-datahub (approved):

PR #15391, opened 2025-11-24
fix(snowflake): support conditional multi-table INSERT statements (#14929)

dinesh-github:
## Summary

Snowflake's conditional multi-table INSERT statements (`INSERT ALL WHEN ... THEN INTO table1 ... WHEN ... THEN INTO table2 ...`) produce multiple entries in the `objects_modified` array of `ACCESS_HISTORY`.

**Before:** Only the last table in a multi-table INSERT had lineage recorded. A warning was logged: *"Unexpectedly got multiple downstream entities from the Snowflake audit log."* Additionally, a pre-existing bug caused `DownstreamColumnRef.table` to always be `None` for audit-log-derived column lineage (wrong pydantic kwarg name).

**After:** Every downstream table gets its own `PreparsedQuery` entry with correct column lineage, unique query fingerprints, and proper usage attribution. Malformed audit log entries are isolated per-object so one bad entry doesn't discard valid siblings.

### Generator-based `_parse_audit_log_row`

Refactored from `Optional[Union[...]]` return to `Iterable[Union[...]]` generator. The caller (`fetch_query_log`) uses `yield from` instead of isinstance dispatch, simplifying the control flow.

### `_build_downstream_targets` with per-object error isolation

Extracted downstream table processing into a dedicated method. Each object in `objects_modified` is wrapped in its own `try/except (KeyError, ValueError, TypeError)` so one malformed audit log entry doesn't discard valid siblings. When ALL objects fail, falls back to `ObservedQuery` for SQL-based lineage.

### `_DownstreamTarget` frozen dataclass

New internal type with a `__post_init__` invariant ensuring `downstream` and `column_lineage` are always both `None` (usage-only queries) or both non-`None` (queries with a downstream table).

### Unique `query_id` per downstream table

Multi-table INSERTs append `_{objectName}` to `secondary_id` before fingerprinting, so `SqlParsingAggregator` stores separate lineage metadata for each downstream instead of overwriting in `_query_map`.

### Usage double-counting prevention

Only the first yielded entry carries `query_count`; subsequent entries get `0` to avoid inflating upstream read counts when the same query fans out to N downstream tables.

### Bug fix: `DownstreamColumnRef(dataset=...)` &rarr; `DownstreamColumnRef(table=...)`

[...]

PR #12318, opened 2025-01-10
feat(ingestion/snowflake):adds streams as a new dataset with lineage and properties.

brock-acryl:
## Checklist

- [ ] The PR conforms to DataHub's [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [Commit Message Format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#commit-message-format))
- [ ] Links to related issues (if applicable)
- [ ] Tests for the changes have been added/updated (if applicable)
- [ ] Docs related to the changes have been added/updated (if applicable). If a new feature has been added a Usage Guide has been added for the same.
- [ ] For any breaking change/potential downtime/deprecation/big changes an entry has been made in [Updating DataHub](https://github.com/datahub-project/datahub/blob/master/docs/how/updating-datahub.md)
Properly picks up:
- Streams and columns in a stream (including metadata columns)
- Stream upstream lineage from show streams statement
- Stream downstream lineage by parsing queries (tested inserts, CTAS, inserts with unions)
- Stream properties
Screenshot 2025-01-10 at 11 11 23 AM

PR #17272, opened 2026-05-01
fix(ingest/snowflake): preserve multi-target lineage for Stream-driven INSERT ALL

dinesh-verma-datahub:
## Related

- Builds on: [#15391](https://github.com/datahub-project/datahub/pull/15391) (per-downstream emission)
- Refines: [#12318](https://github.com/datahub-project/datahub/pull/12318) (original Stream-bypass)

## Summary

When a Snowflake `INSERT ALL` reads from a **Snowflake Stream** and writes to two or more downstream tables, only one of the downstreams was receiving fresh upstream lineage on every ingestion run.

The bug was a coverage gap between two earlier PRs:

- [PR #12318](https://github.com/datahub-project/datahub/pull/12318) (Feb 2025) introduced a Stream-bypass in `_parse_audit_log_row`: any query whose `direct_objects_accessed` contained an object with `objectDomain="Stream"` was emitted as a single `ObservedQuery` and routed through `sqlglot` parsing. The bypass exists because Snowflake leaks placeholder `$SYS_VIEW_<id>` object names into the audit log for some stream-driven queries (undocumented; observed empirically), and URNs built from those names are unusable.
- [PR #15391](https://github.com/datahub-project/datahub/pull/15391) (Apr 2026) added per-downstream `PreparsedQuery` emission in `_build_downstream_targets` so multi-table `INSERT ALL` queries no longer lost siblings to fingerprint collision. But this code lives **after** the bypass, so it never ran for queries with stream upstreams.
- The aggregator-side fallback path (`SqlParsingAggregator.add_observed_query` → `parsed.out_tables[0]`) only registers the first downstream that sqlglot finds. For the customer's INSERT ALL, sqlglot returns `[jsonschemavalidation_error, customer_profile]` — explaining why exactly one of two downstreams gets fresh lineage on every run.

This PR narrows the Stream-bypass so it only fires when audit-log object names are actually unusable (`$SYS_VIEW_<id>` placeholder or `$`-prefix anomaly) **or** the query is single-target. Clean **multi-target** stream-driven `INSERT ALL` now reaches PR #15391's per-downstream path, and both downstreams emit lineage.

## Behavior change scope

The **only** queries whose behavior changes are: stream-driven INSERTs with `len(objects_modified) > 1` AND clean (non-`$SYS_VIEW_`, non-`$`-prefix) target names. Everything else preserves the original PR #12318 path:

| Query shape | Pre-fix | Post-fix |
| --- | --- | --- |
| Single-target stream INSERT | bypass → sqlglot | bypass → sqlglot (unchanged) |
| Multi-target stream INSERT, `$SYS_VIEW_<id>` in audit | bypass → sqlglot | bypass → sqlglot (unchanged) |
| Multi-target stream INSERT, `$`-prefix anomaly in audit | bypass → sqlglot | bypass → sqlglot (preserved as drift safety net) |
| Multi-target stream INSERT, clean names | bypass → sqlglot (loses N−1 downstreams) | per-downstream → all downstreams emit lineage ✓ |
| Non-stream INSERT (table upstream) | per-downstream | per-downstream (unchanged) |
| `CREATE TEMPORARY VIEW` | bypass → sqlglot | bypass → sqlglot (unchanged) |

## Counters added (additive — no rename)

`SnowflakeQueriesExtractorReport` gains three new counters; the existing `num_stream_queries_observed` is unchanged in name and still increments on bypass:

**Fast-path counter:**

- `num_stream_queries_clean_fast_path` — multi-target stream INSERTs that took the new per-downstream fast path.

**Drift-detection canaries:**

- `num_audit_rows_missing_object_name` — audit rows with missing/empty/whitespace `objectName` (Snowflake schema-drift canary).
- `num_audit_rows_unknown_dollar_prefix` — `$`-prefix object names other than `$SYS_VIEW_` (future-placeholder-shape canary).

A `logger.debug` line on the bypass branch logs `query_id` and the condition flags (`corrupt`, `unknown_dollar`, `multi_target`) so on-call can diagnose individual fallback decisions.

### Unit — `metadata-ingestion/tests/unit/snowflake/test_snowflake_queries.py`

`TestStreamUpstreamMultiTableInsert` (11 tests covering every branch of the new condition):

| Test | Asserts |
| --- | --- |
| `test_stream_upstream_clean_metadata_uses_per_downstream_path` | Customer's case: 2 PreparsedQuery, both with column lineage to the Stream URN. Fast-path counter == 1. |
| `test_stream_upstream_with_sys_view_in_objects_modified_falls_back` | `$SYS_VIEW_42` in `objects_modified` → ObservedQuery. |
| `test_stream_upstream_with_sys_view_in_direct_objects_falls_back` | `$SYS_VIEW_99` in `direct_objects_accessed` → ObservedQuery. |
| `test_no_stream_upstream_uses_per_downstream_path` | Table upstream → per-downstream (regression guard). |
| `test_create_temp_view_still_falls_back` | CREATE TEMPORARY VIEW unaffected. |
| `test_single_target_stream_falls_back_unchanged` | Single-target stream → original bypass (regression guard). |
| `test_multi_target_stream_with_unknown_dollar_prefix_falls_back` | `$FUTURE_PLACEHOLDER` → bypass + drift counter == 1. |
| `test_audit_row_missing_object_name_increments_counter` | None `objectName` in `objects_modified`: counter == 1, single surviving PreparsedQuery, column lineage to the Stream URN preserved. |
| `test_audit_row_none_direct_object_falls_back_to_sql` | None upstream `objectName` → ObservedQuery (avoids overwriting prior lineage with empty upstreams). |
| `test_audit_row_whitespace_direct_object_falls_back_to_sql` | Whitespace-only upstream `objectName` → same ObservedQuery fallback. |
| `test_direct_object_with_missing_columns_key_does_not_crash` | Upstream entry missing the `columns` key → no crash; PreparsedQuery emitted. |

`TestClassifyAuditLogObjects` (11 unit tests on the `_classify_audit_log_objects` helper): empty list, clean object, `$SYS_VIEW_` detected, `$SYS_VIEW` (no trailing underscore) classified as unknown-dollar, unknown-dollar counted separately, None/empty/whitespace name → missing-name counter, no-dot name still classified, mixed objects with one corrupt, mixed missing + unknown-dollar.

### Integration — `metadata-ingestion/tests/integration/snowflake/test_snowflake_queries_lineage.py`

End-to-end runs of `SnowflakeQueriesSource` with a fully mocked Snowflake connection and golden-file comparison of emitted MCPs:

| Test | Asserts |
| --- | --- |
| `test_snowflake_queries_stream_upstream_multi_target_lineage` | Stream upstream + multi-target INSERT ALL → both downstreams emit `upstreamLineage` MCPs (golden file); fast-path counter == 1. |
| `test_snowflake_queries_multi_target_insert_all_table_upstream_lineage` | Table upstream + multi-target INSERT ALL → both downstreams emit `upstreamLineage` MCPs (golden file); stream counters == 0 (regression guard). |


sgomezvillamor:
Overall nice fix — the PR description is excellent, the behavior-change table is exactly the right level of detail, and the 8-branch unit test suite is solid. A few things to address before merge:

---

**🟡 `_classify_audit_log_objects` — documentation + direct unit tests**

The current docstring only describes the return tuple. We need the *why*: why we look at the unqualified name only, what `$SYS_VIEW_*` actually is (an undocumented Snowflake quirk), and why a plain `$`-prefix is a separate canary. Please also add a small dedicated unit test for this function — it has non-trivial edge cases (name with no dot, `$SYS_VIEW` without trailing `_`, multiple objects only one of which is corrupt) that shouldn't rely on indirect coverage through `_parse_audit_log_row`.

**🟡 Removed comment block — replace it with the new strategy**

The deleted block explained *why* the bypass exists. The new variable names tell you *what* the conditions are, but not why streams produce `$SYS_VIEW_*` placeholders, why temp views still need the bypass, or why single-target stream queries intentionally stay on the bypass path. A new comment at the decision point should capture this. The HOW is always recoverable — the WHY is not.

---

**🔁 Required follow-up: connector-level tests for streams**

There are no integration/connector tests covering stream upstreams or multi-downstream `INSERT ALL`. Please open a follow-up ticket tracking both cases (stream upstream golden file, multi-downstream INSERT ALL golden file).

---

**📝 Suggestion: document stream support in the user-facing capabilities page**

This is a good opportunity to update the [Snowflake capabilities docs](https://docs.datahub.com/docs/generated/ingestion/sources/snowflake#capabilities) to note that Snowflake Streams are supported as upstream lineage sources, with a brief note on what is and isn't covered (e.g. single-target queries fall back to SQL parsing, column lineage works for clean multi-target stream INSERTs, etc.).

---

**Minor suggestions:**

- `is_multi_table = len(objects_modified) > 1` is identical to `is_multi_target` computed a few lines earlier — reuse the existing variable.
- `test_audit_row_missing_object_name_increments_counter` only asserts the counter, not how many results are returned. When one target has `None` objectName the fast path still fires and `_build_downstream_targets` silently drops that target — that outcome should be asserted.
- `_classify_audit_log_objects` returns `Tuple[bool, int, int]` — a small `NamedTuple` would make call sites self-documenting and harder to get wrong.
- The `if stream_clean_multi_target:` counter increment after the early-return guard is always-True at that point; a short comment (or folding into an `else`) would make the intent clearer.

---

dinesh-verma-datahub:
@sgomezvillamor thanks for the review, i have made the changes as per review comments, also added the mock integraton tests.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_queries.py
```diff
@@ -888,11 +964,11 @@ def _build_downstream_targets(
                             secondary_id=secondary_id,
                         )
                     )
-                except (KeyError, ValueError, TypeError) as e:
+                except (KeyError, TypeError, ValueError) as e:
                     self.structured_reporter.warning(
                         "Failed to process downstream object",
                         context=f"query_id={query_id}, "
-                        f"object={obj.get('objectName', '<unknown>')}",
+                        f"object={obj.get('objectName') or '<missing>'}",
```
sgomezvillamor: is there any reason for `<unknown>` --> `<missing>` update?
I guess this is mostly cosmetic and the breaking change would be really minor, but if there is no reason, let's just keep previous value

sgomezvillamor (approved):
All raised concerns addressed — great iteration, thank you. The `_AuditLogObjectClassification` dataclass, the new comment block explaining the bypass strategy, the 11-test `TestClassifyAuditLogObjects` suite, the integration tests with golden files, and the `snowflake_post.md` docs section all land well. The additional hardening (empty upstream guard, empty-upstreams fallback to `ObservedQuery`, missing `columns` key guard) was a nice bonus.

One non-blocking follow-up to consider after this merges:

The new integration tests in `test_snowflake_queries_lineage.py` are thorough but require significant knowledge of Snowflake's internal audit log record format to read and maintain — the raw `DIRECT_OBJECTS_ACCESSED` / `OBJECTS_MODIFIED` JSON payloads, the cursor mock wiring, and the golden-file shape are all quite involved. I'd prefer to also have coverage for Snowflake Streams in the **connector-tests** repo, where tests run against a real (or realistic) Snowflake environment and are expressed at a higher level of abstraction. The connector-tests representation would be easier to maintain long-term and would catch regressions that mocked integration tests can't. Could you open a follow-up ticket to add stream upstream lineage (including the multi-target INSERT ALL case) to the connector-tests suite?

---

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_queries.py
dinesh-verma-datahub: reverted

PR #17276, opened 2026-05-01
fix(sigma/ingest): Turn on Sigma Data model ingestion by default.

kyungsoo-datahub:


sgomezvillamor (approved):
LGTM
Minor suggestion: worth adding a short note about the breaking change in `updating-datahub.md`

PR #17284, opened 2026-05-02
feat(ingest/fabric-onelake): extract query usage from queryinsights

aviraj-gour:
## Summary

Adds usage statistics extraction for the Fabric OneLake source by streaming query history from `queryinsights.exec_requests_history` on each Lakehouse/Warehouse SQL Analytics Endpoint and feeding it into the existing `SqlParsingAggregator`.

- New `FabricUsageExtractor` runs per Lakehouse/Warehouse over the same SQL endpoint connection used for schema extraction.
- New `FabricUsageConfig` (extends `BaseUsageConfig`) with:
  - `include_usage_statistics` — master toggle (default `True`).
  - `skip_failed_queries` — server-side filter on `status = 'Succeeded'` (default `True`).
- The aggregator is now constructed unconditionally; the usage flags decide whether `generate_usage_statistics` / `generate_operations` are turned on, so view lineage continues to work when usage is disabled. Operational stats follow the standard `include_operational_stats` flag from `BaseUsageConfig`.
- Stateful redundant-run handling via `RedundantUsageRunSkipHandler`. The window checkpoint is only advanced after a successful aggregator drain, so a partial/failed run doesn't mark the window as covered.
- New `USAGE_STATS` source capability declaration on `FabricOneLakeSource`.
- Validator: `usage.include_usage_statistics=True` requires `sql_endpoint.enabled=True`, since `queryinsights` lives on the SQL Analytics Endpoint. Note: Fabric retains query insights for 30 days.

### DIFF
--- a/metadata-ingestion/docs/sources/fabric-onelake/README.md
+++ b/metadata-ingestion/docs/sources/fabric-onelake/README.md
@@ -2,7 +2,7 @@
 
 Microsoft Fabric OneLake is a storage and lakehouse platform. Learn more in the [official Microsoft Fabric OneLake documentation](https://learn.microsoft.com/fabric/onelake/onelake-overview).
 
-The DataHub integration for Microsoft Fabric OneLake covers workspace, lakehouse, and warehouse containers, table datasets with schema metadata, and view datasets with view definitions and view-to-table lineage parsed from the view SQL. It also captures stateful deletion detection.
+The DataHub integration for Microsoft Fabric OneLake covers workspace, lakehouse, and warehouse containers, table datasets with schema metadata, and view datasets with view definitions and view-to-table lineage parsed from the view SQL. It also extracts query usage statistics from the SQL Analytics Endpoint's `queryinsights` views, and captures stateful deletion detection.
 
 ## Concept Mapping
 
--- a/metadata-ingestion/docs/sources/fabric-onelake/fabric-onelake_post.md
+++ b/metadata-ingestion/docs/sources/fabric-onelake/fabric-onelake_post.md
@@ -266,6 +266,61 @@ source:
 3. **Schema**: Column metadata is reused from the same `INFORMATION_SCHEMA.COLUMNS` query that powers table schema extraction — no extra queries per view.
 4. **Lineage**: View definitions are passed to the SQL parsing aggregator to derive view → upstream table lineage. View URNs and upstream table URNs are resolved within the same workspace and item.
 
+#### Usage Statistics
+
+The connector extracts query usage statistics from each Lakehouse and Warehouse by reading the [`queryinsights.exec_requests_history`](https://learn.microsoft.com/en-us/fabric/data-warehouse/query-insights) view on the SQL Analytics Endpoint. Each captured query is parsed by the SQL parsing aggregator and emitted as:
+
+- `datasetUsageStatistics` aspects — query counts, distinct user counts, top users, top fields, and (when enabled) top SQL queries, bucketed by the configured window.
+- `operation` aspects — per-query operation events (insert, update, delete, etc.) when `usage.include_operational_stats` is enabled.
+
+##### Prerequisites
+
+Usage extraction reuses the SQL Analytics Endpoint connection used by [Schema Extraction](#schema-extraction). The same ODBC driver and permissions apply, plus:
+
+- `sql_endpoint.enabled` must be `true` — `queryinsights` views live on the SQL Analytics Endpoint, so the connector cannot read them otherwise. The configuration validator will reject `usage.include_usage_statistics=true` if the endpoint is not enabled.
+- The authenticated identity needs `SELECT` access on `queryinsights.exec_requests_history`. The view returns rows visible to the calling principal under Fabric's standard query-history visibility rules.
+- Microsoft Fabric retains query insights data for **30 days**. Older history cannot be backfilled, so configure `usage.start_time` accordingly.
+
+##### Configuration
+
+```yaml
+source:
+  type: fabric-onelake
+  config:
+    # Usage extraction is enabled by default. Set to false to skip query usage.
+    usage:
+      include_usage_statistics: true
+
+      # When true, the SQL filter excludes rows where status != 'Succeeded'
+      # (canceled / failed queries are skipped at the source).
+      skip_failed_queries: true
+
+      # Optional: emit per-query operation aspects in addition to aggregated
+      # datasetUsageStatistics. Defaults to true (inherited from BaseUsageConfig).
+      include_operational_stats: true
+
+      # Optional: include top SQL queries in the usage payload.
+      include_top_n_queries: true
+      top_n_queries: 10
+
+      # Optional: window the connector queries from queryinsights. Defaults to
+      # the standard BaseUsageConfig "last bucket" window. Fabric retains
+      # queryinsights for 30 days.
+      bucket_duration: DAY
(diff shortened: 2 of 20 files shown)


sgomezvillamor (approved):
Great implementation overall — clean extractor structure, correct `BaseUsageConfig` extension, solid stateful skip-handler semantics, and a good unit test suite. A few things to address before merge:

**🔴 Blocker: Missing `OPERATION_CAPTURE` capability**
Please add to `source.py`:
```python
@capability(
    SourceCapability.OPERATION_CAPTURE,
    "Optionally enabled via `usage.include_usage_statistics` and `usage.include_operational_stats`",
)
```
`SourceCapability.OPERATION_CAPTURE` was [very recently added in #17366](https://github.com/datahub-project/datahub/pull/17366) (merged after this PR opened) so it's natural this was missed. Also update `connector_registry/datahub.json` — the PR already added `USAGE_STATS` there.

**🟡 Stateful checkpoint advanced on total extraction failure**
When all items fail (SQL endpoint down, permissions change), each per-item exception is swallowed by `extract()`, the aggregator drain succeeds with 0 MCPs, and `update_state_on_success()` still advances the checkpoint — permanently marking that window as covered. Future runs skip it and the data is unrecoverable. `FabricUsageExtractor` should track `_had_item_failures` and skip `update_state_on_success()` if any per-item failure occurred.

**🟡 Pydantic mutable default**
`config.py`: `usage: FabricUsageConfig = Field(default=FabricUsageConfig(), ...)` — should use `default_factory=FabricUsageConfig` (the `extract_schema` and `sql_endpoint` fields immediately above already do this).

**🟡 `exc_info=True` missing in `stream_usage_history`'s `except SQLAlchemyError` block**
Stack trace is discarded for the primary expected failure mode, while the `except Exception` branch below retains it. Same pattern in `get_all_table_columns()` and `get_all_views()`.

**🟡 `report.failures` double-counted**
`stream_usage_history` increments `self.report.failures` before re-raising; `extract()`'s catch block then also reports the failure.

**❓ `query_hash` as deduplication key**
`SqlParsingAggregator` uses `observed.query_hash` over its own sqlglot semantic fingerprint when provided (`aggregator.py:906`). Fabric's `BINARY(8)` hash is syntactic; the aggregator fingerprint normalizes whitespace/case. The tradeoff at scale is unanalyzed — worth a comment and a follow-up ticket.

**❓ SQL-level username filter**
`user_email_pattern` already filters Python-side correctly. Is a TODO for pushing `login_name` filtering into the SQL query worthwhile for high-volume deployments?

**ℹ️ Suggestions**
- The aggregator drain uses `report_warning()` — `report_failure()` would be more appropriate there since it means all lineage/usage output is incomplete for the run. Note: `report_warning()` already calls `logger.warning()` internally, so no separate `logger.*` call is needed.
- `include_operational_stats: True` in the test config is never actually exercised since all tests use SELECT (which never produces operations). A DML test verifying an `operation` aspect is generated would close this gap.
- `generate_queries=False` is hardcoded. Consider exposing `include_queries: bool` (like Snowflake/ClickHouse) to allow query entity generation.
- `batch_size` in `stream_usage_history`: since pyodbc buffers the full result set client-side regardless, renaming to `fetch_chunk_size` with an explicit note that it doesn't bound memory would clarify intent.
- Please verify: sibling sub-configs in this file (`ExtractSchemaConfig`, `SqlEndpointConfig`) use `enabled` as the master toggle. Is `include_usage_statistics` intentional here or should it follow the same convention?

**📋 Test coverage gaps**
- No test for the `schema_client is None` guard in `_extract_item_usage`
- `skip_failed_queries` SQL filter (`AND status = 'Succeeded'`) is tested via mock passthrough only, not at the string-generation level

**Follow-up:** connector-tests should eventually cover usage/operational stats with actual SQL execution (similar to Snowflake/BigQuery smoke tests).

**Re the golden file:** no `operation` aspects for a SELECT query is correct (the aggregator explicitly skips operations for SELECTs), and no query entities is intentional (`generate_queries=False`). But as noted above, the DML path is currently untested.

---

aviraj-gour:
@sgomezvillamor Thanks for review!!

> 🔴 Blocker: Missing OPERATION_CAPTURE capability
> Please add to source.py:
> 
> @capability(
>     SourceCapability.OPERATION_CAPTURE,
>     "Optionally enabled via `usage.include_usage_statistics` and `usage.include_operational_stats`",
> )
> SourceCapability.OPERATION_CAPTURE was https://github.com/datahub-project/datahub/pull/17366 (merged after this PR opened) so it's natural this was missed. Also update connector_registry/datahub.json — the PR already added USAGE_STATS there.
> 

Added

> 🟡 Stateful checkpoint advanced on total extraction failure
> When all items fail (SQL endpoint down, permissions change), each per-item exception is swallowed by extract(), the aggregator drain succeeds with 0 MCPs, and update_state_on_success() still advances the checkpoint — permanently marking that window as covered. Future runs skip it and the data is unrecoverable. FabricUsageExtractor should track _had_item_failures and skip update_state_on_success() if any per-item failure occurred.

Fixed. Followed the framework's intended pattern (same as BigQuery / Snowflake / Redshift / Snaplogic) rather than a local _had_item_failures flag.

Change: In [usage.py:148-155], when extract()'s except block catches a per-item failure, it now calls:

if self.redundant_run_skip_handler:
    self.redundant_run_skip_handler.report_current_run_status(
        f"usage-extraction-{workspace_id}/{item_id}", False
    )

Updates only when there is no error.

> 🟡 Pydantic mutable default
> config.py: usage: FabricUsageConfig = Field(default=FabricUsageConfig(), ...) — should use default_factory=FabricUsageConfig (the extract_schema and sql_endpoint fields immediately above already do this).

Switched default=FabricUsageConfig() to default_factory=FabricUsageConfig, matching the pattern of extract_schema and sql_endpoint above.

> 🟡 exc_info=True missing in stream_usage_history's except SQLAlchemyError block
> Stack trace is discarded for the primary expected failure mode, while the except Exception branch below retains it. Same pattern in get_all_table_columns() and get_all_views().

Fixed: added exc_info=True to the except SQLAlchemyError branches

> 🟡 report.failures double-counted
> stream_usage_history increments self.report.failures before re-raising; extract()'s catch block then also reports the failure.
Fixed the double-counting by removing self.report.failures += 1

> ❓ query_hash as deduplication key
> SqlParsingAggregator uses observed.query_hash over its own sqlglot semantic fingerprint when provided (aggregator.py:906). Fabric's BINARY(8) hash is syntactic; the aggregator fingerprint normalizes whitespace/case. The tradeoff at scale is unanalyzed — worth a comment and a follow-up ticket.

dropped query_hash from the entire usage path rather than just stopping the aggregator override. The aggregator now derives its own sqlglot fingerprint as the dedup key (via parsed.query_fingerprint at [aggregator.py:906], matching the BigQuery and Snowflake convention

> ❓ SQL-level username filter user_email_pattern already filters Python-side correctly. Is a TODO for pushing login_name filtering into the SQL query worthwhile for high-volume deployments?

T-SQL has no native regex. Snowflake's clean rlike(user_name, pattern) pushdown isn't available here — SQL Server / Fabric Warehouse SQL Analytics Endpoint only supports LIKE (%, _, [], [^]). Translating arbitrary AllowDenyPattern regexes to LIKE is lossy. Not implementing for now.

Also implemented suggestion 1, 2,3 and 4
Not implemented 5 because matches other connector pattern for usage extraction.

PR #17286, opened 2026-05-02
perf: Lazy load mui-icons

benjiaming:
Dropping mui-icons from main bundle:
  | Metric | Minified | Gzipped |
  |---|---|---|
  | Before (single chunk) | 11.91 MB | 3.09 MB |
  | After main chunk | 8.24 MB | 2.45 MB |
  | After async MUI chunk | 3.69 MB | 0.64 MB |

  - Main chunk: -3.85 MB (30.84% reduction) 
  - New async chunk: +3.88 MB — the MUI icons, loaded on demand
  - Total delta: +24.61 KB (0.11%) — the small overhead from the async loading wrapper/chunk splitting

The 3.69 MB MUI icons chunk only loads when the user visits a page with domain icons or opens the icon picker.

Note: this should be a temporary fix before we move to phosphor icons.


ani-malgari (approved):
🚀

AdrianMachado (approved):
Makes sense

PR #17289, opened 2026-05-04
fix(redshift): deterministic dedup of OUTBOUND_DATASHARE PlatformResources using ingested_at timestamp

dinesh-verma-datahub:
### Problem

When a Redshift producer cluster is ingested more than once — for example after adding or changing `platform_instance` in the recipe — each run creates a **distinct** `OUTBOUND_DATASHARE` PlatformResource entity because the entity URN encodes `platform_instance`. The stale resource from the previous run is not automatically removed.

The consumer-side ingestion queries for matching resources by share name and producer namespace. When multiple resources exist, the old code returned the first result from the search index without any ordering guarantee. If the stale resource was returned first, the consumer generated upstream dataset URNs based on the old `platform_instance` and `env`, pointing at datasets that do not exist in DataHub — producing **ghost lineage nodes** in the UI.

### Fix

**1. Stamp `ingested_at` at emission time**

`OutboundSharePlatformResource` gains `ingested_at: Optional[datetime] = None`, stamped as `datetime.now(timezone.utc)` whenever the producer writes a resource. The field is `Optional` for backward compatibility with resources written before this change.

**2. Deterministic selection on the consumer side**

`find_upstream_share()` now applies a layered dedup strategy when multiple resources are found:

- **Primary**: Sort by `ingested_at` descending. The most recently ingested resource reflects the producer's current configuration. Legacy resources (no `ingested_at`) sort last via a `datetime.min` UTC sentinel. Naive datetimes are coerced to UTC to prevent a `TypeError` in the comparator.
- **Fallback** (all resources are legacy with no `ingested_at`): Prefer resources that have `platform_instance` set, as those come from newer ingestion runs that had platform instance configured.
- **Tie**: Pick the first candidate and surface a `report.warning` prompting operators to clean up stale duplicates.
- **All resources unparseable**: Surface a `report.warning` so missing lineage is never silent.

**3. Fix datetime serialization in `SerializedResourceValue`**

Changed `model_dump()` → `model_dump(mode="json")` so Pydantic v2 converts `datetime` (and other non-JSON-native types) to JSON-serializable forms before `json.dumps`. Without this, emitting any `BaseModel` containing a `datetime` field raised `TypeError` at write time. Existing connectors that use only primitive fields are unaffected.

### Behavior for Existing Customers

| Scenario | Before | After |
|---|---|---|
| Single PlatformResource per share (normal) | Works | Works — no change |
| Legacy resources only (no `ingested_at`) | Non-deterministic pick | Prefers resource with `platform_instance`; falls back gracefully |
| One new resource + stale legacy resource | Non-deterministic pick | New (timestamped) resource always wins |
| Multiple timestamped resources | Non-deterministic | Most recently ingested wins; warning fires until duplicates are cleaned up |

### Connector Tests Updated - https://github.com/acryldata/connector-tests/pull/788


metadata-ingestion/src/datahub/ingestion/source/redshift/datashares.py
```diff
@@ -23,13 +25,17 @@
...
+
 
 class OutboundSharePlatformResource(BaseModel):
     namespace: str
     platform_instance: Optional[str] = None
     env: str
     source_database: str
     share_name: str
+    # Optional for backward-compat with resources written before this field was added.
+    ingested_at: Optional[datetime] = None
```
treff7es, line 38: _aware() exists specifically to prevent a TypeError at sort time when a naive datetime sneaks in. A regression here would be a hard crash, killing all lineage emission. One test
  exercising the naive→UTC coercion would lock the invariant.

Push the invariant into the type itself with a @field_validator on OutboundSharePlatformResource.ingested_at.
```
@field_validator("ingested_at")
  @classmethod
  def _ensure_aware(cls, v: Optional[datetime]) -> Optional[datetime]:
      if v is None or v.tzinfo is not None:
          return v
      return v.replace(tzinfo=timezone.utc)
```

metadata-ingestion/src/datahub/ingestion/source/redshift/datashares.py
```diff
@@ -165,24 +172,92 @@ def find_upstream_share(
...
+                        title="Multiple matching outbound datashare PlatformResources",
+                        message=(
+                            "Found more than one OUTBOUND_DATASHARE PlatformResource "
+                            "matching this inbound share. Picking the most recently "
+                            "ingested; consider cleaning up duplicates."
+                        ),
+                        context=share.get_description(),
+                    )
+
+                return candidates[0]
```
treff7es: Something like this would be way simpler:

```
_EPOCH = datetime.min.replace(tzinfo=timezone.utc)

def _aware(ts: Optional[datetime]) -> datetime:
      if ts is None:
          return _EPOCH
      return ts if ts.tzinfo else ts.replace(tzinfo=timezone.utc)

  def _priority(p: OutboundSharePlatformResource) -> tuple:
      return (
          p.ingested_at is not None,        # timestamped > legacy
          _aware(p.ingested_at),             # newer > older
          p.platform_instance is not None,   # has platform_instance > not
          p.namespace,                       # stable tiebreaker
      )

  best = max(parsed, key=_priority)

  if len(parsed) > 1:
      self.report.warning(
          title="Multiple matching outbound datashare PlatformResources",
          message=(
              f"Found {len(parsed)} matching resources; picked "
              f"ingested_at={best.ingested_at!r}, "
              f"platform_instance={best.platform_instance!r}. "
              f"Clean up duplicates."
          ),
          context=share.get_description(),
      )
  return best

```

metadata-ingestion/src/datahub/ingestion/source/redshift/datashares.py
dinesh-verma-datahub: added this

metadata-ingestion/src/datahub/ingestion/source/redshift/datashares.py
dinesh-verma-datahub: yeah agreed, replaced with this.

treff7es (approved):

treff7es (approved):

PR #17291, opened 2026-05-04
fix(ui): update required semantic tokens

purnimagarg1:
**Linear ticket:**
https://linear.app/acryl-data/issue/CAT-1786/update-required-tokens-in-oss-and-make-ensure-they-look-okay


annadoesdesign (approved):
approved :)

PR #17292, opened 2026-05-04
feat(cli): add instance2instance migration and extend dataplatform2instance to charts, dashboards, and dataflows

rospe:
## Summary

Add a new `datahub migrate instance2instance` CLI command for migrating entities from one platform instance to another (e.g. consolidating Snowflake instances). Also extends the existing `dataplatform2instance` command to handle charts, dashboards, and dataflows in addition to datasets and containers.

## Motivation

Organizations that need to rename or consolidate platform instances (e.g. merging `sf_a1`, `sf_a2`, `sf_a3` into `sf_shared`) have no built-in way to do this. The existing `dataplatform2instance` only handles the no-instance → instance case and only migrates datasets and containers.

### New command: `instance2instance`

```console
datahub migrate instance2instance \
  --platform snowflake \
  --old-instance old_inst \
  --new-instance new_inst \
  --on-conflict patch \
  --dry-run
```

- Migrates entities from one platform instance to another
- Rewrites URNs by replacing the old instance prefix with the new one
- Updates all incoming relationships (lineage, dashboard refs, etc.)
- **Merge support**: when target URN already exists, merges metadata using the DataHub Patch API for additive aspects (ownership, tags, terms, lineage)
- **Conflict resolution**: `--on-conflict=patch|overwrite|prompt` for non-additive aspects (schema, properties)
- **Error resilience**: `--skip-on-error` to continue past problematic entities
- Supports `--dry-run`, `--keep`, `--hard`, `--force` flags

### Extended `dataplatform2instance`

- Now migrates **charts, dashboards, and dataflows** in addition to datasets and containers
- Added `--on-conflict` and `--skip-on-error` flags (same as instance2instance)
- Env filter correctly skipped for charts/dashboards/dataflows (they lack env/origin fields in ElasticSearch)

### Refactored shared migration logic

- Extracted `_migrate_single_entity` and `_migrate_entities` as shared orchestrators used by both commands
- Unified URN builder via `make_urn_builder(entity_type, new_instance, old_instance)` supporting all entity types
- Shared relationship update, aspect cloning, and merge logic in `migration_utils.py`

### Merge strategy (instance2instance)

When the target entity already exists:

| Aspect category | Strategy |
|---|---|
| **Additive** (ownership, tags, terms, lineage) | Merge via Patch API — deduplicated |
| **Mixed** (datasetProperties, customProperties) | Merge key-by-key, conflict strategy for scalars |
| **Non-additive** (schema, viewProperties) | Apply `--on-conflict` strategy |
| **Migration metadata** (dataPlatformInstance, container, status) | Always overwrite |

### Known limitations

- Merge path uses `DatasetPatchBuilder` — merge only works for dataset entities. Charts/dashboards/dataflows are cloned (not merged) when the target already exists.
- Container migration does not support `--on-conflict` or `--skip-on-error` (containers use key-based GUID recomputation which doesn'"'"'t have the same overlap problem).

## Tests

58 unit tests covering:
- URN builders for all entity types (dataset, chart, dashboard, dataFlow) in both p2i and i2i modes
- Instance prefix replacement (simple, complex, dotted instance names)
- Conflict resolution (patch/overwrite strategies for additive, mixed, and non-additive aspects)
- Merge logic (ownership, tags, terms, lineage, customProperties, description)
- Skip-on-error behavior (continues past failures, tracks errors in report)
- Dry-run correctness (clone_aspect always yields MCPs for reporting)
- Single entity migration (emit, delete, merge paths)
- Edge cases (unsupported entity types, empty aspects)

## Documentation

Updated `docs/cli.md` with:
- Full documentation for both `dataplatform2instance` and `instance2instance` commands
- Flag descriptions, example invocations, merge strategy explanation
- Best practices workflow (dry-run → migrate with --keep → verify → clean up)


rospe:
Hi @sgomezvillamor I just wanted to check if there's anything I can do to speed things up? Thanks for the review!

sgomezvillamor:
Great work on this PR! The architecture is solid — the `ConflictStrategy` enum, the aspect-category separation (`ADDITIVE_ASPECTS`, `NON_ADDITIVE_ASPECTS`, `MIXED_ASPECTS`), and the alignment between `dataplatform2instance` and `instance2instance` are all clean and well-thought-out. The test coverage is thorough for most paths. There are two blockers to fix before merging, plus a few warnings worth addressing.

---

## 🔴 Blockers

**1. `DatasetPatchBuilder` used for non-dataset URNs in merge path**

`merge_additive_aspects` and `merge_mixed_aspects` in `migration_utils.py` both unconditionally instantiate `DatasetPatchBuilder(dst_urn)`, even when `dst_urn` is a chart, dashboard, or dataFlow URN. `DatasetPatchBuilder` validates that its URN is a dataset — passing a chart URN will either silently produce a broken patch or raise at runtime.

Fix: in `_migrate_single_entity`, only enter the merge path when `_get_type_from_urn(src_entity_urn) == "dataset"`. For other entity types, fall back to overwrite (or raise a clear error).

**2. `dataJob` entities not migrated**

`dataJob` URNs embed a reference to their parent `dataFlow` URN (e.g. `urn:li:dataJob:(urn:li:dataFlow:(airflow,instance.myDag,PROD),myTask)`). When a flow URN changes because of an instance migration, the corresponding dataJob URNs become orphaned — they still point to the old flow URN. `make_data_job_urn` already supports `platform_instance`, so the building blocks are there. Either add `dataJob` to the migration scope (migrating jobs alongside their parent flows) or explicitly document why it's excluded and what users must do manually.

---

## 🟡 Warnings

**3. Greedy regex for `dataFlow` URN parsing**

`_DATAFLOW_RE = re.compile(r"urn:li:dataFlow:\((.*),(.*),(.*)\)")` is greedy — if an orchestrator or flow ID contains a comma, the capture groups will split incorrectly. The codebase already uses `DataFlowUrn.from_string(urn_str)` (see `mce_builder.py:337`); use that instead and access `.orchestrator`, `.flow_id`, `.cluster` directly.

**4. Container filter skips un-migrated containers silently**

In the p2i container path, `props.get("instance") == env` will be `None == "PROD"` for any container that has never been migrated, silently skipping it. This is a pre-existing edge case surfaced by the refactor — worth at least a warning log.

**5. `replace_instance_prefix` silently creates wrong URNs**

When the old instance prefix isn't found, the function prepends `new_instance.` anyway. A wrong URN written silently is harder to debug than a noisy warning. Add a `logger.warning(...)` when the prefix isn't found.

**6. Python 3.9+ type annotation**

`merge_mixed_aspects` and `merge_entity` use `tuple[int, int]` as a return annotation. The project targets Python 3.8+; use `Tuple[int, int]` from `typing` instead.

---

## 📝 Docs & minor items

**Docs** (`docs/cli.md`): missing a "when to use which command" comparison (p2i vs i2i), no explanation that the default `--on-conflict` differs between the two commands, the merge limitation (datasets only) is undocumented, and `dataJob` gap is not mentioned.

**`_ENV_ENTITY_TYPES` duplication**: the set `{"dataset"}` is defined identically in both CLI handlers. Define it once as a module-level constant in `migration_utils.py`.

**Test gaps**: no test covers a non-dataset entity type entering the merge path (exactly the blocker scenario), and the relationship-update loop isn't tested with non-empty relationships.

**`--entity-types` flag (optional)**: the hardcoded `["dataset", "chart", "dashboard", "dataFlow"]` list is a reasonable candidate to expose as an optional CLI argument for users who want incremental or partial migrations.

---

Address the two blockers and this is ready to merge. Thanks again for the work — the overall structure is great and will make future migrations much more robust. Happy to discuss any of the above!

---

sgomezvillamor (approved):
Both blockers and all the main warnings are addressed — great turnaround, this is looking solid. Two small things before merge:

**1. Use `guess_entity_type` instead of raw URN splitting**

`datahub.utilities.urns.urn` already exports `guess_entity_type(urn)` which does the same `urn.split(":")[2]` with an assertion. There are three places to clean up:
- `migration_utils.py` (`merge_entity`): `entity_type = dst_urn.split(":")[2]` → import and use `guess_entity_type`
- `migrate.py`: `_get_type_from_urn` is a duplicate of the same utility — replace it and its two call sites with `guess_entity_type` (it's already imported from the same module as `Urn`)

**2. `_warn_dataflow_datajob_coupling` ignores `--force` and `--dry-run`**

`click.confirm()` is called unconditionally, so scripted/automated runs with `--entity-types=dataFlow` (intentionally deferring job migration) will get an interactive prompt and hang. The fix is to skip the confirm when `force=True`, consistent with how the rest of both commands handle the `--force` flag.

Neither is a blocker — approving now so this can move forward once addressed.

---

sgomezvillamor:
Hey @rospe — this looks ready to merge. Are you planning any further updates?

rospe:
@sgomezvillamor you can go ahead and merge - i don't see this option myself.

PR #17294, opened 2026-05-04
fix(ingest/kafka-connect,sql-parsing): surface CLL failures and recover from partial bulk-fetch misses

treff7es:
## Summary

The kafka-connect MongoDB source connector and the shared `SchemaResolver` had two diagnostic blind spots that made per-collection lineage drops hard to investigate.

### 1. Silent CLL failures in `_extract_fine_grained_lineage`

Any exception inside `_extract_fine_grained_lineage` was swallowed at `logger.debug` level — no `report.warning`, no aggregation in the ingestion report. Per-collection schema-resolution failures (URN encoding edge cases, transform interactions, etc.) left no signal in the report.

This PR converts the broad `except Exception` into a structured `self.report.warning(...)` that names the connector, source/target, and exception type. Behavior is unchanged: `_extract_fine_grained_lineage` still returns `None` and the caller still appends asset-level lineage as before — only the visibility of the failure changes.

### 2. `SchemaResolverProvider` disabled per-URN fallback

`SchemaResolverProvider.get()` constructed `SchemaResolver` instances with `graph=None`, which disables the per-URN `get_entities()` fallback inside `SchemaResolver.resolve_table()`. When the bulk pre-population missed any URN — transient GMS error mid-batch, pagination edge case, filter mismatch — those entries were *permanently* absent from the cache for the rest of the ingestion run with no recovery path.

This PR passes `self._graph` through to the constructed resolver so the targeted per-URN fallback works again. The bulk-fetch fast path is preserved: cache hits short-circuit before any per-URN call is made, so populated URNs incur zero extra I/O.

### Tests

Adds two regression tests pinning the contract that asset-level lineage is emitted regardless of CLL outcome:

- `test_mongo_source_asset_lineage_preserved_when_schema_missing` — schema resolver returns `(urn, None)`; asset lineage is still emitted, `fine_grained_lineages is None`.
- `test_mongo_source_asset_lineage_preserved_when_resolver_raises` — schema resolver raises an exception; asset lineage is still emitted, a structured warning is reported with the failing source in context.

## Test plan

- [x] New unit tests pass.
- [x] Full `tests/unit/test_kafka_connect.py` (162) and `tests/unit/test_kafka_connect_schema_resolver.py` + `tests/unit/sdk/test_schema_resolver_cache.py` (29 combined) all pass.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/kafka_connect/common.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/kafka_connect/common.py
@@ -999,9 +999,21 @@ class BaseConnector:
                 return fine_grained_lineages
 
         except Exception as e:
-            logger.debug(
-                f"Failed to extract fine-grained lineage for "
-                f"{source_dataset} → {target_dataset}: {e}"
+            # Surface as a structured warning so per-collection failures show up in the
+            # ingestion report. Asset-level lineage is unaffected (the caller still
+            # appends the KafkaConnectLineage entry; only fine_grained_lineages is None).
+            self.report.warning(
+                title="Failed to extract column-level lineage",
+                message=(
+                    "An error occurred while resolving schema for column-level "
+                    "lineage. Asset-level lineage is unaffected."
+                ),
+                context=(
+                    f"connector={self.connector_manifest.name}, "
+                    f"source_platform={source_platform}, source={source_dataset}, "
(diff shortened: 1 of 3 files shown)


treff7es:
Closing for now — the diagnostics/recovery changes here may not be needed. Can reopen if silent CLL failures or partial schema bulk-fetch misses resurface.

PR #17298, opened 2026-05-04
feat(ingest/glue): add extract_column_parameters to surface column-level Parameters in Properties tab

JohnRTurner:
## Summary

- **Glue ingestion**: adds `extract_column_parameters` config option (default `false`) that ingests AWS Glue column-level `Parameters` as `jsonProps` on each `SchemaField`
- **Frontend**: renders `jsonProps` in the column Properties tab alongside structured properties, with filter-bar search support
- **Feature flag**: `showColumnJsonProperties` (default `false`) lets operators opt in to displaying jsonProps without affecting other sources (Salesforce, Snowflake, Avro, JSON Schema) that also populate the field
- **Bug fix**: wires pre-fetched `fieldProperties` directly into `useStructuredProperties` to eliminate Apollo cache staleness — structured properties applied after page load now appear immediately without requiring a full page reload

## Motivation

AWS Glue stores rich column-level metadata in the `Parameters` map (e.g. Iceberg field IDs, nullability flags). This data was ingested but not surfaced in the UI. The `extract_column_parameters` flag makes it visible as read-only rows in the column Properties tab, intermingled with structured properties.

The `showColumnJsonProperties` feature flag is added because `jsonProps` is populated by multiple sources for different purposes — gating display behind a flag means operators can enable it intentionally rather than having internal schema metadata appear by default.

## Test plan

- [ ] Set `extract_column_parameters: true` in a Glue recipe and re-ingest; verify column `Parameters` appear as rows in the Properties tab when `showColumnJsonProperties=true`
- [ ] Verify filter search works across structured properties and jsonProps rows simultaneously
- [ ] Verify `showColumnJsonProperties=false` (default) hides jsonProps rows entirely
- [ ] Verify structured properties on schema fields display correctly without requiring a page reload after being applied (Apollo cache fix)
- [ ] Unit tests: `test_glue_source.py` covers golden-file output; `parseJsonPropsToRows.test.ts` covers frontend parsing

### DIFF
--- a/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/config/AppConfigResolver.java
+++ b/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/config/AppConfigResolver.java
@@ -288,6 +288,7 @@ public class AppConfigResolver implements DataFetcher<CompletableFuture<AppConfi
             .setMultipleDataProductsPerAsset(_featureFlags.isMultipleDataProductsPerAsset())
             .setGlossaryBasedPoliciesEnabled(_featureFlags.isGlossaryBasedPoliciesEnabled())
             .setShowColumnJsonProperties(_featureFlags.isShowColumnJsonProperties())
+            .setHideNullableColumnJsonProperties(_featureFlags.isHideNullableColumnJsonProperties())
             .build();
 
     appConfig.setFeatureFlags(featureFlagsConfig);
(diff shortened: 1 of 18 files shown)


JohnRTurner:
## Additional change: \`hideNullableColumnJsonProperties\` flag

Added a second feature flag as part of this PR: **\`hideNullableColumnJsonProperties\`** (default \`false\`).

**Motivation**: Glue marks all columns as nullable by default, so the \`nullAllowed\` property appears on every column's Properties tab — it carries no useful signal and adds visual noise.

When set to \`true\`, this flag filters the \`nullAllowed\` row out of the Properties tab while leaving all other column JSON properties visible. It operates only when \`showColumnJsonProperties\` is also \`true\`; if \`showColumnJsonProperties\` is \`false\`, no JSON properties are shown regardless.

**How to enable:**
\`\`\`yaml
# application.yaml / feature flags
feature-flags:
  showColumnJsonProperties: true
  hideNullableColumnJsonProperties: true  # suppress the noisy nullAllowed row
\`\`\`

treff7es:
The python part looks good to me but as it contains mostly UI chanes I would let on UI reviewer to approve or not

JohnRTurner:
Closing in favour of #17325 which uses structured properties on schemaField instead of jsonProps + feature flags — no frontend/backend changes required.

PR #17299, opened 2026-05-05
perf(metadata-io): coalesce same-(urn, aspect) MCLs in V2 update strategy

craig-rueda:
## Summary

Within a batched MCL flush in the MAE consumer, the V2 update-indices strategy was emitting one Elasticsearch upsert per MCL even when many MCLs in the batch targeted the same (urn, aspect). V3 already coalesces (`UpdateIndicesV3Strategy.buildV3SearchDocument`); this brings V2 to parity.

For each (urn, aspect) update group within a batch:
- **Timeseries aspects** are still processed per event (every datapoint matters).
- **Non-timeseries aspects** are coalesced to last-write-wins — only the latest event's `transformAspect` runs and only one `upsertDocument` is issued.

Two correctness details that fall out of coalescing:
- **Diff baseline**: the existing `searchDiffMode` short-circuit compares the survivor's transformed search doc against `previousRecordTemplate`. Under coalescing, that previous template is the in-batch intermediate state, not what ES has. The new path passes the **oldest predecessor's** `previousRecordTemplate` as the explicit baseline so a no-op tail (e.g. `[A→B, B→B]`) cannot incorrectly suppress an upsert that the pre-coalesce path would have done.
- **runId fan-out**: the per-event `appendRunId` call previously fired once per MCL. After coalescing it only fires for the survivor, which would lose older runs from the doc's `runId` array and break rollback-by-run. The new `appendCoalescedRunIds` helper appends each distinct predecessor runId explicitly.

The behavior is gated by a flag for safe rollback:
- `elasticsearch.entityIndex.v2.coalesceBatchUpdates` — default `true`.
- Env var: `ELASTICSEARCH_ENTITY_INDEX_V2_COALESCE_BATCH_UPDATES=false` reverts to the legacy per-event path verbatim.

## Files changed

- `metadata-io/.../service/UpdateIndicesUtil.java` — new `groupUpdatesByAspect` helper.
- `metadata-io/.../service/UpdateIndicesV2Strategy.java` — flag field + 8-arg ctor (7-arg delegates with `true`); gated coalesce in `processBatch`; new `processAspectGroup` and `appendCoalescedRunIds`; split `updateSearchIndicesForEvent` to accept an explicit baseline.
- `metadata-io/.../service/UpdateIndicesV2StrategyTest.java` — 6 new tests (5 coalesce + 1 flag-disabled).
- `metadata-service/configuration/.../application.yaml` — new flag declaration.
- `metadata-service/factories/.../UpdateIndicesStrategyFactory.java` — wires `@Value`-injected flag into the V2 ctor.

## Checklist

- [x] PR conforms to the [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md), particularly [PR Title Format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#pr-title-format)
- [x] Tests added for the changes (6 new tests in `UpdateIndicesV2StrategyTest`; all `UpdateIndices*` tests pass)
- [ ] Linked issues — n/a
- [ ] Docs — n/a (operator-facing flag is self-documenting via the YAML default)
- [ ] Breaking change entry in `updating-datahub.md` — n/a (default behavior is the safer/correct one with the runId fix; legacy behavior preserved behind the flag)

### DIFF
--- a/metadata-io/src/main/java/com/linkedin/metadata/service/UpdateIndicesV2Strategy.java
+++ b/metadata-io/src/main/java/com/linkedin/metadata/service/UpdateIndicesV2Strategy.java
@@ -244,14 +244,22 @@ public class UpdateIndicesV2Strategy implements UpdateIndicesStrategy {
     }
 
     MCLItem survivor = aspectEvents.get(aspectEvents.size() - 1);
-    if (structuredPropertiesHookEnabled) {
-      updateIndexMappings(opContext, survivor);
-    }
     // Use the oldest predecessor's previousRecordTemplate as the diff baseline, since that is
     // what ES actually had before the batch began. Otherwise the diff would compare against the
     // intermediate in-batch state and incorrectly skip the upsert when a no-op tail follows real
-    // changes earlier in the group.
+    // changes earlier in the group. This baseline is also the right one for the structured-
+    // property mapping diff: any entityType added by an earlier MCL in the group must still be
+    // applied to ES even though the survivor's own previousRecordTemplate already contains it.
     RecordTemplate baseline = aspectEvents.get(0).getPreviousRecordTemplate();
+    if (structuredPropertiesHookEnabled) {
+      updateIndexMappings(
+          opContext,
+          survivor.getUrn(),
+          survivor.getEntitySpec(),
+          survivor.getAspectSpec(),
+          survivor.getRecordTemplate(),
+          baseline);
+    }
     updateSearchIndicesForEvent(opContext, survivor, baseline);
     updateTimeseriesFieldsForEvent(opContext, survivor);
     appendCoalescedRunIds(opContext, survivor, aspectEvents);
--- a/metadata-io/src/main/java/com/linkedin/metadata/service/UpdateIndicesV2Strategy.java
+++ b/metadata-io/src/main/java/com/linkedin/metadata/service/UpdateIndicesV2Strategy.java
@@ -35,6 +35,8 @@ import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.util.Collection;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -72,6 +74,7 @@ public class UpdateIndicesV2Strategy implements UpdateIndicesStrategy {
   private final TimeseriesAspectService timeseriesAspectService;
   private final String idHashAlgo;
   private final V2MappingsBuilder mappingsBuilder;
+  private final boolean coalesceBatchUpdates;
 
   // Semantic search configuration (optional - null if semantic search not configured)
   @Nullable private final SemanticSearchConfiguration semanticSearchConfig;
@@ -99,6 +102,32 @@ public class UpdateIndicesV2Strategy implements UpdateIndicesStrategy {
       @Nonnull String idHashAlgo,
       @Nullable SemanticSearchConfiguration semanticSearchConfig,
       @Nonnull IndexConvention indexConvention) {
+    this(
+        v2Config,
+        elasticSearchService,
+        searchDocumentTransformer,
+        timeseriesAspectService,
+        idHashAlgo,
+        semanticSearchConfig,
+        indexConvention,
+        true);
+  }
+
+  /**
+   * Same as the 7-arg constructor, but allows toggling per-(urn, aspect) coalescing of batched
+   * updates. When {@code coalesceBatchUpdates} is true (default), N MCLs targeting the same (urn,
+   * aspect) within a batch produce a single ES upsert (last-write-wins, with predecessor runIds
+   * still appended). When false, the legacy per-event behavior is preserved as a rollback.
+   */
+  public UpdateIndicesV2Strategy(
+      @Nonnull EntityIndexVersionConfiguration v2Config,
+      @Nonnull ElasticSearchService elasticSearchService,
+      @Nonnull SearchDocumentTransformer searchDocumentTransformer,
+      @Nonnull TimeseriesAspectService timeseriesAspectService,
+      @Nonnull String idHashAlgo,
+      @Nullable SemanticSearchConfiguration semanticSearchConfig,
+      @Nonnull IndexConvention indexConvention,
+      boolean coalesceBatchUpdates) {
     this.v2Config = v2Config;
     this.elasticSearchService = elasticSearchService;
     this.searchDocumentTransformer = searchDocumentTransformer;
@@ -106,6 +135,7 @@ public class UpdateIndicesV2Strategy implements UpdateIndicesStrategy {
     this.idHashAlgo = idHashAlgo;
     this.semanticSearchConfig = semanticSearchConfig;
     this.indexConvention = indexConvention;
+    this.coalesceBatchUpdates = coalesceBatchUpdates;
     this.mappingsBuilder =
         new V2MappingsBuilder(
             com.linkedin.metadata.config.search.EntityIndexConfiguration.builder()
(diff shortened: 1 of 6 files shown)


alexsku:
## PR Summary

**Type**: perf + behavior change (gated)
**Size**: +470 / −11 across 5 files

Within `UpdateIndicesV2Strategy.processBatch`, batches with multiple MCLs targeting the same (urn, aspect) used to emit one ES upsert per MCL. This PR coalesces non-timeseries aspects to last-write-wins so a single upsert is emitted; timeseries aspects remain per-event. Two correctness adjustments come with the coalesce: the diff baseline anchors to the oldest predecessor's `previousRecordTemplate` (so a no-op tail can't suppress an upsert that pre-coalesce would have done), and predecessor runIds are appended explicitly so rollback-by-run continues to find every URN touched in coalesced runs. Gated by `elasticsearch.entityIndex.v2.coalesceBatchUpdates` (default `true`).

**Key invariants:**
1. After processing a batch, the V2 ES doc must reflect the survivor's transformed-aspect state — same as if all MCLs had been applied per-event.
2. Every distinct runId carried by any MCL in a coalesced group must be searchable on the URN's doc (so rollback-by-run finds it).
3. Timeseries aspects are still processed per event (every datapoint matters).
4. With `coalesceBatchUpdates=false`, behavior must be identical to the pre-PR per-event path.
5. Structured-property index mapping updates must apply for any new entity type a structured-property definition gains across the batch.

---

## Risk Assessment

| | |
|---|---|
| **Risk** | Medium |
| **Blast radius** | All MAE-driven V2 entity-index writes; rollback-by-run; structured-property mapping propagation |
| **Rollback** | Safe — env flag flips back to legacy per-event path verbatim |
| **Rollout** | Default-on is reasonable since the legacy path had its own runId-loss bug, but the structured-property regression below is worth fixing first or rolling out behind `coalesceBatchUpdates=false` until fixed |
| **CI** | **Failing — `build-metadata-io` fails on `PropertiesCollectorConfigurationTest` (see Blocking Issues)** |

---

## Blocking Issues

### [BLOCKER] New `coalesceBatchUpdates` property is not classified in `PropertiesCollectorConfigurationTest`
**Category**: tests / operability | **Confidence**: high
**Location**: `metadata-io/src/test/java/com/linkedin/metadata/system_info/collectors/PropertiesCollectorConfigurationTest.java:971` (sibling `elasticsearch.entityIndex.v2.cleanup` is listed; the new key is missing)

This is the actual `build-metadata-io` CI failure. The repo's `AGENTS.md` calls this test out as a mandatory security guardrail: every new YAML property must be explicitly classified as sensitive (redacted) or non-sensitive. The test's failure message is explicit:

```
Found unclassified configuration properties. Add them to NON_SENSITIVE_PROPERTIES:
  [elasticsearch.entityIndex.v2.coalesceBatchUpdates (not redacted, …)]
```

**Suggested fix**: Add `"elasticsearch.entityIndex.v2.coalesceBatchUpdates"` to `NON_SENSITIVE_PROPERTIES` in `PropertiesCollectorConfigurationTest`, next to the existing `"elasticsearch.entityIndex.v2.cleanup"` entry on line 971. It's a boolean toggle, not credentials — non-sensitive is the right bucket.

---

## High-Priority Issues

### [HIGH] Coalesced structured-property mapping update can miss entity types added by non-survivor MCLs
**Location**: `metadata-io/src/main/java/com/linkedin/metadata/service/UpdateIndicesV2Strategy.java:247-249` — `processAspectGroup()`

When `structuredPropertiesHookEnabled=true`, the coalesced path calls `updateIndexMappings(opContext, survivor)` only for the survivor. That helper computes the entity-type diff as `survivor.getRecordTemplate().entityTypes − survivor.getPreviousRecordTemplate().entityTypes`, which is the *intermediate-to-final* diff, not the *pre-batch-to-final* diff.

Concrete failure scenario — two MCLs in the same batch on `urn:li:structuredProperty:foo`:
- MCL 1: prev `{entityTypes:[]}` → new `{entityTypes:[X]}`
- MCL 2 (survivor): prev `{entityTypes:[X]}` → new `{entityTypes:[X, Y]}`

Pre-PR per-event: applies mappings for `[X]`, then `[Y]`. ES has both.
Post-PR coalesced: only survivor fires; oldEntityTypes = `[X]`; diff = `[Y]`. ES gets `[Y]` mappings only — `[X]`'s mapping is **never applied** in this batch. Subsequent queries scoped to entity type X won't see this property's field.

This is also inconsistent with `UpdateIndicesV3Strategy.processUrnBatch` (`UpdateIndicesV3Strategy.java:257-269`), which iterates *all* events to apply mappings even though the rest of V3 is coalesced.

The same `aspectEvents.get(0).getPreviousRecordTemplate()` baseline that the search path uses (line 254) is the correct baseline here too.

**Suggested fix** (preferred — single coalesced call):
```java
if (structuredPropertiesHookEnabled) {
  updateIndexMappings(
      opContext,
      survivor.getUrn(),
      survivor.getEntitySpec(),
      survivor.getAspectSpec(),
      survivor.getRecordTemplate(),
      aspectEvents.get(0).getPreviousRecordTemplate());
}
```

Alternative — match V3 by iterating all `aspectEvents` and calling `updateIndexMappings(opContext, event)` for each. Slightly wasteful but a literal preservation of pre-PR behavior.

**Suggested test**: Add a `processBatch` test that creates two `STRUCTURED_PROPERTY_DEFINITION` MCLs for the same property URN in one batch (entityTypes `[]→[X]`, then `[X]→[X,Y]`) with `structuredPropertiesHookEnabled=true`, and verifies `buildReindexConfigsWithNewStructProp` is invoked with a definition whose entityTypes contain **both** X and Y (or, with the iterate-all variant, that it is invoked twice — once for X, once for Y).

---

## Other Issues

- **[LOW / edge-case]** `UpdateIndicesV2Strategy.java:255-257` — `appendCoalescedRunIds` runs unconditionally even if `updateSearchIndicesForEvent` short-circuited because `transformAspect` returned empty. Pre-PR, an event whose transform yielded empty did not append its runId. Post-PR, predecessor runIds are still tagged even when the survivor's transform was empty (and thus no upsert happened). Tiny semantic drift — mostly affects aspects with no searchable fields. Fix: gate the call on whether the survivor's upsert actually fired (e.g., by having `updateSearchIndicesForEvent` return a boolean), or accept the drift and document it.
- **[LOW / dx]** `UpdateIndicesStrategyFactory.java:40-41` — `coalesceBatchUpdates` is plumbed as a separate constructor arg rather than added as a field on `EntityIndexVersionConfiguration` (where `enabled` and `cleanup` already live). Minor — the existing factory already reads `cleanup` via `@Value` and rebuilds the config too, so this is consistent with what's there. Worth tidying when the next v2 flag lands.
- **[LOW / dx]** `UpdateIndicesV2StrategyTest.java:921` — `testProcessBatch_CoalesceUsesOldestPredecessorAsDiffBaseline` constructs the events but never verifies that `upsertDocument` was called the expected number of times when `T(O3) == T(O0)` (the actual no-op-tail-following-real-changes case the comment in the implementation calls out). The existing assertion only verifies that `transformAspect` ran for the baseline, not that the upsert was/wasn't suppressed. Consider adding a stronger assertion or a sibling test where the survivor's transformed doc equals the baseline's to assert the upsert is suppressed correctly.

---

## What's Missing

- The classification entry in `PropertiesCollectorConfigurationTest` (the BLOCKER above).
- A test for the structured-property mapping path under coalescing (the HIGH above).
- No entry in `docs/how/updating-datahub.md`. Defensible — the default behavior is the new (more correct) one, the operator-facing flag is self-documenting via the YAML, and the PR description argues this is not a breaking change. Worth confirming the team's expectation since this changes observable ES write counts and runId-array contents under heavy batching.

---

## Test Plan

| Invariant | Existing coverage |
|---|---|
| 1. Survivor upsert reflects final state | `testProcessBatch_CoalesceSameUrnSameAspect_OnlyLatestUpserted` ✓ |
| 2. Predecessor runIds tagged on URN | `testProcessBatch_CoalesceAppendsRunIdsFromPredecessors` ✓ |
| 3. Timeseries not coalesced | `testProcessBatch_TimeseriesNotCoalesced` ✓ |
| 4. Flag-disabled = legacy behavior | `testProcessBatch_FlagDisabled_PreservesPerEventBehavior` ✓ |
| 5. Structured-property mappings cover all entity types | **Not covered — HIGH issue above** |
| Diff baseline from oldest predecessor | `testProcessBatch_CoalesceUsesOldestPredecessorAsDiffBaseline` ✓ (could be stronger; see Other Issues) |

Worth adding: a test where survivor's `transformAspect` returns empty but predecessors have runIds, to lock in or fix the LOW drift above.

---

## Questions for Author

1. **Structured-property mapping under coalesce** — was the `updateIndexMappings(survivor)` call (vs iterating per-event like V3, or using the oldest-predecessor baseline like the search path does) intentional? It seems to drop intermediate entity-type additions.
2. Should `coalesceBatchUpdates` live on `EntityIndexVersionConfiguration` for symmetry with `cleanup`, or is the explicit constructor arg deliberate to keep it out of the typed config object? Either is fine — just want to know if it was a conscious choice.
3. Any concern that with `default=true`, a hot-loop ingestion that submits N MCLs per (urn, aspect) per second will now emit substantially fewer `appendRunId` calls? The runId array on the doc may shrink in observable ways for downstream consumers that watch it.

david-leifker (approved):

metadata-io/src/main/java/com/linkedin/metadata/service/UpdateIndicesV2Strategy.java
```diff
@@ -181,15 +219,96 @@ public void processBatch(
...
+        if (structuredPropertiesHookEnabled) {
+          updateIndexMappings(opContext, event);
+        }
+        updateSearchIndicesForEvent(opContext, event);
+        updateTimeseriesFieldsForEvent(opContext, event);
+      }
+      return;
+    }
+
+    MCLItem survivor = aspectEvents.get(aspectEvents.size() - 1);
```
RyanHolstien, line 246: I think this should be correct. at this point it should be upsert semantics only so the difference between the first event's previous and the final event's current should be the expected final state.

RyanHolstien (approved):

PR #17301, opened 2026-05-05
refactor(ingest): remove litellm in favour of direct provider transports

nwadams:
## Summary

- Removes the `litellm` dependency (CVE surface, deep transitive footprint) from ingestion. Embedding generation is now routed through a small `EmbeddingProvider` ABC with one impl per provider.
- Bedrock uses **boto3** (SigV4 + AWS credential chain). Cohere, OpenAI, and the local OpenAI-compatible provider use **raw HTTP via `requests`**. Vertex AI uses **`google-auth`** for ADC / OAuth refresh and POSTs the `:predict` call over plain HTTP — mirroring the wire format used by the Java `VertexAiEmbeddingProvider` in GMS.
- `setup.py` `embedding_common` now declares only `aws_common` + `google-auth>=2,<3`. `pyproject.toml`, `uv.lock`, and `constraints.txt` regenerated.

This shrinks the `unstructured` extra by ~30 transitive packages, removes a recurring CVE source, and keeps Python ingestion's wire-level contract aligned with the server-side providers in `metadata-io`.

## Test plan

- [x] `./gradlew :metadata-ingestion:lint` — ruff + mypy clean
- [x] `./gradlew :metadata-ingestion:updateLockFile` — pyproject/uv.lock/constraints regenerated
- [x] 96 unstructured unit tests pass, including 8 new wire-format tests for the Cohere/OpenAI HTTP providers
- [ ] Smoke test: run a real ingestion against each provider (Bedrock, Cohere, OpenAI, Vertex, Ollama) before merge


david-leifker (approved):

metadata-ingestion/src/datahub/ingestion/source/unstructured/embedding_providers/bedrock.py
```diff
@@ -0,0 +1,71 @@
...
+_COHERE_PREFIX = "cohere."
+_TITAN_PREFIX = "amazon.titan-embed"
+
+
+class BedrockEmbeddingProvider(EmbeddingProvider):
+    """Embedding via AWS Bedrock Runtime (boto3)."""
+
+    def __init__(self, model: str, aws_region: Optional[str]):
+        try:
+            import boto3
```
alexsku: why do we need this, why can't we import boto3 normally?

metadata-ingestion/src/datahub/ingestion/source/unstructured/embedding_providers/bedrock.py
```diff
@@ -0,0 +1,71 @@
+"""AWS Bedrock embedding provider.
+
+Uses boto3 because SigV4 + the AWS credential chain (IRSA, AssumeRole, IMDSv2)
+is the part where rolling our own would actually hurt.
+"""
+
+from __future__ import annotations
```
alexsku: what do we need __future__ for?

metadata-ingestion/src/datahub/ingestion/source/unstructured/embedding_providers/cohere.py
```diff
@@ -0,0 +1,59 @@
+"""Cohere embedding provider (raw HTTP)."""
+
+from __future__ import annotations
```
alexsku: I would avoid using __future__ if possible

metadata-ingestion/src/datahub/ingestion/source/unstructured/embedding_providers/openai.py
```diff
@@ -0,0 +1,68 @@
...
+        self._session = build_retrying_session()
+        self._session.headers.update(
+            {
+                "Authorization": f"Bearer {api_key or 'local'}",
+                "Content-Type": "application/json",
+            }
+        )
+
+    def embed(self, texts: list[str]) -> EmbeddingResult:
+        payload = post_json(
```
alexsku, line 55: is there a reason we are not using openai sdk here?

alexsku (changes requested):
Thanks for taking on the litellm removal — the wire-format tests are solid and the per-provider split is a clear improvement. A few things I'd like addressed before merge.

## Requested changes

### 1. Drop `from __future__ import annotations` from all six new files

`setup.py` targets `python_requires=">=3.10"`, where `list[str]`, `dict[str, Any]`, `Optional[…]`, and `X | None` all work natively at runtime. Across `src/datahub/ingestion/source/` only **16 of 639 files** import `from __future__ import annotations` — it's not a project convention.

This PR adds it to all six provider files (`base.py`, `bedrock.py`, `cohere.py`, `openai.py`, `vertex_ai.py`, `factory.py`), and none of the annotations actually require it. The cost isn't theoretical: it stringifies every annotation, which silently breaks anything that resolves types at runtime — Pydantic v2 in particular. The rest of `unstructured/` is pydantic-heavy, so anyone moving a model into one of these files later gets a surprise. Removal is mechanical: `factory.py`'s one forward reference (`EmbeddingConfig`) is already string-quoted.

### 2. `boto3` / `google-auth` lazy-import-with-try/except is dead code

`bedrock.py:27-33`:
\`\`\`python
try:
    import boto3
except ImportError as e:
    raise ImportError(
        \"boto3 is required for the bedrock embedding provider. \"
        \"Install with: pip install 'acryl-datahub[unstructured]'\"
    ) from e
\`\`\`

Looking at `setup.py`:
\`\`\`python
embedding_common = { *aws_common, \"google-auth>=2.0.0,<3.0.0\" }
unstructured_lib = { ..., *embedding_common }
\`\`\`

`boto3` and `google-auth` are unconditionally in the `[unstructured]` extra. There's no install path that gives you `embedding_providers/bedrock.py` without `boto3`, so the `ImportError` branch is unreachable, and the \"install with `[unstructured]`\" hint is wrong directionally — the user is *already* on that extra to be running this code.

Also, `factory.py` imports all four provider modules unconditionally at top, so the lazy-inside-`__init__` placement doesn't even achieve lazy loading — `boto3` is imported as soon as anything imports the factory.

Either move the imports to module top (and drop the try/except), or — if the intent is that Bedrock/Vertex are *optional* — split them into separate extras (`[bedrock]`, `[vertex]`) so the ImportError can actually fire. The current shape is the worst of both: dead code that suggests an extra that doesn't exist.

Same applies to `google.auth` in `vertex_ai.py:34-44`.

### 3. `OPENAI_API_KEY` env var is advertised but not honored at provider construction

`_validate_provider_init_requirements` accepts `OPENAI_API_KEY` from `os.environ` (`chunking_source.py:208-216`), and `test_openai_api_key_from_env_var` confirms `__init__` succeeds. But the factory only passes `config.api_key` into the provider (`factory.py:75-77`) — the env var is never consulted — and the provider's constructor raises if `api_key` is falsy:

\`\`\`python
# openai.py:38-42
if not api_key and base_url is None:
    raise ValueError(
        \"OpenAI API key is required. Set embedding.api_key or \"
        \"OPENAI_API_KEY, or configure provider='local' with an endpoint.\"
    )
\`\`\`

Walk-through with `OPENAI_API_KEY=sk-…` set, no `embedding.api_key` in recipe:

1. `_validate_provider_init_requirements` → passes (env set).
2. Source initializes; ingestion starts processing documents.
3. First call to `_get_provider()` → `OpenAIEmbeddingProvider(api_key=None, …)`.
4. ValueError raised, telling the user to set the env var they already set.

Cohere does this correctly: `cohere.py:28` reads `os.environ.get(\"COHERE_API_KEY\")` inside the constructor. Vertex AI works because the factory itself reads `VERTEX_AI_PROJECT_ID` (`factory.py:106`) before passing it. OpenAI is the inconsistent one.

Fix: mirror Vertex AI's pattern in the factory — `api_key=api_key or os.environ.get(\"OPENAI_API_KEY\")` for the `openai` branch — and extend `test_openai_api_key_from_env_var` to exercise `_get_provider()` so the lazy construction path is tested, not just `__init__`. Note: this finding also disappears for free if you adopt the OpenAI SDK (see below), since the SDK reads the env var by default.

## Strong recommendation: use the OpenAI SDK, and SDKs in general

The PR's framing of \"no SDKs needed for HTTP-based providers\" doesn't survive the diff. The PR adds:

- `boto3` (large, with full AWS surface) for Bedrock
- `google-auth` (medium) for Vertex AI

…but reimplements OpenAI and Cohere over raw `requests`. The CVE argument that motivated dropping `litellm` was specifically about *litellm's* footprint (~30 transitive packages); the `openai` SDK is a much smaller and more curated package, and using it would give you for free:

- `OPENAI_API_KEY` env var resolution (the bug above)
- `Organization` / `Project` header handling (org-scoped keys are common)
- `base_url` override for OpenAI-compatible servers (the local/Ollama path is one config knob, not a separate codepath)
- Tested response parsing across endpoint variants
- Retry-After honoring on 429 (more nuanced than the urllib3 retry)
- A maintained wire format you don't have to track

The `OpenAIEmbeddingProvider` reimplementation is ~70 lines of code + ~150 lines of tests verifying URL/headers/body shape. Replacing that with the SDK is roughly 10 lines and the wire-format tests collapse into \"the SDK does what the SDK does.\"

Same logic for Cohere — `cohere` SDK is small, handles auth, and matches the server-side embedding contract.

If the goal is to minimize attack surface, the next target should be `boto3`, not OpenAI. The current mix — SDK for AWS + Vertex, hand-rolled HTTP for OpenAI + Cohere — is hard to defend on either footprint or maintainability grounds. I'd push back on this direction as the file structure being established here will get copied for any future provider.

Either go all-in on SDKs (boto3 + google-auth + openai + cohere — one consistent pattern) or, if SDK avoidance is genuinely the goal, go all-in on `requests` (drop boto3, sign SigV4 by hand — painful but at least consistent). The middle path is the most code to maintain for the least benefit.

alexsku (commented):
Follow-up review notes from the same pass — separating these out so the request-changes review stays focused on the four blocker-style items. None of these are merge-blockers, but the MEDIUMs are worth fixing in the same PR while the code is fresh.

## Medium

### `_validate_provider_init_requirements` no longer checks model presence — silent regression vs old code

Old code (pre-PR):
\`\`\`python
if self.config.embedding.provider == \"bedrock\":
    assert self.config.embedding.model is not None
    self.embedding_model = f\"bedrock/{self.config.embedding.model}\"
\`\`\`

New code (`chunking_source.py:158-162`):
\`\`\`python
self._validate_provider_init_requirements(self.config.embedding)
self.embedding_model = derive_model_id(
    self.config.embedding.provider, self.config.embedding.model
)
\`\`\`

`_validate_provider_init_requirements` only checks API keys for Cohere/OpenAI and `vertex_project_id` for Vertex — it never checks that `model` is set. `derive_model_id(provider, None)` returns `None`, so `self.embedding_model = None`, and `_generate_embeddings` becomes a silent no-op via the `if self.embedding_model:` guard.

Repro: `provider=\"bedrock\"` + no `model` + `allow_local_embedding_config=true`. Old behavior: `AssertionError` at init. New behavior: source initializes cleanly, every document logs \"Skipping embedding generation\" at debug level, no error is reported, no `SemanticContent` aspect is emitted. The user thinks semantic search is wired up but it isn't.

The `from_server` and `validate_against_server` paths both `assert self.model is not None`, so this only bites users with explicit local config + break-glass. Niche but a real correctness regression. Fix: add a `provider and not embedding_config.model` check to `_validate_provider_init_requirements`.

### Auto-derived `model_embedding_key` contains `:` for Bedrock Titan models

`chunking_source.py:822-825`:
\`\`\`python
if self.config.embedding.model_embedding_key:
    model_key = self.config.embedding.model_embedding_key
elif \"embed-english-v3\" in self.config.embedding.model:
    model_key = \"cohere_embed_v3\"
else:
    model_key = self.config.embedding.model.replace(\"-\", \"_\").replace(\".\", \"_\")
\`\`\`

For `model=\"amazon.titan-embed-text-v2:0\"` (a real Bedrock Titan model ID), this yields `\"amazon_titan_embed_text_v2:0\"`. The pydantic validator at `chunking_config.py:127-136` only blocks `.`, so the colon survives the auto-derivation and lands in the `SemanticContent.embeddings` map key — which is used as an Elasticsearch field name and rejected at index time.

Fix: extend the replace chain to cover `:` (and probably any non-alphanumeric/underscore), and tighten the validator to match. Or require `model_embedding_key` explicitly when `provider=\"bedrock\"` and the model starts with `amazon.titan-embed`.

## Low

### `tokenCount` is always `None`, so `totalTokens` is always `0`

`chunking_source.py:800` sets `tokenCount=None` for every chunk, then `chunking_source.py:813` does `totalTokens=sum(c.tokenCount or 0 …)` — which is permanently 0. Either compute tokens (Cohere returns `meta.billed_units.input_tokens`; OpenAI returns `usage.prompt_tokens`; both are essentially free to plumb through `EmbeddingResult`), or drop `totalTokens` from the emitted aspect rather than emitting a misleading 0.

### `post_json` echoes up to 500 chars of `response.text` in error messages

`base.py:91-94`:
\`\`\`python
raise RuntimeError(
    f\"Embedding API call failed: {response.status_code} {response.reason}: \"
    f\"{response.text[:500]}\"
)
\`\`\`

If a provider's error response ever echoes the input — some 4xx APIs do, especially auth/rate-limit responses with debugging hints — that text lands in pipeline reports and logs. Low risk in practice (the configured providers don't echo input today), but worth tightening to status + reason in the user-visible message and logging the body snippet at debug level.

### `post_json`'s `session=None` fallback is dead code

`base.py:75-89` accepts `Optional[requests.Session]` and falls back to bare `requests.post()` when `session=None` — which bypasses the retry adapter that's the whole point of `build_retrying_session`. No production caller passes `None`: cohere, openai, and vertex_ai all build a retrying session. The optional parameter is a footgun for any future caller. Either drop the `Optional` and make `session` required, or have `post_json` build one internally if not passed.

### PR test plan: real-provider smoke test row is unchecked

The unit tests with mocked sessions look thorough, but the PR description's last test-plan item — *\"Smoke test: run a real ingestion against each provider (Bedrock, Cohere, OpenAI, Vertex, Ollama) before merge\"* — is unchecked. Given the breadth of the wire-format change, it'd be good to confirm this ran (or run it) before merging. The wire-format mocks verify *intent*; only a live call verifies that each provider actually accepts the shape we're sending.

metadata-ingestion/src/datahub/ingestion/source/unstructured/embedding_providers/openai.py
nwadams: we wanted to have simplicity to avoid importing a ton of extra packages at the moment similar to how we handle the backend GMS api.

nwadams:
@alexsku re the SDKs

- `boto3` is already always included so we use it for bedrock for simplicity. 
- `google-auth` is needed to support gcp auth. We could use a newer `google-genai` package that supports vertex but it still pulls in several new dependencies.
- OpenAI is a fairly small SDK that we could include with few new dependencies
- Cohere pulls in several dependencies including `tokenizers` which is fairly large. 

We could include some or all of these dependencies but would bloat the package somewhat unless we added acryl-datahub[openai] or similar optional dependency packages that allowed us to split further based on what the callers needed.

I'm trying to match what we do on GMS right now and minimize the number of dependencies where we use direct URL calls there as well. Longer term we should move to server side embedding and remove all the client side logic here. Thoughts?

alexsku (approved):
Looks good — approving. Thanks for the focused turnaround. All four CHANGES_REQUESTED items and 5 of 6 follow-up items landed in a single clean commit, with new tests covering each fix.

## One nit before merge

`test_model_key_default_sanitizes_colon_in_titan_model_id` (`tests/unit/ingestion/source/unstructured/test_chunking_source.py:1093`):

\`\`\`python
keys = list(_semantic_embeddings(semantic_wu).keys())
assert any(\":\" not in k for k in keys)
\`\`\`

`any(\":\" not in k for k in keys)` is True if *one* key is colon-free, so it passes vacuously when there's only one key (which is the case here — and would also pass if every key contained a colon, as long as there were two of them). The intent is `all`:

\`\`\`python
assert all(\":\" not in k for k in keys)
\`\`\`

The follow-up `assert \"amazon_titan_embed_text_v2_0\" in _semantic_embeddings(semantic_wu)` does enforce the real invariant, so the bug is just in the helper assertion — not load-bearing for the test's correctness. Worth a one-line fix while you're in the file.

## Other follow-ups (non-blocking)

- The smoke-test row (\"run a real ingestion against each provider before merge\") is still unchecked in the PR description. With the wire-format reimplementation, that's the one piece the mocked tests can't substitute for — recommend running it before merge and ticking the box. Especially Vertex AI's `:predict` path and Bedrock Titan's per-input loop, which are the most format-sensitive.

- I'll let the SDK-adoption question rest — your call on tradeoffs. If you do reconsider in a follow-up PR, the bedrock + vertex paths are the natural places to start (they're already SDK-using, so the inconsistency cost is what's left to address). Happy to dig into specifics in a separate thread if it'd be useful.

- Tangentially, `EmbeddingConfig.input_type` (`chunking_config.py:108-111`) defaults to `\"search_document\"` and is described as \"Input type for Cohere embeddings\", but the new providers hardcode that string and ignore the config field. Pre-existing dead config (predates this PR), but if it's never going to be wired through, removing the field would prevent users from setting it expecting it to do something. Outside this PR's scope.

Nice work overall — the wire-format tests and the env-var coverage are particularly strong.

PR #17304, opened 2026-05-05
fix(SQL Parsing): Added new parsing functions for object construction

pdruley:
## Summary

**Problem:** Column lineage through Snowflake's `OBJECT_CONSTRUCT(*)` pattern failed when the source CTE renamed a column (e.g. `cost_structure_type AS cost_structure`). The JSON object keys use the CTE alias names, but sqlglot's `pushdown_projections` optimizer eliminates aliased columns that aren't referenced elsewhere — discarding the alias→original mapping before lineage resolution could use it.

**Fix (`sqlglot_lineage.py`):**

- Added `_collect_object_construct_source_aliases()` — walks the pre-optimization AST to find CTEs feeding `OBJECT_CONSTRUCT(*)` and builds an alias→original column name map
- Added `_extract_json_path_keys()` — extracts JSON path key strings from semi-structured access expressions like `col[0]:KEY::TYPE`
- Added `_get_star_map_col_upstreams()` — resolves upstream column refs for `StarMap` nodes by matching JSON path keys against schema column names, with alias fallback
- Extended `_get_direct_raw_col_upstreams()` and `_select_statement_cll()` to thread `object_construct_col_aliases` through to the star map resolution
- Fixed mypy type error: widened `_extract_json_path_keys()` parameter from `sqlglot.exp.Expression` to `sqlglot.exp.Expr` to match the actual type of `lineage.Node.expression`

**Tests:**
- Added `test_snowflake_object_construct_star_lineage` — baseline coverage for `OBJECT_CONSTRUCT(*)` lineage through a CTE
- Added `test_snowflake_object_construct_star_alias_lineage` — covers the aliased-CTE case where JSON keys don't match base table column names directly

### DIFF
--- a/metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
+++ b/metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
@@ -1317,10 +1317,7 @@ def _collect_source_cte_chain(
     )
     if not cte_select:
         return
-    from_clause = cte_select.find(sqlglot.exp.From)
-    if not from_clause:
-        return
-    for src_table in from_clause.find_all(sqlglot.exp.Table):
+    for src_table in cte_select.find_all(sqlglot.exp.Table):
         _collect_source_cte_chain(src_table.name.upper(), cte_map, visited)
 
 
@@ -1348,8 +1345,7 @@ def _collect_object_construct_source_aliases(
     }
 
     # Find CTEs that contain OBJECT_CONSTRUCT(*) and collect their full upstream CTE
-    # chain.  find_all on the FROM clause captures all joined sources, not just the
-    # first table.
+    # chain. find_all on the Select captures all tables including JOINed sources.
     source_cte_names: Set[str] = set()
     for cte in statement.find_all(sqlglot.exp.CTE):
         if not cte.find(sqlglot.exp.StarMap):
@@ -1361,10 +1357,7 @@ def _collect_object_construct_source_aliases(
         )
         if not cte_select:
             continue
-        from_clause = cte_select.find(sqlglot.exp.From)
-        if not from_clause:
-            continue
-        for source_table in from_clause.find_all(sqlglot.exp.Table):
+        for source_table in cte_select.find_all(sqlglot.exp.Table):
             _collect_source_cte_chain(
                 source_table.name.upper(), cte_map, source_cte_names
             )
@@ -1509,8 +1502,14 @@ def _get_direct_raw_col_upstreams(
 
     # Build a child→parent map so we can look up what key is accessed on each column
     # when the column comes from OBJECT_CONSTRUCT(*) / StarMap.
+    # StarMap only appears in Snowflake-dialect ASTs (OBJECT_CONSTRUCT(*)), so skip
+    # this traversal for all other dialects.
     parents: Dict[int, sqlglot.lineage.Node] = {}
-    if table_name_schema_mapping:
+    if (
+        table_name_schema_mapping
+        and dialect is not None
+        and is_dialect_instance(dialect, "snowflake")
+    ):
 
         def _build_parents(root: sqlglot.lineage.Node) -> None:
             stack: List[Tuple[sqlglot.lineage.Node, Optional[sqlglot.lineage.Node]]] = [
--- a/metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
+++ b/metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
@@ -1027,6 +1027,7 @@ def _select_statement_cll(
     table_name_schema_mapping: Dict[_TableName, SchemaInfo],
     default_db: Optional[str] = None,
     default_schema: Optional[str] = None,
+    object_construct_col_aliases: Optional[Dict[str, str]] = None,
 ) -> List[_ColumnLineageInfo]:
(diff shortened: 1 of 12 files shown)


metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
```diff
@@ -1234,20 +1243,206 @@ def _column_level_lineage(
...
+            ):
+                path = json_extract.args.get("expression")
+                if isinstance(path, sqlglot.exp.JSONPath):
+                    for part in path.expressions:
+                        if isinstance(part, sqlglot.exp.JSONPathKey):
+                            keys.add(str(part.this))
+    return keys
+
+
+def _collect_object_construct_source_aliases(
```
askumar27, line 1338: (suggestions) This only handles Column aliases, not function aliases (to_date(x) AS x).
- Only walks one level of source CTE chain — if cte1 → cte2 (passthrough) → cte3 (OBJECT_CONSTRUCT), alias from cte1 is missed. 
- `from_clause.find(sqlglot.exp.Table)` returns first table only — multi-source CTEs with joins not fully tracked. 
- Triple-nested loop at the end of _get_star_map_col_upstreams (keys × tables × all_schemas × cols), needs to be reviewed for performance for large schemas

We need to document atleast if not addressed

askumar27 (changes requested):
As this is a snowflake specific change we should add a dialect guard. Should gate with if `is_dialect_instance(dialect, "snowflake")`

`GET()` syntax gap (separate columns): `GET(ACP[0], 'RATE')` parses to a different AST node than `:KEY`, so `DSP_CONTRACT_CLIENT_AD_SERVING_RATE` etc. won't be resolved.

ligfx:
> sqlglot's pushdown_projections optimizer eliminates aliased columns that aren't referenced elsewhere — discarding the alias→original mapping before lineage resolution could use it.

This is a SQLGlot bug and should be fixed there instead of us writing DataHub-specific code to work around it.

I've filed https://github.com/tobymao/sqlglot/issues/7607 for tracking.

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
pdruley: Resolved in https://github.com/datahub-project/datahub/pull/17304/commits/0915ba6161f5527528913cfe40bab4a83aaf39c6

## Summary of Changes

### 1. Function aliases — fixed
Changed `_collect_object_construct_source_aliases` to handle function-expression aliases
like `to_date(col) AS alias`. Previously only `col AS alias` was tracked. Now any alias
whose inner expression isn't a plain column gets walked with `find_all(sqlglot.exp.Column)`
to collect all referenced column names. Return type changed from `Dict[str, str]` to
`Dict[str, List[str]]` throughout, and the consumer in `_get_star_map_col_upstreams`
updated to `candidates.extend(originals)`.

### 2. Multi-level CTE chain — fixed
Added `_collect_source_cte_chain`, a recursive helper that follows the full upstream CTE
graph. Previously only the direct source of an OBJECT_CONSTRUCT CTE was inspected. Now
`cte1 → cte2 (passthrough) → cte3 (OBJECT_CONSTRUCT)` correctly surfaces aliases defined
in `cte1`.

### 3. Multi-source JOINs — fixed
Replaced `from_clause.find(sqlglot.exp.Table)` with `from_clause.find_all(sqlglot.exp.Table)`
in both `_collect_object_construct_source_aliases` and `_collect_source_cte_chain`, so all
joined tables are tracked, not just the first one.

### 4. Nested loop performance — documented
Added a comment at the loop in `_get_star_map_col_upstreams` noting the
O(keys × leaf_tables × schema_entries × cols) complexity, why it's acceptable for typical
workloads, and what the optimization path would be (pre-build an inverted col→table index)
if it ever becomes a bottleneck.

pdruley:
> As this is a snowflake specific change we should add a dialect guard. Should gate with if `is_dialect_instance(dialect, "snowflake")`
> 
> `GET()` syntax gap (separate columns): `GET(ACP[0], 'RATE')` parses to a different AST node than `:KEY`, so `DSP_CONTRACT_CLIENT_AD_SERVING_RATE` etc. won't be resolved.

Resolved in https://github.com/datahub-project/datahub/pull/17304/commits/69cf43b64395e7b96ef194b13da571801ec89eda

## Changes

- **Function aliases**: `_collect_object_construct_source_aliases` now extracts column
  references from function expressions (`to_date(col) AS alias`). Return type widened
  to `Dict[str, List[str]]`.
- **Multi-level CTE chain**: New `_collect_source_cte_chain` recursive helper captures
  aliases from CTEs multiple hops upstream of `OBJECT_CONSTRUCT(*)`.
- **Multi-source JOINs**: `find_all(Table)` replaces `find(Table)` so joined sources
  are fully tracked.
- **`GET()` syntax**: `_extract_json_path_keys` now handles `GetExtract` nodes alongside
  `JSONExtract`.
- **Dialect guard**: All `OBJECT_CONSTRUCT` logic gated on Snowflake dialect.
- **Nested loop**: Complexity documented with optimization path noted for future.
- **Test**: Added `test_snowflake_object_construct_get_lineage` covering the `GET()` path.

pdruley:
Closing as the test case being used was actually expected behavior due to the upstream being an expression that resulted in a string literal. The sqlglot bug may eventually help these cases as well and there are potential improvements for object handling in the future.

PR #17312, opened 2026-05-06
feat(ingest/sigma): parse customSQL DM element definitions

kyungsoo-datahub:
- Parses the `definition` SQL on DM `customSQL` lineage entries via                    
    `SqlParsingAggregator.add_view_definition` and emits `UpstreamLineage`               
    + `FineGrainedLineage` on the backing DM element Dataset.             
  - FGL downstream field names are rewritten from SQL column names to Sigma              
    display names using each column's `[Custom SQL/SQL_COL]` formula ref,                
    so they match the `SchemaMetadata` already emitted for the element.                  
  - `generate_column_lineage=False` kill-switch is untouched.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/sigma/config.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/sigma/config.py
@@ -312,6 +312,19 @@ class SigmaSourceReport(StaleEntityRemovalSourceReport):
 
     element_dm_edge: ElementDmEdgeReport = field(default_factory=ElementDmEdgeReport)
 
+    # DM customSQL element SQL parsing counters.
+    # Elements skipped before aggregator registration (no definition, missing /
+    # unknown connection, unsupported platform, duplicate source_id).
+    dm_customsql_skipped: int = 0
+    dm_customsql_aggregator_invocations: int = 0
+    dm_customsql_aggregator_invocation_errors: int = 0
+    dm_customsql_parse_failed: int = 0
+    dm_customsql_upstream_emitted: int = 0
+    dm_customsql_column_lineage_emitted: int = 0
+    # FGL downstream fields dropped because the SQL column name had no matching
+    # Sigma display column (formula ref absent or element name mismatch).
+    dm_customsql_fgl_downstream_unmapped: int = 0
+
     # Connection registry counters.
     # Records whose Sigma type mapped to a known DataHub platform.
     connections_resolved: int = 0
--- a/metadata-ingestion/src/datahub/ingestion/source/sigma/data_classes.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/sigma/data_classes.py
@@ -272,6 +272,15 @@ class SigmaDataModelElement(BaseModel):
         return values
 
 
+class CustomSqlEntry(BaseModel):
+    """Shape of a ``customSQL`` entry from ``/v2/dataModels/{id}/lineage``."""
+
+    name: str
+    type: str = ""
+    connectionId: str = ""
+    definition: str = ""
+
+
 class SigmaDataModel(BaseModel):
     dataModelId: str  # UUID; stable across renames
     name: str
@@ -289,7 +298,7 @@ class SigmaDataModel(BaseModel):
     badge: Optional[str] = None
     elements: List[SigmaDataModelElement] = []
     # Populated from /lineage ``data-model`` type entries during assembly.
-    # Maps source DM dataModelId (UUID) → [element names from that source DM].
+    # Maps source DM dataModelId (UUID) -> [element names from that source DM].
     # Used by cross-DM entity-level resolution to look up the correct source
     # element name without requiring the consuming element to share that name.
     source_dm_element_names: Dict[str, List[str]] = Field(default_factory=dict)
@@ -300,6 +309,10 @@ class SigmaDataModel(BaseModel):
     warehouse_inodes_by_inode_id: Dict[str, WarehouseInodeRaw] = Field(
         default_factory=dict
     )
+    # Populated from /lineage ``customSQL`` type entries during assembly.
+    # Maps customSQL entry name (the identifier elements reference in sourceIds)
+    # -> raw lineage entry dict (carries connectionId and definition).
+    custom_sql_by_name: Dict[str, CustomSqlEntry] = Field(default_factory=dict)
 
     def get_url_id(self) -> str:
         """Return the DM's URL identifier: explicit ``urlId`` if set,
--- a/metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
@@ -127,6 +127,12 @@ logger = logging.getLogger(__name__)
 # column field (or we add SQL-based inference), swap this out here.
 SIGMA_DM_UNKNOWN_COLUMN_NATIVE_TYPE = "unknown"
 
+# FGL confidence scores for customSQL element lineage.
+_FGL_CONFIDENCE_SQL_PARSED: float = (
+    0.2  # aggregator-derived; matches SqlParsingAggregator
+)
+_FGL_CONFIDENCE_FORMULA_DERIVED: float = 0.1  # SELECT * synthesis from formula refs
+
 
 def _dm_column_ranks_above(
     candidate: SigmaDataModelColumn, incumbent: SigmaDataModelColumn
@@ -1177,7 +1183,7 @@ class SigmaSource(StatefulIngestionSourceBase, TestableSource):
                             downstreams=[
                                 builder.make_schema_field_urn(entity_urn, sigma_col)
                             ],
-                            confidenceScore=0.1,
+                            confidenceScore=_FGL_CONFIDENCE_FORMULA_DERIVED,
                         )
                         for sql_col, sigma_col in col_mapping.items()
                     ]
@@ -1239,11 +1245,16 @@ class SigmaSource(StatefulIngestionSourceBase, TestableSource):
     def _drain_sql_aggregators(self) -> Iterable[MetadataWorkUnit]:
         """Drain all per-platform aggregators and emit lineage workunits.
 
-        Called once at the end of ``get_workunits_internal`` after all DM
-        elements have been registered via ``add_view_definition``.  FGL
-        downstream schemaField URNs are rewritten to match Sigma display names
-        before yielding.  Each aggregator is closed in a finally block so
-        SQLite-backed tempfiles are released even if gen_metadata raises.
+        Intentionally deferred to the end of ``get_workunits_internal``: all DM
+        elements must be registered via ``add_view_definition`` before parsing
+        runs, so the aggregator sees the full view set in one pass.  The
+        consolidated ``UpstreamLineage`` MCP emitted here is the source of truth
+        for customSQL-backed element lineage; any earlier per-element
+        ``UpstreamLineage`` MCP (for non-customSQL upstreams) is superseded by
+        the merged aspect yielded below.  FGL downstream schemaField URNs are
+        rewritten to Sigma display names before yielding.  Each aggregator is
+        closed in a finally block so SQLite-backed tempfiles are released even
+        if gen_metadata raises.
         """
         for cache_key, aggregator in sorted(
             self._sql_aggregators.items(),
--- a/metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
@@ -44,6 +44,7 @@ from datahub.ingestion.source.sigma.connection_registry import (
     SigmaConnectionRegistry,
 )
 from datahub.ingestion.source.sigma.data_classes import (
+    CustomSqlEntry,
     DataModelElementUpstream,
     DataModelKey,
     DatasetUpstream,
@@ -108,6 +109,8 @@ from datahub.metadata.schema_classes import (
     SubTypesClass,
     TagAssociationClass,
 )
+from datahub.metadata.urns import SchemaFieldUrn
+from datahub.sql_parsing.sql_parsing_aggregator import SqlParsingAggregator
 from datahub.sql_parsing.sqlglot_lineage import create_lineage_sql_parsed_result
 from datahub.utilities.urns.dataset_urn import DatasetUrn
 from datahub.utilities.urns.error import InvalidUrnError
@@ -282,6 +285,27 @@ class SigmaSource(StatefulIngestionSourceBase, TestableSource):
         # dataModelIds whose bridge key collided with an earlier DM. The
         # emit loop skips these to avoid unlinked orphan Containers.
         self.dm_collided_data_model_ids: Set[str] = set()
+        # Per-platform SqlParsingAggregator instances for customSQL DM elements,
+        # keyed by (platform, env, platform_instance).
+        self._sql_aggregators: Dict[
+            Tuple[str, str, Optional[str]], SqlParsingAggregator
+        ] = {}
+        # element_dataset_urn -> {sql_col_lower -> sigma_col_name}.
+        # Built from ALL formula refs for FGL rewriting of named-column SELECTs.
+        self._customsql_col_mappings: Dict[str, Dict[str, str]] = {}
+        # Same key, but only single-ref (passthrough) formulas.
+        # Used for SELECT * synthesis to avoid fabricating upstream edges for
+        # computed expressions that reference multiple SQL columns.
+        self._customsql_passthrough_mappings: Dict[str, Dict[str, str]] = {}
+        # Element URNs registered with an aggregator via add_view_definition.
+        # Used to guard against multiple customSQL source_ids on one element
+        # and to scope dm_customsql_upstream_emitted to known registrations.
+        self._customsql_registered_urns: Set[str] = set()
+        # element_urn -> non-customSQL Upstream / FGL objects stashed from the
+        # per-element emit path.  Merged into the aggregator's UpstreamLineage
+        # MCP at drain so the final aspect is consolidated.
+        self._customsql_extra_upstreams: Dict[str, List[Upstream]] = {}
+        self._customsql_extra_fgls: Dict[str, List[FineGrainedLineageClass]] = {}
         # DM urlId → DM dataModelId (UUID). Reverse of get_url_id(); used to
         # correlate ``data-model`` lineage entries (keyed by dataModelId) with
         # source_id prefixes (keyed by urlId) in cross-DM upstream resolution.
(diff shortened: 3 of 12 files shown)


metadata-ingestion/tests/unit/sigma/test_dm_customsql_parsing.py
```diff
@@ -0,0 +1,482 @@
...
+        all_downstreams = [d for fgl in fgls for d in (fgl.downstreams or [])]
+        for ds in all_downstreams:
+            sfu = SchemaFieldUrn.from_string(ds)
+            assert sfu.field_path == "order_id"  # SQL name; not rewritten
+
+    def test_mixed_sources_extra_upstreams_and_fgls_merged(self) -> None:
+        # When an element has both customSQL and non-customSQL source_ids,
+        # both the non-customSQL Upstream edges and their FGL must appear in the
+        # final UpstreamLineage — not silently overwritten by the drain MCP.
+        from datahub.metadata.com.linkedin.pegasus2avro.dataset import (
```
treff7es: 🔴 **BLOCKER — Late/inline imports in test bodies**

DataHub's `testing.md` lists late/inline imports as an absolute blocker absent a documented circular-import reason. Three test methods do this:

- L340-346 (this method): imports `DatasetLineageType`, `FineGrainedLineageClass`, `FineGrainedLineageDownstreamTypeClass`, `FineGrainedLineageUpstreamTypeClass`, `Upstream`
- L410-414 (`test_fgl_rewrite_passes_through_when_ds_parent_not_in_mappings`)
- L423-428 (same method) — these even re-import `MetadataChangeProposalWrapper` and `UpstreamLineage` that are already at the top of the file

Mechanical move to top-of-file fixes this with no behaviour change.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma_api.py
```diff
@@ -1086,6 +1088,16 @@ def _assemble_data_model(
...
+            elif entry_type in ("customSQL", "customSql"):
+                name = entry.get("name")
+                if isinstance(name, str) and name:
+                    if name in data_model.custom_sql_by_name:
+                        self.report.warning(
+                            title="Sigma DM customSQL duplicate entry name",
+                            message="Two customSQL lineage entries share the same name; later entry overwrites earlier — elements sourcing the earlier entry will get the wrong SQL.",
+                            context=f"dataModelId={data_model.dataModelId!r}, name={name!r}",
+                        )
+                    data_model.custom_sql_by_name[name] = CustomSqlEntry(entry)  # type: ignore[misc]
```
treff7es: 🔴 **BLOCKER — `# type: ignore[misc]` lacks a reason and bypasses validation**

`code_style.md` requires `# type: ignore` to be specific AND accompanied by a justifying comment.

Beyond the missing reason, the cast is semantically unsafe at runtime: constructing a `TypedDict(total=False)` from `Dict[str, Any]` is a shallow `dict()` — no field-type validation runs. A Sigma payload regression that swaps a string for a number reaches `add_view_definition` later as a confusing `TypeError`.

Note: `CustomSqlEntry` is nested inside `SigmaDataModel.custom_sql_by_name` (a Pydantic field), but because the population path is imperative assignment (not `model_validate`), Pydantic's nested-TypedDict validation never fires here. The type is providing zero runtime safety today.

**Preferred fix:** convert `CustomSqlEntry` to a Pydantic `BaseModel` and call `CustomSqlEntry.model_validate(entry)`.

**Minimum:** add an inline reason and a small `_coerce_custom_sql_entry(d)` helper that picks the four known keys with `isinstance` guards.

metadata-ingestion/src/datahub/ingestion/source/sigma/data_classes.py
```diff
@@ -260,6 +261,15 @@ def _discard_api_bare_string_columns(cls, values: Dict) -> Dict:
         return values
 
 
+class CustomSqlEntry(TypedDict, total=False):
+    """Shape of a ``customSQL`` entry from ``/v2/dataModels/{id}/lineage``."""
+
+    name: str
```
treff7es, line 278: 🟠 **MEDIUM — `total=False` understates the invariant**

`name` is provably required: the producer at `sigma_api.py:1093` filters on `isinstance(name, str) and name` before storing. Mark it as `Required[str]` so the type documents the actual contract — and consumers can drop the defensive `.get("name")` at `sigma.py:786`.

Better: convert this whole TypedDict to a Pydantic `BaseModel` (see comment on `sigma_api.py:1100`) — that fixes the BLOCKER and this together.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
```diff
@@ -666,6 +690,273 @@ def _bump_cross_dm_failure(self, outcome: Optional[str]) -> None:
...
+                    f"element_urn={element_dataset_urn!r}, "
+                    f"skipped_name={customsql_entry.get('name')!r}, "
+                    f"skipped_definition_prefix={(customsql_entry.get('definition') or '')[:80]!r}"
+                ),
+            )
+            return
+
+        definition = (customsql_entry.get("definition") or "").strip()
+        if not definition:
+            self.reporter.dm_customsql_skipped += 1
```
treff7es, line 1082: 🟡 **HIGH — Silent skip paths collapse 4 distinct misconfigs into one counter**

The `if not definition`, `if not connection_id`, `record is None`, `not record.is_mappable` chain (lines 793-807) all bump `dm_customsql_skipped` with no log, no warning. The latter two are real misconfigurations:

- **Unknown `connection_id`** — `/v2/connections` enumeration missed a connection that customSQL is actively using (scope/permission issue)
- **Unmappable platform** — Sigma connection type isn't in `SIGMA_TYPE_TO_DATAHUB_PLATFORM_MAP`

Operators see warehouse lineage silently absent on a subset of elements with no actionable signal.

**Suggestion:**
- Split into `dm_customsql_skipped_no_definition` / `_no_connection_id` / `_unknown_connection` / `_unmappable_platform` (the docstring at `config.py:316-318` already enumerates these)
- For the latter two, emit a throttled `self.reporter.warning(...)` carrying `connection_id` / `sigma_type` and a sample element URN

The duplicate-source-id warning at L779-790 is the gold-standard pattern to mirror.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
```diff
@@ -666,6 +690,273 @@ def _bump_cross_dm_failure(self, outcome: Optional[str]) -> None:
...
+                            element_dataset_urn,
+                            ref.column,
+                            mapping[sql_col_lower],
+                            col.name,
+                        )
+                    mapping[sql_col_lower] = col.name
+                    # Single-ref formula = direct passthrough of an upstream column.
+                    # Multi-ref = computed expression; exclude from SELECT * synthesis
+                    # to avoid fabricating upstream edges for non-existent columns.
+                    if len(refs) == 1:
```
treff7es, line 1054: 🟠 **MEDIUM — Critical single-ref-vs-multi-ref split is untested**

The `len(refs) == 1` check is the load-bearing logic for `SELECT *` synthesis: computed expressions must be excluded from passthrough, otherwise the synthesised FGL fabricates non-existent upstream column edges.

`TestCustomSqlColMapping` only has `test_formula_refs_produce_mapping`; it doesn't cover:
- A multi-ref formula like `If([Custom SQL/A]>0, [Custom SQL/B], 0)` — assert both `A` and `B` land in `_customsql_col_mappings` but **neither** in `_customsql_passthrough_mappings`.
- Duplicate SQL-column ref triggers the `logger.warning` at L753-761 (use-latter semantics).
- Case-insensitive matching (`[CUSTOM SQL/order_id]` matching a `Custom SQL` element).

A regression here could ship silently.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
```diff
@@ -666,6 +690,273 @@ def _bump_cross_dm_failure(self, outcome: Optional[str]) -> None:
...
+            )
+            self.reporter.dm_customsql_aggregator_invocations += 1
+            self._customsql_registered_urns.add(element_dataset_urn)
+        except Exception as e:
+            self.reporter.dm_customsql_aggregator_invocation_error += 1
+            self.reporter.warning(
+                title="Sigma DM customSQL element registration failed",
+                message="SqlParsingAggregator.add_view_definition raised; this element will be emitted without warehouse lineage.",
+                context=f"element_urn={element_dataset_urn}, customsql_name={customsql_entry.get('name')!r}, connection_id={connection_id!r}, platform={record.datahub_platform}",
+                exc=e,
```
treff7es, line 1130: ℹ️ **NIT — Include SQL definition prefix in context for triage parity**

The duplicate-source-id warning at L787 truncates the SQL `definition` to 80 chars and includes it in `context=`. Doing the same here would let operators identify the failing view from the report alone, without rerunning. Same idea, ~3 lines.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
```diff
@@ -666,6 +690,273 @@ def _bump_cross_dm_failure(self, outcome: Optional[str]) -> None:
...
+        # Extra FGLs come from the non-customSQL path and already use Sigma
+        # display names — keep them separate from the aggregator FGL so the
+        # rewrite loop below doesn't try to remap them.
+        extra_fgls = self._customsql_extra_fgls.get(entity_urn)
+
+        if not aspect.fineGrainedLineages:
+            # For SELECT * on a single upstream, synthesize FGL from passthrough
+            # formula refs: only single-ref formulas are included so computed
+            # expressions (If([X]>0,[Y],0)) don't fabricate non-existent upstream
+            # column edges.  Confidence 0.1 (formula-derived, lower than SQL-parsed 0.2).
```
treff7es, line 1171: 🟠 **MEDIUM — `SELECT *` + computed-column behaviour untested**

`test_select_star_synthesizes_fgl_from_col_mapping` covers the all-passthrough case. There's no test where a passthrough column and a computed-expression column coexist on a `SELECT *` element. The synthesised FGL must include only the passthrough downstream — the invariant the `len(refs) == 1` check at L766 protects.

Without coverage, a regression that includes computed columns in synthesised FGL would only surface in production.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
```diff
@@ -666,6 +690,273 @@ def _bump_cross_dm_failure(self, outcome: Optional[str]) -> None:
...
+                    return mcp
+            return mcp
+
+        rewritten_fgls = []
+        for fgl in aspect.fineGrainedLineages:
+            rewritten_downstreams = []
+            for ds_urn in fgl.downstreams or []:
+                try:
+                    sfu = SchemaFieldUrn.from_string(ds_urn)
+                except InvalidUrnError:
```
treff7es, line 1207: 🟡 **HIGH — `InvalidUrnError` silently drops FGL entries**

```python
except InvalidUrnError:
    continue   # silent — no counter, no log
```

A malformed schemaField URN coming out of `SqlParsingAggregator` is unexpected and worth surfacing. If every downstream of an FGL trips this, the entry is dropped from `rewritten_fgls` (L913 only appends when non-empty) without any signal — a regression in URN encoding wouldn't be caught.

Bump `dm_customsql_fgl_downstream_unmapped` (or a dedicated `_invalid_urn` counter) and `logger.warning` with the offending URN.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
```diff
@@ -666,6 +690,273 @@ def _bump_cross_dm_failure(self, outcome: Optional[str]) -> None:
...
+        for cache_key, aggregator in sorted(
+            self._sql_aggregators.items(),
+            key=lambda kv: tuple(x or "" for x in kv[0]),
+        ):
+            try:
+                for mcp in aggregator.gen_metadata():
+                    yield self._rewrite_fgl_downstreams(mcp).as_workunit()
+                agg_report = aggregator.report
+                self.reporter.dm_customsql_parse_failed += agg_report.num_views_failed
+            except Exception as e:
```
treff7es, line 1268: 🟡 **HIGH — Drain failure path is untested**

This `try/except Exception` + `finally: aggregator.close()` makes a documented operational claim (the docstring at L939 says "SQLite-backed tempfiles are released even if `gen_metadata` raises"). `testing.md` requires error-path coverage.

`test_aggregator_invocation_error_counter` only covers the *registration* failure (`add_view_definition`), not the drain failure.

Suggested test: `test_drain_continues_after_aggregator_gen_metadata_fails` — patch `SqlParsingAggregator.gen_metadata` with `side_effect=RuntimeError`, register elements across two platforms, drain, assert (a) good platform's MCPs yielded, (b) reporter has the drain warning, (c) `aggregator.close()` called for both.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
```diff
@@ -666,6 +690,273 @@ def _bump_cross_dm_failure(self, outcome: Optional[str]) -> None:
...
+
+        Called once at the end of ``get_workunits_internal`` after all DM
+        elements have been registered via ``add_view_definition``.  FGL
+        downstream schemaField URNs are rewritten to match Sigma display names
+        before yielding.  Each aggregator is closed in a finally block so
+        SQLite-backed tempfiles are released even if gen_metadata raises.
+        """
+        for cache_key, aggregator in sorted(
+            self._sql_aggregators.items(),
+            key=lambda kv: tuple(x or "" for x in kv[0]),
```
treff7es, line 1261: ℹ️ **SUGGESTION — Determinism not tested**

`_drain_sql_aggregators` deliberately sorts by `cache_key` for stable golden files across multiple platforms, but the integration test only exercises one Snowflake DM. A future dict-iteration-order regression on a multi-platform deployment wouldn't be caught.

A unit test that registers customSQL elements across two platforms (Snowflake + Postgres, with one `platform_instance`) and asserts the MCP `entityUrn` order is a deterministic function of the cache keys would close this gap.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
```diff
@@ -246,6 +249,27 @@ def __init__(self, config: SigmaSourceConfig, ctx: PipelineContext):
...
+        self._customsql_passthrough_mappings: Dict[str, Dict[str, str]] = {}
+        # Element URNs registered with an aggregator via add_view_definition.
+        # Used to guard against multiple customSQL source_ids on one element
+        # and to scope dm_customsql_upstream_emitted to known registrations.
+        self._customsql_registered_urns: Set[str] = set()
+        # element_urn -> non-customSQL Upstream / FGL objects stashed from the
+        # per-element emit path.  Merged into the aggregator's UpstreamLineage
+        # MCP at drain so the final aspect is consolidated.
+        self._customsql_extra_upstreams: Dict[str, List[Upstream]] = {}
+        self._customsql_extra_fgls: Dict[str, List[FineGrainedLineageClass]] = {}
```
treff7es, line 314: 🟠 **MEDIUM — Six related `_customsql_*` fields → primitive obsession**

`_sql_aggregators`, `_customsql_col_mappings`, `_customsql_passthrough_mappings`, `_customsql_registered_urns`, `_customsql_extra_upstreams`, `_customsql_extra_fgls` form one logical state machine, are reset together via 6 manual `.clear()` calls (L2549-2554), and have implicit cross-dict invariants (e.g. "if URN is in `_registered_urns`, it has an entry in some aggregator") that live nowhere in the type.

Suggest a small helper:

```python
@dataclass
class CustomSqlState:
    aggregators: Dict[Tuple[str, str, Optional[str]], SqlParsingAggregator] = field(default_factory=dict)
    col_mappings: Dict[str, Dict[str, str]] = field(default_factory=dict)
    passthrough_mappings: Dict[str, Dict[str, str]] = field(default_factory=dict)
    registered_urns: Set[str] = field(default_factory=set)
    extra_upstreams: Dict[str, List[Upstream]] = field(default_factory=dict)
    extra_fgls: Dict[str, List[FineGrainedLineageClass]] = field(default_factory=dict)

    def reset(self) -> None:
        for d in (self.aggregators, self.col_mappings, self.passthrough_mappings,
                  self.extra_upstreams, self.extra_fgls):
            d.clear()
        self.registered_urns.clear()
```

`__init__` becomes one line; the reset block becomes `self._customsql.reset()`. No behaviour change.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
```diff
@@ -1192,9 +1501,28 @@ def _gen_data_model_element_workunits(
...
+                    if upstream_lineage.fineGrainedLineages:
+                        self._customsql_extra_fgls[element_dataset_urn] = list(
+                            upstream_lineage.fineGrainedLineages
+                        )
+                else:
+                    # No customSQL registration (pre-flight failed or no
+                    # customSQL on this element): emit immediately.
+                    yield MetadataChangeProposalWrapper(
+                        entityUrn=element_dataset_urn, aspect=upstream_lineage
+                    ).as_workunit()
```
treff7es, line 1906: ℹ️ **SUGGESTION — Extract emit-or-stash policy into a helper**

~35 lines of dual-purpose branching inside the main element-emit loop. A `_emit_or_stash_element_lineage(*, element_dataset_urn, has_customsql, upstream_lineage)` generator helper would name the policy and shrink the loop:

```python
yield from self._emit_or_stash_element_lineage(
    element_dataset_urn=element_dataset_urn,
    has_customsql=has_customsql,
    upstream_lineage=upstream_lineage,
)
```

Keyword-only args because the call site has 3 parameters of related types (urn / bool / aspect).

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma_api.py
```diff
@@ -1086,6 +1088,16 @@ def _assemble_data_model(
                     data_model.source_dm_element_names.setdefault(src_dm_id, []).append(
                         src_name.strip()
                     )
+            elif entry_type in ("customSQL", "customSql"):
+                name = entry.get("name")
+                if isinstance(name, str) and name:
```
treff7es, line 1124: 🟠 **MEDIUM — Missing `else` silently drops customSQL with missing/empty `name`**

The filter has no `else` branch. A Sigma payload regression that omits `name` (or sets it to a non-string / empty string) would cause downstream elements referencing the entry to silently miss customSQL processing — they don't even appear in `dm_customsql_skipped` since they never reach `_process_dm_customsql_element`.

Add a throttled `else: self.report.warning(...)` noting `dataModelId` and entry shape.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma_api.py
```diff
@@ -1086,6 +1088,16 @@ def _assemble_data_model(
...
+                    if name in data_model.custom_sql_by_name:
+                        self.report.warning(
+                            title="Sigma DM customSQL duplicate entry name",
+                            message="Two customSQL lineage entries share the same name; later entry overwrites earlier — elements sourcing the earlier entry will get the wrong SQL.",
```
treff7es, line 1132: 🟠 **MEDIUM — Duplicate-name overwrite behaviour untested**

This warning ships with a sharp comment — *"later entry overwrites earlier — elements sourcing the earlier entry will get the wrong SQL"* — and no test coverage for either the warning or the overwrite semantics.

Suggest a unit test in `tests/unit/sigma/test_sigma_api.py` that drives `_assemble_data_model` with two same-name `customSQL` entries and asserts (a) `report.warnings` contains the duplicate-entry warning, (b) `data_model.custom_sql_by_name[name]["definition"]` equals the **second** SQL.

metadata-ingestion/src/datahub/ingestion/source/sigma/config.py
```diff
@@ -312,6 +312,19 @@ class SigmaSourceReport(StaleEntityRemovalSourceReport):
 
     element_dm_edge: ElementDmEdgeReport = field(default_factory=ElementDmEdgeReport)
 
+    # DM customSQL element SQL parsing counters.
+    # Elements skipped before aggregator registration (no definition, missing /
+    # unknown connection, unsupported platform, duplicate source_id).
+    dm_customsql_skipped: int = 0
+    dm_customsql_aggregator_invocations: int = 0
+    dm_customsql_aggregator_invocation_error: int = 0
```
treff7es: ℹ️ **NIT — Counter naming parity**

`dm_customsql_aggregator_invocation_error` (singular) vs `dm_customsql_aggregator_invocations` (plural) are parallel concepts (success counter / failure counter). Pluralizing the failure to `..._errors` matches the success counter's tense and reads more naturally on dashboards.

metadata-ingestion/tests/unit/sigma/test_dm_customsql_parsing.py
```diff
@@ -0,0 +1,482 @@
...
+        assert extra_urn in upstream_datasets
+        # Non-customSQL FGL must survive the drain merge.
+        all_upstream_fields = {
+            f
+            for fgl in (aspect.fineGrainedLineages or [])
+            for f in (fgl.upstreams or [])
+        }
+        assert extra_field_urn in all_upstream_fields
+
+    def test_mixed_source_preflight_failure_emits_per_element_mcp(self) -> None:
```
treff7es: ℹ️ **NIT — Misleading test name**

`test_mixed_source_preflight_failure_emits_per_element_mcp` promises "emits per-element MCP" but the test only asserts `_ELEMENT_URN not in source._customsql_registered_urns` and `_ELEMENT_URN not in source._customsql_extra_upstreams`. It never drives the per-element emit path or asserts an MCP is emitted.

Either rename to `test_preflight_failure_does_not_stash_extras`, or extend the test to actually drive `_gen_data_model_element_upstream_lineage` and assert immediate emission.

metadata-ingestion/tests/unit/sigma/test_dm_customsql_parsing.py
```diff
@@ -0,0 +1,482 @@
...
+        source._process_dm_customsql_element(
+            _ELEMENT_URN,
+            {
+                "name": "q1",
+                "connectionId": _SNOWFLAKE_CONN.connection_id,
+                "definition": "SELECT FROM",  # invalid SQL
+            },
+        )
+        list(source._drain_sql_aggregators())
+        assert source.reporter.dm_customsql_parse_failed >= 1
```
treff7es: ℹ️ **NIT — Tighten `>= 1` to `== 1` where deterministic**

One element → one parse-failure expected. `>= 1` is defensively safe but loses the chance to catch double-counting regressions. Consider `== 1` here and in similar deterministic spots.

metadata-ingestion/src/datahub/ingestion/source/sigma/data_classes.py
kyungsoo-datahub: Done. CustomSqlEntry is already a Pydantic BaseModel with name: str (required, no default). The total=False TypedDict has been replaced.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma_api.py
kyungsoo-datahub: Done. `else` branch with self.report.warning(...) is in place. The warning includes dataModelId and the entry's keys so operators can identify the payload.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
kyungsoo-datahub: Done.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
kyungsoo-datahub: Done.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
kyungsoo-datahub: Done. Thank you for the comment.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
kyungsoo-datahub: Done.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
kyungsoo-datahub: Done. `test_select_star_with_computed_col_only_synthesizes_passthrough` covers this:  passthrough column lands in synthesised FGL, computed-expression columns (multi-ref) are excluded.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
kyungsoo-datahub: Done. Thanks.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
kyungsoo-datahub: Acknowledged as a valid suggestion. Skipping for now. The sort is a simple tuple comparison on strings, and any non-determinism would surface in golden file tests

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
kyungsoo-datahub: Done. Thanks.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
kyungsoo-datahub: Skipping for now. The branching is self-contained and named well enough inline. No behavior change, and the extraction would add indirection without reducing complexity.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma_api.py
kyungsoo-datahub: Done. Thanks.

metadata-ingestion/tests/unit/sigma/test_dm_customsql_parsing.py
kyungsoo-datahub: Done. Thank you for the comment.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma_api.py
kyungsoo-datahub: Done.  CustomSqlEntry is a Pydantic BaseModel; sigma_api.py uses CustomSqlEntry.model_validate(entry). No type: ignore remains.

metadata-ingestion/src/datahub/ingestion/source/sigma/config.py
kyungsoo-datahub: Done

metadata-ingestion/tests/unit/sigma/test_dm_customsql_parsing.py
kyungsoo-datahub: Done.

metadata-ingestion/tests/unit/sigma/test_dm_customsql_parsing.py
kyungsoo-datahub: Done.

treff7es (approved):

PR #17314, opened 2026-05-06
fix(ingest/snowflake): stages query parsing and handling edge cases

alokr-dhub:
# Summary                                                                                                  
                                                                        
  - Column-list COPY INTO targets — COPY INTO t(a, b) FROM @stage previously returned None because sqlglot
  wraps the target in a Schema node when a column list is present. The parser now unwraps it correctly.    
  - Multi-stage subquery sources — COPY INTO t FROM (SELECT ... FROM @s1 UNION ALL SELECT ... FROM @s2)
  previously captured at most one stage. ParsedCopyInto now holds stage_fqns: Tuple[str, ...], and         
  _extract_stage_references walks the full expression tree to collect all referenced stages. Each resolves
  to a separate upstream dataset in the lineage graph.                                                     
  - Better warning signal fidelity — three cases are now distinguished: 
    - Stage ref present but can't be normalised to a 3-part FQN → user-facing warning with the raw ref in  
  context                                                                                                  
    - COPY INTO starts with COPY but sqlglot can't parse it → warning with pipe FQN                        
    - Pipe body is not a COPY INTO at all (legal but unsupported) → silent debug log, no warning noise     
  - Shared constant — MAX_DEFINITION_LENGTH = 4000 extracted from snowflake_pipes.py into                  
  snowflake_utils.py so downstream files (tasks, etc.) can import it without duplication.                  
                                                                                                           
  # Test plan                                                                                                
                                                                        
  - test_target_with_column_list — column-list target resolves correctly                                   
  - test_from_subquery / test_column_list_and_subquery / test_subquery_with_qualified_stage_and_path —
  subquery forms                                                                                           
  - test_subquery_union_all_collects_multiple_stages / test_repeated_stage_in_subquery_dedups — two-way
  UNION ALL                                                                                                
  - test_three_way_union_all_collects_all_stages — three-way UNION ALL (set assertion; AST traversal order
  not guaranteed for 3+ branches)                                                                          
  - test_union_all_with_mixed_qualifications — 1-part, 2-part, 3-part stage refs in the same UNION
  - test_union_all_with_column_transforms_and_aliases — column casts, aliases, column-list target together 
  - test_union_all_one_branch_has_no_stage_still_collects_valid — plain-table branch doesn't suppress stage
   refs                                                                                                    
  - test_union_all_all_branches_unresolvable_returns_none /                                                
  test_union_all_mixed_resolvable_and_unresolvable — partial resolution                                    
  - test_complex_nested_query_with_metadata_columns_and_functions — real-world shape with
  metadata$filename, SPLIT_PART, MD5_NUMBER_LOWER64, TO_TIMESTAMP, etc.                                    
  - test_pipe_with_union_all_stages_emits_lineage_for_each — extractor emits one input dataset per stage,
  stage_name/stage_type custom properties correct                                                          
  - test_union_all_pipe_partial_stage_lookup_warns_and_emits_partial_lineage — 3 stages in COPY, 2 in
  lookup → 2 input datasets + unresolved-stage warning                                                     
  - test_malformed_copy_into_emits_warning / test_non_copy_definition_skipped_silently /
  test_unresolvable_stage_ref_emits_specific_warning — warning signal fidelity 


treff7es (commented):
lgtm, nice job

treff7es (approved):

PR #17318, opened 2026-05-06
feat(snowflake): extract dataset and column lineage from Snowflake Task SQL bodies

alokr-dhub:
## Summary

Snowflake Task SQL bodies are parsed to produce dataset-level and column-level lineage on the task's `DataJobInputOutput` aspect, alongside the predecessor DAG edges it already carried.

Parsing goes through the same statement-splitting path as stored procedures (`parse_procedure_code`), so:

- `INSERT`, `MERGE` and `CREATE TABLE AS SELECT` bodies emit `inputDatasets`, `outputDatasets` and column-level `fineGrainedLineages`
- **multi-statement bodies** are split and each statement parsed, with the results combined into one aspect
- **`CALL <procedure>`** resolves to the called procedure's own DataJob, via the `resolve_procedure_urn` hook added in #19380 — so the edge carries the procedure's argument-signature hash and the connector's identifier casing, and points at an entity this run actually emitted
- temp/staging tables collapse through, using the source's `_is_temp_table` and a shared session id

Dataset lineage is gated on `include_table_lineage` and the column half on `include_column_lineage`. The job-to-job edges are DAG structure, not table lineage, so they are not gated by either.

### Reporting

A task body has three possible outcomes, and they are now distinguishable:

| Counter | Meaning |
|---|---|
| `tasks_failed` | task skipped entirely — no aspects emitted |
| `tasks_with_sql_parse_failures` | body failed to parse; task still fully ingested, warning raised with the parse error |
| `tasks_without_sql_lineage` | nothing lineage-bearing in the body (`COPY INTO`, `TRUNCATE`, `ALTER TASK`) — counted, deliberately not warned |

That last distinction matters: `COPY INTO` is a canonical Snowflake task body, so warning on it would fire on nearly every run, while staying silent about real parse failures would hide the feature no-oping.

Predecessors excluded by `task_pattern`, and `CALL` targets excluded by `database_pattern` / `schema_pattern` / `procedure_pattern`, produce **no edge** rather than an edge to a DataJob that was never emitted — each reported under its own warning title.

### Also here

`_get_stages_tasks_pipes_workunits` now runs after `discovered_datasets` is populated. Before that point `_is_temp_table` silently no-ops, which would emit a task's staging table as a real dataset instead of collapsing through it; the invariant is enforced in `_is_temp_table` itself rather than by call order. A failed Snowflake edition lookup warns instead of passing silently.

## Stack

Split so the shared/framework changes can be reviewed apart from the connector:

1. #19379 — public `SqlParsingAggregator.schema_resolver` accessor
2. #19380 — stored-procedure CALL-target resolution and parse reporting (cross-connector)
3. **this PR** — the Snowflake task lineage feature, based on #19380

Previous head before the restructure: `2dc7342028d20c73af0f402e42facdeddf60e07f`.

## Test plan

Unit (`tests/unit/snowflake/test_snowflake_tasks.py`, 46 tests) covers dataset/column/multi-statement lineage, both config gates, predecessor resolution (partially-qualified, quoted, cross-schema collisions, filtered-out), procedure-call resolution (casing, overloads by arity, defaulted parameters, commas inside types and default literals, ambiguity, uningested and filtered-out targets, `include_procedures` off), temp-table collapsing, and each report counter. Procedure urns are asserted by equality against `BaseProcedure.to_urn`, never by substring.

Integration (`tests/integration/snowflake/test_snowflake_stages_tasks_pipes.py`) adds a lineage-enabled golden plus four pipeline-level tests: parse failure reported with the task still ingested, no-lineage bodies counted and not warned, `task_pattern`-filtered predecessor emitting no edge, and `procedure_pattern`-filtered CALL emitting no edge. The last two carry positive controls so they cannot pass vacuously.

- [x] 1,366 unit tests + 47 Snowflake integration tests pass
- [x] `./gradlew :metadata-ingestion:lint` clean (ruff + mypy)
- [x] Verified end-to-end against a live Snowflake account: the CALL-derived edge resolves to `...stored_procedures,PROD),LOAD_REGIONAL_1aefcd00833a35307d2a057f0b387851)`, byte-identical to the urn the same run emits for that procedure

## Checklist

- [x] Docs updated (`snowflake_post.md`: the config dependencies and the CALL prerequisites)
- [ ] Breaking changes — none


metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
```diff
@@ -204,3 +255,126 @@ def _gen_data_job(
                     ]
                 ),
             ).as_workunit()
+
+    def _parse_task_definition_for_lineage(
```
alfiyas-datahub, line 487: This emits task table and column lineage whenever `include_tasks` is enabled, even if `include_table_lineage` or `include_column_lineage` are disabled. Since those config flags are respected elsewhere (and `include_column_lineage` even depends on `include_table_lineage`), this feels a bit unexpected. Should task lineage also respect these flags? The extractor already has access to `self.config`, so adding the check should be straightforward.

metadata-ingestion/tests/unit/snowflake/test_snowflake_tasks.py
```diff
@@ -231,6 +266,408 @@ def test_task_name_map_includes_all_tasks_for_predecessor_resolution(self) -> No
...
+            name="proc_task",
+            definition="CALL my_proc('arg1', 'arg2')",
+        )
+        wus, report = _collect_workunits([task])
+
+        ios = _data_job_input_outputs(wus)
+        # No predecessors, no parsed datasets → no DataJobInputOutput.
+        assert len(ios) == 0
+        messages = [w.message for w in report.warnings]
+        assert any("task definition" in m.lower() for m in messages), (
```
alfiyas-datahub: 1. The phrase "task definition" appears in both warning messages, so it's not clear which one the `CALL` case is triggering. Could we assert on `w.title` as well to make the test more specific?

2. The PR mentions skipping both `CALL` statements and multi-statement bodies, but this test only covers `CALL`. Since multi-statement SQL goes through a different `sqlglot` path (`ParseError` instead of `UnsupportedStatementTypeError`), it would be good to add a separate test for that case too.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
alokr-dhub: Agree. I think include_column_lineage should be checked as well

metadata-ingestion/tests/unit/snowflake/test_snowflake_tasks.py
alokr-dhub: ACK

alokr-dhub (commented):
addressing the review comments

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
```diff
@@ -204,3 +250,128 @@ def _gen_data_job(
...
```
treff7es, line 487: **Reuse the aggregator instead of calling `sqlglot_lineage` directly.**

A Snowflake task is the same shape as a stored procedure: a `DataJob` whose lineage
comes from a SQL body. Stored procedures (MSSQL, MySQL) don't parse directly — they
feed statements into a `SqlParsingAggregator` and convert the result with
`to_datajob_input_output` (`datahub/sql_parsing/datajob.py`), which reads the
aggregator's `UpstreamLineage` MCPs and repackages them as a `DataJobInputOutputClass`,
fine-grained lineages included. So the premise in the PR description — that the
aggregator produces the wrong aspect and therefore must be bypassed — doesn't hold; that
conversion is what the helper exists for.

Going direct is also strictly less capable than the existing framework:

| | direct `sqlglot_lineage` (this PR) | `parse_procedure_code` |
|---|---|---|
| single-statement INSERT/MERGE/CTAS | ✅ | ✅ |
| multi-statement bodies | ❌ warns, drops all lineage | ✅ splits & parses each |
| `CALL <proc>` | ❌ warns, no lineage | ✅ resolves to dataJob→dataJob edge |
| temp/transient tables | ❌ no session tracking | ✅ shared `session_id` + `is_temp_table` |

Snowflake tasks are frequently multi-statement or `CALL a_proc()` — the cases this PR
lists as "skipped with a warning" are exactly the ones the framework already extracts.

Suggested approach: call `parse_procedure_code(...)` from
`sql/stored_procedures/lineage.py`, passing the shared `schema_resolver`,
`default_db=db_name`, `default_schema=schema_name`, the task body, and snowflake's
`is_temp_table`. Its TSQL control-flow filtering is gated on the tsql dialect, so it's a
no-op for snowflake. That deletes `_parse_task_definition_for_lineage`,
`_run_sql_parser`, `_build_fine_grained_lineages`, and `TaskLineage` (~130 lines) and
gains multi-statement + CALL + temp-table handling.

If reaching into the stored-procedures module feels too coupled, the minimum is a
per-task `SqlParsingAggregator` (sharing the main `schema_resolver`) +
`to_datajob_input_output` — which still removes the manual fine-grained construction below.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
```diff
@@ -204,3 +250,128 @@ def _gen_data_job(
...
+                title="Task SQL Parsing Failed",
+                message="Failed to parse task definition for lineage",
+                context=task_fqn,
+                exc=e,
+            )
+            return None
+
+        return result
+
+    def _build_fine_grained_lineages(
```
treff7es: This manual `col_lineage → make_schema_field_urn → FineGrainedLineageClass` mapping
(plus the drop/partial-drop bookkeeping) re-implements what the aggregator already
produces. `to_datajob_input_output` harvests `fineGrainedLineages` straight off the
aggregator's `UpstreamLineage` aspect, so if you switch to the aggregator path (see the
comment on `_parse_task_definition_for_lineage`) this whole method goes away.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
```diff
@@ -20,24 +21,42 @@
...
 
-# Truncate the task definition stored in customProperties to stay well within
-# DataHub's aspect size limits.
-_MAX_DEFINITION_LENGTH = 4000
+
+@dataclass
+class TaskLineage:
+    """Dataset-level inputs/outputs and column lineage parsed from a task body."""
+
+    input_datasets: Tuple[Urn, ...] = field(default_factory=tuple)
```
treff7es: Prefer `List[Urn]` / `List[FineGrainedLineageClass]` here. `set` would be wrong (order
matters for deterministic output, and `FineGrainedLineageClass` isn't hashable), but
`Tuple`'s only edge over `List` is immutability, which is unused — the value is built
once and immediately converted with `list(...)` at the emission boundary. `List` is the
idiomatic choice in this codebase.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
```diff
@@ -181,14 +210,31 @@ def _gen_data_job(
                 pred_job_id = self.identifiers.snowflake_identifier(simple_name)
                 pred_job_urn = make_data_job_urn_with_flow(flow_urn, pred_job_id)
                 input_datajobs.append(pred_job_urn)
+            else:
+                unresolved_predecessors.append(predecessor_name)
```
treff7es: Latent (pre-existing) collision: `simple_name` takes the last dotted segment, so a
fully-qualified cross-schema predecessor like `other_db.other_schema.upstream_task`
whose leaf name matches a task in the *current* schema named `upstream_task` will
false-match `task_name_map` and emit a wrong intra-schema `inputDatajobs` edge instead
of landing here as unresolved. Low likelihood, but this path now depends on the
leaf-split — worth a comment or a follow-up.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_v2.py
```diff
@@ -754,6 +754,7 @@ def _get_stages_tasks_pipes_workunits(
                 report=self.report,
                 data_dictionary=self.data_dictionary,
                 identifiers=self.identifiers,
+                schema_resolver=self.aggregator._schema_resolver,
```
treff7es: Reaching into the aggregator's private `_schema_resolver`. It matches existing practice
(there's no public accessor — even the aggregator's own report code does this), so not a
blocker, but a public `schema_resolver` property on `SqlParsingAggregator` would be a
clean tiny follow-up.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_v2.py
```diff
@@ -941,8 +947,8 @@ def _snowflake_clear_ocsp_cache(self) -> None:
         try:
             if os.path.exists(file_path):
                 os.remove(file_path)
-        except Exception:
-            logger.debug(f'Failed to remove OCSP cache file at "{file_path}"')
+        except Exception as e:
+            logger.warning('Failed to remove OCSP cache file at "%s": %s', file_path, e)
```
treff7es: `debug` → `warning` for OCSP cache removal may get noisy — a missing cache file is an
expected, benign condition, and this would surface on every such run. The edition-detection
change above (→ `report.warning`) is a good improvement; this one I'd leave at `debug`.
Also both of these `snowflake_v2.py` changes are unrelated to task lineage — mild scope
creep for the PR.

metadata-ingestion/tests/integration/snowflake/test_snowflake_stages_tasks_pipes.py
```diff
@@ -136,3 +136,61 @@ def test_snowflake_tasks_only(pytestconfig, tmp_path, mock_time, mock_datahub_gr
     assert report.stages_scanned == 0
     assert report.tasks_scanned == 3
     assert report.pipes_scanned == 0
+
+
+def test_snowflake_task_lineage_extracted_end_to_end(
```
treff7es, line 184: This programmatic test is the *only* coverage for the new task lineage — the golden test
above runs on `_base_config` with lineage disabled, so the golden's task
`dataJobInputOutput` entries stay empty. The golden's only diff vs master is a `runId`
inside `systemMetadata` (an ignored path), so the PR description's claim that the golden
was "updated with task lineage entries" isn't accurate. Suggest reverting the golden
`runId` churn and either fixing the description or adding a lineage-enabled golden so the
aspect shape is snapshotted.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
alokr-dhub: Fixed in f6261f94 — task lineage now returns early from `_parse_task_definition_for_lineage` unless `include_table_lineage` is set, and `include_column_lineage` is checked separately to strip `fineGrainedLineages` when column lineage is disabled but table lineage isn't. Covered by `test_lineage_skipped_when_include_table_lineage_disabled` and `test_column_lineage_skipped_when_include_column_lineage_disabled`.

metadata-ingestion/tests/unit/snowflake/test_snowflake_tasks.py
alokr-dhub: Superseded by the refactor in f6261f94 (see the thread on `_parse_task_definition_for_lineage` below) — the whole direct-`sqlglot_lineage` code path these tests were asserting on is gone, so I rewrote the test file rather than patching these two. The CALL case is now asserted via title (`test_call_only_task_emits_warning_and_no_dataset_lineage`), and multi-statement bodies get their own positive test (`test_multi_statement_task_emits_combined_lineage") since they now produce real combined lineage instead of a warning.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
alokr-dhub: Done in f6261f94 — switched to `parse_procedure_code` + the shared aggregator, deleted `TaskLineage`, `_run_sql_parser`, and `_build_fine_grained_lineages` (~150 lines). Went with the direct `parse_procedure_code` call rather than the "minimum" per-task-aggregator alternative since it's less code and already handles statement splitting/filtering. Net effect: multi-statement task bodies now produce real combined lineage (see `test_multi_statement_task_emits_combined_lineage`) instead of being skipped with a warning. One trade-off: the granular table-error vs column-error warning distinction from the old direct-call path collapsed into a single generic "Task Lineage Extraction Failed" warning, since that granularity lived in the code that got deleted.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
alokr-dhub: This whole method is gone as of f6261f94 — deleted along with `_run_sql_parser` when the extractor switched to `parse_procedure_code`/`to_datajob_input_output`.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
alokr-dhub: Moot — `TaskLineage` (and its Tuple fields) no longer exists after the f6261f94 refactor; `parse_procedure_code` returns a `DataJobInputOutputClass` directly.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
alokr-dhub: Fixed in f021ec2e — a fully-qualified predecessor now only matches the current schema's task_name_map if its db/schema segments actually equal the current db_name/schema_name; otherwise it falls through to unresolved even if the leaf name collides. Added `test_predecessor_fully_qualified_cross_schema_name_collision_is_unresolved` to cover it.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_v2.py
alokr-dhub: Added a public `SqlParsingAggregator.schema_resolver` property in f021ec2e and switched both `snowflake_v2.py` call sites (queries extractor + tasks extractor) off `_schema_resolver`.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_v2.py
alokr-dhub: Reverted to `logger.debug` in f021ec2e — kept the edition-detection warning improvement above as-is since that one's a real improvement, agreed this one would just be noise.

metadata-ingestion/tests/integration/snowflake/test_snowflake_stages_tasks_pipes.py
alokr-dhub: Reverted the golden file's `runId`-only diff in f021ec2e (it really was pure snapshot churn against an ignored path). `test_snowflake_task_lineage_extracted_end_to_end` remains the actual coverage for task lineage shape — I left the PR description as-is since it already references that test by name, but happy to tighten the wording further if you think it's still misleading.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
```diff
@@ -172,24 +181,59 @@ def _gen_data_job(
...
+            # task list if it actually points at this db/schema; otherwise a
+            # same-named task in another schema would wrongly match on leaf name
+            # alone.
+            is_current_schema = len(pred_name_parts) == 1 or (
+                len(pred_name_parts) == 3
+                and pred_name_parts[0] == db_name.upper()
+                and pred_name_parts[1] == schema_name.upper()
+            )
+            if is_current_schema and simple_name in task_name_map:
                 pred_job_id = self.identifiers.snowflake_identifier(simple_name)
```
treff7es: **Pre-existing bug, but the fix is already in your hand.** `simple_name` is the *uppercased predecessor literal*, while the task's own `job_id` at line 143 comes from `snowflake_identifier(task.name)`. With the default `convert_urns_to_lowercase=True` these coincide; with it `False` and a lowercase task name, the predecessor edge points at `...,TASK_A)` while the task is emitted as `...,task_a)` — dangling.

`task_name_map` is `Dict[str, SnowflakeTask]` but only ever used for membership (line 200), so the values are currently dead weight. `snowflake_identifier(task_name_map[simple_name].name)` fixes the bug and puts them to work.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
```diff
@@ -204,3 +248,41 @@ def _gen_data_job(
...
+        """Parse task SQL to extract dataset-level inputs/outputs and column lineage.
+
+        Delegates to the same statement-splitting + SqlParsingAggregator path used
+        for stored procedures, so multi-statement task bodies are handled naturally
+        instead of being rejected outright.
+        """
+        if not task.definition or not self.config.include_table_lineage:
+            return None
+
+        datajob_input_output = parse_procedure_code(
```
treff7es: **BLOCKER — this `CALL` edge always points at a nonexistent DataJob.**

Produced here vs. what Snowflake actually emits for the same procedure (`snowflake_golden.json`):
```
this:      ...(snowflake,TEST_DB.TEST_SCHEMA.stored_procedures,PROD),MY_PROCEDURE)
reality:   ...(snowflake,test_db.test_schema.stored_procedures,PROD),my_procedure_d72dbca43341d7a41a7e01b76e2a25d4)
```
Three mismatches:
1. **Flow-name case** — `db_name`/`schema_name` arrive uppercase from Snowflake and never pass through `self.identifiers.snowflake_identifier`. Dataset URNs get normalized via the aggregator; this path bypasses it.
2. **Job-id case** — `call.name` comes verbatim from SQL text.
3. **Missing `_<arg_hash>`** — `argument_signature` is populated from `information_schema.procedures.ARGUMENT_SIGNATURE` (`snowflake_schema.py:2074`), which returns `"()"` even for zero-arg procs, so `get_procedure_identifier()` *always* hashes. Both golden procedure URNs are hashed.

So even with casing fixed the edge can't match. Also nothing checks `include_procedures`.

**Fix (pick one):** resolve the call target against the registered procedure URN in shared code (also fixes Oracle/MSSQL), or normalize case via `snowflake_identifier` and skip CALL-derived edges for Snowflake until the hash is resolvable.

The CALL code is master's, not yours — but this PR routes task bodies into it and documents it as working.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
```diff
@@ -204,3 +248,41 @@ def _gen_data_job(
...
+            schema_resolver=self.schema_resolver,
+            default_db=db_name,
+            default_schema=schema_name,
+            code=task.definition,
+            is_temp_table=self.is_temp_table,
+            procedure_name=task_fqn,
+        )
+
+        if datajob_input_output is None:
```
treff7es, line 539: **BLOCKER — warns on valid task bodies; real failures stay silent.**

`parse_procedure_code` returns `None` both for real failures and for "nothing lineage-bearing here" (`lineage.py:393-398`, debug-only). This warns on both. 15 of 20 realistic bodies I probed hit it:
```
COPY INTO my_tbl FROM @my_stage    TRUNCATE TABLE my_tbl
ALTER TASK other RESUME            DELETE FROM my_tbl WHERE ...
INSERT INTO tgt(a) VALUES (1)      CREATE OR REPLACE STREAM s ON TABLE t
```
`COPY INTO` is the canonical Snowflake task body. So on nearly every run with `include_tasks: true` operators see "Failed to extract lineage" when nothing failed.

Meanwhile actual failures are invisible: `lineage.py:290-294`/`:310-314` log at DEBUG with only `type(e).__name__`; `raise_` isn't passed so it defaults `False`; and the throwaway aggregator's report — which records `observed_query_parse_failures` / `num_observed_queries_failed` / `num_observed_queries_column_failed` — is discarded. For `DELETE FROM tgt WHERE a>1` it captured a precise column-lineage error and this warning went out with no `exc=`.

**Fix:**
```python
if datajob_input_output is None:
    self.report.tasks_without_sql_lineage += 1
    logger.debug(f"No lineage-bearing statements in task body: {task_fqn}")
    return None
```
(new counter on `SnowflakeV2Report`), and reserve `report.warning(title="Task SQL Parse Failure", ..., exc=...)` for the case where DML was present and failed — which needs the sub-report surfaced out of `parse_procedure_code`. That also restores the table-vs-column distinction you noted losing.

metadata-ingestion/tests/unit/snowflake/test_snowflake_tasks.py
```diff
@@ -231,6 +289,237 @@ def test_task_name_map_includes_all_tasks_for_predecessor_resolution(self) -> No
...
+        io = ios[0]
+        assert io.inputDatasets and len(io.inputDatasets) == 2
+        assert io.outputDatasets and len(io.outputDatasets) == 2
+        assert any("source_a" in d for d in io.inputDatasets)
+        assert any("source_b" in d for d in io.inputDatasets)
+        assert any("target_a" in d for d in io.outputDatasets)
+        assert any("target_b" in d for d in io.outputDatasets)
+        assert not report.warnings
+
+    def test_call_only_task_resolves_to_procedure_job_edge(self) -> None:
```
treff7es, line 749: **BLOCKER (test) — these assertions pass while the URN is unusable.**

`"my_proc" in ...` and `"stored_procedures" in ...` can't catch the mismatch flagged on `snowflake_tasks.py`.

**Fix:** assert URN equality against the real thing — build `BaseProcedure(name="my_procedure", argument_signature="(arg1 VARCHAR)")`, compute `procedure.to_urn(database_key, schema_key)` with the keys the source uses, and `assert io.inputDatajobs[0] == that_urn`. One line, catches all three mismatches.

Two cases to add:
- **Uppercase definition** (`CALL TEST_DB.TEST_SCHEMA.MY_PROCEDURE('a')`) — every definition here is lowercase, which is why the casing bug is invisible.
- **`EXECUTE IMMEDIATE`** — assert it yields no `inputDatajobs` (see review summary).

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_v2.py
```diff
@@ -748,6 +748,8 @@ def _get_stages_tasks_pipes_workunits(
                 report=self.report,
                 data_dictionary=self.data_dictionary,
                 identifiers=self.identifiers,
+                schema_resolver=self.aggregator.schema_resolver,
+                is_temp_table=self._is_temp_table,
```
treff7es, line 893: **This wiring is half-inert.** `_get_stages_tasks_pipes_workunits` is called at line 570; `self.discovered_datasets` isn't assigned until line 615. `_is_temp_table` (`:493-511`) has two arms, and the "allowed by patterns but not in `discovered_datasets`" arm can never fire during task parsing.

Effect: a task staging through an undiscovered scratch table emits lineage *to* that phantom dataset instead of collapsing through it — the exact case this callable is for. Queries-v2 works because it runs after line 615.

**Fix:** move the `include_tasks` block to after `self.discovered_datasets` is set. The bound method is already lazy, so ordering is the only change.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
```diff
@@ -172,24 +181,59 @@ def _gen_data_job(
...
+                datajob_input_output = DataJobInputOutputClass(
                     inputDatasets=[],
                     outputDatasets=[],
                     inputDatajobs=input_datajobs,
-                ),
+                )
+            else:
+                datajob_input_output.inputDatajobs = (
+                    list(datajob_input_output.inputDatajobs or []) + input_datajobs
+                )
```
treff7es: **Third copy of this merge, and the only one without dedup.** The others: `lineage.py:446-458` and `stored_procedures/base.py:254-267` — the latter is `additional_input_jobs`, which already models "caller has authoritative `inputDatajobs` to merge with CALL-derived ones", dedup included.

Today a task listing both `TASK_A` and `DB.SCHEMA.TASK_A` appends the same URN twice (both resolve under the new `is_current_schema` logic).

**Fix:** thread `additional_input_jobs` through `parse_procedure_code` and pass `input_datajobs` in. This whole block becomes:
```python
datajob_input_output = self._parse_task_definition_for_lineage(...)  # takes input_datajobs
if datajob_input_output is not None:
    yield MetadataChangeProposalWrapper(entityUrn=job_urn, aspect=datajob_input_output).as_workunit()
```
If you'd rather not touch the shared signature, at least flatten the three-branch `None` dance and dedup.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
```diff
@@ -172,24 +181,59 @@ def _gen_data_job(
...
+            simple_name = pred_name_parts[-1]
+            # A fully-qualified name only resolves against the current schema's
+            # task list if it actually points at this db/schema; otherwise a
+            # same-named task in another schema would wrongly match on leaf name
+            # alone.
+            is_current_schema = len(pred_name_parts) == 1 or (
+                len(pred_name_parts) == 3
+                and pred_name_parts[0] == db_name.upper()
+                and pred_name_parts[1] == schema_name.upper()
+            )
```
treff7es: Correct for the case it targets, but `snowflake_utils.split_qualified_name` (`:275-312`) already does this — quote-aware, doctested, same module you import `MAX_DEFINITION_LENGTH` from.

Two gaps here:
- **Quoted identifiers**: `.upper()` applies *through* the quotes, so `'"MYDB"' != 'MYDB'` and this branch never matches. Pre-existing for the leaf name, but the path went from silent-skip to warning, so it's now loud.
- **2-part names**: `SCHEMA.TASK` in the current schema falls through to unresolved+warning where leaf matching resolved it.

**Fix:** use `split_qualified_name`, and either add a `len == 2 and parts[0] == schema_name.upper()` arm or comment that 2-part is deliberately unsupported. Cleaner still: key `task_name_map` on the full FQN and normalize the reference to 3 parts first — the idiom `snowflake_pipes.py:61-80` (`_stage_reference_to_fqn`) already uses. That drops the boolean entirely.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
treff7es: **Pre-existing bug; the fix is already at hand.** `simple_name` is the uppercased *predecessor literal*, while the task's own `job_id` (line 143) uses `snowflake_identifier(task.name)`. With `convert_urns_to_lowercase=False` and a lowercase task name the edge points at `...,TASK_A)` while the task is emitted as `...,task_a)` — dangling.

**Fix:** `snowflake_identifier(task_name_map[simple_name].name)`. `task_name_map` is `Dict[str, SnowflakeTask]` but only used for membership today, so this also puts its values to work.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
```diff
@@ -84,14 +85,21 @@ def get_workunits(
...
+                    schema_name=schema_name,
+                    task_name_map=task_name_map,
+                )
+            except Exception as e:
+                self.report.warning(
+                    title="Task Extraction Failed",
+                    message="Failed to extract metadata for task; skipping remaining aspects",
+                    context=f"{db_name}.{schema_name}.{task.name}",
+                    exc=e,
+                )
```
treff7es, line 205: This does catch mid-iteration exceptions (`yield from` re-raises inside the `try` — verified), but it's placed too wide and costs an aspect.

`DataJobInfo` (:161), `SubTypes` (:171), `Status` (:178) are yielded *before* the parse at :217; `Ownership` at :239 *after*. So a throw in `parse_procedure_code` leaves the task with no `DataJobInputOutput` **and no `Ownership`** — silently dropping ownership a previous run had.

Only `_parse_task_definition_for_lineage` can realistically throw; the three MCPWs are static constructions.

**Fix:** narrow the guard to that call, inside `_parse_task_definition_for_lineage`, so the other four aspects always emit and the warning can name the real failure. Keep this outer guard as belt-and-braces if you like — it should then be unreachable.

Minor: `tasks_scanned += 1` at :87 counts failed tasks as scanned; a `tasks_failed` counter would make the report readable.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
```diff
@@ -204,3 +248,41 @@ def _gen_data_job(
...
+        db_name: str,
+        schema_name: str,
+    ) -> Optional[DataJobInputOutputClass]:
+        """Parse task SQL to extract dataset-level inputs/outputs and column lineage.
+
+        Delegates to the same statement-splitting + SqlParsingAggregator path used
+        for stored procedures, so multi-statement task bodies are handled naturally
+        instead of being rejected outright.
+        """
+        if not task.definition or not self.config.include_table_lineage:
```
treff7es: Gating on `include_table_lineage` also suppresses the `CALL` -> procedure `inputDatajobs` edge, which is DataJob lineage, not table lineage. Inconsistent two ways:
- Predecessor-derived `inputDatajobs` on the *same aspect* (lines 183-215) are ungated.
- `snowflake_schema_gen.py:2912-2931` runs this same path for procedures gated only on `include_procedures`. So `include_table_lineage: false` keeps procedure->procedure edges but kills task->procedure ones.

`include_table_lineage`'s description (`snowflake_config.py:408-411`) is about table-to-table lineage only.

**Fix:** move the gate so it strips `inputDatasets`/`outputDatasets`/`fineGrainedLineages` and leaves the DataJob edge intact — or make it a deliberate, commented decision.

(The `fineGrainedLineages = None` strip below is fine: `bigquery_v2/lineage.py:417` sets precedent, and `validate_include_column_lineage` only forbids column-on/table-off, so the runtime strip is needed for the valid table-on/column-off state.)

metadata-ingestion/tests/integration/snowflake/test_snowflake_stages_tasks_pipes.py
```diff
@@ -136,3 +136,61 @@ def test_snowflake_tasks_only(pytestconfig, tmp_path, mock_time, mock_datahub_gr
...
+
+    root_task_io = next(
+        e["aspect"]["json"]
+        for e in events
+        if e.get("aspectName") == "dataJobInputOutput"
+        and e["entityUrn"].endswith("root_task)")
+    )
+    assert any("table_2" in u for u in root_task_io["inputDatasets"])
+    assert any("table_1" in u for u in root_task_io["outputDatasets"])
+    assert root_task_io["fineGrainedLineages"]
```
treff7es: **This should be a golden file.** Output is deterministic (same mocked cursor, same `mock_time`) and the sibling test in this file already uses `check_golden_file` with `ignore_paths`. This was last round's comment; I still think it was right.

What hand-picking loses:
- `assert root_task_io["fineGrainedLineages"]` is truthiness only — never checks a column pair or the upstream/downstream URNs.
- Nothing verifies the *absence* of spurious events.
- `child_task_1`'s `inputDatajobs` is never inspected — which is exactly where the dangling-URN blocker lives. A golden would have put `...,MY_PROCEDURE)` in the diff.

**Fix:** snapshot this run with `check_golden_file`. At minimum, extend lines 194-196 to assert on `child_task_1_io["inputDatajobs"]` and cross-check against the procedure `dataJob` URNs the same run emits.

metadata-ingestion/tests/unit/snowflake/test_snowflake_tasks.py
```diff
@@ -73,6 +94,7 @@ def _collect_workunits(
         report=report,
         data_dictionary=data_dict,
         identifiers=identifiers,
+        schema_resolver=_make_schema_resolver(config),
```
treff7es, line 159: `_collect_workunits` never passes `is_temp_table`, so every test runs the `lambda _: False` default and temp-table filtering is completely uncovered. That's what let the ordering gap in `snowflake_v2.py` through.

**Fix:** add one test — `CREATE TEMPORARY TABLE tmp_a AS SELECT * FROM src; INSERT INTO tgt SELECT * FROM tmp_a` with `is_temp_table=lambda name: "tmp_" in name`, asserting `tmp_a` is in neither `inputDatasets` nor `outputDatasets` while `src` -> `tgt` survives. That shared-`session_id` temp-table resolution is the benefit you cite for delegating to `parse_procedure_code`, and nothing currently proves it works.

metadata-ingestion/docs/sources/snowflake/snowflake_post.md
```diff
@@ -428,6 +428,10 @@ stage_pattern:
 
 Tasks are ingested as DataJob entities grouped under a per-schema DataFlow. Predecessor dependencies between tasks are captured as `inputDatajobs` on the DataJobInputOutput aspect, preserving the DAG structure.
 
+Task SQL bodies are also parsed with sqlglot to extract dataset-level and column-level lineage, using the same statement-splitting pipeline as stored procedure lineage. `INSERT`, `MERGE`, and `CREATE TABLE AS SELECT` statements produce `inputDatasets`, `outputDatasets`, and fine-grained (column-to-column) lineage on the DataJobInputOutput aspect, including multi-statement bodies. `CALL <procedure>` statements resolve to an `inputDatajobs` edge pointing at the called procedure's DataJob, matching how procedure-to-procedure calls are resolved for stored procedures.
```
treff7es: Two fixes needed:

1. **State the config dependency.** All of this needs `include_table_lineage: true`, and the column-level half also `include_column_lineage: true`. As written, a reader assumes `include_tasks: true` suffices. The connector's capability annotations (`snowflake_v2.py:126,130`) already disclose this for regular table lineage — match that.

2. **Soften the `CALL` claim.** "matching how procedure-to-procedure calls are resolved for stored procedures" is true of the mechanism but not the outcome — the URN carries no argument-signature hash and isn't case-normalized, so it never matches the procedure DataJob this connector emits. Until that's fixed this shouldn't read as working.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
```diff
@@ -32,20 +34,19 @@
...
-
 
 @dataclass
 class SnowflakeTasksExtractor:
     config: SnowflakeV2Config
     report: SnowflakeV2Report
     data_dictionary: SnowflakeDataDictionary
     identifiers: SnowflakeIdentifierBuilder
+    schema_resolver: SchemaResolver
+    is_temp_table: Callable[[str], bool] = field(default=lambda _: False)
```
treff7es: No runtime hazard — `@dataclass.__init__` always assigns the instance attribute, so there's no implicit-binding trap (verified). But `field(default=...)` is exactly equivalent to a bare `= lambda _: False` and is the only occurrence of that form in the repo, so it reads as a workaround for a nonexistent problem.

The default is the real issue: the one production caller always passes `self._is_temp_table`, so it exists only so tests don't have to.

**Fix:** make it required, like `schema_resolver` above. Tests then pass a real predicate — which is how the `discovered_datasets` ordering gap would have surfaced.

metadata-ingestion/src/datahub/sql_parsing/sql_parsing_aggregator.py
```diff
@@ -637,6 +637,10 @@ def close(self) -> None:
         self._closed = True
         self._exit_stack.close()
 
+    @property
+    def schema_resolver(self) -> SchemaResolver:
```
treff7es, line 641: Good addition — right shape, genuine extension point.

Two follow-ups:
- Add a one-line docstring on ownership: the resolver is registered in the aggregator's `_exit_stack`, so callers must not close it or use it after `close()`. Right now the accessor silently blesses use-after-close.
- Finish the migration: `snowflake_schema_gen.py:1960` and `:2929` still use `self.aggregator._schema_resolver` (the first with a `# NOTE: This uses private members...` comment), as does `sql_common.py:1560`. The two Snowflake ones are one-liners — worth sweeping here; leave `sql_common.py` for a shared-code PR. Two ways to do the same thing in one connector is worse than either alone.

alokr-dhub:
### Restructured into a stacked PR

Per review feedback that the shared/framework changes shouldn't be reviewed inside a connector PR, this has been split into three:

1. #19379 — `refactor(ingest): expose SqlParsingAggregator.schema_resolver`
2. #19380 — `feat(ingest/sql): resolve stored-procedure CALL targets and report parse failures` (the cross-connector changes: `resolve_procedure_urn`, `EXECUTE IMMEDIATE`, parse reporting, aggregator lifetime)
3. **this PR** — the Snowflake task lineage feature only, now based on #19380

This PR's branch has been force-pushed and rebased onto #19380 and current master (it was 291 commits behind). **The previous head was `2dc7342028d20c73af0f402e42facdeddf60e07f`** — the commits referenced in earlier replies (`f6261f94`, `f021ec2e`, `28ba8e3b`) remain reachable from that SHA and from the force-push entry in this PR's timeline.

All existing review threads are intact; GitHub marks the ones whose lines moved as outdated. The follow-up fixes from the last round of review are in the new head, including the five Bugbot findings that had no reply yet:

- **Procedure URN casing** — schema and procedure name now come from the catalog, not the CALL text (the previously computed `database` was dead code)
- **CALL edges ignoring procedure filters** — new `_procedure_in_scope` applies `database_pattern` / `schema_pattern` / `procedure_pattern`, routed through `SnowflakeFilter` the way `snowflake_schema_gen` does
- **Arity rejecting defaulted calls** — the declared signature is now a `(min, max)` range (`DEFAULT` params may be omitted), split paren- and quote-aware so `NUMBER(10,2)` and `DEFAULT ','` don't inflate the count; arity is an overload tiebreaker rather than a veto on a lone match
- **Yield inside `except`** — `_gen_data_job` is drained into a list inside the guard and yielded outside it
- **Silent procedure-list failure** — now `report.warning(title="Procedure Lookup Failed", ..., exc=e)`

And on @treff7es' "real failures stay silent" point, the half that was still open: `parse_procedure_code` now hands back a `ProcedureParseReport`, so a body that failed to parse raises `Task SQL Parse Failure` with the error text and increments `tasks_with_sql_parse_failures`, while a body with nothing lineage-bearing in it stays a quiet `tasks_without_sql_lineage` count. That also restores the table-vs-column distinction. Two further findings from re-review are fixed here: predecessor edges to `task_pattern`-filtered tasks are dropped (they pointed at DataJobs that were never emitted) and reported separately, and the `EXECUTE IMMEDIATE` test is no longer vacuous.

Verified against a live Snowflake account, not just mocks: the `CALL`-derived edge resolves to `...stored_procedures,PROD),LOAD_REGIONAL_1aefcd00833a35307d2a057f0b387851)` — byte-identical to the urn the same run emits for that procedure, argument-signature hash included.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_v2.py
alokr-dhub: Moved to #19379 in the restructure — the public property and all three Snowflake call sites live there now, so nothing in this PR touches `_schema_resolver`.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
alokr-dhub: Fixed, and verified end-to-end rather than by inspection. The target is now resolved against the procedures this run actually ingested, so the urn carries the argument-signature hash and the catalog's casing instead of the CALL text — the `database` variable that was computed and then dropped on the floor is gone. The generic hook lives in #19380 (`resolve_procedure_urn`); `include_procedures` plus the database/schema/procedure patterns are all honoured, so a filtered-out procedure yields no edge rather than a dangling one.

Against a live Snowflake account, a task calling a one-argument procedure emits `...stored_procedures,PROD),LOAD_REGIONAL_1aefcd00833a35307d2a057f0b387851)` — byte-identical to the urn the same run emits for that procedure. Mutation-checked: dropping the hash from the resolved urn fails the test.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
alokr-dhub: Both halves are done now. The false positive went away with the `tasks_without_sql_lineage` counter, as you suggested.

The second half — surfacing the sub-report — is in #19380: `parse_procedure_code` takes a `ProcedureParseReport` carrying statement, query and column failure counts plus the first error message. So a body whose DML failed to parse raises `Task SQL Parse Failure` with the error text and increments `tasks_with_sql_parse_failures`, while a body with nothing lineage-bearing in it stays a quiet count. That also restores the distinction you noted losing: a column-only failure gets its own `Task Column Lineage Parse Failure`. Statement-level failures now log the exception message instead of just `type(e).__name__`.

metadata-ingestion/tests/unit/snowflake/test_snowflake_tasks.py
alokr-dhub: Rewritten to urn equality. `_procedure_urn` builds the expected urn through `BaseProcedure.to_urn` with the same keys the source uses, and the assertions are `io.inputDatajobs == [that_urn]`. Added the uppercase-definition case, and a `convert_urns_to_lowercase=False` case (`test_call_resolves_to_catalog_casing_without_lowercasing`) — the axis a substring check structurally cannot see.

`EXECUTE IMMEDIATE` has its own test, and you were right to ask for it: as first written it was vacuous (asserting over an empty list). It now puts a procedure literally named `IMMEDIATE` in the fixture, so removing the guard produces an edge and fails the test. The composition path that Oracle/MSSQL take is covered separately in #19380.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_v2.py
alokr-dhub: Fixed — `_get_stages_tasks_pipes_workunits` now runs after `discovered_datasets` is assigned. Rather than leave that as a call-order guarantee the next refactor would quietly undo, `_is_temp_table` itself raises when `discovered_datasets` is still None (an explicit `raise AssertionError`, so it survives `python -O`), and the tests pass a real `is_temp_table` predicate through the extractor.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
alokr-dhub: Threaded through, as you suggested: `additional_input_jobs` is a `parse_procedure_code` parameter now and the merge happens there once, deduplicated (#19380). The local copy in `stored_procedures/base.py` is deleted, and this block collapsed to exactly the shape you sketched. The remaining merge in `lineage.py` is now `dict.fromkeys`, and its `elif result.inputDatajobs` arm turned out to be dead — `to_datajob_input_output` only ever builds dataset lineage.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
alokr-dhub: Now uses `split_qualified_name`. Predecessor references are normalised to an upper-cased 3-part FQN before lookup and `task_name_map` is keyed the same way, so the boolean is gone entirely and 2-part `SCHEMA.TASK` resolves. Quoted identifiers come out right because the splitter is quote-aware, which was the other gap you named.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
alokr-dhub: Fixed — the edge is built from `snowflake_identifier(pred_task.name)`, so `task_name_map`'s values are doing work rather than only membership. `test_predecessor_urn_matches_emitted_task_urn_without_lowercasing` pins it.

Related, found while re-reviewing this path: the map was built from *all* tasks, so a predecessor excluded by `task_pattern` resolved to a DataJob that this run never emits. It is now keyed off `allowed_tasks`, and that case reports `Predecessor Task Filtered Out` instead of emitting the edge.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
alokr-dhub: Narrowed. The parse call guards itself inside `_parse_task_definition_for_lineage`, so DataJobInfo/SubTypes/Status/Ownership always emit — `test_sql_parse_failure_still_emits_ownership` pins that. The outer guard stays as belt-and-braces but no longer wraps a `yield`: the workunits are drained into a list inside the `try` and yielded outside it, so an exception the consumer throws back in isn't swallowed.

And the counter you asked for: `tasks_failed` now means only "task skipped entirely", with `tasks_with_sql_parse_failures` for the lineage-only failure. Those were the same field before, which made the report unreadable exactly where it mattered.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
alokr-dhub: Moved the gate, deliberately. It now strips `inputDatasets` / `outputDatasets` / `fineGrainedLineages` and leaves `inputDatajobs` intact, so a task→procedure edge behaves like the predecessor edges on the same aspect and like the procedure→procedure path that is gated only on `include_procedures`. `test_datajob_edge_survives_include_table_lineage_disabled` covers it.

metadata-ingestion/tests/integration/snowflake/test_snowflake_stages_tasks_pipes.py
alokr-dhub: Done — `snowflake_task_lineage_golden.json`, via `check_golden_file` on a lineage-enabled config. It pins the `fineGrainedLineages` column pairs, `child_task_1`'s `inputDatajobs` (where the dangling-urn bug lived), and the absence of spurious events.

The programmatic tests kept what a golden can't asserts: the report counters, plus four pipeline-level tests — parse failure reported with the task still ingested, no-lineage bodies counted and not warned, a `task_pattern`-filtered predecessor emitting no edge, and a `procedure_pattern`-filtered CALL emitting no edge. The last two carry positive controls, because both aspects vanish when the edge is dropped and the assertions would otherwise pass vacuously.

metadata-ingestion/tests/unit/snowflake/test_snowflake_tasks.py
alokr-dhub: Added. `_collect_workunits` takes an `is_temp_table` predicate, and `test_temp_table_staging_collapses_through_to_real_target` uses a plain `CREATE TABLE` with `lambda name: "tmp_" in name` so the injected predicate is the only signal — a self-identifying `CREATE TEMPORARY TABLE` would pass even with the wiring broken. `is_temp_table` is a required field now, so no test can silently fall back to a default.

metadata-ingestion/docs/sources/snowflake/snowflake_post.md
alokr-dhub: Both done. The paragraph now states the `include_table_lineage` / `include_column_lineage` dependency explicitly, and the CALL claim describes what actually happens: the target is resolved against the procedures ingested in the same run, which requires `include_procedures`, and a procedure that was filtered out or lives outside the ingestion's scope yields no edge rather than an edge to something that was never created. Overloads that arity can't separate are described as reported-ambiguous rather than guessed.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
alokr-dhub: Made required, like `schema_resolver` above it. The tests pass a real predicate now — which is what surfaced the `discovered_datasets` ordering gap you flagged in the other thread, so the change earned itself.

metadata-ingestion/src/datahub/sql_parsing/sql_parsing_aggregator.py
alokr-dhub: Both done, in #19379. The docstring states the ownership rule, and the property raises when the aggregator is closed, so use-after-close is a test failure rather than undefined behaviour. The migration is finished on the Snowflake side: `snowflake_schema_gen.py:1960` and `:2929` and the queries-v2 call site all use the public property, and that first one also stopped reaching for the private `_resolve_schema_info` — it uses `SchemaResolver.resolve_urn`, which let the fail-open `except AttributeError` go. `sql_common.py` is left alone per your scoping.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
```diff
@@ -204,3 +344,251 @@ def _gen_data_job(
...
+                # can't parse still leaves DataJobInfo/SubTypes/Status/Ownership
+                # intact — dropping Ownership would silently discard what a
+                # previous run wrote.
+                self.report.tasks_with_sql_parse_failures += 1
+                self.report.warning(
+                    title="Task Lineage Extraction Failed",
+                    message="Failed to parse task definition; task metadata is still ingested",
+                    context=task_fqn,
+                    exc=e,
+                )
```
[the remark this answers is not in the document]
alokr-dhub: Good catch, and it was worse than described: the outer per-task guard in `get_workunits` caught it too, so the fault was softened at two levels. Both now re-raise `AssertionError` (3f4f701), and `test_ordering_fault_is_not_downgraded_to_a_parse_failure` fails if either re-raise is removed — it injects an `is_temp_table` that raises and asserts the exception reaches the caller instead of becoming a per-task warning.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_tasks.py
```diff
@@ -204,3 +344,251 @@ def _gen_data_job(
...
+                f"{parse_report.queries_failed} query(s) failed; first error: {parse_report.first_error}",
+            )
+        elif parse_report.queries_column_failed:
+            # Table-level lineage survived; only the column-level half was lost.
+            self.report.warning(
+                title="Task Column Lineage Parse Failure",
+                message="Table-level lineage extracted, but column-level lineage failed for part of the task body",
+                context=f"{task_fqn}: {parse_report.queries_column_failed} query(s); "
+                f"first error: {parse_report.first_error}",
+            )
```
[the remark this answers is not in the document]
alokr-dhub: Correct — fixed in the same commit. The column warning now requires both `include_table_lineage` and `include_column_lineage`, so a recipe with column lineage off doesn't hear about a column-lineage failure whose fineGrainedLineages would have been stripped anyway. Covered both ways: silent when disabled, still raised when enabled.

PR #17320, opened 2026-05-06
test(playwright): migrate glossary tests

v-tarasevich-blitz-brain:
This PR migrates glossary related cypress tests to playwright

Depends on https://github.com/datahub-project/datahub/pull/17377

### DIFF
--- a/e2e-test/ui/playwright/pages/glossary.page.ts
+++ b/e2e-test/ui/playwright/pages/glossary.page.ts
@@ -3,37 +3,23 @@
  *
  * Covers the full CRUD lifecycle for Glossary Term Groups and Glossary Terms,
  * including navigation, creation, moving, deleting, and adding terms to datasets.
- *
- * Key selectors are taken directly from the React source:
- *   - GlossaryContentProvider.tsx   → glossaryPageV2, add-term-group-button-v2
- *   - GlossarySidebar.tsx           → glossary-browser-sidebar, create-glossary-button
- *   - ChildrenTab.tsx               → add-term-button
- *   - EntityTabs.tsx                → ${name}-entity-tab-header
- *   - CreateGlossaryEntityModal.tsx → create-glossary-entity-modal-name, glossary-entity-modal-create-button
- *   - MoveGlossaryEntityModal.tsx   → move-glossary-entity-modal, glossary-entity-modal-move-button
- *   - EntityDropdown.tsx            → MoreVertOutlinedIcon (three-dot), entity-menu-delete-button, entity-menu-move-button
- *   - SidebarGlossaryTermsSection.tsx → entity-profile-glossary-terms, add-terms-button
- *   - AddTagsTermsModal.tsx         → tag-term-modal-input, tag-term-option, add-tag-term-from-modal-btn
- *   - EntityActions.tsx             → glossary-batch-add
- *   - EntitySearchResults.tsx       → checkbox-{urn} (entity select checkboxes in batch-add modal)
- *   - SearchSelectModal.tsx         → search-select-modal
- *   - EmbeddedListSearch            → search-results-advanced-search, adv-search-add-filter-tags
  */
 
 import { expect, type Locator, type Page } from '@playwright/test';
 import { BasePage } from './base.page';
 import type { DataHubLogger } from '../utils/logger';
+import { GraphQLHelper } from '../helpers/graphql-helper';
+import { ModalComponent } from './common/modal-component';
 
 export class GlossaryPage extends BasePage {
+  readonly modalComponent: ModalComponent;
+
   // ── Navigation ──────────────────────────────────────────────────────────────
-  readonly glossaryPageHeader: Locator;
   readonly sidebarContainer: Locator;
+  readonly activeTabPanel: Locator;
 
   // ── Root-level create buttons (GlossaryContentProvider header) ──────────────
   readonly addTermGroupButtonV2: Locator;
-
-  // ── Glossary sidebar (GlossarySidebar) ──────────────────────────────────────
-  readonly createGlossarySidebarButton: Locator;
 
   // ── Entity contents tab buttons (ChildrenTab) ────────────────────────────────
   readonly addTermButton: Locator;
@@ -44,6 +30,7 @@ export class GlossaryPage extends BasePage {
 
   // ── Move modal (MoveGlossaryEntityModal) ─────────────────────────────────────
   readonly moveModalContainer: Locator;
+  readonly moveModalSelectInput: Locator;
   readonly moveModalSubmitButton: Locator;
 
   // ── Three-dot entity menu (EntityDropdown) ────────────────────────────────────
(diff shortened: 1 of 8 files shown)


priyabratadas-dh (commented):
## Test Inspector Review

### Summary

- 5 files changed: 1 new page object (`glossary.page.ts`), 3 new spec files, 1 seed data file
- 351 lines of test + 356 lines of page object code added
- 1 critical framework compliance issue found
- 5 warnings found
- CI workflow: no changes needed (tests auto-discovered via `testDir: './tests'`)
- Positive: strong overall architecture, correct fixture imports, good use of POM

---

### 🔴 Framework Compliance Issues

| File | Location | Issue | Fix |
|------|----------|-------|-----|
| `v2-glossary-term.spec.ts` | "can search related entities" test, `firstTitle` block | 3× `page.locator('[data-testid^="preview-urn:"] [data-testid="entity-title"]')` called directly in spec | Move to `GlossaryPage` as `firstRelatedEntityTitle: Locator` property and a `getFirstRelatedEntityTitle()` helper |
| `v2-glossary-term.spec.ts` | "can apply filters" test, ~line 87 | `page.locator('input.ant-checkbox-input[data-testid^="facet-tags-urn:li:tag:"]')` directly in spec — mixes a brittle AntD CSS class with a testid in spec code | Add a `getFacetTagCheckbox()` method to `GlossaryPage` that returns this locator |

---

### 🟡 Warnings

| # | File / Location | Issue | Recommendation |
|---|-----------------|-------|----------------|
| 1 | `v2-glossary-navigation.spec.ts` | All 10 steps (create, move × 2, verify tabs, move group, delete × 3) live in a **single test function** with a forced `setTimeout(120000)`. When it fails, there is no way to tell which of the 10 steps broke without reading the full trace. | Split into ≥ 3 focused tests: "create and move terms", "verify tab persistence", "move group and cleanup". Serial mode already declared so sharing state is fine. |
| 2 | `glossary.page.ts` — `batchAddToFirstResult()` | Method is defined but never called from any spec file. | Remove dead code, or add a comment explaining it is reserved for future use. |
| 3 | `glossary.page.ts` — `clickFacetFilterIcon()` | Uses `.anticon-filter` CSS class selector — an AntD internal class that is not part of DataHub's public API surface and can change on AntD upgrades. | Request a `data-testid="facet-filter-button"` on the filter icon in the React component, or use `page.getByRole('button', { name: 'filter' })` if the aria-label is stable. |
| 4 | `glossary.page.ts` — `filterRelatedAssetsByTag()` | Uses `[aria-label="filter"]` and `#search-results-advanced-search` (ID selector) — both fragile and harder to maintain than `data-testid` attributes. Also uses `div.ant-select-selection-overflow input` (AntD internal). | These selectors already exist in other page objects as accepted workarounds for AntD. Document the gap with a `// TODO: request data-testid` comment to track the technical debt. |
| 5 | `glossary.page.ts` — `moveCurrentEntityTo()` | Uses `.ant-select-selector input` (AntD internal class) to find the search input inside the move modal. | This pattern exists elsewhere in the codebase (e.g., `business-attribute.page.ts`) so it is an accepted workaround. Add a comment citing the AntD constraint to match the pattern used by peers. |

---

### ✅ Positive Feedback

- All 3 spec files correctly import from `../../fixtures/base-test` — the right fixture for authenticated tests.
- `featureName: 'glossary'` + `tests/glossary/fixtures/data.json` path matches exactly what `seeding.fixture.ts` expects. The seed data is well-structured with all four prerequisite entities (2 datasets, 1 GlossaryNode, 1 Tag).
- Unique `runId = Date.now()` suffixes on all entity names prevent cross-run and parallel-suite collisions.
- `test.describe.configure({ mode: 'serial' })` is declared explicitly in all suites that share state, which is the correct pattern.
- Inline comments throughout the page object explain the **why** behind AntD workarounds (`pressSequentially`, `force: true`, `moveModalContainer` scoping) — this is exactly the comment standard the codebase asks for.
- `batchAddToEntityBySearch` properly uses `pressSequentially` with a 50 ms delay to trigger AntD's debounced search handler — shows deep understanding of the AntD interaction model.
- No hardcoded credentials, no `waitForTimeout` calls, no `test-context`/`api-mock`/`logging` fixture imports.
- CI: `playwright.config.ts` uses `testDir: './tests'` with no explicit file lists, so all new `.spec.ts` files are automatically discovered. No workflow change is required.

---

### Recommendations

1. **Block on critical issue**: Move the 4 raw `page.locator()` calls in `v2-glossary-term.spec.ts` into `GlossaryPage` before merging.
2. **Strongly consider splitting** the 10-step navigation mega-test — it will be painful to debug in CI with a 2-minute timeout.
3. **Remove or document** `batchAddToFirstResult()` to avoid confusing future contributors.
4. Items 3–5 in the warnings are low-priority and match existing codebase patterns — they can be tracked as tech debt comments rather than blockers.

---

_Reviewed by Test Inspector_

e2e-test/ui/playwright/pages/glossary.page.ts
```diff
@@ -0,0 +1,356 @@
...
+
+    this.sidebarGlossarySection = page.locator('#entity-profile-glossary-terms');
+    this.addTermsButton = page.locator('[data-testid="add-terms-button"]');
+
+    // The AntD Select renders as a div wrapper; we target the inner search input for typing.
+    this.tagTermModalInput = page.locator('[data-testid="tag-term-modal-input"] input');
+    this.tagTermOption = page.locator('[data-testid="tag-term-option"]').first();
+    this.addTagTermConfirmButton = page.locator('[data-testid="add-tag-term-from-modal-btn"]');
+
+    this.batchAddGlossaryButton = page.locator('[data-testid="glossary-batch-add"]').first();
```
devashish2203: Use of .first() and .last() is not recommended. Do we not have any other property that we can rely on?

e2e-test/ui/playwright/pages/glossary.page.ts
```diff
@@ -0,0 +1,356 @@
...
+  async navigateToEntityPropertiesTab(): Promise<void> {
+    await this.clickEntityTabByName('Properties');
+  }
+
+  async navigateToEntityRelatedAssetsTab(): Promise<void> {
+    await this.clickEntityTabByName('Related Assets');
+  }
+
+  async clickEntityTabByName(tabName: string): Promise<void> {
+    const tabSelector = `[data-testid="${tabName}-entity-tab-header"]`;
```
devashish2203: No locators in functions. These should be defined as attributes in the POM.

e2e-test/ui/playwright/pages/glossary.page.ts
```diff
@@ -0,0 +1,356 @@
...
+
+  /**
+   * Creates a root-level Term Group from the glossary home page header button.
+   * Opens the modal, types the name, and confirms.
+   */
+  async createTermGroup(name: string): Promise<void> {
+    this.logger?.step('createTermGroup', { name });
+    await this.addTermGroupButtonV2.click();
+    // Wait for the modal heading (h1 role) specifically to avoid matching the button text.
+    await expect(this.page.getByRole('heading', { name: 'Create Glossary' })).toBeVisible({ timeout: 15000 });
```
devashish2203: All the getByTole, getByText elements should be attributes in the POM.

e2e-test/ui/playwright/tests/glossary/v2-glossary-navigation.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+ * Prerequisites: `PlaywrightNode` glossary node must exist (seeded via fixtures/data.json).
+ */
+
+import { test, expect } from '../../fixtures/base-test';
+import { GlossaryPage } from '../../pages/glossary.page';
+
+test.use({ featureName: 'glossary' });
+
+// Serial: tests build on state created in earlier tests within this suite.
+test.describe.configure({ mode: 'serial' });
```
devashish2203: We should drop this pattern. 
Each test must be independently runnable. If it relies on pre-existing data each test should create it's data independently. APIs can be used for faster data creation in tests, if the data creation itself is not in the scope of the test.

e2e-test/ui/playwright/tests/glossary/v2-glossary-navigation.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+  test('create term group and terms, move to group, verify sidebar', async ({ page, logger, logDir }) => {
+    test.setTimeout(120000);
+
+    const glossaryPage = new GlossaryPage(page, logger, logDir);
+
+    // ── Step 1: Create root-level Term Group ────────────────────────────────
+    await glossaryPage.navigateToGlossary();
+    await glossaryPage.createTermGroup(TERM_GROUP_NAME);
+
+    await glossaryPage.navigateToGlossary();
```
devashish2203: We should use methods that a normal user would use for navigation. First time navigation using URL is fine, but not for subsequent navigation within the test.

e2e-test/ui/playwright/tests/glossary/v2-glossary-navigation.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+    await glossaryPage.expectSidebarContains(TERM_GROUP_NAME);
+
+    // ── Step 2: Create first Term inside the group ──────────────────────────
```
devashish2203: Instead of comments use page.step to define proper test steps.

e2e-test/ui/playwright/tests/glossary/v2-glossary-navigation.spec.ts
devashish2203: Also there is only one test here, so I think we can drop the serial requirement.

e2e-test/ui/playwright/tests/glossary/v2-glossary-term.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * Each test run uses a unique timestamp-based suffix so tests are independent
+ * and can run in parallel without colliding on entity names.
+ *
+ * Prerequisites:
+ *   - `PlaywrightNode` glossary node exists (seeded via fixtures/data.json).
+ *   - A dataset with tag "PlaywrightGlossaryTag" exists in the test environment.
+ *   - A dataset named "PlaywrightGlossaryTaggedDataset" tagged with "PlaywrightGlossaryTag" exists (seeded via fixtures/data.json).
+ *
+ * Note: Tests in this file share a term created in the first test (serial mode).
```
devashish2203: We should stop using this pattern. Create a unique term in beforeEach using APIs and use that in the test. 
Each test should cleanup the same in afterEach.

e2e-test/ui/playwright/tests/glossary/v2-glossary.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+ * Prerequisites: `PlaywrightNode` glossary node and
+ *   `SamplePlaywrightGlossaryDataset` dataset must exist (seeded via fixtures/data.json).
+ */
+
+import { test, expect } from '../../fixtures/base-test';
+import { GlossaryPage } from '../../pages/glossary.page';
+
+test.use({ featureName: 'glossary' });
+
+// Use serial mode because these tests mutate shared state (the same term group and term).
```
devashish2203: This contradicts what's written earlier in the file at line 11.
> Each test run uses a unique timestamp-based suffix so tests are independent and can run in parallel with other suites without colliding on entity names.

devashish2203 (changes requested):

e2e-test/ui/playwright/tests/glossary/v2-glossary-navigation.spec.ts
v-tarasevich-blitz-brain: I see there is a bug in the app. After adding a new term, sidebar refreshes with the stale data (perhaps kafka delay). Not sure if we need to fix it during the migration.

devashish2203 (approved):

PR #17325, opened 2026-05-06
feat(ingest/glue): emit column Parameters as structured properties on schemaField

JohnRTurner:
## Summary

- Adds `extract_column_parameters` config flag to the Glue source (default `false`)
- When enabled, column-level `Parameters` from the Glue catalog are emitted as `StructuredProperties` on each `schemaField` entity — no frontend or backend changes required
- Each unique parameter key's `StructuredPropertyDefinition` is upserted once per recipe run (cached in `_seen_column_param_urns`); repeated columns reuse the existing definition
- Partition key columns are also processed
- Property URNs follow the pattern `urn:li:structuredProperty:io.datahubproject.glue.column.<sanitized-key>`

## Motivation

Glue tables backed by Iceberg (and other formats) carry column-level metadata in `Parameters` (e.g. `iceberg.field.id`, `iceberg.field.optional`). This approach surfaces that metadata in the DataHub Properties tab without any schema changes, using the existing structured properties infrastructure that `schemaField` already supports.

## Test plan

- [ ] 5 new unit tests in `test_glue_source.py` cover: URN sanitization, definition deduplication, columns-without-params skipping, value assignment correctness, and cross-table definition caching
- [ ] All existing Glue unit tests still pass
- [ ] `lintFix` and pre-commit hooks pass

## Checklist

- [x] PR conforms to Contributing Guideline (PR Title Format)
- [x] Tests added/updated
- [x] Docs updated (`glue_post.md`)
- [ ] No breaking changes

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
@@ -2021,8 +2021,10 @@ class GlueSource(StatefulIngestionSourceBase):
                 )
                 assignments = []
                 for key, value in params.items():
-                    qualified_name = f"io.datahubproject.glue.column.{re.sub(r'[^a-zA-Z0-9._]', '_', key)}"
-                    property_urn = f"urn:li:structuredProperty:{qualified_name}"
+                    property_urn = self._column_param_property_urn(key)
+                    qualified_name = property_urn.removeprefix(
+                        "urn:li:structuredProperty:"
+                    )
                     if property_urn not in self._seen_column_param_urns:
                         yield MetadataChangeProposalWrapper(
                             entityUrn=property_urn,
--- a/metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
@@ -2049,8 +2049,8 @@ class GlueSource(StatefulIngestionSourceBase):
                     ).as_workunit()
             except Exception as e:
                 self.report.report_warning(
-                    key=dataset_urn,
-                    reason=f"Failed to emit column parameters for column {column.get('Name', '?')!r}: {e}",
+                    message="Failed to emit column parameters for column",
+                    context=f"dataset={dataset_urn} column={column.get('Name', '?')!r}: {e}",
                 )
 
     def _get_s3_tags(self, table: Dict, dataset_urn: str) -> Optional[GlobalTagsClass]:
--- a/metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
@@ -13,6 +13,7 @@ from typing import (
     Literal,
     Mapping,
     Optional,
+    Set,
     Tuple,
 )
 from urllib.parse import urlparse
@@ -123,6 +124,9 @@ from datahub.metadata.schema_classes import (
     QueryLanguageClass,
     QueryStatementClass,
     SchemaMetadataClass,
+    StructuredPropertiesClass,
+    StructuredPropertyDefinitionClass,
+    StructuredPropertyValueAssignmentClass,
     TagAssociationClass,
     TimeStampClass,
     UpstreamClass,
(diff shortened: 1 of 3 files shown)


metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
```diff
@@ -1980,6 +1999,48 @@ def _get_ownership(owner: str) -> Optional[OwnershipClass]:
...
 
+    @staticmethod
+    def _column_param_property_urn(key: str) -> str:
+        sanitized = re.sub(r"[^a-zA-Z0-9._]", "_", key)
+        return f"urn:li:structuredProperty:io.datahubproject.glue.column.{sanitized}"
+
+    def _get_column_param_workunits(
+        self, table: Dict, dataset_urn: str
+    ) -> Iterable[MetadataWorkUnit]:
+        columns = table.get("StorageDescriptor", {}).get("Columns", [])
```
treff7es, line 2012: Some error handling would make sense othewise any KeyError, TypeError, or pydantic validation error in this generator aborts

metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
```diff
@@ -1980,6 +1999,48 @@ def _get_ownership(owner: str) -> Optional[OwnershipClass]:
...
+        for column in columns:
+            params = column.get("Parameters")
+            if not params:
+                continue
+            field_urn = mce_builder.make_schema_field_urn(dataset_urn, column["Name"])
+            assignments = []
+            for key, value in params.items():
+                property_urn = self._column_param_property_urn(key)
+                if property_urn not in self._seen_column_param_urns:
+                    self._seen_column_param_urns.add(property_urn)
```
treff7es: self._seen_column_param_urns.add(property_urn) is called before the work unit is yielded. If the consumer/sink drops the definition MCP (GMS rejection, validation failure, network blip with no retry), every subsequent assignment in the same run references a propertyUrn whose definition was never persisted. The doc's "upserted once per recipe run" promise is actually "yielded at most once per run."

metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
```diff
@@ -1980,6 +1999,48 @@ def _get_ownership(owner: str) -> Optional[OwnershipClass]:
...
+                    sanitized = re.sub(r"[^a-zA-Z0-9._]", "_", key)
```
treff7es: This is duplicated (the same is above), having copy-pasted risks drift between the two regular expressions

metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
```diff
@@ -1980,6 +1999,48 @@ def _get_ownership(owner: str) -> Optional[OwnershipClass]:
...
+        columns = table.get("StorageDescriptor", {}).get("Columns", [])
+        columns = columns + table.get("PartitionKeys", [])
+        for column in columns:
+            params = column.get("Parameters")
+            if not params:
+                continue
+            field_urn = mce_builder.make_schema_field_urn(dataset_urn, column["Name"])
+            assignments = []
+            for key, value in params.items():
+                property_urn = self._column_param_property_urn(key)
```
treff7es:
```suggestion
                sanitized = re.sub(r"[^a-zA-Z0-9._]", "_", key)
                property_urn = f"urn:li:structuredProperty:{qualified_name}"
```

metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
JohnRTurner: Wrapped the per-column body in try/except Exception that calls self.report.report_warning(...) and continues, so a bad column (bad Parameters value, unexpected type, etc.) won't abort the rest of the table.

metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
JohnRTurner: Moved _seen_column_param_urns.add(property_urn) to after the yield — the URN is only cached once the consumer has accepted the work unit, so a dropped or rejected MCP leaves the key uncached and eligible for retry.

metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
JohnRTurner: The re.sub pattern now lives exclusively in _column_param_property_urn. _get_column_param_workunits calls that method for the URN and derives qualified_name from it via removeprefix("urn:li:structuredProperty:") — no risk of the two drifting.

metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
JohnRTurner: _column_param_property_urn computes qualified_name internally and returns the full URN. The workunits method derives qualified_name back from the URN via removeprefix, so both the definition's qualifiedName field and the URN come from the same source of truth.

treff7es (approved):

PR #17326, opened 2026-05-06
perf(ingest/teradata): filter tables-and-views query by configured databases to reduce peak memory

JohnRTurner:
## Summary

- `cache_tables_and_views()` previously issued a query that loaded **all** tables/views from every user database on the Teradata installation into `_tables_cache`, regardless of what the recipe was configured to ingest. On large enterprise systems with thousands of databases this is the primary driver of OOM crashes — the cache grows proportional to the entire system, not the recipe scope.
- When `config.databases` is explicitly set, this PR pushes an `AND DataBaseName IN (...)` clause into the query so only the configured databases are fetched. This reduces `_tables_cache` size from O(system) to O(recipe), which can be a 10–100× memory reduction on large installations.
- When `config.databases` is not set, behaviour is unchanged — all databases are loaded as before.

## Why it's safe

Non-recipe database entries in `_tables_cache` were never read: the cache is only accessed for schemas currently being iterated, and only recipe databases pass the `database_pattern` filter. Cross-database view lineage resolution uses the `schema_resolver` (not `_tables_cache`) and was already limited to recipe databases. `_make_lineage_queries()` uses `self.config.databases` directly when that field is set, bypassing `_tables_cache` entirely.

## Test plan

- [ ] Added `test_tables_query_database_filter` — verifies that setting `config.databases` produces a query with an `IN` clause containing exactly those databases
- [ ] Existing `test_query_structure_optimization` updated — now calls `_build_tables_and_views_query()` and additionally asserts no `IN` clause is added when `config.databases` is unset
- [ ] All 110 existing Teradata unit tests pass

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
@@ -829,7 +829,7 @@ class TeradataSource(TwoTierSQLAlchemySource):
     ORDER BY "timestamp", "query_id", "row_no"
     """.strip()
 
-    TABLES_AND_VIEWS_QUERY: str = f"""
+    _TABLES_AND_VIEWS_QUERY_TEMPLATE: str = """
 SELECT
     t.DataBaseName,
     t.TableName as name,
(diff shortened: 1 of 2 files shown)


treff7es (approved):
lgtm

PR #8877, opened 2023-09-21
feat(entity-client): enable client side cache for entity-client and usage-client

david-leifker:
Created a client cache for entity-client, enabled on the system-entity-client specifically. Features include per aspect TTL, maximum size (roughly), and cache metrics.

This cache is particularly useful for caching the usage stats calls on search result pages and user information where for a given page load there are multiple calls to fetch the same information.

## Checklist

- [ ] The PR conforms to DataHub's [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [Commit Message Format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#commit-message-format))
- [ ] Links to related issues (if applicable)
- [ ] Tests for the changes have been added/updated (if applicable)
- [ ] Docs related to the changes have been added/updated (if applicable). If a new feature has been added a Usage Guide has been added for the same.
- [ ] For any breaking change/potential downtime/deprecation/big changes an entry has been made in [Updating DataHub](https://github.com/datahub-project/datahub/blob/master/docs/how/updating-datahub.md)

PR #17327, opened 2026-05-06
fix(re-index-loop): prevent perpetual ES8 reindex loop in graph index…

gdatahub:
… (PFP-3594)

On ES8, the mapping diff comparator` (ReindexConfig.java:238–246) `continuously flagged for reindexing during every system update cycle.

Logs showed entriesDiffering for the source, destination, and properties fields:

```
left (ES stored): {properties: {...}, type=object} // includes explicit type
right (DataHub code): {properties: {...}} // missing type
```

Elasticsearch 8 returns mappings with an explicit type: **object** for object fields.
However, DataHub’s GraphRelationshipMappingsBuilder previously omitted this, relying on **Elasticsearch’s implicit** default.

This mismatch caused the mapping comparison to **fail on every round trip**, triggering a reindex on each cycle without ever converging.

  **Summary**

  - ES8+ silently strips doc_values: false from search_as_you_type fields on round-trip, while ES7/OpenSearch persist it. This caused V2MappingsBuilder to emit a config that never matched what the cluster
  returned — every diff flagged a change and triggered a fresh reindex on every restart.
  - V2MappingsBuilder now selects the partial-ngram config based on SearchEngineType: ES8+ omits doc_values: false; legacy engines keep it. engineType is plumbed through MappingsBuilderFactory,
  UpdateIndicesStrategyFactory, and UpdateIndicesV2Strategy.
  - ReindexConfig.normalizeObjectForComparison now injects the implicit type: "object" ES adds for any node containing properties, eliminating another spurious-diff source that fed the same loop.

  **Changes**

  - `V2MappingsBuilder.java` — engine-aware PARTIAL_NGRAM_CONFIG_LEGACY vs PARTIAL_NGRAM_CONFIG_ES8; static methods → instance methods.
  - `ReindexConfig.java `— add implicit type: object during normalization when properties is present without type.
  - `UpdateIndicesV2Strategy.java`, UpdateIndicesStrategyFactory.java, MappingsBuilderFactory.java — propagate SearchEngineType to the mappings builder.
  - Tests added/updated in V2MappingsBuilderTest, ReindexConfigTest, UpdateIndicesV2StrategyTest, UpdateIndicesHookTest, UpdateIndicesServiceTest, DisabledCustomSearchTest.

### DIFF
--- a/metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/index/entity/v2/V2MappingsBuilder.java
+++ b/metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/index/entity/v2/V2MappingsBuilder.java
@@ -23,6 +23,7 @@ import com.linkedin.metadata.models.annotation.SearchableAnnotation.FieldType;
 import com.linkedin.metadata.models.registry.EntityRegistry;
 import com.linkedin.metadata.search.elasticsearch.index.MappingsBuilder;
 import com.linkedin.metadata.search.utils.ESUtils;
+import com.linkedin.metadata.utils.elasticsearch.SearchClientShim.SearchEngineType;
 import com.linkedin.structured.StructuredPropertyDefinition;
 import com.linkedin.util.Pair;
 import io.datahubproject.metadata.context.OperationContext;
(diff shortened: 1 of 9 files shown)


gdatahub:
Test are failing because the changes from the datahub-fork branch have not been merged yet: [PR #8877](https://github.com/acryldata/datahub-fork/pull/8877?utm_source=chatgpt.com)

Dutt23 (approved):

gdatahub:
Not needed, the required changes have already been merged.

PR #17328, opened 2026-05-06
feat(ingest/sigma): qualify chart inputFields with workbook-lineage warehouse tables

kyungsoo-datahub:
- Adds `get_workbook_lineage` to fetch `/v2/workbooks/{id}/lineage` type=table entries                                                                              
- Builds a per-workbook warehouse-table index merged with the existing per-element SQL-parser index before calling `_resolve_chart_formula_upstream` Step 4
- Resolves DM-fed chart formula bracket refs (e.g. `[FIVETRAN_LOG/Last Sync]`) that had empty `dataset_inputs` and previously incremented `chart_input_fields_self_ref_fallback`

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/sigma/sigma_api.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/sigma/sigma_api.py
@@ -42,6 +42,7 @@ from datahub.ingestion.source.sigma.data_classes import (
     SigmaDataset,
     WarehouseInodeRaw,
     Workbook,
+    WorkbookLineageTableEntry,
     Workspace,
 )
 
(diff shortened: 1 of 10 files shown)


metadata-ingestion/src/datahub/ingestion/source/sigma/sigma_api.py
```diff
@@ -1202,6 +1202,78 @@ def get_file_metadata(self, inode_id: str) -> Optional[Dict[str, Any]]:
             )
             return None
 
+    def get_workbook_lineage(self, workbook_id: str) -> Optional[List[Dict[str, Any]]]:
```
treff7es: I wonder if we could use here a typed objectto return, like:
```
@dataclass
class WorkbookLineageTableEntry():
      type: Literal["table"]
      name: str
      connectionId: str
      inodeId: str
```

treff7es (approved):

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma_api.py
kyungsoo-datahub: Thank you for the comment. Done.

PR #17329, opened 2026-05-06
feat(ingest/sigma): emit FineGrainedLineage on DM elements to warehouse

kyungsoo-datahub:
`fineGrainedLineages` for the same elements.

Resolution uses `columnId` (format: `inode-<url_id>/<WAREHOUSE_COL>`) which
encodes the actual warehouse column name. The formula bracket ref carries the
display name (e.g. "Customer Id"), not the identifier ("customer_id"), so
  trusting it verbatim would produce orphan schemaFields.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
@@ -233,17 +233,17 @@ class _WarehouseTableRef:
     """Resolved warehouse table coordinates derived from a /files response."""
 
     connection_id: str
-    db: str
+    db: Optional[str]
     schema: str
     table: str
 
     def fq_name(self, platform: str, *, lowercase: bool = True) -> str:
-        # db is empty for platforms with a 2-segment /files path (e.g. Redshift:
+        # db is None for platforms with a 2-segment /files path (e.g. Redshift:
         # "Connection Root/<SCHEMA>"). In that case emit schema.table; otherwise
         # emit the full db.schema.table used by Snowflake and similar platforms.
         name = (
             f"{self.schema}.{self.table}"
-            if not self.db
+            if self.db is None
             else f"{self.db}.{self.schema}.{self.table}"
         )
         if platform.lower() in _WAREHOUSE_LOWERCASE_PLATFORMS and lowercase:
@@ -772,13 +772,11 @@ class SigmaSource(StatefulIngestionSourceBase, TestableSource):
                 # first, then fall back to the connection registry's default_database.
                 conn_override = self.config.connection_to_platform_map.get(conn_id)
                 conn_record = self.connection_registry.get(conn_id)
-                db = (
-                    (conn_override.default_database if conn_override else None)
-                    or (conn_record.default_database if conn_record else None)
-                    or ""
+                db = (conn_override.default_database if conn_override else None) or (
+                    conn_record.default_database if conn_record else None
                 )
                 schema = parts[1]
-                if not db and conn_id not in self._missing_default_db_warned:
+                if db is None and conn_id not in self._missing_default_db_warned:
                     self._missing_default_db_warned.add(conn_id)
                     self.reporter.warning(
                         title="Sigma warehouse default_database not configured",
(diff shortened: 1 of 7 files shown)


metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
```diff
@@ -1505,7 +1559,78 @@ def _gen_data_model_element_schema_metadata(
...
+    ) -> Optional[FineGrainedLineageClass]:
+        """Attempt to build a warehouse-passthrough FineGrainedLineage entry.
+
+        Resolves the column's warehouse identity via the columnId field, which
+        Sigma encodes as ``inode-<url_id>/<WAREHOUSE_COLUMN_NAME>``.  This is
+        more reliable than the formula bracket ref, which carries the DM display
+        name (e.g. "Customer Id") rather than the warehouse identifier
+        ("CUSTOMER_ID" / "customer_id").
+
+        Returns a FineGrainedLineageClass on success; returns None and bumps a
```
treff7es: In the cross-DM-precursor branch (sigma.py:1734-1754), only one of the helper's failure modes is counted: if (column.columnId or "").startswith("inode-") → bump
  _warehouse_passthrough_deferred. If the helper returned None for any of the other 4 reasons (sep missing, source_ids mismatch, url_id missing from map, parent_urn None, record None —
   but columnId did start with inode-), the failure is correctly counted. The remaining gap: columnId not inode-shaped at all → falls through to cross-DM resolution, which is
  intentional. So the behavior is mostly OK, but the docstring needs to either match the caller's responsibility or the counter bumps need to move into the helper.

Either update the docstring to "Returns None on any resolution failure; counter bookkeeping is the caller's responsibility" — or move the deferred-counter increment inside the helper for symmetric responsibility.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
```diff
@@ -1505,7 +1559,78 @@ def _gen_data_model_element_schema_metadata(
...
+        upstream_field = builder.make_schema_field_urn(parent_urn, normalized_col)
+        return FineGrainedLineageClass(
+            downstreamType=FineGrainedLineageDownstreamTypeClass.FIELD,
+            downstreams=[downstream_field],
+            upstreamType=FineGrainedLineageUpstreamTypeClass.FIELD_SET,
+            upstreams=[upstream_field],
+            confidenceScore=1.0,
+        )
+
+    def _build_dm_element_fine_grained_lineages(  # noqa: C901
```
treff7es: Can we get rid of the complexity here and remove the mypy silencing?

The two _try_emit invocations have nearly identical post-emit bookkeeping. The _warehouse_fgl_attempted flag exists because the two arms duplicate the attempt.

  Suggested refactor (highest-leverage in this PR): Compute the warehouse FGL once per column, outside the per-ref loop, since it depends only on column and element, not on the bracket
   ref:

  warehouse_fgl = self._try_emit_warehouse_passthrough_fgl(
      column=column, element=element,
      downstream_field=downstream_field,
      warehouse_url_id_map=warehouse_url_id_map,
  )
  warehouse_consumed = False
  # ... then in each branch: if warehouse_fgl and not warehouse_consumed: append + count + mark consumed

  This eliminates the flag, the duplicate _try_emit call, and likely drops complexity below the C901 threshold so the noqa goes away. Side benefit: the "future branch may forget to set
   _warehouse_fgl_attempted" silent-dedup-break failure mode becomes structurally impossible.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
treff7es: Shouldn't `db` wouldn't be better to be `Optional[str]`, this represent better lack of `db` than `db=""`

treff7es (approved):

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
kyungsoo-datahub: The docstring is updated. Thank you for the comment.

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
kyungsoo-datahub: Updated. Thank you for the suggestion!

metadata-ingestion/src/datahub/ingestion/source/sigma/sigma.py
kyungsoo-datahub: Done. Thanks!

Issue #17242, opened 2026-04-29
[RFC] Bilateral Data Contracts: multi-party agreements with lifecycle, sign-offs, and version history


jayacryl:
## TL;DR

We're proposing a new `bilateralDataContract` entity to replace today's producer-only `dataContract`. Bilateral contracts are multi-party agreements between data producers and consumers with explicit lifecycle states, optional sign-off requirements, and built-in version history. Delivered in five phases.

📄 **Full technical spec (with detailed schemas, GraphQL surfaces, and edge cases):** [Google Doc](https://docs.google.com/document/d/1-2z6CGzQH4EFSUnOyDUyL1j02OWR1d21o5VPBkS5Ia0/edit) — please leave line-level comments there for ergonomic review. Comment under this issue with your email if you're missing access.

This issue is the shorter overview and the home for community discussion threads.

<img width="1440" height="1024" alt="Image" src="https://github.com/user-attachments/assets/3b1213ea-f5c5-43a0-847d-6ea4f98dc281" />

---

## Why

Today's `dataContract` entity is a producer-oriented wrapper around assertions:

- One contract per dataset (`upsertDataContract` overwrites)
- No producer/consumer ownership model
- No lifecycle: contracts cannot be drafted, activated, expired, or archived
- No validity periods
- No version history beyond raw aspect versioning
- No sign-off mechanism — a contract is a one-sided producer declaration

Real-world data contracts are agreements between parties, not declarations. This RFC introduces a new entity that handles all of the above without disrupting existing ingestion pipelines.

> **Deprecation notice.** The existing `dataContract` entity, `upsertDataContract` mutation, and `dataset.contract` GraphQL field are marked deprecated but preserved for ingestion back-compat. They are hidden from the UI when `BILATERAL_DATA_CONTRACTS_ENABLED` is on. A one-shot upgrade step clones legacy contracts into bilateral contracts on first flag activation.

---

## Proposed Phases

| Phase | Name | Scope | Status |
|---|---|---|---|
| 1 | **Basic Bilateral Contracts** | New entity, data model, lifecycle, GraphQL CRUD, UI, migration | OSS |
| 2 | **Sign-Offs** | Multi-party acknowledgment aspect; optional gate on PENDING → ACTIVE | OSS |
| 3 | **Version History** | Timeline-Service-backed semantic version history; stable `versionToken` | OSS |
[...]

PR #17331, opened 2026-05-06
feat(ingest/odcs): add Open Data Contract Standard ingestion source

### LINKED ISSUE #11892: feat: Add support of ODCS (Open Data Contract Standard)
Hey team!
It would be nice if DataHub can import data contract / asserts definition from the [ODCS yaml files](https://bitol-io.github.io/open-data-contract-standard/latest/).
This project is supported by Linux Foundation and rapidly grows its popularity in the data community.
For example developers of alternative solution datacontract.com have added support of ODCS v3.0 to their [CLI tool](https://github.com/datacontract/datacontract-cli/releases/tag/v0.10.14) and in long-term are going to merge both standards.

rbrenton:
Closes #11892. Dependent PR: #17332 (registers the `odcs` data platform).

## Summary

Adds a new ingestion source `source.type: odcs` that ingests Open Data Contract Standard (ODCS) v3.0 / v3.1 YAML files and emits matching DataHub aspects. ODCS is the Linux Foundation / Bitol open standard for data contracts (Apache 2.0, currently v3.1.0).

- **Fan-out per contract**: each `schema[]` entry in an ODCS file becomes one DataHub Dataset URN bound to one DataContract URN. Multi-table contracts (the upstream Bitol AdventureWorks example bundles 4 tables) fan out cleanly without cross-table metadata bleed.
- **Per-table emission**: dataset properties, ownership, top-level + column-level tags, editable schema metadata, and assertions from `quality[]` rules. The original YAML is preserved in `DataContractProperties.rawContract` (capped at 1 MB by default).
- **Scoped soft-delete**: `@capability(DELETION_DETECTION, ...)` is declared with explicit scope — ODCS soft-deletes only `DataContract` and `Assertion` URNs it owns. Dataset URNs are owned by the platform-of-record source (postgres / snowflake / etc.) and are never marked removed.
- **Assertion routing**: `unique`/`notNull` library rules → `FieldValuesAssertion`; `rowCount` → `VolumeAssertionInfo`; `sql` rules with `query` → `SqlAssertionInfo`; everything else (including `mustNotBeBetween`) → `CustomAssertionInfo` with explicit `logic`. Operators and thresholds are never fabricated; rules with no actionable content are skipped with a warning and tracked in `report.rules_skipped_no_threshold`.
- **Lenient default validation** with extras detection: `strict_validation` defaults to `false` so deprecated-but-real ODCS forms (top-level `quality[]`, library `rule:` keys) ingest cleanly. Unknown YAML fields surface as warnings and increment `report.unknown_fields_count`.
- **Vendored ODCS JSON Schemas** under `metadata-ingestion/src/datahub/ingestion/source/odcs/odcs_schema/` pinned to upstream release tags v3.0.2 and v3.1.0; SHA-256 checksums committed in `CHECKSUMS.sha256`. The `metadata-ingestion/scripts/refresh_odcs_schemas.sh` script downloads from the pinned tag and refuses to overwrite on hash mismatch. Apache 2.0 attribution added to the top-level `/NOTICE`.

## Tests

- **49 unit tests** under `metadata-ingestion/tests/unit/odcs/` covering fan-out resolution, per-schema MCP emission (with cross-table-bleed regression guard), assertion routing trace, scope tagging, URN stability, numeric normalization, path resolution + symlink behavior + file-size cap, both `strict_validation` branches, unknown-field detection, and the soft-delete state machine including refusal to soft-delete Dataset URNs.
- **3 golden-file integration tests** under `metadata-ingestion/tests/integration/odcs/` backed by hand-authored fixtures: `odcs_minimal`, `odcs_full_v31` (multi-table, every MVP-mapped section, `mustNotBeBetween` rule, an unknown field for extras detection), and `odcs_synthetic_ecommerce` (synthetic 4-table fictional schema; no upstream byte-similarity).

```bash
./gradlew :metadata-ingestion:testQuick
# 57 passing in tests/{unit,integration}/odcs/
./gradlew :metadata-ingestion:lintFix
./gradlew :metadata-ingestion:connectorRegistry :metadata-ingestion:docGen
```

## Out of MVP (deferred to follow-ups)

- The matching `bootstrap_mcps/data-platforms.yaml` registry entry for the `odcs` platform id is split into the **dependent PR #17332** to get explicit maintainer sign-off on id and display name before it ships.
- `dataset_urn_overrides` shape widening for per-table overrides under fan-out.
- Optional `DataProductProperties` aspect emission when a contract has `dataProduct` set, with `assets[]` = fan-out URNs.
- Per-table `servers[]` mapping for mixed-platform contracts.
- ODCS export (DataHub → ODCS YAML).
- ODCS v2.x support (declined for MVP — adopters have largely migrated; revisit if user demand surfaces).
- SLA, pricing, support channels, and non-owner team roles mapping.

## Checklist

- [x] PR title format conforms to [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#pr-title-format) (`feat(ingest/odcs): ...`)
- [x] Links to related issues — closes #11892
- [x] Tests added (49 unit + 3 golden integration; all passing)
- [x] Docs added (`metadata-ingestion/docs/sources/odcs/` — README, pre, recipe, post)
- [x] No breaking change (additive new source; no entry needed in `docs/how/updating-datahub.md`)


metadata-ingestion/src/datahub/ingestion/source/odcs/odcs_mapper.py
```diff
@@ -0,0 +1,938 @@
...
+        if not username:
+            continue
+        role_key = (member.role or "").strip()
+        ownership_type = _OWNER_ROLE_MAP.get(
+            role_key, OwnershipTypeClass.TECHNICAL_OWNER
+        )
+        if username.startswith(("urn:li:corpuser:", "urn:li:corpGroup:")):
+            owner_urn = username
+        elif username.startswith("group:"):
+            # `group:` prefix lets a contract author target a group rather than a user.
```
AdrianMachado: this is part of the ODCS standard?

metadata-ingestion/src/datahub/ingestion/source/odcs/odcs_mapper.py
```diff
@@ -0,0 +1,938 @@
...
+      - `(operator, parameters)` for natively-modeled thresholds.
+    """
+    if rule.mustBeBetween and len(rule.mustBeBetween) == 2:
+        return AssertionStdOperatorClass.BETWEEN, AssertionStdParametersClass(
+            minValue=_num(rule.mustBeBetween[0]),
+            maxValue=_num(rule.mustBeBetween[1]),
+        )
+    if rule.mustNotBeBetween and len(rule.mustNotBeBetween) == 2:
+        # No NOT_BETWEEN operator; route to CustomAssertionInfo (D6).
+        return _ROUTE_TO_CUSTOM, None
```
AdrianMachado: we should consider adding support for NOT_BETWEEN as a followup so you can avoid this. Alternatively, explicitly do not support it in this PR, and add support it once we do support NOT_BETWEEN

metadata-ingestion/src/datahub/ingestion/source/odcs/odcs_mapper.py
```diff
@@ -0,0 +1,938 @@
...
+    fallback_index: int,
+) -> str:
+    guid_dict = {
+        "contract": contract_urn,
+        "dataset": dataset_urn,
+        "rule_name": rule_name or f"{rule_kind}_{fallback_index}",
+        "rule_kind": rule_kind,
+        "field": column or "",
+    }
+    return make_assertion_urn(datahub_guid(guid_dict))
```
AdrianMachado, line 974: careful with urn length limits

metadata-ingestion/src/datahub/ingestion/source/odcs/odcs_mapper.py
```diff
@@ -0,0 +1,938 @@
...
+        # `:g` strips the trailing ".0" on whole numbers (1.0 -> "1") so the
+        # rendered logic string matches what a contract author wrote.
+        low = f"{float(rule.mustNotBeBetween[0]):g}"
+        high = f"{float(rule.mustNotBeBetween[1]):g}"
+        logic_override = f"value not between {low} and {high}"
+
+    # Library volume rule.
+    if rule_kind in ("rowCount", "row_count"):
+        op_opt, params_opt = _operator_and_params_from_threshold(rule)
+        if op_opt is None or op_opt == _ROUTE_TO_CUSTOM or params_opt is None:
```
AdrianMachado: Personally, i feel it would be a bit cleaner if you didn't have `_ROUTE_TO_CUSTOM` and explicitly special-cased at this level rather than polluting the return interface of `_operator_and_params_from_threshold`. If it only happens in one case, then seems like it wouldn't muddle the code here that much

metadata-ingestion/src/datahub/ingestion/source/odcs/odcs_mapper.py
```diff
@@ -0,0 +1,938 @@
...
+    logic_override: Optional[str] = None
+    if rule.mustNotBeBetween and len(rule.mustNotBeBetween) == 2:
+        # `:g` strips the trailing ".0" on whole numbers (1.0 -> "1") so the
+        # rendered logic string matches what a contract author wrote.
+        low = f"{float(rule.mustNotBeBetween[0]):g}"
+        high = f"{float(rule.mustNotBeBetween[1]):g}"
+        logic_override = f"value not between {low} and {high}"
+
+    # Library volume rule.
+    if rule_kind in ("rowCount", "row_count"):
```
AdrianMachado: Is `rule_kind` like a fixed set of possibilities? Could we make it an enum or something?

AdrianMachado (commented):
Some initial review, will dig in deeper next week

metadata-ingestion/src/datahub/ingestion/source/odcs/odcs_mapper.py
rbrenton: ODCS does model people via `team[]` (with `roles[]`) and `support[]` channels, but `team[].username` is defined only as "the user's username or email"; there's no notion of a group principal you can target for ownership.

The `group:` prefix here is a DataHub-side escape hatch: it lets a contract author deliberately map ownership to a `corpGroup` instead of a `corpuser`, since ODCS can't express that.

Happy to drop it if you'd rather keep ownership strictly aligned to the spec. Otherwise, I can document explicitly that it's a DataHub extension so it isn't mistaken for ODCS-native.

metadata-ingestion/src/datahub/ingestion/source/odcs/odcs_mapper.py
rbrenton: The _assertion-id_ is a fixed-length 32-char MD5 hash regardless of input size. The emitted URN is always `urn:li:assertion:` + 32 chars = 49 characters.

rbrenton: (same `make_assertion_urn(datahub_guid(...))` pattern DataHub uses in `api/entities/datacontract/datacontract.py`, so we're consistent with existing assertion-URN generation.)

metadata-ingestion/src/datahub/ingestion/source/odcs/odcs_mapper.py
rbrenton: I'll drop it: `_operator_and_params_from_threshold` returns `(None, None)` for any non-native case, and the guards simplify to `if op_opt is None or params_opt is None:`.

metadata-ingestion/src/datahub/ingestion/source/odcs/odcs_mapper.py
rbrenton: Agreed. We should add support as a follow-up; I was trying to avoid modifying the metadata model in the initial PR.

metadata-ingestion/src/datahub/ingestion/source/odcs/odcs_mapper.py
rbrenton: `rule_kind` is `rule.rule`, a free-form ODCS library-rule name, and ODCS doesn't constrain those; anything we don't recognize falls through to a `CustomAssertion`.

I can pull the recognized names into module-level `frozenset` constants (`_FIELD_RULE_KINDS`, `_NOT_NULL_RULE_KINDS`, `_VOLUME_RULE_KINDS`); unknown kinds still pass through. Could add a `Literal[...]` alias too if we want the names in the type system.

metadata-ingestion/src/datahub/ingestion/source/odcs/odcs_mapper.py
rbrenton: Done in f7e0167 — dropped `_ROUTE_TO_CUSTOM`; `_operator_and_params_from_threshold` now returns `(None, None)` for any non-native case and both call sites guard on `op_opt is None or params_opt is None`. Added a unit test pinning the invariant (a real operator is never returned with `None` params).

metadata-ingestion/src/datahub/ingestion/source/odcs/odcs_mapper.py
rbrenton: Done in f7e0167 — recognized rule-kinds are now module-level frozensets (`_FIELD_RULE_KINDS`/`_NOT_NULL_RULE_KINDS`/`_VOLUME_RULE_KINDS`); unrecognized kinds still fall through to a custom assertion.

AdrianMachado:
Thanks for pushing ODCS support forward. I spent a lot of time wrapping my head around Data Contracts, Data Mesh architecture, and how evangelists think about it vs actual data engineers and stewards think about it. 

I think that I've realized that ODCS actually has very little to do with Data Contracts in the way we think about it. For us, a "Data Contract" is closer to a Data Usage Agreement in Data Mesh governance parlance, which we are making more explicit in #17242 

ODCS is closer to like a spec or definition for the dataset, kind of like the OpenAPI for datasets. With that said, I think that `Logical Models` are actually a much better way to model these...

## Concepts

- Logical models represent the canonical schema and metadata for a dataset independent of any single physical table. In DataHub this is modeled as a normal `dataset` plus `logicalParent` relationships from physical datasets and schema fields.
- ODCS describes a dataset/data-product interface: schema, quality rules, SLA, ownership/team metadata, support, infrastructure, custom metadata, and other producer-published expectations. That aligns more naturally with logical models than with bilateral agreements/data contracts
- DataHub’s current `dataContract` entity is a producer-oriented assertion wrapper and is planned to be deprecated for UI/primary use. The bilateral roadmap introduces `bilateralDataContract`, where contracts are relationship entities with producer and consumer owners, sign-off, lifecycle, sections, templates, and validity periods.
- Therefore ODCS should primarily ingest into logical `dataset` entities. ODCS quality rules should be parsed and retained as ODCS-authored metadata on the logical dataset, but DataHub `assertion` entities should only be emitted when there is a resolved physical dataset to execute them against. ODCS may later feed bilateral contract templates, but ODCS itself should not be treated as a bilateral agreement (or Data Contract as many people think of it).

```mermaid
flowchart LR
    odcsFile["ODCS YAML"] --> logicalDataset["Logical Dataset\nplatform: odcs or logical"]
    odcsFile --> ruleSpecs["Quality rule specs\nstored as ODCS spec content"]
    physicalDataset["Physical Dataset\nSnowflake, Postgres, etc"] --> logicalDataset
    physicalDataset --> physicalFieldLinks["Schema field logicalParent links"]
    ruleSpecs --> executableAssertions["Executable assertions\nonly with physical binding"]
    executableAssertions --> physicalDataset
    bilateralContract["Future Bilateral Contract"] --> physicalDataset
    bilateralContract --> executableAssertions
    bilateralContract --> parties["Producer and Consumer\nowners, sign-off, lifecycle"]
```

## Concrete Entity And Aspect Mapping

The MVP should avoid the phrase “ODCS spec content” unless it is tied to concrete DataHub writes. The proposed mapping is:

1. ODCS logical model entity
   - Entity: `dataset`
   - URN: `urn:li:dataset:(urn:li:dataPlatform:odcs,<stable-logical-name>,<env>)`
   - Purpose: canonical logical dataset/interface described by the ODCS file.

2. Aspects on the ODCS logical `dataset`
   - `datasetProperties`: display name, description, and small ODCS provenance fields in `customProperties`, such as `odcs.id`, `odcs.version`, `odcs.apiVersion`, `odcs.status`, `odcs.sourceFile`, `odcs.schemaName`, and `odcs.physicalName` if present.
   - `schemaMetadata`: canonical schema from ODCS `schema[].properties[]`. This should be the source-authored logical schema, not only editable metadata.
   - `globalTags`: ODCS top-level and schema-level tags, with any configured prefix.
   - `ownership`: ODCS `team[]` entries that map cleanly to DataHub owners.
   - `glossaryTerms` / `structuredProperties`: use these only where the mapping is deliberate and configured. Do not invent broad mappings for every ODCS field in the MVP.
   - `institutionalMemory`: optional link back to the ODCS source file or repository location, if available.

3. ODCS quality rules without a physical binding
   - Do not emit `assertion` entities.
   - Retain rule provenance through one of these explicit approaches:
     - MVP/simple: keep the original ODCS YAML externally addressable via `institutionalMemory`, and store only summary counters or identifiers in `datasetProperties.customProperties`, e.g. `odcs.qualityRuleCount`.

4. Physical dataset binding
   - Entity: existing or resolvable physical `dataset`, e.g. Snowflake/Postgres/BigQuery.
   - Aspect on physical dataset: `logicalParent`, pointing to the ODCS logical dataset.
   - Relationship produced: `PhysicalInstanceOf`.
   - Aspect on physical `schemaField` entities: `logicalParent`, pointing to the matching logical `schemaField` where column mapping is known.

5. Assertions with a physical binding
   - Entity: `assertion`
   - Aspects: existing `AssertionInfo` subtypes, e.g. `FieldAssertionInfo`, `VolumeAssertionInfo`, `FreshnessAssertionInfo`, `SqlAssertionInfo`, or `CustomAssertionInfo`.
   - Target entity inside the assertion: the physical dataset URN, not the logical dataset URN.
   - Linkage/provenance: use assertion `customProperties` such as `odcs.id`, `odcs.rule.name`, `odcs.rule.scope`, and the ODCS logical dataset URN so users can trace the assertion back to the logical ODCS model.

rbrenton:
Thanks @AdrianMachado. The reframing makes sense to me. ODCS v3 reads as a producer-published dataset spec rather than a bilateral agreement, so modeling it as a logical dataset rather than a `dataContract` seems like the better fit, especially given the deprecation direction for `dataContract`. Here's the mapping I'm proposing, and I'd like to confirm it with you and @jayacryl before I start.

**1. ODCS to a logical `dataset`** on `urn:li:dataPlatform:odcs`
- `datasetProperties`: name, description, and ODCS provenance in `customProperties` (`odcs.id`, `odcs.version`, `odcs.apiVersion`, `odcs.status`, `odcs.sourceFile`, `odcs.schemaName`, `odcs.qualityRuleCount`, optional `odcs.physicalName`)
- `schemaMetadata`: canonical schema from `schema[].properties[]`, replacing the current `editableSchemaMetadata`
- `globalTags`, `ownership`, `institutionalMemory` (link to the source YAML)
- a `dataPlatformInfo` MCP to register the `odcs` platform

**2. Drop `dataContract`** emission.

**3. Quality rules**: emit assertions only when a physical dataset binding resolves, targeting the physical dataset URN. The existing routing (FieldValues/Volume/Sql/Custom) is target-agnostic, so it carries over. Each assertion keeps provenance in `customProperties` (`odcs.id`, `odcs.rule.name`, `odcs.rule.scope`, logical URN). With no binding, no assertion entities are emitted; provenance is kept via `institutionalMemory` and `odcs.qualityRuleCount`.

**4. Physical binding**: `logicalParent` on the physical dataset (and schemaField where columns map), giving the `PhysicalInstanceOf` relationship to the logical ODCS dataset.

A few open questions before I start, so we don't round-trip:

1. @jayacryl, you're the assigned reviewer. Are you good with this direction, or is there anything you'd frame differently first?
2. Platform name: `odcs` or the generic `logical` platform for the logical dataset URN? I have a slight preference for `odcs` for clearer provenance, but I'm happy either way.
3. Logical Models are in private beta and gated behind `LOGICAL_MODELS_ENABLED` (off by default). For a standalone ODCS file with no physical table yet, a default OSS user would see a logical dataset that doesn't render, and under strict gating, no assertions. I think that's acceptable if we document the "ingest ODCS plus the physical platform, enable the flag" workflow, and I can add a startup warning when the flag is off. Does that match your expectations for the MVP?
4. The `group:` ownership prefix from your earlier thread is a DataHub-side extension; ODCS has no group principal (`team[].username` is "username or email" only). I'd keep it and document it as a DataHub extension so it isn't mistaken for ODCS-native. Let me know if you'd rather drop it.

From the earlier round (`f7e0167`): dropped `_ROUTE_TO_CUSTOM`, moved rule-kinds to module-level frozensets, and deferred NOT_BETWEEN. I'll start the rewrite once we're aligned.

AdrianMachado:
> Thanks @AdrianMachado. The reframing makes sense to me. ODCS v3 reads as a producer-published dataset spec rather than a bilateral agreement, so modeling it as a logical dataset rather than a `dataContract` seems like the better fit, especially given the deprecation direction for `dataContract`. Here's the mapping I'm proposing, and I'd like to confirm it with you and @jayacryl before I start.
> 
> **1. ODCS to a logical `dataset`** on `urn:li:dataPlatform:odcs`
> 
> * `datasetProperties`: name, description, and ODCS provenance in `customProperties` (`odcs.id`, `odcs.version`, `odcs.apiVersion`, `odcs.status`, `odcs.sourceFile`, `odcs.schemaName`, `odcs.qualityRuleCount`, optional `odcs.physicalName`)
> * `schemaMetadata`: canonical schema from `schema[].properties[]`, replacing the current `editableSchemaMetadata`
> * `globalTags`, `ownership`, `institutionalMemory` (link to the source YAML)
> * a `dataPlatformInfo` MCP to register the `odcs` platform
> 
> **2. Drop `dataContract`** emission.
> 
> **3. Quality rules**: emit assertions only when a physical dataset binding resolves, targeting the physical dataset URN. The existing routing (FieldValues/Volume/Sql/Custom) is target-agnostic, so it carries over. Each assertion keeps provenance in `customProperties` (`odcs.id`, `odcs.rule.name`, `odcs.rule.scope`, logical URN). With no binding, no assertion entities are emitted; provenance is kept via `institutionalMemory` and `odcs.qualityRuleCount`.
> 
> **4. Physical binding**: `logicalParent` on the physical dataset (and schemaField where columns map), giving the `PhysicalInstanceOf` relationship to the logical ODCS dataset.
> 
> A few open questions before I start, so we don't round-trip:
> 
> 1. @jayacryl, you're the assigned reviewer. Are you good with this direction, or is there anything you'd frame differently first?
> 2. Platform name: `odcs` or the generic `logical` platform for the logical dataset URN? I have a slight preference for `odcs` for clearer provenance, but I'm happy either way.
> 3. Logical Models are in private beta and gated behind `LOGICAL_MODELS_ENABLED` (off by default). For a standalone ODCS file with no physical table yet, a default OSS user would see a logical dataset that doesn't render, and under strict gating, no assertions. I think that's acceptable if we document the "ingest ODCS plus the physical platform, enable the flag" workflow, and I can add a startup warning when the flag is off. Does that match your expectations for the MVP?
> 4. The `group:` ownership prefix from your earlier thread is a DataHub-side extension; ODCS has no group principal (`team[].username` is "username or email" only). I'd keep it and document it as a DataHub extension so it isn't mistaken for ODCS-native. Let me know if you'd rather drop it.
> 
> From the earlier round (`f7e0167`): dropped `_ROUTE_TO_CUSTOM`, moved rule-kinds to module-level frozensets, and deferred NOT_BETWEEN. I'll start the rewrite once we're aligned.

1. Sounds great to me, and the link to the source YAML sounds useful too. Do you think there should be some sort of snapshot of the YAML inside the `institutionalMemory` as well? I assume the YAML could potentially evolve but maybe the reingestion has not, so there are some temporary unsynced states? 
2. 👍 
3. 👍 
4. 👍 

Now for your questions

1. I chatted with Jay offline and he's good with this approach too
2. Let's keep using ODCS. I left a minor comment on your data platform PR, but I think its good
3. That sounds right. I would say that logical models (and also ODCS) are not for the faint of heart in the data world. We will eventually roll both out more broadly.
4. Ah, i guess resolving an ODCS `team` to a DataHub `group` can get quite tricky so we can skip that part. For individuals, matching on email seems reasonable.

rbrenton:
@AdrianMachado  The PR has been reworked to match the new framing. ODCS now ingests into logical dataset entities, not data contracts:

- Each ODCS `schema[]` entry --> a logical dataset `urn:li:dataset:(urn:li:dataPlatform:odcs,<logical-name>,<env>)`.
- ODCS quality rules are retained as ODCS-authored metadata on the logical dataset.
- DataHub `assertion` entities are emitted only when a physical dataset binding is resolved; unbound entries still emit the logical dataset (with an `unmapped_reason`) but no assertions.
- When bound, physical datasets link to the logical dataset via `logicalParent`.

This keeps ODCS as a spec/interface for the dataset (your "OpenAPI for datasets" framing) and leaves the door open to feeding bilateral contract templates later without treating ODCS itself as a Data Contract. Ready for another look when you have time.

AdrianMachado:
Sorry for letting this languish - had some personal stuff that required me to take some time off. Instead of back and forth with review, I am just going to push the final bits here before merging

AdrianMachado (approved):

AdrianMachado:
@rbrenton heads up — I pushed directly to your branch (thanks for leaving maintainer edits on). Your history is untouched: it's a merge of current master plus a series of fix commits on top of `23d717a`, so a plain `git pull` will fast-forward your local branch. New head is `0230994`.

What changed and why:

**Quality routing is now spec-exact.** The previous routing keyed on `notNull`/`unique`, which aren't ODCS library rules in any version — a conformant v3.1 file (`metric: nullValues` etc.) got zero typed assertions. The model now parses `metric` (v3.1), `rule` (v3.0 canonical / v3.1 deprecated alias), `id`, `arguments`, and `validValues`, and routes the real vocabulary: `nullValues`, `duplicateValues`/`duplicateCount`, `invalidValues`/`validValues` (incl. `pattern` → REGEX_MATCH), `rowCount`. Anything not exactly representable (e.g. `mustNotBeBetween`, duplicate tolerances, `missingValues`) is preserved as a custom assertion rather than approximated. The `column` field and top-level `quality[]` are gone (neither exists in the spec).

**Assertions now attach to the logical dataset, not the physical one.** This is a design change from what we'd discussed earlier in the thread: quality rules and a new `DATA_SCHEMA` schema-compliance assertion (compatibility `SUPERSET` by default) are emitted for every `schema[]` entry regardless of binding, so contract-first files keep their expectations. Propagation onto physical instances happens on the DataHub side via the `PhysicalInstanceOf` relationship — the source never writes assertions against physical URNs. Assertion URNs are seeded from `quality.id`, and each assertion carries `odcs.*` provenance plus a `dataPlatformInstance` aspect.

**Physical binding is derived from the contract itself.** `servers[].type` is required by the spec, so the platform comes from a built-in type map and the table name is fully qualified from the server's own fields (`customers_db.public.customers`, not bare `customers` — the old URNs would never have matched the platform-of-record source). Snowflake lowercases by default; `server_overrides` refines env/platform_instance; `physical_urn_overrides` is now keyed by schema-entry name instead of list position. With a graph attached, derived URNs are verified to exist before the `logicalParent` link is emitted — no stub datasets.

**Stateful ingestion replaces the state file.** The local-JSON state file broke in managed/UI ingestion and could mass-soft-delete a contract's entities when its file hit a transient read error. It's now the standard `stateful_ingestion` block (StatefulIngestionSourceBase + StaleEntityRemovalHandler); the `logicalParent` and platform-info workunits are marked `is_primary_source=false` so stale removal can never touch physical datasets.

**Unknown-field handling is calibrated to the spec.** Valid-but-unmapped fields (`slaProperties`, `support`, `relationships`, …) produce one info per file instead of per-field "check spelling" warnings; a drift-guard test pins the models + whitelists against the vendored JSON schemas. Also: per-table descriptions now win over the contract description, root `authoritativeDefinitions` land in institutionalMemory, and the `group:` owner prefix is dropped as agreed.

Fixtures were rewritten to be spec-conformant — `odcs_full_v31` now passes `strict_validation: true` (fun fact discovered along the way: v3.1 requires exactly one comparison operator per library rule), and there's a new v3.0.2 fixture for the old vocabulary. 111 tests, mypy/ruff clean.

Deferred, deliberately: schemaField-level `logicalParent` column links, v3.1 `relationships` (FKs), SLA → freshness mapping, and multi-server contracts (first mappable server wins). Happy to file follow-ups for these.

---

PR #17336, opened 2026-05-07
perf(frontend): reduce listRecommendations fan-out and consolidate home-page queries

ani-malgari:
- In entityDataProduct fragment, replaced domain { ...entityDomain } (which spread domainEntitiesFields → 4 ES count queries per related DataProduct) with a minimal domain { domain { urn type properties { name } } }. This eliminates the primary source of ES fan-out in listRecommendations — measured locally at −65% server response time (1,040 ms → 368 ms) with only 1–2 DataProducts present; scales linearly with more DataProducts.
- Introduced useHomeRecommendations hook as a single shared listRecommendations caller for the home page. Migrated useGetPlatforms, useGetDomains, PopularGlossaryTerms, useGetRecentActions, and useRecentlyViewedEntities to consume it, consolidating what were previously up to 5 separate {scenario: HOME} GraphQL requests (with different limit values causing distinct Apollo cache keys) into 1 per page load.
- Changed fetchPolicy: 'no-cache' → 'cache-first' on the listRecommendations call in IntroduceYourselfMainContent, which was forcing an unnecessary round-trip on every onboarding page render.

Results with minimal local data. Performance scales linearly with the number of data products & domains.
Metric | Before (original) | After | Delta
-- | -- | -- | --
Server response time | 1,040 ms | 368 ms | −65% / saved 672 ms
Payload size | 3.7 KB | 3.6 KB | −0.1 KB (negligible)

**Before:**
Screenshot 2026-05-06 at 4 48 58 PM

**After:**
Screenshot 2026-05-06 at 4 38 45 PM

**Note:**
Meticulous shows empty Domains/Platforms tiles in "after". This is a known artifact of changing the entityDataProduct fragment, which alters the listRecommendations operation hash; Meticulous's recorded mocks no longer match. Verified locally that tiles render correctly. Diffs accepted in Meticulous

### DIFF
--- a/datahub-web-react/src/app/homeV2/content/recent/useGetRecentActions.tsx
+++ b/datahub-web-react/src/app/homeV2/content/recent/useGetRecentActions.tsx
@@ -5,7 +5,7 @@ import {
     RECOMMENDATION_MODULE_ID_RECENTLY_VIEWED_ENTITIES,
 } from '@src/app/entityV2/shared/constants';
 
-import { CorpUser, Entity, EntityType } from '@types';
+import { Entity, EntityType } from '@types';
 
 const SUPPORTED_ENTITY_TYPES = [
     ...ASSET_ENTITY_TYPES,
@@ -14,7 +14,7 @@ const SUPPORTED_ENTITY_TYPES = [
     EntityType.GlossaryTerm,
 ];
 
-export const useGetRecentActions = (_user?: CorpUser | null) => {
+export const useGetRecentActions = () => {
     const { modules, loading, refetch } = useHomeRecommendations();
 
     const viewedModule = modules?.find(
--- a/datahub-web-react/src/app/homeV2/content/recent/useGetRecentActions.tsx
+++ b/datahub-web-react/src/app/homeV2/content/recent/useGetRecentActions.tsx
@@ -1,11 +1,11 @@
+import { useHomeRecommendations } from '@app/homeV2/useHomeRecommendations';
 import { ASSET_ENTITY_TYPES } from '@app/searchV2/utils/constants';
 import {
     RECOMMENDATION_MODULE_ID_RECENTLY_EDITED_ENTITIES,
     RECOMMENDATION_MODULE_ID_RECENTLY_VIEWED_ENTITIES,
 } from '@src/app/entityV2/shared/constants';
 
-import { useListRecommendationsQuery } from '@graphql/recommendations.generated';
-import { CorpUser, Entity, EntityType, ScenarioType } from '@types';
+import { CorpUser, Entity, EntityType } from '@types';
 
 const SUPPORTED_ENTITY_TYPES = [
     ...ASSET_ENTITY_TYPES,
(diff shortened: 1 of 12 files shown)


datahub-web-react/src/app/homeV2/content/recent/useGetRecentActions.tsx
```diff
@@ -14,35 +14,23 @@ const SUPPORTED_ENTITY_TYPES = [
...
-                requestContext: {
-                    scenario: ScenarioType.Home,
-                },
-                limit: 10,
-            },
-        },
-        fetchPolicy: 'cache-first',
-        skip: !user?.urn,
-    });
+export const useGetRecentActions = (_user?: CorpUser | null) => {
```
chriscollins3456: so this is only ever used on home page v2 which nobody has on by default, maybe one or two are still using this but most are not

wondering why is this getting fired off at all?

maybe what's happening is that before the feature flag loads we load the old homepage briefly and fire off requests that we never use?

maybe that's happening in more places than we know in the app and we're just wasting lots of graphql calls

chriscollins3456 (approved):
makes sense!

i see most of these will help our old home page though. my main wonder is if we were firing them anyways even with the new home page flag on?

datahub-web-react/src/app/homeV2/content/recent/useGetRecentActions.tsx
benjiaming: Nit: if a variable isn't used, we should probably drop it instead of adding `_` prefix.

datahub-web-react/src/app/homeV2/content/tabs/discovery/sections/platform/useGetPlatforms.tsx
```diff
@@ -13,35 +12,20 @@ type PlatformAndCount = {
...
-        fetchPolicy: 'cache-first',
-        skip: !user?.urn,
-    });
+    const { modules, loading } = useHomeRecommendations();
 
-    const platformsModule = data?.listRecommendations?.modules?.find(
-        (module) => module.moduleId === PLATFORMS_MODULE_ID,
-    );
+    const platformsModule = modules?.find((module) => module.moduleId === PLATFORMS_MODULE_ID);
+    const limit = maxToFetch || MAX_PLATFORMS_TO_FETCH;
```
benjiaming: There 's a subtle bug: if a caller (e.g. `homeV3/modules/platforms/PlatformsModule.tsx:29`) passes in optional `maxToFetch`, this will still load 10 only items.

Because `useHomeRecommendations()` hardcodes limit 10 internally.

datahub-web-react/src/app/homeV2/content/recent/useGetRecentActions.tsx
ani-malgari: addressed in the latest commit

datahub-web-react/src/app/homeV2/content/tabs/discovery/sections/platform/useGetPlatforms.tsx
ani-malgari: fixed this too

ani-malgari:
> makes sense!
> 
> i see most of these will help our old home page though. my main wonder is if we were firing them anyways even with the new home page flag on?

useGetPlatforms / useGetDomains / useRecentlyViewedEntities: used by both V2 and V3 components, so consolidation directly benefits the new home page.
useGetRecentActions / PopularGlossaryTerms: V2-only and consolidating them is essentially a no-op for V3 users (a small win for the few V2 holdouts and zero downside, so I left it in this PR for completeness).

datahub-web-react/src/app/homeV2/content/recent/useGetRecentActions.tsx
ani-malgari: With the redesign flag on (the default for everyone now), these don't fire at all on a normal home-page visit, except for when the localstorage's showHomePageRedesign is false and server has it true (when it's just flipped). I left it in this PR for completeness.

benjiaming (approved):
Great work!

PR #17339, opened 2026-05-07
fix(ui): fix table header appearing over modal

purnimagarg1:
**Linear ticket:**
https://linear.app/acryl-data/issue/CAT-1773/visible-bug-with-modals-and-table-headers

**Screenshots:**

Before:
image

After:
image

### DIFF
--- a/datahub-web-react/src/app/entityV2/shared/EntityDropdown/EntityDropdown.tsx
+++ b/datahub-web-react/src/app/entityV2/shared/EntityDropdown/EntityDropdown.tsx
@@ -517,6 +517,7 @@ const EntityDropdown = (props: Props) => {
                     urns={[urn]}
                     onClose={() => setIsDeprecationModalVisible(false)}
                     refetch={refetchForEntity}
+                    zIndexOverride={101}
                 />
             )}
             {isEntityAnnouncementModalVisible && (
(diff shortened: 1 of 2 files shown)


datahub-web-react/src/app/entityV2/shared/EntityDropdown/EntityDropdown.tsx
```diff
@@ -517,6 +517,7 @@ const EntityDropdown = (props: Props) => {
                     urns={[urn]}
                     onClose={() => setIsDeprecationModalVisible(false)}
                     refetch={refetchForEntity}
+                    zIndexOverride={101}
```
chriscollins3456: so i think this is a bigger problem than just this modal but instead any of our modals are going to have a similar problem with tables. 

would there be a way to solve this for everything in one go? i definitely don't want to go through each modal and do something like this though.

purnimagarg1: The issue is not in our Modal component, rather `UpdateDeprecationModal`, which had a default z-index of 10 when not passed as a prop. Updated the default value for it.

chriscollins3456 (approved):

PR #17341, opened 2026-05-07
docs(athena): refine permissions policy

dennisliu-wag:
athena:ListDataCatalogs should be applied to all resources, updating policy.

Reference: https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonathena.html


askumar27 (approved):
Thanks for the contribution, LGTM!

PR #17342, opened 2026-05-07
fix(ingest/kafka-connect): avoid duplicated schema segment in sink lineage URNs

treff7es:
## Summary

`JdbcSinkConnector.extract_lineages` builds the target URN as `{db}.{schema}.{schema}.{table}` when the topic name already encodes the schema. For a topic named `public.users` with `db.schema=public` and the default `table.name.format=${topic}`, this produces a URN like `mydb.public.public.users` — pointing to a phantom dataset and orphaning lineage.

This PR detects when `table_name` already starts with `{schema_name}.` and skips the duplicate prepend, so the lineage URN matches the actual destination table the sink writes to. Affects 3-tier JDBC sinks (Postgres, SQL Server, Oracle).

The Snowflake sink path is unchanged: `SnowflakeSinkConnector` already handles `snowflake.topic2table.map` correctly in both branches (mapped and default `re.sub` substitution). Pin tests for both branches are included so this doesn't silently regress.

## Changes

- `metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py` — `JdbcSinkConnector.extract_lineages` deduplicates the schema prepend.
- `metadata-ingestion/tests/unit/test_kafka_connect_double_schema_prefix.py` — 3 new regression tests.

## Test plan

- [x] `pytest tests/unit/test_kafka_connect_double_schema_prefix.py` — 4 pass
- [x] `pytest -k kafka_connect tests/unit/` — 510 pass, 0 regressions

## Scope

| Sink | Affected? | Notes |
|---|---|---|
| `JdbcSinkConnector` (Postgres, SQL Server, Oracle) | **Yes** — fixed | 3-tier; topic with `.` + default `${topic}` |
| `JdbcSinkConnector` (MySQL) | No | 2-tier — no schema prepend |
| `SnowflakeSinkConnector` | No (already correct) | `topic2table.map` handled; pin tests added |
| `BigQuerySinkConnector` | Theoretical edge case only | Not touched |
| `ClickHouseSinkConnector` | No | 2-tier |
| `ConfluentS3SinkConnector` | No | Path-based URN |

## Caveat

The fix assumes the Kafka Connect sink connector strips the schema prefix from the topic before writing (e.g. via `RegexRouter` SMT) — which is the common pattern. For an unusual setup where the destination table is *literally* named `public.users` (with a dot in the name), the URN would be incorrect, but no such case is known in practice.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
@@ -1243,8 +1243,16 @@ class JdbcSinkConnector(BaseConnector):
                 if parser.schema_name and has_three_level_hierarchy(
                     parser.target_platform
                 ):
-                    # Platform supports schema hierarchy: database.schema.table
-                    table_with_schema = f"{parser.schema_name}.{table_name}"
+                    schema_prefix = f"{parser.schema_name}."
+                    if table_name.startswith(schema_prefix):
+                        # Topic-derived table name already encodes the schema
+                        # (e.g. topic "public.clm" with schema_name="public"
+                        # and the default table.name.format="${topic}").
+                        # Prepending the schema again would duplicate the
+                        # segment in the lineage URN.
+                        table_with_schema = table_name
+                    else:
+                        table_with_schema = f"{parser.schema_name}.{table_name}"
                     target_dataset = get_dataset_name(
                         parser.database_name, table_with_schema
                     )
--- a/metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
@@ -1066,26 +1066,46 @@ class JdbcSinkConnector(BaseConnector):
             self.connector_manifest, self.platform
         )
 
-    def get_table_name_from_topic(self, topic: str, table_format: str) -> str:
+    def get_table_name_from_topic(
+        self,
+        topic: str,
+        table_format: str,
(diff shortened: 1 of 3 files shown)


metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
```diff
@@ -1247,8 +1247,16 @@ def extract_lineages(self) -> List[KafkaConnectLineage]:
                 if parser.schema_name and has_three_level_hierarchy(
                     parser.target_platform
                 ):
-                    # Platform supports schema hierarchy: database.schema.table
-                    table_with_schema = f"{parser.schema_name}.{table_name}"
+                    schema_prefix = f"{parser.schema_name}."
+                    if table_name.startswith(schema_prefix):
```
sgomezvillamor: what about pushing this logic to the `get_table_name_from_topic` method itself?
so `get_table_name_from_topic` already provides a table name without schema, and so here we just pre-append the schema, as we were doing already

sgomezvillamor (approved):
just left a very minor suggestion

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
treff7es: Good call — done in d5e855f85e. The dedup now lives in `get_table_name_from_topic` (it accepts an optional `schema_name` and strips a leading `{schema_name}.` from the derived table name). The caller goes back to its original simple `f"{schema_name}.{table_name}"` concatenation. Tests still pass.

PR #17348, opened 2026-05-07
fix(ingest/glue): fix schemaField URN and definition ordering for extract_column_parameters

JohnRTurner:
## Summary

Follow-up to #17325. Two bugs found during testing that prevented `extract_column_parameters` from working:

- **Wrong schemaField URN**: `_get_column_param_workunits` was calling `make_schema_field_urn(dataset_urn, column["Name"])`, using the bare column name (e.g. `emp_std_id`) as the field path. The actual `schemaField` entity in DataHub uses the v2 typed path generated by `get_schema_fields_for_hive_column` (e.g. `[version=2.0].[type=string].emp_std_id`). This meant all `structuredProperties` MCPs were written to shadow entities that don't correspond to any visible schema field in the UI. Fix: derive `field_path` from `schema_fields[0].fieldPath`, exactly as `_get_schema_metadata` does.

- **Batch ordering**: `StructuredPropertyDefinition` and `structuredProperties` MCPs were both yielded as pipeline workunits and landed in the same async batch to GMS. GMS validates assignments against the current DB state (pre-batch), so definitions submitted in the same batch are not visible when assignments are validated — resulting in 422 rejections for all `structuredProperties` MCPs. Fix: emit definitions synchronously via `ctx.graph` when available, ensuring they are persisted in GMS before the assignment MCPs arrive. Falls back to yielding as a workunit when `ctx.graph` is `None` (file sink, dry-run).

## Test plan

- [x] E2e tested against a real Glue catalog + local DataHub instance: all column `Parameters` now appear as structured properties on the correct `schemaField` entities in the Properties tab
- [x] `./gradlew :metadata-ingestion:lintFix` passes

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
@@ -2016,8 +2016,16 @@ class GlueSource(StatefulIngestionSourceBase):
                 params = column.get("Parameters")
                 if not params:
                     continue
+                schema_fields = get_schema_fields_for_hive_column(
+                    hive_column_name=column["Name"],
+                    hive_column_type=column.get("Type", "string"),
+                    description=column.get("Comment"),
+                    default_nullable=True,
+                )
+                if not schema_fields:
+                    continue
                 field_urn = mce_builder.make_schema_field_urn(
-                    dataset_urn, column["Name"]
+                    dataset_urn, schema_fields[0].fieldPath
                 )
                 assignments = []
                 for key, value in params.items():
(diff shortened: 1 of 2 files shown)


metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
```diff
@@ -2026,17 +2034,23 @@ def _get_column_param_workunits(
...
-                        ).as_workunit()
-                        # Only cache after successful yield so a dropped MCP
-                        # (network blip, GMS rejection) doesn't prevent retry.
+                        )
+                        if self.ctx.graph is not None:
+                            # Emit synchronously so GMS persists the definition
+                            # before the structuredProperties MCP below is
+                            # validated — GMS rejects assignments that reference
+                            # a definition not yet in the database.
+                            self.ctx.graph.emit_mcp(definition_mcp)
```
treff7es: If you want to be absolute sure about it you should define Emit mode as well
```suggestion
                            self.ctx.graph.emit_mcp(definition_mcp, emit_mode=EmitMode.SYNC_PRIMARY)
```

treff7es (approved):

PR #17356, opened 2026-05-08
fix(graphql): add missing SIT and SBX FabricType enum values

rospe:
These were added to the PDL model in #13842 (June 2025) but the GraphQL schema was never updated to match, causing BAD_REQUEST errors when querying entities with SIT or SBX environment fabrics.


chakru-r (approved):

chakru-r:
LGTM, Thank you for the contribution.

PR #17357, opened 2026-05-08
fix(ingest/mode): gate chart fetch on chart_count, not explorations_count

alfiyas-datahub:
## Summary

The Mode connector had a perf optimization that skipped calling the chart API when a query's `explorations_count` was 0. In Mode's API, explorations are private user-level analyses, unrelated to the charts published on a report — the right field is `chart_count`. For any report whose queries had `explorations_count: 0` but real published charts, the connector silently:

- emitted `dashboardInfo` with `charts: []`
- never wrote chart entities for those charts
- never wrote chart-level `upstreamLineage`

so the lineage chain `Dashboard → Charts → Query → Upstream tables` broke and the UI's Lineage tab looked empty.

This PR switches the gate to `chart_count`, exposes `chart_count` in query custom properties for observability, updates the integration fixture to include `chart_count` (mirroring real Mode API responses), and adds focused regression tests.

## Root cause

`mode.py` `_process_report_inner`:

```python
# before
if query.get("explorations_count", None) == 0:
    charts = []
    self.report.chart_api_calls_skipped += 1
else:
    charts = self._get_charts(...)
```

`explorations_count` was introduced in #16300 (Mar 2026) as the gate. The test fixture `tests/integration/mode/setup/queries.json` had `explorations_count: 1`, so the skip path was never exercised in tests.

## Change

```python
# after
# Gate on chart_count (published charts), not explorations_count
# (private user analyses), to avoid silently dropping report charts.
if query.get("chart_count", 0) == 0:
    charts = []
    self.report.chart_api_calls_skipped += 1
else:
    charts = self._get_charts(...)
```

## Refs
- Linear CUS-8603 / ING-2323

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/mode.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/mode.py
@@ -1159,6 +1159,7 @@ class ModeSource(StatefulIngestionSourceBase):
                     "last_run_id",
                     "data_source_id",
                     "explorations_count",
+                    "chart_count",
                     "report_imports_count",
                     "dbt_metric_id",
                 ],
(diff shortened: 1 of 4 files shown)


metadata-ingestion/src/datahub/ingestion/source/mode.py
```diff
@@ -1159,6 +1159,7 @@ def construct_query_or_dataset(
                     "last_run_id",
                     "data_source_id",
                     "explorations_count",
```
puneetagarwal-datahub, line 1161: check if we can remove explorations_count if we dont have any logic to filter by this

puneetagarwal-datahub (approved):

PR #17360, opened 2026-05-08
fix(ingest/dataplex): use search_projects for project_id_pattern discovery

davidwalker-tp:
Cloud Resource Manager v3 `projects.list` requires a `parent` arg (folder/org), so calling it without one returned `400 invalid parent name` and broke `project_id_pattern`-only auto-discovery for the Dataplex and VertexAI sources. Switch the shared helper to `search_projects()`, which has no parent requirement and only needs `resourcemanager.projects.get` per project. Update Dataplex docs to recommend `roles/browser` instead of `roles/resourcemanager.folderViewer`, and add the same guidance to the VertexAI prerequisites.


sgomezvillamor (approved):
Thanks for the fix — this is correct and well-scoped. Approving.

---

sgomezvillamor:
Many thanks for this fix! We were able to confirm internally both the original issue and that the fix resolves it.

**On breaking changes**: there is no risk of a breaking change here. The `project_id_pattern` auto-discovery path (path 3 in the precedence chain) was returning `400 INVALID_ARGUMENT` for every caller since `projects.list` v3 requires a `parent` argument that was never supplied. The feature was entirely non-functional, so switching to `search_projects` restores rather than changes behaviour. The `project_labels` path was already using `search_projects` and is unaffected.

**Suggestions (non-blocking):**

1. **Audit visibility** — consider adding a `report.info` trace at the end of `resolve_gcp_projects` in `gcp_project_filter.py` to surface the final resolved project list in the DataHub ingestion UI. Note that `context` expects `Optional[str]`, so the list needs to be serialised (e.g. `context=str([p.id for p in allowed])`), and ideally the log is emitted at all three resolution paths, not just the fallback.

2. **Minor doc nit** — the docstring on `resolve_gcp_projects` still says *"list all projects"* in the precedence comment; could be updated to *"search all visible projects"* for consistency with the new implementation.

3. **Minor doc nit** — the Dataplex IAM table still lists `roles/resourcemanager.folderViewer` for the *"Term-asset associations"* row. Since `roles/browser` also grants `resourcemanager.projects.get`, it could be mentioned there as a lighter-weight alternative.

---

davidwalker-tp:
Thanks for the review @sgomezvillamor! Pushed `39aadba47d` addressing the suggestions:

1. **Audit visibility** — done. `resolve_gcp_projects` now emits `report.info` at each of the three resolution paths with the resolved project IDs in `context` and a path-tagged `message` (`"...via explicit project_ids"` / `"...via project_labels and project_id_pattern"` / `"...via project_id_pattern"`). Also covered by new assertions in `test_resolve_explicit_project_ids`, `test_resolve_with_labels_and_pattern`, and `test_resolve_all_projects`.

2. **Docstring nit on `resolve_gcp_projects`** — already updated in the original commit (`cca39c0173`). Line 210 reads *"search all visible projects, then apply project_id_pattern"*. Worth a re-check; otherwise no diff in this commit for it.

3. **Term-asset associations row** — done. The Dataplex IAM table now lists `roles/browser` (lighter-weight) as the primary recommendation, with `roles/resourcemanager.folderViewer` as the alternative; both grant `resourcemanager.projects.get`.

sgomezvillamor (approved):

PR #17361, opened 2026-05-08
ci(playwright): integrate Playwright E2E tests into docker-unified CI pipeline

priyabratadas-dh:
## Summary

- Adds **Playwright E2E test jobs** (`playwright_test`, `playwright_report`) to `docker-unified.yml` so Playwright runs in the same pipeline as smoke tests, using the same built images
- Adds a **`playwright-change` path filter** to `.github/actions/ci-optimization/action.yml` so PRs that touch only `e2e-test/**` files still trigger the Playwright jobs
- Refactors `seeding.fixture.ts` to remove the now-redundant `lineage-seeder.ts` import and `LINEAGE_RESEED_FEATURE_NAME` constant (superseded by `extractComplexAspects`/`ingestComplexAspects` in `seeder-utils.ts`)
- Deletes `e2e-test/ui/playwright/utils/lineage-seeder.ts` — its functionality is fully covered by `seeder-utils.ts`

### `docker-unified.yml`

| Addition | Purpose |
|---|---|
| `playwright_shard_count` workflow_dispatch input | Override shard count (default 4) on manual runs |
| `playwright_yarn_cache_key` / `playwright_yarn_cache_key_prefix` setup outputs | Separate Yarn cache keyed on `e2e-test/ui/playwright/package.json`, independent of the Cypress cache |
| `playwright_matrix` setup output | Shard matrix; populated when `backend-change`, `frontend-change`, `smoke-test-change`, or `playwright-change` is true, or on any non-PR event (push/release/workflow_dispatch) |
| `playwright_change` setup output | True when `e2e-test/**` files changed |
| `playwright_test` job | Sharded Playwright runner — pulls built images from depot (or builds locally on forks), starts quickstart, installs Playwright + Chromium, runs `npx playwright test --shard=N/M`; uploads blob reports, JUnit XML, Docker logs, and traces on failure |
| `playwright_report` job | Merges all shard blob reports into a single unified HTML report via `playwright merge-reports` and uploads as `playwright-html-report` |

### `.github/actions/ci-optimization/action.yml`

- New `e2e-test` paths filter: `e2e-test/**`
- New `playwright-change` output: `true` when e2e-test files changed or trigger is `workflow_dispatch`

## Test plan

- [x] Verify `playwright_test` shards appear in CI on a PR that touches `e2e-test/**`
- [x] Verify `playwright_test` runs on push to master
- [x] Verify `playwright_report` produces a merged HTML report artifact
- [x] Verify playwright jobs are skipped on ingestion-only / connector-source-only PRs


.github/actions/ci-optimization/action.yml
```diff
@@ -114,9 +117,14 @@ runs:
             - "docker/kafka-setup/**"
           smoke-test:
             - "smoke-test/**"
+            - "e2e-test/**"
+            - ".github/workflows/docker-unified.yml"
+            - ".github/actions/ci-optimization/**"
```
devashish2203: Remove ci-optimization and docker-unified from here.

.github/actions/ci-optimization/action.yml
```diff
@@ -114,9 +117,14 @@ runs:
...
           actions:
             - "datahub-actions/**"
             - "docker/datahub-actions/**"
           java-client:
             - "metadata-integration/java/datahub-client/**"
             - "metadata-models/**"
+          e2e-test:
```
devashish2203: Make this specific to Playwright for now.

.github/workflows/docker-unified.yml
```diff
@@ -197,6 +206,42 @@ jobs:
...
+      - name: Build Playwright shard matrix
+        id: set-playwright-matrix
+        run: |
+          shard_count=${{ github.event.inputs.playwright_shard_count || '4' }}
+          includes=''
+          # Run playwright on backend/frontend/smoke-test changes, or on any non-PR event.
+          # backend_change and frontend_change already embed the || github.event_name != 'pull_request' expansion.
+          if [[ "${{ steps.ci-optimize.outputs.backend-change }}" == 'true' || \
+                "${{ steps.ci-optimize.outputs.frontend-change }}" == 'true' || \
+                "${{ steps.ci-optimize.outputs.smoke-test-change }}" == 'true' || \
```
devashish2203: Don't need to run playwright for smoke test only changes.

.github/workflows/docker-unified.yml
```diff
@@ -197,6 +206,42 @@ jobs:
...
+                "${{ steps.ci-optimize.outputs.playwright-change }}" == 'true' || \
+                "${{ github.event_name }}" != 'pull_request' ]]; then
+            matrix=''
+            for ((i = 1; i <= shard_count; i++)); do
```
devashish2203: Shard count is a string above, probably won't work in the for loop.

.github/workflows/docker-unified.yml
```diff
@@ -197,6 +206,42 @@ jobs:
...
+        run: |
+          shard_count=${{ github.event.inputs.playwright_shard_count || '4' }}
+          includes=''
+          # Run playwright on backend/frontend/smoke-test changes, or on any non-PR event.
+          # backend_change and frontend_change already embed the || github.event_name != 'pull_request' expansion.
+          if [[ "${{ steps.ci-optimize.outputs.backend-change }}" == 'true' || \
+                "${{ steps.ci-optimize.outputs.frontend-change }}" == 'true' || \
+                "${{ steps.ci-optimize.outputs.smoke-test-change }}" == 'true' || \
+                "${{ steps.ci-optimize.outputs.playwright-change }}" == 'true' || \
+                "${{ github.event_name }}" != 'pull_request' ]]; then
```
devashish2203: Can't this be set as a condition in if: for this step?

.github/workflows/docker-unified.yml
```diff
@@ -197,6 +206,42 @@ jobs:
...
+            matrix=''
+            for ((i = 1; i <= shard_count; i++)); do
+              if [[ -n "$matrix" ]]; then
```
devashish2203: Will this condition be true ever?

.github/workflows/docker-unified.yml
```diff
@@ -842,6 +887,211 @@ jobs:
           path: cypress-html-report/
           retention-days: 5
 
+  playwright_test:
```
devashish2203, line 909: Add an if condition to make this not run on pull_requests for now.

.github/workflows/playwright-e2e-tests.yml
```diff
@@ -19,6 +19,7 @@ concurrency:
 
 env:
   PROFILE_NAME: "${{ github.event.inputs.profileName || 'quickstart-consumers' }}"
+  ECR_REGISTRY: "795586375822.dkr.ecr.us-west-2.amazonaws.com"
```
devashish2203: Only for fork. Should be removed from OSS.

.github/workflows/playwright-e2e-tests.yml
```diff
@@ -84,6 +86,22 @@ jobs:
           echo "yarn_cache_key=playwright-e2e-${{ runner.os }}-yarn-${{ hashFiles('./e2e-test/ui/playwright/package.json') }}" >> "$GITHUB_OUTPUT"
           echo "yarn_cache_key_prefix=playwright-e2e-${{ runner.os }}-yarn-" >> "$GITHUB_OUTPUT"
 
+      - name: Resolve ECR image tag
```
devashish2203: Remove from OSS.
Also getting the PR numbers can be simply `github.event.pull_request.number`

.github/workflows/playwright-e2e-tests.yml
```diff
@@ -121,12 +139,37 @@ jobs:
         with:
           node-version: "20"
 
-      - name: Run quickstart with head images
+      - name: Configure ECR or fall back to OSS images
```
devashish2203: Not required in OSS.

.github/workflows/playwright-e2e-tests.yml
```diff
@@ -121,12 +139,37 @@ jobs:
...
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ECR_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_ECR_SECRET_ACCESS_KEY }}
+          IMAGE_TAG: ${{ needs.setup.outputs.image_tag }}
+        run: |
+          if [[ -n "${AWS_ACCESS_KEY_ID:-}" ]]; then
+            aws configure set aws_access_key_id "$AWS_ACCESS_KEY_ID"
```
devashish2203: Use aws-actions/configure-aws-credentials as done in other places instead of manually calling aws configure.

smoke-test/run-quickstart.sh
```diff
@@ -24,6 +24,16 @@ XPACK_SECURITY_ENABLED="${XPACK_SECURITY_ENABLED:=plugins.security.disabled=true
 ELASTICSEARCH_USE_SSL="${ELASTICSEARCH_USE_SSL:=false}"
 USE_AWS_ELASTICSEARCH="${USE_AWS_ELASTICSEARCH:=true}"
 
+EXECUTOR_SCALE_ARGS=()
```
devashish2203: Don't need this in OSS

smoke-test/run-quickstart.sh
```diff
@@ -43,5 +53,5 @@ DATAHUB_ACTIONS_IMAGE=acryldata/datahub-actions \
 DATAHUB_TOKEN_SERVICE_SIGNING_KEY=${DATAHUB_TOKEN_SERVICE_SIGNING_KEY} \
 DATAHUB_TOKEN_SERVICE_SALT=${DATAHUB_TOKEN_SERVICE_SALT} \
 DATAHUB_LOCAL_ACTIONS_ENV=`pwd`/test_resources/actions/actions.env  \
-docker compose --project-directory ../docker/profiles --profile ${PROFILE_NAME:-quickstart-consumers} up -d --quiet-pull --wait --wait-timeout 900
+docker compose --project-directory ../docker/profiles --profile ${PROFILE_NAME:-quickstart-consumers} up -d --quiet-pull --wait --wait-timeout 900 "${EXECUTOR_SCALE_ARGS[@]}"
```
devashish2203: Don't need in OSS.

smoke-test/run-quickstart.sh
devashish2203: Also when implementing in Fork see if we can avoid this, if Cypress tests don't need this why does Playwright need this?

.github/workflows/playwright-e2e-tests.yml
```diff
@@ -121,12 +139,37 @@ jobs:
...
+            } >> "$GITHUB_ENV"
+          fi
+
+      - name: Run quickstart
         env:
           DATAHUB_TELEMETRY_ENABLED: false
-          DATAHUB_VERSION: head
           ACTIONS_EXTRA_PACKAGES: "acryl-datahub-actions[executor] acryl-datahub-actions"
           ACTIONS_CONFIG: "https://raw.githubusercontent.com/acryldata/datahub-actions/main/docker/config/executor.yaml"
+          UV_EXTRA_INDEX_URL: ${{ secrets.UV_EXTRA_INDEX_URL }}
```
devashish2203: Not required for OSS. 
If you've created a secret remove that.

.github/workflows/docker-unified.yml
```diff
@@ -842,6 +887,211 @@ jobs:
...
+    name: Playwright E2E Tests (Shard ${{ matrix.shard }}/${{ matrix.shard_count }})
```
devashish2203, line 910: Set a reasonble timeout for the overall job.

devashish2203 (changes requested):
Potential Issues / Risks
- Shard count string vs. integer — already flagged above; the for-loop will likely work in Bash but seq or explicit cast is safer.
playwright_matrix empty-string check — the if: condition checks needs.setup.outputs.playwright_matrix != '' and != '{"include":[]}'. If the matrix output is malformed (e.g., missing quotes in JSON), the fromJson call will silently fail. Consider adding a validation step.
- Playwright cache save race — only shard 1 saves the Yarn cache. If shard 1 fails, the cache is never updated even if other shards succeeded.
- playwright_report does npm install -g @playwright/test without pinning a version. This can result in a different Playwright version than the one used in the test shards (which use yarn install --frozen-lockfile). Pin the version to match package.json.
- Blob report flattening logic — the find ... -exec cp {} playwright-blobs-flat/ approach silently overwrites reports if two shards produce a blob with the same filename. Playwright blob filenames are unique per shard, but this is worth a comment.
- PLAYWRIGHT_REUSE_SERVER=1 is set in the test job but there's no explanation of what it does or whether it's safe in a CI shard context. It should be verified it doesn't share state across shards.

.github/actions/ci-optimization/action.yml
priyabratadas-dh: Done

.github/workflows/docker-unified.yml
priyabratadas-dh: Done

.github/actions/ci-optimization/action.yml
devashish2203: You can call it playwright-test:

.github/workflows/docker-unified.yml
```diff
@@ -197,6 +206,33 @@ jobs:
...
 
+      - name: Compute Playwright Yarn Cache Key
+        id: playwright-yarn-cache-key
+        run: |
+          echo "playwright_yarn_cache_key=docker-unified-playwright-${{ runner.os }}-yarn-${{ hashFiles('./e2e-test/ui/playwright/package.json') }}" >> "$GITHUB_OUTPUT"
+          echo "playwright_yarn_cache_key_prefix=docker-unified-playwright-${{ runner.os }}-yarn-" >> "$GITHUB_OUTPUT"
+
+      - name: Build Playwright shard matrix
+        id: set-playwright-matrix
+        if: ${{ steps.ci-optimize.outputs.backend-change == 'true' || steps.ci-optimize.outputs.frontend-change == 'true' || steps.ci-optimize.outputs.playwright-change == 'true' || github.event_name != 'pull_request' }}
```
devashish2203: Isn't this incorrect? The github.event_name != 'pull_request' should be an && condition to ensure it never runs for pull requests, or else if any of the other conditions are true then it'll run on pull requests too.

.github/workflows/docker-unified.yml
```diff
@@ -842,6 +878,213 @@ jobs:
...
+          docker images
+
+      - name: Run quickstart
+        env:
+          DATAHUB_TELEMETRY_ENABLED: false
+          DATAHUB_VERSION: ${{ needs.setup.outputs.tag }}
+          SKIP_INSTALL_DEV: "true"
+          ACTIONS_EXTRA_PACKAGES: "acryl-datahub-actions[executor] acryl-datahub-actions"
+          ACTIONS_CONFIG: "https://raw.githubusercontent.com/acryldata/datahub-actions/main/docker/config/executor.yaml"
+          UV_EXTRA_INDEX_URL: ${{ secrets.UV_EXTRA_INDEX_URL }}
```
devashish2203: Has the secret been set? Why not use the pattern that is used in other jobs and construct the UV_EXTRA_INDEX_URL within the job?

.github/workflows/playwright-e2e-tests.yml
```diff
@@ -216,7 +221,7 @@ jobs:
           override_branch: ${{ github.head_ref || github.ref_name }}
 
       - uses: actions/cache/save@0057852bfaa89a56745cba8c7296529d2fc39830 # v4
-        if: ${{ github.ref == 'refs/heads/master' && matrix.shard == '1' }}
+        if: ${{ github.ref == 'refs/heads/master' }}
```
devashish2203: Saving cache only on the first shard is fine. 
Doing it in parallel might cause problems.

.github/workflows/playwright-e2e-tests.yml
devashish2203: Should this be node 22? That is the version we use for dev.

.github/workflows/playwright-e2e-tests.yml
```diff
@@ -241,14 +246,16 @@ jobs:
           path: playwright-blob-reports
 
       - name: Install Playwright
-        run: npm install -g @playwright/test
+        # Pin to the version resolved by yarn.lock in the test shards so merge-reports is compatible.
+        run: npm install -g @playwright/test@1.59.1
```
devashish2203: If we use the same yarn cache as above job and install playwright from yarn install, then we don't have to maintain the version in two different places and keep them in sync.

Using the cache will ensure we don't download the packages and thus the install is also fast.

devashish2203 (changes requested):

devashish2203 (approved):

.github/workflows/docker-unified.yml
devashish2203: OSS doesn't need this at all. Remove before merging.

PR #17366, opened 2026-05-08
docs(ingestion): Add capability for operations

AdrianMachado:
## Summary

`Operation` aspects can be used to power `FreshnessAssertions` but we do not have a canonical list of sources that actually emit `Operation` as a part of ingestion. In this PR, I scanned through the sources that either 

1. Emit `OperationClass` directly: ABS, S3, Delta Lake, Glue, Salesforce, BigQuery legacy usage, Snowflake legacy usage, Redshift, Databricks Unity.
2. Use the SQL Parser and have a `generate_operations` config: BigQuery queries v2, Snowflake queries v2, Dremio, ClickHouse, Oracle, Teradata, SQL Queries.

Often times if `include_query_lineage` is true, then operations emission seems to be turned on as a by-product (I suppose we use it for lineage)

## Preview

See latest vercel deployment and go to /docs/api/tutorials/operations

image


sgomezvillamor (approved):
🎖️

PR #17369, opened 2026-05-09
feat(ingest/sigma): emit chart entity-level warehouse upstream

kyungsoo-datahub:
Workbook charts that pull directly from warehouse tables now emit entity-level ChartInfo.inputs upstreams. When extract_lineage: true (default), Sigma's BFS node type=table upstreams are resolved to warehouse Dataset URNs and appended to chartInfo.inputs. This enables the DataHub UI to display warehouse table lineage for charts that previously showed empty inputs.


sgomezvillamor (approved):
Good implementation — the dual-index design, deduplication guards, and URN identity test are all solid.

**Suggestion**: The new sections in `sigma_post.md` contain implementation details aimed at developers/maintainers (BFS algorithm, dual lookup index, urlId vs name-based matching, `inode-{urlId}` format references). These are already well-covered by the code comments in `sigma.py` but add noise to a user-facing config reference — consider trimming them from the docs.

**Suggestion**: `chart_warehouse_table_name_ambiguous` is incremented and logged at `DEBUG` when a name-fallback finds multiple candidate URNs and the edge is silently skipped. This case appears user-actionable — the fix is typically `default_database`/`default_schema` in `connection_to_platform_map` — but since there is no `reporter.warning()`, users won't see it in the ingestion report. Adding one with context (workbook, chart, table name, candidate URNs) would help operators self-serve.

**Minor**: In `_build_workbook_warehouse_table_index`, the `url_id_to_table_name` dict is redundant — the same table name is already available as `transient_map[url_id].table` in the second loop.

---

PR #17370, opened 2026-05-09
feat(ingest/sigma): bridge Sigma column names and emit DM customSQL CLL

kyungsoo-datahub:
- Bridge Sigma display names to warehouse-native column names; emit FineGrainedLineage for DM customSQL via SqlParsingAggregator; dedup DM columns by (elementId, columnId)


sgomezvillamor (approved):
LGTM

---

sgomezvillamor:
Good work overall — clean two-path approach, solid counter coverage, and the test suite is thorough.

**Concern: breaking change classification**
Both entries in `updating-datahub.md` read more like bug fixes with visible side effects than intentional API breaks: (1) `fieldPath` was previously emitting wrong values (display names instead of warehouse-native names); (2) the `(elementId, columnId)` dedup fix was silently dropping valid schema columns. Documenting operator-visible side effects is correct, but framing them as bug fixes with side effects would be more accurate. Or even removing breaking change note.

**Suggestion: `_bridge_warehouse_column_name` — add `reporter.warning()` on unresolved**
When `chart_input_fields_warehouse_column_bridge_unresolved` increments, users produce `InputFields` entries whose `fieldPath` points to a non-existent warehouse column (display name instead of native name), silently breaking CLL. Given the PR itself flags this as a meaningful behavior change, a `reporter.warning()` with the affected element/field context would let users detect and act on it at runtime.

**Minor: counter-only paths**
`_extract_customsql_sql_col_name` rejects columnIds and `_bridge_warehouse_column_name` falls back to display names without any accompanying `logger.debug`. Adding traces with the relevant `column_id`/`display_name` in scope would make troubleshooting significantly easier.

---

PR #17371, opened 2026-05-09
fix(ingest/sigma): chart and cross-DM resolver edge cases + multi-ref formula emission

kyungsoo-datahub:
- Restore DM resolution when self-exclusion empties candidates; try cross-DM FGL before  warehouse passthrough on self-named elements; try cross-DM FGL before orphan-drop; emit one InputField per unique resolved formula ref (with dedup for identical refs)

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/sigma/config.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/sigma/config.py
@@ -198,6 +198,11 @@ class SigmaSourceReport(StaleEntityRemovalSourceReport):
     chart_input_fields_skipped_parameter: int = 0
     # Column whose formula refs are exclusively bare sibling refs (e.g. [col]).
     chart_input_fields_skipped_sibling: int = 0
+    # Extra InputFields emitted for columns whose formula resolves to more than
+    # one distinct (upstream_urn, upstream_field) pair. The first resolved pair
+    # is counted in chart_input_fields_resolved; each additional pair increments
+    # this counter. Non-zero means some chart columns have multi-upstream lineage.
+    chart_input_fields_multi_ref_extra: int = 0
     # Sub-bucket of self_ref_fallback: source name that is a case-only mismatch
     # against a workbook element name (warehouse fallback intentionally skipped).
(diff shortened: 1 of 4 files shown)


sgomezvillamor (approved):
Good work overall — the multi-ref formula emission and cross-DM orphan rescue are well-structured, and the new FGL unit tests are solid.

**⚠️ Warnings**

1. **`type: ignore[index]` in `_try_emit_self_named_cross_dm_fgl`** — `FineGrainedLineageClass.upstreams` is `Optional[List[str]]`; the existing pattern in this file is `assert cross_dm_fgl.upstreams` before indexing. Use that instead of suppressing the mypy error.

2. **No exception handling around `_resolve_cross_dm_fgl` in `_try_emit_self_named_cross_dm_fgl`** — an uncaught exception propagates through `_resolve_intra_dm_fgl` and crashes the entire workbook ingestion. Every other cross-DM call site wraps in `try/except` + `self.report.warning(...)`.

**💡 Suggestions**

3. **`dropped_orphan_upstream` counter** — when rescue fails, both `cross_dm_deferred` and `dropped_orphan_upstream` increment for the same ref. The method docstring documents this, but `config.py` doesn't — the counter comment there is misleading without that context.

4. **Early-return observability** — when `not any("/" in sid ...)` fires and rescue is skipped, there is no `logger.debug` or counter. A failed rescue from an orphan branch is invisible in the report.

5. **`resolved_list: List[Tuple[Tuple[str, str], BracketRef]]`** — nested tuple violates the project's dataclass-over-tuple standard. A small `@dataclass` with `upstream_urn`, `upstream_field`, `ref` fields would be cleaner.

6. **O(n²) dedup** — `{p for p, _ in resolved_list}` is rebuilt on every iteration. A `seen: Set[Tuple[str, str]]` maintained alongside resolves this.

**📎 Minor**

7. The `inode-` prefix exclusion in the source-id guard has no test coverage.

8. No test for partial multi-ref resolution (some refs succeed, some return `None`).

9. `test_intra_dm_only_source_ids_not_treated_as_cross_dm` claims "Case A and Case B" but only exercises Case A.

---

PR #17373, opened 2026-05-11
fix(ingest/snowflake): add sweden-central azure region mapping

nicolaiao2:
The Snowsight URL builder constructs broken external URLs for Snowflake
accounts hosted in Azure Sweden Central. The `SNOWFLAKE_REGION_CLOUD_REGION_MAPPING`
dict in `constants.py` is missing an entry for `azure_swedencentral`, so the
fallback in `SnowsightUrlBuilder.get_cloud_region_from_snowflake_region_id`
produces `swedencentral` instead of the hyphenated `sweden-central` that
Snowsight expects in the URL path.

This results in URLs like:
  https://app.snowflake.com/swedencentral.azure/<locator>/    (broken)

instead of:
  https://app.snowflake.com/sweden-central.azure/<locator>/   (correct)

Adds the missing mapping entry, matching the pattern in #8376 for the other Azure regions.


treff7es (approved):

treff7es:
Thanks for the fix

PR #17375, opened 2026-05-11
refactor(ingest/bigquery): stop writing service account secret to env var and disk

rajatoss:
## Summary

`BigQueryConnectionConfig` used to write the service account JSON to a `/tmp` tempfile and set `GOOGLE_APPLICATION_CREDENTIALS` so the SQLAlchemy BigQuery dialect could resolve credentials via Application Default Credentials. Both side effects exposed the secret — the env var was visible via `/proc/<pid>/environ` and inherited by every subprocess; the tempfile sat on disk for the life of the process and was cleaned up only on graceful exit. This was the same class of issue that motivated the Looker fix in #17212.

This PR removes both side effects. The SQLAlchemy path now reuses the in-memory `bigquery.Client` (already built from explicit `service_account.Credentials.from_service_account_info(...)`) via the BigQuery dialect's `?user_supplied_client=true` URL flag plus `connect_args`. The same change is applied to Fivetran's BigQuery log reader, which used the same env-var pattern. Direct google-cloud clients (`get_bigquery_client`, `get_projects_client`, `get_policy_tag_manager_client`, `make_gcp_logging_client`) already used explicit credentials and are unchanged.

### What changed

- `bigquery_v2/bigquery_connection.py` — removed `os.environ[...] = ...` write and the `create_credential_temp_file()` call. `_credentials_path` is gone. The in-memory `_credentials` build is preserved (this is what protects Observe/assertions from cross-tenant credential confusion when multiple service accounts run in the same process).
- `bigquery_v2/profiler.py` — overrides `get_profiler_instance()` so the SQLAlchemy engine for BigQuery is constructed with `?user_supplied_client=true` and `connect_args={"client": self.config.get_bigquery_client()}` when a credential block is provided. When no credential block is provided, the engine is built with empty `connect_args`, preserving ADC fallback for users running with Workload Identity, gcloud auth, or their own `GOOGLE_APPLICATION_CREDENTIALS`.
- `bigquery_v2/bigquery.py` — deleted `cleanup()` and `atexit.register(cleanup, config)`. No tempfile to clean up.
- `common/gcp_credentials_config.py` — deleted `create_credential_temp_file()`. No remaining callers.
- `fivetran/fivetran_log_db_reader.py` — applies the same `user_supplied_client=true` + `connect_args` pattern as the profiler for the BigQuery destination path.
- `setup.py` — bumped `sqlalchemy-bigquery>=1.4.1` to `>=1.5.0`. The `user_supplied_client` flag landed in v1.5.0 (released 2022-11-29). `pyproject.toml`, `uv.lock`, and `constraints.txt` were regenerated via `./gradlew :metadata-ingestion:updateLockFile`.

### Side effects (intentional)

- `BigQueryQueriesSource` and Fivetran's `BigQueryDestinationConfig` previously called `create_credential_temp_file()` without ever registering `atexit` cleanup, leaking tempfiles indefinitely. Both paths are fixed as a consequence of removing the tempfile creation.
- The race condition described in the prior `BigQueryConnectionConfig.__init__` comment (concurrent service accounts overwriting the shared env var) is now structurally impossible — there is no shared env var anymore.

### What did NOT change

- The default URL returned by `get_sql_alchemy_url()` is unchanged (`"bigquery://"` or `"bigquery://<project_on_behalf>"`); the `user_supplied_client=true` flag is appended at each call site rather than baked into the URL helper, so callers that pass no `connect_args` continue to work.
- All direct google-cloud client calls (the majority of BigQuery's ingestion paths) were already secure and are untouched.
- Users who rely on ADC (no `credential:` block in their recipe) are unaffected.

## Test plan

- [x] `./gradlew :metadata-ingestion:lint` — ruff + mypy both pass
- [x] `tests/unit/bigquery/test_bigquery_source.py` — 111 tests pass, including `test_bigquery_explicit_credentials_built` confirming `_credentials` is populated
- [x] `tests/unit/bigquery/test_bigquery_profiler.py` — 8 tests pass, including two new ones:
  - `test_profiler_engine_uses_user_supplied_client_when_credential_set` — asserts URL contains `user_supplied_client=true` and `connect_args["client"]` is the prebuilt `bigquery.Client`
  - `test_profiler_engine_falls_back_to_adc_when_no_credential` — asserts the flag is NOT added when no credential block is provided, preserving ADC for Workload Identity users
- [x] `tests/unit/fivetran/` — 174 tests pass
- [x] End-to-end probe against installed `sqlalchemy-bigquery==1.16.0` confirmed:
  - After `BigQueryConnectionConfig.__init__`: `GOOGLE_APPLICATION_CREDENTIALS` is unchanged, no new tempfile in `/tmp`
  - `BigqueryProfiler.get_profiler_instance()` builds an engine with the expected URL flag and `connect_args`
  - `from_service_account_info` is called once with the dict (the in-memory path), no `from_service_account_file` and no `google.auth.default()` invocations


metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/profiler.py
```diff
@@ -36,6 +44,57 @@ def __init__(
...
+            )
+            return SQLAlchemyProfiler(
+                conn=inspector.bind,
+                report=self.report,
+                config=self.config.profiling,
+                platform=self.platform,
+                env=self.config.env,
+            )
+        else:
+            logger.info(
```
skrydal, line 90: Add a comment that this will be removed in near future (as we are removing GE)

skrydal (approved):
Good catch

metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/profiler.py
rajatoss: added

PR #17376, opened 2026-05-11
feat(ingest/hex): improved hex connector — lineage, components, column-level lineage, context docs

aviraj-gour:
## Summary

Major in-place upgrade of the existing `hex` connector. Lineage, run history, context documents, and component remapping are now driven directly from new Hex REST APIs discovered during this work — no external CLI, no warehouse-side ingestion dependency, no query-tag scraping. Backward-compatible: all existing recipe fields still work; the only removed fields (`lineage_start_time`, `lineage_end_time`, `datahub_page_size`) emit clear deprecation warnings via `pydantic_removed_field` instead of silently no-op'ing.

## New Hex REST APIs used

- `GET /v1/cells?projectId=` — SQL source, markdown, visualisations, and `COMPONENT_IMPORT` detection.
- `GET /v1/data-connections` — platform / database / schema resolution for lineage URN construction.
- `GET /v1/projects/{id}/queriedTables` — **ENTERPRISE tier only**: Hex's own pre-resolved table list (opt-in via `use_queried_tables_lineage`).
- `GET /v1/projects/{id}/runs` — run history for `lastRefreshed` and ERRORED-run surfacing.
- `POST /v1/projects/export` — the only API that exposes which component IDs a project imports, plus separates native project SQL from inlined component SQL.
- `GET /v1/users/me` — optional, only used to auto-discover the workspace UUID for external URLs.

## Entity model changes

- **Hex Components → Chart entities** (previously Dashboard). Validated against the live API: a component defines its own visualisation and the importing project cannot override it, so the Looker `Look` / PowerBI `Tile` analogy applies. A self-healing migration soft-deletes any pre-existing legacy `Dashboard`-typed Components on the first run.
- **Project → Component link** via `DashboardInfo.charts` (built from the export API's component import list).
- Components get their own `ChartInfo.inputs` dataset lineage and `InputFieldsClass` column lineage — full chain: Project (Dashboard) → Component (Chart) → Dataset → Schema Fields.

## Lineage (tiered, safety-first)

Two tiers, both opt-out via `include_lineage: false`:

- **Tier 1 — `queriedTables`** (ENTERPRISE only, opt-in via `use_queried_tables_lineage: true`): Hex's own runtime-proven table list for **published** projects/components. Unpublished entities always fall back to SQL parsing since `queriedTables` is only populated for published runs. A `403` falls back to SQL parsing for everything.
- **Tier 2 — SQL parsing via `sqlglot`** (all workspaces, default): each cell parsed with its connection's dialect. Unqualified `FROM table` refs are resolved using `default_database` / `default_schema` from `/v1/data-connections`, overridable per-connection via `connection_platform_map`. Projects with imported components have native SQL extracted via the export API so inlined component SQL doesn't double-count on the project.

Safety-first behaviour: cells whose `dataConnectionId` cannot be resolved to a known platform are **skipped with a structured warning** rather than emitting URNs against a guessed platform.

## test_connection with tier detection

`test_connection` now probes the live workspace and emits a per-capability report:

- Basic connectivity (handles workspace-token `/users/me` 500s by falling back to `/projects`).
- Lineage via `queriedTables` (ENTERPRISE) — probes `/v1/projects/{id}/queriedTables`; failure mitigation explains the tier requirement.
- Lineage via SQL parsing — probes `/v1/cells` directly (a token can list projects yet 403 on cells), and verifies `/v1/data-connections` access. On failure, the report enriches the error with sample unresolvable connection IDs and representative table names so the user knows exactly what to add to `connection_platform_map`.

## Stale entity removal

Standard `stateful_ingestion.enabled: true` — projects deleted in Hex are soft-deleted in DataHub on the next run. 

## AI context documents

Each project and component gets a `urn:li:document:hex-{id}` entity with structured markdown (SQL queries, visualisations, markdown docs, section structure). 

## Other notable changes

- Code split: `config.py`, `lineage_builder.py`, `document_builder.py` extracted out of the monolithic source.
- Decorator-based API call tracking — every Hex endpoint hit is counted on the source report, separately for outer invocations and inner paginated pages.
- Proactive client-side rate limiter to stay under Hex's per-token QPS without relying on 429-then-retry.
- Removed the deprecated `query_fetcher.py` path that searched DataHub for Hex-tagged Query entities.


sgomezvillamor:
Great work on this PR — the upgrade is substantial and the cross-platform lineage architecture (tiered queriedTables + SQL parsing, safety-first skipping, connection resolution) is well designed. A few things to address before merge:

**🟡 Missing user-facing warning for `use_queried_tables_lineage: true` on non-ENTERPRISE workspaces**
`api.py:fetch_queried_tables` swallows the 403 silently and flips `_queried_tables_tier_available = False` with no `report.warning()`. A user who sets `use_queried_tables_lineage: true` on a non-ENTERPRISE workspace will see lineage in the output and have no idea it came from SQL parsing, not queriedTables. Compare with the `fetch_connections` exception path which does warn. Please add a warning on the 403 branch explaining the tier requirement and the fallback.

**🟡 ❓ Breaking change discussion: Components → Chart URN remapping**
Remapping Components from `urn:li:dashboard:(hex,<id>)` to `urn:li:chart:(hex,<id>)` is semantically correct (the Looker Look analogy holds), but it's a breaking change for existing users: any tags, glossary terms, owners, or descriptions enriched on the old Dashboard URN will be silently lost — the soft-delete migration does not copy metadata to the new Chart entity. A few questions worth discussing:
- Should there be an opt-out flag (e.g. `use_legacy_component_entity_type: bool`) to let teams migrate at their own pace?
- Should the migration attempt to copy existing metadata to the new URN (requires `ctx.graph`)?
- This should be documented in `docs/how/updating-datahub.md` as a breaking change per project conventions.

**🟡 `include_context_documents` defaults to `True`**
This creates a new entity type (`urn:li:document:hex-{id}`) per project and component, which could surprise existing users on upgrade. Worth checking: is this experimental? Is it expensive for large workspaces? If either answer is yes, `default=False` (opt-in) is safer.

**🟡 Removed `query_fetcher.py` — user-facing docs gap**
The removal of the DataHub-query-fetcher lineage path is mentioned in the PR body but not in `hex_post.md`. Users relying on the old mechanism should be told it's gone and what replaces it. A short note in the docs (even just in the Limitations/Migration section) would help.

**🟡 `hex_post.md` — tiered lineage explanation missing**
The PR description has a clear Tier 1 / Tier 2 breakdown with the "unpublished entities always fall back" and "safety-first skipping" callouts. This is exactly what should be in user-facing docs. Please port it to `hex_post.md`.

**ℹ️ `_track_page` attribute injection is fragile**
The `@_api_call(paginated=True)` decorator injects `self._track_page` as a callable before the method body runs. This requires `# type: ignore` and creates invisible coupling. A small helper method or direct counter increment would be simpler and more maintainable.

**ℹ️ Missing ingestion-level test for queriedTables-403 fallback**
`TestHexTestConnection` covers the `test_connection()` path, but there's no test verifying that during actual ingestion, a 403 on queriedTables falls back to SQL-cell lineage (and surfaces the right warning once the fix above is in).

---

aviraj-gour:
Thanks @sgomezvillamor for review, only unresolved concern for now is `❓ Breaking change discussion: Components → Chart URN remapping`.


> 🟡 Missing user-facing warning for use_queried_tables_lineage: true on non-ENTERPRISE workspaces
> api.py:fetch_queried_tables swallows the 403 silently and flips _queried_tables_tier_available = False with no report.warning(). A user who sets use_queried_tables_lineage: true on a non-ENTERPRISE workspace will see lineage in the output and have no idea it came from SQL parsing, not queriedTables. Compare with the fetch_connections exception path which does warn. Please add a warning on the 403 branch explaining the tier requirement and the fallback.

Added a warning before flipping a flag.

>  Removed query_fetcher.py — user-facing docs gap
> The removal of the DataHub-query-fetcher lineage path is mentioned in the PR body but not in hex_post.md. Users relying on the old mechanism should be told it's gone and what replaces it. A short note in the docs (even just in the Limitations/Migration section) would help.

And

> 🟡 hex_post.md — tiered lineage explanation missing
> The PR description has a clear Tier 1 / Tier 2 breakdown with the "unpublished entities always fall back" and "safety-first skipping" callouts. This is exactly what should be in user-facing docs. Please port it to hex_post.md.

For these two updates in the hex_post.md document, I will also update the updating-datahub.md file within the same PR.


```
include_context_documents defaults to True
This creates a new entity type (urn:li:document:hex-{id}) per project and component, which could surprise existing users on upgrade. Worth checking: is this experimental? Is it expensive for large workspaces? If either answer is yes, default=False (opt-in) is safer.
```

Well this is not a expensive option but defaulted to False.

> ℹ️ _track_page attribute injection is fragile
> The @_api_call(paginated=True) decorator injects self._track_page as a callable before the method body runs. This requires # type: ignore and creates invisible coupling. A small helper method or direct counter increment would be simpler and more maintainable.

Turned _track_page into a real method on HexApi that takes the endpoint name explicitly (self._track_page("list_cells") at the call site), and dropped the paginated=True branch + closure injection + the __init__ placeholder lambda + the # type: ignore[attr-defined]. Behavior is unchanged (same counter keys, same increment points — verified by a small test run); the coupling is now visible.


> ℹ️ Missing ingestion-level test for queriedTables-403 fallback
> TestHexTestConnection covers the test_connection() path, but there's no test verifying that during actual ingestion, a 403 on queriedTables falls back to SQL-cell lineage (and surfaces the right warning once the fix above is in).

Added a test for this.

sgomezvillamor (approved):
LGTM

PR #17380, opened 2026-05-11
feat(ingest/unity): add opt-in Databricks Unity Catalog Metric View i…

dinesh-verma-datahub:
## Summary

Adds first-class support for [Databricks Unity Catalog Metric Views](https://docs.databricks.com/aws/en/metric-views/) to the Unity Catalog source. Behaviour is gated behind a new opt-in flag and the existing flag-off path is unchanged, so the upgrade is safe for current users.

## What changes for users

A new config flag is introduced:

```yaml
source:
  type: unity-catalog
  config:
    include_metric_views: true
    metric_view_pattern:
      allow:
        - my_catalog\.analytics\..*
```

When `include_metric_views: true` is set, each Unity Catalog Metric View is ingested with:

- Dataset subtype `Metric View` (distinct from `Table` and `View`).
- View properties carrying the raw YAML body with `viewLanguage: YAML`. The `materialized` flag is set to `true` when the YAML declares `materialization: materialized`.
- Upstream lineage derived from the YAML `source` and `joins[].source` fields. Both 3-part (`catalog.schema.table`) and 2-part (`schema.table`, resolved against the metric view's own catalog) identifiers are supported, including backtick-quoted parts (e.g. `` `db.with.dots`.schema.table ``).
- Column-level lineage parsed from each `dimensions[].expr` / `measures[].expr` using the Databricks SQL dialect. Unqualified columns map to the top `source`; `<join_name>.column` references map through the matching join. Honours the existing `include_column_lineage` flag (on by default).
- A `Dimension` or `Measure` tag on every schema field whose name matches a YAML dimension/measure.
- A `metric_view_filter` custom property surfacing the YAML's top-level `filter`.
- Per-column descriptions taken from `dimensions[].description` / `measures[].description` when present; otherwise the existing Unity Catalog column comment is kept.
- A filtered set of custom properties: the ~150 `view.sqlConfig.spark.*` entries that Unity Catalog injects per metric view (a snapshot of the workspace's Spark engine config, identical across views) are dropped. All other `view.*` properties and the `metric_view.*` properties pass through unchanged.

A new `metric_view_pattern` config (a standard `AllowDenyPattern` matched against `catalog.schema.metric_view_name`) lets operators include or exclude specific metric views; it only applies when `include_metric_views` is enabled.

Metric views also count toward the existing `tables` filter report so stateful soft-delete keeps tracking them regardless of the flag state.

## Fallback and degradation behaviour

- If a metric view's `source` is a SQL subquery, a 1-part name, or otherwise unresolvable, the YAML lineage path is skipped and the existing Unity Catalog table-lineage REST API is consulted instead. An info entry is emitted when neither path produces any upstreams.
- Malformed YAML, unexpected YAML shapes, missing view bodies, unparseable column expressions, unresolved qualifier prefixes, and skipped joins are reported through dedicated counters and warnings in the source report rather than being silently dropped.
- If the installed `databricks-sdk` is too old to expose the `METRIC_VIEW` table type, metric views continue to be emitted as plain `Table` entities and a one-time startup warning is logged. No other behaviour changes.

## Backwards compatibility

`include_metric_views` defaults to `false`. With the flag off (or with an older `databricks-sdk`), Unity Catalog metric views are emitted exactly as in the previous release — plain `Table` subtype, original column tags, no YAML view body, no extra custom properties, and `view.sqlConfig.*` properties preserved. This invariant is pinned by a dedicated golden-file integration test.

## Tests

- 98 unit tests covering identifier parsing, subtype/view-language selection, the YAML lineage extractor, column-level lineage with joins and case-insensitive aliases, dimension/measure tag attachment, description override, materialization parsing, `metric_view_filter` emission, the `view.sqlConfig.*` filter, the `metric_view_pattern` allow/deny path, REST-fallback behaviour, the old-SDK shim, and the full set of report counters/warnings (including the no-regression contract for the legacy column-tag behaviour on struct/array columns).
- 2 golden-file integration tests: one for the flag-on path (full metric-view enrichment) and one for the flag-off path (byte-for-byte legacy behaviour).

### Connector Tests

https://github.com/acryldata/connector-tests/pull/803

## Docs

- New "Unity Catalog Metric Views" section under the Databricks connector page describing the flag, supported identifier shapes, the YAML-to-aspect mapping, and the REST fallback.
- Release-note entry under `docs/how/updating-datahub.md` → "Next" → "Other Notable Changes".


docs/how/updating-datahub.md
```diff
@@ -111,6 +111,7 @@ This file documents any backwards-incompatible changes in DataHub and assists pe
...
... the existing table-lineage REST API. Column-level lineage is emitted from each `dimensions[].expr` / `measures[].expr` using the Databricks SQL dialect (controlled by the existing `include_column_lineage` flag, which is on by default). Schema fields are tagged `Dimension` or `Measure` when their names match the YAML; per-column descriptions are taken from `dimensions[].description` / `measures[].description` when present and otherwise fall back to the Unity Catalog column comment; the top-level `filter` field surfaces as a `metric_view_filter` custom property; and `materialization: materialized` drives `ViewProperties.materialized = true`. On the opt-in path, the noisy Spark engine config snapshot Unity Catalog injects as `view.sqlConfig.spark.*` custom properties (~150 entries per view) is filtered out — other `view.*` properties and Databricks' own `metric_view.*` properties pass through unchanged. The flag remains `false` by default so the upgrade is byte-for-byte identical to the previous release. Requires a `databricks-sdk` recent enough to expose `TableType.METRIC_VIEW` (older SDKs short-circuit silently and the flag has no effect; a startup warning is emitted in that case).
```
```suggestion
- #17380 (Ingestion / Databricks Unity Catalog) New opt-in flag `include_metric_views` (default `false`) for Unity Catalog Metric Views. When enabled, metric views are emitted with subtype `Metric View` (instead of `Table`), their YAML body is preserved as a `ViewProperties` aspect with `viewLanguage: YAML`, and upstream lineage is derived from the YAML `source` and `joins[].source` fields. Lineage resolution supports 3-part (`catalog.schema.table`) and 2-part (`schema.table`, resolved against the metric view's own catalog) identifiers as well as backtick-quoted parts; 1-part names and SQL subqueries fall back to the existing table-lineage REST API. Column-level lineage is emitted from each `dimensions[].expr` / `measures[].expr` using the Databricks SQL dialect (controlled by the existing `include_column_lineage` flag, which is on by default). Schema fields are tagged `Dimension` or `Measure` when their names match the YAML; per-column descriptions are taken from `dimensions[].description` / `measures[].description` when present and otherwise fall back to the Unity Catalog column comment; the top-level `filter` field surfaces as a `metric_view_filter` custom property; and `materialization: materialized` drives `ViewProperties.materialized = true`. On the opt-in path, the noisy Spark engine config snapshot Unity Catalog injects as `view.sqlConfig.spark.*` custom properties (~150 entries per view) is filtered out — other `view.*` properties and Databricks' own `metric_view.*` properties pass through unchanged. The flag remains `false` by default so the upgrade is byte-for-byte identical to the previous release. Requires a `databricks-sdk` recent enough to expose `TableType.METRIC_VIEW` (older SDKs short-circuit silently and the flag has no effect; a startup warning is emitted in that case).
```

metadata-ingestion/docs/sources/databricks/README.md
```diff
@@ -30,20 +30,21 @@ For a deeper look at how to think about DataHub within and across your Databrick
...
+| Catalog                                   | Container (CATALOG)               | Namespace within a metastore. Hive Metastore is ingested as a special catalog type.                                   |
+| Schema                                    | Container (SCHEMA)                | Nested under its Catalog container.                                                                                   |
+| Table (managed, external, Delta, Iceberg) | Dataset (TABLE)                   | All non-view table types including streaming tables. Schema, descriptions, and tags are extracted.                    |
+| View / Materialized View                  | Dataset (VIEW)                    | View definition is captured.                                                                                          |
+| Unity Catalog Metric View                 | Dataset (METRIC_VIEW)             | Opt-in via `include_metric_views: true`. YAML body is preserved; dimensions and measures are tagged as schema fields. |
```
sgomezvillamor:
```suggestion
| Metric View                               | Dataset (METRIC_VIEW)             | Opt-in via `include_metric_views: true`. YAML body is preserved; dimensions and measures are tagged as schema fields. |
```

metadata-ingestion/docs/sources/databricks/unity-catalog_post.md
```diff
@@ -2,6 +2,35 @@
 
 Use the **Important Capabilities** table above as the source of truth for supported features and whether additional configuration is required.
 
+#### Unity Catalog Metric Views
```
sgomezvillamor:
```suggestion
#### Metric Views
```

sgomezvillamor (approved):
Nice work overall — approving with two small follow-ups in the comments.

---

sgomezvillamor:
Two small follow-ups:

**Frontend rendering**: The View Definition tab will show (since `viewLogic` is populated), but does the UI respect `viewLanguage="YAML"` for syntax highlighting? If it always applies SQL highlighting, the YAML body will render poorly — may need a frontend change.

**`_split_databricks_identifier`**: This is a non-trivial parser and a key reusable utility. Please add a docstring, consider making it public/moving it to a utility module, and add isolated unit tests. Also, `proxy.py` still uses bare `split(".")` — worth updating for consistency.

---

dinesh-verma-datahub:
@sgomezvillamor thanks for the review, applied the minor changes as per your suggestions.

PR #17385, opened 2026-05-11
fix(graphql): preserve named fragments in QueryProjector to keep request body small

nwadams:
## Summary

`QueryProjector.adapt_query()` previously called `_inline_fragments()` before running the field-removal visitor, replacing every named fragment spread with its inline definition. For queries with many fragments (e.g. `entity_details.gql` in `datahub-agent-context`), this expanded the on-the-wire body roughly 4× — from ~43 KB to ~183 KB — even when projection removed nothing.

DataHub Cloud edges enforce a body-size limit (~100 KB) and rejected these requests with **413 Payload Too Large**. The result was that `get_entities` on the MCP agent tools failed for all entities on cloud instances.

The original goal of the projector (strip fields the server schema doesn't support) does not require inlining: GraphQL's `TypeInfo` correctly tracks the parent type inside `FragmentDefinitionNode`s via the type_condition, so field removal works directly on fragment definitions.

### Changes

- `adapt_query()` no longer calls `_inline_fragments()`; the visitor walks the parsed document with fragments preserved.
- A pre-pass identifies fragment definitions whose `type_condition` references a type missing from the server schema. The visitor removes both those definitions and any spreads referencing them.
- `_inline_cache` renamed to `_parse_cache` (now caches parsed `DocumentNode`s, not inlined ones).
- `_inline_fragments()` helper retained — the existing graphql-java 200K-whitespace-token regression test exercises it directly, and the helper remains available for callers that want the legacy behavior.

### Verification

Against `acryl.acryl.io` (DataHub Cloud) calling `GetEntity` with `entity_details.gql`:

| | Before | After |
|---|---|---|
| Request body | 183,763 bytes | 17,234 bytes |
| Result | 413 Payload Too Large | 200 OK |

Documents that previously could not be resolved by `get_entities` (e.g. `urn:li:document:analytics-agent-correction-...`) now return entity data normally.

## Test plan

- [x] `pytest tests/unit/utilities/test_graphql_query_adapter.py` — 76 passed (74 existing + 2 new regression tests)
- [x] End-to-end: `get_entities()` succeeds against cloud for the document URNs that previously returned 413
- [x] Existing graphql-java whitespace-token regression test (`TestGraphqlWhitespaceTokenOverflow`) still passes — it exercises `_inline_fragments` directly and is unaffected

### New tests

- `TestAdaptQueryPreservesFragments::test_entity_details_minified_size_under_cloud_body_limit` — pins the cloud body-size budget against the real `entity_details.gql`.
- `TestAdaptQueryPreservesFragments::test_adapt_query_does_not_inline_supported_fragments` — pins that fragment definitions and spreads survive projection when their types are supported.

### Updated tests

- `test_real_world_multi_fragment_query` — asserts fragment defs/spreads are preserved (removed `... on Dataset` substring assertions that encoded the old inlining behavior).
- `test_search_gql_against_old_schema_strips_document` — same.
- `test_inline_cache_reused_across_schema_generations` → `test_parse_cache_reused_across_schema_generations` (rename to match `_parse_cache`).


alexsku (approved):
Approving. Clean fix for the customer-visible 413, and the new regression tests pin both the cloud body budget and the fragment-preservation contract.

## What this changes

`QueryProjector.adapt_query()` used to call `_inline_fragments()` unconditionally before running the field-removal visitor, replacing every `...FragmentName` spread with the fragment body and ~4×ing the on-the-wire body for `entity_details.gql` (43KB → ~183KB). This PR drops that pre-pass. Instead, a small pre-pass collects fragment definitions whose `type_condition` references a missing type, and the visitor removes those (and their spreads) directly via the new `enter_fragment_definition` / `enter_fragment_spread` handlers, relying on graphql-core's `TypeInfo` to track parent type inside `FragmentDefinitionNode`s for field-level pruning.

## Risk

- **Risk**: Low–Medium. Every caller of `graph.execute_graphql(..., strip_unsupported_fields=True)` (currently `datahub-agent-context` and `datahub graphql` / `datahub search`) sees the new projection shape.
- **Rollback**: safe — revert restores the prior (broken-on-cloud) state. No data implications.
- **Coverage**: new tests pin (a) `entity_details.gql` stays under the 60KB cloud body budget after minification, (b) supported fragments + spreads survive projection, (c) the legacy `_inline_fragments` regression test still exercises the 200K whitespace-token guard via the kept helper.

## Findings

### Medium — orphan fragment definitions after pruning

`graphql_query_adapter.py:432-436`. The pre-pass only inspects each fragment's top-level `type_condition`. If a fragment whose `type_condition` *is* supported has its spreads only inside an `... on UnsupportedType` block that gets pruned, the fragment definition is left in the query unreferenced. Pre-PR (inlining) this couldn't happen because `_FragmentInliner.enter_fragment_definition` returned `REMOVE` for every def, so the output never carried fragment definitions. Post-PR, graphql-java's `NoUnusedFragments` validation rule could trip on bespoke queries.

For `entity_details.gql` specifically this looks fine — fragments are heavily reused across many entity types, so the "only spread inside one unsupported block" case is unlikely. The e2e verification against `acryl.acryl.io` confirms current real-world callers work. **Comfortable shipping as-is**; flagging as a follow-up since it's a behavioral change vs. the old algorithm that a future query could hit. A second pass after pruning that drops unreferenced fragment defs would close it cleanly.

Related but pre-existing (not introduced by this PR): if a fragment body's only contents are all pruned (e.g. `fragment X on Entity { ... on UnsupportedType { urn } }`), `print_ast` will emit `fragment X on Entity { }` which is technically invalid per spec. The old inlining code had the same hazard one level up (`... on Entity { }`). Worth a hardening pass eventually.

### Low / nits

- `graphql_query_adapter.py:44, 423` — constant `INLINE_CACHE_MAX_SIZE` was not renamed alongside `_inline_cache` → `_parse_cache`. Now reads as "bound the parse cache by the inline-cache max size." Trivial rename.
- `test_graphql_query_adapter.py:1903-1908` — `TestRealSearchGqlFragments` docstring still says "exercise the full pipeline (inline → project → prune)". No longer accurate.

## Questions

1. Did the e2e verification cover any self-hosted GMS that has `NoUnusedFragments` validation enabled, or just `acryl.acryl.io`? Trying to gauge whether the M1 follow-up is urgent.
2. Was keeping `_inline_fragments` exported (vs. inlining its body into the whitespace-token regression test) a deliberate choice for the perf suite, or just to avoid churn?

nwadams:
Thanks @alexsku. Pushed `513832ec17` addressing the findings:

**Medium — orphan fragment definitions** ✅ Added `_prune_orphan_fragments()` post-pass in `graphql_query_adapter.py`. After the field-removal visitor runs, it collects every `FragmentSpread.name` in the rewritten document and drops any `FragmentDefinitionNode` not referenced — repeated to a fixed point so fragment chains (`Outer` spreads `Inner`, only entry point gets pruned) clean up in one shot. Pruned names are appended to `removed_fields` as `FragmentDefinition(<name>, orphan)`. Two new regression tests:
- `test_adapt_query_prunes_fragments_orphaned_by_unsupported_type` — exactly the reviewer's case (`fragment X on Dataset` spread only inside `... on UnsupportedType`).
- `test_adapt_query_prunes_transitive_orphan_fragments` — fragment-of-fragment chain.

**Nit — `INLINE_CACHE_MAX_SIZE`** ✅ Renamed to `PARSE_CACHE_MAX_SIZE` (both the constant and the call site).

**Nit — stale `TestRealSearchGqlFragments` docstring** ✅ Updated to "parse → project → prune orphan fragments".

**Q1 — self-hosted GMS with `NoUnusedFragments`**: Honest answer, no — e2e was just `acryl.acryl.io`. That gap is exactly why the orphan-prune pass is worth doing now rather than as follow-up. With the pass in place, any spread-only-inside-removed-type case ships valid output regardless of whether the server enforces the rule.

**Q2 — keeping `_inline_fragments` exported**: Deliberate — `TestGraphqlWhitespaceTokenOverflow` exercises it directly to pin the graphql-java 200K whitespace-token regression. If that helper inlines, the test still catches the historical bug. Inlining its body into the test would lose that signal for any future caller that re-enters the inlining path.

Pre-existing hazard re: empty `fragment X on Entity { }` after pruning all selections — left out of scope for this PR (it predates and is unaffected by this change). Filing a follow-up.

Tests: 78 pass, lint clean, e2e against `acryl.acryl.io` still resolves the original document URNs.

PR #17392, opened 2026-05-12
feat(graphql): always minify GraphQL request bodies on the wire

nwadams:
## Summary

Stacks on top of #17385 (preserve named fragments in `QueryProjector`). That PR fixed the 413 by stopping unnecessary fragment inlining; this PR completes the wire-size story by ensuring **every** outbound GraphQL query is minified — including the `strip_unsupported_fields=False` path and the projection fallback path that were still shipping pretty-printed bodies.

### New: `minify_graphql_query()` in `graphql_query_adapter.py`

A token-aware minifier that walks `print_ast(parse(query))` and collapses runs of whitespace **outside** string literals. Single-quoted (`"..."`) and block (`"""..."""`) strings are passed through verbatim, so queries like `search(input: {query: "hello   world"})` retain their literal content.

The previous trick (`" ".join(print_ast(ast).split())`, used internally by `QueryProjector.adapt_query`) collapsed all whitespace including inside string literals. That was safe for `entity_details.gql` (no literal contained interior whitespace) but a footgun for any caller whose query included literal spaces. The new minifier closes that hole.

### `DataHubGraph.execute_graphql` minifies before sending

Until now only the `strip_unsupported_fields=True` path minified, and only as a side effect of the projector. Callers passing `strip_unsupported_fields=False` shipped the raw pretty-printed query — 43 KB for `entity_details.gql` even though its minified form is 17 KB. After this PR, the projector still minifies (skipping a redundant re-parse via an `already_minified` flag), and the fallback / strip-False paths route the query through `minify_graphql_query()` before building the body. Unparseable input is passed through unchanged so server-side errors still surface to the caller.

## Measured wire-body savings (request body bytes)

| Query | Before | After |
|---|---:|---:|
| `entity_details.gql` (`strip=False`) | 44,729 | 17,234 |
| `entity_details.gql` (`strip=True`) | 17,234 | 17,234 |
| `search.gql` | 874 | 639 |

Verified end-to-end against `acryl.acryl.io`: both `strip=True` and `strip=False` now send identical minified 17 KB bodies for `entity_details.gql` and return the same entity data.

## Test plan

- [x] `pytest tests/unit/utilities/test_graphql_query_adapter.py` — 86 passed (84 from #17385 plus 8 new across `TestMinifyGraphqlQuery` and `TestExecuteGraphqlMinifiesWireBody`)
- [x] `./gradlew :metadata-ingestion:lint` — clean
- [x] Live cloud (`acryl.acryl.io`): both `strip=True` and `strip=False` succeed with identical 17,234-byte request bodies

### New tests
- `TestMinifyGraphqlQuery::test_collapses_indentation_outside_strings`
- `TestMinifyGraphqlQuery::test_preserves_whitespace_inside_string_literals` — the regression the old `.split()/.join()` trick was vulnerable to
- `TestMinifyGraphqlQuery::test_preserves_block_string_contents` — `"""..."""` literals are passed through
- `TestMinifyGraphqlQuery::test_idempotent`
- `TestMinifyGraphqlQuery::test_invalid_query_returned_verbatim` — graceful degradation on unparseable input
- `TestMinifyGraphqlQuery::test_semantic_equivalence` — minified AST == original AST
- `TestExecuteGraphqlMinifiesWireBody::test_strip_false_minifies_before_send` — pins the new wire behavior
- `TestExecuteGraphqlMinifiesWireBody::test_invalid_query_passes_through_when_strip_false` — pins the graceful-degradation guarantee

### Updated test
- `TestExecuteGraphqlFailSafe::test_projection_error_falls_back_to_original_query` compares on parsed-AST equivalence rather than byte-for-byte, since the fallback path now also minifies. Semantic invariant ("no projection applied") is unchanged.

## Depends on

#17385 — without that PR, `QueryProjector.adapt_query` still calls `_inline_fragments` and the `_SpreadCollector` symbol referenced by this diff isn't on master. Will rebase onto master after #17385 merges.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/graph/client.py
+++ b/metadata-ingestion/src/datahub/ingestion/graph/client.py
@@ -1258,6 +1258,10 @@ class DataHubGraph(DatahubRestEmitter, OpenApiAPI, EntityVersioningAPI):
                 root-anchored (``searchAcrossEntities``). See
                 :class:`~datahub.utilities.graphql_query_adapter.RequiredFieldUnsupportedError`.
         """
+        # Whether the query was already minified by adapt_query() above —
+        # avoids a redundant parse+print round-trip on the happy path.
+        already_minified = False
+
         if strip_unsupported_fields:
             try:
                 if self._query_projector is None:
@@ -1267,6 +1271,7 @@ class DataHubGraph(DatahubRestEmitter, OpenApiAPI, EntityVersioningAPI):
                 query, removed = self._query_projector.adapt_query(
                     query, self, required_fields=required_fields
                 )
+                already_minified = True
                 if removed:
                     logger.info(f"Stripped unsupported fields from query: {removed}")
             except Exception as e:
@@ -1285,6 +1290,21 @@ class DataHubGraph(DatahubRestEmitter, OpenApiAPI, EntityVersioningAPI):
                     f"Failed to adapt query for schema compatibility, "
                     f"falling back to original query: {e}"
                 )
+
+        # Always minify before sending — covers strip=False, the projection
+        # fallback path, and any caller that hands us a pretty-printed query.
+        # The projector's output is already minified; skip the redundant
+        # parse+print there. Falls back silently to the original query if
+        # minification raises (e.g. unparseable input from a caller).
+        if not already_minified:
+            try:
+                from datahub.utilities.graphql_query_adapter import (
+                    minify_graphql_query,
+                )
+
+                query = minify_graphql_query(query)
+            except Exception:
+                pass
 
         url = f"{self._gms_server}/api/graphql"
 
--- a/metadata-ingestion/src/datahub/utilities/graphql_query_adapter.py
+++ b/metadata-ingestion/src/datahub/utilities/graphql_query_adapter.py
@@ -101,6 +101,76 @@ def _inline_fragments(document: DocumentNode) -> DocumentNode:
     return document
 
 
+def _minify_printed_graphql(printed: str) -> str:
+    """Collapse insignificant whitespace in already-serialised GraphQL text.
+
+    Treats ``"..."`` and ``\"\"\"...\"\"\"`` string literals as opaque so
+    whitespace inside them is preserved. Caller is responsible for passing
+    valid GraphQL — typically ``print_ast(ast)``.
+    """
+    out: List[str] = []
+    i = 0
+    n = len(printed)
+    while i < n:
+        # Block string: """ ... """ — pass through verbatim.
+        if printed.startswith('"""', i):
+            j = printed.find('"""', i + 3)
+            if j == -1:
+                out.append(printed[i:])
+                break
+            out.append(printed[i : j + 3])
+            i = j + 3
+            continue
+        # Regular string: " ... " with backslash escapes.
+        if printed[i] == '"':
+            j = i + 1
+            while j < n:
+                c = printed[j]
+                if c == "\\":
+                    j += 2
+                    continue
+                if c == '"':
+                    j += 1
+                    break
+                j += 1
+            out.append(printed[i:j])
+            i = j
+            continue
+        # Outside a string: collapse runs of whitespace to a single space.
+        if printed[i].isspace():
+            out.append(" ")
+            while i < n and printed[i].isspace():
+                i += 1
+            continue
+        out.append(printed[i])
+        i += 1
+
+    return "".join(out).strip()
+
+
+def minify_graphql_query(query: str) -> str:
+    """Collapse insignificant whitespace in a GraphQL query string.
+
+    Reduces wire size — ``entity_details.gql`` shrinks from ~43KB to ~17KB —
+    by removing the indentation that ``print_ast`` (and most hand-authored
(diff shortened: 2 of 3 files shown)


alexsku (approved):
## PR Summary

**PR**: #17392 — feat(graphql): always minify GraphQL request bodies on the wire
**Author**: nwadams
**Base**: master ← feat/graphql-minify-all-queries
**Type**: behavior change (wire format) + new utility
**Size**: +223 / -6 across 3 files

Replaces the ad-hoc `" ".join(print_ast(ast).split())` minification (which silently corrupted any literal containing interior whitespace) with a token-aware `_minify_printed_graphql` that treats `"..."` and `"""..."""` as opaque. Wires it into `DataHubGraph.execute_graphql` so **every** outbound request body is minified, not just the `strip_unsupported_fields=True` path. Measured savings: `entity_details.gql` body drops 44,729 → 17,234 bytes on the `strip=False` path.

**Key invariants:**
1. `parse(minify(q)) == parse(q)` — semantic equivalence (parsed AST unchanged).
2. Whitespace inside `"..."` and `"""..."""` literals is preserved verbatim.
3. Unparseable input is returned unchanged (graceful degradation, server still produces meaningful error).
4. No double minification: `adapt_query`'s already-minified output skips the redundant parse+print via `already_minified` flag.

---

## Risk Assessment

| | |
|---|---|
| **Risk** | Low |
| **Blast radius** | All `DataHubGraph.execute_graphql` callers (CLI ingestion, mcp-server-datahub via SDK). Pure wire-format change — no API surface change. |
| **Rollback** | Safe — pure revert, no data/state migration. |
| **Rollout** | Ship it. Behavior is strictly better: smaller bodies, fixes the latent `" ".join(...split())` interior-whitespace bug, fallback path preserved. |
| **CI** | All completed jobs green so far (lint, build, dagster, prefect, perf, testIntegrationBatchRecording). testQuick/testIntegrationBatch* and smoke tests still pending at review time — author has already reported 86/86 unit tests local pass. |

---

## Blocking Issues

None found.

## High-Priority Issues

None found.

## Other Issues

- **[LOW / edge-case]** `metadata-ingestion/src/datahub/utilities/graphql_query_adapter.py:117` — `_minify_printed_graphql` treats `\"""` (escaped triple-quote inside a block string) as a block-string boundary, not as an escape. For a block string whose content contains a literal `"""`, `print_ast` emits `\"""`; the minifier then re-tokenizes the rest of the literal as code and will collapse any interior whitespace that follows the escape. Concrete corruption case: `"""abc\"""   def"""` parses with content `abc"""   def`, but the minifier emits `"""abc\""" def"""` which parses with content `abc""" def`. The docstring at line 156 claims "identical AST before/after," which doesn't hold in this case. Probability is very low — block strings with embedded `\"""` essentially never appear in DataHub queries (search terms are passed as variables, not inlined) — and the failure mode is a server parse error rather than silent data damage. Fine to merge; worth a follow-up to either (a) handle the `\"""` escape in the block-string scanner or (b) tighten the docstring caveat.
- **[LOW / dx]** `metadata-ingestion/src/datahub/ingestion/graph/client.py:1306` — bare `except Exception: pass` around the minify call swallows everything. The realistic failure modes are `ImportError` (graphql-core not installed) and any exception bubbling out of the minifier. Narrowing to `(ImportError, Exception)` or splitting them would make the intent clearer without changing behavior; the existing `# Simulate graphql-core not being installed` test (`test_import_error_falls_back_gracefully`) still works because the inner import here also fails when the module is patched out of `sys.modules`.

---

## What's Missing

Nothing significant. The test surface is well-covered:
- `TestMinifyGraphqlQuery` pins the invariants (idempotent, semantic-equivalence, interior-whitespace preservation, block-string contents, graceful degradation on invalid input).
- `TestExecuteGraphqlMinifiesWireBody` pins the wire behavior on both `strip=False` and unparseable-input paths.
- `TestExecuteGraphqlFailSafe::test_projection_error_falls_back_to_original_query` correctly relaxed to AST equality (the fallback path now also minifies, so byte equality no longer holds).

A regression test for the `\"""` block-string edge case noted above would be a nice addition if you want to address it; not blocking.

## Test Plan

| Invariant | Coverage |
|---|---|
| Semantic equivalence (AST in == AST out) | `test_semantic_equivalence`, `test_strip_false_minifies_before_send` |
| Single-quoted literal interior whitespace preserved | `test_preserves_whitespace_inside_string_literals` |
| Block-string contents preserved | `test_preserves_block_string_contents` (without embedded `\"""`) |
| Idempotency | `test_idempotent` |
| Graceful degradation on unparseable input | `test_invalid_query_returned_verbatim`, `test_invalid_query_passes_through_when_strip_false` |
| `already_minified` skips redundant work | Implicit (no direct test); covered by behavior — adapt_query path output bypasses the second minify because the projector emits already-minified text. |
| Fallback path on projection failure still minifies | `test_projection_error_falls_back_to_original_query` |
| ImportError fallback unchanged | `test_import_error_falls_back_gracefully` |

## Questions for Author

None — PR description and tests answered the obvious ones up front.

---

Approving. Looks good to release; the one edge case noted is genuinely low-probability and the behavior is strictly an improvement over the prior `" ".join(...split())` minifier.

PR #17393, opened 2026-05-12
feat(k8s): add KEDA-aware scaling support to KubernetesController

Ashish37-sh:
**Problem**

The scaleDeployment API directly sets replica counts on deployments. For deployments managed by KEDA (e.g. mae-consumer, mce-consumer, pe-consumer), this has no lasting effect — KEDA continuously monitors Kafka consumer lag and overrides any manual replica count within seconds.

## Summary

Extends the `PATCH /openapi/operations/k8/deployments/{name}/scale` endpoint to be aware of autoscaling and to support explicit pause/resume control alongside the existing replicas and resource update functionality.

**What changed:**

- Added `autoscalingMode` field to `DeploymentScaleRequest` (`"active"` or `"paused"`)
- `autoscalingMode=active` — resumes autoscaling by removing the pause annotation from the deployment's ScaledObject; returns `autoScalingMode: "active"` in the response
- `autoscalingMode=paused` — explicitly pauses autoscaling; freezes at the current replica count if `replicas` is omitted, or at the provided count if specified; returns `autoScalingMode: "paused"` in the response
- `replicas` without a mode — pauses autoscaling if a ScaledObject exists (via annotation), otherwise falls back to a direct scale
- All three modes compose freely with `resources` — resource limit/request updates apply regardless of whether autoscaling state is being changed
- Autoscaling operations that target a deployment with no ScaledObject return a clear `400` rather than silently falling back
- Resume logic extracted into `resumeAutoscaling()` helper; all autoscaling-related helpers grouped together
- All user-facing strings (Swagger docs, API error messages) use generic "autoscaling" language — no internal implementation details exposed to callers

## Design decision: unified endpoint vs. separate resume API

The initial design explored adding a dedicated `POST /deployments/{name}/autoscaling/resume` endpoint to handle resume as a distinct operation. 

The final design uses `autoscalingMode` as an intent field in the request body: `"active"` to resume, `"paused"` to pause explicitly. This composes naturally with `replicas` and `resources` in the same request.

## Backward compatibility

The `autoscalingMode` field is nullable/optional. Existing callers sending only `replicas` or `resources` are unaffected.

## Test plan

- [x] `autoscalingMode=active` on paused deployment → removes annotation, returns `"active"`
- [x] `autoscalingMode=active` on already-active deployment → no-op, returns `"active"`
- [x] `autoscalingMode=active` + `resources` → both annotation removal and resource update applied
- [x] `autoscalingMode=active` on non-autoscaled deployment → `400`
- [x] `autoscalingMode=active` + `replicas` → `400`
- [x] `autoscalingMode=paused` without replicas → pauses at current spec replica count
- [x] `autoscalingMode=paused` with replicas → pauses at specified count
- [x] `autoscalingMode=paused` on non-autoscaled deployment → `400`
- [x] `replicas` on KEDA-managed deployment → pauses via annotation, returns `"paused"`
- [x] `replicas` on KEDA-managed deployment + `resources` → both applied
- [x] `replicas` when KEDA unavailable (404/403) → falls back to direct scale
- [x] `resources` only → no KEDA lookup, plain resource update


david-leifker (approved):
Assuming this works when keda is NOT present, then lgtm!

kewats (approved):

PR #17400, opened 2026-05-12
feat(ingest/thoughtspot): add ThoughtSpot connector

treff7es:
A new source connector that ingests metadata from ThoughtSpot via REST API v2.0. Liveboards become Dashboards, Answers and Visualizations become Charts, Worksheets and Tables become Datasets. Entities sit under Workspace containers when the ingestion principal can enumerate Orgs; otherwise the catalog is flat.

## Coverage

- Workspaces, Liveboards, Answers, Visualizations, Worksheets, Tables emitted via SDK V2 with subTypes.
- Column-level lineage from Visualizations and Answers back to their Worksheets, derived from TML ``column.sources[]`` arrays.
- Cross-platform upstream lineage: federated Worksheets emit table- and column-level lineage to the underlying Databricks / Snowflake / BigQuery / Redshift / etc. URN. Routed via ``external_connections`` per TS connection GUID or display name; preserves source case when needed (BigQuery, SQL Server, case-quoted Databricks).
- Optional view counts via ``metadata_search`` ``include_stats`` (the same global counter shown in the TS UI; no extra API round-trip).
- Ownership, ThoughtSpot tags (as GlobalTags), browse paths, and stateful soft-deletion of entities that disappear between runs.

## Authentication

- Trusted auth (``username`` + ``secret_key``) and password auth via a typed ``auth`` discriminated union; bearer tokens minted per run.
- Mid-run 401 triggers an in-place re-auth; bounded retry surfaces a permanently-disabled principal as a hard failure rather than a silent cascade of recovery info entries.

## Scaling

- Streaming pagination via ``iter_liveboards`` / ``iter_answers``; TML enrichment batched (``tml_export_batch_size=100``) with per-batch graceful degradation.
- Usage cache holds ``(id, views)`` tuples rather than full Pydantic models — keeps memory bounded at 10K+ dashboard scale.
- Per-phase wall-clock timers on the ingestion report (workspace / liveboard / answer / dataset / usage extraction).

## Tests

- 250 unit tests covering pagination, TML chunking, error classification, re-auth wrapper, stats emission, cross-platform lineage, container hierarchy, and stateful deletion.
- Integration test with a mocked-REST golden file.


sgomezvillamor (approved):
Great connector — solid SDK V2 usage, well-designed discriminated-union auth, proper mid-run re-auth, and the TML batch degradation pattern is exactly right. A few things to address before merge:

---

## 🔴 Blockers

**1. Silent dataset wipeout — exhausted generator (`source.py:779`)**
`_get_worksheet_columns_lookup` stores `self.client.get_logical_tables()` (a generator) directly. The dict comprehension on the next lines exhausts it. When the dataset phase later calls `_get_logical_tables()`, it sees a non-None cache and iterates zero elements — all datasets are silently dropped on every real run that has Liveboards.
Fix: `self._logical_tables_cache = list(self.client.get_logical_tables())`

**2. Zero `schemaMetadata` aspects in golden file**
The golden has 4 dataset entities but no `schemaMetadata` aspects. `test_thoughtspot_schema_extraction` is `@pytest.mark.skip`. For a connector that extracts column-level schema via TML, the integration test must validate it. Un-skip the test (or explain in the connector capabilities doc why schema is not emitted) and regenerate the golden.

**3. Pydantic v1 deprecated API (`source.py:382, 2316`)**
`ThoughtSpotConfig.parse_obj(config_dict)` → use `ThoughtSpotConfig.model_validate(config_dict)` in both `create()` and `test_connection()`.

---

## 🟡 Warnings

**4. LiteralString violations in error reporting (`source.py`, 11+ sites)**
`SourceReport.warning(title=..., message=...)` types both as `LiteralString`. f-strings are not literal strings — they break error aggregation in the UI (each unique value creates a new group). Dynamic identifiers must go in `context=`. Affected sites include `source.py:569,595,605,656,662,674` and summary messages at `source.py:511–549`.

**5. `regenerate_golden.py` should be deleted**
Golden files must be (re)generated via `pytest ... --update-golden-files`, never via a custom script. This script has already diverged from the real test fixtures (wrong auth shape, missing endpoint mocks, missing `sql-view-1`). Delete it.

**6. Minimal `upstreamLineage` coverage — test skipped**
The golden has 1 lineage aspect (SQL view path). The mock sets `source_tables` on worksheets and answers, but those relationships don't appear in the golden. `test_thoughtspot_lineage_extraction` is `@pytest.mark.skip`. For a connector advertising `LINEAGE_FINE`, this must be validated before merge.

**7. `AttributeError` on `conn.data_source_type.upper()` (`source.py:2167`)**
`data_source_type` can be `None`. The sibling method `_resolve_sql_view_warehouse` already guards this: `(conn.data_source_type or "").upper()`. Apply the same pattern here.

**8. `_get_connection_lookup()` and `_get_tag_lookup()` lack exception handling**
Both make uncaught API calls. A failure propagates through the calling entity's try/except and silently drops the whole entity instead of gracefully falling back to `{}` with a `report.warning()`.

**9. `ExternalConnectionConfig` overrides full `model_config` (`config.py`)**
`model_config = pydantic.ConfigDict(extra="forbid")` overwrites the parent `ConfigModel.model_config`, losing `ignored_types=(cached_property, ...)` and `json_schema_extra`. Remove the override and rely on the inherited settings — `ConfigModel` already forbids extra fields.

**10. `requests.Session` not closed on client teardown (`client.py`)**
`close()` is a no-op. Add `self.ts_client.requests_session.close()` to release HTTP connections promptly.

**11. Dead code: `processed_*_urns` sets (`source.py:324–326`)**
`processed_workspace_urns`, `processed_liveboard_urns`, `processed_answer_urns` are populated but never read. Remove or wire them into the report.

**12. Fragile `"chart" in wu.get_urn()` substring check (`source.py:1095`)**
Prefer `wu.get_urn().startswith("urn:li:chart:")`, or check whether there's a typed URN helper (e.g. `ChartUrn` / `guess_entity_type`) that would be more idiomatic here.

---

## ℹ️ Suggestions

- **Invalid owner URNs**: Verify that `viz-1`/`viz-2` ownership uses `author_name` (login) not `author.display_name` — `urn:li:corpuser:John Doe` (with spaces) is invalid.
- **`DELETE`/`PUT` in retry `allowed_methods`**: The connector never issues mutating requests; tightening to `frozenset(["GET", "POST"])` avoids accidental retry of future SDK side-effects.
- **Late imports in unit tests**: 22 test functions have imports inside the function body — move to module level per project conventions.
- **No `conftest.py`**: The 300-line `mock_thoughtspot_api` fixture inline in `test_thoughtspot.py` can't be reused by other test files. Extract to `conftest.py`.


---

metadata-ingestion/tests/integration/thoughtspot/README.md
sgomezvillamor, line 1: is this relevant or different from integration tests for other sources?

metadata-ingestion/src/datahub/ingestion/source/thoughtspot/report.py
sgomezvillamor, line 1: rename to `report.py`? so it follows the naming pattern of other files in the dir

treff7es:
Thanks for the thorough review @sgomezvillamor! Addressed in two follow-up commits on this branch.

### Re: B1 — exhausted generator at `source.py:779`

Looked into this carefully — the cached value isn't actually a generator. `get_logical_tables` at `client.py:2099-2104` is annotated `List[LogicalTableResponse]` and the implementation builds + returns a real list (not a yield-based iterator), so `self._logical_tables_cache` is iterable repeatedly and there's no exhaustion in practice. I verified this by tracing call sites — the dataset phase iterates the cache fine on real runs.

That said, the concern is reasonable defensively, and the sibling cache-population path at `_get_logical_tables` already wraps with `list()`. So I added the same `list()` wrap here in `be1c32b` for consistency — if `get_logical_tables` is ever rewritten to yield, the two cache-population paths can't drift.

### Tranche A — `be1c32b` (mechanical fixes)

| Item | Fix |
|---|---|
| B1 | Defensive `list()` wrap (see above) |
| B3 | `parse_obj` → `model_validate` (2 sites) |
| W5 | Deleted `regenerate_golden.py` |
| W7 | `(conn.data_source_type or "").upper()` guard |
| W8 | try/except + `report.warning` on `_get_connection_lookup` + `_get_tag_lookup` |
| W9 | Removed `model_config` override on `ExternalConnectionConfig` (inherits from `ConfigModel`) |
| W10 | `client.close()` now closes `ts_client.requests_session` |
| W11 | Removed 3 dead `processed_*_urns` sets + collapsed call sites to `yield from` |
| W12 | `wu.get_urn().startswith("urn:li:chart:")` |
| S2 | Retry `allowed_methods=frozenset(["GET", "POST"])` |
| L2 | Renamed `thoughtspot_report.py` → `report.py` (matches sibling files) |

### Tranche B — `bc2d84f` (un-skip integration tests, B2/W6)

- Un-skipped `test_thoughtspot_schema_extraction` and `test_thoughtspot_lineage_extraction`.
- Extended the mock fixture to provide `metadata_detail.columns` on logical tables in TS v2 wire shape (`header.{id,name,description}` + `dataType` + `type`) — exercises `_parse_columns_from_metadata_detail` end-to-end.
- Fixed `Answer.source_tables` shape from bare GUID strings to `{id, name}` dicts — the `SourceTableRef` coercer drops bare strings, so the previous fixture silently produced no chart inputs.
- Regenerated golden: 86 → 92 events, now includes 3 `schemaMetadata`, 2 `chartInfo.inputs` (Answer→Worksheet lineage), and 3 `tagKey` aspects from TS column-type tags.

All 299 thoughtspot tests pass; `./gradlew :metadata-ingestion:lintFix` clean.

### Not yet addressed
- **W4 (LiteralString in f-strings)**: substantive refactor across 11+ sites — opted to land the high-confidence mechanical fixes first. Can address in a follow-up if you'd like it in this PR.
- **Suggestions** (owner URNs, late imports in tests, conftest.py extraction): all valid; happy to address in a follow-up if you want them gated on this PR.

treff7es:
Follow-up: addressed the remaining 🟡 W4 and all three ℹ️ Suggestions from your review. PR is now at `715e1b1`.

### W4 — `report.warning` LiteralString cleanup — `efd0662`

Moved every f-string out of `title=` and `message=` so DataHub's UI groups duplicate warnings correctly. Dynamic data (counts, entity IDs, sorted key lists, entity_label/entity_type values) goes in `context=`.

source.py sites: External Lineage Config Keys Unmatched, External Lineage Resolution Failed, SQL View Lineage Parse Failures, `_safe_extract`, `_safe_emit_per_entity`, `_fetch_and_filter_entities` (Permission Denied / Failed to Fetch Entities / Tag Filter Matched Zero Entities).

client.py: same cleanup applied to `_report_model_parse_drops` and `_check_tml_item_status` while in the same code area.

### Owner URN whitespace guard — `c6de94c`

Added `_looks_like_login` to reject any candidate containing whitespace before it reaches `make_user_urn`. The connector previously fell back to `author.name` after the `_looks_like_guid` filter, which let display names like `"John Doe"` through and produced `urn:li:corpuser:John Doe` — malformed; DataHub drops it silently. The fixture's viz-1/viz-2 had exactly this shape, so the golden now correctly omits their ownership aspect.

New unit test `test_custom_properties_skips_display_name_in_author_name` pins the defensive behavior.

### Hoist late imports — `5e46824`

Removed 22 function-scoped imports from `test_thoughtspot_source.py` (all duplicates of names that should have been at module level). Added the missing symbols (`_MAX_REAUTH_ATTEMPTS`, `_http_status_code`, `_classify_http_error`, `_resolve_author_login`, `EntityStats`, `AuditStampClass`, `FineGrainedLineageClass`, `logging`, `yaml`) to the module-level imports. Net `−74 lines`.

### Extract conftest.py — `715e1b1`

Moved `mock_thoughtspot_api` and `test_resources_dir` to `tests/integration/thoughtspot/conftest.py` for auto-discovery by future test modules in this directory. `test_thoughtspot.py` shrinks from 1182 → 429 lines (just test bodies + preamble). Mechanical move; identical fixture behavior.

---

All 300 thoughtspot tests pass; `./gradlew :metadata-ingestion:lintFix` clean. Ready for re-review.

PR #17403, opened 2026-05-12
feat: Add PdlEnumMapper for safe PDL-to-GraphQL enum conversion

AdrianMachado:
## Overview
Introduces a safe enum mapping utility and lint test to prevent crashes when converting PDL enum values to GraphQL-generated Java enums.

## Problem
PDL enums use `$UNKNOWN` as a sentinel for forwards compatibility—when an older schema reads an enum value added in a newer version, Pegasus deserializes it as `$UNKNOWN`. However, GraphQL-generated Java enums do not include `$UNKNOWN`, so raw `Enum.valueOf()` calls will throw `IllegalArgumentException` when encountering this value.

## Solution
- **PdlEnumMapper**: New utility class providing two safe mapping methods:
  - `map(Class<T>, Enum<?>, T defaultValue)` — maps PDL enum to GraphQL enum, falling back to a provided default for unrecognized values
  - `mapDefaultNull(Class<T>, Enum<?> pdlValue)` — same behavior but returns null as fallback
  - Both methods log warnings when falling back, aiding in debugging

- **UnsafeEnumValueOfLintTest**: New lint-style test that scans all `*Mapper.java` files for unsafe `SomeEnum.valueOf(pdlValue.toString()/name())` patterns:
  - Detects both import-based patterns (`EnumName.valueOf(...)`) and fully-qualified patterns (`com.linkedin.datahub.graphql.generated.Foo.valueOf(...)`)
  - Uses regex with `[^;]*?` to stay within single statements and handle multi-line calls
  - Includes a `PREEXISTING_VIOLATIONS` allowlist for existing code that needs migration (tech debt)
  - Provides actionable error messages with file paths, line numbers, and enum names

## Testing
- Added `PdlEnumMapperTest` with unit tests covering:
  - Known value mapping
  - `$UNKNOWN` fallback behavior
  - Unrecognized value fallback
  - Both `map()` and `mapDefaultNull()` variants
- Lint test validates no new unsafe patterns are introduced while allowing pre-existing violations to be tracked for future migration

## Migration Path
Existing code with unsafe `valueOf()` calls is tracked in `PREEXISTING_VIOLATIONS`. These should be migrated to use `PdlEnumMapper` over time. New code must use the mapper to pass the lint test.

https://claude.ai/code/session_01WxD9GYuQP69mC8qrZ5oWTo

### DIFF
--- a/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/types/mappers/PdlEnumMapper.java
+++ b/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/types/mappers/PdlEnumMapper.java
@@ -1,0 +1,53 @@
+package com.linkedin.datahub.graphql.types.mappers;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Safe mapping from PDL enum values to GraphQL-generated Java enums.
+ *
+ * <p>PDL enums use {@code $UNKNOWN} as a sentinel for forwards compatibility: when an older schema
+ * reads an enum value added in a newer version, Pegasus deserializes it as {@code $UNKNOWN}. The
+ * GraphQL-generated Java enums do not include {@code $UNKNOWN}, so a raw {@link Enum#valueOf} call
+ * will throw {@link IllegalArgumentException}. Always use this mapper instead.
+ */
+@Slf4j
+public final class PdlEnumMapper {
+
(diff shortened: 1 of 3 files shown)


jayacryl (approved):
makes sense, but also not java goat
@cursor review this with a critical eye - any smarter approaches? what do the java pros suggest for this sort of problem?

jayacryl:
@cursor read this with a critical eye - any smarter approaches? what do the java pros suggest for this sort of problem?

PR #17404, opened 2026-05-12
fix(search): use _source includes format for ES 8 kNN semantic search

alexjst:
## Summary

Four fixes for GCP semantic search with Vertex AI on Elasticsearch 8:

- **ES 8 `_source` format**: Use `{"includes": [...]}` instead of a bare array — the ES 8 Java client can't deserialize the array form
- **`gemini_embedding_001` model config**: Add Vertex AI model entry to the semantic index models map so the `dense_vector` field exists in the mapping
- **3072 as default dimension**: Use `gemini-embedding-001`'s native output dimensionality (3072) as the default, avoiding client/server mismatch
- **Drop 768-only guard**: Remove the hard guard in `EmbeddingProviderFactory` that rejected any `outputDimensionality != 768` — the guard was based on the incorrect assumption that 768 is the model's native output


alexsku (approved):
Approving — same content as acryldata/datahub-fork#9547. Guard removal + 3072 default + Python outputDimensionality propagation addresses the end-to-end dim consistency.

PR #17409, opened 2026-05-12
fix(datahub-web-react): Member Group UI fixes

ani-malgari:
**Summary**
For groups synced from an external IdP (Okta, Azure AD, etc.), DataHub blocks membership edits because they're owned by the source system. The UX around this was inconsistent and confusing:

When the "Add Member" button is disabled because SSO integration is active, there is no explanation provided to the user as to why the button is not clickable. This creates confusion — users don't know if it's a bug, a permissions issue, or an intentional restriction.
The "Add Member" action is inconsistently handled between the main pane and the side pane in the RBAC / Members settings:
Main pane: The "Add Member" button is blocked (disabled), but no tooltip explains why — leaving users confused. See related issue [UX-84](https://linear.app/acryl-data/issue/UX-84/add-member-button-disabled-due-to-sso-integration-needs-a-tooltip) for the tooltip work.
Side pane: The "Add Member" action is allowed, but leads to an error — creating a worse experience than simply being blocked.
This inconsistency means users may encounter a confusing error in the side pane even though the same action is intentionally restricted in the main pane.

After Changes:
Screenshot 2026-05-12 at 3 51 56 PM
Screenshot 2026-05-12 at 3 52 19 PM

### DIFF
--- a/datahub-web-react/src/app/entityV2/group/GroupInfoHeaderSection.tsx
+++ b/datahub-web-react/src/app/entityV2/group/GroupInfoHeaderSection.tsx
@@ -5,6 +5,7 @@ import React from 'react';
 import styled from 'styled-components';
 
 import { MemberCount } from '@app/entityV2/group/GroupSidebar';
+import { getExternalGroupMembershipTooltip } from '@app/entityV2/group/utils';
 
 import { EntityRelationshipsResult } from '@types';
 
@@ -13,9 +14,18 @@ const GroupHeader = styled.div`
     z-index: 2;
 `;
 
+const NameRow = styled.div`
+    display: flex;
+    align-items: center;
+    gap: 6px;
+    min-width: 0;
+`;
+
 const GroupName = styled(Typography.Title)`
     word-wrap: break-word;
     text-align: left;
+    flex: 1;
+    min-width: 0;
     &&& {
         margin-bottom: 0;
         word-break: break-all;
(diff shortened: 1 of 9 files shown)


datahub-web-react/src/app/entityV2/group/GroupInfoHeaderSection.tsx
```diff
@@ -31,8 +41,15 @@ const GroupName = styled(Typography.Title)`
     }
 `;
 
+const ExternalGroupLock = styled(LockOutlined)`
```
benjiaming, line 44: I see you're converting some icons in this PR to phosphor (like PlusOutlined) but not this one. We could do `lock-key` maybe?

datahub-web-react/src/app/entityV2/group/GroupInfoHeaderSection.tsx
```diff
@@ -31,8 +41,15 @@ const GroupName = styled(Typography.Title)`
...
+    flex-shrink: 0;
+    color: ${(props) => props.theme.colors.bg};
```
benjiaming, line 46: We're setting a background color on an icon? Is it still visible?

datahub-web-react/src/app/entityV2/group/GroupMembers.tsx
```diff
@@ -176,15 +178,27 @@ export default function GroupMembers({ urn, pageSize, isExternalGroup, onChangeM
...
                 >
-                    <UserAddOutlined />
-                    <AddMemberText>Add Member</AddMemberText>
-                </AddMember>
+                    {/*
+                     * Wrapper needed so the Tooltip has a hover target: antd's `disabled`
+                     * Buttons get `pointer-events: none`, which swallows mouse events and
+                     * prevents the Tooltip from firing when disabled.
+                     */}
+                    <div style={{ display: 'inline-block', cursor: isExternalGroup ? 'not-allowed' : 'auto' }}>
```
benjiaming, line 190: Any reason for inline styles?

benjiaming (approved):
Approving with comments.

javabrett (approved):

ani-malgari:
> Approving with comments.

@benjiaming  you brought up good points. i'll create a separate PR for your comments as the cloud PR has been merged already with this SOT and it's easy to autoresolve new changes.

PR #17410, opened 2026-05-12
fix: mask invite token in ViewInviteTokenModal for security

AdrianMachado:
## Description
Added `data-meticulous-mask` attribute to the invite link display element in `ViewInviteTokenModal` to prevent meticulous from flagging changes to the URL (which changes on every vercel deployment)


ani-malgari (approved):
Thank you 🚀

PR #17415, opened 2026-05-13
fix(ownership): always materialize ownerTypes from owners list

alfiyas-datahub:
## Summary

The `OwnershipOwnerTypes` mutation hook short-circuited `setOwnerTypes` whenever the grouped-by-type owner map on the incoming proposal equaled the previous aspect's grouped map. Ingestion connectors (dbt and others) routinely emit `Ownership` with `owners` populated and `ownerTypes` omitted; on repeat ingestion the grouped map matched, the hook never set `ownerTypes`, and the empty map on the proposal overwrote the persisted materialization. Because `ownerTypes` is the field Elasticsearch indexes for ownership-type search facets and Metadata Tests, ownership-type filters silently stopped returning the entity even though its `Owner` list was intact.

The hook now always (re)materializes `ownerTypes` from `owners` on the incoming proposal, and only flags a mutation when the newly materialized map differs from the one already on the proposal. This preserves idempotency for no-op writes (no spurious MCL emission) while guaranteeing the denormalization is never lost.

## Root cause

Old logic compared the grouped map of the proposal against the grouped map of the **previous** aspect and used inequality as the gate for re-materialization. Same `owners` → same grouped map → skip → `ownerTypes` left empty on the proposal → empty map persisted. The invariant the hook is supposed to enforce — *`ownerTypes` is a server-managed denormalization of `owners`* — only held on the first write of a given owner set.

## Changes

| File | What |
| --- | --- |
| `entity-registry/.../OwnershipOwnerTypes.java` | Always materialize `ownerTypes` from `owners`; compare against the proposal's existing `ownerTypes` (not the previous aspect's) to decide whether the proposal mutated. |
| `entity-registry/.../OwnershipOwnerTypeTest.java` | New unit test `testRepeatIngestPreservesOwnerTypesWhenProposalOmitsThem` covering the exact regression: stored aspect has populated `ownerTypes`, proposal omits them, same `owners`. |
| `smoke-test/tests/openapi/v3/ownership_owner_types.json` | New declarative fixture driven by the existing `test_openapi` harness. Double-ingests via OpenAPI v3 and asserts `ownerTypes` survives the repeat write. |

## Verification

- `./gradlew :entity-registry:check` — passes; all 11 `OwnershipOwnerTypeTest` cases green.
- **Unit-test bisect**: stashed the source fix and re-ran the new test against the unfixed source. The test fails on the assertion that `ownerTypes` is re-materialized (proves coverage, not vacuous).
- **Smoke-test bisect**: ran the new fixture against a GMS built with the unfixed source. Step 3 fails with `{'dictionary_item_added': ["root['value']['ownerTypes']"]}` — the buggy server returns no `ownerTypes` key after the second write. Re-running against the fixed server passes.
- Live triple-ingest of a dbt-style payload (owners populated, `ownerTypes` omitted) through the REST emitter: `ownerTypes` has the expected 2 entries after every write.

### DIFF
--- a/entity-registry/src/main/java/com/linkedin/metadata/aspect/hooks/OwnershipOwnerTypes.java
+++ b/entity-registry/src/main/java/com/linkedin/metadata/aspect/hooks/OwnershipOwnerTypes.java
@@ -6,6 +6,7 @@ import static com.linkedin.metadata.Constants.OWNERSHIP_ASPECT_NAME;
 import com.linkedin.common.*;
 import com.linkedin.common.urn.Urn;
 import com.linkedin.common.urn.UrnUtils;
+import com.linkedin.data.DataMap;
 import com.linkedin.data.template.RecordTemplate;
 import com.linkedin.metadata.aspect.RetrieverContext;
 import com.linkedin.metadata.aspect.batch.ChangeMCP;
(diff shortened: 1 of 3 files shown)


david-leifker (approved):
lgtm!

PR #17431, opened 2026-05-13
feat(search): add timeseries write throttle for Elasticsearch index u…

david-leifker:
…pdates

Introduces an in-memory Guava cache that rate-limits timeseries aspect writes to both entity search and timeseries Elasticsearch indices. This prevents high-frequency timeseries aspects (e.g. datasetProfile, operation) from overwhelming Elasticsearch with redundant index updates.

Key design decisions:
- Single shared cache keyed by (URN, aspect) across V2, V3, and semantic search strategies with centralized recordWrite to avoid race conditions
- Default refresh period with per-entity, per-aspect JSON overrides via MCL_TIMESERIES_THROTTLE_REFRESH_OVERRIDES
- Independent enable flags for entity index and timeseries index paths
- Disabled by default; cache TTL auto-derived from max configured period
- Batch-level ThrottleSummary logging to avoid per-event log spam

### DIFF
--- a/metadata-io/src/main/java/com/linkedin/metadata/service/UpdateIndicesV2Strategy.java
+++ b/metadata-io/src/main/java/com/linkedin/metadata/service/UpdateIndicesV2Strategy.java
@@ -83,6 +83,9 @@ public class UpdateIndicesV2Strategy implements UpdateIndicesStrategy {
   // Cache for semantic index existence checks to avoid repeated HEAD requests
   private final Cache<String, Boolean> semanticIndexExistsCache;
 
+  // Throttle cache for timeseries aspect writes
+  @Nullable private final TimeseriesWriteThrottleCache timeseriesThrottleCache;
+
   /**
    * Creates an UpdateIndicesV2Strategy with optional semantic search support.
    *
@@ -112,7 +115,8 @@ public class UpdateIndicesV2Strategy implements UpdateIndicesStrategy {
       @Nullable SemanticSearchConfiguration semanticSearchConfig,
       @Nonnull IndexConvention indexConvention,
       boolean coalesceBatchUpdates,
-      @Nonnull V2MappingsBuilder mappingsBuilder) {
+      @Nonnull V2MappingsBuilder mappingsBuilder,
+      @Nullable TimeseriesWriteThrottleCache timeseriesThrottleCache) {
     this.v2Config = v2Config;
     this.elasticSearchService = elasticSearchService;
     this.searchDocumentTransformer = searchDocumentTransformer;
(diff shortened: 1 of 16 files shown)


metadata-io/src/main/java/com/linkedin/metadata/service/UpdateIndicesV2Strategy.java
```diff
@@ -246,6 +268,61 @@ private void processAspectGroup(
...
+    long eventTimeMs =
+        event.getAuditStamp() != null
+            ? event.getAuditStamp().getTime()
+            : System.currentTimeMillis();
+
+    boolean throttled =
+        timeseriesThrottleCache.shouldThrottle(entityName, urnStr, aspectName, eventTimeMs);
+
+    // Entity index path
+    if (throttled && timeseriesThrottleCache.isEntityIndexEnabled()) {
```
craig-rueda: slight perf improvement here - evaluate `timeseriesThrottleCache.isEntityIndexEnabled()` first, then call `shouldThrottle()`. The former is a simple `bool` check, whereas the latter goes to the cache, etc.

metadata-io/src/main/java/com/linkedin/metadata/service/UpdateIndicesV2Strategy.java
```diff
@@ -246,6 +268,61 @@ private void processAspectGroup(
     appendCoalescedRunIds(opContext, survivor, aspectEvents);
   }
 
+  /**
+   * Applies timeseries throttle checks around entity-index and timeseries-index writes. For
+   * non-timeseries aspects, both writes execute unconditionally.
+   */
+  private void processTimeseriesThrottled(
```
nwadams, line 275: Is it worth adding a shadow mode that keeps track of hot writes to log entities or entity types that we should consider setting throttling rules for? This could be something we enable broadly and alert on.

RyanHolstien (approved):

metadata-io/src/main/java/com/linkedin/metadata/service/UpdateIndicesV2Strategy.java
```diff
@@ -246,6 +268,61 @@ private void processAspectGroup(
...
+      }
+    } else {
+      entityIndexWrite.run();
+      if (throttleSummary != null) {
+        throttleSummary.recordWritten(TimeseriesWriteThrottleCache.ThrottleTarget.ENTITY_INDEX);
+      }
+    }
+
+    // Timeseries index path
+    if (throttled && timeseriesThrottleCache.isTimeseriesIndexEnabled()) {
```
RyanHolstien: I think most of the time we probably don't want to throttle the timeseries index itself since it's append only and should be quick as well as it being the source of truth for this data, but good to break it apart just in case we do.

PR #17432, opened 2026-05-13
fix(unity-catalog): accept ui-only authentication_type field

kyungsoo-datahub:


AdrianMachado:
I don't see the point of doing this vs removing from the UI

PR #17434, opened 2026-05-13
fix(ingest/snowflake): fix silently dropped views in batched SHOW VIEWS

rob-1019:
## Summary

When `INFORMATION_SCHEMA` returns null `VIEW_DEFINITION` for views the
ingestion role does not own, DataHub falls back to issuing batched
`SHOW VIEWS LIKE '<prefix>%'` queries — one per prefix group — to
recover those definitions; when this fallback is triggered on a schema
with more than 1,000 such null-definition views with alphabetically
diverse names, a packing bug caused one group per packed batch to be
silently skipped, leaving those views without `viewProperties` (no DDL,
no lineage edges) in DataHub.

Manually verified pre- and post-patch against a real Snowflake account:
a purpose-built 2,401-view schema ingested as a non-owner role (all
definitions null in `INFORMATION_SCHEMA`) showed the buggy branch
issuing only 4 of the 7 required `SHOW VIEWS` queries and producing 800
`viewProperties` aspects with the `PREFIX_B` group entirely absent,
while the patched branch issued all 7 queries and produced 1,200
`viewProperties` aspects covering every non-secure prefix group.

---

Fixes a silent failure in the Snowflake source where view definitions
(and therefore view lineage) were not ingested for a subset of views in
schemas with > 1,000 null-definition views, even when those views
matched the user's `table_pattern`/`view_pattern` filter and
`include_view_definitions: true` was set.

Observed in production: a schema with 1,122 views produced the warning
*"Could not fetch definitions for 287 views in SOME_DB.SCHEMA after
processing all batches."* The dataset URN was created and
schema/columns were populated, but the `viewProperties` aspect
(containing the DDL) and view-lineage edges were **completely absent** —
not empty, not null, simply never emitted.

## Root cause

`_maybe_populate_empty_view_definitions` in `snowflake_schema.py` called
`build_prefix_batches(..., max_groups_in_batch=1)` and then took
`batch[0]` from each returned batch. A comment claimed this was safe
(*"max_groups_in_batch=1 makes it safe to access batch[0]"*) — but the
underlying `_batch_prefix_groups` packer used a strict `>` check on the
group-count axis, so `max_groups_in_batch=N` actually packed up to N+1
groups per batch. The second group of every paired batch was silently
dropped — no `SHOW VIEWS LIKE '<prefix>%'` query ran for those prefixes,
so the `text`/DDL column was never fetched for any view in that group.

The function's intent is one `SHOW VIEWS LIKE '<prefix>%'` query per
prefix group. Pre-fix, `batch[0]` silently discarded the second group in
each packed batch, so those queries were never built or issued. They
were always supposed to happen — the bug acted as an accidental skip.

**Concrete trace** (the production case): a top-level group split by
first letter yielded `A(681)`, `B(55)`, `C(55)`. The greedy packer:

```
A(681) → batch=[A], size=681
B(55)  → 681+55=736 ≤ 1000 and len=1 not > 1 → batch=[A, B], size=736
C(55)  → len=2 > 1 → close [A, B]; start fresh with [C]
```

Result: `batches = [[A, B], [C]]`. Caller takes `batch[0]` of each →
only `A` and `C` get SHOW VIEWS queries; `B` is dropped entirely.
Repeating across the alphabet produced the observed 10-query SHOW VIEWS
sequence (A, C, E, G, K, M, O, R, T, V) with prefixes D/F/H/L/N/P/S/U
silently missing, and 287/1,122 unfetched view definitions.

**Why it went undetected:** the bug requires ALL three conditions
simultaneously:

1. **More than 1,000 null-definition views in a single schema.** The
   batcher only receives views whose `VIEW_DEFINITION` is null in
   `INFORMATION_SCHEMA` — total view count is irrelevant. A 5,000-view
   schema where the ingestion role owns 4,800 views passes only 200
   names into the batcher (no split, no bug). Null definitions occur
   when the ingestion role is not the owner of the views
   (`INFORMATION_SCHEMA` hides `VIEW_DEFINITION` from non-owners even
   for non-secure views in most Snowflake editions).

2. **Alphabetically diverse view names** such that adjacent prefix
   groups each have fewer than 1,000 names and their combined count
   also fits under 1,000 — allowing the packer to pair them into one
   batch. Schemas with uniformly prefixed names (e.g. all `VIEW_NNNN`)
   produce one large group and one tiny group whose combined count
   exceeds 1,000; the size check separates them before the count check
   can mis-pack them.

3. **`fetch_views_from_information_schema: true`** (or the information
   schema code path being taken for another reason), so that
   `_maybe_populate_empty_view_definitions` is invoked at all.

## The fix

Three concerns addressed in one commit:

1. **Caller fix** — flatten the comprehension in
   `_maybe_populate_empty_view_definitions` to iterate every
   `PrefixGroup` instead of slicing `batch[0]`. This alone fixes the
   production bug. The batch structure from the packer becomes
   irrelevant — every group is processed regardless of how many were
   packed per batch.

2. **Library fix** — change `_batch_prefix_groups` to use the same
   `(current_state) + (incoming_delta) > cap` form for the group-count
   axis that the size axis has always used. The parameter name
   `max_groups_in_batch=N` now means exactly N groups per batch.

   The column-fetch call site bumps `max_groups_in_batch` from `5` to
   `6` to preserve the historical actual behavior (the old `>` check
   with value `5` allowed 6 groups per batch; the new `+1 >` check
   with value `6` also allows 6 groups per batch).

3. **Regression tests** — see Test plan below.

## End-to-end validation

Validated against a real Snowflake account with a purpose-built schema
(2,401 views across 6 prefix groups, ingested as a non-owner role so
all definitions are null in `INFORMATION_SCHEMA`):

| | `acryl-main` (buggy) | fix branch |
|---|---|---|
| SHOW VIEWS queries issued | **4** | **7** |
| `viewProperties` aspects emitted | **800** | **1,200** |
| PREFIX_B views with `viewProperties` | **0 (absent)** | **400** |

Main branch log (bug):
```
Found 4 batch(es).
Processing batch 1/4: SHOW VIEWS LIKE 'PREFIX_A%' ...
Processing batch 2/4: SHOW VIEWS LIKE 'PREFIX_C%' ...  ← PREFIX_B skipped
Processing batch 3/4: SHOW VIEWS LIKE 'SECURE_B%' ...  ← SECURE_A skipped
Processing batch 4/4: SHOW VIEWS LIKE 'SECURE_T%' ...  ← SECURE_C skipped
Fetched definitions for 800 out of 2,401 targeted views.
Could not fetch definitions for 1,601 views.
```

Fix branch log:
```
Found 7 batch(es).
Processing batch 1/7: SHOW VIEWS LIKE 'PREFIX_A%' ...
Processing batch 2/7: SHOW VIEWS LIKE 'PREFIX_B%' ...
Processing batch 3/7: SHOW VIEWS LIKE 'PREFIX_C%' ...
Processing batch 4/7: SHOW VIEWS LIKE 'SECURE_A%' ...
Processing batch 5/7: SHOW VIEWS LIKE 'SECURE_B%' ...
Processing batch 6/7: SHOW VIEWS LIKE 'SECURE_C%' ...
Processing batch 7/7: SHOW VIEWS LIKE 'SECURE_T%' ...
Fetched definitions for 1,200 out of 2,401 targeted views.
```

The 1,201 views with no definition after the fix are all secure views —
`SHOW VIEWS` withholds the `text` column from non-owners on secure
views by design; this is expected and correct behavior.

## Test plan

- [x] New snowflake-level regression test verified to **fail** without
      the caller-side fix and pass after.
- [x] New library-level invariant test covers 7 parameter combinations.
- [x] Existing `test_build_prefix_batches_exceeds_max_batch_size`
      still passes under the new semantics.
- [x] 19 unit tests pass (4 pre-existing builder + 7 new invariant
      parametrizations + 7 pre-existing snowflake schema + 1 new
      snowflake regression).
- [x] End-to-end validated against real Snowflake with a non-owner
      ingestion role across a 2,401-view schema; bug reproduced on
      main, absent on fix branch.

## Checklist

- [x] Links to related issues — none filed; this PR documents the
      observed symptom.
- [x] Tests for the changes have been added.
- [x] Docs related to the changes have been added/updated — N/A;
      bug fix to existing behavior, no surface change.
- [x] Breaking change considerations — none. Ingestion produces *more*
      metadata than before; no schema or config API changes.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema.py
@@ -926,13 +926,14 @@ class SnowflakeDataDictionary(SupportsAsObj):
             return []
 
         view_names = [view.name for view in views_with_empty_definition]
-        batches = [
-            batch[0]
+        # build_prefix_batches packs multiple PrefixGroups per batch; we issue one
+        # SHOW VIEWS query per prefix, so flatten and iterate over every group.
+        prefix_groups = [
+            group
             for batch in build_prefix_batches(
                 view_names, max_batch_size=1000, max_groups_in_batch=1
             )
-            if batch
-            # Skip empty batch if so, also max_groups_in_batch=1 makes it safe to access batch[0]
+            for group in batch
         ]
 
         view_map: Dict[str, SnowflakeView] = {
@@ -942,12 +943,14 @@ class SnowflakeDataDictionary(SupportsAsObj):
 
         logger.info(
             f"Fetching definitions for {len(view_map)} views in {db_name}.{schema_name} "
-            f"using batched 'SHOW VIEWS ... LIKE ...' queries. Found {len(batches)} batch(es)."
+            f"using batched 'SHOW VIEWS ... LIKE ...' queries. Found {len(prefix_groups)} batch(es)."
         )
 
-        for batch_index, prefix_group in enumerate(batches):
+        for batch_index, prefix_group in enumerate(prefix_groups):
             query = f'SHOW VIEWS LIKE \'{prefix_group.prefix}%\' IN SCHEMA "{db_name}"."{schema_name}"'
-            logger.info(f"Processing batch {batch_index + 1}/{len(batches)}: {query}")
+            logger.info(
+                f"Processing batch {batch_index + 1}/{len(prefix_groups)}: {query}"
+            )
 
             try:
                 cur = self.connection.query(query)
@@ -1562,7 +1565,7 @@ class SnowflakeDataDictionary(SupportsAsObj):
         else:
             # Build batches for full schema scan
             object_batches = build_prefix_batches(
-                all_objects, max_batch_size=10000, max_groups_in_batch=5
+                all_objects, max_batch_size=10000, max_groups_in_batch=6
             )
 
         # Process batches
--- a/metadata-ingestion/src/datahub/utilities/prefix_batch_builder.py
+++ b/metadata-ingestion/src/datahub/utilities/prefix_batch_builder.py
@@ -55,8 +55,10 @@ def _build_prefix_groups(names: List[str], max_batch_size: int) -> List[PrefixGr
 def _batch_prefix_groups(
     groups: List[PrefixGroup], max_batch_size: int, max_groups_in_batch: int
 ) -> List[List[PrefixGroup]]:
-    """Batch the groups together, so that no batch's total is larger than `max_batch_size`
-    and no group in a batch is larger than `max_group_size`."""
+    """Batch the groups together, so that no batch holds more than
+    `max_groups_in_batch` groups and the total size of names in a batch
+    does not exceed `max_batch_size` (a single oversized group is allowed
+    to exceed `max_batch_size` on its own, since groups are atomic)."""
 
     # A batch is a set of groups.
 
@@ -67,9 +69,14 @@ def _batch_prefix_groups(
     current_batch_size = 0
     batch: List[PrefixGroup] = []
     for group in groups:
+        # Both the name-count check and the group-count check ask the same
+        # question: "would adding this group push us over the cap?" — i.e.
+        # (current state) + (incoming delta) > cap. If either trips, close
+        # the current batch before appending. The `+ 1` is the incoming
+        # group itself.
         if (
             current_batch_size + len(group.names) > max_batch_size
-            or len(batch) > max_groups_in_batch
+            or len(batch) + 1 > max_groups_in_batch
         ):
             batches.append(batch)
             batch = []
--- a/metadata-ingestion/tests/unit/snowflake/test_snowflake_schema.py
+++ b/metadata-ingestion/tests/unit/snowflake/test_snowflake_schema.py
@@ -207,6 +207,72 @@ class TestSnowflakeDataDictionary:
         assert result[0].view_definition == "SELECT * FROM table1"
         assert result[1].view_definition == "SELECT col1 FROM table2"
 
+    def test_maybe_populate_empty_view_definitions_covers_all_prefixes(
+        self, mock_snowflake_data_dictionary_information_schema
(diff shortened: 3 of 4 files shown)


sgomezvillamor (approved):
Great fix — root cause is clear, both the caller and library are corrected, and the test coverage is solid. Approving. Three non-blocking suggestions:

---

**1. Name the magic numbers**

`max_batch_size=10000` and `max_groups_in_batch=6` (and `1000`/`1` in the SHOW VIEWS path) are load-bearing values that aren't self-explanatory at the call site. Consider extracting named constants with a brief doc comment, e.g.:

```python
# Matches SHOW_COMMAND_MAX_PAGE_SIZE; single oversized groups are allowed to exceed this.
_COLUMN_FETCH_MAX_BATCH_SIZE = 10000
# Bumped from 5 → 6 when _batch_prefix_groups adopted correct `+1 >` semantics
# (old `> 5` allowed 6 groups per batch; new `+1 > 6` also allows 6).
_COLUMN_FETCH_MAX_GROUPS_IN_BATCH = 6
```

Same pattern for the SHOW VIEWS constants (`1000` / `1`).

---

**2. Stale comments in the new regression test**

The docstring says _"off-by-one (separately addressed later in this series)"_ — there's no series; both fixes land here. Drop the parenthetical.

The inline comment `# … packer produces [[A], [B, C]]` describes the **old** packer's output. With the fix, `max_groups_in_batch=1` now produces `[[A], [B], [C]]`. Suggest:

```python
# With the old packer, B and C landed in one batch [[A], [B, C]] and the
# old batch[0]-only caller silently dropped C. Both fixes in this commit
# prevent that.
```

---

**3. Test parameter label is stale**

```python
# Production column-fetch parameters.
([f"NAME_{i:05d}" for i in range(50000)], 10000, 5),
```

Production `max_groups_in_batch` was just bumped to **6**. Update to 6 (or reference the new constant once it exists), otherwise this tests a value that no longer matches production.

---

PR #17436, opened 2026-05-14
feat(ingestion): per-connector CLI version matrix + resolution stamp …

puneetagarwal-datahub:
feat(ingestion): per-connector CLI version matrix + audit stamp + test-connection path

Adds per-connector CLI version resolution so a Snowflake fix can ship without
forcing a bump on every other connector. A JSON matrix maps server release ×
connector → version, with optional per-deployment cohort allowlists for canary
rollouts.

Matrix key conventions: Server-version keys are git describe --tags truncated at the first - (so v1.1.0-cloud → v1.1.0); CLI versions (_default, cohort version) are forwarded verbatim to pip install acryl-datahub==… and must use PyPI format (no v prefix).

Matrix schema:
```
{
  "v1.1.0": {                              // server key
    "snowflake": {
      "_default": "1.6.0",                 // CLI: PyPI version, no `v`
     
    },
    "bigquery": { "_default": "1.6.0" }
  },
  "v1.0.2": {
    "snowflake": { "_default": "1.5.0.19" }
  }
}
```

Resolution priority (top wins):
  1. Per-source explicit `config.version`                  (SOURCE_CONFIG_OVERRIDE)
  2. matrix[serverVersion][connectorType]._default         (MATRIX_CONNECTOR_DEFAULT)
  3. defaultCliVersion from application.yaml               (APPLICATION_DEFAULT)

Three execution entry points are covered:
  - CreateIngestionExecutionRequestResolver  (manual triggers)
  - IngestionScheduler.ExecutionRequestRunnable  (scheduled triggers)
  - CreateTestConnectionRequestResolver  (test-connection from UI)

Each one stamps a structured `CliVersionAudit` record on the resulting
ExecutionRequestInput aspect, alongside the resolved version string in
`args.version`. 

Feature is off by default — when no `http.url` is set the factory wires a
no-op source and every `resolveVersion()` returns empty, preserving existing
behavior exactly. 

Config (nested under `ingestion.cliVersionMatrix`):

  INGESTION_VERSION_MATRIX_SOURCE           (default: "http"; "none" disables)
  INGESTION_VERSION_MATRIX_URL              (default: empty → no-op)
  INGESTION_VERSION_MATRIX_REFRESH_SECONDS  (default: 600)
  INGESTION_VERSION_MATRIX_AUTH_TOKEN


sgomezvillamor:
Overall looks good. The ingestion team doesn't usually review/approve backend code, so a code owner from the platform team should weigh in here. Sharing my concerns below.

---

**⚠️ Warning — Naming is too generic in several places**

Some names don't convey enough context. Everything in this feature is specifically about *ingestion CLI version resolution*, and the names should reflect that — e.g. `Matrix` reads like a general data structure but is really an `IngestionCliVersionMatrix`. Same pattern applies across the new classes and the service itself.

Also worth revisiting:
- `CliVersionSource` enum values: `PER_SOURCE` reads oddly, and `WORKSPACE_DEFAULT` isn't terminology we normally use here. Worth a second pass.
- `CliVersionProvenance` — the `Provenance` suffix isn't common in this codebase. `Audit` might be a better fit.

---

**💡 Suggestion — Logging placement**

`CliVersionResolutionHelper` is fine staying logging-free. Instead, add structured `debug`/`warn` logging at the resolver level (e.g. `CreateIngestionExecutionRequestResolver`, `CreateTestConnectionRequestResolver`, `IngestionScheduler`) where the resolution result is consumed and acted upon.

**💡 Suggestion — Linear id references**

The rreferences to linear ids in the code will become stale and require checking another system for having the context. Recommendation: Replace with self-contained explanation

---

**⚠️ Warning — Map key readability**

`Map<String, Map<String, ConnectorEntry>>` — the keys are opaque. Is there a way to make it more readable without a lot of overhead? Even a type alias or a dedicated wrapper would make the intent clear at a glance. Actually, we could define proper java pojos/beans, that would also help with the json/yaml de/serialization.

---

**⚠️ Warning — Config coupling**

HTTP-specific config is hardcoded directly under `ingestion:`, and the source type is implicitly "HTTP if URL is set, NoOp otherwise." Updating the backend or adding a second (GMS aspect, AppConfig, etcd) would require piling more flat properties here. A more extensible structure:

```yaml
ingestion:
  cliVersionMatrix:
    source: http    # http | none (default: none)
    http:
      url: "${INGESTION_VERSION_MATRIX_URL:}"
      refreshSeconds: ${INGESTION_VERSION_MATRIX_REFRESH_SECONDS:600}
      authToken: "${INGESTION_VERSION_MATRIX_AUTH_TOKEN:}"
```

---

**💡 Suggestion — Version validation regex (`HttpUrlMatrixSource.java:72`)**

The regex is intentionally permissive, but this should be explicitly documented as not validating PEP 440 fully. Also worth confirming it actually matches our current internal versioning scheme.

---

**⚠️ Warning — `HttpUrlMatrixSource` executor lifecycle**

The `ScheduledExecutorService` is created in the constructor but never stored, so there's no way to shut it down gracefully on Spring context teardown. Needs a `@PreDestroy` hook.

---

**⚠️ Warning (low) — Thread not named**

The refresh thread gets a generic `pool-N-thread-1` name. Use a named `ThreadFactory` for easier thread dump analysis.

---

**⚠️ Warning (low) — Two JSON libraries**

`org.json` is used in the resolvers while `HttpUrlMatrixSource` uses Jackson. Consider consolidating.

---

**⚠️ Warning (low/informational) — SSRF**

`URLConnection` accepts any protocol (`file://`, `jar://`, etc.). Consider using Spring's `RestTemplate` or Java 11's `HttpClient`, which are HTTP/HTTPS-only by default.

---

sgomezvillamor:
Overall this looks good and ready for platform review. The previous concerns were addressed thoroughly — appreciate the turnaround.

---

**Still open before merge**

- `CliVersionResolutionHelper` (and new `CliVersionResolutionLogger`) were not renamed — everything else got the `Ingestion` prefix treatment but these two were left behind.
- PR body is outdated: still references `cliVersionResolution` (now `cliVersionAudit`), and lists `matchedCohortIndex` / `matrixFetchedAt` fields that were never implemented. Worth updating so reviewers aren't misled.

**Still not resolved — platform to confirm criticality**

- Two JSON libraries (`org.json` in resolvers, Jackson in `HttpUrlIngestionCliVersionMatrixSource`)
- SSRF: `URLConnection` accepts any protocol (`file://`, `jar://`). Low/informational since the URL is operator-configured.

---

**💡 Comment style**

Some Javadoc comments describe development iterations rather than the current state. For example, `ServerEntry`:

> *"Wrapping the inner map in a named POJO instead of exposing `Map<String, ConnectorEntry>` keeps lookup sites self-documenting: `serverEntry.getConnectorEntry(type)` vs the previous `serverEntry.get(type)` where the meaning of either string was opaque."*

The `vs the previous` comparison will be meaningless to a future reader. Comments should explain the current design, not how it evolved.

**💡 Backward-compat inference**

The factory's URL-presence fallback (`unset → auto-infer from http.url`) is documented as backward-compatibility for deployments that pre-date the discriminator. This is a new feature — there are no pre-discriminator deployments in production. The inference adds complexity for zero benefit; just require an explicit `source` when a URL is set (or pick the default cleanly without framing it as compatibility).

---

**❓ Observability**

What do we have beyond logs to know this feature is healthy? Specifically:

- Is there a metric (counter/gauge) tracking matrix fetch successes and failures? Logs alone make it hard to alert on a matrix that's been silently stale for hours.
- Is the resolved tier (`SOURCE_CONFIG_OVERRIDE`, `MATRIX_COHORT`, etc.) surfaced anywhere in the UI or as a metric? Would be useful for operators to spot unexpected fallback patterns across executions without querying the DB directly.

---

puneetagarwal-datahub:
@sgomezvillamor For the observability , I need to talk with analytics team , as current metrics are in hex . Wanted to know how can we expose this data over there

puneetagarwal-datahub:
@sgomezvillamor 
I have fixed all the addressed comments . Please review

sgomezvillamor:
Overall LGTM
Let's ask platform to review

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/ingest/execution/CreateTestConnectionRequestResolver.java
```diff
@@ -26,23 +32,63 @@
...
   private static final String TEST_CONNECTION_TASK_NAME = "TEST_CONNECTION";
   private static final String TEST_CONNECTION_SOURCE_NAME = "MANUAL_TEST_CONNECTION";
   private static final String RECIPE_ARG_NAME = "recipe";
   private static final String VERSION_ARG_NAME = "version";
   private static final String DEFAULT_EXECUTOR_ID = "default";
+  private static final String SOURCE_FIELD = "source";
+  private static final String TYPE_FIELD = "type";
+
+  /** Reused for the small recipe-JSON peek done by {@link #extractSourceType(String)}. */
+  private static final ObjectMapper MAPPER = new ObjectMapper();
```
david-leifker: Don't create more object mappers, use the one in OperationContext

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/ingest/execution/CreateIngestionExecutionRequestResolver.java
```diff
@@ -122,11 +142,34 @@ public CompletableFuture<String> get(final DataFetchingEnvironment environment)
...
-                      ingestionSourceInfo.getConfig().getVersion(),
-                      _ingestionConfiguration.getDefaultCliVersion()));
+              // Per-source version may be null, empty, or whitespace-only — bootstrap YAML
+              // templating renders `version: "{{ config.version }}"` as 3 spaces when the source
+              // has no pin, and Mustache treats missing keys as empty rather than failing. The
+              // helper normalizes all three to "unset" so resolution falls through to the matrix
+              // / application default; without that, the blank would forward to the executor and
+              // silently pin to its bundled CLI.
+              final String explicitVersion =
+                  ingestionSourceInfo.getConfig().hasVersion()
```
david-leifker: There should be a utility for this already, please re-use existing utility.

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/ingest/execution/CreateTestConnectionRequestResolver.java
```diff
@@ -110,4 +171,25 @@ public CompletableFuture<String> get(final DataFetchingEnvironment environment)
...
         "get");
   }
+
+  /**
+   * Best-effort extraction of {@code source.type} from a recipe JSON document. Returns {@code null}
+   * for any malformed input — the resolver falls back to {@code defaultCliVersion} in that case
+   * rather than failing the request, since a malformed recipe will surface a clearer error
+   * downstream when the executor parses it.
+   */
+  static String extractSourceType(final String recipeJson) {
```
david-leifker: This should probably be located in a util class

ingestion-scheduler/src/main/java/com/datahub/metadata/ingestion/IngestionScheduler.java
```diff
@@ -409,11 +417,32 @@ public void run() {
...
-                ingestionSourceInfo.getConfig().getVersion(),
-                ingestionConfiguration.getDefaultCliVersion()));
+        // Per-source version may be null, empty, or whitespace-only — bootstrap YAML templating
+        // renders `version: "{{ config.version }}"` as 3 spaces when the source has no pin, and
+        // Mustache treats missing keys as empty rather than failing. The helper normalizes all
+        // three to "unset" so resolution falls through to the matrix / application default;
+        // without that, the blank would forward to the executor and silently pin to its bundled
+        // CLI.
+        final String explicitVersion =
+            ingestionSourceInfo.getConfig().hasVersion()
```
david-leifker: Use the common util method

metadata-models/src/main/pegasus/com/linkedin/execution/ExecutionRequestInput.pdl
```diff
@@ -7,7 +7,8 @@ import com.linkedin.common.Urn
  * TODO: Determine who is responsible for emitting execution request success or failure. Executor?
  */
 @Aspect = {
-  "name": "dataHubExecutionRequestInput"
+  "name": "dataHubExecutionRequestInput",
+  "schemaVersion": 2
```
david-leifker, line 11: Incrementing a version requires a mutator to upgrade from version 1 to 2. I don't think I see a class implementing this. Please review documentation for zero-downtime upgrade mutators.

metadata-service/configuration/src/main/java/com/linkedin/metadata/config/HttpMatrixSourceConfiguration.java
```diff
@@ -0,0 +1,41 @@
...
+   *
+   * <ul>
+   *   <li>GitHub PAT: {@code "token ghp_xxxxxxxxxxxxxxxx"}
+   *   <li>OAuth / OIDC bearer: {@code "Bearer eyJ..."}
+   * </ul>
+   *
+   * <p>The property name ends with "Token" so it is auto-redacted by the system-info properties
+   * collector (see {@code PropertiesCollectorConfigurationTest}).
+   */
+  private String authToken;
```
david-leifker, line 40: This leads me to ask how this will work in airgapped environments. I had assumed that the configuration was from GMS. We need to support systems with no outside access.

metadata-service/configuration/src/main/java/com/linkedin/metadata/ingestion/HttpUrlIngestionCliVersionMatrixSource.java
```diff
@@ -0,0 +1,425 @@
...
+   *   <li>{@code 1.5.0.6rc1} — release candidate
+   *   <li>{@code 1.5.0.13.post1} — post-release
+   *   <li>{@code 1!0.0.0.dev0} — PEP 440 epoch + dev tag
+   *   <li>{@code acryl-1.6.0+acryl.20251031} — internal build with local version identifier
+   * </ul>
+   *
+   * <p>Rejects: whitespace, embedded JSON like {@code {"version":"…"}}, HTML fragments, URLs,
+   * anything containing characters outside the allowed set.
+   */
+  private static final Pattern VALID_VERSION_PATTERN = Pattern.compile("^[\\w.+!-]+$");
```
david-leifker, line 105: You also need to support "bundled", this is not a complete list of version patterns. There are other special version strings besides that one as well used by ingestion which are not python version strings.

metadata-service/configuration/src/main/resources/application.yaml
```diff
@@ -104,6 +104,29 @@ ingestion:
   defaultCliVersion: "${UI_INGESTION_DEFAULT_CLI_VERSION:@cliVersion@}"
   maxSerializedStringLength: "${INGESTION_MAX_SERIALIZED_STRING_LENGTH:16000000}" # Indicates the maximum allowed JSON String length Jackson will handle, impacts the maximum size of ingested aspects
   batchRefreshCount: ${INGESTION_BATCH_REFRESH_COUNT:100} # The number of entities to refresh in a single batch when refreshing entities after ingestion
+  # Per-connector CLI version matrix. When configured, overrides defaultCliVersion on a per-connector
+  # basis. Update the matrix file externally (S3/CDN/gist) without redeploying GMS to change
+  # connector versions. Set source=none (or leave url empty) to disable.
+  cliVersionMatrix:
+    # Source backend: "http" (default) or "none". Setting "none" is an explicit kill-switch even
+    # when an http.url is configured. With "http" but no url, the factory wires a no-op source.
+    source: "${INGESTION_VERSION_MATRIX_SOURCE:http}"
```
david-leifker, line 113: Ideally there is a way to source this matrix from GMS itself, while a centralized external source is useful for Saas, there are on-prem scenarios where an individual operator may want to have control of this matrix without relying on an external system.

metadata-service/factories/src/main/java/com/linkedin/gms/factory/ingestion/IngestionCliVersionMatrixServiceFactory.java
```diff
@@ -0,0 +1,95 @@
...
+  @Qualifier("gitVersion")
+  private GitVersion gitVersion;
+
+  /**
+   * Picks the storage backend for the matrix from {@code ingestion.cliVersionMatrix.source}.
+   * Explicit {@code "none"} is a kill-switch that always wins. Otherwise the HTTP source is wired
+   * when {@code http.url} is non-empty, and a no-op source when the URL is empty.
+   */
+  @Bean(name = "ingestionCliVersionMatrixSource")
+  @Scope("singleton")
```
david-leifker: nit: singleton should be the default

david-leifker (changes requested):
Overall this looks reasonable. A few corner cases and gaps that need addressing.

metadata-service/configuration/src/main/java/com/linkedin/metadata/config/HttpMatrixSourceConfiguration.java
sgomezvillamor: Good point @david-leifker !

The matrix storage is pluggable and configurable. `HttpMatrixSource` is just the first implementation. AFIK the plan is to serve it from S3 in a follow-up iteration, by implementing a new matrix source.

Neither HTTP nor S3 may work for air-gapped environments, but that's just a matter of providing a matrix source that fits their requirements, whether fetching from GMS itself or somewhere else (e.g. a ConfigMap in k8s?).

Is air-gapped support a blocking requirement for this first iteration?

metadata-service/configuration/src/main/resources/application.yaml
sgomezvillamor: We're certainly the most interested in this feature, managing a fleet of instances. It's true, it can be useful even for smaller onprem setups (dev, pre, prod).

Either way, this ties into my comment here: https://github.com/datahub-project/datahub/pull/17436#discussion_r3325459645 — the key point being that air-gapped support doesn't need to be solved now, as long as we know it can be addressed later with different matrix source implementations.

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/ingest/execution/CreateTestConnectionRequestResolver.java
```diff
@@ -26,23 +32,63 @@
...
+  private static final ObjectMapper MAPPER = new ObjectMapper();
 
   private final EntityClient _entityClient;
   private final IngestionConfiguration _ingestionConfiguration;
+  private final IngestionCliVersionMatrixService _versionMatrixService;
 
+  /** Two-arg constructor — no per-connector version matrix is consulted. */
   public CreateTestConnectionRequestResolver(
       final EntityClient entityClient, final IngestionConfiguration ingestionConfiguration) {
+    this(entityClient, ingestionConfiguration, null);
```
RyanHolstien: This telescoping constructor anti-pattern is unnecessary since it appears you are not actually using the 2 parameter one for creation, you are using the 3 parameter one in GmsGraphQLEngine.java.

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/ingest/execution/CreateIngestionExecutionRequestResolver.java
```diff
@@ -122,11 +142,34 @@ public CompletableFuture<String> get(final DataFetchingEnvironment environment)
...
+                      ? ingestionSourceInfo.getConfig().getVersion()
+                      : null;
+              final IngestionCliVersionResolutionHelper.Result resolution =
+                  IngestionCliVersionResolutionHelper.resolve(
+                      explicitVersion,
+                      ingestionSourceInfo.getType(),
+                      _versionMatrixService,
+                      _ingestionConfiguration.getDefaultCliVersion(),
+                      _versionMatrixService != null
```
RyanHolstien: Why are we passing through the service and a property pulled from the service? Just pull the property from the service in the lower function.

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/ingest/execution/CreateIngestionExecutionRequestResolver.java
```diff
@@ -48,11 +53,26 @@ public class CreateIngestionExecutionRequestResolver
 
   private final EntityClient _entityClient;
   private final IngestionConfiguration _ingestionConfiguration;
+  private final IngestionCliVersionMatrixService _versionMatrixService;
 
+  /** Two-arg constructor — no per-connector version matrix is consulted. */
   public CreateIngestionExecutionRequestResolver(
       final EntityClient entityClient, final IngestionConfiguration ingestionConfiguration) {
+    this(entityClient, ingestionConfiguration, null);
```
RyanHolstien: Same here wrt telescoping constructor

ingestion-scheduler/src/main/java/com/datahub/metadata/ingestion/IngestionScheduler.java
```diff
@@ -363,6 +370,7 @@ public ExecutionRequestRunnable(
       this.systemOpContext = systemOpContext;
       this.entityClient = Objects.requireNonNull(entityClient);
       this.ingestionConfiguration = Objects.requireNonNull(ingestionConfiguration);
+      this.versionMatrixService = Objects.requireNonNull(versionMatrixService);
```
RyanHolstien, line 373: Other classes have this as nullable? This will throw a null pointer exception whenever it does not exist and this is a required bean I believe.

ingestion-scheduler/src/main/java/com/datahub/metadata/ingestion/IngestionScheduler.java
```diff
@@ -409,11 +417,32 @@ public void run() {
...
+            ingestionSourceInfo.getConfig().hasVersion()
+                ? ingestionSourceInfo.getConfig().getVersion()
+                : null;
+        final IngestionCliVersionResolutionHelper.Result resolution =
+            IngestionCliVersionResolutionHelper.resolve(
+                explicitVersion,
+                ingestionSourceInfo.getType(),
+                versionMatrixService,
+                ingestionConfiguration.getDefaultCliVersion(),
+                versionMatrixService != null ? versionMatrixService.getServerVersion() : null);
```
RyanHolstien: As stated above, this is unreachable. Objects.requireNonNull throws an exception if it's null so it would be non-nullable here.

metadata-service/factories/src/main/java/com/linkedin/gms/factory/graphql/GraphQLEngineFactory.java
```diff
@@ -137,6 +138,10 @@ public class GraphQLEngineFactory {
   @Qualifier("gitVersion")
   private GitVersion gitVersion;
 
+  @Autowired
+  @Qualifier("ingestionCliVersionMatrixService")
+  private IngestionCliVersionMatrixService versionMatrixService;
```
RyanHolstien, line 143: Yeah if this is Autowired without required=false this should not have nullability options in the rest of the code.

metadata-models/src/main/pegasus/com/linkedin/execution/ExecutionRequestInput.pdl
puneetagarwal-datahub: The schemaVersion: 2 is applied automatically by the bump_schema_versions.py pre-commit hook (any aspect change → master + 1), so it's not something I set by hand. 
Since cliVersionAudit is an optional additive field, there's no stored payload to rewrite — and AspectMigrationMutatorChain's trailing-gap bridge already advances old v1 records to v2 during the sweep with no transform (it fires even when an aspect has no registered mutator). 

So a v1→v2 mutator here would be a no-op. Happy to add a passthrough mutator anyway if you'd prefer it for convention/traceability — just let me know.

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/ingest/execution/CreateIngestionExecutionRequestResolver.java
puneetagarwal-datahub: sure

metadata-service/configuration/src/main/java/com/linkedin/metadata/config/HttpMatrixSourceConfiguration.java
puneetagarwal-datahub: Hi @david-leifker  we have already mentioned this gap in RFC for the same . https://www.notion.so/acryldata/RFC-Independent-version-deployment-of-connectors-by-customer-and-connector-type-74afc6a6427783508d768150ae08be46?source=copy_link

On a high level thought , We can plugin the matrix inside the codebase which is shipped , start a local server and server json via that or we can think of something else .

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/ingest/execution/CreateIngestionExecutionRequestResolver.java
puneetagarwal-datahub: will Do here too — only the 3-arg constructor remains

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/ingest/execution/CreateIngestionExecutionRequestResolver.java
puneetagarwal-datahub: Dropped the serverVersion parameter

ingestion-scheduler/src/main/java/com/datahub/metadata/ingestion/IngestionScheduler.java
puneetagarwal-datahub: will do

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/ingest/execution/CreateTestConnectionRequestResolver.java
puneetagarwal-datahub: Removed the 2-arg constructor

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/ingest/execution/CreateIngestionExecutionRequestResolver.java
puneetagarwal-datahub: Removed the 2-arg constructor

ingestion-scheduler/src/main/java/com/datahub/metadata/ingestion/IngestionScheduler.java
puneetagarwal-datahub: Removed the dead versionMatrixService != null ? … : null ternary — requireNonNull already guarantees non-null.

metadata-service/factories/src/main/java/com/linkedin/gms/factory/graphql/GraphQLEngineFactory.java
puneetagarwal-datahub: Agreed

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/ingest/execution/CreateTestConnectionRequestResolver.java
puneetagarwal-datahub: Dropped the static ObjectMapper

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/ingest/execution/CreateTestConnectionRequestResolver.java
puneetagarwal-datahub: Moved

metadata-service/factories/src/main/java/com/linkedin/gms/factory/ingestion/IngestionCliVersionMatrixServiceFactory.java
puneetagarwal-datahub: Removed

metadata-service/configuration/src/main/java/com/linkedin/metadata/ingestion/HttpUrlIngestionCliVersionMatrixSource.java
puneetagarwal-datahub: The permissive ^[\w.+!-]+$ already accepts bundled and other non-PyPI sentinels (e.g. no-acryl-datahub) — they're alphanumeric/hyphenated, so they pass and flow through to args.version unchanged. Documented this in the Javadoc (noting the list is intentionally non-exhaustive so future sentinels keep working) and added a regression test permissiveVersionPatternAcceptsSpecialSentinels.

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/ingest/execution/CreateIngestionExecutionRequestResolver.java
puneetagarwal-datahub: We are now using a common method — IngestionCliVersionResolutionHelper

ingestion-scheduler/src/main/java/com/datahub/metadata/ingestion/IngestionScheduler.java
puneetagarwal-datahub: We are now using a common method — IngestionCliVersionResolutionHelper

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/ingest/execution/CreateIngestionExecutionRequestResolver.java
puneetagarwal-datahub: Couldn't route this through resolveIngestionCliVersion — it does blank→default with no matrix lookup, so it'd bypass the feature. Replaced all its callers with the new shared IngestionCliVersionResolutionHelper.resolve(...) and removed the now-unused old util (+ its tests) so there's a single resolution path and no footgun.

metadata-service/configuration/src/main/java/com/linkedin/metadata/config/HttpMatrixSourceConfiguration.java
david-leifker: As long as there is no regression in the current design, per source configuration with `bundled` and a common default configured via GMS's config endpoint, not a blocker.

david-leifker (approved):
Approving, I would have liked to see a comprehensive implementation with GMS' config endpoint being able to serve the matrix configuration, passed through via environment variable containing a json configuration option. It would have been a simple and straightforward addition to serving it from a remote location.

puneetagarwal-datahub:
> Approving, I would have liked to see a comprehensive implementation with GMS' config endpoint being able to serve the matrix configuration, passed through via environment variable containing a json configuration option. It would have been a simple and straightforward addition to serving it from a remote location.

Good call — the source layer is pluggable by design .We kept the HTTP source as the default because the primary goal is updating per-connector versions fleet-wide without redeploying GMS ( edit the matrix once, all instances refresh in ~10min). Inline can't give that, but it's a great fit for self-contained/air-gapped deployments and is cheap to add alongside.

I will work on it once this is shipped and working .

PR #17439, opened 2026-05-14
feat(ingest/airflow-plugin): drop Airflow 2 support, require Airflow 3.x

dinesh-verma-datahub:
## Summary

Removes the dual-Airflow-version code paths from `acryl-datahub-airflow-plugin`. The plugin no longer carries Airflow-2-specific code, the legacy standalone `openlineage-airflow` package, or per-version extractor wiring. It now targets **Airflow 3.0–3.x** on a single code path built on the Task SDK (`airflow.sdk`).

This is a **breaking** release for users still on **Airflow 2.x** — they must pin an older `acryl-datahub-airflow-plugin` (see the compatibility ladder in `docs/lineage/airflow.md`). **Airflow 3.0, 3.1, and 3.2 are supported** and covered by CI.

## Supported versions

`apache-airflow>=3.0.0,<4.0.0`. The plugin imports `BaseOperator`, `BaseSensorOperator`, `Connection`, `Context`, `DAG`, and `Asset` from `airflow.sdk` (all present since 3.0). `BaseHook` is the one symbol that moved into the Task SDK only in 3.1 (`airflow.sdk.bases.hook`); `hooks/datahub.py` imports it from there with a fallback to the legacy `airflow.hooks.base` so **Airflow 3.0.x keeps working**.

### Package dependencies

**`metadata-ingestion-modules/airflow-plugin/setup.py`**

- Floor bumped to `apache-airflow>=3.0.0,<4.0.0` (drops Airflow 2).
- `apache-airflow-providers-openlineage>=2.1.0` is now an **unconditional** base dependency (2.1.0 is the first version shipping the Airflow 3 listener interface).
- The standalone `openlineage-airflow` package is no longer referenced anywhere.
- The `[airflow2]` install extra is removed. **`[airflow3]` is kept as a no-op extra** so existing `pip install 'acryl-datahub-airflow-plugin[airflow3]'` commands and pinned recipes keep working (it resolves to the same closure as the bare package).
- `integration_test_requirements`: dropped the `apache-airflow[snowflake,amazon,google]` bundle in favor of unpinned `apache-airflow-providers-{snowflake,amazon,google}` plus the existing `sqlite`/`teradata` providers.

**`metadata-ingestion/setup.py` (core) — required for Airflow 3.2 install compatibility**

Airflow 3.2's constraints file pins newer versions of several packages that acryl-datahub had capped, making `pip install` under those constraints unsatisfiable. Raised the upper bounds (resolved versions are unchanged; `uv.lock` re-synced via `updateLockFile`):

- `tabulate<0.10.0` → `<0.11.0`
- `ruamel.yaml<0.19.0` → `<0.20.0` (clib now ships wheels for all supported Pythons/platforms)
- `packaging<26.0.0` → `<27.0.0`

`numpy<2` and `pydantic<2.12` were intentionally left untouched — they live in extras the plugin does not install and guard documented scikit-learn / pyiceberg incompatibilities.

### Code removed

- The entire `src/datahub_airflow_plugin/airflow2/` package (~2.5k LOC of legacy `openlineage-airflow` extractor wiring).
- `AIRFLOW_3_MIGRATION.md`.
- `example_dags/airflow2/` and `example_dags/airflow3/` — flattened to a single `example_dags/`.
- `tests/integration/dags/airflow2/` and `tests/integration/dags/airflow3/` — flattened to `tests/integration/dags/`.
- Per-version golden duplicates — kept one canonical golden per scenario (regenerated against Airflow 3.1, and validated to also match 3.0 and 3.2 in CI).
- `tests/unit/test_teradata_extractor.py` (covered the deleted Airflow-2 extractor).
- `get_provider_info()` no longer emits the deprecated `hook-class-names` block.
- Dead Airflow-2-era scaffolding in the listener/shims (`ExtractorManager`, `_task_holder`, `HAS_AIRFLOW_*` flags, unused shim exports, the dead top-level `_airflow_compat.py`, etc.) and per-version try/except import blocks across the operators/hooks.

### Behavioral / config breaking changes

- `datajob_url_link=taskinstance` (the Airflow-2-only URL format) now raises a `ValueError` with a migration-friendly message at config load. Valid values: `tasks` (default) and `grid`; the `DatajobUrl.TASKINSTANCE` enum member is gone.
- The `patch_snowflake_schema` config option is removed (it only patched a Snowflake bug in the deprecated `openlineage-airflow` `SnowflakeExtractor`; the OpenLineage provider handles this natively). Stale `airflow.cfg` entries are silently ignored.

### Code renamed / restructured

- `airflow3/_airflow3_sql_parser_patch.py` → `airflow3/_sql_parser_patch.py`.
- `_airflow_shims.py` / `datahub_listener.py` are now thin re-exports of the `airflow3/` modules (no version branching).
- `_get_base_url()` in `client/airflow_generator.py` now prefers `[api] base_url` over `[webserver] base_url`, falling back to `http://localhost:8080`.
- `DagRunProxy` gains a `run_type` property (`DagRunType.SCHEDULED` when `run_id` starts with `scheduled__`, else `MANUAL`) so the OpenLineage provider's scheduler-vs-manual check works against the proxy.

### CI / tooling

- `.github/workflows/airflow-plugin.yml`: matrix is now **three blocking lanes** — Python 3.11 × `apache-airflow~=3.0.0` / `~=3.1.0` / `~=3.2.0`, each with its constraints file.
- `tox.ini`: `envlist = py311-airflow30, py311-airflow31, py311-airflow32` (factor-conditional Airflow pins/constraints).
- `.github/python-deps-exclusions.txt`: adds `apache-airflow-providers-{snowflake,amazon,google}` to the unpinned list.
- `Dockerfile.test`, `README.docker.md`, `DOCKER_TEST_GUIDE.md`, `run-tests.sh`: describe the three envs (default `py311-airflow31`).

### Tests

- `tests/integration/test_plugin.py`: removed `get_api_version()`, `is_airflow3()`, the v2→v1 URL fallback, and the `DAGS_FOLDER_AIRFLOW2/3` split.
- `tests/unit/test_teradata_openlineage_patch.py`: removed the Airflow-2-provider-path fallback test.
- `tests/unit/test_airflow.py`: added `test_datajob_url_link_taskinstance_rejected_with_migration_message` and `test_basehook_falls_back_to_legacy_location_on_airflow_30` (stubs the 3.0 import surface and proves the `BaseHook` fallback keeps the hook working); removed the Airflow-3 skip from `test_dags_load_with_no_errors`.

### Docs

- `docs/lineage/airflow.md`: compatibility section now lists 2.x ❌ and 3.0+ ✅; the version-pin ladder reflects 3.0+ support.
- `docs/how/updating-datahub.md`: new "Next / Breaking Changes" entry (Airflow 2 dropped, 3.0+ required, `[airflow3]` retained as a no-op).
- `metadata-ingestion/developing.md` and the airflow-plugin README/Docker docs: updated to the three supported tox envs.

## Test plan

- [x] CI: the `airflow-plugin` workflow passes on all three lanes (Airflow 3.0.0 / 3.1.0 / 3.2.0).
- [ ] `tox -e py311-airflow30 | py311-airflow31 | py311-airflow32` run unit + integration tests locally (or via `./run-tests.sh`).
- [ ] `python -c "from datahub_airflow_plugin.hooks.datahub import DatahubRestHook; from datahub_airflow_plugin.operators.datahub import DatahubEmitterOperator"` works after `pip install -e .` on Airflow 3.0 and 3.2 venvs.
- [ ] Setting `datajob_url_link=taskinstance` raises a `ValueError` on listener init.

## Migration

Users on **Airflow 2.7–2.10** should pin the last `acryl-datahub-airflow-plugin` release before this change (see the compatibility table in `docs/lineage/airflow.md`).

Users on **Airflow 3.0+** can upgrade directly. Optional cleanups:

1. Drop `openlineage-airflow` from any pinned constraints — it is no longer used.
2. `[airflow3]` still works, but `pip install acryl-datahub-airflow-plugin` (no extra) is now equivalent.
3. If `airflow.cfg` sets `datajob_url_link=taskinstance`, change it to `tasks` (default) or `grid`.
4. Remove `patch_snowflake_schema` from `airflow.cfg` if present.


sgomezvillamor (approved):
Clean drop of Airflow 2 support. Verified no regression for existing Airflow 3 users — all removed code paths were already no-ops on Airflow 3, and the new `DagRunProxy`/`BaseHook` fallbacks improve forward compatibility across 3.0–3.2. The three-lane CI matrix is a solid improvement over the single 3.1 lane we had before.

---

sgomezvillamor:
The new three-lane matrix (3.0 / 3.1 / 3.2) is a great improvement over the single 3.1 lane we had before. One gap worth considering: **Python 3.12 is not covered**. Airflow 3.x officially supports Python 3.9–3.12, and a growing share of users are already on 3.12. Since nothing in the plugin is Python-version-specific, adding a 3.12 lane (even just for one Airflow version, e.g. 3.2) would be low-effort coverage. Not a blocker for this PR.

---

PR #17443, opened 2026-05-14
feat(ingest/mongodb): add DocumentDB as a separate data platform

hhobson:
## Why

The MongoDB source can be used to ingest AWS DocumentDB (due to DocumentDB's MongoDB-compatible API), but the DataHub platform is given as MongoDB.

Users have not been able to find DocumentDB collections in DataHub as they didn't realise they would be included in with the as MongoDB collections. DocumentDB is not a hosted MongoDB, but a separate database engine and in environments where both are being used it makes sense to clearly differentiated in DataHub.

## What changed

- Add a DocumentDB platform to data-platforms.yaml
- New opt-in config flag `emit_as_documentdb` (default: False) on MongoDBConfig. 
  - When `emit_as_documentdb: True` and `hostingEnvironment: AWS_DOCUMENTDB`, the connector emits entities under a DocumentDB data platform.
  - Report warning if `emit_as_documentdb: True` is set without `hostingEnvironment:  AWS_DOCUMENTDB`, so misconfiguration is visible in the ingestion report.
- No change to @platform_name("MongoDB") — recipes still use type: mongodb. The connector's identity stays MongoDB; only the emitted data-platform URN switches.
- Tests added for the four possible combinations of `hostingEnvironment` and `emit_as_documentdb`

## Design choices

- Two-condition gate (flag + hosting environment) prevents accidental mislabeling if the flag is set against a genuine MongoDB instance.
- Default stays False — existing recipes are unaffected and continue emitting mongodb URNs.


sgomezvillamor:
Thanks for the contribution — the motivation is solid and the test coverage (all four `emit_as_documentdb` × `hostingEnvironment` combinations, plus the `platform_instance` propagation test) is genuinely good work.

---

**🔴 Blocker — Design: replace two-condition gate with a `platform` Literal field**

The current design requires two fields to line up (`emit_as_documentdb: true` + `hostingEnvironment: AWS_DOCUMENTDB`) before anything changes. Setting just one is either a silent no-op or a warning. This is more complexity than needed.

The established pattern across DataHub connectors is a single `platform` field with a constrained type:

```python
# sql_generic.py, s3/config.py, abs/config.py all do this
platform: Literal["mongodb", "documentdb"] = Field(
    default="mongodb",
    description="Platform to emit entities under. Use 'documentdb' for AWS DocumentDB.",
)
```

If you still want to guard against misconfiguration, add a `@model_validator` that rejects `platform="documentdb"` unless `hostingEnvironment=AWS_DOCUMENTDB`. That gives you the same safety gate without the implicit two-condition logic, and it fails loud at config parse time rather than silently at runtime.

In `MongoDBSource.__init__` this then becomes just `self.platform = config.platform` — no conditional, no warning.

---

**🟡 Warning — DocumentDB platform type doesn't match MongoDB**

In `data-platforms.yaml`, the new entry uses `type: KEY_VALUE_STORE`. MongoDB itself is registered as `type: OTHERS`. DocumentDB is a document database, not a key-value store — it should be `OTHERS` to stay consistent.

---

**ℹ️ Suggestion — Container URN migration not mentioned**

The migration note covers dataset URNs, but switching platforms also orphans **container** URNs (the database-level containers emitted via `gen_containers`). Worth a one-liner in `updating-datahub.md`.

---

**📝 Docs — follow-on from fixing the blocker**

Once `emit_as_documentdb` is replaced by `platform: Literal[...]`, the `mongodb_post.md` section "Emitting as the DocumentDB platform" can be removed (it describes the old flag). Instead, a short paragraph in `docs/mongodb/README.md` explaining that the same connector works for DocumentDB by setting `platform: documentdb` would be the natural home for that guidance.

---

sgomezvillamor:
**🔴 Blocker — Frontend: ingest source picker logo missing**

`datahub-web-react/src/app/ingestV2/source/builder/constants.ts` maintains a hardcoded `PLATFORM_URN_TO_LOGO` map for the ingestion UI source picker. DocumentDB needs to be added there (import the logo + add `urn:li:dataPlatform:documentdb` → logo entry), otherwise the platform will appear without a logo in the ingest builder. Entity pages are fine — those resolve the logo dynamically from GraphQL.

---

hhobson:
> 🟡 Warning — DocumentDB platform type doesn't match MongoDB
>
> In data-platforms.yaml, the new entry uses type: KEY_VALUE_STORE. MongoDB itself is registered as type: OTHERS. DocumentDB is a document database, not a key-value store — it should be OTHERS to stay consistent.

MongoDB is listed as a `KEY_VALUE_STORE` in `metadata-service/configuration/src/main/resources/bootstrap_mcps/data-platforms.yaml`, so have left DocumentDB type as it is.

hhobson:
@sgomezvillamor thanks for the review.

Using `platform` Literal field is much cleaner and I've implemented this. Also added missing platform URN to logo mapping.

sgomezvillamor (approved):
Thanks for the contrib!

sgomezvillamor:
@hhobson this is ready to be merged, are you planning any further work here?

hhobson:
> @hhobson this is ready to be merged, are you planning any further work here?

@sgomezvillamor I'm not planning to add anything else

PR #14304, opened 2025-08-01
feat(ingestion/snowflake): internal marketplace support

acrylJonny:
## Summary

Adds Snowflake Internal Marketplace (private data sharing) ingestion support. Each listing is surfaced as a DataHub **Data Product** with:

- Ownership derived from listing provider / support contacts
- Structured properties mapping listing metadata to DataHub properties
- Optional enrichment via `DESCRIBE AVAILABLE LISTING`
- Optional usage statistics from `SNOWFLAKE.DATA_SHARING_USAGE.LISTING_TELEMETRY_DAILY`
- Domains auto-created per listing provider

**Depends on:**
- #17449 — `fix: tag patch path/APK ordering (Parsson JsonException)` (PR A)
- #17450 — `feat(ingest): patch-based writes for user-editable aspects` (PR B)

## Configuration

```yaml
source:
  type: snowflake
  config:
    marketplace:
      mode: INTERNAL_MARKETPLACE   # or ALL
      internal_marketplace_listing_pattern:
        allow: [".*"]
      fetch_internal_marketplace_listing_details: true
      ingest_marketplace_usage: true
```

## New files

| File | Description |
|---|---|
| `snowflake_marketplace.py` | Core marketplace ingestion logic |
| `datahub-web-react/src/app/entityV2/shared/utils.ts` | Snowflake URL recognition for listing entities |
| `tests/unit/snowflake/test_snowflake_marketplace*.py` | Unit tests (behaviour + data-quality) |
| `tests/integration/snowflake/test_snowflake_marketplace.py` | Integration golden test |
| `examples/library/dataproduct_create_with_owners.py` | Example script |
| Snowflake docs | Updated quick-ingestion guide and connector docs |

## Code quality fixes applied

- `report.report_dropped()` misuse replaced with `report.report_marketplace_listing_filtered()`
- `_enrich_listing_custom_properties` return type changed from `Dict[str, Any]` to `_ListingEnrichmentResult` TypedDict
- `marketplace_listings_filtered` counter added to `SnowflakeV2Report`

## Test plan

- [ ] `./gradlew :metadata-ingestion:testQuick -k snowflake_marketplace`
- [ ] Integration test: `pytest tests/integration/snowflake/test_snowflake_marketplace.py --update-golden`
- [ ] Manual verification against a Snowflake account with internal marketplace listings

## Checklist

- [x] PR conforms to Contributing Guideline
- [x] Tests added/updated
- [x] Docs added/updated
[...]

PR #17449, opened 2026-05-14
fix(metadata/patch): scaffold trailing-empty JSON Pointer tokens (Parsson JsonException)

acrylJonny:
## Summary

Fix the Parsson `JsonException` thrown by `TemplateUtil.populateTopLevelKeys` when an `add`/`replace` patch op targets a JSON Pointer path whose final reference token is the empty string (e.g. `/tags/<urn>/`).

This is the wire format produced by every unattributed patch builder that uses compound array primary keys with a nullable component:

- `GlobalTagsPatchBuilder.addTag(urn)`
- `GlossaryTermsPatchBuilder.addTerm(urn)`
- `OwnershipPatchBuilder.addOwner(owner, type, typeUrn)`
- `StructuredPropertiesPatchBuilder.setNumberProperty` / `setStringProperty`

Before this PR, calling any of these on an entity whose target aspect did not yet exist would throw `jakarta.json.JsonException` server-side, blocking patch-based writes for first-touch entities (e.g. fresh Snowflake ingestion).

## Approach

Per [@kyungsoo-datahub's review](https://github.com/datahub-project/datahub/pull/17449#pullrequestreview-...), the fix lives in `TemplateUtil.populateTopLevelKeys` instead of per-builder client-side workarounds. The scaffolding loop now extends through trailing-empty tokens, which fixes every affected builder at once with no wire-format change, no new client API surface, and no per-builder churn.

The earlier client-side workaround (path/APK reorder, `wildcardRemoveOps` split, `buildAll()`, `IllegalStateException` on mixed batches, Python `HasTagsPatch` rewrite, fixture regen) is reverted as part of this commit.

## Files changed

| File | Change |
| --- | --- |
| `entity-registry/.../template/TemplateUtil.java` | Scaffold through trailing-empty path tokens (~10 lines) |
| `entity-registry/.../template/TemplateUtilTest.java` | New — unit tests for the scaffolding loop |
| `entity-registry/.../template/GlobalTagsTemplateTest.java` | Regression test using `GlobalTagsPatchBuilder.addTag` wire format on a fresh aspect |
| `entity-registry/.../template/GlossaryTermsTemplateTest.java` | Same, for `GlossaryTermsPatchBuilder.addTerm` |
| `entity-registry/.../template/OwnershipTemplateTest.java` | Same, for `OwnershipPatchBuilder.addOwner` (4-key APK with two intermediate empty tokens) |
| `entity-registry/.../template/StructuredPropertiesTemplateTest.java` | Same, for `StructuredPropertiesPatchBuilder.setNumberProperty`/`setStringProperty` |
| `docs/how/updating-datahub.md` | Replace the earlier breaking-change entry with a non-breaking server-side fix note |
| 9 client-side files | Revert to `origin/master` (see commit body for the full list) |

## Test plan

- [x] `./gradlew :entity-registry:test --rerun-tasks` — 509 tests pass, including the new `TemplateUtilTest` and the four new template-level regressions.
- [x] `metadata-ingestion/venv/bin/python -m pytest tests/unit/patch/` — 15 tests pass (these files revert to master, so they pass for the same reason CI passes on master).
- [x] `./gradlew :metadata-integration:java:datahub-client:test --tests datahub.client.v2.entity.DatasetTest --rerun-tasks` — passes (also reverted to master).

## Checklist

- [x] PR conforms to Contributing Guideline
- [x] Tests added for the server-side fix
- [x] No breaking changes — `docs/how/updating-datahub.md` entry rewritten as non-breaking


kyungsoo-datahub (commented):
Thanks for the clear Parsson diagnosis. Aware this is a prerequisite, so noting the context up front.

I'd like to hold this and push for a server-side fix instead. Three reasons:

1. The bug isn't localized to one builder. Same trailing-slash pattern lives in GlossaryTermsPatchBuilder.addTerm (L43), OwnershipPatchBuilder.addOwner (L41–48), and StructuredPropertiesPatchBuilder.setNumberProperty (L91, L113). #17450 is already shipping this PR's workaround for structured properties. #14304's marketplace pipeline will keep surfacing more. Per-builder client-side fixes scale linearly with builders touched — each one needs its own path/APK reorder, dual-state split, buildAll(), IllegalStateException on mixed batches, and fixture regen.

2. Server is the right layer. Root cause is TemplateUtil.populateTopLevelKeys stopping one path-token short for add ops — fine in general, but fails when the final token is the empty string. Extending the scaffolding through trailing-empty tokens (~5 lines in entity-registry/.../TemplateUtil.java) fixes every bug site at once — no wire-format change, no new client API surface, no per-builder churn.

3. If marketplace timeline forces a client-side approach anyway, two notes:
  - This PR has a silent regression: removeTag(urn).getJsonPatch() now returns an empty patch list because wildcardRemoveOps was split from pathValues but getJsonPatch() still only  reads pathValues. Affects EntityServiceTest.java:2729, :2975 and any external caller using PatchItemImpl.builder().patch(GenericJsonPatch). Must plug before landing.
  - A connector-side read-modify-write transformer (analogous to the existing HasTags.transformPatchToFullAspect) lets #14304 fetch-merge-UPSERT instead of PATCH, sidestepping Parsson entirely with no patch-builder dependency at all.

acrylJonny:
@kyungsoo-datahub thanks for the feedback — agreed on all three points. Pushed the server-side approach as `c53f9f1d87` on top of the existing branch:

- **Server-side fix in `TemplateUtil.populateTopLevelKeys`**: scaffolding now extends through trailing-empty path tokens. ~10 lines including the comment.
- **Covers every affected builder at once**: regression tests added for `GlobalTagsPatchBuilder.addTag`, `GlossaryTermsPatchBuilder.addTerm`, `OwnershipPatchBuilder.addOwner` (4-key APK), and `StructuredPropertiesPatchBuilder.setNumberProperty` / `setStringProperty`, each exercising the exact wire format that previously threw `JsonException` on a fresh aspect. Plus a focused `TemplateUtilTest` for the scaffolding loop itself.
- **`removeTag().getJsonPatch()` regression is gone**: the `wildcardRemoveOps` split is reverted along with the rest of the client-side workaround — `GlobalTagsPatchBuilder`, `AbstractMultiFieldPatchBuilder`, `Entity.buildAccumulatedPatches`, `HasTagsPatch`, the test fixtures, and the `updating-datahub.md` breaking-change entry are all back to master.
- **No wire-format change, no new client API surface** — connectors using `addTag` / `addTerm` / `addOwner` / `setNumberProperty` against fresh entities now just work.

The branch history still has the earlier client-side commits and this new commit reverts + supersedes them; happy to squash to a single clean commit if you'd prefer that for the merge.

On the downstream PRs: we'd rather land this fix and have **#17450** and **#14304** rely on it directly than carry per-builder workarounds.

kyungsoo-datahub (approved):

PR #17450, opened 2026-05-14
feat(ingest): patch-based writes for user-editable aspects

acrylJonny:
## Summary

Adds opt-in PATCH semantics for `structuredProperties` writes from ingestion sources. **Default behavior is unchanged** — every existing caller (third-party connectors included) still emits `changeType: UPSERT`. Snowflake users can opt into PATCH per-recipe to preserve UI- and pipeline-added properties across runs.

### What changed

- **`mcp_builder.py`** — new `StructuredPropertyWriteMode(StrEnum)` with `upsert` / `patch`. `add_structured_properties_to_entity_wu` now takes `write_mode: StructuredPropertyWriteMode = UPSERT`. Original UPSERT body restored; PATCH branch composes `HasStructuredPropertiesPatch` + `MetadataPatchProposal`.
- **`SnowflakeConfig`** — new `structured_properties_write_mode: StructuredPropertyWriteMode = UPSERT` field, description names the trade-off. Threaded through at the two `snowflake_schema_gen.py` call sites.
- **Snowflake golden** — reverted to UPSERT (matches the new default; the integration recipe doesn't set `write_mode`).
- **Tests** — cover both default UPSERT and explicit `write_mode=PATCH` paths.
- **`updating-datahub.md`** — opt-in shape, PATCH trade-off (recipe removals no longer propagate), and the dependency on the server-side TemplateUtil fix in #17449 for PATCH to work end-to-end on fresh aspects.

### Split out of this PR

Per reviewer feedback, two pieces moved to their own PRs:

- **`HasInstitutionalMemoryPatch` mixin** — different goal, URL-encoding deserves isolated review.
- **`mce_builder.GlossaryTerms` alias removal** — needs a deprecation cycle, not a hard delete (external imports break silently).

### PATCH trade-off (documented in three places)

Opt into PATCH only when you want UI/pipeline edits to survive ingestion. Removals from the recipe no longer propagate to DataHub — clean those up via the UI or API.

### Server-side dependency

PATCH mode requires #17449 (server-side `TemplateUtil` fix for trailing-empty JSON Pointer tokens). Without it, first-PATCH on a fresh aspect throws `JsonException`. UPSERT (default) is unaffected.

## Test plan

- `./gradlew :metadata-ingestion:lint` — clean (ruff + mypy)
- `metadata-ingestion/venv/bin/python -m pytest tests/unit/patch/test_structured_properties_patch.py` — 7 passed (default UPSERT + explicit PATCH)
- Snowflake integration golden unchanged (recipe leaves `write_mode` at default UPSERT)

## Checklist

- [x] PR conforms to Contributing Guideline
- [x] Tests added/updated
- [x] Change documented in `docs/how/updating-datahub.md`


metadata-ingestion/src/datahub/emitter/mcp_builder.py
```diff
@@ -227,22 +229,20 @@ def add_tags_to_entity_wu(
     ).as_workunit()
 
 
+class _StructuredPropertiesPatcher(HasStructuredPropertiesPatch, MetadataPatchProposal):
```
kyungsoo-datahub, line 248: Suggest gating this PR on the server-side Parsson resolution proposed on #17449. That fix sits in TemplateUtil.populateTopLevelKeys below every patch builder, so it covers StructuredPropertiesPatchBuilder automatically with no additional work. 

Without it, preserve_structured_properties: true isn't honestly enableable. It passes unit tests but throws in real Snowflake ingestion.

kyungsoo-datahub, line 248: The propagation issues here (recipe edits, Snowflake tag deletions) are inherent to PATCH-without-state, not bugs in this PR. PATCH is incremental by design.

Suggest documenting the trade-off prominently in the PATCH enum description, SnowflakeConfig  field description, and updating-datahub.md, so users opting into PATCH mode understand what they're choosing — something like: "preserves user edits across runs, but does not propagate removals from the recipe or Snowflake-side tag deletions; clean up via UI or API."

Future iterations can add stateful tracking or RMW to close those gaps if/when needed.

kyungsoo-datahub (commented):
Suggest narrowing this to a Snowflake-only opt-in, with the helper itself defaulting to UPSERT.

The current shape changes add_structured_properties_to_entity_wu for every caller, including third-party connectors that didn't opt in. UPSERT → PATCH is a real semantic change (properties removed from the recipe no longer propagate to DataHub) — the updating-datahub.md one-liner undersells the impact for users treating their recipe as source of truth.

Concrete shape:

1. Add write_mode to add_structured_properties_to_entity_wu, defaulting to UPSERT. Existing callers get current behavior — backward compatible. New callers opt into PATCH by passing the mode explicitly.

class StructuredPropertyWriteMode(str, Enum):
    UPSERT = "upsert"   # Default — full-aspect replacement
    PATCH = "patch"     # Per-property add; preserves user-added properties

def add_structured_properties_to_entity_wu(
    entity_urn: str,
    structured_properties: Dict[StructuredPropertyUrn, str],
    write_mode: StructuredPropertyWriteMode = StructuredPropertyWriteMode.UPSERT,
    ...

2. Add structured_properties_write_mode: StructuredPropertyWriteMode = UPSERT to SnowflakeConfig, with a description documenting the trade-off (recipe-as-source-of-truth lost when PATCH is selected).
3. Pass the config through at the two Snowflake call sites (snowflake_schema_gen.py):

yield from add_structured_properties_to_entity_wu(
    entity_urn, properties,
    write_mode=self.config.structured_properties_write_mode,
)

Also suggest splitting out:
- HasInstitutionalMemoryPatch mixin (different goal, has its own URL-encoding concerns worth reviewing in isolation).
- mce_builder.GlossaryTerms alias removal (stealth break for external imports — needs a deprecation cycle, not a hard delete).

Default behavior is preserved for every existing caller; Snowflake users opt in explicitly by name (not a boolean); and the trade-off lives at the function signature and the recipe config where it's visible.

metadata-ingestion/src/datahub/emitter/mcp_builder.py
acrylJonny: Done. We landed the server-side fix on #17449 (`TemplateUtil.populateTopLevelKeys` now scaffolds through trailing-empty JSON Pointer tokens), which fixes `StructuredPropertiesPatchBuilder` plus `GlobalTags`/`GlossaryTerms`/`Ownership` at the same time. `updating-datahub.md` in c1f7239 notes the dependency: PATCH mode requires the GMS in #17449 to be deployed for first-PATCH on a fresh aspect to work. Default UPSERT path is unaffected.

acrylJonny: Trade-off is now documented in three places per the suggestion:

- `StructuredPropertyWriteMode` enum docstring in `mcp_builder.py` — names UPSERT as recipe-as-source-of-truth, PATCH as user-edit-preserving with the recipe-removal caveat.
- `SnowflakeConfig.structured_properties_write_mode` field description — same trade-off in the recipe-author's line of sight.
- `docs/how/updating-datahub.md` — opt-in shape, caveat ("properties removed from the recipe no longer propagate to DataHub — clean those up via the UI or API"), and dependency on #17449.

Default is UPSERT, so users only opt into the trade-off by name.

acrylJonny:
Reshape applied in c1f7239 to address your top-level review.

**`write_mode` opt-in (default UPSERT)**
- Added `StructuredPropertyWriteMode(StrEnum)` with `upsert` (default) and `patch` values.
- `add_structured_properties_to_entity_wu` now takes `write_mode` and the default body is the original UPSERT — every existing caller (third-party connectors included) is unchanged. PATCH is opt-in only.
- `SnowflakeConfig.structured_properties_write_mode: StructuredPropertyWriteMode = UPSERT` threaded through at both `snowflake_schema_gen.py` call sites. Field description names the trade-off.

**Split-outs**
- `HasInstitutionalMemoryPatch` mixin reverted; will land in its own PR (URL-encoding deserves isolated review).
- `mce_builder.GlossaryTerms` alias removal reverted; will go through a deprecation cycle in a separate PR to avoid silently breaking external imports.

**Server-side dependency (second comment)**
- Server-side `TemplateUtil` fix is in #17449 — that covers `StructuredPropertiesPatchBuilder` plus `GlobalTags`/`GlossaryTerms`/`Ownership`, so no per-builder client workarounds needed. `updating-datahub.md` notes the dependency.

**Trade-off documentation (third comment)**
- Documented in three places: `StructuredPropertyWriteMode` enum docstring, `SnowflakeConfig` field description, and `updating-datahub.md`.

Snowflake golden file reverted to UPSERT (matches the new default; the integration recipe doesn't set `write_mode`). Tests cover both default UPSERT (`test_default_emits_upsert`) and explicit PATCH paths.

kyungsoo-datahub (approved):

PR #17456, opened 2026-05-15
feat(docker): add published DataHub Postgres extensions image

david-leifker:
## Summary

Adds a **published `acryldata/datahub-postgres`** image (extensions on top of upstream `postgres`) plus **Gradle** and **GitHub Actions** wiring to build/push it. **Quickstart compose is unchanged** in this PR; consumers can adopt the image in a follow-up once a tag exists in the registry.


.github/workflows/publish-datahub-postgres.yml
```diff
@@ -0,0 +1,117 @@
...
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  id-token: write
+
+env:
+  DOCKER_CACHE: DEPOT
+  DEPOT_PROJECT_ID: s0gr1cr3jd
```
devashish2203: Use ${{ vars.DEPOT_PROJECT_ID }} so that no changes are required in Fork.

devashish2203 (approved):

.github/workflows/publish-datahub-postgres.yml
david-leifker: Will also apply to unified

PR #17459, opened 2026-05-15
fix(precommit): smarter PDL schema version bump detection

chakru-r:
Uses git merge-base to find the true divergence point from the base branch, then diffs only this branch's commits to detect PDL changes (comparing working tree to merge-base so uncommitted changes are included).

If any of those PDL files also changed on the base branch since divergence, blocks with a clear error listing the conflicting files and asking the user to merge or rebase. Unrelated PDL changes on the base branch do not block. 

Also reads base file content via refs/remotes/origin/… so no local checkout of the base branch is required.

### DIFF
--- a/.github/scripts/bump_schema_versions.py
+++ b/.github/scripts/bump_schema_versions.py
@@ -66,8 +66,29 @@ def detect_default_branch() -> str:
     return "master"
 
 
-def get_changed_pdl_files(base_branch: str) -> list[str]:
-    """Return repo-relative paths of PDL files that differ from base_branch."""
+def get_merge_base(remote_ref: str) -> str:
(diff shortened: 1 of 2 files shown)


.github/scripts/bump_schema_versions.py
```diff
@@ -66,16 +66,66 @@ def detect_default_branch() -> str:
     return "master"
```
devashish2203, line 66: Should this be acryl-main for fork?

devashish2203 (approved):

.github/scripts/bump_schema_versions.py
chakru-r: still detects the correct branch, so keeping it consistent

PR #17460, opened 2026-05-15
feat(i18n): infrastructure for i18n (feature flag + user settings backend part)

v-tarasevich-blitz-brain:
This PR includes:
- adding a new feature flag `i18nEnabled` and related react hook with caching in local storage
- adding user localization settings (storing and updating) and related react hook with caching in local storage


metadata-models/src/main/pegasus/com/linkedin/identity/CorpUserSettings.pdl
```diff
@@ -6,7 +6,8 @@ import com.linkedin.event.notification.settings.NotificationSettings
  * Settings that a user can customize through the datahub ui
  */
 @Aspect = {
-  "name": "corpUserSettings"
+  "name": "corpUserSettings",
+  "schemaVersion": 2
```
benjiaming, line 10: Do we really need version bump? Adding an optional field to a Pegasus record is backwards-compatible by design, and this is typically done without a schema version bump. Bumping `schemaVersion` triggers different schema compatibility semantics and can affect upgrade paths for users running older versions. If we need to force a migration, it needs to be documented in `docs/how/updating-datahub.md`. I would probably remove it unless you have a good reason for it.

datahub-web-react/src/app/sharedV2/hooks/useIsI18nEnabled.ts
```diff
@@ -0,0 +1,5 @@
+import { useFeatureFlag } from '@app/sharedV2/hooks/useFeatureFlag';
```
benjiaming, line 1: v1 probably shouldn't import from v2. Should we move this to v2?

benjiaming (approved):
The two comments should be resolved but not necessarily block this PR.

metadata-models/src/main/pegasus/com/linkedin/identity/CorpUserSettings.pdl
v-tarasevich-blitz-brain: This version bump was added automatically via git commit hooks. Also I've seen that a few other schemes have `schemaVersion` updated over the past few days (I guess it's because that hook).

PR #14415, opened 2025-08-11
feat(ingestion): lineage for stored procedures across SQL databases with temp table filtering

acrylJonny:
*Summary*

Adds *stored procedure lineage extraction* across 6 major SQL databases: *MySQL, MariaDB, PostgreSQL, Oracle, Snowflake, and SQL Server*. This enables DataHub to capture column-level lineage from stored procedure logic, including sophisticated *temporary table collapsing* to provide clean, actionable lineage graphs.

*Key Features*

*🔍 *Multi-Database Support**
- *MySQL & MariaDB*: Extracts procedures from `INFORMATION_SCHEMA.ROUTINES` with full lineage parsing
- *PostgreSQL*: Parses `pg_proc` for function/procedure definitions
- *Oracle*: Supports `ALL_PROCEDURES` and `ALL_SOURCE` views
- *Snowflake*: Integrates with `INFORMATION_SCHEMA.PROCEDURES` and access history
- *SQL Server*: Reads from `sys.procedures` and `sys.sql_modules`

*🧹 *Intelligent Temp Table Filtering**
Automatically detects and filters temporary tables from lineage graphs:
- *MySQL/MariaDB*: `CREATE TEMPORARY TABLE` patterns
- *PostgreSQL*: `CREATE TEMP TABLE` patterns  
- *Oracle*: `CREATE GLOBAL TEMPORARY` patterns
- *Snowflake*: Configurable regex patterns for ETL tool temp tables (Fivetran, DBT, etc.)

*Example*: Procedure with `A → temp_table → B` now produces clean lineage `A → B`

*📊 *Column-Level Lineage**
Extracts fine-grained column transformations:
```sql
CREATE PROCEDURE calculate_priority_customers()
BEGIN
  CREATE TEMPORARY TABLE temp_priority AS
    SELECT customer_id, total_spent FROM orders;
  
  INSERT INTO priority_customers (id, value)
    SELECT customer_id, total_spent FROM temp_priority;
END

## Checklist
Thank you for contributing to DataHub!

Before you submit your PR, please go through the checklist below:

- [x] The PR conforms to DataHub's [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [Commit Message Format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#commit-message-format))
- [x] Links to related issues (if applicable)
- [x] Tests for the changes have been added/updated (if applicable)
[...]

PR #17461, opened 2026-05-15
test(ingest/mariadb): add integration test suite

acrylJonny:
## Summary

Adds an integration test suite for the existing `MariaDBSource` ingestion connector. MariaDB currently has no integration tests in `metadata-ingestion/tests/integration/`; this PR adds:

- A docker-compose harness using `mariadb:10.6` exposed on port `53300:3306`.
- A `setup.sql` that creates raw/processed tables plus three stored procedures (including one that `CALL`s the others) so the golden file exercises both schema metadata and stored-procedure ingestion / lineage.
- `mariadb_to_file.yml` recipe with profiling enabled and `procedure_pattern` filtering.
- `test_mariadb.py` with two integration tests:
  - `test_mariadb_ingest_no_db` — full ingestion → file → golden comparison.
  - `test_mariadb_test_connection` — verifies `TestConnectionReport` returns the expected success/failure for valid and invalid configs.
- The recorded `mariadb_mces_golden.json`.

`MariaDBSource` (`metadata-ingestion/src/datahub/ingestion/source/sql/mariadb.py`) is unchanged and continues to inherit from `MySQLSource` with `MySQLConfig`; this PR is test-only.

## Depends on #17466

Running this test against current `master` first surfaced a latent bug in the shared stored-procedure flow-name construction for two-tier sources: it emitted

```
urn:li:dataFlow:(mariadb,test_db.test_db.stored_procedures,PROD)
```

(duplicated database). The bug fix lives in #17466 ("fix(ingest/sql): don't duplicate database in two-tier stored-procedure flow name"). The golden file in this PR uses the correct, single-segment format (`test_db.stored_procedures`); please merge #17466 first so this golden matches the runtime output.

## Context

One of several follow-up PRs splitting up the now-stale #14415 into focused, reviewable changes. The MariaDB integration test scaffolding is the largest piece of net-new content from that branch that doesn't already exist in `master`.

## Checklist


acrylJonny:
Pushed a follow-up commit (cd7dbebcee) implementing procedure-to-procedure (`CALL`/`EXEC`) lineage extraction. This was the last remaining gap for two-tier procedure lineage.

**What changed**

- `stored_procedures/lineage.py`: parses `CALL` (MySQL/MariaDB/PostgreSQL/Oracle/Snowflake — sqlglot falls back to `Command` literal, regex-extracted) and `EXEC`/`EXECUTE` (TSQL — structured `Execute` node) and resolves each to a DataJob URN using the same composition rules as `BaseProcedure.to_urn` / `get_procedure_flow_name`.
- `stored_procedures/base.py`: `generate_procedure_lineage` now deduplicates when merging `additional_input_jobs` (Oracle's `ALL_DEPENDENCIES` path) with the body-parsed `inputDatajobs`, so the two paths can coexist.
- `tests/unit/sql_parsing/test_procedure_call_lineage.py`: 12 unit tests covering MySQL/MariaDB/Postgres `CALL`, TSQL `EXEC`/`EXECUTE`, bracketed identifiers, qualified vs unqualified, two-tier vs three-tier defaults, dedup, combined DML+CALL, and a no-false-positives case.
- `tests/integration/mariadb/mariadb_mces_golden.json` regenerated: `refresh_customer_analytics` now emits `inputDatajobs` pointing at `process_customer_data` and `generate_order_summaries`, demonstrating the feature end-to-end on the existing fixture.

**Known limitation** (documented in the URN-builder docstring): the called procedure's argument signature isn't available at the call site, so emitted URNs never carry the `_<hash>` suffix that `BaseProcedure.get_procedure_identifier` adds for procedures with arguments. Two-tier sources never hash, so this only matters for three-tier sources that populate `argument_signature` (Oracle, MSSQL) and have overloaded procedures.

The Oracle source already has a separate, more authoritative path that reads procedure dependencies from `ALL_DEPENDENCIES` and feeds them in via `additional_input_jobs` — that keeps working unchanged (and now de-dupes against the body-parsed entries if both paths happen to find the same call).

PR #14274, opened 2025-07-30
feat(mysql): add support for stored procedures in mysql ingestion

EmmetAVS:
Golden files for mysql integration have been updated to account for the change and the added support for stored procedures.

The key change here is adding support for stored procedures directly in SQLAlchemySource, it does not interferere with any existing classes that inherit from SQLAlchemy source and if a class chooses to add support for stored procedures all they need to do is override get_procedures_for_schema and return a list of BaseProcedures which is handled by the existing code within SQLAlchemySource. I had originally made a [pull request](https://github.com/datahub-project/datahub/pull/14102) that added support directly in PostgresSource but I got rid of that and moved it over to SQLAlchemy (except for get_procedures_for_schema) to account for the change.

PR #17462, opened 2026-05-15
test(mysql): add integration test for stored procedure ingestion

acrylJonny:
## Summary

MySQL stored-procedure support shipped in #14274 (merged 2025-08-11) but no integration test was added at the time. This PR fills in that gap.

- `setup/setup.sql` is extended with a `customer_stats` output table, an `order_date` column on `orders`, and two procedures:
  - `calculate_customer_stats` — table-level lineage from `customers`/`orders` to `customer_stats`.
  - `analyze_customer_priority(min_priority)` — exercises the temp-table filter via `CREATE TEMPORARY TABLE temp_priority_customers ...`.
- A new recipe `mysql_stored_procedures.yml` opts into stored-procedure ingestion with `procedure_pattern` filtering.
- Existing recipes now set `include_stored_procedures: false` so their goldens remain procedure-free; only the new recipe records procedure aspects.
- New golden file `mysql_mces_with_stored_procs_golden.json` records the expected output for the new recipe.
- Existing goldens are regenerated to reflect the additions to the shared `northwind` setup (new `customer_stats` table, `order_date` column on `orders`).
- `test_mysql.py` adds the new parametrization plus a small `ignore_paths` list for nondeterministic profiling sample values and DDL timestamps that vary across docker runs.

## Depends on #17466

MySQL is a two-tier source (`MySQLConfig(MySQLConnectionConfig, TwoTierSQLAlchemyConfig)`), so it hits the same flow-name duplication bug fixed in #17466. The committed `mysql_mces_with_stored_procs_golden.json` uses the corrected format (`northwind.stored_procedures`, single segment); without #17466 the runtime would produce `northwind.northwind.stored_procedures` and this golden would not match.

## Draft note

Marking as draft because the goldens were originally generated from a now-stale branch. They should match upstream's current stored-procedure output format (the in-tree `mysql.py` flow that landed in #14274 and the shared `generate_procedure_workunits` helpers from #16067) once #17466 lands, but a fresh CI golden run is the easiest way to confirm and refresh if anything has drifted. Happy to regenerate locally or via CI before un-drafting.

## Context

One of several focused follow-up PRs splitting up the now-stale #14415; the rest of that PR's content is either already in upstream or covered by separate PRs (MariaDB integration tests, MSSQL temp-table filtering, Oracle perf flag, two-tier flow-name fix).

## Checklist

### DIFF
--- a/metadata-ingestion/tests/integration/mysql/mysql_mces_no_db_golden.json
+++ b/metadata-ingestion/tests/integration/mysql/mysql_mces_no_db_golden.json
@@ -16,7 +16,7 @@
         }
     },
     "systemMetadata": {
-        "lastObserved": 1586847600000,
+        "lastObserved": 1615443388097,
         "runId": "mysql-test",
         "lastRunId": "no-run-id-provided"
     }
@@ -32,7 +32,7 @@
         }
     },
     "systemMetadata": {
-        "lastObserved": 1586847600000,
+        "lastObserved": 1615443388097,
         "runId": "mysql-test",
         "lastRunId": "no-run-id-provided"
     }
@@ -48,7 +48,7 @@
         }
     },
     "systemMetadata": {
-        "lastObserved": 1586847600000,
+        "lastObserved": 1615443388097,
         "runId": "mysql-test",
         "lastRunId": "no-run-id-provided"
     }
@@ -66,7 +66,7 @@
         }
     },
     "systemMetadata": {
-        "lastObserved": 1586847600000,
+        "lastObserved": 1615443388097,
         "runId": "mysql-test",
         "lastRunId": "no-run-id-provided"
     }
@@ -82,7 +82,7 @@
         }
     },
     "systemMetadata": {
-        "lastObserved": 1586847600000,
+        "lastObserved": 1615443388097,
         "runId": "mysql-test",
         "lastRunId": "no-run-id-provided"
     }
@@ -98,7 +98,7 @@
         }
     },
     "systemMetadata": {
-        "lastObserved": 1586847600000,
+        "lastObserved": 1615443388097,
         "runId": "mysql-test",
         "lastRunId": "no-run-id-provided"
     }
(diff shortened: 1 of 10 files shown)


acrylJonny:
Closing as redundant. Quick summary of why and where the coverage now lives:

**Why the CI was failing**

The `ci (mysql, tests/integration/mysql/, 3.11)` job fails for three independent reasons:

1. **URN duplication** – the committed golden uses `urn:li:dataFlow:(mysql,northwind.stored_procedures,...)` (the corrected, post-#17466 format), but master still emits the duplicated `northwind.northwind.stored_procedures`. This was an acknowledged dependency on #17466 in the PR description.
2. **`customProperties` populated in the golden** – `definer` / `security_type` / `sql_data_access` appear on each procedure in the golden, but `MySQLSource.get_procedures_for_schema` in master selects only `name, definition, language` and never sets `extra_properties`, so the runtime always produces `{}`. The golden was generated against a local working tree with additional extraction code that wasn't part of this PR.
3. **Empty `description: ""` keys** – master hardcodes `comment=None`, so the runtime omits the field entirely.

**Why it's redundant**

- #17466 already regenerates the existing MySQL goldens to the fixed URN format and adds the `language` fallback so native SQL procedures emit lineage. Its updated goldens are consistent with what master actually produces (no `definer` / `security_type` / `sql_data_access`).
- #17461 (MariaDB integration test) exercises the same `TwoTierSQLAlchemySource` / `MySQLSource` code path against MySQL-compatible SQL, including multi-statement procedure bodies and procedure-to-procedure (`CALL`) lineage that this PR also targeted.
- The temp-table filter scenario in `analyze_customer_priority` is the only unique addition here. That logic is already covered at the unit level in `tests/unit/sql_parsing/test_stored_procedure_lineage_integration.py` and the temp-table aggregator goldens (`test_multistep_temp_table.json`, `test_temp_table.json`, etc.), so an extra integration-level check doesn't materially raise the bar.

If we later want to add `DEFINER` / `SECURITY_TYPE` / `SQL_DATA_ACCESS` extraction and procedure comments to the MySQL source, that's a follow-up PR (source + test) on top of #17466 — independent of this one.

acrylJonny:
Closing per the analysis above — coverage is provided by #17466 (golden + URN fix + language fallback) and #17461 (multi-statement / procedure-to-procedure lineage on the same two-tier code path).

PR #17463, opened 2026-05-15
fix(ingest/mssql): resolve 2-part-name lineage references against discovered_datasets

acrylJonny:
## Summary

MSSQL stored-procedure bodies frequently reference tables as `schema.table` even though discovery records them fully qualified as `db.schema.table`. Previously, `is_discovered_table` treated any name with fewer than `MSSQL_QUALIFIED_NAME_PARTS` segments as undiscovered, which caused the temp-table filter (`is_temp_table=lambda name: not self.is_discovered_table(name)`) to drop real lineage edges from procedures that use 2-part references.

This adds a lookup pass for 2-part names: when `schema.table` matches the trailing two segments of a fully-qualified entry in `discovered_datasets`, treat the reference as a real table. 1-part names (and 2-part names that don't match anything in `discovered_datasets`) keep the previous undiscovered behaviour.

Adds unit-test coverage for the match and no-match cases on top of the existing `is_discovered_table` test.

## Context

This is one of several focused follow-up PRs splitting up the now-stale #14415; this is the targeted MSSQL filtering fix that originally landed there.

## Checklist


metadata-ingestion/src/datahub/ingestion/source/sql/mssql/source.py
```diff
@@ -1566,6 +1566,28 @@ def is_discovered_table(self, name: str) -> bool:
...
+            if len(parts) == 2:
+                standardized_parts = standardized_name.split(".")
+                schema_name_std = standardized_parts[0]
+                table_name_std = standardized_parts[1]
+
+                for discovered_name in self.discovered_datasets:
+                    discovered_parts = discovered_name.split(".")
+                    if (
+                        len(discovered_parts) >= 3
+                        and discovered_parts[-2] == schema_name_std
```
treff7es: 🟡 **Cross-database ambiguity — the branch's real contract, currently undocumented & untested.**

The match compares only the last two segments (schema + table) and discards the DB, returning on the first hit. MSSQL runs routinely span multiple databases and `dbo.<table>` is ubiquitous, so a 2-part `dbo.orders` resolves `True` against *any* `*.dbo.orders`. That's acceptable for a keep/drop gate, but it should be an intentional, pinned contract:

- Add a one-line comment noting the cross-DB collision is intentionally ignored (first-match-wins).
- Pin it with a test:
```python
source.discovered_datasets = {"db_a.dbo.regular_table", "db_b.dbo.regular_table"}
assert source.is_discovered_table("dbo.regular_table") is True
```
- Follow-up (separate PR): if the procedure's owning DB is known at call time, qualifying the 2-part name with it before matching would make this precise rather than heuristic.

metadata-ingestion/src/datahub/ingestion/source/sql/mssql/source.py
```diff
@@ -1566,6 +1566,28 @@ def is_discovered_table(self, name: str) -> bool:
...
+                        and discovered_parts[-1] == table_name_std
+                    ):
+                        result = True
+                        self._discovered_table_cache[name] = result
+                        return result
```
treff7es: ℹ️ **Simplification:** this whole match reduces to a single `endswith` against the fully-qualified name. `standardized_name` is guaranteed to be `"schema.table"` here (`len(parts) == 2`, and `standardize_identifier_case` only lowercases), so the leading `.` gives you the segment boundary *and* the "must be db-qualified" guarantee for free — letting you drop `standardized_parts`, the two temp vars, `len(discovered_parts) >= 3`, and the `[-2]`/`[-1]` indexing. Behaviorally identical to the current comparison.

```suggestion
                for discovered_name in self.discovered_datasets:
                    # The leading "." enforces a segment boundary, so a
                    # reference to schema "dbo" won't match discovered
                    # schema "xdbo"; it also requires a db-qualified entry.
                    if discovered_name.endswith(f".{standardized_name}"):
                        result = True
                        self._discovered_table_cache[name] = result
                        return result
```

(Still an O(n) scan, but cached per-name in `_discovered_table_cache`, so it runs at most once per distinct 2-part reference — fine as-is. A precomputed `{(schema, table)}` suffix set would make it O(1) if you ever want it.)

metadata-ingestion/src/datahub/ingestion/source/sql/mssql/source.py
```diff
@@ -1566,6 +1566,28 @@ def is_discovered_table(self, name: str) -> bool:
                     self._discovered_table_cache[name] = result
                     return result
 
+            # For 2-part names (schema.table), try to qualify against
+            # `discovered_datasets` by comparing the last two segments. TSQL
+            # procedure bodies often reference tables as `schema.table` even
+            # though discovery records them fully qualified as
+            # `db.schema.table`; without this we mark every such reference as
+            # undiscovered and the temp-table filter drops real lineage.
+            if len(parts) == 2:
```
treff7es, line 1575: ℹ️ **Pattern-filtering asymmetry (worth a comment).** The 3-part branch above applies `database_pattern`/`schema_pattern`/`table_pattern` before accepting a match; this 2-part branch doesn't. Almost certainly fine — entries in `discovered_datasets` already passed those patterns at discovery time — but a one-line note here prevents a future reader from reading it as an oversight.

metadata-ingestion/src/datahub/ingestion/source/sql/mssql/source.py
```diff
@@ -1566,6 +1566,28 @@ def is_discovered_table(self, name: str) -> bool:
...
+
+                for discovered_name in self.discovered_datasets:
+                    discovered_parts = discovered_name.split(".")
+                    if (
+                        len(discovered_parts) >= 3
+                        and discovered_parts[-2] == schema_name_std
+                        and discovered_parts[-1] == table_name_std
+                    ):
+                        result = True
+                        self._discovered_table_cache[name] = result
```
treff7es, line 1581: ℹ️ **Pre-existing DRY (out of scope for this PR).** The `result = ...; self._discovered_table_cache[name] = result; return result` epilogue is repeated ~8× across this method. A `_compute_*` helper wrapped by a single cache-and-return would remove the repetition and the risk of a future branch forgetting to populate the cache. The new block just follows the existing pattern — flag only if you want a follow-up cleanup.

metadata-ingestion/tests/unit/test_mssql.py
```diff
@@ -69,6 +69,21 @@ def test_is_discovered_table(mssql_source):
     assert mssql_source.is_discovered_table("invalid_table_name") is False
 
 
+def test_is_discovered_table_two_part_name_matches_discovered(mssql_source):
+    """A 2-part name (schema.table) should resolve to a discovered table when
+    the schema/table pair matches a fully-qualified entry in
+    `discovered_datasets`. Procedure bodies frequently reference tables this
+    way (e.g. ``UPDATE dbo.regular_table FROM ...``) and without this match
+    the temp-table filter would drop real lineage."""
+    assert mssql_source.is_discovered_table("dbo.regular_table") is True
```
treff7es, line 75: 🟡 **Two cheap test gaps worth closing (~8 lines).**

1. **Cross-DB ambiguity** (the branch's riskiest behavior) — see the inline note on the match in `source.py`. Add a two-DB fixture asserting the 2-part name still resolves `True`, to lock the contract.
2. **Temp-precedence for 2-part names** — the temp filters run before this branch; a future reorder would let 2-part temp refs leak into lineage. Guard it:
```python
assert mssql_source.is_discovered_table("dbo.temp_regular") is False    # ^temp_ pattern
assert mssql_source.is_discovered_table("dbo.#regular_table") is False   # # temp table
```

Both new tests are otherwise good — behavioral, well-named, clear docstrings, reuse the shared fixture.

treff7es (approved):

metadata-ingestion/src/datahub/ingestion/source/sql/mssql/source.py
acrylJonny: Documented and pinned in ccdde75. Added a comment on the 2-part branch noting the cross-DB collision is intentional (first-match-wins, a keep/drop gate rather than a precise resolver), and a `test_is_discovered_table_two_part_name_cross_database` test asserting `dbo.regular_table` still resolves `True` against both `db_a.dbo.regular_table` and `db_b.dbo.regular_table`. Agreed the DB-qualified precise match is a good follow-up.

metadata-ingestion/src/datahub/ingestion/source/sql/mssql/source.py
acrylJonny: Done in ccdde75 — collapsed to `if discovered_name.endswith(f".{standardized_name}")`, dropping `standardized_parts`, the two temp vars, the `len(discovered_parts) >= 3` check, and the `[-2]`/`[-1]` indexing. Kept a short comment noting the leading `.` gives both the segment boundary and the db-qualified guarantee.

metadata-ingestion/src/datahub/ingestion/source/sql/mssql/source.py
acrylJonny: Added a one-line note in ccdde75: no pattern filtering is needed on this branch because every entry in `discovered_datasets` already passed the database/schema/table patterns at discovery time.

metadata-ingestion/src/datahub/ingestion/source/sql/mssql/source.py
acrylJonny: Agreed this is pre-existing and out of scope here — the new block just follows the established cache-and-return pattern. I'll leave it for a separate cleanup PR that introduces a `_compute_*` helper wrapped by a single cache-and-return.

metadata-ingestion/tests/unit/test_mssql.py
acrylJonny: Both gaps closed in ccdde75:

1. **Cross-DB** — `test_is_discovered_table_two_part_name_cross_database` locks first-match-wins against a two-DB fixture.
2. **Temp precedence** — `test_is_discovered_table_two_part_name_temp_precedence` puts both temp variants (`temp_regular` and `#regular_table`) into `discovered_datasets` so they *would* resolve `True` via the endswith match if the temp filters didn't run first, then asserts both stay `False`. (Note: the shared fixture's `^temp_` is anchored and never matches a schema-qualified name, so the test sets a `.*\.temp_.*` pattern to make the precedence assertion meaningful.)

PR #17465, opened 2026-05-15
feat(ingestion): make Great Expectations profiler optional and default to SQLAlchemy

treff7es:
## Summary

Makes Great Expectations (GE) an **opt-in dependency** for DataHub ingestion. SQL connector extras (`[snowflake]`, `[bigquery]`, `[postgres]`, etc.) no longer install `acryl-great-expectations` by default. Users that want the GE profiler must install the new `[profiling-ge]` extra.

Also flips the default `profiling.method` from `ge` to `sqlalchemy` — the SQLAlchemy profiler has feature parity with the GE one and is now the default everywhere.

### Why

The pinned `acryl-great-expectations==0.15.50.1` fork is a long-term maintenance burden — pydantic 2 broke upstream GE, and upgrading further is a large effort because the GE API changed significantly. The SQLAlchemy profiler already coexisted in the codebase with parity claims; this PR finishes the migration by:

1. Making the SQLAlchemy profiler the default (no GE dependency in the import graph).
2. Allowing builds/images to ship without GE entirely.
3. Keeping the GE profiler available behind an explicit `[profiling-ge]` extra for users that need it.

## What changes for users

**Most users — no action needed.** Recipes with `profiling.enabled: true` (no `method:` specified) will now use the SQLAlchemy profiler automatically. Statistics emitted are equivalent for all dataset/column-level metrics (`null_count`, `distinct_count`, min/max, mean/median/stddev, quantiles, sample_values).

**Users that want the GE profiler:**

```bash
pip install 'acryl-datahub[profiling-ge]'
```

```yaml
source:
  config:
    profiling:
      enabled: true
      method: ge
```

If `method: ge` is set without `[profiling-ge]` installed, ingestion fails with a clear `ConfigurationError` pointing at the fix.

**Known difference:** the SQLAlchemy profiler uses 10 equal-width histogram buckets vs GE's adaptive partitioning (only affects `include_field_histogram`, which is off by default).

## Test plan

- [x] New `tests/unit/test_no_ge_dependency.py` parametrizes over 11 modules (3 profiler + 8 source) — uses a subprocess + meta_path blocker to assert each imports without `great_expectations`
- [x] New regression test `test_sql_common_lazy_imports_are_ge_free` pins the `SQLAlchemySource.loop_profiler_requests` contract (caught a critical bug during final review)
- [x] New `test_ge_method_raises_helpful_error_when_ge_missing` asserts the `ConfigurationError` flow emits both fix hints
- [x] `test_profiling_method_defaults_to_sqlalchemy` (in `test_ge_profiling_config.py` and `test_adapters.py`) pins the new default
- [x] New CI step in `.github/workflows/metadata-ingestion.yml` installs `[postgres]` in a fresh venv and asserts GE is absent (and that `[postgres,profiling-ge]` does pull GE)
- [x] Manual fresh-venv verification across `[snowflake]` and `[postgres]` confirmed end-to-end:
  - `acryl-datahub[snowflake]` alone → GE absent
  - `acryl-datahub[snowflake,profiling-ge]` → GE installed (v0.15.50.1)
  - SQLAlchemy profiler + source modules import cleanly under meta_path GE block
- [x] All 166 tests in `test_no_ge_dependency.py` + `test_ge_profiling_config.py` + `sqlalchemy_profiler/` pass

## Implementation notes

- Top-level `from datahub.ingestion.source.ge_data_profiler import DatahubGEProfiler` removed from `sql_generic_profiler.py`, `snowflake/snowflake_profiler.py`, `bigquery_v2/profiler.py`, `sql_common.py`, and `vertica.py` — moved inside `TYPE_CHECKING` or wrapped in `try/except ImportError -> ConfigurationError`
- `ProfilerRequest` dataclass extracted to a new GE-free module at `src/datahub/ingestion/source/profiling/profiler_request.py` (the legacy `GEProfilerRequest` name is preserved as a back-compat alias)
- Shared error message centralized in `src/datahub/ingestion/source/profiling/ge_profiler_loader.py`
- `setup.py`: removed `*great_expectations_lib`, `scipy`, `traitlets`, `IPython` from `sql_common`; created a new `profiling_ge` extra; `greenlet<4.0.0` correctly retained in `sqlalchemy_lib` (used by `sqlalchemy_query_combiner`)
- `pyproject.toml`/`uv.lock` regenerated via `./gradlew :metadata-ingestion:updateLockFile`

## Docs

- `docs/how/updating-datahub.md` — breaking-change entry (PR number to fill in)
- `metadata-ingestion/docs/dev_guides/sql_profiles.md` — replaced the "Rollout Plan / Phase 1/2/3" section with a description of the current state
- `metadata-ingestion/docs/sources/databricks/unity-catalog_recipe.yml` + `_pre.md` — updated to reflect the new default and the install requirement for the GE option

## Breaking change

Existing recipes that explicitly relied on `method: ge` AND installed only `acryl-datahub[snowflake]` (etc.) will now fail at ingestion start with a `ConfigurationError`. The error message tells them exactly how to fix it. Recipes with `method:` unset will silently switch to the SQLAlchemy profiler (which has parity).


metadata-ingestion/src/datahub/ingestion/source/profiling/common.py
```diff
@@ -50,3 +54,36 @@ def convert_to_cardinality(
...
+
+GE_PROFILER_MISSING_MESSAGE = (
+    "The Great Expectations profiler is not installed. Either install "
+    "the optional dependency with `pip install 'acryl-datahub[profiling-ge]'`, "
+    "or switch to the SQLAlchemy profiler by setting "
+    "`profiling.method: sqlalchemy` in your recipe."
+)
+
+
+def create_datahub_ge_profiler(**kwargs: Any) -> "DatahubGEProfiler":
```
sgomezvillamor, line 75: 🏅

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/postgres.py
```diff
@@ -64,6 +67,49 @@ def get_median_expr(self, column: str) -> Optional[ColumnElement[Any]]:
...
+                ).label("percentile")
+                query = sa.select([percentile_expr]).select_from(table)
+                result = conn.execute(query).scalar()
+                results.append(float(result) if result is not None else None)
+            except SQLAlchemyError as e:
+                logger.warning(
+                    f"Failed to compute quantile {q} for {column}: "
+                    f"{type(e).__name__}: {str(e)}"
+                )
+                results.append(None)
```
sgomezvillamor, line 110: Double check whether silently swallowing exceptions and returning nulls is the right pattern here.

IIRC it's fine to throw and let upper layers catch and handle it to continue profiling other columns/tables.

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/redshift.py
```diff
@@ -67,6 +70,51 @@ def get_median_expr(self, column: str) -> Optional[ColumnElement[Any]]:
...
+            row = conn.execute(query).fetchone()
+            if row is None:
+                return [None] * len(quantiles)
+            return [float(v) if v is not None else None for v in row]
+        except SQLAlchemyError as e:
+            logger.warning(
+                f"Failed to compute Redshift quantiles for {column}: "
+                f"{type(e).__name__}: {str(e)}"
+            )
+            return [None] * len(quantiles)
```
sgomezvillamor: same comment about "silently swallowing exceptions"

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/snowflake.py
```diff
@@ -201,34 +201,39 @@ def get_column_quantiles(
...
+            row = conn.execute(query).fetchone()
+            if row is None:
+                return [None] * len(quantiles)
+            return [float(v) if v is not None else None for v in row]
+        except SQLAlchemyError as e:
+            logger.warning(
+                f"Failed to compute Snowflake quantiles for {column}: "
+                f"{type(e).__name__}: {str(e)}"
+            )
+            return [None] * len(quantiles)
```
sgomezvillamor: same comment about "silently swallowing exceptions"

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/base_adapter.py
```diff
@@ -477,24 +483,41 @@ def get_column_median(self, table: sa.Table, column: str, conn: Connection) -> A
...
         """
         expr = self.get_median_expr(column)
-        if expr is None:
+        if expr is not None:
+            query = sa.select([expr]).select_from(table)
+            result = conn.execute(query).scalar()
+            # Return raw result to preserve database-native formatting (like GE does)
+            return result
+
+        # Python-side fallback (mirrors GE's get_column_median for dialects
```
sgomezvillamor, line 494: we may have a logger.debug to have visibility we go through the fallback

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/base_adapter.py
sgomezvillamor, line 1: Platform-specific code should live in the adapters
Some updates here aren't strictly platform-specific, more like fallbacks that apply to some platforms.
So it's a bit on the boundary of the criteria 😅 Fine enough!

sgomezvillamor (approved):

PR #17466, opened 2026-05-15
fix(ingest/sql): fix two-tier stored procedure ingestion to build correct urns (urn format + lineage)

acrylJonny:
## Summary

Two paired fixes that together make stored-procedure ingestion useful for two-tier SQL connectors (MySQL, MariaDB, Hive, Clickhouse, Teradata, and any other source extending `TwoTierSQLAlchemySource`):

1. **URN-format fix** (originally the entire scope of this PR): flow/job URNs no longer duplicate the database name.
2. **Language-fallback fix** (newly cherry-picked from #17461): MySQL/MariaDB native SQL procedures actually emit lineage. They previously had the right URNs but no `dataJobInputOutput` aspect because `information_schema.ROUTINES.EXTERNAL_LANGUAGE` is `NULL` for native procedures and the lineage path was gated on `language == 'SQL'`.

Each change is useful in isolation, but the URN fix without the language fix gives you correctly-named procedure entities with no lineage — i.e. only half the value. Bundling them so the release that fixes the URN shape also turns on lineage for the same sources.

### Bug

For two-tier SQL sources, `loop_stored_procedures` passes `schema == db_name` because the "schema layer" is the database itself. `_process_procedures` then built a `SchemaKey` from those two equal values and handed it to `get_procedure_flow_name`, which assembles the flow name as

```
{schema_key.database}.{schema_key.db_schema}.stored_procedures
```

That produced flow / job URNs of the form

```
urn:li:dataFlow:(mariadb,test_db.test_db.stored_procedures,PROD)
urn:li:dataJob:(urn:li:dataFlow:(mariadb,test_db.test_db.stored_procedures,PROD),generate_order_summaries)
```

duplicating the database name and creating a phantom schema layer in the browse UI that doesn't exist in the source.

### Fix

Move the database-key and schema-key construction in `_process_procedures` / `_process_procedure` behind two helpers (`_get_procedure_database_key`, `_get_procedure_schema_key`) and override the schema-key hook in `TwoTierSQLAlchemySource` to return `None`. With no `SchemaKey`, `get_procedure_flow_name` falls back to `{database}.stored_procedures`. Container assignment for the procedure DataJob itself already correctly falls back to the database container via the existing `container_key = schema_key or database_key` line — that branch was just never exercised before.

Three-tier sources (Postgres, Oracle, MSSQL, DB2, Snowflake via base, …) keep the existing `SchemaKey`-based behaviour.

### After

```
urn:li:dataFlow:(mariadb,test_db.stored_procedures,PROD)
urn:li:dataJob:(urn:li:dataFlow:(mariadb,test_db.stored_procedures,PROD),generate_order_summaries)
```

### Bug

`generate_procedure_lineage` checks `procedure.language == QueryLanguageClass.SQL` before parsing the procedure body. MySQL/MariaDB pulls `language` from `information_schema.ROUTINES.EXTERNAL_LANGUAGE`, which is `NULL` for natively-written SQL procedures (the common case) and is only populated for MySQL 8.0+ MLE procedures written in JavaScript or Java. So every native MySQL/MariaDB procedure was silently skipping lineage extraction even when the body contained parsable DML.

### Fix

Default `BaseProcedure.language` to `QueryLanguageClass.SQL` when the column is `NULL`/empty. Also replace the bare `"SQL"` string literal with the `QueryLanguageClass.SQL` constant at the comparison sites in `base.py` and the construction sites in `mssql/job_models.py` and `oracle.py`, so future renames of the enum surface as type errors instead of silent regressions.

## Tests

URN fix is covered by three focused tests in `test_sql_common.py`:

- `test_three_tier_source_emits_schema_key_for_procedures` — three-tier base returns a `SchemaKey` with `database`/`db_schema` populated (unchanged behaviour).
- `test_two_tier_source_omits_schema_key_for_procedures` — two-tier override returns `None`.
- `test_procedure_flow_name_two_tier_does_not_duplicate_database` — end-to-end protection: feeds the helpers' output through `get_procedure_flow_name` and asserts `"test_db.stored_procedures"`.

Language fix is exercised by the existing MySQL integration goldens (unchanged because the only fixture procedure, `AddCustomer`, uses `INSERT ... VALUES` and produces no parseable inputs either way) and by the MariaDB integration goldens on #17461, which include procedures with `INSERT ... SELECT` bodies that now produce `dataJobInputOutput` aspects on every ingest.

## Impact on existing goldens

The four committed MySQL goldens in this PR have been regenerated to drop the duplicated database segment. The MariaDB integration golden lives on the follow-up PR #17461 (which adds the fixture) and was regenerated there to reflect both fixes.

## Updating DataHub

`docs/how/updating-datahub.md` now has a combined entry for #17466 covering both the breaking URN change and the additive lineage change, with upgrade instructions for cleaning up the orphaned old URNs.

## Checklist


acrylJonny:
Expanded scope as discussed on #17461: cherry-picked the language-fallback fix from `feat/mariadb-integration-tests` (originally `75606fdd8e`, now `e3bf94e2e8` on this branch). The two fixes together make two-tier stored-procedure ingestion fully useful — entity URNs are correctly shaped *and* lineage actually populates — so it's a more complete release-note bullet.

The MariaDB golden regeneration stays on #17461 because the fixture itself lives there. On this branch the existing MySQL goldens were already regenerated for the URN fix and need no further change (verified: the only MySQL fixture procedure, `AddCustomer`, uses `INSERT ... VALUES` and produces no parseable inputs even with lineage extraction enabled, so the goldens are byte-identical with or without the language fix).

`docs/how/updating-datahub.md` extended to cover both aspects of the change.

Once #17466 merges, #17461 rebases cleanly to a pure test-suite PR (MariaDB fixture + procedure-call lineage feature + golden).

treff7es (approved):
lgtm

PR #11133, opened 2024-08-09
[DO NOT MERGE] - Sap hana improvements

acrylJonny:
## Checklist

- [ ] The PR conforms to DataHub's [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [Commit Message Format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#commit-message-format))
- [ ] Links to related issues (if applicable)
- [ ] Tests for the changes have been added/updated (if applicable)
- [ ] Docs related to the changes have been added/updated (if applicable). If a new feature has been added a Usage Guide has been added for the same.
- [ ] For any breaking change/potential downtime/deprecation/big changes an entry has been made in [Updating DataHub](https://github.com/datahub-project/datahub/blob/master/docs/how/updating-datahub.md)

## Summary by CodeRabbit

- **New Features**
	- Enhanced metadata ingestion support for SAP HANA databases, including new configuration options for improved flexibility.
	- Introduced methods to retrieve detailed information about views and their lineage, enhancing data visibility.
	- Implemented concurrent processing for better performance during metadata ingestion.

- **Bug Fixes**
	- Improved handling of various data types and structures specific to HANA to ensure accurate metadata capture.

PR #17472, opened 2026-05-15
feat(ingest/hana): production-ready SAP HANA connector — calc-view lineage, stored procedures, query usage

acrylJonny:
## Summary

Brings the SAP HANA connector to production readiness. The original 40-line `hana.py` shell becomes a focused multi-file package, and the source now extracts calculation-view lineage, stored procedures, and query usage on top of the existing `SQLAlchemySource` table/view/profiling path.

Everything new is **opt-in** via flags on `HanaConfig`. Default ingestion behaviour matches the previous shell.

### Calculation views (`include_calculation_views`, off by default)

- Discovers activated views from `_SYS_REPO.ACTIVE_OBJECT`.
- Emits 5 aspects per view: `status`, `schemaMetadata`, `datasetProperties`, `subType=SAP HANA Calculation View`, `viewProperties`.
- Parses the XML definition with a pure `SAPCalculationViewParser` and feeds downstream-column → upstream-(source, column) mappings into the `SqlParsingAggregator` via `add_known_query_lineage` — same path as Oracle / Redshift / Snowflake.
- Falls back to regex-based table-level lineage for `SqlScriptView` calc views whose lineage isn't expressible in the XML DAG.

### Stored procedures (`include_stored_procedures`, on by default)

- Reads `SYS.PROCEDURES` + `SYS.PROCEDURE_PARAMETERS`, yields `BaseProcedure` rows into the shared stored-procedure lineage path (provides procedure-to-table and procedure-to-procedure lineage; also closes the loop for calc views that `CALL` procedures).

### Query usage and operations (`include_query_usage` / `include_usage_stats`, off by default)

- Mines `_SYS_STATISTICS.HOST_SQL_PLAN_CACHE` (deduped by `(STATEMENT_HASH, LAST_EXECUTION_TIMESTAMP)`, system users / `SYS.*` / `_SYS_*` schemas filtered server-side).
- Feeds `ObservedQuery` rows through the aggregator, producing `DatasetUsageStatistics` and (with `include_operational_stats`) operation aspects.
- Mixes in `BaseUsageConfig` for the standard bucket / time-window / `top_n_queries` knobs; adds `usage_max_queries` to bound the per-run row count.
- Requires `MONITORING` role or `CATALOG READ` system privilege (clearly stated in the prereqs).

## Hardening and code quality

- **`defusedxml`** replaces stdlib `ElementTree` for XML parsing — protects against XXE and billion-laughs payloads served from `ACTIVE_OBJECT`.
- **`StrEnum` constants** in `hana/constants.py` replace string literals across parser dispatch (`HanaSourceType`, `CalcViewNodeType`, `HanaObjectType`).
- **Pydantic models** in `hana/models.py` replace anonymous dicts for parser outputs (`ColumnLineage`, `UpstreamColumnRef`, `ScriptViewDefinition`, `ScriptTableRef`, `HanaObservedQueryRow`, `CalcViewModel`).
- **`SAP_HANA_CALCULATION_VIEW`** added to common `DatasetSubTypes`.
- **Structured warnings**: every `except` block now uses `self.report.warning(title=, message=, context=, exc=)` — single call that records the structured warning, logs through the source logger, and preserves the exception traceback. Drops the paired `logger.warning` + `report_warning` anti-pattern.
- **No re-exports**: `hana/__init__.py` is empty; entry point points directly at `datahub.ingestion.source.sql.hana.hana:HanaSource`.
- **Bound parameters** everywhere for catalog queries (no SQL-injection surface).

## Package layout

| File | Purpose |
| --- | --- |
| `hana/hana.py` | `HanaSource(SQLAlchemySource)` — orchestrator. Constructs the aggregator with the right usage / operations hooks. |
| `hana/hana_config.py` | `HanaConfig` — extends `BaseUsageConfig`, adds calc-view / stored-procedure / usage flags and a deny-list `schema_pattern`. |
| `hana/hana_query.py` | All HANA catalog SQL as `Final[TextClause]` constants (parameterless) or `text(...)` builder functions (parameterised). |
| `hana/hana_schema.py` | Dataclasses for calc views and their columns (precise native-type formatting). |
| `hana/hana_data_dictionary.py` | Catalog queries running on a SQLAlchemy 2.0–style `Connection`. |
| `hana/hana_schema_gen.py` | `HanaCalculationViewExtractor` — workunit emission + lineage feed. |
| `hana/hana_utils.py` | `HanaIdentifierBuilder` (URN construction for calc views and all three upstream source types). |
| `hana/hana_calculation_view_parser.py` | Pure XML → column-lineage parser. |
| `hana/hana_script_lineage.py` | Regex-based table-reference extraction for SQLScript view bodies. |
| `hana/constants.py` | `StrEnum`s pinned to SAP HANA XML schema values. |
| `hana/models.py` | Pydantic models for parser / dictionary outputs. |

## Documentation

New canonical-style docs under `metadata-ingestion/docs/sources/hana/`:

- **`README.md`** — concept mapping (HANA → DataHub).
- **`hana_pre.md`** — deployment compatibility matrix (HANA on-prem / XS-classic vs HANA Cloud / HDI-only — calc-view discovery only works on the former), a copy-paste `GRANT` script, and a capability-to-grant table covering tables/views, calc views, stored procedures, and usage.
- **`hana_post.md`** — lineage internals (XML DAG resolution, SQLScript fallback, observed-query flow), usage computation details, and troubleshooting steps (insufficient privilege, zero rows, etc.).
- **`hana_recipe.yml`** — example with usage extraction commented out.

## Tests

37 focused unit tests covering:

- Config (URL building, schema-deny defaults including `SYS`/`_SYS_*` but allowing `_SYS_BIC`, flag semantics).
- `HanaCalcViewColumn.get_precise_native_type` for parameterised types.
- `HanaIdentifierBuilder` URN construction for calc views and all three upstream source types (`DATA_BASE_TABLE`, `TABLE_FUNCTION`, `CALCULATION_VIEW`), plus negative cases (unknown source type, empty source name).
- `SAPCalculationViewParser` against fixture XML: simple projection, union with two branches, TREE_BASED dimension via DataSource alias, formula referencing input column in the same node, column literally named `type` (real-world edge case that previously broke the discriminator), `SqlScriptView` non-walking, malformed XML, and an XML-bomb rejection that proves the `defusedxml` swap is wired.
- SQLScript table-reference extraction: FROM/JOIN, `DUMMY` skip, dedup, CALL-only empty.
- `HanaCalculationViewExtractor` end-to-end with a mocked engine: 5-aspect workunit shape, URN consistency, `customProperties` content, `NVARCHAR(100)` native-type spelling, lineage feed, engine disposal — separate test asserts the SQLScript fallback path also emits table lineage with empty column lineage.
- Stored-procedure extraction: `BaseProcedure` row mapping with definition / language / `default_schema`.
- Query usage: observed-query typing, empty-statement skip, structured warning on extraction failure, aggregator usage/operations wiring on and off.

Removed self-referential constant assertions and Pydantic-dump tests that were testing the framework rather than connector behaviour.

`./gradlew :metadata-ingestion:lint` (ruff + format + mypy) is clean.

## Supersedes

[acrylJonny/datahub#11133](https://github.com/datahub-project/datahub/pull/11133) (the original `[DO NOT MERGE]` exploration on a personal fork).

## Checklist

- [x] PR conforms to the Contributing Guideline (Conventional Commits format).
- [x] Tests added (37 unit tests; `./gradlew :metadata-ingestion:lint` clean; lineage coverage re-verified at 99.97% on the 2,885-view corpus).
- [x] Docs added — new canonical-style README + pre / post / recipe under `metadata-ingestion/docs/sources/hana/`.
- [x] Breaking changes — none. All new behaviour is opt-in via config flags; default ingestion matches the previous shell. Stored-procedure ingestion is on by default for parity with other SQL connectors (Snowflake, Oracle, MSSQL).


metadata-ingestion/src/datahub/ingestion/source/sql/hana/hana_config.py
```diff
@@ -0,0 +1,71 @@
...
+        description="Regex patterns matched against `<schema>.<procedure_name>` to filter stored procedures.",
+    )
+    include_query_usage: bool = Field(
+        default=False,
+        description="If true, mine query history from `_SYS_STATISTICS.HOST_SQL_PLAN_CACHE`. Requires the `MONITORING` role or `CATALOG READ` system privilege.",
+    )
+    include_usage_stats: bool = Field(
+        default=False,
+        description="If true, emit `DatasetUsageStatistics` aspects. Requires `include_query_usage`.",
+    )
```
treff7es, line 53: **🔴 C2: `include_usage_stats=True` silently accepts incoherent state.**

The description says it requires `include_query_usage`, but nothing enforces it. A user setting `include_usage_stats=True` (and leaving `include_query_usage=False` default) gets no usage stats and no diagnostic.

**Fix** — Pydantic v2 idiom:

```python
@model_validator(mode="after")
def _usage_stats_requires_query_usage(self) -> "HanaConfig":
    if self.include_usage_stats and not self.include_query_usage:
        raise ValueError(
            "`include_usage_stats=true` requires `include_query_usage=true`; "
            "usage statistics are computed from mined queries."
        )
    return self
```

Separately, this field's description could mention privilege requirements (`MONITORING` role or `CATALOG READ`) — it inherits the requirement from `include_query_usage` but users reading this field in isolation won't know.

metadata-ingestion/src/datahub/ingestion/source/sql/hana/hana_script_lineage.py
```diff
@@ -0,0 +1,47 @@
...
+            key = (schema, name)
+            if key in seen:
+                continue
+            seen.add(key)
+            refs.append(ScriptTableRef(schema_name=schema, name=name))
+    except Exception:
+        logger.warning(
+            "Failed to extract table refs from SQLScript body", exc_info=True
+        )
+        return []
```
treff7es: **🔴 C3: `except Exception` here is paranoia that hides regressions.**

`re.finditer` over a compiled pattern on a string has essentially no failure surface — the `if not sql: return []` guard at line 26 already handles the only realistic `TypeError`. The blanket `except Exception` would silently swallow future regressions in `ScriptTableRef` construction (e.g., if it adds Pydantic validation), the `frozenset` lookup, or the `seen` set logic — SQLScript lineage would silently disappear for affected views with no telemetry.

Also `logger.warning` only — no `self.report.warning(...)`, no context (the view name is lost in the log line).

**Fix:** drop the try/except entirely, OR narrow to `(TypeError,)` and accept a report (or callback) so failures are visible.

metadata-ingestion/src/datahub/ingestion/source/sql/hana/models.py
```diff
@@ -0,0 +1,101 @@
...
+    OutputMappingKind,
+)
+
+
+class DataSourceRef(BaseModel):
+    """A ``<DataSource>`` declared inside a calculation view."""
+
+    name: str = ""
+    path: str = ""
+    type: str = ""
```
treff7es: **🟡 H2: Empty-string-as-null antipattern.**

`name: str = ""`, `path: str = ""`, `type: str = ""` use empty strings as sentinels for "absent". Downstream code re-checks this with `if not source_name` (in `HanaIdentifierBuilder.upstream_urn_for_calc_view_source`), which is symptomatic — the type lies about whether the field is present.

Worse, `_collect_data_sources` in `hana_calculation_view_parser.py:217-238` constructs `DataSourceRef(type=...)` and then **mutates the BaseModel after construction** (`source.name = ...`, `source.path = ...`). This defeats Pydantic's validation contract — `model_validate` only runs once at construction.

**Fix:**
- Change to `Optional[str] = None`.
- Refactor `_collect_data_sources` to build a dict and call `DataSourceRef.model_validate(d)` once after the loop.
- Same applies to `OutputMapping` (line 37) where `_build_output_mapping` produces `OutputMapping(source_column="", source_node="", kind=kind)` for the missing-element case — return `None` and skip in the caller, or make those fields `Optional[str] = None`.

metadata-ingestion/src/datahub/ingestion/source/sql/hana/models.py
```diff
@@ -0,0 +1,101 @@
...
+    nodes: Dict[str, CalcViewNode] = Field(default_factory=dict)
+    scripts: Dict[str, str] = Field(default_factory=dict)
+
+
+class UpstreamColumnRef(BaseModel):
+    """A resolved upstream column from lineage tracing."""
+
+    column: str
+    source_name: str
+    source_type: str
```
treff7es: **🟡 H3 (related): `source_type: str` should be `HanaSourceType`.**

Callers already pass `HanaSourceType.DATA_BASE_TABLE` etc., so the `str` annotation here (and on `HanaIdentifierBuilder.upstream_urn_for_calc_view_source`'s `source_type` parameter) is artificially loose. A future enum addition (e.g. `HanaSourceType.HDI_FUNCTION`) would silently produce `None` URNs.

Tightening this to `HanaSourceType` (with `assert_never` on the URN builder's default branch) lets mypy surface enum drift at typecheck time.

metadata-ingestion/src/datahub/ingestion/source/sql/hana/hana_utils.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def calc_view_urn(self, calc_view: HanaCalculationView) -> str:
+        """URN for an activated calculation view (``_SYS_BIC.<package>.<name>``)."""
+        return self._dataset_urn(calc_view.qualified_identifier)
+
+    def upstream_urn_for_calc_view_source(
+        self,
+        *,
+        source_type: str,
+        source_name: str,
```
treff7es, line 43: **🟡 H3: `source_type: str` should be `HanaSourceType`.**

See the matching comment on `models.py:61`. The function dispatches on `HanaSourceType.DATA_BASE_TABLE` / `TABLE_FUNCTION` / `CALCULATION_VIEW` and falls through to `None` for anything else — a future HANA-side type addition would be silently dropped. Use the enum here and pair with `typing.assert_never` (or an explicit `else: raise`) so the typechecker catches enum drift.

metadata-ingestion/src/datahub/ingestion/source/sql/hana/hana_schema_gen.py
```diff
@@ -0,0 +1,284 @@
...
+    ColumnLineageInfo,
+    ColumnRef,
+    DownstreamColumnRef,
+)
+
+# Mapping from SAP HANA native type names (as they appear in
+# ``SYS.VIEW_COLUMNS.DATA_TYPE_NAME``) to DataHub schema type classes.
+# Unmapped types fall through to ``NullTypeClass`` rather than guessing string
+# so downstream consumers can distinguish "we don't know" from "actually text".
+_HANA_TYPE_MAP: dict = {
```
treff7es: **🟡 H5: `_HANA_TYPE_MAP: dict` is loosely typed and missing `FLOAT`.**

The annotation `dict` admits string values or non-class objects. Use a concrete type like `Dict[str, Type[object]]` or `Mapping[str, Callable[[], SchemaFieldTypeBase]]`.

Coverage gap: **`FLOAT(n)` is missing.** HANA's `FLOAT(n)` carries a bit-width (1–53) in `SYS.VIEW_COLUMNS.LENGTH`, so `HanaCalcViewColumn.get_precise_native_type` (in `hana_schema.py`) should also append `(length)` for `FLOAT`. Worth verifying `ST_POINT(srid)` / `ST_GEOMETRY(srid)` while you're there.

Refactoring suggestion to make the invariant explicit — replace the two literal tuples in `get_precise_native_type` with named frozensets:
```python
_NUMERIC_WITH_PRECISION_SCALE = frozenset({"DECIMAL", "NUMERIC", "SMALLDECIMAL"})
_LENGTH_BEARING_STRING = frozenset({
    "VARCHAR", "NVARCHAR", "ALPHANUM", "SHORTTEXT", "VARBINARY",
})
_PRECISION_ONLY = frozenset({"FLOAT"})
```

M8 (related, line 282): when `_HANA_TYPE_MAP` falls through to `NullTypeClass`, no report is generated. Adding a `LossyList[str]` of unmapped types to the report would help users notice when new HANA types arrive.

metadata-ingestion/src/datahub/ingestion/source/sql/hana/hana_calculation_view_parser.py
```diff
@@ -0,0 +1,336 @@
...
+                    INPUT_NODE_REF_SIGIL
+                )[-1]
+                node.union_branches.append(branch)
+                node.branches[branch] = _read_input_mappings(input_el)
+        else:
+            for input_el in child.iter(CalcViewXmlElement.INPUT):
+                branch = input_el.attrib[CalcViewXmlAttribute.NODE].split(
+                    INPUT_NODE_REF_SIGIL
+                )[-1]
+                node.branches[branch] = _read_input_mappings(input_el)
```
treff7es: **🟠 M2: UNION vs non-UNION branches differ by one line.**

The two `for input_el in child.iter(...)` blocks (lines 273-285) are identical except UNION appends to `union_branches`. Trivial unification:

```python
is_union = node_type == CalcViewNodeType.UNION
for input_el in child.iter(CalcViewXmlElement.INPUT):
    branch = input_el.attrib[CalcViewXmlAttribute.NODE].split(
        INPUT_NODE_REF_SIGIL
    )[-1]
    if is_union:
        node.union_branches.append(branch)
    node.branches[branch] = _read_input_mappings(input_el)
```

metadata-ingestion/src/datahub/ingestion/source/sql/hana/hana_calculation_view_parser.py
```diff
@@ -0,0 +1,336 @@
...
+                    edge = branch_cols.get(column)
+                    if edge is not None:
+                        sources.extend(
+                            self._traverse(model, edge, branch, visited.copy())
+                        )
+                formula_expr = node_info.formulas.get(column)
+                if formula_expr is not None:
+                    for ref_col in self._extract_columns_from_formula(formula_expr):
+                        sources.extend(
+                            self._find_all_sources(model, ref_col, node, visited.copy())
```
treff7es: **🟠 M3: `visited.copy()` x4 — push/pop would be O(1) per call.**

Four `visited.copy()` calls in `_find_all_sources` + `_traverse` (lines 150, 157, 163, 211) cost O(N) per recursion. The current semantics — path-based cycle detection that permits formula re-entry — can be preserved with try/finally enter/exit on a shared mutable set:

```python
visited.add(key)
try:
    # ... existing body, no .copy() ...
finally:
    visited.discard(key)
```

Same semantics, no copies, asymptotically faster on wide diamond DAGs. The docstring on lines 130-133 currently explains the *what* (cycles per path); please add a one-line *why* if you adopt this — the trade-off was deliberate to permit formula re-entry.

metadata-ingestion/src/datahub/ingestion/source/sql/hana/hana.py
```diff
@@ -0,0 +1,219 @@
...
+                    user=None,
+                    default_db=None,
+                    default_schema=row.schema_name,
+                )
+
+    def _populate_aggregator_from_queries(self) -> None:
+        if not self.config.include_query_usage:
+            return
+
+        engine = self._build_engine()
```
treff7es: **🟠 M7: `_build_engine()` is outside the `try` block.**

If `create_engine` raises (bad URL, missing driver, auth failure), the user gets a raw stack trace and the entire run aborts — including dropping the metadata-extraction work units already yielded earlier in `get_workunits_internal`. Moving `engine = self._build_engine()` inside the try and routing the failure through `self.report.warning(title="Query usage extraction unavailable", ...)` would let metadata ingestion complete cleanly even when the usage path can't connect.

Separately, anything thrown from `self.aggregator.add(observed)` inside the loop (line 193) also propagates. Per-record extraction failures should warn-and-continue per the standards.

metadata-ingestion/src/datahub/ingestion/source/sql/hana/hana.py
```diff
@@ -0,0 +1,219 @@
...
+        # roll-up gets the right time-bucketed counts.
+        self._populate_aggregator_from_queries()
+
+        if self.calc_view_extractor is None:
+            yield from super()._generate_aggregator_workunits()
+            return
+
+        with self.report.new_stage(LINEAGE_EXTRACTION):
+            yield from self.calc_view_extractor.get_workunits_internal()
+            yield from super()._generate_aggregator_workunits()
```
treff7es: **🟠 M5: Duplicated aggregator flush across two branches.**

`super()._generate_aggregator_workunits()` is yielded on both line 214 (early return path) and line 219 (LINEAGE_EXTRACTION path). A single linear flow avoids the duplication:

```python
with self.report.new_stage(METADATA_EXTRACTION):
    yield from super().get_workunits_internal()

self._populate_aggregator_from_queries()

if self.calc_view_extractor is not None:
    with self.report.new_stage(LINEAGE_EXTRACTION):
        yield from self.calc_view_extractor.get_workunits_internal()

yield from super()._generate_aggregator_workunits()
```

Minor preference question: should the final aggregator flush be inside `LINEAGE_EXTRACTION` (current behavior when calc views are enabled) or outside (this proposal)? Either is defensible — worth a one-line comment either way.

metadata-ingestion/src/datahub/ingestion/source/sql/hana/hana_data_dictionary.py
```diff
@@ -0,0 +1,195 @@
...
+                return_type=None,
+                # HANA exposes both ``SQLSCRIPT`` (the dialect we want
+                # sqlglot to ignore) and ``RUNTIME`` for L-language
+                # procedures; we expose them as-is and let the lineage
+                # parser short-circuit non-SQL languages.
+                language=str(mapping.get("LANGUAGE") or "SQLSCRIPT"),
+                extra_properties=None,
+                default_schema=schema,
+            )
+
```
treff7es, line 144: **ℹ️ M6: Repeated `try/execute/except/report.warning/iterate` pattern in all four methods.**

A small helper would save ~30 lines and make each method's intent more visible:

```python
def _execute_or_warn(
    self,
    query: TextClause,
    *,
    title: str,
    message: str,
    params: Optional[Dict[str, Any]] = None,
    context: Optional[str] = None,
) -> Iterable[Mapping]:
    try:
        result = self.connection.execute(query, params or {})
    except Exception as e:
        self.report.warning(title=title, message=message, context=context, exc=e)
        return
    for row in result:
        yield row._mapping
```

Trade-off: makes per-call exception handling implicit. With exactly 4 callers and uniform warning shape, the helper is a net win.

metadata-ingestion/src/datahub/ingestion/source/sql/hana/hana_config.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+class HanaConfig(BasicSQLAlchemyConfig, BaseUsageConfig):
+    scheme: HiddenFromDocs[str] = Field(
+        default="hana+hdbcli",
+        description="SQLAlchemy scheme. Defaults to the official SAP `hdbcli` driver.",
+    )
+    host_port: str = Field(
+        default="localhost:39041",
+        description="SAP HANA host and port, e.g. `myhost.example.com:30015`.",
```
treff7es: **ℹ️ S1: `host_port` default `"localhost:39041"` is unusual.**

Most other DataHub SQL connectors require host explicitly (e.g. Snowflake `account_id`). The default suits the docker-fixture but a production user copy-pasting a config without setting `host_port` would silently try `localhost`.

Trade-off: keeping the default makes `datahub init`-style scaffolding work without edits. Defer to your judgement.

Separately (S2): the field could mention TLS — HANA Cloud requires `?encrypt=true` in the SQLAlchemy URL, which isn't obvious from the description.

metadata-ingestion/src/datahub/ingestion/source/sql/hana/hana_schema.py
```diff
@@ -0,0 +1,61 @@
...
+
+    package_id: str
+    name: str
+    definition: str
+    columns: List[HanaCalcViewColumn] = field(default_factory=list)
+
+    @property
+    def runtime_view_name(self) -> str:
+        """Name HANA uses for the activated view in ``_SYS_BIC``."""
+        return f"{self.package_id}/{self.name}"
```
treff7es, line 64: **ℹ️ S4: One-line comment requested — `runtime_view_name` preserves case, `qualified_identifier` (line 59) lowercases.**

The asymmetry is intentional (`runtime_view_name` must match HANA tooling exactly, `qualified_identifier` is URN-safe so must be lowercased), but a future maintainer noticing one method `.lower()`s and the other doesn't could break things "fixing" the inconsistency. A one-line comment on `runtime_view_name` would prevent that.

metadata-ingestion/src/datahub/sql_parsing/sql_parsing_common.py
```diff
@@ -34,6 +34,14 @@ def get_dialect_str(platform: str) -> str:
         return "tsql"
     elif platform_lower == "athena":
         return "trino"
+    elif platform_lower == "hana":
+        # sqlglot does not ship a dedicated SAP HANA dialect. HANA SQL is
+        # closest to PostgreSQL among the available dialects (window
+        # functions, ANSI joins, double-quoted identifiers, schema-qualified
+        # names); using "postgres" lets the parser resolve table references
+        # in calculation-view SQL fragments without forcing every query to
```
treff7es, line 42: **ℹ️ S3: The `"hana" → "postgres"` mapping is documented but inherently fuzzy.**

HANA SQL diverges from PostgreSQL in subtle ways (CALL semantics, `NCLOB`, MERGE syntax, `WITH` parameters, type-cast literals). The comment is honest about the trade-off, and the 99.97% lineage coverage on the 2,885-view corpus is empirically reassuring.

Worth a follow-up issue tracking either (a) proper HANA dialect support upstream in sqlglot, or (b) documented known parse-failure modes — so future failure reports have a known-issues page to land against.

metadata-ingestion/src/datahub/ingestion/source/sql/hana/hana_config.py
acrylJonny: Added `@model_validator(mode="after")` `_usage_stats_requires_query_usage` that raises `ValueError`. Description now flags the inherited `MONITORING` / `CATALOG READ` requirement. Test: `test_include_usage_stats_without_query_usage_rejected`.

metadata-ingestion/src/datahub/ingestion/source/sql/hana/hana_script_lineage.py
acrylJonny: Dropped the `try/except` and the now-unused `logger` import.

metadata-ingestion/src/datahub/ingestion/source/sql/hana/models.py
acrylJonny: `DataSourceRef.name/path` and `OutputMapping.source_column/source_node` are now `Optional[str] = None`. `_collect_data_sources` builds locals first and calls `DataSourceRef.model_validate(...)` once after the loop — no post-construction mutation. `_build_output_mapping` returns `OutputMapping(kind=kind)` for the missing-element case and callers skip on None.

metadata-ingestion/src/datahub/ingestion/source/sql/hana/models.py
acrylJonny: `UpstreamColumnRef.source_type` and `DataSourceRef.type` are now `HanaSourceType`. `_collect_data_sources` skips unknown XML type strings at parse time so the URN builder only ever sees a known enum value.

metadata-ingestion/src/datahub/ingestion/source/sql/hana/hana_utils.py
acrylJonny: `source_type` param is now `HanaSourceType`, with `assert_never(source_type)` on the default branch so mypy catches enum drift.

metadata-ingestion/src/datahub/ingestion/source/sql/hana/hana_schema_gen.py
acrylJonny: Typed as `Mapping[str, Type[Union[...]]]`, added `FLOAT`. `get_precise_native_type` now switches on named frozensets `_NUMERIC_WITH_PRECISION_SCALE`, `_LENGTH_BEARING_STRING`, `_PRECISION_ONLY` (FLOAT appends length). Unmapped types are collected on the extractor and surfaced once at end-of-run via `self.report.info(...)`; test `test_calc_view_extractor_reports_unmapped_hana_types`.

metadata-ingestion/src/datahub/ingestion/source/sql/hana/hana_calculation_view_parser.py
acrylJonny: Unified behind `is_union = node_type == CalcViewNodeType.UNION`.

metadata-ingestion/src/datahub/ingestion/source/sql/hana/hana_calculation_view_parser.py
acrylJonny: Switched to push/pop: `visited.add(key)` on entry, `visited.discard(key)` in `finally`. All four `.copy()` calls gone. Docstring updated to note the push/pop semantics.

metadata-ingestion/src/datahub/ingestion/source/sql/hana/hana.py
acrylJonny: Moved `_build_engine()` inside the `try`. Top-level `except` routes engine/extraction failures through `self.report.warning(title="Query usage extraction unavailable", ...)` so metadata workunits already yielded complete cleanly. The per-record `aggregator.add(observed)` is also wrapped with warn-and-continue.

metadata-ingestion/src/datahub/ingestion/source/sql/hana/hana.py
acrylJonny: Restructured to a single linear flow: metadata stage → `_populate_aggregator_from_queries()` → optional `LINEAGE_EXTRACTION` → single final `super()._generate_aggregator_workunits()`. Comment notes the flush runs after both `add_known_query_lineage` and `add(ObservedQuery)`.

metadata-ingestion/src/datahub/ingestion/source/sql/hana/hana_data_dictionary.py
acrylJonny: Extracted `_execute_or_warn(query, *, title, message, params, context)` — all four catalog methods now go through it.

metadata-ingestion/src/datahub/ingestion/source/sql/hana/hana_config.py
acrylJonny: Kept the default for docker-fixture compatibility but added a TLS note (`encrypt=true` via `options.connect_args`) to the description for HANA Cloud users. Happy to flip to required if you prefer.

metadata-ingestion/src/datahub/ingestion/source/sql/hana/hana_schema.py
acrylJonny: Added a docstring note on `runtime_view_name` calling out the case-preserving asymmetry vs `qualified_identifier`.

metadata-ingestion/src/datahub/sql_parsing/sql_parsing_common.py
acrylJonny: Extended the comment with the known mis-parse classes (CALL, NCLOB, MERGE, SQLScript-only constructs) and noted that table-ref extraction in `hana_script_lineage.py` restricts itself to quoted schema-qualified identifiers to sidestep them. Happy to file a follow-up tracking proper HANA dialect support upstream in sqlglot if useful.

treff7es (approved):
lgtm

PR #17478, opened 2026-05-17
fix(structured-properties): handle hard-deleted definitions on upsert and delete

david-leifker:
## Summary
- Emit a companion `propertyDefinition` DELETE MCL before hard-deleting a structured property so assignment cleanup side effects still run.
- On upsert, drop assignments whose property definition was hard-deleted, skip no-op writes when values are unchanged, and guard against null aspect maps.
- Add `EntityClient.filterExistingUrns` for batch existence checks; apply UI source on create proposals.

## Test plan
- [ ] `./gradlew :datahub-graphql-core:test --tests '*StructuredProperties*'`
- [ ] `./gradlew :metadata-io:test --tests PropertyDefinitionDeleteSideEffectTest`
- [ ] Manual: hard-delete a structured property definition and verify asset assignments are removed
- [ ] Manual: upsert structured properties on an asset with orphaned assignments succeeds without validation errors


datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/structuredproperties/UpsertStructuredPropertiesResolver.java
```diff
@@ -113,6 +134,95 @@ public CompletableFuture<com.linkedin.datahub.graphql.generated.StructuredProper
         "get");
   }
 
+  /** Removes value assignments whose structured property entity no longer exists (hard-deleted). */
```
RyanHolstien: Is this something that should be done in a validator? Seems like fairly hefty logic to be done at the GraphQL resolver and these drops are all invisible to the user that sent in the request

metadata-io/src/main/java/com/linkedin/metadata/entity/EntityServiceImpl.java
```diff
@@ -2958,6 +2958,12 @@ RollbackResult deleteAspectWithoutMCL(
           collectMetrics(opContext.getMetricUtils().orElse(null), exceptions).toString());
     }
 
+    // Hard delete wipes all aspects in one shot; capture propertyDefinition before deleteUrn so
+    // PropertyDefinitionDeleteSideEffect can scroll ES and emit PATCH REMOVE MCPs (see
+    // docs/api/tutorials/structured-properties.md).
+    final RecordTemplate[] propertyDefinitionBeforeHardDelete = {null};
```
RyanHolstien: These should not be initialized with null in them, it kinda seems like they don't even need to be arrays?

metadata-io/src/main/java/com/linkedin/metadata/entity/EntityServiceImpl.java
```diff
@@ -3146,7 +3171,23 @@ RollbackResult deleteAspectWithoutMCL(
...
+            constructMCL(
+                null,
+                urnToEntityName(entityUrn),
+                entityUrn,
+                ChangeType.DELETE,
+                STRUCTURED_PROPERTY_DEFINITION_ASPECT_NAME,
+                auditStamp,
+                null,
+                null,
+                propertyDefinitionBeforeHardDelete[0],
```
RyanHolstien: Yeah instead of pulling hard coded element 0 for the arrays just use a single RecordTemplate object

metadata-service/restli-client/src/main/java/com/linkedin/entity/client/RestliEntityClient.java
```diff
@@ -971,6 +972,20 @@ public boolean exists(
...
   }
 
+  @Override
+  @Nonnull
+  public Set<Urn> filterExistingUrns(
+      @Nonnull OperationContext opContext, @Nonnull Collection<Urn> urns)
+      throws RemoteInvocationException {
+    Set<Urn> existing = new HashSet<>();
+    for (Urn urn : urns) {
+      if (exists(opContext, urn)) {
```
RyanHolstien: I know we don't really want to do more work on the RestliEntityClient, but this will fire off N network calls when there is a method that accepts a list downstream, if we can just have AI add the exists with list to the resource I think that would be preferable.

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/structuredproperties/UpsertStructuredPropertiesResolver.java
david-leifker: Really, I think I'll just add an option to the mutator to do this based on config. Drop + warn on hard deleted instead of fail.

RyanHolstien (approved):

PR #17483, opened 2026-05-18
fix(ingestion): test connection falls back to defaultCliVersion when version is unset (follow-up to #17471)

puneetagarwal-datahub:
> **Follow-up to #17471.** That PR fixed the manual-ingestion path; this PR applies the same fix to the test-connection path which #17471 missed.

## Summary

#17471 routed the manual-ingestion path through `IngestionUtils.resolveIngestionCliVersion`, so blank/whitespace `config.version` values cleanly fall back to `defaultCliVersion`. The matching test-connection resolver (`CreateTestConnectionRequestResolver`) was *not* updated in #17471 — it still omits `args.version` entirely when `input.version` is null, empty, or whitespace-only.

With `args.version` absent from the request, the executor falls back to the CLI version baked into its docker image at build time (`BUNDLED_CLI_VERSION`), ignoring the configured `defaultCliVersion`. Test connections therefore run on a different CLI than the actual ingestion would — hiding compatibility issues that surface in production.

## Behavior change

| `input.version` | Before #17483                                                     | After #17483                          |
| --------------- | ----------------------------------------------------------------- | ------------------------------------- |
| `"0.8.44"`      | `args.version = "0.8.44"`                                         | unchanged                             |
| `null`          | `args.version` omitted; executor falls back to bundled CLI        | `args.version = defaultCliVersion`    |
| `""`            | `args.version` omitted                                            | `args.version = defaultCliVersion`    |
| `"   "`         | `args.version` omitted                                            | `args.version = defaultCliVersion`    |

After this PR, the test-connection path mirrors the contract #17471 enforced for the manual-ingestion path. Customers who override `UI_INGESTION_DEFAULT_CLI_VERSION` (or who set per-source `config.version` overrides) will see test-connection results that finally match what their actual ingestion will run on.

## Test plan

- [x] Existing tests in `CreateTestConnectionRequestResolverTest` still pass (verify resolver execution + auth gating)
- [x] Helper-level coverage already exists in `IngestionUtilsTest.resolveIngestionCliVersion*` (added by #17471) — covers all four input shapes (`null` / `""` / `"   "` / valid)
- [x] `./gradlew :datahub-graphql-core:spotlessJavaCheck` passes

## Why no new resolver tests

Following the pattern established by #17471: the helper function (`IngestionUtils.resolveIngestionCliVersion`) is exhaustively tested in `IngestionUtilsTest`. Adding resolver-level tests for the same input shapes would duplicate that coverage without adding unique value — see [AGENTS.md testing principles](./AGENTS.md#testing-principles-focus-on-value-over-coverage).

### DIFF
--- a/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/ingest/execution/CreateTestConnectionRequestResolver.java
+++ b/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/ingest/execution/CreateTestConnectionRequestResolver.java
@@ -80,10 +80,17 @@ public class CreateTestConnectionRequestResolver implements DataFetcher<Completa
                 RECIPE_ARG_NAME,
                 IngestionUtils.injectPipelineName(
                     input.getRecipe(), executionRequestUrn.toString()));
-            String testVersion = input.getVersion();
-            if (testVersion != null && !testVersion.trim().isEmpty()) {
-              arguments.put(VERSION_ARG_NAME, testVersion.trim());
-            }
+            // Mirror the manual-ingestion path (CreateIngestionExecutionRequestResolver) which
+            // routes the same call through IngestionUtils.resolveIngestionCliVersion. Without
+            // this, a test-connection request with no input.version (or a blank one) silently
+            // omits args.version, causing the executor to fall back to its bundled CLI version
+            // rather than the configured defaultCliVersion. That divergence makes test
+            // connections run on a different CLI than the actual ingestion will use — hiding
+            // compatibility issues that surface in production.
+            arguments.put(
+                VERSION_ARG_NAME,
+                IngestionUtils.resolveIngestionCliVersion(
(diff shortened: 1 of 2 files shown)


sgomezvillamor (approved):
LGTM
May you add tests for the different scenarios listed in the PR description?

puneetagarwal-datahub:
Added resolver-level regression tests in 720a477284 — one test per row of the behavior table in the PR description:

| Test | input.version | Expected args.version |
|---|---|---|
| testExplicitVersionPreserved | `"0.8.44"` | `"0.8.44"` |
| testNullVersionFallsBackToDefault | `null` | `defaultCliVersion` |
| testEmptyVersionFallsBackToDefault | `""` | `defaultCliVersion` |
| testWhitespaceVersionFallsBackToDefault | `"   "` | `defaultCliVersion` |

Each uses `ArgumentCaptor<MetadataChangeProposal>` to deserialize the persisted aspect and assert on the actual `args.version` string — verifying both the resolver-level wiring AND that the helper's contract holds through the call chain.

PR #17484, opened 2026-05-18
fix(bigquery): auto-detect region_qualifiers from dataset locations

dinesh-verma-datahub:
# fix(bigquery): auto-detect region_qualifiers from dataset locations

## Summary

The BigQuery connector scans `INFORMATION_SCHEMA.JOBS` per region to extract usage, lineage, queries, and operations. The list of regions scanned is controlled by the `region_qualifiers` config, which has historically defaulted to `["region-us", "region-eu"]`. When a customer's BigQuery data lives in a single-region location (for example `europe-west1`, `us-central1`, `asia-northeast1`) — or in any region outside the two multi-region defaults — the scan returns zero rows with no error and no warning. Usage and lineage extraction appears completely empty, which is indistinguishable from "no activity in the time window," and customers have spent hours troubleshooting permissions, project setup, and job-history settings before realizing the region filter was the cause.

This PR fixes the silent failure by introducing a new opt-in flag (`region_qualifiers_auto_discovery`) that detects the regions where the customer's data actually lives and extends the scan list accordingly. The flag defaults to `false` to avoid unexpected BigQuery query costs for existing users.

## Problem

- The shipped defaults only cover the `US` and `EU` multi-regions. Datasets located in single-region locations or in non-US/EU multi-regions are silently excluded from usage, lineage, and queries extraction.
- The connector previously emitted no signal in this scenario: no warning, no info entry, no failure. Customers had to discover the cause by elimination.
- A separate but related class of misconfigurations — malformed values in `region_qualifiers` — was surfaced only indirectly, as a per-region SQL build error, with limited diagnostic context.

## Solution

**New flag: `region_qualifiers_auto_discovery` (default: `false`).** When enabled, the connector detects the locations of datasets it discovers during schema extraction and merges any newly found regions into `region_qualifiers` at runtime. The configured `region_qualifiers` list is always used as the starting set — auto-discovery only adds to it, never removes from it. Defaults to `false` to avoid unexpected BigQuery query costs for users who do not need multi-region coverage.

To enable for a project with datasets in non-US/EU regions:

```yaml
source:
  type: bigquery
  config:
    region_qualifiers_auto_discovery: true
```

Alternatively, add the required regions explicitly to `region_qualifiers` without enabling auto-discovery.

**Scope is never silently broadened.** When `region_qualifiers_auto_discovery` is `false` (the default), the configured list is used exactly as given. If discovery surfaces regions not in the configured list, those are reported as an informational breadcrumb only — never scanned, never added automatically.

**BigLake / Omni dataset locations are skipped.** Datasets stored on AWS or Azure via BigLake/Omni report locations like `aws-us-east-1` or `azure-eastus`. These are not valid `INFORMATION_SCHEMA` region qualifiers and would produce per-run failures if treated as such. The connector skips them from auto-detection and increments a dedicated counter on the ingestion report so operators can see how many datasets were skipped for this reason.

**Visibility for previously silent or noisy paths.**

- Malformed `region_qualifiers` entries are reported as a structured failure with the rejected values in the context. This preserves the ERROR-level severity that pre-PR behavior had (via per-region SQL build errors) and gives operators an actionable signal earlier in the run.
- If the effective region list ends up empty (all configured entries malformed, or `region_qualifiers` set to an empty list), a structured failure is emitted with a cause-specific message — the extractor would otherwise silently produce no rows.
- Users with `region_qualifiers_auto_discovery: true` whose schema discovery yields no usable signal receive an informational breadcrumb noting that `region_qualifiers` was not extended, with guidance to set it explicitly if their data lives elsewhere.
- Users with `region_qualifiers_auto_discovery: false` whose discovery surfaces regions outside their configured list receive an informational breadcrumb listing the uncovered regions. This fires alongside any rejection failure so operators see both the typo and the regions outside the scan.
- A per-region exception in the query log fetch no longer prevents other regions from being scanned: each region is wrapped independently, each error is reported with project and region in context, and the "all scanned regions returned zero rows" warning correctly excludes errored regions from its gate to avoid misleading guidance.

**Stable, additive reporting.** The ingestion report gains additive fields covering the configured / auto-discovered / used qualifier breakdown, per-region row counts, a bounded list of any unparseable discovered locations, and the BigLake-skipped counter. Discovered regions are sorted before being merged so the report contents are deterministic across runs.

## Customer impact

- **All existing users:** no behavior change by default. `region_qualifiers_auto_discovery` defaults to `false`.
- **Users with data outside `US` / `EU` (the original bug):** set `region_qualifiers_auto_discovery: true` to have those regions detected and scanned automatically. Without the flag, an informational breadcrumb now lists the uncovered regions discovered during schema ingestion.
- **Users with custom `region_qualifiers`:** scope is never silently broadened. Informational breadcrumbs appear if schema discovery finds regions outside their configured list.
- **Users using BigLake / Omni:** BigLake locations are skipped from auto-detection and tracked on the report.
- **Users with typos in `region_qualifiers`:** previously surfaced as a per-region SQL failure at scan time; now surfaced as a structured failure with the rejected values listed, earlier and more specifically.
- **Standalone queries source users (no schema discovery in the pipeline):** no spurious "auto-detection skipped" entries — that breadcrumb only fires when discovery was actually attempted.

## Notes 

- All new report fields are additive — no existing fields are renamed, repurposed, or changed in type.
- `region_qualifiers_auto_discovery` replaces the earlier `on_defaults` set-equality heuristic (which compared the configured list against the shipped defaults constant). The explicit flag is unambiguous — it doesn't require inferring user intent from the configured value — and is immune to future changes in the default region list.
- Per-region exception handling now uses an explicit try/except that records each error individually rather than the prior approach which suppressed exceptions without distinguishing between "errored" and "returned zero rows". This is necessary to avoid the "all regions empty" warning from firing misleadingly when every region in fact errored.


sgomezvillamor:
Haven't reviewed implementation details yet (which look fine at first scan) — focused on config usability.

**Concern: `on_defaults` heuristic**

The gate `set(configured_normalized) == set(DEFAULT_REGION_QUALIFIERS)` is fragile. If DataHub ever adds a region to the defaults (e.g. `region-asia`), users on the old defaults silently shift from "auto-extend" to "custom user, no extension" — their config didn't change but their behaviour regressed. The heuristic also can't distinguish "never set it" from "explicitly wrote the defaults."

**Proposal: drop the heuristic, accept the breaking change cleanly**

```python
region_qualifiers: List[str] = ["region-us", "region-eu"]
enable_region_discovery: bool = True  # set False to pin to exact list
```

`enable_region_discovery=True` always extends with discovered regions — no comparison against defaults needed. Custom-list users who want the old behaviour set `enable_region_discovery: false`. Migration path is one config line; the guarantee is much clearer to document.

**Visibility** — structured report fields (`region_qualifiers_configured/auto_discovered/used`, per-region counts, BigLake counter) are the right call regardless of design direction.

---

dinesh-verma-datahub:
@sgomezvillamor as discussed, choosing the conservative way to avoid sudden cost increase for users, added the 'region_qualifiers_auto_discovery' flag with default as False.

metadata-ingestion/docs/sources/bigquery/bigquery_post.md
```diff
@@ -31,9 +31,21 @@ source:
...
+##### Multi-Region Configuration
+
+`INFORMATION_SCHEMA.JOBS` is scoped per region. By default DataHub scans `region-us` and `region-eu`. If your project has datasets in other regions (e.g. `europe-west1`, `asia-northeast1`), usage and lineage for those regions will be silently missing unless you configure additional regions.
+
+Two options:
+
+- **Auto-discovery** (recommended for multi-region projects): set `region_qualifiers_auto_discovery: true`. DataHub detects dataset locations during schema ingestion and merges any newly found regions into `region_qualifiers`. The configured `region_qualifiers` list is always used as the starting set — auto-discovery only adds to it, never removes from it.
+- **Explicit list**: add regions directly to `region_qualifiers`, e.g. `["region-us", "region-eu", "region-asia-northeast1"]`.
+
+> **Note:** `region_qualifiers_auto_discovery` defaults to `false` to avoid unexpected BigQuery query costs. Enable it only if you have datasets outside `region-us` / `region-eu`.
```
sgomezvillamor:
```suggestion
:::info

`region_qualifiers_auto_discovery` defaults to `false` to avoid unexpected BigQuery query costs. Enable it only if you have datasets outside `region-us` / `region-eu`.
:::
```

sgomezvillamor (approved):
Great work — clean implementation, solid test coverage, and good structured-report visibility. A few notes for follow-up:

**Minor**
- `bigquery_post.md` recipe shows `region_qualifiers_auto_discovery: false` (the default). Remove it or flip to `true` to show the opt-in.
- The "all scanned regions empty" warning only tells users to set `region_qualifiers` explicitly — should also mention `region_qualifiers_auto_discovery: true` as an alternative.

**⚠️ Warning**
- `^[a-z0-9-]+$` is too permissive: BigLake locations (`aws-us-east-1`, `azure-eastus`) pass through as `region-aws-us-east-1` / `region-azure-eastus`; leading/trailing dashes (`-us`, `us-`) also slip through. Fix: `^[a-z0-9]([a-z0-9-]*[a-z0-9])?$` plus a blocklist for `aws-`/`azure-` prefixes.
- Tests miss the BigLake case — `_normalize_location_to_region_qualifier("aws-us-east-1")` currently returns a value instead of `None`.

**Suggestion**
Consider making `_normalize_location_to_region_qualifier` public — it's a pure utility with no side effects and could replace the `is_biglake_dataset()` proxy in schema_gen (more precise: "does it produce a valid qualifier?" rather than "is it BigLake?").

---

PR #17490, opened 2026-05-18
feat(ingest/teradata): added exponential backoff to transient errors, …

NehaGslab:
Exponential-backoff retry for transient errors
- Added module-level retry infrastructure (_engine_connect_with_retry, _execute_with_retry, _fetchmany_with_retry) wrapping all three SQLAlchemy DB interaction points: engine.connect(), conn.execute(), and result.fetchmany()
- Retries up to 3 attempts with exponential backoff starting at 1 second (1s → 2s)

Configurable connection pool timeout
- Added connection_pool_timeout_ms config field (default: 60000 ms / 60 s) replacing the previous hardcoded value
- WARNING log is emitted with per-thread context
- Pool exhaustion events are counted in TeradataReport (num_pool_exhaustion_events)

Retry
- Added retry logic for get_schema_names and check_historical_table_exists methods

Resource Cleanup
- Added dispose engine logic in get_inspectors to avoid memory leaks

Configurable max pool size
- Added max_pool_size configuration with validation for SQLAlchemy connection pool

Testing
- Added TestConnectionPoolRetry test class covering:
- Default and custom connection_pool_timeout_ms config values
- Pool exhaustion retry behaviour and WARNING log content (thread name/ID)


metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -158,6 +167,218 @@
...
+# ---------------------------------------------------------------------------
+# Exponential-backoff retry helpers
+# ---------------------------------------------------------------------------
+
+_RETRY_MAX_ATTEMPTS = 3
+_RETRY_INITIAL_BACKOFF_SECONDS = 1.0
+
+# Substrings found in Teradata / network error messages that are safe to retry
+# on an *existing* connection (execute / fetchmany).
+_RETRYABLE_ERROR_SUBSTRINGS: Tuple[str, ...] = (
```
treff7es, line 230: 🟡 **I1: Substring-based retry classification can match permanent errors**

`"please try again"` is generic English — a Teradata error like `"Authentication failed. Please contact your administrator and try again."` (a plausible message used by some Teradata variants) would match and the connector would retry an auth failure 3 times. Same with `"tdgss"` — a message like `"tdgss configuration is invalid"` (a permanent config error from the TDGSS subsystem) would be retried.

**Inverse problem (false negatives):** `"connect timed out"` requires those exact three words in that order. A driver printing `"Connection has timed out while connecting"` would NOT match — real transient failures get re-raised immediately.

**Suggestion:** prefer `_RETRYABLE_ERROR_CODE_RE` (numeric codes are stable) as the primary signal. Add a permanent-error deny-list checked first (`auth`, `does not exist`, `permission denied`, `invalid user`, `password`). Document each substring with the exact Teradata version / error string it targets so future maintainers can prune false matches.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -158,6 +167,218 @@
...
+_RETRYABLE_CONNECT_EXTRA_SUBSTRINGS: Tuple[str, ...] = (
+    "connection reset",
+    "broken pipe",
+    " eof",
+    "socket closed",
+)
+
+# Teradata numeric error codes that are retryable (matched in the
+# stringified exception which typically contains "[Error NNNN]").
+_RETRYABLE_ERROR_CODE_RE: re.Pattern = re.compile(
```
treff7es, line 261: ℹ️ **S5: Document the Teradata error codes**

A future maintainer has zero context for whether `3120` is correctly classified. Add an inline comment explaining each:

```python
# Teradata error codes that are safe to retry. Match against str(exc) because
# the driver formats messages as "[Error NNNN]: ...".
#
#   2631 — deadlock; transaction aborted, retry safe after backoff
#   3111 — transaction aborted due to system restart
#   3120 — "cannot rollback/modify while sessions logged on" (transient during DBA work)
#   3597 — restart in progress, resubmit
#   3598 — concurrent change conflict on table, retry
#   3897 — database is being restored
#
# Source: Teradata Database Messages reference, doc B035-1096.
```

Also minor: `re.IGNORECASE` is dead here — the pattern has no letters that change case. Can be dropped.

`3120` is the most arguably-wrong inclusion — usually a transient DBA-operation collision, but could mask user errors. Worth a second look.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -158,6 +167,218 @@
...
+def _jittered_backoff(attempt: int, initial_backoff_seconds: float) -> float:
+    """Return a full-jitter exponential backoff duration for *attempt* (0-based).
+
+    Full jitter (``uniform(0, cap)``) is the recommended strategy when many
+    concurrent workers share the same resource: pure exponential backoff causes
+    a thundering-herd re-surge because all threads sleep for the *same* duration
+    and then hammer the server simultaneously.  Full jitter spreads retries
+    uniformly across the window so the server load stays roughly constant.
+    """
+    cap = min(initial_backoff_seconds * (2**attempt), 30.0)
```
treff7es: ℹ️ **S4: Magic number `30.0` should be a named constant**

Lift to a module-level constant alongside `_RETRY_MAX_ATTEMPTS`:

```python
_RETRY_BACKOFF_CAP_SECONDS = 30.0  # full-jitter ceiling per attempt; protects against runaway sleeps if user sets large retry_initial_backoff_seconds
```

And mention it in the `retry_initial_backoff_seconds` Field description — currently the cap is invisible to users who configure a large initial backoff and expect attempt 2 to wait ~80s.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -158,6 +167,218 @@
...
+        raise ValueError(f"max_attempts must be >= 1, got {max_attempts}")
+    conn: Optional[Connection] = None
+    for attempt in range(max_attempts):
+        try:
+            conn = engine.connect()
+            break
+        except PoolTimeoutError as exc:
+            thread = current_thread()
+            if report is not None:
+                report.increment_pool_exhaustion_events()
```
treff7es: 🟡 **I2: Counter semantics for pool exhaustion are misleading**

`num_pool_exhaustion_events` is incremented on **every** `PoolTimeoutError`, including the final raise. So `retry_max_attempts=3` produces `num_pool_exhaustion_events=3` for **one** logical event. Meanwhile `num_db_retries` is `2` (only successful retries). Operators correlating these numbers will be confused.

**Fix (one of):**
1. Move `increment_pool_exhaustion_events()` inside the retry branch (after the `if attempt == max_attempts - 1: raise` check), so it counts only retried attempts.
2. Or rename to `num_pool_timeout_retries` to reflect what it actually counts.

If the double-increment semantics (events vs retries) is intentional, please add a comment naming it: `# pool_exhaustion is logged per attempt; db_retries only counts non-terminal attempts`.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -182,7 +403,9 @@ def get_schema_columns(
     start_time = time.time()
     columns: Dict[str, List[Any]] = {}
     columns_query = f"select * from dbc.{dbc_columns} where DatabaseName (NOT CASESPECIFIC) = :schema (NOT CASESPECIFIC) order by TableName, ColumnId"
-    rows = connection.execute(text(columns_query), {"schema": schema}).fetchall()
+    rows = _execute_with_retry(
```
treff7es, line 564: 🔴 **B2: `report=self.report` not forwarded — retry telemetry is silent here**

This call and the analogous ones at lines 436 (`get_schema_pk_constraints`) and 640 (`get_schema_foreign_keys`) use `_execute_with_retry(...)` without passing `report=`. Because `_execute_with_retry`'s `report` parameter defaults to `None`, every retry inside these largest/slowest extraction paths is invisible:

- `num_db_retries` is never incremented for column/PK/FK extraction.
- A flaky network during column extraction can produce up to `retry_max_attempts - 1` silent retries per call, multiplied across every schema, with no aggregated count in the ingestion report.

**Fix:** forward `self.report` (the function already receives `self` as first arg):
```python
rows = _execute_with_retry(
    connection, text(columns_query), {"schema": schema},
    report=getattr(self, "report", None),
).fetchall()
```
Better: route through the bound `self._retry_execute` wrapper, which already passes `self.report`. Same applies to lines 436 and 640.

This is a one-line fix per call site and directly restores the telemetry the PR sets out to add.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -491,6 +714,12 @@ def optimized_get_view_definition(
 
 @dataclass
 class TeradataReport(SQLSourceReport, BaseTimeWindowReport):
+    # Column extraction metrics
+    num_columns_processed: int = 0
+    num_column_extraction_failures: int = 0
+    num_primary_keys_processed: int = 0
+    column_extraction_duration_seconds: float = 0.0
```
treff7es, line 893: 🔴 **B1: New counter fields are mutated unguarded from worker threads — race on `float +=`**

These new fields are mutated via bare `+=` at five pre-existing sites that run inside the `ThreadPoolExecutor` workers (functions are bound to `TeradataDialect.get_columns` / `get_pk_constraint` via `setattr` at lines 1294/1317):

- Line 424: `self.report.column_extraction_duration_seconds += extraction_time`  ← **float — race**
- Line 485: `self.report.num_primary_keys_processed += 1`
- Line 609: `self.report.num_columns_processed += 1`
- Line 616: `self.report.num_column_extraction_failures += 1`
- Line 622: `self.report.column_extraction_duration_seconds += end_time - start_time`  ← **float — race**

Unlike `int +=` (mostly atomic under CPython's GIL), **`float +=` is not atomic** — it can lose updates under contention. The `_lock` introduced by this PR at line 757 exists to guard exactly this kind of mutation, but the existing increment sites bypass it.

This isn't strictly a regression (the `+=` sites pre-date this PR), but the PR is the right place to fix it because (a) you're newly declaring these fields and could land both halves together, and (b) you've added `_lock` infrastructure that they now should use.

**Fix:** Add helpers to `TeradataReport` mirroring the existing `increment_db_retries` pattern:
```python
def increment_columns_processed(self) -> None:
    with self._lock:
        self.num_columns_processed += 1

def increment_column_extraction_failures(self) -> None:
    with self._lock:
        self.num_column_extraction_failures += 1

def increment_primary_keys_processed(self) -> None:
    with self._lock:
        self.num_primary_keys_processed += 1

def add_column_extraction_duration(self, seconds: float) -> None:
    with self._lock:
        self.column_extraction_duration_seconds += seconds
```

Then replace the bare `+=` at all five sites.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -519,6 +749,21 @@ class TeradataReport(SQLSourceReport, BaseTimeWindowReport):
     # Audit query processing statistics
     num_audit_query_entries_processed: int = 0
 
+    # Retry statistics
+    num_db_retries: int = 0
+
+    # Single internal lock — not serialised, not compared.  Protects all report
+    # fields that are mutated from ThreadPoolExecutor worker threads.
+    _lock: Lock = field(default_factory=Lock, init=False, repr=False, compare=False)
```
treff7es, line 929: ℹ️ **`_lock` is a leaky abstraction (used at ~7 external call sites)**

The underscore prefix declares "private by convention", but `TeradataSource` reaches in and uses `with self.report._lock:` at multiple sites for ad-hoc multi-field atomic updates. Only two specialised helpers are exposed (`increment_db_retries`, `increment_pool_exhaustion_events`).

Consider exposing a single context manager that names the intent:
```python
@contextmanager
def atomic(self) -> Generator[None, None, None]:
    """Take the report lock for a multi-field atomic update.

    Use for multi-field updates that must observe a consistent snapshot. For
    single-field counter bumps prefer the dedicated increment_* helpers.
    """
    with self._lock:
        yield
```

Then replace `with self.report._lock:` with `with self.report.atomic():` everywhere — `_lock` becomes genuinely private and the call sites read better.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -600,6 +845,24 @@ class TeradataConfig(BaseTeradataConfig, BaseTimeWindowConfig):
...
+        "capped to max_pool_size. For example, max_workers=10 with max_pool_size=13 "
+        "creates a pool of 10, not 13. Teradata drivers can be sensitive to high "
+        "connection counts; keep this conservative. Must be between 1 and 50.",
+    )
+
+    @field_validator("max_pool_size")
+    @classmethod
+    def validate_max_pool_size(cls, v: int) -> int:
+        if not 1 <= v <= 50:
+            raise ValueError(f"max_pool_size must be between 1 and 50, got {v}")
```
treff7es: 🟡 **I7: Replace `@field_validator` with declarative `Field(ge=..., le=...)`**

Four new validators (`validate_max_pool_size`, `validate_connection_pool_timeout_ms`, `validate_retry_max_attempts`, `validate_retry_initial_backoff_seconds`) all do pure bounded-range checks. Per DataHub standards, prefer declarative constraints because (a) the bound is harvested by the connector docs generator (`metadata-ingestion/docs/sources/teradata/`) and surfaces in user docs, (b) ~24 LOC of validator boilerplate disappears, (c) pydantic's error message already names the field, the limit, and the observed value.

```python
max_pool_size: int = Field(default=13, ge=1, le=50, description="...")
connection_pool_timeout_ms: int = Field(default=60000, ge=1, description="...")
retry_max_attempts: int = Field(default=_RETRY_MAX_ATTEMPTS, ge=1, description="...")
retry_initial_backoff_seconds: float = Field(default=_RETRY_INITIAL_BACKOFF_SECONDS, gt=0, description="...")
```

Then delete the four `validate_*` methods.

Also minor: the upper bound `50` is unexplained — is it Teradata's `MAXSESSIONS` cap? An empirical safe ceiling? Worth a one-line comment naming the source.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -686,6 +949,52 @@ def _validate_column_extraction_options(self) -> "TeradataConfig":
...
+    @field_validator("connection_pool_timeout_ms")
+    @classmethod
+    def validate_connection_pool_timeout_ms(cls, v: int) -> int:
+        if v < 1:
+            raise ValueError(f"connection_pool_timeout_ms must be >= 1, got {v}")
+        return v
+
+    retry_max_attempts: int = Field(
+        default=_RETRY_MAX_ATTEMPTS,
+        description=(
```
treff7es, line 1164: ℹ️ **S3: Field descriptions for the three retry-related fields don't meet the DataHub bar**

`max_pool_size` (line 850–856) is excellent — explains what/format/where/when/effect. The other three are perfunctory.

Example for `retry_max_attempts`:
```python
retry_max_attempts: int = Field(
    default=_RETRY_MAX_ATTEMPTS,
    ge=1,
    description=(
        "Maximum total attempts (initial + retries) for retryable database operations "
        "(connect, execute, fetchmany). Retryable conditions: pool exhaustion, "
        "transaction-aborted messages, dead-socket signals at connect time, and Teradata "
        "error codes 2631/3111/3120/3597/3598/3897. "
        "Worst-case wall time per operation is "
        "retry_max_attempts * connection_pool_timeout_ms + sum of backoff sleeps (capped at 30s each). "
        "Increase when ingesting from a busy cluster; decrease to surface persistent errors faster."
    ),
)
```

Same treatment for `connection_pool_timeout_ms` and `retry_initial_backoff_seconds` — mention the **30s backoff cap** which is currently invisible to users.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -1086,6 +1398,85 @@ def create(cls, config_dict, ctx):
...
+        """Fetch a batch with config-driven retry / backoff."""
+        return _fetchmany_with_retry(
+            result,
+            batch_size,
+            max_attempts=self.config.retry_max_attempts,
+            initial_backoff_seconds=self.config.retry_initial_backoff_seconds,
+            report=self.report,
+        )
+
+    def _get_schema_names_with_retry(self, engine: Engine) -> List[str]:
```
treff7es, line 1649: 🟡 **I4: Final raise has no `report.failure` — exhausted-retry context is lost**

When the final attempt fails, the bare `raise` re-raises silently — no `self.report.failure(...)` or `self.report.warning(...)`. The caller (`get_inspectors`) doesn't catch it either, so the user sees a raw `OperationalError` in stderr as the visible failure cause. They don't learn from the report that 3 attempts were made.

Per DataHub standards: "After retries are exhausted, the original error context (stack trace / cause) must reach the user." Today it only reaches stderr.

**Fix:** before the final `raise` (line 1478), emit:
```python
self.report.failure(
    title="Could not list schemas after retries",
    message=f"Failed to list schemas after {max_attempts} attempts. "
            "Check network connectivity, authentication, and database permissions.",
    context=str(exc),
    exc=exc,
)
```

Same pattern applies to the other retry helpers' final-raise paths.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -1121,60 +1512,66 @@ def get_inspectors(self):
...
+            # check against. An empty cache means either discovery was disabled
+            # (include_tables/include_views both False) or it ran and genuinely
+            # found nothing — in both cases there is no oracle to validate against
+            # and we fall back to trusting the user's list.
+            have_db_inventory = (
+                self.config.include_tables or self.config.include_views
+            ) and bool(self._tables_cache)
+
+            # Create separate connections for each database to avoid connection lifecycle issues
+            for db in databases:
```
treff7es, line 1762: 🟡 **I3: Per-database connect failure aborts entire ingestion**

If any single database's `_retry_connect(engine)` exhausts retries (pool timeout, transient network), the exception propagates out of the generator and **every remaining database in the loop is silently skipped**. The user gets no warning saying "we never even tried database X, Y, Z because the connect to database D failed."

This violates the project standard "Don't fail entire ingestion on a single error — continue processing."

**Fix:**
```python
for db in databases:
    if not self.config.database_pattern.allowed(db):
        continue
    # ... existing user-supplied validation ...
    try:
        with self._retry_connect(engine) as conn:
            db_inspector = inspect(conn)
            db_inspector._datahub_database = db
            yield db_inspector
    except Exception as e:
        self.report.warning(
            title="Failed to inspect database",
            message=f"Could not acquire a connection to database {db!r} "
                    f"after {self.config.retry_max_attempts} attempts. Skipping.",
            context=str(e),
            exc=e,
        )
        continue
```

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -2177,10 +2610,11 @@ def _execute_with_cursor_fallback(
                 # Fall through to regular execution
```
treff7es, line 2884: 🔴 **B3: Blanket `except Exception` masks permanent errors and can OOM ingestion**

This catch covers:
- Auth failures (after `_retry_execute` exhausts retries and re-raises)
- "Database does not exist" / permission denied
- SQL syntax errors in lineage queries
- Out-of-memory on the client
- KeyboardInterrupt's subclasses that inherit from Exception

After catching, it logs `logger.warning` (which goes to logs, **not** the ingestion report) and silently re-executes the same query *without* `stream_results=True`. The fallback may then pull the entire query result into memory, OOM'ing the ingestion process — for what should have been a clean failure.

**Fix:**
1. Catch only the specific exception families that indicate cursor unsupported (e.g. SQLAlchemy's `NotSupportedError`, or check the message for a documented "streaming not supported" signal).
2. Route to `self.report.warning(title=..., message=..., context=..., exc=e)` so it appears in the structured report.
3. Include `traceback.format_exc()` in `context` (or use `exc=e`) to preserve the original stack.

metadata-ingestion/tests/unit/test_teradata_source.py
```diff
@@ -2799,3 +2976,917 @@ def test_no_validation_when_databases_not_user_supplied(
...
+
+        assert mock_engine.connect.call_count == 2
+        assert report.num_db_retries == 1
+        good_conn.close.assert_called_once()
+
+
+class TestRetryConfig:
+    """retry_max_attempts and retry_initial_backoff_seconds config fields."""
+
+    def test_retry_max_attempts_default(self):
```
treff7es: 🔴 **B4: Trivial config-default tests violate DataHub testing standards**

Per `metadata-ingestion`'s testing standards (and DataHub CLAUDE.md): testing that a pydantic field has its declared default value is an **🚨 ABSOLUTE BLOCKER**. These re-verify pydantic, not your code:

- `test_retry_max_attempts_default` (line 3142) — asserts `config.retry_max_attempts == 3` against `Field(default=3)`
- `test_retry_initial_backoff_seconds_default` (line 3146) — same pattern
- `test_max_pool_size_default` / `_min_boundary` / `_max_boundary` (lines 3200, 3204, 3208)
- `test_connection_pool_timeout_ms_default` (line 3281) / `_min_boundary` (line 3293)
- The `_zero_is_invalid` / `_negative_is_invalid` / `_above_limit_is_invalid` tests re-verify pydantic's `@field_validator` raising `ValidationError` — pure framework-behavior tests.

**Recommendation:** delete these ~12 tests. The valuable test in this area is `TestEffectiveMaxWorkers` (line 3221) which tests the *consequence* (`min(max_workers, max_pool_size)` flowing through to `_effective_max_workers`) — keep and extend that pattern.

Note: this is independent of I7 above — even if you keep the `@field_validator` methods, the *defaults* and *boundaries* tests are still anti-patterns.

metadata-ingestion/tests/unit/test_teradata_source.py
```diff
@@ -2799,3 +2976,917 @@ def test_no_validation_when_databases_not_user_supplied(
...
+
+class TestTeradataReportFields:
+    """Newly declared TeradataReport fields have the correct default values."""
+
+    def test_column_extraction_fields_default_to_zero(self):
+        report = TeradataReport()
+        assert report.num_columns_processed == 0
+        assert report.num_column_extraction_failures == 0
+        assert report.num_primary_keys_processed == 0
+        assert report.column_extraction_duration_seconds == 0.0
```
treff7es: 🟡 **I5: Missing thread-contention test for `TeradataReport._lock`**

`test_column_extraction_fields_default_to_zero` (the only test in `TestTeradataReportFields`) asserts defaults — which itself is a trivial test (see B4). More importantly, the whole reason `_lock` exists (line 757 of teradata.py) is to make `num_db_retries` / `num_pool_exhaustion_events` safe to increment from `ThreadPoolExecutor` workers, but no test actually races increments against each other.

Anyone removing `with self._lock:` from `increment_db_retries` / `increment_pool_exhaustion_events` would not break the test suite.

**Suggested test:**
```python
def test_concurrent_increments_are_atomic(self):
    """Under N threads × M increments each, the final count must equal N*M."""
    import concurrent.futures
    report = TeradataReport()
    n_threads, m_per_thread = 16, 1000

    def worker(_: int) -> None:
        for _ in range(m_per_thread):
            report.increment_db_retries()
            report.increment_pool_exhaustion_events()

    with concurrent.futures.ThreadPoolExecutor(max_workers=n_threads) as ex:
        list(ex.map(worker, range(n_threads)))

    assert report.num_db_retries == n_threads * m_per_thread
    assert report.num_pool_exhaustion_events == n_threads * m_per_thread
```

This is the only test that would catch someone removing the lock. Combined with B1, also add a contention test for the *float* counters that currently bypass the lock.

metadata-ingestion/tests/unit/test_teradata_source.py
```diff
@@ -2799,3 +2976,917 @@ def test_no_validation_when_databases_not_user_supplied(
...
+                return_value=fixed_backoff,
+            ),
+            patch("time.sleep") as mock_sleep,
+            _engine_connect_with_retry(mock_engine, max_attempts=2),
+        ):
+            pass
+
+        mock_sleep.assert_called_once_with(fixed_backoff)
+
+    def test_backoff_grows_with_attempt_number(self):
```
treff7es, line 3642: 🟡 **I6: 30-second backoff cap is dead code in tests**

`teradata.py:215` caps backoff at `min(initial * 2**attempt, 30.0)`. The cap is critical because at `retry_initial_backoff_seconds=20.0, attempt=2` the raw value would be 80s — the cap keeps a worker from sleeping minutes. But every test in `TestJitteredBackoff` and `test_backoff_grows_with_attempt_number` (here) uses `initial=1.0, attempt ∈ [0..3]`, where `min(8.0, 30.0) = 8.0`. The 30s cap is never exercised — a regression that changed `30.0` to `300.0` (or removed the cap) would not fail any test.

**Suggested test:**
```python
def test_backoff_capped_at_30_seconds(self):
    """Cap kicks in once initial*2^attempt exceeds 30 s."""
    with patch("random.uniform", side_effect=lambda lo, hi: hi):
        # initial=20.0, attempt=2 -> raw = 80.0 -> capped to 30.0
        assert _jittered_backoff(2, 20.0) == 30.0
        # initial=5.0, attempt=10 -> raw = 5120.0 -> capped to 30.0
        assert _jittered_backoff(10, 5.0) == 30.0
```

treff7es (changes requested):

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Added 4 levels of checks:
1. Permanent error codes
2. Permanent substrings
3. Retryable error codes 
4. Retryable substrings
Added documentation for errors from Teradata docs

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Added documentation for errors from Teradata docs
For error 3120 is retryable according to Teradata docs

https://docs.teradata.com/r/Enterprise_IntelliFlex_Lake_VMware/Teradata-Call-Level-Interface-Version-2-Reference-for-Workstation-Attached-Systems-20.00/Return-Error-and-Failure-Codes/Responding-to-Error-and-Failure-Parcels

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Added constant for max attempts

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Moved exhaustion events calculation to same block as db retries

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Passes report as argument to retry methods

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Added helpers to TeradataReport to avoid conflict

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Added atomic contextmanager to make call private

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Removed field validators and added validations as per Datahub standards 
Added comment in documentation for MAXSESSIONS cap

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Added documentation for connection_pool_timeout_ms and retry_initial_backoff_seconds

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Added self.report.failure before raise statement

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Fixed exception logic inside loop to avoid entire ingestion failure

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Removed catching Exception class and added specific class for exception handling
Added report. warning for exception logging
Passed exception to preserve stacktrace

metadata-ingestion/tests/unit/test_teradata_source.py
NehaGslab: Removed invalid tests

metadata-ingestion/tests/unit/test_teradata_source.py
NehaGslab: Added suggested test

metadata-ingestion/tests/unit/test_teradata_source.py
NehaGslab: Added suggested test

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -1086,6 +1524,104 @@ def create(cls, config_dict, ctx):
...
+                )
+                self.report.increment_db_retries()
+                time.sleep(backoff)
+                self.report.failure(
+                    title="Could not list schemas after retries",
+                    message=f"Failed to list schemas after {max_attempts} attempts. "
+                    "Check network connectivity, authentication, and database permissions.",
+                    context=str(exc),
+                    exc=exc,
+                )
```
treff7es: 🔴 **N1 (BLOCKER): `report.failure` fires on every successful retry**

This `self.report.failure(...)` sits inside the `for`-loop body, **after** `time.sleep(backoff)`. Control flow:

- Attempt 0 hits a transient error → log warning, sleep, **then call `report.failure("Could not list schemas after retries"...)`**, then iterate.
- Attempt 1 succeeds → `return` schema names — but the report now contains a spurious failure with a stale exception.

So any single transient blip that succeeds on retry marks the whole ingestion as failed. The message also lies: at attempt 0 of `max_attempts=3`, it claims "Failed to list schemas after 3 attempts" when only one attempt has failed.

**Fix:** Delete lines 1616–1622 entirely. The genuine-exhaustion case is already handled by the `if attempt == max_attempts - 1` branch at 1594–1607.

**Why the new tests miss this:** `test_retries_on_transient_error_then_succeeds` doesn't assert on `source.report.failures` — it only checks `num_db_retries >= 1`. Please add an assertion that `source.report.failures == []` after a successful retry.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -1121,60 +1657,77 @@ def get_inspectors(self):
...
+                        yield db_inspector
+                except Exception as e:
+                    self.report.warning(
+                        title="Failed to inspect database",
+                        message=f"Could not acquire a connection to database {db!r} "
+                        f"after {self.config.retry_max_attempts} attempts. Skipping.",
+                        context=str(e),
+                        exc=e,
+                    )
+                    continue
```
treff7es: 🔴 **N11 (BLOCKER): try/except around `yield` swallows downstream consumer errors**

`get_inspectors` is a generator. `yield db_inspector` transfers control to the consumer; **any** exception raised inside the consumer's iteration body propagates back **through** `yield` and is caught here, downgraded to `report.warning("Failed to inspect database")`, and the per-DB loop just continues.

This means a KeyError, AttributeError, auth failure during column extraction, lineage-emission bug — anything raised downstream while iterating that inspector — gets reclassified as "connection failure" with the wrong title, wrong stage, and the original cause is buried.

The I3 fix was supposed to make per-DB **connect** failures non-fatal; instead it now masks all errors.

**Fix:** Split the connect step from the yield so the try/except only wraps the connect:

```python
try:
    conn_cm = self._retry_connect(engine)
    conn = conn_cm.__enter__()
except Exception as e:
    self.report.warning(title="Failed to acquire connection", ..., exc=e)
    continue
try:
    db_inspector = inspect(conn)
    db_inspector._datahub_database = db
    yield db_inspector
finally:
    conn_cm.__exit__(None, None, None)
```

Also add a behavioural test: `[db1 ok, db2 raises auth error, db3 ok]` → `yielded == ["db1", "db3"]` and `len(report.warnings) == 1`.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -158,6 +168,311 @@
...
+    "invalid logon",
+    "invalid user",
+    "password",
+    "permission denied",
+    "access denied",
+    "does not exist",
+    "no access",
+    "tdgss configuration",  # permanent TDGSS mis-configuration (vs. transient handshake)
+    "invalid configuration",
+)
```
treff7es, line 209: 🟡 **N12: `_PERMANENT_ERROR_SUBSTRINGS` includes over-broad terms**

Two substrings here will deny retry on genuinely retryable errors:

- **`"password"`** matches any error mentioning the word — including transient messages like "your session was terminated; please re-enter username and password to reconnect" (TRANSIENT, post network reset).
- **`"does not exist"`** can fire transiently when CREATE/DROP races a SELECT, or against a temp object that's being recreated.

The numeric error-code regex (`_PERMANENT_ERROR_CODE_RE` with 8017/3003/3523/3524) is already authoritative and unambiguous. Consider removing these broad substrings, or qualifying them:

```python
# Only deny if BOTH password AND invalid/incorrect appear:
if "password" in msg and ("invalid" in msg or "incorrect" in msg):
    return False
```

Same caution applies to `"invalid configuration"` — risk of matching a transient "invalid configuration session ID" message.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -158,6 +168,311 @@
...
+                        title="Database execute failed after retries",
+                        message=(
+                            f"A retryable error persisted after {attempt + 1} attempts. "
+                            "The operation has been abandoned. "
+                            "Check Teradata connectivity and cluster stability."
+                        ),
+                        context=str(exc),
+                        exc=exc,
+                    )
+                raise
```
treff7es: 🟡 **N13: Permanent errors re-raise with no report breadcrumb**

The guard `_should_retry(exc) and attempt > 0` means a `report.failure` is only emitted when (a) we already retried at least once AND (b) the error is still retryable. **Permanent errors (auth, access denied, "does not exist") are re-raised without any `report.failure`, no log line, no exception context at all.**

The comment says "permanent errors are left for callers ... which have the higher-level context", but several callers (`get_schema_columns`, `get_schema_pk_constraints`, `get_schema_foreign_keys`, `cache_tables_and_views`) do **not** wrap these in a try/except — the exception bubbles up bare. From the user's perspective, a Teradata permission denial silently aborts metadata extraction with only the SQLAlchemy traceback visible.

Same pattern at lines 452–465 in `_fetchmany_with_retry`.

**Suggested fix:** before `raise`, always emit `self.report.warning(title=f"Database {op_name} failed", message=..., exc=exc)` — the helper has the SQL statement and attempt count which callers lack.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -2155,32 +2747,57 @@ def _execute_with_cursor_fallback(
...
+                    raise
+                self.report.warning(
+                    title="Server-side cursor not supported — falling back to client-side execution",
+                    message=(
+                        "Streaming cursor mode failed with a driver-level error. "
+                        "Falling back to client-side buffering. "
+                        "Consider setting use_server_side_cursors: false to suppress this warning."
+                    ),
+                    exc=e,
                 )
```
treff7es, line 2883: ℹ️ **N16: Substring filter on error message can mis-classify**

The match is `str(e).lower()`, so an `OperationalError` whose embedded SQL text happens to contain `"cursor"` or `"stream"` (e.g. a CTE named `stream_events`, a query referencing `dbc.QryLogV` that mentions cursors in `RequestText`) will be misclassified as "cursor mode failed" and quietly switch to client-side buffering.

Low-frequency in practice, but the misclassification means the warning message lies about the cause.

Consider tightening: require both `"stream"` AND `"not supported"` to co-occur, or check exception subclass / SQLSTATE instead of message contents.

Also missing: a test that exercises the **propagation** branch (line 2783-2784 `raise`) — i.e. that a permission-denied `OperationalError` propagates and does NOT fall back. That's the critical safety guard from the B3 fix.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -686,6 +1071,56 @@ def _validate_column_extraction_options(self) -> "TeradataConfig":
...
+            "Permanent errors (auth failures, permission denied, object does not exist) "
+            "are never retried regardless of this setting. "
+            "Worst-case added latency per operation is approximately "
+            "retry_max_attempts × connection_pool_timeout_ms plus backoff sleeps "
+            "(each capped at 30 seconds by retry_backoff_cap_seconds). "
+            "Increase when ingesting from a busy or flaky cluster; "
+            "decrease to surface persistent errors faster. "
+            "Default is 3."
+        ),
+    )
```
treff7es, line 1179: ℹ️ **N6: Missing upper bounds on retry/timeout fields**

`max_pool_size` got `le=50` (line 975), but the I7 sweep missed upper bounds on:

- `connection_pool_timeout_ms`: only `ge=1` — `60_000_000` (16 h) silently accepted
- `retry_max_attempts`: only `ge=1` — `retry_max_attempts=1000` runs a 1000-attempt loop

Reasonable ceilings would prevent footguns:
```python
connection_pool_timeout_ms: int = Field(default=60000, ge=1, le=600_000, ...)  # 10 min cap
retry_max_attempts: int = Field(default=3, ge=1, le=10, ...)
```

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -158,6 +168,311 @@
...
+                raise
+            backoff = _jittered_backoff(attempt, initial_backoff_seconds)
+            logger.warning(
+                f"Retryable DB error on fetchmany (attempt {attempt + 1}/{max_attempts}): {exc}. "
+                f"Retrying in {backoff:.2f}s..."
+            )
+            if report is not None:
+                report.increment_db_retries()
+            time.sleep(backoff)
+    raise AssertionError("unreachable")  # loop always raises or returns
```
treff7es: ℹ️ **N10: Three retry helpers duplicate ~150 lines of identical scaffolding**

`_engine_connect_with_retry`, `_execute_with_retry`, and `_fetchmany_with_retry` share:
- `if max_attempts < 1: raise ValueError(...)`
- `for attempt in range(max_attempts): try: ... except Exception as exc:`
- terminal-attempt check + `raise`
- `_jittered_backoff` + `logger.warning` + `report.increment_db_retries()` + `time.sleep`
- `raise AssertionError("unreachable")`

The duplication is what allowed N1 (the spurious failure bug) to live in the one helper not covered by the shared pattern. Extracting a single `_retry_call(op, classify, ...)` primitive would remove ~50 lines and eliminate the drift hazard. Not a blocker for this PR, but worth a follow-up.

Also: `_engine_connect_with_retry` has near-identical `except PoolTimeoutError` and `except Exception` branches (lines 347-373). `PoolTimeoutError` is already a subclass of `Exception` and `_should_retry` returns `True` for it. Merging them removes ~15 LOC and fixes the missed final-attempt `increment_pool_exhaustion_events` accounting.

metadata-ingestion/tests/unit/test_teradata_source.py
```diff
@@ -2799,3 +2976,915 @@ def test_no_validation_when_databases_not_user_supplied(
...
+
+class TestTeradataReportFields:
+    """Newly declared TeradataReport fields have the correct default values."""
+
+    def test_column_extraction_fields_default_to_zero(self):
+        report = TeradataReport()
+        assert report.num_columns_processed == 0
+        assert report.num_column_extraction_failures == 0
+        assert report.num_primary_keys_processed == 0
+        assert report.column_extraction_duration_seconds == 0.0
```
treff7es: 🟡 **N4: Trivial pydantic-default test still present (violates B4)**

This just asserts dataclass int fields default to `0`/`0.0` — pure dataclass behaviour, not application logic. B4 from the previous review explicitly forbade this pattern. Please delete the whole method.

The meaningful sibling — `test_concurrent_increments_are_atomic` directly below — already proves the fields exist and behave correctly under load.

metadata-ingestion/tests/unit/test_teradata_source.py
```diff
@@ -2799,3 +2976,915 @@ def test_no_validation_when_databases_not_user_supplied(
...
+        with (
+            patch("time.sleep"),
+            pytest.raises(PoolTimeoutError),
+            _engine_connect_with_retry(mock_engine, max_attempts=3, report=report),
+        ):
+            pass  # should never reach here
+
+        # All 3 attempts counted as exhaustion events; only 2 retries (last attempt just raises).
+        assert report.num_pool_exhaustion_events == 2
+        assert report.num_db_retries == 2
```
treff7es: 🟡 **N2: Test comment contradicts the assertion**

```python
# All 3 attempts counted as exhaustion events; only 2 retries (last attempt just raises).
assert report.num_pool_exhaustion_events == 2
```

The assertion is correct under the I2 fix (counts only retries, not the final raise), but the comment says "All 3 attempts counted." Either:

1. Fix the comment: `# 2 retries (attempts 0 and 1 sleep+increment); attempt 2 raises without counting.`
2. Or accept that `num_pool_exhaustion_events` now counts retries, not events, and **rename** the field to `num_pool_timeout_retries` — the current name still suggests the original semantics.

Option 2 is cleaner long-term; operators correlating this with `num_db_retries` will continue to be confused by the apparent mismatch with the field name.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Removed duplicate error reporting logic

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Split the connect step as per suggestion. Added test case to check behaviour.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Removed broad terms from _PERMANENT_ERROR_SUBSTRINGS, these removed errors are already handled by _PERMANENT_ERROR_CODE_RE section

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Added self.report.warning code for both paths if and else

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Fixed (dual-token _not_supported and _cursor_or_stream)

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Fixed upper bounds for connection_pool_timeout_ms and retry_max_attempts

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Added common code in _retry_call method to avoid code repetition

metadata-ingestion/tests/unit/test_teradata_source.py
NehaGslab: Removed test case

metadata-ingestion/tests/unit/test_teradata_source.py
NehaGslab: Renamed to num_pool_timeout_retries

metadata-ingestion/tests/unit/test_teradata_source.py
```diff
@@ -2799,3 +2976,1193 @@ def test_no_validation_when_databases_not_user_supplied(
...
+    def test_concurrent_increments_are_atomic(self):
+        """All lock-protected helpers produce exact counts under thread contention.
+
+        N threads each call every increment_* / add_* helper M times.  Without
+        the _lock, float += is not atomic under the GIL and int += can lose
+        updates under high contention.  Any removal of 'with self._lock:' from
+        a helper will cause this test to fail non-deterministically (and very
+        reliably at n_threads=16, m_per_thread=1000).
+        """
+        import concurrent.futures
```
treff7es: 🔴 **S1 (BLOCKER): Late import inside test method**

`import concurrent.futures` inside a function body is an **ABSOLUTE BLOCKER** per DataHub testing standards.

This import is also **redundant** — `ThreadPoolExecutor` is already imported at the top of this file:
```python
from concurrent.futures import FIRST_COMPLETED, Future, ThreadPoolExecutor, wait
```

**Fix** — remove this line and update the `with` statement:
```python
with ThreadPoolExecutor(max_workers=n_threads) as ex:
    list(ex.map(worker, range(n_threads)))
```

treff7es (changes requested):
## Third-pass review — nearly there 🏁

All issues from rounds 1 and 2 are resolved. One new finding blocks merge, plus one optional suggestion.

### ✅ Resolved (rounds 1 & 2)

| ID | Issue |
|----|-------|
| B1 | Race on `float +=` in worker threads |
| B2 | `report=` not forwarded to retry helpers |
| B3 | Blanket `except Exception` in cursor-fallback path |
| B4 | Trivial config-default tests |
| N1 | `report.failure` fired on every successful retry |
| N11 | `try/except` around `yield` swallowed consumer errors |
| I1–I7 | Retry classification, pool counter semantics, per-DB connect isolation, report.failure on last attempt, thread-contention test, backoff-cap test, declarative field constraints |
| N2 | Test comment contradicted assertion |
| N4 | Trivial dataclass-defaults test |
| N6 | Missing upper bounds on retry/timeout fields |
| N10 | Three retry helpers duplicated ~150 lines |
| N12 | Over-broad permanent-error substrings |
| N13 | Permanent errors had no report breadcrumb |
| N16 | Substring filter could mis-classify cursor-mode errors |

---

### 🔴 S1 — Must fix: late `import` inside test method

See inline comment below. One-line fix — remove the late import and use `ThreadPoolExecutor` that is already in scope from the module-level `from concurrent.futures import ... ThreadPoolExecutor ...`.

---

### ℹ️ S2 — Optional: `get_inspectors` manual `__enter__`/`__exit__`

The manual context-manager dance is correct but non-idiomatic:

```python
conn_cm = self._retry_connect(engine)
conn = conn_cm.__enter__()
...
yield db_inspector
finally:
    conn_cm.__exit__(None, None, None)
```

A `with` block inside a generator works identically and is clearer:

```python
with self._retry_connect(engine) as conn:
    db_inspector = inspect(conn)
    db_inspector._datahub_database = db
    yield db_inspector
```

Not blocking — just a readability note.

---

Fix S1 → this is ready to merge.

metadata-ingestion/tests/unit/test_teradata_source.py
NehaGslab: Fixed the import.

treff7es (approved):

PR #17387, opened 2026-05-12
fix(search): skip search hits with missing/invalid URN instead of crashing

creazyfrog:
## Summary

Fixes #13181

When a document in Elasticsearch is missing the `urn` field — such as the `datahub-gc` bootstrap ingestion source document that was persisted in v1.0.0 — the call to `getResult(hit)` throws an exception. This exception propagates up and crashes the **entire** search/scroll operation, making the Ingestion Sources page completely fail to load for affected deployments.

### Root cause

`UrnExtractionUtils.extractUrnFromSearchHit()` throws when the `urn` field is absent or null. This exception was not caught in either:
- `getRestrictedResults()` — used by the standard search path
- `extractScrollResult()` — used by the scroll/pagination path

### Fix

Wrap the `getResult(hit)` call in a per-hit try-catch in both methods:
- `getRestrictedResults()`: switch from `.map(this::getResult)` to `.flatMap(...)` with try-catch, emitting an empty stream for bad hits
- `extractScrollResult()`: wrap in try-catch with `continue` to skip bad hits

Bad hits are skipped with a `warn`-level log including the index and document ID so operators can identify and clean up the offending documents.

### Impact

- **No regression**: valid documents are returned normally
- **Affected deployments**: Ingestion Sources page becomes functional again without requiring manual index cleanup
- **Operator visibility**: warning log identifies the bad document for optional cleanup

## Test plan

- [ ] Deploy against an Elasticsearch index containing a document missing the `urn` field
- [ ] Verify the Ingestion Sources page loads successfully and lists valid sources
- [ ] Verify a `WARN` log is emitted identifying the skipped document
- [ ] Verify existing unit tests for `SearchRequestHandler` still pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #17494, opened 2026-05-18
test(search): add unit tests for #17387 [CAT-2018]

alexjst:
Adds unit tests for the skip-bad-urn behavior added in #17387, which was merged without test coverage.

Two tests in `SearchRequestHandlerTest`:
- `testExtractResultSkipsHitsWithMissingOrInvalidUrn` — non-scroll path
- `testExtractScrollResultSkipsHitsWithMissingOrInvalidUrn` — scroll path

Each feeds a mocked response with one valid hit, one missing `urn`, and one with an unparseable URN, and asserts only the valid hit comes back.


alexjst:
Superseded by #17668, which includes this test coverage plus the actual follow-up fixes from the #17387 review: result truncation in `CacheableSearcher`, narrowed exception handling (`InvalidSearchHitException`), and improved skip logging + metric. Closing in favor of that PR.

Issue #4854, opened 2022-05-06
Can not log out when using OIDC


twuyts:
**Describe the bug**
I enabled OIDC authentication using Google. I can log in using my Google account. However, when I try to log out, I am immediately logged back in. There is no way of selecting a different account.

**To Reproduce**
1. Enable [Google Authentication](https://datahubproject.io/docs/how/auth/sso/configure-oidc-react-google/)
2. Log in
3. select `Logout` in the top right  dropdown menu
4. you are redirected to the homepage and still logged in

**Expected behavior**
I would expect a page stating that I was logged out successfully, preferably with a link allowing me to log back in. When logging back in, I should be able to select the previous account, or use a different one.

**Screenshots**
![logout](https://user-images.githubusercontent.com/3224432/167105519-4a111b9b-8e3a-42dc-aff8-f691b851e023.gif)

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04.4 LTS
 - Browser Firefox
 - Version 99.0 (64-bit)

**Additional context**
Same behaviour in other browsers.

twuyts:
might be related to https://github.com/datahub-project/datahub/issues/2735 ?

xiphl:
related to https://datahubspace.slack.com/archives/CV2UVAPPG/p1650916737226709.
i do get redirected to the keycloak login page in my case though.

jjoyce0510:
This there, can you please send the link for your Oidc Discovery URL? I want to confirm that it contains the "end_session_endpoint" param, which we use to log folks out when logout is clicked.

jjoyce0510:
I notice that there is a revoke token endpoint present in some cases, which is not being used. My thought it that fixing this will be a matter of supporting token revocation, but still doing some investigation

sunli1223:
I have a workaround solution, just only edit the code(change the href url)
```html
<li class="ant-dropdown-menu-item ant-dropdown-menu-item-danger ant-dropdown-menu-item-only-child sc-pNWdM XJxhY" role="menuitem" tabindex="0" data-menu-id="rc-menu-uuid-97193-1-logout"><a href="/">Sign Out</a></li>
```
 to 
```html
<li class="ant-dropdown-menu-item ant-dropdown-menu-item-danger ant-dropdown-menu-item-only-child sc-pNWdM XJxhY" role="menuitem" tabindex="0" data-menu-id="rc-menu-uuid-97193-1-logout"><a href="/logOut">Sign Out</a></li>
```

[...]

PR #17497, opened 2026-05-19
feat(auth): revoke session token on logout

### LINKED ISSUE #17068: Authentication gaps: logout token reuse, concurrent sessions, native login lockout, and independent SSO/DataHub sessions
## Summary

I reviewed DataHub authentication/session behavior in both non-SSO and SSO environments and found a few gaps that may matter for deployments with stricter security requirements.

This is not a report about a single crash or defect, but a request to confirm current behavior and discuss whether some of these should be improved or made configurable.

### 1. Logout does not appear to revoke the server-side session token
On logout, the browser session is cleared, but the previously issued DataHub session token does not appear to be explicitly revoked server-side.

As a result, if that token had already been copied or intercepted, it may remain usable until expiration.

### 2. Concurrent login appears to be allowed
A new login appears to issue a new session without invalidating previous sessions for the same user.

This means multiple active sessions for the same user are possible.

### 3. Native login failure threshold / lockout
For native authentication, I could not find a built-in configurable failed-login threshold / lockout policy.

This may be acceptable for SSO deployments where the IdP enforces lockout, but it seems like a gap for non-SSO/native login deployments.

### 4. SSO session and DataHub session are independent
When SSO is enabled, DataHub still manages its own application session separately from the SSO/IdP session.

As a result, expiration of the DataHub session does not necessarily mean the user is fully logged out. If the SSO/IdP session is still active, the user may be authenticated again automatically.

## Questions

1. Is the current logout behavior (clear browser session only, no explicit session-token revoke) intentional?
2. Is there already a recommended way to enforce single-session login?
3. Is failed-login threshold/lockout expected to be handled only by external IdPs?
4. Is the separation between SSO/IdP session and DataHub session the intended design?
5. Would maintainers be open to PRs for:
   - revoking session tokens on logout
   - optional single-session enforcement
   - configurable failed-login threshold for native authentication

## Suggested improvements

Potential improvements could be:

- revoke current DataHub session token on logout
- add optional single-session enforcement per user
- add configurable lockout / failed-login threshold for native authentication

sand116:
## Summary

- revoke the current UI session token during logout before clearing the Play session
- reject reuse of revoked session tokens in `StatefulTokenService`
- add focused controller/client/service tests for the revoke flow

## Related Issues

- closes #17068 (`[Feature] Improve Native Authentication Security (Account Lockout & Password Reset Cooldown)`)
- related to #4854 (`OIDC logout and token invalidation follow-up context`)

## Test Plan

- `git diff --check`
- Attempted: `./gradlew :metadata-service:auth-impl:test --tests "com.datahub.authentication.token.StatefulTokenServiceTest"`
- Local Gradle verification is blocked in this environment because the workspace is running on Java 17 while `metadata-models`, `metadata-service:auth-config`, and `metadata-io` require Java 21 variants

## Docs

- no user-facing docs required


sakethvarma397:
@sand116 Thanks for creating such a detailed issue and for this PR!

sand116:
@RyanHolstien could you take a look when you have a chance? I'd appreciate your review.

RyanHolstien:
There are several fundamental problems with this PR and I don't think we can accept it in the current state.

This will store a new entity for every single session a user creates without any mechanism for clean-up. Over time this will result in significant database bloat. 

This creates an entirely new entity model just for revoked tokens when a model for tokens already exists.

It makes a core flow change forcing through the stateful token flow which was designed just for separate PATs with no option for configuration.

This also seems like a generally non-standard way of handling sessions, stateless session tokens are a design choice in the current implementation and switching away from that needs some more intentional design. A side effect of their stateless nature is an inability to do a hard revocation. [Instead they are valid for a limited amount of time](https://docs.datahub.com/docs/authentication/concepts#what-is-a-datahub-token-service-what-are-access-tokens)  which is [configurable](https://docs.datahub.com/docs/authentication/guides/sso/configure-oidc-react#2-configure-datahub-frontend-server). A different typical approach that does allow limited revocation with shorter session token durations is usage of a refresh token. There are definitely reasonable arguments against this method of using JWTs as session tokens and I don't think we're necessarily against moving more towards stateful support as an option, but I don't think storing all the revoked sessions and still generating "stateless" (with stored revoked sessions) session tokens like we do is the right approach. The cardinality of active sessions is going to be much smaller than all revoked sessions in existence and it makes more sense to store from that side. Either way needs automatic clean up though.

sand116:
@RyanHolstien 

Thanks for the candid feedback. I reworked the PR around the concerns you raised.

- Default UI session behavior remains stateless.
- Added an opt-in `authentication.statefulSessionTokensEnabled` path for teams that want hard revocation support.
- Removed the revoked-session entity/model entirely.
- Reused the existing `dataHubAccessToken` model for stateful session tracking instead of creating a separate model.
- When enabled, we store active session tokens, not revoked-session tombstones.
- Logout only revokes stored v2 session tokens; stateless v1 session tokens remain a no-op on revoke.
- Expired-token cleanup now includes session tokens so stored session records are automatically cleaned up.

I also kept standard token management views focused on PAT/service tokens by excluding session tokens unless explicitly requested.

sand116:
@lakshay-nasa Hi! Could you take a look at this PR when you have a chance? Would really appreciate your review. Thanks!

metadata-service/configuration/src/main/resources/bootstrap_mcps/ingestion-datahub-gc.yaml
```diff
@@ -19,7 +19,7 @@
         source:
           type: 'datahub-gc'
           config:
-            cleanup_expired_tokens: {{cleanup_expired_tokens}}{{^cleanup_expired_tokens}}false{{/cleanup_expired_tokens}}
+            cleanup_expired_tokens: {{cleanup_expired_tokens}}{{^cleanup_expired_tokens}}true{{/cleanup_expired_tokens}}
```
RyanHolstien, line 22: Should this be dependent on enabling the stateful tokens? Or why is the default being changed for all cases? Don't necessarily think this is a bad thing to update, just curious.

metadata-service/auth-servlet-impl/src/main/java/com/datahub/auth/authentication/AuthServiceController.java
```diff
@@ -236,6 +244,42 @@ CompletableFuture<ResponseEntity<String>> generateSessionTokenForUser(
         });
   }
 
+  @PostMapping(value = "/revokeSessionToken")
+  CompletableFuture<ResponseEntity<Void>> revokeSessionToken(final HttpEntity<String> httpEntity) {
+    final String authorizationHeader = httpEntity.getHeaders().getFirst(HttpHeaders.AUTHORIZATION);
+    if (authorizationHeader == null || !authorizationHeader.startsWith(BEARER_PREFIX)) {
```
RyanHolstien: Doesn't check auth, but only allows revoking your own session token after validating so I think this is probably okay. Makes sense to always be able to revoke your own token. 

Missing header seems like it should result in UNAUTHORIZED though rather than BAD_REQUEST though.

metadata-service/configuration/src/main/resources/bootstrap_mcps/ingestion-datahub-gc.yaml
sand116: The change in default from `false` to `true` is independent of stateful session tokens. Even without stateful session tokens, expired PAT tokens are stored in the `dataHubAccessToken` entity and need cleanup. The previous default of `false` meant those were never being purged. This is a general improvement that benefits all installations.

metadata-service/auth-servlet-impl/src/main/java/com/datahub/auth/authentication/AuthServiceController.java
sand116: Good catch — split the check so a missing `Authorization` header returns `401 UNAUTHORIZED` (no credentials provided), while a malformed or non-Bearer value returns `400 BAD_REQUEST`. Fixed in 58fe405634 along with the corresponding test update.

PR #17500, opened 2026-05-19
feat(ingestion/unity-catalog): extend metric view with v1.1 composable lineage and agent metadata

dinesh-verma-datahub:
## Summary

Extends the Databricks Unity Catalog connector's existing Metric View ingestion with the v1.1 semantic-layer feature set so the catalog can serve as a full source of truth for Databricks metric views — including composable lineage and the agent-metadata fields used by Databricks Genie.

### Composable lineage

- **Within a single metric view** — measures defined in terms of other measures via the Databricks `MEASURE(name)` syntax now produce field-to-field lineage edges inside the metric view's own dataset. For example, an "average order value" measure expressed as `MEASURE(total_revenue) / MEASURE(order_count)` now has two upstream lineage edges to `total_revenue` and `order_count`. Matching is case-insensitive (per SQL convention), and the emitted column URN always uses the canonical-case name as defined in the spec — so referencing a measure with a different letter case still resolves to the correct field. References to measures that don't exist in the same view are reported as a warning and counted.
- **Across joined tables (snowflake-style joins)** — nested join hierarchies are now traversed recursively. Every level of the hierarchy contributes its join target as a distinct upstream of the metric view, and the alias defined at every level becomes resolvable in dimension and measure expressions. A column expression like `customer.nation.n_name` (referencing a nested join's table) correctly resolves the immediate alias to the right upstream column.

### Agent metadata (Databricks Runtime 17.3+, YAML 1.1)

Per-field metadata used by Databricks Genie now lands on the metric view dataset:

- **Display name** — a human-readable label for the field, truncated to 255 characters with an observability counter recording any truncation events.
- **Synonyms** — alternative names for the field. Capped at 10 entries (per the published spec), each truncated to 255 characters, with non-string and empty entries filtered out. An overflow counter records when the 10-entry cap fires.
- **Format** — the display-formatting spec for a measure. All six format types in the v1.1 spec are supported: `number`, `currency`, `percentage`, `byte`, `date`, `date_time`. Each subkey known for its format type lands as an individually queryable custom property, including nested fields (`decimal_places.type`, `decimal_places.places`). Subkeys not recognized for the given format type are dropped and recorded in an observability counter, so non-standard YAML doesn't silently slip through.

Each agent-metadata field is exposed as a dataset-level custom property keyed by lowercase field name. Display name is **not** stored on the schema field's deprecated `label` attribute (which the model documents as "not surfaced in the UI"); using a custom property keeps the value queryable and visible.

### Window measures

When a measure declares a non-empty `window` block:

- The measure's schema field is additionally tagged "Window Measure" (alongside the existing "Measure" tag).
- A single-entry window block surfaces `order`, `range`, and `semiadditive` as individually queryable custom properties.
- A multi-entry window block is preserved verbatim as a single JSON-encoded custom property so no data is lost.

### Spec-level surface

Additional dataset-level properties make the metric view spec inspectable without re-reading the YAML body:

- **Spec version** — surfaced directly so consumers can branch on it.
- **Filter expression** — already present from the base PR; now driven through a single shared property builder so it stays in sync with the rest.
- **Joins JSON** — the entire joins hierarchy is preserved as a JSON property, including the `on:` join predicate, `using:` shorthand, and nested joins. (See "YAML 1.2 loader" below for why the `on:` key needed special handling.)
- **Materialization** — when materialization is configured as an object (v1.1 form), its `schedule`, `mode`, and `materialized_views` list each become individually queryable. The legacy string form (`materialization: materialized`) is still recognized and continues to set the view's `materialized` flag.

### YAML 1.2 loader

PyYAML's default `safe_load` follows YAML 1.1, under which the bareword `on` (along with `off`, `yes`, `no`) is coerced to a boolean. Since Databricks metric view joins use `on:` as a key for the join predicate, parsing under YAML 1.1 would silently rename that key to `True` when downstream code serializes the joins structure — losing the predicate's identity.

A YAML loader matching YAML 1.2 semantics (only `true`/`false` spellings recognized as booleans) is now used for metric view definitions. The `on:` join predicate now round-trips intact through to the joins custom property.

### Dataset description prefers YAML

A metric view's top-level `comment` field is its own documentation. When present, it now drives the dataset description directly, falling back to Unity Catalog's table comment when absent. The two can drift on the Databricks side, and the YAML value is closer to the metric view author's intent.

## What's fixed

- **Field-level descriptions on v1.1 specs** — v1.1 renamed the per-dimension/measure description key from `description` to `comment`. Both spellings are now accepted, with `comment` taking precedence so v1.1 specs no longer lose their descriptions.
- **Materialization recognition on v1.1 specs** — v1.1 changed the `materialization` value from a string to an object. The view's `materialized` flag is now correctly set in both forms; the v0.1 string form remains supported.

## New observability counters

Four new counters surface in the ingestion report so unusual specs are visible rather than silent:

- Unresolved `MEASURE()` references — how many composable-measure references couldn't be resolved within the same view.
- Display-name truncations — how many display names exceeded 255 characters.
- Synonym overflows — how many fields had more than 10 synonyms.
- Unknown format subkeys — how many `format` blocks contained subkeys not recognized for their format type (including unknown nested `decimal_places` subkeys).

## Tests

- **Unit tests** — broad coverage of every new behavior including: YAML 1.2 round-trip of the `on:` key, all format types and subkey combinations, nested `decimal_places` dict flattening, synonym overflow, display-name truncation, window measure tagging (including the empty-list edge case), composable measure lineage with case-mismatched references, nested join lineage, top-level `comment` preference, and the bug-fix behavior for both `comment` vs `description` and dict vs string materialization. Includes one regression test specifically asserting that the joins custom property preserves the `on:` key as a string (would have caught the boolean coercion bug).
- **Integration tests** — the existing mock metric view fixture is upgraded to a richer v1.1 YAML that exercises the full agent-metadata surface (display name, synonyms, currency format, window measure, composable measure, snowflake join with `on:` predicate, v1.1 materialization object). Golden files regenerated.
- **Golden file refresh** — also resolves a pre-existing drift on master where the metric view goldens hadn't been refreshed after the recent change to default-emit dataset properties as full upserts.


metadata-ingestion/src/datahub/ingestion/source/unity/source.py
```diff
@@ -1634,6 +1711,62 @@ def _extract_metric_view_column_lineage(
...
+                            downstreamType=FineGrainedLineageDownstreamType.FIELD,
+                            downstreams=[downstream_field],
+                        )
+                    )
+                else:
+                    unresolved_refs.add(ref_name)
+
+        if unresolved_refs:
+            self.report.num_metric_view_unresolved_measure_refs += len(unresolved_refs)
+            self.report.report_warning(
```
sgomezvillamor: use `report.warning` with proper `title`, `message`, `context`, ...

sgomezvillamor:
Great PR — excellent test coverage and the YAML 1.2 loader fix is particularly well-handled. A few items before merge:

---

🔴 **BLOCKER (please confirm)** — `MEASURE()` expressions generate spurious cross-dataset field edges

When sqlglot parses an expression like `MEASURE(total_revenue) / MEASURE(order_count)`, it extracts `total_revenue` and `order_count` as unqualified `Column` nodes and resolves them against the source table via `alias_map[""]`. This produces wrong edges like `orders.total_revenue + orders.order_count → revenue_metrics.aov` — those fields don't exist on `orders`, so they'd be dangling URNs in DataHub.

This is visible in `metric_view_mces_golden.json` alongside the correct intra-MV edges emitted by `_extract_intra_mv_measure_lineage`.

Suggested fix in `_extract_metric_view_column_lineage`, before calling `_parse_metric_view_expr`:
```python
clean_expr = _MEASURE_REF_RE.sub("0", expr)
```
This strips `MEASURE()` calls before SQL parsing — the correct intra-MV edges already handle them. **Please confirm whether this is intentional or an oversight.**

---

🟡 **WARNING** — `unity-catalog_post.md` not updated

The docs still say descriptions come from `dimensions[].description` / `measures[].description`, but `comment` now takes precedence. None of the new v1.1 features are documented (composable lineage, agent metadata custom properties, `Window Measure` tag, nested join traversal, new `metric_view.*` custom properties). Worth adding an `#### Metric Views v1.1` subsection under `### Capabilities`.

---

ℹ️ **SUGGESTION** — non-string synonym entries dropped silently

In `_add_synonyms_props`, invalid-type entries are filtered without incrementing any counter, unlike all other truncation/overflow paths in this PR. A small counter increment would keep the observability story consistent.


---

sgomezvillamor (commented):
left some comments

dinesh-verma-datahub:
@sgomezvillamor resolved all the above issues.

sgomezvillamor (approved):
LGTM 🎖️

PR #17503, opened 2026-05-19
feat(ingest/dbt): add `add_structured_property` operation to meta_mapping

alfiyas-datahub:
## Summary

- Adds a new `add_structured_property` operation to `meta_mapping` and `column_meta_mapping` so dbt `meta` fields can be mapped natively to DataHub Structured Properties during ingestion — no second pipeline, no race condition.
- Implemented in the shared `OperationProcessor` (`metadata-ingestion/src/datahub/utilities/mapping.py`), so every source that consumes `meta_mapping` (dbt-core, dbt-cloud, glue, snowflake, etc.) gets it for free.
- Wired into the dbt source for both **dataset-level** and **column-level** emission, including a new helper that produces standalone MCPs for `schemaField` URNs (since `StructuredProperties` is not part of `DatasetSnapshot`).

### New `meta_mapping` operation

```yaml
source:
  type: dbt
  config:
    meta_mapping:
      data_load_frequency:
        match: ".*"
        operation: "add_structured_property"
        config:
          structured_property_urn: "urn:li:structuredProperty:io.acme.data_load_frequency"
      retention_days:
        match: ".*"
        operation: "add_structured_property"
        config:
          structured_property_urn: "urn:li:structuredProperty:io.acme.retention_days"
          value_type: "number"
    column_meta_mapping:
      classification:
        match: ".*"
        operation: "add_structured_property"
        config:
          structured_property_urn: "urn:li:structuredProperty:io.acme.column_classification"
```

### Capabilities

- Works at **both** dataset and column level (column-level rules emit standalone MCPs on `schemaField` URNs)
- `value` defaults to the raw meta value (preserves `int` / `float` types — important for `double` PrimitivePropertyValue)
- `value` accepts an explicit literal or a template with `{{ $match }}` substitution
- `value_type: number` for explicit numeric coercion of string inputs
- List values are supported (multi-value structured properties)
- Multiple rules targeting the same property URN are aggregated into a single value-list assignment, with duplicates dropped
- Invalid URNs and missing config are logged and the rule is skipped — never breaks the run
- The `datahub:` shorthand bag also accepts `structured_properties: {urn-or-id: value}` for inline definitions

### Prerequisite for users

The Structured Property definition itself must already exist in DataHub (created via UI, CLI, or YAML). This operation only **assigns values** — same as `add_tag` / `add_term` today, which don't dynamically create tag/term entities either.


metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
```diff
@@ -3104,6 +3123,47 @@ def get_schema_metadata(
...
+    def _create_column_structured_property_mcps(
+        self, node: DBTNode, dataset_urn: str
+    ) -> Iterable[MetadataChangeProposalWrapper]:
+        """Emit a StructuredProperties MCP for each column with a matching
+        column_meta_mapping `add_structured_property` rule. The aspect attaches
+        to the column's schemaField URN, not the dataset URN."""
+        if not self.config.column_meta_mapping:
+            return
+
+        action_processor = OperationProcessor(
```
puneetagarwal-datahub: dont process this again as we have already done in get_schema_metadata

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
```diff
@@ -3104,6 +3123,47 @@ def get_schema_metadata(
...
+            match_nested_props=True,
+        )
+
+        for column in node.columns:
+            if not column.meta:
+                continue
+            try:
+                meta_aspects = action_processor.process(column.meta)
+            except Exception as e:
+                logger.warning(
```
puneetagarwal-datahub: use self.report.warning(...)

puneetagarwal-datahub (dismissed):

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
treff7es: **Duplicate work (DRY + perf).** `get_schema_metadata` (above, ~L3040) already builds a byte-for-byte identical `OperationProcessor(column_meta_mapping, …)` and calls `.process(column.meta)` on every column to extract tags/terms — it just discards the `add_structured_property` aspect. This method rebuilds the same processor and runs `.process(column.meta)` a **second time** on every column. For a project with many columns + `column_meta_mapping` set, that doubles the regex work. The field-name lowercasing is duplicated too (L3158-3160 vs L3088-3090).

Suggest capturing `(schema_field_urn, sp_aspect)` during the existing `get_schema_metadata` loop and yielding from that, or at minimum hoisting the processor construction out of the per-node path (its args are constant for the run).

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
treff7es: This is new code in a class where `self.report` is available — prefer `self.report.warning(title=…, message=…, context=f"{node.dbt_name}.{column.name}")` over `logger.warning` so a column whose structured property fails to emit surfaces in the ingestion summary rather than only in raw logs (matches the convention used elsewhere in this file).

metadata-ingestion/src/datahub/utilities/mapping.py
```diff
@@ -273,6 +334,20 @@ def process(self, raw_props: Mapping[str, Any]) -> Dict[str, Any]:
...
 
+    def _add_datahub_structured_properties_to_map(
+        self,
+        raw_properties: Dict[str, _StructuredPropertyValueT],
+        operations_map: Dict[str, list],
+    ) -> None:
+        """Translate `datahub.structured_properties` shorthand entries into the
+        flat `{"urn": ..., "value": ...}` form that `convert_to_aspects` consumes."""
+        for prop_id, prop_value in raw_properties.items():
+            urn = make_structured_property_urn(prop_id)
```
treff7es: `make_structured_property_urn(prop_id)` is called bare here, inside `process()`'s first `try` whose handler is a generic `logger.error("Error while processing datahub property…")`. One typo'd key in the `datahub.structured_properties` bag (e.g. `urn:li:tag:…`) raises and aborts the loop, so the **remaining** structured properties in the bag are silently dropped, with a log that names neither the property nor that structured properties were the cause.

(Tags/terms/owners/domain are added before this block, so those survive — the loss is scoped to this bag.) Mirror the `get_operation_value` branch: wrap per-property in try/except, `logger.warning` naming `prop_id`, and `continue`. That also makes the two config paths behave consistently.

metadata-ingestion/src/datahub/utilities/mapping.py
```diff
@@ -444,6 +549,57 @@ def get_operation_value(
...
+                    raw_value = float(raw_value)
+                except ValueError:
+                    logger.warning(
+                        f"add_structured_property rule '{operation_key}' configured with "
+                        f"value_type=number but value '{raw_value}' is not numeric."
+                    )
+                    return None
+
+            coerced_values = _coerce_structured_property_values(raw_value)
+            if not coerced_values:
```
treff7es, line 630: The rule matched (`maybe_match is not None`) but produced no usable value, so it's dropped here with no log — from the user's view the property should have been assigned. A `logger.warning` naming `operation_key` and the resolved URN turns a silent no-op into a debuggable signal. Same applies to the malformed/`value is None` `continue`s in `convert_to_aspects`.

metadata-ingestion/src/datahub/utilities/mapping.py
```diff
@@ -320,6 +395,35 @@ def convert_to_aspects(self, operation_map: Dict[str, list]) -> Dict[str, Any]:
...
+                values_list = grouped_values.setdefault(urn, [])
+                if value not in values_list:
+                    values_list.append(value)
+
+            if grouped_values:
+                structured_properties_aspect = StructuredPropertiesClass(
+                    properties=[
+                        StructuredPropertyValueAssignmentClass(
+                            propertyUrn=urn,
+                            values=cast(List[Union[str, float]], values_list),
```
treff7es: Minor: this `cast(...)` is redundant — `values_list` is already typed `List[Union[str, float]]` (from `grouped_values`). More broadly, the `{"urn", "value"}` record is produced in two sites and re-validated defensively here; a small `TypedDict`/`NamedTuple` would let mypy verify both producers and the consumer, drop the defensive `isinstance`/`.get()` checks, and make this cast unnecessary (per the repo's "prefer typed objects over loose dicts" standard).

metadata-ingestion/tests/unit/dbt/test_dbt_source.py
```diff
@@ -3829,6 +3831,276 @@ def test_dbt_common_semantic_model_subtype_assignment():
...
+    assert len(sp_aspects) == 1
+    assignment = sp_aspects[0].properties[0]
+    assert (
+        assignment.propertyUrn
+        == "urn:li:structuredProperty:io.acme.data_load_frequency"
+    )
+    assert list(assignment.values) == ["hourly"]
+
+
+def test_dbt_meta_mapping_add_structured_property_disabled_when_meta_mapping_off():
```
treff7es, line 3912: This test doesn't actually exercise the flag it names. It passes `meta_aspects={}`, so the guard `if meta_structured_properties_aspect and self.config.enable_meta_mapping` short-circuits on the empty dict (`…aspect` is `None`) **before** `enable_meta_mapping` is ever checked — the test would still pass if `and self.config.enable_meta_mapping` were deleted. To test the guard, pass a non-empty `meta_aspects` containing a structured-property aspect with the flag off and assert it's dropped.

metadata-ingestion/tests/unit/dbt/test_dbt_source.py
```diff
@@ -3829,6 +3831,276 @@ def test_dbt_common_semantic_model_subtype_assignment():
...
+        dbt_file_path="/models/orders.sql",
+        dbt_package_name="my_project",
+        catalog_type=None,
+        missing_from_catalog=False,
+    )
+
+
+def test_dbt_meta_mapping_add_structured_property_model_level():
+    """meta_mapping with add_structured_property emits a StructuredProperties aspect on the dataset."""
+    from datahub.utilities.mapping import OperationProcessor
```
treff7es: Late import inside a test. `OperationProcessor` isn't imported at module top and there's no circular-import reason (`dbt_common.py` imports it at module scope), so this is flagged as a blocker by the repo's testing conventions (`AGENTS.md` / `testing.md`). Move it to the top of the file.

metadata-ingestion/src/datahub/utilities/mapping.py
```diff
@@ -88,18 +132,27 @@ class Constants:
     TAG_DIST_KEY = "DIST_KEY"
     TAG_SORT_KEY = "SORT_KEY"
     SEPARATOR = "separator"
+    STRUCTURED_PROPERTY_URN = "structured_property_urn"
+    STRUCTURED_PROPERTY_VALUE = "value"
+    STRUCTURED_PROPERTY_VALUE_TYPE = "value_type"
+    STRUCTURED_PROPERTY_VALUE_TYPE_STRING = "string"
+    STRUCTURED_PROPERTY_VALUE_TYPE_NUMBER = "number"
```
treff7es, line 139: What's up with the date type, rich text type, or was it not a requirement? 
What was the exact request?

treff7es: Image

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
```diff
@@ -3104,6 +3175,29 @@ def get_schema_metadata(
...
+        if not self.config.column_meta_mapping:
+            return
+        if column_meta_aspects is None:
+            column_meta_aspects = self._extract_column_meta_aspects(node)
+
+        for field_name, meta_aspects in column_meta_aspects.items():
+            sp_aspect = meta_aspects.get(Constants.ADD_STRUCTURED_PROPERTY_OPERATION)
+            if not sp_aspect:
+                continue
+            yield MetadataChangeProposalWrapper(
```
treff7es, line 3201: I guess this assumes the structured property already exists, right? 
Is it ok to fail if a structured property does not exist on the server side, or should we do validation before setting structured properties for existence?
If we need it, we will have to precache the structured properties. With that, we can validate if we want to set an invalid type or if structured property is missing.

metadata-ingestion/src/datahub/utilities/mapping.py
treff7es: CleanShot 2026-06-19 at 20 44 22@2x

alfiyas-datahub: value_type is only used as a hint for client-side conversion. Since structured property values are stored as either strings or numbers, number is the only case that requires special handling (string → float). DATE, URN, and RICH_TEXT are all stored as strings and already work through the default path. The actual type validation is handled server-side by StructuredPropertiesValidator. For the Nordstrom use case, all requested fields were string-based, so adding number covered the only non-trivial conversion case.

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
alfiyas-datahub: yeah, this feature only assigns structured property values and assumes the property definition already exists in DataHub. Validation isn't skipped, GMS still checks that the property exists and validates its type, cardinality, and allowed values when the write happens. I avoided adding client-side validation because it would require a graph connection and would break graph-less use cases like file sinks, dry runs, and CI tests.

treff7es (approved):

PR #17505, opened 2026-05-19
feat(docker): replace head tag with quickstart and sha-* tags

david-leifker:
## Summary

DataHub master builds used to publish a floating `:head` tag on every image. Each service was tagged as its own build finished, so a `docker compose pull` at any moment could pull a **mixed set of commits** across GMS, frontend, upgrade, and so on. That is awkward for local quickstart and unsafe for production-style deployments (Kubernetes, Helm), where every pod in a rollout should share one known build.

This PR replaces that model with two explicit tag types:

- **`sha-<short_sha>`** — immutable, per-commit tags published during the build (for pinning in K8s/Helm and production).
- **`quickstart`** — a coordinated floating tag applied **only after** the full image matrix passes smoke tests on `master`, so all services move together for Compose and `datahub docker quickstart`.

The legacy `:head` registry tag is removed. Bare short SHA tags (`:abc1234`) become `:sha-abc1234` for clarity.

The CLI also gains **confirmation before silent version fallbacks**, so typos and stale cached manifests are less likely to pull the wrong images without the user noticing.

## Why

| Problem with `:head` | What we do instead |
| --- | --- |
| Images tagged at different times → mismatched commits across services | `quickstart` retagged once per green master build (post-smoke) |
| Floating tag unsuitable for K8s rollouts that can happen anytime | Pin `sha-*` or release `v*` in clusters |
| Short SHA tags ambiguous (`abc1234` vs other identifiers) | Prefix commit tags as `sha-abc1234` |
| Unrecognized `--version` or deprecated `head` mapping applied silently | Interactive y/n prompt (or explicit opt-in for scripts) before fallback |

## User-facing changes

**Docker Compose / local quickstart**

- Compose defaults: `DATAHUB_VERSION=quickstart` (was `head`).
- `datahub docker quickstart --version head` still works: the quickstart version manifest maps `head` → `docker_tag: quickstart` + `composefile_git_ref: master`, so older CLIs pick this up from GitHub without a new pip release (cached local manifest may still say `head` until refetched).
- Explicit `--version quickstart` is supported via the manifest.
- Pin a specific master build: `DATAHUB_VERSION=sha-<7-char>` (registry-only; not a git tag).
- **Omitting `--version`** still uses the manifest `default` entry with **no prompt**.
- **Unrecognized non-release `--version` values** (e.g. typos like `hed`) prompt before falling back to the suggested configuration (pinned `default` from the manifest, or `master` + `quickstart` for missing magic aliases like `head`).
- **Release-like tags** (`v1.2.0`) and **`sha-*`** continue to pass through as git ref + docker tag without prompting.
- **`--accept-version-default`** skips the prompt for scripts/CI (non-interactive runs without this flag abort with a clear error when confirmation would be required).

**Kubernetes / Helm / production**

- Use release tags (`v*`) or immutable `sha-*` tags.
- Do **not** use `quickstart` in cluster manifests; it changes on the next green `master` build.

## Breaking change

Documented in `docs/how/updating-datahub.md`:

- `:head` is no longer published.
- `:abc1234` → `:sha-abc1234`.
- `:quickstart` added for coordinated local/Compose use only.


craig-rueda (approved):
lgtm

metadata-ingestion/src/datahub/cli/quickstart_versioning.py
```diff
@@ -150,6 +152,20 @@ def get_quickstart_execution_plan(
             ),
         )
 
+        if requested_version == "head" and "head" not in self.quickstart_version_map:
```
devashish2203: Will there be a plan to remove this code after a certain release?

devashish2203 (approved):

metadata-ingestion/src/datahub/cli/quickstart_versioning.py
david-leifker: Let me clean this up. No reference to head in the code. If not matching, we'll prompt for reverting to default config, allowing the user to correct a typo version or accept the default configuration using quickstart tag (as defined in the mapping)

PR #17516, opened 2026-05-20
fix(document): gate lastIngested to external documents only

shirshanka:
## Summary

- **`DocumentMapper.java`**: The `lastObserved` fallback (used when DataHub writes with `DEFAULT_RUN_ID`) now only fires when the document's source type is `EXTERNAL`. Native documents will never receive a `lastIngested` value from the mapper, so "Last Synced" stays hidden for them.
- **`LastIngestedProperty.tsx`**: Added an `isExternal` guard as a second layer of defence — the property renders empty for non-external entities even if `lastIngested` is somehow populated.
- **`DocumentMapperTest.java`**: Added 13 tests covering the new `lastIngested` logic (external fallback, native no-op, no-source no-op, no systemMetadata) and previously untested mapper paths (`DocumentSettings`, `BrowsePathsV2`, empty `SubTypes`/`Domains`, external source without URL/ID).
- **`LastIngestedProperty.test.tsx`**: New test file with 4 cases covering external vs native visibility and null/missing value handling.

## Context

`fe8f8c0456` shipped the "Last Synced" property for external documents. The backend mapper populated `lastIngested` from `SystemMetadataUtils`, but that method filters out `DEFAULT_RUN_ID` — the run ID DataHub uses when writing documents via the UI/API. This PR adds the fallback to `lastObserved` for that case, while ensuring the fallback (and the frontend property) only applies to external documents.

## Test plan

- [ ] `./gradlew :datahub-graphql-core:test --tests "com.linkedin.datahub.graphql.types.knowledge.DocumentMapperTest"` — 31 tests pass
- [ ] `yarn test src/app/entityV2/summary/properties/property/properties/__tests__/LastIngestedProperty.test.tsx --run` — 4 tests pass
- [ ] Verify "Last Synced" shows on an external (Confluence/Notion) document in the UI
- [ ] Verify "Last Synced" does not appear on a native document


annadoesdesign (approved):
FE looks good

PR #15691, opened 2025-12-16
feat(hive): Add Thrift connection mode with Kerberos support to hive-metastore connector

max-datahub:
# feat(hive): Add Thrift connection mode with Kerberos support to hive-metastore connector

## Summary

Extends the `hive-metastore` connector to support direct Thrift API access via a new `connection_type: thrift` configuration option. This enables metadata extraction from Hive Metastore with Kerberos/SASL authentication, addressing enterprise environments where direct database access is unavailable but Thrift API access (port 9083) with Kerberos authentication is possible.

## ⚠️ Breaking Change: Deprecated WHERE Clause Options

**The following configuration options have been deprecated and will raise an error if used:**
- `tables_where_clause_suffix`
- `views_where_clause_suffix`  
- `schemas_where_clause_suffix`

These options allowed arbitrary SQL fragments to be injected into queries, creating a SQL injection vulnerability. They have been removed for security reasons.

**Migration:** Use `database_pattern` and `table_pattern` for filtering instead:

```yaml
# Before (DEPRECATED - will error)
source:
  type: hive-metastore
  config:
    tables_where_clause_suffix: 'AND d."name" in (''production_db'')'

# After (RECOMMENDED)
source:
  type: hive-metastore
  config:
    database_pattern:
      allow:
        - "^production_db$"
```

## Motivation

The existing `hive-metastore` connector requires direct SQL access to the HMS backend database (PostgreSQL/MySQL). Many enterprise Hadoop deployments:
- Only expose the Thrift API (port 9083)
- Require Kerberos authentication
- Don't allow direct database access for security reasons

This extension fills that gap by adding a Thrift connection mode with SASL/GSSAPI authentication.

## Configuration

```yaml
# SQL mode (default, existing behavior)
source:
  type: hive-metastore
  config:
    # connection_type: sql  # default
    host_port: "hive-metastore-db:5432"
    # ... SQL connection settings

# Thrift mode (NEW)
source:
  type: hive-metastore
  config:
    connection_type: thrift
    host_port: "hive-metastore:9083"
    use_kerberos: true
    kerberos_service_name: "hive"
    # Optional: kerberos_hostname_override for load balancer scenarios
    # Optional: catalog_name for HMS 3.x multi-catalog deployments
```

## Architecture

The implementation uses a **composition-based design** with the Protocol pattern for clean separation of concerns:

```
HiveMetastoreSource (unified entry point)
    │
    ├── HiveDataFetcher Protocol    ─── Abstracts data access
[...]

Issue #10094, opened 2024-03-21
Custom Properties are deleted when a dataset is re-ingested


BrendanPC:
**Describe the bug**
After adding custom properties to a dataset, the properties all disappear the next time the ingestion runs for that dataset.

**To Reproduce**
1. Ingest a dataset. (I have verified the bug with source types Kafka and Clickhouse, but I believe it is independent of source type)
2. Add custom properties to the dataset. [Example code adding properties](https://github.com/datahub-project/datahub/blob/9659d6086739990dff7aa4551361b30a9e7993ee/metadata-ingestion/examples/library/dataset_add_properties.py#L37)
3. Verify that the custom properties have appeared in the DataHub UI for that dataset.
4. Trigger an ingestion run of the dataset.
5. Observe that all the custom properties have disappeared.

**Expected behavior**
Custom properties should not disappear.

**Setup:**
 - DataHub v.0.12.1

ylwi:
We experience this with a Postgres source and we can't find this behaviour documented anywhere.

As mentioned in [this question on Slack](https://datahubspace.slack.com/archives/CUMUWQU66/p1715164498576389) there's an option in another source type ([Hive Metastore](https://datahubspace.slack.com/archives/CUMUWQU66/p1715164498576389)) to enable properties merge. It would be useful to have a similar option for other sources.

The workaround [mentioned in the same Slack thread](https://datahubspace.slack.com/archives/CUMUWQU66/p1717181871587249?thread_ts=1715164498.576389&cid=CUMUWQU66) about using a transformer works for us:
```
transformers:
  - type: "simple_add_dataset_properties"
    config:
      semantics: PATCH
      properties: {}
```

PR #10086, opened 2024-03-20
fix(ingestion/unity-catalog): patch owners and properties

dushayntAW:
## Checklist

- [ ] The PR conforms to DataHub's [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [Commit Message Format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#commit-message-format))
- [ ] Links to related issues (if applicable)
- [ ] Tests for the changes have been added/updated (if applicable)
- [ ] Docs related to the changes have been added/updated (if applicable). If a new feature has been added a Usage Guide has been added for the same.
- [ ] For any breaking change/potential downtime/deprecation/big changes an entry has been made in [Updating DataHub](https://github.com/datahub-project/datahub/blob/master/docs/how/updating-datahub.md)

PR #9293, opened 2023-11-22
fix(ingest/redshift): patch instead of replace redshift custom properties

ethan-cartwright:
## Testing
* added tests
* tested this locally to ensure it works for:
    * custom properties
    * that table descriptions and other `set` methods function as expected
    * the `patch_custom_properties` flag works as expected

## Checklist

- [ ] The PR conforms to DataHub's [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [Commit Message Format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#commit-message-format))
- [ ] Links to related issues (if applicable)
- [ ] Tests for the changes have been added/updated (if applicable)
- [ ] Docs related to the changes have been added/updated (if applicable). If a new feature has been added a Usage Guide has been added for the same.
- [ ] For any breaking change/potential downtime/deprecation/big changes an entry has been made in [Updating DataHub](https://github.com/datahub-project/datahub/blob/master/docs/how/updating-datahub.md)

PR #17527, opened 2026-05-20
feat(ingest/glue): add PATCH mode for dataset properties

slydon:
## Summary

- Add `incremental_properties` config (default `false`) to the Glue connector that emits dataset properties as incremental PATCH operations instead of full overwrites
- Custom properties added via the DataHub API, SDK, or enrichment pipelines now survive Glue re-ingestion when enabled
- Uses the shared `IncrementalPropertiesConfigMixin` and `auto_incremental_properties` workunit processor, following the same pattern established by [Redshift (#9293)](https://github.com/datahub-project/datahub/pull/9293), [Unity Catalog (#10086)](https://github.com/datahub-project/datahub/pull/10086), and [Hive Metastore (#15691)](https://github.com/datahub-project/datahub/pull/15691)

## Motivation

When external processes enrich Glue datasets with custom properties (e.g., ownership annotations, data quality scores), re-running Glue ingestion would overwrite those properties with only the values from the Glue catalog. This is the same problem addressed by
[datahub-project/datahub#10094](https://github.com/datahub-project/datahub/issues/10094) (closed as stale without a fix).

The Glue connector is the last major AWS source without PATCH support for dataset properties.

## Changes

- **`metadata-ingestion/src/datahub/ingestion/source/aws/glue.py`**: Added `IncrementalPropertiesConfigMixin` to `GlueSourceConfig`; registered `auto_incremental_properties` as a workunit processor that converts dataset properties to PATCH MCPs when `incremental_properties: true`
- **`metadata-ingestion/tests/unit/glue/test_glue_source.py`**: Added `test_incremental_properties_emits_patch_mcps` verifying patch MCPs with correct `changeType` and JSON Patch operations; existing tests explicitly set `incremental_properties=False` to preserve golden file compatibility
- **`metadata-ingestion/docs/sources/glue/glue_post.md`**: Documented the new capability with usage example
- **`docs/how/updating-datahub.md`**: No breaking change entry needed — default is `false`, preserving existing behavior

## Test plan

- [x] All existing Glue unit tests pass
- [x] New test verifies patch MCPs are emitted with correct JSON Patch operations and `changeType == "PATCH"`
- [x] New test verifies `DatasetPropertiesClass` is absent from dataset snapshots in patch mode
- [x] `ruff check`, `ruff format`, and `mypy` all pass
- [x] Tests work without AWS credentials (mocked via botocore Stubber)
- [ ] Manual validation with a Glue catalog (set custom properties via API, run ingestion, confirm they persist)

## Checklist

- [x] Links to related issues
- [x] Tests added
- [x] Docs added
- [x] ~~Breaking change entry~~ Not needed — defaults to `false`, no behavior change for existing users


metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
```diff
@@ -287,6 +288,11 @@ class GlueSourceConfig(
         ),
     )
 
+    enable_properties_merge: bool = Field(
```
alokr-dhub: **`default=True` should be `default=False`**

The rest of the codebase uses opt-in for this feature:
- `IncrementalPropertiesConfigMixin` → `incremental_properties: bool = False`  
- Snowflake and Unity Catalog (post-#16873) → `default=False`

Shipping `default=True` silently changes behavior for every existing Glue user on upgrade. Properties removed from Glue will also no longer be removed from DataHub in PATCH mode — a non-obvious side-effect not covered in the docs.

`default=False` with opt-in is consistent with the mixin convention and makes this a non-breaking change 

Nit: aligning the field name to `incremental_properties` would match `IncrementalPropertiesConfigMixin` used by Unity Catalog and Snowflake, avoiding a third naming convention alongside `patch_custom_properties` (Redshift) and `incremental_properties` (Unity/Snowflake).

alokr-dhub (changes requested):
Thanks for the PR. The motivation is correct and the prior art in Redshift/Hive/Unity is a good reference. I have six inline comments below, two of which are blockers before merge:

1. **Default  should be ** — matches the codebase convention, avoids a breaking change, and avoids a silent surprise for users who remove properties from Glue.
2. **No-op test assertion** — the snapshot exclusion check always passes regardless of whether the bug is present or not.

The rest are suggestions to improve correctness and coverage.

metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
```diff
@@ -1233,13 +1239,11 @@ def get_all_databases_and_tables(
         return all_databases, all_tables
 
     def get_lineage_if_enabled(
-        self, mce: MetadataChangeEventClass
+        self,
+        mce: MetadataChangeEventClass,
+        dataset_properties: Optional[DatasetPropertiesClass],
```
alokr-dhub: **`Optional` annotation is misleading here**

The sole call site passes `dataset_properties` from `self._get_dataset_properties(table)`, which always returns a non-`None` `DatasetPropertiesClass`. The `Optional` was justified in the old code because it was extracted from the MCE (which can return `None`) — but that extraction is now gone.

Declaring it `Optional` invites future callers to pass `None` as a silent "skip lineage" signal, which would drop S3/Iceberg lineage with no log output or warning.

Suggest: `dataset_properties: DatasetPropertiesClass` (non-optional).

metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
```diff
@@ -1738,12 +1742,21 @@ def _extract_record(
...
             urn=dataset_urn,
-            aspects=[
-                Status(removed=False),
-                dataset_properties,
-            ],
+            aspects=[Status(removed=False)],
         )
 
+        # Emit dataset properties as patch or full aspect
+        if self.source_config.enable_properties_merge:
```
alokr-dhub: **Missing error handling — `datasetProperties` silently dropped on any failure**

`create_dataset_props_patch_builder` or `patch_builder.build()` can raise (e.g., a non-JSON-serializable value in `customProperties`). A failure leaves the dataset in DataHub with no name, description, or custom properties — with nothing surfaced in the ingestion report.

Suggest either:
1. Wrapping in `try/except` with a fallback to `dataset_snapshot.aspects.append(dataset_properties)` and a `self.report.report_warning()` call, or
2. Using the `auto_incremental_properties` workunit processor referenced by Redshift (cleaner, avoids duplicating the debt).

metadata-ingestion/tests/unit/glue/test_glue_source.py
```diff
@@ -1559,6 +1563,82 @@ def test_glue_redact_job_script_secret_fields(secret_name):
...
+        if wu.metadata
+        and isinstance(wu.metadata, models.MetadataChangeEventClass)
+        and wu.metadata.proposedSnapshot
+    ]
+    for wu in mce_wus:
+        assert isinstance(wu.metadata, models.MetadataChangeEventClass)
+        snapshot = wu.metadata.proposedSnapshot
+        if snapshot and hasattr(snapshot, "aspects"):
+            for aspect in snapshot.aspects:
+                assert (
```
alokr-dhub: **This assertion is a no-op — it always passes**

`DatasetPropertiesClass` inherits from `DictWrapper`, not from `dict`. So `isinstance(aspect, dict)` is `False` for every typed Avro aspect class, making `not isinstance(aspect, dict)` always `True` and the whole condition vacuously pass — even if `DatasetPropertiesClass` is still in the snapshot.

The primary invariant of this feature (no double-emission of `datasetProperties` as both a PATCH and a full UPSERT in the MCE snapshot) is completely unguarded.

Replace with:
```python
for aspect in snapshot.aspects:
    assert not isinstance(aspect, models.DatasetPropertiesClass), (
        "DatasetPropertiesClass should not be in MCE snapshot in PATCH mode"
    )
```

metadata-ingestion/tests/unit/glue/test_glue_source.py
```diff
@@ -1559,6 +1563,82 @@ def test_glue_redact_job_script_secret_fields(secret_name):
...
+    ]
+    assert len(patch_wus) > 0, "Expected patch MCPs for datasetProperties"
+
+    # Verify patch MCPs contain JSON Patch operations
+    for wu in patch_wus:
+        assert isinstance(wu.metadata, models.MetadataChangeProposalClass)
+        assert wu.metadata.aspect is not None
+        patch_ops = json.loads(wu.metadata.aspect.value)
+        assert isinstance(patch_ops, list)
+        assert all("op" in op and "path" in op for op in patch_ops)
```
alokr-dhub, line 2063: **Missing `changeType == "PATCH"` check**

The test verifies that the patch ops list is well-formed but not that the MCP is actually a PATCH operation. A regression where `create_dataset_props_patch_builder` emits `UPSERT` instead of `PATCH` MCPs would go undetected.

Add to the per-`wu` loop:
```python
assert wu.metadata.changeType == "PATCH"
```

metadata-ingestion/tests/unit/glue/test_glue_source.py
```diff
@@ -1559,6 +1563,82 @@ def test_glue_redact_job_script_secret_fields(secret_name):
     )
 
 
+# ── enable_properties_merge (PATCH mode) ──────────────────────────────────────
+
+
+def test_enable_properties_merge_emits_patch_mcps(
```
alokr-dhub: **Two important scenarios not covered**

**1. `emit_storage_lineage=True` + `enable_properties_merge=True`**

The refactoring of `get_lineage_if_enabled` to accept `dataset_properties` as a parameter exists precisely because in PATCH mode the properties are no longer in the MCE snapshot — the old `mce_builder.get_aspect_if_available(mce, DatasetPropertiesClass)` call would silently return `None`, dropping all S3/Iceberg lineage. If that parameter-passing were ever reverted, every user with both flags enabled would silently lose storage lineage with no indication in the report. No test currently exercises this combination.

**2. Stateful ingestion + PATCH mode**

`test_glue_stateful` sets `enable_properties_merge: False` to preserve golden files, so stale entity removal with PATCH mode is completely untested. The important scenario is: run ingestion twice with `enable_properties_merge=True` and stateful ingestion enabled; a table is removed from Glue between runs. The second run should emit a soft-delete workunit. This is particularly worth covering since PATCH MCPs are emitted before the MCE, and their `is_primary_source` flag is what registers the URN in the checkpoint state.

metadata-ingestion/tests/unit/glue/test_glue_source.py
```diff
@@ -149,6 +151,7 @@ def glue_source_with_profiling(
             use_s3_object_tags=use_s3_object_tags,
             extract_delta_schema_from_parameters=extract_delta_schema_from_parameters,
             profiling=profiling_config,
+            incremental_properties=False,
```
alokr-dhub: The default case for incremental_properties should handle this scenario. I don't think we should set this explicitly in tests to test for the False case

metadata-ingestion/tests/unit/glue/test_glue_source.py
```diff
@@ -444,6 +447,7 @@ def test_glue_stateful(pytestconfig, tmp_path, mock_time, mock_datahub_graph):
     source_config_dict: Dict[str, Any] = {
         "extract_transforms": False,
         "aws_region": "eu-east-1",
+        "incremental_properties": False,
```
alokr-dhub: Similary, here we can test without explicilty setting for incremental_properties as False

alokr-dhub:
@slydon LGTM. The PR description seems to be outdated. Please fix it

alokr-dhub (approved):

PR #17529, opened 2026-05-20
refactor(glossary): redesign cards and sidebar with semantic tokens

annadoesdesign:
Before:
Screenshot 2026-05-20 at 4 23 49 PM
After: 
Screenshot 2026-05-20 at 3 35 20 PM

Refresh the visual treatment of the glossary card grid and sidebar navigation, and bring the touched files in line with the project's design-system conventions (alchemy components + semantic theme tokens).

UI changes
- Glossary cards: drop the colored header and diagonal ribbon in favor of a 40x40 colored icon container (full-color Phosphor icon on a lighter tinted background), built on top of the shared alchemy Card. Lay out the grid as 3 equal columns with an 8px gap aligned to the PageTitle, and clamp/tooltip long descriptions so hover never causes a horizontal page shift from scrollbar reflow.
- Glossary sidebar: replace MUI carets with Phosphor CaretRight / CaretDown at 14px regular weight, preserve the caret slot when a node has no children so rows stay aligned, and swap the old ribbon badge for the same colored icon container pattern. Child nodes and terms inherit the parent node's color and render the singular BookmarkSimple icon.
- Root glossary page: hide the bottom term list; it now only renders inside a specific glossary node.

Cleanup
- New shared GlossaryColoredIcon component replaces three near-identical inline icon-container implementations (card, NodeItem, TermItem).
- Remove dead code from GlossaryListCard (EntityDetails, EntityDetailsLeftColumn, EntityDetailsRightColumn, BookmarkIconWrapper, BookmarkRibbon, leftover saffron hover fill, duplicate align-items, malformed 'overflow hidden' rule) and collapse the duplicated GlossaryNode / GlossaryTerm branches.
- Replace antd usage in all touched files: antd Tooltip ->
  alchemy Tooltip, antd Typography -> plain themed div.
- Migrate every color in the touched files to semantic theme tokens (text, textSecondary, textDisabled, icon, iconDisabled, b

### DIFF
--- a/datahub-web-react/src/app/glossaryV2/GlossaryBrowser/NodeItem.tsx
+++ b/datahub-web-react/src/app/glossaryV2/GlossaryBrowser/NodeItem.tsx
@@ -1,5 +1,7 @@
-import { LoadingOutlined } from '@ant-design/icons';
-import { KeyboardArrowDownRounded, KeyboardArrowRightRounded } from '@mui/icons-material';
+import { BookmarkSimple } from '@phosphor-icons/react/dist/csr/BookmarkSimple';
+import { BookmarksSimple } from '@phosphor-icons/react/dist/csr/BookmarksSimple';
+import { CaretDown } from '@phosphor-icons/react/dist/csr/CaretDown';
+import { CaretRight } from '@phosphor-icons/react/dist/csr/CaretRight';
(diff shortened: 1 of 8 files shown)


chriscollins3456 (approved):
heck yeah! also i think the PR description is backwards - the before and after photo that is

PR #17531, opened 2026-05-21
fix(ui): Show glossary term icon next to parent-less terms in glossary

chriscollins3456:
Previously terms that were parent-less would show up in the nav sidebar without an icon. Now they do:

image


annadoesdesign (commented):
approved :)

annadoesdesign (approved):

PR #17532, opened 2026-05-21
refactor(ui): consolidate tag, term, and entity-sidebar pills with sh…

annadoesdesign:
Before:
Screenshot 2026-05-20 at 7 32 08 PM
Screenshot 2026-05-20 at 7 32 13 PM

After:
Screenshot 2026-05-20 at 7 14 34 PM
Screenshot 2026-05-20 at 7 14 43 PM
Screenshot 2026-05-20 at 7 14 40 PM
Screenshot 2026-05-20 at 7 14 36 PM

## Summary

Follow-up to #17529 — extends the new glossary pill pattern to tags, entity-sidebar links (Domain / Data Product / Application), and the Add Tags/Terms modal, and consolidates five copies of the same chip styling into shared primitives.

### New shared primitives
- **`TagPill`** (`app/sharedV2/tags/TagPill.tsx`) — capsule-shaped tag chip with a colored dot, optional remove icon, and a `borderless` variant for inline-with-text usage (replaces the legacy `TagLabel`).
- **`GlossaryTermPill`** (`app/glossaryV2/GlossaryTermPill.tsx`) — rounded-rect term chip with a colored icon container. Supports a `proposed` variant (dashed border for SaaS proposal flow) and a `borderless` variant that replaces the legacy `TermLabel`.
- **`PillRemoveIcon`** (`app/sharedV2/icons/PillRemoveIcon.tsx`) — single Phosphor `X` (12px / regular / semantic icon+iconHover) used by every chip and entity-sidebar link. Layout overrides applied via `className` to keep the primitive layout-agnostic.

### Testable utility extraction
- **`getGlossaryTermColor(term, generateColor)`** and **`getDeepestParentNode(parentNodes)`** in `glossaryV2/colorUtils.ts` — the "term inherits the deepest parent node's color, falling back to a palette pick" convention previously duplicated in 5 places (`TermContent`, `GlossaryTermMiniPreview`, `TagTermLabel`, `SearchFilterLabel`, `AddTagsTermsModal`) is now a single pure function backed by 10 unit tests.

### Entity sidebar consistency
- `DomainLink`, `DataProductLink`, `ApplicationLink`:
  - Phosphor `X` close icon (replacing antd `CloseOutlined`).
  - Link text stays gray (`theme.colors.text`) on hover / focus / active — no more accidental purple.
  - 16px icons in `theme.colors.icon`.
  - 4px gap between icon and text.
- All sibling-pill gaps (`TagTermGroup`, `SidebarOwnerSection`, `SidebarApplicationSection`, `SidebarDomainSection`) standardized to 4px.

### "Add Tags/Terms" modal redesign
- Migrated from antd `Select` to alchemy `Modal` + `SimpleSelect` with checkboxes.
- Term dropdown now renders as a nested tree (parent-node headers + indented term children) matching the business glossary sidebar.
- Expand/collapse carets on parent nodes (Phosphor `CaretRight` / `CaretDown`, 14px / regular).
- Recommendations seed the dropdown on open so the user no longer has to type to see anything.
- Already-applied URNs are filtered out so the dropdown is strictly "add new" — consistent with the previous antd flow.
- Selected items render in their respective sidebar pill format (`TagPill` for tags, `GlossaryTermPill` for terms).
- Node-only rows hide their checkbox since they aren't selectable.

### Other cleanup
- Deleted `app/shared/TagLabel.tsx` and `app/shared/TermLabel.tsx` — merged into the `borderless` variants of the new pills.
- Toast notifications migrated from antd `message` to alchemy `toast` across the affected files.
- Phosphor icons imported directly (no `@components/Icon` wrapper).

## Files changed

**New (4):** `TagPill.tsx`, `GlossaryTermPill.tsx`, `PillRemoveIcon.tsx`, `colorUtils.test.ts`
**Deleted (2):** `TagLabel.tsx`, `TermLabel.tsx`
**Modified (24):** entity-sidebar links, sidebar sections, tag/term wrappers, search filter labels, mini preview, modal

## Test plan

- [ ] **Tags sidebar**: Add a tag, remove a tag, search for a tag — pill renders as a capsule with a colored dot and a Phosphor X.
- [ ] **Glossary terms sidebar**: Add a term, remove a term, search for a term — pill renders with the colored icon container inheriting the root parent node's color.
- [ ] **Domain sidebar**: Domain link is gray (not purple) on hover. 16px icon. Phosphor X.
- [ ] **Data Product sidebar**: Same as Domain — gray link, 16px icon, Phosphor X.
- [ ] **Applications sidebar**: Vertically aligned, gray link, 16px icon, Phosphor X.
- [ ] **Add Tags modal**: Opens with recommendations populated; selecting a tag shows the capsule pill; already-applied tags are hidden from the dropdown.
- [ ] **Add Terms modal**: Tree view shows parent-node headers with carets; expanding/collapsing works; only terms (not nodes) are selectable; already-applied terms are hidden.
- [ ] **Home sidebar mini preview**: Glossary term shows in the new colored-icon container (no ribbon).
- [ ] **Search filter labels**: Tag, term, domain, and user filter pills render in their new pill formats with 10px text.
- [ ] **Glossary term page**: Term mini-previews and breadcrumbs render with the new pill.
- [ ] **`./gradlew :datahub-web-react:lint`** passes.
- [ ] **`yarn test --run src/app/glossaryV2/__tests__/colorUtils.test.ts`** — 10 new unit tests pass.

## Checklist

- [x] PR conforms to the [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md)
- [x] Tests added for the new color utility
- [ ] No breaking changes — purely visual / structural consolidation
- [ ] No docs changes needed


datahub-web-react/src/app/entityV2/glossaryTerm/profile/AddRelatedTermsModal.tsx
```diff
@@ -120,7 +122,19 @@ function AddRelatedTermsModal(props: Props) {
         const newUrns = [...selectedUrns, urn];
         setSelectedUrns(newUrns);
         const selectedSearchOption = tagSearchOptions.find((option) => option.props.value === urn);
-        setSelectedTerms([...selectedTerms, { urn, component: <TermLabel name={selectedSearchOption?.props.name} /> }]);
+        setSelectedTerms([
+            ...selectedTerms,
+            {
+                urn,
+                component: (
+                    <GlossaryTermPill
```
chriscollins3456, line 137: we're likely going to get some conflicts here when we OSS merge i think there was some work done to the related terms stuff in saas that hasn't been brought back to oss

datahub-web-react/src/app/glossaryV2/GlossaryTermPill.tsx
```diff
@@ -0,0 +1,132 @@
...
+    color,
+    icon = BookmarkSimple,
+    children,
+    rightAdornment,
+    onRemove,
+    highlight,
+    proposed,
+    fontSize,
+    clickable,
+    borderless,
```
chriscollins3456: this is a lot of props for different customizations.. i wonder if we need this sort of customizability? at least these ones seem like they all control different aspects of visual display: 
- rightAdornment
- highlight
- proposed
- fontSize
- borderless

i know in general we try to keep it simpler with different `variants` for different situations, with a different `size` for different use cases as well.

not a blocker just something i'd like to call out

datahub-web-react/src/app/shared/tags/AddTagsTermsModal.tsx
```diff
@@ -531,62 +597,25 @@ export default function EditTagTermsModal({
...
-                                tagSearchOptions
-                            )}
-                        </Select>
-                        <BrowserWrapper isHidden={!isShowingGlossaryBrowser}>
-                            <GlossaryBrowser isSelecting selectTerm={selectTermFromBrowser} />
-                        </BrowserWrapper>
-                    </ClickOutside>
-                </Form.Item>
-            </Form>
+            <SimpleSelect
```
chriscollins3456: oh interesting we can use the SimpleSelect for a nested situation? are we just rebuilding NestedSelect?

that being said NestedSelect is a mess and hard to deal with in general

datahub-web-react/src/app/glossaryV2/GlossaryTermPill.tsx
chriscollins3456: i see we have the same thing with the tag pill as well, again not blocking but curious if it could be simpler

datahub-web-react/src/app/shared/tags/AddTagsTermsModal.tsx
```diff
@@ -127,169 +132,312 @@ export default function EditTagTermsModal({
...
-        if (text.length > 0) {
-            tagTermSearch({
-                variables: {
-                    input: {
-                        type,
-                        query: text,
-                        limit: 10,
+    const [tagTermSearch, { data: searchData, loading: searchLoading }] = useGetAutoCompleteResultsLazyQuery();
+    // Seed the dropdown with the top entities of the relevant type so the modal isn't empty before the user types.
+    const { recommendedData, loading: recommendationsLoading } = useGetRecommendations([type]);
```
chriscollins3456: okay so it took a sec since there's a lot of business logic in this AddTagsTermsModal, but it looks like we are not using the tree that we create for the business glossary when a user first goes to add a term, instead just gets an initial list of recommended terms

are we sure we want to move away from displaying the glossary itself like we were before just with new styles?

i think it could work like it does with domains now where we basically just render a small version of the left browser on the business glossary page and let them select terms

datahub-web-react/src/app/shared/tags/AddTagsTermsModal.tsx
```diff
@@ -127,169 +132,312 @@ export default function EditTagTermsModal({
...
-        querySelectorToExecuteClick: '#addTagButton',
-    });
+    const excludeSet = useMemo(() => new Set(existingUrns || []), [existingUrns]);
+
+    // Build a tree-ordered flat list of options for glossary terms.
+    // Each term carries its parent nodes; we synthesise a row per unique parent node so the dropdown
+    // reads like the sidebar tree (node header → indented term children). Node rows are disabled so
+    // the user can't accidentally select a node (we're only adding terms). Already-applied terms are
+    // dropped, and node headers are only emitted if they have at least one selectable child.
+    const termTreeOptions = useMemo<EntityOption[]>(() => {
```
chriscollins3456: i'd also rather not have all the logic for building this tree here in AddTagsTermsModal, this component was bloated and messy before and now it's much more so

if we're able to pull out the glossary tree into its own component and use that that would be ideal.

honestly maybe it's time we finally split this component into 2 - AddTagsModal and AddTermsModal. there's no reason these two need to be in one component anymore. if you don't want to take that on since i know this is supposed to be UI polish that's alright

chriscollins3456 (commented):
my main concern here is getting rid of rendering the glossary tree and just rendering some recommendations. i don't know if that functionality should change, but instead just updating the styles

i suggested we follow a similar pattern for what we do for domains, where we can render the left glossary browser and let users select from there, and if they search for terms, render them flat instead of showing them nested

datahub-web-react/src/app/entityV2/glossaryTerm/profile/AddRelatedTermsModal.tsx
annadoesdesign: Fair flag. The change here is small — `TermLabel` → `GlossaryTermPill` with `variant="borderless"`, plus the `BrowserWrapper` import path now that the combined modal was split up. Kept it because the point of this PR is visual consistency across pill components, and `AddRelatedTermsModal` was the last spot still rendering raw `TermLabel`s. The merge conflict should be a one-line resolution per occurrence — happy to revert this file if you'd rather take the consistency hit until SaaS lands.

datahub-web-react/src/app/glossaryV2/GlossaryTermPill.tsx
annadoesdesign: Agreed, addressed in 8a00d56. The pill API is now `variant: 'default' | 'borderless' | 'highlighted'` + `size: 'sm' | 'md'`. `highlight`, `borderless`, `fontSize`, and `proposed` are all gone — `proposed` removed entirely so it doesn't conflict with the SaaS flow. Only remaining ad-hoc props are composition slots (`children`, `rightAdornment`) and the standard `className` / `dataTestId`.

datahub-web-react/src/app/shared/tags/AddTagsTermsModal.tsx
annadoesdesign: Yeah, partly. The difference is we lean entirely on `SimpleSelect`'s native checkbox UX and only layer on depth + a caret column — no separate keyboard model, no portal popovers, no menu state machine. The whole tree algorithm now lives in `useTermTreeOptions` (~165 lines, unit-tested in `__tests__/useTermTreeOptions.test.tsx`). If `NestedSelect` ever gets the cleanup treatment, this can collapse back into it without touching consumers.

datahub-web-react/src/app/glossaryV2/GlossaryTermPill.tsx
annadoesdesign: Same cleanup applied to `TagPill` in 8a00d56 — `variant` + `size` only, no more `highlight` / `borderless` / `fontSize` / `proposed`.

datahub-web-react/src/app/shared/tags/AddTagsTermsModal.tsx
annadoesdesign: Conscious choice — wanted to preserve the `SimpleSelect`-with-checkbox UX so users get the same selection ergonomics for tags and terms (multi-select with chip strip + remove icons), rather than introducing a separate left-pane browser only for terms.

To meet you partway in 8a00d56:

- On initial load we still surface recommendations (no API change), but they now render **nested** via `useTermTreeOptions` — root → leaf indentation, expand/collapse carets per node, dedup of shared ancestors.
- On search we flatten as you suggested.
- Already-applied terms are excluded from the tree so users can only add.

Happy to layer in the full glossary-browser pattern as a follow-up if you still want it, but it felt like a bigger UX shift than this PR's scope.

datahub-web-react/src/app/shared/tags/AddTagsTermsModal.tsx
annadoesdesign: Both done in 8a00d56:

- Tree-building logic extracted into `useTermTreeOptions` (+ 9 unit tests in `__tests__/useTermTreeOptions.test.tsx`).
- `AddTagsTermsModal` split into separate `AddTagsModal.tsx` + `AddTermsModal.tsx`.
- Picker state + autocomplete plumbing extracted into `useEntityPickerState` (shared by both modals).
- Batch add/remove mutation flow extracted into `useBatchTagTermMutation`.
- Hidden `GlossaryBrowser` portal lifted into its own `BrowserWrapper` file.

datahub-web-react/src/app/entityV2/glossaryTerm/profile/AddRelatedTermsModal.tsx
chriscollins3456: no, claude, that's okay we'll handle in the merge

datahub-web-react/src/app/shared/tags/AddTagsTermsModal.tsx
chriscollins3456: yeah i'm just not fully convinced we want to move away from the browse experience for glossary terms by default. that's how it's been in the app for 3+ years and i think it's how people think about the business glossary

the biggest risk is that sometimes there are terms with the same name in different parent nodes and without the browse capability it will be hard or impossible to find them

that being said i'm concerned with the searching we have now that we lose the parent nodes breadcrumbs - right? for the same reason i'm concerned - knowing exactly which term it is that you're selecting.

chriscollins3456 (approved):
looking solid - did some manual review with you and i think we're good to go here to get this along

PR #17536, opened 2026-05-21
chore(prefect-plugin): upgrade Prefect v2→v3 and clean up pydantic v1 usage

dinesh-verma-datahub:
## Summary

Upgrades the Prefect plugin from Prefect v2 → v3 and removes all Pydantic v1 shim usage.

**Prefect v2 → v3**
- Dependency pin updated to `>=3.0.0,<4.0.0`
- All `asyncio.run(...)` calls replaced with `run_coro_as_sync(...)`
- Entry point group changed from `prefect.block` to `prefect.collections`
- `_block_type_name` fixed to use `ClassVar`

**Task emission rewrite**
- Removed iteration over `flow_run_ctx.task_run_futures` (dropped in Prefect v3)
- Replaced with `read_task_runs` API calls with offset pagination (`_TASK_RUN_PAGE_SIZE = 200`) and stable-count polling to handle Prefect's async task run state persistence
- Eliminated redundant `read_task_run` call per task — `_emit_task_run` now accepts a `TaskRun` directly from the already-fetched batch
- Cross-page dedup added (UUID sort order means a row can appear on multiple pages during concurrent inserts)

**Pydantic v1 → v2**
- Removed `pydantic.v1` shim (`from pydantic.v1 import SecretStr` → `from pydantic import SecretStr`)
- `_datajobs_to_emit` changed from a mutable class-level dict to `PrivateAttr(default_factory=dict)` with tightened type `Dict[str, DataJob]`
- `FlowRun.parse_obj()` / `TaskRun.parse_obj()` replaced with `model_validate()`
- Dead `V1RootValidator`/`V1Validator` type references and `TYPE_CHECKING` guards removed across 8 files

**Tests**
- `fast_sleep` fixture eliminates real `asyncio.sleep` delays from polling loop tests
- Upstream lineage assertions added to `test_emit_flow` (verifies `inputDatajobs` URNs)
- New test files: error paths, GMS failure isolation, pagination edge cases, cross-page dedup

## Test plan
- [ ] `pytest metadata-ingestion-modules/prefect-plugin/tests/unit/` passes (Python 3.10, 3.11)
- [ ] CI prefect-plugin checks green


sgomezvillamor (approved):
Great work on the Prefect v2→v3 and pydantic v1→v2 migration! The `asyncio.run()` → `run_coro_as_sync()` approach is exactly right for Prefect v3's event loop model, and replacing `task_run_futures` with paginated `read_task_runs` calls is a clean, correct fix. A few items before merge:

---

**🔴 BLOCKER — Breaking change not documented in `updating-datahub.md`**
This PR drops Prefect v2 support (`>=2.0.0,<3.0.0` → `>=3.0.0,<4.0.0`) and changes the entry point group from `prefect.block` to `prefect.collections`. Both are user-facing breaking changes. Please add an entry to `docs/how/updating-datahub.md` before merge.

---

**🟡 WARNING — Silent timeout in `_fetch_all_task_runs_stable`**
The 50-iteration × 0.1 s polling loop exits silently when exhausted. If Prefect's state persistence is slow, tasks could be silently dropped from DataHub lineage. Please add a warning log when the loop ends without stabilising.

---

**🟡 WARNING — `# type: ignore[arg-type]` on `FlowRunFilterId`**
`flow_run_id` is typed as `str` but `FlowRunFilterId.any_` likely expects `List[UUID]`. Worth fixing the type rather than suppressing — e.g. `any_=[UUID(flow_run_id)]`.

---

**ℹ️ SUGGESTION — `import asyncio` inside method body**
`asyncio` was removed from the top-level imports but re-introduced locally inside `_fetch_all_task_runs_stable`. It should stay at the module level per project conventions.

---

**❓ QUESTION — Can `-> Any` be narrowed for the validator factory functions?**
In files like `connection_resolver.py`, `validate_field_*.py`, etc. the return type changed from `V1RootValidator`/`V1Validator` → `Any`. In pydantic v2, `model_validator(mode="before")(fn)` wraps validators as `classmethod` descriptors — would `-> classmethod` be a viable narrower annotation here?

---

docs/how/updating-datahub.md
```diff
@@ -44,6 +44,8 @@ Requirements:
 
 ### Breaking Changes
 
+- #17536: **(prefect-plugin)** The `acryl-datahub-prefect-plugin` package now requires **Prefect 3.x** (`>=3.0.0,<4.0.0`); Prefect 2.x is no longer supported. The Prefect entry point group also changed from `prefect.block` to `prefect.collections` to match Prefect 3's plugin discovery. **Action:** Upgrade Prefect to 3.x and re-register the DataHub block (`DatahubEmitter`) before upgrading the plugin.
```
```suggestion
- #17536: **(prefect-plugin)** The `datahub-prefect` package now requires **Prefect 3.x** (`>=3.0.0,<4.0.0`); Prefect 2.x is no longer supported. The Prefect entry point group also changed from `prefect.block` to `prefect.collections` to match Prefect 3's plugin discovery. **Action:** Upgrade Prefect to 3.x and re-register the DataHub block (`DatahubEmitter`) before upgrading the plugin.
```

PR #17550, opened 2026-05-22
feat(dev): multi-worktree isolation and remote runner plugin for datahub-dev

shirshanka:
## Summary

- **Multi-worktree isolation**: each git worktree gets its own Docker project name, volumes, and ports automatically — no more collisions when running DataHub in multiple branches simultaneously. Slot-based port allocation (+1000 per slot) across up to 12 concurrent instances (2 local, 10 remote). State is tracked in `~/.datahub/dev/instances.json`.

- **Remote runner plugin**: a pluggable interface that proxies all `datahub-dev.sh` operations to a remote machine (EC2, K8s pod, etc.). Set `runner` in `~/.datahub/dev/config.json` or `DATAHUB_RUNNER` env var. New `suspend`/`resume` stop/start the underlying compute so you're not billed when idle.

- **Reference K8s runner** at `scripts/dev/runners/k8s.sh` implementing all 6 interface verbs (`init`, `sync`, `exec`, `tunnel`, `resume`, `suspend`).

- **39 new tests** in `scripts/dev/tests/test_runner.py` covering slot allocation, tunnel-pair computation, registry round-trips, call sequence, error propagation, and idempotency.

- **AGENTS.md** updated with the full remote runner lifecycle, port assignment table, and runner interface contract.

Also fixes two previously hardcoded ports in `docker-compose.gms.yml` (MAE consumer 9091, MCE consumer 9090 + debug 5009) that blocked parallel local instances, and makes Gradle's Docker Compose `projectName` configurable via `COMPOSE_PROJECT_NAME`.

## New commands

```bash
# Multi-instance
datahub-dev.sh instances list    # show all local + remote instances with status/ports
datahub-dev.sh instances clean   # remove stale registry entries
eval $(datahub-dev.sh shell-env) # set DATAHUB_GMS_URL to this instance's port

# Remote runner lifecycle
datahub-dev.sh setup --remote    # one-time: provision remote environment (runner init)
datahub-dev.sh start             # sync changed files → quickstartDebug on remote → tunnel ports
datahub-dev.sh suspend           # stop containers + halt compute (no billing while idle)
# 'start' automatically resumes a suspended instance
```

## Runner interface

A runner is any executable that speaks 6 verbs:

```bash
runner init                        # one-time bootstrap
runner sync                        # push changed local files to the remote
runner exec -- <cmd> [args...]     # execute a command in the remote workspace
runner tunnel <local:remote> ...   # set up port forwarding
runner resume                      # start compute if stopped (no-op if running)
runner suspend                     # stop containers + halt compute
```

## Test plan

- [ ] `uv run --python 3.11 --no-project --with pytest pytest scripts/dev/tests/test_runner.py scripts/dev/tests/test_datahub_dev.py -v` — all 54 tests pass
- [ ] `datahub-dev.sh start` in two different worktrees simultaneously — each gets isolated ports/volumes
- [ ] `datahub-dev.sh instances list` — shows both instances with correct ports
- [ ] `eval $(datahub-dev.sh shell-env)` — sets correct `DATAHUB_GMS_URL` per worktree
- [ ] With a runner configured: `start` → `stop` → `suspend` round-trip

### DIFF
--- a/docker/build.gradle
+++ b/docker/build.gradle
@@ -10,7 +10,7 @@ apply from: "../gradle/versioning/versioning.gradle"
 
 ext {
     compose_base = "profiles/docker-compose.yml"
-    project_name = "datahub"
+    project_name = System.getenv("COMPOSE_PROJECT_NAME") ?: "datahub"
 
     backend_profile_modules = [
             ':datahub-upgrade',
--- a/metadata-integration/java/datahub-client/src/main/resources/MetadataChangeProposal.avsc
+++ b/metadata-integration/java/datahub-client/src/main/resources/MetadataChangeProposal.avsc
@@ -173,11 +173,6 @@
         "doc" : "Aspect version\n   Initial implementation will use the aspect version's number, however stored as\n   a string in the case where a different aspect versioning scheme is later adopted.",
         "default" : null
       }, {
-        "name" : "schemaVersion",
-        "type" : [ "null", "long" ],
-        "doc" : "Schema version of the aspect data model. Used to determine aspects that need migrations.\nDefaults to 1 when not present as a baseline. Incremented when an aspect is modified.",
-        "default" : null
-      }, {
         "name" : "aspectCreated",
         "type" : [ "null", {
           "type" : "record",
--- a/metadata-integration/java/datahub-event/src/main/resources/MetadataChangeProposal.avsc
+++ b/metadata-integration/java/datahub-event/src/main/resources/MetadataChangeProposal.avsc
@@ -173,11 +173,6 @@
         "doc" : "Aspect version\n   Initial implementation will use the aspect version's number, however stored as\n   a string in the case where a different aspect versioning scheme is later adopted.",
         "default" : null
       }, {
-        "name" : "schemaVersion",
-        "type" : [ "null", "long" ],
-        "doc" : "Schema version of the aspect data model. Used to determine aspects that need migrations.\nDefaults to 1 when not present as a baseline. Incremented when an aspect is modified.",
-        "default" : null
-      }, {
         "name" : "aspectCreated",
         "type" : [ "null", {
           "type" : "record",
--- a/scripts/dev/datahub-dev.sh
+++ b/scripts/dev/datahub-dev.sh
@@ -4,23 +4,35 @@
 #   scripts/datahub-dev.sh rebuild --wait
 # etc.
 
-# Inject runner path from ~/.datahub/dev/config.json when not already set and
-# not running on a remote machine (DATAHUB_REMOTE_EXEC=1 means we are already
-# on the remote side of a runner invocation — no further proxying).
-if [ -z "$DATAHUB_RUNNER" ] && [ "$DATAHUB_REMOTE_EXEC" != "1" ]; then
-  DATAHUB_RUNNER=$(python3 - <<'EOF' 2>/dev/null
-import json, pathlib
+# Read per-user config from ~/.datahub/dev/config.json.
+# Reads two keys (runner, compose_project) in a single Python call to avoid
+# paying the interpreter startup cost twice.
+if [ "$DATAHUB_REMOTE_EXEC" != "1" ]; then
+  _dh_cfg=$(python3 - <<'EOF' 2>/dev/null
+import json, pathlib, sys
 cfg = pathlib.Path.home() / ".datahub/dev/config.json"
 try:
     d = json.loads(cfg.read_text())
     r = d.get("runner", "")
+    p = d.get("compose_project", "")
     if r:
-        print(str(pathlib.Path(r).expanduser()))
+        r = str(pathlib.Path(r).expanduser())
+    print(r)
+    print(p)
 except Exception:
-    pass
+    print("")
+    print("")
 EOF
 )
-  [ -n "$DATAHUB_RUNNER" ] && export DATAHUB_RUNNER
+  if [ -z "$DATAHUB_RUNNER" ]; then
+    DATAHUB_RUNNER=$(echo "$_dh_cfg" | sed -n '1p')
+    [ -n "$DATAHUB_RUNNER" ] && export DATAHUB_RUNNER
+  fi
+  if [ -z "$COMPOSE_PROJECT_NAME" ]; then
+    COMPOSE_PROJECT_NAME=$(echo "$_dh_cfg" | sed -n '2p')
+    [ -n "$COMPOSE_PROJECT_NAME" ] && export COMPOSE_PROJECT_NAME
+  fi
+  unset _dh_cfg
 fi
 
 # When a runner is configured and we're not already on the remote side,
--- a/scripts/dev/datahub-dev.sh
+++ b/scripts/dev/datahub-dev.sh
@@ -3,4 +3,44 @@
 #   scripts/datahub-dev.sh status
 #   scripts/datahub-dev.sh rebuild --wait
 # etc.
+
+# Inject runner path from ~/.datahub/dev/config.json when not already set and
+# not running on a remote machine (DATAHUB_REMOTE_EXEC=1 means we are already
+# on the remote side of a runner invocation — no further proxying).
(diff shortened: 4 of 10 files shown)


docker/build.gradle
```diff
@@ -10,7 +10,7 @@ apply from: "../gradle/versioning/versioning.gradle"
 
 ext {
     compose_base = "profiles/docker-compose.yml"
-    project_name = "datahub"
+    project_name = System.getenv("COMPOSE_PROJECT_NAME") ?: "datahub"
```
devashish2203, line 13: Let's use gradle.properties to define overridable settings and then users can set ORG_GRADLE_PROJECT_<propertyName> to override the value on their local.

https://docs.gradle.org/current/userguide/build_environment.html#sec:project_properties:~:text=Environment%20Variables%3A%20You%20can%20set%20project%20properties%20with%20environment%20variables.%20If%20the%20environment%20variable%20name%20looks%20like%20ORG_GRADLE_PROJECT_myProperty%3D%27Hi%2C%20world%27%2C%20then%20Gradle%20will%20set%20a%20myProperty%20property%20on%20your%20project%20object%2C%20with%20the%20value%20of%20Hi%2C%20world.

metadata-integration/java/datahub-client/src/main/resources/MetadataChangeProposal.avsc
```diff
@@ -172,11 +172,6 @@
         "type" : [ "null", "string" ],
         "doc" : "Aspect version\n   Initial implementation will use the aspect version's number, however stored as\n   a string in the case where a different aspect versioning scheme is later adopted.",
         "default" : null
-      }, {
```
devashish2203, line 175: Is this a related change?

metadata-integration/java/datahub-event/src/main/resources/MetadataChangeProposal.avsc
```diff
@@ -172,11 +172,6 @@
         "type" : [ "null", "string" ],
         "doc" : "Aspect version\n   Initial implementation will use the aspect version's number, however stored as\n   a string in the case where a different aspect versioning scheme is later adopted.",
         "default" : null
-      }, {
```
devashish2203, line 175: Same as above?

scripts/dev/datahub-dev.sh
```diff
@@ -3,4 +3,44 @@
 #   scripts/datahub-dev.sh status
 #   scripts/datahub-dev.sh rebuild --wait
 # etc.
+
+# Inject runner path from ~/.datahub/dev/config.json when not already set and
```
devashish2203: When this goes to fork, how do we differentiate between OSS and fork config?

scripts/dev/datahub_dev.py
devashish2203, line 119: nit: This will keep creating .env files for every branch. Do we need some cleanup mechanism as well or we expect people to delete these manually at some point?

scripts/dev/datahub_dev.py
```diff
@@ -112,15 +122,271 @@ def _get_branch_slug() -> str:
...
+                if proj.startswith("dh-") and proj != COMPOSE_PROJECT:
+                    projects.add(proj)
+                break
+    return len(projects)
+
+
+def _write_ports_to_env_file(ports: Dict[str, int]) -> None:
+    """Upsert all port env vars into DEV_ENV_FILE so docker compose picks them up."""
+    if not DEV_ENV_FILE.exists():
+        DEV_ENV_FILE.touch()
```
devashish2203, line 365: Based on the file name it gets created for every branch. 
What is the expected behaviour if someone changes branch in the same worktree or root location? Do the ports change then?

scripts/dev/datahub_dev.py
```diff
@@ -112,15 +122,271 @@ def _get_branch_slug() -> str:
...
+REMOTE_EXEC: bool = os.environ.get("DATAHUB_REMOTE_EXEC") == "1"
+
+# Path to the runner executable.  Cleared when running remotely so a runner
+# configured on the remote machine doesn't cause recursive proxying.
+RUNNER: str = "" if REMOTE_EXEC else os.environ.get("DATAHUB_RUNNER", "")
+
+# Docker Compose project name. Allows explicit override (power-user escape
+# hatch); defaults to a per-worktree identifier so all Docker objects
+# (containers, volumes, networks) are automatically namespaced.
+COMPOSE_PROJECT: str = os.environ.get("COMPOSE_PROJECT_NAME") or f"dh-{WORKTREE_ID}"
```
devashish2203, line 157: Use a sha in COMPOSE_PROJECT could be a little confusing when trying to find which container belongs to which worktree as the sha is not readable.

Maybe we need helper functions to lookup/list WORKTREE_ID/COMPOSE_PROJECT details for a worktree.

docker/build.gradle
shirshanka: Thanks for the pointer! The `ORG_GRADLE_PROJECT_` pattern is the right approach for custom Gradle settings, but `COMPOSE_PROJECT_NAME` is a [Docker Compose standard environment variable](https://docs.docker.com/compose/how-tos/project-name/) that the Gradle docker-compose plugin also reads natively. `datahub-dev.sh` injects it once (via `_dev_env()`) and it flows to both docker compose and the Gradle plugin from a single source — if we moved it to a Gradle property, docker compose itself would stop seeing it and the per-worktree isolation would break.

metadata-integration/java/datahub-client/src/main/resources/MetadataChangeProposal.avsc
shirshanka: Not intentional — this branch was cut before `schemaVersion` was added to master in #16778. The rebase I just pushed resolves it.

metadata-integration/java/datahub-event/src/main/resources/MetadataChangeProposal.avsc
shirshanka: Same — stale base, resolved by the rebase.

scripts/dev/datahub-dev.sh
shirshanka: No fork-specific logic needed in the script itself — the runner plugin is the extension point. The runner is purely a user-level choice of execution environment (local Docker, K8s, cclaude, whatever), completely orthogonal to which repo you are working in. You configure your preferred runner once in `~/.datahub/dev/config.json` and it works the same regardless of OSS or fork. The config is user-level so it never touches the repo.

scripts/dev/datahub_dev.py
shirshanka: This is the same lifecycle as any other branch artifact — same as worktree directories accumulating when you work across multiple worktrees. The env files live inside the worktree at `docker/datahub-dev-<branch>.env`, so they are cleaned up automatically when the worktree itself is removed. Within a single worktree, a few tiny gitignored files per branch is normal dev hygiene.

scripts/dev/datahub_dev.py
shirshanka: Ports do **not** change on a branch switch. `WORKTREE_ID` is derived from the repo root path, not the branch, so the same Docker project, volumes, and port assignment survive a `git checkout`. Only the env file name changes (it is branch-specific so each branch carries its own feature-flag/port overrides), but the running containers are unaffected. This is intentional — you want your DataHub instance to keep running while you switch branches.

scripts/dev/datahub_dev.py
shirshanka: The discovery commands already exist: `datahub-dev.sh instances list` prints the full mapping (worktree path → compose project → ports) for every registered instance, and `datahub-dev.sh shell-env` gives the GMS URL for the current worktree. For raw `docker ps` output the short hash is admittedly opaque, but `instances list` is the intended lookup tool. Happy to make the output more prominent in the docs if that helps.

docker/build.gradle
devashish2203: Understood.

devashish2203 (approved):

PR #17553, opened 2026-05-22
fix(ingest/dataplex): fix lookupEntryLinks request and attach glossary terms to containers

sgomezvillamor:
## Summary

Dataplex glossary term-to-asset associations never worked end to end. Three defects sat on the path, each masking the next: every `lookupEntryLinks` call was rejected with a 403; linked BigQuery datasets could neither be matched nor emitted; and when a link failed to bind, nothing said so.

This builds on the project number → project ID normalization already in the connector (`_normalize_entry_project_id`), which maps the project number Dataplex returns in entry references back to the project ID that ingested entries are keyed by. That normalization was correct but unreachable, because the API call preceding it always failed. This PR fixes the call, extends matching and emission to Container entities, and makes unmatched links visible in the report.

## 1. Malformed `lookupEntryLinks` request → 403 on every call

The glossary term entry path mixed representations:

```
projects/{project_id}/locations/global/entryGroups/@dataplex/entries/projects/{project_number}/locations/{gl_location}/glossaries/...
         ^^^ needs the project NUMBER  ^^^ hardcoded, needs the term's location
```

`lookupEntryLinks` requires the location in the URL path and the location inside the entry resource name to be the same region. Requests in this shape are refused at the edge, before IAM evaluation — which is why they produced 403s with **no Cloud Audit Log entries at all**, while the same service account succeeded via `curl`. That is what made this look like a permissions problem.

The entry path now carries the project number and the term's own location in both halves:

```
projects/{project_number}/locations/{gl_location}/entryGroups/@dataplex/entries/projects/{project_number}/locations/{gl_location}/glossaries/...
```

The `(project × entries_location)` loop is also gone, replaced by a single call at the term's location — `terms × projects × entries_locations` requests become `terms`. This is not a reduction in coverage: the loop varied only the URL location while reusing one fixed `entry` parameter, so at most one iteration — the one matching the term's location — could ever succeed, and the rest were guaranteed 403s. Dataplex stores a term's entry links alongside the term and returns linked assets by full resource name, so assets in other locations still resolve.

**Behavioural change worth noting:** previously a failure at one location let another location still supply links. With a single call, a failure now zeroes that term. The per-term warning is retained and the new counters below make it visible.

## 2. Container assets were unreachable

A BigQuery **dataset** maps to a DataHub **Container**, not a Dataset. A term attached to one could never be matched, and phase 2 would have raised `InvalidUrnError` from `DatasetUrn.from_string` had it matched — aborting the whole stage rather than tagging anything.

- `DataplexContext` now carries `entry_name_to_urn`, registered for **both** Dataset and Container entities at the single point in the entries stage that every built entity passes through. Kept separate from `entry_data`, which is a lineage-only index: widening `EntryDataTuple` would drag containers into `get_lineage_for_entry` and burn a Lineage API call per container per project/location pair.
- Phase 2 emits a bare `glossaryTerms` aspect, which works for any entity type. The workunit id and aspect value are **unchanged** from what `Dataset.set_terms` produced (verified empirically); only a stray `dataPlatformInstance` aspect the SDK entity emitted alongside it is gone. An SDK `Container` was not an option — it requires a display name and would emit `containerProperties` over what the entries stage wrote earlier in the same run.

## 3. Unmatched links were invisible

A link that resolved but never bound was logged at `DEBUG` only, leaving a clean report with `term_associations_emitted: 0` and no indication of why. The report now has `term_links_matched`, `term_links_unmatched` and a sample of unmatched entry names, and warns **once** when links came back but none of them bound.

## Smaller fixes

- **`project_numbers` `KeyError` guard.** `self._ctx.project_numbers[project_id]` was an unguarded index running in a worker thread; an unresolved project raised and silently dropped that term's associations. It now warns and names the required permission.
- **Dead `project_ids` parameter dropped** from `process_term_associations`. It fed the removed location loop from `config.project_ids` rather than the resolved project list, so under `project_labels` / `project_id_pattern` discovery that list is empty and **zero** lookups were made regardless of the path being correct. The failure-path `context=` in `dataplex.py` now reports the resolved list too.

## Testing

- 318 Dataplex unit + integration tests pass; `lint` (ruff + mypy) clean.
- The association fixtures set the mocked `entryReferences[].name` to the same project-ID string as the tracked entry, which made the project-number mismatch **structurally invisible** — those tests passed whether or not `_normalize_entry_project_id` worked. They now use the project-number form the API actually returns, in a realistic entry-name shape. Verified by temporarily neutering the helper: four tests that previously passed regardless now fail.
- New coverage: Container assets receive terms; only `glossaryTerms` is emitted (no stray aspect); the request URL and entry path are asserted, along with one call per term rather than one per configured location; an unresolved project number warns instead of raising; unmatched links are counted and warned; both Dataset and Container entries are registered in the index while only Datasets reach the lineage index.
- `tests/integration/dataplex/test_dataplex_glossary_reconcile.py` is updated because it seeds the glossary lookup index directly; the reconciliation behaviour it covers is unchanged.

## Checklist

- [x] Tests added/updated
- [x] Docs updated — entity-type support for lineage vs term associations in `dataplex_post.md`, and a GCS example corrected (this connector has no GCS support)
- [x] No breaking changes, so no `docs/how/updating-datahub.md` entry


aviraj-gour:
Nice diagnosis — the malformed-resource-name-rejected-before-IAM finding explains a class of 403 that is genuinely hard to attribute, and deleting the `(project × location)` loop is clearly right once you see only one iteration could ever have matched.

A few things worth picking up.

## Stale comments this PR leaves behind

All three now contradict code in the same diff:

- **`dataplex_glossary.py:264-266`** — the class docstring still says `process_term_associations` calls `lookupEntryLinks` "across all entries_locations" (the loop this PR deletes) and emits "on those DataHub **Dataset** entities" (Containers are the point of the PR). Doubly stale, and it is the first thing a reader of this class sees.
- **`dataplex_context.py:29`** — "consumed by lineage + glossary stages". `grep entry_data dataplex_glossary.py` now returns nothing, and it contradicts the new comment eight lines below that correctly calls `entry_data` "a lineage-only index".
- **`test_dataplex_glossary.py:614`** — docstring says an unresolved project number "must not raise inside the worker thread", but the executor loop already catches worker exceptions. The contract the test asserts is the warning title and that no call is made.

## The location half of the entry-path fix is untested

The number half is covered; the location half is not. Every glossary term fixture uses `location="global"` (`test_dataplex_glossary.py:113`, `:307`, `:430`, `:436`), so `gl_location` is always the literal `"global"` and reverting `locations/{gl_location}` to `locations/global` is indistinguishable. The URL assertion at `:602` bakes in `locations/global` on both halves too.

`glossary_locations` is user-configurable and its own description says glossaries "can exist in any location" (`dataplex_config.py:247-252`), so a regional glossary is a supported config that reproduces the exact bug this PR fixes.

One fixture change covers it — set the term location to something regional and assert the decoded `entry=` param. That single test subsumes both halves.

Two smaller gaps nearby, lower priority:

- Nothing makes `authed_session.get` raise, so the new per-term `except` at `:737` is uncovered. Slightly more relevant now that a single transient 500 zeroes a term instead of falling through to another location.
- `term_links_matched` is never positively asserted. The missing case is one matched + one unmatched, where the "no links matched" warning must **not** fire — as written, dropping the `report_link_matched()` call keeps the suite green while arming a false warning on any run with one out-of-scope link.

## `_terms_audit_stamp()` duplicates an existing SDK helper

`HasTerms._terms_audit_stamp()` at `sdk/_shared.py:613-618` is byte-identical, and the new 12-line docstring (`dataplex_glossary.py:74-87`) does not mention it. Two knock-ons:

- "what the SDK writes for terms, tags and owners" is wrong for tags and owners — `set_tags` emits `GlobalTagsClass(tags=[...])` (`_shared.py:556`) and `set_owners` emits `OwnershipClass(owners=[...])` (`_shared.py:360`), neither with an audit stamp. Terms only.
- The "new instance per call because the aspect classes are mutable" paragraph documents a hazard the code does not have — the value is constructed fresh at the single call site regardless.

Relatedly, `from datahub.sdk._utils import DEFAULT_ACTOR_URN` makes this connector the only importer of `sdk._utils` from outside `sdk/`. The cited `TODO` is real, but it argues for exporting the constant publicly rather than reaching into a private module. If you would rather not touch the SDK here, fine — but then the helper and its docstring could shrink to a line or two.

## `dataplex_post.md:227` over-states Cloud SQL support

The only Cloud SQL mappers are `cloudsql-mysql-{instance,database,table}` (`dataplex_mappers.py:821`, `:841`, `:865`) — no Postgres or SQL Server. "Cloud SQL tables" and "Cloud SQL instances and databases" should both say "Cloud SQL for MySQL". The rest of the list matches `ENTRY_MAPPERS` exactly.

## Minor: comment density

~40 lines of prose on ~157 added lines, some of it PR description rather than code context — the audit-stamp docstring, the import justification, and a third copy of the Dataset-vs-Container fact at `dataplex_entries.py:332-334`. The fixture comments at `test_dataplex_glossary.py:257-268` are the model: one non-obvious API quirk, stated once, at the definition. Taste call — and the *why*-comments on the entry path and the bare-MCP choice are the good kind, those should stay.

sgomezvillamor:
Thanks — all of this was right, and I verified each point rather than taking them on faith. Addressed in `a9543813bc847624039f0c6ce4571087783ed050`.

**Stale comments** — all three fixed. The class docstring was the worst of them: it described the deleted loop *and* claimed Dataset-only, which is the opposite of what this PR adds. `grep entry_data dataplex_glossary.py` returns 0, confirming the `dataplex_context` comment was wrong too.

**The location half being untested** — this was the most valuable catch and I'd missed it entirely. You're right that every fixture pinned `location="global"`, so `gl_location` was always the literal and the URL assertion baked it into both halves. `_setup_processor_with_terms` now takes a `location` parameter and the URL test uses a regional glossary. Confirmed it bites: reverting *only* the location half (keeping `project_number`) now fails that test, where before the suite stayed green.

**`term_links_matched` never positively asserted** — also correct, and the failure mode you describe is the nasty one. Added a one-matched/one-unmatched test that asserts both counters and that the aggregate warning stays silent. Verified: deleting `report_link_matched()` fails it, where previously that change kept the suite green while arming a false "nothing matched" warning on any run with an out-of-scope link.

**Uncovered `except`** — added a test that makes the session raise. Agreed it matters more now that one failure zeroes a term instead of falling through.

**`_terms_audit_stamp()`** — you're right on both sub-points. `set_tags` emits `GlobalTagsClass(tags=[...])` and `set_owners` emits `OwnershipClass(owners=[...])`, neither with an audit stamp, so "terms, tags and owners" was simply wrong. And the aliasing paragraph described a hazard that doesn't exist at a single call site. The docstring now points at `HasTerms._terms_audit_stamp` and keeps only the `time=0` rationale.

On the private import: I've left it, but shrunk the justification to one line. Exporting the constant publicly is the right answer and I agree it belongs in the SDK rather than here — happy to do it as a follow-up if you'd prefer that over the current state.

**`dataplex_post.md` Cloud SQL** — correct, that was my error. Only `cloudsql-mysql-{instance,database,table}` exist; both mentions now say "Cloud SQL for MySQL".

**Comment density** — fair. Trimmed the audit-stamp docstring, the import justification, and the third copy of the Dataset-vs-Container fact in `dataplex_entries.py`. Kept the entry-path and bare-MCP comments as you suggested.

One thing arriving separately: a warning in `dataplex_post.md` that term associations replace the entity's full `glossaryTerms` list, since the aspect has no server-side merge — so UI-curated terms are dropped. That's pre-existing behaviour shared with Dremio, Cube and Kafka rather than something this PR introduces, but this PR activates the path, so it needed saying. Tracked separately for a real fix via `set_attribution`.

aviraj-gour (approved):

PR #17562, opened 2026-05-22
feat(i18n): extract strings from home pages

v-tarasevich-blitz-brain:
This PR extracts translation keys from homeV2/homeV3 folders


datahub-web-react/src/app/homeV2/content/tabs/discovery/sections/domains/DomainCard.tsx
```diff
@@ -80,9 +82,9 @@ export const DomainCard = ({ domain, assetCount }: Props) => {
                 <Text>
                     <Name>{name}</Name>
                     <Counts>
-                        <Count>{formatNumber(assetCount)} assets</Count>
+                        <Count>{t('domains.assetCount', { count: formatNumber(assetCount) })}</Count>
```
benjiaming: Wait does that work? i18next requires a number but `formatNumber` returns something like "1.2k".

I think we need `t('domains.assetCount', { count: assetCount, formattedCount: formatNumber(assetCount) })`

datahub-web-react/src/app/homeV2/content/tabs/discovery/sections/platform/PlatformCard.tsx
```diff
@@ -75,7 +77,7 @@ export const PlatformCard = ({ platform, count }: Props) => {
 
     return (
         <Tooltip
-            title={`View ${formatNumberWithoutAbbreviation(count)} ${name} assets`}
+            title={t('platforms.viewAssetsTooltip', { count: formatNumberWithoutAbbreviation(count), name })}
```
benjiaming: also needs a number

benjiaming (changes requested):

datahub-web-react/src/i18n/locales/en/module.dataProducts.json
```diff
@@ -0,0 +1,4 @@
+{
+    "moduleName": "Data Products",
+    "moduleDescription": "View the data products inside of this domain"
+}
```
benjiaming: Some of these namespaces are quite short. I would prefer merging them into `modules.json` with a prefix. Not a blocker.

benjiaming (approved):

PR #17567, opened 2026-05-22
docs(assertion): clarify when to use AssertionStdOperator._NATIVE_

jayacryl:
## Summary

`AssertionStdOperator._NATIVE_` was documented as just "Other", which leaves callers (and LLM-driven tooling composing assertions) guessing about its intended use.

This PR spells out:

- **What it's for:** a catch-all for assertions whose check can't be expressed with one of the structured operators (mainly external-system importers like dbt tests and Great Expectations expectations).
- **What to pair it with:** the corresponding native payload — `DatasetAssertionInfo.nativeType` / `nativeParameters` / `logic`, or the free-form fields on `CustomAssertionInfo`. Without that payload the assertion is unparameterised.
- **The gotcha that motivated this:** assertion shapes that don't have a native escape hatch (`FieldValuesAssertion`, `FieldMetricAssertion`) produce an assertion with no executable predicate when `_NATIVE_` is used, and have no rendered description in the UI. Callers should pick a structured operator instead.

This came out of debugging an integration where an LLM, asked to set up a uniqueness check, classified it as `FieldValuesAssertion` ("checks each row") and then reached for `_NATIVE_` because there's no row-level uniqueness predicate — the right shape is `FieldMetricAssertion` with `metric=UNIQUE_PERCENTAGE`. Clearer docs would have prevented that.

### About the `AssertionInfo.schemaVersion` bump

The second file in the diff was modified by the repo's own `.github/scripts/bump_schema_versions.py` pre-commit hook, which conservatively bumps any aspect transitively depending on a changed PDL (`AssertionInfo` includes `AssertionStdOperator`). Happy to drop this from the diff if maintainers prefer to skip the bump for a comment-only change — let me know.

## Checklist

- [x] PR conforms to the Contributing Guideline (`docs` type, no breaking changes).
- [x] No related issues to link.
- [x] N/A — comment-only doc change, no tests to add.
- [x] Docs updated (this PR *is* the doc update).
- [x] No breaking change — no `updating-datahub.md` entry needed.


AdrianMachado (approved):

jayacryl:
@claude fix CI

AdrianMachado:
@claude explain CI issues and fix them

PR #17570, opened 2026-05-23
feat(ingestion/bigid): add BigID privacy classification connector

datahub-akshay:
## What

Adds a new `bigid` ingestion source that syncs BigID's privacy/classification findings into DataHub.

BigID scans structured and unstructured data sources for PII and sensitive data. This connector reads those findings via BigID's REST API and surfaces them in DataHub, **enrichment-first**: by default (`create_datasets: false`) it only enriches Dataset entities that native connectors already created, and never emits structural aspects.

## What it emits

- **Field-level GlossaryTerms** on structured sources (Snowflake, Postgres, MySQL, …) — one term per BigID classifier that fired on the column, applied as a non-destructive PATCH on `editableSchemaMetadata` with full `MetadataAttribution` (finding count, confidence, scan timestamp, connection).
- **Dataset-level GlossaryTerms** on unstructured sources (SharePoint, Google Drive, O365, file shares, AI vector DBs) — aggregated from object-level `attribute_details`, applied as a PATCH.
- **GlobalTags** (BigID sensitivity) and a **riskScore StructuredProperty** on datasets.
- **DatasetProfile** — row/field counts from BigID findings; field-level profiling for structured sources.
- **GlossaryTerm + GlossaryNode entities** — BigID Business Glossary terms under a `bigid` root node; auto-generated classifier terms under a `bigid > Classifier` node; IDSoR attributes under a `bigid > IDSoR` node via three-path resolution (linked term / friendly-name auto-gen / raw-name auto-gen). Term and node URNs are **deterministic GUIDs** derived from stable path keys.
- **Domain entities** (opt-in) — BigID domain/sub-domain hierarchy as DataHub domains, keyed by `DomainKey` GUID and scoped by `env` + `platform_instance`.

## Design notes

- **Non-destructive enrichment.** Field and dataset enrichment use `DatasetPatchBuilder` (PATCH), so the connector adds terms/tags without clobbering aspects owned by other sources.
- **No ghost entities.** Enrichment workunits are emitted with `is_primary_source=False`, so they neither create placeholder datasets nor get tracked by stale-entity removal.
- **Streaming catalog.** The catalog API is processed in a single streaming pass, holding only `O(distinct tags)` in memory rather than buffering the full catalog — safe for 100k+ object deployments.
- **Auth.** Provide `user_token` (long-lived, exchanged for a short-lived session token and auto-refreshed — recommended for scheduled ingestion) or `access_token` (short-lived, used directly). If both are set, `user_token` takes precedence because it can auto-refresh.
- Entity types without SDK V2 builder coverage (GlossaryTerm, GlossaryNode, Domain, StructuredProperty) are emitted via `MetadataChangeProposalWrapper`; dataset enrichment uses `DatasetPatchBuilder`.

## Key config options

| Field | Default | Description |
|---|---|---|
| `bigid_url` | required | BigID instance URL |
| `user_token` / `access_token` | required (one) | Auth credentials (`user_token` wins if both set) |
| `create_datasets` | `false` | Also create dataset stubs for BigID objects |
| `sync_tags` | `true` | Emit sensitivity tags and risk scores |
| `sync_idsor` | `true` | Sync IDSoR attributes as GlossaryTerms |
| `sync_unlinked_classifiers` | `true` | Auto-generate terms for unlinked classifiers |
| `sync_unstructured_enrichment` | `false` | Emit dataset-level GlossaryTerms for unstructured sources |
| `minimum_confidence_threshold` | `0.0` | Skip findings below this confidence (0.0–1.0; HIGH=0.75, MEDIUM=0.50, LOW=0.25) |
| `domain_mode` | `none` | `auto_namespaced` to sync BigID domains as DataHub domains |

## Dependencies

- Depends on **#18138** (`fix(ingestion): emit fieldPath add op for editableSchemaMetadata field patches`). The field-level enrichment PATCH path relies on that `FieldPatchHelper` fix to avoid a GMS `422 fieldPath is required`. That change is split into its own PR for isolated review; it is also included on this branch so this connector is testable standalone until #18138 merges.

## Testing

- **Unit tests**: platform mapping, GUID URN construction, IDSoR three-path resolution, classifier deduplication, confidence-threshold filtering, domain resolution, MetadataAttribution, token precedence, streaming catalog error handling.
- **Integration tests**: golden-file comparison plus structural-invariant tests; fixtures captured from a live BigID instance and sanitized.
- **Live validation**: run against a BigID instance with ~18k catalog objects; 0 failures.

## Checklist

- [x] `@platform_name`, `@config_class`, `@support_status(INCUBATING)` decorators
- [x] `StatefulIngestionSourceBase` + `StaleEntityRemovalHandler`
- [x] Paginated API calls (`PAGE_SIZE=500`), `requests.Session` reuse, generator-based emission
- [x] Unit + integration tests with golden files
- [x] Docs (`bigid_pre.md`, `bigid_post.md`, README) and integrations-catalog logo


treff7es:
## Code + Modeling Review: BigID connector

_Full review across architecture, error handling, type safety, tests, and BigID entity modeling._

---

### Verdict: 🟡 NEEDS CHANGES

Two items should be fixed before merge; everything else is a suggestion.

---

## 🔴 Must Fix

### 1. `exc=` never passed to any report call (17 sites)

Every `self.report.warning()` / `self.report.failure()` call in `bigid_source.py` discards the live exception object:

```python
# Current (all 17 sites):
self.report.warning("ds-connections unavailable", context=str(exc))

# Should be:
self.report.warning("ds-connections unavailable", context=str(exc), exc=exc)
```

The framework's `warning()`/`failure()` methods accept `exc: Optional[BaseException]` and use it to log the full stack trace at DEBUG level. Without it, Sentry receives only a flat string, and post-incident debugging requires reproducing the error. The sole stack trace in the connector is the `logger.exception()` call at the outer `except Exception` branch — the least likely path to fire.

Mechanical fix: add `exc=exc` to all 17 call sites.

---

### 2. Confirmed bug — duplicate term URN per schema field

`tests/integration/bigid/golden/test_bigid_source.json` contains a duplicate: the `customer_email` field has `urn:li:glossaryTerm:bigid.bt_email` appearing **twice** in its `terms` array — once from the classifier path and once from the IDSoR path-1 resolution (which is supposed to *reuse* the existing term, not add it again).

Root cause: `_emit_schema_field_enrichment` appends to `editable_fields[field_path].terms` without deduplicating by URN before appending.

Fix: deduplicate by `urn` before the append, regenerate the golden file, and add a regression test:

```python
def test_no_duplicate_term_urns_per_field():
    mcps = _run_source()
    for mcp in mcps:
        if mcp.get("aspectName") != "editableSchemaMetadata":
            continue
        for field in _parse_aspect(mcp).get("editableSchemaFieldInfo", []):
            urns = [t["urn"] for t in field.get("glossaryTerms", {}).get("terms", [])]
            assert len(urns) == len(set(urns)), (
                f"Duplicate term URNs on {mcp['entityUrn']} field {field['fieldPath']}: {urns}"
            )
```

---

## 🟡 Should Fix

### 3. Two `logger.warning()` calls bypass the ingestion report UI

Users inspecting the DataHub ingestion report page won't see these:

- **Unknown connection type** (`_load_registries`): logged via `logger.warning()` + `report_connection_no_platform()`, but never produces a `StructuredLogEntry`. Users won't know their connection is producing non-canonical URNs.
- **Invalid (non-numeric) `riskScore`** (`_extract_dataset_tag_associations`): logged via `logger.warning()`, structured property silently dropped.

Replace both `logger.warning()` calls with `self.report.warning(...)`.

### 4. `DatasetProfile` modeling has two issues

**a) `fieldCount` → `rowCount` is semantically wrong.** BigID's `fieldCount` is a per-column scan count (sample size), not a table row count. Using `columns[0].fieldCount` as the table-level `rowCount` is inaccurate for partial scans and assumes all columns have the same count.

**b) Overwrite risk.** This connector's UPSERT on `DatasetProfile` will overwrite profiles emitted by native connectors (Snowflake, BigQuery, etc.) and vice versa — run ordering is undefined in multi-source DataHub. Consider either (1) limiting `DatasetProfile` emission to `create_datasets=True` mode, or (2) emitting only field-level profiles without touching table-level `rowCount`/`columnCount`.

---

## ℹ️ Suggestions

### Error handling

- `_emit_domain_entities`: the `try/except` wraps `yield` statements, but the `except` block is unreachable in generator semantics — MCP construction (what can actually throw) happens before the yield. Also, if the first `yield` succeeds and an exception fires before `_emitted_domain_urns.add()`, the domain is partially emitted but untracked. Move MCP construction outside the `try`:

  ```python
  try:
      props_wu = MetadataChangeProposalWrapper(entityUrn=domain_urn, aspect=props).as_workunit()
      status_wu = MetadataChangeProposalWrapper(entityUrn=domain_urn, aspect=StatusClass(removed=False)).as_workunit()
  except Exception as exc:
      self.report.warning("domain-entity-failed", context=f"domain={domain_val!r}: {exc}", exc=exc)
      continue
  yield props_wu
  yield status_wu
  self._emitted_domain_urns.add(domain_urn)
  ```

- The outer `except BigIDAPIError` in `get_workunits_internal` should also call `logger.exception()` (same as the `except Exception` branch below it).
- `(ValueError, TypeError): pass` in `_emit_dataset_profile` for `fieldCount` has no debug log — unlike the identical pattern in `_build_field_profile` which does. Add `logger.debug()` for consistency.

### Tests

- `confidence_level_tag=True` code path is completely untested.
- `owner_type="group"` branch is untested (only `user` and `none` are covered).
- `registries-empty` failure path (both API calls fail → connector emits nothing) has no test.
- `get_idsor_attribute_map` parsing edge cases have no direct unit test (empty `attributeName`, `{}` `friendlyName`, non-dict response).
- The `ds_connections.json` fixture uses `"type": "sharepoint-online"` which is not in `BIGID_TYPE_TO_PLATFORM` (the mapping has `"sharepoint-online-v2"`). The integration test is silently exercising the "unknown type → raw string" fallback path for SharePoint rather than the canonical platform mapping.

### Type design

- `dict[str, tuple[str, Optional[str]]]` for `_idsor_attr_map` — the two tuple positions are both `str | None`, so swapping them is a silent bug mypy can't catch. A small named dataclass is cleaner:
  ```python
  @dataclass(frozen=True)
  class IDSoRAttributeInfo:
      friendly_name: str
      glossary_id: Optional[str]
  ```
- `minimum_confidence_threshold`: prefer `Field(ge=0.0, le=1.0)` over the separate `@field_validator` — one mechanism is better than two, and the constraint becomes visible in the generated JSON Schema.
- The four-position anonymous tuple returned by `_resolve_term_urn_and_pending` and `_resolve_idsor_term` would be safer as a small dataclass — positional destructuring is fragile when the return type is extended.

### Code organization

- `_rank_to_float` and `_CONFIDENCE_FLOAT` are pure stateless utilities imported with a leading underscore from `bigid_report.py` into `bigid_source.py`. They belong in `bigid_utils.py` alongside the other pure helpers.
- The pending-term dispatch loop (`if ":bigid.idsor." in term_urn: ... elif ":bigid.classifier." ...`) is duplicated verbatim in two places (unstructured path ~line 875, schema-field path ~line 1097). Extract to `_emit_pending_terms()`.
- `self.client: Optional[BigIDClient] = None` followed immediately by `self.client = BigIDSource._build_client(config)` — the `None` assignment is dead code; remove it.

### Platform mapping gaps

Based on BigID's documented data source coverage, these are likely missing:
- `mongodb` → `mongodb` (BigID has an official MongoDB partnership)
- `azure-sql` → `mssql`
- `sap-hana` → `saphana` (BigID-SAP partnership documented)
- `adls-v2` → `adls-gen2`

Missing entries degrade gracefully (raw type string used as platform with a warning), so not blocking.

---

## Modeling assessment (BigID concepts → DataHub entities)

Researched against BigID's developer portal and API docs to validate the design choices.

| Decision | Verdict |
|---|---|
| Business Glossary → `GlossaryTerm` entities | ✅ Correct |
| Classifiers (linked) → `GlossaryTerm` + `MetadataAttribution` on schema fields | ✅ Correct — idiomatic DataHub PII classification pattern |
| Classifiers (unlinked) → auto-generated `GlossaryTerm` | ✅ Defensible |
| IDSoR attributes → `GlossaryTerm` | 🟡 Acceptable but semantically imprecise — IDSoR is *correlation* (whose data?) not classification (what type?). `StructuredProperty` would be more accurate; GlossaryTerms win on UI discoverability. The `bigid.idsor` sub-node namespace makes the distinction visible. |
| Sensitivity tags → `GlobalTag` | ✅ Correct |
| Risk score → `StructuredProperty` | ✅ Correct |
| Domains → `Domain` entities | ✅ Correct |
| `DatasetProfile` from column profiles | 🟡 See above |

All six API endpoints are verified against BigID's developer portal. The auth quirk (raw token, no `Bearer ` prefix) and `totalRowsCounter`-echoes-`limit` pagination behavior are real BigID-specific behaviors the connector handles correctly.

---

## Positive highlights

- Excellent file decomposition — config, API client, report, and utils all in separate files with clear responsibilities.
- `requests.Session` + `Retry` adapter with `allowed_methods=["GET"]` is textbook HTTP client setup.
- `time_machine` (not `freezegun`) used correctly throughout all tests.
- IDSoR three-path resolution is well-tested with distinct cases for path 1 (linked term reuse), path 2 (friendly-name auto-gen), and path 3 (raw-name fallback).
- No `cast()`, no bare `# type: ignore`, all public methods typed — clean mypy hygiene.
- The class docstring reads as a clear architecture decision record: SDK V2 deviation, enrichment semantics, `dataPlatformInstance` omission — all explained with reasoning. This is exactly the right approach for documented deviations.
- `MetadataAttribution` on GlossaryTermAssociations is the correct DataHub idiom for preserving scanner provenance (confidence, count, timestamp, connection).

gabe-lyons:
@acrylJonny will review

max-datahub:
## Review — BigID connector

Thanks for this — it's a genuinely well-built connector: clean file separation (`config`/`constants`/`models`/`api`/`utils`/`report`), unusually honest capability declarations, strong error-isolation tests, careful MCP emission ordering, and clean credential handling (`SecretStr`, `.get_secret_value()` only at the client boundary, never logged or emitted). The comments below are ranked most-severe first; the blockers are worth resolving before merge, and I've tried to make each one actionable with a file/line reference.

### 🔴 Blockers

**1. Enrichment aspects on externally-owned datasets are emitted as `is_primary_source=True` → stale-removal can soft-delete real datasets**
Every dataset-targeting workunit in `_emit_platform_instance`, `_emit_dataset_tags`, `_emit_schema_field_enrichment`, `_emit_dataset_profile`, and `_emit_unstructured_dataset_enrichment` uses the default `is_primary_source=True`. Because `BigIDSourceReport` extends `StaleEntityRemovalSourceReport`, the base `Source.get_workunit_processors()` auto-wires `AutoStaleEntityRemovalProcessor`, so those target-dataset URNs (Snowflake/Postgres/etc. — owned by *other* connectors) get tracked in this connector's checkpoint state. When `stateful_ingestion` is enabled (the connector advertises `DELETION_DETECTION`), a later run where a table doesn't reappear — confidence-threshold change, `connection_pattern` change, an incomplete scan, or a mid-pagination API error — will emit `status(removed=True)` on that live, externally-owned dataset.

This does **not** trigger in the default config (`stateful_ingestion` unset → checkpointing disabled), but it does on the advertised stateful path. Fix: pass `is_primary_source=False` on all enrichment aspects that target datasets this connector doesn't create; keep the `create_datasets=True` creation path as primary. This mirrors how dbt/airbyte/matillion handle foreign-owned datasets.

**2. `DatasetFieldProfile` emits raw column min/max (and `sampleValues`) with no opt-out and no exclusion of classified columns**
`bigid_utils._build_field_profile` (~L158-186) sets `min`/`max` from `minNum`/`maxNum` or `minLexStr`/`maxLexStr`, and for strings sets `sampleValues=[min_lex_str, max_lex_str]` — i.e. the actual lexicographically smallest/largest values in the column. `_emit_dataset_profile` runs for any structured object with columns; there's no profiling on/off config flag and no exclusion for columns BigID has classified as sensitive. For a privacy connector specifically, republishing raw column extreme values (surfaced in the UI as "Sample Values") to everyone with profile read access is risky-by-default. Note: the current test fixtures leave the PII string columns' `columnProfile` empty, so the golden files don't exercise this — but nothing in the code prevents it when BigID does profile a classified string column. Fix: drop string `min`/`max`/`sampleValues` (individual record values); keep only aggregates (`uniqueCount/Proportion`, `nullCount/Proportion`), and/or gate value-derived stats behind a default-off flag whose description states it can surface real column values.

**3. Glossary-term definition clobbering: `urn:li:glossaryTerm:bigid.bt_email` is emitted with two conflicting `glossaryTermInfo` aspects**
The business-glossary sync (`_emit_single_glossary_term`) emits `name="Email Address"`, `parentNode=…:bigid`, while the IDSoR path-1 linked-term path (`_emit_idsor_linked_term`) emits `name="Email"`, `parentNode=…:bigid.idsor` for the same URN (visible in both golden files). The code comment assumes linked IDSoR terms aren't present in `business_glossary_items`, but the fixtures show the realistic case where they are. Under UPSERT with the real emission order, last-write-wins renames a governed business term and reparents it under the IDSoR node on every run — and existing tests assert this as expected. Fix: share a "term URN already defined by glossary sync" set and skip the IDSoR linked-term emission for those URNs; add an invariant that no `(entityUrn, "glossaryTermInfo")` pair is emitted twice.

**4. Silent catalog failures**
`BigIDClient.get_catalog_objects` reads `data.get("results", [])`, so a 200 response that's a dict without a `results` key (error envelope / API drift) yields zero objects, no report entry, and a "successful" run — unlike the sibling methods, which raise `BigIDAPIError` on an unexpected shape. Separately, a mid-pagination `BigIDAPIError` is downgraded to `report.warning` rather than `report.failure`. Combined with #1, both paths become silent under-enrichment (always) and mass soft-deletion (on the stateful path, since the stale-removal fail-safe only skips on `report.failures`, not warnings). Fix: raise on a missing `results` key, and treat an incomplete catalog fetch as a failure (or suppress stale removal for that run).

### 🟡 Should-fix

- **SDK V2 usage.** New connectors are expected to use the SDK v2 builders. The PR states SDK v2 lacks builders for GlossaryTerm/GlossaryNode/Domain/StructuredProperty, but v2 builders do exist for `GlossaryTerm`, `GlossaryNode`, `Tag`, and `Dataset` — only `Domain` and `StructuredProperty` genuinely have no SDK v2 entity class today (only the `HasDomain`/`HasStructuredProperties` assignment mixins). The cleanest path is to add SDK v2 `Domain` and `StructuredProperty` entity classes upstream (the `DomainUrn`/`StructuredPropertyUrn` URNs and `DomainProperties`/`StructuredPropertyDefinition` aspects already exist, so each is a thin wrapper analogous to `sdk/glossary_node.py`) and migrate the whole connector to v2; a reasonable interim is to migrate the already-supported entities now and keep a narrow, clearly-commented MCPW exception for Domain/StructuredProperty. As a bonus, v2 auto-emits the `status` aspect, removing most of the manual `StatusClass(removed=False)` boilerplate. (Note: this is independent of blocker #1 — the `is_primary_source` fix still applies to enrichment patches.)
- **Pydantic `ValidationError` isn't caught** alongside `BigIDAPIError` in `_load_registries` / the catalog fetch loop, so one malformed API record aborts the whole pass. The models are permissive, so it's uncommon, but the asymmetry with the well-guarded processing loop is real.
- **Plain `http://` is accepted silently** by the `bigid_url` validator; the token and sensitive classification metadata then travel in cleartext. (TLS verification itself is on — no `verify=False`.) Consider warning on `http`, or an explicit opt-in.
- **`DatasetProfile.columnCount` is still emitted unconditionally** regardless of `create_datasets`, which can clobber profiles from the native connector that owns the dataset.
- **Dedup-by-first-seen** in `_emit_schema_field_enrichment` keeps whichever attribution appears first when two findings resolve to the same term URN; the golden shows a HIGH-confidence IDSoR attribution being dropped in favor of a MEDIUM classifier purely by ordering. Prefer the highest-confidence attribution on merge.
- **Deletion detection has no test**, and the integration test uses a custom `UPDATE_GOLDEN=1` mechanism rather than `mce_helpers.check_golden_file`. A two-run stale-removal test would protect the connector's headline stateful feature and catch #1.

### ℹ️ Suggestions
- `TermResolution.clf_type: str` → `Literal[...]`; add a URN validator for `risk_score_structured_property_urn`; `connection_pattern` should use `default_factory=`; bound `timeout`/`max_retries`.
- Some emission boilerplate can be collapsed (repeated `MCPW(...).as_workunit()` + `StatusClass` pairs, duplicated tag/term-URN literals, an unused `attr` param in `_resolve_idsor_term`).
- Consider `builder.make_dataset_urn_with_platform_instance` instead of the hand-rolled URN percent-encoding in `bigid_utils`.
- Minor: the PR description says `minimum_confidence_threshold` defaults to `MEDIUM`, but the code defaults to `0.0` (no filtering).

### CI
`ci (… testQuick)` is currently failing with mypy (172 errors, concentrated in `tests/unit/bigid/test_bigid_source.py` and `test_bigid_config.py` — untyped mocks: `method-assign`, `.side_effect`/`.return_value` on plain callables, `SecretStr | None` union-attr). `pytest` passes; this is a typing-only blocker in the test files.

Happy to help talk through any of these, particularly the stateful/`is_primary_source` interaction and the SDK v2 path.

acrylJonny (approved):
Some changes made - now this is looking good to merge

PR #17575, opened 2026-05-25
feat(ingestion/gcs): add workload_identity auth type for GKE Workload Identity

alokr-dhub:
## Summary

Adds a `workload_identity` auth type to the GCS connector that uses
[Application Default Credentials](https://cloud.google.com/docs/authentication/application-default-credentials)
via `google.auth.default()`. This is the recommended authentication option for
DataHub running inside GKE with Workload Identity enabled — no credential files
or secrets are required; the pod's Kubernetes Service Account is used automatically.

**Before:** GCS supported two auth types:

- `hmac` — long-lived HMAC keys (default)
- `workload_identity_federation` — keyless auth for workloads _outside_ GCP (AWS, Azure, on-premises) using an explicit WIF configuration JSON

**After:** A third type is added:

- `workload_identity` — keyless auth for GKE pods with Workload Identity enabled. Set `auth_type: workload_identity` and nothing else; credentials are sourced automatically from the GKE metadata server.

### Key changes

**Source (`gcs_source.py`)**

- New `GCSAuthType.WORKLOAD_IDENTITY` enum value and config validator that rejects WIF config fields when this auth type is set
- `_setup_adc_credentials()` calls `google.auth.default()` with `cloud-platform` scope; wraps `DefaultCredentialsError` and `GoogleAuthError` in descriptive `ValueError`s matching the WIF error convention; logs a warning when no project ID is returned
- Extracted `_build_gcs_oauth_aws_config()` and `_build_s3_config()` helpers to eliminate duplication between the WIF Federation and ADC branches
- `_apply_oauth()` now asserts credentials are set rather than silently skipping hook registration (which would produce opaque 403s)
- Instance attributes renamed `_wif_credentials` → `_gcp_credentials`, `_wif_project_id` → `_gcp_project_id` (both auth types populate them)
- `credentials` parameter of `_register_gcs_oauth_before_send` tightened from `Any` to `Credentials`
- `get_report()` annotated with `-> GCSSourceReport`

**Tests (`test_gcs_source.py`)**

- Happy-path tests: `workload_identity` source creation with and without project ID
- Error-path tests: `DefaultCredentialsError` and `GoogleAuthError` both produce clear `ValueError`
- Config validator test parametrized over all three WIF fields
- All imports moved to module level (no late imports in function bodies)

**Docs**

- `gcs_pre.md`: updated Prerequisites to document all three auth types with step-by-step GKE Workload Identity setup instructions
- `gcs_recipe.yml`: added `workload_identity` recipe example

## Test plan

- [x] `./gradlew :metadata-ingestion:lintFix` — passes clean
- [x] `pytest tests/unit/test_gcs_source.py` — 33/33 passed
- [x] Docs pass `mdPrettierCheck`

## Checklist

- [x] Tests added/updated
- [x] Docs added/updated
- [x] No breaking changes — additive config only; existing `hmac` and `workload_identity_federation` recipes are unchanged

### DIFF
--- a/metadata-ingestion/docs/sources/gcs/gcs_pre.md
+++ b/metadata-ingestion/docs/sources/gcs/gcs_pre.md
@@ -10,10 +10,11 @@ and uses DataHub S3 Data Lake integration source under the hood. Refer section [
 
 Before running ingestion, ensure network connectivity to the source, valid authentication credentials, and read permissions for metadata APIs required by this module.
 
-The GCS source supports two authentication methods:
+The GCS source supports three authentication methods:
 
-1. **HMAC Keys** (default): Service account key-based authentication using Google Cloud Storage HMAC keys.
-2. **Workload Identity Federation**: Keyless, token-based authentication. Recommended for Kubernetes/GKE workloads (avoids storing service account keys), cross-cloud authentication (AWS/Azure → GCP), and environments requiring automatic credential rotation.
+1. **HMAC Keys** (`hmac`, default): Long-lived key-based authentication using Google Cloud Storage HMAC keys. Suitable for simple setups and service accounts outside GCP.
+2. **GKE Workload Identity** (`workload_identity`): Keyless authentication using [Application Default Credentials](https://cloud.google.com/docs/authentication/application-default-credentials). The recommended option when DataHub runs inside GKE with Workload Identity enabled — no credential config required.
+3. **Workload Identity Federation** (`workload_identity_federation`): Keyless, token-based authentication for workloads running _outside_ GCP (AWS, Azure, on-premises). Exchanges an external identity token for short-lived GCP credentials via GCP's STS endpoint.
 
 #### HMAC Authentication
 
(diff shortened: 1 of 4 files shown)


sgomezvillamor (approved):
**⚠️ Warning — credential file handling**
We recently fixed BigQuery to never write credentials to a tempfile or env var. The new `workload_identity` path here is clean — fully in-memory. However, the pre-existing `gcp_wif_configuration` option (file path) means the *user* has to write a WIF JSON file to disk as part of their setup, so end-to-end someone is still writing credentials to a file. Please confirm with Piotr this is acceptable before it surfaces as a security concern. If it is, we should consider limiting WIF support to `gcp_wif_configuration_json` and `gcp_wif_configuration_json_string` only.

**💡 Suggestion — reject `credential` with `workload_identity`**
If `auth_type: workload_identity` is set alongside a `credential` (HMAC key), it is silently ignored. WIF config fields are correctly rejected in this case; HMAC fields should be too — a clear `ValueError` avoids user confusion.

**💡 Suggestion — stale class comment on `GCSOAuthAwsConnectionConfig`**
The comment still implies this class is only used for WIF federation. Now that it also serves the ADC path, it should mention both OAuth-based auth types.

**💡 Suggestion — `assert` vs explicit `raise` in `_apply_oauth`**
`assert creds is not None` is stripped in optimised builds (`python -O`). Use `raise RuntimeError(...)` explicitly so the fail-fast guarantee holds in production.

**❓ Question — connector tests**
Should we add a case exercising `workload_identity` in the integration/connector test suite?

---

alokr-dhub:
> **⚠️ Warning — credential file handling** We recently fixed BigQuery to never write credentials to a tempfile or env var. The new `workload_identity` path here is clean — fully in-memory. However, the pre-existing `gcp_wif_configuration` option (file path) means the _user_ has to write a WIF JSON file to disk as part of their setup, so end-to-end someone is still writing credentials to a file. Please confirm with Piotr this is acceptable before it surfaces as a security concern. If it is, we should consider limiting WIF support to `gcp_wif_configuration_json` and `gcp_wif_configuration_json_string` only.
> 
> **💡 Suggestion — reject `credential` with `workload_identity`** If `auth_type: workload_identity` is set alongside a `credential` (HMAC key), it is silently ignored. WIF config fields are correctly rejected in this case; HMAC fields should be too — a clear `ValueError` avoids user confusion.
> 
> **💡 Suggestion — stale class comment on `GCSOAuthAwsConnectionConfig`** The comment still implies this class is only used for WIF federation. Now that it also serves the ADC path, it should mention both OAuth-based auth types.
> 
> **💡 Suggestion — `assert` vs explicit `raise` in `_apply_oauth`** `assert creds is not None` is stripped in optimised builds (`python -O`). Use `raise RuntimeError(...)` explicitly so the fail-fast guarantee holds in production.
> 
> **❓ Question — connector tests** Should we add a case exercising `workload_identity` in the integration/connector test suite?
> 

@sgomezvillamor I think you are referring to the PR #[17375](https://github.com/datahub-project/datahub/pull/17375). The fix removed the side-effects where the SQLAlchemy engine wrote creds to a /tmp folder implicitly. However, `gcp_wif_configuration` is still a supported option for bigquery and other gcp connectors. We can plan to deprecate the file_path config option as follow up issue. cc @skrydal

sgomezvillamor:
> However, `gcp_wif_configuration` is still a supported option for bigquery and other gcp connectors.

good call! if `gcp_wif_configuration` was not a concern before, it shouldn't be now. looks good!

PR #17582, opened 2026-05-26
fix(ingest/powerbi): emit lineage for paginated reports with embedded RDL datasources

alfiyas-datahub:
## Summary

PowerBI paginated reports (RDL) that connect directly to a SQL Server / Oracle / PostgreSQL / OleDb backend instead of a shared PowerBI dataset were silently emitting zero upstream lineage. The connector treats every report the same and only resolves lineage via `report.dataset_id`, which the workspace scan API never returns for embedded RDL datasources. Result: 0/N paginated reports get a `datasetEdges` aspect, with no warning in the ingestion report.

This PR adds a fallback path that calls `GET /groups/{groupId}/reports/{reportId}/datasources` for `PaginatedReport`s without a `dataset_id`, and turns the returned `datasourceType` + `connectionDetails.{server,database}` into upstream `datasetEdges` using the existing platform-mapping machinery (`SupportedDataPlatform`, `server_to_platform_instance`).

### What changes

- New `ReportDatasource` dataclass + `Report.datasources` field for the new payload (`data_classes.py`).
- New `Constant.REPORT_DATASOURCES` endpoint key and `get_report_datasources()` on `DataResolverBase`, with a regular-API implementation and an admin-API no-op (`config.py`, `data_resolver.py`). PowerBI doesn't expose an admin variant of this endpoint, so `admin_apis_only` deployments get an empty list.
- New branch in `PowerBiAPI.get_reports()` that only fetches `/datasources` for `PaginatedReport`s without a `dataset_id`, so zero extra HTTP traffic for any existing code path (`powerbi_api.py`).
- New `paginated_report_datasource_urns()` helper that maps each `ReportDatasource` to a DataHub URN via `SupportedDataPlatform` + `server_to_platform_instance`, and folds the URNs into the existing `dataset_edges` set in `report_to_datahub_work_units()` (`powerbi.py`). No new aspect type, no new MCP shape - the lineage rides the same `DashboardInfo.datasetEdges` path regular reports use today.
- New structured `Missing Lineage For Paginated Report` info event so silent skips surface in the ingestion report. The existing `Missing Lineage For Report` event only fires when `dataset_id` is present but unresolvable, which is why this bug class has been invisible.
- New integration test `test_powerbi_paginated_report_rdl_lineage` + dedicated mock fixture covering exactly the bug-case shape (no `datasetId`, with `/datasources` returning SQL Server connection details).

### Observability

After this change, every paginated report that doesn't resolve to upstream lineage produces a structured info event:

- `Missing Lineage For Paginated Report` - report has no `dataset_id` and `/datasources` returned nothing (admin-only mode, permission gap, or genuinely empty)
- `Unmapped Paginated Report Datasource` - `/datasources` returned a `datasourceType` that has no entry in `SupportedDataPlatform` (e.g., a Microsoft-only type we haven't mapped yet)

Both messages name the report and (where applicable) the server so operators can diagnose missing lineage without re-reading the connector source.


metadata-ingestion/src/datahub/ingestion/source/powerbi/rest_api_wrapper/data_resolver.py
```diff
@@ -467,6 +475,32 @@ def get_pages_by_report(self, workspace: Workspace, report_id: str) -> List[Page
...
+        logger.debug(f"Request to paginated-report datasources URL={endpoint}")
+        response = self._request_session.get(
+            endpoint,
+            headers=self.get_authorization_header(),
+        )
+        if is_http_failure(
+            response,
+            f"Unable to fetch datasources for paginated report {report_id}",
+        ):
+            return []
```
treff7es: Shouldn't we need to log something here? This way we just silently ignore this.

metadata-ingestion/src/datahub/ingestion/source/powerbi/rest_api_wrapper/powerbi_api.py
```diff
@@ -292,6 +293,37 @@ def get_reports(self, workspace: Workspace) -> Dict[str, Report]:
...
                     )
+            elif report.type is ReportType.PaginatedReport:
+                # RDL reports with embedded datasources don't carry a datasetId
+                # in the scan response; fall back to /reports/{id}/datasources.
+                try:
+                    report.datasources = self._get_resolver().get_report_datasources(
+                        workspace=workspace, report_id=report.id
+                    )
+                except Exception:
+                    self.log_http_error(
```
treff7es: We should pass Exception and log as well otherwise we won't be able to see the root cause

metadata-ingestion/src/datahub/ingestion/source/powerbi/rest_api_wrapper/powerbi_api.py
```diff
@@ -292,6 +293,37 @@ def get_reports(self, workspace: Workspace) -> Dict[str, Report]:
...
+                        message=(
+                            f"Unable to fetch datasources for paginated report "
+                            f"{report.name}({report.id}) in workspace {workspace.name}"
+                        )
+                    )
+                    report.datasources = []
+
+                if not report.datasources:
```
treff7es: After the try/except, both paths set report.datasources = []. The subsequent if not report.datasources: can't tell them apart — users see "grant Report.Read.All" even when the API call succeeded and the report genuinely has no datasources.

metadata-ingestion/tests/integration/powerbi/test_powerbi.py
```diff
@@ -197,6 +197,89 @@ def test_powerbi_ingest(
...
+                    and "datasetEdges" in op.get("path", "")
+                    and isinstance(op.get("value"), dict)
+                    and "destinationUrn" in op["value"]
+                ):
+                    edge_destinations.append(op["value"]["destinationUrn"])
+
+    # When no server_to_platform_instance mapping exists, the default
+    # PlatformDetail returns env=PROD (matching existing behavior in
+    # extract_directlake_lineage), regardless of the source-level env.
+    expected = (
```
treff7es: Every comparable integration test (test_powerbi_lineage, test_powerbi_report, etc.) verifies against a committed golden file via mce_helpers.check_golden_file. The new test only checks a single string in edge_destinations, leaving the full MCP structure (DashboardInfo aspects, UpstreamLineage, SubTypes) unvalidated. Future regressions that corrupt the output format won't be caught.

metadata-ingestion/src/datahub/ingestion/source/powerbi/rest_api_wrapper/data_classes.py
```diff
@@ -306,6 +306,28 @@ class ReportType(Enum):
...
+    Populated from ``GET /groups/{groupId}/reports/{reportId}/datasources``;
+    not surfaced by the workspace scan API.
+    """
+
+    datasource_type: Optional[str]
+    server: Optional[str]
+    database: Optional[str]
+
+    @staticmethod
+    def from_raw(raw: Dict[str, Any]) -> "ReportDatasource":
```
treff7es: Consider having from_raw return Optional["ReportDatasource"], logging a DEBUG message and returning None when datasource_type or server is absent. This eliminates the degenerate all-None state and makes the calling code cleaner.

metadata-ingestion/src/datahub/ingestion/source/powerbi/powerbi.py
```diff
@@ -633,6 +633,66 @@ def to_datahub_dataset(
...
+        """Resolve a paginated report's embedded datasources to upstream URNs.
+
+        Produces coarse-grained (server[.database]) URNs; table-level lineage
+        would require parsing the .rdl XML.
+        """
+        urns: List[str] = []
+        if not report.datasources:
+            return urns
+
+        pbi_to_dh = {
```
treff7es: SupportedDataPlatform is a stable enum; the {powerbi_name: pair} dict is rebuilt on every report. Promote to a module-level constant and reuse it (and in validate_dataset_type_mapping).

treff7es:
The test calls:
```
read_mock_data(
    pytestconfig.rootpath
    / "tests/integration/powerbi/mock_data/paginated_report_rdl_datasources.json"
)
```
read_mock_data does a plain open() with no fallback. The file paginated_report_rdl_datasources.json is not in the PR. Every CI run will raise FileNotFoundError before a single assertion executes. This is an unconditional blocker — the test cannot pass.

Fix: Add the mock JSON file to tests/integration/powerbi/mock_data/ in the PR.

metadata-ingestion/src/datahub/ingestion/source/powerbi/powerbi.py
```diff
@@ -633,6 +634,52 @@ def to_datahub_dataset(
...
+        report: powerbi_data_classes.Report,
+    ) -> List[str]:
+        """Resolve a paginated report's embedded datasources to upstream URNs.
+
+        Produces coarse-grained (server[.database]) URNs; table-level lineage
+        would require parsing the .rdl XML.
+        """
+        urns: List[str] = []
+        for ds in report.datasources:
+            pair = POWERBI_TYPE_TO_DATA_PLATFORM_PAIR.get(ds.datasource_type)
```
treff7es: The existing M-Query lineage path (line 373) uses `self.__config.is_platform_in_dataset_type_mapping(platform_name)` and `get_from_dataset_type_mapping()`, which respect the user-configurable `dataset_type_mapping` config and apply space-normalisation (e.g. `"Amazon Redshift"` → `"AmazonRedshift"`). This path calls `POWERBI_TYPE_TO_DATA_PLATFORM_PAIR.get(ds.datasource_type)` directly, bypassing both.

Two consequences:
1. A user who has narrowed `dataset_type_mapping` will still get lineage for excluded platforms from paginated reports.
2. Datasource type names from `/datasources` that include spaces (if the API ever returns them) won't match the enum keys and will silently fall through to the "Unmapped" info event.

Either use `get_from_dataset_type_mapping(ds.datasource_type)` for consistency with the existing lineage path, or add a comment explaining why paginated reports intentionally bypass `dataset_type_mapping`.

metadata-ingestion/src/datahub/ingestion/source/powerbi/rest_api_wrapper/data_resolver.py
```diff
@@ -826,6 +859,13 @@ def get_tiles_endpoint(self, workspace: Workspace, dashboard_id: str) -> str:
     def get_pages_by_report(self, workspace: Workspace, report_id: str) -> List[Page]:
         return []  # Report pages are not available in Admin API
 
+    def get_report_datasources(
+        self, workspace: Workspace, report_id: str
+    ) -> List[ReportDatasource]:
+        # No admin-API variant exists for /reports/{id}/datasources;
+        # callers in admin_apis_only mode will get empty lineage.
+        return []
```
treff7es, line 871: The stub returns `[]` with no log, so `admin_apis_only=True` is indistinguishable from "the report genuinely has no datasources" at the call site. Both paths land in the same `reporter.info("Missing Lineage For Paginated Report", ...)` block, with the admin-only hint buried in the message body.

At minimum add a `logger.debug("get_report_datasources is a no-op in admin_apis_only mode")` here. Better: check `self.__config.admin_apis_only` before the `try` in `powerbi_api.py` and emit a distinct event (e.g. `"Paginated Report Lineage Unavailable"`) so users in admin-only mode get an unambiguous, actionable message.

metadata-ingestion/src/datahub/ingestion/source/powerbi/rest_api_wrapper/data_classes.py
```diff
@@ -306,6 +306,36 @@ class ReportType(Enum):
...
+        if not datasource_type or not server:
+            logger.debug(
+                "Skipping report datasource without datasourceType/server: %r",
+                raw,
+            )
+            return None
+        return ReportDatasource(
+            datasource_type=datasource_type,
+            server=server,
+            database=(raw.get("connectionDetails") or {}).get("database"),
```
treff7es: `raw.get("connectionDetails") or {}` is evaluated three times across lines 325 and 335. Extract it once:

```python
connection = raw.get("connectionDetails") or {}
server = connection.get("server")
if not datasource_type or not server:
    ...
return ReportDatasource(
    datasource_type=datasource_type,
    server=server,
    database=connection.get("database"),
)
```

treff7es (approved):

PR #17583, opened 2026-05-26
feat(i18n): extract keys from entity validations tab

v-tarasevich-blitz-brain:
This PR extracts i18n keys from `src/app/entityV2/shared/tabs/Dataset/Validations/`

Some functions use string building which are not translatable right now without refactoring of them.


datahub-web-react/src/app/entityV2/shared/tabs/Dataset/Validations/assertion/profile/shared/result/AssertionResultPopoverContent.tsx
```diff
@@ -210,29 +223,35 @@ const MessageSection = ({ errorMessage, show }: { errorMessage?: string; show: b
...
 
 const ExternalResultsSection = ({ assertion, result }: { assertion: Assertion; result?: AssertionResult }) => {
+    const { t } = useTranslation('entity.validations');
     if (!isExternalAssertion(assertion)) return null;
 
     const externalResultsSections: React.ReactNode[] = [];
     if (result?.nativeResults?.length) {
         externalResultsSections.push(
             <ThinDivider key="external-results-divider" />,
+            /* eslint-disable i18next/no-literal-string -- key:value colon separator is universal formatting */
```
benjiaming: Can we create a followup ticket to address this.

benjiaming:
This would be easier to review if the base branch is temporarily set to vt--cat-2139--extract-keys-from-home-pages.

datahub-web-react/src/app/entityV2/shared/tabs/Dataset/Validations/assertion/profile/shared/result/AssertionResultPopoverContent.tsx
v-tarasevich-blitz-brain: I've updated the eslint rule

datahub-web-react/src/i18n/locales/en/common.labels.json
```diff
@@ -0,0 +1,9 @@
+{
+    "category": "Category",
+    "details": "Details",
+    "na": "N/A",
+    "name": "Name",
+    "status": "Status",
+    "tags": "Tags",
+    "type": "type"
```
benjiaming, line 8: Nit: questionable if lowercase labels should be here or moved to `entity.validations`.

benjiaming (approved):

PR #17584, opened 2026-05-26
fix(ingest/aerospike): pass socket_timeout via query.results() policy

dotan-mor:


metadata-ingestion/src/datahub/ingestion/source/aerospike.py
```diff
@@ -289,15 +289,18 @@ def construct_schema_aerospike(
     if sample_size:
         query.max_records = sample_size
     query.records_per_second = records_per_second
+
+    policy: Dict[str, Any] = {"max_retries": 0}
```
rajatoss, line 291: Dict[str, int]

metadata-ingestion/src/datahub/ingestion/source/aerospike.py
```diff
@@ -669,6 +678,7 @@ def xdr_sets(self, namespace: str, sets: List[str]) -> Dict[str, List[str]]:
                     .split("\n")[0]
                 )
             except Exception as e:
+                logger.exception("XDR per-DC query failed for %s/%s", namespace, dc)
```
rajatoss, line 675: why we are adding this when we already do 
```
 self.report.warning(
                    message="Failed to retrieve XDR config for DC",
                    context=f"{namespace}/{dc}",
```
similarly above line 659

metadata-ingestion/tests/unit/test_aerospike_source.py
```diff
@@ -0,0 +1,690 @@
...
+        # When the timeout knob is unset, only max_retries=0 remains; the
+        # aerospike client's own defaults apply for socket_timeout/total_timeout.
+        assert query.policy_seen == {"max_retries": 0}
+
+    def test_aerospike_query_rejects_socket_timeout_attribute(self):
+        # Regression: aerospike.Query is a C extension that rejects unknown
+        # attribute assignment. _StrictQuery mirrors that contract via __slots__.
+        # If the fix ever regresses to `query.socket_timeout = X`, the policy
+        # tests above will fail with the same AttributeError as the original bug.
+        query = _StrictQuery([])
```
rajatoss: I think this test is repetitive, and can be deleted as something like this is already covered in `test_passes_socket_timeout_via_policy`

metadata-ingestion/tests/unit/test_aerospike_source.py
```diff
@@ -0,0 +1,690 @@
...
+# 2. Hosts validator                                                          #
+# --------------------------------------------------------------------------- #
+
+
+class TestHostsValidator:
+    def test_rejects_single_element_tuple(self):
+        with pytest.raises(ValueError, match="at least a hostname and port"):
+            AerospikeConfig.model_validate({"hosts": [("h",)]})
+
+    def test_rejects_non_integer_port(self):
```
rajatoss: is there a better way to test this, pinning to exact error message is a bit brittle

rajatoss:
I have reviewed it, if possible please check the tests you have added for duplicate nature, I think some of them can be removed or at-least combined

metadata-ingestion/src/datahub/ingestion/source/aerospike.py
dotan-mor: Aerospike policy is `Dict[str, Any]`
https://aerospike-python-client.readthedocs.io/en/latest/client.html#aerospike-policies
It can have a few different types, this way it supports future changes in case there will be a need to add policy settings of other types like **expressions**

rajatoss:
LGTM @kevinkarchacryl approving. Thanks @dotan-mor.

rajatoss (approved):

PR #17591, opened 2026-05-26
fix(graphql): always register semantic-search schema; stub resolver when disabled

nwadams:
## Summary

When `SemanticSearchService` is null (the default in OSS / local dev with `ELASTICSEARCH_SEMANTIC_SEARCH_ENABLED=false`), `SemanticSearchPlugin` was registering neither the GraphQL schema nor the resolvers. Clients calling `semanticSearchAcrossEntities` got a `FieldUndefined` validation error — a schema-level error that varies by GMS configuration and is indistinguishable from a typo on the client side.

This makes the schema always-defined and registers stub `DataFetcher`s that throw `SemanticSearchDisabledException` when the service is unavailable. Clients now see a stable runtime error (`"Semantic search is disabled in this environment"`) and can reliably detect the disabled state to fall back to keyword search.

Discovered while debugging silent zero-hit failures in Ask DataHub on OSS quickstart environments (see internal report; the integrations service couldn't tell apart "semantic search is off" from "field is misspelled" and swallowed both as empty results).

## Changes

- `SemanticSearchPlugin.getSchemaFiles()` unconditionally returns `semantic-search.graphql`.
- `SemanticSearchPlugin.configureExtraResolvers()` registers stub fetchers (lambda → throws `SemanticSearchDisabledException`) for both `semanticSearch` and `semanticSearchAcrossEntities` when the service is null.
- Updated `SemanticSearchPluginTest` to reflect the new contract — schema is always loaded; invoking either stub fetcher throws `SemanticSearchDisabledException`.

The "happy path" (service available) is unchanged.

## Test plan

- [x] `./gradlew :datahub-graphql-core:test --tests SemanticSearchPluginTest` — 10 passing
- [ ] Verify against a GMS with semantic search enabled — existing semantic search behavior unchanged
- [ ] Verify against a GMS with semantic search disabled — calling `semanticSearchAcrossEntities` returns a clear runtime error mentioning "disabled", not `FieldUndefined`

### DIFF
--- a/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/plugins/SemanticSearchPlugin.java
+++ b/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/plugins/SemanticSearchPlugin.java
@@ -8,9 +8,11 @@ import com.linkedin.datahub.graphql.resolvers.semantic.SemanticSearchResolver;
 import com.linkedin.datahub.graphql.types.EntityType;
 import com.linkedin.datahub.graphql.types.LoadableType;
 import com.linkedin.entity.client.EntityClient;
+import com.linkedin.metadata.search.SemanticSearchDisabledException;
 import com.linkedin.metadata.search.SemanticSearchService;
 import com.linkedin.metadata.service.FormService;
 import com.linkedin.metadata.service.ViewService;
+import graphql.schema.DataFetcher;
 import graphql.schema.idl.RuntimeWiring;
 import java.util.Collection;
 import java.util.Collections;
@@ -69,11 +71,11 @@ public class SemanticSearchPlugin implements GmsGraphQLPlugin {
 
   @Override
   public List<String> getSchemaFiles() {
-    // Only load the schema if semantic search is available
-    if (semanticSearchService != null) {
-      return List.of(SEMANTIC_SEARCH_SCHEMA_FILE);
-    }
-    return Collections.emptyList();
+    // Always load the schema so the field is defined for clients. When the
+    // service is unavailable, the resolver throws SemanticSearchDisabledException
+    // which surfaces as a clear runtime GraphQL error instead of FieldUndefined
+    // — giving clients a stable contract for feature detection.
+    return List.of(SEMANTIC_SEARCH_SCHEMA_FILE);
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


alexsku (approved):
## Review: Approve ✅

Reviewed the full files on the PR branch (not just the diff) and traced the runtime path end-to-end. The change is correct, minimal, and well-targeted at a real problem. Good to release.

### What this does
Before, when `SemanticSearchService` is null (OSS default / `ELASTICSEARCH_SEMANTIC_SEARCH_ENABLED=false`) the plugin loaded neither the schema nor resolvers, so `semanticSearch` / `semanticSearchAcrossEntities` failed **query validation** with `FieldUndefined` — a config-dependent error indistinguishable from a client-side typo. After, the schema is always loaded and the disabled path registers stub `DataFetcher`s that throw `SemanticSearchDisabledException`, so clients get a stable **execution-time** error instead. The happy path is unchanged.

### Verification
- **Exception → client-message path confirmed.** `SemanticSearchDisabledException` is a `RuntimeException` (message: *"Semantic search is disabled in this environment"*). `DataHubDataFetcherExceptionHandler` — wired at `GraphQLEngine.java:115` via `.defaultDataFetcherExceptionHandler(...)` — falls through to its `RuntimeException` branch and surfaces exactly that message (`errorCode=SERVER_ERROR`). So the stub genuinely produces the advertised runtime error.
- **No duplicate-field / dangling-type risk.** These two fields are defined only in `semantic-search.graphql`, which `extend`s `Query` referencing types (`SearchInput`, `SearchAcrossEntitiesInput`, `SearchResults`) that live in the always-loaded core schema. Loading it unconditionally can't create duplicate definitions or unresolved refs.
- **Stub is necessary, not redundant.** Without a registered fetcher, graphql-java's default `PropertyDataFetcher` would return `null` (a silent zero-hit) — the throwing stub is what converts that into an explicit signal.
- **Real consumers benefit, no UI impact.** The GraphQL consumers are the `datahub search` CLI (`semantic_search.gql`) and agent-context (`document_semantic_search.gql`); there are no `datahub-web-react` consumers, so no frontend behavior change.
- Tests assert the new contract behaviorally (capture both stubs, confirm each throws). `codecov/patch` reports modified lines covered, and CI build & test is green.

### Risk: Low
Blast radius is the two semantic-search Query fields. Rollback is a clean revert (no data/schema migration). For a disabled GMS the response shape changes from a validation error to `data: null` + `errors[]`, which is strictly more standard and is what the intended consumers want.

### Non-blocking suggestions (LOW — fine to merge as-is)
1. **Stale log in `init()`** (`SemanticSearchPlugin.java:65-66`): the null path still logs *"Semantic search GraphQL endpoints will not be available."*, but they're now registered as throwing stubs. This slightly contradicts the accurate log you added at line ~102 — worth tweaking for operator clarity.
2. **Feature-detection is message-string-based.** Because the error maps to the generic `SERVER_ERROR` code, clients must match on the message text to detect the disabled state. That's a clear improvement over `FieldUndefined`, but a dedicated error code/extension (e.g. a `DataHubGraphQLException` subtype) would be a more robust contract if you want to harden this later.
3. **End-to-end path isn't auto-tested.** The unit tests confirm the fetcher throws, and I confirmed the handler mapping by reading the code, but no integration/smoke test asserts the actual GraphQL error payload for the disabled path (the two manual-verification boxes in the test plan are unchecked). Optional given the unit coverage.

PR #17592, opened 2026-05-27
feat(ingest/kinesis): AWS Kinesis Data Streams + Amazon Data Firehose source

treff7es:
## Summary

Adds a new ingestion source covering both AWS streaming services under two DataHub platforms that mirror AWS's own service split:

- **`kinesis`** (display name: *Amazon Kinesis Data Streams*) — KDS streams as Datasets (`Stream` subtype) under a regional Container.
- **`kinesis-firehose`** (display name: *Amazon Data Firehose*) — Firehose delivery streams as DataJobs (`Firehose Delivery Stream` subtype) under one regional DataFlow per recipe.

The two platforms share one connector, one IAM policy, and one ingestion run.

## Capabilities

- `DESCRIPTIONS`, `CONTAINERS`, `TAGS`, `OWNERSHIP`
- `LINEAGE_COARSE` — Firehose `dataJobInputOutput` edges from the source KDS stream to the destination platform (S3, Redshift, OpenSearch/Elasticsearch, Snowflake, Apache Iceberg, MongoDB)
- `SCHEMA_METADATA` (opt-in via `glue_schema_registry.enabled`) — Avro / JSON / Protobuf schemas from AWS Glue Schema Registry
- `DELETION_DETECTION` via stateful ingestion
- Account-id auto-resolved via `sts:GetCallerIdentity` for cross-account URN safety
- Cross-platform lineage via `destination_platform_map` with per-destination `platform_instance` / `env` / `convert_urns_to_lowercase` overrides
- Firehose Parquet/ORC format-conversion lineage — surfaces the referenced Glue Catalog table as an additional upstream

## Tests

- 108 unit tests across 13 files (config, identity, region, report, schema registry, tagging, stateful, firehose, firehose destinations, stream, aws_utils, connection)
- 1 LocalStack integration test against a 24KB / 805-event golden file exercising KDS + Firehose + S3 lineage + tags + ownership end-to-end

## Documentation

- Full docs under `metadata-ingestion/docs/sources/kinesis/`
- Both platforms registered in `data-platforms.yaml` with official AWS architecture icons
- Discoverability: dedicated `kinesis-firehose` entry in the integrations catalog and the docs sidebar (so users searching "Firehose" or "Amazon Data Firehose" find the connector)

## Test plan

- [x] Unit suite: ``pytest tests/unit/kinesis/`` — 108 pass
- [x] Integration test: ``pytest tests/integration/kinesis/`` against LocalStack 3.0 — passes with the bundled golden file
- [x] Real-AWS smoke: ingestion against a sandbox account (KDS + Firehose + Glue Catalog table lineage)
- [x] Docs site build (`./gradlew :docs-website:yarnStart`) — page renders cleanly, sidebar shows both platforms

## Checklist

- [x] Tests for the changes have been added/updated (108 unit + 1 integration)
- [x] Docs related to the changes have been added/updated (README + pre + post + recipe + integrations catalog + sidebar entry)
- [ ] For any breaking change/potential downtime/deprecation/big changes an entry has been made in [Updating DataHub](https://github.com/datahub-project/datahub/blob/master/docs/how/updating-datahub.md) — N/A (new connector, no breaking change)

## Not supported (see docs Limitations section)

- `USAGE_STATS` via CloudWatch
- Multi-region ingestion in a single recipe
- Multi-registry GSR support in a single recipe
- Schema inference from record sampling
- Lambda consumer discovery
- Kinesis Data Analytics (KDA / managed Flink)
- Firehose destinations beyond the six listed (HTTP, Datadog, Splunk, New Relic, Coralogix, LogicMonitor, Dynatrace, Honeycomb, Sumo Logic)
- Glue Schema Registry cross-schema references (\`\$ref\` / named imports)


treff7es:
## Screenshots from local DataHub

Ingested against an AWS sandbox account (us-west-2) with 2 KDS streams and 1 Firehose delivery stream, producing the full chain `Stream → Firehose Delivery Stream → S3` with the Firehose `SchemaConfiguration` Glue Catalog table also surfaced as an upstream input.

### Home page — both platforms registered with the right names + icons

home

### Browse `kinesis` platform — *Amazon Kinesis Data Streams*

kinesis platform

### Browse `kinesis-firehose` platform — *Amazon Data Firehose*

firehose platform

### KDS Dataset: `TestStream`

Owner derived from AWS `owner` resource tag (now always a `corpuser` URN), tag `owner:John Doe`, downstream Firehose lineage detected, "Synced from Amazon Kinesis Data Streams" status:

TestStream dataset

### Custom properties from `DescribeStream`

stream ARN, status, retention, encryption, stream mode, shard count:

TestStream properties

### Regional Firehose DataFlow

`Firehose` subtype, `Amazon Data Firehose` platform:

Firehose DataFlow

### Firehose Delivery Stream DataJob: `KDS-S3-SrICX`

`Firehose Delivery Stream` subtype, lineage detected (upstream 1 stream):

DataJob overview

### Lineage graph

Stream → Firehose Delivery Stream → S3 destination, with the Glue Catalog table from the Parquet `SchemaConfiguration` also as an upstream input:

Lineage graph

---

metadata-ingestion/docs/sources/kinesis/kinesis_post.md
```diff
@@ -0,0 +1,191 @@
+### Capabilities
+
+:::note
+Stateful Ingestion (deletion detection) requires a `platform_instance` to be set on this source.
+:::
+
+Use the **Important Capabilities** table above the page as the source of truth for which capabilities are supported and which require additional configuration.
```
sgomezvillamor: if we have a table above as the source of truth, we could remove the one below, no?

metadata-ingestion/docs/sources/kinesis/kinesis_post.md
```diff
@@ -0,0 +1,191 @@
+### Capabilities
+
+:::note
+Stateful Ingestion (deletion detection) requires a `platform_instance` to be set on this source.
```
sgomezvillamor: I don't recall this being a constraint in other connector
how is it is here?

metadata-ingestion/docs/sources/kinesis/kinesis_post.md
```diff
@@ -0,0 +1,191 @@
...
+| Snowflake                         | `snowflake`      | `urn:li:dataset:(urn:li:dataPlatform:snowflake,<db>.<schema>.<table>,...)` |
+| Apache Iceberg                    | `iceberg`        | `urn:li:dataset:(urn:li:dataPlatform:iceberg,<namespace>.<table>,...)`     |
+| MongoDB                           | `mongodb`        | `urn:li:dataset:(urn:li:dataPlatform:mongodb,<database>.<collection>,...)` |
+
+URN names are lowercased by default (matching the Snowflake source's `convert_urns_to_lowercase=True` default). Override per destination via `destination_platform_map.<platform>.convert_urns_to_lowercase: false` — see [Cross-platform lineage](#cross-platform-lineage-with-destination_platform_map) below.
+
+Unsupported Firehose destinations (HTTP, Datadog, Splunk, New Relic, Coralogix, LogicMonitor, Dynatrace, Honeycomb, Sumo Logic, etc.) do **not** produce a lineage edge — the connector logs a warning (_"Unsupported Firehose destination"_) and records the destination configuration as a custom property on the DataJob. The DataJob itself is still emitted.
+
+#### Cross-platform lineage with `destination_platform_map`
```
sgomezvillamor, line 22: do we want to introduce this when we are about to implement the URN case insensitve resolver?

metadata-ingestion/docs/sources/kinesis/kinesis_post.md
```diff
@@ -0,0 +1,191 @@
...
+`stream_pattern` and `delivery_stream_pattern` use the standard DataHub `AllowDenyPattern` shape. A common deny rule excludes internal / audit / debug streams:
+
+```yaml
+stream_pattern:
+  deny:
+    - "^_.*"
+    - ".*-debug$"
+```
+
+#### Tags-as-ownership
```
sgomezvillamor: isn't that what this transform does?
https://docs.datahub.com/docs/metadata-ingestion/docs/transformer/dataset_transformer#extract-ownership-from-tags

metadata-ingestion/docs/sources/kinesis/kinesis_post.md
```diff
@@ -0,0 +1,191 @@
...
+
+The connector reads owner information from a single AWS resource-tag key (default `owner`). Set `owner_tag_key` to whatever convention your organization uses (e.g. `team`, `data_owner`). The tag value is used verbatim as the owner identifier and emitted as a `urn:li:corpuser:<value>` owner — DataHub's identity layer handles group-membership mapping at the platform level if the value names a group rather than an individual user.
+
+#### Glue Schema Registry
+
+GSR is **opt-in** because it requires extra IAM permissions (`glue:Get*` / `glue:List*`). To enable it:
```
sgomezvillamor: it's optional
and we should document what we get/miss if enabled/disabled

metadata-ingestion/docs/sources/kinesis/kinesis_post.md
sgomezvillamor: I mean the per-platform convert_urns_to_lowercase

metadata-ingestion/docs/sources/kinesis/kinesis_pre.md
```diff
@@ -0,0 +1,117 @@
+### Overview
+
+> **Looking for Amazon Data Firehose (formerly Kinesis Data Firehose)?**
+> You're in the right place — Firehose delivery streams are ingested by this same `kinesis` connector. See the `kinesis-firehose` platform section below.
```
sgomezvillamor:
```suggestion
:::info Looking for Amazon Data Firehose (formerly Kinesis Data Firehose)?

You're in the right place — Firehose delivery streams are ingested by this same `kinesis` connector. See the `kinesis-firehose` platform section below.
:::
```

metadata-ingestion/docs/sources/kinesis/kinesis_pre.md
```diff
@@ -0,0 +1,117 @@
...
+
+### Prerequisites
+
+#### Installation
+
+The Kinesis source ships in the base `acryl-datahub` package — `boto3` is already a core dependency, so there is no separate `[kinesis]` extra:
+
+```shell
+pip install 'acryl-datahub'
+```
```
sgomezvillamor: AFAIK the pip install command is automatically generated by the docs gen
so we may have two commands... and maybe not consistent each other

Also, how is is no separate `[kinesis]` extra?
Even if no "additional" dependency, we should have it, so all sources have the same approach.
If we ever need to introduce a dependency for kinesis, that would require to define and so that would be a breaking change for all users installing kinesis just with `acryl-datahub`

metadata-ingestion/docs/sources/kinesis/README.md
```diff
@@ -0,0 +1,30 @@
...
+AWS Kinesis is a real-time streaming and data-delivery service. See the [official AWS Kinesis page](https://aws.amazon.com/kinesis/) for product details.
+
+This connector covers both AWS streaming services with one recipe and one IAM policy:
+
+- **Amazon Kinesis Data Streams (KDS)** — emitted under the `kinesis` platform (display name: _Amazon Kinesis Data Streams_) as **Datasets** (`Stream` subtype).
+- **Amazon Data Firehose** (formerly _Amazon Kinesis Data Firehose / KDF_) — emitted under the `kinesis-firehose` platform (display name: _Amazon Data Firehose_) as **DataJobs** under one regional **DataFlow** per recipe, with cross-platform lineage edges to the destination (S3, Redshift, OpenSearch, Snowflake, Apache Iceberg, MongoDB).
+
+AWS resource tags become DataHub tags and ownership. Glue Schema Registry can be opted in to attach Avro / JSON / Protobuf schemas to streams.
+
+> **Looking specifically for Amazon Data Firehose?** Firehose delivery streams are ingested by this same connector — see the [Concept Mapping](#concept-mapping) table below or the [Limitations](#limitations) section for cross-platform lineage configuration.
```
sgomezvillamor:
```suggestion
:::info Looking specifically for Amazon Data Firehose?

Firehose delivery streams are ingested by this same connector — see the [Concept Mapping](#concept-mapping) table below or the [Limitations](#limitations) section for cross-platform lineage configuration.
:::

```

metadata-ingestion/src/datahub/ingestion/source/kinesis/kinesis.py
```diff
@@ -0,0 +1,326 @@
...
+    @classmethod
+    def create(cls, config_dict: dict, ctx: PipelineContext) -> "KinesisSource":
+        config = KinesisSourceConfig.model_validate(config_dict)
+        return cls(config, ctx)
+
+    def get_workunit_processors(self) -> List:
+        return [
+            *(super().get_workunit_processors() or []),
+            self.stale_entity_removal_handler.workunit_processor,
+        ]
```
sgomezvillamor: This is not the way anymore :-) 
Please, check recent refactor on workunit processors
https://docs.datahub.com/docs/metadata-ingestion/docs/dev_guides/add_stateful_ingestion_to_source#3-modify-the-source

metadata-ingestion/src/datahub/ingestion/source/kinesis/kinesis_aws_utils.py
```diff
@@ -0,0 +1,36 @@
...
+each one repeats the same isinstance discriminator. This module exposes one
+helper, ``aws_error_code``, so those callsites read as a single expression.
+"""
+
+from typing import Union
+
+from botocore.exceptions import BotoCoreError, ClientError
+
+
+def aws_error_code(e: Union[ClientError, BotoCoreError]) -> str:
```
sgomezvillamor: is this kinesis specific or could be moved to some aws shared folder?

metadata-ingestion/tests/integration/kinesis/__init__.py
sgomezvillamor, line 1: How satisfied are you with these Docker-based integration tests? Do you think they're sufficient, or would connector-tests still be needed on top?

metadata-ingestion/tests/integration/kinesis/setup_test_data.py
sgomezvillamor, line 1: this setup test data script seems to be doing a cusom strategy for the setup of the tests


why not following some standard such as the ones below?

```
For LocalStack integration tests in Python, here are the best approaches ranked by practicality:
1. Pytest Fixtures (Recommended)
Use conftest.py with session-scoped fixtures so infra is created once per test run:
python# conftest.py
import boto3
import pytest

LOCALSTACK_URL = "http://localhost:4566"

@pytest.fixture(scope="session")
def aws_credentials():
    """Mocked AWS credentials for LocalStack."""
    import os
    os.environ["AWS_ACCESS_KEY_ID"] = "test"
    os.environ["AWS_SECRET_ACCESS_KEY"] = "test"
    os.environ["AWS_DEFAULT_REGION"] = "us-east-1"

@pytest.fixture(scope="session")
def s3_client(aws_credentials):
    return boto3.client("s3", endpoint_url=LOCALSTACK_URL)

@pytest.fixture(scope="session")
def setup_buckets(s3_client):
    """Create all required buckets once for the whole test session."""
    buckets = ["my-uploads", "my-processed", "my-archive"]
    for bucket in buckets:
        s3_client.create_bucket(Bucket=bucket)
    yield
    # Optional teardown
    for bucket in buckets:
        s3_client.delete_bucket(Bucket=bucket)

# Use in tests:
def test_upload(s3_client, setup_buckets):
    s3_client.put_object(Bucket="my-uploads", Key="file.txt", Body=b"hello")
    ...
2. Docker Compose + Startup Script
Start LocalStack via Docker Compose and run an init script before tests:
yaml# docker-compose.yml
services:
  localstack:
    image: localstack/localstack
    ports:
      - "4566:4566"
    environment:
      - SERVICES=s3,sqs,dynamodb
    volumes:
      - ./localstack-init:/etc/localstack/init/ready.d  # auto-runs on startup
bash# localstack-init/init.sh
#!/bin/bash
awslocal s3 mb s3://my-uploads
awslocal sqs create-queue --queue-name my-queue
awslocal dynamodb create-table \
  --table-name my-table \
  --attribute-definitions AttributeName=id,AttributeType=S \
  --key-schema AttributeName=id,KeyType=HASH \
  --billing-mode PAY_PER_REQUEST
This is great for CI pipelines since the infra is ready the moment the container is healthy.
```

metadata-ingestion/tests/integration/kinesis/test_kinesis.py
```diff
@@ -0,0 +1,119 @@
...
+        ) as resp:
+            body = resp.read().decode("utf-8")
+            return (
+                '"kinesis": "available"' in body
+                or '"kinesis":"available"' in body
+                or '"kinesis": "running"' in body
+                or '"kinesis":"running"' in body
+            )
+    except Exception:
+        return False
```
sgomezvillamor: this could be pushed down to docker compose, as the healthcheck

metadata-ingestion/tests/integration/kinesis/test_kinesis.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+@pytest.fixture(scope="module")
+def kinesis_runner():
```
sgomezvillamor: we have utilities in the codebase for doing the docker compose startup/shutdown

metadata-ingestion/tests/integration/kinesis/test_kinesis.py
```diff
@@ -0,0 +1,119 @@
...
+    # running DataHub. `OUTPUT_FILE` is interpolated by the recipe loader.
+    env = {
+        **os.environ,
+        "AWS_ENDPOINT_URL": "http://localhost:4566",
+        "AWS_DEFAULT_REGION": "us-east-1",
+        "AWS_ACCESS_KEY_ID": "test",
+        "AWS_SECRET_ACCESS_KEY": "test",
+        "OUTPUT_FILE": str(output),
+    }
+    # Prefer DATAHUB_BIN if set (avoids picking up a stale global `datahub`).
```
sgomezvillamor: stale global datahub?

what are we preventing here?

sgomezvillamor:
## Connector Review — AWS Kinesis (KDS + Firehose)

### 🔴 Blockers

#### 1. LiteralString violations in `report.warning()` / `report.failure()` (~13 call sites)

The `message` and `title` parameters of `self.report.warning()` / `self.report.failure()` are typed as `LiteralString` — they must be compile-time constant strings. F-strings with runtime variables violate this constraint, break error aggregation (each unique `{code}` value becomes a new aggregation key), and will fail mypy.

Affected files: `kinesis.py`, `kinesis_stream.py`, `kinesis_firehose.py`, `kinesis_schema_registry.py`.

Example fix:
```python
# ❌ WRONG
self.report.warning(
    title="Could not resolve AWS account ID",
    message=f"sts:GetCallerIdentity returned {code}; URNs will not include account ID.",
)

# ✅ CORRECT — dynamic value goes in context=
self.report.warning(
    title="Could not resolve AWS account ID",
    message="sts:GetCallerIdentity failed; URNs will not include account ID.",
    context=f"error_code={code}",
)
```

Apply this pattern to all ~13 violations (every `message=f"..."` or `title=f"..."` call site across the four files).

#### 2. Low test coverage (36.43% patch coverage)

Codecov reports 36.43% patch coverage (410 lines uncovered). Per-file:
- `kinesis_firehose.py`: 20%
- `kinesis_stream.py`: 29%
- `kinesis.py`: 37%
- `kinesis_firehose_destinations.py`: 32%

Critical gaps:
- No unit tests for `kinesis_stream.py`, `kinesis_schema_registry.py`, or `kinesis.py`
- Most `DestinationHandler` subclasses in `kinesis_firehose_destinations.py` are untested (Redshift, Snowflake, Iceberg multi-URN, OpenSearch, MongoDB)
- `test_connection()` paths not covered
- Only 1 integration test scenario

---

### 🟡 Should Fix

**Remove `# Task N` comments from source code** (`kinesis.py`, `kinesis_firehose.py`). References like `# Task 2 fix-up`, `# Task 10`, `# Task 11` belong in the PR description, not production code — they have no meaning to future readers.

**Ownership inconsistency between KDS and Firehose** — `kinesis_firehose.py` correctly calls `kinesis_tagging.build_ownership_aspect()`, but `kinesis_stream.py`'s `_emit_dataset()` imports `OwnerClass`/`OwnershipTypeClass` and builds the aspect inline. Use `build_ownership_aspect()` in both places.

**Add `[kinesis]` extra to `setup.py`** — even with no new dependencies (boto3 is already a core dep), all connectors should define their own extra for consistency and future-proofing: `"kinesis": []`.

**`convert_urns_to_lowercase` config** — adding a per-destination case-folding config surface now may create conflict or redundancy once the URN case-insensitive resolver ships. Worth aligning with the team on timing before this becomes a permanent API.

**Ownership-from-tags overlap with `ExtractOwnershipFromTags` transformer** — the connector bakes tag-based ownership extraction with a fixed key convention and always emits `corpuser` (never `corpGroup`). Consider either removing it and pointing users to the transformer, or documenting the limitations clearly.

---

### ℹ️ Suggestions

- `DatasetSubTypes.KINESIS_STREAM = "Stream"` is very generic. Consider `"Kinesis Stream"` to match the pattern of `SNOWFLAKE_STREAM = "Snowflake Stream"` — it's clearer in the UI.
- Expand integration tests to cover: empty state (no streams), GSR enabled, and allow/deny pattern filtering.

---

### ✅ What's Done Well

- Correct base classes (`StatefulIngestionSourceBase` + `TestableSource`) with solid `test_connection()` per-service probing
- First-page vs. mid-pagination failure distinction is exactly right for stateful ingestion correctness
- `DestinationHandler` plugin pattern is clean and extensible
- `kinesis_aws_utils.py` correctly unifies `ClientError` + `BotoCoreError` handling
- `KinesisSourceReport` is thorough — good per-entity tracking, GSR counters, config snapshot
- Pydantic v2 used consistently throughout; correct mixin inheritance in config
- `time_machine` used in integration tests (not `freezegun`) ✓


---

metadata-ingestion/docs/sources/kinesis/kinesis_pre.md
aviraj-gour: `setup.py` and `pyproject.toml`, which add a [kinesis]

```
# setup.py:720-722
# kafka_protobuf reused for Glue Schema Registry PROTOBUF support
# (kinesis_schema_registry imports protobuf_util at module load).
"kinesis": aws_common | kafka_protobuf,
```

`pip install 'acryl-datahub'` will hit ModuleNotFoundError: No module named `grpc` the moment the source loads. The doc should say pip install `acryl-datahub[kinesis]`.

metadata-ingestion/docs/sources/kinesis/kinesis_pre.md
```diff
@@ -0,0 +1,117 @@
...
+      ],
+      "Resource": "*"
+    }
+  ]
+}
+```
+
+Notes on each statement:
+
+- **`AccountIdentityRead`** — used at source init to call `sts:GetCallerIdentity` and resolve the AWS account ID. The connector uses the account ID as part of the default `platform_instance` (`<account_id>-<region>`) so URNs disambiguate across accounts and regions. If you omit this statement (or `sts:GetCallerIdentity` is denied), the connector logs a warning and continues with `platform_instance=None`; URNs then won't include the account ID, so cross-account collision-safety depends on you setting `platform_instance` explicitly in the recipe.
```
aviraj-gour: This says the default platform_instance is `<account_id>-<region>`, but the code sets it to account_id alone (`kinesis.py:150`), and the comment above it explicitly keeps region out of platform_instance

metadata-ingestion/src/datahub/ingestion/source/kinesis/kinesis_config.py
```diff
@@ -0,0 +1,251 @@
...
+DestinationPlatform = Literal[
+    "s3", "redshift", "elasticsearch", "snowflake", "iceberg", "mongodb", "glue"
+]
+
+# Closed set of AWS service names that `make_client` knows how to configure
+# (endpoint override + retry config). Narrowing this from `str` catches
+# typos at typecheck time: `make_client(session, "firhose")` fails the
+# Literal check instead of returning a misconfigured boto3 client whose
+# first API call surfaces an unhelpful botocore error. "sts" intentionally
+# omitted — `_resolve_account_id` constructs it directly via
```
aviraj-gour: Minor: `_resolve_account_id` builds STS via `self._session.client("sts")`, so `aws_config.aws_endpoint_url` isn't applied to STS (only the `AWS_ENDPOINT_URL` env var is) — unlike every other client, which goes through `make_client`. It degrades gracefully (warning + `platform_instance=None`), so not a blocker, but it's inconsistent, and against a custom data endpoint with public STS reachable it could resolve a real account id that doesn't match the data source. The comment in `kinesis_config.py:36-38` ("runs before the rest of the config is loaded") is also inaccurate — this runs in `__init__` with full config. Consider adding `"sts"` to the `AwsService` Literal and routing through `make_client`.

metadata-ingestion/src/datahub/ingestion/source/kinesis/kinesis.py
```diff
@@ -0,0 +1,316 @@
...
+            yield from self.firehose_extractor.get_workunits()
+
+    def _emit_region_container(self) -> Iterable[MetadataWorkUnit]:
+        region = self.config.aws_config.aws_region
+        assert region is not None  # validated in __init__
+        region_key = self._region_key()
+        yield from gen_containers(
+            container_key=region_key,
+            name=region,
+            sub_types=["Region"],
```
aviraj-gour: `Minor`: Should we add this to `common.subtypes.py`.

metadata-ingestion/src/datahub/ingestion/source/kinesis/kinesis.py
```diff
@@ -0,0 +1,316 @@
...
+                    "AWS account_id. Set platform_instance in the recipe to disambiguate "
+                    "across accounts, or grant sts:GetCallerIdentity to the principal."
+                ),
+                exc=e,
+            )
+            return None
+
+    def _region_key(self) -> KinesisRegionKey:
+        return KinesisRegionKey(
+            region=self.config.aws_config.aws_region or "unknown",
```
aviraj-gour: `or "unknown"` here is unreachable — `__init__` (kinesis.py:133-138) raises `ValueError` if the region can't be resolved, so `aws_region` is always set by the time `_region_key` runs


```suggestion
            region=self.config.aws_config.aws_region",
```

metadata-ingestion/src/datahub/ingestion/source/kinesis/kinesis_config.py
```diff
@@ -0,0 +1,251 @@
...
+            "to `False` when your destination's source recipe set "
+            "`convert_urns_to_lowercase=False` and your identifiers are "
+            "case-sensitive (Iceberg, MongoDB, or quoted-identifier "
+            "Snowflake / Redshift)."
+        ),
+    )
+
+    @pydantic.field_validator("env")
+    @classmethod
+    def validate_env(cls, v: Optional[str]) -> Optional[str]:
```
aviraj-gour, line 89: **`DestinationPlatformDetail.validate_env` diverges from standard env handling.** It rebuilds the env set via `vars(FabricTypeClass)` reflection with no `.upper()` normalization, so `env: prod` is accepted at the top level (`EnvConfigMixin` uppercases it) but raises `ValueError` inside a `destination_platform_map` entry. 
`Suggestion` reuse `ALL_ENV_TYPES` from `datahub.emitter.mce_builder` and `.upper()`-normalizing.

metadata-ingestion/src/datahub/ingestion/source/kinesis/kinesis_config.py
```diff
@@ -0,0 +1,251 @@
...
+    glue_schema_registry: KinesisGlueSchemaRegistryConfig = Field(
+        default_factory=KinesisGlueSchemaRegistryConfig,
+        description="Glue Schema Registry integration (disabled by default).",
+    )
+
+    destination_platform_map: Dict[DestinationPlatform, DestinationPlatformDetail] = (
+        Field(
+            default_factory=dict,
+            description=(
+                "Map destination platform name (one of 's3', 'redshift', 'snowflake', "
```
aviraj-gour, line 226: The `DestinationPlatform` Literal includes `"glue"`, and the docstring doesn't.

metadata-ingestion/src/datahub/ingestion/source/kinesis/kinesis_firehose.py
```diff
@@ -0,0 +1,470 @@
...
+
+if TYPE_CHECKING:
+    from boto3.session import Session
+    from mypy_boto3_firehose import FirehoseClient
+    from mypy_boto3_firehose.type_defs import DeliveryStreamDescriptionTypeDef
+
+logger = logging.getLogger(__name__)
+
+# KDS streams (the *upstream* in Firehose lineage) live under platform="kinesis".
+PLATFORM_NAME = "kinesis"
```
aviraj-gour: **Nitpick:** This is also defined in `kinesis.py` — consider using the same reference to avoid duplication. (Not necessary)

metadata-ingestion/src/datahub/ingestion/source/kinesis/kinesis_schema_registry.py
```diff
@@ -0,0 +1,185 @@
...
+            if data_format == "JSON":
+                return list(
+                    JsonSchemaTranslator.get_fields_from_schema(
+                        json.loads(schema_def),
+                        is_key_schema=False,
+                    )
+                )
+            if data_format == "PROTOBUF":
+                return protobuf_util.protobuf_schema_to_mce_fields(
+                    ProtobufSchema(f"{stream_name}.proto", schema_def),
```
aviraj-gour: Stream names can contain dots (AWS allows `[a-zA-Z0-9_.-]`), and `grpc.protos()` derives a module
name from this filename — dots break it, so `orders.v2` silently fails to parse (caught by the
broad `except` and reported as parse-failed). The Confluent SR reference sanitizes with
`topic.replace(".", "_")` before building the `ProtobufSchema` name; suggest mirroring that here.

metadata-ingestion/src/datahub/ingestion/source/kinesis/kinesis_config.py
```diff
@@ -0,0 +1,251 @@
...
+        ),
+    )
+    env: Optional[str] = Field(
+        default=None,
+        description=(
+            "env (e.g., PROD, DEV) of the destination. If unset, inherits from "
+            "this source's env."
+        ),
+    )
+    convert_urns_to_lowercase: bool = Field(
```
aviraj-gour, line 73: Should we keep default as `False`?

metadata-ingestion/src/datahub/ingestion/source/kinesis/kinesis_schema_registry.py
treff7es: thanks, I resolved it

metadata-ingestion/src/datahub/ingestion/source/kinesis/kinesis_firehose.py
treff7es: fixed

metadata-ingestion/src/datahub/ingestion/source/kinesis/kinesis_config.py
treff7es: fixed

metadata-ingestion/src/datahub/ingestion/source/kinesis/kinesis_config.py
treff7es: True is the default in other platforms to make sure urns going to match

metadata-ingestion/tests/integration/kinesis/__init__.py
treff7es: Ideally, localstack works well and requires way less setup and less fragile

metadata-ingestion/tests/integration/kinesis/setup_test_data.py
treff7es: thanks, I updated

sgomezvillamor:
**Reminder:** F-strings remain in `message=` (and some `title=`) parameters across 10+ call sites — including the new `report_listing_failure` helper in `kinesis_report.py` which introduced new `title=f"..."` violations. `message` and `title` must be string literals (they're used as aggregation keys); dynamic values like `{code}`, `{data_format}`, `{dest_type}` belong in `context=`.

---

sgomezvillamor (approved):
LGTM — all major review items addressed. See inline reminder about the remaining `LiteralString` violations in `message=`/`title=`.

---

metadata-ingestion/src/datahub/ingestion/source/kinesis/kinesis_tagging.py
```diff
@@ -0,0 +1,76 @@
...
+) -> Optional[GlobalTagsClass]:
+    """Flatten AWS resource tags to a DataHub GlobalTags aspect.
+
+    Each `{Key, Value}` becomes a `urn:li:tag:{Key}:{Value}`. Returns None when the
+    tag list is empty so callers can decide whether to emit the aspect at all.
+    """
+    if not tags:
+        return None
+    associations = [
+        TagAssociationClass(tag=make_tag_urn(f"{t['Key']}:{t['Value']}"))
```
aviraj-gour: AWS allows tags with empty Value (key-only tags).  
 The filter if t.get("Key") and t.get("Value") silently drops them. A user tagging a stream with           `Key=pii, Value=""` would not see it in DataHub with warning.                                                                    
  Fix: Either include tags with empty values, or log when tags are dropped.

aviraj-gour (approved):

metadata-ingestion/src/datahub/ingestion/source/kinesis/kinesis_tagging.py
treff7es: Good catch — fixed in c15e899a20b. `build_global_tags_from_aws_tags` now requires only `Key` and emits a key-only `urn:li:tag:<Key>` when the value is empty (instead of dropping it), so a `Key=pii, Value=""` tag still surfaces in DataHub. Added unit tests for the empty-value and key-less cases.

treff7es:
Re the LiteralString reminder — done in c15e899a20b. Made `title=`/`message=` static literals and moved the interpolated detail (error codes, stream / delivery-stream names, page counts) into `context=` across all 13 `report.warning`/`report.failure` call sites, including the `report_listing_failure` helper in `kinesis_report.py`. This keeps the structured-logger dedup key (`title`-`message`) stable so identical issues group under one entry with their distinct contexts. No change to emitted MCPs — golden file is byte-identical and the integration test still passes.

PR #17594, opened 2026-05-27
fix(spark-lineage): Add flag to suppress noisy INDIRECT column upstreams

aviraj-gour:
## Summary

- In `acryl-spark-lineage`, every output column showed the same wide set of upstream columns (JOIN keys, WHERE filters, GROUP BY columns) in DataHub's column lineage panel, making the panel unreadable.
- Adds a Spark config flag `spark.datahub.column_lineage.include_indirect` (default `true`, preserves existing behaviour). When set to `false`, input columns whose only role is INDIRECT (JOIN/FILTER/GROUP BY) are dropped from column-level lineage.
- The default is unchanged so existing users see no behaviour change on upgrade. Opt-in to the cleaner lineage by setting the flag to `false`.

## Behaviour comparison

Query from the ticket:

```sql
SELECT c.customer_id, c.email, c.country,
       COUNT(o.order_id) AS total_orders,
       SUM(o.total_amount) AS lifetime_value
FROM customers c
JOIN orders o ON c.customer_id = o.customer_id
WHERE o.status = 'completed'
GROUP BY c.customer_id, c.email, c.country
```

| Output column    | Default (`include_indirect=true`)  | Opt-in (`include_indirect=false`)  |
| ---------------- | ---------------------------------- | ---------------------------------- |
| `customer_id`    | 5 upstreams (incl. JOIN/WHERE/GROUP BY) | 1 upstream — `customers.customer_id` |
| `email`          | 5 upstreams                        | 1 upstream — `customers.email`     |
| `country`        | 5 upstreams                        | 1 upstream — `customers.country`   |
| `total_orders`   | 6 upstreams                        | 1 upstream — `orders.order_id`     |
| `lifetime_value` | 6 upstreams                        | 1 upstream — `orders.total_amount` |

Table-level lineage is unchanged — source datasets still appear upstream of the output dataset.

## Implementation

The change is intentionally minimal — one skip check inside the existing input-field loop in `OpenLineageToDataHub.getFineGrainedLineage`:

```java
if (!includeIndirect && isIndirectOnly(inputField)) {
    return;  // skip this input field entirely
}
```

Plus a small `isIndirectOnly` helper that returns true when every transformation on the input field has `type == "INDIRECT"`. Mixed DIRECT+INDIRECT fields are kept as-is. No bucketing, no multi-entry-per-column splitting, no transformation-text filtering — single-pass union emission, same shape as before.

## What changed

| File                                                                  | Change                                                                                                  |
| --------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------- |
| `openlineage-converter/.../OpenLineageToDataHub.java`                 | Added one `if (!includeIndirect && isIndirectOnly(inputField))` skip + `isIndirectOnly` helper          |
| `openlineage-converter/.../DatahubOpenlineageConfig.java`             | Added `includeIndirectColumnLineage` field (default `true`)                                             |
| `acryl-spark-lineage/.../SparkConfigParser.java`                      | Wired `spark.datahub.column_lineage.include_indirect` (default `true`)                                  |
| `acryl-spark-lineage/.../OpenLineageEventToDatahubTest.java`          | Added test for the new flag dropping INDIRECT-only contributors                                         |
| `acryl-spark-lineage/README.md`                                       | Documented the new flag                                                                                 |

## Test plan

- [x] `./gradlew :metadata-integration:java:openlineage-converter:test` — 20 passing
- [x] `./gradlew :metadata-integration:java:acryl-spark-lineage:test --tests "datahub.spark.OpenLineageEventToDatahubTest"` — 33 passing
- [x] `spotlessApply` + `mdPrettierWrite` clean
- [x] Manual end-to-end smoke test: ran a local PySpark job (JOIN + WHERE + GROUP BY) with the new jar emitting MCPs to a file. Confirmed default behaviour is byte-for-byte identical to the pre-fix jar. Confirmed `--conf spark.datahub.column_lineage.include_indirect=false` drops the JOIN/WHERE/GROUP BY columns from column-level lineage while keeping table-level lineage intact.

## Backwards compatibility

**No default behaviour change.** Existing Spark users upgrading the jar will see the same column lineage they had before. Users who want the cleaner lineage panel opt in via `--conf spark.datahub.column_lineage.include_indirect=false`.


metadata-integration/java/acryl-spark-lineage/src/main/java/datahub/spark/conf/SparkConfigParser.java
```diff
@@ -85,6 +85,7 @@ public class SparkConfigParser {
       "metadata.dataset.enableEnhancedMergeIntoExtraction";
 
   public static final String CAPTURE_COLUMN_LEVEL_LINEAGE = "captureColumnLevelLineage";
+  public static final String COLUMN_LINEAGE_INCLUDE_INDIRECT = "column_lineage.include_indirect";
```
treff7es: Camelcase is the preferred

treff7es (approved):
Please, change the config value but overall it looks good

PR #17595, opened 2026-05-27
fix(ingest/s3): handle file names with dots in the stem when resolving format extension

### LINKED ISSUE #19690: S3: "unsupported extension" warning for file names containing dots when default_extension is set
### Describe the bug

`S3Source.get_fields` uses `pathlib.Path(...).suffix`, which is purely lexical and returns everything after the last dot. For S3 files whose names contain dots — e.g. `events.account.update-2026-05-27-<hash>.gz` — stripping the `.gz` compression suffix leaves `.update-2026-05-27-<hash>` as the apparent format extension. That is not a real format, so the inferrer emits an `"unsupported extension"` warning.

Because the leftover value is non-empty, the `default_extension` fallback on the path_spec is never reached — so even with `default_extension: json` set, every such file warns and is not resolved.

### To Reproduce

1. Configure an S3 path_spec with `default_extension: json`.
2. Point it at files whose names contain dots before the compression suffix, e.g. `s3://.../events.account.update-2026-05-27-11-<hash>.gz`.
3. Observe an `"unsupported extension"` warning for every such file; `default_extension` is not applied.

### Expected behavior

When a file name has no recognised format extension (`csv`, `tsv`, `json`, `parquet`, `avro`), the configured `default_extension` should be used so the file is ingested, instead of warning. Files with a recognised extension (including compressed `*.json.gz`, `*.parquet.gz`) should behave as before.

### Version

DataHub ingestion (S3 source), current master.

rospe:
Fixes #19690

## Summary

`pathlib.Path(...).suffix` is purely lexical — it returns everything after the last dot. For S3 files like `events.account.update-2026-05-27-<hash>.gz` this means stripping the `.gz` compression suffix leaves `.update-2026-05-27-<hash>` as the apparent format extension, which is not a real file format. The inferrer then receives a meaningless extension and emits an `"unsupported extension"` warning even when `default_extension` is set on the path_spec.

This PR makes `S3Source.get_fields` only keep an extension after compression stripping (and for uncompressed files) if it matches one of the supported file types (`csv`, `tsv`, `json`, `parquet`, `avro`). Otherwise it falls through to `default_extension` so the inferrer can still be selected correctly.

## Why

We hit this on a real bucket where event names contain dots in their file names, e.g.

`s3://.../event=events.account.update/legalEntity=.../events.account.update-2026-05-27-11-<hash>.gz
`

Even with `default_extension: json` set on the path_spec, every one of these files emitted an `"unsupported extension"` warning because `pathlib.suffix` returned `.update-2026-05-27-11-<hash>` after the `.gz` was stripped — a non-empty value, so the `default_extension` fallback was never reached.

### Preserved

| Input | Result |
|---|---|
| `data.json` | `.json` |
| `data.json.gz` | `.json` (compression stripped, inner extension kept) |
| `data.parquet` | `.parquet` |
| `data.parquet.gz` | `.parquet` |
| `data.gz` (default=json) | `.json` |
| `data.gz` (no default) | `""` → warn (unchanged) |
| `data` (no extension, default=json) | `.json` |
| `data` (no extension, no default) | `""` → warn (unchanged) |

### New

| Input | Old | New |
|---|---|---|
| `foo.bar.baz-<hash>.gz` (default=json) | `.baz-<hash>` → warn | `.json` ✓ |
| `foo.bar.baz-<hash>` (default=json) | `.baz-<hash>` → warn | `.json` ✓ |
| `data.txt` (default=json) | `.txt` → warn | `.json` ✓ |
| `data.JSON` (uppercase, default=json) | `.JSON` → warn | `.json` ✓ |

The behaviour change is one-directional and gated by `default_extension`: when a user sets it, it now applies to **any** file whose name does not have a recognised format extension, not just files with no dot at all. Without `default_extension`, behaviour is unchanged. There is no panic or data-loss risk — the worst case is a `"could not infer schema"` warning instead of the previous `"unsupported extension"` warning when `default_extension` is set and a stray file's actual content does not match the chosen format.

## Changes

- New `_resolve_format_extension` helper in `s3/source.py` extracted from `S3Source.get_fields` so the resolution logic can be unit-tested directly.
- `S3Source.get_fields` now delegates to the helper.
- `PathSpec.default_extension` field description updated to reflect that it applies to any file whose format cannot be inferred from the file name.
- New "File type detection" section in `metadata-ingestion/docs/sources/s3/s3_post.md` documenting how extensions are resolved and when `default_extension` kicks in.
- Parametrized unit tests in `tests/unit/s3/test_s3_source.py` covering preserved behaviour (plain `.json`, `.json.gz`, `.parquet.gz`, `.gz`-only with/without default, compression disabled) and new behaviour (dotted-stem compressed/uncompressed, with/without default).

## Checklist

- [x] PR title follows the conventional commit format
- [x] Tests added (10 parametrized cases on the new helper)
- [x] Docs updated (S3 connector docs + `PathSpec.default_extension` description)
- [x] No breaking change — see behaviour table above

## Notes for reviewers

- The pre-existing `test_get_folder_info_returns_expected_folder` failure on master (1-hour timezone offset between `tzutc()` and `datetime.timezone.utc`) is unrelated to this change.

### DIFF
--- a/metadata-ingestion/docs/sources/s3/s3_post.md
+++ b/metadata-ingestion/docs/sources/s3/s3_post.md
@@ -254,6 +254,20 @@ Schemas for schemaless formats (CSV, TSV, JSONL, JSON) are inferred. For CSV, TS
 JSON file schemas are inferred on the basis of the entire file (given the difficulty in extracting only the first few objects of the file), which may impact performance.
 We are working on using iterator-based JSON parsers to avoid reading in the entire JSON object.
 
+##### File type detection
+
+The format of a file is detected from its name. The connector first checks the apparent extension (everything after the last dot) and accepts it only if it matches one of the supported file types listed above. For files compressed with `.gz` or `.bz2`, the compression suffix is stripped and the inner extension is checked the same way (so `data.json.gz` is treated as JSON).
+
+This means file names whose stem contains dots — for example `events.account.update-2026-05-27-<hash>.gz` — are **not** misinterpreted as having an extension of `.update-2026-05-27-<hash>`. Such files fall back to `path_spec.default_extension` if it is set, and are skipped otherwise.
+
+If your files have no real format extension (or have only a compression extension like `.gz`), set `default_extension` on the path_spec to tell the connector how to parse them:
+
+```yaml
+path_specs:
+  - include: s3://my-bucket/{table}/
+    default_extension: json
+```
+
 #### S3-specific mapping details
 
 This ingestion source maps the following source system concepts to DataHub concepts:
--- a/metadata-ingestion/src/datahub/ingestion/source/data_lake_common/path_spec.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/data_lake_common/path_spec.py
@@ -99,7 +99,7 @@ class PathSpec(ConfigModel):
 
     default_extension: Optional[str] = Field(
         None,
-        description="For files without extension it will assume the specified file type. If it is not set the files without extensions will be skipped.",
+        description="Assume this file type for files where the format cannot be inferred from the file name. This includes files with no extension and files whose apparent extension (the part after the last dot) is not a recognised format (which can happen for file names whose stem contains dots, e.g. ``foo.bar.baz-<hash>.gz``). If unset, such files are skipped.",
     )
 
     table_name: Optional[str] = Field(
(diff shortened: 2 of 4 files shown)


rospe:
Hi @maggiehays , can you please check if there's someone in your team that can take a look at this PR?

PR #17599, opened 2026-05-27
fix(ui): darken UserProfileInfoCard GradientContainer background

liannab:
The UserProfileInfoCard background color is not dark enough behind the white text in order to meet accessibility contrast ratio 4:5:1

Screenshot 2026-05-27 at 10 44 50 AM

This PR just adjusts that color to be darker.

### DIFF
--- a/datahub-web-react/src/app/entityV2/user/UserProfileInfoCard.tsx
+++ b/datahub-web-react/src/app/entityV2/user/UserProfileInfoCard.tsx
@@ -1,4 +1,4 @@
-import { Avatar } from '@components';
+import { Avatar, colors } from '@components';
 import { Col } from 'antd';
 import React, { useState } from 'react';
 import styled from 'styled-components';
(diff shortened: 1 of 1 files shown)


datahub-web-react/src/app/entityV2/user/UserProfileInfoCard.tsx
```diff
@@ -65,7 +65,7 @@ export const UserProfileInfoCard = ({ sidebarData, refetch, dataHubRoleName, isP
     return (
         <>
             <CustomAvatarContainer>
-                <GradientContainer />
+                <GradientContainer gradient={colors.primary[900]} />
```
chriscollins3456, line 68: so this will cause lint to fail since we're avoiding using `colors` directly now, instead we want to use semantic tokens or our themes more properly.

for this what you can do is add `const theme = useTheme();` and then:
```
<GradientContainer gradient={theme.styles['primary-color']} />
```
will give you this:
Image

or you can do:
```
<GradientContainer gradient={theme.colors.textActive} />
```
to get this:
Image

feel free to play with the tokens or the styles! as long as it's a dark primary we're good to go

chriscollins3456 (commented):
thank you for this! a few suggestions on how you could fix the linting issue we're seeing

PR #17606, opened 2026-05-27
fix(ingestion/PowerBi): Improves comment and CTE handling for PowerBI parser

pdruley:
## Summary

- **Bug**: PowerBI `Odbc.Query` ingestion produced wrong upstream URNs when the custom SQL was a single CTE query containing blank lines before `SELECT` keywords and a semicolon inside a SQL comment. A CTE alias (e.g. `my_cte`) was emitted as a real upstream table instead of the actual source tables referenced inside the CTE bodies.

- **Root cause**: Two interacting bugs in `parse_custom_sql` (`native_sql_parser.py`):
  1. The old normalization regex (`re.sub(r"\n{2,}(\s*SELECT\b)", ...)`) blindly inserted `;` before any `SELECT` preceded by blank lines — including `SELECT`s inside CTE bodies and the closing `SELECT` of a `WITH` clause. This fragmented single-statement CTE queries into multiple pieces.
  2. The check `if ";" in normalized` fired on semicolons inside SQL comments (e.g. `-- done; continue`), sending single-statement queries down the multi-statement parsing path. With the query fragmented, the final fragment had no CTE definitions, so the outer CTE alias resolved as a real table against the default database.

- **Fix**: Replaced both mechanisms with two new helper functions:
  - `_has_real_semicolons(query)` — strips `--` line comments and `/* */` block comments before checking for `;`, so comment content never triggers multi-statement parsing.
  - `_insert_statement_separators(query)` — replaces the naive regex with a line-by-line parser that tracks parenthesis depth and an `in_cte_query` flag. Separators are only inserted before `SELECT` at depth 0 when no `WITH` clause is open, ensuring CTE bodies and the closing `SELECT` of a `WITH` query are never split.

- **Tests**: Added `tests/unit/powerbi/test_native_sql_parser.py` covering `_has_real_semicolons` (comment vs. real semicolons), `_insert_statement_separators` (CTE body, CTE final SELECT, standalone SELECTs, nested CTE with comment semicolon), and an end-to-end regression test verifying CTE aliases do not leak into upstream URNs.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/native_sql_parser.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/native_sql_parser.py
@@ -288,7 +288,7 @@ def parse_custom_sql(
         # separators only at depth-0 positions that are not the closing SELECT
         # of a CTE query, then decide which parser to use.
         normalized = _insert_statement_separators(query)
-        if ";" in normalized:
+        if _has_real_semicolons(normalized):
             result = create_lineage_from_sql_statements(
                 queries=normalized,
                 default_schema=schema,
--- a/metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/native_sql_parser.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/native_sql_parser.py
@@ -182,11 +182,17 @@ def _insert_statement_separators(query: str) -> str:
         if blank_count >= 1 and paren_depth == 0:
             if re.match(r"SELECT\b", stripped, re.IGNORECASE):
                 if in_cte_query:
-                    # This SELECT closes the open WITH clause — it is part of
-                    # the same statement, so do NOT insert a separator.
+                    # This SELECT is the final query of the WITH clause.
+                    # Inserting a separator here would detach it from the CTE
+                    # definitions, causing sqlglot to treat CTE aliases as real
+                    # tables and generate spurious upstream URNs.
                     in_cte_query = False
                 else:
-                    result.append(";")
+                    # Append to the last non-blank line for conventional SQL style.
+                    for i in range(len(result) - 1, -1, -1):
+                        if result[i].strip():
+                            result[i] += ";"
+                            break
         elif paren_depth == 0 and in_cte_query:
             # CTE closing SELECT with no preceding blank line — the blank_count
             # gate above never fires, so reset the flag here.  Without this,
--- a/metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/native_sql_parser.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/native_sql_parser.py
@@ -132,6 +132,92 @@ def remove_drop_statement(query: str) -> str:
     return remove_tsql_control_statements(query)
 
 
+def _has_real_semicolons(query: str) -> bool:
+    """Return True if the query contains semicolons outside of SQL comments.
+
+    Prevents comment content (e.g. '-- already done; continue') from
+    incorrectly triggering the multi-statement parsing path.
+    """
+    no_line_comments = re.sub(r"--[^\n]*", "", query)
+    no_comments = re.sub(r"/\*.*?\*/", "", no_line_comments, flags=re.DOTALL)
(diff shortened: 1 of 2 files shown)


aviraj-gour:
CTE-alias-emitted-as-upstream behaviour is a genuine bug and the regression-test direction is good. Before we land it though, I'd like to discuss the approach, because I think the fix addresses a narrow symptom and the new heuristic introduces its own edge cases.

## Core concern: dialect-blind preprocessing is doing sqlglot's job

This parser is shared across **every** PowerBI source platform — `Sql.Database` (mssql), `Snowflake.Databases`, `GoogleBigQuery`, `PostgreSQL`, `Oracle`, `MySQL`, `Redshift`, `Databricks`, and `Odbc.Query` (which resolves to any of them via the DSN). So the embedded SQL is **not** T-SQL in general — its dialect is whatever the source is, and that dialect is already known (`platform_pair.datahub_data_platform_name`) and handed to sqlglot in `create_lineage_*`.

But the preprocessing layer — `remove_tsql_control_statements` + the new `_has_real_semicolons` / `_insert_statement_separators` — is dialect-blind and string/line-based. It's effectively doing **statement-boundary detection by regex**, which is the job sqlglot already does (with the correct dialect) one stage later. The original bug only exists because `remove_tsql_control_statements` strips `GO`/`USE`/etc. and leaves bare blank lines behind; the new code then tries to reconstruct the boundaries from those blank lines. We're cleaning up ambiguity that an earlier step manufactured.

## Concrete cases where the new heuristic is wrong

I ran these against the functions in this PR. It both **over-splits** (breaks single statements) and **under-splits** (merges separate ones):

### Over-split — set operators (very common)

```sql
SELECT a FROM t1
UNION ALL

SELECT b FROM t2
```

→ inserts `;` after `UNION ALL`, splitting one statement into two invalid ones. Same failure class as the original bug. (`EXCEPT` / `INTERSECT` too.)

### Under-split — comment between statements

```sql
SELECT a FROM t1

-- next query
SELECT b FROM t2
```

→ 0 separators inserted (the comment line resets the blank-line counter), so two real statements get merged.

### Under-split — statement followed by a CTE

```sql
SELECT a FROM t1

WITH c AS (SELECT b FROM t2)
SELECT * FROM c
```

→ no separator before `WITH` (insertion only fires before `SELECT`), so they merge.

### Original bug still reproduces with a leading comment

```sql
/* report */ WITH c AS (
    SELECT x FROM realtab
)

SELECT * FROM c
```

→ CTE detection is `re.match(r"WITH\b", ...)` (requires `WITH` first on the line), so `in_cte_query` is never set, the closing `SELECT` looks like a new statement, the CTE is split, `realtab` is lost and `c` is emitted as a fake upstream — i.e. exactly the bug this PR targets.

### `_has_real_semicolons` false positive

```sql
SELECT 'a; b -- c' FROM t
```

→ reports a real semicolon (the comment-strip runs before the string-strip and eats the closing quote), routing a single statement to the multi-statement path.

## Why I'd avoid the line-heuristic approach

Each of these is a new special case to bolt on (`UNION`, comments, `WITH`-after-`SELECT`, leading comments, string handling…), and there will be more (`MERGE`, `INSERT…SELECT`, derived tables, dialect quoting). The root issue is that the layer reasons about SQL structure from whitespace and line prefixes rather than from the grammar.

## Suggested direction

1. **Preserve boundaries at the point of removal** instead of reconstructing them: have `remove_tsql_control_statements` emit `;` where it strips `GO` / `DROP` / `USE` / `SET`, so genuine multi-statement queries arrive already terminated and splitting becomes "split on real semicolons." A single CTE then parses correctly as one statement and the alias bug can't occur, because nothing splits on blank lines.
2. **Make the T-SQL stripping dialect-aware** — it currently applies T-SQL-only assumptions (`GO`, `#temp`, `;WITH`, `SET…ON/OFF`) to Snowflake / BigQuery / Postgres queries unconditionally.

This keeps statement-splitting where the dialect is actually known and lets sqlglot do the parsing it's built for.

bmiller-dh:
@aviraj Thanks for the thorough review. 

Tracking the architectural rework in [ING-2798]https://linear.app/acryl-data/issue/ING-2798

aviraj-gour:
FYI @bmiller-dh,I have raised a seprate PR to fix this.

PR #17613, opened 2026-05-28
feat(s3/abs/gcs): add infer_schema config flag to skip schema inference

rospe:
Add an `infer_schema` boolean field (default: `True`) to the S3, ABS, and GCS source configs. When set to `False`, the source skips schema inference entirely and does not emit a `schemaMetadata` aspect for any dataset.

This is useful when an authoritative schema pipeline (e.g. a JSON Schema registry ingestion) already owns the schema aspect and the file-based source should not overwrite it with data-driven inference. All other aspects (datasetProperties, partitions, containers, lineage, tags) are still emitted normally.

- S3: flag added to `DataLakeSourceConfig`, guarded in `S3Source.ingest_table`
- ABS: same change in ABS's own `DataLakeSourceConfig` and `ABSSource.ingest_table`
- GCS: flag added to `GCSSourceConfig` and forwarded to the underlying `DataLakeSourceConfig` in both HMAC and WIF auth paths

Docs are auto-generated from the pydantic Field description annotation.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/abs/config.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/abs/config.py
@@ -97,6 +97,17 @@ class DataLakeSourceConfig(
         description="Number of files to list to sample for schema inference. This will be ignored if sample_files is set to False in the pathspec.",
     )
 
+    infer_schema: bool = Field(
+        default=True,
+        description=(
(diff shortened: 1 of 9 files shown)


rospe:
Hi @maggiehays , can you please check if there's someone in your team that can take a look at this PR? Thank you!

PR #17615, opened 2026-05-28
fix(ingest): skip empty columns for CLL

kevinkarchacryl:
The sqlglot SQL parser upgrade from v29 to v30 (shipped in 1.5.0.19) changed how certain SQL patterns are analyzed during CLL extraction, causing some upstream column names to be empty strings. These empty strings were being passed directly to make_schema_field_urn(), producing invalid URNs of the form urn:li:schemaField:(urn:li:dataset:(...),) that GMS rejects, causing the entire ingestion to fail. The fix adds guards in two places in dbt_common.py.


kyungsoo-datahub (approved):

bmaquet:
Thank you @kevinkarchacryl for this fix!

PR #11230, opened 2024-08-23
fix(ingestion/tableau): restructure the tableau graphql datasource query

sid-acryl:
Fixes 
- Restructure upstream GraphQl queries to avoid node limit error 
- Rewrite `emit_project_container` to emit project in topological order

PR #17617, opened 2026-05-28
fix(ingest/tableau): apply lineage_overrides to URN construction

acrylJonny:
## Summary

Fixes Tableau lineage when `lineage_overrides` remaps a three-tier source platform (e.g. `presto`) to a two-tier target (e.g. `athena`), adds a new config to route per-environment when connections share a hostname, and surfaces undiscoverable database connections in the ingestion report.

### Bugs fixed

Both `TableauUpstreamReference.make_dataset_urn` (`tableau_common.py`) and `_create_database_table_urn` (`tableau.py`) built fully-qualified names against `original_platform` instead of the overridden `platform`. The resulting URN advertised the target platform (`athena`) but kept the source platform's URN shape (`catalog.schema.table` with database segment), so it never matched the URNs the Athena ingestion source actually emits (`schema.table`). Lineage broke silently — URNs were valid but resolved to nothing.

The `database_override_map` two-tier check had the same problem: it consulted `original_platform` to decide whether to strip the upstream database, so a presto→athena remap kept the database segment in place.

Both paths now use the overridden `platform`, matching the target platform's URN format.

### New feature: `database_id_to_platform_instance_map`

Adds an id-keyed map alongside the existing hostname-keyed map. Tableau database `id` values are unique per connection even when several share a hostname — the common case for AWS Athena workgroups behind a regional endpoint, or multiple Snowflake accounts behind a proxy. With this map, a single Tableau recipe can route distinct connections to distinct `platform_instance` values; id matches take precedence over hostname matches.

The feature is platform-agnostic — Athena/presto is the motivating case but the same routing works for any connector.

### Discoverability: log every server, warn on unrouted ones

When either routing map is configured, the connector now:

- Logs `id / name / hostName / connectionType` at INFO for every Tableau database server it discovers, so operators can locate ids for `database_id_to_platform_instance_map` without needing direct Tableau Metadata API access.
- Emits `report.warning` for any server that matches neither map, surfacing partially-configured routing in the ingestion report instead of silently dropping the connection off the routing path.

### Tests

- Existing `test_lineage_overrides` updated to assert the correct two-tier shape (`platform_instance.schema.table`).
- New `test_lineage_overrides_presto_to_athena` covers the presto→athena URN override end-to-end.
- New `test_database_id_to_platform_instance_map_routes_per_id` covers single-recipe multi-environment routing.
- New `test_database_id_to_platform_instance_map_is_platform_agnostic` pins that the feature works for Snowflake (no `lineage_overrides` involved).
- New `test_database_id_to_platform_instance_map_falls_back_on_miss` pins fall-through behavior when an id has no map entry.
- New `test_database_id_to_platform_instance_map_wins_over_hostname` pins the precedence rule.
- New `test_overridden_info_drives_target_platform_url_shape` guards both URN-building sites by pinning the `get_overridden_info` + `get_fully_qualified_table_name` composition they share.

Full Tableau unit suite: **230 passed**.


aviraj-gour:
Fix looks right. Both URN-building sites passed `original_platform` into `get_fully_qualified_table_name`, so a `presto → athena` override produced a URN stamped `athena` but shaped like presto (`catalog.schema.table`) — which never matches what the Athena source emits, so the edge resolves to nothing and nothing errors.

Worth noting these URNs were dangling before the fix. One side effect though: DataHub will have auto-created stub entities for the old bad URNs, and lineage now re-points away from them. Anyone who hand-applied tags or docs to those stubs will find them orphaned. Probably worth a line in the updating_datahub.

metadata-ingestion/src/datahub/ingestion/source/tableau/tableau.py
```diff
@@ -1281,16 +1293,50 @@ def maybe_parse_hostname():
...
+                database_server_id,
+                name,
+                host_name,
+                connection_type,
+            )
+
+            routed_by_hostname = bool(host_name and host_name in hostname_map)
+            routed_by_id = bool(database_server_id and database_server_id in id_map)
+            if not (routed_by_hostname or routed_by_id):
+                self.report.warning(
```
aviraj-gour: Should this be WARN? If someone has 50 connections and only needs to route 2 Snowflake accounts, the other 48 land here — so a recipe that's working as intended reports a permanent warning on every run.

is this intentional? log line above is not enough or should we add `log=False here`?

acrylJonny: Not intentional at that breadth — you're right, partial routing is a normal setup and warning per unrouted connection punishes it. Fixed in d5306003.

Unrouted connections now go to a `database_servers_not_routed` report field instead of a warning, so the 48 still show up in the report for discoverability but don't raise anything.

I kept a warning, but moved it to the case that's actually always a mistake: a map key that matched no connection at all. That's a typo or a connection that no longer exists, and it's silently doing nothing — you'd otherwise only notice when the platform_instance you expected never appears. It fires at most once, and the entry count is bounded by the size of your config rather than by how many connections you have. So the 50-connection/2-mapped recipe is now silent, while a recipe with `athena.us-east-1.amazonaws.com` typo'd gets told.

Went with this over `log=False` because that would keep it in the report UI for the same 48 connections — quieter console, same false alarm where operators actually look.

Three tests pin it: the 50/2 shape stays silent, an unmatched key warns, and a fully-routed config is silent.

Also added the `updating_datahub` line from your other comment — covers the orphaned stubs, since the auto-created entities for the old bad URNs keep any hand-applied tags or docs once lineage re-points.

aviraj-gour: I think my original comment misled you and you built too much on it, sorry.

`report_log` keys entries on `title-message`, so the 48 unrouted connections were never 48 warnings. They were one entry, with each connection appended to that entry's context LossyList (capped at `DATAHUB_REPORT_WARNING_SAMPLE_SIZE`, default 10). Same shape and same retention as database_servers_not_routed — one grouped entry, 10 sampled, total count. So the rewrite didn't actually reduce anything, did it?

Which makes me wonder whether we need the report side at all. The Tableau database server: INFO line already prints every connection — routed and unrouted, unsampled, every run. Anything we put in the report is a random 10-item subset of something the log already has in full, and for debugging you'd go to the log anyway, because the sample is exactly where the connection you're chasing might be missing.

can we keep the log line and drop the rest.
WDYT?

acrylJonny: Agreed — that rewrite didn't buy anything. Kept the INFO line, dropped `database_servers_not_routed` and the unused-map-key warning.

aviraj-gour (commented):
Routing changes look good, thanks for unwinding the report fields.

Hit one thing re-reading the URN change though.

The goal on this line is that the URN matches what the target platform's own source emits — both the shape and what goes in each segment. The switch to the remapped platform gets that right in one direction only.

**3-tier → 2-tier** (e.g. presto → athena): db gets dropped, `db.schema.table` → `schema.table`. Correct, and that's the fix.

**2-tier → 3-tier**: the same line now keeps the db, so we emit three segments. But that db is the *source* platform's database — for a three-tier target the first segment is its catalog, which is a different thing, and nothing in the two-tier source supplies it. We only ever had two meaningful parts, so filling a third from a mismatched concept gives a URN the target's source will never emit. That's the same class of bug this PR is fixing, pointed the other way.

Before the change that case came out two-part, because the check read the original platform.
If the answer is that there's no third segment to supply and it should stay two-part (catalog arriving via `platform_instance`) 

we should focus on constructing correct urn, or if we can not construct warning will help in config/docs/reciepe as a limitation, workaround and current behaviour for 2 to 3 tier.

WDYT?

acrylJonny:
You're right, and the reverse direction was worse than "keeps a mismatched segment" — for a two-tier source Tableau's `database` names the same container the schema already names, so `hive` -> `presto` was emitting `presto_catalog.test-schema.test-schema.test-table`. Same dangling-URN class of bug, pointed the other way, as you said.

Fixed in 47f4c6f. The database is now stripped when **either** side is two-tier, rather than just the target:

- **3-tier -> 2-tier** (presto -> athena): dropped because the target has no slot for it. Unchanged, that's the fix this PR is for.
- **2-tier -> 3-tier** (hive -> presto): dropped because the source has no catalog to give. The URN stays `schema.table` and the catalog has to arrive via `platform_instance`, which is what the pre-existing `hive` -> `presto` test was already doing with `{hive: "my_presto_instance.presto_catalog"}`. So this is back to the pre-PR behaviour for that direction.
- Same tier both ways, or no override: untouched.

Went with constructing the correct URN rather than warning, since there is a right answer here — two parts — and it's the same one master produced. A warning would fire on a config that's working.

`test_lineage_overrides_two_tier_to_three_tier_keeps_two_part_name` pins it; it fails with the duplicated segment if the gate goes back to checking only the target.

Also documented the asymmetry, since "override to a three-tier platform and you still won't get three segments" isn't obvious from the config: a Limitations bullet in the Tableau docs pointing at `platform_instance_map` for the catalog, and the `updating-datahub` note narrowed to the direction whose URNs actually change.

acrylJonny:
Follow-up in 6b6c8d2 — trimmed the comments around this change, and while doing that I dug into why `clickhouse` sits in the trimming list but not the database-stripping one, since deriving one from the other made the asymmetry look like an oversight. It isn't:

| connection | schema present | schema absent |
| --- | --- | --- |
| clickhouse (trimmed, not stripped) | `my_schema.my_table` | `my_db.my_table` |
| mysql (trimmed and stripped) | `my_schema.my_table` | `my_table` |

Trimming to two parts already drops the redundant database when there's a schema, so stripping clickhouse would be a no-op in that case — but keeping the database is what gives a schemaless upstream `db.table` instead of a bare table name. That's now in the comment with the #11230 reference, so nobody collapses the two lists on sight.

That mysql row is a real bug of the same family — a schemaless two-tier upstream loses its only container and comes out one-part. It predates this PR and fires with no `lineage_overrides` involved, so changing it here would move URNs for hive/mysql/athena/teradata users this PR otherwise doesn't touch, and it wouldn't be covered by the upgrade note. Left it alone; happy to raise it separately if you'd like it fixed.

aviraj-gour (approved):

PR #17618, opened 2026-05-28
test(playwright/views): migrate views tests to playwright

purnimagarg1:
**Linear ticket:**
https://linear.app/acryl-data/issue/PFP-4008/migrate-views-tests

**Description:**

This PR migrates views tests to Playwright.

**New/Updated Page Objects:**
```
- view-select.page.ts
- manage-views.page.ts
```

**New test files in t`ests/views/`:**
```
- v2-view-select.spec.ts
- v2-manage-views.spec.ts
```


priyabratadas-dh (commented):
## Test Inspector Review

### Summary
- ✅ 2 new spec files added (v2-manage-views.spec.ts, v2-view-select.spec.ts)
- ✅ 2 new page objects added (manage-views.page.ts, view-select.page.ts)
- ✅ Test data seeding correctly wired in data.setup.ts
- ✅ Frontend source updated with matching `data-testid` attributes
- 🔴 1 critical issue (CSS class-based selectors in page object)
- 🟡 4 improvement observations
- 📝 1 PR title typo

---

### 🔴 Framework Compliance Issues

| File | Lines | Issue | Fix |
|------|-------|-------|-----|
| `view-select.page.ts` | 94–109 | `waitForPopoverOpen` / `waitForPopoverClose` use `document.querySelector('.view-select-popover')` and check `.ant-popover-hidden` CSS class — brittle Ant Design internals that break across upgrades | Replace with a `data-testid` attribute on the popover element in the frontend source (e.g. `data-testid="view-select-popover"`) and use `page.getByTestId(...).waitFor()` instead of `waitForFunction` |

**Note:** The `waitForFunction` with CSS class inspection is in the **page object** (not the spec), so it avoids raw selectors in the spec itself — that architecture is correct. The issue is specifically the reliance on an Ant Design internal class (`ant-popover-hidden`) which is an implementation detail that can change.

---

### 🟡 Improvement Observations

**1. PR Title Typo**
Title reads `test(playwight/views)` — "playwight" should be "playwright". Recommend updating the PR title before merge.

**2. Missing assertions after `setViewAsDefault` / `removeViewAsDefault` in manage-views spec**
In `v2-manage-views.spec.ts`, the test calls `setViewAsDefault(editedViewName)` and `removeViewAsDefault(editedViewName)` but does not assert the outcome (e.g. a "default" badge appearing/disappearing in the row). These steps are untestably passing — they could silently fail without detection.

Suggested assertion in `ManageViewsPage`:
```typescript
async expectViewIsDefault(viewName: string): Promise<void> {
  const viewRow = this.getViewRowByName(viewName);
  await expect(viewRow.getByTestId('default-view-badge')).toBeVisible({ timeout: TIMEOUTS.MEDIUM });
}
```

**3. `v2-manage-views.spec.ts` uses `featureName: 'views'` but doesn't consume seeded data**
`ManageViewsPage` creates its own views from scratch via UI — it doesn't need the glossary term or dataset fixtures. Using `featureName: 'views'` triggers seeding unnecessarily, adding ~1-2s of overhead per worker. If the seeding is needed for future tests in the views describe block, this is fine to keep. Otherwise, removing `test.use({ featureName: 'views' })` from `v2-manage-views.spec.ts` would speed up that test slightly.

**4. `viewsTypeSelect` locator in `ViewSelectPage` is declared and used in `createViewFromSelect` but `views-type-select` may not always be visible**
The `viewsTypeSelect.waitFor({ state: 'visible' })` call (line 130) may be flaky if the type select isn't always shown. This is a minor concern — worth verifying the element is consistently present in the create-view flow.

---

### ✅ Positive Feedback

- Excellent separation of page objects from spec files — all selectors live in page objects, specs are clean and readable
- Correct fixture import: `from '../../fixtures/base-test'` (authenticated tests)
- Page object instantiation pattern (`new ManageViewsPage(page, logger, logDir)` in `beforeEach`) is correct
- All `data-testid` attributes were added to the frontend source in the same PR — eliminates any test ID mismatch risk
- `withRandomSuffix` used for view names — prevents cross-test pollution
- `views-constants.ts` centralizes test data constants — good maintainability practice
- `TIMEOUTS` constant correctly introduced to `utils/constants.ts` and shared across both page objects
- `waitForPopoverOpen/Close` is properly extracted as private methods, not duplicated
- CI integration is automatic — `testDir: './tests'` in `playwright.config.ts` picks up the new `tests/views/` directory without any workflow changes

---

### Recommendations

1. **Must fix before merge:** Replace `ant-popover-hidden` CSS class check in `waitForPopoverOpen/Close` with a proper testable selector
2. **Recommended:** Add assertions after `setViewAsDefault` and `removeViewAsDefault` in the manage-views spec
3. **Nice to have:** Fix the "playwight" typo in the PR title
4. **Consider:** Removing `featureName: 'views'` from `v2-manage-views.spec.ts` if that test doesn't rely on the seeded data

---

_Reviewed by Test Inspector_

priyabratadas-dh (approved):

PR #17623, opened 2026-05-28
fix(snowflake): fix cross-database semantic view lineage

datahub-akshay:
## Summary

Semantic views whose base tables live in a database outside the recipe scope silently produced no `upstreamLineage`. Three separate bugs combined to cause this:

- **DDL fallback for missing REFERENCES privilege**: `INFORMATION_SCHEMA.SEMANTIC_TABLES` requires at least REFERENCES on base-table databases to return cross-database rows. When it returns 0 rows, the connector now falls back to recovering base-table identifiers from the `TABLES ( ... )` clause of the semantic view DDL (already fetched during ingestion, so no additional privileges are needed). `CREATE SEMANTIC VIEW` is not part of sqlglot's grammar (`parse_one` raises, and lenient mode yields an opaque `Command` with no extractable tables), so instead of a hand-rolled regex the parser walks sqlglot's **token stream**. Splitting on tokens rather than text means quoted identifiers, doubled-quote escapes, string comments, and nested parens (`PRIMARY KEY (...)`, `WITH SYNONYMS (...)`, subqueries) are handled by the tokenizer rather than by fragile pattern matching. This fallback only runs when `SEMANTIC_TABLES` returns 0 rows, so it is not a hot path.

- **Upstream URN filter removed**: the pre-compute loop was calling `is_dataset_pattern_allowed()` on each base table URN, silently dropping any table whose database was excluded from `database_pattern`. `sql_parsing_aggregator` does not apply the pattern filter to upstreams; semantic view lineage now matches that behavior.

- **Structured warning on empty upstream set**: the silent `logger.debug` skip is replaced with `structured_reporter.warning` so the gap surfaces in the ingestion report regardless of the `column_lineage` config flag.

## Test plan

- [x] Unit tests pass: `pytest metadata-ingestion/tests/unit/snowflake/`
- [x] `test_parse_base_tables_from_ddl_*` covers plain identifiers, alias-first syntax (`alias AS db.sch.tab`), quoted identifiers, comment-only entries (no PRIMARY KEY), all trailing clauses combined (PRIMARY KEY + WITH SYNONYMS + COMMENT), SQL-query logical tables (skipped), non-qualified and over-qualified names (ignored), doubled-quote escapes inside comments, missing TABLES clause, empty DDL, and malformed DDL (tokenizer error returns an empty mapping instead of crashing ingestion)
- [x] `test_populate_semantic_view_base_tables_ddl_fallback` verifies the fallback triggers when SEMANTIC_TABLES returns 0 rows and populates both `base_tables` and the alias-keyed `logical_to_physical_table`
- [x] `test_process_semantic_views_cross_db_urns_not_filtered_by_database_pattern` verifies the upstream filter regression cannot be silently reintroduced
- [x] `test_process_semantic_views_emits_structured_warning_when_no_upstream_urns` is parametrized over `column_lineage=True/False` to verify the warning fires unconditionally
- [x] End-to-end ingestion of tables + SQL views + semantic view with cross-DB base tables into a test DataHub instance: all 5 entities landed with correct `upstreamLineage` and `fineGrainedLineages`


puneetagarwal-datahub:
Check with the ddl which are present in the zendesk and validate

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_schema.py
```diff
@@ -52,6 +53,25 @@ class SnowflakeTaskState(StrEnum):
...
+    rf"^\s*({_SEM_VIEW_IDENT})\s+AS\s+(.+)$", re.IGNORECASE | re.DOTALL
+)
+_SEM_VIEW_TABLES_BLOCK_RE = re.compile(r"\bTABLES\s*\(", re.IGNORECASE)
+# Optional trailing clauses after a logical table; always preceded by whitespace, so
+# the lookbehind avoids clipping a table whose name ends in one of these words
+# (e.g. DB.SCHEMA.TAG, where TAG follows a '.').
+_SEM_VIEW_TRAILING_CLAUSE_RE = re.compile(
+    r"(?<=\s)(PRIMARY\s+KEY|UNIQUE|CONSTRAINT|WITH\s+SYNONYMS|WITH\s+TAG|COMMENT|TAG)\b",
+    re.IGNORECASE,
+)
```
treff7es: These regular expressions are not just expensive but also error-prone. 
Can't we do it with sqlglot instead?

alfiyas-datahub: sqlglot can't fully understand **CREATE SEMANTIC VIEW** statements yet. It throws a parse error, or in lenient mode returns a generic Command object that doesn't contain the referenced tables. So we can't use its SQL parser to extract lineage. The tokenizer is the best available option in sqlglot.

treff7es (approved):

PR #17625, opened 2026-05-29
fix(ingest/looker): handle glue platform in LookML hive-style lineage

alokr-dhub:
## Summary

Fixes broken lineage and column-level lineage for LookML views that reference Glue Data Catalog tables. In LookML, Glue tables queried via Athena/Presto use the same `hive.<db>.<table>` 3-part naming as Hive, but the ingestion code only special-cased the `hive` platform — Glue URNs ended up with the `hive.` prefix baked into the dataset id, and sqlglot failed to parse hive-flavored SQL because there is no `glue` dialect.

Two small changes, both mirroring existing Hive behaviour:

- `sql_parsing_common.get_dialect_str`: map `glue` → `hive` (sqlglot has no dedicated glue dialect; Glue via Athena/Presto is hive-dialect SQL).
- `looker.view_upstream._drop_hive_dot`: also strip the `hive.` prefix from dataset URNs whose platform is `glue`, so upstream and fine-grained lineage point at `urn:li:dataset:(urn:li:dataPlatform:glue,<db>.<table>,...)` instead of `...,hive.<db>.<table>,...`.

Other platforms are untouched.

## Test plan

- [x] Unit test for `_drop_hive_dot` covering glue (with and without `hive.` prefix), hive, snowflake, and a no-op case — `metadata-ingestion/tests/unit/looker/test_view_upstream.py`
- [x] Unit test confirming `get_dialect("glue")` resolves to the hive dialect — `metadata-ingestion/tests/unit/sql_parsing/test_sqlglot_utils.py::test_glue_resolves_to_hive_dialect`
- [x] Integration test `test_drop_hive_glue` mirroring the existing `test_drop_hive`, with a golden file (`drop_hive_dot_glue_golden.json`) that asserts:
  - upstream URN is `urn:li:dataPlatform:glue,employee_db.income_source,PROD` (glue platform, no `hive.` prefix)
  - fine-grained column lineage edges land on the glue dataset's schema fields
- [x] Re-ran existing `test_drop_hive` — no regression

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/looker/view_upstream.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/looker/view_upstream.py
@@ -152,14 +152,20 @@ def _platform_names_have_2_parts(platform: str) -> bool:
 
 def _drop_hive_dot(urn: str) -> str:
     """
-    This is special handling for hive platform where "hive." is coming in urn's id because of the way SQL
-    is written in lookml.
+    This is special handling for hive and glue platforms where "hive." is coming in urn's id because of the way SQL
+    is written in lookml. Glue tables queried via Presto/Athena use the same `hive.<db>.<table>` syntax.
 
     Example: urn:li:dataset:(urn:li:dataPlatform:hive,hive.my_database.my_table,PROD)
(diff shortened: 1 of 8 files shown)


alokr-dhub:
Superseded by #17626 — same changes, branch now hosted on datahub-project/datahub rather than acryldata/datahub.

PR #17627, opened 2026-05-29
fix(ingest): treat Hive `list<>` as a synonym for `array<>`

acrylJonny:
## Summary

The `HiveColumnToAvroConverter` only recognised `array<T>` as the Hive
array shape. Iceberg, and several Athena / Glue clients exposing Iceberg
tables, emit `list<T>` instead — semantically the same.

When the converter hit `list<...>` it fell through to the
unknown-type path and the entire column collapsed to `NullType`. This
showed up downstream as missing or empty column names for any complex
field with a nested list, which broke schema visibility and column-level
lineage extraction for those columns.

### Change

- Add `list` to `_COMPLEX_TYPE` and accept `list<T>` alongside `array<T>` in `_parse_datatype_string`.
- Read the inner type using `s.index("<") + 1` so the same code path handles both prefix lengths.
- Add two unit tests:
  - `test_list_type_treated_as_array` — direct synonym.
  - `test_list_of_structs_inside_struct_does_not_collapse_to_null` — the Iceberg-on-Glue-via-Athena shape that motivated the change.

### Refs

ING-2085

## Checklist

- [x] Tests added
- [x] No breaking changes

### DIFF
--- a/metadata-ingestion/src/datahub/utilities/hive_schema_to_avro.py
+++ b/metadata-ingestion/src/datahub/utilities/hive_schema_to_avro.py
@@ -45,7 +45,7 @@ class HiveColumnToAvroConverter:
         "byteint": "int",
     }
 
-    _COMPLEX_TYPE = re.compile("^(struct|map|array|uniontype)")
+    _COMPLEX_TYPE = re.compile("^(struct|map|array|list|uniontype)")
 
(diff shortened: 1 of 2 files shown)


acrylJonny:
Superseded by #16999 which already contains this fix (along with the NullType fieldPath fix, Pandas dtype handling, and S3 Tables discovery). Closing in favour of moving #16999 forward.

PR #17628, opened 2026-05-29
fix(ingest): give NullType columns a unique fieldPath suffix

acrylJonny:
## Summary

Defence-in-depth complement to #17627 (the `list<>` Hive synonym).

When SqlAlchemy can't reflect a column type — either because the dialect
lacks a mapping or because the engine returns a shape the dialect doesn't
expose — it surfaces the column as a real `types.NullType()` instance.
Each such column was suffixed only with `[version=2.0].[type=null]` (no
column name), so two unmapped columns on the same table collided on a
single path and presented downstream as missing or duplicate columns.

### Change

- Add `types.NullType` to the set of types that get the column name
  appended to their `fieldPath` in `get_schema_fields_for_sqlalchemy_column`.
- Add `test_null_type_columns_get_unique_field_paths` covering the
  collision scenario directly.

Existing behaviour for the unknown-string fallback (which passes a `str`
through `get_schema_fields_for_sqlalchemy_column`) is preserved — that
path doesn't hit the new branch because `type("foo") is str`, not
`NullType`.

### Stack

1. [#17627](https://github.com/datahub-project/datahub/pull/17627) — `list<>` Hive synonym
2. **this PR** — NullType fieldPath uniqueness
3. (next) Pandas nullable dtype handling in the Athena dialect

### Refs

ING-2085

## Checklist

- [x] Tests added
- [x] No breaking changes (existing assertions retained)

### DIFF
--- a/metadata-ingestion/src/datahub/utilities/sqlalchemy_type_converter.py
+++ b/metadata-ingestion/src/datahub/utilities/sqlalchemy_type_converter.py
@@ -226,12 +226,19 @@ def get_schema_fields_for_sqlalchemy_column(
             )
         ]
 
-    # for all non-nested data types an additional modification of the `fieldPath` property is required
+    # For all non-nested data types we additionally suffix `fieldPath` with the column name
+    # so each column has a unique path. NullType is included for two reasons:
+    #   1. It is the fallback for any column whose type SqlAlchemy can't reflect (rare dialect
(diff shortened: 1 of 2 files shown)


acrylJonny:
Superseded by #16999 (see the comparison table in that PR's thread).

PR #17634, opened 2026-05-29
fix(graph): key getMergedEdges by Edge to avoid Duplicate key on hashCode collision

### LINKED ISSUE #17635: `GraphIndexUtils.getMergedEdges` throws `IllegalStateException: Duplicate key` on Edge hashCode collisions
# Issue: `GraphIndexUtils.getMergedEdges` throws `IllegalStateException: Duplicate key` on Edge hashCode collisions

## Summary

When the graph-index update path processes a relationship aspect (e.g. `upstreamLineage`)
whose edge set contains two **distinct** edges that happen to share the same 32-bit
`hashCode()`, `GraphIndexUtils.getMergedEdges` throws
`java.lang.IllegalStateException: Duplicate key …`. This fails `UpdateIndicesHook`
(and `PlatformEventGeneratorHook`) for that `MetadataChangeLog`, so the graph index
is not updated for the affected entity.

## Affected version

DataHub `v1.6.0`; the code is unchanged on `master`
(`metadata-service/services/src/main/java/com/linkedin/metadata/graph/GraphIndexUtils.java`).

## Root cause

`getMergedEdges` builds a `Map<Integer, Edge>` keyed on the raw `Edge#hashCode()`:

```java
final Map<Integer, Edge> oldEdgesMap =
    oldEdgeSet.stream()
        .map(edge -> Pair.of(edge.hashCode(), edge))
        .collect(Collectors.toMap(Pair::getFirst, Pair::getSecond));   // no merge function
...
if (oldEdgesMap.containsKey(newEdge.hashCode())) {                     // lookup by int hashCode
  final Edge oldEdge = oldEdgesMap.get(newEdge.hashCode());
```

Two problems:
1. `Collectors.toMap` with no merge function throws `IllegalStateException` when two
   edges in `oldEdgeSet` collide on `hashCode()`.
2. The lookup matches on `hashCode()` rather than `Edge#equals`, so even without the
   exception, a collision would match the wrong edge.

`Edge` (`com.linkedin.metadata.aspect.models.graph.Edge`) is a Lombok `@Data` type whose
`hashCode`/`equals` include `source`, `destination`, `relationshipType`, `lifecycleOwner`,
`via`. Distinct edges (e.g. to different `schemaField` destinations) can collide in the
32-bit hash space.

## Stack trace (abbreviated, from v1.6.0)

```
ERROR c.l.m.k.l.AbstractKafkaListener - Failed to execute hook with name UpdateIndicesHook
java.lang.IllegalStateException: Duplicate key 1895094058 (attempted merging values Edge(...) and Edge(...))
    at java.base/java.util.stream.Collectors.duplicateKeyException(...)
    at com.linkedin.metadata.graph.GraphIndexUtils.getMergedEdges(GraphIndexUtils.java:493)
    at com.linkedin.metadata.graph.GraphIndexUtils.computeAspectEdgeDiff(GraphIndexUtils.java:459)
    at com.linkedin.metadata.service.UpdateGraphIndicesService.updateGraphServiceDiff(...)
    ...
```

## How to reproduce

Ingest a dataset with **fine-grained (column-level) lineage** that produces many
`schemaField` edges, such that two distinct edges collide on `Edge#hashCode()`. Observed in
production on Iceberg datasets with column-level lineage to a large number of upstream
`schemaField`s — two edges to different destination fields collided (key `1895094058`),
e.g. `…/AGAS/2/… FaultWord2` vs `…/AGAS/3/… FaultWord1`.

A minimal deterministic reproduction (used in the regression test) relies on the classic
`"Aa"`/`"BB"` String hashCode collision via `relationshipType`.

## Impact

- The graph-index update for the affected `MetadataChangeLog` is dropped (lineage edges
  not refreshed in the graph index). The source-of-truth aspect in the local DB is fine.
- Severity depends on the consumer error-handling configuration; in the worst case repeated
  failures generate significant error-log noise.

## Proposed fix

Key the map by the `Edge` itself (relying on `Edge#equals`/`#hashCode`) instead of the raw
int hashCode, and add a merge function for safety. Fixed by #17634

daha:
## What

`GraphIndexUtils.getMergedEdges` built a `Map<Integer, Edge>` keyed on `Edge#hashCode()`
using `Collectors.toMap` with no merge function. When an aspect's edge set contained two
distinct edges that share a 32-bit `hashCode()`, `toMap` threw
`IllegalStateException: Duplicate key`, which failed `UpdateIndicesHook` for that
`MetadataChangeLog` and skipped the graph-index update for the entity. The subsequent
lookup also matched edges by `hashCode()` rather than `equals`.

This change keys the map by the `Edge` itself (using `Edge#equals`/`#hashCode`, the correct
identity for matching old vs. new edges) and adds a defensive merge function.

## Why

Observed in production (v1.6.0) on datasets with **fine-grained (column-level) lineage**:
two distinct `schemaField` edges collided on `Edge#hashCode()`, throwing
`IllegalStateException: Duplicate key 1895094058 …` from `getMergedEdges` and breaking the
`UpdateIndicesHook` for the affected lineage updates.

## How

```diff
- final Map<Integer, Edge> oldEdgesMap =
-     oldEdgeSet.stream()
-         .map(edge -> Pair.of(edge.hashCode(), edge))
-         .collect(Collectors.toMap(Pair::getFirst, Pair::getSecond));
+ final Map<Edge, Edge> oldEdgesMap =
+     oldEdgeSet.stream()
+         .collect(Collectors.toMap(edge -> edge, edge -> edge, (existing, replacement) -> existing));
...
- if (oldEdgesMap.containsKey(newEdge.hashCode())) {
-   final Edge oldEdge = oldEdgesMap.get(newEdge.hashCode());
+ if (oldEdgesMap.containsKey(newEdge)) {
+   final Edge oldEdge = oldEdgesMap.get(newEdge);
```

`getMergedEdges` is made package-private (`@VisibleForTesting`) so the regression test can
call it directly.

## Tests

Adds `GraphIndexUtilsTest#testGetMergedEdgesHandlesHashCodeCollision`, which constructs two
distinct edges that collide on `hashCode()` (via the classic `"Aa"`/`"BB"` String collision
on `relationshipType`), asserts the collision precondition, and verifies `getMergedEdges`
no longer throws and merges the matching edge.

Run:
```
./gradlew :metadata-service:services:test --tests "com.linkedin.metadata.graph.GraphIndexUtilsTest"
```
## Other info
Fixes #17635 

The fix and description has been generated with Claude Code (Opus 4.8 with 1M context). It was generated from analysis of an error log in our local installation of Datahub. 

### DIFF
--- a/metadata-service/services/src/main/java/com/linkedin/metadata/graph/GraphIndexUtils.java
+++ b/metadata-service/services/src/main/java/com/linkedin/metadata/graph/GraphIndexUtils.java
@@ -3,6 +3,7 @@ package com.linkedin.metadata.graph;
 import static com.linkedin.metadata.Constants.*;
 
 import com.datahub.util.RecordUtils;
+import com.google.common.annotations.VisibleForTesting;
 import com.linkedin.common.InputField;
 import com.linkedin.common.InputFields;
 import com.linkedin.common.urn.Urn;
@@ -486,17 +487,22 @@ public class GraphIndexUtils {
     return new HashSet<>(edgeAndRelationTypes.getFirst());
   }
 
-  private static List<Edge> getMergedEdges(final Set<Edge> oldEdgeSet, final Set<Edge> newEdgeSet) {
-    final Map<Integer, Edge> oldEdgesMap =
-        oldEdgeSet.stream()
-            .map(edge -> Pair.of(edge.hashCode(), edge))
-            .collect(Collectors.toMap(Pair::getFirst, Pair::getSecond));
+  @VisibleForTesting
+  static List<Edge> getMergedEdges(final Set<Edge> oldEdgeSet, final Set<Edge> newEdgeSet) {
+    // Key by the Edge itself (using Edge#equals/#hashCode), not by the raw
+    // Edge#hashCode() value. Two distinct edges can share a 32-bit hashCode, which
+    // previously made Collectors.toMap throw "IllegalStateException: Duplicate key"
+    // and made the lookup below match colliding edges instead of equal ones.
+    // oldEdgeSet is a Set, so its elements are already distinct by Edge#equals and
+    // cannot produce duplicate keys; no merge function is needed.
+    final Map<Edge, Edge> oldEdgesMap =
+        oldEdgeSet.stream().collect(Collectors.toMap(edge -> edge, edge -> edge));
 
     final List<Edge> mergedEdges = new ArrayList<>();
     if (!oldEdgesMap.isEmpty()) {
       for (Edge newEdge : newEdgeSet) {
-        if (oldEdgesMap.containsKey(newEdge.hashCode())) {
-          final Edge oldEdge = oldEdgesMap.get(newEdge.hashCode());
+        if (oldEdgesMap.containsKey(newEdge)) {
+          final Edge oldEdge = oldEdgesMap.get(newEdge);
           final Edge mergedEdge = GraphIndexUtils.mergeEdges(oldEdge, newEdge);
           mergedEdges.add(mergedEdge);
         }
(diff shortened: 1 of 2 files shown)


david-leifker (approved):
lgtm! Thank you!

PR #17641, opened 2026-05-29
fix(search): prevent ES8 reindex loops and normalize legacy range queries

david-leifker:
## Summary
- Add engine-specific index settings comparison via `IndexSettingsComparison` so ES8-injected `type=custom` on analyzers and case-only boolean diffs no longer trigger spurious reindex cycles
- Normalize legacy OpenSearch `RangeQueryBuilder` JSON (`from`/`to` with `include_lower`/`include_upper`) to ES8-compatible bounds (`gte`/`gt`/`lte`/`lt`) before query conversion
- Wire the ES8 shim into `ReindexConfig` through `ESIndexBuilder` and add unit/integration tests covering round-trip settings and mapping stability

### DIFF
--- a/docs/how/updating-datahub.md
+++ b/docs/how/updating-datahub.md
@@ -170,9 +170,11 @@ Requirements:
 
 ### Known Issues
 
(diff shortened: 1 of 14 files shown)


docs/how/updating-datahub.md
```diff
@@ -170,9 +170,11 @@ Requirements:
 
 ### Known Issues
 
+- **(Operations / Elasticsearch 8) Perpetual reindex loop on system-update.** Deployments backed by **Elasticsearch 8.x** can get stuck in a loop where every system-update pass detects spurious index mapping drift against what ES8 persisted (for example ES8-injected `type: object` on nested object fields) and triggers another full reindex. This is most likely when **`ELASTICSEARCH_INDEX_BUILDER_MAPPINGS_REINDEX=true`** (including Docker quickstart, which enables it for initial index builds). **Workaround:** Let the **first** system-update after install or upgrade finish so indices are created, then set **`ELASTICSEARCH_INDEX_BUILDER_MAPPINGS_REINDEX=false`** on **`datahub-gms`** and **`datahub-upgrade`** before later system-update runs until a fixed release is available.
```
RyanHolstien: This comment reads kind of weirdly, is the issue fixed here? I'm not sure we should be recommending setting this setting to false for people since it will skip doing the reindexing when it should do so, no?

RyanHolstien: I think the "loop" description is also more worrisome than this should probably indicate. It's not a loop of reindexing, it's that it will always reindex for a specific helm deploy.

RyanHolstien (approved):

docs/how/updating-datahub.md
david-leifker: This is under v1.6.0 section which is released with the bug. I will update the description to be clear about the fact that it only applies to helm update or system-update run. Indicating clearly that its not a loop. Good callout

PR #17645, opened 2026-05-31
fix(utilities): make LossyList.__contains__ unwrap stored tuples

acrylJonny:
## Summary

`LossyList` stores items internally as `(index, item)` tuples for reservoir sampling and overrides `__getitem__` / `__iter__` to hide that detail. But it never overrode `__contains__`, so the inherited `list.__contains__` compared against the raw tuples — making `x in lossy_list` always `False` for the un-tupled item that callers actually have a reference to.

This silently breaks membership tests across every caller, most visibly anything inspecting `report.filtered` to decide whether an entity was already reported on.

## Changes

- Override `LossyList.__contains__` to iterate via the existing unwrap path (`__iter__`) so there is a single source of truth for the tuple → item projection.
- Add two regression tests:
  - positive/negative membership coverage on the small-list path (pre-reservoir-sampling),
  - survivors-must-be-findable check after enough appends to trigger reservoir sampling.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/metadata-ingestion/src/datahub/utilities/lossy_collections.py
+++ b/metadata-ingestion/src/datahub/utilities/lossy_collections.py
@@ -83,13 +83,9 @@ class LossyList(List[T], Generic[T]):
         yield from [elem[1] for elem in sorted(super().__iter__())]  # type: ignore
 
     def __contains__(self, item: object) -> bool:
-        """Membership test against unwrapped items.
-
-        Internally items are stored as (index, item) tuples for reservoir
-        sampling, so the inherited list.__contains__ would only match against
-        those tuples. Iterate via __iter__ (which unwraps) so callers can use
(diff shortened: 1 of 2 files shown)


acrylJonny:
Applied the cosmetic suggestion in dca8081bee — swapped `self.__iter__()` for `any(elem == item for elem in self)`. Iterating `self` goes through our overridden `__iter__` (single source of truth for unwrapping), so the cleaner spelling actually makes the intent more obvious. All 19 LossyList tests still pass; ruff clean.

Agreed on the rest of your read — the "no existing callers were using `x in lossy_list` in production code" framing is the right one. This was a latent silent bug that would have bitten the next person who tried the natural idiom against an instance returned from a `report.failures` / `processed_entities` field.

treff7es (approved):

PR #17646, opened 2026-05-31
fix(ingestion/dremio): respect schema_pattern and dataset_pattern in lineage emission

acrylJonny:
## Summary

Query lineage (from `sys.jobs_recent`) and view-parent lineage discovered URNs were emitted unconditionally, bypassing the catalog walk's `schema_pattern` / `dataset_pattern` / `_accelerator_` gates. DataHub then materialized **ghost dataset entities** for filtered schemas — entities with no parent container and no browse path, the exact opposite of what the operator configured. This is the same anti-pattern that Snowflake and Redshift fixed via `is_allowed_table` callbacks.

## Changes

- `SqlParsingAggregator` is now constructed with an `is_allowed_table` callback so lineage / usage / operations emission gates on the same filters used by the catalog walk. Discovered names short-circuit through `catalog_dataset_names`; others are tested against `dataset_pattern`, `schema_pattern` (container portion), and the `_accelerator_` exclusion.
- `process_query` and `generate_view_lineage` pre-filter their upstreams/downstream before handing them to the aggregator, so filtered URNs never reach the query map. `add_observed_query` still runs so usage stats for in-scope datasets are preserved.
- New report counter `num_lineage_dropped_filtered` so operators can see how many lineage references were skipped.

## Tests

- `tests/unit/dremio/test_dremio_lineage_filter.py`:
  - `TestPassesDremioFilters` — pure-logic coverage of `_passes_dremio_filters` (schema/dataset patterns, `_accelerator_` exclusion, `catalog_dataset_names` shortcut).
  - `TestDremioSourceLineageFilter` — integration coverage of `_is_allowed_table`, `process_query`, and `generate_view_lineage` against a stubbed aggregator, asserting filtered URNs never reach `add_known_query_lineage` / view-lineage emit paths.

## Docs

- Added an "Other Notable Changes" entry to `docs/how/updating-datahub.md` describing the new gating behavior and the report counter.

## Checklist

- [x] Tests added
- [x] Docs added (`updating-datahub.md`)
- [x] Behavior-change entry in `updating-datahub.md`


acrylJonny:
Thanks for the careful review. Addressed in ba88717fba:

**Minor suggestions**
- **Promoted `_passes_dremio_filters` → `passes_dremio_filters`.** It has multiple callers and is imported directly by tests; the underscore was no longer accurate.
- **Renamed report counter `num_lineage_dropped_filtered` → `lineage_dropped_filtered`.** Matches the existing `containers_filtered` / `containers_scanned` style in `DremioSourceReport`. Updated `updating-datahub.md` accordingly.
- **Added a clarifying comment on the asymmetry** at the `elif not downstream_allowed:` branch. The case where `downstream_allowed=True` but all upstreams are filtered is already counted per-edge in the `queried_datasets` loop, so not counting it again there avoids double-counting — explicit comment now says so.
- **Added the missing assertion** in `test_is_allowed_table_rejects_reflection` — it now checks `lineage_dropped_filtered` is incremented, matching `test_is_allowed_table_rejects_filtered_schema`.

**Needs Human Eyes — `add_observed_query` + `is_allowed_table` interaction**
Walked through `SqlParsingAggregator` to confirm no leak. The aggregator consults `is_allowed_table` at three gates before emitting any aspect:
1. `sql_parsing_aggregator.py:1434` — `if not self.is_allowed_table(downstream_urn): return` blocks per-downstream lineage emission for filtered URNs, incrementing `num_lineage_skipped_due_to_filters`.
2. `sql_parsing_aggregator.py:978` — in the usage emission path, each upstream URN parsed from the query SQL is dropped if `not self.is_allowed_table(upstream_urn, allow_external=False)`.
3. `sql_parsing_aggregator.py:1643` — whole-query gate: if no allowed table is involved, the query is skipped entirely.

So the case the reviewer flagged (downstream allowed, all explicit upstreams pre-filtered, `add_known_query_lineage` skipped, `add_observed_query` still runs, parser re-discovers the same upstreams) is safe: the re-discovered URNs are filtered at gate (2) for usage and gate (1) for lineage. Replaced the brief safety comment above `add_observed_query` with one that enumerates these three gates explicitly so future readers don't have to redo the audit.

All 15 unit tests in `test_dremio_lineage_filter.py` pass; ruff clean.

metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_source.py
```diff
@@ -611,34 +672,59 @@ def process_query(self, query: DremioQuery) -> None:
...
+                    KnownQueryLineageInfo(
+                        query_text=query.query,
+                        upstreams=upstream_urns,
+                        downstream=downstream_urn,
+                    ),
+                    merge_lineage=True,
+                )
+            elif not downstream_allowed:
+                # Per-upstream drops are already counted in the loop above.
+                self.report.lineage_dropped_filtered += 1
```
treff7es, line 727: Counter semantics: when downstream is filtered AND upstreams are also filtered, both counts accumulate — lineage_dropped_filtered can exceed the number of edges dropped. Add a comment clarifying the unit is "lineage references dropped" not "edges."

treff7es (approved):

metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_source.py
acrylJonny: Addressed in a follow-up commit on the same PR (merged with master at 40a06350f8). The counter is now documented as a "is the filter biting?" signal rather than an edge count:

```python
# Counter for lineage edges/URNs dropped by the
# schema_pattern / dataset_pattern. Increments once per filtered
# view-parent, once per filtered query upstream, once per query whose
# downstream is filtered (regardless of how many upstreams that
# implicitly drops), and once per URN the SqlParsingAggregator
# rediscovers during SQL parsing. A single dropped edge can therefore
# contribute multiple increments (e.g. when both sides of an edge are
# filtered); use it as a "is the filter biting?" signal, not an exact
# edge count.
lineage_dropped_filtered: int = 0
```

And the call site that increments for a filtered downstream has an inline comment cross-referencing the docstring:

```python
# One increment per query whose downstream is filtered.
# Per-upstream drops were already counted in the loop above;
# this is intentionally not multiplied by the upstream count
# — see lineage_dropped_filtered's docstring for the "lineage
# references, not edges" semantics.
self.report.lineage_dropped_filtered += 1
```

PR #17647, opened 2026-05-31
fix(ingestion/dremio): auto-detect ARRAY<VARCHAR> queried_datasets on Dremio 26.1.0+

acrylJonny:
## Summary

Dremio Software [26.1.0](https://docs.dremio.com/current/release-notes/version-260-release/) changed `sys.jobs_recent.queried_datasets` from `VARCHAR` (a comma-separated bracket string) to `ARRAY<VARCHAR>`. The existing EE/CE jobs query used `LENGTH(queried_datasets)` and selected the column raw, both of which fail at runtime against the array type with errors like:

```
Function ARRAY_SIZE not found with arguments [VARCHAR]
```

This broke usage/lineage ingestion for users upgrading their Dremio cluster to 26.1.0+, with no opt-out short of disabling the extraction entirely.

## Changes

- Add a one-shot runtime probe `_supports_array_queried_datasets` that issues a minimal `SELECT ARRAY_SIZE(queried_datasets) FROM SYS.JOBS_RECENT LIMIT 1` and caches the result on the API client. Probe runs once per ingestion process; subsequent calls hit the cached boolean.
- The probe catches `Exception` broadly: `execute_query_iter` can raise `DremioAPIException` (SQL-level type-mismatch on legacy VARCHAR), `RuntimeError` (job FAILED/CANCELED), or network/transport errors. Any failure -> assume legacy; the real jobs query will raise a meaningful error later if the underlying problem isn't the schema version.
- On probe success, use new `get_query_all_jobs_array` which mirrors the cloud query (`CONCAT('[', ARRAY_TO_STRING(queried_datasets, ','), ']')`) so the downstream parser (`DremioQuery._get_queried_datasets`) consumes a stable bracket-string regardless of server version.
- On probe failure (Dremio < 26.1.0, or no jobs visible), fall back to the existing VARCHAR-shaped query unchanged.
- Cloud Dremio always uses `get_query_all_jobs_cloud` — no probe needed there.

## Tests

- `tests/unit/dremio/test_dremio_jobs_array.py`:
  - `TestQueriedDatasetsArrayQuery` — SQL string assertions on the new array-form query.
  - `TestSupportsArrayQueriedDatasetsProbe` — probe caching, cloud edition short-circuit, the `DremioAPIException` fallback path, and the `RuntimeError` fallback path (covers `execute_query_iter`'s FAILED/CANCELED job-state raises).
  - `TestExtractAllQueriesDispatch` — verifies the right SQL is dispatched for each (edition, probe-result) combination.
- `test_extract_all_queries` in `test_dremio_api_pagination.py` pins `_queried_datasets_is_array=True` so the new probe doesn't consume the mocked iterator under existing tests.

## Docs

No `updating-datahub.md` entry: this is a forward-compat fix that auto-detects the server schema and keeps the externally-observable output (bracket-string `queried_datasets`) identical across Dremio versions. There is no behavior or configuration change for users to act on.

## Checklist

- [x] Tests added
- [ ] Docs added (`updating-datahub.md`) — N/A, see Docs section above
- [ ] Behavior-change entry in `updating-datahub.md` — N/A, no behavior change

### DIFF
--- a/docs/how/updating-datahub.md
+++ b/docs/how/updating-datahub.md
@@ -60,6 +60,8 @@ Requirements:
 
 - **(Ingestion / dbt)** dbt test assertion entities now emit an `ownership` aspect when the dbt test node has explicit owner metadata (`meta.owner` / `config.meta.owner`).
 
+- **(Ingestion / Dremio)** The Dremio Software connector now auto-detects whether `sys.jobs_recent.queried_datasets` is exposed as `VARCHAR` (pre-26.1.0) or `ARRAY<VARCHAR>` (26.1.0+) by probing the column once per ingestion and selecting an appropriate jobs query. Users upgrading their Dremio cluster to 26.1.0 or newer no longer have to disable usage/lineage extraction to avoid a `Function ARRAY_SIZE not found` error. No recipe changes are required; the probe runs lazily before query extraction.
+
 ## v1.6.0
 
 Requirements:
--- a/docs/how/updating-datahub.md
+++ b/docs/how/updating-datahub.md
@@ -60,8 +60,6 @@ Requirements:
 
 - **(Ingestion / dbt)** dbt test assertion entities now emit an `ownership` aspect when the dbt test node has explicit owner metadata (`meta.owner` / `config.meta.owner`).
 
-- **(Ingestion / Dremio)** The Dremio Software connector now auto-detects whether `sys.jobs_recent.queried_datasets` is exposed as `VARCHAR` (pre-26.1.0) or `ARRAY<VARCHAR>` (26.1.0+) by probing the column once per ingestion and selecting an appropriate jobs query. Users upgrading their Dremio cluster to 26.1.0 or newer no longer have to disable usage/lineage extraction to avoid a `Function ARRAY_SIZE not found` error. No recipe changes are required; the probe runs lazily before query extraction.
-
 ## v1.6.0
 
 Requirements:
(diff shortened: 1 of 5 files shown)


acrylJonny:
Thanks for the careful review — all three points addressed in fd70832576 + PR-description edit:

**Bug — exception catch too narrow.** You're right and the impact is exactly what you described. Confirmed `execute_query_iter` raises `RuntimeError` directly for FAILED (`dremio_api.py:471`) and CANCELED (`dremio_api.py:473`) job states, and can also surface transport errors from the underlying session. The whole purpose of the probe is "any failure → assume legacy", so the catch is now `Exception` with a brief comment enumerating what can actually escape. If the underlying issue isn't the schema version (e.g. permissions, missing table, transport), the real jobs query will raise a meaningful error a few steps later — same surface a user without this PR would see.

**Test gap.** Added `test_probe_falls_back_on_non_dremio_exception` that raises a `RuntimeError("Query failed: planner exception")` from the mocked `execute_query_iter` and asserts the probe returns `False`. Mirrors the existing `test_varchar_column_probe_raises_and_falls_back` shape. All 14 unit tests pass; ruff clean.

**Checklist inconsistency.** Edited the PR description: removed the now-stale "Docs added" check, replaced it with an explicit N/A note explaining why this is a forward-compat fix with no externally-observable behavior change for users.

treff7es (approved):

PR #17648, opened 2026-05-31
feat(ingestion/dremio): map RESTCATALOG / Polaris / Unity / BigQuery / HANA source types

acrylJonny:
## Summary

Dremio's [Catalog API](https://docs.dremio.com/dremio-cloud/api/catalog/source/) publishes a fixed enum of supported source types. Several of those — notably the unified `RESTCATALOG` connector (Apache Polaris OSS, Nessie+REST, AWS Glue Iceberg REST, S3 Tables, Confluent Tableflow, Microsoft OneLake), plus `BIGQUERY`, `SAPHANA`, `SNOWFLAKEOPENCATALOG`, and `UNITY` — were missing from the Dremio → DataHub platform map.

Without explicit entries, `get_datahub_platform` lowercased the unknown source type (so a Polaris source became `urn:li:dataPlatform:restcatalog`, Unity Catalog became `unity`, etc.), and `get_category` returned `"unknown"`, which silently disables the container-vs-file dispatch used by URN building and cross-platform lineage.

## Changes

Add the documented mappings to `DremioToDataHubSourceTypeMapping`:

| Dremio source type | DataHub platform | Category | Notes |
|---|---|---|---|
| `BIGQUERY` | `bigquery` | database | Dremio Cloud only |
| `RESTCATALOG` | `iceberg` | database | Polaris OSS, Nessie+REST, Glue Iceberg REST, S3 Tables, Tableflow, OneLake |
| `SAPHANA` | `hana` | database | EE only |
| `SNOWFLAKEOPENCATALOG` | `iceberg` | database | Snowflake-hosted Polaris (Iceberg REST) |
| `UNITY` | `databricks` | database | Databricks Unity Catalog |

## Behavior change

Existing deployments that ingested any of these sources will see their dataset URNs migrate to the correct `urn:li:dataPlatform:<correct>` value. Previously these sources lived under URN platforms like `restcatalog` and `unity` which don't exist in DataHub's `DataPlatformInfo` registry — so cross-platform lineage was not working anyway.

To avoid orphaned entities post-upgrade, enable stateful ingestion with stale entity removal on the next run, or perform a one-time URN rewrite for the affected datasets.

## Tests

- `tests/unit/dremio/test_dremio_source_type_mapping.py`:
  - `test_new_source_types_map_to_expected_platforms` — each new type returns its target platform.
  - `test_new_source_types_categorized_as_database` — each is classified as `database` for URN/path dispatch.
[...]

PR #17649, opened 2026-05-31
feat(ingestion/dremio): wire domain aspect emission and replace add_mapping with config

acrylJonny:
## Summary

Two related half-wired features cleaned up:

1. **`domain` recipe field never emitted a `Domains` aspect.** `DremioSourceConfig.domain` was documented and plumbed through to `DremioAspects.__init__` but `create_domain_aspect()` was never called from `populate_container_mcp` or `populate_dataset_mcp`, so the setting silently did nothing.
2. **`DremioToDataHubSourceTypeMapping.add_mapping` was dead code with a footgun.** Never called anywhere in DataHub, but the pattern (class-level mutation of the static map) leaked overrides across DremioSource instances in the same pipeline.

> **Stacking note:** This PR is stacked on **#17648** (`dremio-source-mappings-polaris`), which adds the Polaris/Iceberg REST/Unity/BigQuery/HANA entries to the same `SOURCE_TYPE_MAPPING`. Merge #17648 first.

### Domain aspect wiring (`dremio_aspects.py`)

`populate_container_mcp` and `populate_dataset_mcp` now call `self.create_domain_aspect()` and yield the resulting `DomainsClass` MCP when `domain` is set.

### `source_type_mappings` config field (replaces `add_mapping`)

- New recipe field `source_type_mappings: Dict[str, DremioSourceTypeOverride]` where each entry declares `platform` and an optional `category` (`database` / `file_object_storage`).
- `DremioToDataHubSourceTypeMapping` now supports per-instance overlays via `__init__(extra_mappings=...)`. Built-in mappings stay class-level; overrides live on the instance and never mutate the shared dict.
- Public API:
  - `lookup_datahub_platform` / `lookup_category` (instance methods) — honor recipe overrides.
  - `get_datahub_platform` / `get_category` (`@staticmethod`s) — built-in mapping only, kept for callers without a config in hand.
- Wired through `DremioAPIOperations` → `DremioSource.source_type_mapper` → `build_dremio_source_map`.
- `add_mapping` static method **removed**; replaced by the safer config-driven path above.

Example recipe:
```yaml
source:
  type: dremio
  config:
    source_type_mappings:
      MYORG_KAFKA: { platform: kafka, category: database }
      NESSIE: { platform: iceberg-rest } # also overrides a built-in
```

## Behavior changes

- Recipes already setting `domain` will start emitting the configured domain — verify the URN/name is what you intend before upgrading.
- `add_mapping` callers (none in DataHub itself) will see `AttributeError`. The documented migration path is the new `source_type_mappings` config field.

## Tests

- `tests/unit/dremio/test_dremio_domain_aspect.py`:
  - `create_domain_aspect()` behavior — `None` when unset, URN passthrough, stable hashed URN for bare names.
  - Domain aspect emission for both containers and datasets.
  - Regression guard `test_add_mapping_removed`.
- `tests/unit/dremio/test_dremio_source_type_overrides.py`:
  - Built-in mapping intact with no overrides.
  - Unknown source-type fallback to lowercased name.
  - Override registration, case-insensitivity, override-wins-over-builtin, optional category, file-object-storage category.
  - **Per-instance scoping** — overrides on one instance must not leak to another (the `add_mapping` footgun this replaces).
  - Static lookups unaffected by instance overrides.

## Docs

- Two "Other Notable Changes" entries in `docs/how/updating-datahub.md` (domain wiring, `source_type_mappings` replacing `add_mapping`).

## Checklist

- [x] Tests added
- [x] Docs added (`updating-datahub.md`)
- [x] Behavior-change entries in `updating-datahub.md`


acrylJonny:
Thanks for the careful review! Pushed `7230aff` addressing both major issues and three of the four minor items. The fourth (`AZURE_STORAGE` dual-membership) was already addressed in the stacked B4 PR #17648 with an inline comment + pin test, so it auto-shows up here after the rebase onto the new B4 tip.

**Major #1 — Replace UUID5 hash with `DomainRegistry`**

You're right, the UUID5 path was non-standard and effectively a dead-letter office. The new `create_domain_aspect` now:

- Full URN (`urn:li:domain:<id>`) → pass through directly, no registry needed.
- Bare name (e.g. `marketing`) → resolved via `DomainRegistry.get_domain_urn`, which tries `urn:li:domain:<name>` first then falls back to `graph.get_domain_urn_by_name`. Identical pattern to Snowflake / BigQuery / `sql_common.py`.
- Bare name + no graph in the recipe → log a warning and skip the aspect rather than emit a synthetic URN that won't match anything in DataHub.

Construction happens in `DremioSource.__init__` only when the user actually set a bare name (full URNs skip the registry to avoid an unnecessary graph dependency).

`uuid` import dropped from `dremio_aspects.py`. `updating-datahub.md` expanded to call out the new graph requirement for bare names.

**Major #2 — Type guard in `_collect_domain_aspects`**

Replaced the comprehension with an explicit loop that filters on `isinstance(wu.metadata, MetadataChangeProposalWrapper)` before touching `.aspect`. An MCE workunit can no longer trip the assertion silently.

Also rewrote the bare-name test cases:

- `test_hashes_bare_name_to_stable_urn` → deleted (synthetic-hash contract is gone).
- `test_bare_name_resolved_via_domain_registry` → new, uses a `DomainRegistry` pre-populated with `{"marketing": "urn:li:domain:abc-123"}` (via `__new__` to skip the graph) and asserts the emitted URN is the real one.
- `test_bare_name_without_registry_warns_and_skips` → new, asserts the no-graph path logs a warning and returns `None`.

**Minor — `dremio_recipe.yml` example**

Added a commented-out `source_type_mappings` block with inline comments documenting the Dict-keyed-by-Dremio-type shape and the `database` / `file_object_storage` categories.

**Minor — End-to-end `build_dremio_source_map` test**

Added `test_build_source_map_honors_recipe_overrides_end_to_end` which constructs a `DremioToDataHubSourceTypeMapping` with `extra_mappings={...}` exactly like `DremioAPIOperations` does at runtime, hands it to `build_dremio_source_map`, and asserts the override flows through to both `platform` and `dremio_source_category` for both `database` and `file_object_storage` cases.

Also added `test_build_source_map_default_mapper_marks_unknown_arp_types` as a negative control: without overrides, unknown ARP types collapse to `category='unknown'` + lowercased platform — pinning the exact gap `source_type_mappings` is designed to close.

**Minor — Naming consistency (`dremio_to_datahub_source_mapper` → `source_type_mapper`)**

Renamed `DremioAPIOperations.dremio_to_datahub_source_mapper` to `source_type_mapper`. All call sites updated. Now `DremioAPIOperations.source_type_mapper`, `DremioSource.source_type_mapper`, and `build_dremio_source_map(source_type_mapper=...)` all use the same name — no more cognitive switch when navigating the call chain.

**Minor — `AZURE_STORAGE` dual-membership**

Already addressed in stacked PR #17648 — added an inline comment at the `FILE_OBJECT_STORAGE_TYPES` site documenting the precedence (`get_category` checks DATABASE first → always resolves to "database") plus a `test_azure_storage_resolves_to_database_despite_dual_membership` pin test. After the rebase onto the new #17648 tip, both show up in this PR's diff.

**Local checks**

- `ruff format --check` + `ruff check` on all edited files: clean.
- `mypy` on the 4 changed source files: `Success: no issues found in 4 source files`.
- All 106 Dremio unit tests pass (`pytest tests/unit/dremio/`).

**Re: merge-base / CI**

Base is still `dremio-source-mappings-polaris` (#17648 — stacked). When #17648 lands first, this PR will retarget cleanly to `master`. CI on this branch runs against the actual merge of B5 on B4's tip, so the override changes from #17648 are exercised in this PR's tests too.

acrylJonny:
Thanks for the re-review — addressed the one new minor finding in `d602b7ed1f`.

**Minor — `domain_registry is None` branch in `create_domain_aspect` is dead in production**

You're right. The path is genuinely unreachable: `DremioSource.__init__` builds `DomainRegistry` when `config.domain` is a bare name, and `DomainRegistry.__init__` raises `ValueError` if there's no graph (`domain_registry.py:25-28`). The warn-and-skip fallback was carrying its weight only in a unit test that directly constructed `DremioAspects(domain="marketing", domain_registry=None)` — i.e. testing an impossible production state.

Changes:

- `dremio_aspects.py`: replaced the dead branch with `assert self.domain_registry is not None` to document the invariant and satisfy mypy's narrowing on the next line.
- `test_dremio_domain_aspect.py`: replaced the unreachable-path test with `test_bare_name_without_registry_in_production_path_is_impossible`, which pins the real failure surface — `DomainRegistry(cached_domains=["marketing"], graph=None)` raises `ValueError`.

Verification: `ruff check`, `ruff format`, `mypy` clean on the touched files; full Dremio unit suite (106 tests) green.

metadata-ingestion/tests/unit/dremio/test_dremio_domain_aspect.py
```diff
@@ -0,0 +1,146 @@
...
+    assert len(domain_aspects) == expected_count
+    if expected_count:
+        assert domain_aspects[0].domains == ["urn:li:domain:eng"]
+
+
+def test_add_mapping_removed():
+    # Regression guard: add_mapping mutated class state; replaced by the
+    # per-instance source_type_mappings config field. If reintroduced,
+    # route callers to the config instead.
+    from datahub.ingestion.source.dremio.dremio_datahub_source_mapping import (
```
treff7es: Imports should be at the top if there is no strong reason to do it

treff7es (approved):

metadata-ingestion/tests/unit/dremio/test_dremio_domain_aspect.py
acrylJonny: Addressed in the merged version. All imports — including the previously local `DomainRegistry` import — sit at module top now:

```python
from datahub.ingestion.source.dremio.dremio_datahub_source_mapping import (
    DremioToDataHubSourceTypeMapping,
)
from datahub.ingestion.source.dremio.dremio_entities import (
    DremioDataset,
    DremioDatasetType,
    DremioSpace,
)
from datahub.metadata.schema_classes import DomainsClass
from datahub.utilities.registries.domain_registry import DomainRegistry
```

PR #14890, opened 2025-09-30
feat(ingestion/dremio): stateful incremental ingestion, N+1 query fix, and code quality improvements

acrylJonny:
## Summary

This PR improves the Dremio connector across three areas: stateful ingestion parity with other
connectors, fixing a significant N+1 query pattern in catalog scanning, and code quality cleanup.

### Fix: N+1 SQL queries during catalog scan

Previously, `get_all_tables_and_columns` issued one SQL query per container. For catalogs with
many containers, this meant thousands of Dremio SQL jobs — each incurring Dremio's query planning
overhead and poll-wait cycle. Replaced with a single global query across the entire catalog,
chunked with LIMIT/OFFSET to stay within Dremio's memory limits.

Accompanying changes:
- Replace fixed `sleep(3)` job-polling with adaptive backoff (1 s → 10 s)
- Size the HTTP connection pool to `max_workers` to avoid pool saturation during parallel
  catalog traversal
- Cache `/catalog/{id}` responses in-memory to avoid redundant GETs during tree traversal
- Reuse the dataset list collected during the main pass for profiling, removing a redundant
  second call to `get_all_tables_and_columns()`

### Stateful ingestion

Brings Dremio to parity with Snowflake/BigQuery/Redshift connectors:

- **`RedundantQueriesRunSkipHandler`** — on repeated runs, advances the query lineage start time
  to the previous run's end time so `SYS.JOBS_RECENT` is not re-scanned from scratch each time
- **`ProfilingHandler` + `profile_if_updated_since_days`** — skips re-profiling tables within
  the configured window; Dremio has no table modification timestamp so time-since-last-profile
  is used instead
- **`auto_incremental_lineage`** — emits lineage as PATCH operations, preserving manually-added
  lineage edges in the DataHub UI across runs
- **`auto_incremental_properties`** — emits `DatasetProperties` as PATCH operations, preserving
  descriptions and custom properties edited in the DataHub UI

### Correctness fixes

- **`created` timestamp** — was emitting `{"time": 0}` (epoch/1970) when Dremio provides no
  creation timestamp; now omitted entirely
- **`format_type`** — emitted into `customProperties` when available (Enterprise/Cloud editions)
- **`DremioDataset`** — attributes now initialised to safe defaults before conditional assignment,
  preventing `AttributeError` on Community edition
- **`ProfileConfig`** — now inherits from `GEProfilingConfig` rather than `GEProfilingBaseConfig`,
[...]

PR #17651, opened 2026-05-31
fix(ingestion/dremio): correctness fixes — created timestamp, format_type, Community defaults, ProfileConfig parent

acrylJonny:
> **Part 1 of 3** splitting [#14890](https://github.com/datahub-project/datahub/pull/14890) (the `dremio-enhancements-sep-25` omnibus PR) into reviewable slices. **#14890 will be closed once this stack lands.** Subsequent slices:
>
> - **A2 — performance overhaul + code-quality cleanup** (stacked on this PR)
> - **A3 — stateful ingestion + docs** (stacked on A2)

## Summary

Four correctness bugs in the Dremio connector that surface in production but are silent in unit tests against in-memory mocks.

### Correctness fixes

| Bug | Behavior before | Behavior after |
|---|---|---|
| `created` timestamp | Emitted `{"time": 0}` (epoch / 1970-01-01) when Dremio provides no creation timestamp, polluting the UI with "Created 1970" | Field is omitted entirely |
| `format_type` | Dremio's dataset `format_type` (`PARQUET`, `ICEBERG`, …) was discarded for Enterprise / Cloud editions | Surfaced into `customProperties.format_type` when available |
| `DremioDataset` attribute init | Attributes referenced unconditionally but only assigned inside Enterprise / Cloud branches → `AttributeError` on Community edition | Initialised to safe defaults before the conditional branches |
| `ProfileConfig` parent class | Inherited from `GEProfilingBaseConfig` which omits `profile_if_updated_since_days` and several other knobs Dremio recipes couldn't reach | Now inherits from `GEProfilingConfig`, aligning Dremio with Snowflake / BigQuery / Redshift |

### Supporting changes

- Add `DatasetContainerSubTypes.DREMIO_FOLDER` so folders surface with a distinct subtype rather than reusing `DREMIO_SOURCE`.
- Extract Dremio Pydantic response models and the `DremioEntityContainerType` / `DremioDatasetType` enums into a new `dremio_models.py`. This eliminates an existing circular-import workaround between `dremio_entities` and `dremio_aspects` and lets later slices import them without dragging in the heavier modules.

## Tests

- `test_dremio_aspects.py` — `created` omission, `format_type` passthrough, `customProperties` shape on the dataset aspect.
- `test_dremio_entities.py` — safe-default initialisation on Community-edition responses, parsing round-trip through the new `dremio_models` Pydantic models.

## Behavior changes

- Recipes that previously emitted `created: {time: 0}` will stop doing so — the field is now omitted when Dremio gives no creation date. Downstream consumers must tolerate a missing `created` aspect field (this is already the convention everywhere else in DataHub).
- `format_type` will appear in `customProperties` for Enterprise / Cloud Dremio sources. No-op for Community edition.

## Checklist

- [x] Tests added
- [ ] Docs added — N/A (behavior changes documented in the stacked A3 PR's `updating-datahub.md` once landed)
- [ ] Breaking changes — N/A; the `created: {time: 0}` removal is a bug fix to align with upstream convention


acrylJonny:
Thanks for the detailed pass! Force-pushed `c8fb73be5a` addressing the review.

### Major

- **Major #1 — `root_container_type` dead in A1**: took your suggested route. Added `# TODO(#17652 / A2)` markers at both consumer sites (`DremioFolder.root_container_type` field declaration in `dremio_entities.py` and the dead folder branch in `DremioAspects._create_browse_paths_containers`) plus a comment explaining that A1 alone is still strictly better than master (master emitted no folder browse paths at all). #17652 lands the producer in the recursive catalog walk; both markers get dropped when that slice merges.
- **Major #2 — Forward-scaffolding models**: added explicit "Consumed in #17652 / A2" docstring lines on `DremioContainerResponse` and `DremioJobState`, and "Consumed in #17653 / A3" on `DremioColumnStats` / `DremioProfilingResult`. Reviewers no longer have to grep across the stack to figure out why those classes sit unused in A1.
- **Major #3 — `location_id` required**: relaxed to `Optional[str] = Field(default=None, alias="LOCATION_ID")`. The legacy dict path used `.get("LOCATION_ID", "")` which silently tolerated Community-edition view payloads that omit it; the pydantic port was 500-ing on those. Preserved the `""` contract at the `DremioDataset` assignment site (`self.location_id = self._dataset_response.location_id or ""`) so downstream consumers keep their `str` typing.

### Minor / AI Slop

- **`except (ValueError, AttributeError)`**: dropped `AttributeError` — `datetime.strptime` only raises `ValueError`, and the `if dataset.created:` gate already filters `None`. Existing `test_malformed_created_produces_no_timestamp` covers the realistic failure mode.
- **`.value` on StrEnum comparisons**: dropped from all three `DatasetContainerSubTypes` checks in `_create_browse_paths_containers` (including the two pre-existing ones) — `subclass` is typed as the enum already.
- **URL reference comment**: `# From https://docs.dremio.com/cloud/reference/sql/data-types/` is still present on line 76 — no removal happened. Either a stale view or a paste from a different PR; leaving as-is.
- **`extract_name_from_path_if_missing` "unknown" fallback**: replaced with a `ValueError` so the "I don't have a usable name" case fails loudly instead of forging a synthetic entity. Silent corruption there would propagate into URN building with no signal back to the operator.

### Tests

- New `test_dremio_models.py` (7 tests):
  - `DremioDatasetResponse` minimal required-field round-trip + missing-required-field raises.
  - Community-edition null `LOCATION_ID` accepted (regression guard for Major #3).
  - `DremioContainerResponse` name backfill from list path + string path.
  - Both empty-path and missing-name + missing-path cases now raise (pins the strict fallback).
- 134 unit tests pass, ruff + mypy clean.
- `# TODO(A2)` markers consciously leave the folder-prefix path uncovered in A1 — the producer side ships in #17652 and brings its own coverage there.

### Stack

Cascade-rebased #17652 (A2) and #17653 (A3) onto the new A1 tip. A2 picks up the typed `DremioContainerResponse` consumer (`container.root_container_type` instead of `.get(...)`) and drops the now-stale `# TODO(A2)` markers since both sides land atomically there. A3 cherry-picked clean. All three stacks force-pushed.

acrylJonny:
## Round-2 review response (commit `33de0aa93e`)

### Major #1 — folder browse-path regression ✅ fixed

Root cause: `auto_browse_path_v2` (a framework workunit processor on every `Source`) synthesizes child container browse paths from the `Container` parent aspect — but only when the child does *not* emit its own `BrowsePathsV2`. Master gated emission on `if not container.path:`, so folders fell back to the auto processor and inherited `["Sources", source_urn]` from their parent. A1 removed that guard, so folders emitted a partial (no-prefix) `BrowsePathsV2` that the auto processor then preferred over the parent-inherited path — dropping the root prefix.

Fix: restored `if not container.path:` in `populate_container_mcp` so only top-level (`path=[]`) containers emit `BrowsePathsV2` directly. Also dropped the now-unreachable `DREMIO_FOLDER` branch and partial-path expansion in `_create_browse_paths_containers` (dead under the guard). The `root_container_type` field on `DremioFolder` is kept as forward-compat scaffolding for #17652 (A2), which will reintroduce a folder-aware emission path once the catalog walk stamps the field.

Goldens regenerated locally; Sources/Spaces prefix tallies now match master exactly:

| golden | master Sources/Spaces | A1 (this commit) |
|---|---|---|
| `dremio_mces_golden.json` | 24 / 10 | 24 / 10 |
| `dremio_platform_instance_mces_golden.json` | 24 / 10 | 24 / 10 |
| `dremio_schema_filter_mces_golden.json` | 10 / 0 | 10 / 0 |

### Major #2 — DATA_MANIPULATION operators 🛑 not a real regression

The reviewer note says removing the `DATA_MANIPULATION` bucket changes `query_subtype` for `INSERT INTO` / `MERGE INTO` / `CREATE TABLE`. That's not actually the case — that bucket was unused in master too:

```python
# master dremio_entities.py
def _get_query_subtype(self) -> str:
    for query_operator in (
        QUERY_TYPES["SELECT"] + QUERY_TYPES["DML"] + QUERY_TYPES["DDL"]
    ):
        if self.query_without_comments.upper().startswith(query_operator):
            return query_operator
    return "UNDEFINED"
```

`_get_query_subtype` iterates `SELECT + DML + DDL` only. `DATA_MANIPULATION` is referenced exactly once in master — its own dict definition. The TODO in `_get_affected_tables` references a *different* name (`_data_manipulation_queries`) that doesn't exist. So `INSERT INTO foo` has always matched `"INSERT"` from `DML` first (subtype = `"INSERT"`), `MERGE INTO foo` → `"MERGE"`, `CREATE TABLE foo` → `"CREATE"`. Master and A1 produce identical subtypes.

Added a parametrized regression test (`test_subtype_for_data_manipulation_unchanged_after_bucket_removal`) that pins each of those three queries to its current subtype, so if anyone later wires `DATA_MANIPULATION` back in and changes the behavior the test will fail loudly.

### Minor — `table_schema: str` required 🛑 same risk as master, no change

Master had the same null-throw risk via `self.path = dataset_details.get("TABLE_SCHEMA", "")[1:-1].split(", ")[:-1]` — `None[1:-1]` would `AttributeError`. In practice `TABLE_SCHEMA` is sourced from `INFORMATION_SCHEMA.COLUMNS` and is non-nullable by SQL contract (it's the table's namespace), so neither master nor A1 has ever observed a null in the wild. The `location_id` defensive fix in round 1 was justified because Community-edition *does* return null `LOCATION_ID` for views; `TABLE_SCHEMA` has no equivalent CE quirk. Leaving as-is.

### Minor — `is_nullable` bool → str silently 🛑 fixing a lying annotation

Master's `DremioDatasetColumn` declared `is_nullable: bool` but the runtime value was always a string. The API path was:

```python
# master dremio_api.py — raw API value passed through verbatim
"is_nullable": record["IS_NULLABLE"],   # SQL standard "YES" / "NO" string

# master dremio_entities.py — assigned to the bool field with no coercion
is_nullable=col.get("is_nullable", False),

# master dremio_aspects.py — already compared as a string
nullable=column.is_nullable == "YES",
```

The `bool` annotation was aspirational. A1's `str` annotation makes the type honest. Added `TestDremioDatasetColumnNullability` pinning the actual `== "YES"` emission semantics on YES / NO / default-from-omitted-field paths, so the contract is now backed by tests.

### Minor — `GEProfilingConfig` exposes Snowflake/BigQuery fields 🛑 pre-existing, out of scope

`GEProfilingConfig` is a shared mixin used by every connector with profiling support. The same Snowflake/BigQuery-specific fields are exposed for Redshift, Trino, Athena, Druid, etc. A1 doesn't introduce or expand this — Dremio used to compose the parent `ProfileConfig`, which is exactly the bug A1 fixes (parent had no `profile_table_size_warning_threshold`, `profile_if_updated_since_days`, etc. that the Dremio profiler actually reads). Pruning the unused fields would need to happen at the `GEProfilingConfig` definition site and would affect every connector. Not actionable in A1.

metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_aspects.py
```diff
@@ -1,525 +1,521 @@
...
+        container_class = self._create_container_class(dataset)
+        if container_class:
+            mcp = MetadataChangeProposalWrapper(
+                entityUrn=dataset_urn,
+                aspect=container_class,
+            )
+            yield mcp.as_workunit()
+
+        if dataset.dataset_type == DremioDatasetType.VIEW:
+            view_definition = self._create_view_properties(dataset)
```
treff7es, line 287: This will fail when `self._create_view_properties(dataset)` returns `None`

metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_entities.py
```diff
@@ -272,38 +260,42 @@ def __init__(
 
 
 class DremioContainer:
-    subclass: str = "Dremio Container"
+    subclass: DatasetContainerSubTypes = DatasetContainerSubTypes.DREMIO_FOLDER
```
treff7es: If any new subclass fails to override subclass, it silently emits DREMIO_FOLDER SubType. Consider ClassVar with no default (forcing subclasses to declare it) or use the generic FOLDER.

treff7es (approved):

metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_aspects.py
acrylJonny: Good catch — fixed in 26ca1bc4e2. The call site now guards on the return:

```python
view_definition = self._create_view_properties(dataset)
if view_definition:
    mcp = MetadataChangeProposalWrapper(...)
```

Added `TestDremioAspectsViewProperties` covering both paths.

metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_entities.py
acrylJonny: Done in 26ca1bc4e2. `DremioContainer.subclass` is now `ClassVar[DatasetContainerSubTypes]` with no default, so a subclass that forgets to declare it gets an `AttributeError` at first access (and mypy flags it statically). The three current subclasses (`DremioSourceContainer`, `DremioSpace`, `DremioFolder`) all declare their own subtype explicitly, and `TestDremioContainerSubclassDeclaration` guards the contract.

PR #16163, opened 2026-02-10
feat(security): add TransparentSecretStr for safe credential serialization for ingestions

esteban:
feat(security): add TransparentSecretStr for safe credential serialization for ingestion

Migrating from String to SecretStr fields break serialization boundaries (Kafka, SQS, JSON)
because model_dump() returns SecretStr objects (TypeError in json.dumps) and model_dump_json()
return '**********' (silent data corruption).

TransparentSecretStr serializes as plain str in model_dump()/model_dump_json()
while preserving SecretStr masking in repr/str and secret registry registration.

This PR converts plaintext credential fields to TransparentSecretStr across connectors

- AWS: aws_secret_access_key, aws_session_token
- Azure: account_key, sas_token, client_secret
- Cassandra: password, ssl_certfile
- DBT Cloud: token
- Dremio: password
- Fivetran: api_key, api_secret
- Kafka Connect: password, kafka_api_secret
- Looker: client_id, client_secret (with env var cleanup)
- Neo4j: password
- Okta: okta_api_token
- PowerBI: tenant_id (config), password (report server)
- Qlik Sense: api_key
- Sigma: client_id, client_secret
- Tableau: password
- Unity Catalog: token
- Elasticsearch: password
- GCP Credentials: private_key (with model_dump serialization fix)
- LDAP: ldap_password
- MLflow: password
- MongoDB: password
- NiFi: password, client_key_password
- OpenAPI: password, token, bearer_token
- Preset: api_key, api_secret
- Pulsar: client_secret, token
- Redash: api_key
- Salesforce: password, private_key, security_token, access_token, consumer_key
- Snowflake: private_key, token, encoded_oauth_private_key
- Superset: password
- Unstructured: api_key (partition), api_key (embedding), token (connection)
- Circuit Breaker: datahub_token
- Metadata Change Sync Action: gms_auth_token

PR #12, opened 2015-12-24
port data quality module into wherehows

czbernard:
Add and refactor data quality code into wherehows code base

PR #17652, opened 2026-05-31
perf(ingestion/dremio): N+1 query fix, catalog cache, adaptive polling, code-quality cleanup

acrylJonny:
> **Part 2 of 3** splitting [#14890](https://github.com/datahub-project/datahub/pull/14890). **Stacked on #17651** (correctness fixes). The remaining slice **A3** (stateful ingestion + docs) will be opened against this branch once it lands.

## Summary

This slice carries the bulk of the original #14890: the **N+1 SQL fix** that motivated the PR plus the code-quality cleanup that accompanies it. Correctness fixes were extracted to **#17651**; stateful ingestion will land in **A3**.

## Performance overhaul

- **N+1 → single global query.** `get_all_tables_and_columns(containers)` used to issue one SQL query per Dremio container. For catalogs with thousands of containers and ~10s of Dremio query-planning overhead per call, that was the dominant ingestion-time bottleneck. Replaced with `_get_all_tables_global_chunked` driving three edition-specific SQL templates (`QUERY_DATASETS_{CE,EE,CLOUD}_GLOBAL`).
- **Chunked aggregation accumulates across chunk boundaries.** Per-column metadata is accumulated across the LIMIT/OFFSET stream and only flushed when a different `FULL_TABLE_PATH` appears (the SQL ORDER BYs by `TABLE_SCHEMA, TABLE_NAME, ORDINAL_POSITION`, so within-table rows are contiguous). A wide table whose columns straddle a chunk boundary now emits once with all of its columns instead of twice with partial sets.
- **Catalog response cache.** GET `/catalog/{id}` calls hit the same node IDs 4-6x during traversal + dataset fetch; they now share results.
- **Adaptive job polling.** Fixed `sleep(3)` replaced with backoff (1s → 10s). Short queries don't wait 3s; long queries don't hammer the status endpoint. Wall-clock timeouts in `execute_query` / `execute_query_iter` now cover both the poll and fetch phases via a shared deadline.
- **HTTP pool sizing.** Pool now sized to `max(max_workers, 15)` to eliminate "Connection pool is full" warnings that were serialising parallel catalog traversal.
- **Profiling reuses the main pass.** The dataset list collected during the main walk is reused for profiling, removing a redundant second global catalog query when profiling is enabled.
- **Consolidated job polling.** Duplicate polling logic across `submit_query`, `execute_query`, and the profiling path collapsed into a single `_wait_for_job()`. `_fetch_all_results` now delegates to `_fetch_results_iter` so the pagination + error-handling logic only lives in one place.
- **`get_dataset_id` memoised.** The `/catalog/by-path/...` endpoint is path-keyed and not part of the id-keyed catalog cache; Community-edition ingestion was previously calling this twice per table with the same `(schema, \"\")` key per schema. Results are now cached for the lifetime of the API instance.

## Code-quality cleanup

- **Filter-report wiring restored.** `report_container_scanned` / `report_container_filtered` are now wired at the three `schema_pattern` enforcement sites (source / space / folder) so `report.filtered` and the counters reflect reality again.
- **`profiling.query_timeout` actually used.** Previously the config field was set but ignored; now forwarded to `execute_query` from `_profile_chunk`.
- **`_wait_for_job` tolerates malformed status responses.** Partial JSON / rate-limit HTML responses missing `jobState` no longer crash mid-query with a `KeyError`; the wall-clock timeout is the safety net.
- **`authenticate()` retries only transient errors.** Credential / config errors (`AssertionError`, "no token" / `DremioAPIException`) now fail fast instead of burning all retry attempts.
- **Dead code removal.** Superseded SQL constants (`QUERY_DATASETS_{CE,EE,CLOUD}`, `PROFILE_COLUMNS`) now covered by `_GLOBAL` variants; unused `get_column_stat` helper; empty `report_upstream_latency` placeholder; unused source-type-mapping attribute on `DremioAPIOperations`; the `DATA_MANIPULATION` dict key from `DremioSQLQueries`; the `_get_container_tables_chunked` per-source helper made obsolete by the global query.
- **Dead `HOME → SPACE` ternary removed.** Both branches returned `DremioEntityContainerType.SPACE`; collapsed to the unconditional mapping.
- **`logging.info` → `logger.info`.** `get_tags_for_resource` / `get_description_for_resource` / catalog walk now use the configured logger.
- **Renames.** `QUERY_TYPES` → `_QUERY_OPERATORS` to reflect its actual role (a lookup of leading SQL operators per query category).

## Config additions

| Field | Default | Purpose |
|---|---|---|
| `include_system_tables` | `True` | Whether `INFORMATION_SCHEMA` / `SYS` containers and their datasets are ingested. Matches `include_*` options on other connectors. |
| `usage` | wrapped in `HiddenFromDocs` | Inner fields hidden from generated docs without losing backward compatibility for recipes that still set them. |
| `password` validator | `TransparentSecretStr` | Integrates with the secret-masking pipeline from #16163; plaintext serialisation still works for cross-process emitter pass-through; `repr()` masks. |

## Tests

| Test file | Coverage |
|---|---|
| `test_dremio_chunking.py` (new + updated) | Global-query path, cross-chunk column accumulation for tables that straddle a LIMIT/OFFSET boundary, OOM-warning path |
| `test_dremio_schema_filter.py` (rewritten) | All four scanned/filtered outcomes at every enforcement site (source / space / folder) |
| `test_dremio_system_tables.py` (new) | `include_system_tables` behaviour and pattern dispatch |
| `test_dremio_config.py` (new) | Password validator + `TransparentSecretStr` serialisation semantics |
| `test_dremio_api_pagination.py` (updated) | Global-query path tests, `_wait_for_job` tolerance for responses missing `jobState`, `get_dataset_id` memoisation |
| `test_dremio_timestamp_filtering.py` (updated) | WHERE-clause regression guards for `get_query_all_jobs` and `get_query_all_jobs_cloud` (STATUS, user_name, query_type, queried_datasets size, cloud ARRAY projection) |
| Integration tests | Moved `batch_1` → `batch_4` (test parallelism rebalancing) |
| Goldens | Regenerated to reflect corrected `created` behaviour from #17651 and the new emission paths from this slice |

## Behavior changes

- Ingestion time on multi-thousand-container catalogs drops from hours to a single global query plus walk.
- `include_system_tables` defaults to `True` (no change in behavior); set `False` to skip `INFORMATION_SCHEMA` and `SYS`.
- Secrets in `password` are now masked in `repr()` and logs but still pass-through to downstream emitters.

## Checklist

- [x] Tests added/updated
- [ ] Docs added — N/A (recipe-facing docs land in A3 with the stateful ingestion config; this slice is mostly internal refactoring)
- [ ] Breaking changes — N/A; the new config defaults preserve existing behavior


acrylJonny:
## Review feedback addressed

Pushed `f06c39f1b5` covering everything in the review.

### Blockers
- **#1 cross-chunk column accumulator** — `column_dictionary` / `table_metadata` were declared inside the `while True:` loop and reset every 1,000 rows. Moved them out and added a streaming flush: emit the previous table the moment a different `FULL_TABLE_PATH` appears (the query is `ORDER BY TABLE_SCHEMA, TABLE_NAME, ORDINAL_POSITION` so within-table rows are contiguous), then flush the final in-flight table after the chunk loop. New test `test_get_all_tables_global_chunked_straddling_table` forces `chunk_size=2` against a 3-column `table1` spanning two chunks and confirms it emits once with all three columns.
- **#2 profiling `query_timeout` actually wired** — `_profile_chunk` now passes `self.QUERY_TIMEOUT` to `execute_query`. Updated the PR description to reflect the actual change instead of a claim about an older commit.
- **#3 fetch-phase timeout regression** — `execute_query` / `execute_query_iter` stamp a single `deadline = time() + timeout` and thread it through `_fetch_results_iter`, which now re-checks the wall clock before each page request and cancels the job + raises on overrun. Restores the poll+fetch coverage that the master `ThreadPoolExecutor` wrapper used to give.

### Majors
- **#4 SQL structure regression tests restored** — `test_query_structure_unchanged` and `test_cloud_query_structure_unchanged` are back in `test_dremio_timestamp_filtering.py`, pinning `STATUS = 'COMPLETED'`, `user_name != '\$dremio\$'`, `query_type not like '%INTERNAL%'`, the cloud `ARRAY_SIZE(queried_datasets)>0` / `LENGTH(queried_datasets)>0` size guards, and the cloud-specific `CONCAT('[', ARRAY_TO_STRING(queried_datasets, ','), ']')` projection.
- **#5 second full catalog scan during profiling** — `get_workunits_internal` now materialises the dataset list during the emission loop *only when profiling is enabled* and reuses it for the profile pass, so we run the global catalog query exactly once whether or not profiling is on.
- **#6 `_wait_for_job` `KeyError` guard** — `status.get(\"jobState\")` plus a warning log on responses that don't include the field (partial JSON, rate-limit HTML, transient proxy errors). The wall-clock timeout is the safety net. New test `test_wait_for_job_tolerates_missing_state_field` passes two malformed responses followed by `COMPLETED`.

### Minors
- **#7 dead `HOME → SPACE` ternary** collapsed to the unconditional mapping it always was.
- **#8 `_fetch_all_results` / `_fetch_results_iter` dedup** — `_fetch_all_results` now does `return list(self._fetch_results_iter(...))`; deadline-checking + error-handling live in one place.
- **#9 empty `DremioAPIException()`** — both call sites now `raise DremioAPIException(str(e)) from e` so the underlying message is preserved.
- **#10 `authenticate()` retry burns retries on credential errors** — `AssertionError` and `DremioAPIException` re-raise immediately; only transient errors trigger the retry+sleep loop.
- **#11 `logging.info` → `logger.info`** in `get_tags_for_resource`, `get_description_for_resource`, and the catalog walk.
- **#12 CE `O(tables)` by-path traffic** — `get_dataset_id` is now memoised per `(schema, dataset)` on the API instance. CE ingestion repeatedly resolved the same `(schema, \"\")` key for every table's `LOCATION_ID`; that collapses to one resolution per schema. New test `test_get_dataset_id_memoised` pins this.

### PR description
Removed the \"Made with Cursor\" footer and replaced two claims that were false against the *original* commit set:
- \"Chunked aggregation now spans LIMIT/OFFSET boundaries\" — now actually true thanks to Blocker #1's fix; description updated to describe the new accumulator behaviour.
- \"`profiling.query_timeout` actually used\" — now actually wired thanks to Blocker #2's fix.

### Validation
- `ruff format` + `ruff check`: clean
- `mypy` on `src/datahub/ingestion/source/dremio/`: clean
- `pytest tests/unit/dremio/`: 147 passed (was 143; +1 cross-chunk, +1 `jobState`-guard, +1 memoisation, +1 SQL-structure cloud variant)

### Stack
A3 (#17653) was rebased onto the updated A2 tip — the conflict was the dataset-collection loop, resolved by keeping the conditional-list approach from A2 and dropping the unconditional `datasets_seen` list A3 had introduced for the same purpose. A3 ran `170 passed`, ruff + mypy clean, force-pushed.

### Outside scope (called out for the record)
- `stateful_ingestion` field removal from `DremioSourceConfig` — confirmed inherited from `StatefulIngestionConfigBase`; recipes using `stateful_ingestion:` continue to load.
- `get_pattern_condition` SQL injection — pre-existing, not introduced here. Worth tracking as a separate issue.

acrylJonny:
Round-2 review fixes pushed as `b6ed0805f6`.

## Fixed

### Major — `get_glossary_terms()` third full catalog scan
Confirmed. The harvest is now inline during the main dataset pass — same loop that buffers `datasets_for_profiling`. Each `DremioDataset.__init__` already populates `self.glossary_terms` via `get_tags_for_resource(resource_id=…)`, so the catalog-level iterator was always going to re-run the global `get_all_tables_and_columns()` query just to discover what we'd already loaded. Now we never call `dremio_catalog.get_glossary_terms()` from `get_workunits_internal` at all.

Bonus: switched the dedup key from object identity to `glossary_term.glossary_term` (the string). The legacy `glossary_terms_seen` set was identity-based, but each dataset's `__init__` constructs fresh `DremioGlossaryTerm` instances, so the same tag string from N datasets emitted N times anyway (idempotent at the workunit framework level, but wasted work). String dedup gives us one emission per unique tag.

Regression test (`test_glossary_terms_harvested_from_dataset_pass`):
- Asserts `dremio_catalog.get_glossary_terms` is never invoked.
- Two `Mock` datasets carrying the same tag string (via two distinct `DremioGlossaryTerm` instances) produce exactly one `process_glossary_term` call.

### Minor — `_wait_for_job` poll/fetch deadline arithmetic
Concrete fix: changed `_wait_for_job(job_id, timeout: int)` to `_wait_for_job(job_id, deadline: float)`. `execute_query` / `execute_query_iter` now hand the same absolute `deadline = time() + timeout` to both the poll phase and the fetch phase, so the budget is shared by construction rather than by coincidence.

(In the old code the budgets weren't strictly additive — the poll's internal `start = time()` was set microseconds after the outer `deadline = time() + timeout`, so they aligned in practice. But threading absolute deadline through is a clarity win and removes the redundancy reviewers were flagging.)

Updated `test_execute_query_iter_timeout` to match the simpler one-deadline-check flow (was three `time()` mocks for `[deadline_stamp, wait_for_job_start, timeout_check]`, now two for `[deadline_stamp, deadline_check]`) and the timeout message text.

## Pushing back

### `DremioColumnStats` / `DremioProfilingResult` as dead forward-declared code
Same scaffolding pattern A1 reviewer accepted in round 2 of #17651 (`Pydantic DremioDatasetResponse … is clean`). Both classes carry docstrings explicitly naming `#17653 / A3` as the consumer, and they were introduced in A1's commit `b8d613f` — A2 inherits them via the stacked diff but doesn't add them. Removing them from A2 means deleting from the A1 base then re-adding in A3, which contradicts the stacked-PR review the reviewer already signed off on. Happy to revisit if you want them gated behind a feature flag in A3 instead.

## Verification
- `pytest tests/unit/dremio/` → 148 passed, including the new regression test.
- `ruff check` / `ruff format --check` clean.
- `mypy` clean on `dremio_api.py` + `dremio_source.py`.
- No integration goldens contain `glossaryTerm` entities (the CE integration fixture defines no tags), so no golden regen was required.

metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_api.py
```diff
@@ -1,34 +1,43 @@
...
-from datahub.ingestion.source.dremio.dremio_models import DremioEntityContainerType
 from datahub.ingestion.source.dremio.dremio_reporting import DremioSourceReport
 from datahub.ingestion.source.dremio.dremio_sql_queries import DremioSQLQueries
 from datahub.utilities.perf_timer import PerfTimer
 
+logger = logging.getLogger(__name__)
+
 if TYPE_CHECKING:
-    from datahub.ingestion.source.dremio.dremio_entities import DremioContainer
+    pass
```
treff7es: Why are you doing this?

metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_api.py
```diff
@@ -269,15 +406,65 @@ def _request(self, method: str, url: str, data: Union[str, None] = None) -> Dict
...
+        Responses for /catalog/{id} (no sub-path) are cached in memory because
+        the catalog tree traversal and subsequent dataset fetch both walk the same
+        nodes, producing 4-6 redundant round-trips per catalog entry.
+        """
+        _cacheable = url.startswith("/catalog/") and url.count("/") == 2
+        if _cacheable and url in self._catalog_cache:
+            return self._catalog_cache[url]
+        result = self._request("GET", url)
+        if _cacheable:
+            self._catalog_cache[url] = result
```
treff7es, line 432: The check-then-write is not atomic. get_all_containers runs process_source_and_containers in a ThreadPoolExecutor — two threads can both pass the check and make duplicate network calls. Fix: either use threading.Lock, functools.lru_cache, or document as best-effort (overwrites are idempotent; add a comment).

metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_api.py
```diff
@@ -931,46 +1058,94 @@ def get_all_containers(self):
...
+            if container_type == DremioEntityContainerType.SOURCE.value:
+                source_resp = {}
+                source_config = {}
+                if source.get("id"):
+                    try:
+                        source_resp = self.get(
+                            url=f"/catalog/{source.get('id')}",
+                        )
+                        source_config = source_resp.get("config", {})
+                    except Exception:
```
treff7es: Is it ok to fall back silently?

metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_api.py
acrylJonny: Dropped in 404cdc4ffa — it was leftover scaffolding from a draft that planned to add forward-string-ref imports under `TYPE_CHECKING`. The unused `TYPE_CHECKING` import is gone too.

metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_api.py
acrylJonny: Documented as best-effort in 404cdc4ffa. The race is benign here:

- `/catalog/{id}` is idempotent, so the worst case is one extra HTTP round-trip per racing pair.
- Single-key dict access is atomic under the GIL, so the cache itself won't corrupt.
- Adding a `Lock` would serialise every hot-path GET behind one mutex and erase most of the win the cache is here to deliver.

The new docstring spells this out so future readers don't re-litigate it.

metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_api.py
acrylJonny: Fixed in 404cdc4ffa. The bare `except Exception: pass` is replaced with:

```python
except Exception as exc:
    logger.warning(
        f"Failed to fetch source detail for {…}; falling back to basic catalog info: {exc}"
    )
```

So cross-tenant `/catalog/{id}` failures (and any other underlying error) now surface in the log naming the offending source, while the walk continues with the basic catalog info as before.

treff7es (approved):

PR #17653, opened 2026-05-31
feat(ingestion/dremio): stateful incremental ingestion, incremental lineage/properties, profile-skip

acrylJonny:
> **Part 3 of 3** splitting [#14890](https://github.com/datahub-project/datahub/pull/14890). **Stacked on #17652** (perf overhaul + cleanup), which is itself stacked on **#17651** (correctness fixes). Once this lands, **#14890 will be closed**.

## Summary

Brings the Dremio connector to parity with Snowflake / BigQuery / Redshift on **stateful ingestion**, **incremental lineage**, and **incremental properties**. All three are wired through upstream framework mixins so behavior matches other connectors.

## Stateful ingestion handlers

| Handler | Effect | New config | Default |
|---|---|---|---|
| `RedundantQueriesRunSkipHandler` | On repeated runs, advances query-lineage start time to previous run's end time so `SYS.JOBS_RECENT` history is not re-scanned each run | `enable_stateful_time_window` | `False` (opt-in to preserve existing behavior) |
| `ProfilingHandler` | Skip re-profiling tables within `profile_if_updated_since_days`. Dremio has no table modification timestamp, so this is compared against when DataHub last profiled rather than when the table changed | `profile_if_updated_since_days` (already on `GEProfilingConfig`, now reachable since A1 fixed `ProfileConfig` parent) | unset (always profile) |

## Incremental emitters

| Emitter | Effect | Config | Default |
|---|---|---|---|
| `auto_incremental_lineage` | Lineage aspects emitted as PATCH ops instead of full overwrites — manually-curated lineage edges survive across runs | `incremental_lineage` | `False` (opt-in to preserve existing full-overwrite behavior) |
| `auto_incremental_properties` | `DatasetProperties` emitted as PATCH ops — UI-edited descriptions and `customProperties` survive re-ingestion | `incremental_properties` (via upstream `IncrementalPropertiesConfigMixin`) | mixin default |

## Supporting changes

- `dremio_source.py` imports `DremioDatasetType` from `dremio_models` (where it actually lives after the A1 extraction) so there's one canonical import path.
- Small comment/whitespace cleanup in `DremioSource.__init__`.

## Tests

| Test file | Coverage |
|---|---|
| `test_dremio_stateful.py` | `RedundantQueriesRunSkipHandler` time-window advance + `enable_stateful_time_window` gate |
| `test_dremio_profiler.py` | `ProfilingHandler` skip logic against `profile_if_updated_since_days` (including the `None`-means-always-profile path and the time-window override) |
| `test_dremio_config.py::TestIncrementalLineageWiring` | Behavioral pin that `incremental_lineage` actually re-routes upstream-lineage workunits through the framework's PATCH converter when enabled, and passes through as UPSERT when disabled |

## Docs

- `dremio_pre.md` — new "Stateful Ingestion" section with feature/config reference table
- `dremio_post.md` — new "Protecting UI Edits Between Runs" section covering `incremental_properties` / `incremental_lineage` use cases
- `dremio_recipe.yml` — full stateful-ingestion examples

## Behavior changes for existing deployments

- **`incremental_lineage` defaults to `False`** — opt in by setting `incremental_lineage: true` in the recipe to emit lineage as PATCH ops (preserves manually-curated lineage edges between runs). Existing recipes get no behavior change.
- **`enable_stateful_time_window` defaults to `False`** — opt-in; existing recipes get no behavior change unless they enable it.
- **`profile_if_updated_since_days` is now respected on Dremio** — unset by default, so no behavior change unless configured.

Because all three new knobs default to off, no `updating-datahub.md` entry is needed for this PR — adding a config field with the existing default behavior is new functionality, not a breaking change.

## Checklist

- [x] Tests added
- [x] Docs added (`dremio_pre.md`, `dremio_post.md`, `dremio_recipe.yml`)
- [x] No breaking-change `updating-datahub.md` entry needed (all new behavior is opt-in)

## Closing #14890

Once this stack (A1 → A2 → A3) lands, #14890 will be closed with a redirect comment. The combined diff of A1 + A2 + A3 is identical in spirit to #14890 but split into three reviewable pieces with clear concerns per slice.


acrylJonny:
## Review fixes (commit `2aef72b92e`)

### Blockers
1. **`time.time()` → `time_machine.travel`** in `test_dremio_profiler.py`. All three skip-threshold tests now run against a fixed `_FROZEN_NOW = 2024-01-10 12:00:00 UTC` so the offsets are deterministic on slow CI runners.
2. **`row_count` doubling bug** fixed in `_combine_profile_results`. Each chunk re-queries `COUNT(*)` on the same table, so summing was wrong by a factor of `ceil(columns / 800)`. Now takes `max(...)`. The test that previously asserted the buggy `200` now asserts the correct `100`.

### Major
3. **`enable_stateful_time_window` no-op footgun** — added a `@model_validator` that logs a clear warning if the flag is set without `stateful_ingestion.enabled`. Three new tests pin the behaviour (warns when no-op, silent when disabled, silent when stateful is properly configured).
4. **Memory pressure from buffering full `DremioDataset` objects** — introduced a slim `ProfileTarget` `NamedTuple(dataset_urn, full_table_name, resource_name, columns)`. The emission loop projects to that immediately so we keep the catalog-walk-elimination win without retaining the full entity for 10k+ catalogs. New `TestBuildProfileTarget` test covers the projection.
5. **`incremental_lineage: true` default → `false`** — matches the standard `IncrementalLineageConfigMixin` baseline used by Snowflake/BigQuery and avoids a silent breaking change for self-hosted GMS without patch support. With the new default this is plain new functionality so I also removed the `updating-datahub.md` entry. `dremio_recipe.yml` and `dremio_post.md` updated accordingly. New test pins the default.
6. **Trailing space** in `ge_profiling_config.profile_if_updated_since_days` description removed.

### Minor
- `TestCombineProfileResults::test_single_chunk_passthrough` — kept as a regression guard for the chunk-result merge path; trivial but cheap.
- `DremioDatasetType` import — `dremio_entities` re-exports it from `dremio_models`, so the old `from dremio_entities import DremioDatasetType` path still works (used in `test_dremio_validation.py`).
- `dremio_post.md` reviewed and the `incremental_lineage` paragraph rewritten to match the new default.

### Human-review (`profile_if_updated_since_days` semantic divergence)
Rather than rename the shared field, I documented the Dremio-specific semantic divergence inline in the `GEProfilingConfig` description: Snowflake/BigQuery compare against table modification time; Dremio compares against the DataHub-side last-profile time because Dremio exposes no table-mod timestamp. Renaming would require breaking the shared `GEProfilingConfig` API across all profiling sources.

### Verification
- `ruff check` + `ruff format` clean for `dremio/`, `tests/unit/dremio/`, `ge_profiling_config.py`
- `mypy` clean for the same 12 source files
- 175/175 Dremio unit tests pass locally

acrylJonny:
Round-2 review fixes pushed as `e5eb0d0638`.

## Blockers

### Inline `import logging` in test bodies
Moved to the top of `test_dremio_config.py`.

### `test_incremental_lineage_default_is_false` tested a pydantic default
Replaced with `TestIncrementalLineageWiring` — two behavioral tests that exercise the actual workunit-processor pipeline:

- `test_default_emits_upsert_full_overwrite`: feeds an `UpstreamLineageClass` MCPW through Dremio's processor chain with `incremental_lineage=False` (default). Asserts the workunit comes out as an MCPW with `changeType=UPSERT` (i.e. unchanged — full overwrite semantics preserved).
- `test_flag_enabled_converts_to_patch`: same input with `incremental_lineage=True`. Asserts the framework's `auto_incremental_lineage` has rewritten it into a PATCH MCP targeting the same downstream URN.

This catches regressions to both the default and the wiring itself — if someone removes the `functools.partial(auto_incremental_lineage, ...)` line from `get_workunit_processors`, both tests fail.

## Major — Dead `StatefulLineageConfigMixin` / `StatefulProfilingConfigMixin` inheritance
Confirmed and fixed. `git grep enable_stateful_lineage_ingestion src/datahub/ingestion/source/dremio` and the profiling equivalent both return nothing — those fields are only consumed by Snowflake / BigQuery / Redshift / Dataplex / Snaplogic. Dremio uses:

- `enable_stateful_time_window` + `RedundantQueriesRunSkipHandler` for lineage / usage skipping.
- `profile_if_updated_since_days` + `RedundantProfilingRunSkipHandler` for profile skipping.

Both are wired through `stateful_ingestion` directly without needing the mixin fields. Removed both mixins from `DremioSourceConfig`'s bases and left an inline comment explaining why (to stop a future hand from re-adding them). The two `model_validator(mode="after")` warnings (`"Stateful ingestion is disabled, disabling enable_stateful_*…"`) no longer fire on default Dremio runs.

Pin: `test_no_stateful_lineage_or_profiling_mixin_warnings_on_default_run` constructs `DremioSourceConfig(**base_kwargs)` with `caplog.set_level(WARNING)` and asserts no record mentions either field name.

## Minors

### Self-contradictory comment at `dremio_config.py:222`
Rewrote — the old `# … this flag silently no-ops.` was the original (pre-A3-rev2) intent; once we added the explicit warning the "silently" half stopped being true. New comment: `# Surface a warning when the user has asked for time-window skipping but left stateful_ingestion off — otherwise the flag would no-op invisibly and they'd see no run-to-run skipping.`

### PR description claimed `incremental_lineage` default=True
Updated the PR body — the table entry now reads `False (opt-in to preserve existing full-overwrite behavior)` and the "Behavior changes for existing deployments" section now matches the code default. Also dropped the `updating-datahub.md` framing: since all three new knobs are opt-in (`incremental_lineage=False`, `enable_stateful_time_window=False`, `profile_if_updated_since_days` unset), there's no behavior change for existing recipes, which falls under "new functionality" rather than "breaking change" per the project's `updating-datahub.md` guidance.

## Verification

- `pytest tests/unit/dremio/` → 177 passed (incl. the three new tests).
- `ruff check` / `ruff format --check` clean.
- `mypy` clean on `dremio_config.py`.
- No integration goldens reference the dead mixin fields, so no regen required.

metadata-ingestion/tests/unit/dremio/test_dremio_config.py
```diff
@@ -62,3 +76,164 @@ def test_transparent_secret_str_masked_in_repr(self):
...
+        # checkpointing wiring this test doesn't set up); just compose
+        # the lineage / property handlers we're asserting on.
+        from functools import partial
+
+        from datahub.ingestion.api.incremental_lineage_helper import (
+            auto_incremental_lineage,
+        )
+        from datahub.ingestion.api.incremental_properties_helper import (
+            auto_incremental_properties,
+        )
```
treff7es: Imports should be at the top

treff7es (approved):

metadata-ingestion/tests/unit/dremio/test_dremio_config.py
acrylJonny: Hoisted in d8aa9f5cad. `functools.partial` and the two `auto_incremental_*` helpers are now at module top with the rest of the imports.

PR #17655, opened 2026-06-01
feat(ingestion/teradata): Error categorization with separate counts added, slow query detection added

NehaGslab:
## Summary

Teradata ingestion failures and slow lineage queries were previously hard to diagnose: all view-processing errors were lumped into a single `num_view_processing_failures` counter, and a lineage query against a large `DBC.QryLogV` could run for several minutes with no signal other than the overall run feeling slow. This PR closes those observability gaps by (1) categorizing view-processing errors into actionable buckets with separate counts, and (2) adding slow lineage query detection with per-query DB-time timing.

### What changed

**Error categorization with separate counts**

- Added `_categorize_view_error` + a `ViewErrorCategory` `StrEnum` (`TIMEOUT`, `PERMISSION`, `PARSE`, `UNKNOWN`) that classifies view-processing exceptions using the same signals as the existing retry logic (`PoolTimeoutError`/`TimeoutError`, Teradata error codes, and message keywords). Timeout is checked first so a timed-out query wrapped in an `OperationalError` isn't misclassified.
- Added Teradata-code/keyword patterns for parse errors (`3706`/`3707`, "syntax error", …) and permission/auth errors (`3523`/`3524`/`8017`/`3003`, "access denied", "logon failed", …).
- Extended `TeradataReport` with per-phase, thread-safe counters: `schema_discovery_failures`, `historical_lineage_check_failures`, `view_timeout_errors`, `view_parse_errors`, `view_permission_errors`, `view_unknown_errors`, plus an `increment_view_error` helper that updates both the category counter and the existing `num_view_processing_failures` under a single lock.
- Centralized warning emission in `_warn_view_error`, which looks up actionable guidance from `WARNING_VIEW_ERROR_MAP` (e.g. "Permission denied processing view — check database access rights for the ingestion user.") so operators get a clear next step instead of a raw traceback. Dynamic data (schema/view) goes in `context`.
- Hardened against double-counting: `process_single_view` now documents and enforces a "never re-raise" contract, the `CancelledError` path (stall-timeout cancellation) skips the increment since it was already counted, and hung-view timeouts are counted in the `TIMEOUT` bucket.

**Slow lineage query detection**

- Added `lineage_slow_query_log_seconds` config (default `60`, `0` disables). When the total **DB time** for a single lineage query (the `execute` call plus all `fetchmany` calls — downstream sqlglot processing is excluded) exceeds the threshold, the connector emits a `report.warning(...)` and a `WARNING` log line including the query label, elapsed DB time, and the first 500 chars of the (connector-generated, PII-free) SQL.
- Timing is accumulated per DB operation rather than measured end-to-end, so pipeline/yield latency doesn't skew the number. Timing is recorded in a `finally` block so a query that runs for minutes before failing still surfaces.
- New report fields: `lineage_query_timings` (per-query DB timings for post-run analysis) and `lineage_slow_queries_detected` (count).
- Introduced a `LineageQuery` dataclass (`sql` + human-readable `label`) so each lineage query carries a stable label for timing/reporting.

**Other cleanups**

- Reworked schema-discovery, historical-lineage-check, and view-timeout warnings to the structured `title` / `message` / `context` form, and removed manual `traceback.format_exc()` logging in favor of `exc_info=True`.
- Removed repetitive stalled-view warning spam via a `stall_warned` flag in the watchdog.
- Docs: documented slow lineage query detection in `teradata_post.md`.

### Behavior change

`lineage_slow_query_log_seconds` defaults to `60`, so lineage queries taking longer than 60s of DB time will now emit a `WARNING` (logging only — no change to extracted metadata). Set it to `0` to disable, or raise it if your environment expects long-running lineage queries. Note: if the driver retries a failed `fetchmany`, the retry backoff sleep is included in the measured DB time, so set the threshold well above the expected base query time.

## Test plan

- Unit tests in `tests/unit/test_teradata_source.py`:
  - `_categorize_view_error` correctly classifies timeout, permission, parse, and unknown exceptions (parameterized), including the timeout-first precedence.
  - View-processing failures increment both `num_view_processing_failures` and the correct per-category counter, with no double-counting across the single-threaded, thread-pool, `CancelledError`, and hung-view-timeout paths.
  - Slow lineage query detection: queries over the threshold increment `lineage_slow_queries_detected`, record `lineage_query_timings`, and emit a warning; timings are still recorded when a query fails; `lineage_slow_query_log_seconds: 0` disables detection.

## Checklist

- [x] Tests added/updated
- [x] Docs added/updated (`teradata_post.md`)
- [ ] Breaking changes documented in `docs/how/updating-datahub.md` — N/A (logging-only behavior change, no metadata impact)


metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -262,6 +264,93 @@
...
+
+
+class ViewErrorCategory(StrEnum):
+    TIMEOUT = "timeout"
+    PERMISSION = "permission"
+    PARSE = "parse"
+    UNKNOWN = "unknown"
+
+
+def _categorize_view_error(exc: BaseException) -> ViewErrorCategory:
```
treff7es, line 329: **🔴 Blocker (#8): classify by numeric error code first; keep keywords only as a code-less fallback.**

The keyword lists (`_PARSE_ERROR_KEYWORDS`, `_PERMISSION_ERROR_KEYWORDS`, the `"timeout"`/`"timed out"` checks) match anywhere in the stringified exception — including SQL text, object names, or a wrapped inner message — so a view whose SQL or error context merely *contains* "access denied", "cursor", or "timeout" is miscategorized, and that drives the wrong operator guidance from `WARNING_VIEW_ERROR_MAP`. The delicate "timeout-first ordering" is itself a symptom.

The Teradata `[Error NNNN]` code is authoritative. Classify by code first, and keep the keyword/`isinstance` checks **as a fallback for code-less errors only** — do not delete `_PERMISSION_ERROR_KEYWORDS`: auth failures can surface without a bracketed code (e.g. a driver-level "logon failed" before a session, or a wrapped error where the `[Error 8017/3003]` tag didn't propagate), and dropping the permission fallback would regress those to `UNKNOWN`.

```python
_ERROR_CODE_RE = re.compile(r"\[Error (\d+)\]")
_PARSE_CODES = frozenset({3706, 3707})
_PERMISSION_CODES = frozenset({3523, 3524, 8017, 3003})

def _error_code(exc):
    m = _ERROR_CODE_RE.search(str(exc))
    return int(m.group(1)) if m else None

def _categorize_view_error(exc):
    if isinstance(exc, (PoolTimeoutError, TimeoutError)):
        return ViewErrorCategory.TIMEOUT

    code = _error_code(exc)
    if code in _PERMISSION_CODES:
        return ViewErrorCategory.PERMISSION
    if code in _PARSE_CODES:
        return ViewErrorCategory.PARSE

    # Fallback only when the driver omitted the [Error N] tag.
    # Keep timeout-first precedence: a code-less message can carry both
    # "timed out" and "permission denied".
    msg = str(exc).lower()
    if "timed out" in msg or "timeout" in msg:
        return ViewErrorCategory.TIMEOUT
    if any(k in msg for k in _PERMISSION_ERROR_KEYWORDS):
        return ViewErrorCategory.PERMISSION
    if any(k in msg for k in _PARSE_ERROR_KEYWORDS) or isinstance(exc, NotSupportedError):
        return ViewErrorCategory.PARSE
    return ViewErrorCategory.UNKNOWN
```

Note: `teradatasql` (verified v20.0.0.52) exposes no structured code attribute — the code only lives in the message string — so a regex stays, but it now extracts the authoritative numeric code rather than guessing from free text, with keywords demoted to the code-less safety net. The existing `_PERMANENT_ERROR_CODE_RE`/retryable-code regexes could adopt the same extract-int + `frozenset` pattern. Please add a test where a parse `[Error 3706]` whose detail text contains "timeout" still classifies as PARSE (not TIMEOUT).

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -2641,46 +2868,98 @@ def _mark_phase(phase: str, query_index: int = 0) -> None:
...
-                        f"Completed query {query_index}: {query_total_count} lineage entries in {batch_count} batches"
-                    )
+                        if (
+                            self.config.lineage_slow_query_log_seconds > 0
+                            and query_db_elapsed
+                            > self.config.lineage_slow_query_log_seconds
+                        ):
+                            self.report.increment_lineage_slow_query()
+                            self.report.warning(
+                                title="Slow lineage query detected",
```
treff7es: **🔴 Blocker (#1): thread-safety — wrap this `report.warning()` in `self.report.atomic()`.**

Every other cross-thread `report.warning()` in this file takes the report lock (`_warn_view_error`, the timeout warning, and the watchdog warning just above). `report.warning` mutates an unsynchronized `LossyDict`/`LossyList`. This slow-query warning runs on the main lineage thread while the `teradata-lineage-watchdog` daemon thread is concurrently calling `report.warning` under `atomic()` — so the two can corrupt the shared warnings structure. `increment_lineage_slow_query()`/`record_lineage_query_timing()` are already lock-guarded; only this `report.warning(...)` needs wrapping:

```python
with self.report.atomic():
    self.report.warning(title="Slow lineage query detected", ...)
```

(The `logger.warning` below is fine — stdlib logging is thread-safe.)

Also (#3, non-blocking): because `yield from batch` is inside the `try`, this `finally` also runs on `GeneratorExit` if the consumer abandons the generator, emitting a slow-query warning for a query the user cut short. Consider moving the threshold check to the normal-completion path and keeping only `record_lineage_query_timing` here.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -2207,13 +2389,18 @@ def process_single_view(
...
                                     context=(
                                         f"{schema}.{name} did not complete within "
                                         f"{per_view_timeout}s (ran for "
                                         f"{elapsed:.0f}s)"
                                     ),
                                 )
+                            # Count hung views in the same bucket as other timeouts
+                            # so num_view_processing_failures + view_timeout_errors
+                            # give a complete picture alongside num_view_processing_timeouts.
+                            self.report.increment_view_error(ViewErrorCategory.TIMEOUT)
```
treff7es: **🔴 Blocker (#2): hung-view abandon path can double-count `num_view_processing_failures`.**

This increments `increment_view_error(TIMEOUT)` (which also bumps `num_view_processing_failures`) and then calls `fut.cancel()`. But `ThreadPoolExecutor` can't cancel an already-running worker — `cancel()` returns `False` — and the future is then discarded from `remaining_futures`. The worker keeps running `process_single_view`; if its blocking I/O later raises, the worker's own handler calls `increment_view_error(...)` **again**, so one hung-then-failed view is counted twice (TIMEOUT + its real category). The CONTRACT docstring only reasons about the `fut.result()` re-raise path, not about the worker continuing after abandonment.

Pick a single ownership rule and document it — e.g. rely on `num_view_processing_timeouts` + `stalled_views` for hung views and do **not** also call `increment_view_error` here, or track abandoned futures and have the worker's handler skip counting for them. The current `test_hung_view_increments_timeout_counters_consistently` uses a worker that returns cleanly, so this race isn't exercised.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -2257,6 +2444,31 @@ def process_single_view(
...
+        exception context.  Dynamic data (schema, view name) is placed in
+        ``context``.
+        """
+        message = self.WARNING_VIEW_ERROR_MAP.get(
+            error_category, self.WARNING_VIEW_ERROR_MAP[ViewErrorCategory.UNKNOWN]
+        )
+        with self.report.atomic():
+            self.report.warning(
+                title="View processing error",
+                message=message,  # type: ignore[arg-type]  # all dict values are compile-time literals
```
treff7es: **🟡 (#7): this `# type: ignore[arg-type]` is removable.** `Source.warning` requires `message: LiteralString`; the map is declared `Dict[ViewErrorCategory, str]`, so `.get(...)` returns `str`. Declare the values as `LiteralString` (already importable from `typing_extensions`, already imported in this file):

```python
WARNING_VIEW_ERROR_MAP: ClassVar[Dict[ViewErrorCategory, LiteralString]] = { ... }
```

mypy then verifies each value is a literal at definition time and the ignore can be deleted — turning a suppression into an enforced guarantee.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -1259,6 +1395,29 @@ def _validate_column_extraction_options(self) -> "TeradataConfig":
...
+        ),
+    )
+
+
+@dataclass
+class LineageQuery:
+    """A SQL query together with a short human-readable label used for timing reports."""
+
+    sql: str
+    label: str
```
treff7es: **🟡 (#6): `label: str` is too loose.** Only two values ever exist (`"current_only"` / `"historical_union"`) and the label flows into `lineage_query_timings` dict keys. Tighten to a `Literal` so a typo can't drift into report keys:

```python
LineageQueryLabel = Literal["current_only", "historical_union"]

@dataclass
class LineageQuery:
    sql: str
    label: LineageQueryLabel
```

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Enhance error classification by classifying by numeric error code first, kept keywords only as a code-less fallback. Updated test cases for same.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Added slow query reporting inside try block, added report.warning inside atomic block

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Refined ownership logic for abandoned views to prevent double-counting in error metrics

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Updated type hint for WARNING_VIEW_ERROR_MAP to use LiteralString for improved type safety

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Updated LineageQuery label type to use Literal for enhanced type safety

treff7es (approved):

PR #17656, opened 2026-06-01
fix(ingest): prevent DEBUG log explosion from per-aspect curl logging

aviraj-gour:
## Summary

Enabling `DATAHUB_DEBUG=1` could balloon ingestion logs from MBs to GBs. The root cause was in `DataHubRestEmitter._emit_generic`: for **every** aspect on the **success** path it eagerly built a full `make_curl_command` string (URL + all session headers + the entire serialized payload) and logged it at `DEBUG`. On a large ingestion that is one large multi-line log record per aspect, plus the CPU cost of constructing it.

This PR makes the curl-equivalent string lazy — it is now built **only inside the failure handlers**, where it is genuinely useful for debugging — and takes the opportunity to make the surrounding emit-path logging consistent and actually useful for tracing a write.

## Problem

`_emit_generic` ran this unconditionally, per aspect, at `DEBUG`:

```python
curl_command = make_curl_command(self._session, method, url, payload)
logger.debug(
    "Attempting to emit aspect (size: %s) to DataHub GMS; using curl equivalent to:\n%s",
    payload_size, curl_command,
)
```

Result: `DATAHUB_DEBUG=1` became impractical for real troubleshooting — the signal was buried under per-aspect curl dumps, and the log volume alone made it costly.

### Core fix — `_emit_generic`

- Removed the unconditional success-path `make_curl_command` construction and log.
- The curl-equivalent is now built **lazily, only in the `except` handlers**, so the happy path pays neither the CPU nor the log-volume cost.
- When DATAHUB_EMITTER_TRACE is enabled: make_curl_command output is always emitted at TRACE level, regardless of success or failure. This gives a dedicated, opt-in escape hatch for deep per-request debugging without polluting DEBUG logs during normal operation.

### Added post-flight success logging

These paths previously logged nothing on success; each now emits a single concise `DEBUG` line after a successful send:

- `emit_mce` → `Sent MCE successfully urn=… status=…`.
- `emit_mcp` → one `Sent MCP successfully urn=… aspect=… status=… trace_id=…` line shared by both the openapi and restli branches (guarded by `if response is not None`).
- `emit_usage` → `Sent usage stats successfully resource=… bucket=… status=…`.
- For batch emission paths (`_emit_openapi_mcps`, `_emit_restli_mcps`), log a single `DEBUG` line per chunk in the format `Sent MCP batch chunk=<i>/<total> items=<N> status=<status> trace_id=<trace_id> urns=[(<urn>, <aspect>), ...]`, where `chunk=<i>/<total>` helps distinguish split batches (especially in sync mode where logs may otherwise look identical), with OpenAPI computing `total` globally across all `(method, url)` groups to remain accurate, and `urns` enumerating each MCP’s `(entityUrn, aspectName)` so you can precisely verify what was sent and easily grep by URN.

For synchronous paths (MCE, usage, restli) a `2xx` is the commit proof; MCP paths additionally carry `trace_id` for async write correlation.

### `_await_status`

- Added an entry log `awaiting async write trace_id=… urns=N` at the start of verifying each trace, giving visibility into the async write-verification wait (this loop only runs in `ASYNC_WAIT` mode). `urns` is the count of URNs awaiting verification — deliberately not labeled `pending`, since nothing has been polled yet and no GMS write status is known at that point.

## Behavioral impact

- `DATAHUB_DEBUG=1` no longer emits `make_curl_command` output on successful emits. On failures, the curl-equivalent is still logged (at `DEBUG`) — that's where it's actually needed, and failures are rare so volume is a non-issue.
- New success lines are one short line per emit / per chunk at `DEBUG`.
- No public API or return-type changes.

### DIFF
--- a/metadata-ingestion/src/datahub/emitter/rest_emitter.py
+++ b/metadata-ingestion/src/datahub/emitter/rest_emitter.py
@@ -835,15 +835,6 @@ class DataHubRestEmitter(Closeable, Emitter):
             )
 
         if trace_data:
-            if _DATAHUB_EMITTER_TRACE:
-                try:
-                    logger.info(
-                        f"MCP trace_id={trace_data.trace_id} "
-                        f"timestamp={trace_data.extract_timestamp()} "
-                        f"urn={mcp.entityUrn} aspect={mcp.aspectName}"
-                    )
-                except Exception as e:
-                    logger.debug(f"Failed to log trace data: {e}")
             if self._should_trace(emit_mode):
                 self._await_status(
                     [trace_data],
@@ -947,15 +938,6 @@ class DataHubRestEmitter(Closeable, Emitter):
                     data.trace_id if data else None,
                 )
                 if data is not None:
-                    if _DATAHUB_EMITTER_TRACE:
-                        try:
-                            logger.info(
-                                f"MCP batch trace_id={data.trace_id} "
-                                f"timestamp={data.extract_timestamp()} "
-                                f"urns={list(data.data.keys())}"
-                            )
-                        except Exception as e:
-                            logger.debug(f"Failed to log trace data: {e}")
                     trace_data.append(data)
 
         if trace_data and self._should_trace(emit_mode):
@@ -1042,15 +1024,6 @@ class DataHubRestEmitter(Closeable, Emitter):
                 data.trace_id if data else None,
             )
             if data is not None:
-                if _DATAHUB_EMITTER_TRACE:
-                    try:
-                        logger.info(
-                            f"MCP batch trace_id={data.trace_id} "
-                            f"timestamp={data.extract_timestamp()} "
-                            f"urns={list(data.data.keys())}"
-                        )
-                    except Exception as e:
-                        logger.debug(f"Failed to log trace data: {e}")
                 trace_data.append(data)
 
         return trace_data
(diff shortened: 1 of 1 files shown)


metadata-ingestion/src/datahub/emitter/rest_emitter.py
```diff
@@ -441,15 +441,21 @@ def _get_user_agent_string(self, session: requests.Session) -> str:
 class _Chunk:
     items: List[str]
     total_bytes: int = 0
+    # (urn, aspect) for each item, kept parallel to `items` so the batch emit
```
sgomezvillamor: Something like this would improve readability, WDYT?


```
@dataclass
class _ChunkItem:
    """A single serialized item in a batch chunk."""
    payload: str
    urn: Optional[str] = None
    aspect_name: Optional[str] = None

    @property
    def byte_size(self) -> int:
        return len(self.payload.encode())


@dataclass
class _Chunk:
    items: List[_ChunkItem] = field(default_factory=list)
    total_bytes: int = 0

    def add_item(
        self,
        payload: str,
        urn: Optional[str] = None,
        aspect_name: Optional[str] = None,
    ) -> None:
        item = _ChunkItem(payload=payload, urn=urn, aspect_name=aspect_name)
        self.items.append(item)
        self.total_bytes += item.byte_size
```

metadata-ingestion/src/datahub/emitter/rest_emitter.py
```diff
@@ -819,16 +831,16 @@ def emit_mcp(
...
+                mcp.entityUrn,
+                mcp.aspectName,
+                response.status_code,
+                trace_data.trace_id if trace_data else None,
+            )
+
         if trace_data:
-            if _DATAHUB_EMITTER_TRACE:
-                try:
-                    logger.info(
```
sgomezvillamor, line 825: IMO we should keep this log info when _DATAHUB_EMITTER_TRACE=true

metadata-ingestion/src/datahub/emitter/rest_emitter.py
```diff
@@ -917,16 +934,17 @@ def _emit_openapi_mcps(
...
+                    method,
+                    len(chunk.items),
+                    response.status_code,
+                    data.trace_id if data else None,
+                    chunk.urn_aspects,
+                )
                 if data is not None:
-                    if _DATAHUB_EMITTER_TRACE:
-                        try:
-                            logger.info(
```
sgomezvillamor, line 923: IMO we should keep this log info when _DATAHUB_EMITTER_TRACE=true

sgomezvillamor (commented):
Overall, this LGTM
* When DATAHUB_EMITTER_TRACE flag is enabled, we always emit the full curl command regardless of success or failure.
* When DATAHUB_EMITTER_TRACE flag is not enabled, curl command only emits if failure.

I'm just concerned about some logger.info traces that may have been lost (or not tracked as info) when DATAHUB_EMITTER_TRACE flag is enabled.

metadata-ingestion/src/datahub/emitter/rest_emitter.py
aviraj-gour: We can keep this, adding back.

metadata-ingestion/src/datahub/emitter/rest_emitter.py
aviraj-gour: We can keep this, adding back.

metadata-ingestion/src/datahub/emitter/rest_emitter.py
aviraj-gour: This looks much better, Implemented.
Thanks for Suggestion

sgomezvillamor (approved):
LGTM

PR #17660, opened 2026-06-01
fix(vertexai): gracefully skip projects with no Vertex AI Metadata Store

eylulyr:
**Description**
Listing experiments fails with 404 Requested Metadata Store default not found for any project that has never used Vertex AI Experiments / Pipelines / ML Metadata. The default Metadata Store is created lazily the first time one of those features is used in a given project + region — so a 404 is the normal, expected state for projects that only run training jobs or use the Model Registry. Today this aborts the whole project's experiment extraction.

**Changes**
Catch google.api_core.exceptions.NotFound in VertexAIExperimentExtractor._list_experiments_rate_limite, return an empty list so the rest of the project's metadata (models, training jobs, endpoints, …) still ingests, and surface the skip via self.report.warning(...) so it appears in the structured run summary instead of only in logs.
Plumb SourceReport into VertexAIExperimentExtractor via its constructor.

**Testing**
Unit test in tests/unit/vertexai/test_vertexai_source.py asserts that a NotFound from the SDK results in (1) an empty experiment list and (2) a structured warning on SourceReport with title "Vertex AI Metadata Store not found". All existing Vertex AI unit tests continue to pass.


alokr-dhub:
LGTM - approving

alokr-dhub (approved):

PR #17668, opened 2026-06-01
fix(search): prevent result truncation when skipping invalid-URN hits [CAT-2018]

### LINKED ISSUE #13181: Failed to load ingestion sources
**Describe the bug**
After updating to the recent 1.0.0 version, ingestion sources do not work anymore.

**To Reproduce**
Go to "Admin / Data Sources" and recognize a red error message at the top and an empty list of sources.

**Expected behavior**
Both existing data sources and newly added ones should appear.

**Additional context**
The following errors show in container logs:

`
2025-04-11 06:58:26,277 [ForkJoinPool.commonPool-worker-1779] ERROR c.l.m.s.e.query.ESSearchDAO:165 - Search query failed
java.lang.NullPointerException: null
2025-04-11 06:58:26,277 [ForkJoinPool.commonPool-worker-1779] ERROR c.l.d.g.e.DataHubDataFetcherExceptionHandler:45 - Failed to execute
java.util.concurrent.CompletionException: java.lang.RuntimeException: Failed to list ingestion sources
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:315)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:320)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1770)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1760)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:373)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1182)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1655)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1622)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:165)
Caused by: java.lang.RuntimeException: Failed to list ingestion sources
	at com.linkedin.datahub.graphql.resolvers.ingest.source.ListIngestionSourcesResolver.lambda$get$1(ListIngestionSourcesResolver.java:119)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1768)
	... 6 common frames omitted
Caused by: com.datahub.util.exception.ESQueryException: Search query failed:
	at com.linkedin.metadata.search.elasticsearch.query.ESSearchDAO.lambda$executeAndExtract$1(ESSearchDAO.java:166)
	at io.datahubproject.metadata.context.TraceContext.withSpan(TraceContext.java:110)
	at io.datahubproject.metadata.context.OperationContext.withSpan(OperationContext.java:391)
	at com.linkedin.metadata.search.elasticsearch.query.ESSearchDAO.executeAndExtract(ESSearchDAO.java:147)
	at com.linkedin.metadata.search.elasticsearch.query.ESSearchDAO.search(ESSearchDAO.java:338)
	at com.linkedin.metadata.search.elasticsearch.ElasticSearchService.search(ElasticSearchService.java:173)
	at com.linkedin.metadata.search.elasticsearch.ElasticSearchService.search(ElasticSearchService.java:155)
	at com.linkedin.metadata.client.JavaEntityClient.search(JavaEntityClient.java:458)
	at com.linkedin.datahub.graphql.resolvers.ingest.source.ListIngestionSourcesResolver.lambda$get$1(ListIngestionSourcesResolver.java:83)
	... 7 common frames omitted
Caused by: java.lang.NullPointerException: null
2025-04-11 06:58:26,278 [ForkJoinPool.commonPool-worker-1773] ERROR c.datahub.graphql.GraphQLController:153 - Errors while executing query: query listIngestionSources($input: ListIngestionSourcesInput!) {
  listIngestionSources(input: $input) {
    start
    count
    total
    ingestionSources {
      urn
      name
      type
      config {
        recipe
        version
        executorId
        debugMode
        extraArgs {
          key
          value
          __typename
        }
        __typename
      }
      schedule {
        interval
        timezone
        __typename
      }
      platform {
        urn
        __typename
  ..., result: {errors=[{message=An unknown error occurred., locations=[{line=2, column=3}], path=[listIngestionSources], extensions={code=500, type=SERVER_ERROR, classification=DataFetchingException}}], data={listIngestionSources=null}, extensions={tracing={version=1, startTime=2025-04-11T06:58:26.265025322Z, endTime=2025-04-11T06:58:26.278222119Z, duration=13199038, parsing={startOffset=622316, duration=594755}, validation={startOffset=1183528, duration=514351}, execution={resolvers=[{path=[listIngestionSources], parentType=Query, returnType=ListIngestionSourcesResult, fieldName=listIngestionSources, startOffset=1879409, duration=10967547}]}}}}, errors: [DataHubGraphQLError{path=[listIngestionSources], code=SERVER_ERROR, locations=[SourceLocation{line=2, column=3}]}]
`

alexjst:
## Summary

Follow-up to #17387 (which fixed #13181). That PR stopped search from crashing on documents with a missing/invalid `urn` — e.g. the `datahub-gc` bootstrap document persisted in v1.0.0 — by skipping those hits. It resolved the crash but introduced three problems, fixed here.

### 1. Silent result truncation in batched/cached search

`CacheableSearcher.getSearchResults()` ended pagination on any batch smaller than `batchSize`. Since skipped hits shrink a batch, a single invalid-URN document in a full, non-final batch made the loop stop early and silently drop every subsequent batch (and `numEntities` then disagreed with the entities actually returned). The loop now ends only on an empty batch, at the cost of at most one extra empty fetch on a full scan.

`numEntities` (total hits) is intentionally not used as the stop signal: `track_total_hits` is disabled/capped on several paths (KNN, etc.), so it is not a reliable count.

### 2. Over-broad exception handling

The skip caught bare `Exception`, so unrelated failures inside `getResult` (e.g. in `extractMatchedFields`/`extractFeatures`) were silently swallowed — dropping valid results and masking real bugs. A dedicated `InvalidSearchHitException` is now thrown by `UrnExtractionUtils.extractUrnFromSearchHit` and is the only thing caught; any other exception propagates.

### 3. Undiagnosable warn log

The skip logged only `e.getMessage()` with no throwable, so the `NullPointerException` case this code exists to handle produced `Error: null` with no stacktrace. It now logs the exception itself and increments a `skippedInvalidSearchHit` metric.

The two duplicated skip blocks (scroll path and standard path) are consolidated into one `getResultSafely` helper.

## Supersedes #17494

#17494 added unit tests for the skip behavior but no production fix. This PR includes equivalent and broader coverage, so #17494 is being closed in favor of it.

## Test coverage

- `SearchRequestHandlerTest` — both search paths skip invalid-URN hits while returning valid ones; all-invalid response returns empty without crashing.
- `CacheableSearcherTest` — a skipped hit in a full batch no longer truncates pagination (this test fails against the previous behavior and passes here).
- `UrnExtractionUtilsTest` — updated to assert the dedicated exception type.

### DIFF
--- a/metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/query/request/SearchRequestHandler.java
+++ b/metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/query/request/SearchRequestHandler.java
@@ -511,7 +511,14 @@ public class SearchRequestHandler extends BaseRequestHandler {
 
   @Nonnull
   private List<MatchedField> extractMatchedFields(@Nonnull SearchHit hit) {
+    // getHighlightFields() and getMatchedQueries() can be null for a valid hit (no highlight / no
+    // named-query match). Default them to empty: now that getResultSafely only catches
+    // InvalidSearchHitException, an unguarded NPE here would fail the whole search instead of
+    // skipping a single bad hit.
     Map<String, HighlightField> highlightedFields = hit.getHighlightFields();
+    if (highlightedFields == null) {
+      highlightedFields = Map.of();
+    }
     // Keep track of unique field values that matched for a given field name
     Map<String, Set<String>> highlightedFieldNamesAndValues = new HashMap<>();
     for (Map.Entry<String, HighlightField> entry : highlightedFields.entrySet()) {
@@ -528,7 +535,9 @@ public class SearchRequestHandler extends BaseRequestHandler {
       }
     }
     // fallback matched query, non-analyzed field
-    for (String queryName : hit.getMatchedQueries()) {
+    String[] matchedQueries =
+        hit.getMatchedQueries() != null ? hit.getMatchedQueries() : new String[0];
+    for (String queryName : matchedQueries) {
       if (!highlightedFieldNamesAndValues.containsKey(queryName)) {
         if (hit.getFields().containsKey(queryName)) {
           for (Object fieldValue : hit.getFields().get(queryName).getValues()) {
--- a/metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/query/request/SearchRequestHandler.java
+++ b/metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/query/request/SearchRequestHandler.java
@@ -36,6 +36,7 @@ import com.linkedin.metadata.search.elasticsearch.query.filter.QueryFilterRewrit
 import com.linkedin.metadata.search.features.Features;
 import com.linkedin.metadata.search.utils.ESAccessControlUtil;
 import com.linkedin.metadata.search.utils.ESUtils;
+import com.linkedin.metadata.search.utils.InvalidSearchHitException;
 import com.linkedin.metadata.search.utils.UrnExtractionUtils;
 import com.linkedin.util.Pair;
 import io.datahubproject.metadata.context.OperationContext;
@@ -458,17 +459,11 @@ public class SearchRequestHandler extends BaseRequestHandler {
     List<SearchEntity> results = new ArrayList<>(searchHits.length);
     for (SearchHit hit : searchHits) {
       // Build base SearchEntity — skip hits with missing/invalid URN rather than crashing
-      SearchEntity entity;
-      try {
-        entity = getResult(hit);
-      } catch (Exception e) {
-        log.warn(
-            "Skipping scroll hit with invalid or missing URN. Index: {}, ID: {}. Error: {}",
-            hit.getIndex(),
-            hit.getId(),
-            e.getMessage());
+      Optional<SearchEntity> maybeEntity = getResultSafely(opContext, hit);
+      if (maybeEntity.isEmpty()) {
         continue;
       }
+      SearchEntity entity = maybeEntity.get();
       // Compute per-hit scrollId using this hit's sort values
       Object[] sort = hit.getSortValues();
       String perHitScrollId =
@@ -617,6 +612,34 @@ public class SearchRequestHandler extends BaseRequestHandler {
   }
 
   /**
+   * Builds a {@link SearchEntity} for a hit, returning empty (and skipping the hit) only when its
+   * URN is missing or invalid — e.g. documents created by older bootstrap code (see issue #13181).
+   *
+   * <p>Any other failure is left to propagate: silently swallowing it would mask real bugs and
+   * could drop valid results without surfacing the cause. The narrow {@link
+   * InvalidSearchHitException} catch ensures only the known, recoverable data-quality condition is
+   * tolerated.
+   */
+  private Optional<SearchEntity> getResultSafely(
+      @Nonnull OperationContext opContext, @Nonnull SearchHit hit) {
+    try {
+      return Optional.of(getResult(hit));
+    } catch (InvalidSearchHitException e) {
+      log.warn(
+          "Skipping search hit with invalid or missing URN. Index: {}, ID: {}",
+          hit.getIndex(),
+          hit.getId(),
+          e);
+      opContext
+          .getMetricUtils()
+          .ifPresent(
+              metricUtils ->
+                  metricUtils.increment(SearchRequestHandler.class, "skippedInvalidSearchHit", 1));
+      return Optional.empty();
+    }
+  }
+
+  /**
    * Gets list of entities returned in the search response, skipping any hits with missing or
    * invalid URN fields (e.g. documents created by older bootstrap code) instead of crashing the
    * entire search operation.
@@ -629,19 +652,7 @@ public class SearchRequestHandler extends BaseRequestHandler {
       @Nonnull OperationContext opContext, @Nonnull SearchResponse searchResponse) {
     List<SearchEntity> results =
         Arrays.stream(searchResponse.getHits().getHits())
-            .flatMap(
-                hit -> {
-                  try {
-                    return Stream.of(getResult(hit));
-                  } catch (Exception e) {
-                    log.warn(
-                        "Skipping search hit with invalid or missing URN. Index: {}, ID: {}. Error: {}",
-                        hit.getIndex(),
-                        hit.getId(),
-                        e.getMessage());
-                    return Stream.empty();
-                  }
-                })
+            .flatMap(hit -> getResultSafely(opContext, hit).stream())
             .collect(Collectors.toList());
     return ESAccessControlUtil.restrictSearchResult(opContext, results);
   }
--- a/metadata-io/src/main/java/com/linkedin/metadata/search/cache/CacheableSearcher.java
+++ b/metadata-io/src/main/java/com/linkedin/metadata/search/cache/CacheableSearcher.java
@@ -55,14 +55,7 @@ public class CacheableSearcher<K> {
           do {
             batchedResult = getBatch(opContext, batchId);
             int currentBatchSize = batchedResult.getEntities().size();
-            // An empty batch means we've reached the end of the results.
-            //
-            // NOTE: we must NOT stop just because currentBatchSize < batchSize. Hits with
-            // invalid/missing URNs are dropped before reaching this point (see
-            // SearchRequestHandler#getResultSafely), so a batch can legitimately yield fewer
-            // entities than batchSize while more results remain in subsequent batches. Treating a
-            // short batch as the last one would silently truncate the result set. We instead rely
-            // on an empty batch to detect the end, at the cost of at most one extra empty fetch.
+            // If the number of results in this batch is 0, no need to continue
             if (currentBatchSize == 0) {
               break;
             }
@@ -72,6 +65,19 @@ public class CacheableSearcher<K> {
                   Math.min(currentBatchSize, startInBatch + size - resultEntities.size());
               resultEntities.addAll(batchedResult.getEntities().subList(startInBatch, endInBatch));
               foundStart = true;
+            }
+            // Stop once the search engine has no more matching documents beyond this batch.
+            //
+            // We compare the raw hit positions scanned so far ((batchId + 1) * batchSize) against
+            // the total hit count, rather than checking (currentBatchSize < batchSize). Hits with
+            // invalid/missing URNs are dropped before reaching here (see
+            // SearchRequestHandler#getResultSafely), so a full page of batchSize raw hits can yield
+            // fewer than batchSize entities. Using the entity count to detect the last page would
+            // either silently truncate results when a hit is skipped, or force an unnecessary extra
+            // batch fetch. The total hit count is exact within the from/size pagination window
+            // (which the search engine caps at index.max_result_window).
+            if ((long) (batchId + 1) * batchSize >= batchedResult.getNumEntities()) {
+              break;
             }
             resultsSoFar += currentBatchSize;
             batchId++;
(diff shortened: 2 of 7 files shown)


metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/query/request/SearchRequestHandler.java
```diff
@@ -616,6 +611,34 @@ private SearchEntity getResult(@Nonnull SearchHit hit) {
...
+   * <p>Any other failure is left to propagate: silently swallowing it would mask real bugs and
+   * could drop valid results without surfacing the cause. The narrow {@link
+   * InvalidSearchHitException} catch ensures only the known, recoverable data-quality condition is
+   * tolerated.
+   */
+  private Optional<SearchEntity> getResultSafely(
+      @Nonnull OperationContext opContext, @Nonnull SearchHit hit) {
+    try {
+      return Optional.of(getResult(hit));
+    } catch (InvalidSearchHitException e) {
```
nwadams, line 636: are there a kind of exception we want to bubble up here?

nwadams (approved):

alexsku (changes requested):
Thanks for the thorough write-up and the targeted unit tests — the underlying truncation bug is real and the fix to `CacheableSearcher` is conceptually correct. Requesting changes because **CI is currently red on a pre-existing test that this change breaks**, plus a couple of behavioral points worth deciding consciously before merge.

## 🔴 Blocking: this breaks an existing integration test (`build-metadata-io`)

`LineageServiceTestBase.testSearchService` fails deterministically on all three backends (OpenSearch, ES8, ES7):

```
TooManyActualInvocations: searchClientShim.search(...)
Wanted 1 time … But was 2 times
  -> at ESSearchDAO.lambda$executeAndExtract$1(ESSearchDAO.java:145)   (x2)
  at LineageServiceTestBase.testSearchService(LineageServiceTestBase.java:366)
```

Root cause is the removed `if (currentBatchSize < batchSize) break;` in `CacheableSearcher`. The loop now fetches an extra empty "probe" batch to confirm end-of-results whenever the collected count hasn't reached `from + size`. In that test, a lineage search returns 1 result with page size 10 (`batchSize`=100), so `resultsSoFar(1) < from+size(10)` → batch 1 (`from=100`) is fetched, comes back empty → `search()` is invoked twice, but `LineageServiceTestBase.java:366` asserts `times(1)`.

That test isn't part of this PR's diff, so it wasn't updated. It looks like the ES/OpenSearch testcontainer suite wasn't run locally — worth doing before re-pushing, since these don't run in the lighter checks.

**Fix:** depends on the decision below. If the extra probe query is accepted, update the assertion to `times(2)` and use `searchRequestCaptor.getAllValues().get(0)` (first request is the real one; second is the empty probe). If you adopt the raw-count approach below, the test passes unchanged.

## 🟠 The "extra empty fetch" is more frequent than the description implies

The PR says "at most one extra empty fetch **on a full scan**." In practice the probe fires **whenever total results < `from + size`** — i.e. for every small/filtered result set and **the last page of every result set**. Since invalid-URN docs are rare, that extra ES round-trip almost always returns empty and is pure overhead. For a system where search/lineage queries can be high-volume and ES is a shared resource, ~doubling the query count for low-result searches is worth a deliberate decision rather than a side effect.

**Cleaner alternative that fixes the bug *without* the extra query:** the original break was wrong only because it compared the **post-skip** entity count to `batchSize`. The reliable end-of-results signal is the **raw** hit count from ES (`searchResponse.getHits().getHits().length`): ES returns a full `batchSize` page whenever more results exist, so `rawHits < batchSize` means "last batch" regardless of how many hits were skipped. If the raw per-batch hit count is plumbed through `SearchResult` (or a side channel) and the loop breaks on `rawHits < batchSize`, you keep the truncation fix, avoid the probe in the common case, and the existing test stays green. If you'd rather keep "break on empty" for simplicity, that's defensible — just please accept the cost explicitly and adjust the description.

## 🟡 The probe batch can cross `max_result_window`

There's no `max_result_window` override in the repo, so the OpenSearch default of **10000** applies, and `batchSize`=100. For a result set of ~9901–10000 entities paged to its **last** page, the probe batch issues `from=10000` and ES rejects it ("result window too large") — where the old code returned the final partial page. Narrow trigger (deep offset pagination to the end of a ~10k set), and that region is already fragile, but it's a new failure mode. The raw-count approach above also eliminates it.

## 🟢 Minor / question

- Narrowing `catch (Exception)` → `catch (InvalidSearchHitException)` is the right move. One thing to confirm: `extractMatchedFields` does `for (String q : hit.getMatchedQueries())` with no null guard, so any non-`InvalidSearchHitException` throw there now fails the whole search instead of skipping one hit. Almost certainly fine (search ran without any catch before #17387), but can `getMatchedQueries()`/`getHighlightFields()` ever be null for a valid hit?

## What's good

- The truncation fix itself is sound and the streaming/concatenation logic is internally consistent.
- The **scroll path** correctly keys `nextScrollId` off the raw `searchHits.length`, so it's already robust to skips — nice.
- `numEntities` from the trailing empty probe batch is still correct (total hits are query-global), so no count/metadata regression.
- Null-`_source` guard, the dedicated exception type, full-stacktrace logging, and the `skippedInvalidSearchHit` metric are all good, and the new `CacheableSearcherTest` / `SearchRequestHandlerTest` cases are well targeted.

### Summary of asks
1. **Get CI green** — update/repair `LineageServiceTestBase.testSearchService` (blocking).
2. Decide whether the extra probe query is acceptable or switch to raw-hit-count end detection (which also resolves the `max_result_window` edge and keeps the test green).
3. Quick confirm on the `getMatchedQueries()` null question.

alexjst:
Thanks for the thorough review — all four points are addressed in `f63168dbca` (CacheableSearcher rework) and `b50eba492a` (null guard).

**🔴 #1 — `LineageServiceTestBase.testSearchService` breakage (build-metadata-io).** Fixed. Reworked `CacheableSearcher` to detect end-of-results from the raw hit positions scanned (`(batchId + 1) * batchSize >= numEntities`) rather than the post-filter entity count. There's no probe batch anymore, so `search()` is invoked once again and the test passes **unchanged** (no `times(2)` edit needed). Verified `build-metadata-io` green in CI and ran the ES/OpenSearch testcontainer lineage suites locally.

**🟠 #2 — extra empty fetch frequency.** Resolved by the same rework — no probe batch in the common case or on the last page. Added `CacheableSearcherTest.testCacheableSearcherDoesNotOverFetchWhenResultsSmallerThanWindow`, which asserts a sub-window result set requires exactly one fetch.

**🟡 #3 — probe crossing `max_result_window`.** Gone with the probe; pagination never issues a fetch past the total hit count.

**🟢 #4 — null `getHighlightFields()` / `getMatchedQueries()`.** Good catch. `extractMatchedFields` now defaults both to empty (`b50eba492a`), so a valid hit with no highlight / no named-query match is returned instead of crashing the search under the narrowed catch. Regression test added (`testExtractResultHandlesNullHighlightsAndMatchedQueries`).

On total-vs-raw-count: I used the total hit count (`numEntities`, already on `SearchResult`) instead of plumbing a new raw per-batch count, to avoid a `SearchResult` model change / wider `EntitySearchService` signature change. It's exact within the from/size window (which the engine caps at `max_result_window`), so it gives the same single-fetch behavior the raw-count approach would.

Only remaining red is the Playwright E2E suite, which hits the 20-minute job cap during the fork image build (no depot cache) — unrelated to this change; all build / unit / integration checks are green.

metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/query/request/SearchRequestHandler.java
alexjst: Yes — intentionally. `getResultSafely` catches only `InvalidSearchHitException` (thrown by `UrnExtractionUtils.extractUrnFromSearchHit` for a missing/invalid URN), so that one known data-quality case is skipped while any other exception propagates and fails the search rather than being silently swallowed — which was the issue with the original `catch (Exception)` from #17387. I also added a null guard in `extractMatchedFields` (`b50eba492a`) so the most likely "unexpected" NPE on that path — null `getHighlightFields()`/`getMatchedQueries()` on a valid hit — can't crash the search either.

alexsku (approved):
Re-reviewed the rework in `f63168dbca` + `b50eba492a`. All four points from my previous review are addressed, CI is green (`build-metadata-io` + Playwright), and the new termination approach is cleaner than what I'd suggested. **Approving.** One non-blocking nit below (stale PR description) that I'd like fixed before the squash-merge since it becomes the permanent commit message.

Switching end-of-results detection to `(long)(batchId + 1) * batchSize >= numEntities` — comparing the raw scan position to the total hit count — is the right call: it fixes the truncation, removes the probe query (so `search()` is invoked once again and `LineageServiceTestBase.testSearchService` passes unchanged), and stops exactly at the `max_result_window` boundary instead of fetching past it. The `(long)` cast handles the overflow correctly. Traced it through small result sets, skip-in-a-full-page, exact-multiple, and deep-pagination-at-the-boundary — all correct.

## Backward compatibility (the part I scrutinized most)

Holds up well:

- **No API / model / cache-format change.** `getResultSafely` is private, `InvalidSearchHitException` is additive, and `extractUrnFromSearchHit` still throws (now a `RuntimeException` subtype, so existing `catch` sites still catch it). Reusing the existing `numEntities` field instead of adding one to `SearchResult` keeps cached entries and the wire format compatible — nice call-out in your reply.
- **Returned results are behavior-preserving** except the intended fix (no truncation): `numEntities` / `metadata` / `pageSize` semantics are unchanged.
- **ES query count is the same or fewer than master** (one query for result sets ≤ `batchSize`, no probe) — no perf regression.
- **The unreliable-count path doesn't reach this loop.** I confirmed the only production consumer of `CacheableSearcher` is `CachingEntitySearchService` via the standard `"entitySearchService"` bean, where `numEntities` = ES total hits under the default `track_total_hits` (= `max_result_window`), exact within the retrievable window. The KNN/semantic path (`track_total_hits=false`) is `SemanticEntitySearchService implements SemanticEntitySearch`, builds its own `SearchResult`, and never flows through `CacheableSearcher` — so the "numEntities is unreliable for KNN" concern from the original description doesn't apply here.

## Non-blocking

- **Please update the PR description before merge.** Section 1 still describes the abandoned v1 approach — "the loop now ends only on an empty batch, at the cost of at most one extra empty fetch" and "`numEntities` … is intentionally **not** used … `track_total_hits` is disabled/capped (KNN, etc.)". The final code does the opposite (uses `numEntities`, no probe). Since this is the squash-merge message, it'd be good to describe the raw-position-vs-total-hits approach and note *why* it's safe (KNN/semantic don't use this path).
- **Minor coupling note:** the loop now depends on `numEntities` being the accurate raw total within the window. True for the only current consumer and documented in the code comment, but routing a `track_total_hits=false` (or sub-window-capped) searcher through `CacheableSearcher` later would silently truncate. A one-line guard/note at the searcher lambda would lock the assumption in — optional.

## One confirm
- Any deployment that lowers `track_total_hits` below `max_result_window` for the *standard* search path would break the "exact within the window" assumption. I found no such config in-repo — just confirming none is set via env/runtime.

Thanks for the quick and thorough turnaround on this.

PR #17671, opened 2026-06-02
fix(ingestion): respect sample size env var for per-entry context LossyList

datahub-akshay:
## Summary

The `report_log` method was creating the per-entry context `LossyList` with no `max_elements` argument, causing it to default to 10 regardless of the configured `DATAHUB_REPORT_FAILURE/WARNING/INFO_SAMPLE_SIZE` environment variables.

### The problem

The structured log system has two layers of sampling:

1. **Outer `LossyDict`** — tracks unique error message types. Max size is controlled by the env vars (e.g. `DATAHUB_REPORT_FAILURE_SAMPLE_SIZE`).
2. **Inner `LossyList` per entry** — tracks individual dataset names within each grouped message. This was hardcoded to 10 with no env var override.

When many datasets share the same error message (e.g. a permission error hitting an entire schema), they are grouped into one `StructuredLogEntry`. Only 10 dataset names would appear in the context list even if the env vars were set to a large value. The truncation sentinel `... sampled of N total elements` would still appear.

### The fix

Pass `max_elements=entries.max_elements` when creating the context `LossyList`, so it inherits the same limit as the outer `LossyDict` for that log level. No new configuration is needed. The existing env vars now control both layers.

### Changes

- `source.py`: `LossyList()` replaced with `LossyList(max_elements=entries.max_elements)`
- New test `test_context_list_respects_sample_size` added to `test_report_sample_size.py`

## Testing

All 3 unit tests in `test_report_sample_size.py` pass.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/api/source.py
+++ b/metadata-ingestion/src/datahub/ingestion/api/source.py
@@ -186,7 +186,7 @@ class StructuredLogs(Report):
             logger.log(level=level.value, msg=log_content, stacklevel=stacklevel)
 
         if log_key not in entries:
-            context_list: LossyList[str] = LossyList()
+            context_list: LossyList[str] = LossyList(max_elements=entries.max_elements)
             if context is not None:
                 context_list.append(context)
             entries[log_key] = StructuredLogEntry(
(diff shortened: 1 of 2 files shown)


treff7es (approved):

treff7es:
This is a duplicate of https://github.com/datahub-project/datahub/issues/17590

PR #17680, opened 2026-06-02
fix(i18n): remove i18next.d.ts to avoid OOM errors

v-tarasevich-blitz-brain:
Example of OOM error - https://github.com/datahub-project/datahub/actions/runs/26821171552/job/79075948034?pr=17679

We already added NODE_OPTIONS=--max-old-space-size=8192 but with time it started failing again.

This problem is related to i18next version https://github.com/i18next/i18next/discussions/2246. But we can't upgrade it because the support of our version of TypeScript was removed -  https://github.com/i18next/i18next/blob/v26.3.0/CHANGELOG.md#2400.

Upgrading our TypeScript version is out of scope of this project. So we need to remove i18next.d.ts for now. TS will not check if translation keys exist in the json files after that.


benjiaming (approved):
I suppose this makes sense temporarily but I think we need to remember to put this back after TypeScript is upgraded. Can you create a follow up ticket?

PR #17681, opened 2026-06-02
feat(ui): add metadata tests filter

mihai103:
Hello, 

This PR adds a new search v2 filter for assets with failing metadata tests and makes the same capability available in the view builder. The code is very similar to the one for the incidents and assertions filters.


sakethvarma397:
@mihai103 Thanks for the contribution! Could you also please share the screenshots of the UI here? I understand that the changes should look petty much same as the existing filters, but wanted to do a quick sanity check.

mihai103:
hey @sakethvarma397,

Please find below a printscreen:

image

PR #17683, opened 2026-06-02
feat(ingest/dbt): add skip_missing_upstreams_in_lineage config option

JohnRTurner:
## Summary

Adds a new `skip_missing_upstreams_in_lineage: bool = False` config field to `DBTCommonConfig`. When enabled, upstream dataset URNs that don't already exist in DataHub are excluded from the `UpstreamLineageClass` aspect before it is emitted, preventing dangling graph edges ("ghost" nodes) from appearing in the lineage UI.

## How stub nodes are created (background)

When dbt ingestion emits an `UpstreamLineageClass` aspect for a downstream model, GMS routes it through `UpdateGraphIndicesService`, which calls `graphService.addEdge()` for every upstream entry. The `GraphService.addEdge` contract explicitly creates source and destination nodes if they don't exist (Neo4j: `MERGE`; Elasticsearch: writes the edge document with the URN as a dangling endpoint). This makes missing upstream datasets appear as ghost nodes in the lineage graph UI, even though no entity record exists in the entity store.

Existing workarounds (`skip_sources_in_lineage: true` + `entities_enabled.sources: NO`) prevent explicit source entity emission but do not prevent the graph edge from being written when the downstream model's lineage aspect still references the missing URN.

## How the fix works

Before emitting the `UpstreamLineageClass` for each dbt node, `_create_lineage_aspect_for_dbt_node` now checks each upstream URN against the DataHub entity store via `graph.exists()`. URNs absent from the entity store are dropped from the aspect. With no reference to the missing URN in the emitted aspect, `UpdateGraphIndicesService` never writes a graph edge for it, and no ghost node appears.

Existence results are cached per-run (one HTTP round-trip per unique upstream URN, regardless of how many dbt models reference it).

## Configuration

```yaml
source:
  type: dbt
  config:
    skip_sources_in_lineage: true      # suppress explicit source entity emission
    entities_enabled:
      sources: "NO"
    skip_missing_upstreams_in_lineage: true  # new — omit missing upstreams from lineage
```

**Requirements:** a DataHub graph connection is required. Using this flag without one (e.g. with a `file` sink) raises a clear error at startup via `ctx.require_graph()`.

## Caveats

These are worth understanding before enabling the flag:

1. **Out-of-order ingestion loses lineage edges.** If dbt is ingested before upstream source systems, those lineage edges are omitted and remain absent until dbt is re-ingested after the upstreams are present. Projects that rely on ingesting dbt first (or accept incomplete lineage temporarily) should not enable this flag.

2. **Adds latency proportional to unique upstream count.** Each unique upstream URN incurs one HTTP call to GMS per run. The per-run cache means a URN shared across hundreds of models is only checked once, but a project with many hundreds of distinct upstream tables will feel this.

3. **Parallel ingestion race.** If an upstream source is being ingested concurrently, `graph.exists()` may return `False` for a URN that is about to be created. That edge is dropped for the current run and restored on the next dbt re-run.

4. **Soft-deleted entities are treated as present.** `graph.exists()` checks for a key aspect in the entity store; soft-deletes retain their key aspect. Lineage to soft-deleted upstream entities is still emitted.

5. **Only covers dbt-entity lineage.** This flag filters upstream URNs in `_create_lineage_aspect_for_dbt_node` (the dbt platform lineage aspect). The sibling relationship between a dbt entity and its target platform counterpart emitted by `create_target_platform_mces` is not affected.

## Test plan

- [x] 4 new unit tests in `tests/unit/dbt/test_dbt_source.py`:
  - Config field parses correctly
  - Missing upstream filtered, existing upstream retained, report counter incremented
  - All upstreams missing → `None` returned (no empty aspect emitted)
  - `graph.exists()` called exactly once per unique URN across multiple models (cache verified)
- [x] E2E test against a live DataHub instance (scratch, not checked in): runs a full dbt pipeline via the `datahub-rest` sink and asserts the stored `upstreamLineage` aspect — verifies the missing URN is present with the flag off and absent with the flag on

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
@@ -149,6 +149,7 @@ from datahub.utilities.lossy_collections import LossyList
 from datahub.utilities.mapping import Constants, OperationProcessor
 from datahub.utilities.time import datetime_to_ts_millis
 from datahub.utilities.topological_sort import topological_sort
+from datahub.utilities.urns.urn import Urn
 
 logger = logging.getLogger(__name__)
 DBT_PLATFORM = "dbt"
@@ -547,7 +548,13 @@ class DBTCommonConfig(
     skip_missing_upstreams_in_lineage: bool = Field(
         default=False,
         description="When enabled, upstream datasets that do not already exist in DataHub are excluded from "
-        "lineage, preventing the creation of empty stub entities. Requires a DataHub graph connection.",
+        "lineage, preventing dangling graph edges from appearing in the lineage UI. "
+        "Typically used together with `skip_sources_in_lineage` and `entities_enabled.sources: NO`. "
+        "Important caveats: (1) if dbt is ingested before its upstream source systems, those lineage "
+        "edges will be silently omitted until dbt is re-ingested after the upstreams are present; "
+        "(2) adds one graph.exists() round-trip per unique upstream URN per run (cached within the run); "
+        "(3) soft-deleted upstream entities are treated as present. "
+        "Requires a DataHub graph connection.",
     )
     tag_prefix: str = Field(
         default=f"{DBT_PLATFORM}:", description="Prefix added to tags during ingestion."
@@ -1445,10 +1452,23 @@ class DBTSourceBase(StatefulIngestionSourceBase):
         return self._query_timestamp_cache
 
     def _upstream_exists_in_datahub(self, urn: str) -> bool:
-        """Check whether an upstream URN exists in DataHub, with per-run caching."""
+        """Check whether an upstream URN exists in DataHub, with per-run caching.
+
+        Fails open: if the existence check itself fails (transient GMS error, timeout,
+        etc.) the edge is kept to avoid silent lineage loss.
+        """
         if urn not in self._upstream_exists_cache:
             assert self.ctx.graph  # caller ensures graph is available
-            self._upstream_exists_cache[urn] = self.ctx.graph.exists(urn)
+            try:
+                self._upstream_exists_cache[urn] = self.ctx.graph.exists(urn)
+            except Exception as e:
+                self.report.report_warning(
+                    title="Upstream existence check failed",
+                    message="Could not verify upstream existence; keeping the lineage edge to avoid silent lineage loss.",
+                    context=urn,
+                    exc=e,
+                )
+                self._upstream_exists_cache[urn] = True  # fail open
             if not self._upstream_exists_cache[urn]:
                 self.report.lineage_upstreams_skipped_missing += 1
         return self._upstream_exists_cache[urn]
@@ -3224,12 +3244,30 @@ class DBTSourceBase(StatefulIngestionSourceBase):
                     )
                 ]
 
-            if self.config.skip_missing_upstreams_in_lineage and self.ctx.graph:
+            if self.config.skip_missing_upstreams_in_lineage:
                 upstream_urns = [
                     urn
                     for urn in upstream_urns
                     if self._upstream_exists_in_datahub(urn)
                 ]
+                # Also filter CLL: schemaField edges embed the parent dataset URN,
+                # so a missing upstream with CLL still creates the ghost node via
+                # graphService.addEdge(). Keep only entries whose upstream datasets
+                # are all in the surviving set.
+                if cll:
+                    existing_upstream_set = set(upstream_urns)
+                    filtered_cll = []
+                    for entry in cll:
+                        filtered_upstreams = [
+                            sf_urn
+                            for sf_urn in (entry.upstreams or [])
+                            if Urn.from_string(sf_urn).entity_ids[0]
+                            in existing_upstream_set
+                        ]
+                        if filtered_upstreams:
+                            entry.upstreams = filtered_upstreams
+                            filtered_cll.append(entry)
+                    cll = filtered_cll
 
             if not upstream_urns:
                 return None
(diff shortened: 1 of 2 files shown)


metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
```diff
@@ -542,6 +544,11 @@ class DBTCommonConfig(
         "Requires that `entities_enabled.sources` is set to `NO`. "
         "This is mainly useful when you have multiple, interdependent dbt projects. ",
     )
+    skip_missing_upstreams_in_lineage: bool = Field(
```
treff7es, line 548: **W3 — description omits the operational caveats you documented in the PR body.** A recipe author reads only this string. Worth folding in the most important gotchas:
- **Out-of-order ingestion silently drops edges** until the next dbt re-run (the biggest operational surprise).
- Adds one `graph.exists()` round-trip per distinct upstream (cached per run).
- Soft-deleted upstreams are treated as present.

Also consider cross-referencing `skip_sources_in_lineage` so users can tell the two apart.

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
```diff
@@ -1435,6 +1444,15 @@ def _get_query_timestamp(self) -> int:
         self.report.query_timestamps_fallback_used = True
         return self._query_timestamp_cache
 
+    def _upstream_exists_in_datahub(self, urn: str) -> bool:
+        """Check whether an upstream URN exists in DataHub, with per-run caching."""
+        if urn not in self._upstream_exists_cache:
+            assert self.ctx.graph  # caller ensures graph is available
+            self._upstream_exists_cache[urn] = self.ctx.graph.exists(urn)
```
treff7es: **C2 (robustness) — unguarded network call in a per-node loop can abort the whole run.** This is reached once per dbt node via `_create_lineage_aspect_for_dbt_node`. A transient GMS 5xx/timeout, token expiry, or a malformed URN raises and propagates up, killing the entire ingestion — discarding all the unrelated dbt metadata (schemas, descriptions, tags) too. A flag meant to *trim* lineage shouldn't be able to take down a healthy run on an infra blip. Suggest wrapping it and **failing open** (keep the edge you couldn't disprove) with a reported warning:

```python
try:
    exists = self.ctx.graph.exists(urn)
except Exception as e:
    self.report.report_warning(
        title="Upstream existence check failed",
        message="Could not verify upstream; keeping the edge to avoid silent lineage loss.",
        context=urn, exc=e,
    )
    exists = True
```

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
```diff
@@ -3202,6 +3224,13 @@ def _translate_dbt_name_to_upstream_urn(dbt_name: str) -> str:
                     )
                 ]
 
+            if self.config.skip_missing_upstreams_in_lineage and self.ctx.graph:
+                upstream_urns = [
+                    urn
+                    for urn in upstream_urns
+                    if self._upstream_exists_in_datahub(urn)
+                ]
```
treff7es, line 3261: **C1 (correctness) — column-level lineage is not filtered, so the ghost node can still be created.** This filters only `upstream_urns` (the coarse `upstreams[]`). The `cll`/`fineGrainedLineages` list is built independently above (from `sql_parsing_result.column_lineage` or `node.upstream_cll`) and emitted unfiltered in the same `UpstreamLineageClass`. A `schemaField` URN embeds its parent dataset URN, and `GraphService.addEdge()` materializes both endpoints — so a missing upstream that has CLL still ships schema-field edges that re-create the exact ghost node this flag targets. Please **filter `cll` against the same existence set**, or call this out explicitly in the caveats (the PR mentions siblings but not CLL). This path is also untested (no test sets `upstream_cll`/`include_column_lineage`).

**S1 (minor) — redundant guard.** Given `require_graph(...)` at startup, the `and self.ctx.graph` here is redundant and slightly muddies the "graph guaranteed at startup" contract. Keep the narrowing in one place.

treff7es (approved):

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
JohnRTurner: Good catch — expanded the field description to include the out-of-order ingestion caveat, per-URN latency note, soft-delete behaviour, and a cross-reference to `skip_sources_in_lineage`. Also added a `logger.warning` when `skip_missing_upstreams_in_lineage` is used without `skip_sources_in_lineage`, explaining why source lineage may be silently dropped on a first run.

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
JohnRTurner: Wrapped in try/except — on any exception the URN is cached as existing (fail open), the edge is retained, and a structured warning is recorded on the report so the issue is visible without aborting the run. Added a unit test and E2E test covering the fail-open path.

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
JohnRTurner: Fixed both — after filtering `upstream_urns`, the same existence set is used to strip `fineGrainedLineages` entries whose upstream `schemaField` URNs embed a missing dataset URN (`Urn.from_string(sf_urn).entity_ids[0]` extracts the parent dataset URN cleanly). `FineGrainedLineage` entries with no remaining upstreams are dropped entirely. Redundant `and self.ctx.graph` guard removed. Added a unit test specifically for the CLL filtering path.

PR #17684, opened 2026-06-02
perf(ingest): back UsageAggregator with FileBackedDict to bound memory

treff7es:
## Summary

`UsageAggregator` accumulated usage statistics entirely in memory
(`Dict[datetime, Dict[ResourceType, GenericAggregatedDataset]]`). Each
`GenericAggregatedDataset` holds three `Counter`s, and `queryFreq` is keyed by full
SQL query text — so for high-cardinality workloads this grows unbounded and is a
dominant driver of remote-executor OOMs (ING-2749). This actions the long-standing
`# TODO: Replace with FileBackedDict` on that class.

This change backs the aggregation with a SQLite-backed `FileBackedDict` so values spill
to disk, bounding RAM. **Emitted workunits are unchanged** (verified by the existing
golden/aggregator tests).

### What changed
- **`usage_common.py`**: `UsageAggregator` stores aggregates in a `FileBackedDict` keyed
  by `"<bucket_millis>@<resource>"`. `aggregate_event` uses `for_mutation` (load-or-create
  + dirty bit) so in-place counter updates survive cache eviction. The class is now
  `Closeable`. In-memory cache size is tunable via `DATAHUB_USAGE_AGGREGATOR_CACHE_SIZE`
  (default 2000), validated positive.
- **`sql_parsing_aggregator.py`**: shares its existing SQLite connection with the usage
  aggregator and closes it via the exit stack (single DB file).
- **`unity/usage.py`**: closes its own temp DB in a `finally` (guarded so cleanup can't
  mask an in-flight exception).

### Behavior / scope notes
- The `(time bucket, str(resource))` pair is the aggregation identity; resources whose
  `str()` are equal are intentionally merged. This contract is documented and pinned by a
  regression test. Safe for the current callers (`UrnStr`; usage-path `TableReference`,
  whose `last_updated` is unset).
- Memory is bounded ≤ the previous behavior. Trade-off: under very high cardinality with
  poor access locality there is a serialization cost on cache misses — tune the cache size
  env var if needed.
- Scope: this fixes usage paths that flow through `SqlParsingAggregator` (e.g. Snowflake
  `use_queries_v2`, the default) and Unity. Bespoke per-source usage aggregators
  (redshift/usage, clickhouse_usage, starburst_trino_usage) still use the in-memory pattern
  and are not migrated here.

## Test Plan
- [x] `pytest tests/unit/test_usage_common.py` — 18 passed (incl. eviction-persistence,
  non-str resource round-trip, equal-`str` merge contract, idempotent close,
  shared-connection-not-closed, env-var parser, non-positive cache-size guard)
- [x] `pytest tests/unit/sql_parsing/test_sql_aggregator.py` — passes (shared-connection
  wiring + unchanged usage goldens)
- [x] `pytest tests/unit -k unity` — 281 passed
- [x] `./gradlew :metadata-ingestion:lintFix` clean; mypy clean on changed files

## Checklist
- [x] Links to related issues — ING-2749
- [x] Tests added/updated
- [ ] Docs added/updated — n/a (no user-facing config beyond an internal tuning env var)
- [ ] Breaking changes — none

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/usage/usage_common.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/usage/usage_common.py
@@ -1,6 +1,5 @@
 import dataclasses
 import logging
-import os
 from datetime import datetime
 from typing import (
     Callable,
@@ -19,6 +18,7 @@ from pydantic.fields import Field
 
 import datahub.emitter.mce_builder as builder
 from datahub.configuration.common import AllowDenyPattern, HiddenFromDocs
+from datahub.configuration.env_vars import get_usage_aggregator_cache_size
 from datahub.configuration.time_window_config import (
     BaseTimeWindowConfig,
     BucketDuration,
@@ -46,35 +46,7 @@ ResourceType = TypeVar("ResourceType")
 # The total number of characters allowed across all queries in a single workunit.
 DEFAULT_QUERIES_CHARACTER_LIMIT = 24000
 
-# Default in-memory LRU cache size (number of GenericAggregatedDataset objects) for the
-# SQLite-backed usage aggregation store. Larger than FileBackedDict's library default (900)
-# because usage events for many resources interleave, so a bigger cache reduces
-# deserialize/re-pickle churn on the hot path. Override via env var for heavy workloads.
-DEFAULT_USAGE_AGGREGATOR_CACHE_SIZE = 2000
-
-
-def _get_usage_aggregator_cache_size() -> int:
-    raw = os.environ.get("DATAHUB_USAGE_AGGREGATOR_CACHE_SIZE")
-    if raw is None:
-        return DEFAULT_USAGE_AGGREGATOR_CACHE_SIZE
-    # A non-integer or non-positive override is treated as invalid: FileBackedDict
-    # requires a positive cache size (for_mutation asserts > 0), so falling back to the
-    # default avoids crashing on every event from a misconfigured env var.
-    try:
-        value = int(raw)
-    except ValueError:
-        value = 0
-    if value <= 0:
-        logger.warning(
-            "Ignoring invalid DATAHUB_USAGE_AGGREGATOR_CACHE_SIZE=%r; using default %d",
-            raw,
-            DEFAULT_USAGE_AGGREGATOR_CACHE_SIZE,
-        )
-        return DEFAULT_USAGE_AGGREGATOR_CACHE_SIZE
-    return value
-
-
-USAGE_AGGREGATOR_CACHE_SIZE = _get_usage_aggregator_cache_size()
+USAGE_AGGREGATOR_CACHE_SIZE = get_usage_aggregator_cache_size()
 
 
 def default_user_urn_builder(email: str) -> str:
@@ -308,7 +280,9 @@ class UsageAggregator(Generic[ResourceType], Closeable):
         # UrnStr (identity) and for usage-path TableReference (str() omits last_updated,
         # which the usage flow never sets — see TableReference.__str__). A ResourceType
         # whose str() is lossy w.r.t. its __eq__ would silently merge distinct resources.
-        return f"{int(floored_ts.timestamp() * 1000)}@{resource}"
+        # FileBackedDict keys must be str (it is a MutableMapping[str, _VT] over a SQLite
+        # TEXT column), so we encode the (bucket_millis, resource) identity as a string.
+        return f"{builder.make_ts_millis(floored_ts)}@{resource}"
 
     def aggregate_event(
         self,
--- a/metadata-ingestion/src/datahub/ingestion/source/usage/usage_common.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/usage/usage_common.py
@@ -1,11 +1,10 @@
 import dataclasses
 import logging
-from collections import defaultdict
+import os
 from datetime import datetime
 from typing import (
     Callable,
     Counter,
-    Dict,
     Generic,
     Iterable,
     List,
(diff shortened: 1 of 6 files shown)


treff7es:
## Benchmark: in-memory vs FileBackedDict

Fed an identical stream of **10M usage events** to the old in-memory storage (`defaultdict` + `GenericAggregatedDataset`, faithfully reproducing master) and the new `FileBackedDict`-backed `UsageAggregator`, each in its own process. Measured wall-time, peak process RSS, and (new mode) on-disk SQLite size. `cache_max_size=2000` (the default).

| Scenario | Mode | Wall time | Peak RAM | On-disk | Aggregates |
|---|---|---|---|---|---|
| Low cardinality (1k resources) | old | 38.3 s | 74.8 MB | — | 1,000 |
| Low cardinality (1k resources) | new | 63.5 s | 74.6 MB | 0.5 MB | 1,000 |
| High cardinality (1M resources) | old | 44.3 s | **1,425.7 MB** | — | 1,000,000 |
| High cardinality (1M resources) | new | 430.7 s | **85.0 MB** | 536.4 MB | 1,000,000 |

**Head-to-head**

| Regime | RAM (old → new) | Time (old → new) |
|---|---|---|
| Low cardinality (working set fits cache) | 74.8 → 74.6 MB (~same) | 38 → 63 s (1.7×) |
| High cardinality (working set ≫ cache) | 1,426 → 85 MB (**16.8× less**) | 44 → 431 s (9.7×) |

### Takeaways
- **Memory scales with cardinality, not event count.** The old path holds one `GenericAggregatedDataset` per unique (bucket, resource) in RAM, so it grows linearly — 1.4 GB at 1M resources and climbing (→ OOM). The new path stays flat at ~85 MB by spilling aggregates to a SQLite file (536 MB here). This is the OOM fix.
- **The cost is CPU time, and only when the working set exceeds the cache.** With resources < cache, RAM is identical and it's ~1.7× slower (dirty-bit bookkeeping + occasional flush). The 9.7× slowdown is the pathological all-misses case (working set 500× the cache), CPU-bound on pickle + SQLite (~43 µs/event vs 4.4 µs).
- **It's tunable:** `DATAHUB_USAGE_AGGREGATOR_CACHE_SIZE` trades RAM back for speed when needed.

**Caveat:** 1M distinct resources is a deliberate worst case (cache useless). Real usage workloads have far fewer tables with many queries each — i.e. the low-cardinality regime (~1.7× time, flat RAM), which is the realistic operating point.

metadata-ingestion/src/datahub/ingestion/source/usage/usage_common.py
```diff
@@ -42,6 +46,36 @@
...
 DEFAULT_QUERIES_CHARACTER_LIMIT = 24000
 
+# Default in-memory LRU cache size (number of GenericAggregatedDataset objects) for the
+# SQLite-backed usage aggregation store. Larger than FileBackedDict's library default (900)
+# because usage events for many resources interleave, so a bigger cache reduces
+# deserialize/re-pickle churn on the hot path. Override via env var for heavy workloads.
+DEFAULT_USAGE_AGGREGATOR_CACHE_SIZE = 2000
+
+
+def _get_usage_aggregator_cache_size() -> int:
```
sgomezvillamor: should this be moved to `metadata-ingestion/src/datahub/configuration/env_vars.py` ?

metadata-ingestion/src/datahub/ingestion/source/usage/usage_common.py
```diff
@@ -240,14 +274,41 @@ def ensure_top_n_queries_is_not_too_big(cls, v: int, info: ValidationInfo) -> in
...
+        self._aggregation: FileBackedDict[GenericAggregatedDataset[ResourceType]] = (
+            FileBackedDict(
+                shared_connection=shared_connection,
+                tablename="usage_aggregation",
+                cache_max_size=cache_max_size,
+            )
+        )
+
+    @staticmethod
+    def _make_key(floored_ts: datetime, resource: ResourceType) -> str:
```
sgomezvillamor, line 276: minor both: 
we have a `def make_ts_millis(ts: datetime) -> int` method that we can use here
and as a key, instead of `str`, we may use `Tuple[int, str]`

metadata-ingestion/src/datahub/ingestion/source/usage/usage_common.py
```diff
@@ -240,14 +274,41 @@ def ensure_top_n_queries_is_not_too_big(cls, v: int, info: ValidationInfo) -> in
...
+        if cache_max_size <= 0:
+            raise ValueError(f"cache_max_size must be positive, got {cache_max_size}")
         self.config = config
-        self.aggregation: Dict[
-            datetime, Dict[ResourceType, GenericAggregatedDataset[ResourceType]]
-        ] = defaultdict(dict)
+        # Map of "<bucket_millis>@<resource>" -> GenericAggregatedDataset.
+        # Backed by SQLite so high-cardinality query/user/column counters spill to disk
+        # instead of growing unbounded in memory.
+        self._aggregation: FileBackedDict[GenericAggregatedDataset[ResourceType]] = (
```
sgomezvillamor, line 267: the FileBackedDict does not specify a key because it is always str ?

sgomezvillamor (approved):

metadata-ingestion/src/datahub/ingestion/source/usage/usage_common.py
treff7es: Moved it — `DATAHUB_USAGE_AGGREGATOR_CACHE_SIZE` is now read via `get_usage_aggregator_cache_size()` in `configuration/env_vars.py` (alongside `get_sql_parse_cache_size`), and `usage_common` imports it. Done in 681ab2bdf9c.

metadata-ingestion/src/datahub/ingestion/source/usage/usage_common.py
treff7es: Switched to `builder.make_ts_millis(floored_ts)`. Kept the key as `str` rather than `Tuple[int, str]`: `FileBackedDict` is a `MutableMapping[str, _VT]` backed by a SQLite `key TEXT` column, so it only supports str keys (which is your point in the other comment). 681ab2bdf9c.

metadata-ingestion/src/datahub/ingestion/source/usage/usage_common.py
treff7es: Correct — `FileBackedDict` keys are always `str` (`MutableMapping[str, _VT]`, SQLite `key TEXT UNIQUE`). That is why the aggregation key is the `f"{millis}@{resource}"` string and a `Tuple[int, str]` key is not possible.

treff7es:
Superseded by #17703 (merged), which fixes the same OOM (ING-2749) with a row-level `FileBackedCounter` instead of spilling whole aggregates via `FileBackedDict`. That approach benchmarked 2.6× faster and is robust to a single hot resource. Closing in favor of the merged work.

PR #17692, opened 2026-06-02
fix(masking): mask at handler level and scope secrets per execution

kyungsoo-datahub:
## Problem

The filter was attached to the root *logger*. Python doesn't consult a logger's own
filters for records propagated from child loggers, so it never saw ingestion output.
Masking worked only because `_update_existing_handlers()` repointed console handler
streams at the wrapped `sys.stderr`. Two consequences:

- **Secrets reached log files in plain text.** Repointing covers stdout/stderr only;
  a `FileHandler` got child-logger records unmasked. The console looked correct, so
  this was silent.
- **Logging stopped under celery.** Celery's `sys.stderr` proxy turns writes back into
  log records, so a handler pointed at it cycles: handler → proxy → same handler.
  Output is dropped. Teardown restored `sys.stderr` but not handler streams, so it
  persisted. Symptom in worker mode: after an ingestion with secrets, celery/root logs
  stop appearing while ingestion-subprocess logs continue.

Review surfaced a third: with overlapping executions, the first to finish removed the
filter and cleared the shared registry, unmasking the one still running.

## Changes

- Remove `_update_existing_handlers()`; attach the filter to each handler instead.
  Handler filters *are* consulted for propagated records, so child-logger output —
  including file output — is masked. Teardown is symmetric and identity-based.
- Keep the stdout/stderr wrapper for raw `print()` only, skipped when the stream isn't
  OS-backed (`fileno()` raises). Wrapping a proxy re-enters logging; skipping loses
  nothing, since proxied writes become log records.
- Scope secrets per execution. Teardown drops only that group, fully uninstalling when
  none remain. `initialize_secret_masking()` returns a token
  `shutdown_secret_masking(token)` accepts, for cross-thread teardown.

Coverage gaps that were unreachable while the filter sat on root:

- Records masked once, not once per handler (the second truncation ate the first
  suffix and reported a wrong byte count).
- `extra={...}` masked, recursing into containers on a copy. Depth cap and cycle guard
  emit a placeholder, not the raw subtree.
- Exceptions no longer rebuilt via `type(exc)(*masked_args)` — that raised whenever
  `__init__` didn't match `args` (common in DB drivers) and the fallback returned a
  bare `RuntimeError`, destroying the traceback, class, and `__cause__` chain. The
  traceback is materialized to `exc_text` and masked there; `exc_info` untouched.
- `logging_utils` captures `sys.__stderr__`, not import-time `sys.stderr` — already the
  celery proxy when imported at task time.
- Capacity guard counts expanded keys, not raw secrets, in both registration paths.

## Behavior changes

- `max_message_size` (5000) previously applied only to records logged directly on root;
  now to every record reaching a handler.
- Error reporters (Sentry, Datadog) read the unmasked `exc_info`. That's what preserves
  tracebacks; mask at the reporter integration instead.
- Structured formatters emit plain-text tracebacks (`exc_text` becomes authoritative).
- ~2.3× on the logging path (12.7 → 29 µs/record, 20 secrets), i.e. ~0.16s per 10k
  lines. Flat across handlers.

## Known limitations (documented in-code)

- Handler coverage is a snapshot at install; a `FileHandler` added afterwards in a
  non-celery process is uncovered.
- Arbitrary objects in `extra=` aren't stringified-and-masked (`__str__` side effects).
- A formatter with a custom `formatException` bypasses the masked `exc_text`.
- `shutdown_secret_masking()` without a token falls back to the ambient context and
  no-ops if it has no scope (debug-logged; fails safe).

## Testing

- `TestHandlerCoverageAndCelerySafety` — child-logger masking; celery feedback stream
  no longer silences logs. Confirmed failing before the fix.
- `TestConcurrentExecutions` — teardown doesn't unmask or wipe a peer execution;
  sequenced through the `_teardown_hook()` seam, not sleeps. Includes a test that
  re-introduces the old semantics and asserts the invariant breaks.
- `TestReviewFixes` — idempotency, extras recursion and non-mutation, cycle/depth
  placeholders, traceback preservation, registry rebind.
- N masking unit tests pass; ruff clean.

---

> [!NOTE]
> **Overview**
> Reworks secret masking so **child-logger and file output are actually covered**, Celery workers don’t lose logs, and **overlapping ingestions** don’t tear down each other’s secrets.
> 
> **Installation model:** Stops repointing handler streams to wrapped `stdout`/`stderr` (the Celery feedback loop). Masking now attaches **`SecretMaskingFilter` to each handler** (plus a `Handler.__init__` hook for handlers created later). `sys.stdout`/`stderr` are wrapped only for raw `print()` when `fileno()` works—proxies like Celery’s `LoggingProxy` are skipped. Install/uninstall track a single `_installed_filter` by identity.
> 
> **Lifecycle:** `initialize_secret_masking()` returns an **execution token** and opens a distinct registry scope; `shutdown_secret_masking(token)` drops only that scope and fully tears down when no scopes remain. Bootstrap uses an **RLock**, clears `_bootstrap_completed` when the filter is gone, and **raises `RuntimeError`** (with critical logging) on install failure instead of failing silently. Disabled masking no longer latches “bootstrapped” without installing a filter.
> 
> **Filter behavior:** Mask **before** truncate with a pattern-length-aware pre-cut; recursive masking of `extra={...}` and non-string `msg`; traceback text via `exc_text` while leaving `exc_info` intact; string idempotency token `_MASKED`; marker-first regex with `lastgroup` dispatch. Shared `REDACTED_FORMAT` in `constants.py`. Masking-safe logging binds to **`sys.__stderr__`**.
> 
> **Registry:** Secrets grouped per execution (COW snapshots, incremental publish on register); cross-thread register via `exec_id`; capacity on expanded keys; registration rejects marker-shaped values.
> 
> Adds a **Cursor rule** for masking invariants and expands unit tests (concurrent executions, bootstrap latch, force teardown, handler coverage).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c827a36ced3c23f442fcc51dc4d2ef9cac8d35ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/metadata-ingestion/src/datahub/masking/secret_registry.py
+++ b/metadata-ingestion/src/datahub/masking/secret_registry.py
@@ -8,6 +8,10 @@ active executions' secrets (via get_all_secrets/get_version), so masking is
 process-global and always-on — it can over-mask during overlap (safe) but never
 under-mask. See datahub.masking.bootstrap for the lifecycle.
 
+Secrets registered outside any execution scope land in a catch-all ``__global__``
+group; it is dropped once the last real execution ends (so it lives at most as
+long as concurrent execution activity, not for the whole process).
+
 Concurrency: copy-on-write. Writers (register/begin/end/clear) hold the lock and
 atomically swap in freshly-built dicts; readers (the masking hot path) read those
 immutable snapshots without a lock.
--- a/metadata-ingestion/src/datahub/masking/secret_registry.py
+++ b/metadata-ingestion/src/datahub/masking/secret_registry.py
@@ -132,10 +132,6 @@ class SecretRegistry:
             self._rebuild_locked()
             return bool(active)
 
-    def has_active_executions(self) -> bool:
-        with self._registry_lock:
-            return any(g != _GLOBAL_GROUP for g in self._groups)
-
     def _current_group_locked(self) -> Dict[str, str]:
         exec_id = _current_exec.get() or _GLOBAL_GROUP
         return self._groups.setdefault(exec_id, {})
--- a/metadata-ingestion/src/datahub/masking/secret_registry.py
+++ b/metadata-ingestion/src/datahub/masking/secret_registry.py
@@ -108,6 +108,15 @@ class SecretRegistry:
         _current_exec.set(exec_id)
         return exec_id
 
+    def ensure_execution(self) -> str:
+        """Open a secret scope for the current context only if it doesn't already
+        have one; returns the active execution id. Idempotent within a context,
+        so a repeated initialize_secret_masking() won't start a second scope."""
+        exec_id = _current_exec.get()
+        if exec_id is not None:
+            return exec_id
+        return self.begin_execution()
+
     def end_execution(self) -> bool:
         """Drop the current execution's secrets. Returns True if other
         executions are still active (so the caller should NOT fully tear down)."""
--- a/metadata-ingestion/src/datahub/masking/secret_registry.py
+++ b/metadata-ingestion/src/datahub/masking/secret_registry.py
@@ -1,19 +1,35 @@
 """
 Thread-safe registry for managing secrets used in masking.
 
-This module provides a singleton registry for storing and managing secrets.
-Secrets are automatically registered from ConfigModel fields and environment
-variables, then used by logging filters to mask sensitive data.
+Secrets are grouped by the execution that registered them, so overlapping
+in-process executions don't interfere: one execution ending removes only its own
+secrets, never another's. The masking filter always sees the *union* of all
+active executions' secrets (via get_all_secrets/get_version), so masking is
+process-global and always-on — it can over-mask during overlap (safe) but never
+under-mask. See datahub.masking.bootstrap for the lifecycle.
 
+Concurrency: copy-on-write. Writers (register/begin/end/clear) hold the lock and
+atomically swap in freshly-built dicts; readers (the masking hot path) read those
+immutable snapshots without a lock.
 """
 
 import os
 import threading
+import uuid
+from contextvars import ContextVar
 from typing import Dict, Optional
 
 from datahub.masking.logging_utils import get_masking_safe_logger
 
 logger = get_masking_safe_logger(__name__)
+
+# Secrets registered outside an explicit execution scope land here. Dropped when
+# no real executions remain active.
+_GLOBAL_GROUP = "__global__"
+
+# Identifies the execution registering/owning secrets in the current context.
+# Set by begin_execution(), read by register_secret() and end_execution().
+_current_exec: ContextVar[Optional[str]] = ContextVar("masking_exec_id", default=None)
 
 
 def is_masking_enabled() -> bool:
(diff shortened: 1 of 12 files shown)


metadata-ingestion/tests/conftest.py
```diff
@@ -44,6 +46,21 @@ def mock_time():
...
+
+    Masking installs a filter on every logging handler and populates a singleton
+    SecretRegistry. Tests that trigger it (the ingest CLI, initialize_secret_masking,
+    or SecretStr config validation) don't always tear it down, which would mask
+    later tests' captured log output. Reset after each test for isolation
+    (shutdown also clears bootstrap state so a later init re-installs cleanly).
+    """
+    yield
+    shutdown_secret_masking()
+    SecretRegistry.reset_instance()
```
skrydal: Do we absolutely need this on autouse across all tests?

skrydal:
I do not agree with the statement:
> Repeat install re-scans for handlers added since; logger/handler iteration is concurrency-safe.

I do not think this PR addresses concurrency issue raised initially:

```
Thread A (exec 1)                                                                                     Thread B (exec 2)
  ───────────────────────                 ───────────────────────
  initialize_secret_masking()
    ├ installs filter on root logger
    └ registers secrets
                                                                                                                initialize_secret_masking(force=True)
                                                                                                                  ├ filter exists → returns early
                                                                                                                  └ registers its own secrets
    ⋮ execution runs
    ⋮
  shutdown_secret_masking()
    ├ removes filter from root logger ──▶   ⚠ filter gone, secrets unmasked
    ├ clears registry ───────────────▶   ⚠ Thread B's secrets wiped
    └ restores handler streams
                                            ⋮ execution still running
                                            ⋮ log output may contain secrets

```
Can we test for it reliably?

skrydal (approved):
I am approving to not block merging later, but please address my comments.
Overall solution seems to me overcomplicated and I wondered whether we should take a step back to how logging is handled. Also considering we are logging in so many different formats now.

metadata-ingestion/tests/conftest.py
kyungsoo-datahub: Yes, but it's cheap. The fixture is guarded, so it's a no-op (two quick checks) for any test that doesn't touch masking, and only resets when masking was actually left active.

metadata-ingestion/src/datahub/masking/secret_registry.py
```diff
@@ -50,166 +132,303 @@ def get_instance(cls) -> "SecretRegistry":
...
+
+    # --- Registration (writers) --------------------------------------------
+
+    def _admit_locked(
+        self,
+        variable_name: str,
+        raw_value: str,
+        group: Dict[str, str],
+        pending_keys: set,
+        warn: bool = True,
```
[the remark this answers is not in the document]
kyungsoo-datahub: eproduced the mechanism, but disputing the severity and the causal claim.

The revival is real — `_current_group_locked()` setdefaults the dead id back in:

```
after cross-thread teardown: filter installed = False, root filters = 0
register_secret('B', ...) -> groups = ['<dead id>']
```

But the cleartext isn't caused by it. Control, registering from a fresh context with no
stale ambient id at all:

```
CONTROL fresh-context register -> cleartext = True | groups = ['__global__']
```

Identical outcome. The cleartext comes from registering *after a completed teardown*,
when the filter is deliberately uninstalled — which bucket the secret lands in doesn't
affect masking. Changing `_current_group_locked` wouldn't change this.

The revived id also doesn't poison later teardowns. The next `initialize_secret_masking()`
reuses it (`ensure_execution` finds it live again) and it drains normally:

```
AFTER proper shutdown: filter installed = False | groups = [] | count = 0
```

So: not a leak, and self-healing. There is a real hygiene point underneath — registration
silently succeeding while no filter is installed gives false confidence that masking is
on. Adding a debug log for that instead of changing the registration path.

PR #17701, opened 2026-06-03
feat(i18n): extract keys from entity type pages

max-datahub:
## Summary

Extracts hardcoded user-visible strings from the **entityV2 entity-type pages** into a new `entity.types` react-i18next namespace, covering **30 entity types** (dataset, chart, dashboard, container, domain, glossary term/node, tag, business attribute, data product, application, user, group, the ML entities, and the data-pipeline entities).

- New `en/entity.types.json` (**306 keys**), structured to dedupe aggressively:
  - `tab.*` — shared tab labels reused across all types (`properties`, `lineage`, `documentation`, `summary`, `incidents`, …)
  - `sidebar.*` — shared sidebar section descriptions
  - `shared.*` — cross-type strings (stat percentiles, counts, run columns), with `_one`/`_other` pluralization and `<Trans>` for the document change-history sentences
  - `<type>.name` / `<type>.namePlural` — entity **display names** (see note below)
  - per-type page strings
- Reuses `common.actions` / `common.feedback` / `common.labels`; adds `common.feedback.changesSaved` and `common.labels.createdAt` / `lastModifiedAt`.
- Registers the namespace in `i18n.ts` and `setupTests.ts`.
- Extends the ESLint untranslatable-attribute exclude list for non-text props (`language`, `fill`, `justifyContent`, `field`, `tab`, `*Path`, `*TestID`).
- Fixes 3 pre-existing English typos in the extracted copy (`Updates`→`Updated Data Product!`; dashboard `More views often than`→`More views than`; `Failed to group members`→`Failed to add group members`).
- Entity classes use `i18next.t(...)`; module-scope column configs are built in-render so they can translate; intentionally-untranslated strings are marked; CSS fixes applied where translated text could overflow.

**Entity-type display names are extracted but kept in English** — treated as do-not-translate proper nouns and centralized for consistency (the German pass keeps them English). **EN-only**; German translation follows in a separate PR.

> Note: `i18next.d.ts` is intentionally not touched/re-added — it was removed upstream in #17680 to avoid the `tsc` OOM caused by its typed-resources augmentation.

## Pages to verify

Open with a non-English locale and scan for overflow/truncation or untranslated text:

- **Each entity-type profile page** (Dataset, Chart, Dashboard, Container, Domain, Glossary Term/Node, Tag, Business Attribute, Data Product, Application, User, Group, ML Model/Group/Feature/Feature Table/Primary Key, Data Job/Flow/Process Instance) — tab labels, sidebar section headers + descriptions, summary/stats sections, preview cards
- **Browse sidebar & search filters** — entity-type collection names
- **Document** — change-history timeline (the `<Trans>` sentences), and create/move/restore document toasts
- **User / Group** — edit-profile modals, member/owner add-remove toasts, membership counts
- **Glossary** — related-terms add/remove; **Domain** — entity/sub-domain/data-product counts

## Checklist

- [x] `tsc --noEmit`, ESLint, and Prettier pass locally
- [ ] Docs added/updated (n/a — UI string extraction)
- [ ] Breaking changes (none)


datahub-web-react/src/app/entityV2/glossaryTerm/profile/GlossaryRelatedTermsResult.tsx
```diff
@@ -12,6 +13,7 @@ import { Button } from '@src/alchemy-components';
 
 import { TermRelationshipType } from '@types';
 
+/* untranslated-text -- dual-use enum: UI label + comparison key; needs label/key decouple */
 export enum RelatedTermTypes {
     hasRelatedTerms = 'Contains',
     isRelatedTerms = 'Inherits',
```
benjiaming, line 19: We do have to translate these. It's fine to do them as followup but please create a ticket.

datahub-web-react/src/app/entityV2/group/types.ts
```diff
@@ -1,4 +1,6 @@
 export enum TabType {
+    /* untranslated-text -- enum doubles as route path + identity key; translating breaks routing */
     Assets = 'Owner Of',
+    /* untranslated-text -- enum doubles as route path + deep-link target */
     Members = 'Members',
```
benjiaming, line 5: Same, we do have to translate this but ok if it's a follow up.

benjiaming (approved):

datahub-web-react/src/app/entityV2/glossaryTerm/profile/GlossaryRelatedTermsResult.tsx
max-datahub: Filed #17725 to track this. These are dual-use enums (the value is also a route path / comparison key), so the fix is a small decouple refactor — keep the value as the stable identity and add a translated label map (the `Record<Enum, () => string>` + `i18next.t()` pattern already used by `getDataContractCategoryLabel`). Keeping the `/* untranslated-text */` skip here for now; will address in that follow-up.

datahub-web-react/src/app/entityV2/group/types.ts
max-datahub: Filed #17725 to track this. These are dual-use enums (the value is also a route path / comparison key), so the fix is a small decouple refactor — keep the value as the stable identity and add a translated label map (the `Record<Enum, () => string>` + `i18next.t()` pattern already used by `getDataContractCategoryLabel`). Keeping the `/* untranslated-text */` skip here for now; will address in that follow-up.

PR #17702, opened 2026-06-03
feat(i18n): extract keys from common shared components

max-datahub:
## Summary

Extracts hardcoded user-visible strings from the common shared component areas (`src/app/shared` and `src/app/sharedV2`) into react-i18next namespaces. **English only** — German translations follow in a separate PR.

The bulk of these directories had not yet been migrated. This PR covers the files that actually contain user-visible strings (~41 files); the remaining shared files are utils/types/hooks/contexts with no user-visible text and are left for a future enforcement pass.

### Namespaces

- **11 new `shared.*` namespaces** — `shared.tags`, `shared.business-attribute`, `shared.share`, `shared.error`, `shared.health`, `shared.product`, `shared.time`, `shared.confirmation`, `shared.search`, `shared.propagation`, `shared.misc` — registered in `src/i18n/i18n.ts` (`NAMESPACES`) and `src/setupTests.ts`. (Following #17680, namespaces are no longer declared in `i18next.d.ts`.)
- Generic verbs/labels/feedback reuse `common.actions`/`common.feedback`/`common.labels` (with a few new generic keys: `dismiss`, `exit`, `home`, `somethingWentWrong`, `none`).

### Changes

- Migrate 41 component/util files to `t()`/`tc()`/`<Trans>` with `useTranslation` (or `i18next.t('ns:key')` in non-component modules).
- Pluralized durations/counts use `_one`/`_other` with `count`; sentences with inline formatting use `<Trans>` (error "Need support?" line, `LastUpdated` legends + platform-name label).
- Replace deprecated `ANTD_GRAY`/`REDESIGN_COLORS` imports **and** pre-existing hardcoded hex/rgb colors in the touched files with semantic theme tokens (`props.theme.colors.*` / `useTheme().colors.*`).
- Add the migrated files to `translated-files.txt` to enforce `no-literal-string` on them going forward.

## Pages to verify (with a non-EN locale)

- **Any entity profile** — tag/term pills (add/remove, "Create" tag, color save), business-attribute section, "Applied to"/owners empty states, the share menu (Copy Link/Name/URN, Email), "Last updated …" tooltip (legend + platform name/logo).
- **Service-account / settings flows** that surface the shared error boundary/fallback ("Whoops!", "Something went wrong"), upload failures, and the discard-unsaved-changes confirmation.
- **Search/list toolbars** — Download/Edit buttons (tooltips, "Downloading…"), pagination, and the "404 / page not found" page.

## Checklist

- [x] Tests — existing unit tests for the touched components pass (no behavior change; two pre-existing `localStorage` test-isolation failures are unrelated and reproduce on master)
- [ ] Docs (n/a)
- [ ] Breaking changes (none)


datahub-web-react/src/app/shared/tags/AddTagsTermsModal.tsx
```diff
@@ -336,7 +338,10 @@ export default function EditTagTermsModal({
             .catch((e) => {
                 message.destroy();
                 message.error(
-                    handleBatchError(urns, e, { content: `Failed to add: \n ${e.message || ''}`, duration: 3 }),
+                    handleBatchError(urns, e, {
+                        content: t('addTags.error', { error: e.message || '' }),
```
benjiaming: Wait isn't this for `EntityType.GlossaryTerm`? shouldn't it be `addTerms.error`?

datahub-web-react/src/i18n/locales/en/shared.misc.json
```diff
@@ -0,0 +1,11 @@
+{
+    "backButton.tooltip": "Go back",
+    "confirmationModal.areYouSure": "Are you sure?",
+    "copyUrn.tooltip": "Copy URN. An URN uniquely identifies an entity on DataHub.",
+    "customIcon.alt": "Icon",
+    "entityCountCard.viewCountTooltip": "View {{formattedCount}} {{tooltipDescriptor}}",
+    "multiStepForm.disabledNextTooltip": "Please complete all required fields before moving to the next step",
+    "noPageFound.backToHome": "Back to Home",
+    "noPageFound.subtitle": "The page you requested was not found,",
```
benjiaming: There's an extra comma at the end. We should fix it even if it's a preexisting bug.

benjiaming (changes requested):

datahub-web-react/src/app/shared/tags/AddTagsTermsModal.tsx
max-datahub: Good catch on the inconsistency. I went with a slightly different fix than `addTerms.error` — reasoning:

The success toasts branch because the copy genuinely differs (`Added Tags!` vs `Added Terms!`). The error toasts, though, were always type-agnostic — the original (pre-i18n) code was `` `Failed to add: \n ${e.message}` `` for both tags and terms, with no tag/term noun. So an `addTerms.error` key would just hold the same English as `addTags.error`, making the branch cosmetic.

Instead I renamed the misleadingly tag-specific keys to generic `add.error` / `remove.error` and updated all four call sites (the remove handlers had the same pattern). Happy to switch to per-type keys if you'd prefer symmetry with the success messages.

datahub-web-react/src/i18n/locales/en/shared.misc.json
max-datahub: fixed

benjiaming (approved):

PR #17703, opened 2026-06-03
perf(ingest): back UsageAggregator with SQL-side incremental counting

treff7es:
## Summary

Backs `UsageAggregator` with **SQLite-side incremental counting** so the usage aggregation hot path is fast *and* memory stays bounded for high-cardinality workloads (ING-2749), and extracts the counting machinery into a reusable, swappable `FileBackedCounter` primitive. The public interface (`aggregate_event` / `generate_workunits` / `close`) and the emitted `DatasetUsageStatistics` workunits are **behavior-equivalent**. `GenericAggregatedDataset` and `make_usage_workunit` are unchanged (other extractors depend on them).

**Supersedes #17684** (the earlier FileBackedDict-of-whole-aggregates approach, which benchmarked as CPU-bound on per-event pickling).

### What's here
1. **`FileBackedCounter`** (`file_backed_collections.py`) — a new member of the `FileBacked*` family: a disk-backed two-level counter `(group_key, item_key) -> count` with batched increment-upserts and grouped most-common reads (deterministic `cnt DESC, rowid ASC`, insertion-order tie-break — matches `Counter.most_common`). It is **thread-safe** (a lock guards the in-memory write buffer), **validates `tablename`** as an identifier (it's interpolated into SQL — SQLite can't bind identifiers), mirrors `FileBackedDict`'s SQLite ≥3.24 `ON CONFLICT` guard/fallback, and **registers itself in the shared connection's dependents** so closing the connection flushes buffered increments.
2. **`GroupedItemCounter` protocol + `InMemoryGroupedCounter`** — the counter backend is a clean swap seam: `UsageAggregator(config, counter=...)` accepts any implementation (defaults to `FileBackedCounter`; an in-memory backend is available for tests / small workloads). The two are proven interchangeable by a dual-backend parity test.
3. **`UsageAggregator` rewired** (`usage_common.py`) — composes a `FileBackedCounter` (counts) + a `FileBackedDict` (resource objects); group existence is folded into a sentinel row so denied/empty events still emit. The `(group_key, item_key)` encoding is centralized in a single `_UsageKeys` codec, and the per-event recording rule lives in one `_add_read_entry` method. No hand-rolled SQL remains in `usage_common.py`.
4. `SqlParsingAggregator` shares its SQLite connection with the aggregator (closed via its exit stack); Unity closes its own temp DB in a guarded `finally`.

### Benchmark (10M events, 1M resources, worst case)
| Approach | Time | Peak RAM |
| --- | --- | --- |
| Old in-memory | 44 s | 1,426 MB |
| FileBackedDict-of-aggregates (#17684) | 431 s | 85 MB |
| **This PR (SQL counting)** | **166 s** | **93 MB** |

2.6× faster than #17684, flat ~90 MB vs 1.4 GB in-memory.

### Why this over the FileBackedDict approach (#17684)?

The core difference is **the storage granularity matches the operation**. Both approaches spill to disk (both fix the OOM), but they differ in what a single usage event costs:

| | FileBackedDict-of-aggregates (#17684) | This PR (SQL counting) |
| --- | --- | --- |
| Stores | `key -> GenericAggregatedDataset` (a whole object: 3 `Counter`s) | `(bucket, resource, dim, item) -> count` rows |
| Per-event cost | load + unpickle the whole aggregate → mutate one counter → re-pickle the whole object on eviction = **O(aggregate size)** | a batched `INSERT ... ON CONFLICT DO UPDATE SET cnt = cnt + 1` = **O(1) increment**, done by SQLite |

The usage workload is *"increment a count," millions of times*. `FileBackedDict` is built for *get/put a whole value*, so every increment paid the cost of (de)serializing the entire aggregate just to bump one number — which is why #17684 was CPU-bound on pickle. Concretely, this version is:

- **2.6× faster** (166 s vs 431 s) — no per-event object (de)serialization; aggregation happens in SQLite.
- **Robust to a single hot resource** — `FileBackedDict` spills at the granularity of a whole aggregate, so one resource with millions of unique queries is a single giant object that must fit in RAM to be mutated (disk can't help). SQL-counting spills at the row level, so even that case stays bounded.
- **Smaller on disk** (344 MB vs 536 MB) — compact count rows vs pickled dataclasses-with-`Counter`s.

Trade-off: it adds a new primitive (`FileBackedCounter`) and rewires the aggregator rather than reusing `FileBackedDict` as-is — but the primitive is reusable and the swap seam keeps the aggregator clean. Net across the three options: in-memory is fastest but OOMs; FileBackedDict bounds memory but is pickle-bound; SQL counting bounds memory **and** recovers most of the speed.

### Concurrency contract
`FileBackedCounter` is thread-safe. `UsageAggregator` as a whole is **single-writer per instance** (its `FileBackedDict` resource store is not thread-safe) — documented in its docstring; for parallel ingestion, use one aggregator per worker. This matches how the sources drive it today (sequential per source).

Independent of #17705 (FileBackedDict iteration/cache improvements); they touch different regions of `file_backed_collections.py`.

## Test Plan
- [x] `pytest tests/unit/utilities/test_file_backed_counter.py` — **16 passed**: dual-backend (file + in-memory) most-common/tie-break/sentinel/group-order/default-count, cross-flush accumulation, shared-connection-not-closed, **thread-safety** (8 threads), **flush-on-connection-close**, idempotent close, unsafe-tablename rejection.
- [x] `pytest tests/unit/test_usage_common.py` — **26 passed**: semantic-parity vs an in-memory `GenericAggregatedDataset` reference (denied-user, empty events, multi-bucket, equal-`str()` merge, **colons in keys**, non-str resource round-trip), multi-flush parity (`batch_size` 1/3/50000), **dual-backend parity**, top-N tie-break exact match, `_UsageKeys` codec round-trips, shared-connection.
- [x] `pytest tests/unit/sql_parsing/test_sql_aggregator.py` — **31 passed** (incl. `test_basic_usage`, a full-pipeline golden carrying a real `datasetUsageStatistics` aspect generated with the old logic → byte-identical output).
- [x] `pytest tests/unit -k unity` — **281 passed**.
- [x] `pytest tests/integration/sql_queries` — **9 passed**: full `sql_queries` ingestion (usage aspects included) byte-identical to goldens generated with the old in-memory logic.
- [x] `./gradlew :metadata-ingestion:lintFix` clean; mypy clean on changed files.

## Checklist
- [x] PR title format
- [x] Related issue — ING-2749; supersedes #17684
- [x] Tests added/updated
- [ ] Docs — n/a (internal storage change; no recipe-facing config)
- [ ] Breaking changes — none

### DIFF
--- a/metadata-ingestion/src/datahub/utilities/file_backed_collections.py
+++ b/metadata-ingestion/src/datahub/utilities/file_backed_collections.py
@@ -607,7 +607,9 @@ class FileBackedCounter(Closeable):
         self._batch_size = batch_size
         self._closed = False
         # Write buffer, pre-summing repeated keys within a flush window so they collapse
-        # to a single upsert.
+        # to a single upsert. _lock guards _buf so concurrent increment()/_flush() calls
+        # don't lose counts or mutate the dict mid-iteration.
+        self._lock = threading.Lock()
         self._buf: Dict[Tuple[str, str], int] = {}
         self._upsert = (
             f"INSERT INTO {tablename}(group_key, item_key, cnt) VALUES (?, ?, ?) "
@@ -632,14 +634,21 @@ class FileBackedCounter(Closeable):
 
     def increment(self, group_key: str, item_key: str, count: int = 1) -> None:
         key = (group_key, item_key)
-        self._buf[key] = self._buf.get(key, 0) + count
-        if len(self._buf) >= self._batch_size:
+        with self._lock:
+            self._buf[key] = self._buf.get(key, 0) + count
+            over_batch = len(self._buf) >= self._batch_size
+        if over_batch:
             self._flush()
 
     def _flush(self) -> None:
-        if not self._buf:
-            return
-        rows = [(group, item, cnt) for (group, item), cnt in self._buf.items()]
+        # Snapshot and clear the buffer under the lock, then write outside it: the SQL
+        # write is serialized separately by ConnectionWrapper's own lock, and holding
+        # _lock across I/O would needlessly block concurrent increment() calls.
+        with self._lock:
+            if not self._buf:
+                return
+            rows = [(group, item, cnt) for (group, item), cnt in self._buf.items()]
+            self._buf.clear()
         if self._use_sqlite_on_conflict:
             self._conn.executemany(self._upsert, rows)
         else:
@@ -658,7 +667,6 @@ class FileBackedCounter(Closeable):
                         "WHERE group_key = ? AND item_key = ?",
                         (cnt, group, item),
                     )
-        self._buf.clear()
 
     def most_common_by_group(self) -> Iterator[Tuple[str, List[Tuple[str, int]]]]:
         """Yield (group_key, items) for each group in group_key order, where items are
--- a/metadata-ingestion/src/datahub/utilities/file_backed_collections.py
+++ b/metadata-ingestion/src/datahub/utilities/file_backed_collections.py
@@ -573,3 +573,116 @@ class FileBackedList(Generic[_VT], Closeable):
 
     def __del__(self) -> None:
         self.close()
+
+
+class FileBackedCounter(Closeable):
+    """A disk-backed two-level counter: counts items within groups.
+
+    Stores ``(group_key, item_key) -> count`` in SQLite, incrementing via batched
+    upserts, and reads back the most-common items per group. A member of the
+    FileBacked* family, for when the set of counted keys is too large to hold in memory.
+
+    Callers that need a composite group or item encode it into the string key with a
+    separator their prefix cannot contain.
+
+    Unlike FileBackedDict/FileBackedList, FileBackedCounter does NOT register itself in
+    ``ConnectionWrapper._dependent_objects``. When sharing a connection, the owner must
+    therefore call ``close()`` explicitly to flush buffered increments (the
+    ``UsageAggregator`` consumer does this).
+    """
+
+    def __init__(
+        self,
+        *,
+        shared_connection: Optional[ConnectionWrapper] = None,
+        tablename: str = "counter",
+        batch_size: int = 50000,
+    ) -> None:
+        self._conn = shared_connection or ConnectionWrapper()
+        self._owns_connection = shared_connection is None
+        self._tablename = tablename
+        self._batch_size = batch_size
+        self._closed = False
+        # Write buffer, pre-summing repeated keys within a flush window so they collapse
+        # to a single upsert.
+        self._buf: Dict[Tuple[str, str], int] = {}
+        self._upsert = (
+            f"INSERT INTO {tablename}(group_key, item_key, cnt) VALUES (?, ?, ?) "
+            f"ON CONFLICT(group_key, item_key) DO UPDATE SET cnt = cnt + excluded.cnt"
+        )
+        self._use_sqlite_on_conflict = True
+        if sqlite3.sqlite_version_info < (3, 24, 0):
+            # ON CONFLICT was added in SQLite 3.24.0 (2018-06-04).
+            # See https://www.sqlite.org/lang_conflict.html
+            if _get_sqlite_version_override():
+                self._use_sqlite_on_conflict = False
+            else:
+                raise RuntimeError("SQLite version 3.24.0 or later is required")
+        # Always use IF NOT EXISTS: this counter may share a connection or be re-opened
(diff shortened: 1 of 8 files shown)


treff7es:
Follow-up commits (pushed): extracted the counting machinery into a reusable **`GroupedItemCounter` protocol** with two implementations — `FileBackedCounter` (disk, default) and `InMemoryGroupedCounter` (memory) — so `UsageAggregator`'s backend is a clean swap seam:

```python
UsageAggregator(config)                                   # disk-backed (default, unchanged)
UsageAggregator(config, counter=InMemoryGroupedCounter()) # in-memory (small workloads / tests)
```

Both backends are proven interchangeable by a dual-backend parity test against the in-memory `GenericAggregatedDataset` reference (`InMemoryGroupedCounter` works because `Counter.most_common()` shares the same count-desc/insertion-order tie-break as the SQL `cnt DESC, rowid ASC`). The duplicated dim encoding was centralized into a single `_encode_item`/`_decode_item` codec. Production behavior is unchanged (default is file-backed; the swap is a constructor seam, no recipe config).

metadata-ingestion/src/datahub/utilities/file_backed_collections.py
```diff
@@ -573,3 +578,170 @@ def close(self) -> None:
...
+    """
+
+    def increment(self, group_key: str, item_key: str, count: int = 1) -> None: ...
+
+    def most_common_by_group(self) -> Iterator[Tuple[str, List[Tuple[str, int]]]]: ...
+
+    def close(self) -> None: ...
+
+
+class InMemoryGroupedCounter:
```
sgomezvillamor, line 730: nit:
we may rename these implementations of the protocol:
InMemoryGroupedItemCounter
and
FileBackedGroupedItemCounter

metadata-ingestion/src/datahub/utilities/file_backed_collections.py
```diff
@@ -573,3 +578,170 @@ def close(self) -> None:
...
+    When given a shared connection, it registers itself in the connection's dependent
+    objects (like FileBackedDict/FileBackedList), so closing the connection flushes any
+    buffered increments before the connection is torn down.
+    """
+
+    def __init__(
+        self,
+        *,
+        shared_connection: Optional[ConnectionWrapper] = None,
+        tablename: str = "counter",
```
sgomezvillamor, line 602:
```suggestion
        tablename: str = "grouped_item_counter",
```

metadata-ingestion/src/datahub/utilities/file_backed_collections.py
```diff
@@ -573,3 +578,170 @@ def close(self) -> None:
...
+            if _get_sqlite_version_override():
+                self._use_sqlite_on_conflict = False
+            else:
+                raise RuntimeError("SQLite version 3.24.0 or later is required")
+        # Always use IF NOT EXISTS: this counter may share a connection or be re-opened
+        # against an existing table, unlike FileBackedDict which gates table reuse via
+        # allow_table_name_reuse.
+        self._conn.execute(
+            f"CREATE TABLE IF NOT EXISTS {tablename}("
+            "group_key TEXT, item_key TEXT, cnt INTEGER, "
```
sgomezvillamor, line 640: TEXT vs VARCHAR ?

sgomezvillamor (approved):
Solid PR — clean primitive, good swap seam, thorough tests, and the benchmark tells a clear story. A few minor notes:

**🟡 WARNING — `generate_workunits()` usable after `close()`**
Calling `generate_workunits()` after `close()` hits an opaque SQLite error from the closed connection. An early guard (`if self._closed: raise RuntimeError(...)`) would surface a clearer failure.

**ℹ️ SUGGESTION — `InMemoryGroupedCounter.close()` stays usable after close**
Unlike `FileBackedCounter`, the in-memory backend has no `_closed` guard in `close()`, so it silently remains usable afterward. Not a bug, but slightly asymmetric.

**ℹ️ IDEA — 3 (or 4) separate counters instead of key composition**
An alternative design would use `_user_counts`, `_query_counts`, `_column_counts` as separate `GroupedItemCounter` instances, avoiding the `dim:` prefix encoding in `_UsageKeys`. The trade-off: you'd need a 4th structure (an existence counter/set) to handle denied-user events that must still emit an empty workunit, plus 3 sweeps in `generate_workunits()` instead of 1. Worth considering for a follow-up if the prefix encoding ever becomes a maintenance pain point.

---

PR #17721, opened 2026-06-03
fix(ingest/snowplow): use non-deprecated pipelines/v1 enrichments endpoint

treff7es:
## Summary

The Snowplow connector fetched pipeline enrichments from the BDP Console API endpoint `GET organizations/{org}/resources/v1/pipelines/{pipelineId}/configuration/enrichments`, which Snowplow has **deprecated** (confirmed via `deprecated: true` in their OpenAPI spec). This switches to the supported endpoint `GET organizations/{org}/pipelines/v1/{pipelineId}/enrichments` — the same `pipelines/v1` namespace already used by `get_pipeline()`.

## Details

The replacement endpoint returns a different payload shape (`EnrichmentResource`: `name`/`enabled`/`schemaKey`/`userData`) instead of the legacy `EnrichmentDefinitionResource` (`id`/`filename`/`content`/`lastUpdate`). To keep all downstream lineage/PII logic and existing tests unchanged, the new payload is adapted into the internal model via a new `Enrichment.from_resource()` anti-corruption mapping:

- `userData` **is** the enrichment parameter object directly (e.g. `{"geo": {...}}`, `{"pii": [...], "strategy": {...}}`) — it is not wrapped in a `parameters` key. This matches the deprecated endpoint's `content.data.parameters` exactly.
- `vendor` is derived from the Iglu `schemaKey` (`iglu:<vendor>/<name>/...`), since the new payload no longer returns it.
- The new payload has no per-enrichment UUID, so enrichment **DataJob URNs are now derived from the enrichment name** instead of the UUID.

## Breaking change

Enrichment DataJob URNs change from UUID-based (`...,07409eac-...)`) to name-based (`...,campaign-attribution)`). Existing enrichment DataJobs ingested under the old URNs will be orphaned and re-created under the new name-based URNs on the next run. No recipe changes required. Documented in `docs/how/updating-datahub.md`.

## Testing

- Updated unit tests (`test_snowplow_client.py`, `test_snowplow_client_extended.py`) and the integration fixture + golden file to the new shape.
- Full Snowplow unit + integration suite passes (668 tests).
- **Validated end-to-end against the live BDP Console API**: authenticated, listed pipelines, and ran a full file-sink ingestion (0 failures, 0 warnings). The new endpoint returns a superset of the deprecated one (19 vs 17 enrichments), and all enrichment parameters — including PII field configuration used for lineage — are correctly captured.

## Checklist

- [x] Docs updated (breaking change entry in `updating-datahub.md`)


metadata-ingestion/src/datahub/ingestion/source/snowplow/models/snowplow_models.py
```diff
@@ -893,19 +893,74 @@ class EnrichmentContent(BaseModel):
...
 class Enrichment(BaseModel):
-    """Enrichment from BDP API (represents a DataJob in DataHub)."""
+    """Enrichment from BDP API (represents a DataJob in DataHub).
 
-    id: str = Field(description="Enrichment UUID")
+    The internal shape predates the pipelines/v1 API. Instances are built from the
+    non-deprecated endpoint via :meth:`from_resource`; ``id``/``filename`` hold the
+    enrichment ``name`` (the new payload has no UUID) and ``last_update`` is empty
+    (the new payload does not return it).
+    """
```
askumar27, line 902: **ℹ️ SUGGESTION (S1) — `name` is now the DataJob URN key.** Since the payload no longer carries a UUID, `id` (→ `job_id` in `_emit_enrichment_datajob`) is derived from the enrichment `name`. Two latent risks when a URN key moves from an opaque server ID to a natural key:

- **Collision**: two enrichments in one pipeline sharing a `name` collapse into one DataJob URN (last-write-wins), silently dropping one. UUIDs guaranteed this couldn't happen.
- **Rename churn**: renaming an enrichment now orphans its old DataJob and creates a new one (rename-stable before).

Snowplow enrichment names are generally unique-per-pipeline so real-world risk is low, but a one-line dedupe/`report.warning()` on duplicate `job_id` before emit — or at least a comment noting the uniqueness assumption — would make the failure visible rather than silent. (cf. the URN case-sensitivity incident pattern, CUS-8792.)

metadata-ingestion/src/datahub/ingestion/source/snowplow/models/snowplow_models.py
```diff
@@ -893,19 +893,74 @@ class EnrichmentContent(BaseModel):
...
     model_config = ConfigDict(populate_by_name=True)
 
+    @staticmethod
+    def _vendor_from_schema_key(schema_key: str) -> str:
+        """Derive the enrichment vendor from an Iglu schema key.
+
+        Schema keys look like ``iglu:<vendor>/<name>/jsonschema/<version>``. The
+        pipelines/v1 enrichments payload no longer returns a vendor field, so we
+        recover it from the schema key.
+        """
```
askumar27, line 924: **ℹ️ SUGGESTION (S2) — non-string `schemaKey` bypasses the per-item catch.** `get_enrichments` widened its guard to `except (ValidationError, KeyError, TypeError)` so "one bad item doesn't lose all." But if the API ever returns a non-string `schemaKey` (e.g. a number/list), `schema_key.startswith("iglu:")` raises **`AttributeError`**, which isn't in that tuple — so it escapes the per-item handler and aborts the whole enrichment loop, defeating the intent.

Cheap fix — type-guard at the top of this method:
```python
if not isinstance(schema_key, str):
    return ""
```
(or add `AttributeError` to the caught tuple in `snowplow_client.py`).

docs/how/updating-datahub.md
```diff
@@ -114,6 +114,8 @@ Requirements:
 
 - **(Java / Plugin API)** `SecretService.encrypt` and `SecretService.decrypt` no-context overloads are removed; use `encrypt(OperationContext, String)` and `decrypt(OperationContext, String)`. Custom GMS plugins or extensions that call `SecretService` directly must pass `OperationContext` from the request, or `null` for background jobs (allowed by the guard). With `ENFORCE`, `decrypt` throws `SecurityException` for human browser/mobile callers and callers other than trusted ingestion workers (**datahub-actions** in OSS).
 
+- #17721: **(Ingestion / Snowplow)** The connector now fetches pipeline enrichments from the supported `pipelines/v1/{pipelineId}/enrichments` endpoint instead of the deprecated `resources/v1/.../configuration/enrichments` endpoint Snowplow is retiring. Because the new payload no longer returns the per-enrichment UUID, enrichment **DataJob URNs are now derived from the enrichment name** (e.g. `...,campaign-attribution)` instead of `...,07409eac-...)`. Existing enrichment DataJobs ingested under the old UUID-based URNs will be orphaned (left as stale/soft-deleted) and re-created under the new name-based URNs on the next run. No recipe changes are required.
```
askumar27: **ℹ️ SUGGESTION (S3) — make the cleanup precondition explicit.** "orphaned (left as stale/soft-deleted)" is accurate only when stateful ingestion is enabled — the source supports `DELETION_DETECTION` via the stale-entity-removal handler, but users running without `stateful_ingestion` configured will accumulate the old UUID-based DataJobs as permanent phantoms (cf. CUS-8995/8984/8839). Consider: "…re-created under the new name-based URNs; the old UUID-based DataJobs are soft-deleted if stateful ingestion is enabled, otherwise they persist as stale DataJobs."

askumar27 (approved):
Approving — clean, well-scoped fix for the upstream `pipelines/v1` deprecation. The `Enrichment.from_resource()` anti-corruption mapping is the right seam: it absorbs the new payload shape while leaving all downstream lineage/PII/description logic untouched. Breaking change is honestly documented with the PR number, and tests were updated *and* validated end-to-end against the live BDP API.

No blockers. A few non-blocking suggestions inline — worth a look before merge but fine to defer:

- **S1** (models): `name` is now the URN key — worth guarding/noting the per-pipeline uniqueness assumption.
- **S2** (models): the widened per-item catch can still be bypassed by a non-string `schemaKey` (`AttributeError`).
- **S3** (docs): stale-DataJob cleanup is conditional on stateful ingestion being enabled.

Thanks for verifying against the live API — the 19-vs-17 enrichment superset check is exactly the kind of validation that gives confidence here.

PR #17728, opened 2026-06-04
fix(dbt): use dbt adapter type as SQL dialect for column lineage parsing

aviraj-gour:
## Summary

When running the dbt connector with `target_platform: glue` (or any platform whose name is not a valid sqlglot dialect), SQL parsing for column-level lineage fails with `ValueError: Unknown dialect 'glue'`.

The root cause is that `schema_resolver.platform` (the storage platform, e.g. `"glue"`) was being passed as the SQL dialect to sqlglot. These are two different concepts:

- **Storage platform** (`"glue"`) — where the data lives; used for URN construction. Not a SQL dialect.
- **dbt adapter** (`"trino"`) — how SQL is written against those tables; the correct SQL dialect for parsing.

## Changes

- `_parse_cll`: use `node.dbt_adapter` as the SQL dialect for `parse_statements_and_pick`, `detach_ctes`, and `sqlglot_lineage` instead of `schema_resolver.platform`
- `_get_dbt_cte_names`: use `upstream_node.dbt_adapter` for adapter-specific CTE name lookup instead of `schema_resolver.platform`
- `_create_view_properties_aspect`: use `node.dbt_adapter` for SQL formatting instead of `self.config.target_platform`

All three changes fall back to the original platform value if `dbt_adapter` is `None`, preserving existing behaviour for all other setups.

## Root cause

Ingestion report showed `ValueError: Unknown dialect 'glue'` for nodes that had `compiled_code` present, and `sql_parser_successes: 0` across models in a Glue-backed dbt project.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
@@ -2243,7 +2243,8 @@ class DBTSourceBase(StatefulIngestionSourceBase):
                     ]
                     if upstream_node.is_ephemeral_model()
                     for cte_name in _get_dbt_cte_names(
-                        upstream_node.name, schema_resolver.platform
+                        upstream_node.name,
+                        upstream_node.dbt_adapter or schema_resolver.platform,
                     )
                 }
                 if cte_mapping:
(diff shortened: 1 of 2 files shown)


sgomezvillamor (approved):
LGTM

PR #17732, opened 2026-06-04
fix(domains): cross-check OpenSearch results against primary store in hasChildDomains()

javabrett:
## Summary

`DeleteDomainResolver` calls `DomainUtils.hasChildDomains()` before deleting a domain. That method queries OpenSearch via `entityClient.filter()`, which is eventually consistent. When a child domain is deleted and the parent is immediately deleted in the same request cycle, the child's deletion may not yet be indexed in OpenSearch. `hasChildDomains()` then incorrectly returns `true` and rejects the parent delete with `Cannot delete domain <urn> which has child domains`, even though the child no longer exists in the primary store.

This is reproducible in any programmatic workflow that deletes multiple levels of a domain hierarchy in sequence — for example `terraform destroy` on a nested domain tree requires multiple retries (one per hierarchy level) before it succeeds.

## Root Cause

`DomainUtils.hasChildDomains()` calls `entityClient.filter()` which queries the ElasticSearch/OpenSearch index. This index is updated asynchronously via the Kafka MAE pipeline and can lag seconds behind the primary store (MySQL). The lag is the window in which this bug manifests.

## Fix

After fetching candidate child URNs from OpenSearch, cross-check each one against the primary store via `entityClient.filterExistingUrns()`, which delegates to `EntityService.exists()` → `aspectDao.batchGet()` — a single batched MySQL read that is strongly consistent. Stale OpenSearch entries are discarded; deletion is blocked only if at least one child still exists in MySQL.

The filter limit is raised from `1` to `200` so that candidate URNs are available for cross-checking. No new constructor dependencies are required — `EntityClient` already exposes `filterExistingUrns`.

## Tests

- Added `smoke-test/tests/domains/test_domain_delete_race.py`: creates a parent domain, creates a child domain under it, deletes the child, then immediately attempts to delete the parent with no sleep between the two calls. The test uses the underlying HTTP session directly to bypass `TestSessionWrapper`'s consistency sleep, which would otherwise mask the race. Confirmed failing before the fix and passing after against a locally-built quickstart.
- Updated `DeleteDomainResolverTest` to add `testDeleteWithStaleChildDomains`: mocks OpenSearch returning a stale child while the primary store confirms the child no longer exists, and asserts that deletion succeeds.
- Updated existing `testGetSuccess` and `testDeleteWithChildDomains` tests to match the new filter limit (`200`) and to wire the `filterExistingUrns` mock correctly.


david-leifker (approved):
lgtm!

PR #17733, opened 2026-06-04
perf(ingest/omni): remove get_model_yaml + replace throttling with tenacity

sgomezvillamor:
## Summary

Performance optimization and lineage improvements for the Omni connector.

### Performance: Remove `get_model_yaml` calls

The Omni connector called `get_model_yaml` for every model to discover topics and views via YAML parsing — yet it produced zero results. The YAML returned by Omni uses a different format (`base_view:` for topics, comments for views) than what the parser expected (`type: topic` / `type: view`).

All topic data was already being fetched through the `get_topic` API during dashboard tile processing. The YAML path was entirely redundant.

This change removes `get_model_yaml` calls, the YAML parsing code, and the fallback paths that depended on it. Topic discovery now happens exclusively through the dashboard processing stage via `get_topic`.

### Performance: Replace throttling with tenacity
- Removed custom `_throttle()` mechanism and sleep-based concurrency control. The API client now uses tenacity exponential backoff on 429/5xx responses, letting the server control the rate.

### Fix: Document processing order
- Process dashboard documents before workbooks: dashboards discover topics via `get_topic` and populate `_semantic_fields`. Workbook-only documents depend on that data for fine-grained lineage, so they run in a second batch.

### Feature: View → Physical Table column-level lineage
- Each dimension/measure in a semantic view now produces a field-level lineage edge mapping `physical_table.field → semantic_view.field`, gated by the `include_column_lineage` flag.

### Observability improvements
- Added `OmniClientReport` tracking per-method API call counts and timing.
- Replaced generic `filtered` list with `filtered_models` and `filtered_documents`.
- Replaced catch-all `semantic_datasets_emitted` with paired counters: `connections_emitted`, `models_emitted`, `topics_emitted`, `views_emitted`, `dashboards_emitted`, `charts_emitted`.
- Added `view_to_physical_column_lineage_edges` counter.

### Testing
- Strengthened fine-grained lineage test to verify all 4 exact field-level edges.
- Added view → physical column lineage tests (exact edges + flag-off behavior).
- Added 20 unit tests for `omni_lineage_parser` (`extract_field_refs`, `normalize_field_name`, `parse_field_list`).

### Validated against production environment
- Runtime dropped from 2.5 days to 9.5 minutes (~370x faster).
- API calls dropped from ~25k to ~2.3k (11x fewer), with 22k `get_model_yaml` calls eliminated.
- Zero entity regression — all entity counts equal or higher (topics: 508→577, views: 4.5k→5.2k, dashboards: 640→707).
- New view→physical column lineage producing 189 field-level edges (previously 0).
- 6,220 YAML fetch warnings eliminated.

## Test plan

- [x] All 40 tests pass (20 integration + 20 unit)
- [x] Linting passes (`./gradlew :metadata-ingestion:lintFix`)
- [x] Golden file regenerated and stable
- [x] End-to-end run against production Omni instance confirmed zero regressions

## Summary by CodeRabbit

* **New Features**
  * Added parallel Omni ingestion with configurable worker limits for improved performance.
  * Added automatic retries with exponential backoff for rate limits and transient API errors.
  * Added detailed ingestion metrics, API timing information, and stage-based reporting.
  * Improved fine-grained lineage for semantic views, physical tables, dashboards, and charts.
  * Added lowercase dataset URN normalization support.

* **Documentation**
  * Clarified Omni concept mappings, rate limiting behavior, retry handling, and troubleshooting guidance.

* **Bug Fixes**
  * Improved lineage accuracy and handling of filtered models and documents.
  * Removed reliance on model YAML fallback processing.


metadata-ingestion/src/datahub/ingestion/source/omni/omni.py
```diff
@@ -1298,18 +1514,20 @@ def _ingest_documents(self) -> Iterator[MetadataWorkUnit]:
...
+                        chart_ids,
+                        chart_inputs,
+                        chart_titles,
+                        chart_urls,
+                        dashboard_topics,
+                        dashboard_topic_urns,
+                        view_to_topic_urns,
+                    )
                 )
                 model_id_from_dashboard = self._current_tile_model_id
```
treff7es: **Blocker — data race.** `_current_tile_model_id` is a single instance attribute (init at line 170) that `_collect_tile_data` writes (1283/1287) and this line reads back. `_process_document_worker` runs concurrently in the `ThreadPoolExecutor` at line 1652, so another worker can overwrite it between the write and this read — the dashboard gets attributed to the wrong model, or dropped.

Return `model_id` as part of `_collect_tile_data`'s result (a small dataclass alongside the collections it already builds) instead of stashing it on `self`, and delete the attribute.

metadata-ingestion/src/datahub/ingestion/source/omni/omni_report.py
```diff
@@ -6,22 +7,101 @@
...
-    dashboards_scanned: int = 0
-    semantic_datasets_emitted: int = 0
-    physical_datasets_emitted: int = 0
+    dashboards_emitted: int = 0
+    charts_emitted: int = 0
+
+    # Lineage metrics
     dataset_lineage_edges_emitted: int = 0
+    view_to_physical_column_lineage_edges: int = 0
     fine_grained_lineage_edges_exact: int = 0
```
treff7es, line 62: **Blocker — lost-update race.** These three counters are bumped with a raw `+=` inside `_emit_inferred_view_datasets` (omni.py:1401, 1422, 1424, 1426), which runs in the document `ThreadPoolExecutor`. `increment_counter()` right here is the thread-safe path — and it's already used for `views_emitted` ten lines above those raw increments. Route all four through `increment_counter`.

metadata-ingestion/src/datahub/ingestion/source/omni/omni.py
```diff
@@ -137,11 +166,17 @@ def __init__(self, config: OmniSourceConfig, ctx: PipelineContext) -> None:
...
+        self._model_context_by_id: Dict[str, _ModelContext] = {}
         self._current_tile_model_id: Optional[str] = None
 
+        # Lock for source-level collections (URN sets, context dicts, registries)
+        # We use a single lock because:
+        # 1. Lock contention is minimal - most time is spent in I/O (API calls)
+        # 2. Simpler code - no need to reason about lock ordering/deadlocks
+        # 3. Critical sections are small - just set.add() or dict.__setitem__()
+        # If profiling shows lock contention, we can split into per-collection locks
+        self._state_lock = threading.Lock()
```
treff7es: **Blocker — unguarded shared state.** `_state_lock` only guards `_model_dataset_urns` (963, 994). The workers also do check-then-add on `_connection_dataset_urns`, `_topic_ingested_keys`, `_topic_dataset_urns`, `_topic_urn_by_key`, `_semantic_dataset_urns`, `_semantic_fields`, `_physical_dataset_urns`, and `_folder_dataset_urns` with no lock. Under real parallelism that means duplicate emissions, inconsistent maps, and the occasional `set changed size during iteration` (which the broad `except` then downgrades to a warning). Guard these too, or make the caches worker-local and merge after join.

Worth noting the golden test won't catch any of this — the fake client is fast and deterministic and runs with `ignore_order=True`.

metadata-ingestion/src/datahub/ingestion/source/omni/omni.py
```diff
@@ -200,11 +231,42 @@ def _emit_upstream_lineage(
...
+        """Build field-level lineage edges from a semantic view to its physical table.
+
+        For each dimension/measure in the view, creates an edge mapping:
+          physical_table.field_name → semantic_view.field_name
+        """
+        semantic_urn = self._semantic_dataset_urn(model_id, view_name)
+        edges: List[FineGrainedLineageClass] = []
+        for field_name in sorted(view_fields):
+            upstream_field = make_schema_field_urn(physical_urn, field_name)
+            downstream_field = make_schema_field_urn(semantic_urn, field_name)
```
treff7es, line 268: **Correctness — phantom upstream columns.** This maps `physical.<name> → view.<name>` with the same field name on both sides (`seen_fields` is the view-side names). For a computed measure like `total_revenue = SUM(amount)` that emits an upstream `physical.total_revenue` — a column that doesn't exist on the physical table (the real upstream is `amount`). `extract_field_refs(expr)` is already computed for measures at line 790 but never passed here, and `test_view_to_physical_table_column_lineage` asserts the phantom `ORDERS.total_revenue` edge, so the test locks the wrong behavior in.

Suggest gating on confidence: emit the same-name edge only for true passthrough fields (no `dialect_sql`/`display_sql`) and skip/resolve the rest, counting the unresolved ones. Related nit: the plain-dot fallback in `extract_field_refs` also matches things like `3.14` → `FieldRef('3','14')`.

metadata-ingestion/src/datahub/ingestion/source/omni/omni_config.py
```diff
@@ -133,7 +136,8 @@ def validate_base_url(cls, v: str) -> str:
         default=True,
         description=(
             "Upper-case database, schema, and table name components in URNs when the "
-            "resolved platform is Snowflake. Snowflake identifiers are case-insensitive "
-            "and DataHub's Snowflake connector stores them in upper case by default."
+            "resolved platform is Snowflake. Set convert_urns_to_lowercase=True when "
```
treff7es, line 139: This tells users to set `convert_urns_to_lowercase=True` to stitch lineage, but that field (inherited from the newly-added `LowerCaseDatasetUrnConfigMixin`) is never applied to any URN builder — the only place it's read is a `logger.debug` at omni.py:702. The physical-URN path only ever upper-cases via `normalize_snowflake_names` (omni.py:408); there's no lowercasing anywhere. Either wire `convert_urns_to_lowercase` into `_physical_dataset_urn` (see sigma.py for the pattern) or drop this guidance so it doesn't point at an inert flag.

metadata-ingestion/src/datahub/ingestion/source/omni/omni_api.py
```diff
@@ -106,46 +125,82 @@ def paginate_records(
 
     def test_connection(self) -> bool:
```
treff7es: `test_connection` catches everything and returns `False`, and the caller reports a fixed `"API returned an unsuccessful response."` (omni.py:204-208) — so a 401, a 403, a DNS failure and a timeout all look identical to the operator. Let the exception propagate (or return the message) so the outer handler's `failure_reason=str(exc)` surfaces the real cause.

metadata-ingestion/src/datahub/ingestion/source/omni/omni_api.py
```diff
@@ -14,60 +25,68 @@
...
+        """Context manager to track API call timing."""
+        start_time = time.monotonic()
+        try:
+            yield
+        finally:
+            duration = time.monotonic() - start_time
+            if self._report:
+                self._report.record_call(method_name, duration)
+
+    @retry(
```
treff7es, line 60: The predicate only retries `HTTPError` with status in (429,500,502,503,504). Transient `ConnectionError`/`ReadTimeout` — common near rate limits, which is exactly this PR's target scenario — aren't retried, so a single blip drops the item. Also worth honoring the `Retry-After` header on 429s instead of pure exponential backoff. (`reraise=True` and `before_sleep_log` are the right calls.)

metadata-ingestion/setup.py
```diff
@@ -671,7 +671,8 @@
     },
     "flink": {"requests<3.0.0", "tenacity>=8.0.1,<9.0.0"},
     "grafana": {"requests<3.0.0", *sqlglot_lib},
-    "omni": {"requests<3.0.0", "PyYAML>=5.4"},
+    "omni": {"requests<3.0.0", "PyYAML>=5.4", "tenacity>=8.0.1,<9.0.0"},
+    "glue": aws_common | cachetools_lib | sqlglot_lib,
```
treff7es: Duplicate `glue` key — the next line already defines `glue` with `usage_common`, so this one is dead and looks like a rebase artifact (ruff F601 will flag it). Drop it; it's unrelated to Omni anyway.

metadata-ingestion/src/datahub/ingestion/source/omni/omni_config.py
```diff
@@ -57,13 +59,14 @@ def validate_base_url(cls, v: str) -> str:
         le=100,
         description="Number of records per page for paginated Omni API endpoints. Lower values reduce memory usage; higher values speed up ingestion.",
     )
-    max_requests_per_minute: int = Field(
-        default=50,
+    max_workers: int = Field(
```
treff7es, line 62: `max_workers` is the right knob for the new concurrency, but `max_requests_per_minute` was removed outright. `ConfigModel` forbids extra fields, so any existing recipe still setting it will fail validation with no migration path. Worth an entry in docs/how/updating-datahub.md, not just the connector doc.

metadata-ingestion/pyproject.toml
```diff
@@ -783,6 +783,7 @@ okta = [
 omni = [
     "PyYAML>=5.4",
```
treff7es: With the YAML path gone, `yaml` is no longer imported anywhere in the connector — `PyYAML` looks like a dead dependency now (same in setup.py). Safe to drop.

metadata-ingestion/src/datahub/ingestion/source/omni/omni_api.py
```diff
@@ -1,9 +1,20 @@
...
 import requests
+from tenacity import (
+    before_sleep_log,
+    retry,
+    retry_if_exception,
+    stop_after_attempt,
+    wait_exponential,
+)
+
+if TYPE_CHECKING:
```
treff7es: `TYPE_CHECKING` isn't needed here — `omni_report` doesn't import `omni_api`, so there's no cycle to break. A plain top-level import works and matches the connector convention of avoiding `TYPE_CHECKING`.

metadata-ingestion/src/datahub/ingestion/source/omni/omni_report.py
```diff
@@ -6,22 +7,101 @@
...
+    list_documents_time_seconds: float = 0.0
+    get_dashboard_document_time_seconds: float = 0.0
+    get_document_queries_time_seconds: float = 0.0
+    list_folders_time_seconds: float = 0.0
+
+    def __post_init__(self) -> None:
+        # Lock for thread-safe metric updates
+        self._metrics_lock = threading.Lock()
+
+    def record_call(self, method_name: str, duration_seconds: float) -> None:
```
treff7es, line 28: `record_call` resolves fields by string (`getattr(self, f"{name}_calls", 0)` / `setattr`), and the `getattr` default means a mistyped or renamed method silently creates a phantom attribute instead of erroring. With 16 parallel hand-maintained fields that's easy to drift. Collapsing to `Dict[str, _CallMetric]` keyed by method drops the reflection and the bookkeeping. (Snowplow and Fabric each have a near-identical client-metrics class — a shared mixin could be a nice follow-up.)

treff7es (changes requested):

metadata-ingestion/src/datahub/ingestion/source/omni/omni.py
sgomezvillamor: Fixed in a8a4e9f82a — replaced `self._current_tile_model_id` with a `_TileCollectionResult` dataclass returned from `_collect_tile_data`. Workers no longer write/read a shared instance attribute.

metadata-ingestion/src/datahub/ingestion/source/omni/omni_report.py
sgomezvillamor: Fixed in f15d796f24 — all four raw `+=` on report counters now go through `increment_counter()` which holds `_report_lock`.

metadata-ingestion/src/datahub/ingestion/source/omni/omni.py
sgomezvillamor: Fixed in a6b41d5e95 — replaced `_state_lock` with structurally thread-safe collections.

Added `GuardedSet` and `GuardedDict` to `datahub.utilities.guarded_collections`. `GuardedSet.check_and_add(item) -> bool` atomically tests membership and inserts, eliminating the check-then-add race. `GuardedDict` includes `compute_if_absent(key, factory)` for the dict equivalent.

All 7 sets and 4 dicts in `OmniSource` now use guarded variants. `_state_lock` is deleted — thread safety is enforced by the data structures, not developer discipline. 17 unit tests cover both classes including concurrent correctness.

metadata-ingestion/src/datahub/ingestion/source/omni/omni.py
sgomezvillamor: Fixed in b55c9b2b5e — computed measures (fields with SQL expressions like `SUM(amount)`) are now skipped in view→physical column lineage. Only passthrough dimensions (no `expression`) emit same-name edges.

Changes:
- `_build_view_to_physical_lineage` looks up `_SemanticField` and skips fields with non-empty `expression`
- Added `view_to_physical_column_lineage_skipped_computed` report counter
- Added `test_computed_measures_skipped_in_view_physical_lineage` unit test
- Updated `omni_post.md` to document passthrough-only behavior

metadata-ingestion/src/datahub/ingestion/source/omni/omni_config.py
sgomezvillamor: It's mentioned because `convert_urns_to_lowercase` is a workunit processor that may be relevant for having linkable lineage; being handled in the workunit processor, it does not need to be handled in the Omni connector source code itself.

metadata-ingestion/src/datahub/ingestion/source/omni/omni_api.py
sgomezvillamor: Fixed in c68c636414 — `test_connection()` now returns `None` on success and raises on failure, so the caller's `except` block surfaces `str(exc)` as the real `failure_reason` (401, DNS failure, timeout, etc.) instead of a generic message.

metadata-ingestion/src/datahub/ingestion/source/omni/omni_api.py
sgomezvillamor: Fixed in 1dfdede018 — retry predicate now includes `ConnectionError` and `ReadTimeout` (unconditional) alongside the existing `HTTPError` status-code check. This matches the standard `urllib3.Retry` pattern.

Re: `Retry-After` header — skipping for now; production experiments showed Omni's 429 responses don't include it.

metadata-ingestion/setup.py
sgomezvillamor: Fixed in fce6af4440 — removed the duplicate `glue` key. Was a merge artifact; the second entry (with `usage_common`) is the correct one.

metadata-ingestion/src/datahub/ingestion/source/omni/omni_config.py
sgomezvillamor: Fixed in 37fefbe295 — added breaking change entry in `docs/how/updating-datahub.md` for the `max_requests_per_minute` removal.

metadata-ingestion/pyproject.toml
sgomezvillamor: Fixed in 8ea389bcbf — removed `PyYAML` from the `omni` extra in `setup.py` and regenerated `pyproject.toml`, `uv.lock`, and `constraints.txt`.

metadata-ingestion/src/datahub/ingestion/source/omni/omni_api.py
sgomezvillamor: Fixed in 418fede911 — replaced `TYPE_CHECKING` guard with a plain top-level import. No circular dependency exists between `omni_api` and `omni_report`.

metadata-ingestion/src/datahub/ingestion/source/omni/omni_report.py
sgomezvillamor: Fixed in 0fe2ee9f94 — replaced the 16 flat fields + `getattr`/`setattr` with `Dict[str, _CallMetric]` keyed by method name. Adding a new API method no longer requires touching the report class.

metadata-ingestion/docs/sources/omni/README.md
```diff
@@ -6,13 +6,13 @@ The DataHub integration for Omni covers BI entities such as dashboards, charts,
...
+| `Tile`          | [Chart](../../metamodel/entities/chart.md)         | Each query presentation within a dashboard                                                                      |
+| `Topic`         | [Dataset](../../metamodel/entities/dataset.md)     | SubType `"Topic"` — the semantic join graph entry point                                                         |
+| `View`          | [Dataset](../../metamodel/entities/dataset.md)     | SubType `"View"` — semantic layer table with dimensions and measures as schema fields                           |
+| `Workbook`      | [Dataset](../../metamodel/entities/dataset.md)     | SubType `"Workbook"` — unpublished personal exploration document                                                |
+| Warehouse table | Lineage reference only                             | Not emitted as entities; referenced via URN in upstream lineage of Omni Views (e.g. Snowflake, BigQuery tables) |
+| Document owner  | Ownership relationship                             | URN reference only; propagated as `TECHNICAL_OWNER` ownership aspect to Dashboard and Chart entities            |
```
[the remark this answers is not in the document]
sgomezvillamor: Fixed in 228235c40c — the doc was wrong. The code emits `DATAOWNER` in all three places (datasets, dashboards, charts), and that's the correct intent: Omni document owners represent data ownership, not technical ownership. Updated the concept mapping table to match.

sgomezvillamor:
@treff7es great feedback, really helpful! 🙌
I've addressed all your comments, could you please take another look?

metadata-ingestion/src/datahub/ingestion/source/omni/omni.py
```diff
@@ -793,103 +861,150 @@ def _ingest_topic_payload(
...
-                "connectionId": model.get("connectionId") or "",
+                "connectionId": str(connection_id),
                 "createdAt": model.get("createdAt") or "",
                 "updatedAt": model.get("updatedAt") or "",
                 "deletedAt": model.get("deletedAt") or "",
                 "entityType": "model",
             }
+
             model_urn = self._model_dataset_urn(model_id)
             self._model_dataset_urns.add(model_urn)
```
treff7es, line 971: **Nondeterministic base-model emission (race).** The primary model is registered with an unguarded `.add()` here and emitted unconditionally below (1020-1028), while the base model uses `check_and_add` (994). When a model is both in the filtered set *and* another model's `baseModelId` (e.g. `shared-model-1`), whichever worker wins decides whether an extra `inferred=true` base stub gets emitted — a 6th aspect on the same URN, so the output isn't deterministic. Guard the primary emit with `check_and_add` too, or drop model-phase threading (see the comment on 914), which removes the race entirely.

metadata-ingestion/src/datahub/ingestion/source/omni/omni.py
```diff
@@ -1271,18 +1482,20 @@ def _ingest_documents(self) -> Iterator[MetadataWorkUnit]:
...
 
             if folder_id:
-                yield from self._ensure_inline_folder(
-                    folder_id, folder_name, folder_path
+                work_units.extend(
+                    self._ensure_inline_folder(str(folder_id), folder_name, folder_path)
                 )
 
-            document_connection_id = str(document.get("connectionId") or "")
+            document_connection_id = str(document.get("connectionId"))
```
treff7es: **`connection.None` bug.** `str(document.get("connectionId"))` becomes the string `"None"` when the field is missing, which is truthy and slips past the `if not connection_id` guard in `_ensure_connection_dataset`, emitting a junk `connection.None` dataset. Line 1613 and the model path (1062) already use `... or ""` — do the same here: `str(document.get("connectionId") or "")`.

metadata-ingestion/src/datahub/ingestion/source/omni/omni.py
```diff
@@ -1158,8 +1356,7 @@ def _emit_inferred_view_datasets(
             )
             semantic_view_urn = self._semantic_dataset_urn(model_id, field_ref.view)
             view_topic_urns = view_to_topic_urns.get(field_ref.view, set())
-            if semantic_view_urn not in self._semantic_dataset_urns:
-                self._semantic_dataset_urns.add(semantic_view_urn)
+            if self._semantic_dataset_urns.check_and_add(semantic_view_urn):
```
treff7es, line 1367: **Partial / nondeterministic inferred-view schema.** `inferred_fields_by_view` is built incrementally in this loop, but the view is emitted the first time it's seen (this `check_and_add`), so its `schemaMetadata` only carries the fields seen so far. `fields_by_dashboard` is an unordered set, so a genuinely inferred multi-field view gets an incomplete, run-varying field list. Accumulate all fields per view first, then emit. (Masked in tests today because real views always win via the dashboard-first batch.)

metadata-ingestion/src/datahub/ingestion/source/omni/omni_config.py
treff7es, line 62: **Removed config field needs a graceful deprecation.** `max_requests_per_minute` was dropped but there's no `pydantic_removed_field` for it. `ConfigModel` uses `extra=forbid`, so any existing recipe still setting it will hard-fail validation on upgrade rather than log a deprecation. Add `_max_requests_per_minute = pydantic_removed_field("max_requests_per_minute")` (pattern used in mode.py / unity/config.py). The updating-datahub.md note alone doesn't prevent the crash.

metadata-ingestion/src/datahub/ingestion/source/omni/omni.py
```diff
@@ -793,103 +861,150 @@ def _ingest_topic_payload(
...
             self.report.warning(
-                "connections-fetch",
-                f"Failed to fetch Omni connections; proceeding with config overrides only: {exc}",
+                title="Connections fetch error",
+                message="Failed to fetch Omni connections; proceeding with config overrides only",
+                exc=exc,
             )
 
-        for model in self.client.list_models(page_size=self.config.page_size):
+    def _process_model_worker(self, model: Dict[str, Any]) -> List[MetadataWorkUnit]:
```
treff7es, line 922: **Model-phase threading is dead weight.** This worker makes no `self.client.*` calls — it only reads pre-populated context and builds MCPs in memory, which the GIL serializes anyway. The executor at 1107 adds submit/future overhead for no speedup and is what creates the base-model race above. A plain sequential loop is simpler and deterministic. Per the PR's own numbers the 370x came from removing `get_model_yaml`, not from threading.

metadata-ingestion/src/datahub/ingestion/source/omni/omni.py
```diff
@@ -902,82 +1017,142 @@ def _ingest_semantic_model(self) -> Iterator[MetadataWorkUnit]:
...
+        except Exception as exc:
+            self.report.warning(
+                title="Models fetch error",
+                message="Failed to fetch models from Omni API",
+                exc=exc,
+            )
+            return
+
+        # Phase 2: Process filtered models in parallel
+        with ThreadPoolExecutor(max_workers=self.config.max_workers) as executor:
```
treff7es: **Reinvents existing infrastructure.** If parallelism stays, `datahub/utilities/threaded_iterator_executor.py` (`ThreadedIteratorExecutor.process`) already does "parallel workers that *yield* workunits, streamed through a bounded queue" with backpressure + exception propagation — `mode.py` (a comparable BI connector) uses it. This hand-rolled `ThreadPoolExecutor` + `as_completed` builds eager per-worker lists (no memory bound) and is most of why the guarded collections are needed. Converting workers to generators + this util removes both.

metadata-ingestion/src/datahub/utilities/guarded_collections.py
```diff
@@ -0,0 +1,114 @@
+import threading
+from typing import Callable, Dict, Generic, Iterator, List, Optional, TypeVar, overload
+
+K = TypeVar("K")
+V = TypeVar("V")
+
+
+class GuardedSet(Generic[V]):
```
treff7es, line 8: **Single-consumer utility in a shared location.** This is used only by Omni. `datahub/utilities` is for infra with proven cross-connector reuse; the connector's own precedent for shared-state locking is a plain `threading.Lock` on the state object (`OmniSourceReport._report_lock`, and `dataplex_lineage.py:142`). Suggest moving this into the omni package (e.g. `omni/_concurrency.py`) until a second consumer earns the abstraction. Note the docstring's own point — only the compound ops (`check_and_add`/`compute_if_absent`) actually need the lock under the GIL.

metadata-ingestion/src/datahub/utilities/guarded_collections.py
```diff
@@ -0,0 +1,114 @@
...
+    def get(self, key: K) -> Optional[V]: ...
+
+    @overload
+    def get(self, key: K, default: V) -> V: ...
+
+    def get(self, key: K, default: Optional[V] = None) -> Optional[V]:
+        with self._lock:
+            return self._data.get(key, default)
+
+    def compute_if_absent(self, key: K, factory: Callable[[K], V]) -> V:
```
treff7es, line 71: `compute_if_absent` isn't called anywhere in the repo, and `setdefault` (104) is only ever called on plain local dicts, not guarded instances. Trim both to the surface that's actually used.

metadata-ingestion/src/datahub/ingestion/source/omni/omni.py
```diff
@@ -902,82 +1017,142 @@ def _ingest_semantic_model(self) -> Iterator[MetadataWorkUnit]:
...
+
     def _ingest_folders(self) -> Iterator[MetadataWorkUnit]:
+        """Ingest Omni folder hierarchy as DataHub containers.
+
+        NOTE: Folders are processed sequentially (not parallelized) because they have
+        parent/child relationships that must be resolved in order. Parallelizing would
+        require a two-pass approach (collect all, sort by depth, then emit), which adds
+        complexity for minimal benefit given typical folder counts (~10-100).
+        """
         for folder in self.client.list_folders(page_size=self.config.page_size):
```
treff7es: **Only stage without an error guard.** Unlike connections/models/documents, this loop has no `try/except`. If `list_folders` throws, it propagates to `get_workunits_internal`'s handler, which calls `report.failure` and re-raises — aborting the *documents/dashboards* stage. Wrap it in the same warning+return pattern so the least-critical entity can't kill the most-critical stage.

metadata-ingestion/src/datahub/ingestion/source/omni/omni_report.py
```diff
@@ -1,27 +1,90 @@
...
+
+    def __post_init__(self) -> None:
+        super().__post_init__()  # type: ignore
+        # Lock for thread-safe counter updates
+        self._report_lock = threading.Lock()
+        # Ensure client_report is initialized
+        if not isinstance(self.client_report, OmniClientReport):
+            self.client_report = OmniClientReport()
+
+    def increment_counter(self, name: str, delta: int = 1) -> None:
```
treff7es, line 75: **Stringly-typed counter.** `getattr(self, name)` with no default raises `AttributeError` on a typo, and the workers' broad `except Exception` swallows it as a generic "processing error" — a bug that surfaces in the wrong place and is invisible to mypy. Prefer incrementing named fields directly under the existing lock.

metadata-ingestion/src/datahub/ingestion/source/omni/omni.py
```diff
@@ -1041,26 +1242,20 @@ def _ingest_topic_from_dashboard(
...
-                    f"Failed to fetch topic {topic_name} for model {model_id}: {exc}",
-                )
+            self.report.warning(
+                title="Topic fetch from dashboard error",
+                message="Failed to fetch topic from dashboard model",
+                context=f"topic_name={topic_name}, dashboard_model={model_id}",
+                exc=exc,
+            )
 
     def _collect_tile_data(
```
treff7es, line 1271: **11 params, 8 caller-owned collections mutated in place.** This contradicts the docstring ("avoids shared mutable state") and the return-a-dataclass convention. `chart_ids`/`chart_inputs`/`chart_titles`/`chart_urls` are really one `List[ChartTile]`. Build and return the (already-defined) `_TileCollectionResult` instead of writing through the signature.

treff7es (approved):

metadata-ingestion/src/datahub/ingestion/source/omni/omni.py
sgomezvillamor: Fixed in 79e605d932 — changed primary model registration from `.add()` to `.check_and_add()` so both paths use the same atomic guard.

The primary model always emits unconditionally (it has the real properties and overwrites any inferred stub via last-writer-wins). The inferred stub only emits if `check_and_add` returns `True` (no other worker claimed the URN). With `check_and_add` on both paths, the most common race (primary worker B registers before worker A's inferred stub check) is now deterministic: the stub is skipped.

metadata-ingestion/src/datahub/ingestion/source/omni/omni.py
sgomezvillamor: Fixed in 79bd0f4dd5 — added `or ""` so a missing `connectionId` coerces to empty string instead of the truthy `"None"`, matching the model path at line 1062.

metadata-ingestion/src/datahub/ingestion/source/omni/omni_config.py
sgomezvillamor: I do hard removal in INCUBATING connectors

metadata-ingestion/src/datahub/ingestion/source/omni/omni.py
sgomezvillamor: Fixed in 91653c675d — removed model-phase threading entirely and replaced with a plain sequential loop. This also eliminates the base-model race from r3629346902.

The threading was a pre-optimization design that wasn't cleaned up when `get_model_yaml` (the I/O) was removed. With the YAML path gone, model workers make zero `self.client.*` calls — they only read pre-populated context and build MCPs in memory, so the GIL serializes them anyway. The executor was adding submit/future overhead for no speedup.

Document-phase threading is retained since document workers make real HTTP calls (`get_dashboard_document`, `get_topic`, `get_document_queries`).

Re: GuardedSet/GuardedDict — `_model_dataset_urns` no longer strictly needs guarding, but keeping them all as GuardedSet/GuardedDict uniformly is arguably simpler than maintaining a mix of plain and guarded collections. The overhead is negligible.

metadata-ingestion/src/datahub/ingestion/source/omni/omni.py
sgomezvillamor: Fixed in 327e46fac4 (on top of 91653c675d which removed model-phase threading).

**Model-phase threading removed** — this was a pre-optimization design that wasn't cleaned up when the I/O (`get_model_yaml`) was removed. Model workers make no `self.client.*` calls, so the GIL serialized them anyway. Now a plain sequential loop.

**Document-phase switched to `ThreadedIteratorExecutor`** — workers are now generators that yield into a bounded queue (backpressure), consistent with mode.py. In practice a single document produces a small handful of MCPs (dashboard + a few charts + lineage), so there was no real memory pressure from the eager lists, but happy to reuse existing infra.

**GuardedSet/GuardedDict retained** — document workers concurrently discover shared entities (topics, views, folders, connections) that need atomic `check_and_add()` deduplication. `ThreadedIteratorExecutor` solves work distribution, not application-level dedup — these are orthogonal concerns. Keeping them all as GuardedSet/GuardedDict uniformly is arguably simpler than maintaining a mix of plain and guarded collections.

metadata-ingestion/src/datahub/utilities/guarded_collections.py
sgomezvillamor: This is standard, commonly reusable functionality — `GuardedSet.check_and_add` and `GuardedDict` are generic thread-safe primitives that any connector with parallel workers could benefit from. Keeping them in the omni module would prevent other sources from using them, which is why they were placed in `datahub/utilities/`. Follow-up PRs from other connectors can adopt them directly.

metadata-ingestion/src/datahub/utilities/guarded_collections.py
sgomezvillamor: `compute_if_absent` is not used yet but is quite standard functionality and an expected feature for such a class — it's the thread-safe equivalent of `dict.setdefault` with a factory callable (same as Java's `ConcurrentHashMap.computeIfAbsent`). Added for illustrative purposes mainly; happy to remove if the preference is to trim to the used surface only.

metadata-ingestion/src/datahub/ingestion/source/omni/omni.py
sgomezvillamor: Fixed in e3da803953 — wrapped `_ingest_folders` in the same `try/except` + `report.warning` + `return` pattern used by connections, models, and documents. Folder failures no longer abort the documents/dashboards stage.

metadata-ingestion/src/datahub/ingestion/source/omni/omni_report.py
sgomezvillamor: Fixed in 1fc6071fab — `increment_counter` now validates the counter name against the dataclass's own `int` fields (derived via `dataclasses.fields(self)`), so a typo raises `ValueError` with the list of valid counters instead of an `AttributeError` swallowed by the worker's broad `except`. No manual list to maintain — adding a new `int` field to the report automatically makes it a valid counter.

metadata-ingestion/src/datahub/ingestion/source/omni/omni.py
sgomezvillamor: Addressed — moved all 8 mutable collection params into the `_TileCollectionResult` dataclass. `_collect_tile_data` now creates and populates collections internally, returning them via the result object. The caller unpacks from `tile_result.xxx` instead of passing pre-allocated containers.

Signature went from 11 params → 3 (`doc_id`, `dashboard_url`, `dashboard_title`).

8b19dbdf96

metadata-ingestion/src/datahub/ingestion/source/omni/omni.py
sgomezvillamor: Good catch — split into two passes: first accumulate all fields per view, then emit. This ensures `schemaMetadata` always contains the complete field set regardless of set iteration order.

Added `test_inferred_view_schema_contains_all_fields` — a regression test that exercises the inferred path with a non-topic view ("analytics") having 3 fields (`page_views`, `sessions`, `bounce_rate`). The test verifies all 3 fields appear in the emitted schema.

c08b4b36c5

PR #17740, opened 2026-06-04
feat: add output ports module

mihai103:
Hello,

Output ports can be added to Data Products but there is no module to display in the Data Product page.

This PR:
- add a dedicated Output Ports summary module for Data Products in the new dp summary experience
- show Output Ports alongside Assets in the default Data Product summary layout, while hiding the module entirely when no output ports exist - nothing changes for users not using output ports

image


max-datahub:
Thanks for the contribution! Output Ports really do deserve a spot on the Data Product summary page, and the frontend half of this is in good shape — the hook faithfully mirrors the existing `useGetDataProductAssets` pattern, and the module wiring (`Module.tsx`, `LARGE_MODULE_TYPES`, `DEFAULT_MODULE_URNS` on both sides, default-template test update) is all correct.

However, adding a new `DataHubPageModuleType` touches a few more hand-maintained sites across the stack, and two backend pieces are missing that will cause real breakage:

### Blocking

**1. The PDL (storage-layer) enum is missing `OUTPUT_PORTS`.**
The GraphQL enum in `module.graphql` was updated, but not `metadata-models/src/main/pegasus/com/linkedin/module/DataHubPageModuleType.pdl`. The consequence is concrete — `UpsertPageModuleResolver` maps the GraphQL value to the PDL enum via:

```java
com.linkedin.module.DataHubPageModuleType gmsType =
    com.linkedin.module.DataHubPageModuleType.valueOf(type.toString());
```

`valueOf("OUTPUT_PORTS")` throws `IllegalArgumentException`, so any attempt to create/upsert an Output Ports module through the API fails, and the `urn:li:dataHubPageModule:output_ports` entity can never actually be created.

**2. The default module is never bootstrapped, so it silently vanishes after any summary-page edit.**
Every sibling default module (`assets`, `lineage`, `columns`, …) has an entry in `metadata-service/configuration/src/main/resources/bootstrap_mcps/page-modules.yaml`, which creates the module entity in GMS (the file header describes the recipe: add an entry + bump the version in `bootstrap_mcps.yaml`).

Without it, the pure-default rendering works (the client composes the default template with the inlined `OUTPUT_PORTS_MODULE` fragment), but the moment a user personalizes their Data Product summary layout, the template is persisted server-side referencing `urn:li:dataHubPageModule:output_ports` — an entity that doesn't exist. On the next load, hydration of that module returns nothing and the Output Ports module disappears from that user's page. This is exactly why `urn:li:dataHubPageModule:assets` is bootstrapped.

**Fix for both:** add `OUTPUT_PORTS` to the PDL enum, add the bootstrap YAML entry, and bump the bootstrap version — the same recipe the other nine default modules follow.

### Non-blocking suggestions

- **Stale data on output-port changes:** existing flows refresh the Assets module via `reloadByKeyType(MODULE, DataHubPageModuleType.Assets)` (e.g. `SetDataProductModal`) when data product membership changes, but nothing fires a reload for `DataHubPageModuleType.OutputPorts`, and the hook uses `cache-first`. Marking/unmarking an output port won't refresh the module until remount — worth wiring the same reload key where output-port status is toggled.
- **`PageModuleServiceTest.testDeletePageModuleDefaultModulePrevention`** keeps its own hardcoded URN list that wasn't extended with `output_ports`. The test still passes (it loops over its own list), but the delete-prevention guard silently doesn't cover the new value.
- **Scope noise:** the `RowLayout.tsx` change is whitespace-only and could be dropped.
- **Minor UX:** the module returns `null` while the first fetch is in flight (`total` is 0 until data arrives), so it pops in after load rather than showing a loading state. Consistent with the hide-when-empty intent, just a visible layout shift to be aware of.
- **Test coverage:** the component tests are reasonable, but `useGetOutputPorts` itself is untested — the sibling `useGetAssets.test.ts` is a good template.

Happy to help iterate on the backend pieces if useful!

chriscollins3456 (commented):
nice! yeah the biggest two comments are points 1 and 2 that max makes above with needing more backend wiring to bootstrap this new module and support the new type in PDL

so this module obviously just shows you output ports. it doesn't support adding/removing output ports right?

max-datahub:
Thanks for the quick turnaround — your latest commit resolves both blocking items from my earlier review (the PDL enum value and the bootstrap entry + version bump). 👍

I pushed one follow-up commit (`667a30d`) on top to green CI and tighten backend coverage:

- **CI schema-version check (was failing):** `DataHubPageModuleProperties` references the `DataHubPageModuleType` enum, so adding `OUTPUT_PORTS` requires bumping the aspect's `schemaVersion` (v1 → v2). Applied via `.github/scripts/bump_schema_versions.py`; `--check` now passes.
- **Delete-prevention coverage:** the production `PageModuleService.DEFAULT_MODULE_URNS` already includes `output_ports` (👍), but `PageModuleServiceTest` keeps a separate hardcoded list that didn't, so the guard wasn't actually exercised for the new default module. Added it.
- **Hook test:** added `useGetOutputPorts.test.ts` (covers the `isOutputPort=true` filter, pagination via refetch, and loading state). The component test already existed; the hook itself was the uncovered file Codecov flagged.
- **Minor:** dropped a stray blank line in `RowLayout.tsx`.

Local gates green: `:metadata-service:services:test` (17 passing), the new vitest suite (4/4), `datahub-web-react` lint, and Java spotless.

### End-to-end verification (fresh quickstart)

Rebuilt and booted a clean instance from this branch and verified the full flow:

| Check | Result |
|---|---|
| Model codegen after the `schemaVersion` bump | ✅ up-to-date — no stale generated/committed artifact |
| Bootstrap creates `urn:li:dataHubPageModule:output_ports` | ✅ `name:"Output Ports"`, `type:OUTPUT_PORTS`, `scope:GLOBAL` (confirmed via GraphQL) |
| PDL enum through the write path | ✅ `upsertPageModule(type: OUTPUT_PORTS)` gets past `DataHubPageModuleType.valueOf("OUTPUT_PORTS")` with no enum crash |
| Default-module handling | ✅ that upsert is then correctly rejected by `validateInput` as a default module — same treatment as `ASSETS`; default modules come from bootstrap + the default template, not user upserts |

Net: since users can't create this module type via the API, the bootstrapped global entity is the only source of it, so a personalized template referencing `urn:li:dataHubPageModule:output_ports` now resolves instead of dangling — the scenario the bootstrap entry protects against.

### Non-blocking follow-ups

- `PageModuleServiceTest` maintains a list parallel to the production `DEFAULT_MODULE_URNS` (it's also missing `platforms`/`lineage`); I kept my change surgical, but pointing the test at the production constant would prevent this drift going forward.
- Optional: a quick browser pass (personalize a Data Product summary → reload) for visual confirmation; the mechanism it relies on is now verified.

datahub-web-react/src/app/homeV3/template/components/addModuleMenu/useAddModuleMenu.tsx
```diff
@@ -78,6 +78,17 @@ export const ASSETS_MODULE: PageModuleFragment = {
     },
 };
 
+export const OUTPUT_PORTS_MODULE: PageModuleFragment = {
+    urn: 'urn:li:dataHubPageModule:output_ports',
+    type: EntityType.DatahubPageModule,
+    properties: {
+        name: 'Output Ports',
```
chriscollins3456: we should translate this like our other module names as well

chriscollins3456 (approved):
this is looking good! request to translate a string but not blocking me approving

datahub-web-react/src/app/homeV3/template/components/addModuleMenu/useAddModuleMenu.tsx
max-datahub: linter missed that, thanks for pointing it out!

PR #17741, opened 2026-06-04
feat(i18n): extract keys from entity v1 shared containers

v-tarasevich-blitz-brain:
This PR extracts i18n keys from entity/shared/containers


datahub-web-react/src/i18n/locales/en/entityV1.shared.containers.json
```diff
@@ -0,0 +1,57 @@
+{
```
benjiaming: Do we actually need version in the filename?

benjiaming (approved):

datahub-web-react/src/i18n/locales/en/entityV1.shared.containers.json
v-tarasevich-blitz-brain: removed

PR #17746, opened 2026-06-04
feat(ui): rebuild Alchemy Alert layout, add actionPlacement, errorMessage, and Storybook

annadoesdesign:
Screenshot 2026-06-04 at 10 54 39 AM
Screenshot 2026-06-04 at 10 54 52 AM
Screenshot 2026-06-04 at 10 54 45 AM
Screenshot 2026-06-04 at 10 54 42 AM

## Summary

Rebuilds the Alchemy `Alert` component into a column layout where the
description and other body content span the full container width, independent
of what's on the right side of the header. Previously the description was
constrained by whatever sat to its right (close button + any actions), which
caused awkward wrapping for longer text.

Also adds several capabilities to bring the OSS Alert in line with how it's
already being used downstream, plus first-time Storybook coverage.

### New layout

```
[icon] Title                         [topRight action] [X]
       Description spans the full container width…
       [errorMessage]
       [inline action]
```

- Header row: icon + title on the left, optional `topRight` action + close X on the right.
- Body section (description, `errorMessage`, inline action) is indented 28px so it aligns flush with the title text rather than the icon's left edge.

### New props / variants

- `actionPlacement?: 'inline' | 'topRight'` — where to render the optional `action` element. Defaults to `'inline'` (under the description). `'topRight'` places it next to the close X without sacrificing description width.
- `errorMessage?: string` — technical error detail rendered in a monospace block beneath the description.
- `unknown` variant — neutral surface + Question icon for indeterminate states.
- `data-testid` is now forwarded to the container so test selectors work.

### Polish

- Variant-matched colors for in-Alert buttons via `VARIANT_BUTTON_COLOR_MAP` (close X reads as the same hue as the banner).
- Description bumped from 12px → 14px to match the title size.
- `AlertActions` is pulled left by 12px so the inline button label sits flush with the description while the button itself keeps its normal hit-target padding.
- `AlertBody` pulls `-4px` to tighten the title → description leading.

### Side fix

`Button.tsx` now imports `Icon` via the specific path
(`@components/components/Icon`) instead of the barrel (`@components`). This
breaks a circular-dependency chain (`Alert` → `Button` → `@components` →
`FileDragAndDropArea` → `styled(Button)`) that previously caused
`Cannot create styled-component for component: undefined` during testing.

### Backward compatibility

Public API is fully additive — every existing prop (`variant`, `title`,
`description`, `icon`, `action`, `onClose`, `className`, `style`) keeps the
same behavior. No consumer changes required.

### Storybook

First-time Storybook coverage for `Alert`:
- `sandbox` (interactive playground with toggles for every prop, including `actionPlacement`)
- `variants`, `dismissible`, `withAction`, `topRightAction`, `withErrorMessage`, `customIcon`, `titleOnly`, `kitchenSink`

## Checklist

- [x] PR conforms to the [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (PR Title Format: `feat(ui): …`)
- [x] Tests for the changes have been added/updated (16/16 passing in `Alert.test.tsx`, including new coverage for `errorMessage` rendering, `unknown` variant, and `actionPlacement="topRight"`)
- [x] Storybook coverage added (`Alert.stories.tsx`)
- [x] No breaking changes — public API is fully backward compatible

## Test plan

- [x] `tsc --noEmit -p .` clean
- [x] `eslint` clean (only the pre-existing `react-refresh/only-export-components` warning, consistent with the existing `Button.tsx` `buttonDefaults` co-export pattern)
- [x] `vitest run Alert.test.tsx` — 16/16 pass
- [x] `prettier --check` clean on all touched files
- [ ] Verify Storybook renders all stories locally (`yarn storybook`)
- [ ] Visual regression: confirm long-description case wraps correctly with topRight action + close button present

### DIFF
--- a/datahub-web-react/src/alchemy-components/components/Alert/components.ts
+++ b/datahub-web-react/src/alchemy-components/components/Alert/components.ts
@@ -2,7 +2,7 @@ import styled from 'styled-components';
 
 import { AlertVariant } from '@components/components/Alert/types';
 
-import { radius, spacing } from '@src/alchemy-components/theme';
+import { radius, spacing, typography } from '@src/alchemy-components/theme';
 
 const VARIANT_THEME_MAP: Record<AlertVariant, { bg: string; text: string; icon: string }> = {
     success: { bg: 'bgSurfaceSuccess', text: 'textSuccess', icon: 'iconSuccess' },
@@ -99,7 +99,7 @@ export const AlertErrorMessage = styled.div(({ theme }) => ({
     backgroundColor: theme.colors.bg,
     borderRadius: radius.md,
     padding: `${spacing.xxsm} ${spacing.xsm}`,
-    fontFamily: "'SF Mono', 'SFMono-Regular', Consolas, 'Liberation Mono', Menlo, Courier, monospace",
+    fontFamily: typography.fonts.mono,
     fontSize: '13px',
     overflowWrap: 'break-word' as const,
     wordBreak: 'break-all' as const,
--- a/datahub-web-react/src/alchemy-components/components/Alert/components.ts
+++ b/datahub-web-react/src/alchemy-components/components/Alert/components.ts
@@ -10,19 +10,67 @@ const VARIANT_THEME_MAP: Record<AlertVariant, { bg: string; text: string; icon: 
     warning: { bg: 'bgSurfaceWarning', text: 'textWarning', icon: 'iconWarning' },
     info: { bg: 'bgSurfaceInfo', text: 'textInformation', icon: 'iconInformation' },
     brand: { bg: 'bgSurfaceBrand', text: 'textBrand', icon: 'iconBrand' },
+    unknown: { bg: 'bgSurface', text: 'textSecondary', icon: 'icon' },
 };
 
-export const AlertContainer = styled.div<{ $variant: AlertVariant }>(({ $variant, theme }) => {
-    const tokens = VARIANT_THEME_MAP[$variant];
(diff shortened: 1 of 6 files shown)


datahub-web-react/src/alchemy-components/components/Alert/components.ts
```diff
@@ -31,25 +79,28 @@ export const AlertIconWrapper = styled.span<{ $variant: AlertVariant }>(({ $vari
...
     flexShrink: 0,
     whiteSpace: 'nowrap',
+    marginLeft: '-12px',
 });
+
+export const AlertErrorMessage = styled.div(({ theme }) => ({
+    backgroundColor: theme.colors.bg,
+    borderRadius: radius.md,
+    padding: `${spacing.xxsm} ${spacing.xsm}`,
+    fontFamily: "'SF Mono', 'SFMono-Regular', Consolas, 'Liberation Mono', Menlo, Courier, monospace",
```
benjiaming: Shouldn't we use `typography.fonts.mono ` from `@components/theme`?

benjiaming (approved):

datahub-web-react/src/alchemy-components/components/Alert/components.ts
annadoesdesign: Good call — switched to `typography.fonts.mono` in f730d9eab23. Thanks!

PR #17759, opened 2026-06-05
feat(ingestion): reusable OAuth token providers for DataHub Python clients

max-datahub:
## Summary

Adds a small, reusable **OAuth token-provider layer** to the Python SDK so DataHub clients (CLI, SDK, ingestion) can authenticate to GMS with a **fresh, short-lived OAuth JWT per request** instead of a static token.

GMS already validates externally-issued OAuth JWTs as a resource server, so **no GMS change is required** — the gap was client-side, where the `Authorization` header is frozen into the HTTP session at construction.

> Draft: the design is still under internal discussion and may be reworked; opening early for visibility/feedback.

## What's included

- `datahub.emitter.token_provider`: a `TokenProvider` interface, `StaticTokenProvider` (back-compat), `CachingTokenProvider` (refresh-before-expiry, thread-safe), and `TokenProviderAuth` — a `requests.auth.AuthBase` that re-evaluates the token **per request** and retries once on a 401.
- `DataHubRestEmitter` accepts an injectable `auth`; the existing static `token=` path is unchanged (backward compatible).
- Pluggable providers via a `datahub.token_provider.plugins` entry-point group:
  - `k8s_oidc` — Kubernetes projected service-account token (cluster OIDC; stdlib only)
  - `azure_entra` — Entra token via workload identity or client secret (new `azure-auth` extra)
  - `oidc_client_credentials` — generic OIDC client-credentials grant (Okta, Keycloak, any OIDC)
- Declarative `auth: {type, config}` on `DatahubClientConfig` (mutually exclusive with `token`), resolved in `DataHubGraph`.
- Unit tests for all of the above + a local Keycloak-based OAuth smoke test under `smoke-test/tests/oauth/`.

## How it works

A client's job is to obtain a JWT from an issuer GMS trusts and present it. This adds a pluggable way to acquire that token (per cloud / IdP / grant) with per-request freshness + refresh, behind one declarative config. Adding a new provider is a small class implementing `get_token()` plus an entry point — no SDK or consumer change.

## Testing

- Unit tests under `metadata-ingestion/tests/unit/auth/` (interface, caching/refresh, `AuthBase` + 401 retry, emitter wiring, registry resolution, all three providers, declarative config).
- A docker-compose + Keycloak smoke test (`smoke-test/tests/oauth/`) validating that a real client-credentials token is **accepted** (correct audience) and **rejected** (wrong audience) by GMS.

## Checklist

- [x] Conventional Commit title
- [x] Tests added (unit + smoke)
- [ ] Docs — follow-up
- [x] Backward compatible (static `token=` unchanged)

### DIFF
--- a/metadata-ingestion/pyproject.toml
+++ b/metadata-ingestion/pyproject.toml
@@ -148,6 +148,10 @@ aws-secret-manager = [
 
 azure-ad = []
 
+azure-auth = [
+    "azure-identity>=1.21.0,<2.0.0",
+]
+
 azure-data-factory = [
     "azure-identity>=1.21.0,<2.0.0",
     "azure-mgmt-datafactory>=9.0.0,<10.0.0",
@@ -1962,6 +1966,11 @@ debug = [
 
 [project.scripts]
 datahub = "datahub.entrypoints:main"
+
+[project.entry-points."datahub.token_provider.plugins"]
+k8s_oidc = "datahub.ingestion.auth.k8s_projected:K8sProjectedTokenProvider"
+azure_entra = "datahub.ingestion.auth.azure_entra:AzureEntraTokenProvider"
+oidc_client_credentials = "datahub.ingestion.auth.oidc_client_credentials:OidcClientCredentialsTokenProvider"
 
 [project.entry-points."sqlalchemy.dialects"]
 "doris.pymysql" = "datahub.ingestion.source.sql.doris.doris_dialect:DorisDialect"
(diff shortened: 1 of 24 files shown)


max-datahub:
Superseded by #17939, which carries the same work on a branch in this repository (no personal fork), rebased onto current master so it builds on #17868, plus a review refinement (token expiry is now taken from the provider's response rather than decoded from the JWT). Closing in favor of #17939.

PR #17761, opened 2026-06-05
feat(ingest/tableau): ingest Initial SQL as lineage and custom property

treff7es:
## Summary

Adds opt-in ingestion of a published Tableau data source's **Initial SQL** (the SQL Tableau runs when first opening a connection).

> **Stacked on #17807** (tokenizer-based `split_statements` rewrite + `#<digit>` T-SQL temp-table parse fix). This PR is based on that branch and targets it; it will retarget to `master` automatically once #17807 merges. Review #17807 first.

Initial SQL is not exposed by the Tableau REST `connections` endpoint, the `tableauserverclient` `ConnectionItem` model, or the GraphQL Metadata API — it only exists in the authored data source definition as the `one-time-sql` attribute of each `<connection>` in the downloaded `.tds`/`.tdsx`. This PR therefore downloads each published data source (opt-in, because it is one REST download per data source), parses `one-time-sql`, and:

- emits the tables referenced in the Initial SQL as **upstream lineage** of the data source (reusing the existing Custom SQL parsing pipeline: `get_overridden_info` → `create_lineage_sql_parsed_result` → `make_upstream_class`), merged de-duplicated into the data source's single `UpstreamLineage` aspect; and/or
- stores the raw Initial SQL text as the `initialSql` **custom property**.

Initial SQL is frequently multiple statements with no `;` separators (idiomatic T-SQL), which is why this work motivated the shared splitter rewrite in #17807; the `#<digit>` temp-table fix there is what lets these statements recover their real `FROM` source on SQL Server.

### Config (all on the `tableau` source)

- `ingest_initial_sql` (default `false`) — master switch; disabled by default due to the per-data-source download cost.
- `initial_sql_as_lineage` (default `true`)
- `initial_sql_as_custom_property` (default `true`)

### Notes / scope

- Published data sources only; embedded data sources are not supported (no standalone download).
- XML parsed with `defusedxml`. Download/corrupt-zip/malformed-XML/unparseable-SQL all warn-and-skip with report counters (`num_initial_sql_*`); nothing crashes ingestion.
- No behavior change when `ingest_initial_sql` is false (the default).

## Test Plan

- [x] New unit tests in `tests/unit/tableau/test_tableau_initial_sql.py` cover the pure parser (happy/empty/malformed/corrupt-zip/no-tds), the orchestration (`get_initial_sql_lineage`: emit/no-luid/download-failure/lineage-disabled), and `emit_datasource` integration (asserts one merged `UpstreamLineage` + the `initialSql` custom property).
- [x] Full Tableau unit suite green (`pytest tests/unit/tableau/` — 266 passed).
- [x] `./gradlew :metadata-ingestion:lint` passes (ruff + mypy).
- [ ] Live end-to-end against Tableau Cloud (golden) tracked as a follow-up in the connector-tests repo.

## Checklist

- [x] Tests added/updated
- [x] Docs added/updated (`metadata-ingestion/docs/sources/tableau/tableau_post.md`)
- [x] No breaking changes (feature is opt-in, off by default)


sgomezvillamor (approved):
Nice feature, clean implementation. A few notes:

---

**⚠️ `_TWO_TIER_PLATFORMS` inconsistency (pre-existing, but now active on both paths)**

The new constant was correctly extracted from the existing inline check in `get_fully_qualified_table_name` — no regression there. However, the two lists that govern two-tier behaviour in `tableau_common.py` now disagree:

```python
# _TWO_TIER_PLATFORMS (drives _normalize_parsed_upstream_urn / _dataset_name_max_parts)
("athena", "hive", "mysql", "clickhouse")

# get_overridden_info (sets upstream_db = None)
("athena", "hive", "mysql", "teradata")
```

`clickhouse` is treated as two-tier for name trimming but **not** for `upstream_db` nulling; `teradata` is the reverse. With this PR, both lists run in sequence on the Initial SQL path, so for Teradata a SQL-parsed reference like `db.schema.table` (3 parts, since `max_parts=3`) won't match the GraphQL-derived URN where `upstream_db` was nulled. Suggest unifying both into a single constant in `tableau_common.py` before or alongside merge.

---

**📋 Follow-up / out of scope for this PR**

Two items worth a dedicated follow-up:

1. **Centralize two-tier platform definitions across the codebase.** `_TWO_TIER_PLATFORMS` variants exist independently in at least `looker/view_upstream.py`, `matillion_dpc/constants.py`, `sql/trino.py`, and now `tableau_common.py`, and they all disagree on membership. A single authoritative set (probably in `sql_parsing_common.py` or a new `platform_utils.py`) would eliminate the drift.

2. **Platform-specific URN builders.** `_normalize_parsed_upstream_urn` is the right short-term fix — it mirrors `get_fully_qualified_table_name`'s truncation for the SQL-parsed path and the comment explains it well. Longer term, per-platform URN builder classes would remove the need for this kind of post-hoc normalization scattered across connectors.

---

Everything else looks solid: `defusedxml` for XML security, `TemporaryDirectory` + `include_extract=False` keep disk impact minimal, workbook download caching is correct (including negative caching on failure), and test coverage is thorough.

---

PR #17772, opened 2026-06-05
chore(sql-parsing): remove redundant unnest_subqueries null-predicate patch

aviraj-gour:
## Summary

- Removes `_patch_unnest_subqueries()` from `_sqlglot_patch.py`, which guarded two `_replace(parent_predicate, ...)` call sites against `parent_predicate` being `None`.

The patch targets the last for loop in decorrelate, specifically the part after if key in group_by.

original sqlglot

```
if key in group_by:
    key.replace(nested)
elif isinstance(predicate, exp.EQ):
    parent_predicate = _replace(
        parent_predicate,
        f"({parent_predicate} AND ARRAY_CONTAINS({nested}, {column}))",
    )
else:
    key.replace(exp.to_identifier("_x"))
    parent_predicate = _replace(
        parent_predicate,
        f"({parent_predicate} AND ARRAY_ANY({nested}, _x -> {predicate}))",
    )

```

with patch applied
```

if key in group_by:
    key.replace(nested)
elif isinstance(predicate, exp.EQ):
    if parent_predicate:                          # <-- guard added
        parent_predicate = _replace(
            parent_predicate,
            f"({parent_predicate} AND ARRAY_CONTAINS({nested}, {column}))",
        )
else:
    key.replace(exp.to_identifier("_x"))
                                                  # <-- blank line added
    if parent_predicate:                          # <-- guard added
        parent_predicate = _replace(
            parent_predicate,
            f"({parent_predicate} AND ARRAY_ANY({nested}, _x -> {predicate}))",
        )

```

## Why it's safe to remove

sqlglot v30.0.0 (https://github.com/tobymao/sqlglot/pull/7300) introduced an early return in `decorrelate()`:
```python
if parent_predicate is None and not is_subquery_projection:
    return
```
Combined with an existing continue in the loop [for the is_subquery_projection=True path](https://github.com/tobymao/sqlglot/blob/22ee9336e2f198ebea38640caa59f8f551fb1982/sqlglot/optimizer/unnest_subqueries.py#L288), parent_predicate can never be None when reaching the elif/else branches in v30+. We're on v30.8.0.

Note: Our patch and sqlglot's fix are not the same approach — ours was a broad call-site guard, theirs closes the gap through two separate mechanisms. They produce equivalent safety for all reachable paths in v30+.

History
The patch originated in the [hsheth2/sqlglot fork (PR)](https://github.com/hsheth2/sqlglot/pull/2) before sqlglot had any fix. It was ported into _sqlglot_patch.py in #11693 when we migrated from the fork to mainline sqlglot + patchy.

### DIFF
--- a/metadata-ingestion/src/datahub/sql_parsing/_sqlglot_patch.py
+++ b/metadata-ingestion/src/datahub/sql_parsing/_sqlglot_patch.py
@@ -8,7 +8,6 @@ import sqlglot
 import sqlglot.expressions
 import sqlglot.lineage
 import sqlglot.optimizer.scope
-import sqlglot.optimizer.unnest_subqueries
 
 from datahub.utilities.is_pytest import is_pytest_running
 from datahub.utilities.unified_diff import apply_diff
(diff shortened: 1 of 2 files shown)


sgomezvillamor (approved):
LGTM

PR #17774, opened 2026-06-05
ci(playwright): increase timeout for fork PRs on ubuntu-latest runners

priyabratadas-dh:
## Summary

Fork PRs (community contributions) run on `ubuntu-latest` without a warm Yarn or Chromium cache. The cold start — yarn install + browser download + Docker quickstart — reliably exceeds the 20-minute `timeout-minutes` budget, causing **all Playwright shards to be cancelled before any tests run**.

Observed concretely on PR #17668 (`fix/search-skip-pagination-truncation`): all 5 shards hit the 20-minute wall, produced zero artifacts, and required a maintainer to:
1. Add the `depot` label to opt the fork into Depot runners
2. Push a dummy commit to re-trigger CI

This PR eliminates that manual intervention for the cold-start case.

### Changes

- Emit a `playwright_timeout` output from the `set-runner` step:
  - **20 min** — internal PRs and fork PRs with the `depot` label (Depot runners, warm cache)
  - **45 min** — fork PRs on `ubuntu-latest` (cold start needs headroom)
- Expose `playwright_timeout` as a `setup` job output
- Use `${{ fromJson(needs.setup.outputs.playwright_timeout || '20') }}` as `timeout-minutes` in `playwright_test`

No change in behaviour for internal PRs or `depot`-labelled fork PRs.

## Test plan

- [ ] Verify a fork PR without the `depot` label gets `timeout-minutes: 45` (check the `setup` job output in the Actions log)
- [ ] Verify an internal PR still gets `timeout-minutes: 20`
- [ ] Verify a fork PR with the `depot` label still gets `timeout-minutes: 20`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #17775, opened 2026-06-05
ci(playwright): increase timeout for non-Depot runners in OSS workflow

priyabratadas-dh:
## Summary

The standalone `playwright-e2e-tests.yml` workflow has the same `timeout-minutes: 20` problem as the main CI pipeline: on non-Depot runners (`ubuntu-latest` / `ubuntu-24.04`) a cold start — yarn install + Chromium download + Docker quickstart — can exceed the budget before any tests run, producing zero artifacts.

Companion to #17774, which applied the same fix to `docker-unified.yml`.

### Changes

- In the `Select runner` step, emit `playwright_timeout`:
  - **20 min** when `depot-config-set == true` (Depot runners available, warm cache)
  - **45 min** otherwise (cold start on `ubuntu-latest` needs headroom)
- Expose `playwright_timeout` as a `setup` job output
- Use `${{ fromJson(needs.setup.outputs.playwright_timeout || '20') }}` as `timeout-minutes` in `playwright_test`

No behaviour change when Depot is configured.

## Test plan

- [ ] Confirm a run on a non-Depot environment gets `timeout-minutes: 45` (check `setup` job output in the Actions log)
- [ ] Confirm a Depot-backed run still sees `timeout-minutes: 20`

### DIFF
--- a/.github/workflows/playwright-e2e-tests.yml
+++ b/.github/workflows/playwright-e2e-tests.yml
@@ -28,6 +28,7 @@ jobs:
     runs-on: ubuntu-slim
     outputs:
       test_runner_type: ${{ steps.set-runner.outputs.test_runner_type }}
+      playwright_timeout: ${{ steps.set-runner.outputs.playwright_timeout }}
       yarn_cache_key: ${{ steps.yarn-cache-key.outputs.yarn_cache_key }}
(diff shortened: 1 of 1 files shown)


priyabratadas-dh:
Closing in favour of a non-fork PR

PR #17781, opened 2026-06-05
fix(ingestion/json-schema): prevent combinatorial explosion with recursive $ref + oneOf

### LINKED ISSUE #14358: JSON Schema Ingestion Fails on $ref Loop in Definitions
**Describe the bug**
When ingesting a JSON Schema using the `json-schema` source type, the ingestion process fails to detect and handle $ref loops correctly. This results in an infinite loop during import, and the process never completes.

**To Reproduce**
Run an ingestion of recipe for source type `json-schema` with the following stripped down schema:

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "$id": "https://schema.com/schema/1.0",
  "type": "object",
  "properties": {
    "test": {
      "$ref": "#/definitions/condition"
    }
  },
  "required": [
    "test"
  ],
  "definitions": {
    "condition": {
      "type": "object",
      "properties": {
        "condition": {
          "$ref": "#/definitions/condition"
        }
      }
    }
  }
}
```

**Expected behavior**
The ingestion process should detect the `$ref` loop and either:

- Resolve it safely (e.g., by limiting recursion depth), or
- Fail gracefully with a clear error message indicating the loop.

**Actual behavior**
The ingestion process enters an infinite loop and does not complete.

- Docker Image: acryldata/datahub-ingestion:v1.1.0
- DataHub version: v1.2.0

**Additional context**
This kind of $ref loop is valid in some schema use cases (e.g., recursive structures), and many JSON Schema parsers handle it by tracking visited references. It would be helpful if DataHub could implement similar handling or provide guidance on supported schema patterns.

rospe:
## Checklist

- [x] PR conforms to the [Contributing Guideline](https://datahubproject.io/docs/contributing)
- [x] Links to related issue: Fixes #14358
- [x] Tests added/updated
- [ ] Docs added/updated (N/A — internal fix, no user-facing config changes)

## Summary

JSON Schema ingestion hangs or takes minutes/hours on schemas with recursive `$ref` references combined with `oneOf`/`anyOf`/`allOf` constructs. The existing recursion detection (string comparison) fails when `_retain_parent_schema_props_in_union` merges parent properties into each `oneOf` branch, creating new dicts with different string representations while the underlying `$ref` target is the same definition.

**Fix:** Add object-identity-based recursion detection to `FieldPath`. The `jsonref` library guarantees that all proxy objects pointing to the same `$ref` target share the same `__subject__` (resolved Python dict). By tracking `id(schema.__subject__)` in a per-path visited set, cycles are caught immediately regardless of how `oneOf`/`anyOf` merging reshapes the enclosing dict. The check runs at the top of `_field_from_complex_type` before type dispatch.

**Additional improvements:**
- Type fidelity: recursive marker uses the actual `datahub_field_type` instead of hardcoding `RecordTypeClass`
- Depth safety net (`MAX_COMPLEX_DEPTH=50`): guarantees termination even against unforeseen vectors
- Descriptive recursive leaves: injects "Recursive reference to \<definition-name\>" when description is empty

**Performance:** 37 fields in <1ms, down from 218,976 fields in 25 seconds.

## What was tested

- New regression test `test_json_schema_complex_recursive_refs_with_oneof` using the exact schema from the issue
- All 27 existing `test_json_schema_util.py` tests pass
- Deployed and validated against production schemas with recursive `$ref` + nested `oneOf` structures


rospe:
Hi @maggiehays , can you please check if there's someone in your team that can take a look at this PR? Thank you!

PR #17785, opened 2026-06-05
feat(i18n): extract keys from entity v1 entity shared (misc)

v-tarasevich-blitz-brain:
This PR extracts i18n keys from folders:

entity/shared/propagation
entity/shared/entity
entity/shared/* (root)
entity/shared/tabs/Dataset
entity/shared/tabs/Documentation
entity/shared/tabs/Incident
entity/shared/tabs/Properties
entity/shared/tabs/ERModelRelationship


datahub-web-react/src/app/entity/shared/tabs/Incident/components/IncidentListItem.tsx
```diff
@@ -229,33 +232,46 @@ export default function IncidentListItem({ incident, refetch }: Props) {
...
+                                            i18nKey="list.createdBy"
+                                            t={t}
+                                            values={{ time: toRelativeTimeString(createdDate) }}
+                                            components={{
+                                                actor: createdActor?.corpUser ? (
+                                                    <Link
+                                                        to={entityRegistry.getEntityUrl(
+                                                            EntityType.CorpUser,
+                                                            createdActor?.corpUser?.urn,
+                                                        )}
```
benjiaming, line 261: Bug: if createdActor?.corpUser is null, we would see something like "Created 5 minutes ago by ". Not a regression though.

benjiaming (approved):

datahub-web-react/src/app/entity/shared/tabs/Incident/components/IncidentListItem.tsx
max-datahub: Good catch — fixed in 3e87b22. Since the "by" is baked into the translated string (`list.createdBy` = "Created {{time}} by <actor></actor>"), I added a separate key without it (`list.created` = "Created {{time}}") and switch to it when the actor's `corpUser` is absent, so it now renders "Created 5 minutes ago" instead of the dangling "by ". Applied the same fix to the parallel resolved-by case (`list.resolved`).

PR #17787, opened 2026-06-05
feat(forms): assign compliance forms to owners of a specific type

acrylJonny:
## Motivation

Compliance forms can already be assigned to entity owners, but every owner gets every form regardless of ownership type. Teams that distinguish *technical*, *business*, *data steward* (etc.) owners on the same entity have no way to route a given form to the right cohort.

This PR adds an optional `ownershipTypes` list on `FormActorAssignment` that narrows owner-based assignment to owners whose ownership type matches the form's filter. Empty / unset preserves today's behaviour.

## What changed

| Layer | Change |
| --- | --- |
| Model | `FormActorAssignment.ownershipTypes: optional array[Urn]` — searchable, relationship-mapped to the `ownershipType` entity. |
| Java util | `OwnershipUtils.isOwnerOfEntityWithType` mirrors `OwnerServiceUtils.isOwnerEqual`'s legacy-enum fallback so owners persisted with only the deprecated `Owner.type` enum (no `typeUrn`) still match. |
| Service | `FormService.isFormAssignedToUser` calls into it; tests cover both `typeUrn`-only and enum-only paths. |
| Patch builder | `FormInfoPatchBuilder.{add,remove}AssignedOwnershipType`. Folds the existing user/group add/remove paths onto a shared helper (also removes the copy-paste error message bug in the SaaS-side ` setAssignedGroups`). |
| GraphQL | `FormActorAssignment.ownershipTypes` (typed), `FormActorAssignmentInput.ownershipTypes` (create-form), `FormActorAssignmentUpdateInput.ownershipTypes{ToAdd,ToRemove}` (update). Wired through `FormMapper`, `FormUtils`, `UpdateFormResolver`, and the loadable-type batch resolver in `GmsGraphQLEngine`. |
| Frontend | `useIsUserAssigned` reads `formAssociation.form.info.actors.ownershipTypes` and filters owners; `formFields` fragment fetches the new block. |
| Python SDK | `Actors.ownership_types` on the YAML recipe model (so the field documented under `actors:` is actually accepted by Pydantic), plus `FormPatchBuilder.{add,remove}_assigned_ownership_type`. |
| Docs | `forms.md` + compliance-forms guides document the new field and the built-in ownership-type URNs. |

## Behaviour

```yaml
# in a Forms recipe
actors:
  owners: true
  ownership_types:
    - urn:li:ownershipType:__system__technical_owner
    - urn:li:ownershipType:__system__data_steward
```

A user is treated as assigned only if (a) they're explicitly named, or (b) they own the entity *with one of the listed ownership types*. Empty / unset list preserves the existing "any owner is assigned" behaviour.

## Tests

- `OwnershipUtilsTest` — covers no-filter, single-type, multi-type, no-match, group-membership, and the enum-only legacy-owner fallback.
- `FormServiceOwnershipTest` — end-to-end test that `isFormAssignedToUser` returns the expected result across all of those combinations against mocked `FormInfo` + `Ownership` aspects.
- `FormInfo/__tests__/utils.test.ts` — covers `useIsUserAssigned` for typed and untyped ownership filtering.

## Checklist

- [x] PR conforms to the [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (especially [PR Title Format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#pr-title-format))
- [x] Tests added/updated
- [x] Docs updated
- [ ] Breaking changes: none — the new field is optional with a backward-compatible default.

## Notes

Originally proposed as #7076 (closed); this is a clean revival rebased against current `master`.


chriscollins3456 (commented):
nice i think this is making a good amount of sense!

obviously this would just be API-supported for now since i don't see the frontend changes for adding/updating these fields in this PR (or maybe this is in your next PR).

one major thing i think we're missing here is updating the path in order to fetch forms for users in the Task Center - that is done through the GetFormsForActorResolver. You'll have to look into that and how we get ownership forms and the forms you need to fill out because you own an entity that has an ownership form on it (like at `_formService.getFormsAssignedByOwnership`).

acrylJonny:
Thanks @chriscollins3456! Both points are SaaS-side:

- The add/update UI for these fields is the compliance-forms builder under `app/govern/Dashboard/Forms`, which lives in the DataHub Cloud codebase rather than OSS. This OSS PR is the model + GraphQL + SDK layer plus the `isFormAssignedToUser` (entity sidebar) check.
- `GetFormsForActorResolver` / `getFormsAssignedByOwnership` are Cloud-only as well, so there's nothing to change for them here. Good catch on the Task Center path though — I've updated `getFormsAssignedByOwnership` on the Cloud side so an ownership form is only returned when the user owns a matching entity with one of the form's allowed `ownershipTypes` (falling back to the existing "any owner" behaviour when no types are set), with unit coverage.

datahub-web-react/src/app/entity/shared/containers/profile/sidebar/FormInfo/useIsUserAssigned.ts
```diff
@@ -2,24 +2,54 @@ import { useUserContext } from '@app/context/useUserContext';
...
+    if (!owners) return false;
+
+    const userOwnership = owners.find((owner) => owner.owner.urn === userUrn);
+    if (!userOwnership) return false;
+
+    if (!ownershipTypeUrns || ownershipTypeUrns.length === 0) {
+        return true;
+    }
+
+    return !!userOwnership.ownershipType && ownershipTypeUrns.includes(userOwnership.ownershipType.urn);
```
chriscollins3456: one thing here that's an edge case but could happen is that a user could be an owner more than once on the asset and have multiple ownership types.

so we could change the logic here to get a list of the user ownerships (usually length of 0 or 1 but could be more) and then check if any of the ownerships matches any of the specified ownership type urns

chriscollins3456 (commented):
only one real comment for an edge case otherwise things look good!

thanks for clarifying on the OSS vs SaaS difference i didn't fully register i was looking at OSS and forgot about the differences between the two

datahub-web-react/src/app/entity/shared/containers/profile/sidebar/FormInfo/useIsUserAssigned.ts
acrylJonny: Good catch — updated so we collect all of the user's ownership entries and return true if any of their types matches any specified type, rather than only checking the first owner entry (a user owning the same entity under multiple types could otherwise be missed). Added tests covering the multiple-type case where a non-first type matches.

chriscollins3456 (approved):
nice

PR #17789, opened 2026-06-05
feat(gms): add distributed rate limiting across REST, GraphQL, and OpenAPI

david-leifker:
## Summary

- Add a new `metadata-service:rate-limit` module with a Hazelcast-backed `RateLimitEngine`, servlet filter, rule selector, adaptive capacity limiter, and Micrometer metrics
- Wire rate limiting into GMS across RestLI, GraphQL, and OpenAPI with standard throttle response headers
- Add configuration (`RateLimitProperties`, YAML + env var support), documentation (`docs/deploy/gms-rate-limiting.md`), and unit tests (1,341 passing across affected modules)

Design Doc: https://github.com/datahub-project/datahub/blob/1dd3e1c19249e74fc6eb8d2f31f5710e4edb4049/docs/deploy/gms-rate-limiting.md


metadata-service/rate-limit/src/main/java/com/linkedin/metadata/ratelimit/RateLimitMetrics.java
```diff
@@ -0,0 +1,71 @@
...
+                ? decision.getDenyingRuleId()
+                : decision.getCapacityRuleId() != null ? decision.getCapacityRuleId() : "none",
+            "type",
+            decision.getDenyingType() != null
+                ? decision.getDenyingType().name()
+                : decision.getCapacityRuleId() != null ? "capacity" : "none",
+            "outcome",
+            decision.isAllowed() ? "allow" : "deny",
+            "graphql_operation",
+            decision.getGraphqlOperation() != null ? decision.getGraphqlOperation() : "none");
```
nwadams: do we worry about tag explosion on the operationName at all here?

metadata-service/graphql-servlet-impl/src/main/java/com/datahub/graphql/GraphQLController.java
```diff
@@ -103,6 +112,29 @@ CompletableFuture<ResponseEntity<String>> postGraphQL(
...
+            new ResponseEntity<>(
+                mapper.writeValueAsString(Map.of("error", "Rate limit exceeded")),
+                headers,
+                HttpStatus.TOO_MANY_REQUESTS));
+      } catch (JsonProcessingException e) {
+        return CompletableFuture.completedFuture(
+            new ResponseEntity<>(HttpStatus.TOO_MANY_REQUESTS));
+      }
+    }
+    final RateLimitLease rateLimitLease = rateLimitEngine.toLease(rateLimitDecision);
```
nwadams: do we want to run everything in a try/finally block so we release the ratelimitLease if we fail to parse variables or fail to auth successfully before we run `GraphQLConcurrencyUtils.supplyAsync(`

metadata-service/graphql-servlet-impl/src/main/java/com/datahub/graphql/GraphQLController.java
```diff
@@ -136,62 +168,76 @@ CompletableFuture<ResponseEntity<String>> postGraphQL(
...
+
+              /*
+               * Format & Return Response
+               */
+              try {
+                long totalDuration = submitMetrics(executionResult);
+                // Remove tracing from response to reduce bulk, not used by the frontend
+                executionResult.getExtensions().remove("tracing");
+                String responseBodyStr =
+                    new ObjectMapper().writeValueAsString(executionResult.toSpecification());
```
nwadams: nit: this is pre-existing but should we use a singleton objectMapper to avoid extra memory allocations

alexsku (commented):
## Review: distributed rate limiting (REST / GraphQL / OpenAPI)

Thorough read of the new `metadata-service:rate-limit` module and its wiring. Overall this is a well-structured feature, cleanly **off by default**, and the REST filter path (`RateLimitFilter`) is solid and well-tested. Two issues should be addressed before anyone enables `capacity.enabled` in production, plus some smaller items.

> Note: I reviewed against PR head `1dd3e1c`. Severity labels are my opinion, not gate requirements.

---

### 🔴 Blocker — Capacity permit leaks when request processing throws between acquire and the async stage

**`GraphQLController.java` (lines 117–147; release at line 232)**

`evaluateAndAcquireGraphQL` acquires an adaptive-concurrency permit at line 117, but the permit is only released in `executionFuture.whenComplete(...)` at line 232. The code in between — `variables` parsing (142–147), `SpringQueryContext` construction (154–164), and `supplyAsync` — is **not** wrapped in try/finally. If anything there throws synchronously, the method returns before `whenComplete` is attached and **the permit is never released**.

This is trivially reachable from a client: a request with a non-object `variables`, e.g.

```json
{"query":"query X{__typename}","variables":42}
```

makes `new ObjectMapper().convertValue(variablesJson, Map.class)` throw `IllegalArgumentException` at line 145. The client correctly gets a 400 (via `GlobalControllerExceptionHandler`), but the GraphQL capacity inflight counter is permanently incremented — `SimpleLimiter` has no timeout/auto-release. Repeat N times → inflight saturates → the limiter denies *all* GraphQL with 429 until GMS restart. A handful of malformed requests can brick GraphQL on a pod.

It's gated behind `capacity.enabled=true` (off by default), but `capacity.graphql.enabled` defaults true within that, so this is the feature's primary path. The `RateLimitFilter` (REST path) handles this correctly with try/finally + an `AtomicBoolean` (`RateLimitFilter.java` lines 52–66) — the controller just doesn't mirror that discipline.

**Suggested fix** — release the lease on synchronous failure:

```java
final RateLimitLease lease = rateLimitEngine.toLease(rateLimitDecision);
try {
    // variables parse, context build, supplyAsync, whenComplete(...)
    return executionFuture;
} catch (RuntimeException ex) {
    rateLimitEngine.release(lease, false);
    throw ex;
}
```

**Validate**: with `RATE_LIMITS_CAPACITY_ENABLED=true`, POST a few requests with `"variables": 42`, then `GET /openapi/v1/rate-limits/status` → `adaptive._graphql_capacity.inflight` stays elevated and never returns to 0. There is currently no `GraphQLController` test covering this.

---

### 🟠 High — Unbounded metric cardinality from user-controlled GraphQL operation name

**`RateLimitMetrics.java` (lines 23–42), fed from `GraphQLController.java:115`**

`recordDecision` tags the `gms.rate_limit.requests` counter with `graphql_operation = decision.getGraphqlOperation()`. That value is `resolvedOperationName`, derived from the request's `operationName` (or the first operation name parsed from the query document) — fully client-controlled and unbounded. Denied decisions are sampled 100% (`shouldSample` returns true for any deny), so an attacker sending distinct operation names (`query a1{…}`, `query a2{…}`, …) creates a new meter series per name.

There's no global cardinality cap — the only `MeterFilter`s (`MicrometerMetricsFactory.java` lines 107–145) filter by metric origin, not tag cardinality. Result: unbounded `Meter` objects in the GMS heap (OOM risk) and a Prometheus cardinality blow-up — driven by exactly the abusive traffic this feature exists to stop.

**Suggested fix**: only emit the operation tag when the operation matches a configured operation-scoped rule (bounded set), else use a constant like `"other"`; or gate it behind `metrics.detailed=true`; or register a `maximumAllowableTags` filter for this meter.

---

### Other items

- **Medium (behavior change)** — `GraphQLController.java:115`: when `operationName` is omitted, `GraphQLOperationNameResolver.resolve` now parses the query *before* execution. A malformed query throws `InvalidSyntaxException` → 400. Previously it went straight to `_engine.execute`, which returns a standard GraphQL **200 with an `errors` array**. This is a client-visible regression for malformed queries that don't send `operationName`. Consider wrapping `resolve` in try/catch and falling back to `"graphql"` so syntax handling stays with the engine. (It also double-parses the document — here and in the engine.)

- **Medium (docs)** — The PR description mentions "a runtime admin API for endpoint **overrides**," but `RateLimitController` is read-only (`GET /config`, `GET /status`) and `EndpointRateLimitStore` has no runtime mutator. Was override descoped? If so, the description should be updated.

- **Low (correctness)** — `RateLimitProperties.java:16-19`: `failOpen=true`, `defaultRetryAfterSeconds=60`, `excludedPaths` have field initializers but no `@Builder.Default`. Anything built via `RateLimitProperties.builder().build()` gets `failOpen=false` (**fail-closed**) and `excludedPaths=null`. Production binds via setters so it's safe today, but it's a fail-closed footgun for tests/future callers. Same for `CapacityLimitConfig`.

- **Low (validation)** — `RateLimitConfigLoader` validates presence of `initialLimit`/`maxLimit` but not ordering/positivity (`minLimit ≤ initialLimit ≤ maxLimit`, all > 0). Bad values flow straight into `Gradient2Limit.newBuilder()`.

- **Low (DX)** — minor duplication: `RateLimitHeaderWriter` vs `ThrottleResponseHeaderWriter`; `HazelcastInstanceBootstrapCondition` and `RateLimitEndpointEnabledCondition` redefine the same constant. Also `RateLimitConfigLoader` (lines 80–85) logs "found in filesystem" *before* the stream is opened, so it prints even when the open fails.

---

### Test gaps

- **No `GraphQLController` test at all** — the new acquire/release lifecycle (success, GraphQL-errors, exception-before-async) is untested, which is exactly where the Blocker lives. The filter equivalent *is* well covered (`RateLimitFilterTest`).
- `GraphQLOperationNameResolverTest` covers only happy paths — no malformed-document case (the behavior change above).
- `codecov/patch` is at 48%, consistent with the untested controller branches.

### One process note

`build-except-metadata-ingestion` is currently red. I couldn't surface the exact Java failure (the failed-log output truncates mid-build; Python suites all pass and the annotation is just "exit code 1"). Worth confirming it isn't a real test failure in a metadata-service module before merge.

---

#### Questions
1. Was the "endpoint overrides" admin API intentionally descoped to read-only?
2. Releases use `onSuccess`/`onIgnore` only (`AdaptiveCapacityLimiter.java:43-47`) — timeouts/5xx map to `onIgnore`, never `onDropped`. Deliberate (don't want overload signals to shrink the Gradient2 limit)?

Nice work overall — the leak and the cardinality vector are both cheap fixes, and with controller tests this is in good shape.

metadata-service/configuration/src/main/resources/application.yaml
```diff
@@ -154,6 +154,44 @@ datahub:
...
+    # GMS HTTP service rate limiting (disabled by default).
+    # Limits incoming GMS API traffic — NOT MCP ingestion or Kafka lag throttle.
+    # Guide: docs/deploy/gms-rate-limiting.md
+    #
+    # Hierarchy: capacity.enabled | endpoint.enabled (by type)
+    #            → capacity.default / capacity.graphql (sub-pools) / capacity.rules
+    #            → endpoint.rules (cluster-wide buckets via Hazelcast when endpoint.enabled)
+    rateLimits:
+      failOpen: ${RATE_LIMITS_FAIL_OPEN:true}
+      defaultRetryAfterSeconds: ${RATE_LIMITS_DEFAULT_RETRY_AFTER:60}
```
alexsku: Optional / non-blocking idea: for **endpoint** (token-bucket) denials, `Retry-After` is currently the flat `defaultRetryAfterSeconds`. Bucket4j can give you the exact recovery time, so you could honor the real wait whenever it exceeds the configured backoff:

```java
ConsumptionProbe probe = bucket.tryConsumeAndReturnRemaining(1);
if (!probe.isConsumed()) {
    long waitSec = (long) Math.ceil(probe.getNanosToWaitForRefill() / 1_000_000_000.0);
    retryAfter = Math.max(config.getDefaultRetryAfterSeconds(), waitSec); // config = minimum backoff
}
```

This reframes `defaultRetryAfterSeconds` as a **minimum** backoff rather than a constant: a deeply-drained bucket that genuinely needs ~10 min reports 10 min instead of an optimistic 60 (which would just yield a handful of failed retries + 429s before it actually recovers). Short shortfalls still get the floor, which is reasonable for a limiter — you generally don't want to invite sub-minute retry loops. Probably worth a little jitter on top so a wave of throttled clients doesn't all retry on the same tick.

This would mean `EndpointRateLimitStore.tryConsume` surfaces the `ConsumptionProbe` instead of a bare `boolean`. It doesn't apply to the capacity/Gradient2 path (no refill clock there), so the constant stays as the fallback for that side.

metadata-service/rate-limit/src/main/java/com/linkedin/metadata/ratelimit/RateLimitMetrics.java
david-leifker: Yes I think we do. I will mitigate this concern by only recording the operation name if it is defined in a rule. Practically this means that we'll maintain separate tracking for the operations we care about protecting the most.

metadata-service/graphql-servlet-impl/src/main/java/com/datahub/graphql/GraphQLController.java
david-leifker: good catch

metadata-service/configuration/src/main/resources/application.yaml
david-leifker: I like this idea

nwadams (approved):

PR #16029, opened 2026-01-30
fix: Add retry logic when constructing kafka producer (#86)

redblackcoder:
Intermittent network issue could cause datahub-gms (and other services) to fail to startup successfully while trying to configure kafka producer.

This change allows for retry logic to go past such intermittent errors.

PR #17791, opened 2026-06-05
feat(kafka): Kafka producer retry

david-leifker:
Updates from https://github.com/datahub-project/datahub/pull/16029


metadata-service/factories/src/main/java/com/linkedin/gms/factory/kafka/DataHubKafkaProducerFactory.java
```diff
@@ -71,7 +74,48 @@ protected Producer<String, String> createDUEProducer(
...
+  /**
+   * Creates a KafkaProducer with retry logic for initialization failures. This handles transient
+   * issues like DNS resolution failures that can occur during container startup in Kubernetes
+   * environments.
+   *
+   * @param props the producer configuration properties
+   * @param producerConfiguration the producer configuration containing retry settings
+   * @return the created KafkaProducer
+   */
+  public static <K, V> Producer<K, V> createProducerWithRetry(
```
RyanHolstien, line 89: nit: This function isn't being used and can be removed, the usage of it is already passing through the KafkaProducer::new which I think is preferred to having this telescoping function.

RyanHolstien (approved):

PR #17794, opened 2026-06-05
fix(looker): set HTTPAdapter pool size to match max_threads

EladLeev:
Hey 👋 
While using the Looker source ingestion, the connector emitted repeated `urllib3` warnings during ingestion:
```
[2026-06-03 08:43:16,208] WARNING  {urllib3.connectionpool:327} - Connection pool is full, discarding connection: XYZ.cloud.looker.com. Connection pool size: 10
```
Seems like the root cause was that `HTTPAdapter` was only configured with `max_retries`, leaving the connection pool at urllib3's [default of 10](https://requests.readthedocs.io/en/latest/api/#requests.adapters.HTTPAdapter). While we made parallel API calls, the pool was consistently exhausted, causing connections to be discarded and re-created.

This PR:
1. Moves the `max_threads` from `LookerDashboardSourceConfig` to `LookerAPIConfig`.
2. Sets `pool_connections` and `pool_maxsize` on the `HTTPAdapter` to `max_threads + 10` (same as the Superset source connector)
3. Add a unit test that verifies the pool size.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/looker/looker_config.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/looker/looker_config.py
@@ -1,5 +1,4 @@
 import dataclasses
-import os
 import re
 from typing import Any, ClassVar, Dict, List, Optional, Tuple, Union
 
@@ -272,10 +271,6 @@ class LookerDashboardSourceConfig(
         False,
         description="Whether to skip ingestion of dashboards in personal folders. Setting this to True will only "
         "ingest dashboards in the Shared folder space.",
-    )
-    max_threads: int = Field(
-        default_factory=lambda: os.cpu_count() or 40,
-        description="Max parallelism for Looker API calls. Defaults to cpuCount or 40",
     )
     external_base_url: Optional[str] = Field(
         None,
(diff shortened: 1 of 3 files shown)


shirshanka:
Thanks for the contribution @EladLeev 🙌 


Here's a quick review:

🐛 Root Cause — Correctly Identified

The HTTPAdapter was initialized with only max_retries, leaving urllib3's connection pool at the default of 10. With max_threads defaulting to os.cpu_count() or 40, up to 30 worker threads were competing for 10 pool slots → connections discarded and re-created → "Connection pool is full" warnings and degraded throughput.✅ 

What the PR Does 

looker_lib_wrapper.py — Moves max_threads into LookerAPIConfig and sets pool_connections = pool_maxsize = max_threads + 10 on the HTTPAdapter. Since LookerDashboardSourceConfig inherits from LookerAPIConfig, existing references continue to work.looker_config.py — Removes the duplicate max_threads from LookerDashboardSourceConfig. Clean.

test_looker_lib_wrapper.py — New unit test verifying pool size = max_threads + 10.

💬 Minor Suggestions

The + 10 buffer follows the Superset connector pattern — a short inline comment explaining why (e.g. "extra headroom for auth/metadata calls outside the main thread pool") would help future readers.

Worth checking: does the LookML source share the same LookerAPIConfig-based HTTP session? If yes, this fix covers it. If not, it may need similar treatment.

Verdict: LGTM with minor nits — real bug fix with proper test coverage.

 @askumar27 tagging you for a final sign-off since you last worked on the Looker source! 🙏

treff7es (approved):
thanks for the contribution, it looks good!

PR #17797, opened 2026-06-08
fix(ingest/dbt): restore column-level lineage for two-tier warehouses (catalog-prefixed SQL + v2 schema fieldPaths)

aviraj-gour:
## Summary

Restores dbt **column-level lineage (CLL)** for projects whose warehouse is a two-tier / catalog-prefixed platform (Glue, Trino, Athena, Iceberg) used with `include_database_name: false`. In an affected project, model CLL went from **0/174 models to all of them**; table-level lineage was unaffected throughout.

Two independent bugs both broke the same step — resolving upstream column schemas during SQL parsing — so this PR fixes both.

### 1. Catalog prefix in compiled SQL didn't match registered URNs

With `include_database_name: false`, dbt registers nodes as `schema.table` (2-part). But dbt's compiled SQL, running through Trino/Athena, references tables as `catalog.schema.table` (3-part, e.g. `"iceberg_prod_usw2"."schema"."table"`). No 3-part URN is ever registered, so the lookup failed and upstreams couldn't be resolved.

**Fix:** `_CatalogStrippingSchemaResolver` strips the catalog component before every lookup when `include_database_name` is disabled, so a 3-part SQL reference resolves to the registered 2-part URN.

### 2. v2 fieldPaths from the graph weren't normalized

When upstream schemas are fetched from the graph (a Glue dataset's `SchemaMetadata`, e.g. with `datahub_api` configured), their fieldPaths are v2-encoded — `[version=2.0].[type=string].event_id`. `_to_schema_info` used these verbatim as column names, so the SQL parser (which matches on bare `event_id`) found no match, and every output column came back with no upstreams → no CLL.

**Fix:** `_to_schema_info` now strips the v2 wrapper via `get_simple_field_path_from_v2_field_path`, mirroring what the rest of the SQL-parsing stack already does in `_convert_schema_field_list_to_info`. Bare names (from the dbt catalog) pass through unchanged.


metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
```diff
@@ -149,11 +151,33 @@
...
+
+    With include_database_name=False all URNs are registered as schema.table
+    (2-part), but dbt compiled SQL running on Trino/Athena references tables as
+    catalog.schema.table (3-part). Since no 3-part URN is ever registered, the
+    catalog component is stripped before every lookup so the resolved URN always
+    matches the 2-part URN that dbt and target_platform_urn_to_dbt_name use.
+    """
+
+    def resolve_table(self, table: _TableName) -> Tuple[str, Optional[SchemaInfo]]:
+        if table.database:
```
puneetagarwal-datahub: if table.database: check not required

puneetagarwal-datahub (changes requested):

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
```diff
@@ -2079,7 +2115,16 @@ def _infer_schemas_and_update_cll(  # noqa: C901
 
         graph: Optional[DataHubGraph] = self.ctx.graph
 
-        schema_resolver = SchemaResolver(
+        resolver_class: Type[SchemaResolver]
+        if not self.config.include_database_name:
+            logger.debug(
```
puneetagarwal-datahub: debug log not required

puneetagarwal-datahub (changes requested):

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
aviraj-gour: resolved, removed check.

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
aviraj-gour: removed log

puneetagarwal-datahub (approved):

PR #17804, opened 2026-06-08
ci: skip metadata model upload when DataHub secret is unset instead of failing

alokr-dhub:
## Summary

The `metadata model generate` workflow (`.github/workflows/metadata-model.yml`) gated its publishing steps on whether the run is on the canonical repo (`needs.setup.outputs.publish`), but **not** on whether the DataHub upload target was actually configured.

When the `DataHubServer` secret is empty or expired, the `Upload metadata to DataHub` step falls back to the DataHub CLI default of `http://localhost:8080`, retries for ~28s, and then fails the entire workflow with a connection-refused error — even though model generation and the S3 upload both succeeded.

Observed in [this run](https://github.com/datahub-project/datahub/actions/runs/27132469737/job/80076595458):

```
Failed to configure the sink (datahub-rest): 💥 Failed to connect to DataHub
with DataHubRestEmitter: configured to talk to http://localhost:8080
...
ConnectionRefusedError: [Errno 111] Connection refused
```

## Changes

- Resolve the `DataHubServer` secret into a `datahub_configured` boolean output in the `setup` job. (Secrets cannot be referenced directly in step `if:` conditions, so the check is done once in a `run:` step and exposed as an output, mirroring the existing `publish` gate.)
- Gate the `Upload metadata to DataHub` step on `datahub_configured == 'true'`.
- Emit a `::warning::` annotation when publishing is enabled but the secret is missing, so the skip is visible in the Actions summary without hard-failing the run.

## Behavior

| Scenario | Before | After |
| --- | --- | --- |
| Secret set | uploads | uploads |
| Secret empty/expired | **workflow fails** (connection refused) | model still generated + uploaded to S3; upload step skipped with a warning |

Note: this makes the workflow resilient to a missing secret — it does not replace setting/refreshing the `DataHubServer` / `DataHubToken` secrets when the live publish is actually desired.

### DIFF
--- a/.github/workflows/metadata-model.yml
+++ b/.github/workflows/metadata-model.yml
@@ -19,12 +19,22 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       publish: ${{ steps.publish.outputs.publish }}
+      datahub_configured: ${{ steps.publish.outputs.datahub_configured }}
     steps:
       - name: Check whether upload to datahub is enabled
         id: publish
+        # secrets cannot be referenced directly in step `if:` conditions, so we
+        # resolve the DataHub target here and expose a boolean output instead.
+        env:
+          DATAHUB_SERVER: ${{ secrets.DataHubServer }}
         run: |
           echo "Enable publish: ${{ github.repository == 'datahub-project/datahub' }}"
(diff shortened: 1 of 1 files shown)


alokr-dhub:
Superseded by #17806, which uses a branch on datahub-project/datahub directly (same-repo) instead of a fork.

PR #17805, opened 2026-06-08
feat(ui): add lazy Phosphor icon registry for MFE nav icons

rob-1019:
A previous tree-shaking commit (cdb1ba7b) replaced the dynamic barrel import with a hardcoded <AppWindow />, causing user-specified icons not present in the tree-shaken bundle to not show the configured icon.

This PR introduces a lazy-loading registry so any of the 1500+ Phosphor icons can be used without bundling them all upfront.

- scripts/generate-lazy-icon-stubs.js writes one thin re-export per icon into src/app/mfeframework/lazy-icons/ (gitignored, generated at build time)
- assertLazyIconsGenerated() Vite plugin fails fast if stubs are missing
- New Gradle task generateLazyIconStubs wired into yarnServe, yarnBuild, yarnTest, yarnLint
- Vite builds each stub as a separate async chunk via import.meta.glob
- getLazyIcon(name) loads only the chunk for the configured icon

MFE authors continue using navIcon: "Database" (any Phosphor name) unchanged.

### DIFF
--- a/datahub-web-react/.eslintignore
+++ b/datahub-web-react/.eslintignore
@@ -3,3 +3,4 @@ dist/
 build/
 node_modules/
 **/*.generated.ts
+scripts/
--- a/datahub-web-react/.gitignore
+++ b/datahub-web-react/.gitignore
@@ -30,6 +30,9 @@ yarn-error.log*
 
 # gql codegen
 *.generated.ts
+
+# generated lazy icon stubs (produced by scripts/generate-lazy-icon-stubs.js)
+/src/app/mfeframework/lazy-icons/
 /.vscode
 
 .yarn-test-sentinel
--- a/datahub-web-react/build.gradle
+++ b/datahub-web-react/build.gradle
@@ -104,17 +104,32 @@ task yarnGenerate(type: Copy, dependsOn: yarnGenerateCacheable) {
   includeEmptyDirs = false
 }
 
-task yarnServe(type: YarnTask, dependsOn: [yarnInstall, yarnGenerate]) {
+// Generate per-icon stub files so Vite can build them as separate lazy chunks.
+// Client downloads only the icon chunks it actually renders — none are in the main bundle.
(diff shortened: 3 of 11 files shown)


rob-1019:
The failing Pytest Smoke Tests (Batch 3/7) is the test_datapack_smoke.py test is currently failing for reasons not related to this patch because the showcase-ecommerce datapack it loads contains MCPs with a newer schemaVersion field that the server schema on master doesn't yet recognize — an Avro deserialization mismatch (1,246 records rejected as  "cannot be parsed as union schema"), entirely in the metadata-model/ingestion layer.

benjiaming:
This PR is adding 5MB because of all the 1,500 async stub icons. I wonder if we could keep current tree shaking and load any missing icons from a new endpoint which would return them as svg. We would need to introduce the icons into gms jar. The browser would load it on demand (most users will never need it) and cache it.

```
GET /api/v2/icons/phosphor/sprite.svg

Returns one SVG with all icons as <symbol> elements:
<svg xmlns="http://www.w3.org/2000/svg">
  <symbol id="AppWindow" viewBox="0 0 256 256">...</symbol>
  <symbol id="Globe" viewBox="0 0 256 256">...</symbol>
  ...
</svg>

Frontend renders with <use>:
<svg width={size} height={size}>
  <use href="/api/v2/icons/phosphor/sprite.svg#AppWindow" />
</svg>
```

benjiaming (approved):
Conditionally approving as long as the bundle does not increase by 5MB.

rob-1019:
@benjiaming The individual chunks in question here are lazy-loaded - the clients do not download the whole bundle, just the chunks that contain something they need. I checked what clients are actually downloading using Chrome Network Tools and no large loads are being introduced.

Note that the below is on a dev install so some assets will get smaller in prod builds due to build-time optimizations but they won't get any bigger - there's not any 5MB chunks sneaking in nor do I see lots of mid/small/tiny ones either when I scroll through.

image

PR #17807, opened 2026-06-08
refactor(ingest/sql-parsing): tokenizer-based statement splitter + #<digit> temp-table fix

treff7es:
## Summary

Splits the shared SQL-parsing changes out of #17761 (Tableau Initial SQL) into a standalone PR, since they are a library-level change that benefits every SQL connector, not just Tableau. #17761 is now stacked on top of this PR.

Two changes:

### 1. Tokenizer-based `split_statements` rewrite

Replaces the bespoke character-state-machine statement splitter with a dialect-aware one built on sqlglot's tokenizer. It is shared by stored-procedure lineage (MSSQL/Oracle/Postgres/MySQL/DB2/HANA/Snowflake), PowerBI native-SQL parsing, and Tableau Initial SQL, including semicolon-less T-SQL/PL-SQL batches.

- **Dialect-aware lexing** — boundary detection runs on a sqlglot token stream produced by the caller's dialect, so quoting (`` `x` ``/`[x]`/`"x"`), comment styles (`--`/`/* */`/`#`), dollar-quoting and the `;` rule are handled per dialect. `split_statements` now takes an optional `dialect`, threaded through from the stored-procedure and multi-statement lineage callers.
- **Single linear pass, offset slicing** — records cut points as source offsets and returns each statement as a verbatim slice of the original SQL (comments/whitespace/casing preserved); comment-only chunks are detected by the tokenizer emitting no tokens, not by regex.
- **Boundary heuristics** — bracket/CASE-END depth, control-flow peeling (`BEGIN/END/TRY/CATCH/IF/WHILE/LOOP/ELSIF/EXCEPTION`), DML starters, default-split `SELECT`/`WITH` with continuation guards (set-op `ALL/DISTINCT`, `INSERT`/`MERGE` source, `CREATE … AS`, CTE main query), `SET @v`/`DECLARE @v`, `MERGE … WHEN … THEN`. Guiding rule: a keyword is a boundary **only when the previous token can end a statement** (a continuation blocklist that errs toward over-splitting, never merging two real statements → never silently dropping lineage).
- **`GO` batch separator** — gated to T-SQL dialects only (`mssql`/`fabric-onelake` → sqlglot `tsql`); other dialects treat `GO` as an ordinary token.
- **Never raises, never loses SQL** — tokenizer or walk failure yields the whole input unchanged.

~30–45× faster than the old regex-per-character splitter (which ran `re.match` at every character position).

### 2. `#<digit>` T-SQL temp-table parse fix

sqlglot's tokenizer lexes the leading digit run of a T-SQL temp table named `#<digit>…` (e.g. `#63114Actual`, `##2025Totals`) as a `NUMBER`, so the parser raises `Expected table name but got NUMBER` and loses **all** lineage for that statement — including the real `FROM` source. A tsql-gated `_sanitize_tsql_temp_tables()` in `parse_statement` (mirroring the existing `_sanitize_snowflake_ddl` path) wraps such names in `[…]` so sqlglot reads them as one delimited identifier. The normalized name still starts with `#`, so downstream temp-table detection is unaffected; string literals and comments are passed through untouched. Scoped to T-SQL because the `#` temp-table prefix exists only there (other dialects treat `#` as a comment or reject it).

## Test Plan

- [x] `tests/unit/sql_parsing/test_split_statements.py` (77 tests) + recall (12/12, each chunk run through the real `create_lineage_sql_parsed_result`) and precision (10/10) gates in `test_split_statements_recall.py`.
- [x] `tests/unit/sql_parsing/test_sqlglot_utils.py` — pure sanitizer cases + behavioral lineage-recovery test for `#<digit>` temp tables.
- [x] Full `tests/unit/sql_parsing/` suite green (212 passed, 3 skipped); stored-proc/MSSQL lineage suites green.
- [x] `./gradlew :metadata-ingestion:lint` passes (ruff + mypy).

## Checklist

- [x] Tests added/updated
- [x] No breaking changes (`split_statements` gains an optional `dialect` param; behavior is equivalent-or-better, verified by the recall/precision gates)

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/sql/stored_procedures/lineage.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/sql/stored_procedures/lineage.py
@@ -14,7 +14,7 @@ from datahub.sql_parsing.sql_parsing_aggregator import (
     ObservedQuery,
     SqlParsingAggregator,
 )
-from datahub.sql_parsing.sql_parsing_common import QueryType
+from datahub.sql_parsing.sql_parsing_common import QueryType, get_dialect_str
 from datahub.sql_parsing.sqlglot_utils import (
     get_dialect,
     is_dialect_instance,
@@ -168,7 +168,7 @@ def parse_procedure_code(
     dialect = get_dialect(platform)
 
     # Split statements using split_statements()
-    statements = list(split_statements(code))
+    statements = list(split_statements(code, dialect=get_dialect_str(platform)))
 
     # Filter out non-DML statements
     dml_statements = _filter_dml_statements(
--- a/metadata-ingestion/src/datahub/sql_parsing/split_statements.py
+++ b/metadata-ingestion/src/datahub/sql_parsing/split_statements.py
@@ -1,338 +1,483 @@
 import logging
 import re
-from enum import Enum
-from typing import Iterator, List, Tuple
+from typing import Final, Iterator, List, Optional, Tuple
+
+from sqlglot import Dialect
+from sqlglot.dialects.tsql import TSQL
+from sqlglot.tokens import Token, TokenType
 
 logger = logging.getLogger(__name__)
-SELECT_KEYWORD = "SELECT"
-CASE_KEYWORD = "CASE"
-END_KEYWORD = "END"
 
-_WORD_BOUNDARY_PATTERN = re.compile(r"\w\W|\W\w")
-_FOR_CLAUSE_PATTERN = re.compile(r"\bFOR\s*$")
+# GO is a client batch separator (sqlcmd/SSMS), not a SQL statement: it appears alone on a
+# line. We split on it before tokenizing because sqlglot's tsql tokenizer can otherwise
+# swallow `GO\n<rest>` into a single STRING token. GO lines carry no lineage and are dropped.
+_GO_BATCH_SEPARATOR = re.compile(r"(?im)^[ \t]*GO[ \t]*;?[ \t]*\r?$")
 
-CONTROL_FLOW_KEYWORDS = [
-    # MSSQL/T-SQL control flow
-    "GO",
-    r"BEGIN\s+TRY",
-    r"BEGIN\s+CATCH",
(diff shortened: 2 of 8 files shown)


sgomezvillamor (approved):
**Context:** This replaces the character-state-machine `split_statements()` with a sqlglot-tokenizer-based implementation. Blast radius is very high — this function is called by stored-procedure lineage for MSSQL, Oracle, Postgres, MySQL, DB2, HANA, Snowflake, PowerBI, and Tableau.

**What's solid:** The T-SQL recall/precision gates are thorough, the continuation-blocklist design correctly errs toward over-splitting (never merges two real statements), `_uses_go_batches` correctly uses `isinstance(..., TSQL)` to cover all T-SQL aliases, and the tokenizer-failure fallback (yield whole input) prevents crashes.

---

**Must-fix before merge:**

1. **Silent failure chain** — `logger.debug` in the `except Exception` block should be `logger.warning`. When the tokenizer fails, the whole batch is yielded as one blob, `parse_statement()` then fails on it, and that's also caught silently. End result: a procedure's entire lineage is lost with no operator-visible signal.

2. **No Oracle/non-T-SQL recall gate** — `test_split_statements_recall.py` uses `dialect="tsql"` exclusively. Oracle PL/SQL (`LOOP`/`ELSIF`/`EXCEPTION`/`END LOOP`), Snowflake, and MySQL are all served by this function but have zero behavioral recall tests.

3. **GO-batch path untested end-to-end** — none of the `parse_procedure_code()` integration tests use a `GO`-separated MSSQL batch, so the new dialect-threaded code path through `parse_procedure_code` is untested at that level.

---

**Should fix:**

4. `test_splits_no_semicolon_tsql_batch` and `test_split_oracle_plsql_function` call `split_statements()` without a `dialect` argument — they test the `dialect=None` fallback, not the production path.

5. Precision tests: `_tables()` returns `set()` when lineage parsing fails, so `set() == set()` silently passes. Add `assert len(_tables(sql)) > 0` before comparing to prevent false positives.

---

Issue #16173, opened 2026-02-11
Glue source uses ShowAssignedLFTags=True, excluding inherited LakeFormation tags


csilgo-straumann:
**Describe the bug**

The Glue source's `get_table_lf_tags` and `get_database_lf_tags` methods call the AWS `get_resource_lf_tags` API with `ShowAssignedLFTags=True`. This causes the API to return only directly assigned LF-tags, excluding tags inherited from parent resources (database → table inheritance).

In most AWS LakeFormation setups, tags are assigned at the database level and inherited by all tables. With `ShowAssignedLFTags=True`, these inherited tags are invisible to DataHub.

```python
# glue.py - get_table_lf_tags()
response = self.lf_client.get_resource_lf_tags(
    ...
    ShowAssignedLFTags=True,  # <-- should be False or configurable
)
```

**To Reproduce**

1. In AWS LakeFormation, assign LF-tags to a database (e.g., `domain=CS_Sales`)
2. Verify that tables in that database inherit the tag (visible in AWS Console)
3. Run Glue ingestion with `extract_lakeformation_tags: true`
4. Observe that only tables with directly assigned tags get `globalTags` — tables relying on inheritance get none

**Expected behavior**

The Glue source should extract both directly assigned and inherited LF-tags. Either change `ShowAssignedLFTags` to `False` (the AWS API default), or add a config option like `extract_inherited_lakeformation_tags: true/false`.

**Desktop**

- DataHub version: v1.4.0
- CLI version: 1.4.0.2
- Source type: `glue`

**Additional context**

When `ShowAssignedLFTags=False`, the AWS API returns all effective tags on the resource, including inherited ones. This is the default behavior of the API.

PR #17812, opened 2026-06-08
feat(ingestion/glue): propagate Lake Formation database tags to tables and extract column tags (ING-2163)

treff7es:
## Summary

Makes the Glue connector surface the **Lake Formation tags that tables and columns effectively carry**, addressing ING-2163 / #16173.

Two gaps, both fixed from data AWS already returns in a single call:

1. **Tables didn't show tags inherited from their database.** `GetResourceLFTags` returns only *directly-assigned* tags per resource (`ShowAssignedLFTags=True`).
2. **Column-level LF-tags weren't extracted at all.**

## Key realization (scalable by construction)

One `GetResourceLFTags` call on a **Table** resource returns all three layers — `LFTagOnDatabase`, `LFTagsOnTable`, `LFTagsOnColumns` — so everything comes from the **same per-table call we already make**: no extra API calls, no per-column calls, no caching.

## Behavior

Config (default-safe, gated by existing `extract_lakeformation_tags`):

- **`propagate_lakeformation_tags`** (default `false`): the database's tags are inherited by tables **and columns**. A more-specific assignment overrides an inherited value for the same key (table over database, column over table/database).
- **`extract_lakeformation_column_tags`** (default `true`): attach directly-assigned column tags to the matching `SchemaField`.

This mirrors the **Snowflake connector's `with_lineage` inheritance** (database → table → column, most-specific wins) so the two connectors behave consistently.

| Database | Table (direct) | Column (direct) | Column result (propagation on) |
|---|---|---|---|
| `Env:Prod` | — | — | `Env:Prod` (inherited) |
| `Env:Prod` | `Class:Sensitive` | — | `Env:Prod`, `Class:Sensitive` |
| `Env:Prod` | `Class:Sensitive` | `pii:true` | `pii:true`, `Env:Prod`, `Class:Sensitive` |
| — | `Class:Sensitive` | `Class:Public` | `Class:Public` (column wins) |

### Propagated tags are marked as propagated

Inherited tags are emitted with **propagation attribution** (`MetadataAttributionClass` with `sourceDetail={"propagated":"true","origin":<urn>}`) **and** a legacy `context` JSON string — the same mechanism the DataHub tag-propagation action uses — so propagated tags are distinguishable from directly-assigned ones (the DataHub UI renders a **Propagated** indicator with the origin entity). Origin is the database container (for table tags) or the table dataset (for column tags). Directly-assigned tags carry no attribution.

Verified in the emitted aspects:

```
TABLE  globalTags: datahub_propagation_test:db_level  context={"propagated":"true","origin":"urn:li:container:..."}  attribution=yes
COLUMN customer_id: datahub_col_test:pii            context=None            # direct — not marked
COLUMN customer_id: datahub_propagation_test:db_level context={"propagated":"true","origin":"urn:li:dataset:(...sales,PROD)"}
COLUMN region:      datahub_propagation_test:db_level context={"propagated":"true","origin":"urn:li:dataset:(...sales,PROD)"}
```

glue-lf-column-inheritance

glue-lf-column-tag-sales

glue-lf-firehosetest-inherited

glue-lf-propagated-tooltip

glue-lf-sales-override

glue-lf-tag-db-level

glue-lf-tag-table-override

## Testing

TDD unit tests: table propagation on/off, table-wins override, dedupe, column-tag parsing, column→`SchemaField` attachment, column inheritance from table/database, column-direct override, propagated-tag marking (table + column), multi-value tag expansion, malformed-tag skip, error→`report.warnings`, and a full-pipeline e2e. Full Glue unit suite: **136 passed**; `ruff` + `mypy` clean.

Verified end-to-end against live AWS Glue + Lake Formation, ingested into a running DataHub: inherited tags appear on tables/columns with the **Propagated** badge and correct origin; directly-assigned tags are unmarked.

## Review follow-ups (addressed)

- Surfaced previously-silent `except` blocks in `get_table_lf_tags` / `get_database_lf_tags` via `report.warning`; removed a stray `print()` and `catalog_id` variable shadowing.
- Added a `model_validator` warning when `propagate_*` is set without `extract_lakeformation_tags`.
- Guarded `_parse_lf_tag_dicts` against entries missing `TagKey`.
- Documented the new `extract_lakeformation_column_tags` default behavior in `docs/how/updating-datahub.md`.

## Checklist

- [x] Links to related issues (ING-2163, #16173)
- [x] Tests added/updated
- [x] Docs: new config fields self-document; upgrade note added to `updating-datahub.md`
- [ ] Breaking changes: none (opt-in; column tags default-on but only emit when LF extraction is enabled)


sgomezvillamor (approved):
Clean implementation — zero extra API calls, correct LF override semantics, and solid behavioral test coverage. A few items before merging:

| Issue | Severity | Location | Action |
|---|---|---|---|
| `extract_lakeformation_column_tags` defaults to `True` — existing users with `extract_lakeformation_tags: true` will silently gain column tags after upgrade | 🔴 | `glue.py:267` | Change default to `False`, or document in `docs/how/updating-datahub.md` |
| Silent failure in `get_table_lf_tags` — bare `except Exception` with no logging; permission errors disappear with no user-visible indication | 🟡 | `glue.py:~540` | Use `self.report.warning(title=..., message=..., context=f"{catalog_id}.{db}.{table}", exc=e)` |
| `get_database_lf_tags` uses `print()` and shadows `catalog_id` inside the tag loop (corrupts error messages) | 🟡 | `glue.py:~465` | Switch to `self.report.warning()`; rename loop variable to avoid shadowing |
| `propagate_database_lakeformation_tags_to_tables` also causes table+database tags to cascade to all columns when `extract_lakeformation_column_tags` is on — the name doesn't capture this | 🟡 | `glue.py:260` | Either add a separate `propagate_lakeformation_tags_to_columns` flag for independent control, or update the description to make the column-inheritance side-effect explicit |
| No validation that the sub-flags require `extract_lakeformation_tags: true`; silently has no effect if parent is off | ℹ️ | `glue.py` | Add `@model_validator` with a logger warning |
| `_parse_lf_tag_dicts` doesn't guard against a missing `TagKey` in malformed responses | ℹ️ | `glue.py` | Add `if not tag_key: continue` with a warning |
| No test for multi-value `TagValues` per key; no test that errors surface to `report.warnings` | ℹ️ | `test_glue_source.py` | Add 2 tests |

The `default=True` change and the flag-naming question are the two I'd resolve before merging.

---

PR #17815, opened 2026-06-08
fix(lookml): Preserve cross-project view assignments in Looker ingestion

sgomezvillamor:
## Description

Fixes a bug in the Looker ingestion source where cross-project view imports were incorrectly having their project assignments overridden by the explore's project name.

### Changes

1. **Added `extract_project_from_imported_file_path()` function**: Extracts the project name from imported_projects/ file paths to identify cross-project imports.

2. **Updated `preprocess_file_path()` method**: Now uses the project name embedded in the file path itself (for cross-project imports) when stripping the `imported_projects/` prefix, rather than always using `self.project_name`. This ensures the prefix strip succeeds even when processing an explore from a different project.

3. **Simplified `create_view_project_map()` logic**: Removed the override logic that was incorrectly replacing field-level project assignments with the explore's project name. Now the function respects the project assignment from the source file path for all cross-project imports. Same-project views (with `project_name=None`) continue to fall back to the explore's project via the `BASE_PROJECT_NAME` sentinel.

### Testing

Added comprehensive unit tests covering:
- `extract_project_from_imported_file_path()`: Tests extraction from various file path formats
- `create_view_project_map()`: Regression tests ensuring cross-project primary views are not overridden, cross-project non-primary views are preserved, and same-project views are excluded from the map

All new tests pass and validate the fix.

https://claude.ai/code/session_01Ak3UPKVEjgoAZiJNEpJf41

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/looker/looker_common.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/looker/looker_common.py
@@ -219,8 +219,13 @@ class LookerViewId:
         for pattern in str_to_remove:
             new_file_path = re.sub(pattern, "", new_file_path)
 
+        # Use the project embedded in the file path itself when it is a cross-project import so
+        # that the prefix strip succeeds even if self.project_name is the explore's project.
+        project_in_path = (
+            extract_project_from_imported_file_path(new_file_path) or self.project_name
+        )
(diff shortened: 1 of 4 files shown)


alokr-dhub:
closing this as the changes are rebased here: https://github.com/datahub-project/datahub/pull/19306 with follow up fixes.

PR #17820, opened 2026-06-09
feat(ingestion/quicksight): add Amazon QuickSight source connector

alfiyas-datahub:
## Summary
Adds a new pull-based ingestion connector for Amazon QuickSight.

The connector extracts the QuickSight asset graph and maps it onto DataHub entities:

- **Dashboards & Analyses** — ingested as DataHub dashboards with their charts/visuals.
- **Datasets** (subtype `Dataset`). Data Sources (the raw warehouse connections) are
  **not** emitted as entities — like Tableau/Looker/PowerBI, they're used only as
  resolution metadata to stitch a dataset's lineage to its upstream platform.
- **Folder hierarchy** — shared (and restricted) folders modeled with `browsePathsV2`,
  with an opt-in synthetic "Shared folders" root container (`add_shared_folders_container`)
  so top-level folders nest under it instead of appearing flattened. Personal
  ("My folders") folders are not exposed by the QuickSight API and are not ingested.
- **Lineage** — table-level and column-level, including custom SQL parsing (via sqlglot)
  to resolve upstream source tables/columns. Lineage stitches across platforms
  (e.g. Athena, Redshift). S3-backed datasets are an exception: QuickSight only exposes
  the manifest location, not the underlying data path, so S3 upstream lineage is skipped
  (with a report warning) rather than emitting a dangling edge.
- **Ownership & Tags** — ownership derived from `Describe*Permissions` (owner = principal
  with `Update*Permissions`), AWS resource tags from `ListTagsForResource`. SSO/IAM-federated
  principals are normalised to email-based `CorpUser` ids, consistent with the
  PowerBI/Tableau/Sigma/Looker/Informatica connectors.
- **Users & Groups** — `CorpUser` / `CorpGroup` entities with memberships (opt-in).
- **Stateful ingestion** — stale entities are automatically soft-deleted across runs.

Also fixes a UI bug where facet labels were double-pluralized (Analysiss, Shared Folderss,
Axiss) by teaching `pluralizeIfIrregular` about irregular plurals and already-plural words.

## Changes
- New source package under `metadata-ingestion/src/datahub/ingestion/source/quicksight/`
  (config, API client, lineage, URN helpers, report, and per-asset extractors).
- Registered the `quicksight` plugin entry point and dependency extra in `setup.py` and
  `pyproject.toml` (incl. `quicksight` in the `boto3-stubs` extra for type checking).
- Added `QUICKSIGHT_DATASET = "Dataset"` subtype.
- Connector docs (`README.md`, `quicksight_pre.md`, `quicksight_post.md`, recipe) and
  `integrations_catalog.json` updated to supported.
- Frontend pluralization fix in `textUtil.ts` + tests.

## Test plan
- [x] Unit tests for config, constants, lineage, enrichment, users/groups, stateful
      ingestion, and each asset extractor
- [x] Golden-file integration test (`test_quicksight.py`) using a mocked QuickSight client
- [x] Frontend pluralization tests (vitest)
- [x] End-to-end validated against a live QuickSight account: subtypes, folder hierarchy,
      charts, table + column lineage, ownership URN format, live tag extraction, and
      stateful checkpoint round-trip


metadata-ingestion/src/datahub/ingestion/source/quicksight/extractors/containers.py
```diff
@@ -0,0 +1,365 @@
...
+            yield from self._emit_shared_folders_root()
+
+        for summary in ordered:
+            folder_id = summary["FolderId"]
+            name = str(summary.get("Name") or folder_id)
+            if not self.config.folder_pattern.allowed(name):
+                self.report.containers.dropped(f"folder:{folder_id}")
+                continue
+
+            parent_folder_id = self._parent_folder_id(folder_id)
```
puneetagarwal-datahub: describe_folder ends up getting called twice per folder here: once for the sort key via _parent_folder_ids(s["FolderId"]), and again inside the loop via _parent_folder_id(folder_id) → _parent_folder_ids. Since quicksight_post.md calls out QuickSight's per-API TPS throttling, this doubles DescribeFolder calls for no benefit. Could we memoize _parent_folder_ids (a Dict[str, List[str]] cache keyed by folder_id) so each folder is described once?

puneetagarwal-datahub: Whenever we are calling QuickSight's API . Try to memoize

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_config.py
```diff
@@ -0,0 +1,211 @@
...
+        default=True,
+        description="Whether to fetch full dashboard definitions (sheets/visuals). "
+        "These payloads are large; disabling this skips Chart entity emission.",
+    )
+    extract_analysis_definitions: bool = Field(
+        default=True,
+        description="Whether to fetch full analysis definitions (sheets/visuals). "
+        "These payloads are large.",
+    )
+    extract_usage_stats: bool = Field(
```
puneetagarwal-datahub: not being used

datahub-web-react/src/app/shared/textUtil.ts
```diff
@@ -33,17 +33,24 @@ export function pluralize(count: number, noun: string, suffix = 's') {
...
 
-    if (irregularPlurals.hasOwnProperty(noun?.toLowerCase())) {
-        return irregularPlurals[noun?.toLowerCase()];
+    const lower = noun?.toLowerCase();
+    if (irregularPlurals.hasOwnProperty(lower)) {
+        return irregularPlurals[lower];
+    }
+    // Don't double-pluralize names that are already plural (e.g. "Shared Folders",
+    // "Data Sources"). Only applies to the default "s" suffix.
+    if (suffix === 's' && lower?.endsWith('s')) {
```
puneetagarwal-datahub, line 62: This will pass nouns like example : axis 

also every noun cant be pluralize by appending s

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_urn.py
```diff
@@ -0,0 +1,71 @@
...
+Centralized so the Dataset URNs referenced by dashboards/analyses lineage stay
+byte-identical to the URNs the dataset processor emits.
+"""
+
+from typing import Optional
+
+import datahub.emitter.mce_builder as builder
+from datahub.metadata.urns import DashboardUrn
+
+PLATFORM = "quicksight"
```
puneetagarwal-datahub, line 12: this is defined in 4-5 places . can we define it once ?

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_config.py
```diff
@@ -0,0 +1,211 @@
...
+    )
+
+    stateful_ingestion: Optional[StatefulStaleMetadataRemovalConfig] = Field(
+        default=None,
+        description="Stateful ingestion configuration (enables stale entity removal).",
+    )
+
+    def get_quicksight_client(self) -> "QuickSightClient":
+        # adaptive retry mode is recommended for QuickSight's per-API TPS throttling.
+        return self.get_session().client("quicksight", config=self._aws_config())
```
puneetagarwal-datahub, line 218: how are we making adaptive retry mode working here ?

metadata-ingestion/src/datahub/ingestion/source/quicksight/processors/data_sources.py
```diff
@@ -0,0 +1,182 @@
...
+
+            if not self.config.data_source_pattern.allowed(name):
+                self.report.data_sources.dropped(f"{data_source_id} ({name})")
+                continue
+
+            platform = DATA_SOURCE_TYPE_TO_PLATFORM.get(quicksight_type)
+            dialect = DATA_SOURCE_TYPE_TO_DIALECT.get(quicksight_type)
+            if platform is None:
+                self.report.num_unknown_data_source_types += 1
+                logger.info(
```
puneetagarwal-datahub: Use structured self.report.warning(title=…, message=…, context=…, exc=e) for recoverable errors and failure() for critical ones.

puneetagarwal-datahub: check all loggers

metadata-ingestion/src/datahub/ingestion/source/quicksight/processors/users_groups.py
```diff
@@ -0,0 +1,160 @@
...
+            email=user.get("Email") or None,
+            groups=group_names or None,
+        )
+        self.report.corp_users.processed(user_id)
+        # override_editable=False keeps us on the ingestion-owned CorpUserInfo
+        # aspect rather than the UI-editable one.
+        for mcp in corp_user.generate_mcp(
+            generation_config=CorpUserGenerationConfig(override_editable=False)
+        ):
+            yield mcp.as_workunit()
```
puneetagarwal-datahub: QuickSight emits CorpUser/CorpGroup via mcp.as_workunit() (is_primary_source=True), so they enter the stale-entity-removal checkpoint and can be soft-deleted when extract_users_and_groups is toggled off or the identity APIs degrade. PowerBI handles this by emitting user MCPs with is_primary_source=False precisely to keep global identities out of stale removal. Recommend matching that here: mcp.as_workunit(is_primary_source=False) in users_groups.py.

Check it once

puneetagarwal-datahub (changes requested):

puneetagarwal-datahub:
We need to refactor our code according to work/refactoring done by @sgomezvillamor 
https://github.com/datahub-project/datahub/pull/17852

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_urn.py
```diff
@@ -0,0 +1,71 @@
...
+    )
+
+
+def make_dashboard_urn(asset_id: str, platform_instance: Optional[str]) -> str:
+    """Build a Dashboard URN for a QuickSight analysis or dashboard.
+
+    Both analyses and dashboards map to DataHub Dashboard entities; the URN is
+    ``(quicksight, {id})`` with the platform instance folded in when configured.
+    """
+    return DashboardUrn.create_from_ids(
```
puneetagarwal-datahub, line 75: Dataset URNs embed {account_id}.{id} + env, but dashboards/analyses/charts use the bare (quicksight, {id}) with no account_id and no env (confirmed in the golden: urn:li:dashboard:(quicksight,analysis-sales)). So two AWS accounts — or two envs of one account — ingested without a distinct platform_instance will silently collapse distinct dashboards/charts onto the same URN and overwrite each other.

puneetagarwal-datahub:
Since this quicksight feature is API Limit sensitive  , I would like see number of calls to aws per flow when you prepare test doc

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_lineage.py
```diff
@@ -0,0 +1,228 @@
...
+    def _handle_s3_source(self, entry: dict) -> List[str]:
+        target = self._target_for(entry.get("DataSourceArn", ""))
+        if target is None or not target.resolved.s3_manifest_uri:
+            return []
+        # The manifest URI already encodes bucket/key; S3 datasets carry no
+        # platform instance.
+        return [
+            builder.make_dataset_urn_with_platform_instance(
+                platform=S3_PLATFORM,
+                name=target.resolved.s3_manifest_uri,
```
puneetagarwal-datahub: S3 lineage points at the manifest file, not the data. 
quicksight_lineage.py:161 uses s3_manifest_uri (bucket/manifest.json) as the upstream S3 dataset name. DataHub's S3 source keys datasets by the data path/prefix, so this edge resolves to urn:li:dataset:(s3, bucket/manifest.json) — which won't match any real S3-ingested dataset. Every S3-backed QuickSight dataset gets dangling/phantom lineage. Objectively wrong output (scoped to S3 data sources).

puneetagarwal-datahub (changes requested):

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_urn.py
alfiyas-datahub: The golden file already includes the AWS account ID in the asset name (for example, 123456789012.analysis-sales), so assets from different accounts won't conflict with each other.

The missing env field for Dashboard and Chart URNs is a limitation of the current DataHub model and is the same behavior used by other sources like PowerBI and Tableau. For separating multiple ingestions from the same account, platform_instance is the recommended approach. I can add a note to the documentation if that would help.

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_urn.py
alfiyas-datahub: Done, single definition in quicksight_urn.py all extractors import PLATFORM from there. The only other occurrence is QuickSightSource.platform in quicksight.py, which is the required Source class attribute.

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_config.py
alfiyas-datahub: Yes, the QuickSight connector already uses adaptive retry mode by default. The configuration sets aws_retry_mode to "adaptive" and passes it to the AWS client through the shared AWS configuration. No additional changes are needed in get_quicksight_client() since it automatically uses this shared configuration.

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_config.py
```diff
@@ -0,0 +1,218 @@
...
+from datahub.ingestion.source.aws.aws_common import AwsConnectionConfig
+from datahub.ingestion.source.state.stale_entity_removal_handler import (
+    StatefulStaleMetadataRemovalConfig,
+)
+from datahub.ingestion.source.state.stateful_ingestion_base import (
+    StatefulIngestionConfigBase,
+)
+
+
+class ExternalDataSourceConfig(PlatformInstanceConfigMixin, EnvConfigMixin):
```
puneetagarwal-datahub: how are we using this ?

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_config.py
```diff
@@ -0,0 +1,218 @@
...
+    AwsConnectionConfig,
+):
+    """Configuration for the Amazon QuickSight ingestion source.
+
+    Reuses :class:`AwsConnectionConfig` for AWS authentication (explicit keys,
+    named profile, IAM role assumption, or instance-role auto-detection).
+    QuickSight is regional, so a single ingestion run targets one ``aws_region``.
+    """
+
+    aws_account_id: Optional[str] = Field(
```
puneetagarwal-datahub, line 78: use same structure as other aws connectors

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_config.py
```diff
@@ -0,0 +1,218 @@
...
+        "Enable it for Enterprise accounts that use multiple namespaces. Mirrors "
+        "Tableau's `add_site_container`. Account separation is handled via "
+        "`platform_instance`, not a container — matching the Glue / Redshift / "
+        "PowerBI / Informatica convention.",
+    )
+
+    # --- Filtering ---
+    namespace_pattern: AllowDenyPattern = Field(
+        default=AllowDenyPattern.allow_all(),
+        description="Regex patterns to filter QuickSight namespaces (only relevant "
```
puneetagarwal-datahub, line 160: how are we filtering namespaces here ?

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_config.py
```diff
@@ -0,0 +1,218 @@
...
+        "when `add_namespace_container` is enabled).",
+    )
+    folder_pattern: AllowDenyPattern = Field(
```
puneetagarwal-datahub, line 163: check for my folder inclusion in this

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_config.py
```diff
@@ -0,0 +1,218 @@
...
+    )
+    dataset_pattern: AllowDenyPattern = Field(
+        default=AllowDenyPattern.allow_all(),
+        description="Regex patterns to filter QuickSight datasets by name.",
+    )
+    data_source_pattern: AllowDenyPattern = Field(
+        default=AllowDenyPattern.allow_all(),
+        description="Regex patterns to filter QuickSight data sources by name.",
+    )
+    tag_pattern: AllowDenyPattern = Field(
```
puneetagarwal-datahub, line 185: check this tag_pattern

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_lineage.py
alfiyas-datahub: Fixed it. QuickSight only provides the S3 manifest location, not the actual data file paths, so the lineage we generated couldn't correctly link to S3 datasets in DataHub. Instead of creating incorrect lineage, the connector now skips S3 upstream lineage and logs a warning.

puneetagarwal-datahub (changes requested):

datahub-web-react/src/app/shared/__tests__/pluralize.test.ts
```diff
@@ -1,4 +1,4 @@
-import { pluralize } from '@app/shared/textUtil';
+import { forcePluralize, pluralize, pluralizeIfIrregular } from '@app/shared/textUtil';
 
 describe('pluralize text based on the count', () => {
     it('pluralize regular word with count greater than 1', () => {
```
puneetagarwal-datahub, line 4: Check the flow where if a external user has shared a folder , we are able to get that information

metadata-ingestion/docs/sources/quicksight/quicksight_pre.md
```diff
@@ -0,0 +1,79 @@
...
+      "Effect": "Allow",
+      "Action": [
+        "sts:GetCallerIdentity",
+        "quicksight:ListDashboards",
+        "quicksight:ListAnalyses",
+        "quicksight:ListDataSets",
+        "quicksight:ListDataSources",
+        "quicksight:ListFolders",
+        "quicksight:ListFolderMembers",
+        "quicksight:ListNamespaces",
```
puneetagarwal-datahub, line 38: hows namespace being used ?

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_config.py
alfiyas-datahub: This is an optional configuration used to help QuickSight lineage connect correctly to datasets ingested from other systems like Snowflake or Redshift. It lets you specify settings such as platform_instance, env, naming conventions, and default database/schema so the generated URNs match the upstream connector.

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_config.py
alfiyas-datahub: QuickSightSourceConfig already uses the same common AWS configuration components as other AWS connectors. It reuses shared settings for authentication, region, profiles, roles, environment, platform instance, and stateful ingestion, keeping it consistent with connectors like Glue and SageMaker.

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_config.py
alfiyas-datahub: Namespaces appear only if you set **add_namespace_container**: **true**  then you’d see Container entities with subtype "Namespace" (e.g. default, marketing). By default that level is off, so you won’t see namespace containers in DataHub either.

In QuickSight UI there’s no user-facing namespace list, use list-namespaces via CLI/API; gives **default** only unless you’re on Enterprise with extra namespaces configured.

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_config.py
alfiyas-datahub: Verified, **Shared Folders** are folders that have been shared with you and appear under "Shared folders" in QuickSight. They can contain dashboards, datasets, analyses, and other assets that multiple users can access. Private folders under "My folders" belong only to the user and are not visible through the shared folder APIs.

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_config.py
alfiyas-datahub: tag_pattern filters which AWS resource tags (as key=value) are copied from QuickSight into DataHub when extract_tags is enabled. Users then search, filter assets by those tags in the DataHub UI.

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_lineage.py
treff7es: **Must fix — claimed "Fixed" but the code is unchanged.** Your Jun 22 reply states the connector now *skips* S3 upstream lineage and logs a warning, but `_handle_s3_source` still emits `urn:li:dataset:(s3, <bucket>/manifest.json)` from `s3_manifest_uri`. DataHub's S3 source keys datasets by the **data path**, not `manifest.json`, so this upstream is a phantom node that won't stitch to the real S3 dataset.

Please either (a) implement the described skip-and-warn, or (b) keep manifest-based lineage as a documented best-effort limitation — but in either case reconcile the resolved review thread so it matches the code.

metadata-ingestion/src/datahub/ingestion/source/quicksight/extractors/data_sets.py
```diff
@@ -0,0 +1,171 @@
...
+    def get_workunits(self) -> Iterable[MetadataWorkUnit]:
+        for summary in self.api.list_data_sets():
+            data_set_id = summary.get("DataSetId")
+            name = summary.get("Name") or data_set_id or ""
+            if not data_set_id:
+                continue
+            if not self.config.dataset_pattern.allowed(name):
+                self.report.datasets.dropped(f"{data_set_id} ({name})")
+                continue
+            yield from self._emit_data_set(data_set_id, name, summary)
```
treff7es: **Must fix — no per-asset error isolation.** This loop (and the equivalent loops in `dashboards.py`, `analyses.py`, `data_sources.py`, `visuals.py`, `containers.py`) calls `_emit_*` without a surrounding try/except. The `_describe`/enrichment helpers self-guard, but anything outside them — a malformed ARN in `make_dataset_urn`/`id_from_arn`, SDK validation inside `Dataset(...)`, a `KeyError` on an unexpected response shape — propagates up. Since `get_workunits_internal` isn't wrapped, one bad asset tears down this extractor **and every extractor scheduled after it**, so a failure here also loses analyses, dashboards and users/groups.

Wrap each per-asset iteration so a single failure degrades to a warning and continues:
```python
try:
    yield from self._emit_data_set(data_set_id, name, summary)
except Exception as e:
    self.report.warning(
        title="Failed to process dataset",
        message="Skipping this dataset; the rest of the run continues.",
        context=f"{data_set_id} ({name})",
        exc=e,
    )
```

metadata-ingestion/src/datahub/ingestion/source/quicksight/extractors/dashboards.py
```diff
@@ -0,0 +1,152 @@
...
+                self.api.aws_account_id,
+                id_from_arn(source_arn),
+                self.config.platform_instance,
+            )
+        ]
+
+    def _describe(self, dashboard_id: str) -> Dict[str, Any]:
+        try:
+            return self.api.describe_dashboard(dashboard_id)
+        except ClientError as e:
```
treff7es: These `_describe` / `_extract_charts` helpers catch only `botocore.exceptions.ClientError`, but botocore raises siblings that are **not** `ClientError`: `EndpointConnectionError`, `ConnectionClosedError`, `ReadTimeoutError`, `BotoCoreError` — all common transient conditions. Combined with the missing per-asset isolation above, any of these aborts the run.

`data_sources.py` and `enrichment.py` already catch broad `Exception` for the same class of degradable describe/enrich call. Please standardize on that convention here (and in `analyses.py` / `data_sets.py` `_describe`).

metadata-ingestion/setup.py
```diff
@@ -842,6 +842,7 @@
     "dlt": {"dlt>=1.0.0,<2.0.0"},
     "snaplogic": set(),
     "qlik-sense": sqlglot_lib | {"requests<3.0.0", "websocket-client<2.0.0"},
+    "quicksight": aws_common | sqlglot_lib,
```
treff7es, line 852: The QuickSight connector annotates its client as `QuickSightClient` (`from mypy_boto3_quicksight import ...`), but `quicksight` isn't in the `boto3-stubs[...]` extra (line ~912: `[s3,glue,sagemaker,sts,dynamodb, lakeformation]`). Because `ignore_missing_imports` is on, mypy silently treats the client as `Any`, so the typed annotations and the `cast(Dict[str, Any], ...)` block in `quicksight_api.py` provide no real checking.

Add `quicksight` to the `boto3-stubs` extra and run `updateLockFile`; the casts can then mostly be dropped. (Also worth fixing the stray space in `dynamodb, lakeformation`.)

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_config.py
```diff
@@ -0,0 +1,218 @@
...
+        "via `sts:GetCallerIdentity` when not provided.",
+    )
+
+    # --- Lineage & enrichment toggles ---
+    extract_lineage: bool = Field(
+        default=True,
+        description="Whether to extract upstream lineage from QuickSight Datasets to "
+        "their backing warehouse/database tables.",
+    )
+    extract_column_lineage: bool = Field(
```
treff7es: Naming alignment: the ecosystem-dominant name for this sub-toggle is `include_column_lineage` (used by BigQuery, Unity Catalog, dbt, Fivetran, ADF, …) rather than `extract_column_lineage`. Since the connector is unreleased, a straight rename is cheapest.

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_config.py
```diff
@@ -0,0 +1,218 @@
...
+        default=True,
+        description="Whether to extract column-level lineage for CustomSql datasets "
+        "via sqlglot. Requires `extract_lineage` to be enabled.",
+    )
+    extract_ownership: bool = Field(
+        default=True,
+        description="Whether to extract ownership from QuickSight resource permissions.",
+    )
+    strip_user_email_domain: bool = Field(
```
treff7es: Naming alignment: the dominant BI-connector name for this is Looker's `strip_user_ids_from_email` (27 uses) rather than `strip_user_email_domain` (6, mostly Informatica). Your own description already notes it "Mirrors … Looker's `strip_user_ids_from_email`" — consider just adopting that name (with a `pydantic_renamed_field` alias).

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_config.py
```diff
@@ -0,0 +1,218 @@
...
+    """Per-data-source overrides used when stitching QuickSight Datasets to
+    their upstream warehouse/database tables.
+
+    Keyed in :class:`QuickSightSourceConfig.external_data_sources` by the stable
+    QuickSight ``DataSourceId`` (UUID) — preferred because it survives renames in
+    the QuickSight UI — with the data source display name accepted as a fallback.
+    UUID matches take precedence over name matches.
+    """
+
+    convert_urns_to_lowercase: bool = Field(
```
treff7es, line 46: `convert_urns_to_lowercase` is hand-redefined here, but `LowerCaseDatasetUrnConfigMixin` exists precisely for this field. Prefer inheriting the mixin so the default/description/semantics don't drift independently:
```python
class ExternalDataSourceConfig(
    LowerCaseDatasetUrnConfigMixin,
    PlatformInstanceConfigMixin,
    EnvConfigMixin,
):
```
You can keep your richer description as a field override after inheriting if you want to preserve the "must match the upstream connector" note.

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_config.py
treff7es, line 163: Open question from earlier review re: "My Folder" / PERSONAL folders. `list_folders()` doesn't filter by `FolderType`, and the prior reply only discussed SHARED folders. Could you clarify the scope in the docs — are PERSONAL folders ingested, intentionally excluded, or controllable via `folder_pattern`? That was the substance of the reviewer's question.

metadata-ingestion/src/datahub/ingestion/source/quicksight/extractors/data_sources.py
```diff
@@ -0,0 +1,186 @@
...
+            "dataSourceId": data_source_id,
+        }
+        custom_properties.update(self._connection_properties(data_source))
+
+        # Data sources are never QuickSight folder members, so this resolves to
+        # the namespace container (if enabled) or the platform root. An explicit
+        # empty list (rather than None) overwrites any stale browse path when the
+        # asset attaches directly to the platform root.
+        parent = self.parent_resolver(data_source_id)
+        dataset = Dataset(
```
treff7es: **Modelling — don't emit QuickSight Data Sources as entities.** Right now each Data Source (`ds-athena`, `ds-redshift`) becomes a standalone `dataset` (subType `Data Source`) with no schema, no container, and no lineage in or out — the QuickSight Dataset's lineage points directly at the warehouse table (`redshift,dev.public.orders`), so these nodes sit orphaned in the graph.

No other BI connector models the raw warehouse *connection* as a dataset entity. Tableau/Looker/PowerBI treat the connection purely as resolution metadata for the upstream table's platform — it never becomes its own node. (The thing that maps to Tableau's "Published/Embedded Data Source" is the QuickSight *Dataset*, which this connector already models well: schema + table/column lineage + container.)

Recommend dropping the Data Source entity emission entirely — don't emit the `dataset` and don't surface the connection params as custom properties either. Keep building the `ResolvedDataSource` map (line ~94): it's still needed to resolve the upstream platform/dialect for lineage. Just stop the `Dataset(...)` / `as_workunits()` emission here (lines ~132-145), drop `SUBTYPE_DATA_SOURCE`, and remove the data-source golden entities/tests.

metadata-ingestion/src/datahub/ingestion/source/quicksight/extractors/users_groups.py
```diff
@@ -0,0 +1,178 @@
...
+    QuickSightSourceConfig,
+)
+from datahub.ingestion.source.quicksight.quicksight_report import (
+    QuickSightSourceReport,
+)
+from datahub.ingestion.source.quicksight.quicksight_urn import quicksight_user_id
+
+# QuickSight identity APIs are Enterprise-edition only and need extra IAM
+# actions; degrade gracefully on these codes rather than failing the run.
+_GRACEFUL_DEGRADE_CODES = frozenset(
```
treff7es: Nice-to-have (DRY): this `_GRACEFUL_DEGRADE_CODES` frozenset + `_is_graceful` is byte-identical to `_GRACEFUL_DEGRADE_CODES` + `_graceful_code` in `containers.py:35`. Two copies that must stay in sync. Consider hoisting one `_graceful_code(error) -> Optional[str]` helper into `quicksight_api.py` (or a small `quicksight_errors.py`) and importing it in both.

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_api.py
```diff
@@ -0,0 +1,226 @@
...
+        )
+        return cast(Dict[str, Any], response.get("Folder") or {})
+
+    def list_folder_members(self, folder_id: str) -> Iterator[Dict[str, Any]]:
+        """Yield the members (assets) of a folder.
+
+        Each member is ``{"MemberId": <asset_id>, "MemberArn": <arn>}`` where the
+        ARN's resource type (dashboard/analysis/dataset) identifies the asset
+        kind. ``list_folder_members`` has no boto3 paginator, so we follow
+        ``NextToken`` manually.
```
treff7es, line 96: Nice-to-have (test gap): this is the only hand-rolled paginator in the connector, but both the integration `FakeQuickSightClient` and the unit mocks return a single page with no `NextToken`, so the loop never iterates. A folder with >100 members (common in real accounts) would silently drop membership beyond page 1 and mis-parent assets. Worth a focused unit test driving a two-page response.

datahub-web-react/src/app/shared/textUtil.ts
```diff
@@ -49,10 +49,17 @@ export function pluralizeIfIrregular(noun: string, suffix = 's'): string {
     const irregularPlurals: Record<string, string> = {
         query: 'queries',
         match: 'matches',
+        analysis: 'analyses',
```
treff7es, line 52: Nice-to-have: the earlier review flagged `axis` by name, but it's still not handled — the irregular map has `query`/`match`/`analysis` but not `axis`, and the `endsWith('s')` fallback returns `"axis"` unchanged instead of `"axes"`. Add `axis: 'axes'` (and a test case) if QuickSight surfaces that facet.

alfiyas-datahub: Added axis: 'axes' to the irregular-plurals map in textUtil.ts plus a test case

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_api.py
alfiyas-datahub: Done — added test_list_folder_members_follows_next_token_across_pages, which drives a two-page response and asserts members from both pages are returned and the NextToken is forwarded.

metadata-ingestion/src/datahub/ingestion/source/quicksight/extractors/users_groups.py
alfiyas-datahub: Done, I moved the shared error handling code into a new `quicksight_errors.py` file. Both `containers.py` and `users_groups.py` now reuse it instead of having duplicate code.

metadata-ingestion/src/datahub/ingestion/source/quicksight/extractors/data_sources.py
alfiyas-datahub: Done,  the connector no longer creates Data Source entities or their custom properties. It still builds the internal data source map needed to generate correct dataset-to-warehouse lineage. I also removed the related subtype, tests, and golden outputs, and verified that lineage for sources like Redshift continues to work correctly.

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_config.py
alfiyas-datahub: Updated the doc. ListFolders only returns **SHARED** and **RESTRICTED** folders — both are ingested and filterable via folder_pattern. Personal "My folders" aren't exposed by any QuickSight API, so they can't be ingested; assets living only there still ingest and attach to the namespace/platform root. Added this as a Limitations note.

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_config.py
alfiyas-datahub: **ExternalDataSourceConfig** now reuses the shared **LowerCaseDatasetUrnConfigMixin** instead of defining the setting itself. The field is kept only to provide connector-specific documentation about matching the upstream connector's configuration.

metadata-ingestion/src/datahub/ingestion/source/quicksight/extractors/data_sets.py
alfiyas-datahub: wrapped each per-asset iteration in try/except

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_lineage.py
alfiyas-datahub: The previous skip-and-warn change was accidentally lost during a rebase and wasn't included in the branch. I've fixed it now: `_handle_s3_source` no longer creates incorrect S3 lineage from the manifest file, and instead skips it, logs a warning, and tracks how many were skipped. The manifest location is still stored as a custom property, and the tests and documentation have been updated accordingly.

metadata-ingestion/src/datahub/ingestion/source/quicksight/extractors/dashboards.py
alfiyas-datahub: The error handling now catches all AWS SDK exceptions, not just ClientError, so connection and timeout errors are handled gracefully as well. The special handling for FILE datasets is unchanged, and I added a test to verify that EndpointConnectionError is handled correctly.

metadata-ingestion/setup.py
alfiyas-datahub: I added quicksight to the boto3-stubs dependency and updated the lock files. I also renamed `extract_column_lineage` to `include_column_lineage` and `strip_user_email_domain` to `strip_user_ids_from_email` (with backward compatibility). I kept the type casts in `quicksight_api.py` because they're still needed to satisfy mypy type checking.

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_config.py
```diff
@@ -0,0 +1,232 @@
...
+        "via `sts:GetCallerIdentity` when not provided.",
+    )
+
+    # --- Lineage & enrichment toggles ---
+    extract_lineage: bool = Field(
+        default=True,
+        description="Whether to extract upstream lineage from QuickSight Datasets to "
+        "their backing warehouse/database tables.",
+    )
+    include_column_lineage: bool = Field(
```
treff7es, line 90: Missing dependency validator. The description says this "Requires `extract_lineage` to be enabled," but nothing enforces it. With `include_column_lineage=True, extract_lineage=False`, `get_upstream_lineage` early-returns (`quicksight_lineage.py:80`) and column-level lineage is **silently dropped** with no warning. Add a `model_validator(mode="after")` that errors (or at least warns) when `include_column_lineage and not extract_lineage`. Highest-value of the inline items since it's a real silent misconfiguration.

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_lineage.py
```diff
@@ -0,0 +1,229 @@
...
+        resolved = self.data_source_map.get(data_source_arn)
+        if resolved is None:
+            self.report.warning(
+                title="Data source not resolved for lineage",
+                message="A dataset references a data source that was not ingested "
+                "(e.g. filtered out or cross-account); skipping its upstream lineage.",
+                context=data_source_arn,
+            )
+            return None
+        if resolved.platform is None:
```
treff7es, line 136: Silent lineage drop. When a resolved data source has no upstream-platform mapping, `_target_for` returns `None` here with no counter or warning — so the per-dataset lineage drop is invisible. The unsupported-type warning fires once at map-build time, but at resolution time the user just sees a dataset with missing upstreams and no indication why. Suggest bumping a report counter (e.g. `num_lineage_skipped_unsupported_platform`) so the drop is observable.

metadata-ingestion/src/datahub/ingestion/source/quicksight/extractors/containers.py
```diff
@@ -0,0 +1,382 @@
...
+        """
+        cached = self._parent_folder_ids_cache.get(folder_id)
+        if cached is not None:
+            return cached
+        try:
+            folder = self.api.describe_folder(folder_id)
+        except Exception as e:
+            if graceful_code(e) is None:
+                raise
+            self._parent_folder_ids_cache[folder_id] = []
```
treff7es, line 352: Silent graceful-degrade. On a graceful error (e.g. `AccessDenied` on a single `DescribeFolder`) this caches `[]` and returns with **no `report.warning`** — unlike every other `graceful_code` site in this file. The folder is then treated as top-level (wrong parent) and its emit-ordering sort key is wrong, all silently. Add a `self.report.warning(title="Could not resolve folder ancestry", message="Folder may be placed at the wrong level in the hierarchy.", context=folder_id, exc=e)` before returning `[]`.

metadata-ingestion/src/datahub/ingestion/source/quicksight/extractors/containers.py
```diff
@@ -0,0 +1,382 @@
...
+        try:
+            for ns in self.api.list_namespaces():
+                name = ns.get("Name")
+                if name:
+                    namespace_names.append(name)
+        except Exception as e:
+            code = graceful_code(e)
+            if code is None:
+                raise
+            self.report.warning(
```
treff7es, line 186: Pass `exc=e` so the stack trace is retained (Sentry/debug). Right now only the AWS error code string is passed as `context` and the exception object is dropped. Same fix needed in `_emit_folders` below (`containers.py:247`).

treff7es (approved):
Re-review — thanks for the thorough turnaround. I re-verified every item from the previous round against the current code and they're **all genuinely fixed** (not just claimed): S3 manifest lineage now skips + warns instead of emitting a dangling edge, per-asset error isolation is in every loop, the `except` clauses now catch botocore siblings (not just `ClientError`), `quicksight` is in the `boto3-stubs` extra, the config renames + `LowerCaseDatasetUrnConfigMixin` are in, Data Sources are now a resolution-only map (no standalone entities — golden confirms), the PERSONAL/"My folders" scope is documented, the shared error helper is hoisted into `quicksight_errors.py`, users/groups are kept out of the stale-removal checkpoint via `is_primary_source=False`, and folder ancestry is memoized. Lineage uses the shared sqlglot `create_lineage_sql_parsed_result`, and the golden file is solid (58KB, 95 events).

**Approving.** The inline comments below are all non-blocking polish — they cluster around the silent-failure / missing-validator class and can be follow-ups.

**Suggestions (no inline comment, optional):**
- Dead fields: `ResolvedDataSource.arn` and `.quicksight_type` are write-only — never read (`arn` also duplicates the map key). Safe to drop.
- `analyses.py` and `dashboards.py` are ~90% duplicated (`get_workunits`/`_extract_charts`/`_describe`); a shared base would roughly halve ~300 lines. Behavior is correct, so optional.
- Type tightening: make `_UpstreamTarget` a frozen dataclass with non-`Optional` `platform` (deletes the two `# type: ignore[arg-type]`); consider TypedDicts for the `physical_table_map` boto3 shapes.
- Test gaps (coverage is otherwise strong): the `describe_folder` memoization isn't asserted via `call_count`, and the `_record_folder_members` graceful-failure branch is untested.
- Nit: the PR description still lists "Data Sources (subtype Data Source)" as ingested entities — stale now that they're map-only.

metadata-ingestion/src/datahub/ingestion/source/quicksight/quicksight_config.py
alfiyas-datahub: removed the unused extract_usage_stats

metadata-ingestion/src/datahub/ingestion/source/quicksight/extractors/containers.py
alfiyas-datahub: parent_folder_ids is now memoized via a Dict[str, List[str]] cache (_parent_folder_ids_cache, containers.py:108) keyed by folder_id, so each folder is described exactly once even though the result feeds both the emit-ordering sort key and the per-folder parent lookup.

datahub-web-react/src/app/shared/textUtil.ts
alfiyas-datahub: Added the irregular forms to the map (analysis → analyses, axis → axes) and the lookup now runs before the +s fallback, plus an already-plural guard so Shared Folders/Data Sources aren't double-pluralized.

metadata-ingestion/src/datahub/ingestion/source/quicksight/processors/data_sources.py
alfiyas-datahub: The unsupported-data-source path now uses self.report.warning(title="Unsupported data source type", message=..., context=...)

metadata-ingestion/src/datahub/ingestion/source/quicksight/processors/users_groups.py
alfiyas-datahub: Done, matching PowerBI. User and group metadata is now marked as non-primary, so DataHub won't accidentally delete existing users or groups if user/group extraction is turned off or the identity APIs are temporarily unavailable. This keeps user and group information safe during ingestion.

puneetagarwal-datahub (changes requested):

datahub-web-react/src/app/shared/__tests__/pluralize.test.ts
alfiyas-datahub: Yes, folders created or shared by other users are ingested as long as the ingesting service account has access to them. DataHub discovers those folders, their contents, and identifies the actual owner from the folder permissions, not the service account.

metadata-ingestion/docs/sources/quicksight/quicksight_pre.md
alfiyas-datahub: A QuickSight namespace is like a separate workspace inside a QuickSight account, used to keep different groups of users and their assets isolated. Most accounts only have the default namespace, while Enterprise accounts can have multiple. DataHub uses namespaces to place assets in the correct location and to discover users and groups from each namespace. If enabled, namespaces can also appear as a separate level in the browse hierarchy, but this is off by default since most customers only use the default namespace.

puneetagarwal-datahub (approved):

PR #17821, opened 2026-06-09
test(arch): add ArchUnit rule enforcing OperationContext as first param on AspectDao public methods

Dutt23:
## Summary

Adds an architectural enforcement test that requires every public method on `AspectDao` to accept `OperationContext` as its first parameter.

- New `@OperationContextExempt(reason="...")` annotation for methods that legitimately don't need a context (transaction helpers, lifecycle toggles, config accessors)
- New `AspectDaoOperationContextTest` using ArchUnit — one-liner to enforce any interface: `checkArch(AspectDao.class, Map.of(0, OperationContext.class))`
- 5 methods annotated as exempt with reasons

## Why

Prevents regression: any new `AspectDao` method added without `OperationContext` will fail CI immediately.


build.gradle
```diff
@@ -156,6 +156,7 @@ project.ext.externalDependency = [
     'pekkoJackson': "org.apache.pekko:pekko-serialization-jackson_$playScalaVersion:$pekkoVersion",
     'antlr4Runtime': 'org.antlr:antlr4-runtime:4.9.3',
     'antlr4': 'org.antlr:antlr4:4.9.3',
+    'archunit': 'com.tngtech.archunit:archunit:1.3.0',
```
kewats: Should we use 1.4.2, that being latest?

metadata-io/src/main/java/com/linkedin/metadata/entity/AspectDao.java
```diff
@@ -275,6 +275,9 @@ PartitionedStream<EbeanAspectV2> streamAspectBatches(
    * @return partitioned stream of matching rows
    */
   @Nonnull
+  @OperationContextExempt(
+      reason =
+          "TODO: Needs a bigger refactor, will he handled later. Streams need to fllow a consumer pattern")
```
kewats: nit - typo at `fllow`. Same for streamAspects method.

kewats (approved):

build.gradle
Dutt23: Updated version.

metadata-io/src/main/java/com/linkedin/metadata/entity/AspectDao.java
Dutt23: Thanks , fixed the typo.

PR #17826, opened 2026-06-09
fix(ingest/sql-parsing): guard against stack-overflow crashes on deeply nested SQL

treff7es:
## Summary

`sqlglot`'s parser, optimizer, lineage traversal, and SQL rendering are all recursive over a statement's nesting structure. Pathologically nested SQL — for example dynamic SQL emitted by stored procedures (concatenated `EXECUTE IMMEDIATE`, deeply nested CTEs/subqueries) — can recurse deep enough to overflow the stack.

In the pure-Python build this surfaces as a catchable `RecursionError`, but in the **mypyc-compiled `sqlglot[c]` build used in production** it overflows the **native C stack** into an **uncatchable `SIGSEGV`** that kills the entire `datahub ingest` subprocess with no traceback. No `except` can catch an OS signal, so the only defense is to stop *before* the recursion runs.

## Changes

Two complementary guards in `sqlglot_utils.py`, applied at the parse chokepoints (`parse_statement`, `parse_statements_and_pick`, `generalize_query`):

- **Pre-parse text-nesting guard** — a cheap, string-literal-aware bracket-depth scan that protects the recursive parser itself from "paren bombs". A `str.count` fast path (max depth can't exceed the number of opening brackets) keeps the common case nearly free, so the per-query fingerprinting path pays ~0.04–0.22%.
- **Post-parse AST-depth guard** — an **iterative** (non-recursive) AST walk that protects the optimizer / lineage / rendering from deep trees such as nested `CASE` expressions that carry no brackets for the text scan to catch.

When a statement exceeds the limits, it is logged (preview at `WARNING`, full SQL at `DEBUG` for later debugging) and a `StatementTooComplexError` is raised. It subclasses `sqlglot.errors.SqlglotError` so existing handlers degrade gracefully: lineage is skipped and reported via `debug_info`, and fingerprinting falls back to a string-based hash. Both limits are environment-tunable:

- `DATAHUB_SQL_PARSE_MAX_TEXT_NESTING` (default 500)
- `DATAHUB_SQL_PARSE_MAX_AST_DEPTH` (default 600)

Native stack frames are smaller than CPython frames, so limits safely below the pure-Python recursion ceiling (~1000) also protect the compiled build, while sitting well above any realistic query (real SQL nests <~30 deep).

## Testing

New unit tests in `test_sqlglot_utils.py` and `test_sqlglot_lineage.py` cover: the text-nesting scan (incl. string-literal handling), paren-bomb rejection at the parse chokepoints, deep-`CASE` rejection via the AST guard, graceful fingerprint degradation, and that the offending SQL is logged. Full `sql_parsing` unit suite passes (378 passed), `mypy` and `ruff` clean.

## Checklist

- [x] Tests added/updated
- [ ] Docs added/updated

### DIFF
--- a/metadata-ingestion/src/datahub/sql_parsing/sqlglot_utils.py
+++ b/metadata-ingestion/src/datahub/sql_parsing/sqlglot_utils.py
@@ -2,6 +2,7 @@ from datahub.sql_parsing._sqlglot_patch import SQLGLOT_PATCHED
 
 import functools
 import logging
+import os
 import re
 from typing import Dict, Iterable, List, Optional, Tuple, Union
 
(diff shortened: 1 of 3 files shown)


treff7es:
Closing in favor of #17788, which takes the same approach (StatementTooComplexError guard in parse_statement). Consolidating to avoid duplicate work. Note: investigation of a live SIGSEGV showed the crash is a native C-stack overflow *inside* sqlglot[c]'s optimizer during column-level lineage on a shallow view — so the input depth/length guard in either PR does not catch that class; the durable fix is subprocess isolation of CLL. The crash-diagnosis breadcrumb has been added to #17788.

PR #17828, opened 2026-06-09
feat(i18n): add support of i18n to remirror

v-tarasevich-blitz-brain:
Adds internationalization to the rich-text editor on two fronts:

1. DataHub's own editor UI strings (toolbar, menus, modals) are extracted into the alchemy i18next namespace.
2. Remirror's built-in labels (command / table / keyboard / color labels) are now translatable through Remirror's own Lingui i18n, with a German bundle supplied by DataHub. English continues to ship with @remirror/i18n, so it isn't duplicated.


datahub-web-react/src/i18n/locales/en/alchemy.json
```diff
@@ -9,6 +9,39 @@
...
+    "editor.table.deleteColumn": "Delete column",
+    "editor.table.deleteRow": "Delete row",
+    "editor.table.deleteTable": "Delete table",
+    "editor.table.insertColumnLeft": "Insert column left",
+    "editor.table.insertColumnRight": "Insert column right",
+    "editor.table.insertRowAbove": "Insert row above",
+    "editor.table.insertRowBelow": "Insert row below",
+    "editor.upload.chooseFile": "Choose File",
+    "editor.upload.failedTitle": "Upload Failed",
+    "editor.upload.maxSize": "Max size: 2GB",
```
benjiaming: We already have this one (with capital S) below. Let's just reuse it.

datahub-web-react/.storybook/i18n.ts
```diff
@@ -1,23 +1,30 @@
...
 import { initReactI18next } from 'react-i18next';
 
-// Eagerly bundle every English namespace so stories render real strings instead of raw keys.
-// (The app lazy-loads these via a backend; Storybook has no backend, so we load them synchronously.)
-const modules = import.meta.glob('../src/i18n/locales/en/*.json', { eager: true, import: 'default' });
+// Eagerly bundle every namespace for every language so stories render real strings
+// instead of raw keys, and the toolbar language switcher works offline.
+// (The app lazy-loads these via a backend; Storybook has no backend, so we load them
+// synchronously.) Namespaces without a translation for a given language fall back to en.
+const modules = import.meta.glob('../src/i18n/locales/*/*.json', { eager: true, import: 'default' });
```
benjiaming, line 8: re Eagerly bundling all namespaces: lazy loading would be a potential performance optimization. Can you create a follow-up ticket to be picked up when we add the next locale (at the latest).

benjiaming (changes requested):

datahub-web-react/.storybook/i18n.ts
v-tarasevich-blitz-brain: https://linear.app/acryl-data/issue/CAT-2398/optimize-storybook-i18n-loading

benjiaming (approved):

PR #17834, opened 2026-06-09
feat(glossary): add color picker, modernize parent selector, and inherit colors from parent groups

annadoesdesign:
## Summary

Adds modern color pickers to glossary terms, glossary nodes, and domains, and reworks how those colors flow through the UI.

**Color picker on create + edit**

- Domains, glossary nodes, and glossary terms now expose the alchemy `ColorPicker` in both the create modals and the entity-profile icon (`IconColorPicker`). The picker pre-fills with the color you actually see on the page (saved `colorHex`, falling back to the palette color we derive from the URN), so editing never starts from a gray default.
- On create, the picker initially tracks the selected parent's color (so a child term inherits its group's identity by default). Once the user explicitly picks, that choice locks in and ignores subsequent parent changes.
- The default empty-state color was switched to `#533fd1` to match the brand violet instead of the legacy gray placeholder. We only persist `colorHex` to the backend when the user has actually picked a color, so unchanged entities still hash deterministically off their URN.

**Glossary term color inheritance**

- New `getGlossaryTermColor(term, generateColor)` utility resolves a term's color via a single precedence chain: term's own `colorHex` → root parent's `colorHex` → palette of root URN → palette of term URN.
- Every term-rendering surface that previously hashed the term URN directly now goes through this utility (sidebar `TermItem`, `useTermTreeOptions`, `DefaultEntityHeader`, `GlossaryListCard`), so a term's profile page header, the sidebar, and the term-picker chips all agree on the color.
- Backend: registered `displayProperties` as a valid aspect on `glossaryTerm`, added the field to the `GlossaryTerm` GraphQL type, wired `GlossaryTermMapper`/`GlossaryTermType` to resolve and return it, and broadened `UpdateDisplayPropertiesResolver`'s authorization so glossary editors don't also need `MANAGE_DOMAINS`. Frontend fragments (`glossaryTerm`, `childGlossaryTerm`, `rootGlossaryNodeWithFourLayers`, `autoCompleteFields`, `facetFields`, `searchResultsWithoutSchemaField`) all now request `displayProperties`.

**Modernized glossary parent selector**

- Replaced the antd `Select` + `GlossaryBrowser` popup with a single alchemy `SimpleSelect`-based `NodeParentSelect` that browses the glossary tree (`useGlossaryTreeEntities`) and switches to `useGetAutoCompleteResultsLazyQuery` while the user is searching. V1 file now re-exports V2 so both modal flavors share the same picker.
- Each option renders the node's `GlossaryColoredIcon` + caret + label, indented by depth, with selectable nodes (the picker overrides `useTermTreeOptions`'s default `disabledValues`). Move-mode still excludes the entity itself and its descendants.
- Optional parent on create: the parent field is always shown and non-required, so root-level terms/nodes can be created directly from the entity-page dropdown.

**Collapsible glossary sidebar**

- `GlossarySidebar` is now collapsible with the same affordance as the domains sidebar. When collapsed, the create-glossary button is hidden, the toggle moves to center, and a vertical strip of `GlossaryColoredIcon`s renders the root nodes/terms (each linking to the entity), so users can still navigate without expanding.

**Documentation editor**

- The create modal's documentation field used to open a nested `DescriptionModal` over the create modal. Inlined the alchemy `Editor` directly into the create modal so it's one continuous form instead of modal-in-modal.

**Bug fixes along the way**

- `DefaultEntityHeader` was resolving glossary-term colors with `displayProperties?.colorHex || generateGlossaryColor(urn)`, skipping the parent inheritance step. A term with no own color was rendered with a palette hash of the term URN in the page header even though the sidebar showed the inherited parent color. Routed through `getGlossaryTermColor` instead so all surfaces agree.
- `GlossaryBrowser/NodeItem` was rendering nested nodes with `BookmarkSimple` (the term icon) instead of `BookmarksSimple` (the group icon) — `iconColor ? BookmarkSimple : BookmarksSimple` confused inheritance with leaf-ness. Always uses `BookmarksSimple` for nodes now.
- `EntityPage` (V1 and V2) had a Rules-of-Hooks violation: `entityRegistry.renderProfile(...)` was called conditionally based on `canViewEntityPage`, but some entity profiles inline hooks via `getProfileTabs()` (e.g. `useShowAssetSummaryPage`), which changed the hook count when privileges resolved. Hoisted the call so it runs on every render.
- Fixed a pre-existing slf4j-placeholder-in-`String.format` bug in `UpdateDisplayPropertiesResolver` (`{}` → `%s`) and started passing the original exception as a cause for better debugging.

**i18n**

- Added missing translation keys: `color`, `icon`, `optional` to `common.labels.json`; `sidebar.expand` / `sidebar.collapse` to `governance.glossary.json`; `createGlossary.searchParentPlaceholder` to `entity.shared.entityDropdown.json`; `createModal.namePlaceholder` to `entityV1.shared.entityDropdown.json`. All previously hardcoded English strings introduced by these features are now keyed.

## Test plan

- [x] \`yarn lint\` (eslint + prettier + tsc) clean
- [x] \`yarn test src/app/glossaryV2 src/app/shared/tags src/app/entityV2/shared/EntityDropdown\` — 66/66 passing, including new tests for the term-color precedence chain (\`colorUtils.test.ts\`) and tree-options term color (\`useTermTreeOptions.test.tsx\`)
- [ ] Create a new domain via the menu, pick a color, verify the saved color shows in the listing and the entity-header icon
- [ ] Create a new glossary term group with no parent, verify it appears at the root of the sidebar and its color sticks after reload
- [ ] Create a glossary term under an already-colored group without picking a color; verify it inherits the group color in the sidebar, on its own profile page header, and in any term-pill that references it
- [ ] Create a glossary term under an already-colored group and pick a different color; verify the term shows the chosen color everywhere, overriding the parent's
- [ ] Edit a term's color via the header icon-color picker; verify it updates in place without a hard reload
- [ ] Collapse the glossary sidebar; verify root nodes/terms appear as colored icons, that each navigates correctly, and that the create button hides while the toggle stays accessible
- [ ] Move a glossary node via the move modal; verify the picker excludes the node itself and its descendants


annadoesdesign:
Reopening from internal branch — see new PR.

PR #17839, opened 2026-06-10
feat(openapi/lineage): Add lineage scroll endpoint

asikowitz:
Adds a new `/openapi/v3/relationship/scrollLineage` endpoint that automatically adds the lineage filter (based on lineage registry -- creates triplets). To facilitate this, updates `GraphFilters` to have a `allowedEdgeTriplets` that gets read by `GraphQueryUtils`.

Also refactors `RelationshipControllerTest` because calls changed to `scrollRelatedEntities`.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/graph/openapi.py
+++ b/metadata-ingestion/src/datahub/ingestion/graph/openapi.py
@@ -364,8 +364,91 @@ class OpenApiAPI(OpenAPIGraphProtocol):
             - scroll_id: cursor to pass in the next call (None when exhausted)
             - relationships: list of Relationship objects with source/destination info
         """
-        url = f"{self._gms_server}/openapi/v3/relationship/scroll"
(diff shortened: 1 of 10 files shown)


david-leifker (changes requested):
I'd suggest a slightly different approach to the API which I think gives a more clear differentiation between relationships and lineage. Approach: re-use the backend scroll logic once but expose two endpoints with different defaults and validation. That gives lineage callers a purpose-built API without duplicating backend code or forcing everyone through relationship-type bookkeeping.

POST `/openapi/v3/lineage/scroll`
Purpose-built bulk scroll for lineage edges only. Auto-restricts relationship types via LineageRegistry; callers do not need to pass relationshipTypes and those that are filtering a subset of lineage relationships also get proper validation against the lineage registry.

david-leifker (approved):
Stamped assuming endpoint rename.

PR #17842, opened 2026-06-10
feat(ingest/sql-parsing): temp-aware query fingerprinting across all SQL sources

treff7es:
## Summary

On large warehouse deployments the **Query entity index dominates the metadata store** (frequently the largest ES index by doc count), driving reindex failures during upgrades and elevated cost. A major contributor is **per-run ephemeral object names**: ELT/transform tooling (Fivetran/Stitch staging tables, Segment, Snowpark, Great Expectations, …) embeds a fresh UUID/hash in table names on every run, so two otherwise-identical statements fingerprint differently and mint a brand-new Query entity each run.

The connector already detects these as temp tables **for lineage**, but that detection was never consulted when computing the fingerprint that becomes the Query URN — fingerprinting was "temp-blind".

This PR makes fingerprinting **temp-aware across all SQL-parsing sources**. The behavior is **always on** (there is no configuration flag), and per-run ephemeral query variants — including queries whose *output* is a temp table — **roll up to a single Query entity**.

### Marker-preserving fingerprint normalization

When fingerprinting, a per-run ephemeral token is normalized to a fixed value **while the marker that identifies the kind of temp table is kept**, so per-run variants collapse but the entity stays legible and distinct temp kinds are not merged:

- `ge_temp_aa91f1fd` → `ge_temp_0`
- `…fivetran_x_staging.entity-staging-<uuid>` → `…fivetran_x_staging.entity-staging-?`
- `ge_temp_…` ≠ `snowpark_temp_table_…` (different markers stay distinct)

This applies on two paths:

- **Fast precompute path** (Snowflake/BigQuery fingerprint without parsing): `DEFAULT_TEMP_TABLE_FINGERPRINT_RULES` are applied **unconditionally inside the fingerprint functions** — they are no longer threaded through each call site, so every fast-path caller (Snowflake queries, BigQuery queries, Unity usage, Snowflake lineage, stored-proc lineage) gets consistent normalization. Quote characters bound the match naturally, so no per-dialect quoting is enumerated.
- **Slow/AST path:** `generalize_query` replaces any table whose resolved `catalog.db.name` matches a connector's `temporary_tables_pattern` with a placeholder before fingerprinting — one source of truth, dialect-agnostic. `SqlParsingAggregator.temp_table_name_patterns` threads those patterns through the lru-cached `sqlglot_lineage` (hashable tuple), so every aggregator-based source (Redshift, MSSQL, Oracle, `sql-queries`, …) is temp-aware once it supplies temp patterns.

### Roll-up of temp-output queries (not dropped)

A query whose **output** is a temp table (`CREATE [OR REPLACE] TEMP … AS …`) is **emitted as a single rolled-up Query entity**, not dropped:

- Per-run variants collapse to one via the fingerprint normalization above.
- It carries only its **real upstream subjects**; the ephemeral temp table is **not** emitted as a (phantom) downstream dataset subject.
- Its lineage still resolves through the temp table into the consuming query, so real source→destination lineage is unaffected.

This preserves the query record and its upstream-usage signal (vs. dropping it outright), while still removing the per-run explosion.

## Why it's lineage-safe

Temp/staging tables are already excluded from lineage and never persisted as datasets. Collapsing only the per-run identity merges per-run duplicates of the *same logical job* into one entity while preserving real source→destination lineage. Distinct real targets and stable dbt models (`<model>__dbt_tmp`) keep separate fingerprints.

## Validation (real production data)

Validated on a real `metadata_aspect_v2` export (200,000-query reservoir sample; margin of error ~±0.2%). Each query's SQL was re-fingerprinted with the temp-aware rules **off** (baseline = current behavior) vs **on** (this PR):

| Variant | Distinct Query entities | Reduction |
|---|---|---|
| Baseline (current) | 199,330 | — |
| **Roll-up (this PR)** | 127,523 | **−36.0%** |
| Drop alternative | 121,039 | −39.3% |

- **~52% of queries** are touched by an ephemeral-marker rule; these collapse to a **36% reduction in distinct Query entities**.
- Roll-up keeps **6,484** distinct temp-output entities (all temp-only, no real-query collision). Dropping them instead would add only **+3.3 pts** (36.0% → 39.3%) — roll-up captures ~92% of the maximum reduction while preserving the query records.
- Existing lineage golden tests unchanged (dataset→dataset lineage preserved across all integration goldens).

**This is *prevention***, not retroactive cleanup: it reduces entities created on future ingestion runs. Because `query_id` changes, the existing backlog requires the companion age-based GC to reclaim the stale tail (count briefly rises before it falls).

## Breaking change

Temp-aware fingerprinting is always on, so `query_id` — and therefore Query URNs — change for affected queries. On the next ingestion run you get a new set of collapsed Query entities while the old per-run ones become stale and are reclaimed by `datahub-gc`. Documented in `docs/how/updating-datahub.md`. No action required to benefit.

## Checklist

- [x] PR conforms to the title format
- [x] Tests added/updated (marker-preserving fingerprint rules, slow-path AST collapse + lineage-safety, dialect coverage, aggregator temp-output roll-up)
- [x] Breaking-change entry in `docs/how/updating-datahub.md`


treff7es:
### Measured impact on a real production Snowflake metadata sample

Validated the entity-count reduction on a real `metadata_aspect_v2` export (200,000-query reservoir sample; margin of error ~±0.2%). Each query's SQL was re-fingerprinted with the temp-aware rules **off** (baseline = current behavior) vs **on** (this PR).

| Variant | Distinct Query entities | Reduction |
|---|---|---|
| Baseline (current) | 199,330 | — |
| **Roll-up (this PR)** | 127,523 | **−36.0%** |
| Drop alternative | 121,039 | −39.3% |

- **~52% of queries** are touched by an ephemeral-marker rule; these collapse to a **36% reduction in distinct Query entities**.
- Temp-output `CREATE TEMP … AS …` queries are **rolled up to one** (not dropped): 6,484 distinct temp-output entities in the sample, all with no real-query collision. Dropping them instead would add only **+3.3 pts** (36.0% → 39.3%) — roll-up captures ~92% of the maximum reduction while preserving the query records and their upstream-usage signal.

**Note:** this is *prevention* — it reduces entities created on future ingestion runs. Because `query_id` changes, the existing backlog still requires the companion age-based GC to reclaim the stale tail (count briefly rises before it falls).

treff7es:
### Single-day vs. cumulative impact (why the number depends on the window)

Following up on the impact numbers above with a per-day view, since the reduction depends heavily on the measurement window.

Re-fingerprinting **one calendar day** of query activity at a time (all queries last-active that day):

| Window | Queries | Roll-up reduction | Marker-touched |
|---|---|---|---|
| Single day A | 6,923 | −5.6% | 16.4% |
| Single day B | 5,774 | −10.5% | 22.1% |
| Single day C | 4,807 | −8.4% | 26.4% |
| **Full store** | ~1.13M | **−36.0%** | 51.6% |

**The bloat is a cross-day accumulation problem, not an intra-day one.** A recurring ETL query with an ephemeral name runs once (or a few times) *within* a day, so there's little to collapse that day — hence ~5–10%. But it runs again the next day with a *new* ephemeral name, and under today's temp-blind fingerprinting that mints a brand-new Query entity **every day**. Over months those daily variants accumulate — which is what produces the 51.6%/−36% at full-store scale.

Steady-state mechanism:

- **Temp-blind (today):** a recurring ephemeral query adds **~1 new entity per day, indefinitely**.
- **Temp-aware (this PR):** it adds 1 entity the first day, then **~0 thereafter** (later runs match the existing fingerprint).

So the durable effect is that the **daily growth rate of the Query index flattens** for the ~50% of queries that are ephemeral-recurring. The single-day figure understates this; the cumulative **−36%** reflects how much smaller the index becomes over time.

PR #17845, opened 2026-06-10
feat(ingest/sql-parsing): centralize temp-aware fingerprinting across all SQL sources

treff7es:
> **Stacked on #17842** (`fix/temp-aware-query-fingerprint`). Base is that branch, so the diff shows only the generalization. Merge after #17842 (or it will retarget to `master` once #17842 lands).

## Summary

#17842 added temp-aware query fingerprinting for **Snowflake only**, on the fast (regex) path. This PR generalizes it to every source that parses SQL through the `SqlParsingAggregator` — and, per review feedback, does so by **reusing each connector's existing `temporary_tables_pattern`** instead of a parallel set of per-tool / per-dialect text regexes.

## Design (after review)

- **Slow / AST path (all aggregator-based sources):** `generalize_query` walks the parsed expression and replaces any table whose resolved `catalog.db.name` matches the connector's existing `temporary_tables_pattern` with a placeholder before fingerprinting. Because the AST yields **bare identifier names**, there are **no dialect-quoting regexes** here — it's dialect-agnostic and consistent with how those tables are already excluded from lineage. **One source of truth.**
- **Fast precompute path (Snowflake / BigQuery, no parse):** a *minimal* text-regex set (`DEFAULT_TEMP_TABLE_FINGERPRINT_RULES`) is retained only here, since these connectors fingerprint without parsing. Clearly scoped as the perf fallback.
- **Threading:** `SqlParsingAggregator.temp_table_name_patterns` (default `None`) carries the connector's resolved-name patterns through the lru-cached `sqlglot_lineage` as a hashable tuple. Snowflake passes its `_compiled_temporary_tables_pattern`; BigQuery's aggregator needs none (it precomputes).

## Why this is less fragile than the first pass

The earlier version matched ephemeral names in **raw SQL text** with per-dialect quoting regexes (double-quote/backtick/bracket) — a second temp-detection system that could drift from `temporary_tables_pattern`. Matching on the **resolved AST name** removes the quoting regexes and the duplication.

## Default-on & breaking change

Snowflake and BigQuery are **default-on** (`temp_aware_query_fingerprinting`, default `true`). Other aggregator-based sources benefit once they supply temp-table name patterns. Lineage-safe, but **changes `query_id`/Query URNs** for affected queries; documented in `docs/how/updating-datahub.md`.

## Coverage note (honest)

Snowflake/BigQuery (the validated high-volume Query emitters) stay covered via the fast precompute path. The slow-path AST normalization benefits other sources **that configure `temporary_tables_pattern`**; it is not a universal hardcoded ETL list.

## Tests & validation

- Unit tests: AST name-based collapse + lineage-safety on the slow path; dialect-quoting coverage for the retained fast-path set; an aggregator-level test where a connector-supplied `temporary_tables_pattern` collapses per-run staging to one entity while a distinct real target stays split. Full `sql_parsing` suite: 450 passed, 3 skipped; ruff + mypy clean; golden files unchanged.
- Real-data validation of the double-quote/dialect path was done in #17842. Other dialects are covered by unit tests (no real BigQuery/Redshift dumps were available).

## Checklist

- [x] PR conforms to the title format
- [x] Tests added/updated
- [x] Breaking-change entry updated in `docs/how/updating-datahub.md`


treff7es:
Folded into #17842 — the centralization is now part of the single temp-aware-fingerprinting PR. Closing this stacked PR; deleting its branch.

PR #17851, opened 2026-06-10
fix: handle oversized aspects — raise Jackson maxNameLength (GMS) and budget full schemaMetadata when trimming (ingestion)

puneetagarwal-datahub:
## Summary

Two related oversized-content problems prevent metadata from being ingested, on two different code paths. Both stem from oversized Iceberg/dbt-style schemas (deeply-nested struct field paths).

**1. GMS — oversized JSON property names in patches.** Deeply-nested struct field paths carried as keys in `upstreamLineage` patches can exceed Jackson's default `maxNameLength` of 50,000 characters. Applying such a patch fails deserialization:

```
Name length (62775) exceeds the maximum allowed (50000, from `StreamReadConstraints.getMaxNameLength()`)
    at PatchItemImpl.applyTemplatePatch(...)
```

Every `StreamReadConstraints` config raised `maxStringLength` but left `maxNameLength` at Jackson's default, so a single over-long property name fails ingestion even when the overall payload is small. The exception never includes the offending name, making it hard to diagnose.

**2. Ingestion — `schemaMetadata` trimming under-counts the aspect.** `EnsureAspectSizeProcessor.ensure_schema_metadata_size` trims schema fields to fit the payload limit, but it only summed `schema.fields` and ignored the rest of the aspect — notably `platformSchema.rawSchema`, plus top-level keys and JSON-escaping overhead. So it trimmed fields to ~15.49 MB, then the non-field content pushed the serialized aspect back over the limit and GMS still rejected it:

```
String value length (16033303) exceeds the maximum allowed (16000000, from `StreamReadConstraints.getMaxStringLength()`)
```

Observed on CLI 1.5.0.13: the existing truncation removed 497 fields but the aspect was still 16,033,303 chars — **33 KB over** the 16 MB limit — purely because ~540 KB of non-field content was never charged against the budget.

### GMS (Jackson name-length)

- Add `MAX_JACKSON_NAME_LENGTH` and `INGESTION_MAX_SERIALIZED_NAME_LENGTH` (default 16 MB, env-tunable), mirroring the existing string-length limit.
- Set `maxNameLength` on every aspect-deserialization mapper: `TemplateUtil`, `PatchItemImpl`, `ObjectMapperFactory` (primary + API-sanitizing), `RAPServletFactory`, `ServletConfig`, `GraphQLController`, `UsageServiceUtil`, `OperationsServiceUtil`.
- Log `urn` / aspect / location + length when a patch carries a property **name** over the name limit, or a string **value** over the value limit, so oversized names/values are identifiable from logs (the exception itself omits them). Both are purely diagnostic and tunable via `INGESTION_OVERSIZED_NAME_WARN_THRESHOLD` (default 50,000) and `INGESTION_OVERSIZED_VALUE_WARN_THRESHOLD` (default 10,000,000). The scan is a single allocation-light pass and is fully wrapped in a `try/catch` that downgrades to debug — it can never affect ingestion. Note: unlike `maxNameLength`, the string-value limit is **not** raised — the value warning only flags fields approaching the existing 16 MB limit.
- Add a `TemplateUtil` regression test and an `updating-datahub.md` note.

### Ingestion (schemaMetadata trimming)

- `EnsureAspectSizeProcessor.ensure_schema_metadata_size` now budgets the **full serialized aspect**, not just the fields list. A new `_schema_size_without_fields` helper measures the non-field content (top-level keys + `platformSchema.rawSchema`), and the field-trimming loop starts from that baseline — so accepted fields **plus** non-field content stay within the constraint.
- When `rawSchema` alone exceeds the budget, it is dropped (with a warning) so the structured fields can still be retained rather than failing the whole aspect.
- Regression tests covering both the budget accounting and the oversized-`rawSchema` case.

## Example: GMS diagnostic logging

The warning fires only when a patch carries a property name (a JSON-Pointer path segment) over the limit — normal patches stay silent.

Normal patch — no warning:

```json
[{ "op": "add",
   "path": "/glossaryTerms/urn:li:glossaryTerm:PII",
   "value": { "urn": "urn:li:glossaryTerm:PII" } }]
```

Oversized column-level lineage patch — the `schemaField` key for a deeply-nested struct field is ~62,775 chars:

```json
[{ "op": "add",
   "path": "/fineGrainedLineages/DOWNSTREAM/urn:li:schemaField:(urn:li:dataset:(urn:li:dataPlatform:dbt,my_model,PROD),[version=2.0].[type=struct].[type=struct]...<~62,775 chars>...field)",
   "value": { } }]
```

logs (and the write still succeeds, because `maxNameLength` is now raised):

```
WARN Oversized JSON key in patch: urn=urn:li:dataset:(urn:li:dataPlatform:dbt,my_model,PROD),
     aspect=upstreamLineage, keyLength=62775 (warn threshold 50000).
     Key prefix: '...'. The write will succeed (maxNameLength raised), but a key this large is
     likely unusable downstream (UI / Elasticsearch); review the source.
```

## Example: schemaMetadata trimming budget

With `schema_size_constraint ≈ 15,492,710` and ~540 KB of non-field content (`rawSchema` + keys):

```
Before: budget for fields = 15,492,710 (non-field ignored)
        final aspect = 15,492,710 (fields) + 540,593 (non-field) = 16,033,303  → REJECTED (33 KB over 16 M)

After:  total_fields_size starts at 540,593 (non-field charged first)
        fields trimmed until 540,593 + fields < 15,492,710
        final aspect = 15,492,710  → ACCEPTED
```

## Notes

- The GMS change stops the failure by accepting these names. A field name in the tens of thousands of characters is still awkward to render (UI) and index (Elasticsearch); the cleaner long-term fix is on the source side (saner field paths for ultra-nested structs) and is intentionally out of scope here.
- The ingestion change makes trimming actually fit the destination limit; it does not raise any limit. For destinations with a hard cap (e.g. an external catalog), trimming is the only lever.
- 16 MB is generous headroom — observed offending names were ~51K–65K characters.

## Checklist


puneetagarwal-datahub:
Superseded by #17892. This PR was auto-closed when its branch `fix/jackson-max-name-length` was renamed to `fix/oversized-aspect-handling` (the scope now also covers the ingestion-side schemaMetadata trimming fix). Continuing in #17892.

PR #17854, opened 2026-06-10
perf(graphql): compute only selected EntityPrivileges fields

gdatahub:
## What & why

`EntityPrivilegesResolver` serves the `privileges` field on the ~21 entity types that expose it. It
eagerly computed the **entire** `EntityPrivileges` object — ~22 independent authorization checks for
a Dataset — regardless of which sub-fields the query selected. The lineage `canEditLineageFragment`
selects **only** `canEditLineage`, so every node in a lineage graph paid for all ~22 checks to
produce one boolean: an N×~22 authorization fan-out.

This makes the resolver **selection-set aware** — it computes only the requested privilege booleans.
The privileges are mutually independent, so this is behaviour-preserving (unselected fields stay
null and are never serialized).

## How the selected fields are detected (and why not `getSelectionSet()`)

The selected sub-fields are read from **this field's own AST selection set** — `getMergedField()` +
its selection set, resolving fragment spreads (`...entityPrivileges`) and inline fragments. This is
local to the `privileges` field and cheap.

It deliberately avoids `DataFetchingEnvironment.getSelectionSet()`: that API normalizes the selection
set under graphql-java's 100k max-field-count guard, which a large recursive query
(`getBulkEntityLineageV2`) exceeds — throwing `AbortExecutionException` and aborting the **entire**
lineage query (the lineage graph then fails to render). Reading the immediate selection from the AST
is not subject to that global limit. (An earlier revision used `getSelectionSet()` and hit exactly
this; commit `fix(graphql): read selected EntityPrivileges sub-fields from the AST` corrects it.)

## Review feedback addressed

Per review: the repeated hardcoded field-name strings are now `EntityPrivilegesFields` constants
(single source of truth), guarded by `EntityPrivilegesFieldsTest` — a **two-way** reflective check
against the generated `EntityPrivileges` type (every constant maps to a generated field, and every
generated field has a constant), so a typo or future drift fails the build.

## Validation — end-to-end A/B (master vs this PR)

Benchmarked on a live GMS across **all 21 privilege-bearing entity types × 6 users** with
resource-scoped policies (domain / tag / owner / urn):

| | result |
|---|---|
| **Equivalence** | privilege matrix **byte-for-byte identical** to master across every user/entity/field — no behaviour change |
| **Lineage query p50** | **−49%** (consistent across all users) |
| **Full 24-field fragment p50** | neutral (all fields still computed when all are selected) |
| **Lineage render** | the previously-failing Cypress lineage spec (`v3_lineage_column_path`) now passes locally against the fixed build |

Unit tests: `EntityPrivilegesResolverTest` updated + new selection-awareness, field-name guard, and
two-way drift tests; full `:datahub-graphql-core:test` green (1603).

## Alternatives considered

A urn-keyed `privileges` DataLoader (batching the per-node checks) was prototyped and A/B-tested — it
measured a ~5× regression because it serialized into one task the per-entity work graphql-java runs
in parallel, without reducing total work. Reverted in favour of this parallelism-preserving,
selection-aware approach.

## Checklist

- [x] Docs — N/A (internal resolver optimization, no API/behaviour change)
- [x] Breaking changes — none (behaviour-preserving)

### DIFF
--- a/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/entity/EntityPrivilegesResolver.java
+++ b/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/entity/EntityPrivilegesResolver.java
@@ -1,5 +1,6 @@
 package com.linkedin.datahub.graphql.resolvers.entity;
 
+import static com.linkedin.datahub.graphql.resolvers.entity.EntityPrivilegesFields.*;
 import static com.linkedin.metadata.Constants.WILDCARD_URN;
 import static com.linkedin.metadata.authorization.ApiGroup.LINEAGE;
 import static com.linkedin.metadata.authorization.ApiOperation.UPDATE;
@@ -115,7 +116,7 @@ public class EntityPrivilegesResolver implements DataFetcher<CompletableFuture<E
     final EntityPrivileges result = new EntityPrivileges();
     computeIfSelected(
         selected,
-        "canManageEntity",
+        CAN_MANAGE_ENTITY,
         () -> BusinessAttributeAuthorizationUtils.canManageBusinessAttribute(context),
         result::setCanManageEntity);
     addCommonPrivileges(result, urn, context, selected);
@@ -126,7 +127,7 @@ public class EntityPrivilegesResolver implements DataFetcher<CompletableFuture<E
       Urn termUrn, QueryContext context, Set<String> selected) {
     final EntityPrivileges result = new EntityPrivileges();
     addCommonPrivileges(result, termUrn, context, selected);
-    if (selected.contains("canManageEntity")) {
+    if (selected.contains(CAN_MANAGE_ENTITY)) {
       result.setCanManageEntity(computeTermCanManageEntity(termUrn, context));
     }
     return result;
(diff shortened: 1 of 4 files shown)


datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/entity/EntityPrivilegesResolver.java
```diff
@@ -50,81 +55,123 @@ public CompletableFuture<EntityPrivileges> get(DataFetchingEnvironment environme
...
+
+  private EntityPrivileges getBusinessAttributePrivileges(
+      Urn urn, QueryContext context, Set<String> selected) {
     final EntityPrivileges result = new EntityPrivileges();
-    result.setCanManageEntity(
-        BusinessAttributeAuthorizationUtils.canManageBusinessAttribute(context));
-    addCommonPrivileges(result, urn, context);
+    computeIfSelected(
+        selected,
+        "canManageEntity",
```
chakru-r: There are too many repeating hardcoded strings here -- can we have consts for them and avoid repeating? Unfortunately though these are defined in the graphql definition, these are used to generate fields in EntityPrivileges class and not available as strings. 
Hardcoding this can have mistakes -- and even if correct now, can drift if the grapqhl definition is updated.  
One way is have a EntityPrivilegesFields class that just defines these consts and have a unit test that ensures all consts in it map to the generated fields -- two way check, all defined consts exist in the generated class and all generated fields have a const defined -- so that mismatches as well as drift can be caught.

chakru-r (changes requested):

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/entity/EntityPrivilegesResolver.java
gdatahub: Created a constants class containing all the field names and added test cases to validate the fields value  in both directions (generated fields vs constant).

Dutt23 (approved):

PR #17860, opened 2026-06-11
feat(sdk): env-var fallback for client cert paths in RequestsSessionConfig

rajatoss:
## Summary

`RequestsSessionConfig.build_session()` now falls back to `DATAHUB_CLIENT_CERT_PATH` / `DATAHUB_CLIENT_KEY_PATH` from the environment when the caller doesn't supply explicit cert paths. This mirrors the `__from_env__` pattern already used for the server URL and auth token, and gives mTLS-configured processes a single, SDK-owned resolution point.

## Why

Downstream callers (the DataHub remote executor in particular) need every outbound HTTPS path — direct `RequestsSessionConfig` users, `DataHubRestEmitter`, `DataHubGraph`, and ingestion sinks — to present the same client certificate when an mTLS perimeter (e.g. an AWS ALB with `mutual-authentication: verify`) sits in front of GMS. Previously each layer had to thread `client_certificate_path` / `client_key_path` kwargs through (or reinvent it with a bare `requests.Session()`). With this change, setting the env vars once in the process environment is enough.

Placing the fallback in `build_session()` rather than `DataHubRestEmitter.__init__` is intentional: callers who build a session via `RequestsSessionConfig` directly (without going through the rest emitter) get the same resolution — important for non-emitter HTTP clients in downstream code (e.g. the executor's integrations-service sessions).

## Behavior

| Explicit cert paths | Env vars set        | Result                         |
| ------------------- | ------------------- | ------------------------------ |
| both                | (any)               | tuple `(cert, key)` (explicit) |
| cert only           | (any)               | string `cert` (explicit)       |
| neither             | both                | tuple `(cert, key)` (env)      |
| neither             | cert only           | string `cert` (env, combined PEM) |
| neither             | neither             | `session.cert` unset           |

Explicit paths still take precedence — backwards compatible.

## Test plan

- [x] Unit tests in `tests/unit/sdk/test_rest_emitter.py::TestClientCertEnvFallback`:
  - via `RequestsSessionConfig().build_session()` directly: both env vars set, neither set, explicit overrides env, only cert env var set
  - via `DataHubRestEmitter(...)` transitively: env vars reach `session.cert`
- [x] Existing `TestRequestsSessionConfig` (cert-handling tests) still pass — no regression
- [x] End-to-end verified on a dev instance via the executor's mTLS rollout (acryldata/datahub-fork#10153) — Snowflake ingestion successfully pushed metadata through an mTLS-enforcing ALB after the executor switched to env-fallback resolution


skrydal (approved):
🚀

PR #17861, opened 2026-06-11
feat(ingest/data-lake): profile S3/GCS/Azure/Delta with DuckDB instead of Spark/Deequ

treff7es:
## Summary

Replaces the Spark + PyDeequ profiling engine in the data-lake sources (S3, GCS, Azure Blob, Delta Lake) with **DuckDB**, driven through the existing `SQLAlchemyProfiler`. This **removes the JVM/Java requirement** and drops the heavy `pyspark` + `pydeequ` dependencies from DataHub entirely — the data-lake profiler was their last consumer.

Profiling now runs via a new `DuckDBAdapter` (in the shared `sqlalchemy_profiler`) plus a thin `DuckDBProfiler` "attach layer" that registers each table's files as a DuckDB view (`read_parquet`/`read_csv_auto`/`read_json`/`read_avro`) and delegates statistics collection to the same profiler all SQL sources already use.

### Key changes
- **New `DuckDBAdapter`** — native `approx_count_distinct`/`quantile_cont`, plus a `SUMMARIZE` fast-path that computes the base numeric + cardinality + null block for all columns in a single scan.
- **New `DuckDBProfiler`** — temp-file-backed DuckDB connection, cloud credentials injected via `CREATE SECRET` (never env vars), per-table fail-soft, and stream-by-default / sample-over-threshold using the same sampling knobs as the warehouse sources.
- **GCS** rides on this for free (it delegates to an internal S3 source); **Avro** is supported via DuckDB's `avro` extension; messy CSVs are read with `strict_mode=false`.
- Data-lake profiling config now subclasses `GEProfilingConfig` while **preserving historical defaults**; `spark_driver_memory` / `spark_config` are **removed** (supplying either emits a removed-field warning and is otherwise ignored).
- Dead Spark/Deequ profiler modules removed; `pyspark`/`pydeequ` dropped everywhere — including from the `databricks` extra, since Unity Catalog never imported PySpark (it uses `databricks-sql-connector` + SQL parsing), so `acryl-datahub[databricks]` no longer pulls PySpark either.

### Follow-up review fixes
- **Error classification**: object-store access errors (bad/expired credentials, unreachable endpoint) now escalate to a single deduped `report_failure` instead of N per-table warnings; unsupported file formats get a dedicated, actionable warning category.
- **Azure Blob (abs)**: dropped the misleading `DATA_PROFILING` capability — abs has no profiler wired and no Azure credential path — and `profiling.enabled: true` now fails fast with a clear error rather than silently emitting nothing.
- **Typing/config**: `DuckDBProfiler` is typed on `DataLakeProfilerConfig` (no `getattr`); a blank `duckdb_extension_directory` coerces to `None`.
- Refactored `get_table_profile` (extracted materialize + report-forwarding helpers), deduped path/extension handling, and narrowed broad `except` blocks so real bugs surface.

### Behavior parity
Row/column/null counts are identical to the previous engine; distinct counts differ within approximation tolerance (HyperLogLog); DuckDB additionally emits numeric min/max/mean/median/stdev. Integration golden files were re-baselined and the diffs verified to be sensible.

## Test Plan
- [x] Unit tests for the adapter (native exprs, SUMMARIZE cache + fallback, type conversion), the attach layer (formats, sampling, avro, quote-escaping, cleanup), and credential SQL generation
- [x] Integration profiling tests (`test_data_lake_local_ingest`, `test_s3_profiling_coverage`) pass against re-baselined golden files
- [x] `ruff` + `mypy` clean on all changed source and test files

## Checklist
- [x] Tests added/updated
- [x] Docs updated (connector docs drop the Java/Spark prerequisites; DuckDB profiling documented)
- [x] Breaking change documented in `docs/how/updating-datahub.md` (Java/Spark no longer required; profile values may shift slightly; `spark_driver_memory`/`spark_config` removed; PySpark dropped from the `databricks` extra)


docker/datahub-ingestion/Dockerfile
```diff
@@ -191,6 +191,19 @@ RUN --mount=type=cache,target=$HOME/.cache/uv,uid=1000,gid=1000,sharing=private
...
+# object-store reads) and avro (Avro files), into a fixed, version-keyed
+# extension directory. Build-time has network access and buildx runs this on the
+# target platform, so the correct per-platform binary is baked in. DuckDB stores
+# them under <dir>/v<duckdb-version>/<platform>/, so a future duckdb bump rebuilds
+# this layer (it depends on the install above) and re-populates the matching
+# version dir. At runtime the profiler reads DATAHUB_DUCKDB_EXTENSION_DIRECTORY,
+# loads from this directory, and never downloads — so profiling works in
+# air-gapped deployments of this image with no extra configuration.
+ENV DATAHUB_DUCKDB_EXTENSION_DIRECTORY=$HOME/.duckdb-extensions
+RUN python -c "import os, duckdb; d = os.environ['DATAHUB_DUCKDB_EXTENSION_DIRECTORY']; os.makedirs(d, exist_ok=True); conn = duckdb.connect(); conn.execute(\"SET extension_directory='\" + d + \"'\"); [conn.execute(f'INSTALL {ext}') for ext in ('httpfs', 'avro')]; conn.close()" \
```
skrydal: Let's make the list `httpfs` and `avro` explicit as ENV variable.
Is the list of extensions complete? e.g. don't we need `AWS` and `Azure` extensions as well?
Also if `duckdb` package is dependency of `data_lake_profiling` extra, how do we ensure that it is available in the moment we RUN this particular command?

docker/datahub-ingestion/Dockerfile
```diff
@@ -191,6 +191,19 @@ RUN --mount=type=cache,target=$HOME/.cache/uv,uid=1000,gid=1000,sharing=private
     && datahub --version
 RUN rm -rf $HOME/.config/uv
 
+# Pre-install the DuckDB extensions used by data lake profiling: httpfs (remote
+# object-store reads) and avro (Avro files), into a fixed, version-keyed
+# extension directory. Build-time has network access and buildx runs this on the
+# target platform, so the correct per-platform binary is baked in. DuckDB stores
+# them under <dir>/v<duckdb-version>/<platform>/, so a future duckdb bump rebuilds
```
skrydal: Does it mean that if different `duckdb` version is used by the ingestion then we have to download new extensions versions?
Maybe this procedure should be part of the bundling script then.

skrydal:
Correct me if I am wrong (I do not have azure credentials at hand to test). It seems to me `abs` source did not actually support profiling and still does not do it. Unlike GCP, it does not inherit from S3 source, so it doesn't get functionality from there.
Also - shouldn't the profiler code be stored is a more general location that S3 source dirs? Even the naming suggests it does not belong there.

metadata-ingestion/src/datahub/ingestion/source/s3/duckdb_secrets.py
```diff
@@ -0,0 +1,58 @@
...
+        parts.append(f"REGION '{_esc(aws.aws_region)}'")
+    if aws.aws_endpoint_url:
+        # DuckDB's ENDPOINT wants host[:port] without the scheme; USE_SSL reflects https.
+        # urlparse mishandles scheme-less strings like "minio.example.com:9000" —
+        # it treats the whole string as a path. Detect this and prepend a dummy
+        # scheme so urlparse gives us a usable netloc.
+        url = aws.aws_endpoint_url
+        if "://" not in url:
+            parsed = urlparse(f"http://{url}")
+            use_ssl = False  # no scheme → assume plain HTTP
```
skrydal: I think in 2026 we should default to SSL rather than plain. If somebody wants to use plain connection they should be explicit about it.

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/duckdb_profiler.py
```diff
@@ -0,0 +1,371 @@
...
+        aws_config: Optional[AwsConnectionConfig],
+        report: DataLakeSourceReport,
+        profiling_config: GEProfilingConfig,
+        platform: str = "s3",
+    ) -> None:
+        self.aws_config = aws_config
+        self.report = report
+        self.profiling_config = profiling_config
+        self.platform = platform
+        self._tmpdir = tempfile.mkdtemp(prefix="datahub-duckdb-profile-")
```
skrydal, line 141: We do not seem to be ensuring this tmp dir is removed in case of a crash.

docker/datahub-ingestion/Dockerfile
treff7es: Good catches — addressed in bbabd4667c:
- The baked set is now explicit in `DATAHUB_DUCKDB_BAKE_EXTENSIONS="httpfs avro aws"`.
- **`aws` was indeed missing**: when no explicit keys are configured the S3 secret uses `PROVIDER credential_chain` (instance-profile/role-based access), and that provider lives in the `aws` extension — so role-based remote profiling would have failed. It is now loaded on demand (only for the credential_chain path) and baked into the image.
- **No `azure`**: abs/Azure remote profiling is not wired (no Azure secret builder / credential path), so baking it would be dead weight. Happy to add it when we wire Azure remote reads.
- **duckdb at RUN time**: yes — `final-full` installs `[all]`, which includes the `s3`/`gcs`/`abs` plugins (each pulls in `data_lake_profiling` → `duckdb`) and none are in `all_exclude_plugins`. The slim/locked stages do not run this bake step, which is intentional.

docker/datahub-ingestion/Dockerfile
treff7es: Right that extensions are version+platform-keyed (`<dir>/v<duckdb-version>/<platform>/`). Within the image this is safe: `duckdb` is pinned via build-constraints and the bake layer depends on the install layer, so a duckdb bump rebuilds and re-bakes. And if a deployment somehow runs a different duckdb than was baked, the profiler now self-heals at runtime — it falls back to `INSTALL` (download) and only fails, with actionable guidance, when the network is unavailable. Expanded the Dockerfile comment to spell this out.

metadata-ingestion/src/datahub/ingestion/source/s3/duckdb_secrets.py
treff7es: Agreed — changed in bbabd4667c. A scheme-less endpoint (`minio.example.com:9000`) now defaults to `USE_SSL true`; plaintext must be opted into explicitly with an `http://` URL.

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/duckdb_profiler.py
treff7es: Fixed in bbabd4667c with `weakref.finalize`, so the temp dir is removed on crash/GC/interpreter-exit even if `close()` is never called (`close()` still triggers it early; it is idempotent). Added a test asserting cleanup on GC without `close()`.

treff7es:
> shouldn't the profiler code be stored in a more general location than S3 source dirs? / abs profiling

Both addressed:

- **Profiler location**: agreed — moved `duckdb_profiler.py` and `duckdb_secrets.py` to `source/data_lake_common/` (alongside `path_spec`/`data_lake_utils`) in 4f4fdb141b, with their unit tests relocated to match. To keep `data_lake_common` free of any source-package dependency, the profiler now types its report against a small structural `Protocol` (`report_warning` + `report_failure`) instead of importing s3's `DataLakeSourceReport`; the concrete source reports satisfy it.
- **abs profiling**: you're right. `abs` does not inherit from `S3Source` (unlike GCS, which routes through it via `create_equivalent_s3_source`), so it wires no profiler — profiling there is effectively a no-op, and the dead Spark `abs/profiling.py` was removed in this PR. That state is pre-existing and unchanged here. Wiring `abs` through the DuckDB profiler needs an Azure extension + an Azure secret builder, so I'd suggest doing it as a dedicated follow-up rather than expanding this PR.

docker/datahub-ingestion/Dockerfile
```diff
@@ -191,6 +191,29 @@ RUN --mount=type=cache,target=$HOME/.cache/uv,uid=1000,gid=1000,sharing=private
...
+# Build-time has network access and buildx runs this on the target platform, so
+# the correct per-platform binary is baked in. DuckDB stores them under
+# <dir>/v<duckdb-version>/<platform>/. duckdb is pinned via build-constraints and
+# this layer depends on the install above, so a duckdb bump rebuilds it and
+# re-populates the matching version dir. If a deployment somehow runs a different
+# duckdb than was baked, the profiler self-heals at runtime — it falls back to
+# INSTALL (a download) and only fails, with actionable guidance, when the network
+# is unavailable. At runtime the profiler reads DATAHUB_DUCKDB_EXTENSION_DIRECTORY,
+# loads from this directory, and never downloads — so profiling works in
+# air-gapped deployments of this image with no extra configuration.
```
skrydal: I have read this part several times and I am convinced that it is self-contradictory. First it says:
> If a deployment somehow runs a different
> duckdb than was baked, the profiler self-heals at runtime — it falls back to
> INSTALL (**a download**) and only fails, with actionable guidance, when the network
> is unavailable.

And then it says:
> At runtime the profiler reads DATAHUB_DUCKDB_EXTENSION_DIRECTORY,
> loads from this directory, and **never downloads** — so profiling works in
> air-gapped deployments of this image with no extra configuration.

So, at the time we are building the image, we have X version of duckdb installed and for this version, below code will download the plugins and store them in `${DATAHUB_DUCKDB_BAKE_EXTENSIONS}/X`.
Then when we are running an ingestion, for, let's say, acryl-datahub 1.6.0.1, which requires Y version, it will need to **download** a new plugin version (I understand it will happen, even if `INSTALL` is not invoked explicitly). In such case, if datahub executor is running in air-gapped environment, it will fail.
Thus this process would need to always be a part of bundling script (if you are bundling anything touching duckdb, then below needs to be triggered). 

This is the biggest problem I am seeing with this PR. It introduces an eternal architectural and maintenance burden. It will be very easy to break this fragile procedure in the future, as people will not be aware such a specialized step needs to be done, if and only if, duckdb-dependent source is installed.

I also think this should be available in the slim image, effectively this PR should introduce s3/gcs profiling capability back into the slim image, and remove the very idea of `s3-slim` and `gcs-slim` sources (there is no need for them at this moment).

metadata-ingestion/src/datahub/ingestion/source/abs/config.py
```diff
@@ -141,6 +140,19 @@ def _infer_platform_from_path_spec(
 
         return "abs" if is_abs_uri(include) else "file"
 
+    @model_validator(mode="after")
+    def reject_unsupported_profiling(self) -> "DataLakeSourceConfig":
+        # Profiling is not implemented for the ABS source: no profiler is wired
+        # in ABSSource, and there is no Azure credential path for the DuckDB
+        # engine that backs the S3/GCS data-lake sources. Fail loudly instead of
+        # silently accepting `profiling.enabled: true` and emitting nothing.
```
skrydal: Imho this comment is unnecessary and will be misleading in the future, I suggest removing it.

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/duckdb_profiler.py
```diff
@@ -0,0 +1,646 @@
...
+    """
+
+
+class UnsupportedDuckDBFormatError(ValueError):
+    """The file's format is not one DuckDB profiling can read.
+
+    A per-table data issue (this object's format is unsupported), not a
+    structural failure — so ``get_table_profile`` reports it as a deduped
+    warning rather than failing the run. Subclasses ``ValueError`` and is raised
+    by ``_open_relation`` for an unknown/empty extension.
```
skrydal: Comment like this "leaks" logic from the actual function (`get_table_profile`) into this place. It will age very badly. Please remove this part of the comment. Same applies for the above.

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/duckdb_profiler.py
```diff
@@ -0,0 +1,646 @@
...
+            # timer before yielding, so the recorded per-table duration measures
+            # profiling — not the downstream sink time during the yields.
+            profiles = list(
+                profiler.generate_profiles([request], max_workers=1, platform="duckdb")
+            )
+            timer.finish()
+            for _req, profile in profiles:
+                if profile is not None:
+                    if row_estimate is not None:
+                        profile.rowCount = row_estimate
```
skrydal, line 591: This is the case of _sampling_ being used. Should we capture also appropriate information in partition spec? i.e.

```python
    if profile.partitionSpec and profile.partitionSpec.type == PartitionTypeClass.PARTITION:
        profile.partitionSpec.partition += " SAMPLE"
    else:
        profile.partitionSpec = PartitionSpecClass(
            type=PartitionTypeClass.QUERY, partition="SAMPLE"
        )
```

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/duckdb_profiler.py
```diff
@@ -0,0 +1,646 @@
...
+        self._finalizer = weakref.finalize(
+            self, shutil.rmtree, self._tmpdir, ignore_errors=True
+        )
+        self._engine: Optional[sa.engine.Engine] = None
+        self._secrets_done = False
+        self._loaded_extensions: Set[str] = set()
+
+    def _engine_lazy(self) -> sa.engine.Engine:
+        if self._engine is None:
+            self._engine = sa.create_engine(f"duckdb:///{self._db_path}")
```
skrydal: `sa.create_engine("duckdb:///…")` here uses SQLAlchemy's default pool. Combined with the instance-level `_secrets_done` / `_loaded_extensions` booleans a few lines above, this is a latent bug on pool churn:

DuckDB's non-`PERSISTENT` secrets are session-scoped (per DBAPI connection). If any code path invalidates the current connection (transaction error, `pool_recycle`, overflow spillover), the next `engine.begin()` returns a fresh connection whose session has no `datahub_s3` secret and no `httpfs` loaded — but `_secrets_done=True` skips re-issue. The next remote read then fails with a cryptic "no HTTP secret" / IO error.

Practical risk is low today: `max_workers=1` + default LIFO pool + no `pool_recycle` means the same connection is normally reused. But any future change to those assumptions (or a mid-run error that invalidates the connection) trips it, and no current test exercises it.

Simplest fix that matches the single-threaded, single-file-DB usage model — pin to one connection:

```python
from sqlalchemy.pool import StaticPool

self._engine = sa.create_engine(
    f"duckdb:///{self._db_path}",
    poolclass=StaticPool,
)
```

That deletes the whole class of "did a different connection come out of the pool" latent bug and lets the existing `_secrets_done` / `_loaded_extensions` guards stay as-is.

metadata-ingestion/src/datahub/ingestion/source/abs/datalake_profiler_config.py
skrydal, line 1: There doesn't seem to be any reason for this file to exist. Please remove it.

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/duckdb_profiler.py
treff7es: Good catch — fixed in `014a0d1`. The engine is now created with `poolclass=StaticPool`, so the single DBAPI connection is reused for the profiler's lifetime and the `_secrets_done` / `_loaded_extensions` guards can't be defeated by a connection coming back fresh from the pool. Kept the guards as-is per your suggestion.

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/duckdb_profiler.py
treff7es: Done in `014a0d1`. Since sampling is applied upstream (`USING SAMPLE` in the materialize step), the delegated `SQLAlchemyProfiler` sees a full table and won't annotate it — so I mark it in `get_table_profile` via a small `_mark_profile_sampled` helper that sets `PartitionSpecClass(type=QUERY, partition="SAMPLE")` (or appends `" SAMPLE"` to an existing PARTITION spec), matching `ge_data_profiler` / `sqlalchemy_profiler`. Extended the sampling unit test to assert it.

metadata-ingestion/src/datahub/ingestion/source/abs/datalake_profiler_config.py
treff7es: Removed in `014a0d1`. `abs/config.py` now imports `DataLakeProfilerConfig` straight from `data_lake_common.datalake_profiler_config`; no re-export shim.

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/duckdb_profiler.py
treff7es: Fixed in `014a0d1`. Both exception docstrings now describe only what the exception means and where it's raised — the narration of how `get_table_profile` routes each one (failure vs. deduped warning) is gone.

metadata-ingestion/src/datahub/ingestion/source/abs/config.py
treff7es: Removed in `014a0d1`. The `ValueError` message already states profiling isn't supported for abs, so the comment added nothing durable.

docker/datahub-ingestion/Dockerfile
treff7es: Thanks — split this into the things you raised:

**1. Self-contradictory comment — fixed** (`014a0d1`). Reworded to the real behavior: at runtime the profiler `SET`s `extension_directory` then `INSTALL … LOAD …`; a matching baked binary makes `INSTALL` a no-network no-op, and only a version/platform mismatch attempts a download.

**2. Air-gapped / "eternal maintenance burden" — I'd push back that it's a burden on us.** It only bites in a narrow, self-inflicted case:
- *Default path — no issue.* The image bakes extensions for the duckdb it ships; runtime duckdb == baked duckdb, so profiling loads locally and never downloads. Holds for executor venvs too, as long as the recipe doesn't pin a different version.
- *Drift path — operator-provisioned by definition.* The only failure is running a **different** duckdb than shipped. But getting a newer `acryl-datahub` into an air-gapped box already requires a deliberate provisioning step (internal mirror / staged wheels) — you can't `pip install` across an air gap. Staging the matching duckdb extensions is one artifact in that same, already-necessary process, and `profiling.duckdb_extension_directory` exists for exactly that. (Extensions aren't PyPI packages — duckdb fetches them from its own repo — so they're a separate artifact the operator stages; that's what the config knob documents.)

And practically this is a **very rarely exercised** combination: air-gapped **and** data-lake profiling **and** a duckdb version differing from the image. I don't think it warrants restructuring the pipeline; the corrected docs + the config knob cover it.

**3. Profiling in the slim image — done in this PR** (`0bf9ee9`). Since DuckDB is a lightweight, JVM-free dep (the whole point of the swap), `final-slim` now installs the full `s3`/`gcs`/`abs` extras and bakes the extensions like `final-full`. I also pulled the bake out of the inline one-liner into a shared `bake_duckdb_extensions.py` (bind-mounted into both stages) so there's a single, discoverable place for it.

**4. Removing the `s3-slim`/`gcs-slim`/`abs-slim` variants — I'd like to keep them for now and do the removal as a focused follow-up.** I traced what depends on them and it's more than an install convenience:
- `PLUGINS_WITH_SLIM_VARIANT = ["s3", "gcs", "abs"]` drives the `datahub-actions` `BUNDLED_VENV_SLIM_MODE` bundling feature, `final-locked` installs them, and a unit test asserts their presence — so removing them is a cross-cutting change to the actions image + bundling, not a one-line `setup.py` delete.
- It's a user-facing breaking change for anyone on `pip install acryl-datahub[s3-slim]`.
- They now double as the duckdb-free install path for platforms without a DuckDB wheel (`armv7l`, musl/Alpine self-builds).

You're right their original rationale (dodging PySpark) is gone, so I agree they're on borrowed time. I'd rather retire them cleanly in a dedicated follow-up — extras + `BUNDLED_VENV_SLIM_MODE` cleanup + deprecation notice + refreshing the now-stale `docs/PYSPARK.md` — than expand this engine-swap PR with that blast radius. The profiling capability you asked for is in the slim image regardless.

skrydal:
PR superseded.

PR #17863, opened 2026-06-11
perf(frontend): slim down sibling fragment in search and data product queries

JohnRTurner:
## Summary

- Adds a new `siblingSearchDatasetFields` GraphQL fragment for sibling entity sub-fetches in search and data product pages
- Updates `searchResultsWithoutSchemaField` to use the slim fragment instead of the full `nonSiblingsDatasetSearchFields` fragment for both the `siblings.siblings` and `siblingsSearch.searchResults` blocks
- Removes three heavyweight fields from sibling fetches: `institutionalMemory`, `lastProfile: datasetProfiles(limit: 1)`, and `statsSummary.topUsersLast30Days`

These fields are not used by the client-side sibling merge logic (`combineEntityWithSiblings` in `siblingUtils.ts`), which only merges ownership, tags, terms, health, subTypes, assertions, custom properties, schema fields, and lightweight stats counts.

Affected queries (all search and data product flows):
- `search`, `searchAcrossEntities`, `scrollAcrossEntities`
- `listDataProductAssets` (data product entity page)

Not affected: the dataset profile page (`getDataset` query in `dataset.graphql`), which intentionally fetches full fields for the primary entity.

## Benchmark

Measured locally against 20 dataset pairs, each with 5 institutional memory docs and 1 dataset profile (5 rounds per variant):

| Metric | OLD | NEW | Improvement |
|--------|-----|-----|-------------|
| Response time (per entity) | 18 ms | 13 ms | **~25% faster** |
| Payload size (per entity) | 4.0 KB | 3.1 KB | **~21% smaller** |
| Payload (20-result search page) | ~79 KB | ~63 KB | **~16 KB saved** |

## Test plan

- [ ] Run `yarn generate` — verifies GraphQL fragment is valid (no unknown fields)
- [ ] Run `yarn type-check` — no TypeScript regressions
- [ ] Open a search results page and verify sibling-bearing datasets display correctly (tags, terms, ownership still shown; merge still works)
- [ ] Open a data product page and verify asset list loads with correct sibling data

### DIFF
--- a/datahub-web-react/src/graphql/search.graphql
+++ b/datahub-web-react/src/graphql/search.graphql
@@ -598,6 +598,66 @@ fragment nonSiblingsDatasetSearchFields on Dataset {
     ...datasetStatsFields
 }
 
+fragment siblingSearchDatasetFields on Dataset {
+    exists
+    name
+    origin
+    platform {
+        ...platformFields
+    }
+    editableProperties {
+        name
+        description
+    }
+    platformNativeType
+    properties {
+        name
+        description
+        qualifiedName
+        customProperties {
+            key
+            value
+        }
+        externalUrl
+        lastModified {
+            time
+            actor
+        }
+    }
+    ownership {
+        ...ownershipFields
+    }
+    globalTags {
+        ...globalTagsFields
+    }
+    glossaryTerms {
+        ...glossaryTerms
+    }
+    subTypes {
+        typeNames
+    }
+    parentContainers {
+        ...parentContainersFields
+    }
+    deprecation {
+        ...deprecationFields
+    }
+    health {
+        ...entityHealth
+    }
+    lastOperation: operations(limit: 1) {
+        lastUpdatedTimestamp
+        timestampMillis
+    }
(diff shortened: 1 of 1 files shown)


chriscollins3456 (approved):
this seems safe to me. i don't think it's going to push the needle much but it does remove some bloat from siblings requests

gabe-lyons:
@benjiaming can you review?

benjiaming (changes requested):
This could be a nice perf win but there are two regressions. We should restore `lastProfile` and `topUsersLast30Days` on `siblingSearchDatasetFields`. Search cards read those off the merged entity, so dbt-primary + warehouse-sibling cases will lose row counts / top users otherwise.

JohnRTurner:
rebased onto master and put lastProfile and topUsersLast30Days back on siblingSearchDatasetFields - you're right, search cards need those for the dbt-primary + warehouse-sibling case. good to go on my end.

benjiaming (approved):

PR #17869, opened 2026-06-11
fix(ingest/fivetran): map Managed Data Lake Glue destinations to <schema>.<table> URNs

treff7es:
## Summary

Fivetran's Managed Data Lake → Glue integration registers **each destination schema as its own AWS Glue database** and each table as a Glue table. The connector previously modeled Glue as a relational warehouse and emitted a three-part `glue.<database>.<schema>.<table>` URN that:

- required a user-supplied `database` (which the REST API can't discover), and
- **never matched** the URNs that DataHub's own Glue source produces (`<glue_db>.<table>`), so Fivetran→Glue lineage silently failed to render.

This routes `glue` through the same two-part `<schema>.<table>` path as `iceberg` — the schema *is* the Glue database. `database` is no longer required or used for Glue, and **multi-schema destinations** (one Glue database per connector — the normal case) now resolve correctly because each lineage row's own schema leads the URN.

### Before / after

| | URN |
|---|---|
| Before | `urn:li:dataset:(glue, <database>.<schema>.<table>, env)` |
| After | `urn:li:dataset:(glue, <schema>.<table>, env)` |

### Verified end-to-end

Against a live Fivetran MDL→Glue destination (Glue database `fivetran_log`, table `account`): the connector now emits `glue,fivetran_log.account,PROD` — byte-identical to a DataHub Glue source ingestion of the same catalog — and lineage stitches automatically (the Glue dataset shows the Fivetran sync job as its upstream).

## Checklist

- [x] Tests added/updated — `TestBuildDestinationUrnGlueMdl` rewritten to assert the two-part shape, including a multi-schema case that config alone cannot express
- [x] Docs updated — `fivetran_pre.md` Glue routing section
- [ ] Breaking changes — n/a (the Glue MDL routing shipped in 1.6.0; this corrects its URN shape before broad adoption)

### DIFF
--- a/metadata-ingestion/docs/sources/fivetran/fivetran_pre.md
+++ b/metadata-ingestion/docs/sources/fivetran/fivetran_pre.md
@@ -128,6 +128,35 @@ destination_to_platform_instance:
 
 The user override always wins; REST destination discovery still runs in the background to fill in any fields the user didn't pin (e.g. `database` from the discovered config when not overridden). Glue routing needs no `database` — the destination schema is the Glue database (see the Glue section above).
 
+**When you must set `platform` yourself.** Discovery only knows how to map the destination services DataHub recognizes (Snowflake, BigQuery, Databricks, and the Managed Data Lake variants). If a REST call succeeds but reports a service the connector doesn't map — or if you run without `api_config` and the recipe-level `destination_platform` doesn't fit a particular destination — the connector cannot guess a platform. It then skips that destination's lineage edges with a one-time structured warning, and you must pin `destination_to_platform_instance.<id>.platform` explicitly to resolve it.
+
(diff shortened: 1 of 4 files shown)


sgomezvillamor (approved):
Clean fix — the two-part URN shape is correct and the tests cover the motivating multi-schema case well.

Two minor nits (non-blocking):

1. **Stale test comment** — `test_managed_data_lake_auto_detects_glue_from_toggle` still says "They still need to set `database` separately…", which is the opposite of what this PR proves. Worth removing.

2. **`_TWO_PART_NAMESPACE_DESTINATION_PLATFORMS` could be derived** — the comment already documents the relationship (`_NON_RELATIONAL - _PATH_STYLE`); making it computed rather than hand-maintained would prevent the two sets from drifting silently when new platforms are added.

---

Issue #7733, opened 2023-03-31
OIDC settings in docker-compose file throw error in datahub frontend


jakobhanna:
**Describe the bug**
Using the compose quickstart file with OIDC customization the datahub frontend container throws an error.

**To Reproduce**
1. Go to docker/quickstart/ and configure `docker-compose-without-neo4j.quickstart`
2. Change the OIDC settings
3. Start datahub with `python3 -m datahub docker quickstart --quickstart-compose-file docker/quickstart/docker-compose-without-neo4j.quickstart.yml`
4. Show the logs

**Expected behavior**
Datahub frontend should work as expected. Without OIDC settings the service starts normal.

**Desktop (please complete the following information):**
 - OS: [Ubuntu 22.04]

**Additional context**
Here the error log: 
```
datahub-frontend-react        | WARNING: An illegal reflective access operation has occurred
datahub-frontend-react        | WARNING: Illegal reflective access by com.google.inject.internal.cglib.core.$ReflectUtils$1 (file:/datahub-frontend/lib/com.google.inject-guice-4.2.3.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)
datahub-frontend-react        | WARNING: Please consider reporting this to the maintainers of com.google.inject.internal.cglib.core.$ReflectUtils$1
datahub-frontend-react        | WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
datahub-frontend-react        | WARNING: All illegal access operations will be denied in a future release
datahub-frontend-react        | Oops, cannot start the server.
datahub-frontend-react        | com.google.inject.CreationException: Unable to create injector, see the following errors:
datahub-frontend-react        | 
datahub-frontend-react        | 1) No implementation for play.cache.SyncCacheApi was bound.
datahub-frontend-react        |   at auth.AuthModule.configure(AuthModule.java:81) (via modules: com.google.inject.util.Modules$OverrideModule -> auth.AuthModule)
datahub-frontend-react        | 
datahub-frontend-react        | 1 error
datahub-frontend-react        |         at com.google.inject.internal.Errors.throwCreationExceptionIfErrorsExist(Errors.java:554)
datahub-frontend-react        |         at com.google.inject.internal.InternalInjectorCreator.initializeStatically(InternalInjectorCreator.java:161)
[...]

PR #17872, opened 2026-06-12
fix(ingest/sql-parsing): preserve case-sensitive column lineage, drop the `lineage()` normalization patch

aviraj-gour:
## Summary

Column-level lineage was silently dropped for mixed-case columns whenever DataHub preserved the column's real casing (case-sensitive/quoted columns on Snowflake, and mixed-case columns on Databricks/Spark/etc.). The old workaround monkey-patched `sqlglot.lineage.lineage` at import time to globally disable its identifier-normalization step. This PR deletes that patch and instead fixes our one call site: we hand the output column to `lineage()` with its identifier marked **case-sensitive**, so sqlglot leaves the name verbatim. Emitted lineage is preserved across all dialects; we just stop hacking a third-party function. Regression tests lock it in.

## Why the patch existed

`lineage(column, sql, ...)` re-normalizes the lookup `column` before matching it against the query's projections, and raises if nothing matches ([`lineage.py#L156`](https://github.com/tobymao/sqlglot/blob/v30.8.0/sqlglot/lineage.py#L156)):

```python
column_name = normalize_identifiers.normalize_identifiers(column, dialect=dialect).name
if not any(select.alias_or_name == column_name for select in selectable.selects):
    raise SqlglotError(f"Cannot find column '{column_name}' in query.")
```

That normalization folds the identifier's case based on its **quoted** flag and the dialect's strategy ([`Dialect.normalize_identifier`](https://github.com/tobymao/sqlglot/blob/v30.8.0/sqlglot/dialects/dialect.py#L1001)):

- **Unquoted** identifier → folded on every case-folding dialect (Snowflake → `UPPER`, Postgres → `lower`).
- **Quoted** identifier → preserved on `UPPERCASE` / `LOWERCASE` / `CASE_SENSITIVE`, but **still folded** on `CASE_INSENSITIVE`.

| Strategy | Examples | Folds a **quoted** identifier? |
| --- | --- | --- |
| `UPPERCASE` | Snowflake, Oracle | No (preserved) |
| `LOWERCASE` | Postgres, default (no dialect) | No (preserved) |
| `CASE_SENSITIVE` | MySQL | No (never folds) |
| `CASE_INSENSITIVE` | Databricks, Spark, Hive, Redshift, Trino/Presto, DuckDB, BigQuery | **Yes — folds even when quoted** |

DataHub does its own normalization **with** schema context — it builds the scope with `MappingSchema(normalize=False)`, so its projections keep real casing (`betStatusId`, `CustomerName`). This lookup is the exception: it gets only a `dialect`, never the schema, so it folds by the dialect default and disagrees with those verbatim projections:

- Snowflake `"FullName"` (quoted, mixed-case) → `FULLNAME` → no match.
- Databricks `betStatusId` (mixed-case, from `SELECT *`) → `betstatusid` → no match.

Either way the raise fires, the call site catches it, and the edge is silently dropped. The patch deleted that fold for every caller.

## The fix

At our single call site in [`sqlglot_lineage.py`](metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py):

```python
output_col_expr = sqlglot.expressions.column(output_col)
output_col_expr.this.meta["case_sensitive"] = True
lineage_node = sqlglot.lineage.lineage(
    output_col_expr,
    statement, dialect=dialect, scope=root_scope, trim_selects=False, copy=False,
)
```

`meta["case_sensitive"]` is sqlglot's opt-out: `normalize_identifiers` prunes and skips any node carrying it ([`normalize_identifiers.py#L66-L67`](https://github.com/tobymao/sqlglot/blob/v30.8.0/sqlglot/optimizer/normalize_identifiers.py#L66)), so the lookup name is matched verbatim on every dialect — the same result the deleted patch produced.

Why it's safe:

- The marker sits on a **throwaway** expression we build only for the lookup; it never enters the AST/scope `lineage()` walks. `to_node` then receives the verbatim string `column_name` — identical to what the old `column_name = column` patch produced — so downstream lineage is unchanged.

## Why the maintainer's "just quote it" is not enough

The sqlglot maintainer's suggested workaround ([#7733](https://github.com/tobymao/sqlglot/issues/7733)) is to pass the column as a quoted identifier — `exp.column(output_col, quoted=True)`. That works on dialects that **exempt quoted identifiers** from folding (`UPPERCASE`/`LOWERCASE`/`CASE_SENSITIVE`/default): on Snowflake, `"CustomerName"` is preserved and the lookup matches.

It is **not** enough on `CASE_INSENSITIVE` dialects (Databricks, Spark, Hive, Redshift, Trino, DuckDB), which fold identifiers **even when quoted** (last table row above). The failure surfaces precisely when the resolved output column is still mixed-case — and that is the `SELECT *` path: DataHub expands `*` from its schema (built `normalize=False`), so the projection keeps its real casing `betStatusId`. `quoted=True` then folds the lookup name to `betstatusid`, which no longer matches the projection `betStatusId` → `Cannot find column` → the edge is dropped. The Unity Catalog view `CREATE VIEW view1 AS SELECT * FROM bet` is exactly this case, and `view1` lost its `fineGrainedLineages` under `quoted=True`.

(BigQuery is also `CASE_INSENSITIVE` but did **not** break: it is in DataHub's `DIALECTS_WITH_CASE_INSENSITIVE_COLS`, so DataHub pre-folds its names to a canonical case and both sides agree. Explicit references like `SELECT betStatusId` also survive `quoted=True`, because qualify already folded the written name — so the fold lands on the same value. Only the schema-expanded mixed-case `*` projection diverges.)

The `case_sensitive` marker suppresses the fold on **every** dialect, so the resolved name is matched verbatim regardless — Snowflake and Databricks alike.

| Resolved column (after qualify) | `quoted=True` alone | `case_sensitive` marker |
| --- | --- | --- |
| `"CustomerName"` on Snowflake — `UPPERCASE` | preserved ✅ | preserved ✅ |
| `betStatusId` via `SELECT *` on Databricks — `CASE_INSENSITIVE` | folded → **dropped** ❌ | preserved ✅ |
| already-folded / lowercase names | unchanged ✅ | unchanged ✅ |

## Tests

Added in `tests/unit/sql_parsing/test_sqlglot_lineage.py` (with goldens):

- `test_snowflake_case_sensitive_quoted_column_lineage` — single projection; asserts the upstream edge keeps casing (`CustomerName`).
- `test_snowflake_case_sensitive_column_through_cte` — same column through a CTE, so casing must survive a multi-hop walk, not just the top-level lookup.
- `test_databricks_case_sensitive_column_lineage` — `SELECT *` over a mixed-case schema on a `CASE_INSENSITIVE` dialect (the Unity Catalog view path). This is the case `quoted=True` silently dropped; it fails if the `case_sensitive` marker is removed.

All three were verified to **fail** without the marker (the revert reproduces `Cannot find column '…'` and empties `column_lineage`) and to pass once restored, so they guard the mechanism, not just the happy path. The Unity Catalog integration tests (`test_ingestion`, `test_ml_model_with_signature_and_run_details`) pass with goldens unchanged.

### DIFF
--- a/metadata-ingestion/src/datahub/sql_parsing/_sqlglot_patch.py
+++ b/metadata-ingestion/src/datahub/sql_parsing/_sqlglot_patch.py
@@ -149,20 +149,6 @@ def _patch_lineage() -> None:
     lineage_py.Node = Node  # type: ignore
     sqlglot.lineage.Node = Node  # type: ignore
 
-    patchy.patch(
-        lineage_py.lineage,
-        """\
-@@ -68,4 +68,6 @@
-     if column is not None:
--        column_name = normalize_identifiers.normalize_identifiers(column, dialect=dialect).name
-+        # column_name = normalize_identifiers.normalize_identifiers(column, dialect=dialect).name
-+        assert isinstance(column, str)
-+        column_name = column
-         if not any(select.alias_or_name == column_name for select in selectable.selects):
-             raise SqlglotError(f"Cannot find column '{column_name}' in query.")
-""",
-    )
-
     # Patch 1: Change set to list for source_columns (preserve column order)
     patchy.patch(
         lineage_py.to_node,
--- a/metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
+++ b/metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
@@ -22,6 +22,7 @@ from typing import (
 import pydantic.dataclasses
 import sqlglot
 import sqlglot.errors
+import sqlglot.expressions
 import sqlglot.lineage
 import sqlglot.optimizer
 import sqlglot.optimizer.annotate_types
@@ -1119,7 +1120,12 @@ def _select_statement_cll(
 
             try:
                 lineage_node = sqlglot.lineage.lineage(
-                    output_col,
+                    # Pass the already-resolved output column as a quoted identifier so
+                    # sqlglot's normalize_identifiers preserves its casing instead of
+                    # case-folding it (which would break the lookup for case-sensitive
+                    # columns). identify=True during qualification makes every projection
+                    # a quoted identifier, so quoted=True is uniformly correct here.
+                    sqlglot.expressions.column(output_col, quoted=True),
                     statement,
                     dialect=dialect,
                     scope=root_scope,
--- a/metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
+++ b/metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
@@ -1119,13 +1119,19 @@ def _select_statement_cll(
                 continue
 
             try:
+                # output_col was resolved by our qualify pass (schema built with
+                # normalize=False), so it already holds the column's real casing.
+                # sqlglot.lineage re-normalizes the lookup name using dialect rules
+                # only -- on case-insensitive dialects (Databricks, Spark, Hive,
+                # Redshift, Trino, ...) that folds the identifier (quoting does not
+                # exempt it), so the lookup no longer matches and lineage is silently
+                # dropped for mixed-case columns. Marking the identifier case_sensitive
+                # makes that normalization a no-op on every dialect (sqlglot's
+                # documented hook); lineage() matches on the identifier name only.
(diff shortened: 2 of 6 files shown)


skrydal (approved):
👍

sgomezvillamor (commented):
🥇

PR #17875, opened 2026-06-12
feat(i18n): add Brazilian Portuguese (pt-BR) locale

benjiaming:
## Summary

- Adds Brazilian Portuguese (pt-BR) as a supported UI language
- 67 i18next namespace JSON files translated under src/i18n/locales/pt-BR/
- Registers the locale in the language selector as "Português (Brasil)"
- Adds a 129-key Remirror editor bundle (src/i18n/remirror/pt-BR.json) so the rich-text editor labels are also translated
- Wires up Ant Design's pt_BR locale and dayjs's pt-br locale for date/time formatting throughout the app
 
Screenshot 2026-06-12 at 2 26 12 AM


v-tarasevich-blitz-brain (commented):
Also we need to add translation import in datahub-web-react/src/utils/dayjs.ts

v-tarasevich-blitz-brain (approved):
I've pushed dayjs import

PR #17880, opened 2026-06-12
fix(ingestion): RecursionError on Avro records inside optional unions

### LINKED ISSUE #17699: Kafka ingestion hits RecursionError on Avro schemas with recursive references in unions
Title: Kafka ingestion hits RecursionError on Avro schemas with recursive references in unions

  ---
  Describe the bug

  The AvroToMceSchemaConverter in schema_util.py fails to parse Avro schemas containing recursive record references
  when the recursion is wrapped in an optional union (e.g., ["null", "demo.TreeNode"]). The converter tracks visited
  record types via _record_types_seen to prevent infinite recursion for direct self-references, but when the recursive
  reference is inside a union, the visited-check does not trigger before re-entering the record conversion. This
  causes unbounded recursion until Python's stack limit is hit.

  The ingestion run is still marked as successful, the Kafka topic entity is created in DataHub, but the
  Schema/Columns tab is empty for the affected topic. Logs contain RecursionError: maximum recursion depth exceeded in
  comparison.

  To Reproduce

  1. Register the following Avro schema against a Kafka topic:
  {
    "type": "record",
    "name": "TreeNode",
    "namespace": "demo",
    "fields": [
      { "name": "value", "type": "string" },
      { "name": "left", "type": ["null", "demo.TreeNode"], "default": null },
      { "name": "right", "type": ["null", "demo.TreeNode"], "default": null }
    ]
  }
  2. Run ingestion with the kafka source targeting that topic
  3. Observe RecursionError: maximum recursion depth exceeded in comparison in the ingestion logs
  4. Navigate to the ingested topic in the DataHub UI and open the Schema tab
  5. Schema/Columns tab is empty despite ingestion being marked successful

  Expected behavior

  The converter should detect the recursive reference inside the union, stop descending, and emit the schema field
  (possibly with a marker indicating a recursive reference back to the parent record). The schema should render
  correctly in the UI.

  Screenshots

  N/A

  Desktop (please complete the following information):

  - OS: Linux (ingestion container)
  - DataHub Version: v1.4.0.3 (also confirmed to exist on master)

Itsdarkhere:
The AvroToMceSchemaConverter only ran its visited-records guard against the outer schema, so a recursive reference wrapped in [null, RecordName] slipped past it and recursed until Python's stack limit was reached. Because avro_schema_to_mce_fields swallows exceptions by default, the RecursionError was silently absorbed: the ingestion run was reported successful, the Kafka topic entity was created, but the Schema/Columns tab showed empty for the affected topic.
The fix computes the unwrapped actual_schema first and runs the visited-records guard against whichever side (outer or unwrapped) is a RecordSchema. Cleanup is also tightened so a call only removes from _record_types_seen the fullname it actually added. Added a regression test, test_avro_schema_with_recursion_inside_optional_union, covering the TreeNode-via-optional-union case from the issue. Fixes #17699


askumar27 (approved):
LGTM. The fix is correct, minimal, and well-tested. The regression test directly reproduces the reported issue ( #17699 )

PR #17884, opened 2026-06-12
feat(domains): Add attribution to domains; add domains to glossary nodes

asikowitz:
Add attribution info to domains just like tags, terms, ownership, structured properties, etc. Because we have the existing `domains` field, this is done by adding a new `domainAssociations` field that is kept in sync with the existing field via a mutation hook (and migration mutator to handle migrating old data).  Also updates change event generator to pass context and patching logic to operate on domainAssociations.

### DIFF
--- a/metadata-io/src/main/java/com/linkedin/metadata/aspect/hooks/DomainsSyncMutationHook.java
+++ b/metadata-io/src/main/java/com/linkedin/metadata/aspect/hooks/DomainsSyncMutationHook.java
@@ -18,7 +18,6 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.LinkedHashSet;
 import java.util.List;
-import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Stream;
 import javax.annotation.Nonnull;
@@ -34,10 +33,10 @@ import lombok.extern.slf4j.Slf4j;
  * <p>Write path:
  *
  * <ul>
- *   <li>If both fields are present and inconsistent (non-propagated URNs from associations ≠
- *       domains) AND domains changed from previous → reject with 400.
- *   <li>If {@code domainAssociations} is provided → derive {@code domains} from non-propagated
- *       entries.
+ *   <li>If both fields are present and inconsistent (all URNs from associations ≠ domains) AND
+ *       domains changed from previous → reject with 400.
+ *   <li>If {@code domainAssociations} is provided → derive {@code domains} from all entries;
+ *       reorder so manual associations (and their URNs) precede propagated ones.
  *   <li>If only {@code domains} is provided (legacy path) → diff against previous and update {@code
  *       domainAssociations}.
  * </ul>
@@ -80,11 +79,12 @@ public class DomainsSyncMutationHook extends MutationHook {
 
     if (associationsProvided) {
       // Both fields present — only allowed if they are already in sync.
-      // "In sync" means the non-propagated URNs from associations equal the domains set.
-      Set<Urn> nonPropagatedUrns = getNonPropagatedUrns(proposed.getDomainAssociations());
-      Set<Urn> domainUrns = new LinkedHashSet<>(proposed.getDomains());
+      // "In sync" means the full set of URNs from associations equals the domains set.
+      Set<Urn> allAssociationUrns = getAllUrns(proposed.getDomainAssociations());
+      Set<Urn> domainUrns = toSet(proposed.getDomains());
 
-      if (!nonPropagatedUrns.equals(domainUrns)) {
+      // When domains is empty, the caller relies on domainAssociations alone — no conflict.
+      if (!domainUrns.isEmpty() && !allAssociationUrns.equals(domainUrns)) {
         boolean domainsChanged = didDomainsChange(proposed, previous);
         if (domainsChanged) {
           throw new ValidationException(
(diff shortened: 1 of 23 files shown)


alexsku:
## Review — feat(domains): Add attribution to domains

**Type:** data model + behavior change (PDL schema, mutation hook, migration, patch builder, ingestion SDK) · **Size:** +1916 / −117 across 22 files

Reviewed the full feature: parallel `domainAssociations` field kept in sync with the legacy `domains` array via `DomainsSyncMutationHook`, a flag-gated `DomainsMigrationMutator` for v1→v2 backfill, compound-key `(attribution.source, domain)` patch semantics (Java + Python), and per-domain context in the change-event generator. The design is coherent and well-tested. One hard blocker (also why CI is red), plus a docs gap and a few edge-case clarifications.

---

### 🔴 Blocking

**`Domains.pdl` schemaVersion not bumped to 2**
`metadata-models/.../domain/Domains.pdl` — the `@Aspect` block is still `{ "name": "domains" }` (implicit `schemaVersion: 1`), but the PR adds `domainAssociations` and ships `DomainsMigrationMutator` with `getTargetVersion() = 2`. CI `check-schema-versions` fails:

```
ERROR: The following aspect(s) changed but their schemaVersion was not bumped:
  metadata-models/src/main/pegasus/com/linkedin/domain/Domains.pdl
NEEDS BUMP  ...Domains.pdl  v1 → v2  [direct change]
```

Beyond the red check, the declared "current" version must equal the migration target (2) to stay consistent with how new writes are stamped. Fix: add `"schemaVersion": 2` to the `@Aspect` block (cf. `DynamicFormAssignment.pdl`) and run `python .github/scripts/bump_schema_versions.py`.

Also failing: `codecov/patch` at 58% (9 uncovered lines in `aspect_helpers/domains.py`).

---

### 🟡 Other issues

- **[medium / operability]** `SpringStandardPluginConfiguration` — backfill migration is gated behind `featureFlags.aspectMigrationMutatorEnabled` (**default false**), but `domainsSyncMutationHook` is registered `enabled(true)` with no flag. At merge, new/edited entities get `domainAssociations`, but existing entities stay un-backfilled until the flag is enabled. Please confirm this staged rollout is intended and that all `domains` readers fall back correctly when `domainAssociations` is empty.

- **[medium / edge-case]** `DomainsSyncMutationHook.java:87-96` — when both fields are present and inconsistent but `domains` is *unchanged* vs. previous, no exception is thrown and `deriveDomainsFromAssociations` silently rewrites `domains` to match associations (e.g. `domains=[A,B]` + `domainAssociations=[A]` silently drops `B`). Likely acceptable since associations are source of truth, but worth a log/comment.

- **[medium / edge-case]** An unattributed `addDomain` (source `""`) followed by an attributed `addDomain` for the same domain URN yields two association entries for one domain. `domains` is deduped by URN, but `domainAssociations` is never deduped by domain — so a domain can appear twice with different sources. Intended, or should the sync hook dedup?

- **[low / dead code]** `DomainsSyncMutationHook.java:209` `getNonPropagatedUrns(...)` is never called; and the inline comment at line 98 ("derive domains from non-propagated entries") contradicts the implementation, which includes propagated URNs in `domains`. Remove the method and fix the comment.

- **[low / dx]** `aspect_helpers/domains.py` `remove_domain` docstring says omitting `attribution_source` uses `"*"` as the source "which the backend expands to every source." The implementation actually reorders the primary keys (domain-first) and truncates the path via `determine_array_primary_keys` — no `*` wildcard. Update the docstring to match the golden.

---

### What's missing

- The schemaVersion bump above.
- No `docs/how/updating-datahub.md` entry — a new aspect field + schema migration warrants one.
- No GraphQL/frontend changes to surface attribution (presumably a follow-up — worth stating explicitly).

### Test coverage
Parity (`PatchBuilderParityTest` + golden), sync hook (489 lines), and migration are well covered. Gaps: error/edge branches of `remove_domain` in `domains.py` (codecov), and confirming `SingleDomainChangeEventGenerator`'s `get(0)` still picks the first *manual* domain with the new ordering.

---

Once `schemaVersion: 2` lands and CI is green, this is in good shape pending answers on rollout and association dedup.

<sub>🤖 Review generated with Claude Code</sub>

metadata-io/src/main/java/com/linkedin/metadata/aspect/hooks/DomainsSyncMutationHook.java
```diff
@@ -0,0 +1,228 @@
...
+        assoc.setDomain(urn);
+        newAssociations.add(assoc);
+      }
+    }
+
+    proposed.setDomainAssociations(newAssociations);
+  }
+
+  /** Reorders associations so that manual (non-propagated) entries come before propagated ones. */
+  private static void sortAssociations(Domains domains) {
```
RyanHolstien, line 168: nit: This being a void instead of returning the sorted list makes the logic look a bit more awkward imo

RyanHolstien (approved):

metadata-io/src/main/java/com/linkedin/metadata/aspect/hooks/DomainsSyncMutationHook.java
asikowitz: Tried having it return a new list but it was a bit more awkward. I did change the logic though to use a comparator and think that's a lot simpler.

PR #17890, opened 2026-06-15
feat(ingestion/iceberg): add ingest_empty_namespaces to skip namespaces with no selected tables

puneetagarwal-datahub:
## Summary

Iceberg REST catalogs can expose a very large number of namespaces. Because namespace emission is gated only by `namespace_pattern` (not `table_pattern`), a recipe that selects a single table via `table_pattern` still emits a container — 5 aspects (status / containerProperties / subTypes / dataPlatformInstance / browsePathsV2) — for **every** namespace allowed by `namespace_pattern`, plus a `list_tables` + `load_namespace_properties` round-trip each. For a catalog with thousands of namespaces that is tens of thousands of MCPs (and a lot of wasted catalog calls) for a single ingested table, and it adds payload pressure on size-limited emit paths.

This is the same behavior the SQL sources have today — `sql_common` emits schema/database containers per `schema_pattern`/`database_pattern`, not `table_pattern`. The difference is scale: SQL schemas number in the tens, while Iceberg REST namespaces can be in the thousands.

## Change

Adds an `ingest_empty_namespaces` config, **default `True`** — so the default behavior is unchanged and stays consistent with the SQL sources. Set it to `False` to emit a namespace container only when at least one of its tables matches `table_pattern`:

- `table_pattern` filtering moves up into table listing (rejected tables are reported as dropped there), so namespaces with no selected tables are skipped before their container is built.
- A namespace whose table listing *fails* still emits its container, as before.

## Notes

- This is **opt-out**: no behavior change for existing users; high-cardinality catalogs can set `ingest_empty_namespaces: false`.
- Not a cosmetic fix — the UI already hides childless browse containers. The point is to avoid emitting / storing / indexing (and making per-namespace catalog calls for) thousands of unrelated namespace entities on a narrow ingestion, and to ease payload pressure on size-capped emit paths.

## Checklist

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/iceberg/iceberg.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/iceberg/iceberg.py
@@ -3,7 +3,7 @@ import logging
 import threading
 import uuid
 from functools import partial
-from typing import Any, Dict, Iterable, List, Optional, Tuple
+from typing import Any, Dict, Generator, Iterable, List, Optional, Tuple
 
 from dateutil import parser as dateutil_parser
 from pyiceberg.catalog import Catalog
@@ -160,7 +160,6 @@ class IcebergSource(StatefulIngestionSourceBase):
         self.stamping_processor = AutoSystemMetadata(
             self.ctx
         )  # single instance used only when processing namespaces
-        self.namespaces: List[Tuple[Identifier, str]] = []
 
     @classmethod
     def create(cls, config_dict: Dict, ctx: PipelineContext) -> "IcebergSource":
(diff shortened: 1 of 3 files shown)


sgomezvillamor (commented):
Is this common practice across other sources? Eg, sources with database and table patterns: do we filter out empty databases there too? Worth double-checking for consistency. If not, is there a specific reason to do this for Iceberg only?

Also, if the concern is cosmetic, AFAIK the UI already filters out browse path containers with no children, so this ingestion-side filtering may not be needed at all.

puneetagarwal-datahub:
> Is this common practice across other sources? Eg, sources with database and table patterns: do we filter out empty databases there too? Worth double-checking for consistency. If not, is there a specific reason to do this for Iceberg only?
> 
> Also, if the concern is cosmetic, AFAIK the UI already filters out browse path containers with no children, so this ingestion-side filtering may not be needed at all.

Iceberg catalogs can have thousands of namespaces (the motivating case had ~7,100) while the recipe selected a single table, so gating containers on namespace_pattern alone emitted a container (5 aspects) + catalog calls for every namespace — tens of thousands of MCPs for one table.

You're right it's inconsistent with sql_common (which emits schema/db containers per schema_pattern, not table_pattern), and the UI already hides childless containers. So I'll keep the current behavior as the default and make empty-namespace skipping opt-out via the ingest_empty_namespaces flag (default True) — no behavior change for anyone, just a lever for high-cardinality catalogs.

sgomezvillamor:
> Iceberg catalogs can have thousands of namespaces (the motivating case had ~7,100) while the recipe selected a single table, so gating containers on namespace_pattern alone emitted a container (5 aspects) + catalog calls for every namespace — tens of thousands of MCPs for one table.

If users are concerned about that, the solution is to configure the appropriate `namespace_pattern`, right? I'm concerned on adding sort of redundant configs.

If the worry is that this issue is subtle and easy to miss for final users because the issue is not even clearly visible to them, adding a separate and redundant config won't really solve it. Better to improve the docs for both database+table and namespace+table patterns, emphasizing that keeping them aligned avoids unnecessary processing of databases/namespaces with no tables. WDYT?

puneetagarwal-datahub:
> is subtle and easy to miss for final users because the issue is not even clearly visible to them, adding a separate and redundant config w

Yes I think educating them and keeping them aligned  to avoid unnecessary processing of databases/namespaces with no tables and not to exploit the receipe makes sense to me too

PR #17891, opened 2026-06-15
feat(ingestion/teradata): skip profiling of oversize tables via profi…

NehaGslab:
## Summary

Large Teradata tables can make profiling extremely slow — a single PROFILE/sample scan over a multi-GB table can run for a very long time and stall profiling for the rest of the schema. This PR implements the existing profile_table_size_limit config for Teradata so oversize tables are skipped before any profiling query is issued.

Teradata previously did not implement generate_profile_candidates, so profile_table_size_limit was silently ignored. This brings Teradata in line with Snowflake, BigQuery, Databricks, and Oracle, which already honor this setting — including their fail-open behavior (tables whose size can't be determined are still profiled).

### What changed

- **`TeradataSource.generate_profile_candidates`**: starts from the full table list (inspector.get_table_names) and removes only the tables DBC positively reports as oversize. Tables absent from DBC.TableSizeV (new/zero-perm tables, or permission asymmetry on DBC) are therefore treated as eligible (fail-open), matching Oracle's handling of NULL stats and BigQuery/Snowflake's handling of unknown size. Matching is case-insensitive, since Teradata object names are case-insensitive and DBC casing may differ from the dialect.
- **Sizing query**(`PROFILE_OVERSIZED_TABLES_QUERY`): selects only the tables at/above the limit using DBC space accounting (`SUM(CurrentPerm) >= :size_limit_bytes` in `DBC.TableSizeV`). Unlike Oracle's stats-based estimate, this reflects real stored size and does not require `COLLECT STATS`.
- **Resilient to DBC permission failures**: if the sizing query fails (commonly: ingestion user lacks `SELECT on DBC.TableSizeV`), we emit a `report.warning(...)` and return None (no filtering) so profiling proceeds for all tables instead of aborting the entire run.
- **Distinct skip reporting**: added a `profiling_skipped_size_limit` report counter (keyed by schema) so size-based skips are visible and not conflated with the `generic profiling_skipped_other` bucket — aligning with the counter BigQuery/Snowflake already emit.
- **Config docs**: added teradata to the `SupportedSources` for `profile_table_size_limit` and documented that Teradata derives size from DBC space accounting.
- Row-count filtering (`profile_table_row_limit`) is intentionally not supported,
  since Teradata has no cheap/reliable DBC row count without `COLLECT STATS`.

### Behavior change

`profile_table_size_limit` defaults to **5 GB**. Because Teradata now honors it, tables
larger than 5 GB will no longer be profiled by default. Users who want to profile large
tables should set `profile_table_size_limit: null` (or a higher value) under `profiling`.

## Test plan

- [x] Unit tests in `tests/unit/test_teradata_source.py`:
  - excludes only oversize tables: candidates are the full table list minus DBC-reported oversize tables, the GB limit is converted to bytes, schema/limit query params are bound, CHAR padding is stripped, and profiling_skipped_size_limit is recorded
  - fail-open on missing size data: a table absent from DBC.TableSizeV stays a candidate (not silently dropped)
  - fail-open on query failure: a DBC.TableSizeV error returns None (profile everything) and emits a warning rather than failing the run
  - falls back (raises NotImplementedError) when no size limit is set, so the base profiler treats all tables as eligible
- [ ] Verified against a real Teradata instance with a large table (profiling skips it
      with a log line; smaller tables still profile)

## Checklist

- [x] Tests added/updated
- [x] Docs updated (connector config reference is auto-generated from the field description)


metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -1436,6 +1449,18 @@ class TeradataSource(TwoTierSQLAlchemySource):
...
+    # Returns only the tables whose total size is at or above the limit. We then
+    # subtract these from the full table list, so any table missing from
+    # DBC.TableSizeV (new/zero-perm tables, permission asymmetry) remains a
+    # profiling candidate (fail-open)
+    PROFILE_OVERSIZED_TABLES_QUERY: str = """
+SELECT TableName AS name
+FROM DBC.TableSizeV
+WHERE DatabaseName (NOT CASESPECIFIC) = :schema (NOT CASESPECIFIC)
+GROUP BY TableName
+HAVING SUM(CurrentPerm) >= :size_limit_bytes
```
treff7es: 🟡 **Boundary inconsistent with Snowflake/Redshift/BigQuery.** Those connectors skip only when `size_in_bytes / 2**30 > profile_table_size_limit` (strict `>`, see `sql_generic_profiler.py:312`), so a table *exactly* at the limit is still profiled. This query uses `>=`, so at exactly N GB Teradata **skips** while the others **profile** — opposite behavior on the same config field.

Suggest `HAVING SUM(CurrentPerm) > :size_limit_bytes`. As a bonus this also softens the `0` footgun (`> 0` skips only tables with stored data instead of literally everything), making the `_validate_profile_table_size_limit` guard less load-bearing.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -3018,3 +3070,98 @@ def close(self) -> None:
...
+            raise NotImplementedError("Teradata only supports size-based candidates")
+
+        size_limit_bytes = size_limit_gb * 1024**3
+
+        try:
+            with self._retry_connect(inspector.engine) as conn:
+                rows = self._retry_execute(
+                    conn,
+                    text(self.PROFILE_OVERSIZED_TABLES_QUERY),
+                    {"schema": schema, "size_limit_bytes": size_limit_bytes},
```
treff7es, line 3441: 🟡 **Double-reported warning on DBC permission failure.** `_retry_execute` defaults to `warn_on_permanent_failure=True`, so on a permanent failure (the exact `SELECT on DBC.TableSizeV` case cited below) `_on_terminal` already emits a generic `report.warning`, and then the `except` block emits a second, more actionable one — two warnings per schema for one root cause.

The codebase has a purpose-built escape hatch for exactly this: `_check_historical_table_exists` passes `warn_on_permanent_failure=False` because it re-reports the exception itself. Recommend doing the same here:
```python
                    {"schema": schema, "size_limit_bytes": size_limit_bytes},
                    warn_on_permanent_failure=False,
```

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -3018,3 +3070,98 @@ def close(self) -> None:
...
+        `threshold_time` (derived from `profile_if_updated_since_days`) is intentionally
+        ignored: this connector currently implements size-based filtering only. The SQL
+        base's `is_dataset_eligible_for_profiling` does not apply freshness either, so
+        time-based filtering is effectively a no-op for Teradata today. It could be added
+        later using `TeradataTable.last_alter_timestamp` from the table cache.
+        """
+        size_limit_gb = self.config.profiling.profile_table_size_limit
+        if size_limit_gb is None:
+            # Nothing we can filter on -> let the base treat all tables as eligible.
+            raise NotImplementedError("Teradata only supports size-based candidates")
```
treff7es: ℹ️ **`raise NotImplementedError` as normal control flow.** The method *is* implemented; raising "not implemented" to signal "no filtering" overloads the exception and relies on the base loop's `except NotImplementedError`. `return None` is behaviorally identical (both → `profile_candidates = None` → all eligible) and honors the `Optional[List[str]]` tri-state contract — note the DBC-failure path a few lines down already uses `return None` for the same "no filtering" outcome, so this would make the two consistent. Minor / optional.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -1507,6 +1532,33 @@ def __init__(self, config: TeradataConfig, ctx: PipelineContext):
...
+            )
+
+        # profile_if_updated_since_days has no effect on Teradata: candidate
+        # selection is size-based only (see generate_profile_candidates), so warn
+        # rather than let users assume freshness filtering is applied.
+        if (
+            self.config.is_profiling_enabled()
+            and self.config.profiling.profile_if_updated_since_days is not None
+        ):
+            logger.warning(
```
treff7es: ℹ️ Consider `self.report.warning(title=..., message=...)` instead of `logger.warning` here. A "your setting is being silently ignored" message is exactly the kind of thing that should appear in the structured ingestion report, not only in logs where it's easy to miss. Minor.

metadata-ingestion/tests/unit/test_teradata_source.py
```diff
@@ -4205,3 +4205,168 @@ def test_no_idcoltype_leaves_autoincrement_alone(self):
             library_col_info={"name": "col1", "nullable": True, "autoincrement": False},
         )
         assert cols[0]["autoincrement"] is False
+
+
+class TestGenerateProfileCandidates:
+    """Size-based profiling candidate filtering (skips large tables)."""
```
treff7es, line 5543: ℹ️ **Test gap (priority ~5/10): pattern-vs-size counter attribution.** `is_dataset_eligible_for_profiling` passes `profile_candidates=None` to `super()` specifically so a pattern-denied table counts into `profiling_skipped_table_profile_pattern`, not `profiling_skipped_size_limit`. That invariant is currently untested — `test_eligibility_counts_size_skip_once...` only exercises the size side with allow-all patterns. A ~10-line test (configure `profile_pattern` to deny an in-candidates table; assert the pattern counter increments and the size counter stays empty) would lock it in and guard against a future refactor mis-attributing the bucket.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Updated the condition to '>'

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Added warn_on_permanent_failure to function call to avoid duplicate warning

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Returned None instead of raising exception

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Added self.report.warning

metadata-ingestion/tests/unit/test_teradata_source.py
NehaGslab: Added test case for table without size skip

treff7es (approved):

PR #17894, opened 2026-06-15
fix(gms): configurable Jackson max JSON property-name length (maxNameLength)

puneetagarwal-datahub:
## Summary

Deeply-nested struct field paths — e.g. dbt column-level lineage carried as keys in `upstreamLineage` patches — can exceed Jackson's default `maxNameLength` of 50,000 characters. When GMS applies such a patch, deserialization fails:

```
Name length (62775) exceeds the maximum allowed (50000, from `StreamReadConstraints.getMaxNameLength()`)
    at PatchItemImpl.applyTemplatePatch(...)
```

Every `StreamReadConstraints` config in the codebase raised `maxStringLength` but left `maxNameLength` at Jackson's default, so a single over-long property name fails ingestion even when the overall payload is small. The exception never includes the offending name, making it hard to diagnose.

## Changes

- Add `MAX_JACKSON_NAME_LENGTH` and `INGESTION_MAX_SERIALIZED_NAME_LENGTH` (env-tunable), mirroring the existing string-length limit, and set `maxNameLength` on every aspect-deserialization mapper: `TemplateUtil`, `PatchItemImpl`, `ObjectMapperFactory` (primary + API-sanitizing), `RAPServletFactory`, `ServletConfig`, `GraphQLController`, `UsageServiceUtil`, `OperationsServiceUtil`.
- Diagnostic logging: log `urn` / aspect / location + length when a patch carries a property **name** over the name limit or a string **value** over the value limit, so the offending content is identifiable from logs (the exception omits it). Tunable via `INGESTION_OVERSIZED_NAME_WARN_THRESHOLD` (default 50,000) and `INGESTION_OVERSIZED_VALUE_WARN_THRESHOLD` (default 10,000,000). The scan is a single allocation-light pass, wrapped in try/catch → debug, so it can never affect ingestion. Note: the string-value limit is **not** raised — the value warning only flags fields approaching the existing 16 MB limit.
- `TemplateUtil` + `PatchItemImpl` regression tests and an `updating-datahub.md` note.

## For review / open question on the default

The limit is **configurable** via env var — that's the core ask. The current **default is 16 MB**, chosen as generous headroom (observed offending names were ~51K–65K chars). Happy to discuss a smaller default if platform prefers a more conservative ceiling; the env var lets deployments tune it either way.

## Notes

- This accepts these names rather than failing the write. A name in the tens of thousands of characters is still awkward to render (UI) and index (Elasticsearch); the cleaner long-term fix is source-side (saner field paths for ultra-nested structs) and is out of scope here.

## Checklist


skrydal:
@puneetagarwal-datahub are you actively working on this PR or just keeping it up-to-date while waiting for the review?
@david-leifker considering scope of the changes, could you take a look?

puneetagarwal-datahub:
> keeping it up-to-date while waiting for the review?

 I am keeping it up-to-date while waiting for the review

david-leifker (approved):
Test coverage, then you're good to go!

puneetagarwal-datahub:
Hi @david-leifker 

One thing which I observed while testing it and the propogation of data till ES. 

`The lineage record is stored in (SQL), but that one column-level edge isn't written to the graph index, so it won't show in lineage — data preserved, visibility lost for that edge. And this ES graph limit isn't new: ES caps keyword terms at 32,766 chars, so any schemaField key between ~32K and 50K already failed to index in ES — it just didn't crash GMS (the crash only happened above the old 50K Jackson limit). So the maxNameLength fix doesn't introduce the ES gap; it just moves the >50K cases from "GMS crash / whole MCP fails" into the same already-existing "edge silently skipped in ES" behavior.`

Error produced Locally:

`index [graph_service_v1], id [...], 
type=illegal_argument_exception, reason=Document contains at least one immense term
in field="source.urn" (whose UTF8 encoding is longer than the max length 32766) ...
type=max_bytes_length_exceeded_exception, reason=bytes can be at most 32766 in length; got 60090`

WDYT ?

cc: @sgomezvillamor @skrydal

david-leifker:
There is previous work for document ids for schema fields entity index. There is a limit to the size of the document id as well. A hash was introduced. It was never really deployed because of the overhead required to translate all of the documents to use the hash. In this case we likely need a similar solution, all document ids and references (at least schema fields) could use a hash as the keyword field values. This will then hit the painful requirement to migrate existing data.

An alternative to consider might be a size based trigger to using the hash in addition to non-hash. This is a fairly large change to make and we'd want a full tech spec or design document review. The surface area includes not just the storage but all possible query paths. We'd need some intelligent interception at the API layer to detect and translate to/from the hash so it is non-trivial.

puneetagarwal-datahub:
> There is previous work for document ids for schema fields entity index. There is a limit to the size of the document id as well. A hash was introduced. It was never really deployed because of the overhead required to translate all of the documents to use the hash. In this case we likely need a similar solution, all document ids and references (at least schema fields) could use a hash as the keyword field values. This will then hit the painful requirement to migrate existing data.
> 
> An alternative to consider might be a size based trigger to using the hash in addition to non-hash. This is a fairly large change to make and we'd want a full tech spec or design document review. The surface area includes not just the storage but all possible query paths. We'd need some intelligent interception at the API layer to detect and translate to/from the hash so it is non-trivial.

Agreed the full fix is the hash approach and warrants a tech spec . This PR ([#17894](https://github.com/datahub-project/datahub/issues/17894)) is scoped only to the Jackson maxNameLength crash — it turns a GMS parse crash into "aspect stored, one oversized column edge not indexed," which is already the pre-existing behavior for 32K–50K keys. Proposing we merge it as the crash fix, and track the ES hashing separately.

PR #17895, opened 2026-06-15
fix(ingestion): budget full schemaMetadata aspect when trimming, not just fields

puneetagarwal-datahub:
## Summary

`EnsureAspectSizeProcessor.ensure_schema_metadata_size` trims schema fields to fit the payload limit, but it only summed `schema.fields` and ignored the rest of the aspect — notably `platformSchema.rawSchema`, plus top-level keys and JSON-escaping overhead. So it trimmed fields to ~the limit, and then the non-field content pushed the serialized aspect back over and GMS rejected it:

```
String value length (16033303) exceeds the maximum allowed (16000000, from `StreamReadConstraints.getMaxStringLength()`)
```

Observed in the field on a deeply-nested Iceberg schema: the existing truncation removed 497 fields but the aspect was still 16,033,303 chars — **33 KB over** the 16 MB limit — purely because ~540 KB of non-field content was never charged against the budget.

## Change

- `ensure_schema_metadata_size` now budgets the **full serialized aspect**, not just the fields list. A new `_schema_size_without_fields` helper measures the non-field content (top-level keys + `platformSchema.rawSchema`), and the field-trimming loop starts from that baseline — so accepted fields **plus** non-field content stay within the constraint.
- When `rawSchema` alone exceeds the budget, it is dropped (with a warning) so the structured fields can still be retained rather than failing the whole aspect.
- Regression tests covering both the budget accounting and the oversized-`rawSchema` case.

## Example

With `schema_size_constraint ≈ 15,492,710` and ~540 KB of non-field content:

```
Before: budget for fields = 15,492,710 (non-field ignored)
        final aspect = 15,492,710 (fields) + 540,593 (non-field) = 16,033,303  → REJECTED (33 KB over 16 M)

After:  field budget starts at 540,593 (non-field charged first)
        final aspect = 15,492,710  → ACCEPTED
```

## Checklist

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
+++ b/metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
@@ -14,6 +14,7 @@ from datahub.ingestion.api.workunit_processor import (
 )
 from datahub.metadata.schema_classes import (
     DatasetProfileClass,
+    OtherSchemaClass,
     QueryPropertiesClass,
     QuerySubjectsClass,
     SchemaFieldClass,
@@ -100,10 +101,48 @@ class EnsureAspectSizeProcessor(WorkunitProcessor[EnsureAspectSizeProcessorRepor
         if truncated:
             self._record_truncation("datasetProfile")
 
+    def _schema_size_without_fields(self, schema: SchemaMetadataClass) -> int:
+        """Serialized size of the aspect excluding the ``fields`` list.
+
+        The ``fields`` list is the part we trim, but the rest of the aspect is
+        emitted too and must count against the budget — in particular
+        ``platformSchema.rawSchema``, which for deeply-nested schemas can be
+        several MB on its own. Budgeting fields alone lets the final aspect
+        exceed the limit even after truncation.
+        """
+        original_fields = schema.fields
+        try:
+            schema.fields = []
+            return len(json.dumps(pre_json_transform(schema.to_obj())))
+        finally:
+            schema.fields = original_fields
+
     def ensure_schema_metadata_size(
         self, dataset_urn: str, schema: SchemaMetadataClass
     ) -> None:
-        total_fields_size = 0
+        # Budget against the full serialized aspect, not just the fields list.
+        # Top-level keys and especially platformSchema.rawSchema are emitted as
+        # well; ignoring them lets the aspect exceed the limit.
+        raw_schema_dropped = False
+        non_fields_size = self._schema_size_without_fields(schema)
+        if (
+            non_fields_size >= self.schema_size_constraint
+            and isinstance(schema.platformSchema, OtherSchemaClass)
+            and schema.platformSchema.rawSchema
+        ):
+            # The raw schema blob alone exceeds the budget. Drop it so the
+            # structured fields can still be retained instead of failing the
+            # whole aspect.
+            self.ctx.source_report.warning(
+                title="Schema truncated due to size constraint",
+                message="Dataset schema contained too much data and would have caused ingestion to fail",
+                context=f"Raw schema was removed from schema for {dataset_urn} due to aspect size constraints",
+            )
+            schema.platformSchema.rawSchema = ""
+            non_fields_size = self._schema_size_without_fields(schema)
+            raw_schema_dropped = True
+
+        total_fields_size = non_fields_size
         accepted_fields: List[SchemaFieldClass] = []
         for schema_field in schema.fields:
             field_size = len(json.dumps(pre_json_transform(schema_field.to_obj())))
@@ -116,7 +155,7 @@ class EnsureAspectSizeProcessor(WorkunitProcessor[EnsureAspectSizeProcessorRepor
                     message="Dataset schema contained too much data and would have caused ingestion to fail",
                     context=f"Field {schema_field.fieldPath} was removed from schema for {dataset_urn} due to aspect size constraints",
                 )
-        if len(accepted_fields) < len(schema.fields):
+        if raw_schema_dropped or len(accepted_fields) < len(schema.fields):
             self._record_truncation("schemaMetadata")
         schema.fields = accepted_fields
 
--- a/metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
+++ b/metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
@@ -14,7 +14,6 @@ from datahub.ingestion.api.workunit_processor import (
 )
 from datahub.metadata.schema_classes import (
     DatasetProfileClass,
-    OtherSchemaClass,
     QueryPropertiesClass,
     QuerySubjectsClass,
     SchemaFieldClass,
@@ -105,10 +104,15 @@ class EnsureAspectSizeProcessor(WorkunitProcessor[EnsureAspectSizeProcessorRepor
         """Serialized size of the aspect excluding the ``fields`` list.
 
         The ``fields`` list is the part we trim, but the rest of the aspect is
-        emitted too and must count against the budget — in particular
-        ``platformSchema.rawSchema``, which for deeply-nested schemas can be
-        several MB on its own. Budgeting fields alone lets the final aspect
-        exceed the limit even after truncation.
+        emitted too and must count against the budget — chiefly the
+        ``platformSchema`` blob, which for deeply-nested schemas can be several
+        MB on its own, plus top-level scalars (``hash``, ``schemaName``, …).
+
+        We measure by temporarily clearing ``fields`` and serializing the whole
+        aspect, rather than sizing ``platformSchema`` alone, so the budget
+        reflects *all* non-field content the server will parse (nothing is missed
+        if new top-level fields are added later). The mutation is local and
+        restored in the ``finally``.
         """
         original_fields = schema.fields
         try:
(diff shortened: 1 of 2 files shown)


sgomezvillamor:
Nice catch! The root cause diagnosis is solid and the regression tests are a good addition. A few points:

---

**⚠️ Warning — priority ordering**

`platformSchema` content should be trimmed/dropped *before* `fields` are trimmed. `fields` contain the structured metadata that users actually query and depend on; the raw platform schema blob is the least valuable part of the aspect. The PR gets this right for `OtherSchemaClass` (drops `rawSchema` first), but for every other union variant the recovery block is skipped, `total_fields_size` starts at an already-oversized `non_fields_size`, and the field loop rejects every single field — leaving 0 fields retained *and* the aspect still over the limit.

The correct sequence: drop/trim the `platformSchema` string → re-measure → only then start trimming fields.

---

**💡 Suggestion — union coverage gap**

The drop-rawSchema recovery path only handles `OtherSchemaClass`. The other union variants also carry potentially large string fields:

- `KafkaSchema`: `documentSchema` (often a full Avro schema), `keySchema`
- `BinaryJsonSchema`, `OrcSchema`: `schema`
- `MySqlDDL`, `OracleDDL`: `tableSchema`
- `PrestoDDL`: `rawSchema`
- `EspressoSchema`: `documentSchema`, `tableSchema`
- `KeyValueSchema`: `keySchema`, `valueSchema`

(Note: `AvroSchema` is defined but is **not** in the `platformSchema` union, so it's a non-issue.)

If fixing all variants is out of scope for this PR, please add a TODO comment with this list so the gap is visible.

---

**💡 Suggestion — `_schema_size_without_fields` mutation approach**

Zeroing out `schema.fields` and serializing the whole aspect is clever but surprising for future readers. An alternative worth considering:

```python
len(json.dumps(pre_json_transform(schema.platformSchema.to_obj())))
```

The concern is that it misses other top-level fields (`foreignKeys`, `primaryKeys`, `cluster`, `dataset`, …). But those are typically small scalars; the 1.5% buffer from `schema_size_constraint = payload_constraint * 0.985` likely covers them. If you want to be explicit you could sum them individually, or slightly tighten the 0.985 factor. Either way, a short comment explaining why the mutation was chosen over the simpler alternative would help future readers.

---

puneetagarwal-datahub:
> Nice catch! The root cause diagnosis is solid and the regression tests are a good addition. A few points:
> 
> **⚠️ Warning — priority ordering**
> 
> `platformSchema` content should be trimmed/dropped _before_ `fields` are trimmed. `fields` contain the structured metadata that users actually query and depend on; the raw platform schema blob is the least valuable part of the aspect. The PR gets this right for `OtherSchemaClass` (drops `rawSchema` first), but for every other union variant the recovery block is skipped, `total_fields_size` starts at an already-oversized `non_fields_size`, and the field loop rejects every single field — leaving 0 fields retained _and_ the aspect still over the limit.
> 
> The correct sequence: drop/trim the `platformSchema` string → re-measure → only then start trimming fields.
> 
> **💡 Suggestion — union coverage gap**
> 
> The drop-rawSchema recovery path only handles `OtherSchemaClass`. The other union variants also carry potentially large string fields:
> 
> * `KafkaSchema`: `documentSchema` (often a full Avro schema), `keySchema`
> * `BinaryJsonSchema`, `OrcSchema`: `schema`
> * `MySqlDDL`, `OracleDDL`: `tableSchema`
> * `PrestoDDL`: `rawSchema`
> * `EspressoSchema`: `documentSchema`, `tableSchema`
> * `KeyValueSchema`: `keySchema`, `valueSchema`
> 
> (Note: `AvroSchema` is defined but is **not** in the `platformSchema` union, so it's a non-issue.)
> 
> If fixing all variants is out of scope for this PR, please add a TODO comment with this list so the gap is visible.
> 
> **💡 Suggestion — `_schema_size_without_fields` mutation approach**
> 
> Zeroing out `schema.fields` and serializing the whole aspect is clever but surprising for future readers. An alternative worth considering:
> 
> ```python
> len(json.dumps(pre_json_transform(schema.platformSchema.to_obj())))
> ```
> 
> The concern is that it misses other top-level fields (`foreignKeys`, `primaryKeys`, `cluster`, `dataset`, …). But those are typically small scalars; the 1.5% buffer from `schema_size_constraint = payload_constraint * 0.985` likely covers them. If you want to be explicit you could sum them individually, or slightly tighten the 0.985 factor. Either way, a short comment explaining why the mutation was chosen over the simpler alternative would help future readers.
> 

Priority ordering (the regression): Fixed. The platform schema is now dropped before fields are trimmed, then re-measured, so a variant whose blob alone exceeds the budget no longer leaves 0 fields + an oversized aspect.

Union coverage: Generalized rather than TODO'd. New _drop_platform_schema blanks every string member of platformSchema via its backing store, so it covers all variants (rawSchema / documentSchema / tableSchema / keySchema / valueSchema) without an isinstance chain to keep in sync. Dropped the now-unused OtherSchemaClass import. Added a regression test with a KafkaSchema whose documentSchema alone exceeds the budget — asserts the blob is dropped, all fields retained, and the aspect fits (this is the case that previously regressed to 0 fields).

_schema_size_without_fields mutation: Kept the clear-fields-and-serialize approach (it captures all non-field top-level content, not just platformSchema, so it stays correct if new top-level fields are added) and expanded the docstring to explain why it's used over sizing platformSchema alone, noting the mutation is local and restored in finally.

sgomezvillamor (approved):
All feedback addressed — great iteration. Approved.

---

sgomezvillamor:
**💡 Suggestion — observability metric for platform schema drops**

`EnsureAspectSizeProcessorReport` already tracks `num_truncations_by_aspect` for field-level truncations. It would be useful to also track the cases where the platform schema blob itself is dropped, e.g.:

```python
@dataclass
class EnsureAspectSizeProcessorReport(WorkunitProcessorReport):
    num_truncations_by_aspect: Dict[str, int] = field(default_factory=dict)
    num_platform_schema_drops: int = 0  # incremented by _drop_platform_schema
```

This makes it easier for users to spot in the ingestion report when `_drop_platform_schema` silently blanks content — right now the only signal is the warning log. A dedicated counter makes it immediately visible in the report summary without having to grep through warnings. The name `num_platform_schema_drops` (or `platform_schema_drops_count`) feels self-explanatory, but open to alternatives.

---

puneetagarwal-datahub:
> **💡 Suggestion — observability metric for platform schema drops**
> 
> `EnsureAspectSizeProcessorReport` already tracks `num_truncations_by_aspect` for field-level truncations. It would be useful to also track the cases where the platform schema blob itself is dropped, e.g.:
> 
> ```python
> @dataclass
> class EnsureAspectSizeProcessorReport(WorkunitProcessorReport):
>     num_truncations_by_aspect: Dict[str, int] = field(default_factory=dict)
>     num_platform_schema_drops: int = 0  # incremented by _drop_platform_schema
> ```
> 
> This makes it easier for users to spot in the ingestion report when `_drop_platform_schema` silently blanks content — right now the only signal is the warning log. A dedicated counter makes it immediately visible in the report summary without having to grep through warnings. The name `num_platform_schema_drops` (or `platform_schema_drops_count`) feels self-explanatory, but open to alternatives.
> 

Added num_platform_schema_drops to EnsureAspectSizeProcessorReport (commit cfb03e84b4f), incremented whenever an oversized platform schema is dropped — so it surfaces in the report summary, not just the warning log. Used num_platform_schema_drops (consistent with num_truncations_by_aspect)

metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
```diff
@@ -100,10 +103,75 @@ def ensure_dataset_profile_size(
...
+        """
+        platform_schema = schema.platformSchema
+        if platform_schema is None:
+            return False
+        cleared = False
+        for key, value in platform_schema._inner_dict.items():
+            if isinstance(value, str) and value:
+                platform_schema._inner_dict[key] = ""
+                cleared = True
+        return cleared
```
skrydal, line 146: Why can't we just set `platformSchema` to `None`?
@puneetagarwal-datahub

puneetagarwal-datahub: platformSchema is a required union (no optional keyword) in metadata-models/src/main/pegasus/com/linkedin/schema/SchemaMetadata.pdl . None fails serialization .

metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
```diff
@@ -100,10 +103,75 @@ def ensure_dataset_profile_size(
...
+            self.ctx.source_report.warning(
+                title="Schema truncated due to size constraint",
+                message="Dataset schema contained too much data and would have caused ingestion to fail",
+                context=f"Raw platform schema was removed from schema for {dataset_urn} due to aspect size constraints",
+            )
+            self.report.num_platform_schema_drops += 1
+            non_fields_size = self._schema_size_without_fields(schema)
+            truncated = True
+
+        total_fields_size = non_fields_size
```
skrydal, line 174: let's analyze the logic here. We are protecting ourselves from cases where raw platform schema itself exceeds 16 MB (in such case it will be removed - because we want to have more columns rather than raw schema). What if platform schema accounts for 10 MB and fields also 10 MB? In such case we will keep platform schema and then remove approximately 4 MB of columns. Shouldn't we rather start trimming with the schema itself? That would be adjusted to the previous statements that we value actual columns over raw platform schema.

metadata-ingestion/tests/unit/api/source_helpers/test_ensure_aspect_size.py
```diff
@@ -386,10 +387,10 @@ def test_ensure_size_of_too_big_schema_metadata(processor):
     )
     assert len(schema.fields) < 1004, "Schema has not been properly truncated"
     assert schema.fields[-1].fieldPath == "dddd", "Small field was not added at the end"
-    # +100kb is completely arbitrary, but we are truncating the aspect based on schema fields size only, not total taken
-    # by other parameters of the aspect - it is reasonable approach though - schema fields is the only field in schema
-    # metadata which can be expected to grow out of control
-    assert len(json.dumps(schema.to_obj())) < INGEST_MAX_PAYLOAD_BYTES + 100000, (
+    # The truncated aspect must fit within the budget. The budget now accounts
+    # for the full serialized aspect (not just the fields list), so the result
+    # is bounded by the payload constraint rather than an arbitrary slop.
```
skrydal, line 392: This comment is AI slop, please remove it. Whenever Claude puts "now" in the comment it is a bad comment really, it describes transition and not the state of the code.
The +100k took into account additional bytes which are added due to JSON serialization. The new size estimation procedure still does not take them into the account (not that it is a problem). The only reason we are not hitting this now is the fact that scaling factor is used (0.985).

metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
skrydal: Ok got it, I have taken a look at the model. Indeed we can not just set it to None, however clearly we want to do it here. Maybe we should consider changing the model to make `platformSchema` optional?
Moreover, each type of schema (`OrcSchema`, `OtherSchema`) seems to just have one field in the record - either `rawSchema` or `schema`. Maybe we should just refer to them here explicitly? That is if platformSchema belongs to those types then set to empty string `.schema`, if other types then set to empty string `.rawSchema`, otherwise log the fact that we do not recognize the schema and won't be "trimming" it?
This would gives us some predictability to how the procedure works, instead of iterating over _inner_dict, note that if there is non-string field it won't be emptied in current approach, but we have no guarantees when will it happen.

metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
puneetagarwal-datahub: @sgomezvillamor @skrydal  

Before [#17895](https://github.com/datahub-project/datahub/pull/17895): budget counted fields only — fields were trimmed but the raw schema on top still pushed the total over the limit → ingestion failed.

[#17895](https://github.com/datahub-project/datahub/pull/17895) (merged): budget now counts the whole aspect, so it actually fits — but in a mixed case it keeps the raw schema and trims columns.

**One more angle for fix:** when over budget, drop the raw schema first (only when fields would fit without it), then trim fields — so we keep columns over the raw schema

Should we do this one ? WDYT

metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
puneetagarwal-datahub: Review here for the fix: https://github.com/datahub-project/datahub/pull/17934

metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
puneetagarwal-datahub: Review here for the fix: https://github.com/datahub-project/datahub/pull/17934

metadata-ingestion/tests/unit/api/source_helpers/test_ensure_aspect_size.py
puneetagarwal-datahub: Review here for the fix: https://github.com/datahub-project/datahub/pull/17934

metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
sgomezvillamor: Yes, definitely needs to be addressed.

The goal should be to remove as few fields as possible. And so:
- Remove the platform schema first
- Then, if further trimming is needed, remove fields gradually

PR #17896, opened 2026-06-15
ci(i18n): add to CI i18n keys parity script

v-tarasevich-blitz-brain:
This PR adds i18n locales parity script and adds it to CI


datahub-web-react/src/i18n/locales/de/entityV1.shared.entityDropdown.json
v-tarasevich-blitz-brain, line 1: This file is unused after #17743/#17782

datahub-web-react/build.gradle
```diff
@@ -166,6 +166,24 @@ task yarnLint(type: YarnTask, dependsOn: [yarnInstall, yarnGenerate]) {
   }
 }
 
+// Verifies non-EN locale JSON files stay in sync with the `en` source of truth (stale keys,
+// placeholder drift). Only reads JSON, so it skips the expensive yarnGenerate codegen step.
+task yarnI18nParityCheck(type: YarnTask, dependsOn: [yarnInstall]) {
+  args = ['run', 'check-i18n-parity']
+  inputs.files(project.fileTree(dir: 'src/i18n/locales', include: ['**/*.json']))
```
chriscollins3456, line 173: minor but we could probably also add `inputs.file('scripts/check-i18n-parity.mjs')` so that if this script file changes, we know to run this task as well as opposed to just the json files in src/i18n/locales

chriscollins3456 (approved):
nice i think this makes sense! thanks for this

one minor comment not blocking

datahub-web-react/build.gradle
v-tarasevich-blitz-brain: I'll add another issue for that. Thanks

v-tarasevich-blitz-brain: https://linear.app/acryl-data/issue/CAT-2501/follow-up-to-trigger-ci-when-check-i18n-parity-script-changed

PR #17900, opened 2026-06-15
test(playwright/ml): migrate ML tests to playwright

purnimagarg1:
**Linear ticket:**
https://linear.app/acryl-data/issue/PFP-4088/migrate-ml-tests

**Description:**

Migrate ML entity Playwright tests with full feature parity and code quality improvements.

**New Files:**

- `e2e-test/ui/playwright/pages/ml-entities.page.ts` — Shared page object for ML entity pages
- `e2e-test/ui/playwright/tests/ml-entities-v2/v2-experiment.spec.ts`
- `e2e-test/ui/playwright/tests/ml-entities-v2/v2-model-mlflow.spec.ts`
- `e2e-test/ui/playwright/tests/ml-entities-v2/v2-model-sagemaker.spec.ts`


priyabratadas-dh (commented):
## Playwright Code Sentinel Review

**Verdict: PASS** — No critical violations found. The PR follows POM conventions, uses the correct `base-test` fixture, adds `data-testid` attributes to production components, and correctly marks the Cypress tests as migrated.

---

### Warnings to address before merge

**1. Coverage gap in `v2-experiment.spec.ts` — Recommend fixing**

The Cypress source (`experiment.js`) verified specific metric values for `accuracy` and `learning_rate`. The Playwright spec navigates to the Metrics tab but only asserts tab visibility, not the metric values themselves. The PR description says "full feature parity" but this assertion is missing.

Suggested addition:
```typescript
// In the metrics tab section
await expect(page.getByText('accuracy')).toBeVisible();
await expect(page.getByText('learning_rate')).toBeVisible();
```

**2. Double `waitForPageLoad()` in `v2-experiment.spec.ts`**

The page object's navigation method already calls `waitForPageLoad()` internally, but the spec calls it again immediately after. One call is redundant and adds unnecessary wait time.

**3. Missing `logger.step()` instrumentation — all 3 spec files**

All three specs receive `logger` in `beforeEach` but never call `logger.step()` or `logger.info()`. Without step markers, CI failures produce logs with no breadcrumbs to trace which step broke. Example pattern from other migrated tests:
```typescript
logger.step('Navigating to ML experiment page');
await mlPage.goToMlExperiment(...);
```

**4. Fat test in `v2-model-mlflow.spec.ts`**

The single `'MLflow model page has correct data'` test performs 8+ sequential operations across Summary, Properties, and Group tabs. If any assertion fails, it's hard to isolate which tab/operation caused it. Consider splitting into one `it` block per tab.

**5. Confusing navigate-away-and-back pattern in `v2-model-sagemaker.spec.ts`**

The test clicks a model group link then immediately calls `page.goBack()`. If the intent is to assert the link exists and is navigable, the click is fine but a comment explaining the intent would help. If the navigation is unintentional, remove it.

---

### Not migrated (track separately)

`feature_table.js` (mlFeatureTable entities) is still in Cypress — should be tracked in a follow-up ticket.

---

> Reviewed by Playwright Code Sentinel · [DataHub E2E Standards](https://github.com/datahub-project/datahub/blob/master/e2e-test/ui/playwright/PLAYWRIGHT.md)

priyabratadas-dh (approved):

PR #17901, opened 2026-06-15
fix(i18n): add missed de/es/pt-BR translations

v-tarasevich-blitz-brain:
Adds missing translations after adding https://github.com/datahub-project/datahub/pull/17896


chriscollins3456 (approved):
cool - this will obviously need a rebase once the one ahead is merged to remove those commits but the new commits seem fine if you feel good about the translations

PR #17904, opened 2026-06-15
fix(auth): enforce authorization for logicalParent, Data Product and Query entity

david-leifker:
## Summary

Moves authorization for three previously under-enforced aspects into the MCP validator layer so GraphQL, Rest.li, and direct ingestion all share identical rules. Auth failures now return 403 instead of 500 on ingest paths.

---

### Logical Parent (`logicalParent` aspect)

Authorization is checked against **both** the child dataset (the entity whose `logicalParent` is being written) and the parent dataset being referenced.

| Scenario | Result |
|---|---|
| Edit Entity on both child and parent | ✅ allowed |
| Edit Entity on the child only | ❌ denied |
| Edit Entity on the parent only | ❌ denied |
| No Edit Entity on either | ❌ denied (UI + GraphQL `setLogicalParent`) |

`setLogicalParent` has no resolver-level pre-check — the aspect validator is the sole enforcement point across all APIs.

---

### Data Product membership (`dataProductProperties.assets`)

Membership changes travel two independent paths. Either path is sufficient for MCP/ingestion; GraphQL routes each mutation through the path matching its UI surface.

**Product-side** (`batchSetDataProduct` with a product URN — Data Product page)
- Requires **Manage Data Products** on at least one domain associated with the product.
- The product and changed assets do not need to share a domain.
- Multi-domain products: manage on **any one** product domain suffices.
- Product with no domain → denied (fail-closed).
- Edit Data Product on an asset alone is not sufficient.

**Asset-side** (`batchAddToDataProducts`, `batchRemoveFromDataProducts`, `batchSetDataProduct` with null URN — asset profile / sidebar)
- Requires **Edit Data Product** on each asset being linked or unlinked.
- Manage Data Products on a domain does not authorize these flows by itself.

**MCP / direct ingestion**
- Allowed when either product-side manage or asset-side edit succeeds for every asset in the diff (additions and removals use the same symmetric check).

**Bootstrap / admin**
- Default `datahub` user and Admin/Editor roles can ingest and manage data product membership without extra policies after deploy/restart.
- Fresh installs receive new domain-scoped "Manage Data Products on All Domains" policies for root user, Admin, and Editor (`metadata-service/war/src/main/resources/boot/policies.json`).

---

### Query entity reads

Applies when view authorization is enabled.

| Scenario | Result |
|---|---|
| View Entity Page on **all** subject datasets | ✅ allowed |
| Edit Dataset Queries on **all** subject datasets | ✅ allowed |
| Edit Entity on **all** subject datasets | ✅ allowed |
| Access to some but not all subject datasets | ❌ entire query hidden |
| Query has no subjects | ❌ denied (fail-closed) |

Schema field subjects are resolved to their parent dataset for authorization. Query write rules (Edit Dataset Queries or Edit Entity on every subject) are unchanged.

---

### Known limitation — domain drift

Authorization runs only when `dataProductProperties.assets` changes. Re-domaining a Data Product or asset does not retroactively re-check or remove existing membership. Cross-domain membership can persist after a move; this is accepted behavior under the OR-path model and is tracked separately.


asikowitz (changes requested):
Would like setting logical parent to require permissions on both parent and child

craig-rueda:
## PR Review

**PR**: #17904 — fix(auth): enforce authorization for logicalParent, Data Product and Query entity
**Author**: david-leifker
**Base**: master ← fix/ac-002-ac-003-central-auth
**Type**: behavior change + security fix
**Size**: +3832 / -514 across 43 files

Moves authorization enforcement for three previously under-enforced aspects — `logicalParent`, `dataProductProperties.assets`, and Query entity reads — into the MCP validator layer so GraphQL, Rest.li, and direct ingestion all share identical rules. Auth failures now surface as 403 instead of 500 on ingest paths. Adds bootstrap policies for root/Admin/Editor to manage data products on all domains. Comprehensive tests (unit + smoke) and extensive docs updates accompany the changes.

**Key invariants:**
1. `logicalParent` writes require `EDIT_ENTITY` on both the child and the proposed parent (clearing requires only child)
2. Data product membership changes require either `MANAGE_DATA_PRODUCTS` on any product domain (product-side) OR `EDIT_DATA_PRODUCT` on every changed asset (asset-side)
3. Query entity reads require `VIEW_ENTITY_PAGE` or `EDIT_DATASET_QUERIES` on ALL subject datasets (fail-closed when no subjects)
4. System/internal writes (`SYSTEM_UPDATE_SOURCE`) bypass auth validators
5. MCP validator auth failures propagate as 403 through `IngestProposalExceptionUtils`

---

## Risk Assessment

| | |
|---|---|
| **Risk** | Medium |
| **Blast radius** | All DataHub API surfaces (GraphQL, Rest.li, MCP ingestion); existing ingestion pipelines writing `logicalParent` or `dataProductProperties` |
| **Rollback** | Safe — validators are gated by config properties with kill switches; boot policies are additive |
| **Rollout** | New validators are enabled by default (`matchIfMissing = true`). Existing deployments gain auth enforcement on upgrade. Deployments relying on unauthenticated ingestion of logicalParent/dataProductProperties will break unless policies are in place or the validators are disabled via config. |
| **CI** | In progress; lint and security checks pass; codecov/patch shows 69.6% patch coverage |

---

## Blocking Issues

### [BLOCKER] Logical-models docs say parent auth is NOT checked — code checks both child and parent
**Location**: `docs/features/feature-guides/logical-models/overview.md:161`

The docs state:
> "Authorization requires **Edit Entity** on the **child dataset** being updated, **not on the proposed parent URN**."

This directly contradicts the code in `LogicalParentAuthorizationValidator.java:44-46` which adds the parent URN to the auth check set. It also contradicts the PR description ("checked against **both** the child dataset and the parent dataset") and `docs/authorization/policies.md` which correctly says both are checked.

**Suggested fix**: Replace the sentence with: "Authorization requires **Edit Entity** on the **child dataset** being updated and, when setting a parent, on the **proposed parent** dataset as well. Clearing a logical parent requires Edit Entity on the child dataset only."

---

## High-Priority Issues

### [HIGH] `UpdateNameResolver.updateDataProductName` wraps AuthorizationException in RuntimeException
**Location**: `datahub-graphql-core/.../resolvers/mutate/UpdateNameResolver.java:330-332` — `updateDataProductName()`

The `catch (Exception e)` at line 330 catches the `AuthorizationException` thrown at lines 311/315 and re-wraps it as `RuntimeException`, losing the 403 status code. Users renaming a Data Product without permission will see a 500 error instead of 403.

**Suggested fix**: Either narrow the catch to only expected exceptions, or use `IngestProposalExceptionUtils.toGraphQLException()` as done in other resolvers, or re-throw `AuthorizationException` before the generic catch:
```java
} catch (AuthorizationException e) {
    throw e;
} catch (Exception e) {
    throw new RuntimeException(..., e);
}
```

### [HIGH] `batchUnsetDataProduct` doesn't use `IngestProposalExceptionUtils` — auth failures return 500
**Location**: `datahub-graphql-core/.../resolvers/dataproduct/BatchSetDataProductResolver.java:88` — `batchUnsetDataProduct()`

`batchSetDataProduct` (line 74) correctly uses `IngestProposalExceptionUtils.toGraphQLException()`, but `batchUnsetDataProduct` (line 88) still uses bare `new RuntimeException(...)`. If the MCP `DataProductMembershipAuthorizationValidator` rejects an unset operation, the client gets 500 instead of 403.

**Suggested fix**: Replace `new RuntimeException(...)` with `IngestProposalExceptionUtils.toGraphQLException(...)` to match the set path.

---

## Other Issues

- **[MEDIUM / correctness]** `SpringStandardPluginConfiguration.java:457-492` — Both `privilegeConstraintsValidator()` and `tagPrivilegeConstraintsValidator()` are `@Bean` with the same `@ConditionalOnProperty`, creating two bean entries for the same validator. When validators are collected via `List<AspectPayloadValidator>`, the tag validator could appear twice, causing double validation. Fix: remove the `@Bean` annotation from `privilegeConstraintsValidator()` or make it a simple alias without `@Bean`.

- **[MEDIUM / edge-case]** `IngestProposalExceptionUtils.java:83-84` — `isUnauthorizedMessage` uses `message.toLowerCase().contains("unauthorized")` as a fallback to detect auth failures. Any validation message incidentally containing "unauthorized" (e.g., "Field X unauthorized-format is invalid") would be misclassified as an auth failure and return 403 instead of 400. Fix: tighten the substring check or remove this fallback if `ValidationSubType.AUTHORIZATION` is reliably set.

- **[MEDIUM / dx]** `ListQueriesResolver.java:121` — `result.setTotal(gmsResult.getNumEntities())` reports the unfiltered total when query entities are filtered by view auth. Clients may under-paginate, thinking there are fewer results than exist. This is a known limitation documented in the test, but worth a follow-up to adjust total or document in the API.

---

## What's Missing

- No smoke test for `updateName` on Data Products — the smoke tests cover `setLogicalParent`, `batchSetDataProduct`, and query reads, but not the `updateName` path which has its own auth logic.
- No test for the `UI_SOURCE` behavioral change — the `TagPrivilegeConstraintsValidator` no longer skips `UI_SOURCE` writes (the old `PrivilegeConstraintsValidator` did). The unit test was updated but no E2E test confirms the behavior change.

## Questions for Author

1. **Double-bean registration**: Was the intent for `privilegeConstraintsValidator()` to serve as a backward-compatible bean alias? If so, should it lose its `@Bean` annotation to avoid registering the same validator twice?

2. **UI_SOURCE no longer skipped for tags**: The old `PrivilegeConstraintsValidator` skipped `UI_SOURCE` events with a comment "handled by LabelUtils." The new `AbstractAspectAuthorizationValidator.requiresValidation()` only skips `SYSTEM_UPDATE_SOURCE`. Was removing the UI_SOURCE skip intentional? Are there UI flows that relied on this bypass?

3. **Rollout risk for existing ingestion**: Both new validators default to enabled (`matchIfMissing = true`). Existing ingestion pipelines writing `logicalParent` or `dataProductProperties` without appropriate policies will start failing with 403. Is this the desired upgrade behavior, or should these default to disabled for the first release?

4. **`isAuthorizedToManageDataProductsOnAllDomains` naming**: This deprecated method delegates to `OnAnyDomain` — was the original intent truly "all domains" (AND) or was this a naming correction? Callers expecting AND semantics would silently get OR behavior.

craig-rueda (approved):
lgtm

asikowitz (approved):

PR #17906, opened 2026-06-15
perf(auth): resolve each resource once per authorization request

acrylJonny:
## Summary

Authorization decisions are correct today, but **expensive to compute on entity-heavy pages**. Rendering a list/search/domain page calls the GraphQL `EntityPrivileges` resolver per result, which fans out into ~15–22 `authorize()` calls per entity (one per privilege). Two compounding inefficiencies made each of those calls do avoidable work:

1. **Wasteful check order.** `PolicyEngine#isPolicyApplicable` evaluated the *resource* match before the *actor* match. Applicability is `actor AND resource`, so the order never changes the decision — but resolving whether a resource is in scope can trigger an uncached, recursive domain/container hierarchy walk against GMS, while the user/group/role actor predicates are resource-independent and already memoized per request. Policies that could never apply to the requester still paid for resource resolution.

2. **No resource-spec reuse across the fan-out.** Every `authorize()` call rebuilt a fresh `ResolvedEntitySpec`, re-resolving the *same* resource's domain/container/owner attributes from scratch. Measured live, a single domain-scoped policy caused one dataset's domain to be resolved **9 times within a single privileges query**.

This PR fixes both without changing any authorization outcome or touching the policy cache (freshness is unaffected).

## What changed

**1. Cheapest-first, owner-deferred ordering** (`PolicyEngine`)
Evaluate the resource-independent actor predicates (user/group/role) first; only fall through to resource scope + ownership when a policy can actually apply to the actor. Resource-ownership policies are handled so they cost no more than the original ordering.

**2. Request-scoped resolved-spec memoization** (the holistic fix)
A new **opt-in** capability interface, `ResourceSpecCachingAuthorizer`, lets the per-request `AuthorizationContext` share one `ResolvedEntitySpec` cache across every check in that request. `DataHubAuthorizer` and `AuthorizerChain` implement it and resolve each spec via `computeIfAbsent`, so a resource's attributes (and any hierarchy walk) are resolved **at most once per request**.

- The public `Authorizer` SPI is **unchanged**. Authorizers that don't implement the new interface resolve per-call exactly as before, and cache-less callers pass `null`.
- Because all production authorization funnels through `AuthorizationContext → AuthorizerChain → DataHubAuthorizer`, every call site picks up the benefit automatically — no per-resolver changes.

## Impact

| | Before | After |
| --- | --- | --- |
| Resource (domain) resolutions for one entity in a single privileges query | **9×** | **1×** |
| Authorization decisions | — | **identical** |
| Policy cache freshness | — | **unchanged** |

## No functional change

Applicability remains `actor AND resource`; reordering and memoization only remove redundant work. This is asserted directly:

- Two `DataHubAuthorizerTest` cases were tightened to prove resource resolution is *skipped* when the actor can't match (`getV2` called `times(0)`).
- A new deterministic guard, `testResourceResolvedOncePerRequestAcrossPrivilegeChecks`, authorizes one resource for multiple privileges in a shared request cache and asserts the domain is resolved exactly once (`getV2` `times(1)`) while every decision stays `ALLOW`.

## Testing

- `auth-impl` unit suite green, including the new guard and the two tightened cases.
- New smoke test `test_policy_eval_perf.py`: loads a domain's contents for a many-group user as domain-scoped policies accumulate (10 groups/20 policies and 100 groups/50 policies) and asserts page latency does not scale with policy count.
- Validated end-to-end on a live stack: the query that produced 9 domain resolutions now produces 1.

## Checklist

- [x] PR title follows the `perf:` convention
- [x] Tests added/updated (unit guard + smoke perf regression)
- [ ] Docs — n/a (internal performance change, no behavior change)
- [ ] Breaking changes — none

### DIFF
--- a/metadata-auth/auth-api/src/main/java/com/datahub/plugins/auth/authorization/ResourceSpecCachingAuthorizer.java
+++ b/metadata-auth/auth-api/src/main/java/com/datahub/plugins/auth/authorization/ResourceSpecCachingAuthorizer.java
@@ -1,0 +1,32 @@
+package com.datahub.plugins.auth.authorization;
+
+import com.datahub.authorization.AuthorizationRequest;
+import com.datahub.authorization.AuthorizationResult;
+import com.datahub.authorization.EntitySpec;
+import com.datahub.authorization.ResolvedEntitySpec;
+import java.util.Map;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * Opt-in capability for {@link Authorizer}s that can share resolved resource specs across the many
+ * authorization checks issued within a single request.
+ *
+ * <p>Rendering a page's per-result action affordances authorizes each entity for ~15-22 privileges,
+ * and resolving a resource's authorization attributes (domain, container, owner) can force a
+ * recursive hierarchy walk against GMS. That resolution is actor-independent, so an implementor can
+ * resolve each resource at most once per request via the supplied cache.
+ *
+ * <p>The public {@link Authorizer} contract is unchanged: non-implementors resolve per call, and
+ * callers without a cache pass {@code null}.
+ */
+public interface ResourceSpecCachingAuthorizer {
+
+  /**
+   * @param resourceSpecCache request-scoped cache of resolved specs, or {@code null} to resolve
+   *     without caching
+   */
+  AuthorizationResult authorize(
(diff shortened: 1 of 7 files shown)


acrylJonny:
Superseded by #17908, opened from a branch on datahub-project/datahub directly (same-repo) instead of the fork. Identical changes.

PR #17907, opened 2026-06-15
fix(glue): encode / in database and table names to produce valid DataHub URNs

rob-1019:
## Summary

- Glue allows `/` in database and table names, but `UrnEncoder` does not encode `/` (only `(`, `)`, and `,` are reserved). A literal `/` in the URN name field breaks DataHub lookups.
- Added `_glue_name_to_urn_part()` which pre-encodes name components with `urllib.parse.quote` before URN construction. Names without special chars are unchanged; `(` and `)` continue encoding as `%28`/`%29`; `/` now encodes as `%2F`.
- Also documents (via test) that `AllowDenyPattern` treats `allow` entries as regexes — users must escape `(` and `)` when specifying literal database names in `database_pattern.allow` (engineering workaround: `raw__user-provided_\(upd/upo\)_framework`).

## Test plan

- [ ] 17 new unit tests in `tests/unit/glue/test_glue_special_chars.py` — all pass
- [ ] 83 existing Glue unit tests — all still pass
- [ ] Covers the exact reported database name `raw__user-provided_(upd/upo)_framework` and the engineers' suggested escaped pattern
- [ ] End-to-end test against a running DataHub instance (TODO)

## Checklist

- [x] PR conforms to the Contributing Guideline (PR title uses `fix(glue):` prefix)
- [x] Tests added/updated
- [ ] Docs added/updated (not required — config behavior unchanged)
- [ ] Breaking changes documented (none — URN encoding change only affects tables with `/` in names, which were previously broken anyway)

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/aws/glue.py
@@ -16,7 +16,7 @@ from typing import (
     Set,
     Tuple,
 )
-from urllib.parse import quote as _urlquote, urlparse
+from urllib.parse import quote, urlparse
 
 import botocore.exceptions
 import yaml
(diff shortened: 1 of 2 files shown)


rob-1019:
Closing — investigation showed the root cause is user configuration (regex metacharacters in database_pattern.allow need escaping) rather than a code bug. No code change warranted.

PR #17909, opened 2026-06-15
refactor(ingest/dremio): drop unused query classification + comment stripping

acrylJonny:
## Summary

`DremioQuery` computed three attributes on every construction — `query_without_comments`, `query_type`, `query_subtype` — but **nothing in the connector ever reads them** (only the unit tests did). They are not used for filtering, emission, lineage, or operation types.

The comment-stripping that fed them (`get_raw_query`) ran a **full sqlglot parse + regenerate per query**. Because it passed no dialect, it used sqlglot's **base** dialect, which:
- mis-tokenizes Dremio-specific SQL → `Invalid expression / Unexpected token` warnings, and
- drops unsupported args — e.g. rewrites `TO_CHAR(ts, 'YYYY-MM-DD')` to `CAST(ts AS TEXT)` and logs `Argument 'format' is not supported for expression 'ToChar' when targeting Dialect`.

Since the outputs are dead, this PR removes the whole chain rather than just fixing the dialect:
- `get_raw_query`, `_get_query_type`, `_get_query_subtype`
- the `DremioQueryType` enum and the operator lookup table
- the now-unused `re` / `parse_one` / `StrEnum` imports
- the corresponding unit tests

This eliminates the warning **and** the per-query sqlglot parse on large Dremio query histories.

Separately, sets `format_queries=False` on the Dremio `SqlParsingAggregator` (mirroring `snowflake-queries`) to skip the per-query reformatting step, reducing CPU/memory pressure.

No user-facing/emitted behavior changes, so no `updating-datahub.md` entry.

## Test plan

- [x] `pytest tests/unit/dremio/` — 238 passed.
- [x] `./gradlew :metadata-ingestion:lintFix` — clean.

## Checklist

- [ ] Documentation update (if applicable)

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_entities.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_entities.py
@@ -1,10 +1,7 @@
 import logging
-import re
 import uuid
 from datetime import datetime
 from typing import Any, ClassVar, Dict, Iterator, List, Optional
-
-from sqlglot import parse_one
 
 from datahub.emitter.mce_builder import make_term_urn
 from datahub.ingestion.source.common.subtypes import DatasetContainerSubTypes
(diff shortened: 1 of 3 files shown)


acrylJonny:
Superseded by #17910 (same branch, pushed to the upstream repo instead of a fork).

PR #17910, opened 2026-06-15
perf(ingest/dremio): skip query reformatting and remove dead query-processing code

acrylJonny:
## Summary

Removes redundant per-query work in the Dremio connector that hurt performance on large query histories, and fixes spurious sqlglot warnings. Three related changes plus a golden refresh.

### 1. Skip query reformatting (`format_queries=False`)

Sets `format_queries=False` on the Dremio `SqlParsingAggregator` (mirroring `snowflake-queries`) to skip the per-query parse-and-regenerate reformatting step, reducing CPU/memory pressure on large histories. As a result, `queryProperties` now stores the **raw** query text rather than the sqlglot pretty-printed form.

### 2. Drop dead query classification + comment stripping

`DremioQuery` computed `query_without_comments`, `query_type`, `query_subtype` on every construction, but nothing in the connector ever read them. The comment-stripping that fed them (`get_raw_query`) ran a full sqlglot parse + regenerate per query with **no dialect**, falling back to sqlglot's base dialect, which:

- mis-tokenizes Dremio-specific SQL → `Invalid expression / Unexpected token` warnings, and
- drops unsupported args — e.g. logs `Argument 'format' is not supported for expression 'ToChar' when targeting Dialect`.

Removed the whole chain (`get_raw_query`, `_get_query_type`, `_get_query_subtype`, the `DremioQueryType` enum and operator table, the now-unused `re` / `parse_one` / `StrEnum` imports, and the corresponding unit tests). Eliminates the warnings **and** the redundant per-query parse.

### 3. Drop inert affected-table lineage path

`_get_affected_tables()` always returned `""`, so the `add_known_query_lineage` branch in `process_query` never fired. The `SqlParsingAggregator` already extracts both upstreams and the affected/downstream table from each observed query via sqlglot (mirroring the mssql/oracle/postgres SQL connectors), so connector-side downstream extraction is redundant. Removed the `affected_dataset` scaffolding and the dead branch; kept the query/catalog format-validation warning by running it unconditionally on `queried_datasets`.

### Golden files

Regenerated the Dremio integration goldens. The only meaningful change is that `queryProperties` statements go from multi-line sqlglot output to single-line raw text, a direct consequence of change (1).

## Test plan

- [x] `pytest tests/unit/dremio/` — passed.
- [x] `pytest tests/integration/dremio/test_dremio.py` — 3 passed (goldens regenerated).
- [x] `./gradlew :metadata-ingestion:lintFix` — clean.

## Checklist

- [ ] Documentation update (if applicable)

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_entities.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_entities.py
@@ -1,10 +1,7 @@
 import logging
-import re
 import uuid
 from datetime import datetime
 from typing import Any, ClassVar, Dict, Iterator, List, Optional
-
-from sqlglot import parse_one
 
 from datahub.emitter.mce_builder import make_term_urn
 from datahub.ingestion.source.common.subtypes import DatasetContainerSubTypes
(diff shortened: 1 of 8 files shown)


metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_source.py
```diff
@@ -295,6 +294,7 @@ def __init__(self, config: DremioSourceConfig, ctx: PipelineContext):
             usage_config=self.config.usage,
             # Gate SQL-discovered URNs through the catalog-walk filters.
             is_allowed_table=self._is_allowed_table,
+            format_queries=False,
```
treff7es, line 297: ✅ Verified: `format_queries` is a real `SqlParsingAggregator` param (`:419`) gating `_maybe_format_query` (`:642`); `False` stores raw query text. The 7 golden `statement` diffs confirm it (`"SELECT\n  *\nFROM ..."` → `"SELECT * FROM ..."`). Mirrors `snowflake-queries`. No concern.

metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_source.py
```diff
@@ -729,67 +729,16 @@ def _validate_query_lineage_format(self, query: DremioQuery) -> None:
...
-                # — see lineage_dropped_filtered's docstring for the "lineage
-                # references, not edges" semantics.
-                self.report.lineage_dropped_filtered += 1
-
-        # Always register the observed query; usage stats aren't gated here.
-        # Aspects for filtered URNs that the SQL parser rediscovers are blocked
-        # by SqlParsingAggregator's is_allowed_table gate at emission time.
+        # Surface query/catalog naming mismatches (e.g. S3/HDFS paths or
+        # versioned refs that won't resolve against catalog metadata).
+        self._validate_query_lineage_format(query)
```
treff7es, line 732: Good change — calling this unconditionally is safe (it only reads `queried_datasets`/`job_id`, never `affected_dataset`) and is actually an **improvement**: it was previously gated behind the always-falsy `affected_dataset`, so the S3/HDFS/path/versioned-ref mismatch warnings never fired. They now surface.

ℹ️ Optional (pre-existing, not a regression): aggregator-side `is_allowed_table` drops don't feed `report.lineage_dropped_filtered`, so the filtered-lineage count may under-report vs. the old (dead) code's intent. Worth a follow-up ticket if accurate filtered counts matter — the old branch never ran, so nothing regresses here.

metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_entities.py
```diff
@@ -102,32 +38,19 @@ def set_glossary_term(self, glossary_term: str) -> str:
...
-        self.query_without_comments = self.get_raw_query(query)
-        self.query_type = self._get_query_type()
-        self.query_subtype = self._get_query_subtype()
         self.queried_datasets = self._get_queried_datasets(queried_datasets)
-        if affected_datasets:
-            self.affected_dataset = affected_datasets
-        else:
-            self.affected_dataset = self._get_affected_tables()
 
     def get(self, attr):
```
treff7es: ℹ️ Optional cleanup: `DremioQuery.get()` has no callers — the `query.get(field)` in `is_valid_query` operates on a raw `dict`, not this class. It predates this PR, but since this PR removes the last attributes any dynamic `.get()` could have targeted, it's now unambiguously dead scaffolding and fits the cleanup intent if you want to drop it.

metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_entities.py
```diff
@@ -138,50 +61,11 @@ def _get_submitted_ts(self, timestamp: str) -> datetime:
     def _get_query(self, query: str) -> str:
```
treff7es, line 58: ℹ️ Out of scope (pre-existing) — flagging only so it isn't mistaken for something this PR missed: `.replace("'", "'")` looks like a no-op but is likely smart-quote normalization (U+2019 → U+0027). Worth a separate look to confirm it does what's intended.

treff7es (approved):

PR #17918, opened 2026-06-16
feat(i18n): cover UserListV2 utils and ingest utils

v-tarasevich-blitz-brain:
This PR extracts i18n keys from 
- src/app/identity/user/UserListV2.utils.ts
- src/app/ingest/source/utils.ts

### DIFF
--- a/datahub-web-react/src/app/identity/user/UserListV2.utils.ts
+++ b/datahub-web-react/src/app/identity/user/UserListV2.utils.ts
@@ -21,7 +21,7 @@ export const getUserStatusText = (userStatus: CorpUserStatus | undefined | null,
     }
 
     // If no status, default to Inactive
-    return i18next.t('users.statusInactive', { ns: 'entity.identity' });
+    return i18next.t('entity.identity:users.statusInactive');
 };
 
 type StatusFilterOption = {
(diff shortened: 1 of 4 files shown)


datahub-web-react/src/app/identity/user/UserListV2.utils.ts
```diff
@@ -31,9 +31,24 @@ type StatusFilterOption = {
 };
 
 export const STATUS_FILTER_OPTIONS: StatusFilterOption[] = [
```
chriscollins3456, line 33: yeah i don't think this is ever even used anyways

datahub-web-react/src/app/ingest/source/utils.ts
```diff
@@ -414,7 +416,10 @@ export const extractEntityTypeCountsFromFacets = (
             .forEach((agg) =>
                 finalCounts.push({
                     count: agg.count,
-                    displayName: pluralize(agg.count, capitalizeFirstLetterOnly(agg.value) || ''),
+                    displayName: i18next.t('ingestion:source.entityTypeNameCount', {
+                        count: agg.count,
+                        type: capitalizeFirstLetterOnly(agg.value) || '',
+                    }),
```
chriscollins3456, line 422: this one is kind of odd since we're just passing in the agg value from the backend but i don't see any other way here than to use that value

chriscollins3456 (approved):

PR #4244, opened 2022-02-24
fix(ingest) kafka-connect: Pass the env variable as part of making dataset

arunvasudevan:
## Checklist
- [ X] The PR conforms to DataHub's [Contributing Guideline](https://github.com/linkedin/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [Commit Message Format](https://github.com/linkedin/datahub/blob/master/docs/CONTRIBUTING.md#commit-message-format))

Datasets created as part of kafka connect ingestion are created with default `PROD` env. This fix is to pass the env variable from the recipe.

PR #17920, opened 2026-06-16
chore(ingest/sql-parsing): remove redundant scope-traverse cycle-detection patch

aviraj-gour:
## Summary

Removes DataHub's `_patch_scope_traverse()` monkey-patch from `_sqlglot_patch.py`. It reimplemented sqlglot's `Scope.traverse()` in pure Python with cycle detection to turn an infinite loop into a raised `OptimizeError`. The hang it guarded against is now fixed at its root cause upstream [sqlglot](https://github.com/tobymao/sqlglot/issues/7295), present in our pinned 30.8.0), so the patch is dead code for any known input.

## What caused the hang

DataHub runs sqlglot's optimizer with `validate_qualify_columns=False`. On certain `UNION ALL` queries (repro: sqlglot [#4244](https://github.com/tobymao/sqlglot/pull/4244#issuecomment-2415249989), unresolved columns made `unnest_subqueries.decorrelate()` treat a `UNION` branch as a correlated subquery and run [`select.parent.replace(alias)`](https://github.com/tobymao/sqlglot/blob/v30.8.0/sqlglot/optimizer/unnest_subqueries.py#L273) — replacing a whole `SELECT` branch with a bare column `_u_0.user_id`. That single line is what led to the infinite loop, via this chain:

1. The `UNION`'s left operand is now a `Column` where a `SELECT` must be.
2. In [`scope._traverse_union`](https://github.com/tobymao/sqlglot/blob/v30.8.0/sqlglot/optimizer/scope.py#L702-L738), building the scope for that branch calls `_traverse_scope`, which has no case for a `Column`: it [logs `Cannot traverse scope … with type Column` and returns **without yielding**](https://github.com/tobymao/sqlglot/blob/v30.8.0/sqlglot/optimizer/scope.py#L689-L691).
3. Because nothing is yielded, the loop variable in `_traverse_union` is never reassigned and still points at the **parent union scope**. The next branch then runs [`union_scope.union_scopes = [prev_scope, scope]`](https://github.com/tobymao/sqlglot/blob/v30.8.0/sqlglot/optimizer/scope.py#L732), recording the union scope **as its own child**.
4. [`Scope.traverse()`](https://github.com/tobymao/sqlglot/blob/v30.8.0/sqlglot/optimizer/scope.py#L568-L589) is a DFS over `union_scopes`; it steps into the scope that contains itself and loops forever.

So the cycle was never a `traverse()` problem — it was a corrupted AST from `decorrelate()`.

## Why DataHub's earlier fix did NOT fix this

DataHub previously carried a second patch, `_patch_unnest_subqueries()` (removed in [#17772](https://github.com/datahub-project/datahub/pull/17772)), which guarded `decorrelate()`'s later `_replace(parent_predicate, …)` calls with `if parent_predicate:`. We verified that guard, applied alone to pre-fix sqlglot, **still hangs** on the [4244 query](https://github.com/tobymao/sqlglot/pull/4244#issuecomment-2415249989) — because the corrupting line is `select.parent.replace(alias)`, which runs *before* the guarded calls and doesn't touch `parent_predicate`. That guard only ever fixed a *crash* sibling of the same defect, not this hang — which is why the separate `_patch_scope_traverse()` net existed at all.

### DIFF
--- a/metadata-ingestion/tests/unit/sql_parsing/test_sqlglot_patch.py
+++ b/metadata-ingestion/tests/unit/sql_parsing/test_sqlglot_patch.py
@@ -4,9 +4,7 @@ import time
 
 import pytest
 import sqlglot
-import sqlglot.errors
 import sqlglot.lineage
-import sqlglot.optimizer
 import sqlglot.optimizer.unnest_subqueries
 from sqlglot import exp
 
@@ -33,19 +31,6 @@ def test_cooperative_timeout_sql() -> None:
             time.sleep(0.0001)
 
     assert 0.6 <= timer.elapsed_seconds() <= 1.2
-
-
-def test_scope_circular_dependency() -> None:
-    scope = sqlglot.optimizer.build_scope(
-        sqlglot.parse_one("WITH w AS (SELECT * FROM q) SELECT * FROM w")
-    )
-    assert scope is not None
-
-    cte_scope = scope.cte_scopes[0]
-    cte_scope.cte_scopes.append(cte_scope)
-
-    with pytest.raises(sqlglot.errors.OptimizeError, match="circular scope dependency"):
-        list(scope.traverse())
 
 
 def test_lineage_node_subfield() -> None:
--- a/metadata-ingestion/src/datahub/sql_parsing/_sqlglot_patch.py
+++ b/metadata-ingestion/src/datahub/sql_parsing/_sqlglot_patch.py
@@ -7,7 +7,6 @@ import patchy.api
 import sqlglot
 import sqlglot.expressions
 import sqlglot.lineage
-import sqlglot.optimizer.scope
 
 from datahub.utilities.is_pytest import is_pytest_running
 from datahub.utilities.unified_diff import apply_diff
@@ -25,7 +24,6 @@ from datahub.utilities.unified_diff import apply_diff
 # With sqlglot[c] (C tokenizer), patchy works for pure-Python functions but fails on those
 # compiled by mypyc to .so extensions. For compiled targets we use alternative strategies:
 # - __deepcopy__ in expression_core.py: wrapper function
-# - Scope.traverse in optimizer/scope.py: full Python reimplementation
 # - lineage.lineage / lineage.to_node: load lineage.py as a pure-Python module and patch that
 
 _DEBUG_PATCHER = is_pytest_running()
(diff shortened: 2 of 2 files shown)


metadata-ingestion/tests/unit/sql_parsing/test_sqlglot_patch.py
```diff
@@ -35,19 +33,6 @@ def test_cooperative_timeout_sql() -> None:
     assert 0.6 <= timer.elapsed_seconds() <= 1.2
 
 
-def test_scope_circular_dependency() -> None:
-    scope = sqlglot.optimizer.build_scope(
-        sqlglot.parse_one("WITH w AS (SELECT * FROM q) SELECT * FROM w")
```
sgomezvillamor, line 40: why removing this query?
what's outcome of parsing this query with scope traverse patching removed?

aviraj-gour: test only existed to verify the patch exist. It manually injects a cycle (cte_scope.cte_scopes.append(cte_scope)) and asserts traverse() raises OptimizeError — but that error was raised only by the patch's cycle-detection code.

Outcome with the patch removed: The test no longer fails — it hangs forever. Upstream's traverse() has no cycle detection, so on that hand-injected self-referential scope it loops infinitely (CI would have to kill it on timeout). It can never pass, since the OptimizeError it expects is gone with the patch. That's why it's deleted rather than kept.

sgomezvillamor: I'm confused here.

My understanding is that `"WITH w AS (SELECT * FROM q) SELECT * FROM w"` was intentionally malformed to force the test scenario — and removing the patch causes it to hang. Does that mean by removing the patch a malformed user query could cause ingestion to hang and eventually fail?

aviraj-gour: The removed test_scope_circular_dependency didn't test that a query hangs — it manually appended a scope to its own cte_scopes to fabricate a cycle, then asserted the patch raised. It only exercised the patch's own code, so it dies with the patch. 

Query is not malformed, if you check test code we add a scope to its own cte_scopes to fabricate a cycle

sgomezvillamor: ah ok, I see

```
    cte_scope = scope.cte_scopes[0]
    cte_scope.cte_scopes.append(cte_scope)
```

sgomezvillamor (approved):

PR #12840, opened 2025-03-11
feat(ingest/salesforce): include formula in calculated field descript…

mayurinehate:
…ions

## Checklist

- [ ] The PR conforms to DataHub's [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [Commit Message Format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#commit-message-format))
- [ ] Links to related issues (if applicable)
- [ ] Tests for the changes have been added/updated (if applicable)
- [ ] Docs related to the changes have been added/updated (if applicable). If a new feature has been added a Usage Guide has been added for the same.
- [ ] For any breaking change/potential downtime/deprecation/big changes an entry has been made in [Updating DataHub](https://github.com/datahub-project/datahub/blob/master/docs/how/updating-datahub.md)

PR #17926, opened 2026-06-16
fix(ingest/salesforce): guard against empty record count during profiling

brock-acryl:
## Summary

When Salesforce profiling is enabled, the connector calls `/limits/recordCount?sObjects=<Object>` for each object and unconditionally indexed the first element of the response's `sObjects` list:

```python
sobject_record_counts = sObject_record_count_response.get("sObjects", [])
return sobject_record_counts[0]
```

Salesforce omits some objects from the `recordCount` response (returning `{"sObjects": []}`) depending on the object and org configuration. In that case the `[0]` access raises `IndexError: list index out of range`. Because the exception propagated unhandled, a single object with no available record count **aborted the entire ingestion run** — every remaining object was lost.

This regressed in #12840, which refactored `SalesforceSource` into a `SalesforceApi` helper and replaced a safe iteration loop with the unsafe `[0]` index access.

## Changes

- **`get_approximate_record_count`** now returns `Optional[SObjectRecordCount]`. On an empty `sObjects` list it records a warning and returns `None` instead of crashing.
- **`get_profile_workunit`** emits a column-only profile (`columnCount` set, `rowCount` omitted) when the record count is unavailable, so profiling still produces useful metadata. `rowCount` is already optional on `DatasetProfileClass`.
- **Per-object guard**: each object's workunit generation is wrapped so any single object's unexpected failure is reported as a warning and skipped, rather than aborting the whole pipeline.

## Testing

- Added `test_profiling_with_empty_record_count`, which mocks an empty `recordCount` response and asserts the run completes, a profile is still emitted with `rowCount=None` / `columnCount` set, and a warning is recorded.
- Existing salesforce golden-file tests are unaffected and still pass.
- `./gradlew :metadata-ingestion:lintFix` and `mypy` are clean.

## Checklist

- [x] Tests added/updated
- [ ] Docs added/updated — n/a (bug fix, no user-facing behavior change)


metadata-ingestion/tests/integration/salesforce/test_salesforce.py
```diff
@@ -264,3 +274,44 @@ def test_salesforce_ingest_with_lineage(pytestconfig, tmp_path):
...
+    mocked_call = mock.Mock()
+    mocked_call.side_effect = side_effect_call_salesforce_empty_record_count
+    mock_sf._call_salesforce = mocked_call
+    mock_sdk.return_value = mock_sf
+
+    config = SalesforceConfig.model_validate(
+        {
+            "auth": "DIRECT_ACCESS_TOKEN",
+            "instance_url": "https://mydomain.my.salesforce.com/",
+            "access_token": "access_token`",
```
sgomezvillamor: is the backtick at the end on purpose?

```
access_token`
```

sgomezvillamor (approved):
LGTM

PR #17927, opened 2026-06-16
feat(ingest/matillion): OpenLineage-driven lineage, foldered containers, run history, and unpublished-pipeline fixes

acrylJonny:
## Summary

Fixes two ingestion bugs in the Matillion DPC connector, reworks how lineage is produced so it no longer depends on recent pipeline executions, and improves how pipelines/components are organized and how run history is surfaced.

### Bugs fixed
- **`include_unpublished_pipelines: false` emitted no pipelines.** Pipeline/lineage discovery was driven off pipeline executions; with unpublished pipelines disabled the connector fetched pipelines but emitted none. Discovery is now bounded by the configured `start_time`/`end_time` window and decoupled from executions.
- **`include_unpublished_pipelines: true` could hang.** The executions/lineage scan was effectively unbounded. It is now bounded by the time window and resilient to slow endpoints (see below).

### Lineage rework
- Lineage (DataFlow/DataJob plus dataset and column-level) is now emitted **directly from OpenLineage job events**, decoupled from pipeline executions, so lineage is produced even for pipelines that have not run recently.
- **Lineage components are reconciled to their discovered pipeline.** The OpenLineage `job.name` already carries the full pipeline path (folders included), identical to the discovered published-pipeline path. Lineage reconciles on `(project_id, full_path)` and reuses the discovered pipeline's DataFlow URN, so components (DataJobs) nest under the correct foldered pipeline instead of floating as separate bare-named flows. The environment in the lineage namespace is an opaque UUID that no API maps back to the environment name, so it is dropped from the match key; reconciliation keys purely on the path, which both code paths report identically.
- **`pipeline_patterns` is applied to the lineage job's full path directly**, identically to discovery — removing the earlier resolve-or-allow fallback (and its filtering leak where unresolvable jobs were waved through).
- Dataset lineage to external platforms (Snowflake, Salesforce, …) is emitted with **`is_primary_source=False`**, so Matillion records the lineage edge without materializing a `status` aspect or claiming ownership of datasets it only references — eliminating "ghost" entities for externally-owned datasets.
- **SQL parsing re-wired as a gap-fill:** SQL from OpenLineage job facets is handed to the `SqlParsingAggregator` (`add_observed_query`) only for outputs that lack OpenLineage column lineage and already exist in DataHub. This enriches without clobbering richer OpenLineage lineage and without creating stub upstreams.

### Container hierarchy
- Pipelines and components are now organized into a browsable container tree that mirrors their path in Matillion (`Project › Environment › <folder> … › Pipeline › Component`), instead of all pipelines sitting flat under the environment. Folder levels are derived from the pipeline path, so the browse tree lines up with the paths used in `pipeline_patterns`. Components (DataJobs) carry a `container` aspect and browse directly under their pipeline.
- Environment and folder containers are **always** built; `extract_projects_to_containers` (default `true`) now only controls whether they nest under a top-level **Project** container or hang at the root. Browse paths no longer include the entity itself as a trailing node.
- Lineage jobs whose environment cannot be resolved (not published and not seen in execution history) are now **skipped** rather than emitted as flows floating under the bare project. Enabling `extract_run_history` (or `include_unpublished_pipelines`) helps resolve environments for unpublished child orchestrations. A `lineage_jobs_skipped_no_environment` report counter records skips.

### Run history
- Per-execution `DataProcessInstance` entities are now emitted for the pipeline (**DataFlow**) as well as each component (**DataJob**), populating the "Runs" tab at both levels.
- **New `extract_run_history` flag** (default `false`): emit run history for published pipelines without also discovering unpublished ones (implied when `include_unpublished_pipelines` is enabled).

### New configuration
- **`include_dependent_pipelines`** (default `true`, preserves prior behavior): when disabled, child pipelines that appear only in lineage events (and were not discovered as project pipelines) are no longer created as their own DataFlows/DataJobs; lineage for discovered pipelines is unaffected. A `lineage_dependent_pipelines_skipped` counter records skips.

### External links
- Entity external URLs point at the Data Productivity Cloud (Maia) console on the fixed, global host `https://app.matillion.com`. The console is a single host with the account selected at login (not encoded in the URL), so links are fully derivable and require no configuration.
- Fixed two broken links: the pipeline observability link dropped an invalid `timeFrame=*` filter (the dashboard only accepts tokens such as `24h`/`7d`/`30d`/`custom`) and now URL-encodes the search value; streaming pipelines, which have no dedicated console deep-link, no longer emit a link that was built from the API host and a non-existent route.

### Reliability & API handling
- Lineage is fetched in ≤31-day windows (Matillion API limit).
- `Timeout`/`ConnectionError` on the slow pipeline-executions and lineage endpoints are caught with actionable report warnings; the run degrades gracefully instead of failing.
- Docs updated with the lineage API's offset-pagination degradation and timeout-tuning guidance.

### Cleanup
- Internal lineage state moved to Pydantic models (`MatillionJobNamespace`, `OutputLineageAccumulator`, and a frozen `SqlAggregatorKey`); the SQL aggregator cache and schema preloading are keyed by `(platform, platform_instance, env)` to prevent cross-instance collisions.
- Removed dead lineage code, the unused `PipelineLineageResult` model, and the base-name resolution helper now obsoleted by full-path reconciliation.
- Genericized test/doc fixtures to remove customer-identifying data.

## Checklist
- [x] PR conforms to the [PR Title Format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#pr-title-format)
- [x] Tests added/updated
- [x] Docs added/updated
- [x] No breaking changes


max-datahub (commented):
Thanks for the thorough rework — the OpenLineage-driven lineage, foldered containers, and run-history changes are well-structured, the changelog is detailed, and the reconciliation/container/gap-fill logic is well tested. A few things I'd like to sort out before this merges. Not approving yet, but most of this is small.

## 🔴 Must fix before merge

**`AGENTS.md` — looks like personal agent memory leaked into the diff.** The change appends `## Learned User Preferences` / `## Learned Workspace Facts` sections (references to local `~/.datahubenv` workflow, nuking the local quickstart, "AI slop cleanup" habits, etc.). This is unrelated to Matillion and reads like local assistant state that got committed by accident — can we drop it from this PR entirely?

## ❓ Please confirm — multi-environment lineage attribution

`_discovered_env_by_path` is keyed on `(project_id, full_path)` with no environment, and `events_by_job` buckets OpenLineage events on `(project_id, job_name)` — the parsed `MatillionJobNamespace.environment_id` is discarded (`matillion.py` ~L792). So if the **same pipeline path exists in more than one environment**, last-write-wins picks a single env in `_discovered_env_by_path`, and *all* lineage for that path attaches to that one environment's DataFlow (`matillion.py` ~L862/L888); the other environment's discovered flow gets no lineage/DataJobs.

I understand the env UUID can't be mapped back to an environment *name* — but it can still be used to keep dev/prod event buckets separate. The key question: in Matillion DPC, are published pipelines effectively project-scoped (same file visible under every environment) or genuinely distinct per environment? If the former, this mis-attributes lineage routinely rather than in a rare edge case. Existing tests only exercise a single environment, so this path is uncovered. Could you confirm the intended behavior (and add a two-environment fixture if it's meant to be handled)?

## 🟡 Test coverage gaps (nice to close before merge)

The tests that exist are good — differential reconciliation tests and the new `test_matillion_container.py` are excellent. But a few claimed behaviors have no coverage:

- **`is_primary_source=False`** (`matillion.py` L957, L1661) — zero tests, and because it's a sink-level `MetadataWorkUnit` attribute (not in the serialized MCE JSON) the golden-file tests can't catch a regression here. A single `assert wu.is_primary_source is False` on the upstream-lineage workunit would protect it.
- **`_fetch_lineage_events` timeout path** (`lineage-events-timeout` warning) — untested, while the sibling `pipeline-executions-timeout` path is.
- **Multi-page pagination within a lineage window** — `test_lineage_events_fetched_in_max_31_day_windows` always returns an empty page, so the inner page loop never advances past page 0.
- **`SqlAggregatorKey` env isolation** — the tests were mechanically updated to the new key shape but don't assert that two environments get distinct cached aggregators (the collision this key change fixes).

## 🔵 Non-blocking follow-ups

- **Dead code (pre-existing, but this PR leans on it more):** `redundant_run_skip_handler` is read via `getattr(self, "redundant_run_skip_handler", None)` in `_get_lineage_time_window` / `_update_lineage_state` / `_fetch_lineage_events`, but it's never instantiated anywhere in the class hierarchy — so the stateful lineage-window skip path is always inert even though `stateful_ingestion` config implies incremental lineage works. Worth either wiring it (à la redshift/bigquery lineage) or removing the dead branches.
- **P3:** DataFlow/DPI URNs differ between `include_unpublished_pipelines` on/off (base-name key vs full-path key) for the same pipeline, so flipping that flag can orphan previously-ingested entities. Not a single-run bug, but worth a note.
- **Behavior flips worth a line in the docs/changelog:** `include_external_urls` now defaults `false` (existing console links disappear on re-ingest), and `extract_projects_to_containers=false` now still emits environment/folder containers. Both reasonable — just user-visible.

## ✅ Verified correct

For what it's worth, I dug into several of the riskier claims and they hold up: the "no ghost entities" behavior (all aggregator + external-dataset lineage emitted `is_primary_source=False`), the gap-fill existence gate (`output_urn in self._registered_urns`, populated only from real schemas), no `SqlAggregatorKey` None-env collision, no double-fetch of executions at default config, and graceful (surfaced) timeout degradation.

Happy to re-review quickly once the `AGENTS.md` section is removed and the multi-env question is settled.

max-datahub (approved):
Thanks for the thorough follow-ups — everything from the earlier review is addressed:

- The accidental `AGENTS.md` personal-memory additions are reverted.
- Multi-environment lineage now nests a flow/component under **every** discovered environment for a project-scoped path (with env-independent dataset lineage emitted once), instead of collapsing onto a last-write-wins env. `test_lineage_emitted_under_each_environment_for_shared_path` covers it cleanly.
- New coverage added for the gaps: `is_primary_source=False` on the upstream workunit, the lineage-events timeout path, in-window pagination (page 0→1), and per-environment SQL-aggregator isolation.
- Dead code removed (`redundant_run_skip_handler`, `MatillionUrnBuilder`, `_fetch_steps_for_pipeline`), bare `except Exception` narrowed to specific request exceptions, and the behavior flips (`include_external_urls` default, always-on env/folder containers, multi-env lineage) plus the `include_unpublished_pipelines` URN caveat are documented in the changelog.

Code and tests read well and CI is green. LGTM 🚀

PR #17934, opened 2026-06-17
fix(ingestion): trim oversized schemaMetadata by dropping platform schema before fields

puneetagarwal-datahub:
## Summary

Follow-up to #17895, addressing the review comments there. Two changes to `EnsureAspectSizeProcessor.ensure_schema_metadata_size` / `_drop_platform_schema`:

### 1. Drop the platform schema *before* trimming fields (priority fix)

Previously the platform schema was dropped only when it **alone** exceeded the budget; otherwise it was kept and fields were trimmed to make room. In the mixed case — raw schema and fields each under the limit but over together — that kept the raw schema and dropped columns, backwards from the "fields are more valuable than the raw platform schema" priority.

Now: if the whole aspect fits, do nothing (fast path); otherwise drop the platform schema first and re-check, trimming fields only if they still don't fit. So a mixed aspect keeps **all** columns and sheds the raw schema instead.

| | raw schema 10 MB + fields 10 MB |
|---|---|
| Before | keep raw schema, drop ~4.5 MB of columns |
| After | drop raw schema, keep all columns |

### 2. Blank the platform schema per-variant via an explicit field map

`_drop_platform_schema` iterated `platformSchema._inner_dict` and blanked every string member. That reaches into a semi-internal attribute and would also wipe small type markers (e.g. `KafkaSchema.documentSchemaType="AVRO"`), not just the raw blob.

Replaced with an explicit map of each `platformSchema` union variant to its schema-content string field(s) (`rawSchema` / `documentSchema` / `tableSchema` / `keySchema` / `valueSchema`), blanking only those via the public accessors. Markers are left intact, and an unrecognized variant is logged and skipped.

Also removed two transitional ("now"/"previous") comments flagged in review.

## Tests

- New mixed-case test: raw schema + fields each under budget but over together → raw schema dropped, **all** fields retained.
- The `KafkaSchema` drop test now sets `documentSchemaType="AVRO"` and asserts it **survives** while the blob is blanked.
- 37 tests pass; mypy + ruff clean.

## Checklist

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
+++ b/metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
@@ -148,30 +148,30 @@ class EnsureAspectSizeProcessor(WorkunitProcessor[EnsureAspectSizeProcessorRepor
     def ensure_schema_metadata_size(
         self, dataset_urn: str, schema: SchemaMetadataClass
     ) -> None:
-        # Budget against the full serialized aspect, not just the fields list:
-        # the platformSchema blob (and other top-level content) is emitted too.
-        truncated = False
-        non_fields_size = self._schema_size_without_fields(schema)
-
-        # If the non-field content alone exceeds the budget, the platformSchema
-        # blob is the culprit and the least valuable content. Drop it BEFORE
-        # trimming fields (which carry the metadata users query), then re-measure.
-        # Otherwise the field budget would start already-oversized and reject
-        # every field while the aspect stays over the limit.
+        # Fast path: if the whole aspect already fits, there's nothing to trim.
         if (
-            non_fields_size >= self.schema_size_constraint
-            and self._drop_platform_schema(schema)
+            len(json.dumps(pre_json_transform(schema.to_obj())))
+            < self.schema_size_constraint
         ):
+            return
+
+        # Over budget. Shed the least-valuable content first — the platformSchema
+        # blob (an opaque source-format dump) — so we keep as many fields (the
+        # structured, queryable metadata) as possible. Only after that do we trim
+        # fields, and only if they still don't fit. This keeps "fields > raw
+        # platform schema" even when neither part is individually over the limit
+        # (e.g. a ~half-schema / ~half-fields aspect).
+        truncated = False
+        if self._drop_platform_schema(schema):
             self.ctx.source_report.warning(
                 title="Schema truncated due to size constraint",
                 message="Dataset schema contained too much data and would have caused ingestion to fail",
                 context=f"Raw platform schema was removed from schema for {dataset_urn} due to aspect size constraints",
             )
             self.report.num_platform_schema_drops += 1
-            non_fields_size = self._schema_size_without_fields(schema)
             truncated = True
 
-        total_fields_size = non_fields_size
+        total_fields_size = self._schema_size_without_fields(schema)
         accepted_fields: List[SchemaFieldClass] = []
         for schema_field in schema.fields:
             field_size = len(json.dumps(pre_json_transform(schema_field.to_obj())))
--- a/metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
+++ b/metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
@@ -235,13 +235,17 @@ class EnsureAspectSizeProcessor(WorkunitProcessor[EnsureAspectSizeProcessorRepor
             if total_fields_size + field_size < self.schema_size_constraint:
                 accepted_fields.append(schema_field)
                 total_fields_size += field_size
-            else:
-                self.ctx.source_report.warning(
-                    title="Schema truncated due to size constraint",
-                    message="Dataset schema contained too much data and would have caused ingestion to fail",
-                    context=f"Field {schema_field.fieldPath} was removed from schema for {dataset_urn} due to aspect size constraints",
-                )
-        if truncated or len(accepted_fields) < len(schema.fields):
+
+        dropped_field_count = len(schema.fields) - len(accepted_fields)
+        if dropped_field_count:
+            # Report a single warning per entity with the dropped count, rather
+            # than one warning per dropped field.
+            self.ctx.source_report.warning(
+                title="Schema truncated due to size constraint",
+                message="Dataset schema contained too much data and would have caused ingestion to fail; some fields were dropped",
+                context=f"Dropped {dropped_field_count} of {len(schema.fields)} fields from schema for {dataset_urn} due to aspect size constraints",
+            )
+        if truncated or dropped_field_count:
             self._record_truncation("schemaMetadata")
         schema.fields = accepted_fields
 
--- a/metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
+++ b/metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
@@ -2,7 +2,7 @@ import json
 import logging
 import os
 from dataclasses import dataclass, field
-from typing import Dict, Iterable, List, Optional
+from typing import Dict, Iterable, List, Optional, Tuple
 
 from datahub.emitter.rest_emitter import INGEST_MAX_PAYLOAD_BYTES
 from datahub.emitter.serialization_helper import pre_json_transform
@@ -13,10 +13,20 @@ from datahub.ingestion.api.workunit_processor import (
     WorkunitProcessorReport,
 )
 from datahub.metadata.schema_classes import (
+    BinaryJsonSchemaClass,
     DatasetProfileClass,
+    EspressoSchemaClass,
+    KafkaSchemaClass,
+    KeyValueSchemaClass,
+    MySqlDDLClass,
+    OracleDDLClass,
+    OrcSchemaClass,
+    OtherSchemaClass,
+    PrestoDDLClass,
     QueryPropertiesClass,
     QuerySubjectsClass,
     SchemaFieldClass,
+    SchemalessClass,
     SchemaMetadataClass,
     UpstreamLineageClass,
     ViewPropertiesClass,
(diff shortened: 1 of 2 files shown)


sgomezvillamor (approved):
The strategy is correctly implemented and the core behavioral change is well tested. A few things worth addressing:

---

**⚠️ Warning: Missing cascade test (platform schema drop + fields still need trimming)**

The end-to-end path where the platform schema is dropped *and* fields still don't fit is only covered implicitly by `test_ensure_size_of_too_big_schema_metadata`, which uses `rawSchema="aaa"` — a trivially small value. That test never asserts `rawSchema == ""` or `num_platform_schema_drops == 1`, so a regression that swapped the order (fields trimmed first, then schema dropped) would still pass.

A dedicated test is needed: meaningful platform schema (~40% budget) + 1000 large fields, asserting both `rawSchema == ""` and `num_platform_schema_drops == 1` and `len(fields) < 1000` and final size within budget.

---

**ℹ️ Suggestion: `_PLATFORM_SCHEMA_RAW_FIELDS` field names are unverifiable at type-check time**

`setattr`/`getattr` bypass mypy, so a mistyped field name silently does nothing (`getattr` returns `None`, the field is skipped, `cleared` stays `False`). The KafkaSchema test provides runtime coverage for one variant; the others (EspressoSchema, KeyValueSchema, OrcSchema, etc.) have no test exercising the new explicit-map path. A lightweight smoke test per variant (or a parametrized test) would catch a mistyped field name.

---

**ℹ️ Suggestion: Two pre-existing test names don't describe the new behaviour**

- `test_ensure_size_of_too_big_schema_metadata` → `test_schema_metadata_trims_fields_when_fields_are_too_large`
- `test_ensure_size_of_proper_schema_metadata` → `test_schema_metadata_no_op_when_fits_within_budget`

Neither was introduced by this PR, so not a blocker, but renaming while touching the file would make the intent obvious.

---

puneetagarwal-datahub:
> The strategy is correctly implemented and the core behavioral change is well tested. A few things worth addressing:
> 
> **⚠️ Warning: Missing cascade test (platform schema drop + fields still need trimming)**
> 
> The end-to-end path where the platform schema is dropped _and_ fields still don't fit is only covered implicitly by `test_ensure_size_of_too_big_schema_metadata`, which uses `rawSchema="aaa"` — a trivially small value. That test never asserts `rawSchema == ""` or `num_platform_schema_drops == 1`, so a regression that swapped the order (fields trimmed first, then schema dropped) would still pass.
> 
> A dedicated test is needed: meaningful platform schema (~40% budget) + 1000 large fields, asserting both `rawSchema == ""` and `num_platform_schema_drops == 1` and `len(fields) < 1000` and final size within budget.
> 
> **ℹ️ Suggestion: `_PLATFORM_SCHEMA_RAW_FIELDS` field names are unverifiable at type-check time**
> 
> `setattr`/`getattr` bypass mypy, so a mistyped field name silently does nothing (`getattr` returns `None`, the field is skipped, `cleared` stays `False`). The KafkaSchema test provides runtime coverage for one variant; the others (EspressoSchema, KeyValueSchema, OrcSchema, etc.) have no test exercising the new explicit-map path. A lightweight smoke test per variant (or a parametrized test) would catch a mistyped field name.
> 
> **ℹ️ Suggestion: Two pre-existing test names don't describe the new behaviour**
> 
> * `test_ensure_size_of_too_big_schema_metadata` → `test_schema_metadata_trims_fields_when_fields_are_too_large`
> * `test_ensure_size_of_proper_schema_metadata` → `test_schema_metadata_no_op_when_fits_within_budget`
> 
> Neither was introduced by this PR, so not a blocker, but renaming while touching the file would make the intent obvious.
> 

Taken cared of

metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
```diff
@@ -127,51 +155,58 @@ def _schema_size_without_fields(self, schema: SchemaMetadataClass) -> int:
...
+        (``_PLATFORM_SCHEMA_RAW_FIELDS``), leaving small markers such as
+        ``KafkaSchema.documentSchemaType`` intact. An unrecognized variant is
+        logged and left untouched. Returns ``True`` if any content was cleared.
         """
         platform_schema = schema.platformSchema
         if platform_schema is None:
             return False
+        raw_fields = _PLATFORM_SCHEMA_RAW_FIELDS.get(type(platform_schema))
+        if raw_fields is None:
+            logger.info(
```
skrydal: Let's bump it to a warning

metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
```diff
@@ -33,6 +43,24 @@
...
+    PrestoDDLClass: ("rawSchema",),
+    MySqlDDLClass: ("tableSchema",),
+    OracleDDLClass: ("tableSchema",),
+    OrcSchemaClass: ("schema",),
+    BinaryJsonSchemaClass: ("schema",),
+    KafkaSchemaClass: ("documentSchema", "keySchema"),
+    EspressoSchemaClass: ("documentSchema", "tableSchema"),
+    KeyValueSchemaClass: ("keySchema", "valueSchema"),
+    SchemalessClass: (),
+}
```
skrydal: I fully understand the sentiment to follow DRY rule. I propose a type-safe approach here. Consider:
```python
  def _drop_platform_schema(self, schema: SchemaMetadataClass) -> bool:
      ps = schema.platformSchema
      if ps is None or isinstance(ps, SchemalessClass):
          return False
      if isinstance(ps, (OtherSchemaClass, PrestoDDLClass)):
          return _blank_raw_schema(ps)
      if isinstance(ps, KafkaSchemaClass):
          return _blank_kafka_schema(ps)
      # the rest here ...
      logger.warning(
          "Unrecognized platformSchema variant %s; raw blob retained",
          type(ps).__name__,
      )
      return False


  def _blank_raw_schema(ps: Union[OtherSchemaClass, PrestoDDLClass]) -> bool:
      if ps.rawSchema:
          ps.rawSchema = ""
          return True
      return False


  def _blank_kafka_schema(ps: KafkaSchemaClass) -> bool:
      cleared = False
      if ps.documentSchema:
          ps.documentSchema = ""
          cleared = True
      if ps.keySchema:
          ps.keySchema = ""
          cleared = True
      return cleared
```
This gives us type safety, at a small repetition cost. Also, if model changes we will have an error coming straight from mypy.

skrydal (changes requested):
Thank you for the change, it makes the previous approach much better.
Please consider my 2 comments, as well as, adding a test which would go through all the schemas in the model and confirm we cover all the schemas:

```python
  def test_drop_platform_schema_handles_every_union_variant(processor):
      # Pull the live union members of SchemaMetadata.platformSchema from the
      # generated model. When a new variant is added to the schema, this test
      # fails with a clear "add a constructor and a branch in
      # _drop_platform_schema" message — preventing the silent
      # "Unrecognized variant" log line + retained oversized blob in production.
      #
      # The constructor map is in the test (not in src/) on purpose: a new
      # variant should force both a dispatcher branch AND a behavioral check
      # that the new branch actually clears something.
      constructors: Dict[str, Callable[[], DictWrapper]] = {
          "OtherSchema":      lambda: OtherSchemaClass(rawSchema="x" * 100),
          "PrestoDDL":        lambda: PrestoDDLClass(rawSchema="x" * 100),
          "MySqlDDL":         lambda: MySqlDDLClass(tableSchema="x" * 100),
          "OracleDDL":        lambda: OracleDDLClass(tableSchema="x" * 100),
          "OrcSchema":        lambda: OrcSchemaClass(schema="x" * 100),
          "BinaryJsonSchema": lambda: BinaryJsonSchemaClass(schema="x" * 100),
          "KafkaSchema":      lambda: KafkaSchemaClass(documentSchema="x" * 100),
          "EspressoSchema":   lambda: EspressoSchemaClass(
                                  documentSchema="x" * 100, tableSchema="x" * 100),
          "KeyValueSchema":   lambda: KeyValueSchemaClass(
                                  keySchema="x" * 100, valueSchema="x" * 100),
          "Schemaless":       lambda: SchemalessClass(),
      }

      model_union_members = {
          s.name
          for s in SchemaMetadataClass.RECORD_SCHEMA
              .fields_dict["platformSchema"].type.schemas
      }
      test_covers = set(constructors.keys())

      missing_in_test = model_union_members - test_covers
      extra_in_test = test_covers - model_union_members
      assert not missing_in_test, (
          f"New platformSchema variant(s) in the model are not exercised here: "
          f"{missing_in_test}. Add a constructor above AND a branch in "
          f"_drop_platform_schema."
      )
      assert not extra_in_test, (
          f"Test references variants no longer in the model: {extra_in_test}"
      )

      for name, make in constructors.items():
          schema = SchemaMetadataClass(
              schemaName="t", version=1, platform="p", hash="h",
              platformSchema=make(), fields=[],
          )
          cleared = processor._drop_platform_schema(schema)

          if name == "Schemaless":
              assert not cleared, "Schemaless carries no content to clear"
          else:
              assert cleared, (
                  f"_drop_platform_schema returned False for {name}; "
                  f"the dispatcher likely has no branch for it"
              )
```

metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
puneetagarwal-datahub: Done

skrydal:
On a separate note - I think we could improve how we are reporting the fact of trimming of the fields. We should report one warning per entity, with the count of columns dropped.

puneetagarwal-datahub:
> Thank you for the change, it makes the previous approach much better. Please consider my 2 comments, as well as, adding a test which would go through all the schemas in the model and confirm we cover all the schemas:
> 
> ```python
>   def test_drop_platform_schema_handles_every_union_variant(processor):
>       # Pull the live union members of SchemaMetadata.platformSchema from the
>       # generated model. When a new variant is added to the schema, this test
>       # fails with a clear "add a constructor and a branch in
>       # _drop_platform_schema" message — preventing the silent
>       # "Unrecognized variant" log line + retained oversized blob in production.
>       #
>       # The constructor map is in the test (not in src/) on purpose: a new
>       # variant should force both a dispatcher branch AND a behavioral check
>       # that the new branch actually clears something.
>       constructors: Dict[str, Callable[[], DictWrapper]] = {
>           "OtherSchema":      lambda: OtherSchemaClass(rawSchema="x" * 100),
>           "PrestoDDL":        lambda: PrestoDDLClass(rawSchema="x" * 100),
>           "MySqlDDL":         lambda: MySqlDDLClass(tableSchema="x" * 100),
>           "OracleDDL":        lambda: OracleDDLClass(tableSchema="x" * 100),
>           "OrcSchema":        lambda: OrcSchemaClass(schema="x" * 100),
>           "BinaryJsonSchema": lambda: BinaryJsonSchemaClass(schema="x" * 100),
>           "KafkaSchema":      lambda: KafkaSchemaClass(documentSchema="x" * 100),
>           "EspressoSchema":   lambda: EspressoSchemaClass(
>                                   documentSchema="x" * 100, tableSchema="x" * 100),
>           "KeyValueSchema":   lambda: KeyValueSchemaClass(
>                                   keySchema="x" * 100, valueSchema="x" * 100),
>           "Schemaless":       lambda: SchemalessClass(),
>       }
> 
>       model_union_members = {
>           s.name
>           for s in SchemaMetadataClass.RECORD_SCHEMA
>               .fields_dict["platformSchema"].type.schemas
>       }
>       test_covers = set(constructors.keys())
> 
>       missing_in_test = model_union_members - test_covers
>       extra_in_test = test_covers - model_union_members
>       assert not missing_in_test, (
>           f"New platformSchema variant(s) in the model are not exercised here: "
>           f"{missing_in_test}. Add a constructor above AND a branch in "
>           f"_drop_platform_schema."
>       )
>       assert not extra_in_test, (
>           f"Test references variants no longer in the model: {extra_in_test}"
>       )
> 
>       for name, make in constructors.items():
>           schema = SchemaMetadataClass(
>               schemaName="t", version=1, platform="p", hash="h",
>               platformSchema=make(), fields=[],
>           )
>           cleared = processor._drop_platform_schema(schema)
> 
>           if name == "Schemaless":
>               assert not cleared, "Schemaless carries no content to clear"
>           else:
>               assert cleared, (
>                   f"_drop_platform_schema returned False for {name}; "
>                   f"the dispatcher likely has no branch for it"
>               )
> ```

Implemented

puneetagarwal-datahub:
> On a separate note - I think we could improve how we are reporting the fact of trimming of the fields. We should report one warning per entity, with the count of columns dropped.

Yes , Implemented

skrydal (approved):
Awesome

PR #17939, opened 2026-06-17
feat(ingestion): reusable OAuth token providers for DataHub Python clients

max-datahub:
## Summary

Adds a small, reusable **OAuth token-provider layer** to the Python SDK so DataHub clients (CLI, SDK, ingestion, and the long-running Remote Executor) can authenticate to GMS with a **fresh, short-lived OAuth JWT per request** instead of a single static token frozen into the HTTP session at construction.

GMS already validates externally-issued OAuth JWTs as a resource server, so **no GMS change is required** — the gap was purely client-side. This is the machine-to-machine (RFC 6749 §4.4 client-credentials, RFC 7523 JWT-assertion) counterpart to the interactive `datahub init --oauth` PKCE flow added in #17868; the two coexist (see below).

> Supersedes #17759 (which was opened from a personal fork). This PR carries the same work on a branch in this repository, rebased onto current `master` (so it builds on #17868), plus the review refinement below.

## What's included

- `datahub.emitter.token_provider`: a `TokenProvider` interface, `TokenResult` (token + reported expiry), `StaticTokenProvider` (back-compat), `CachingTokenProvider` (refresh-before-expiry, thread-safe), and `TokenProviderAuth` — a `requests.auth.AuthBase` that re-evaluates the token **per request** and retries once on a 401.
- `DataHubRestEmitter` accepts an injectable `auth`; the existing static `token=` path is unchanged (backward compatible).
- Pluggable providers via a `datahub.token_provider.plugins` entry-point group:
  - `k8s_oidc` — Kubernetes projected service-account token (cluster OIDC; stdlib only)
  - `azure_entra` — Entra token via workload identity or client secret (new `azure-auth` extra)
  - `oidc_client_credentials` — generic OIDC client-credentials grant (Okta, Keycloak, any OIDC)
- Declarative `auth: {type, config}` on `DatahubClientConfig` (mutually exclusive with `token`), resolved in `DataHubGraph`.
- Unit tests for all of the above + a local Keycloak-based OAuth smoke test under `smoke-test/tests/oauth/`.

## Token expiry is taken from the provider, not decoded from the token

OAuth2 access tokens are opaque to the client, so the SDK **never decodes the JWT** to learn when to refresh. Each provider reports expiry from its native, standards-defined source:

| Provider | Expiry source |
| --- | --- |
| `oidc_client_credentials` | `expires_in` from the token response (RFC 6749 §5.1) |
| `azure_entra` | `expires_on` from `azure-identity`'s `AccessToken` |
| `k8s_projected` | none reported — the file is re-read each call (cheap; kubelet rotates in place) |

`CachingTokenProvider` refreshes ~300s before the reported expiry; when no expiry is reported it re-fetches per call.

## Relationship to #17868 (interactive PKCE login)

#17868 added `datahub init --oauth` (authorization-code + PKCE, with refresh-token auto-refresh stored in `~/.datahubenv`). That path is for a **human** logging into DataHub Cloud; this PR is for a **machine identity**. They coexist cleanly: `refresh_oauth_token_if_needed()` is a no-op unless a stored `oauth:` section exists, so a machine-flow `auth:` config flows through `load_client_config()` untouched. A regression test guards this.

## How it works

A client's job is to obtain a JWT from an issuer GMS trusts and present it. This adds a pluggable way to acquire that token (per cloud / IdP / grant) with per-request freshness + refresh, behind one declarative config. Adding a new provider is a small class implementing `get_token()` plus an entry point — no SDK or consumer change.

## Testing

- Unit tests under `metadata-ingestion/tests/unit/auth/` (interface, caching/refresh, `AuthBase` + 401 retry, emitter wiring, registry resolution, all three providers, declarative config) and a `load_client_config` coexistence test.
- A docker-compose + Keycloak smoke test (`smoke-test/tests/oauth/`) validating that a real client-credentials token is **accepted** (correct audience) and **rejected** (wrong audience) by GMS.
- `./gradlew :metadata-ingestion:lint` (ruff + mypy) and `:metadata-ingestion:checkLockFile` pass.

## References

- RFC 6749 §4.4 (Client Credentials Grant), §5.1 (token response fields)
- RFC 7523 (JWT profile for client authentication / authorization grants) — Azure Workload Identity
- RFC 9068 (JWT Profile for OAuth 2.0 Access Tokens)

## Checklist

- [x] Conventional Commit title
- [x] Tests added (unit + smoke)
- [ ] Docs — follow-up
- [x] Backward compatible (static `token=` unchanged)


alexsku:
## Review — OAuth token providers

> Initial review pass. **Nick Adams will do the final review.**

**Type:** additive, client-side SDK feature · **Risk:** Low · **Blast radius:** Python SDK clients only — `token=` path untouched, new behavior opt-in via `auth:`, no GMS/server change · **Rollback:** safe (purely additive).

Clean, well-tested, low-risk. Notably it adopts the **"expiry from the provider, never decode the JWT"** design (`TokenResult.expires_at` sourced from `expires_in` / `azure-identity`'s `expires_on`), which was the main design concern on the spec. No blockers.

### Other issues (all low severity)

- **[LOW / correctness]** `cli/config_utils.py:153-155` — `load_client_config()` does `datahub_config.model_copy(update={"token": refreshed_token})`. Pydantic v2 `model_copy` does **not** re-run validators, so if a user ever has *both* an `auth:` block and a stored interactive `oauth:` session, this silently sets `token` alongside `auth`, bypassing the `_validate_auth_exclusive` invariant. Benign in effect (the emitter prefers `auth`, so the stale interactive token is just ignored), but it's a silent invariant violation, and the new coexistence test only covers the *no stored oauth* case. Fix: re-validate after the copy, or only apply the token update when `auth is None`.

- **[LOW / operability]** No env-var path supplies `auth:` yet. `load_client_config()` returns early on `DATAHUB_GMS_URL` (`:133-135`) reading only `server`+`token`, so a file-based `auth:` block is ignored whenever that env var is set. Matches the spec (the `DATAHUB_AUTH_TYPE` shorthand was deferred to the executor follow-up), but worth a doc note so executor deployments don't set `DATAHUB_GMS_URL` and silently lose `auth:`.

- **[LOW / dx]** `CachingTokenProvider.get_token()` (`token_provider.py:88-93`) holds the lock across `self._fetch()`, so during a refresh all concurrent callers block — including ones that would otherwise be cache hits. This is the intended single-flight behavior and is correct; a one-line comment ("cache reads intentionally block during refresh") would stop a future well-meaning optimization from breaking it.

- **[LOW / note]** `datahub init` → `write_gms_config()` (`:182`) shallow-merges and wholesale-replaces the `gms` block, so a hand-authored `auth:` block is wiped on re-init. This is **pre-existing** shallow-merge behavior (not introduced here), but `auth:` is now subject to it — worth a docs caveat that machine-auth config should be managed out-of-band.

### What's missing
- **Docs** — acknowledged follow-up (operator guide); gating item for adoption, fine to land after.
- The 401-retry-via-`response.connection.send` path is exercised only through `requests_mock` (mirrors the canonical `HTTPDigestAuth.handle_401` pattern, so correctness isn't in doubt) — confirming it's been validated against a real GMS via the smoke test closes the loop.

### Test coverage
Genuinely good and behavior-focused: back-compat, mutual exclusivity (constructor path), caching/refresh/unknown-expiry/invalidate, 401 retry (once / no-loop / non-401 passthrough / no-invalidate→no-retry), lazy optional-dep import, registry resolution, per-provider expiry, plus the Keycloak right/wrong-audience smoke test.

### Questions
1. In `load_client_config`, should the `auth:` + stored-`oauth:` combination be explicitly rejected (or `auth` win deterministically with re-validation), rather than relying on the emitter's precedence?
2. Is the executor/Helm env wiring (`DATAHUB_AUTH_TYPE`) intentionally out of this PR (the spec's Slice-1 follow-up)?
3. Has the Keycloak smoke test been run end-to-end against a live GMS locally, or does it first execute in CI here?

**Bottom line:** No blockers; the low items are worth a quick pass but shouldn't hold merge once CI is green. Final review to come from **Nick Adams**.

nwadams (approved):

PR #17945, opened 2026-06-17
fix(ingest/datahub-documents): seed chunking checkpoint from last committed state

alexjst:
## Summary

`DocumentChunkingStateHandler.create_checkpoint()` built each run's checkpoint from a brand-new empty `DocumentChunkingCheckpointState()`. Since `update_document_state()` only records documents processed during the current run, the committed checkpoint **overwrote** the previous state instead of **accumulating** it.

Effect on batch (content-hash) mode with stateful ingestion enabled:

- Run N processes set P, commits a checkpoint containing only P.
- Run N+1 loads {P}, skips those, re-embeds the complement, commits only the complement.
- Run N+2 re-embeds P again.

With static content the skip set never converges — documents ping-pong between runs and keep getting re-embedded.

## Fix

Seed the new checkpoint from the last committed state, carrying forward `document_state` and `event_offsets` (`deepcopy` to isolate the prior state object from in-run mutation):

```python
state = DocumentChunkingCheckpointState()
last_state = self.get_last_state()
if last_state:
    state = DocumentChunkingCheckpointState(
        document_state=copy.deepcopy(last_state.document_state),
        event_offsets=dict(last_state.event_offsets),
    )
```

Each run's committed checkpoint becomes (everything previously known) + (this run's updates), so once a document is embedded its hash persists, skips accumulate, and re-embedding converges to ~0 for unchanged content. `get_last_state()` still honors `ignore_old_state`, so that path intentionally seeds from empty.

## Checklist

- [x] The PR conforms to DataHub's Contributing Guideline (particularly the Commit Message Format)
- [x] Tests added (regression tests for checkpoint seeding and state read/update behavior)

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/datahub_documents/document_chunking_state_handler.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/datahub_documents/document_chunking_state_handler.py
@@ -1,5 +1,6 @@
 """State handler for document chunking stateful ingestion."""
 
+import copy
 import logging
 from typing import Optional, cast
 
@@ -85,12 +86,26 @@ class DocumentChunkingStateHandler(
         if not self.is_checkpointing_enabled() or self._ignore_new_state():
             return None
 
+        # Seed the new checkpoint from the last committed state so previously
+        # embedded document hashes (and event offsets) carry forward. Otherwise
+        # each run starts from an empty state and only records the documents it
+        # processed, so the committed checkpoint overwrites rather than
+        # accumulates. That makes unchanged documents re-embed on later runs
+        # (the skip set never converges) instead of being skipped.
+        state = DocumentChunkingCheckpointState()
+        last_state = self.get_last_state()
+        if last_state:
+            state = DocumentChunkingCheckpointState(
+                document_state=copy.deepcopy(last_state.document_state),
+                event_offsets=dict(last_state.event_offsets),
+            )
+
         assert self.pipeline_name is not None
         return Checkpoint(
             job_name=self.job_id,
             pipeline_name=self.pipeline_name,
             run_id=self.run_id,
-            state=DocumentChunkingCheckpointState(),
+            state=state,
         )
 
     def get_current_state(
--- a/metadata-ingestion/tests/unit/test_datahub_documents_source.py
+++ b/metadata-ingestion/tests/unit/test_datahub_documents_source.py
@@ -3,7 +3,7 @@
 import hashlib
 import json
 import sys
-from typing import Any
+from typing import Any, Optional
 from unittest.mock import Mock, patch
 
 import pytest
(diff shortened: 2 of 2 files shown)


nwadams:
Can we also check whether the content has changed, and not re-embed unless text content has changed?

alexjst:
Good call — that check already exists, and this PR is what makes it effective.

`_should_process()` computes a sha256 of the document text (`_calculate_text_hash`) and compares it against the hash stored from the previous run (`state_handler.get_document_hash(urn)`), skipping and reporting `num_documents_skipped_unchanged` when they match. So a document is only re-embedded when its text content actually changes.

The problem was that the stored hash wasn't surviving between runs. `create_checkpoint()` started from an empty state each run and only recorded the documents processed that run, so the committed checkpoint overwrote the prior hashes instead of accumulating them. `get_document_hash()` then returned `None` for everything not touched in the previous run, and the content comparison treated those as "changed" and re-embedded them.

This PR seeds the new checkpoint from the last committed state, so the hashes persist and the existing content-change check works as intended — re-embedding converges to ~0 for unchanged content. No separate content check needed on top.

nwadams:
> Good call — that check already exists, and this PR is what makes it effective.
> 
> `_should_process()` computes a sha256 of the document text (`_calculate_text_hash`) and compares it against the hash stored from the previous run (`state_handler.get_document_hash(urn)`), skipping and reporting `num_documents_skipped_unchanged` when they match. So a document is only re-embedded when its text content actually changes.
> 
> The problem was that the stored hash wasn't surviving between runs. `create_checkpoint()` started from an empty state each run and only recorded the documents processed that run, so the committed checkpoint overwrote the prior hashes instead of accumulating them. `get_document_hash()` then returned `None` for everything not touched in the previous run, and the content comparison treated those as "changed" and re-embedded them.
> 
> This PR seeds the new checkpoint from the last committed state, so the hashes persist and the existing content-change check works as intended — re-embedding converges to ~0 for unchanged content. No separate content check needed on top.

@alexjst is there no way to check whether the document text itself has been updated? we don't store a hash in the embedded document?

metadata-ingestion/src/datahub/ingestion/source/datahub_documents/document_chunking_state_handler.py
```diff
@@ -85,12 +86,26 @@ def create_checkpoint(
         if not self.is_checkpointing_enabled() or self._ignore_new_state():
             return None
 
+        # Seed the new checkpoint from the last committed state so previously
+        # embedded document hashes (and event offsets) carry forward. Otherwise
+        # each run starts from an empty state and only records the documents it
+        # processed, so the committed checkpoint overwrites rather than
+        # accumulates. That makes unchanged documents re-embed on later runs
+        # (the skip set never converges) instead of being skipped.
+        state = DocumentChunkingCheckpointState()
```
jjoyce0510, line 95: is this checkpoint state always honored?

jjoyce0510: or are there some cases where it is not honored?

jjoyce0510: can you confirm that the state is correctly aggregating content hashes for a given document over multiple runs? 

if not, we could get into a reindexing situation.

jjoyce0510 (approved):
Let's go ahead and get this in.

PR #17946, opened 2026-06-18
feat(ingestion): Predictable schema size trimming

skrydal:
We want to truncate the aspect starting with the deepest fields first. This also ensures that we always have upper-level fields.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
+++ b/metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
@@ -203,7 +203,7 @@ class EnsureAspectSizeProcessor(WorkunitProcessor[EnsureAspectSizeProcessorRepor
         )
         return False
 
-    def _extract_field_path_length(self, field: SchemaFieldClass) -> int:
+    def _extract_field_path_depth(self, field: SchemaFieldClass) -> int:
         return len(
             [t for t in re.sub(r"\[[^]]*]", "", field.fieldPath).split(".") if t]
         )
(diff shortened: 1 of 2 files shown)


metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
```diff
@@ -230,11 +236,13 @@ def ensure_schema_metadata_size(
 
         total_fields_size = self._schema_size_without_fields(schema)
         accepted_fields: List[SchemaFieldClass] = []
-        for schema_field in schema.fields:
+        sorted_fields = sorted(schema.fields, key=self._extract_field_path_length)
```
puneetagarwal-datahub: When we sort this , do we have anything downstream which is dependent on original sorted order as this line will sort them according to field-path-length ?

puneetagarwal-datahub (changes requested):

metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
```diff
@@ -202,6 +203,11 @@ def _drop_platform_schema(self, schema: SchemaMetadataClass) -> bool:
         )
         return False
 
+    def _extract_field_path_length(self, field: SchemaFieldClass) -> int:
```
sgomezvillamor: Some suggestions for a more descriptive method name:

```
_(get|count|extract)_field_path_depth
_(get|count|extract)_field_path_nesting_level
```

sgomezvillamor (approved):

metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
skrydal: Not to my knowledge. We are actually introducing more determinism here as to how we will trim (still there is undefined order for items with the same depth - but to address this we would need to add some index to the fields - but this would involve actually redesigning schema fields model, which might be a very good idea in a long term, but requiring significant effort).

metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
skrydal: Good idea

puneetagarwal-datahub (approved):

PR #17950, opened 2026-06-18
feat(ingest): add SDK utility for server entity/aspect specs capability detection

aviraj-gour:
## Summary

Gives the Python SDK / ingestion framework a first-class way to ask a running DataHub server **what it actually supports**, so SDK code can make runtime decisions from the backend's real capabilities instead of assuming.

It answers two questions about the live server:

1. Which `(entityType, aspectName)` pairs are registered?
2. Which `schemaVersion` does the server advertise for a given aspect?

This is a shared building block for upcoming features that must feature-detect a new aspect before relying on it (and fall back to local computation otherwise).

## What changed

**Value type — `datahub/utilities/entity_aspect_specs.py`**

`EntityAspectSpecs`: a parsed, queryable view of the server's specs, with **no I/O** — pure data plus parsing/serialization.

- `supports(entityType, aspectName) -> bool` — raises `ValueError` for an unregistered entity type rather than guessing.
- `schema_version(aspectName) -> Optional[int]` — the advertised version; `None` only for the genuine "registered but no version advertised" (older server) case; raises `ValueError` if the aspect isn't registered at all.
- `from_registry_elements(...)` to build from the registry payload; `to_dict`/`from_dict` for caching.

**Graph client — `DataHubGraph.get_entity_aspect_specs()`**

All fetch logic lives on the client. It pages the registry `/openapi/v1/registry/models/entity/specifications` endpoint and is **two-tier cached**:

- **In-memory memo** for the client's lifetime (returns `None` on failure without memoizing, so a transient GMS blip doesn't permanently disable detection).
- **On-disk** (below), so short-lived processes (CLI, datapack loads) skip re-fetching across restarts.

Pagination is extracted into a reusable `_paginate_offset()` helper for offset-paginated (`start`/`count`/`total`) GMS endpoints, routed through the existing `_get_generic` request path.

**New reusable primitive — `datahub/utilities/server_state_disk_cache.py`**

`ServerStateDiskCache`: a best-effort JSON disk cache keyed by **server URL + commit hash**.

- The commit hash in the key makes entries **self-invalidate on server upgrade** — stale data is unreachable by construction, no explicit invalidation needed.
- **Atomic writes** (temp file + rename), **symlink rejection**, and **never raises** — any I/O error is treated as a cache miss. Auto-cleans entries older than 7 days and orphaned temp files.
- Inlines `~/.datahub` rather than importing from `datahub.cli`, keeping `utilities/` free of any dependency on the higher-level `cli` package (same approach as `graphql_query_adapter`). Serialization stays at the call site, so it stays a generic primitive reusable by other utils.

**Consolidation — removes `cli/datapack/schema_compat.py`**

The data-pack loader had a near-duplicate paginated registry fetch + disk cache. `_apply_schema_filter` now uses `DataHubGraph.get_entity_aspect_specs()` directly; a small local `_is_mcp_compatible` keeps datapack's deliberately permissive "unknown entity type → allow, let the server decide" policy. (~170 lines of duplicated fetch/cache logic deleted.)

## Testing

**Test relocation — graph-client smoke tests moved out of `tests/cli/`**

`smoke-test/tests/cli/datahub_graph_test.py` tested `DataHubGraph` client methods (`get_aspect_v2`, `get_entities`, `graph_relationships`), not the `datahub` CLI — it only lived under `cli/` for historical reasons. Since this PR adds another graph-client smoke test, the tests now have a proper home:

- Moved → `smoke-test/tests/graph/test_graph_client.py` (+ its `graph_data.json` fixture), mirroring the unit layer at `metadata-ingestion/tests/unit/ingestion/graph/test_client.py`.
- `tests/cli/` now holds only genuine CLI-command tests (`*_cmd/` + the rollback flow); new graph-client tests no longer land there.
- Updated the matching `pytest_test_weights.json` keys to the new module path.

No behavior change: nothing imported the module, and the shared `graph_client` fixture lives in `conftest.py` (unaffected). Pure organization — smoke batching is by test-count, not folder, so execution is identical.

**New coverage added**

- **Unit** (`test_server_state_disk_cache.py`): the `ServerStateDiskCache` save → cleanup → load cycle — stale entries past `max_age_days` are evicted on the next `put()` while fresh ones survive, and orphaned temp files from crashed writes are swept.
- **Unit** (`test_client.py`): in-memory memo invalidates when the server's commit hash changes (upgrade → refetch), and the disk cache is shared across client instances.
- **Smoke / e2e** (`tests/graph/test_graph_client.py`): `get_entity_aspect_specs()` against the live `/specifications` endpoint, asserting both parsing branches (`keyAspectName` and `aspectSpecs`) so an API-response-shape break is caught end to end.


sgomezvillamor:
Really good iteration — the `ServerStateDiskCache` extraction is the right move and going through `_session`/`_get_generic` means auth + retry are now inherited for free. Kudos!

**🟡 In-memory cache has no invalidation on server upgrade.** `_entity_aspect_specs` is memoized without a commit hash guard — a long-running client will serve stale specs indefinitely after a server upgrade. `QueryProjector` in `graphql_query_adapter.py` already handles this with a `_schema_commit_hash` check; worth following the same pattern.

**🟡 Module placement.** `EntityAspectSpecs` is now a pure data model whose only consumer is `DataHubGraph`. `datahub.ingestion.graph` would be a more natural home than `datahub.utilities`.

**🟡 TypedDict for API response shapes.** `entity`/`aspect_spec` carry external API contract — `TypedDict` annotations on `_EntityElement` / `_AspectSpec` / `_AspectAnnotation` would document the expected shape and let mypy catch key typos at the schema boundary.

**🟡 Missing test coverage.**
- Smoke test: e2e against the live `/specifications` endpoint.
- Smoke test: server-upgrade scenario (commit hash change → cache miss → refetch).
- `ServerStateDiskCache`: integration test (or smoke if needed) covering the full save → cleanup → load cycle.

**💡 `_paginate_offset` is new — is it reusable?** Clean addition. Worth checking whether existing manual while-loops elsewhere in `DataHubGraph` for offset-paginated endpoints could be consolidated here.

**💡 `from_registry_elements` → `from_registry_api_elements`** — makes the origin clearer.

**💡 `_all_aspects()` — `aspect_schema_versions.keys()` is NOT equivalent.** Older servers register aspects without advertising a `schemaVersion`, so `.keys()` would be a strict subset. A short comment here would prevent future readers from making the same incorrect simplification.

---

aviraj-gour:
Thanks for Review @sgomezvillamor,
This was really helpful

> In-memory cache has no invalidation on server upgrade. _entity_aspect_specs is memoized without a commit hash guard — a long-running client will serve stale specs indefinitely after a server upgrade. QueryProjector in graphql_query_adapter.py already handles this with a _schema_commit_hash check; worth following the same pattern.

Fixed. get_entity_aspect_specs() now pairs the in-memory memo with the commit hash it was built against and only short-circuits when the current hash matches.

> Module placement. EntityAspectSpecs is now a pure data model whose only consumer is DataHubGraph. datahub.ingestion.graph would be a more natural home than datahub.utilities.

Moved entity_aspect_specs.py (and its test) into datahub.ingestion.graph, sibling to client.py.

> TypedDict for API response shapes. entity/aspect_spec carry external API contract — TypedDict annotations on _EntityElement / _AspectSpec / _AspectAnnotation would document the expected shape and let mypy catch key typos at the schema boundary.

Added _AspectAnnotation / _AspectSpec / _EntityElement (all total=False, The parser treats every key as optional — it reads all four with .get() and tolerates absence/null, also schemaVersion is absent on older servers.)

> Missing test coverage.
> Smoke test: e2e against the live /specifications endpoint.
> Smoke test: server-upgrade scenario (commit hash change → cache miss → refetch).
> ServerStateDiskCache: integration test (or smoke if needed) covering the full save → cleanup → load cycle.

Added a Test for all the scenario.

Also, datahub_graph_test.py tested DataHubGraph client methods, not the CLI, yet lived under tests/cli/. Relocate it (and its graph_data.json fixture) to a dedicated tests/graph/ package, mirroring the unit layer at tests/unit/ingestion/graph/test_client.py, so CLI command tests andgraph-client tests have distinct homes. Update the test-weight keys to match the new module pat

> _paginate_offset is new — is it reusable? Clean addition. Worth checking whether existing manual while-loops elsewhere in DataHubGraph for offset-paginated endpoints could be consolidated here.

Checked all other paginated sites in DataHubGraph — the rest are cursor-based (scrollAcrossEntities) or single-page POST searches with a different response shape, so none fit the start/count/total/elements GET contract. Left the helper as-is; nothing to consolidate today.

> from_registry_elements → from_registry_api_elements — makes the origin clearer.

Renamed across all three sites (definition, caller, test) to make the origin clearer.

>  _all_aspects() — aspect_schema_versions.keys() is NOT equivalent. Older servers register aspects without advertising a schemaVersion, so .keys() would be a strict subset. A short comment here would prevent future readers from making the same incorrect simplification.

Added a comment

smoke-test/tests/graph/test_graph_client.py
```diff
@@ -149,3 +149,12 @@ def test_graph_relationships(graph_client, auth_session):
...
     _ensure_dataset_present_correctly(auth_session, graph_client)
+
+
+def test_get_entity_aspect_specs(graph_client):
+    # E2E against the live specifications endpoint; a broken response shape
+    # yields empty specs and fails here.
+    specs = graph_client.get_entity_aspect_specs()
+    assert specs is not None
+    assert specs.supports("dataset", "datasetKey")  # keyAspectName branch
+    assert specs.supports("dataset", "datasetProperties")  # aspectSpecs branch
```
sgomezvillamor, line 160: should `schemaVersion` be added here?
should some negative case be added here too? eg non existing entity or aspect?

sgomezvillamor (approved):
LGTM 🎖️

smoke-test/tests/graph/test_graph_client.py
aviraj-gour: Added a assertions for `schemaVersion`

PR #17951, opened 2026-06-18
feat(ingest/snowflake): resolve external stage URLs via S3 path specs for lineage

alokr-dhub:
## Summary

Snowflake external-stage URLs (e.g. `s3://my-bucket/events/year=2024/month=01/`) are directory prefixes more granular than the logical dataset. Previously the stage URL was turned into a lineage URN as-is, which never matched the table-level URN that the S3 source emits (`s3://my-bucket/events`), so lineage didn't stitch.

This change pipes Snowflake external stages through the same `path_specs` machinery the S3 source uses, folding the stage URL up to the `{table}` boundary so the resulting URN matches. It works for S3, GCS, and Azure stages.

## Changes

- **Shared data lake lineage config** — Lifted the lineage-provider config out of `redshift/config.py` into `data_lake_common/config.py` as `DataLakeLineageProviderConfig` / `DataLakeLineageProviderConfigBase` (grouped under a `datalake_lineage_config` property) so Snowflake and any future source can reuse it. Added `get_path` / `get_urn_for_lineage` helpers that fold a path to its `{table}` boundary and dispatch URN generation by scheme (`s3://`, `gcs://`/`gs://`, Azure HTTPS / `azure://`). The old `S3LineageProviderConfig` / `S3DatasetLineageProviderConfigBase` names are kept as back-compat aliases (Redshift still consumes `s3_lineage_config` unchanged).
- **`PathSpec.fold_dir_to_table`** — New method that resolves a directory prefix (which doesn't satisfy the file-oriented `allowed()` check) to the dataset boundary marked by `{table}`. Honours the spec's `exclude` list and `tables_filter_pattern`.
- **`PathMode` enum** — Replaces a boolean `strip_filename` parameter. `PathMode.FILE` handles COPY/UNLOAD-style file paths; `PathMode.DIRECTORY` handles stage-style directory prefixes. (`S3PathMode` is retained as an alias.)
- **Snowflake stages** — `_resolve_external_stage_url` now calls `config.datalake_lineage_config.get_urn_for_lineage(url, env, mode=PathMode.DIRECTORY)`. Stages filtered out by `ignore_non_path_spec_path` are tallied in a new `num_stage_external_lineage_dropped` report counter.
- **Docs** — `snowflake_post.md` explains when to configure `datalake_lineage_config` for stages and gives an example.

## Test plan

- [x] New unit tests for `DataLakeLineageProviderConfig.get_path` / `get_urn_for_lineage` covering file mode, directory mode, `ignore_non_path_spec_path` interactions, prefix folding, multi-spec ordering, and S3/GCS/Azure scheme dispatch
- [x] New unit tests for `PathSpec.fold_dir_to_table` covering missing `{table}` token, shallow prefix, glob mismatch, and `exclude` enforcement
- [x] New unit tests for `_resolve_external_stage_url` asserting:
  - A nested stage URL folds to the dataset URN
  - A stage URL at exactly the `{table}` boundary still resolves under `ignore_non_path_spec_path`
  - A non-matching stage URL is dropped and the new report counter increments
- [x] All existing tests in `tests/unit/data_lake/`, `tests/unit/redshift/`, `tests/unit/s3/` still pass

## Checklist

- [x] Conforms to the [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#pr-title-format)
- [x] Tests added
- [x] Docs added (`metadata-ingestion/docs/sources/snowflake/snowflake_post.md`)
- [x] No breaking changes (default `datalake_lineage_config` is a no-op; existing behaviour preserved when `path_specs` is empty)


sgomezvillamor:
## Review

Good work overall — clean refactor, solid test coverage, and the problem statement is clear.

---

### 🟡 Strong suggestion: generalize `s3_lineage_config` → `datalake_lineage_config`

The path spec folding only helps S3 stages. GCS and ABS are left raw:

```python
if url.startswith("s3://"):
    urn = self.config.s3_lineage_config.get_s3_urn_for_lineage(...)  # ✅ path spec folding
if url.startswith("gcs://"):
    return make_dataset_urn_with_platform_instance(...)  # ❌ raw, no folding
if url.startswith("azure://"):
    return make_abs_urn(abs_url, self.config.env)  # ❌ raw, no folding
```

Rather than `s3_lineage_config`, consider introducing `datalake_lineage_config` with a unified interface that dispatches by scheme. This avoids landing in a state where we later need to add `gcs_lineage_config`, `abs_lineage_config`, etc. It's also the right abstraction to eventually reuse in Redshift and other sources that touch data lake paths.

---

### 🟡 Warning: trailing slash not normalized in DIRECTORY mode fallback

In `get_s3_path`, when no path spec matches in DIRECTORY mode and `ignore_non_path_spec_path=False`, the method returns the original `path` — potentially with a trailing slash (e.g. `s3://bucket/events/year=2024/`). `fold_dir_to_table` always normalizes via `rstrip("/")`, so the two code paths are inconsistent. If `make_s3_urn_for_lineage` doesn't strip the slash, the resulting URN won't match what the S3 source emits and lineage won't stitch.

---

### ℹ️ Suggestion: add a positive counter alongside `num_stage_lineage_dropped_s3_path`

A `num_stage_lineage_resolved_s3_path` counter would let operators verify the config is actually working — right now you only know how many stages were dropped, not how many were successfully folded.

---

### ℹ️ Minor: type annotation in test helper

```python
config: SnowflakeV2Config = None,  # type: ignore[assignment]
```

→ `config: Optional[SnowflakeV2Config] = None`

---

alokr-dhub:
@sgomezvillamor addressed the comments. Please review.

sgomezvillamor (approved):
All concerns from the previous round have been addressed — the unified `DataLakeLineageProviderConfig` / `datalake_lineage_config`, trailing-slash normalisation, and both report counters are in good shape. A few follow-up suggestions below, none of them blockers.

---

**Suggestion — move `get_path` / `get_urn_for_lineage` out of the config class**

`DataLakeLineageProviderConfig` subclasses `ConfigModel`, which is conventionally a pure value object (validation + defaults). Having `get_path` and `get_urn_for_lineage` as instance methods makes it also a service. Consider moving them to free functions in `path_spec.py` or a new `data_lake_common/lineage_utils.py` in a follow-up, keeping the config as a plain data holder. Not a blocker here.

---

**Suggestion — TODO comment on the backward-compat aliases**

`S3LineageProviderConfig` and `S3DatasetLineageProviderConfigBase` are kept for backward compat but there is no signal in the code that they are transitional. A short comment on each would help:

```python
# TODO: migrate all sources that expose `s3_lineage_config` (e.g. Redshift)
# to use DataLakeLineageProviderConfig / DataLakeLineageProviderConfigBase
# and remove these aliases once all callers are updated.
class S3LineageProviderConfig(DataLakeLineageProviderConfig): ...
```

---

**Minor — two low-value tests to consider removing**

- `test_unsupported_scheme_returns_none`: the assertion is `assert ... is None`, which is fine, but the comment in the test implies the intent is to verify the `logger.debug` path. Testing that a logger call does not raise is not meaningful; the real contract (returning `None` for unknown schemes) is already covered by the GCS/ABS/S3 positive cases.

- `test_prefix_folding_not_applied_to_file_inputs`: this verifies that `FILE` mode does not call `fold_dir_to_table`. That is an implementation detail, not observable behaviour. The test would survive any correct reimplementation but break on an equally-correct one that happens to share the code path differently.

---

**Also worth getting a review from Tamas (@treff7es) given the cross-source impact of the `DataLakeLineageProviderConfig` refactor.**

---

alokr-dhub:
@treff7es Please review this once and let me know if its good to merge.

PR #17952, opened 2026-06-18
chore(i18n): synchronize eslint patterns to exclude

v-tarasevich-blitz-brain:
This PR synchronize eslint i18n excluding patterns with SaaS


chriscollins3456 (approved):
i would double check this is up to date with saas i saw a few minor differences

PR #17954, opened 2026-06-18
feat(ingestion/unity): derive usage and query entities via SqlParsingAggregator

treff7es:
## Summary

Reworks Unity Catalog **usage extraction** to run query history through DataHub's shared
**`SqlParsingAggregator`** — the same component used by the Snowflake, BigQuery, and
Redshift connectors — replacing the connector's bespoke SQL-parsing-and-aggregation path.

When `usage_data_source` resolves to system tables (explicit `SYSTEM_TABLES`, or `AUTO`
with a SQL warehouse configured), the connector reads statements from
`system.query.history` (streamed) and feeds each to the aggregator, which
fingerprint-deduplicates, parses every **unique** statement once, and emits usage
statistics, operations, and Query entities.

**Table and column lineage are unchanged** — they continue to be sourced natively from
`system.access.table_lineage` / `system.access.column_lineage`, with the aggregator run
under `generate_lineage=False` so it never competes with or duplicates the native lineage.

This removes the connector's main CPU bottleneck (a single-threaded, pure-Python parse
loop that re-parsed every occurrence of every statement), deduplicates redundant parsing,
and brings Unity Catalog usage in line with the rest of DataHub's SQL connectors.

### What changed

- **Usage via `SqlParsingAggregator` (all-observed).** Every statement from the existing
  fast fetch is fed via `add_observed_query`; the aggregator dedups by fingerprint and
  parses each unique query once (internal sqlglot caches). The bespoke `_parse_query`,
  Spark-plan fallback (and its `pyspark` dependency), `UsageAggregator`, and hand-rolled
  Query-entity / operation emission are removed.
- **Lineage untouched.** Table and column lineage stay on the `system.access` path;
  `generate_lineage=False` on the aggregator avoids duplicate/competing lineage.
- **Schema resolver** populated from the UC table schemas ingested in the same run, so
  partial (2-part `schema.table`) references resolve; single-catalog recipes use that
  catalog as the parser's `default_db`.
- **Streamed, bounded-memory fetch.** Query history is read with `fetchmany` batching
  (`_execute_sql_query_streaming`) instead of `fetchall`, so peak memory no longer scales
  with the size of the audit log.
- **Query entities** are emitted for warehouse statements (`include_queries`, default `true`),
  including `queryProperties`, `querySubjects`, and `queryUsageStatistics`.
- **History window** relaxed from 30 to 365 days on the system-tables path (the system
  tables retain ~1 year); the 30-day cap still applies to the REST API path.
- **Robust error handling.** A per-query guard skips and counts a single malformed
  statement instead of aborting the whole usage run; the aggregator close is guarded; and
  fetch/emit failures are reported rather than swallowed.

### Performance

Measured on a representative Unity Catalog workspace across three query-history windows
(same recipe, same warehouse, current connector = "master" vs this PR = "head"). Runtime
is the pipeline processing time; end-to-end wall-clock improvement was even larger
(2.0–2.7×) but is more sensitive to sink/network timing, so processing time is reported as
the conservative figure.

| Window | Statements | Unique (parsed by head) | master runtime | head runtime | Speedup |
| ------ | ---------: | ----------------------: | -------------: | -----------: | ------: |
| 14-day |      4,005 |     604 (6.6× fewer)    |       57.3 s   |     37.7 s   |  1.52×  |
| 28-day |      7,146 |   1,146 (6.2× fewer)    |       68.9 s   |     50.2 s   |  1.37×  |
| 90-day |     15,761 |   1,324 (11.9× fewer)   |      118.1 s   |     79.4 s   |  1.49×  |

The speedup comes from eliminating redundant parsing: on master, SQL parsing alone took
**32.7 / 44.4 / 87.7 s** (single-threaded) for the three windows; head parses only the
unique fingerprints, so parse cost grows with the *distinct* query set, not the total
volume.

**Output is equivalent or richer** — identical coverage, plus Query entities:

| Aspect                   | master | head |
| ------------------------ | -----: | ---: |
| `datasetUsageStatistics` |     22 |   22 |
| `operation`              |      1 |    1 |
| Query entities (`queryProperties` etc.) | 0 | ✓ |

**Memory.** Peak RSS was lower for head in every window (≈ −40 / −75 / −26 MB), but the
more important change is *shape*, not the absolute delta at this scale. The old path calls
`cursor.fetchall()`, which materializes the entire `system.query.history` result as one
Arrow table **and** a full `List[Row]` before processing begins; the new path streams via
`cursor.fetchmany()` and discards each batch, so resident memory is bounded by one driver
buffer regardless of total row count. At these volumes (4k–16k statements) the result fits
comfortably either way, so the difference is modest; the divergence grows with audit-log
size. For a busy workspace (e.g. ~500k statements) the old path's `fetchall` is a
multi-hundred-MB-to-GB materialization with real OOM risk on constrained executors, while
the streamed path stays bounded — the scenario that motivated this change.

### Compatibility

`usage_data_source: SYSTEM_TABLES` (and `AUTO` + warehouse) behavior changes in place:
usage/operations are now produced by the shared aggregator rather than the bespoke parser,
and the default backfill window widens to 365 days. Existing ingestions may see usage
results recomputed and a longer first backfill. Documented in `docs/how/updating-datahub.md`.
Opt out with `usage_data_source: API`.

**Known limitation.** Databricks query history does not record the catalog/schema that was
active when a statement ran. Statements that reference tables by a short (one-part) name
only — typical of ad-hoc/notebook queries — cannot be attributed to a specific table, so
usage/operation counts for those statements may be lower than before. (A future
enhancement could recover some of these by combining native `system.access` lineage as
pre-parsed input; tracked separately.)

### What to expect in DataHub after upgrade

On the system-tables usage path, with default settings:

- **New — `Query` entities.** Warehouse SQL statements are now surfaced as DataHub `Query`
  entities (`queryProperties`, `querySubjects`, `queryUsageStatistics`), gated by
  `include_queries` (default `true`). These did **not** appear on this path before. They are
  **deduplicated by SQL fingerprint** — one entity per *distinct* statement, not one per
  execution — so the count is far smaller than the raw query volume (≈1.3k Query entities
  from ≈16k statements in the benchmark workspace). Set `include_queries: false` to suppress
  them.
- **Unchanged — dataset usage coverage.** Per-table `datasetUsageStatistics` is equivalent
  to before (identical dataset count at every window in the benchmark — 22 vs 22).
- **`operation` aspects** are still emitted but are now **fingerprint-deduplicated**:
  repeated identical statements collapse to a single `operation` instead of one per
  execution (so raw per-execution operation counts will look lower, while the distinct
  output is preserved).
- **Possible small decrease** — usage/operation attribution for queries that reference
  tables by a one-part (bare) name (typical ad-hoc/notebook queries) may be lower, per the
  known limitation above.
- **Net effect:** overall emitted volume is slightly **higher**, not lower — the new
  (deduplicated) `Query` entities outweigh the operation dedup. Use `include_queries: false`
  to keep only usage/operations, or `usage_data_source: API` to opt out of the
  system-tables path entirely.

## Testing

- **Unit** (`tests/unit/test_unity_catalog_usage_join.py`): aggregator is built with the
  correct flags and fed each fetched statement; observed-query timestamps normalized to
  UTC; `is_allowed_table` scoping (incl. the platform-instance-prefix case);
  single-catalog `default_db`; schema-resolver wiring; `auto_empty` reset for unqueried
  tables; zero-query path emits nothing. Error paths: `gen_metadata()` raising does not
  propagate, `close()` raising does not propagate, and a single bad query is skipped while
  the rest still feed.
- **Streaming/proxy** (`tests/unit/test_unity_catalog_proxy.py`): `fetchmany` batching,
  warning-on-error without raising, deterministic connection close on early abort,
  no-warehouse handling.
- **Golden integration** (`tests/integration/unity/test_unity_catalog_ingest.py`): drives
  routing → aggregation → emission through the mocked proxy seam and asserts the emitted
  `datasetUsageStatistics`, `operation`, and Query-entity aspects.
- `ruff` + `mypy` clean.

## Golden file changes (expected)

This PR **adds one new golden** and modifies **no existing golden**:

- **New:** `tests/integration/unity/unity_catalog_usage_aggregator_mces_golden.json` — the
  first golden that actually exercises usage *output*. There was previously **no golden
  coverage of usage**: the existing unity goldens (`unity_catalog_mces_golden.json`,
  `unity_catalog_ml_model_mces_golden.json`, constraints, metric-view) contain zero
  `datasetUsageStatistics` / `operation` aspects — usage was either disabled in those
  tests or, on the two that left it enabled, silently produced nothing against the mocked
  client (the old fail-silent path). The new golden feeds a fixed set of queries through
  the `SqlParsingAggregator` and asserts `datasetUsageStatistics`, `operation`,
  `queryProperties`, and `querySubjects`.
- **Two existing tests** now set `include_usage_statistics: False` (matching the rest of
  the suite). They never asserted usage, and the explicit flag is now required because
  query-history fetch failures surface as `report_failure` (a deliberate fix in this PR)
  instead of being swallowed — so leaving usage enabled against a no-query-history mock
  would now correctly fail the run.

**Why a reviewer should expect golden churn on this path:** usage, operations, and query
entities are now produced by the shared `SqlParsingAggregator` in *its* emission format
rather than the connector's bespoke path. Any environment that enables usage on the
system-tables path (or regenerates goldens) will see `datasetUsageStatistics` /
`operation` / `queryProperties` / `querySubjects` aspects appear where there were none —
and, due to fingerprint dedup, repeated identical statements collapse to a single Query
entity instead of one aspect per execution. These differences are intended, not
regressions.

## Checklist

- [x] The PR conforms to DataHub's Contributing Guideline (particularly the Commit Message and PR Title formats)
- [x] Links to related issues (N/A)
- [x] Tests added/updated
- [x] Docs added/updated (`docs/how/updating-datahub.md`)


treff7es:
### Note: scalability & the intentional double-read of `system.access.table_lineage`

When **both lineage and usage** extraction run on the system-tables path, `system.access.table_lineage` is read by **two separate queries**, by design — they are not merged:

- **Lineage** reads it **aggregated per catalog** (`GROUP BY` source/target → distinct table/column edges, `max(event_time)`), and does **not** require `statement_id` — so it captures edges from all compute, including classic/all-purpose clusters.
- **Usage** reads it **per statement** (`INNER JOIN system.query.history ON statement_id`, `statement_id IS NOT NULL`, `direct_access = true`) to attribute reads/writes per query and emit usage stats + Query entities.

These cannot be collapsed into one scan without losing coverage: the usage join excludes non-warehouse lineage (`statement_id IS NULL`) that the lineage path needs, the two use different granularity (distinct edges vs. per-statement), and lineage additionally needs `system.access.column_lineage`. Both queries prune on the `event_date` partition column.

### Memory safety on large audit logs

The usage path is designed to bound memory on workspaces with millions of query statements:

- The system-tables usage/history queries stream results via `fetchmany` batches (no full `fetchall` materialization), with the cursor deterministically closed via `contextlib.closing()` even on early consumer abort.
- The per-run de-duplication structures (`matched_ids`, query fingerprints) are `FileBackedDict`-backed (spill to disk), and the usage aggregator is already SQLite-backed.

For very large histories, narrowing the time window / running incrementally remains the recommended operational approach.

metadata-ingestion/src/datahub/ingestion/source/unity/config.py
```diff
@@ -367,6 +367,12 @@ class UnityCatalogSourceConfig(
         ),
     )
 
+    emit_queries: bool = pydantic.Field(
+        default=True,
+        description="Emit DataHub Query entities for statements reconstructed from "
+        "system tables. Only effective on the system-tables usage path.",
+    )
```
askumar27: Not sure if I understand the config by name

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
```diff
@@ -668,16 +683,11 @@ def get_query_history_via_system_tables(
...
-            logger.error(
-                f"Error fetching query history from system tables: {e}", exc_info=True
-            )
-            self.report.report_failure(
-                title="Failed to fetch query history from system tables",
-                message="Error querying system.query.history table",
-                context=f"Query period: {start_time} to {end_time}",
-            )
+                    self.report.report_warning(
+                        "query-parse-system-table",
```
askumar27: This can be more human readable, as UI propagates this message to the user.

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
```diff
@@ -1310,83 +1320,141 @@ def _create_query(self, info: QueryInfo) -> Optional[Query]:
...
+            ]
+        ):
+            logger.error(
+                "SQL query failure appears to be proxy-related. "
+                "Please check proxy configuration and authentication. "
+                f"Proxy environment variables detected: {list(proxy_env_debug.keys())}"
+            )
+
+        self.report.report_warning(
+            "sql-query-failed",
```
askumar27: Can we make this well formatted titles?

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
```diff
@@ -1310,83 +1320,141 @@ def _create_query(self, info: QueryInfo) -> Optional[Query]:
...
-                rows = cursor.fetchall()
-                logger.debug(
-                    f"SQL query executed successfully, returned {len(rows)} rows"
-                )
-                return rows
+                while True:
+                    batch = cursor.fetchmany(batch_size)
+                    if not batch:
+                        break
+                    yield from batch
```
askumar27: 🔴 **BLOCKER (B1):** `yield from batch` inside `try/except Exception` — catches downstream consumer errors, not yours.

Per DataHub ingestion review guidelines, `try/except` around `yield` is an auto-rejection blocker. When a consumer of this generator throws an exception (e.g., the `Query(...)` constructor fails in `get_query_history_via_system_tables`), Python propagates it into the generator via `.throw()`. The `except Exception` catches it, records a spurious `"sql-query-failed"` warning, and silently increments `num_usage_query_fetch_failures` — masking the real error.

**Fix:** Separate connection setup (protected by try/except) from the streaming yield loop:

```python
try:
    connection = connect(**sql_connection_params)
    cursor = connection.cursor()
    cursor.execute(query, list(params))
except Exception as e:
    self._report_sql_query_failure(e, query, params, proxy_env_debug)
    self.report.num_usage_query_fetch_failures += 1
    return

with closing(connection), closing(cursor):
    while True:
        batch = cursor.fetchmany(batch_size)
        if not batch:
            break
        yield from batch   # now outside try/except
```

askumar27 (commented):
## Deep Review: Unity Catalog SqlParsingAggregator Migration

**Risk:** Medium | **Blast radius:** All UC recipes using `SYSTEM_TABLES` or `AUTO` + warehouse

The architecture of this refactor is excellent — replacing bespoke parsing with the shared aggregator is the right call, the streaming approach is well-motivated, and the test coverage is thorough. The PR description is exemplary (performance benchmarks, compatibility notes, known limitations).

However, there are two `yield`-inside-`try/except` sites that are the module's explicit auto-rejection pattern per DataHub ingestion review guidelines, plus a mid-stream failure gap that could produce silently incomplete data in production.

### Summary

| # | Severity | Location | Issue |
|---|----------|----------|-------|
| B1 | 🔴 BLOCKER | `proxy.py:1453` | `yield from batch` inside `try/except Exception` catches consumer errors |
| B2 | 🔴 BLOCKER | `usage.py:170` | `yield from auto_empty_dataset_usage_statistics(...)` inside `try/except Exception` |
| W1 | 🟡 WARNING | `usage.py:141` | Mid-stream fetch failure after `num_queries > 0` bypasses failure detection |
| W2 | 🟡 WARNING | `config.py:370` | `emit_queries=True` default — new feature should default to `False` |
| W3 | 🟡 WARNING | `proxy.py:1456` | Fetch failure downgraded from `report_failure` to `report_warning` |
| W4 | 🟡 WARNING | `report.py:40` | `num_operational_stats_workunits_emitted` field is dead (never incremented) |
| S1 | ℹ️ SUGGESTION | `usage.py:127` | Per-query `except Exception` should re-raise `MemoryError`/`SystemExit` |
| S2 | ℹ️ SUGGESTION | `usage.py:185` | `aggregator.close()` failure should surface via `report_warning` |
| S3 | ℹ️ SUGGESTION | `proxy.py:685` | Missing `exc_info=True` on row-parse warning |

### Test coverage

Overall excellent. Two gaps worth noting:
- **API path** (`usage_uses_system_tables=False`) has no unit test routing through `proxy.query_history`
- **Mid-stream partial fetch** at the usage-extractor layer is untested (streaming layer is tested)

metadata-ingestion/src/datahub/ingestion/source/unity/usage.py
```diff
@@ -1,329 +1,189 @@
...
-                ],
-                target_tables=[
-                    self._parse_sqlglot_table(table)
-                    for table in sql_parser_in_tables.out_tables
-                ],
+                # Skip resetting per-table usage when we couldn't read any queries at
+                # all (empty history or missing permission).  Emitting zero-usage aspects
+                # in that case would wrongly wipe existing usage stats in DataHub.
+                return
+            yield from auto_empty_dataset_usage_statistics(
```
askumar27, line 191: 🔴 **BLOCKER (B2):** `yield from auto_empty_dataset_usage_statistics(...)` inside `try/except Exception` — same pattern as B1.

This `yield from` is a lazy generator chain. Any exception thrown by the downstream consumer (e.g., serialization error, `GeneratorExit` from early cancellation) gets caught and misreported as `"usage-extraction"` failure. The real traceback is lost.

**Note:** This pattern existed in the old code too (`yield from self._get_workunits_internal(table_refs)` inside try/except), so it's not a regression — but since this is a rewrite, it's an opportunity to fix it.

**Fix:** Move setup into the try block, then yield outside:

```python
try:
    aggregator = self._build_aggregator(...)
    for query in self._fetch_queries():
        ...  # non-yielding work
    if self.report.num_queries == 0:
        ...  # early returns
        return
except Exception as e:
    self.report.report_failure("usage-extraction", ...)
    return

# yield outside try/except — consumer exceptions propagate correctly
try:
    yield from auto_empty_dataset_usage_statistics(
        auto_workunit(aggregator.gen_metadata()), ...
    )
finally:
    if aggregator is not None:
        try:
            aggregator.close()
        except Exception:
            logger.warning("Failed to close SqlParsingAggregator", exc_info=True)
```

metadata-ingestion/src/datahub/ingestion/source/unity/usage.py
```diff
@@ -1,329 +1,189 @@
...
-                return None
-            else:
-                return QueryTableInfo(
-                    source_tables=self._resolve_tables(
-                        table_info.source_tables, table_map
-                    ),
-                    target_tables=self._resolve_tables(
-                        table_info.target_tables, table_map
-                    ),
+            if self.report.num_queries == 0:
```
askumar27, line 170: 🟡 **WARNING (W1):** Mid-stream fetch failure produces silently incomplete data.

This failure-detection gate only fires when `num_queries == 0`. If `fetchmany` succeeds for batch 1 (yielding rows, incrementing `num_queries`) but fails for batch 2, `num_queries > 0` and this branch is never entered. The run completes "successfully" with silently incomplete usage data — operators see normal query counts and no error.

This is particularly risky because the streaming generator (`_execute_sql_query_streaming`) reports fetch failures as `report_warning` (not `report_failure`), and the only escalation to failure is here.

**Fix:** Check `num_usage_query_fetch_failures` unconditionally *after* the loop, regardless of `num_queries`:

```python
# After the for loop, before the num_queries == 0 check:
if self.report.num_usage_query_fetch_failures > fetch_failures_before:
    self.report.report_failure(
        "usage-fetch-incomplete",
        f"Query-history fetch failed after reading {self.report.num_queries} queries. "
        "Usage statistics may be incomplete. See 'sql-query-failed' warning for details.",
    )
    # Continue to emit partial data, but the failure ensures operators are alerted.
```

metadata-ingestion/src/datahub/ingestion/source/unity/config.py
```diff
@@ -367,6 +367,12 @@ class UnityCatalogSourceConfig(
         ),
     )
 
+    emit_queries: bool = pydantic.Field(
```
askumar27: 🟡 **WARNING (W2):** New feature defaults to `True`.

Per DataHub conventions: *"New features default to false — Shipping default=True silently changes behavior."*

Every existing UC recipe using system tables will suddenly emit Query entities without opt-in (~1.3k entities from ~16k statements in the benchmark). While the `updating-datahub.md` entry documents this, the opt-out friction is high — users must add `emit_queries: false` to suppress a new class of entities they didn't ask for.

Consider defaulting to `False` for the initial release, then flipping to `True` in a follow-up after users have a cycle to observe.

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
```diff
@@ -1310,83 +1320,141 @@ def _create_query(self, info: QueryInfo) -> Optional[Query]:
...
-                ]
-            ):
-                logger.error(
-                    "SQL query failure appears to be proxy-related. "
-                    "Please check proxy configuration and authentication. "
-                    f"Proxy environment variables detected: {list(proxy_env_debug.keys())}"
-                )
-
-            return []
+            self._report_sql_query_failure(e, query, params, proxy_env_debug)
```
askumar27: 🟡 **WARNING (W3):** Fetch failure downgraded from `report_failure` to `report_warning`.

The old `get_query_history_via_system_tables` called `report.report_failure()` when the query-history fetch failed. The new shared `_report_sql_query_failure` helper uses `report.report_warning("sql-query-failed", ...)`. A permission error on `system.query.history` was previously a red failure in the ingestion UI — now it's a yellow warning that operators may miss.

The usage extractor compensates when `num_queries == 0`, but that gate is fragile (see W1 — partial fetches bypass it entirely).

Consider adding a `severity` parameter to `_report_sql_query_failure` so the streaming path can escalate to `report_failure`, or have `_execute_sql_query_streaming` call `report_failure` directly instead of sharing the helper.

metadata-ingestion/src/datahub/ingestion/source/unity/report.py
```diff
@@ -44,14 +33,8 @@ class UnityCatalogReport(SQLSourceReport):
...
-    num_queries_duplicate_table: int = 0
-    num_queries_parsed_by_spark_plan: int = 0
-    usage_perf_report: UnityCatalogUsagePerfReport = field(
-        default_factory=UnityCatalogUsagePerfReport
-    )
+    num_queries_dropped: int = 0
+    num_usage_query_fetch_failures: int = 0
 
     # Distinguish from Operations emitted for created / updated timestamps
     num_operational_stats_workunits_emitted: int = 0
```
askumar27: 🟡 **WARNING (W4):** Dead field — never incremented after this PR.

The code that incremented this counter (`_generate_operation_workunit`) was removed. The `SqlParsingAggregator` generates operations internally and doesn't increment this field. It will always be 0, misleading operators who monitor it.

Either remove the field or wire it up to the aggregator's internal operation count.

metadata-ingestion/src/datahub/ingestion/source/unity/usage.py
```diff
@@ -1,329 +1,189 @@
...
+                                else None
+                            ),
+                            default_db=default_db,
+                            default_schema=None,
+                        )
                     )
-                else:
-                    logger.info(
-                        "Using API for usage query history (AUTO mode, no warehouse)"
+                except Exception as per_query_exc:
```
askumar27: ℹ️ **SUGGESTION (S1):** Per-query `except Exception` should re-raise system errors.

`add_observed_query` is a local in-memory operation — the realistic failure mode is a sqlglot parse exception. But `except Exception` also catches `MemoryError`/`SystemExit`/`KeyboardInterrupt`, which would be silently counted as "dropped queries" rather than propagating.

```suggestion
                except (MemoryError, SystemExit, KeyboardInterrupt):
                    raise
                except Exception as per_query_exc:
```

metadata-ingestion/src/datahub/ingestion/source/unity/usage.py
```diff
@@ -1,329 +1,189 @@
...
-                missing_table = True
-            else:
-                refs = table_map[table]
-                if len(refs) == 1:
-                    output.append(refs[0])
-                else:
+        finally:
+            if aggregator is not None:
+                try:
+                    aggregator.close()
```
askumar27, line 199: ℹ️ **SUGGESTION (S2):** `aggregator.close()` failure should surface in the ingestion report.

Currently only logged at `logger.warning` — users see the ingestion report, not server logs. Add a `report_warning` so operators are aware of resource leaks:

```python
except Exception:
    logger.warning("Failed to close SqlParsingAggregator", exc_info=True)
    self.report.report_warning(
        "aggregator-close-failed",
        "SqlParsingAggregator failed to close cleanly. Resources may have leaked.",
    )
```

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
```diff
@@ -668,16 +683,11 @@ def get_query_history_via_system_tables(
                     )
                 except Exception as e:
                     logger.warning(f"Error parsing query from system table: {e}")
```
askumar27: ℹ️ **SUGGESTION (S3):** Missing `exc_info=True` on row-parse warning.

This `logger.warning` uses an f-string without `exc_info=True`, so the traceback is lost — making it harder to diagnose which field caused the `Query(...)` constructor to fail. Compare with the REST API path at line ~585 which correctly includes `exc_info=True`.

```suggestion
                    logger.warning(f"Error parsing query from system table: {e}", exc_info=True)
```

askumar27 (approved):

askumar27:
## PR validation: ingest comparison vs. `acryl-datahub` 1.6.0.4

  Ran the same Unity Catalog recipe (workspace `dbc-a2674c21-c551`, catalog `main`, `platform_instance: core_finance`, `include_notebooks: false`) against two builds with `file` sink, then diffed with `datahub check metadata-diff`.

  | | Latest (1.6.0.4) | PR build (`0.0.0.dev1`) | Δ |
  |---|---|---|---|
  | Total MCPs | 271 | 282 | **+11** |
  | Unique URNs | 57 | 59 | +2 |

  ### Semantic changes (attributable to this PR)

  - **+2 new `query` entities** emitted via `SqlParsingAggregator`:
    - `urn:li:query:3042f9bb…`, `urn:li:query:7dc48dcb…`
    - Each contributes `queryProperties`, `querySubjects`, `queryUsageStatistics`, `dataPlatformInstance`, `status` (12 aspects total).
  - **`operation` aspect reshaped**: 2 emissions with `sourceType` + `queries` → 1 emission with `affectedDatasets` (−1 aspect).
  - **`datasetUsageStatistics` enriched** on tables referenced by parsed queries — e.g. `core_finance.main.lineagedemo.dinner` and `dinner_with_two_column_lineage` go from `uniqueUserCount=0, totalSqlQueries=0` to `=1`, with `topSqlQueries` and `userCounts` populated.

  Reconciliation: `+12 query aspects − 1 operation aspect = +11 events` ✅

  ### Non-semantic noise

  All remaining diff entries are `attribution.time` microsecond drift between runs and tag-list ordering shuffles within `globalTags` / `schemaMetadata`. No content changes.

  ### Aspect counts unchanged

  `container.*`, `dataset.{schemaMetadata, datasetProperties, datasetUsageStatistics, browsePathsV2, container, dataPlatformInstance, globalTags, siblings, status, subTypes, upstreamLineage, viewProperties}`, `tag.tagKey` — all identical in count.

PR #17955, opened 2026-06-18
fix(i18n): fix more dropdown with zero

v-tarasevich-blitz-brain:
Fixes More with zero

image


chriscollins3456 (commented):
i've already merged this fix in here actually: https://github.com/datahub-project/datahub/pull/18006

so we can close this PR

PR #17956, opened 2026-06-18
feat: add dataObject entity for cataloging files/objects in storage

max-datahub:
## Summary

Adds a new first-class **`dataObject`** entity for cataloging files/objects in storage — audio, video, images, documents, and other binary/unstructured files — as fully governable assets (owners, tags, glossary terms, domains, structured properties, documentation, access policies), **without** storing their content. Folders/buckets remain modeled as existing `Container` subtypes.

The **S3 connector** gains an opt-in `unstructured_file_extensions` allowlist: matching files are emitted as `dataObject`s (with storage metadata — MIME type, size, storage class, ETag), while structured files (CSV/JSON/Parquet/Avro) continue to be ingested as datasets. GCS inherits via the S3 adapter.

image

## Motivation

DataHub has no first-class way to catalog non-tabular files in object storage. Bending `dataset` (its schema/lineage/profile aspects don't apply to media) or the knowledge-oriented `document` entity (which assumes text content) both fit poorly. A thin, **content-by-reference** entity fills the gap and is shaped to extend to sub-file "chunks" (self-nesting `parentDataObject`) and to lineage/access later without model churn.

### Entity, GraphQL, frontend, ingestion
- **Model:** `dataObject` + `dataObjectKey` (mirrors `DatasetKey`), `dataObjectProperties`, `objectStorageProperties` (filterable `mimeType`; not a top-level facet), `parentDataObject` (self-nesting). Reuses the standard governance aspects. Not `searchGroup: primary` (matches `document`).
- **GraphQL:** `DataObject` type + mapper + engine registration (read-only entity; governance edits flow through shared `Entity` resolvers).
- **Frontend:** entity registration; search/preview card showing the **subtype** (Audio / Video / Image / Document File / Spreadsheet / Presentation / Archive / …) with "Data Object" as fallback; profile with Documentation + Properties tabs and sidebar sections for owners, tags, glossary terms, domain, **Data Product**, **Application**, and **Structured Properties**, plus status. Supports forms, deprecation, and soft-delete.
- **Ingestion:** S3 `emit_data_object()` behind the opt-in `unstructured_file_extensions` config (empty default = no behavior change); extension→subtype + MIME classification; reuses S3 tag extraction + container hierarchy. GCS inherited. Emission is **streamed** (memory-bounded regardless of object count) and capped by **`max_data_objects`** (per-path-spec; default `100_000`; `null` = unlimited) — on reaching the cap it stops emitting for that path spec and logs a warning (no silent truncation). Folder-level governance via the existing `Container` remains available for users who don't need per-file entities.

### Registered across all entity-enumeration surfaces
A new entity must be wired into many hand-maintained enumeration sites or it breaks silently in some surface. `dataObject` is now registered in: search-hit hydration (`UrnToEntityMapper`), `EntityTypeMapper`/`EntityTypeUrnMapper`, searchable/autocomplete/browse lists, recommendations, telemetry, forms assignment, operation reporting, editable description, **policy Resource Types** (`PoliciesConfig`), **structured-property "Applies to"**, **view builder**, **Data Product / Application membership** (via `supportedCapabilities`), and the frontend search/autocomplete GraphQL fragments.

### Guard tests — prevent the "forgot to register a new entity" regression class
- **Category A** — `UrnToEntityMapperCompletenessTest`: every search/autocomplete/browse entity **must** be hydratable, so a future entity added to a search list but not the mapper fails CI instead of breaking search at runtime.
- **Category B (registry-aspect-derived)** — `ReportOperationResolverEntityCoverageTest`, `SearchBasedFormAssignmentManagerEntityCoverageTest`, `RecentlyEditedSourceEntityCoverageTest`: each asserts a hardcoded allowlist equals the set of registry entities declaring the gating aspect (`operation` / `forms` / `browsePathsV2 ∧ dataPlatformInstance`), ± a documented exception set.
- Capability-derived surfaces (Add-Assets) need no guard — the `supportedCapabilities()` declaration is the single source of truth.

## Scale

Per-file granularity is intentional: the target files (contract PDFs, slide decks, meeting recordings, diagrams) carry metadata that varies file-by-file within a folder and need to be individually searchable — which folder/collection-level governance can't express (and folder-level is already covered by `Container`). To keep that safe at volume, the connector **streams** emission (bounded memory) and enforces the **`max_data_objects`** cap above; the cardinality risk is a general connector concern (an over-partitioned prefix already produces many dataset entities) rather than something specific to this entity. Genuinely large-scale unstructured ingestion (orders of magnitude beyond typical) remains a separate storage/indexing workstream.

## Testing

- **Unit:** extension/MIME classification; `emit_data_object` aspect emission + the empty-config no-op safety property; routing (structured extensions stay datasets); **volume bounding** (cap stops + warns, streaming, structured-files-unaffected).
- **GraphQL:** `DataObject` type/mapper/privilege wiring; `UpdateDescriptionResolver` routing; the guard tests above; full `datahub-graphql-core` suite green.
- **Integration/smoke:** golden-file test in the S3 suite with mp4/png/pdf fixtures (S3 + GCS); a `dataObject` smoke test (GraphQL get/search); full S3 suite green with zero changes to existing goldens.
- **Manual (local quickstart):** verified end-to-end with synthetic `dataObject`s — search + container Contents, subtype display, policy Resource Type, Data Product / Application membership, and structured-property assignment.

## Notes

- **Purely additive** — no migration; existing ingestions unaffected unless `unstructured_file_extensions` is set.
- This is Phase 1; ADLS Gen2 / Databricks Volumes connectors, sub-file chunks, and lineage/access population are deliberately out of scope — the model is shaped to support them.

## Checklist

- [x] Tests added/updated (unit + GraphQL + integration/smoke + registry-derived guard tests)
- [ ] Docs: connector config documented inline on the new fields; a dedicated usage guide will follow
- [x] No breaking changes (additive) — no `updating-datahub.md` entry needed


sgomezvillamor:
I've had a quick look at the proposed model and have some questions before diving deeper into the implementation.

**Why individual-object granularity?**

I understand the motivation for a new entity type — `Dataset` implies structured/tabular data with a schema, and forcing audio/video/image files into that model may be semantically wrong. That said, `Dataset` doesn't *require* a schema, and its governance aspects (ownership, tags, glossary, domains) work fine without one. The distinguishing characteristic of `Dataset` is that it represents a **collection** — a table, a file set, etc.

`Document` is individual-granularity, but that's justified because the content itself needs to be indexed for search/RAG use cases — the individual unit carries unique semantic value.

For unstructured binary files (audio, video, images), what's the governance use case that requires individual-file granularity rather than collection-level? A `SchemalessDataset` (or a `Dataset` subtype) representing the collection — e.g. "all customer call recordings in s3://bucket/calls/" — would still support owners, tags, glossary terms, domains, and structured properties, while avoiding the cardinality explosion.

**Scale concern**

With `dataObject` at individual-file granularity, enabling `unstructured_file_extensions: [mp4, wav, png]` on a bucket with eg 1M objects creates 1M DataHub entities. That's a significant and unguarded impact on:
- GMS write throughput during ingestion
- Elasticsearch index size and search performance
- UI responsiveness across search, browse, and lineage surfaces

The PR doesn't discuss upper bounds, load testing at scale, or a `max_data_objects` safety cap. Before merging, I'd want to understand: what's the expected scale for target use cases?

Happy to discuss — just want to make sure the model decision is intentional before the implementation lands.

max-datahub:
Hey @sgomezvillamor, thanks for the feedback!

I mostly agree on the scale concerns, but I believe the case for a new entity type is strong enough (also based on internal discussions that I will loop you in separately).
Here is how I would respond to your points: 
> These are the right questions to settle before the implementation lands. Two parts: the granularity/model decision, and scale.
> 
> ### Why individual-object granularity
> 
> The governance need is **per-file metadata that varies _within_ a folder, plus individual discoverability**. The files we're targeting are things like contract PDFs, slide decks, meeting recordings, and architecture diagrams. Within a single bucket/prefix these are heterogeneous: one contract is under legal hold and another isn't; one recording contains PII and another doesn't; a specific diagram needs an owner, domain, and structured properties that don't apply to its neighbors. Users also need to find and link *the* specific file, not just the folder it lives in.
> 
> Collection-level governance can't express that heterogeneity — a single entity for "everything in `s3://bucket/contracts/`" yields one set of tags/owners/properties for the whole set. And the folder-level case is **already covered today by `Container`**: S3 prefixes are containers and carry owners/tags/glossary/domains/structured properties. So a user who only needs "manage the folder" doesn't need this entity at all — it's specifically for the per-file layer that containers can't represent.
> 
> On `Dataset` specifically: a Dataset semantically connotes a **structured/tabular asset** — even though a schema is technically optional, the model's contract is "structured data," and a *schemaless dataset* for an MP4 or PNG is semantically incoherent. `Document` is the other individual-granularity entity, but it earns that granularity by **storing and indexing text content** for search/RAG (`DocumentContents.text` is required); binary files have no such content, and the intent here is **metadata-by-reference** — DataHub never stores the bytes. So neither existing entity fits cleanly, which is why this is a thin, content-by-reference entity that reuses all the standard governance aspects rather than a subtype of either.
> 
> It's also designed to self-nest (`parentDataObject`) so finer-grained sub-file units can attach later without a model change — a deliberate forward-looking choice.
>
> ### Scale
> 
> Worth separating the **model** from the **connector** here:
> 
> - **The model adds zero entities** until something ingests them — so the cardinality question is about the connector's emission behavior, not the entity type.
> - **This cardinality risk isn't new or unique to this entity.** The existing S3 source can already emit millions of *dataset* entities from an over-partitioned prefix, and an over-engineered DB schema does the same. It's a general ingestion concern.
> - **What we'll harden — and the file path is actually easier to make safe than the structured path**, because there's no schema inference, SQL parsing, or lineage to hold in memory: stream/batch entity emission so memory stays bounded regardless of object count; add a configurable cap that fails with a clear error (no silent truncation) and a warning when a path-spec would exceed a threshold; and document the bounds. Emission stays opt-in (`unstructured_file_extensions` is empty by default), and folder-level `Container` governance remains the answer for anyone who doesn't need per-file granularity.
> 
> Very-large-scale unstructured ingestion (orders of magnitude beyond typical) is a separate storage/indexing workstream we're scoping independently — this PR's scope is the model plus a safe, bounded connector path.
> 
> Happy to add the cap + streaming and the bounds docs to this PR so the safety story is concrete rather than asserted. Does that resolve the concern?

sgomezvillamor:
I agree that Dataset and Document don't fit. That said, I still have concerns on the granularity decision, and I think it's worth slowing down here because this sets a precedent for how DataHub models unstructured assets going forward.

**Are we optimizing for the edge case?**

The scenario you describe — heterogeneous files in a flat prefix, each needing independent governance — is real, but I'd argue it's the exception, not the norm. Many organizations do manage these assets with structure: contracts live in /legal/contracts/2025/, recordings in /calls/team-a/, diagrams in /arch/service-x/. In those cases, the folder hierarchy already carries meaningful governance context, and Container captures it today.

If we introduce dataObject as the standard model, organizations with well-governed storage get no benefit from the individual granularity — they're just forced to manage 1M entities instead of a handful of containers. The corner case becomes the default.

What if we introduced an UnstructuredDataset (or similar) at collection level, and treated individual-file granularity as an opt-in escape hatch for the heterogeneous case? Users with well-governed folder structures get a clean collection entity. Users with the "everything in one bucket" problem can, if truly necessary, map each file to its own UnstructuredDataset — but that's the edge case paying its own cost, not the default.

The key question may be: for users with well-governed folder structures, is there concrete value in per-file entities beyond what Container or UnstructuredDataset would provide?

**Connector coupling**

A separate concern: should this live in the existing S3 connector at all? Structured and unstructured ingestion serve different users with different expectations. A dedicated s3-unstructured (or object-store-unstructured) connector would make the scope explicit, keep the existing connector's complexity contained, and allow the new one to evolve independently (its own config, filtering, scale controls, and eventually chunk-level lineage). Coupling them together via a config flag feels like a convenience shortcut that accumulates debt over time.

**Overlap with ML training data**

A related modeling tension worth raising: today, a collection of audio files used to train an ML model is represented as a Dataset — because that's how we express "a corpus of data consumed by a process." If dataObject becomes the standard representation of individual audio files, what's the right entity for the training corpus? Do we keep using Dataset (now inconsistently sitting alongside dataObject for the same underlying files)? Do we expect users to aggregate dataObject instances into a Dataset via lineage? This adjacency to ML asset modeling should be thought through before we land the individual-file abstraction, otherwise we risk ending up with two overlapping representations for the same physical reality.

metadata-models/src/main/pegasus/com/linkedin/dataobject/ParentDataObject.pdl
```diff
@@ -0,0 +1,13 @@
+namespace com.linkedin.dataobject
+
+import com.linkedin.common.Urn
+
+/**
+ * The hierarchical parent data object (e.g. the file a chunk belongs to).
+ */
+@Aspect = { "name": "parentDataObject" }
+record ParentDataObject {
+  @Relationship = { "name": "IsChildOf", "entityTypes": ["dataObject"] }
```
david-leifker, line 10: Don't add relationship annotations for entities of this scale without special consideration in the graph indices

metadata-models/src/main/pegasus/com/linkedin/dataproduct/DataProductProperties.pdl
```diff
@@ -39,7 +40,7 @@ record DataProductProperties includes CustomProperties, ExternalReference  {
   @Relationship = {
     "/*/destinationUrn": {
       "name": "DataProductContains",
-      "entityTypes": [ "dataset", "dataJob", "dataFlow", "chart", "dashboard", "notebook", "container", "mlModel", "mlModelGroup", "mlFeatureTable", "mlFeature", "mlPrimaryKey", "document" ],
+      "entityTypes": [ "dataset", "dataJob", "dataFlow", "chart", "dashboard", "notebook", "container", "mlModel", "mlModelGroup", "mlFeatureTable", "mlFeature", "mlPrimaryKey", "document", "dataObject" ],
```
david-leifker, line 43: Data Products doesn't scale to the required scale of dataObject

david-leifker: I would leave this off and consider relying on containers which contain dataObject only, but its still a concern.

PR #17961, opened 2026-06-19
feat(databricks): optimize usage extraction via system-table lineage

askumar27:
## Summary

Speed up Unity Catalog usage extraction by joining system query history with table lineage to skip sqlglot parsing where possible, adding Snowflake-parity catalog pushdown, and fetching SELECT-only queries when operational stats are disabled.

## Motivation

Unity Catalog usage extraction was a bottleneck for large workspaces because every query in history was parsed with sqlglot, even when Databricks system tables already expose table-level read/write lineage. This change reduces CPU time and warehouse load during usage ingestion while preserving metadata correctness via a sqlglot fallback path.

## Changes Overview

**New Features:**
- Join `system.query.history` with `system.access.table_lineage` and emit `PreparsedQuery` when lineage is available (Snowflake access-history pattern)
- Optional catalog pushdown via `push_down_database_pattern_access_history` (maps to existing `catalog_pattern`, Snowflake database-parity)
- `skip_sqlglot_when_system_table_lineage_missing` to skip (instead of sqlglot-parse) queries with no matching `system.access.table_lineage` rows on the system-tables path
- SELECT-only query fetch when `include_operational_stats: false` (BigQuery-style read-side optimization)
- `include_column_usage_stats` (default `false`) to force full sqlglot parsing and emit column-level usage statistics (`fieldCounts`); bypasses the preparsed path and overrides push-down/skip flags when enabled
- Ingestion report counters and summary warnings for preparsed fallback, missing/skipped lineage, unresolvable table names, and lineage row read errors

**Modifications:**
- `UnityCatalogUsageExtractor` routes queries through preparsed, observed (sqlglot), or skipped paths based on system-table lineage availability and config
- System-table query fetch streams and groups multi-row lineage results per `statement_id`
- Statement-type filtering applied consistently to both system-table SQL and warehouse API fallback paths
- INFO-level routing summary on the system-tables path; quoted identifiers parsed via `split_databricks_identifier`
- Unity Catalog connector docs for usage source selection, pushdown, skip flag, column usage stats opt-in, and system-table permissions

**Bug Fixes:**
- Row parse errors during lineage grouping no longer corrupt lineage for subsequent statements
- Warning context no longer re-raises when `statement_id` is unreadable on a failed row
- Catalog pushdown RLIKE filters honor `catalog_pattern.ignoreCase`
- Required query-history row fields validated before grouping; parse failures summarized once per fetch
- Fetch failures return before emitting usage workunits (no partial usage on incomplete history reads)
- Catalog pushdown RLIKE patterns are bound as SQL parameters (not string-interpolated), preventing injection and special-character breakage
- Usage routing warnings use stable message keys with context for report deduplication

**Refactoring:**
- Extracted `usage_statement_types()` helper and catalog filter SQL builders in the proxy layer
- Lineage table resolution uses public `SchemaResolver.resolve_table_parts()` instead of private internals
- Documented `SchemaResolver.resolve_table_parts()` contract (URN is best-effort; `SchemaInfo` is the resolution signal); DRY'd uniform usage-lineage warning blocks

## Architecture/Design Notes

- **Preparsed vs sqlglot:** When system-table lineage is present and at least one lineage table name resolves to an ingested dataset (`SchemaInfo` present in the schema resolver), queries are fed to `SqlParsingAggregator.add_preparsed_query()`; otherwise they fall back to `ObservedQuery`. The preparsed path is table-level only (no column `fieldCounts`).
- **Column usage stats opt-in:** `include_column_usage_stats: true` forces every query through sqlglot on the system-tables path, restoring column-level usage statistics at the cost of bypassing the preparsed shortcut; push-down and skip flags are ignored.
- **Skip path:** `skip_sqlglot_when_system_table_lineage_missing` applies only when lineage rows are absent for a statement; URN-resolution failures still fall back to sqlglot.
- **Defense in depth:** Catalog pushdown narrows SQL fetch volume; Python-side `is_allowed_table` still scopes emitted usage to ingested tables.
- **Pushdown semi-join:** When enabled, only statements with at least one lineage row in the time window are fetched; lineage-less queries are omitted entirely (documented, not sqlglot-fallbacked).
- **Operational stats gate:** `include_operational_stats` controls both operational aspect emission and whether DML/DDL statement types are fetched from history.
- **Backward compatible:** All new config is additive with safe defaults (`push_down_database_pattern_access_history: false`, `skip_sqlglot_when_system_table_lineage_missing: false`, `include_column_usage_stats: false`; existing statement-type behavior preserved when ops stats enabled).

## Testing

- **Unit tests:** 152 tests across `test_unity_catalog_usage_join.py`, `test_unity_catalog_proxy.py`, and `test_unity_catalog_config.py` covering streaming, lineage grouping, preparsed path, sqlglot fallback, skip flag, `include_column_usage_stats` routing and flag precedence, catalog pushdown (including ignoreCase and parameterized bind params), injection-style catalog patterns, SELECT-only filtering, multi-target fan-out, partial URN resolution, the `SchemaInfo`-based resolution contract, fingerprint fallback, row-field-error row-skip (no statement split), parse-error recovery, fetch-failure gating, over-restrictive-pushdown and all-unparseable diagnostics, system-table flag-misuse and column-stats override warnings, and mixed routing counters
- **Lint:** `./gradlew :metadata-ingestion:lintFix` passes
- **Live validation (summary):** Compared develop-mode build against `acryl-datahub` 1.6.0.5rc2 using file-sink ingestion with `include_operational_stats` toggled on/off over a 7-day window on a real Unity Catalog workspace:
  - **End-to-end:** dev branch ~37% faster with ops on (45s vs 72s wall time); usage stage ~14% faster (21.5s vs 25.1s)
  - **SELECT-only:** dev ops-off fetched 463 fewer queries than ops-on (2,177 vs 2,640); baseline rc2 still fetched all statement types
  - **Preparsed path:** queries bypass sqlglot only when their `system.access.table_lineage` tables are themselves ingested by the recipe (per the post-review `SchemaInfo` resolution contract). On a single-`main`-catalog test recipe, ~16 of ~2,925 queries (~0.5%) qualify; ~286 with lineage pointing at out-of-catalog tables (e.g. `acryl_demo.*`) correctly fall back to sqlglot. An earlier ~317 (~12%) preparsed count before the SchemaInfo fix counted synthesized URNs and is not representative.
  - **Column usage stats:** toggling `include_column_usage_stats` on/off over the same window — off: preparsed=16 / sqlglot=2,909; on: sqlglot=2,925 (100%), 0 failures; usage MCP output equivalent (table-level stats identical; `fieldCounts` empty for both on this workspace's query patterns)
  - **Metadata equivalence:** `datahub check metadata-diff` between dev and baseline for matching ops settings showed no semantic usage differences (only attribution timestamp drift from re-ingestion)
  - **Post-review re-run:** re-ran the latest branch (system-tables path, 7-day window, file sink, ops on/off); both pipelines finished with 0 failures, with `num_lineage_row_field_read_errors`, `num_queries_preparsed_fingerprint_fallback`, and `num_queries_missing_info` all 0 and out-of-catalog lineage tables routed to sqlglot fallback as expected

## Impact Assessment

**Affected Components:** Unity Catalog ingestion — usage extraction path only (`usage.py`, `proxy.py`, `proxy_types.py`, `schema_resolver.py`, config/report types)

**Breaking Changes:** None

**Performance Impact:** Positive for workspaces using system-table usage path; largest gains when operational stats are enabled (preparsed + full history) and additional fetch reduction when ops stats are disabled (SELECT-only)

**Risk Level:** Medium — touches hot usage path for large deployments, but sqlglot fallback preserves prior behavior when lineage is absent or unresolvable

## Deployment Notes

- Optional recipe knobs:
  - `push_down_database_pattern_access_history: true` with `catalog_pattern` to push catalog filtering into `table_lineage` SQL (note semi-join excludes lineage-less statements)
  - `skip_sqlglot_when_system_table_lineage_missing: true` to avoid sqlglot cost for queries Databricks did not record in `system.access.table_lineage`
  - `include_column_usage_stats: true` to force full sqlglot parsing and emit column-level usage statistics (`fieldCounts`); slower, overrides push-down/skip flags
- No feature flags or migrations required
- Monitor ingestion report warnings/counters for preparsed fallback, missing/skipped lineage, and unresolvable table names

## References

- Related prior work: #17954 (SqlParsingAggregator usage migration)


metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
```diff
@@ -629,69 +734,177 @@ def get_query_history_via_system_tables(
...
             WHERE
-                start_time >= %s
-                AND end_time <= %s
-                AND execution_status = 'FINISHED'
-                AND statement_type IN ({statement_type_filter})
-            ORDER BY start_time
+                qh.start_time >= %s
+                AND qh.end_time <= %s
+                AND qh.execution_status = 'FINISHED'
+                AND qh.statement_type IN ({statement_type_filter}){catalog_pushdown_clause}
```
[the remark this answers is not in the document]
askumar27: @AikidoSec ignore: fixed

treff7es:
Took a careful look at this and validated the approach against a live Unity Catalog workspace (7-day window, ~3.7k finished statements). Overall this is a solid, well-built optimization — the `PreparsedQuery`-from-system-tables direction is exactly the Snowflake access-history pattern, the SQL is properly parameterized, the sqlglot fallback + report counters are clean, and the defaults are safe. A few findings worth weighing before merge, mostly about how much this actually buys on real Databricks data and a fidelity tradeoff on the preparsed path.

### 1. Coverage of the system-table path is inherently low (~9–13%)

Measured on the live workspace (statements joined to `system.access.table_lineage` by `statement_id`):

| Statement type | total | with `table_lineage` |
|---|---:|---:|
| SELECT | 2,082 | 268 (13%) |
| CREATE | 342 | 0 |
| DESCRIBE/SET/ANALYZE/REFRESH/DROP/OTHER | ~1,240 | 0 |
| **All finished statements** | **3,667** | **320 (8.7%)** |

`statement_id` lineage only exists for warehouse/serverless queries that emit a lineage event, so ~87–91% of statements have no row in either lineage table and stay on the sqlglot path. So this is a targeted accelerator for a minority of queries (consistent with the PR's own ~12% preparsed figure), not a primary path — worth setting expectations accordingly.

### 2. `column_lineage` doesn't widen coverage — only fidelity

| source | distinct statements covered |
|---|---:|
| `table_lineage` | 320 |
| `column_lineage` | 194 |
| union | 320 |
| in `column_lineage` not in `table_lineage` | **0** |

`column_lineage` is a strict subset of `table_lineage` (same capture mechanism, finer grain). So no combination gets past the ~13% ceiling.

### 3. Preparsed path drops column-level usage + query subjects (fidelity)

The `PreparsedQuery` is built from `table_lineage` only, leaving `column_usage=None` and `column_lineage=None`. Both are consumed downstream (and on by default for UC):

- `column_usage` → `fields` in `aggregate_event` → **`fieldCounts`** in `DatasetUsageStatistics` (`sql_parsing_aggregator.py:982,996`)
- `column_usage`/`column_lineage` → schemaField subjects in **`querySubjects`** via `get_subjects(include_fields=…)` (`:177,:190`); `generate_query_subject_fields` defaults `True` and isn't overridden in `usage.py`.

So every query routed to preparsed loses its per-column usage breakdown and column-level query subjects (the all-observed/sqlglot path computes both). Notably, Databricks *does* expose column-level read access — `system.access.column_lineage` carries source-only rows (target NULL) for SELECTs — so this is recoverable by also joining `column_lineage` (source-only → `column_usage`, source→target → `column_lineage`). Caveat: that only helps the ~5% of statements (194/3,667) that have column lineage; the rest still need sqlglot.

### 4. `skip_…` + `push_down…` can silently drop most usage

Given ~87–91% of statements have no `table_lineage` row, both `skip_sqlglot_when_system_table_lineage_missing` and the `push_down_database_pattern_access_history` semi-join would discard the large majority of usage/operations on a workspace like this (the semi-join drops lineage-less statements even with the skip flag off). Defaults-off is the right call; suggest making the docs very explicit that enabling either narrows usage to the lineage-bearing minority — it's a bigger effect than "skip parsing for a few queries."

### Snowflake comparison (for context)

Snowflake `ACCESS_HISTORY` is a single view carrying column-level **reads** (`DIRECT_OBJECTS_ACCESSED[].columns`) and **writes** (`OBJECTS_MODIFIED[].columns[].directSources`) with ~full coverage, so its connector makes `PreparsedQuery` the *primary* path with full column fidelity. Databricks splits the same info across `table_lineage` + `column_lineage` keyed by `statement_id` and only for ~13% of statements — so on Databricks the system-table path is an accelerator layered on top of sqlglot, not a replacement.

### Suggestions (non-blocking)

1. Join `system.access.column_lineage` to populate `column_usage`/`column_lineage` on the preparsed path, so `fieldCounts` and column-level `querySubjects` aren't silently dropped for the queries it accelerates.
2. Make the docs emphatic that `skip_…`/`push_down…` trade usage *coverage* for speed (most statements lack lineage rows), not just parse time.

None of this is a correctness bug — the defaults are safe and the fallback is intact. These are fidelity/coverage tradeoffs that I think deserve to be conscious + documented.

metadata-ingestion/src/datahub/ingestion/source/unity/usage.py
```diff
@@ -57,11 +76,292 @@ def _build_aggregator(
...
+        if parts is None or len(parts) != 3:
+            logger.debug("Skipping unexpected table full name: %s", full_name)
+            self.report.num_lineage_tables_unresolvable += 1
+            self.report.lineage_tables_unresolvable_sample.append(full_name)
+            return None
+        catalog, schema, table = parts
+        urn, _ = self.schema_resolver.resolve_table_parts(
+            database=catalog, db_schema=schema, table=table
+        )
+        if urn is None:
```
treff7es: 🔴 **Critical — this fallback is dead code, and the unresolved-table safety net never fires.**

`resolve_table_parts()` → `resolve_table()` returns `Tuple[str, Optional[SchemaInfo]]` where the **URN (1st element) is never `None`** — on a cache/graph miss it returns a *synthesized* URN + `None` (`schema_resolver.py` final `return (urn_lower if ... else urn), None`). So:

- This `if urn is None:` branch is unreachable → `mypy --warn-unreachable` (CI runs it on src) will likely fail here.
- The PR's documented "URN-resolution failures still fall back to sqlglot" **can't happen**: a `catalog.schema.table` not present in DataHub still yields a URN, so it's treated as resolved and emitted as a `confidence_score=1.0` `PreparsedQuery` pointing at a possibly-nonexistent dataset. `num_queries_preparsed_fallback_to_sqlglot` won't trigger for not-in-catalog tables, and `num_lineage_tables_unresolvable` undercounts (only the malformed-name case at L113 counts).

**Fix:** use the 2nd element as the resolution signal:
```python
urn, schema_info = self.schema_resolver.resolve_table_parts(...)
if schema_info is None:
    self.report.num_lineage_tables_unresolvable += 1
    self.report.lineage_tables_unresolvable_sample.append(full_name)
    return None
return urn
```
(Or, if emitting lineage to not-yet-ingested URNs is intentional, delete the dead branch + counters and document that — the current code is internally contradictory.)

metadata-ingestion/src/datahub/sql_parsing/schema_resolver.py
```diff
@@ -232,6 +232,18 @@ def resolve_table(self, table: _TableName) -> Tuple[str, Optional[SchemaInfo]]:
 
         return (urn_lower if self._prefers_urn_lower() else urn), None
 
+    def resolve_table_parts(
```
treff7es, line 243: ℹ️ Related to the Critical in `usage.py`: please document this contract — the returned URN is **best-effort and always non-None** (synthesized on a miss); callers must test the *second* element (`SchemaInfo is None` ⇒ not found), not the URN. The current docstring ("Resolve a table URN…") reinforces the wrong reading and is what the consumer got wrong.

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
```diff
@@ -629,69 +734,177 @@ def get_query_history_via_system_tables(
...
+                            row, "source_table_full_name", self.report
+                        )
+                        if source_name:
+                            source_names.add(source_name)
+                        target_name = _optional_row_field(
+                            row, "target_table_full_name", self.report
+                        )
+                        if target_name:
+                            target_names.add(target_name)
+                    except _ROW_PARSE_EXCEPTIONS:
```
treff7es: 🟡 **High — a single lineage-row error splits the statement and inflates counts.**

On a row-field error within an already-started group this does `yield from _flush_current()` then `current_statement_id = None`. Because the statement boundary is reset, the *next* row carrying the **same** `statement_id` is treated as a brand-new statement and re-parsed — so the statement is **emitted twice, each with a partial set of source/target tables**, and `query_count` is double-counted. Downstream `_to_preparsed_queries` then emits incomplete upstreams/downstreams with no warning that the lineage is partial.

**Fix:** on a field-read error within an established statement, increment `num_lineage_row_field_read_errors` and `continue` (keep accumulating into the same group); don't flush+reset the boundary.

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
```diff
@@ -184,6 +185,107 @@ class TableLineageInfo:
...
+        parts.append(f"({source_filter} AND tl2.source_table_catalog IS NOT NULL)")
+        bind_params.extend(source_params)
+    if target_filter != "TRUE":
+        parts.append(f"({target_filter} AND tl2.target_table_catalog IS NOT NULL)")
+        bind_params.extend(target_params)
+    sql = parts[0] if len(parts) == 1 else f"({' OR '.join(parts)})"
+    return sql, bind_params
+
+
+_ROW_PARSE_EXCEPTIONS = (AttributeError, KeyError, TypeError, ValueError)
```
treff7es: 🟡 **High — exception tuple diverges from what's actually caught, mis-bucketing errors.**

This tuple `(AttributeError, KeyError, TypeError, ValueError)` guards the row loop, but `_optional_row_field` only catches `(AttributeError, ValueError)`. A `KeyError`/`TypeError` reaching the outer handler is counted as `num_queries_missing_info` (a *query-parse* counter) rather than `num_lineage_row_field_read_errors` (the *field-read* counter) — wrong bucket — and `TypeError`/`KeyError` silently swallow genuine logic bugs in this hot loop at `debug` level.

**Fix:** narrow the outer handler to the specific read it guards (`row.statement_id`), and keep `_optional_row_field` the owner of field-level catches; log at `warning`, not `debug`.

metadata-ingestion/src/datahub/ingestion/source/unity/config.py
```diff
@@ -384,6 +384,31 @@ class UnityCatalogSourceConfig(
         ),
     )
 
+    push_down_database_pattern_access_history: bool = Field(
```
treff7es, line 387: 🟡 **High — the "only on the system-tables path" invariant is prose-only.**

This flag (and `skip_sqlglot_when_system_table_lineage_missing` below) is documented as a no-op unless usage is fetched via system tables, but nothing warns if a user sets it with `usage_data_source: API` — they get silent no-op behavior.

**Fix:** add a single `@model_validator(mode="after")` emitting a `report.warning` (not an error — don't break recipes) when these flags are set but the effective usage path isn't system tables.

metadata-ingestion/src/datahub/ingestion/source/unity/usage.py
```diff
@@ -167,22 +444,45 @@ def get_usage_workunits(
...
-                        hint = (
-                            "verify CAN_MANAGE privilege on the SQL warehouse "
-                            "and that the time window covers recent activity"
+                        if self._use_system_tables_join():
+                            hint = (
+                                "verify SELECT privilege on system.query and system.access "
+                                "and that the time window covers recent activity"
+                            )
+                            if (
+                                self.config.push_down_database_pattern_access_history
```
treff7es, line 514: 🟡 **High (test gap) — this user-facing diagnostic is untested.**

The catalog-pushdown "over-restrictive" hint here, and the `num_queries_missing_info > 0` "history existed but was unparseable" warning at ~L450, are exactly the messages operators rely on when usage comes back empty — but no unit test exercises either branch. Please add: system-tables path + `push_down_database_pattern_access_history=True` + non-null `catalog_pattern` + proxy returns `[]` ⇒ assert the over-restrictive hint fires; and an all-rows-unparseable case ⇒ assert the parse-failure warning fires and "No queries found for usage" does not.

metadata-ingestion/src/datahub/ingestion/source/unity/usage.py
```diff
@@ -57,11 +76,292 @@ def _build_aggregator(
...
+            self.report.num_lineage_tables_unresolvable,
+        )
+
+    def _to_preparsed_queries(self, query: Query) -> List[PreparsedQuery]:
+        upstreams = self._resolve_table_urns(query.source_table_full_names)
+        targets = self._resolve_table_urns(query.target_table_full_names)
+        ts = self._normalize_timestamp(query.start_time)
+        user = self._user_urn(query)
+        query_type = self._query_type(query.statement_type)
+        query_id = get_query_fingerprint(query.query_text, self.platform, fast=True)
```
treff7es: 🟠 **Medium — a fingerprint hiccup drops a query that *had* valid system-table lineage.**

`get_query_fingerprint(...)` is unguarded; malformed SQL can still throw inside sqlglot tokenization even in `fast` mode, propagating to the per-query handler in `get_usage_workunits` → the query is counted as `num_queries_dropped`. That defeats the whole point of the preparsed path: a query whose lineage was already resolved from the system table is lost to a fingerprinting error.

**Fix:** wrap it; on failure fall back to a deterministic id from `query.statement_id`/`query_id` (always present on the system-tables path) so the resolved lineage is still emitted.

metadata-ingestion/src/datahub/ingestion/source/unity/usage.py
```diff
@@ -57,11 +76,292 @@ def _build_aggregator(
...
+                query=query.query_text,
+                timestamp=self._normalize_timestamp(query.start_time),
+                user=self._user_urn(query),
+                default_db=default_db,
+                default_schema=None,
+            )
+        )
+        self.report.num_queries_observed_sqlglot += 1
+
+    def _report_usage_lineage_warnings(self) -> None:
```
treff7es, line 371: ℹ️ **Suggestion (DRY).** These are four near-identical `if counter > 0: report_warning(title=…, message=…, context=f"count={counter}")` blocks. A small data-driven loop over `(counter, title, message)` tuples (with the unresolvable case handling its own context) would cut ~50 lines to ~15 and make the parallelism obvious.

askumar27:
Thanks for the thorough review @treff7es — all findings are addressed. Summary below; I've also replied inline on each thread.

| Finding | Resolution | Commit |
|---|---|---|
| 🔴 Critical — dead fallback / unresolved-table net never fires (`usage.py`) | `_full_name_to_urn` now uses `SchemaInfo` (2nd element) as the resolution signal; not-in-catalog tables return `None` and fall back to sqlglot, and `num_lineage_tables_unresolvable` counts them | `09162e2` |
| 🟠 Medium — unguarded `get_query_fingerprint` drops a query with valid lineage | `_query_fingerprint` wrapper falls back to a deterministic `unity-stmt-<id>` + new `num_queries_preparsed_fingerprint_fallback` counter | `09162e2` |
| 🟡 High — single row error splits a statement / double-emits | field-read error now `continue`s without flushing the boundary | `98c1d83` |
| 🟡 High — exception tuple mis-buckets errors | outer handler narrowed to match `_optional_row_field`; field reads vs query-parse bucket separately | `98c1d83` |
| 🟡 High — no-op flags off the system-tables path are silent | `@model_validator` warns (not errors) when the flags are set but usage won't use system tables | `04d67d1` |
| 🟡 High — over-restrictive-pushdown / all-unparseable diagnostics untested | added both unit tests | `ccf3fb1` |
| ℹ️ Resolver URN contract undocumented | documented `resolve_table` / `resolve_table_parts`: URN is best-effort/always non-None; check `SchemaInfo` | `ea50704` |
| ℹ️ DRY the 4 warning blocks | collapsed the uniform warnings into a data-driven loop (unresolvable case keeps its sample context) | `ea50704` |

**Live validation** (system-tables usage path, 7-day window, run on the latest branch — file sink, both with and without operational stats): both pipelines finished with **0 failures**. On a ~2.9K-statement window the routing was 15 preparsed-from-lineage and **286 not-in-catalog lineage tables that now correctly fall back to sqlglot** — exactly the case the critical fix targets (previously these were emitted as `confidence=1.0` lineage to synthesized URNs). `num_lineage_row_field_read_errors`, `num_queries_preparsed_fingerprint_fallback`, and `num_queries_missing_info` were all `0`, and `mypy --warn-unreachable` confirms the formerly-dead branch is now reachable.

metadata-ingestion/src/datahub/ingestion/source/unity/usage.py
askumar27: Fixed in `09162e2` — `_full_name_to_urn` now uses the 2nd element (`SchemaInfo`) as the resolution signal: a not-in-catalog `catalog.schema.table` returns `None` and falls back to sqlglot, and `num_lineage_tables_unresolvable` counts it. `mypy --warn-unreachable` confirms the branch is now reachable. Live run: of a ~2.9K-statement window, 286 not-in-catalog lineage tables now correctly fall back instead of emitting `confidence=1.0` lineage to synthesized URNs.

metadata-ingestion/src/datahub/sql_parsing/schema_resolver.py
askumar27: Done in `ea50704` — documented the contract on `resolve_table` and `resolve_table_parts`: the returned URN is best-effort and always non-None (synthesized on a miss); callers must use `SchemaInfo` (2nd element) as the resolution signal, never `urn is None`.

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
askumar27: Fixed in `98c1d83` — a field-read error on a row no longer flushes/resets the statement boundary; it increments `num_lineage_row_field_read_errors` and `continue`s, so the statement stays in one group and is not emitted twice with partial lineage.

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
askumar27: Fixed in `98c1d83` — the outer handler is narrowed to `_ROW_FIELD_READ_EXCEPTIONS` (kept identical to what `_optional_row_field` catches), so field reads bucket into `num_lineage_row_field_read_errors`; query-construction failures bucket separately via `_ROW_PARSE_EXCEPTIONS` into the query-parse counter.

metadata-ingestion/src/datahub/ingestion/source/unity/config.py
askumar27: Fixed in `04d67d1` — added an `@model_validator(mode="after")` that emits a warning (not an error, so recipes still run) when `push_down_database_pattern_access_history` / `skip_sqlglot_when_system_table_lineage_missing` are set but usage will not use system tables. Covered by tests in `ccf3fb1`.

metadata-ingestion/src/datahub/ingestion/source/unity/usage.py
askumar27: Added in `ccf3fb1` — `test_no_queries_hint_mentions_overrestrictive_pushdown` (system-tables path + pushdown + non-null `catalog_pattern` + empty proxy ⇒ asserts the over-restrictive hint fires) and `test_all_unparseable_history_warns_and_suppresses_no_queries` (all rows unparseable ⇒ asserts the parse-failure warning fires and "No queries found" does not).

metadata-ingestion/src/datahub/ingestion/source/unity/usage.py
askumar27: Fixed in `09162e2` — `_query_fingerprint` wraps `get_query_fingerprint`; on failure it falls back to a deterministic `unity-stmt-<id>` and increments `num_queries_preparsed_fingerprint_fallback`, so a query whose lineage was already resolved from system tables is no longer dropped on a tokenizer hiccup.

metadata-ingestion/src/datahub/ingestion/source/unity/usage.py
askumar27: Done in `ea50704` — the three uniform count-only warnings now come from a data-driven `(count, title, message)` loop; the unresolvable-tables case keeps its own sample-augmented context.

metadata-ingestion/src/datahub/ingestion/source/unity/proxy.py
```diff
@@ -629,69 +740,196 @@ def get_query_history_via_system_tables(
...
             WHERE
-                start_time >= %s
-                AND end_time <= %s
-                AND execution_status = 'FINISHED'
-                AND statement_type IN ({statement_type_filter})
-            ORDER BY start_time
+                qh.start_time >= %s
+                AND qh.end_time <= %s
+                AND qh.execution_status = 'FINISHED'
+                AND qh.statement_type IN ({statement_type_filter}){catalog_pushdown_clause}
```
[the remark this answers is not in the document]
askumar27: @AikidoSec ignore: false positive — the values interpolated at this line are a fixed internal enum allowlist (QueryStatementType members), not user input. The only user-controlled values in this query (catalog_pattern RLIKE patterns) are bound as query parameters (%s / bind params), not concatenated.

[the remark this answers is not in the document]
askumar27: @AikidoSec ignore: fixed

treff7es (approved):

Issue #14989, opened 2025-10-13
acryl-spark-lineage jar contains multiple copies of dependencies


umartin:
**Describe the bug**
acryl-spark-lineage jar contains both relocated dependencies and non-relocated dependencies. Sometimes with multiple versions. The non-relocated dependencies can cause class loading conflict with other spark extensions.

**To Reproduce**
Three copies of antlr is packaged in the jar.
```sh
$ zipinfo -1 metadata-integration/java/acryl-spark-lineage/build/libs/acryl-spark-lineage_2.12-1.3.0rc1-SNAPSHOT.jar | grep  'antlr*\/$'
org/antlr/
io/acryl/shaded/st4hidden/org/antlr/
antlr/
META-INF/maven/org.antlr/
```
A more complete listing of the jar content
```sh
$ zipinfo -1 metadata-integration/java/acryl-spark-lineage/build/libs/acryl-spark-lineage_2.12-1.3.0rc1-SNAPSHOT.jar | grep  '^[[:alnum:]]*\/[[:alnum:]]*\/$'
io/openlineage/
datahub/spark/
io/datahubproject/
datahub/client/
io/acryl/
org/publicsuffix/
com/linkedin/
datahub/event/
org/apache/
legacyPegasusSchemas/com/
pegasus/com/
com/datahub/
io/confluent/
software/amazon/
common/message/
datahub/spark2/
io/micrometer/
io/github/
nonapi/io/
org/eclipse/
jakarta/json/
com/google/
org/codehaus/
org/antlr/
io/swagger/
com/eclipsesource/
org/reactivestreams/
antlr/actions/
antlr/ASdebug/
antlr/build/
antlr/collections/
antlr/debug/
antlr/preprocessor/
aix/ppc64/
darwin/aarch64/
freebsd/amd64/
freebsd/i386/
linux/aarch64/
linux/amd64/
linux/arm/
linux/i386/
linux/loongarch64/
linux/mips64/
linux/ppc64/
linux/ppc64le/
linux/riscv64/
linux/s390x/
win/aarch64/
win/amd64/
win/x86/
net/jpountz/
org/xerial/

```

**Expected behavior**
Non-relocated dependencies is fine in an executable binary. For a library it is a nightmare since dependencies are hidden from any resolution mechanism in dependency resolvers. Class loading is not deterministic. Antlr 4.5 in acryl-spark-lineage:0.2.18 might get loaded before the antlr version used by delta-lake which causes non recoverable errors.

[...]

PR #17962, opened 2026-06-19
feat(spark-lineage): resolve per-connection platform_instance for cross-platform Spark lineage

treff7es:
## Summary

A single Spark job that reads catalog/warehouse tables from **multiple AWS accounts, regions, or hosts** could only set one global `spark.datahub.metadata.dataset.platformInstance`. Every upstream collapsed onto that one instance, so tables owned by different accounts (or the same account in another region, or different warehouse hosts) got identical/wrong dataset URNs — producing duplicate/ghost entities and lineage edges that dangle because they don't match the URN the upstream's own connector emits.

The connection identity is already present in the OpenLineage namespace (the Glue symlink ARN `arn:aws:glue:{region}:{account}`, `snowflake://{account}`, `postgres://{host}:{port}`, …) but the converter discarded it. This PR resolves `platform_instance`/`env` **per upstream connection**, keyed by that canonical OpenLineage namespace.

## Changes

- **Per-connection instance resolution** in `openlineage-converter`: recover the connection identity (the Glue ARN from the OpenLineage symlink, which is otherwise rewritten to `glue`; the URI authority for other namespaces) and resolve it against a new map, falling back to the existing platform-keyed behavior. The platform is implied by the namespace scheme, so no platform field is needed. **Default behavior is unchanged when the new config is unset.**
- New Spark config `spark.datahub.metadata.dataset.connections.<olNamespace>.{platformInstance,env}`, documented in the spark-lineage README with a per-platform namespace-key reference.
- **Fix:** streaming (Kafka) dataset URNs now carry the configured `platform_instance`, so they match the batch-emitted URNs for the same topic.
- Emit an independent `DataPlatformInstance` aspect on the DataJob (previously only the parent DataFlow carried it).

Notes:

- Glue is keyed by **account + region** (the ARN authority) — account alone is not unique, since the same account can own a Glue catalog in another region.
- Namespaces with no authority (BigQuery — project in the name; S3/GCS — keyed by bucket) keep their existing mechanisms (`path_spec_list`).
- Automatic derivation of the instance from connection identity is intentionally left to a follow-up (a connection→instance registry), so this PR only ever uses operator-supplied values.

## Checklist

- [x] Tests added/updated — `OpenLineageEventToDatahubTest` (per-connection Glue + non-Glue Postgres resolution, config parsing, streaming-Kafka instance, DataJob aspect)
- [x] Docs added/updated — spark-lineage README (new "Cross-platform platform instances (`connections`)" section)
- [ ] No breaking changes — additive config; default behavior unchanged


metadata-integration/java/openlineage-converter/src/main/java/io/datahubproject/openlineage/converter/OpenLineageToDataHub.java
```diff
@@ -222,16 +236,77 @@ private static Optional<DatasetUrn> getDatasetUrnFromOlDataset(
...
+        connectionKeyOverride != null ? connectionKeyOverride : toConnectionKey(namespace);
+    // When URNs are lowercased the dataset namespace is too (above), so the lookup key must match.
+    // The symlink override is derived from the raw symlink namespace, which skips that lowercasing,
+    // so normalize here to keep both paths consistent (matters for mixed-case hosts, e.g. Hive).
+    if (connectionKey != null && mappingConfig.isLowerCaseDatasetUrns()) {
+      connectionKey = connectionKey.toLowerCase();
+    }
+    // Resolve the per-connection detail once; both platform_instance and env key off it.
+    ConnectionInstanceDetail connectionDetail =
+        connectionKey == null ? null : mappingConfig.getConnectionInstanceMap().get(connectionKey);
```
askumar27, line 265: **Observability (should-fix):** when `connectionKey` doesn't exactly match a configured entry (trailing slash, omitted port, `sqlserver://` vs `mssql://`, case when `lowerCaseUrns=off`, or an authority-less namespace like `hive`/`bigquery`), this silently returns `null` and falls through to the global `platformInstance`/`env` — so the URN dangles, which is the exact failure mode this feature exists to prevent, with no signal at any log level.

Suggest at minimum:
```java
log.debug("connection lookup: namespace={} -> key={} -> matched={}", namespace, connectionKey, connectionDetail != null);
```
and ideally a once-per-run `log.warn` for configured connection keys that matched zero datasets. This is the highest-value signal to add — without it, "I configured it but lineage still dangles" will be the main support burden.

metadata-integration/java/openlineage-converter/src/main/java/io/datahubproject/openlineage/converter/OpenLineageToDataHub.java
```diff
@@ -222,16 +236,77 @@ private static Optional<DatasetUrn> getDatasetUrnFromOlDataset(
       }
     } else {
       platform = namespace;
+      // Bare-namespace FS datasets (e.g. "file", "dbfs") have no scheme, so they don't flow through
+      // HdfsPathDataset where file_partition_regexp is normally applied. Apply the opt-in regexp
+      // here too — otherwise it is silently ignored for these platforms. Scoped to FS platforms so
+      // non-FS bare namespaces (hive, the symlink-flattened glue) are unaffected.
+      if (HdfsPlatform.isFsPlatformPrefix(platform)
```
askumar27, line 243: This applies `file_partition_regexp` to bare-namespace FS datasets (`file`, `dbfs`) that previously bypassed it. Correctly gated (unset pattern → no change), and arguably a bugfix — but for users who **have** set the regexp, existing `file`/`dbfs` URNs will change (partition stripped) and prior lineage edges can orphan. Please add an entry to `docs/how/updating-datahub.md` documenting this behavior change.

metadata-integration/java/acryl-spark-lineage/build.gradle
```diff
@@ -246,6 +262,41 @@ scalaVersions.each { sv ->
     relocate 'com.google.errorprone', 'io.acryl.shaded.com.google.errorprone'
     relocate 'com.sun.jna', 'io.acryl.shaded.com.sun.jna'
 
+    // Additional relocations to prevent classloading conflicts (issue #14989)
```
askumar27: **Scope/risk:** these ~30 relocations (#14989) land in the production shadowJar consumed by *all* Spark/Databricks/Glue/EMR users, in the same PR as an unrelated `platform_instance` feature. Several are broad (`software.amazon`, `org.apache.kafka`, `org.apache.http`, `io.micrometer`, `io.confluent`, `io.datahubproject`). The `com.linkedin`/`com.datahub` carve-out shows care, but this is only exercised by `check_jar.sh`/smoke tests, not real-runtime downstream integration. Consider splitting #14989 into its own PR so it can be reviewed/rolled back independently — or at least call out the coupling in the description and confirm the shaded jar was validated on a real Databricks/Glue runtime.

metadata-integration/java/acryl-spark-lineage/src/main/java/datahub/spark/conf/SparkConfigParser.java
```diff
@@ -226,6 +229,55 @@ public static String getCommonPlatformInstance(Config datahubConfig) {
...
+                  + "Ignoring this connection's env; its datasets will use the job default fabric.",
+              envValue,
+              namespace,
+              DATASET_CONNECTIONS_KEY,
+              namespace,
+              FABRIC_TYPE_KEY,
+              Arrays.toString(FabricType.values()));
+        }
+      }
+      map.put(namespace, detailBuilder.build());
```
askumar27: nit / silent-config: a typo'd sub-key (e.g. `platform_instance` instead of `platformInstance`) or an otherwise-empty block produces a `ConnectionInstanceDetail` with both fields empty and still gets `put` into the map — a phantom "configured" connection that resolves to nothing, with no warning. Consider `log.warn`-ing (and skipping the `put`) when both `platformInstance` and `env` are absent; it also makes the lookup-miss case above cleaner.

(The `env` validation just above is a model handler, by the way — ERROR severity with the value, namespace, full recipe path, and valid enum set. 👍)

metadata-integration/java/acryl-spark-lineage/src/main/java/datahub/spark/converter/SparkStreamingEventToDatahub.java
```diff
@@ -223,9 +222,15 @@ public static Optional<DatasetUrn> generateUrnFromStreamingDescription(
           return Optional.empty();
         }
       }
+      // Stamp the configured platform_instance so streaming (e.g. Kafka) URNs match the
+      // batch-emitted URNs for the same dataset. Without it, the same topic resolves to two
+      // different URNs and cross-platform lineage dangles.
+      String platformInstance =
+          sparkLineageConf.getOpenLineageConf().getCommonDatasetPlatformInstance();
```
askumar27, line 229: nit: the streaming path uses `getCommonDatasetPlatformInstance()` directly and doesn't consult `connectionInstanceMap`, so per-connection instances won't apply to Kafka streaming URNs (only the global common instance). Fine for this PR's scope — just flagging the asymmetry with the batch path in case it's surprising later.

askumar27 (approved):
Approving — the core design is sound, null-safe, and additive (absent config = byte-identical URN output). Nicely tested with real testcontainers/Spark smoke tests (Postgres JDBC + Glue-via-moto).

A few non-blocking notes inline. The highest-leverage one is **observability on the connection-key lookup** (`OpenLineageToDataHub`): today, if a configured `connections` key never matches an emitted namespace, the URN silently falls back to the global instance and lineage dangles — which is exactly the failure this feature prevents — with no log at any level. A one-line debug log there (and ideally a once-per-run warn for configured keys that matched nothing) would likely eliminate the #1 future support question for this feature.

Other notes: the build.gradle shade relocations (#14989) are a larger blast radius than the feature and arguably belong in their own PR (or at least a runtime-tested callout), and the `file_partition_regexp` change for bare-namespace FS datasets (file/dbfs) is a URN-changing behavior for opted-in users that should get an `updating-datahub.md` entry. None block merge.

PR #17971, opened 2026-06-19
fix(ingest/unity): drain query history before parsing to avoid operation-handle eviction

treff7es:
## Summary

Hardening for the Unity Catalog **usage** path: drain query history into a disk-backed
`FileBackedList` **first**, then route/parse from the buffer — so usage no longer holds the
warehouse cursor open across the slow sqlglot parse.

> **#17961 has merged to master.** This PR now targets `master` and builds on its
> `_add_query_to_aggregator` routing.

## Why

`get_usage_workunits` previously iterated the live warehouse cursor while routing/parsing
each query (`_add_query_to_aggregator` inside the `for query in self._fetch_queries()`
loop). That holds the cursor open for the entire parse, which can take many minutes on a
large window. Databricks can drop a query's **server-side operation handle** mid-fetch (the
maintainers document eviction after ~20–30 min of fetch-inactivity in
[databricks-sql-python#785](https://github.com/databricks/databricks-sql-python/pull/785));
the next fetch then fails with a **non-retryable** `RESOURCE_DOES_NOT_EXIST` ("Command
&lt;id&gt; does not exist").

Fetching is cheap (≈0.1s for ~1k rows); the cost is the parse. Draining first releases the
cursor in seconds, and the buffer is disk-backed so memory stays bounded regardless of
history size. **This is the pattern the Snowflake and BigQuery query extractors already
use** — both fetch the audit log into a `FileBackedList` (`audit_log.sqlite`) before parsing.
This change brings Unity in line with them.

## Honest scope: hardening, not a proven fix

We observed a real `RESOURCE_DOES_NOT_EXIST` failure at ~22 min on a serverless SQL
warehouse, but **could not reproduce it deterministically** — controlled idle holds of
11 min and ~56 min both survived. So the trigger appears intermittent (likely a serverless
warehouse recycle/scale event rather than a simple idle timer). This PR therefore **removes
the connector's exposure** to the whole class of "held-open cursor dies mid-fetch" failures
(operation-handle eviction, idle timeout, mid-fetch recycle) rather than claiming to fix one
specific error. It is low-risk and aligns with existing connectors, so it's worth landing as
defensive hardening.

## What changed

- `usage.py` `get_usage_workunits`: split the single fetch+route loop into **drain** (append
  every `Query` to a `FileBackedList[Query]`, releasing the cursor once the fetch generator
  is fully consumed) → existing fetch-failure / no-queries checks → **parse** (route each
  buffered query through `_add_query_to_aggregator`). Routing-summary and lineage-warning
  reporting moved after the parse loop; `buffered_queries.close()` added to `finally`.
- Counters, routing decisions, and emitted aspects are unchanged — only *when* the cursor is
  released differs.

## Follow-ups from review

- **Idle windows now record zero-usage.** When a query-history read *succeeds* but yields no
  usable queries, `get_usage_workunits` no longer returns early — it emits a diagnostic
  warning and falls through to `auto_empty_dataset_usage_statistics`, stamping a
  current-bucket zero datapoint. This matches Snowflake/BigQuery/Redshift, which call
  `auto_empty` whenever the read succeeded and only short-circuit on a read/permission
  failure. Fetch *failures* still short-circuit with `report_failure` (no false zeros).
  `datasetUsageStatistics` is a timeseries aspect and the empty aspect is UPSERTed, so this
  records a zero datapoint for the window — it does not delete prior usage history.

- **Usage step timers.** Added `usage_query_fetch_timer` (drain) and `usage_parsing_timer`
  (parse) to the report, mirroring BigQuery's `query_log_fetch_timer` / `sql_parsing_sec`.
  This makes the drain-before-parse behavior measurable and surfaces eviction risk (fetch
  time creeping toward parse time).

- **Optional audit-log caching (`local_temp_path`).** Mirrors the Snowflake/BigQuery
  `use_cached_audit_log` optimization. When the hidden, dev/advanced `local_temp_path` is
  set, the drained query history is kept in a named, **window-keyed** SQLite file under
  that directory and reused across runs: the next run over the same window reloads it and
  skips re-fetching, **including after a crash**. The filename is keyed by usage source and
  time window, so a leftover file is never reused for a stale window (closing the
  `# TODO: check if the cached audit log is stale` that Snowflake/BigQuery carry). When
  unset, behavior is unchanged — an ephemeral, self-cleaning buffer with no caching.

## Benchmark &amp; correctness (master vs this branch)

Measured on a live Databricks workspace with identical dependencies — only the connector
source files were swapped, so the code path is the only variable. Usage-only
(lineage/notebooks off), with fixed time windows so both versions fetch the same queries.

**Correctness:** output is equivalent to master — identical aspect-type counts and **zero**
`datasetUsageStatistics` value differences across every emitted usage aspect.

**Memory:** equivalent (~360 MB at ~2.4k queries, ~420 MB at ~16k queries) — the drain
buffer is disk-backed (SQLite), so peak RSS stays bounded as history grows.

**Runtime — fetch vs parse, via the new timers:**

| Window | queries | fetch/drain | parse (sqlglot) |
| --- | --- | --- | --- |
| 7-day | ~2,400 | ~12 s | ~7 s |
| 90-day | ~16,100 | ~15 s | ~68 s |

Parse scales linearly with query volume; fetch/drain stays nearly flat (dominated by the
system-table query, not row count). This quantifies the drain rationale: the warehouse
cursor is held only for the **fetch/drain** window, then released, so the **O(query_count)**
parse runs cursor-free. On the pre-drain (interleaved) path the cursor stays open for fetch
**+** parse (~68 s and growing at 16k queries) — which is what walks into the ~20–30 min
operation-handle eviction window on large histories. (At small volumes the drain is a few
seconds of overhead; the payoff appears as parse time grows.)

## Testing

- New `test_queries_drained_before_parsing` — asserts the fetch generator is fully consumed
  before the first aggregator interaction (and all queries are still routed). Fails on the
  old interleaved code (first route happens mid-fetch).
- New `test_zero_queries_emits_empty_usage_for_idle_window` — a successful empty read stamps
  zero-usage for ingested tables and warns; `test_fetch_failure_*` still asserts **no** usage
  on a fetch failure, so the idle vs failure paths cannot collapse.
- Full unity suite: **152 passed, 1 skipped**; `ruff` + `mypy` clean.

## Checklist

- [x] PR conforms to the Contributing Guideline (Commit Message and PR Title formats)
- [x] Tests added/updated
- [x] Docs added/updated (N/A — internal behavior change)

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/unity/usage.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/unity/usage.py
@@ -412,6 +412,48 @@ class UnityCatalogUsageExtractor:
                 context=context,
             )
 
+    def _report_no_queries(self) -> None:
+        # Called when query history yielded zero usable queries (and no fetch
+        # failure occurred). Distinguishes "rows were present but unparseable"
+        # from "no rows at all", and tailors the permission hint to the path.
+        if self.report.num_queries_missing_info > 0:
+            self.report.report_warning(
+                title="Query history rows could not be parsed",
+                message=(
+                    "Statements from system.query.history could not be "
+                    "parsed and were skipped, so no usage was extracted."
+                ),
+                context=f"count={self.report.num_queries_missing_info}",
+            )
+            return
+
+        if self._use_system_tables_join():
+            hint = (
+                "verify SELECT privilege on system.query and system.access "
+                "and that the time window covers recent activity"
+            )
+            if (
+                self.config.push_down_database_pattern_access_history
+                and self.config.catalog_pattern is not None
+            ):
+                hint += (
+                    "; if catalog pushdown is enabled, also verify "
+                    "catalog_pattern allow/deny rules are not over-restrictive"
+                )
+        else:
+            hint = (
+                "verify CAN_MANAGE privilege on the SQL warehouse "
+                "and that the time window covers recent activity"
+            )
+        self.report.report_warning(
+            title="No queries found for usage",
+            message=(
+                "No queries were found in the configured time range "
+                "for usage extraction."
+            ),
+            context=hint,
+        )
+
     def get_usage_workunits(
         self, table_refs: Set[TableReference]
     ) -> Iterable[MetadataWorkUnit]:
@@ -452,11 +494,12 @@ class UnityCatalogUsageExtractor:
             try:
                 aggregator = self._build_aggregator(is_allowed_table=is_allowed_table)
                 buffered_queries = FileBackedList[Query]()
-                for query in self._fetch_queries():
-                    # Drain only — the cursor/connection is released once the
-                    # generator is fully consumed here.
-                    self.report.num_queries += 1
-                    buffered_queries.append(query)
+                with self.report.usage_query_fetch_timer:
+                    for query in self._fetch_queries():
+                        # Drain only — the cursor/connection is released once the
+                        # generator is fully consumed here.
+                        self.report.num_queries += 1
+                        buffered_queries.append(query)
             except (MemoryError, SystemExit, KeyboardInterrupt):
                 raise
             except Exception as e:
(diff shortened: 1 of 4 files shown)


metadata-ingestion/src/datahub/ingestion/source/unity/usage.py
```diff
@@ -439,42 +558,74 @@ def get_usage_workunits(
...
+        use_cached_audit_log = audit_log_file is not None and audit_log_file.exists()
+
+        buffered_queries: Optional[FileBackedList[Query]] = None
         try:
             try:
                 aggregator = self._build_aggregator(is_allowed_table=is_allowed_table)
-                for query in self._fetch_queries():
-                    self.report.num_queries += 1
+                if use_cached_audit_log:
+                    assert audit_log_file is not None
```
askumar27, line 578: 🟡 **W1: `assert` for runtime type narrowing.** This (and the two other `assert … is not None` on lines 619 and 654) will be stripped by `python -O`, which would let `None` propagate silently. Safer to use an explicit guard:

```python
if audit_log_file is None:
    raise RuntimeError("audit_log_file unexpectedly None after use_cached_audit_log check")
```

Same pattern for `buffered_queries` on lines 619 and 654.

metadata-ingestion/src/datahub/ingestion/source/unity/usage.py
```diff
@@ -494,54 +645,42 @@ def get_usage_workunits(
...
-                        )
-                # Skip resetting per-table usage when we couldn't read any queries at
-                # all (empty history or missing permission).  Emitting zero-usage aspects
-                # in that case would wrongly wipe existing usage stats in DataHub.
-                return
+                # Successful but empty read (a fetch failure returns above): warn, then
+                # fall through to auto_empty so the idle window records a zero datapoint,
+                # matching Snowflake/BigQuery/Redshift. (Timeseries UPSERT — adds a
+                # current-bucket zero, does not delete history.)
+                self._report_no_queries()
```
askumar27, line 652: 🟡 **W2: Behavioral change on zero-queries.** Previously, an empty query-history read emitted **no** usage workunits (return early). Now it falls through to `auto_empty_dataset_usage_statistics`, stamping a zero datapoint.

The rationale is sound (matches Snowflake/BigQuery/Redshift, timeseries UPSERT doesn't delete history), but this is user-visible: connectors on idle workspaces will now emit `datasetUsageStatistics` where they previously emitted none. Worth a note in `updating-datahub.md`?

metadata-ingestion/src/datahub/ingestion/source/unity/usage.py
```diff
@@ -439,42 +558,74 @@ def get_usage_workunits(
...
-                            aggregator, query, default_db=default_db
+                        logger.info(
+                            "Using cached query-history audit log at %s", audit_log_file
                         )
-                    except (
-                        MemoryError,
-                        SystemExit,
-                        KeyboardInterrupt,
-                    ):  # never swallow system-level errors as a "dropped query"
+                        shared_connection = ConnectionWrapper(audit_log_file)
```
askumar27, line 583: ℹ️ **S1: Cache integrity on read.** The corrupt-cache test covers files that can't be opened (not valid SQLite), but if a valid SQLite file contains rows from an incompatible `Query` shape that _deserializes successfully_ with wrong field mappings, the cache would silently serve incorrect data. The format-version bump mitigates this for planned changes, but consider a lightweight read-back check (e.g., verify a known field on the first row) after opening a cached file.

askumar27 (approved):
Nice work — solid PR that brings Unity in line with the Snowflake/BigQuery drain-before-parse pattern. Well-tested with good coverage of cache lifecycle edge cases. See inline comments for a couple of warnings and suggestions.

metadata-ingestion/src/datahub/ingestion/source/unity/usage.py
```diff
@@ -25,9 +26,15 @@
...
 from datahub.sql_parsing.sql_parsing_common import QueryType
 from datahub.sql_parsing.sqlglot_utils import get_query_fingerprint
+from datahub.utilities.file_backed_collections import ConnectionWrapper, FileBackedList
 
 logger = logging.getLogger(__name__)
 
+# Bump whenever the on-disk shape of a buffered Query (or its pickling) changes, so a
+# cache written by an older build resolves to a different filename — a clean miss that
+# re-fetches, rather than a deserialization error when reloading an incompatible Query.
+_AUDIT_LOG_FORMAT_VERSION = 1
```
askumar27, line 36: nit: The comment says "bump whenever the on-disk shape … changes." Consider also noting that `Query` field renames/removals should trigger a bump, since `FileBackedList` uses pickle — a renamed field could silently deserialize to a default rather than erroring.

PR #17980, opened 2026-06-20
feat(ingestion): remove column-level classification feature

treff7es:
## Summary

Removes the column-level **classification** feature from ingestion. It depended on the unmaintained [`acryl-datahub-classify`](https://pypi.org/project/acryl-datahub-classify/) package, which pinned `numpy<2` and an outdated spaCy stack (`blis`/`thinc`) and blocked dependency upgrades across the framework.

This is the first step of a dependency-modernization effort (the remaining `numpy<2` driver is `feast`, handled separately).

## What changed

- **Dependency:** dropped `acryl-datahub-classify` (and its `schwifty`/`pip`/`numpy<2` companions) from `setup.py`; regenerated `pyproject.toml`/`uv.lock`/`constraints.txt`.
- **Framework removed:** deleted `datahub_classifier.py`, `classifier_registry.py`, and `classification_mixin.py`.
- **Deprecation UX:** the `classification` config option is now **hidden** from source config schemas (no declared field). A recipe that still sets it **fails fast** at startup with a clear `ValueError`:
  > Column-level classification has been removed … install the last release that supports it: `pip install 'acryl-datahub==1.6.0.5'`.
  
  (See `ClassificationSourceConfigMixin` in `classifier.py`.) Recipes that don't configure classification are unaffected.
- **Connectors:** stripped the classification handler / workunit processor / report mixin usage from SQL, Snowflake, BigQuery, Redshift, and dynamodb (config mixin kept, now sourced from `classifier.py`, so the deprecation error fires).
- **Tests/docs:** removed classification from integration recipes and regenerated affected goldens; deleted classification unit/integration tests and added one covering the deprecation; updated the classification dev guide to a removal notice and added a breaking-change entry to `updating-datahub.md`.

> Note: the `@capability(SourceCapability.CLASSIFICATION, ...)` declarations are intentionally left in place for now; fully scrubbing the capability surface (and regenerating connector capability docs) can be a follow-up.

## Testing

- `ruff check` + `ruff format --check` → clean
- `mypy src/ tests/` → **Success: no issues found in 2186 source files**
- `checkLockFile` → in sync
- Unit tests (changed connectors + new deprecation test): **1193 passed** (1 pre-existing unrelated `azure.mgmt` collection error in the dev venv)
- Integration goldens regenerated under `TZ=UTC` and verified: **bigquery (9), snowflake (6), dynamodb, mysql/postgres (9, docker), trino (3, docker)** all pass

## Checklist

- [x] Tests added/updated
- [x] Docs updated (classification dev guide + breaking change note)
- [x] Breaking changes documented in `docs/how/updating-datahub.md`


sgomezvillamor:
## Review

**🔴 BLOCKER — Full framework removal needs product approval**

The PR removes more than the `acryl-datahub-classify` integration — it also deletes the abstract `Classifier` base class, the `PluginRegistry`, and `ClassificationConfig`. These are public extension points that third-party users could have implemented independently of `acryl-datahub-classify`. Removing only `datahub_classifier.py` and the `datahub_classify` dependency from `setup.py` would achieve the stated goal (unblocking `numpy>=2`) without breaking custom implementations. The full framework removal needs explicit product approval before merging.

---

**🔴 BLOCKER — Hard fail impact not assessed**

Recipes with `classification.enabled: true` now fail entirely at startup. There is no mention of telemetry on how many active recipes have classification enabled, no customer outreach, and no deprecation grace period. A less disruptive approach: emit a warning and continue ingestion without classification, deferring the hard fail to a subsequent release. At minimum, the risk assessment needs to be done and documented before merging.

---

**🟡 WARNING — `@capability(SourceCapability.CLASSIFICATION, ...)` declarations left in place**

If the feature and its supporting framework are fully removed, the capability declarations on Snowflake, BigQuery, Redshift, and the SQL common source should also be removed. As-is, these sources will advertise classification support while any attempt to use it raises `ValueError` at startup — misleading for users and any tooling that reads capability metadata.

---

**🟡 WARNING — New deprecation test uses a deprecated Pydantic v1 method**

`test_classification_deprecation.py:15` calls `ClassificationSourceConfigMixin.parse_obj(...)`. Per project standards, this should be `ClassificationSourceConfigMixin.model_validate(...)`.

---

sgomezvillamor:
**Follow-up on the `@capability(SourceCapability.CLASSIFICATION, ...)` declarations**

Thanks for the thoughtful refactor — keeping the framework while removing only the built-in implementation is the right call.

One remaining concern: the `@capability(SourceCapability.CLASSIFICATION, ...)` declarations on Snowflake, BigQuery, Redshift, and the SQL common source were restored in the third commit, but these sources no longer support classification **out of the box**. The capability advertisement implies the source can classify data, not that it *could* if a user writes and registers a custom classifier themselves.

In practice: after this PR, none of our sources classify anything without user-supplied code. Claiming the capability is supported is therefore misleading — it sets a wrong expectation for users browsing connector docs or using tooling that reads capability metadata.

Two options worth considering:

1. **Remove the declarations from all sources.** If no source ships a working classifier, no source should claim the capability. The framework (registry, base class, mixin) stays intact for users who build their own integration.
2. **Remove `SourceCapability.CLASSIFICATION` entirely.** If the capability is meaningless without a third-party or custom implementation, keeping the enum value is noise. A follow-up PR could clean this up along with any UI/doc references.

Either way, restoring the capability declarations in a state where they cannot be exercised without extra user work seems like the wrong outcome. Would be good to address this before merging.

---

treff7es:
@sgomezvillamor good catch — agreed. Since no source ships a working classifier out of the box anymore, advertising `SourceCapability.CLASSIFICATION` overstated what they do.

Went with option 1: removed the `@capability(SourceCapability.CLASSIFICATION, ...)` declarations from all the sources that had them — Snowflake, BigQuery, Redshift, the SQL common source, **and DynamoDB** (which had the same declaration). I also removed the now-dead `CLASSIFICATION` references in Snowflake's `check_capabilities` runtime report (it would otherwise have kept reporting a capability the source no longer declares). Kept `SourceCapability.CLASSIFICATION` in the enum so it's still available to anyone registering a custom classifier, and left Hive's `supported=False` ("Not Supported") declaration as-is since it's honest.

Also merged latest master to resolve the conflict. Ready for re-review.

sgomezvillamor (approved):
LGTM

I would also remove this line too
https://github.com/datahub-project/datahub/blob/870b850ed486e8a1f201b58bb22599f24a7ea2f8/metadata-ingestion/src/datahub/ingestion/api/source.py#L79

PR #17981, opened 2026-06-20
feat(ingestion/unity): remove legacy GE profiling method

treff7es:
## Summary

Removes the legacy Great Expectations profiling method (`profiling.method: ge`) from the Unity Catalog source. It is superseded by the SQLAlchemy-based profiler (`method: sqlalchemy`, the default); `analyze` (ANALYZE TABLE) remains as the other option.

This is the first step toward bumping `databricks-sql-connector` (which is gated on dropping the databricks SQLAlchemy-dialect usages — see ING-2895). It removes the legacy GE path; the `sqlalchemy` method and `hive_metastore_proxy` still use the dialect, so this PR does not by itself unblock the connector bump.

## Changes

- Delete `UnityCatalogGEProfilerConfig` and drop it from the discriminated `profiling` union (now `analyze` | `sqlalchemy`).
- Remove the `method: ge` dispatch branch in `source.py`.
- Remove the now-unused `is_ge_profiling()` helper; `uses_table_level_profiler()` now checks only the sqlalchemy method.
- Narrow `UnityCatalogGEProfiler` typing to `UnityCatalogSQLAlchemyProfilerConfig` (the class is retained — it backs the `sqlalchemy` method).
- Update unit tests and the unity-catalog profiling permission docs.

## Migration

Recipes that set `profiling.method: ge` should switch to `sqlalchemy` (the default) or `analyze`.

## Testing

- `mypy src/datahub/ingestion/source/unity/` → Success (18 files)
- `pytest tests/unit/test_unity_catalog_config.py tests/unit/test_unity_catalog_proxy.py` → 94 passed, 1 skipped
- ruff check + format clean

## Checklist

- [x] Tests updated
- [x] Docs updated (unity-catalog profiling permissions)

### DIFF
--- a/metadata-ingestion/docs/sources/databricks/unity-catalog_pre.md
+++ b/metadata-ingestion/docs/sources/databricks/unity-catalog_pre.md
@@ -55,7 +55,6 @@ You can authenticate with Databricks using OAuth, Azure authentication, a Person
   - Otherwise (REST API path): `CAN_MANAGE` on the SQL warehouse: [guide](https://docs.databricks.com/security/auth-authz/access-control/sql-endpoint-acl.html).
 - To `include_table_constraints` (disabled by default), no additional permissions are required beyond the `SELECT` privilege already listed above. The connector uses the same `tables.get()` API endpoint.
 - To ingest `profiling` information with the default SQLAlchemy profiler (`method: sqlalchemy`), you need `SELECT` privilege on tables and views.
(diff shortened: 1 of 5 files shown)


treff7es:
Closing — removing only the unity `method: ge` config doesn't stand on its own (the `sqlalchemy` method and hive_metastore_proxy still use the databricks SQLAlchemy dialect, so it unblocks nothing and just drops a working profiling option). Superseding with the proper change: remove the Great Expectations profiler globally, standardize on the SQLAlchemy profiler (which implements the same interface), and make that profiler SQLAlchemy-2.0-compatible — which also removes the acryl-great-expectations sqlalchemy<2 pin. Tracked in ING-2895.

PR #17982, opened 2026-06-20
chore(ingest): cut over metadata-ingestion to SQLAlchemy 2.0

treff7es:
## Summary

All-or-nothing cutover of `metadata-ingestion` from **SQLAlchemy 1.4 → 2.0**. The base
pin becomes `sqlalchemy>=2.0,<2.1`, so the uv universal lock resolves a single SA version
for the whole package — there is no per-connector version split. All SQLAlchemy-touching
code paths are migrated to the 2.0 API and the 1.4 compatibility routes are removed.

> **Draft:** core migration is complete and validated (see below). A few cross-cutting
> items remain before this is merge-ready — tracked in *Outstanding* at the bottom.

## Motivation & Impact

**Why now — SQLAlchemy 1.4 is EOL and the dialect ecosystem has moved on.** We were pinned to
`SQLAlchemy 1.4.44` (released **Nov 2022**); the entire 1.4 line has had **no release since
Sept 2024** (1.4.54) and is in security-only/EOL status — it was always the temporary
"migration bridge" to 2.0. Meanwhile the database dialects we depend on have hard-pinned to
SA 2.0 in their current releases, so staying on 1.4 freezes us on abandoned dialect versions:

| Dialect | 1.4-era version (SA pin) | Current version (SA pin) |
| --- | --- | --- |
| sqlalchemy-redshift | `0.8.14` (`<2.0.0`) | `1.0.0` (**`>=2.0.0`**) |
| databricks-sqlalchemy | legacy `databricks-dbapi` (deprecated) | `2.0.9` (**`>=2.0.21`**) |
| sqlalchemy-hana (SAP HANA) | `3.0.3` | `4.6.0` (**`>=2`**) |
| sqlalchemy-cockroachdb | `1.4.6` | `2.0.4` (**`>=2.0.47`**) |
| pyathena (Athena) | `2.25.2` (`<2.0.0`) | `3.x` async (**`>=2.0.0`**) |

The single most concrete blocker: **`vertica-sqlalchemy-dialect` (last release Jun 2024, abandoned)
pins `sqlalchemy<=1.4.44`** — one unmaintained dialect that nailed the whole package to a
3.5-year-old SQLAlchemy build. It's removed here in favour of the maintained
`sqlalchemy-vertica-python`.

**Metadata-extraction impact — fidelity-neutral-to-positive, zero regressions.** Of the nine SQL
connectors, six (mysql, postgres, mssql, hive, trino, timescaledb) produce **byte-identical**
metadata — a pure code-path migration. The three that changed only *improved*:
- **clickhouse:** a time column now types as `TimeType` (was `NullType`); a key column now
  correctly reports `isPartOfKey: true`.
- **oracle:** ROWID columns now carry a real `nativeDataType` (was `null`); ALL-mode no longer
  ingests Oracle's internal `mview_*` / `scheduler_*` system-catalog tables (noise reduction).
- **db2:** native type label `INTEGER` → `INT` (the upgraded `ibm-db-sa` driver's rendering; the
  DataHub field type is unchanged — still `NumberType`).

Across all goldens there are **no losses**: nothing downgraded to `NullType`, no fields/primary
keys dropped, no nullability changes.

**Security.** Dropping the Great-Expectations / Jupyter / SciPy stack and the abandoned vertica
dialect, plus bumping every driver, removed 65 transitive packages. The repo security scan
(Aikido) reports the change **resolves 6 critical + 2 medium + 38 low** vulnerabilities.

**Performance.** Secondary for our read/reflection-heavy usage, but real: SA 2.0 has faster
Cython row/result processing and a more mature statement-compilation cache, and the driver bumps
bring throughput gains on large result sets (e.g. `databricks-sql-connector 2.9.6 → 4.3.0` adds
Arrow/CloudFetch retrieval). SA 2.0's `insertmanyvalues` is not material here since extraction is
read-dominated.

**Also enables:** Python 3.12/3.13 support (SA 1.4's C extensions never fully supported them),
native typing via SA 2.0's `py.typed` (we dropped the stopgap `sqlalchemy2-stubs`), and the
`text()`-required execution model that eliminates a class of accidental raw-string SQL.

## Why all-or-nothing

SQLAlchemy can only resolve to one version per environment. Some dialect drivers we depend
on pin `SQLAlchemy<2`, so a partial migration is not installable — bumping the base pin
forces every dialect to 2.0 at once. That also let us delete the dual-version shims rather
than carry them indefinitely.

## What changed

**Dependency / packaging**
- Base pin `sqlalchemy>=2.0,<2.1`; removed `sqlalchemy2-stubs` + the SA mypy plugin (incompatible with SA 2.0's own plugin).
- `unity`: added `databricks-sqlalchemy` (registers the `databricks` dialect on SA 2.0).
- `vertica`: switched to the maintained `sqlalchemy-vertica-python` (the old dialect pinned `SA<=1.4.44`).
- Regenerated `pyproject.toml` / `uv.lock` / `constraints.txt`.
- Removed the probe-only `testSqlAlchemy2` CI lane + `sa2_compat` scaffolding (the main `testQuick` lane now runs those suites under SA 2.0).

**Migration patterns applied across connectors**
- Password masking: `str(URL)` → `URL.render_as_string(hide_password=False)` (load-bearing — without it no SQL connector authenticates).
- Raw SQL wrapped in `text()`; `%s` / `%(name)s` params → named `:name` binds.
- `Engine.execute()` (removed) → `with engine.connect() as conn:`.
- Bind params as kwargs `execute(stmt, k=v)` (removed) → param dict `execute(stmt, {"k": v})`.
- `Row` dict-access (`row["c"]`, `"c" in row`, `row.items()`, removed) → `.mappings()` / `row._mapping`.
- `select([cols])` → `select(*cols)`; dialect/driver import-path updates (pyathena, databricks, vertica).
- Query combiner: transaction rollback before serial fallback (autocommit removed in 2.0).

**Vertica** — hybrid approach: a `VerticaInspector(Inspector)` subclass that copies dialect state and re-implements the Vertica-specific reflection on the 2.0 API, instead of forking the abandoned dialect.

**Profiling** — removed the Great Expectations profiler entirely (sqlalchemy-only profiling). `acryl-great-expectations` pins `SA<2`, so it is uninstallable on the cutover; `profiling.method` is now `Literal["sqlalchemy"]`.

## Breaking changes

- **SQLAlchemy 2.0 is now required** for `metadata-ingestion`.
- **Great Expectations profiler removed.** `profiling.method: ge` is no longer valid; only `sqlalchemy` remains. (The generic SQLAlchemy profiler has been the default and is at parity for the metrics it supports.)
- **Vertica** now requires `sqlalchemy-vertica-python` (installed via the `vertica` extra).

## Validation

- **Full Metadata Ingestion CI is green** — all `testQuick` lanes (3.10/3.11/3.12), `testPerformance`,
  and every `testIntegrationBatch` (0–5 + Recording) pass on the latest commit.
- **Unit tests:** `mypy` clean on both `src/` and `tests/`; ruff + `checkLockFile` green.
- **Real-instance (connector-tests):** mysql, postgres, snowflake, bigquery, redshift, databricks/unity, fivetran.
- **Docker integration:** oracle, hana, clickhouse, trino, timescaledb, hive, db2, mssql.
- Integration tests surfaced (and we fixed) several real SA-2.0 issues rather than papering over them by
  regenerating goldens:
  - trino/presto reflection — bind-params-as-kwargs + `Row` dict-access in `$properties` parsing (was silently dropping table comments / ingesting nothing).
  - timescaledb — raw `Row` string-indexing dropping every hypertable/continuous-aggregate/job.
  - hive/presto — `acryl-pyhive` (SA-1.4-era) import failures + raw-string dialect reflection; fixed with a compat shim + `text()` dialect overrides.
- Goldens: clickhouse + oracle regenerated after manual review (real driver-bump fidelity gains / Oracle
  system-table exclusion); db2 updated for the `ibm-db-sa` native-type label (`INTEGER`→`INT`); trino goldens
  **kept** because the only regen diff was local-timezone frozen-time values that would break CI.

## Outstanding before merge

- [ ] Reconcile `acryldata/connector-tests` `method:ge` variants → `method:sqlalchemy` (cross-repo; GE profiler removed here).
- [ ] Re-validate the query-combiner transaction-rollback path (P10) against a live Postgres/Redshift.

**Resolved during this PR (no longer blocking):**

- ✅ hive/presto on SA 2.0 — handled in-repo via `_pyhive_compat` (import shim) + `text()` dialect overrides; hive integration is green. A future maintained `acryl-pyhive` release could let us drop the shim, but it is not required to merge.
- ✅ db2 runs in CI — `ibm-db-sa` resolves and the db2 integration batch passes (native `ibm_db` only blocked *local* arm64 runs).
- ✅ Breaking-change entry added to `docs/how/updating-datahub.md`.

## Checklist

- [x] PR title follows the Conventional Commits format
- [x] Tests added/updated
- [x] Docs updated — breaking-change entry added to `docs/how/updating-datahub.md`
- [x] Breaking changes documented in `docs/how/updating-datahub.md`


metadata-ingestion/src/datahub/ingestion/source/sql/oracle.py
```diff
@@ -902,7 +923,9 @@ def get_columns(
             "identity_cols": identity_cols,
         }
 
-        c = self._inspector_instance.bind.execute(sql.text(text), params)
+        c = _inspector_connection(self._inspector_instance).execute(
+            sql.text(text), params
```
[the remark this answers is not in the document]
treff7es: @AikidoSec ignore: False positive. This is the upstream SQLAlchemy Oracle dialect `get_columns` reflection pattern. The interpolated fragments are not user input: `dblink` is always "" in DataHub (it chains from the `dblink=""` default and is never set — it is an Oracle database-link reflection feature we do not use), and `char_length_col` / `identity_cols` are hardcoded, server-version-derived constants. The only user-controlled values, `table_name` and `owner`, are passed as bind parameters (`:table_name`, `:owner`). This PR did not introduce the string formatting — it only changed `.bind.execute(...)` to `_inspector_connection(...).execute(...)` for SQLAlchemy 2.0 compatibility.

docs/how/updating-datahub.md
```diff
@@ -108,6 +108,14 @@ Requirements:
...
+
+  - **Great Expectations profiler removed.** The legacy GE-based profiler (`profiling.method: ge`) has been removed across all SQL sources; `acryl-great-expectations` pins `SQLAlchemy<2` and is uninstallable on the cutover. The generic SQLAlchemy profiler (`profiling.method: sqlalchemy`, already the default) is the only profiler. **Migration:** remove `method: ge` from any source's `profiling` block (or set `method: sqlalchemy`) — `ge` is no longer a valid value and will fail config validation. The `profiling_ge` / `great-expectations` install extras have been removed.
```
sgomezvillamor: What about automatically converting `profiling.method: ge` to `profiling.method: sqlalchemy` at runtime, and raising a warning so users are aware of the change? WDYT?

sgomezvillamor (commented):
## SQLAlchemy 2.0 migration review

Overall this is a thorough and well-executed migration. The core SA 2.0 patterns (`engine.execute()` removal, `text()` requirements, `select(*cols)`, `.mappings()`, autocommit removal, `render_as_string`) are applied consistently across all 100+ changed files. The pyhive shim design is particularly well thought out — the two-layer approach (import-time patches in `_pyhive_compat.py` + dialect method replacements in `hive_source.py`) is the right architecture for bridging an unmaintained third-party dialect.

Two issues need to be addressed before merging:

---

### 🔴 `unity/hive_metastore_proxy.py` — `rows.index(Row(...))` will silently return empty table properties

```python
# _execute_sql returns:
return cast(List[Row], conn.execute(text(sql)).fetchall())

# used as:
index = rows.index(Row("# Detailed Table Information", "", ""))
```

`conn.execute(text(sql)).fetchall()` returns `List[sqlalchemy.engine.Row]` at runtime — the `cast()` is a type-checker-only annotation with no runtime effect. `Row` here is `databricks.sql.types.Row`. In SA 2.0, `sqlalchemy.engine.Row` is a named tuple; its `__eq__` compares by value against other tuples, but equality with a `databricks.sql.types.Row` is not guaranteed — it depends on an undocumented cross-type contract between two third-party libraries. If the comparison fails, `list.index()` raises `ValueError`, which is caught by the surrounding `try/except`, and `_get_table_info` silently returns empty properties with no warning logged.

Suggested fix — compare by column value directly instead of relying on cross-type equality:

```python
index = next(
    (i for i, r in enumerate(rows) if r[0] == "# Detailed Table Information"),
    None,
)
if index is None:
    return {}
```

Also in the same file: `self.inspector.bind.close()` — in SA 2.0, `Inspector.bind` returns the `Engine` when created via `inspect(engine)`. `Engine` has no `.close()` method. If the inspector was created with an engine rather than a connection, this will raise `AttributeError`. The `# type: ignore` comment is suppressing the warning about this. Worth verifying how this inspector is constructed and whether `inspector.close()` or a context manager should be used instead.

---

### 🔴 `presto.py` — `PrestoDialect.get_schema_names` not patched for standalone Presto runs

`hive_source.py` patches `HiveDialect.get_schema_names` to use `text()`. Since `PrestoDialect` inherits from `HiveDialect`, this fix propagates via MRO — but only when `hive_source.py` has been imported. In a Presto-only pipeline, `hive_source.py` is never loaded, so `PrestoDialect.get_schema_names` resolves to pyhive's original implementation which calls `connection.execute("SHOW DATABASES")` — a pattern SA 2.0 rejects with `ObjectNotExecutableError`.

Fix: add to `presto.py` (importing `get_schema_names_patched` from `hive_source.py`, or defining it inline):

```python
PrestoDialect.get_schema_names = get_schema_names_patched
```

---

### 🟡 Query combiner rollback path — listed as pending in the PR checklist

The PR itself flags *"Re-validate the query-combiner transaction-rollback path (P10) against a live Postgres/Redshift"* as a pre-merge item. Flagging here to confirm it's still open. The rollback-before-fallback code:

```python
try:
    query_future.conn.rollback()
except Exception as rollback_err:
    logger.debug(...)
```

In SA 2.0, `Connection.rollback()` raises `InvalidRequestError` when called outside an active transaction. The `try/except` swallows this, but the fallback query then runs on a connection in an indeterminate state. This needs live integration validation before merge.

---

### 🟡 `vertica_inspector.py` — N+1 connections per schema introspection

`_fetch_projection_comments` and related methods each open a fresh `engine.connect()`. For a Vertica schema with many projections this creates a new connection per query, which is a significant regression from the original single-connection `vertica-sqlalchemy-dialect`. Consider accepting an optional connection parameter or opening one connection at construction time and reusing it.

---

### 🟡 `hive/hive_sql_fetcher.py` — connection leak on exception

`SQLAlchemyClient.execute_query` yields rows lazily from a connection opened in a lazy property. If the caller raises an exception mid-iteration, `close()` is never called and the connection leaks. The class should implement `__enter__`/`__exit__` (or the call sites should use `contextlib.closing`) to guarantee cleanup.

---

### ℹ️ `sql_config.py` — stale documentation link

Field description still references SA 1.4 docs (`en/14/core/engines.html`). Should be `en/20/`.

---

### ℹ️ `ge_profiling_config.py` — breaking change for `method: ge`

Users with `profiling.method: ge` in existing recipes will get a hard Pydantic validation error after upgrade with no guidance. One option worth considering: rather than a hard failure, a `@field_validator` could auto-migrate the value at runtime and emit a deprecation warning:

```python
@field_validator("method", mode="before")
@classmethod
def migrate_ge_method(cls, v: str) -> str:
    if v == "ge":
        logger.warning(
            "profiling.method 'ge' is no longer supported (acryl-great-expectations removed). "
            "Automatically migrating to 'sqlalchemy'. Please update your recipe."
        )
        return "sqlalchemy"
    return v
```

This preserves the breaking-change nature (ge is gone) while avoiding a hard crash for users who haven't updated their recipes yet, and gives them an actionable warning. Still worth an entry in `docs/how/updating-datahub.md` either way.

---

### ℹ️ `_pyhive_compat.py` — removal trigger

The docstring says to remove this once acryl-pyhive ships SA 2.0 support, but doesn't reference the pinned version. A comment like `# acryl-pyhive==0.6.18 pins us to the SA-1.4 era — revisit when the pin is lifted` directly above the patches would make the removal trigger self-documenting.


---

treff7es:
Thanks for the review — went through each point. Net: **4 code changes**, and verified pushback on the rest (one confirmed empirically against a live Postgres). All landed in a single commit; branch is rebased onto latest master with ruff + mypy + `checkLockFile` green.

### Fixed

**🔴 `presto.py` `get_schema_names`** — confirmed real. pyhive's `PrestoDialect.get_schema_names` runs raw-string `connection.execute('SHOW SCHEMAS')`, which SA 2.0 rejects. One correction on the mechanism: `PrestoDialect` extends `DefaultDialect`, not `HiveDialect` (they're siblings), so the Hive patch never reached it — which is *why* it was missed. Only affects standalone Presto (Trino uses a separate SA-2.0-native dialect, so it never exercised this path). Patched with a `text()` version + a regression test.

**🟡 `vertica_inspector.py` N+1 connections** — reflection now reuses the connection the inspector is already bound to (`get_inspectors()` binds it via `inspect(conn)`), as the original dialect did, instead of a per-method `engine.connect()` checkout. Lifecycle stays owned by `get_inspectors()`.

**🟡 `hive_sql_fetcher.py`** — added `__enter__`/`__exit__` to the SQLAlchemy client and data fetcher and switched the test-connection path to `with`. (Note: it wasn't actually leaking — the connection is owned by the long-lived client and closed in `source.close()`, and the test path already used try/finally — but the context-manager form is cleaner.)

**ℹ️ `sql_config.py`** — doc links `/en/14/` → `/en/20/` (3 spots).

**ℹ️ `ge_profiling_config.py` `method: ge`** — kept as a hard failure (see the updating-datahub.md thread), but it now raises an actionable `ValueError` pointing at the fix instead of the opaque `Literal` error.

### Verified, no change needed

**🔴 `hive_metastore_proxy` `rows.index(Row(...))`** — not a bug. SA 2.0 `Row.__eq__` compares by tuple value, and `databricks.sql.types.Row` is a positional tuple subclass, so the lookup resolves by value (documented at `_execute_sql`). The failure path also isn't silent — `_get_table_info` calls `report_warning`.

**🔴 `inspector.bind.close()`** — not a bug. The inspector is built from `inspect(engine.connect())` — a `Connection`, not an `Engine` (the `assert isinstance(conn, Connection)` in `_execute_sql` relies on it). `Connection.close()` exists; the `# type: ignore` only covers the stub's `Engine | Connection` union.

**🟡 query-combiner rollback** — validated against live Postgres: a failed combined query aborts the transaction (subsequent statements raise `25P02`), and the rollback-before-fallback lets the valid queued query still succeed serially. `Connection.rollback()` with no active transaction is a **no-op** on SA 2.0 (it does not raise `InvalidRequestError`), so it's safe in both cases.

**ℹ️ `_pyhive_compat` removal trigger** — already documented: the module docstring names the pin, *"acryl-pyhive (0.6.18, the latest published release)"*.

docs/how/updating-datahub.md
treff7es: Considered it, but kept this a hard failure rather than auto-migrating — `acryl-great-expectations` is fully removed, so silently rewriting `ge` → `sqlalchemy` would mask that the profiler a user explicitly asked for is gone. Instead, `method: ge` now raises an actionable `ValueError` at config validation ("remove the 'method' line or set it to 'sqlalchemy'") instead of the opaque Literal error, so users still get a clear, fail-fast signal.

PR #17984, opened 2026-06-21
fix(ingest/unity): address post-merge review comments on usage caching

treff7es:
## Summary

Follow-up to #17971 (merged) addressing review comments from @askumar27 that landed just before the squash-merge. No functional behavior change beyond the documented one (W2).

- **W1 — `assert` stripped by `python -O`.** Replaced the three `assert … is not None` runtime narrowings in `get_usage_workunits` with a small `_require()` helper that raises explicitly, so the invariant checks survive `python -O` (which strips `assert` statements). The narrowing now flows through `_require`'s return type — also keeps `get_usage_workunits` under the complexity limit.
- **W2 — document the idle-window behavior change.** Added an entry to `docs/how/updating-datahub.md`: on the system-tables usage path, an idle window (zero queries, fetch succeeded) now emits zero-value `datasetUsageStatistics` for ingested tables where it previously emitted none — matching Snowflake/BigQuery/Redshift. It's a timeseries UPSERT (current-bucket zero, no history deletion); a genuine fetch/permission failure still emits nothing.
- **S1 + nit — cache-format version comment.** Expanded the `_AUDIT_LOG_FORMAT_VERSION` comment to state that field renames/removals **and** semantic/type changes must bump the version, since `FileBackedList` pickles `Query` and pickle does not validate against the current class. A read-back field check cannot detect a structurally-valid-but-semantically-wrong cache, so the version token is the authoritative guard — no separate read-back check was added (it would give false confidence without closing the gap).

## Checklist

- [x] PR conforms to the Contributing Guideline (Commit Message and PR Title formats)
- [x] Links to related issues — follow-up to #17971
- [x] Tests added/updated — covered by #17971's suite; this is a refactor (W1) + docs (W2) + comment (S1/nit), no behavior change. 158 unit tests pass, `ruff` + `mypy` clean.
- [x] Docs updated — `updating-datahub.md` (W2)


askumar27 (approved):
LGTM!

PR #17969, opened 2026-06-19
feat(ingest/mysql): add performance_schema usage and query lineage

acrylJonny:
## Summary

Adds opt-in **usage statistics and query-based lineage** for the MySQL and MariaDB connectors, sourced from `performance_schema.events_statements_summary_by_digest` instead of the general query log.

- New `include_usage_statistics` config (default `false`) on `MySQLConfig` (inherited by MariaDB), plus a `usage: BaseUsageConfig` field for the time window.
- When enabled, normalized statement digests are read from `performance_schema` and fed through a `SqlParsingAggregator` to emit dataset usage statistics, top queries, and table/column lineage. This relies only on the always-on `statements_digest` consumer (on by default in MariaDB 10.11) and `SELECT` on `performance_schema`.
- Two-tier aware: the schema is treated as `default_schema` so digests resolve to `schema.table` URNs.
- Failures reading `performance_schema` (disabled consumer, missing grant) are reported as warnings and do not abort the run.
- Adds the `USAGE_STATS` capability to both `MySQLSource` and `MariaDBSource`.
- Also extracts dataJob-to-dataJob lineage from `CALL`/`EXEC` statements in stored procedures, and adds a MariaDB integration test suite.

### Caveats (documented in the connector docs)

- No per-user attribution — digests are aggregated across all users.
- `COUNT_STAR` is cumulative since the digest table was last reset; executions are attributed to the bucket of each digest's `LAST_SEEN`.
- Digest text is normalized (literals replaced with `?`) and truncated to `performance_schema_max_digest_length`.

## Checklist

- [x] The PR conforms to DataHub's Contributing Guideline (particularly [Commit Message Format](https://datahubproject.io/docs/contributing/#commit-message-format))
- [x] Links to related issues (if applicable)
- [x] Tests added/updated (unit tests for digest mapping/filtering and aggregator wiring; MariaDB integration test suite)
- [x] Docs added/updated (MySQL/MariaDB prerequisites + limitations)

Made with [Cursor](https://cursor.com)

PR #17985, opened 2026-06-21
feat(ingest/mysql): add performance_schema usage and query lineage

acrylJonny:
## Summary

Adds opt-in **usage statistics and query-based lineage** for the MySQL and MariaDB connectors, with a pluggable `usage_source`.

- New `include_usage_statistics` config (default `false`) on `MySQLConfig` (inherited by MariaDB), plus a `usage: BaseUsageConfig` window.
- Pluggable `usage_source`:
  - `performance_schema` (default): reads normalized statement digests from `performance_schema.events_statements_summary_by_digest`. Relies only on the always-on `statements_digest` consumer and `SELECT` on `performance_schema`. No per-user attribution (digests are aggregated across users).
  - `general_log` (opt-in): reads literal statements with the executing user and timestamp from `mysql.general_log`, tracks per-session schema via `Init DB`/`USE` rows, and supports `email_domain` to map LDAP/db logins to user emails (raw username kept when unset).
- Statements are fed through a `SqlParsingAggregator` to emit dataset usage statistics, top queries, and table/column lineage. Two-tier aware: the schema is treated as `default_schema` so digests resolve to `schema.table` URNs.
- Failures reading the usage source (disabled consumer, missing grant) are reported as warnings and do not abort the run.
- Adds the `USAGE_STATS` capability to both `MySQLSource` and `MariaDBSource`, and extracts dataJob-to-dataJob lineage from `CALL`/`EXEC` statements in stored procedures.
- Makes the MySQL/MariaDB profiling golden tests deterministic (`max_workers: 1`, `query_combiner_enabled: false`) to stop the flaky `fieldProfiles`/`rowCount` drops under CI concurrency.

### Caveats (documented in the connector docs)

- `performance_schema` mode has no per-user attribution; digest text is normalized (literals → `?`) and truncated to `performance_schema_max_digest_length`, and MariaDB's `...` variadic truncation can cause lineage loss.
- `COUNT_STAR` is cumulative since the digest table was last reset; executions are attributed to the bucket of each digest's `LAST_SEEN`.

> Supersedes #17969 — same change set, but raised from a branch on the main repo instead of a personal fork.

## Checklist

- [x] Tests added/updated (unit tests for digest/general_log mapping and aggregator wiring; MySQL + MariaDB integration test suites for both usage sources)
- [x] Docs added/updated (MySQL/MariaDB prerequisites + limitations)


max-datahub (commented):
Ran this against live MySQL 8.4.10 and MariaDB 11.4 containers — both usage modes, the stored-proc golden, and the connection tests — plus the unit suites. All green. Notes below are mostly about tightening assertions, not the behavior itself, which looks correct.

### What's genuinely well covered

- **Real backends, not sqlite/mocks.** Both `performance_schema` and `general_log` are exercised end-to-end on MySQL and MariaDB, driven by an actual query workload against the live server.
- **The per-user vs. aggregated distinction is pinned down.** `general_log` asserts `userCounts` includes `root`; `performance_schema` asserts there is none. That's the real semantic difference between the two sources, and it's the right thing to lock in.
- **Stored-proc dataJob→dataJob `CALL` lineage has an exact golden on a real instance**, including the nested case — `refresh_customer_analytics` resolves to `inputDatajobs: [process_customer_data, generate_order_summaries]`. That's the hardest part of the PR and it's the best-covered.
- **Determinism handling is thoughtful:** pinned wide usage window for the live-clock path so a small gap can't push queries out of range, and profiling disabled where it forced the order-sensitive golden comparison and caused drift.

### Gaps — all "produced correctly today, but nothing catches a regression"

**1. Column-level lineage isn't asserted anywhere.** The PR sells table *and column* lineage, and it does work — a live run emits `fineGrainedLineages` `id→id` and `name→name` on `processed_customers`. But `assert_query_lineage_present` in `tests/test_helpers/mysql_usage_helpers.py` only checks the coarse `upstreams`. If CLL silently degraded to coarse-only, every test would still pass. Worth asserting `fineGrainedLineages` in the shared helper.

**2. `topSqlQueries` isn't asserted — and it's the one field that distinguishes the two modes.** `performance_schema` emits normalized digests (`WHERE id = ?`), `general_log` emits literal SQL (`WHERE id = 1`). That difference is the whole reason two sources exist, and nothing checks it — a run that fed the wrong text, or effectively swapped the modes, would stay green. Asserting the normalized form for `performance_schema` and the literal form for `general_log` would nail the contract.

**3. `general_log` session/schema tracking is only mock-tested.** The trickiest new logic — learn a session's current DB from `Connect`/`Init DB`/`USE`, then drop queries from sessions with no known schema — is only hit with mocked rows in the unit tests. The integration `general_log` test runs a single `root` session connected straight into the target DB, so the multi-session / `USE`-switching path never runs against a real `mysql.general_log` table (whose row format also shifts a little across versions). A second session that `USE`s a different schema in the workload would give this real coverage and guard the filtering logic.

Gaps 1 and 2 are ~5 lines in `mysql_usage_helpers.py` and cover both platforms at once, since MySQL and MariaDB share the helper.

max-datahub (approved):

PR #17990, opened 2026-06-21
test(ingest/airflow): tolerate spurious Airflow 3 DagRun-state race

treff7es:
## Summary

The Airflow 3 integration tests (`test_airflow_plugin`) have become flaky, failing with `ValueError: DAG failed` — and recently the flakiness spread to even the most trivial DAGs (e.g. `simple_dag`, whose only tasks are `BashOperator`s that `echo` a string and cannot fail).

Investigating a failing run showed the harness's own task dump reporting **every task instance as `success`** while the **DagRun itself was reported `failed`**:

```
Task execution order for DAG simple_dag:
=== Task: task_1 (State: success; Try: 1) ===
=== Task: run_another_data_task (State: success; Try: 1) ===
```

### Root cause

On Airflow 3, tasks run out-of-process and report their final state to the scheduler asynchronously via the execution API. Under CI load this opens a window where the scheduler finalizes the DagRun as `failed` even though all task instances actually succeeded. This did not occur on Airflow 2 (in-process execution), which is why the flake appeared only after the move to Airflow 3.x.

The test harness's `_wait_for_dag_finish` trusted the DagRun-level state and raised `ValueError("DAG failed")` the moment it saw `failed`, without checking whether any task actually failed — so a scheduler race was reported as a hard test failure.

### Fix

`_wait_for_dag_finish` now inspects the individual task instances when a failure is reported and only raises if a task genuinely failed (`failed`/`upstream_failed`). An all-success DagRun mislabeled `failed` is logged and treated as the spurious scheduler race it is. The `success=False` path (the snowflake DAGs that are *expected* to fail) is unchanged.

## Checklist

- [x] Tests updated (test-only change to the integration harness)
- [ ] Docs updated — n/a
- [ ] Breaking changes — none

### DIFF
--- a/metadata-ingestion-modules/airflow-plugin/tests/integration/test_plugin.py
+++ b/metadata-ingestion-modules/airflow-plugin/tests/integration/test_plugin.py
@@ -151,36 +151,63 @@ def _wait_for_dag_finish(
         # On Airflow 3 the scheduler runs tasks out-of-process and learns their
         # final state asynchronously via the execution API. Under load this opens
         # a window where it finalizes the DagRun as "failed" even though every
-        # task instance actually succeeded. Before treating a failed DagRun as a
-        # real failure, inspect the individual task states and only raise if a
-        # task genuinely failed; otherwise the DagRun-level "failed" is spurious.
-        failed_tasks = _get_failed_task_ids(
+        # task instance actually succeeded (the harness's own task dump shows all
+        # tasks "success" while the DagRun is "failed"). Before trusting the
+        # DagRun-level "failed", inspect the individual task instances:
+        #   - if a task genuinely failed, it's a real failure -> raise;
+        #   - if any task is still settling (non-terminal), the run hasn't really
+        #     finished, so retry instead of declaring a spurious success;
+        #   - only when every task is terminal and none failed do we treat the
+        #     DagRun-level "failed" as the spurious scheduler race it is.
+        # Same "task succeeded but marked failed" scheduler-race symptom:
+        # https://github.com/apache/airflow/issues/34339
+        task_instances = _get_task_instances(
             airflow_instance, dag_id, dag_run["dag_run_id"]
         )
+        failed_tasks = [
+            ti["task_id"] for ti in task_instances if ti["state"] in _FAILED_TASK_STATES
+        ]
         if failed_tasks:
             raise ValueError(f"DAG failed; failed tasks: {failed_tasks}")
-        print(
-            f"DAG {dag_id} reported state=failed but no task instance failed; "
-            "treating as success (spurious Airflow 3 DagRun-state race)."
+
+        unfinished_tasks = [
+            ti["task_id"]
+            for ti in task_instances
+            if ti["state"] not in _TERMINAL_TASK_STATES
+        ]
+        if unfinished_tasks:
+            raise NotReadyError(
+                f"DAG reported state=failed but tasks have not settled yet: "
+                f"{unfinished_tasks}"
+            )
+
+        logger.warning(
+            "DAG %s reported state=failed but every task instance is terminal "
+            "and none failed; treating as success (spurious Airflow 3 "
+            "DagRun-state race).",
+            dag_id,
         )
         return
 
     raise NotReadyError(f"DAG has not finished yet: {state}")
 
 
-def _get_failed_task_ids(
+# Airflow task-instance states. A task in a terminal state will not transition
+# again on its own; the rest are still settling and warrant a retry.
+_FAILED_TASK_STATES = frozenset({"failed", "upstream_failed"})
+_TERMINAL_TASK_STATES = frozenset(
+    {"success", "failed", "upstream_failed", "skipped", "removed"}
+)
+
+
+def _get_task_instances(
     airflow_instance: AirflowInstance, dag_id: str, dag_run_id: str
-) -> list[str]:
+) -> list[dict[str, Any]]:
     res = _make_api_request(
         airflow_instance.session,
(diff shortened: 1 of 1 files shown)


metadata-ingestion-modules/airflow-plugin/tests/integration/test_plugin.py
```diff
@@ -138,13 +138,49 @@ def _wait_for_dag_finish(
...
+        # a window where it finalizes the DagRun as "failed" even though every
+        # task instance actually succeeded. Before treating a failed DagRun as a
+        # real failure, inspect the individual task states and only raise if a
+        # task genuinely failed; otherwise the DagRun-level "failed" is spurious.
+        failed_tasks = _get_failed_task_ids(
+            airflow_instance, dag_id, dag_run["dag_run_id"]
+        )
+        if failed_tasks:
+            raise ValueError(f"DAG failed; failed tasks: {failed_tasks}")
+        print(
```
sgomezvillamor: do we `print`? or should this be a logger debug/warning instead?

metadata-ingestion-modules/airflow-plugin/tests/integration/test_plugin.py
```diff
@@ -138,13 +138,49 @@ def _wait_for_dag_finish(
...
+) -> list[str]:
+    res = _make_api_request(
+        airflow_instance.session,
+        f"{airflow_instance.airflow_url}/api/v2/dags/{dag_id}/dagRuns/{dag_run_id}/taskInstances",
+    )
+    task_instances = res.json()["task_instances"]
+    return [
+        ti["task_id"]
+        for ti in task_instances
+        if ti["state"] in ("failed", "upstream_failed")
```
sgomezvillamor: What if a task is still running or hasn't started yet?
Should we first check all tasks finished and then checking their final status (success, failure, cancelled, etc.)?

metadata-ingestion-modules/airflow-plugin/tests/integration/test_plugin.py
```diff
@@ -138,13 +138,49 @@ def _wait_for_dag_finish(
...
+        # real failure, inspect the individual task states and only raise if a
+        # task genuinely failed; otherwise the DagRun-level "failed" is spurious.
+        failed_tasks = _get_failed_task_ids(
+            airflow_instance, dag_id, dag_run["dag_run_id"]
+        )
+        if failed_tasks:
+            raise ValueError(f"DAG failed; failed tasks: {failed_tasks}")
+        print(
+            f"DAG {dag_id} reported state=failed but no task instance failed; "
+            "treating as success (spurious Airflow 3 DagRun-state race)."
```
sgomezvillamor: is there any public doc about this DagRun-state race issue that we could link here?

metadata-ingestion-modules/airflow-plugin/tests/integration/test_plugin.py
treff7es: Good call — switched to `logger.warning(...)` (with lazy `%s` formatting) instead of `print`. Done in 412fcb6.

metadata-ingestion-modules/airflow-plugin/tests/integration/test_plugin.py
treff7es: Great catch — this was the real gap. The check now categorizes task instances into three buckets:

- any task in `failed`/`upstream_failed` → genuine failure, raise `ValueError`;
- any task still **non-terminal** (`running`/`queued`/`scheduled`/`None`/…) → the run has not actually settled, so raise `NotReadyError` and let the existing `@tenacity.retry` loop wait instead of declaring a false success;
- only when **every** task instance is terminal and none failed do we treat the DagRun-level `failed` as the spurious race.

Terminal states are defined as `{success, failed, upstream_failed, skipped, removed}`. If tasks genuinely never settle, the retry budget (`stop_after_delay(90)`) is exhausted and `NotReadyError` is reraised, so a stuck run surfaces as a timeout rather than a silent pass. Done in 412fcb6.

metadata-ingestion-modules/airflow-plugin/tests/integration/test_plugin.py
treff7es: There is no authoritative upstream doc for the exact Airflow 3 out-of-process variant, but the same "task ran successfully yet the run is marked failed" scheduler-race symptom is reported in apache/airflow#34339. Linked it in the comment. Done in 412fcb6.

treff7es: Correction: #34339 was an Airflow 2.7.1 report, which is the wrong reference since this flake only appears on Airflow 3.0 (out-of-process task execution). Updated the comment to cite the Airflow-3.0-specific execution-API state-mismatch reports instead: apache/airflow#53797 and apache/airflow#52813. Done in bdb05ba.

metadata-ingestion-modules/airflow-plugin/tests/integration/test_plugin.py
```diff
@@ -138,13 +138,78 @@ def _wait_for_dag_finish(
...
+
+# Airflow task-instance states. A task in a terminal state will not transition
+# again on its own; the rest are still settling and warrant a retry.
+_FAILED_TASK_STATES = frozenset({"failed", "upstream_failed"})
+_TERMINAL_TASK_STATES = frozenset(
+    {"success", "failed", "upstream_failed", "skipped", "removed"}
+)
+
+
+def _get_task_instances(
```
sgomezvillamor, line 205: Instead of returning `dict[str, Any]` — which can't be narrowed and forces the caller to know the Airflow API model — why not return `dict[str, str]` with just the state per task?

sgomezvillamor (approved):

PR #17992, opened 2026-06-22
perf(ingest/snowflake): compose allow/deny patterns into a single RLIKE for metadata pushdown

rob-1019:
# perf(ingest/snowflake): compose allow/deny patterns into a single RLIKE for metadata pushdown

## What

When metadata pattern pushdown is enabled, the Snowflake source previously emitted
one `RLIKE` predicate per allow pattern (OR'd) and one `NOT RLIKE` per deny pattern
(AND'd) in the `information_schema` filter. This composes each list into a **single
alternation** — `RLIKE '(p1)|(p2)|...'` and `NOT RLIKE '(p1)|(p2)|...'` — so the WHERE clause has one `RLIKE` evaluation instead of N, with
identical filtering semantics.

**~11× faster** on the pushed-down `information_schema.tables` query in a 500-pattern
benchmark (15.0s → 1.3s, same 498 rows).

## Flag

Gated entirely behind the **existing** `push_down_metadata_patterns` flag (default
`false`) — **no new config**, and the default (flag off) path is unchanged. No other
flags needed.

## Notes

- Each pattern is wrapped in its own group before being joined with `|`, i.e.
  `(p1)|(p2)|...`. This keeps the composed alternation exactly equivalent to the
  original one-RLIKE-per-pattern behavior: it isolates each pattern's anchors
  (`^`/`$`) and any internal `|` from binding across pattern boundaries or
  interacting with Snowflake's implicit whole-pattern anchoring. It is purely a
  regex-grouping/equivalence measure — not a security one; SQL-injection safety is
  handled separately by escaping each literal. Snowflake `RLIKE` is POSIX ERE, so
  the group is a capturing group `(…)`, not a PCRE non-capturing group `(?:…)`
  (which ERE rejects).
- **Deny** is composed server-side as a best-effort row reduction, capped at half the
  per-query byte budget; the push-down table path always re-applies the full
  `table_pattern` client-side (via the existing `is_dataset_pattern_allowed`) as the
  correctness backstop. Unlike allow, deny can't be split across queries without
  leaking excluded objects.
- If the composed filter would exceed Snowflake's per-query size limit, the filter is
  skipped and applied client-side instead (logged at WARNING).

## Testing

- Unit: `test_snowflake_queries.py` + `test_snowflake_schema.py` (255 pass); ruff +
  mypy clean.
- Real Snowflake ingestion with combined allow + deny + pushdown: correct results,
  denied tables excluded.
- Scale: ~3,000 allow + ~3,000 deny on each of database/schema/table lists composes to
  a single ~192 KB query and matches a small-recipe run; randomizing pattern order
  yields byte-identical output (composition is order-independent).
- Over-limit fallback: a ~5 MB allow list triggers the client-side fallback and
  ingests correctly.

## Checklist

- [x] PR Title Format
- [x] Tests added/updated
- [x] Docs updated (`push_down_metadata_patterns` description)
- [x] No breaking changes (opt-in flag, default path unchanged)


sgomezvillamor (approved):
Nice work — the composition logic is correct, the fail-closed semantics are right, and the adversarial test suite is thorough. A few small things before merge:

---

**🟡 Query-size-limit fallback not surfaced in the ingestion report**

In `get_tables_for_schema`, when the composed filter exceeds `SNOWFLAKE_MAX_QUERY_BYTES` the fallback only calls `logger.warning(...)`. Users monitoring ingestion in the DataHub UI will never see it — they'd have to tail container logs to understand why pushdown is slower than expected. Suggest adding a structured report call alongside the logger:

```python
self.structured_reporter.warning(
    "Pushdown filter too large",
    f"table_pattern filter for {db_name} is {len(table_filter.encode())} bytes "
    f"(>{SNOWFLAKE_MAX_QUERY_BYTES}); falling back to client-side filtering (slower).",
)
```

---

**ℹ️ Allow-pattern-dropped warning doesn't mention data loss**

In `_build_pattern_filter`, when `_make_composable` returns `None` for an allow pattern the message says:

> "This pattern will not be applied as a SQL filter."

With pushdown on there's no client-side allow backstop, so tables matching *only* that pattern are silently excluded from ingestion. The message doesn't communicate this consequence. Suggested wording:

> "This pattern will not be applied as a SQL pushdown filter and has no client-side fallback — tables matching only this pattern will be missing from ingestion. Disable `push_down_metadata_patterns` if this pattern is required."

---

**ℹ️ Stale comment in test file**

```python
# ===========================================================================
# Query-optimization unit tests (merged from test_snowflake_query_optimizations.py):
#   _build_pattern_filter composition/security, _make_composable fast path
# ===========================================================================
```

This documents where the tests *came from*, not what they do. The class names below (`TestBuildPatternFilterSecurity`, `TestBuildPatternFilterComposition`, etc.) are already self-describing — fine to drop this block comment.

---

PR #17994, opened 2026-06-22
fix(ingest/confluence): treat pages.allow as crawl seeds so recursive child pages are ingested

nwadams:
## Summary

When the Confluence connector is configured with `pages.allow` and `recursive: true`, only the root/seed pages are ingested — their child pages are silently dropped.

```yaml
source:
  type: confluence
  config:
    pages:
      allow:
        - "123456"   # a root page
    recursive: true   # expected: ingest this page AND its descendants
```

### Root cause

`_get_pages_from_page_allow` correctly crawls the descendants of each configured page (honoring `recursive`). But `_create_document_entity` then re-checked **every** collected page against `pages.allow` via `_is_page_allowed`:

```python
if self.config._parsed_page_allow is not None:
    if page_id not in self.config._parsed_page_allow:
        return False
```

Recursively-discovered children are not literally present in `pages.allow`, so they were filtered out at emission time, leaving only the seeds. (The skip was also mislogged as "Filtered by pages.deny".)

`pages.allow` is documented as selecting "specific pages **and their children**" — i.e. it is a set of *roots* that seed the crawl, not a per-page whitelist.

### Fix

Replace the emission gate `_is_page_allowed` with a deny-only `_is_page_denied`. Any page that reaches emission has already been scoped by the crawler, so it is kept unless explicitly denied via `pages.deny`. Space-based ingestion and `recursive: false` are unaffected — in those paths `pages.allow` is unset, so only `pages.deny` ever applied.

This mirrors the same fix made for the **Notion** connector in #17523, which removed an analogous root-id check that dropped recursively-discovered descendants.

### Testing

- Migrated existing filtering tests to the deny-only contract.
- Added end-to-end coverage:
  - `test_page_allow_recursive_ingests_full_tree` — `pages.allow: ["20001"]` + `recursive: true` now collects the full tree `{20001, 20002, 20003, 20004, 20005, 20006}` and denies none of them.
  - `test_page_allow_non_recursive_ingests_only_seed` — `recursive: false` collects only the seed.
- `ruff format`, `ruff check`, `mypy`, and the focused unit suites (59 tests) all pass.

## Checklist

- [x] Links to related issues (mirrors prior Notion fix #17523)
- [x] Tests added/updated

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/confluence/confluence_source.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/confluence/confluence_source.py
@@ -425,27 +425,28 @@ class ConfluenceSource(StatefulIngestionSourceBase, TestableSource):
 
         return True
 
-    def _is_page_allowed(self, page_id: str) -> bool:
+    def _is_page_denied(self, page_id: str) -> bool:
         """
-        Check if page is allowed by pages.allow/pages.deny lists.
+        Check if a page is explicitly excluded by the pages.deny list.
+
+        pages.allow is intentionally NOT consulted here. It is a list of *root*
+        pages that seed the crawl (see _get_pages_from_page_allow); when
+        recursive=true the crawler walks the descendants of those roots, so
+        every page that reaches this point is in scope by construction.
+        Re-checking membership in pages.allow at emission time would drop those
+        legitimately-discovered child pages and silently defeat recursive
+        ingestion (only the seed/root pages would survive).
 
         Args:
             page_id: Confluence page ID
 
         Returns:
-            True if page should be ingested, False otherwise
+            True if the page is denied and should be skipped, False otherwise
         """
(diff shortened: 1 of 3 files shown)


nwadams:
confirmed recursive documents ingested
```
source:
    type: confluence
    config:
        url: 'https://datahub-integration-testing.atlassian.net/wiki'
        username: me
  
      api_token: '${CONFLUENCE_API_TOKEN}'
        cloud: true
        pages:
            allow:
                - '16351389'
        recursive: true
        max_documents: -1
sink:
    type: datahub-rest
    config:
        server: 'http://localhost:8080'
        token: '${DATAHUB_GMS_TOKEN}'

```

Screenshot 2026-06-23 at 12 30 07 PM

shirshanka (approved):
LGTM!

PR #18001, opened 2026-06-23
fix(ingest/redshift): avoid lineage cursor timeout; usage via SqlParsingAggregator with optional column-level usage

treff7es:
## Summary

Related Redshift ingestion improvements (see commits for the clean split):

1. **Fix lineage cursor timeout** — `RedshiftSqlLineage` streamed lineage/temp-table rows straight from a live Redshift cursor and called the slow `SqlParsingAggregator.add_observed_query(...)` inline per row, holding the cursor open through the entire aggregation. On large query histories Redshift can time out the cursor mid-stream. Now each result set is drained into a local `FileBackedList` first (releasing the cursor quickly), then aggregated from the cache — mirroring the Snowflake queries-v2 pattern. Applied at every chokepoint (`_populate_lineage_agg`, the temp-table loop in `build()`, and the unified feed below).

2. **Generate usage via `SqlParsingAggregator`** — replace the bespoke `RedshiftUsageExtractor` aggregation with the shared aggregator, fed `PreparsedQuery` whose upstream is the already-resolved scanned table from `stl_scan`. This preserves Redshift's table-level accuracy (no reliance on parsing the query text to find tables) while consolidating on the standard usage path. Both old and new paths funnel through the same `make_usage_workunit`, so counts, top queries, buckets, and user URNs are unchanged. One behavioral change: the per-user `userEmail` field is no longer populated (the aggregator builds user identity from a single URN); the CorpUser URN is preserved, so existing links are unchanged. Golden files re-baselined accordingly.

3. **Optional column-level usage** — new `include_column_usage_stats` config (default **false**). When enabled, usage is generated by parsing the query text instead of `stl_scan` attribution, which yields **column-level usage statistics** (`fieldCounts`). The full query text is reconstructed from `STL_QUERYTEXT` / `SYS_QUERY_TEXT` (the same boundary-aware LISTAGG source used by lineage), not the truncated `stl_query.querytxt`.

4. **Unify the parsing path (queries-v2)** — when `include_column_usage_stats` is enabled, lineage **and** usage **and** column-usage **and** Query entities are produced by a **single** `SqlParsingAggregator` from **one** parse of each query, fed by a new DB-scoped `list_all_queries_sql` (every statement in the window, reads + writes, not pre-filtered by table). This removes the previous double-parse where the lineage aggregator and a second usage aggregator each parsed the overlapping write queries. The current run's schemas are registered into the aggregator's resolver (from the already-fetched `all_tables`, no extra DB round-trip) so column references — including `SELECT *` — resolve without a pre-populated backend graph. Unused tables still get an empty usage aspect via `auto_empty_dataset_usage_statistics`. When `include_column_usage_stats` is on with all lineage flags off, the unified feed still runs for usage but emits no lineage edges (`generate_lineage` follows the lineage flags).

5. **Apply `table_pattern` to SQL-parsing lineage/usage** — `is_allowed_table` is now wired into both Redshift `SqlParsingAggregator` instances, so parsed lineage/usage/query entities respect `table_pattern` (consistent with Snowflake, BigQuery, Unity and Dremio, which already pass it). Redshift previously left this aggregator filter dormant. Default `table_pattern` allows everything, so most setups are unaffected.

Operations remain on the existing extractor (table-accurate via `stl_insert`, no per-row SQL parsing → no cursor-timeout exposure).

Note: the column-level usage flag is named `include_column_usage_stats`, matching the equivalent toggle in the Databricks/Unity connector.

## Observability

The source report now distinguishes where time goes and surfaces both usage paths:

- `report.usage_query_fetch_timer` vs `report.usage_parsing_timer` — splits Redshift query-fetch (cursor drain) from aggregator feed/parse, so a slow run is diagnosable as DB-bound vs parse-bound (mirrors Databricks' `usage_query_fetch_timer` / `usage_parsing_timer`).
- `report.usage_aggregator` — surfaces the default (stl_scan) path's `SqlParsingAggregator` report (parse timers, filter skips, query counts), which was previously a blind spot; the v2 path already exposed `report.sql_aggregator`.

## Default (stl_scan) vs `include_column_usage_stats` — live comparison

Ran both routes against a live cluster with identical config except the flag (same window, profiling off):

| Metric | Default (stl_scan) | `include_column_usage_stats` |
| --- | --- | --- |
| Wall clock | 239 s | 286 s (+20%) |
| Peak RSS | 292 MB | 292 MB (flat) |
| `upstreamLineage` aspects | 9 | 9 (unchanged) |
| Query entities | 32 | 32 (unchanged) |
| `datasetUsageStatistics` aspects | 6 | 9 |
| …with `fieldCounts` (column-level) | 0 | 9 |
| `usage_query_fetch_timer` | 17.9 s | 6.1 s |
| `usage_parsing_timer` | 0.01 s | 0.31 s |

Findings / conclusions:

- **Lineage and Query entities are identical** across routes — the unification adds usage/columns without perturbing lineage.
- The usage-target differences are a **semantics** difference, not new entities (the unique dataset set is identical): with parsing, usage is attributed to the **view / external table the query named** (3 views + 1 external table that stl_scan misses, since stl_scan credits the physically-scanned base table), and a write-only `INSERT … VALUES` is **not** miscounted as usage (stl_scan counts its target as a read). So column-usage mode is also marginally more correct for read attribution, not just richer.
- **Cost is fetch-bound, not parse-bound:** parsing is ~0.3 s; the time is in system-table queries. Notably the default path's `stl_scan` usage query (heavy `svv_table_info` join, 17.9 s) is *slower to fetch* than the v2 `list_all_queries` (6.1 s). Memory is flat (FileBackedList drain), so the only real cost of column-usage mode is the extra lineage-stage query time.

## Testing

- Unit tests: lineage cursor drain-order; end-to-end column-level usage through the real aggregator (explicit columns **and** `SELECT *` via registered schema); unified feed produces lineage edges from `INSERT…SELECT`; usage-only-v2 (lineage flags off) produces usage but no lineage edges; `list_all_queries_sql` DB-scoping (provisioned + serverless); full-text reconstruction. Existing usage golden tests pass unchanged on the default path. 97 redshift unit + usage-integration tests pass; the table_pattern filter and usage._user_urn paths are unit-tested.
- `ruff` + `mypy` clean.
- **Validated against a live serverless Redshift cluster**: the unified `list_all_queries_sql` and lineage execute without error; a controlled `SELECT id, name FROM public.best` produced `datasetUsageStatistics` with `fieldCounts [(id,1), (name,1)]` on `dev.public.best` — confirming the full-text query + schema-resolver registration deliver real column-level usage end-to-end through the single aggregator.

## Checklist

- [x] Tests added/updated
- [x] Docs updated (new config option documented inline)
- [ ] Breaking changes: none (default behavior unchanged; `userEmail` omission in usage is the only output delta and only on the modernized usage path)


metadata-ingestion/src/datahub/ingestion/source/redshift/lineage.py
```diff
@@ -600,10 +634,76 @@ def build(
...
+                                    session_id=str(row[idx_session_id]),
+                                )
+                            )
+                        rows = cursor.fetchmany()
+                with self.report.usage_parsing_timer:
+                    for observed_query in observed:
+                        self.aggregator.add_observed_query(observed_query)
+        except Exception as e:
+            self.report.warning(
+                title="Failed to extract some lineage/usage",
```
askumar27: 🟡 **In queries-v2 mode this `except` block is the only producer of both lineage AND usage**, yet a total failure here is reported as a *warning* and the run still exits successfully.

The most likely real-world failure for a brand-new system-table query is a missing `SELECT` grant on `STL_QUERYTEXT`/`SYS_QUERY_HISTORY`/`svl_user_info` — `cursor.execute` throws before the first row, and lineage + usage for the whole database silently collapse to zero. Worse, the v2 empty-usage backfill (`auto_empty_dataset_usage_statistics` in `redshift.py`) then emits a zero-usage aspect for every known table, so a silently-failed extraction is **indistinguishable from a healthy run where nothing was queried**. `report_status("extract-all-queries", False)` only feeds the redundant-run-skip checkpoint; it does not fail the run.

Suggest tracking rows-fetched/parsed counters and escalating to `self.report.failure()` when the feed yields zero observed queries due to a pre-fetch exception, so a permissions error is distinguishable from a genuinely empty window. (Reporting partial per-query parse failures as a warning is fine.)

metadata-ingestion/src/datahub/ingestion/source/redshift/config.py
```diff
@@ -217,6 +228,16 @@ def check_database_is_set(self) -> "RedshiftConfig":
         assert self.database, "database must be set"
         return self
 
+    @model_validator(mode="after")
+    def warn_column_usage_without_usage(self) -> "RedshiftConfig":
```
askumar27: 🟡 This is a user misconfiguration (a feature was enabled that will silently produce nothing), but it's surfaced only to server logs. Per the module convention, user-visible problems should reach the ingestion report — users see the report, not logs.

A pydantic `model_validator` genuinely can't access `self.report` (it doesn't exist at parse time), so the cleanest fix is to move this check into the source's init / connection-test path and emit `self.report.warning(title=..., message=..., context=...)` there. As written, a user who flips `include_column_usage_stats` without `include_usage_statistics` gets zero column usage and no explanation in the report.

metadata-ingestion/src/datahub/ingestion/source/redshift/usage.py
```diff
@@ -215,28 +229,44 @@ def _get_workunits_internal(
...
+                            for event in access_events:
+                                aggregator.add_preparsed_query(
+                                    self._access_event_to_preparsed_query(event)
+                                )
+
+                    for wu in auto_workunit(aggregator.gen_metadata()):
+                        self.report.num_usage_workunits_emitted += 1
+                        yield wu
+                finally:
+                    aggregator.close()
```
askumar27, line 269: 🟡 `aggregator.close()` in `finally` is the right intent, but the `yield` earlier in this `try` means `close()` only runs on full generator exhaustion or GC-driven `GeneratorExit`. If the pipeline abandons this generator mid-stream (e.g. an upstream error), the file-backed `SqlParsingAggregator` leaks its temp files / handles non-deterministically.

This is still better than before (which never closed at all), so non-blocking — but to be fully correct, don't suspend across the resource scope: materialize `gen_metadata()` before closing, or let the caller own the aggregator lifetime.

metadata-ingestion/src/datahub/ingestion/source/redshift/redshift.py
```diff
@@ -425,30 +430,52 @@ def _extract_metadata(
                 self.process_schemas(connection, database)
             )
 
-            # Only extract lineage if at least one lineage flag is enabled.
-            # This addresses a regression introduced in PR #14580 where lineage v1 removal
-            # inadvertently caused lineage extraction to run even when all flags were disabled.
-            if (
+            lineage_enabled = (
```
askumar27: ℹ️ This 6-flag `lineage_enabled` boolean is duplicated verbatim in `lineage.py` (where it drives the aggregator's `generate_lineage`). If a new lineage flag is ever added, both copies must be updated in lockstep or behavior diverges silently. Consider extracting it to a single helper/property on the config.

askumar27 (approved):
Reviewed against the metadata-ingestion PR guidelines. Well-engineered and unusually well-documented refactor — the cursor-drain fix, the aggregator consolidation (queries-v2), and the new observability timers/`usage_aggregator` report are all sound, and the column-level path is correctly gated behind an opt-in flag defaulting to `false`. Behavioral tests (cursor-drain ordering, lineage from the unified feed, `table_pattern` filtering) are high quality, and the `userEmail` / `table_pattern` behavior changes are documented in `updating-datahub.md`.

Approving — none of my notes block merge. The one worth addressing is the error-severity asymmetry in `_populate_unified_queries` (inline): in queries-v2 mode a permissions error silently zeroes both lineage and usage while the run reports success, and the empty-usage backfill disguises it as healthy.

**Non-blocking follow-ups (not inlined):**
- No integration/golden test drives `include_column_usage_stats=True` through the full source (`_extract_metadata` v2 gate + `auto_empty` backfill); unit tests call `_populate_unified_queries` directly and bypass that wiring.
- `test_populate_unified_queries_produces_column_level_usage` comments that `SELECT *` resolves, but the query under test uses explicit columns — the schema-driven `*` expansion isn't actually exercised.
- The usage extractor's own `_is_allowed_table` (default stl_scan path) has no `table_pattern` deny assertion.
- nit: serverless `list_all_queries_sql` hardcodes `qt.sequence < 16` while the provisioned variant templates `_QUERY_SEQUENCE_LIMIT` — use a named constant for parity.
- nit: `test_usage_only_via_sql_parsing_no_lineage_edges` asserts `aggregator.generate_lineage is False` (white-box); the aspect-level assertions already prove the contract.

treff7es:
Thanks for the review @askumar27 — addressed in 7e331e5:

- **Unified-feed failure reported as warning (lineage.py)** — agreed; in queries-v2 mode this feed is the sole producer of usage (and the main SQL-parsed lineage), so a total failure (e.g. a missing `SELECT` grant on `STL_QUERYTEXT`/`SYS_QUERY_HISTORY`) now calls `report.failure` instead of `report.warning`, so the run no longer exits successfully. Added a regression test asserting the failure path is reported as a failure, not a warning.
- **Misconfig surfaced only to logs (config.py)** — moved into the ingestion report: the `include_column_usage_stats` + `include_usage_statistics=false` combination now emits `report.report_warning` from `_warn_deprecated_configs` (where `self.report` exists), and the log-only pydantic validator was removed.
- **Duplicated `lineage_enabled` 6-flag boolean (redshift.py)** — extracted to a single `RedshiftConfig.lineage_enabled` property, now used by both the source-level gating and the aggregator's `generate_lineage`.

- **`aggregator.close()` in `finally` (usage.py)** — kept as-is. This is equivalent to the existing `with RedshiftSqlLineage(...)` lineage path: both are generators whose cleanup runs on `GeneratorExit` (which the pipeline triggers on close) and the `finally` additionally covers exceptions. The only residual case — GC-timed cleanup on abnormal mid-stream abandonment — applies identically to the existing lineage extractor, so there's no improvement here without restructuring the whole generator chain.

---

On the **Aikido "SQL injection via string concatenation"** findings (`query.py`, `list_all_queries_sql`): these are accepted false positives. They are the same controlled-input SQL-string pattern already used ~36× across this file (and in the Snowflake/BigQuery query builders) — flagged only because the lines are new. The interpolated values are not attacker-controllable: `start_time`/`end_time` are `strftime` outputs, and `database` is the recipe/catalog database name (trusted operator config). These system-table queries execute via `get_query_result(conn, query_string)`, which does not accept bind parameters, so parameterizing just this query isn't supported by the execution path and would be inconsistent with the rest of the connector.

PR #18002, opened 2026-06-23
feat(kafka-connect): add Iceberg Sink Connector support

davidwalker-tp:
## Summary

Add lineage extraction support for the Apache Iceberg Sink Connector (`org.apache.iceberg.connect.IcebergSinkConnector`) in the Kafka Connect ingestion source.

This enables DataHub to automatically recognize and extract lineage from Iceberg Sink Connectors, mapping Kafka topics to Iceberg tables.

### Code Implementation
- **New IcebergSinkConnector class** in `sink_connectors.py`:
  - Parses `iceberg.tables` configuration (comma-separated list of fully-qualified table names)
  - Implements topic resolution with runtime intersection logic (consistent with ClickHouse pattern)
  - Emits all-to-all (N×M) lineage for multi-table fan-out scenarios (dynamic routing via record field values)
  - Dataset naming: `namespace.table` format (e.g., `mydb.events`)

- **Configuration constants** in `config_constants.py`:
  - `ICEBERG_TABLES` for the target tables list
  - `ICEBERG_TABLES_DYNAMIC_ENABLED` for dynamic routing flag
  - `ICEBERG_CATALOG` for catalog reference

- **Registry integration** in `connector_registry.py`:
  - Registered connector class mapping in `_get_sink_connector()`

### Testing
- Added comprehensive test suite in `TestIcebergSinkConnector`:
  - ✅ Single topic → single table
  - ✅ No tables configured (empty lineage)
  - ✅ Multi-topic → single table (N×1)
  - ✅ Multi-topic → multi-table (N×M all-to-all)
  - ✅ Dynamic routing emits all-to-all
  - ✅ Platform is "iceberg"
  - All 6 tests pass

### Documentation
- Updated "Supported Sink Connectors" table in `kafka-connect_post.md`
- Marked as ✅ Full support for both self-hosted and Confluent Cloud
- Documented lineage extraction as "Topic → Table mapping"

## Verification

✅ **Unit Tests:** 6/6 passed  
✅ **Linting:** `ruff check --fix` passed (no errors)  
✅ **Formatting:** `mdPrettierWrite` passed (markdown formatting compliant)

### Why All-to-All (N×M) Lineage for Multi-Table Scenarios?

Iceberg's dynamic routing uses a configurable field in the Kafka message record to determine which table receives each message at runtime. Since DataHub connector configuration is static, we cannot determine the exact topic → table mapping without processing actual message data.

To ensure complete lineage coverage, the connector emits all-to-all lineage: every subscribed topic is connected to every configured table. This is:
- **Conservative:** Never misses a real dependency
- **Consistent:** Matches how other connectors handle uncertainty (BigQuery dynamic routing, etc.)
- **Documented:** The multi-table behavior is noted in test coverage

### Related Issues

Fixes support gap identified when reviewing Kafka Connect source for Apache Iceberg integration patterns. The Iceberg Sink Connector is now on par with ClickHouse and other supported sink connectors.

---


metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
```diff
@@ -503,6 +503,111 @@ def get_platform(self) -> str:
...
+                topics_regex,
+                available_topics=self.connector_manifest.topic_names
+                if self.connector_manifest.topic_names
+                else None,
+            )
+
+        return []
+
+    def extract_flow_property_bag(self) -> Dict[str, str]:
+        return dict(self.connector_manifest.config)
```
treff7es: 🔴 **Blocker — credential leak.** This returns the entire connector config unfiltered, and it becomes `DataFlowInfo.customProperties` (emitted to DataHub and shown in the UI). Iceberg sink configs routinely carry secrets — REST-catalog `iceberg.catalog.credential` (OAuth2 secret) and `iceberg.catalog.token`, plus object-store keys like `iceberg.catalog.s3.secret-access-key` — all published in cleartext.

Every other sink in this file redacts credentials (S3 masks `aws.secret.access.key`, Snowflake `snowflake.private.key`, ClickHouse the password, BigQuery `keyfile`, JDBC the connection password). Please apply the same deny-list masking, and make it **prefix-aware** since Iceberg nests secrets under `iceberg.catalog.*` (drop keys containing `secret`/`token`/`credential`/`password`/`.key`).

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
```diff
@@ -503,6 +503,111 @@ def get_platform(self) -> str:
...
+            topic_list = list(available_topics)
+            logger.debug(
+                f"No subscription config found for {connector_manifest.name}, "
+                f"using all {len(topic_list)} available topics"
+            )
+
+        # All-to-all mapping: each topic → all configured tables
+        topics_to_tables: Dict[str, List[str]] = {}
+        for topic in topic_list:
+            topics_to_tables[topic] = tables if tables else []
```
treff7es: 🟡 **Silent empty-lineage.** When `iceberg.tables` is absent/empty, every topic maps to `[]` and `extract_lineages` returns `[]` with no `report.warning` and no counter — the operator sees a green run that emitted nothing. This is a *valid* configuration: `iceberg.tables.dynamic-enabled=true` routes tables per-record at runtime and legitimately leaves `iceberg.tables` empty.

Please emit an observable signal here, distinguishing "dynamic routing not supported for lineage" from "tables missing/misconfigured".

(Minor: `tables if tables else []` is redundant — `parse_comma_separated_list` already returns a list, so this is always just `tables`. Note also every key aliases the *same* list object; prefer `{topic: tables for topic in topic_list}` or, more honestly, store `topics` + `tables` as two flat lists and cross-product in `extract_lineages`.)

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
```diff
@@ -503,6 +503,111 @@ def get_platform(self) -> str:
...
+        return dict(self.connector_manifest.config)
+
+    def extract_lineages(self) -> List[KafkaConnectLineage]:
+        try:
+            lineages: List[KafkaConnectLineage] = []
+            parser = self.get_parser(self.connector_manifest)
+
+            for topic, target_tables in parser.topics_to_tables.items():
+                for target_dataset in target_tables:
+                    fine_grained = self._extract_fine_grained_lineage(
```
treff7es: 🟡 **Column-level lineage is dead code — wrong helper.** `_extract_fine_grained_lineage` is the *source*-direction helper (DB table → Kafka topic) and hard-returns `None` when `source_platform == "kafka"` (`common.py:966-973`). Since this always passes `source_platform=KAFKA`, `fine_grained` is always `None`.

The purpose-built *sink*-direction helper `_extract_sink_fine_grained_lineage(source_topic, target_dataset, target_platform)` already exists and is used correctly by Snowflake (322), BigQuery (857), and JDBC (1384) in this same file. One-line fix:

```python
fine_grained = self._extract_sink_fine_grained_lineage(
    source_topic=topic, target_dataset=target_dataset, target_platform="iceberg",
)
```

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
```diff
@@ -503,6 +503,111 @@ def get_platform(self) -> str:
         return "clickhouse"
 
 
+@dataclass
+class IcebergSinkConnector(BaseConnector):
+    @dataclass
+    class IcebergParser:
+        tables: List[str]
```
treff7es: `IcebergParser.tables` is written at construction but read nowhere — `extract_lineages` only consumes `topics_to_tables`. Drop the field (and the `tables=tables` arg), or fold both into the two-list model suggested above.

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
```diff
@@ -503,6 +503,111 @@ def get_platform(self) -> str:
...
+            )
+
+        # All-to-all mapping: each topic → all configured tables
+        topics_to_tables: Dict[str, List[str]] = {}
+        for topic in topic_list:
+            topics_to_tables[topic] = tables if tables else []
+
+        return self.IcebergParser(tables=tables, topics_to_tables=topics_to_tables)
+
+    def get_topics_from_config(self) -> List[str]:
```
treff7es: This `get_topics_from_config` is a byte-for-byte 6th copy (S3/Snowflake/ClickHouse/BigQuery/JDBC already have identical ones), and the topic intersection/fallback block above (~521-544) is a 4th copy. `BaseConnector` already declares `get_topics_from_config` as an overridable stub (`common.py:725`), and both bodies touch only `BaseConnector` members. Ideally promote both into `BaseConnector` and delete the duplicates — at minimum, please don't add a 6th copy here. (Base-class extraction across all sinks is fine as a follow-up PR.)

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/config_constants.py
```diff
@@ -75,6 +75,11 @@ class ConnectorConfigKeys:
     CLICKHOUSE_PASSWORD: Final[str] = "password"
     CLICKHOUSE_TOPIC2TABLE_MAP: Final[str] = "topic2TableMap"
 
+    # Iceberg configuration
+    ICEBERG_TABLES: Final[str] = "iceberg.tables"
+    ICEBERG_TABLES_DYNAMIC_ENABLED: Final[str] = "iceberg.tables.dynamic-enabled"
+    ICEBERG_CATALOG: Final[str] = "iceberg.catalog"
```
treff7es: `ICEBERG_TABLES_DYNAMIC_ENABLED` and `ICEBERG_CATALOG` are defined but referenced nowhere in the source or tests. Either wire them up (dynamic-mode handling for the silent-lineage case; catalog-aware platform/instance naming) or remove them — unused constants imply support that doesn't exist.

metadata-ingestion/tests/unit/kafka_connect/test_kafka_connect.py
```diff
@@ -704,6 +705,152 @@ def test_clickhouse_sink_stale_topic_excluded_when_runtime_topics_available(
...
+        report: Mock = Mock(spec=KafkaConnectSourceReport)
+
+        connector = IcebergSinkConnector(manifest, config, report)
+        lineages: List = connector.extract_lineages()
+
+        assert len(lineages) == 2
+        tables = {lin.target_dataset for lin in lineages}
+        assert tables == {"mydb.events", "mydb.users"}
+
+    def test_platform_is_iceberg(self) -> None:
```
treff7es: This is a trivial test — `get_platform` is a one-line `return "iceberg"`, so this can only fail on a typo. Platform correctness is already covered implicitly by the lineage tests asserting `target_platform == "iceberg"`. Suggest removing it.

More valuable to add instead: a test where configured `topics` and runtime `topic_names` **differ** (e.g. `topics="events,stale"`, `topic_names=["events"]` → only `events`), since the stale-topic intersection is the connector's main real logic and is currently a no-op in every test (see Snowflake's `test_...stale_topic_excluded` for the pattern).

treff7es:
@davidwalker-tp thanks for the fix. I left a few small comments; if you can fix those, the PR should be fine.

treff7es (changes requested):

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
davidwalker-tp: ✅ **Fixed in commit:** `fix(kafka-connect): use sink-direction fine-grained lineage helper for Iceberg`

Changed from `_extract_fine_grained_lineage` (source-direction, always returns `None` for Kafka sources) to `_extract_sink_fine_grained_lineage`, matching the pattern used by Snowflake, BigQuery, and JDBC connectors.

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
davidwalker-tp: ✅ **Fixed in commit:** `refactor(kafka-connect): remove unused tables field from IcebergParser`

Removed the `tables: List[str]` field and `tables=tables` constructor arg from `IcebergParser`. The field was never read—only `topics_to_tables` is consumed by `extract_lineages`.

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
davidwalker-tp: ✅ **Fixed in commit:** `fix(kafka-connect): mask sensitive Iceberg sink connector properties`

**Blocker resolved.** Replaced verbatim config dict with substring deny-list masking:
```python
sensitive_markers = ('secret', 'token', 'credential', 'password', '.key')
```

This safely redacts all Iceberg catalog secrets (OAuth tokens, REST credentials, S3 keys) without needing fixed property names. Matches the pattern used by other sink connectors (S3, Snowflake, BigQuery, JDBC).

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
davidwalker-tp: ✅ **Fixed in commit:** `fix(kafka-connect): warn when Iceberg lineage cannot be determined from static config`

**Three fixes:**
1. Simplified dict comprehension—removed redundant ternary (`tables if tables else []` → `tables`)
2. Fixed list aliasing—now uses dict comprehension: `{topic: tables for topic in topic_list}`
3. **Added warning** when `iceberg.tables` is empty but topics exist, unless `iceberg.tables.dynamic-enabled=true` (valid dynamic-routing case). Helps operators distinguish misconfiguration from intentional dynamic setup.

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/config_constants.py
davidwalker-tp: ✅ **Fixed in commit:** `chore(kafka-connect): remove unused ICEBERG_CATALOG constant`

Removed `ICEBERG_CATALOG` (never referenced). `ICEBERG_TABLES_DYNAMIC_ENABLED` is now **wired up** in commit 4 to detect dynamic-routing mode and suppress spurious warnings.

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/sink_connectors.py
davidwalker-tp: ✅ **Fixed in commit:** `refactor(kafka-connect): extract shared topic resolution helpers to BaseConnector`

Added two protected methods to `BaseConnector`:
- `_get_topics_from_sink_config()`—topics/topics.regex priority logic (was duplicated 6×)
- `_resolve_subscribed_topics()`—topic/runtime intersection + fallback (was duplicated 4×)

Updated `IcebergSinkConnector` to use these helpers. Other 5 sinks kept their copies per your note that full base-class extraction is a fine follow-up PR. This eliminates duplication for Iceberg immediately.

metadata-ingestion/tests/unit/kafka_connect/test_kafka_connect.py
davidwalker-tp: ✅ **Fixed in commit:** `test(kafka-connect): replace trivial platform test with stale-topic exclusion test`

**Removed:** `test_platform_is_iceberg` (one-line method, only fails on typo, already implicitly covered by other tests)

**Added:** `test_iceberg_sink_stale_topic_excluded_when_runtime_topics_available` — validates the connector's actual untested logic: intersecting configured topics with runtime Kafka list to exclude stale subscriptions. Mirrors the pattern in `TestSnowflakeSinkConnector` and `TestClickHouseSinkConnector`.

davidwalker-tp:
> @davidwalker-tp thanks for the fix. I left a few small comments; if you can fix those, the PR should be fine.

Thank you for the review, I've addressed the comments, with Claude's help. If you could take another look?

treff7es (approved):
Thanks for the fixes, now it looks good.

PR #18004, opened 2026-06-23
feat(ingest): reconcile upstream lineage URN casing against DataHub

puneetagarwal-datahub:
## Summary

DataHub compares URNs as exact, case-sensitive strings. When two sources name the same table with different casing — a warehouse stores `...,db.schema.table,PROD)` (lowercase) while a BI tool references `...,DB.SCHEMA.TABLE,PROD)` (uppercase) — DataHub treats them as two entities, so the lineage edge is never drawn and multi-hop lineage silently breaks.

The only existing mitigation, the per-source `convert_urns_to_lowercase` flag, keeps lineage connected by lowercasing *every* URN — which flattens the warehouse's real casing, can merge genuinely distinct tables, and isn't available on some BI connectors.

This PR adds an **opt-in ingestion work-unit processor** that instead resolves each upstream reference to the casing of the entity that **already exists** in DataHub — per ingestion, no global coordination, preserving the warehouse's original casing.

## What it does

- New `AutoResolveLineageUrnsProcessor`. For each configured upstream platform it bulk-loads that platform's existing entities once, then reconciles every lineage reference locally (no per-reference backend calls).
- Fixes **table-level** references (`UpstreamLineage`, `DashboardInfo`, `ChartInfo`, and `DataJobInputOutput` inputs) and **column-level** references (`FineGrainedLineage` field paths, corrected against the stored schema).
- Only references *to* warehouse assets are rewritten; the entity's own identity and downstream fields are never touched.
- Multiple upstream platforms per source are supported — each reference routes to its own platform's catalog.
- Stamps a `matchType` on each reference so the outcome is visible.

## Match types

Recorded on each reference on a configured platform:

- **`EXACT`** — already matched an existing entity; unchanged.
- **`NORMALIZED`** — rewritten to heal a casing mismatch.
- **`UNRESOLVED`** — no existing entity matched; left unchanged but flagged so broken lineage is visible.

Absence of `matchType` means the reference is out of scope (platform not configured, or data ingested before the feature).

_Column-level note:_ `FineGrainedLineage.upstreams` is a bare URN array, so its `matchType` is one aggregate per record, not per column — a mixed record reports `NORMALIZED` over `UNRESOLVED`. Per-column verdicts would require a model change; deferred.

## Coverage and limitations

- **Heals toward a lowercase-stored warehouse** (the Snowflake/BigQuery default) in any incoming casing. A warehouse that stores a **non-lowercase** identity (UPPER / Pascal / Mixed) isn't reconciled yet, and ambiguous case-collisions aren't detected — both deferred to a backend follow-up (a normalized-URN index).
- **UPSERT lineage only.** A lineage aspect emitted as a PATCH is emitted unchanged, counted (`num_patch_lineage_skipped`) with an end-of-run warning. BI targets emit full aspects and are unaffected; `dataJobInputOutput` (dbt / Airflow / Spark) can emit patches.
- **Requires a backend connection** (no-op offline) and reconciles only against entities that already exist at ingestion time.
- **Does not retroactively heal existing edges** — re-ingest the source after enabling.
- **Not yet covered:** `dataJobInputOutput` outputs (deliberate — the job's own products), column-level field lists on chart / dataJob, and `dataProcessInstance` run lineage.

## Observability

Run-report counters (normalized / unchanged / unresolved / patch-skipped, plus per-work-unit) and aggregated end-of-run warnings for: unresolved references (with a sample), a configured platform that matched nothing (typo catcher), a catalog that failed to load, and patch-based lineage skipped.

## Example

A BI source emits an uppercase upstream; the processor rewrites it to the stored lowercase casing and stamps the verdict:

```json
{
  "upstreams": [
    {
      "dataset": "urn:li:dataset:(urn:li:dataPlatform:snowflake,db.schema.table,PROD)",
      "type": "TRANSFORMED",
      "matchType": "NORMALIZED"
    }
  ],
  "fineGrainedLineages": [
    {
      "upstreamType": "FIELD_SET",
      "upstreams": ["urn:li:schemaField:(urn:li:dataset:(urn:li:dataPlatform:snowflake,db.schema.table,PROD),amount)"],
      "downstreamType": "FIELD",
      "downstreams": ["urn:li:schemaField:(urn:li:dataset:(urn:li:dataPlatform:looker,report,PROD),amount)"],
      "matchType": "NORMALIZED"
    }
  ]
}
```

The BI tool referenced `DB.SCHEMA.TABLE` / `AMOUNT`; the table and column are corrected to the stored casing, and the downstream field is left untouched.

## Configuration

Opt-in, off by default. Enable on a BI-tool ingestion and point it at the upstream warehouse platform(s):

```yaml
flags:
  auto_resolve_lineage_urns:
    enabled: true
    upstream_platforms:
      - platform: snowflake
        platform_instance: my_instance # optional
        env: PROD # optional, defaults to PROD
      - platform: redshift # add one entry per upstream platform referenced
        env: PROD
```

Enable it on BI-tool / cross-platform ingestions, **not** on the warehouse ingestion itself (the warehouse owns its casing). Requires the SQL-parser extra (`acryl-datahub[sql-parser]`), which the intended connectors already bundle; misconfiguration fails fast at config-parse time.

## Related connector fixes (split out)

Two connector-level casing bugs found in the same audit (Power BI, Mode) ship separately in #18181; this PR is framework-only.

## Testing

52 unit tests: healing across casings, the full `EXACT` / `NORMALIZED` / `UNRESOLVED` / out-of-scope matrix, column correction, multiple upstream platforms, MCE / MCP-wrapper / raw-MCP shapes, patch-based skip, observability counters and warnings, and config validation. Validated end-to-end against a running instance (table + column healing, `matchType` persistence).

## Notes

- Additive and backward-compatible: the new aspect field is optional and the behavior is behind an opt-in flag (default off). Embedding aspects' schema versions are bumped and RestLI snapshots regenerated accordingly.

## Checklist

- [x] Docs added/updated — usage guide at `metadata-ingestion/docs/dev_guides/lineage_urn_casing.md`
- [x] No breaking change (additive optional field + opt-in flag, default off)


metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_normalize_lineage_urns.py
```diff
@@ -0,0 +1,482 @@
...
+
+    `index` maps lowercase(urn) -> the real URNs sharing that form, built from the
+    schema-bearing entities the `SchemaResolver` already knows (``get_urns()``). A URN
+    exists exactly iff it appears in its own lowercase bucket. The `resolvers` hold the
+    schemas, used to correct column-name casing. Note: entities that exist in DataHub
+    without a schema are not represented here — reconciling those needs a richer
+    SchemaResolver and is a tracked follow-up.
+    """
+
+    index: Dict[str, List[str]]
```
sgomezvillamor: we agreed to keep "entities that exist in DataHub without a schema" as a known limitation, to be eventually resolved in SchemaResolver
so this `index` here should be removed

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_normalize_lineage_urns.py
```diff
@@ -0,0 +1,482 @@
...
+        return getattr(ctx.pipeline_context, "graph", None) is not None
+
+    def process(self, stream: Iterable[MetadataWorkUnit]) -> Iterable[MetadataWorkUnit]:
+        # We mutate the typed aspect (via get_aspect_of_type) rather than the
+        # transform_urns() helper that auto_lowercase_urns uses: that helper rewrites
+        # *every* URN uniformly (URN -> URN), but we must be selective (only upstream
+        # references, never the entity or downstream fields) AND set a non-URN field
+        # (matchType), neither of which transform_urns can express. For MCE/MCPW
+        # get_aspect_of_type returns the live aspect, so this is already an in-place
+        # edit; _write_back only does extra work for raw MCPs (see below).
```
sgomezvillamor: such a comment

```
_write_back only does extra work for raw MCPs (see below)
```

can be superseded or emphasized if we eg call the method: `_write_back_if_mcp`

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_normalize_lineage_urns.py
```diff
@@ -0,0 +1,482 @@
...
+        super().__init__(ctx)
+        graph = ctx.pipeline_context.graph
+        # Guaranteed non-None by should_enable(); assert for the type checker.
+        assert graph is not None
+        self._graph: DataHubGraph = graph
+        self._config: List[UpstreamPlatformCasing] = (
+            ctx.pipeline_context.flags.normalize_lineage_urn_casing.upstream_platforms
+        )
+        # Lazily bulk-initialized catalogs, keyed by normalized platform name.
+        self._catalog_by_platform: Dict[str, _Catalog] = {}
```
sgomezvillamor: given we have `platform`, `platform_instance` and `env` in `UpstreamPlatformCasing`
how is we only index by `platform`?

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_normalize_lineage_urns.py
```diff
@@ -0,0 +1,482 @@
...
+        """
+        # Deferred import: schema_resolver_provider pulls in sqlglot, which must not
+        # be a module-load-time dependency (see the note at the top of this file).
+        from datahub.sql_parsing.schema_resolver_provider import provide_schema_resolver
+
+        if platform not in self._catalog_by_platform:
+            # Normalize the configured platform: it may be a bare name ("snowflake") or
+            # a full URN ("urn:li:dataPlatform:snowflake"), both of which the resolver
+            # accepts. `platform` here is the normalized name parsed from the dataset
+            # URN, so compare like-for-like.
```
sgomezvillamor: any "normalization" on the input config should be done as a pydantic validator in the config model itself

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_normalize_lineage_urns.py
```diff
@@ -0,0 +1,482 @@
...
+        # some connector tests).
+        return getattr(ctx.pipeline_context, "graph", None) is not None
+
+    def process(self, stream: Iterable[MetadataWorkUnit]) -> Iterable[MetadataWorkUnit]:
+        # We mutate the typed aspect (via get_aspect_of_type) rather than the
+        # transform_urns() helper that auto_lowercase_urns uses: that helper rewrites
+        # *every* URN uniformly (URN -> URN), but we must be selective (only upstream
+        # references, never the entity or downstream fields) AND set a non-URN field
+        # (matchType), neither of which transform_urns can express. For MCE/MCPW
+        # get_aspect_of_type returns the live aspect, so this is already an in-place
```
sgomezvillamor: for MCE/MCPW we return live aspect
however MCP is the most common case and this requires deserialization

that's quite unfortunate, I hope it does not become a performance issue

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
```diff
@@ -0,0 +1,482 @@
...
+    def process(self, stream: Iterable[MetadataWorkUnit]) -> Iterable[MetadataWorkUnit]:
+        # We mutate the typed aspect (via get_aspect_of_type) rather than the
+        # transform_urns() helper that auto_lowercase_urns uses: that helper rewrites
+        # *every* URN uniformly (URN -> URN), but we must be selective (only upstream
+        # references, never the entity or downstream fields) AND set a non-URN field
+        # (matchType), neither of which transform_urns can express. For MCE/MCPW
+        # get_aspect_of_type returns the live aspect, so this is already an in-place
+        # edit; _write_back only does extra work for raw MCPs (see below).
+        for wu in stream:
+            try:
```
sgomezvillamor, line 265: given the potential performance impact, we need to know the rationale of why including the 4 aspects

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_normalize_lineage_urns.py
sgomezvillamor: 🤔 

maybe we can:
- traverse aspect a-la-transform_urns and get "dataset" urns
- resolve these with the schemaresolver
- and only if we need to fix urn, then we may pay the cost of deser-serialization

WDYT? so we only pay that cost if there is something to be fixed

sgomezvillamor:
what about renaming all this as: auto resolve lineage urns
would that be a better name?

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_normalize_lineage_urns.py
puneetagarwal-datahub: without the index the processor can only "heal" when the BI casing already equals the stored casing.

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
```diff
@@ -0,0 +1,592 @@
...
+        # don't target datasets or are the entity's own outputs (see the dev guide).
+        self._normalizers: List[Tuple[Type[_Aspect], Callable[..., bool]]] = [
+            (UpstreamLineageClass, self._normalize_upstream_lineage),
+            (DashboardInfoClass, self._normalize_dashboard_info),
+            (DataJobInputOutputClass, self._normalize_datajob_io),
+            (ChartInfoClass, self._normalize_chart_info),
+        ]
+
+    @classmethod
+    def should_enable(cls, ctx: WorkunitProcessorContext) -> bool:
```
treff7es, line 245: 🔴 **Blocker (CI red):** `should_enable()` isn't robust to a `Mock()` pipeline context. This processor is added to the shared chain for *every* source, and several integration tests construct sources with `ctx=Mock()` (e.g. `tests/integration/salesforce/test_salesforce.py:112,157`). Under a bare `Mock`, `cfg.enabled` and `cfg.upstream_platforms` are truthy Mocks and `getattr(ctx.pipeline_context, "graph", None)` is a non-`None` Mock — so this returns `True`, the processor initializes with a Mock `self._config`, and the run later crashes (see `_warn_unmatched_platforms`).

`ci (3.11, testIntegrationBatch0)` and `integration-tests-gate` are failing on exactly this. Real runs are unaffected (default `enabled=False`), but the shared processor must fail closed under a degenerate/mock config. Suggest requiring a real, enabled config, e.g. `if cfg.enabled is not True: return False` and verifying `upstream_platforms` is a non-empty `list` before enabling.

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
```diff
@@ -0,0 +1,592 @@
...
+            self.report.num_workunits_modified += 1
+
+    def _warn_unmatched_platforms(self) -> None:
+        """Warn about configured platforms that no reference used this run.
+
+        The usual cause is a case/spelling mismatch in the config (platform names are
+        compared case-sensitively), which would otherwise silently heal nothing.
+        """
+        unmatched = {
+            entry.platform for entry in self._config
```
treff7es, line 333: This is the concrete crash site behind the CI failure: `{entry.platform for entry in self._config}` raises `TypeError: 'Mock' object is not iterable` when `self._config` is a Mock (from a `Mock()` ctx that slipped past `should_enable`). It also runs *outside* the per-workunit try/except in `process()`, so it aborts the whole stream. Fixing `should_enable` (above) is the root cause; you could additionally guard this against a non-iterable config for defense in depth.

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
```diff
@@ -0,0 +1,592 @@
...
+        # PipelineContext instance attribute (absent from MagicMock(spec=...) used by
+        # some connector tests).
+        return getattr(ctx.pipeline_context, "graph", None) is not None
+
+    def process(self, stream: Iterable[MetadataWorkUnit]) -> Iterable[MetadataWorkUnit]:
+        for wu in stream:
+            try:
+                self._resolve_workunit(wu)
+            except Exception as e:
+                self.report.num_exceptions += 1
```
treff7es, line 268: 🟡 Per-workunit failures are recorded only as `self.report.num_exceptions` (a counter on `WorkunitProcessorReport`, which surfaces in the sub-report JSON) plus `logger.warning`. They never enter the pipeline `warnings`/`failures` summary operators actually read — so a run that silently fails to reconcile 30% of a source's lineage still reports success. The processor has `self.ctx.source_report` in hand; the sibling `EnsureAspectSizeProcessor` routes recoverable degradations through `self.ctx.source_report.warning(title=..., message=..., context=..., exc=e)`. Recommend doing the same here (keep the counter + logger).

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
```diff
@@ -0,0 +1,592 @@
...
+            # a configured platform; out-of-scope refs get res.match_type=None and are
+            # left untouched.
+            if res.match_type is not None:
+                upstream.matchType = res.match_type
+                changed = True
+            if res.match_type == LineageMatchTypeClass.NORMALIZED:
+                upstream.dataset = res.urn
+                self.report.num_dataset_urns_normalized += 1
+            elif res.match_type == LineageMatchTypeClass.UNRESOLVED:
+                self.report.num_refs_unresolved += 1
```
treff7es: 🟡 `UNRESOLVED` is the feature's primary "this lineage is likely broken" signal, and the comments at ~L409-412 / L565-566 state the intent is that it be *visible*. But it only increments `num_refs_unresolved` on the non-surfacing sub-report — so it isn't visible in the run summary. Recommend an **aggregated** end-of-run `source_report.warning` (with a sampled `LossyList` of unresolved URNs) when `num_refs_unresolved > 0`; a per-reference warning would be too noisy, but an aggregate with samples hits the right altitude.

metadata-ingestion/src/datahub/ingestion/run/pipeline_config.py
```diff
@@ -47,6 +47,56 @@ class FailureLoggingConfig(ConfigModel):
...
+    platform: str = Field(
+        description="Upstream data platform whose assets are referenced by this "
+        "source's lineage (e.g. `snowflake`). References to this platform's assets "
+        "are reconciled against the casing stored in DataHub.",
+    )
+    platform_instance: Optional[str] = Field(
+        default=None,
+        description="Platform instance of the upstream platform, if any.",
+    )
+    env: str = Field(
```
treff7es: 🟡 `env: str` (and `platform_instance`) are hand-rolled here, duplicating the canonical `PlatformDetail` (`configuration/source_common.py`). More importantly, this `env` has **no validator**, unlike `EnvConfigMixin.env_must_be_one_of` (used across the ecosystem), so `env: "prod"` (lowercase) or a typo passes config parsing and then silently under-resolves inside `provide_schema_resolver` — the same silent-failure mode the `platform` validator was added to prevent. Suggest `class UpstreamPlatformCasing(PlatformInstanceConfigMixin, EnvConfigMixin)` and dropping the two hand-written fields (mirrors `sigma`/`qlik_sense` `PlatformDetail`), which restores the `env` validator for free.

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
```diff
@@ -0,0 +1,592 @@
...
+
+
+@dataclass
+class _Resolution:
+    """Outcome of resolving one dataset URN against the configured platform(s)."""
+
+    urn: str  # The (possibly rewritten) URN to emit.
+    schema: Optional[SchemaInfo]  # Cached schema of the resolved entity, if known.
+    # EXACT / NORMALIZED / UNRESOLVED / None (no reconciliation performed).
+    match_type: Optional[str]
```
treff7es: ℹ️ `match_type` carries a closed set (`EXACT`/`NORMALIZED`/`UNRESOLVED`) or `None`, but is typed `Optional[str]`, so mypy can't catch a typo or exhaustiveness-check the `if/elif` verdict chains that drive correctness. A `Literal["EXACT", "NORMALIZED", "UNRESOLVED"]` alias (reused at the `match_types` list ~L455 and the `_resolve_field_urn` return ~L479) documents the domain far better at near-zero cost. A full `enum` isn't worth it since codegen renders these as plain `str`. Optional: `@dataclass(frozen=True)` + a `__post_init__` tying `schema`/`match_type`/`urn` together would make the invariant in the docstring fail fast.

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
```diff
@@ -0,0 +1,592 @@
...
+        # configured platforms that matched nothing (a common cause is a case/spelling
+        # typo in the config, since platform names are compared case-sensitively).
+        self._seen_reference_platforms: Set[str] = set()
+        # (aspect class -> in-place normalizer, returns True iff it mutated the aspect).
+        # These are the aspects a BI / orchestration source emits that carry *upstream
+        # dataset* references — the only refs affected by cross-source casing mismatch:
+        # upstreamLineage (table + fineGrained columns), dashboardInfo / chartInfo inputs,
+        # and dataJobInputOutput inputs (dbt / Airflow / Spark). Other lineage aspects
+        # don't target datasets or are the entity's own outputs (see the dev guide).
+        self._normalizers: List[Tuple[Type[_Aspect], Callable[..., bool]]] = [
```
treff7es, line 231: ℹ️ Minor: `Callable[..., bool]` erases the argument type — each normalizer already has a precise signature. `Callable[[_Aspect], bool]` restores arity/return checking with no runtime cost. Separately, the NORMALIZED/UNRESOLVED/else counter dispatch repeats across `_normalize_upstream_lineage` / `_heal_dataset_urns` / `_heal_dataset_edges`; a small `_tally_table_ref(res) -> bool` helper would keep the three counter paths in sync (leave the schema-aware `_resolve_field_urn` out of it). Both non-blocking.

metadata-ingestion/src/datahub/ingestion/run/pipeline_config.py
```diff
@@ -47,6 +47,56 @@ class FailureLoggingConfig(ConfigModel):
...
+        description="Environment (FabricType) of the upstream platform's assets.",
+    )
+
+    @field_validator("platform")
+    @classmethod
+    def _normalize_platform(cls, v: str) -> str:
+        # Accept either a bare platform name ("snowflake") or a full data-platform URN
+        # ("urn:li:dataPlatform:snowflake") and store the bare name, so downstream
+        # matching against the platform parsed from dataset URNs is like-for-like.
+        from datahub.metadata.urns import DataPlatformUrn
```
treff7es: late imports not prefered if it is not justified

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
```diff
@@ -0,0 +1,592 @@
...
+
+def _field_path(field_urn: str) -> Optional[str]:
+    """Return the field path (column) of a schemaField URN, or None if not parseable."""
+    try:
+        return Urn.from_string(field_urn).entity_ids[1]
+    except Exception:
+        return None
+
+
+def _is_dataset_urn(urn: Optional[str]) -> TypeGuard[str]:
```
treff7es, line 139: ℹ️ Prefer the typed URN primitive over `guess_entity_type` here. `guess_entity_type` is a naive splitter (`urn.split(":")[2]`) that only checks the prefix and one segment, not full structure. `DatasetUrn` is already imported in this module and re-parsed in `_resolve_dataset` anyway, and `DatasetUrn.from_string(urn)` raises `InvalidUrnError` on a malformed/non-dataset URN — giving full structural validation and letting you narrow the bare `except Exception` to `except InvalidUrnError`:

```python
def _is_dataset_urn(urn: Optional[str]) -> TypeGuard[str]:
    if not urn:
        return False
    try:
        DatasetUrn.from_string(urn)
        return True
    except InvalidUrnError:
        return False
```

The same tightening applies to `_parent_dataset_urn` / `_field_path` (bare `except Exception` around `Urn.from_string(...).entity_ids[i]`). This is a refinement, not a correctness bug — behavior is equivalent for the inputs the processor sees.

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
```diff
@@ -0,0 +1,592 @@
...
+class _Resolution:
+    """Outcome of resolving one dataset URN against the configured platform(s)."""
+
+    urn: str  # The (possibly rewritten) URN to emit.
+    schema: Optional[SchemaInfo]  # Cached schema of the resolved entity, if known.
+    # EXACT / NORMALIZED / UNRESOLVED / None (no reconciliation performed).
+    match_type: Optional[str]
+
+
+def _parent_dataset_urn(field_urn: str) -> Optional[str]:
```
treff7es, line 112: ℹ️ Same idea as the `_is_dataset_urn` note — these two helpers use positional `entity_ids[0]`/`entity_ids[1]`, but the repo already has the typed `SchemaFieldUrn` (exported from `datahub.metadata.urns`, same module you import `DatasetUrn` from) with named `.parent` / `.field_path` accessors and full validation:

```python
from datahub.metadata.urns import SchemaFieldUrn
from datahub.utilities.urns.error import InvalidUrnError

def _parent_dataset_urn(field_urn: str) -> Optional[str]:
    try:
        return SchemaFieldUrn.from_string(field_urn).parent
    except InvalidUrnError:
        return None

def _field_path(field_urn: str) -> Optional[str]:
    try:
        return SchemaFieldUrn.from_string(field_urn).field_path
    except InvalidUrnError:
        return None
```

Beyond readability, this closes a latent bug: with `Urn.from_string(field_urn).entity_ids[1]`, if `field_urn` is accidentally a *dataset* URN (not a schemaField URN), `entity_ids[1]` silently returns the dataset **name** (e.g. `DB.SCHEMA.TABLE`) instead of `None` — so a bad ref would be treated as a valid field path. `SchemaFieldUrn.from_string` raises `InvalidUrnError` on any non-schemaField URN, so the guard returns `None` correctly, and the `except` narrows from bare `Exception` to `InvalidUrnError`.

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
```diff
@@ -0,0 +1,597 @@
...
+        granularity. A reference for a platform instance / env that isn't configured
+        simply won't appear in the index and resolves to UNRESOLVED.
+        """
+        # Record every reference platform (even before the load cache) so the
+        # end-of-run check can flag configured platforms that matched nothing.
+        self._seen_reference_platforms.add(platform)
+
+        # Deferred import: schema_resolver_provider pulls in sqlglot, which must not
+        # be a module-load-time dependency (see the note at the top of this file).
+        from datahub.sql_parsing.schema_resolver_provider import provide_schema_resolver
```
treff7es: 🟡 Two related concerns about how the sqlglot dependency is handled here.

**1. Don't bury the sqlglot dependency in a leaf method.** The deferred import is correct in *intent* — this module is imported eagerly by `workunit_processors/__init__` on every source's `get_workunit_processors()` path, so module load must stay sqlglot-free (guarded by `test_module_import_does_not_pull_sqlglot`). But expressing that dependency as inline imports scattered across leaf methods (`provide_schema_resolver` here, `match_columns_to_schema` in `_resolve_field_urn`) hides a real, if conditional, dependency. Prefer hoisting both to a single visible chokepoint — `__init__`, which only runs after `should_enable()` has confirmed the feature is on + a graph exists, so it stays off the module-load path and still passes the guard test — and cache the callables on `self`. One honest "this processor needs sqlglot once constructed" beats two buried imports. (The fully-declared end state — a plain top-level import — requires making `schema_resolver`/`schema_resolver_provider` import sqlglot-free at their source, i.e. deferring `import sqlglot` inside `sql_parsing/_models.py`; that's a separate, higher-blast-radius follow-up.)

**2. Nothing guarantees a source that enables this flag actually ships sqlglot.** `sqlglot` is not in `framework_common` — it's only in specific connector extras. This processor is registered for *every* source and gated only by the pipeline-level `flags.auto_resolve_lineage_urns` flag, which any recipe can set. I checked: all the intended BI/dashboard connectors (looker, powerbi, mode, superset/preset, tableau, redash, metabase, sigma, qlik-sense, hex, grafana, fivetran) do bundle `sqlglot_lib`, so the target use case is fine today. But a source whose extra doesn't pull sqlglot (a pure-API connector) that enables this flag will hit `ModuleNotFoundError: sqlglot` at runtime — and because `process()` swallows per-workunit exceptions (see the other comment), it would fail *silently* rather than telling the user what to install. Suggest catching `ImportError` at the point of the deferred import and surfacing a clear `source_report.failure` ("auto_resolve_lineage_urns requires sqlglot; install the `[datahub-sql-parser]`/relevant extra"), and documenting the sqlglot requirement in the dev guide.

metadata-models/src/main/pegasus/com/linkedin/dataset/Upstream.pdl
```diff
@@ -59,4 +59,15 @@ record Upstream {
...
+  /**
+   * How this upstream reference's URN was resolved against the entities stored in
+   * DataHub. Set by the lineage URN casing normalization processor: EXACT when the
+   * reference already matched an existing entity, NORMALIZED when it was rewritten to
+   * heal a casing mismatch, UNRESOLVED when it could not be resolved to a single
+   * existing entity. Absent when no reconciliation was performed (out of scope).
+   * Reflects DataHub's knowledge at ingestion time and is not re-evaluated later; see
+   * LineageMatchType.
+   */
+  matchType: optional LineageMatchType
```
treff7es, line 72: ❓Question: should we also store the *original* (pre-normalization) reference URN on the aspect, not just `matchType`?

Today, when a ref is `NORMALIZED` the processor overwrites the URN in place (`upstream.dataset = res.urn`) and only records that it was healed — the casing the source actually emitted is discarded. `matchType` tells you *that* an edge was rewritten, but not *from what to what*. For auditing/debugging ("why was this edge rewritten, and what did the connector really emit?") an original→resolved record would be useful.

Note a method enum (LOWERCASE/UPPERCASE/…) wouldn't help here — normalization is a lookup ("match the casing that already exists in DataHub"), not a fixed rule, so mixed-case rewrites don't fit those buckets. If we want more than the current verdict, capturing the original URN is the meaningful thing.

Not asking to add it necessarily — it's an extra URN on every normalized edge (aspect bloat on a hot path), and `EXACT`/`NORMALIZED`/`UNRESOLVED` may well be enough for the goal of heal-and-flag. Just want a conscious decision on whether provenance is in scope now or a deliberate follow-up.

metadata-ingestion/docs/dev_guides/lineage_urn_casing.md
```diff
@@ -0,0 +1,219 @@
...
+- **Resolves only against entities that already exist at ingestion time.** This relies on the warehouse
+  being ingested before the BI tool that references it — the normal order for scheduled pipelines. If a
+  reference's target does not yet exist in DataHub, it is left unchanged and self-heals once the
+  warehouse is ingested and the BI source re-runs.
+- **Does not retroactively heal existing broken edges** already stored in the graph. Re-ingest the
+  affected source after enabling the flag to fix them.
+- **Collision-safe but conservative.** On case-sensitive platforms where two genuinely different tables
+  differ only by case, ambiguous references are left unchanged rather than risk merging distinct
+  entities.
+- **Reconciles against schema-bearing entities only.** Membership/casing comes from what the
```
treff7es: 💬 Discussion: worth documenting a related scope boundary here for **sqlglot-derived** lineage (BI sources that parse SQL via `SqlParsingAggregator`, e.g. Looker derived tables, Mode/Superset query SQL), separate from the direct-emit path this guide focuses on.

This processor heals references that were *emitted*. But column-level (fine-grained) lineage from sqlglot is only produced when the schema resolves, and `SchemaResolver.resolve_table` tries just three casing variants (as-parsed/dialect-normalized, fully-lowercased, and lower-table+non-lower-instance). If the warehouse's stored casing falls outside those — e.g. an uppercase Snowflake table referenced in a casing dialect-normalization doesn't realign, or arbitrary PascalCase — the schema lookup misses, `resolve_table` returns `(synthesized_urn, None)`, and **no CLL is generated at parse time**. Since the fine-grained edge is never emitted, this ingest-time healer can't recover it.

Table-level lineage is unaffected (`resolve_table` always returns a best-effort non-None URN, so the edge is emitted in the parsed casing and then healed here). The gap is specifically parse-time column-level drops.

Is the intent that this case is covered only by the deferred casing-aware `SchemaResolver` follow-up (the same work the "schema-bearing entities only" bullet references)? If so, a one-line note here — "CLL from sqlglot-derived sources can be dropped upstream on a casing miss and is not recoverable by this processor; see the SchemaResolver follow-up" — would set expectations. Not asking to fix it in this PR; just to make the boundary explicit.

treff7es (commented):
## Review summary

(Detailed inline findings are in my earlier review; this adds the overview/verdict that didn't carry over, plus one further discussion point on the dev guide.)

High-quality, well-tested, and well-documented change — correct framework-layer placement, good reuse of existing SQL-parsing infrastructure, opt-in and fail-closed by default, and a genuinely behavioral 41-test suite. One CI-blocking bug needs fixing, plus a few should-fix items.

### 🔴 Blocker (CI is red)
- `should_enable()` isn't robust to a `Mock()` pipeline context. Several integration tests build sources with `ctx=Mock()` (e.g. `tests/integration/salesforce/test_salesforce.py`), where `cfg.enabled`/`cfg.upstream_platforms` are truthy Mocks and `graph` is non-`None`, so the processor enables and then `_warn_unmatched_platforms()` crashes iterating a Mock `self._config`. `ci (3.11, testIntegrationBatch0)` and `integration-tests-gate` are currently failing on this.

### 🟡 Should fix
- Recoverable failures (`num_exceptions`, `num_refs_unresolved`) only land in the processor sub-report + logger, not the pipeline `warnings` summary — undercutting the stated "broken lineage is visible" goal. `EnsureAspectSizeProcessor` shows the `self.ctx.source_report.warning(...)` pattern.
- The sqlglot dependency is expressed via imports buried in leaf methods; hoist to `__init__` (still off the module-load path). Separately, `sqlglot` isn't in `framework_common` and this processor runs for every source behind a pipeline flag — all intended BI connectors do bundle sqlglot, but a non-sqlglot source that enables the flag would `ImportError` at runtime (silently, given the swallowed exceptions). Consider a clear `source_report.failure` on `ImportError`.
- `UpstreamPlatformCasing.env`/`platform_instance` are hand-rolled instead of inheriting `EnvConfigMixin`/`PlatformInstanceConfigMixin`; `env` has no validator, so a lowercase/typo value silently under-resolves.

### ℹ️ Suggestions
- Tighten `match_type: Optional[str]` to a `Literal`; `Callable[..., bool]` to `Callable[[_Aspect], bool]`; a small shared counter-tally helper for the repeated dispatch.
- Use the typed URN primitives the repo already provides: `DatasetUrn.from_string` in `_is_dataset_urn`, and `SchemaFieldUrn.parent`/`.field_path` in `_parent_dataset_urn`/`_field_path` (the latter also closes a latent bug where a stray dataset URN yields a bogus field path instead of `None`).

### ❓ Open questions
- Whether to also store the original (pre-normalization) URN for provenance (a method enum like lowercase/uppercase would be low-value since normalization is lookup-based, not rule-based).
- The sqlglot-derived CLL scope boundary (see inline comment on the dev guide).

### Notes (not defects)
- ✅ The `schemaVersion` bumps are correct and required — `bump_schema_versions.py` (pre-commit + CI) auto-bumps on any structural PDL change and cascades into embedding aspects. No migration mutator is needed for an additive optional field.
- This PR is stacked on the unmerged #18017; the Teradata / `odbc.py` / `powerbi/config.py` / ODBC-two-tier changes belong there and will drop out of the diff once it merges.
- One reviewer thread (removing the local casing index) looks open, and the two "deferred per review" items in the description appear author-declared rather than reviewer-confirmed — worth explicit sign-off.

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
puneetagarwal-datahub: done

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
puneetagarwal-datahub: done

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
puneetagarwal-datahub: done

metadata-ingestion/src/datahub/ingestion/run/pipeline_config.py
puneetagarwal-datahub: done

metadata-ingestion/src/datahub/ingestion/run/pipeline_config.py
puneetagarwal-datahub: done

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
puneetagarwal-datahub: done

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
puneetagarwal-datahub: done

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
puneetagarwal-datahub: done

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
puneetagarwal-datahub: done

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
puneetagarwal-datahub: done

metadata-models/src/main/pegasus/com/linkedin/dataset/Upstream.pdl
puneetagarwal-datahub: @treff7es I am not sure we have that usecase today . We can target if required to show in UI to user .

@sgomezvillamor  WDYT . should we add it on 1st cut  ?

metadata-ingestion/docs/dev_guides/lineage_urn_casing.md
puneetagarwal-datahub: addded

metadata-models/src/main/pegasus/com/linkedin/dataset/Upstream.pdl
puneetagarwal-datahub: Also not only this case , I feel then we should then have original form in case of lowercase processor and other ones if we have to preserve original casing/content ?

sgomezvillamor: Agreed that storing the original would help for auditing/debugging, but an extra URN does add overhead 😢 

My suggestion for covering this would be to store it in the `properties` dict in `UpstreamLineage`. Either way, we can just leave it as a code comment for now — from my side, not a hard requirement for this PR.

metadata-ingestion/docs/dev_guides/lineage_urn_casing.md
```diff
@@ -0,0 +1,235 @@
...
+- For every upstream reference, it looks for the existing entity that matches **ignoring case**:
+  - If an entity with the **exact** URN already exists, the reference is left unchanged (recorded as an
+    exact match). This ensures genuinely distinct entities on case-sensitive platforms are never merged.
+  - Otherwise, if exactly **one** existing entity matches case-insensitively, the reference is rewritten
+    to that entity's URN (recorded as a normalized match). This heals the mismatch in **both**
+    directions — uppercase-reported→lowercase-stored and lowercase-reported→uppercase-stored.
+  - If **no** entity matches, or **more than one** matches (an ambiguous collision on a case-sensitive
+    platform, e.g. both `orders` and `Orders` exist), the reference is left unchanged.
+
+Only references **to** warehouse assets are modified. The entity the aspect is attached to, and
```
sgomezvillamor:
```suggestion
Only references **to** upstream assets are modified. The entity the aspect is attached to, and
```

metadata-ingestion/docs/dev_guides/lineage_urn_casing.md
sgomezvillamor, line 1: Overall the docs look great. One suggestion since this is user-facing: it feels a bit too implementation-heavy — worth simplifying and avoiding too much detail on the internals.

Also, I'd frame the resolver as the current solution and `convert_urns_to_lowercase` as the legacy/deprecated alternative, with a note on why the resolver is preferred: it doesn't require altering the identity of the assets.

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
```diff
@@ -0,0 +1,694 @@
...
+        except ImportError as e:
+            self._enabled = False
+            self.ctx.source_report.failure(
+                title="auto_resolve_lineage_urns could not run",
+                message="Reconciling lineage URN casing requires the sqlglot-based SQL "
+                "parser, which this source's install does not include. Install the SQL "
+                "parsing extra (`pip install acryl-datahub[sql-parser]`) or a connector "
+                "extra that bundles it; lineage is emitted unchanged until then.",
+                exc=e,
+            )
```
sgomezvillamor: Do I understand correctly that we are failing regardless of `AutoResolveLineageUrnsConfig#enabled` being `True` or `False`? We should only fail if resolving lineage is enabled.

I would suggest:

- update description of `AutoResolveLineageUrnsConfig#enabled` flag so it clearly mentions the requirement on the extra dep
- handle this as a pydantic model validation directly in `AutoResolveLineageUrnsConfig`

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
```diff
@@ -0,0 +1,694 @@
...
+            return False
+        if not isinstance(cfg.upstream_platforms, list) or not cfg.upstream_platforms:
+            if isinstance(cfg.upstream_platforms, list):
+                # Genuinely enabled but unconfigured (empty list): every reference would
+                # no-op — skip the bulk catalog load entirely and tell the operator why.
+                logger.warning(
+                    "auto_resolve_lineage_urns is enabled but no upstream_platforms "
+                    "are configured; the processor will not run. Configure the warehouse "
+                    "platform(s) this source references to enable casing reconciliation."
+                )
```
sgomezvillamor: this looks to me that could be solved as a pydantic model validation also, 
when the resolver is enabled, lack of `upstream_platforms` could even be a failure

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
```diff
@@ -0,0 +1,694 @@
...
+                message="Reconciling lineage URN casing requires the sqlglot-based SQL "
+                "parser, which this source's install does not include. Install the SQL "
+                "parsing extra (`pip install acryl-datahub[sql-parser]`) or a connector "
+                "extra that bundles it; lineage is emitted unchanged until then.",
+                exc=e,
+            )
+        else:
+            self._provide_schema_resolver = provide_schema_resolver
+            self._match_columns_to_schema = match_columns_to_schema
+        # Per-platform state, lazily bulk-initialized on first reference. The
```
sgomezvillamor: By default, `SchemaResolver` works in lazy/on-demand mode. `SchemaResolverProvider` in `metadata-ingestion/src/datahub/sql_parsing/schema_resolver_provider.py` provides a bulk-initialized version. What's being implemented here is a third and new initializatino mode: lazily bulk-initialized mode on first reference.

A couple of concerns:

* This lazily bulk-initialized mode feels like a premature optimization — it covers the case where `upstream_platforms` in the config are never actually referenced by downstream assets. That's likely an edge case, and the complexity added may not be worth it. I'd skip this for now unless we have clear signals it's a recurring issue.
* Also, if we do ever add this, it should live in `SchemaResolverProvider` itself.

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
```diff
@@ -0,0 +1,694 @@
...
+        # SchemaResolvers are the source of truth for schema + membership; the casing
+        # index is a derived lowercase(urn) -> real URNs map used to reconcile arbitrary
+        # casings and detect ambiguous collisions. NOTE: this local index duplicates
+        # membership the SchemaResolver could own and only exists until SchemaResolver
+        # gains casing-aware resolution (tracked follow-up), at which point it is deleted.
```
sgomezvillamor: This casing-aware resolution would be solved by the `NormalizedUrn` aspect we've discussed, right? 
If so, I'd remove it here, document the gap, and we can prioritize such a fix directly in `SchemaResolver`.

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
```diff
@@ -0,0 +1,694 @@
...
+        self._resolvers_by_platform: Dict[str, List["SchemaResolver"]] = {}
+        self._casing_index_by_platform: Dict[str, Dict[str, List[str]]] = {}
+        # Platforms whose catalog load has been attempted. Doubles as the set of
+        # platforms actually referenced this run (every reference routes through
+        # _ensure_platform_loaded), so _warn_unmatched_platforms uses it to flag
+        # configured platforms that no reference matched.
+        self._loaded_platforms: Set[str] = set()
+        # A bounded sample of URNs left UNRESOLVED, for the aggregated end-of-run
+        # warning (the counter num_refs_unresolved gives the full total).
+        self._unresolved_sample: LossyList[str] = LossyList()
```
sgomezvillamor: This would fit better in `AutoResolveLineageUrnsProcessorReport`, no?

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
```diff
@@ -0,0 +1,694 @@
...
+            if aspect is None:
+                continue
+            had_lineage = True
+            if normalize(aspect):
+                self._write_back_if_mcp(wu, aspect)
+                modified = True
+        if had_lineage:
+            self.report.num_workunits_with_lineage += 1
+        if modified:
+            self.report.num_workunits_modified += 1
```
sgomezvillamor: is this the same?

```suggestion
            self.report.num_workunits_with_lineage += 1
            if normalize(aspect):
                self._write_back_if_mcp(wu, aspect)
                self.report.num_workunits_modified += 1
```

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
```diff
@@ -0,0 +1,694 @@
...
+
+    num_dataset_urns_normalized: int = 0  # Upstream dataset URNs rewritten
+    num_column_urns_normalized: int = 0  # Fine-grained field URNs rewritten
+    num_refs_unchanged: int = 0  # Left as-is (exact match, or out of scope)
+    num_refs_unresolved: int = 0  # Configured platform, no unique match (flagged)
+    num_exceptions: int = 0  # Failed to process a workunit
+    # Workunit-level observability: how many workunits carried a lineage aspect vs. how
+    # many actually had a reference rewritten. The gap is the no-op share (references
+    # already correct / out of scope).
+    num_workunits_with_lineage: int = 0  # Carried a lineage aspect
```
sgomezvillamor:
```suggestion
    num_workunits_with_lineage_aspect: int = 0
```

with proper naming, we don't need the comment to disambiguate.

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
```diff
@@ -0,0 +1,694 @@
...
+        The usual cause is a case/spelling mismatch in the config (platform names are
+        compared case-sensitively), which would otherwise silently heal nothing.
+        """
+        # Defense in depth: this runs outside the per-workunit try/except, so guard
+        # against a non-list config (should_enable already fails closed on a mock ctx).
+        if not isinstance(self._config, list):
+            return
+        unmatched = {entry.platform for entry in self._config} - self._loaded_platforms
+        if unmatched:
+            logger.warning(
```
sgomezvillamor: Make this warning visible in the UI as a structured report warning.

With this warning, we could suggest the user either fix the casing issues, or also remove the upstream platform if it's unnecessary.

That second case actually reinforces my point that this warning alone might be sufficient mitigation, removing the need for lazy bulk initialization optimization altogether.

metadata-ingestion/docs/dev_guides/lineage_urn_casing.md
puneetagarwal-datahub: Trimmed it down

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
puneetagarwal-datahub: Done. It was already enabled-gated (processor only builds when should_enable() is true), but your way is cleaner — moved it to a model_validator on AutoResolveLineageUrnsConfig that fails fast at config parse, only when enabled, and updated the enabled description to call out the sql-parser extra. Dropped the __init__ fallback since the config now guarantees the dep.

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
puneetagarwal-datahub: Agreed — dropped the lazy-bulk-on-first-reference mode.

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
puneetagarwal-datahub: the local lowercase(urn) -> real URNs map is an interim, client-side stand-in

SchemaResolver only does exact or 3-guess matching and can't detect collisions.the lowercase index adds that (efficiently), and it lives in the processor only to avoid a risky framework change now — moving it into SchemaResolver is the tracked follow-up.

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
puneetagarwal-datahub: done

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
puneetagarwal-datahub: done

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
puneetagarwal-datahub: done

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
puneetagarwal-datahub: Done

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
```diff
@@ -0,0 +1,656 @@
...
+    the entity the aspect is attached to and downstream fields are never touched. It
+    must be enabled on BI-tool / cross-platform ingestions and configured with the
+    upstream platform(s) — never on the warehouse ingestion, whose reported casing and
+    identity must be respected.
+    """
+
+    def __init__(self, ctx: WorkunitProcessorContext) -> None:
+        super().__init__(ctx)
+        graph = ctx.pipeline_context.graph
+        # Guaranteed non-None by should_enable(); assert for the type checker.
```
sgomezvillamor: `should_enable()` runs after `__init__`, so we no have such a warranty

```suggestion
        # assert for the type checker.
```

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
```diff
@@ -0,0 +1,656 @@
...
+            except Exception as e:
+                self.report.num_exceptions += 1
+                # Surface in the pipeline report (not just this processor's sub-report),
+                # so a run that fails to reconcile part of a source's lineage doesn't
+                # look clean. Keeps the processor counter above; logs via the report.
+                self.ctx.source_report.warning(
+                    title="Lineage URN casing not reconciled",
+                    message="Failed to reconcile lineage URN casing for a work unit; "
+                    "its lineage is emitted unchanged.",
+                    context=wu.id,
```
sgomezvillamor: I don't see `id` as a property/method in the `MetadataWorkunit` class
I do see `def get_urn(self) -> str`

```suggestion
                    context=wu.get_urn(),
```

sgomezvillamor (approved):
Reviewed the source changes this round (didn't go deep on the tests). Overall LGTM — framework-level placement is the right call, opt-in and fail-closed by default, and the observability story is solid. Approving.

Two minor points, non-blocking — fine to address here or as a quick follow-up:

**1. Multi-instance / multi-env resolution iterates resolvers per platform and takes the first schema match.**
I initially thought about indexing by `(platform, platform_instance)` instead of a `List[SchemaResolver]`, but `platform_instance` can't be extracted from the URN — it's part of the name, and separating it out would mean fetching the `dataPlatformInstance` aspect, which is overkill. So the iterate-the-resolvers approach makes sense to me. We *could* additionally disambiguate on `env` (that one is recoverable from the URN) to avoid, say, healing a PROD ref to a DEV entity, but that collision is unlikely in practice, so I'm ok leaving it. Please just add a short comment capturing this design reasoning if it isn't already there.

**2. `resolve_table_parts(database=None, db_schema=None, table=name)` leans on SchemaResolver internals.**
It works, but only because it relies on `get_urn_for_table` concatenating the parts back into the name — a bit fragile given that method already has a TODO around 2/3-layer hierarchy. Let's add a code comment noting the concern, and track a proper `resolve_dataset_urn(urn)`-style method on `SchemaResolver` as a separate follow-up PR (additive/read-only, distinct from the `normalizedUrn` work).

---

sgomezvillamor:
One more minor thing I ran into after approving — non-blocking, follow-up is fine.

**Patch-based upstream lineage isn't handled.** The processor pulls aspects via `wu.get_aspect_of_type(...)`, which for a raw MCP goes through `MetadataChangeProposalWrapper.try_from_mcpc()`, and that returns `None` for anything that isn't an upsert:

```python
if mcpc.changeType != ChangeTypeClass.UPSERT:
    # We can only generate MCPWs for upserts.
    return None
```

So an upstream-lineage aspect emitted as a `PATCH` (e.g. `DatasetPatchBuilder.add_upstream_lineage`, `DataJobPatchBuilder`) is skipped entirely — the URNs in the patch ops are never reconciled, and because the whole work unit is skipped they don't even surface as `UNRESOLVED`. It's a silent pass-through.

Two small asks:

- **Observability:** bump a counter when a configured lineage aspect arrives with `changeType != UPSERT`, so the skip shows up in the report instead of being invisible.
- **Docs:** call this out as a known limitation. The "Not covered" list currently only enumerates aspect types, not change types, and the description's "works across all work-unit shapes — MCE, MCP-wrapper, raw MCP" is a bit optimistic since PATCH is skipped.

Low priority — the main BI targets (Looker/Tableau/Superset/Mode) emit full aspects via MCPW so they're unaffected, but the `dataJobInputOutput` (dbt / Airflow / Spark) path can emit patches, so it's a real gap for part of the stated coverage.

---

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
puneetagarwal-datahub: get_urn() is both safer and more useful for debugging . will change

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_resolve_lineage_urns.py
puneetagarwal-datahub: sure

sgomezvillamor (approved):

PR #18008, opened 2026-06-23
fix(search): correct structured-property filtering & display by value type

rob-1019:
## Summary

Improves structured property filter reliability across five areas:

- **URN-type structured properties now show up as filter options in the search sidebar.** These properties were missing a `.keyword` subfield in their OpenSearch mapping, but `toStructuredPropertyFacetName()` appends `.keyword` for aggregation queries. Without it, the aggregation returned zero buckets and the filter widget never appeared. Added `getMappingsForStructuredPropertyUrn()` to include the subfield.

- **Free-form text filters now return results for partial values.** STRING/RICH_TEXT structured properties without `allowedValues` were getting `condition: undefined` (backend defaults to EQUAL), so typing a partial value in the filter input returned nothing. Added `getNewFilterCondition()` to set `FilterOperator.Contain` for text fields, and routed free-form STRING/RICH_TEXT props to `FieldType.TEXT` in `getDynamicFilterField`.

- **Numeric type detection no longer misidentifies UUID values as numbers.** `getStructuredPropFilterDisplayName` used `parseFloat()` to detect numbers, which parses UUID strings (e.g. `"550e8400-..."`) as `Infinity`. Now checks `valueType.urn` against `NUMBER_TYPE_URN` instead.

- **`TextValueMenu` now shows the property's display name instead of its raw field key.** The filter input title was showing `structuredProperties.myProp` rather than the human-readable name. Now uses `useFilterDisplayName`, consistent with the enum and entity filter menus.

- **NUMBER-typed filter buckets now render the formatted value instead of the raw ES double.** `useLoadAggregationOptions` passed `field.entity` to `getStructuredPropFilterDisplayName`, but `field.entity` doesn't reliably carry the structured property's definition (`valueType`) in every dropdown code path. Without it, number buckets rendered as `"42000.0"` in the dropdown, inconsistent with the formatted value the selected-filter chip showed for the same value. Now prefers the aggregation facet's own entity, which does carry the definition, falling back to `field.entity`.

## Testing

- Unit tests added for numeric type detection, UUID safety, and CONTAINS condition routing (`utils.test.tsx`)
- Java unit tests added for URN and STRING structured property keyword subfield mapping (`V2MappingsBuilderTest.java`)

## Checklist

- [x] URN-type structured property filter option appears in the search sidebar
- [x] STRING property with a UUID value displays the UUID correctly
- [x] Free-form STRING property: typing a partial value returns matching results
- [x] NUMBER property filter dropdown shows the formatted value, matching the selected-filter chip
- [x] CI green

### DIFF
--- a/datahub-web-react/src/app/searchV2/filters/__tests__/useSearchFilterDropdown.test.tsx
+++ b/datahub-web-react/src/app/searchV2/filters/__tests__/useSearchFilterDropdown.test.tsx
@@ -1,0 +1,60 @@
+import { act, renderHook } from '@testing-library/react-hooks';
+
+import useSearchFilterDropdown from '@app/searchV2/filters/useSearchFilterDropdown';
+import { STRUCTURED_PROPERTIES_FILTER_NAME } from '@app/searchV2/utils/constants';
+import { STRING_TYPE_URN } from '@src/app/shared/constants';
+
+import { FacetMetadata, FilterOperator } from '@types';
+
+// The hook pulls its search context and the aggregation lazy-query from other hooks; stub both so we
+// can exercise updateFilters in isolation. updateFilters does not run the aggregation query, so the
+// lazy-query mock only needs to satisfy the [trigger, state] tuple shape.
+vi.mock('@app/searchV2/useGetSearchQueryInputs', () => ({
+    default: () => ({ entityFilters: [], query: '*', orFilters: [], viewUrn: undefined }),
+}));
+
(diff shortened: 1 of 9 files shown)


david-leifker:
Prefer https://github.com/datahub-project/datahub/pull/18842

rob-1019:
Closing in favor of #18842, which was merged and covers the same structured-property filtering/URN-keyword-handling scope (plus more) - thanks for pointing to it.

PR #18009, opened 2026-06-23
perf(auth): cache resource owners per authorize() call to remove ownership-fetch.

Dutt23:
## Summary

`PolicyEngine.getOwnersForType()` fetched a resource's `ownership` aspect via
`entityClient.getV2(OWNERSHIP_ASPECT_NAME)` **once per ownership-based policy**, even when
multiple such policies were evaluated against the **same resource** in a single `authorize()`
call. On pages that authorize many assets under several ownership policies this produced a
large N+1 of uncached backend reads (e.g. 25 assets × 10 ownership policies ≈ 250 `getV2`
calls per page load).

This change caches the resolved owner list **per `authorize()` call**, keyed by resource URN,
so each resource's ownership is fetched at most once regardless of how many ownership policies
reference it. The full owner list is cached and ownership-type filtering is applied in memory
(the `getV2` already returns all owners), so different `ownershipTypes` reuse the same fetch.

## Changes

- `PolicyEvaluationContext` gains a `Map<String, List<Owner>> resourceOwnersByUrn` cache
  (request/authorize-call scoped — created once per `authorize()` and shared across the policies
  evaluated in that call).
- `getOwnersForType()` now reads through `computeIfAbsent(resourceUrn, ...)` and filters owners
  by type in memory. Extracted the actual fetch into a new `fetchOwners()` helper.
- Empty/missing ownership is negative-cached (empty list) so resources without owners are not
  re-fetched.

## Behavior

No functional change. Same owner-match semantics, same ownership-type filtering, same
ALLOW/DENY outcomes. Only the number of backend `getV2` calls is reduced.

## Scope

Improves resource-authorization paths (search / browse / list / entity views) where access is
ownership-driven. Does not affect platform-privilege checks (no resource → ownership path never
runs).

### DIFF
--- a/metadata-service/auth-impl/src/main/java/com/datahub/authorization/PolicyEngine.java
+++ b/metadata-service/auth-impl/src/main/java/com/datahub/authorization/PolicyEngine.java
@@ -25,6 +25,7 @@ import io.datahubproject.metadata.context.ServicesRegistryContext;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


david-leifker (approved):
lgtm!

PR #18014, opened 2026-06-24
feat(ingest): default emit mode to async for Airflow, Dagster, Prefect, and Great Expectations plugins

aviraj-gour:
## Summary

The orchestration/validation plugins emitted metadata to GMS synchronously (`SYNC_PRIMARY`), which blocks on a primary-storage commit per write. At high run/validation volume this puts avoidable load on GMS and its primary store. This PR defaults these plugins to **ASYNC** emit so GMS can batch writes, while leaving an explicit opt-out for callers that need synchronous guarantees.

## Why `default_emit_mode` on the emitter/graph client?

`emit_mode` was previously only a per-call argument, defaulting to a single process-global value (`_DEFAULT_EMIT_MODE`). That left no way to say "this client should default to ASYNC" without overriding the global for the whole process or passing `emit_mode` at every single call site.

So the default now lives on the **instance**:

- **`DataHubRestEmitter`** holds a per-instance `_default_emit_mode`. The per-call `emit*` parameters became `Optional[EmitMode] = None`; when `None`, the call resolves to the instance default, which itself falls back to the global env-driven default. So existing SDK/CLI behavior is unchanged unless a caller sets it — the global default is preserved as the final fallback.
- **`DataHubGraph` / `DatahubClientConfig`** — the default is a `default_emit_mode` field on `DatahubClientConfig` rather than a constructor sideband, so it's mapped like every other emitter parameter and can be set declaratively in one place. `DataHubGraph.__init__` forwards it to the underlying emitter, and `DataHubGraph.from_emitter` preserves the source emitter's `default_emit_mode` so `emitter.to_graph()` no longer silently reverts to the global default.
- **`EmitMode` moved to its own leaf module** (`datahub/emitter/emit_mode.py`) so `graph.config` can reference it for the field without a circular import; `rest_emitter` re-exports it so existing `from datahub.emitter.rest_emitter import EmitMode` imports keep working.

This is what makes the per-plugin change tractable: a plugin sets the default once at construction, and every emit the emitter/client performs — including those routed through shared helpers (`DataJob`/`DataFlow`/`DataProcessInstance` emits, validation-result emits) that don't take an `emit_mode` argument — inherits it, without threading the mode through each call path.

## Per-plugin changes

Each plugin defaults to `ASYNC` with an opt-out config:

- **Airflow** — `[datahub] emit_mode` in `airflow.cfg`
- **Dagster** — `emit_mode` on `DatahubDagsterSourceConfig` (forwarded onto the client config)
- **Prefect** — `emit_mode` on the `DatahubEmitter` block (forwarded as `default_emit_mode` to the underlying emitter so all helper emit paths pick it up)
- **Great Expectations** — `emit_mode` kwarg on `DataHubValidationAction`, coerced from the checkpoint YAML string to an `EmitMode` enum (raises a descriptive error listing valid values on an invalid string)

## Behavioral impact

With async, `emit()` no longer raises on a server-side commit/validation rejection and there is no read-after-write guarantee. Transport, auth, and HTTP 4xx/5xx errors still raise as before. Documented as a breaking change in `updating-datahub.md`; users who need the old behavior set `emit_mode = SYNC_PRIMARY` (or `SYNC_WAIT`).


docs/how/updating-datahub.md
```diff
@@ -44,6 +44,8 @@ Requirements:
 
 ### Breaking Changes
 
+- **(Airflow / Dagster / Prefect / Great Expectations plugins)** The default emit mode changed from synchronous to **ASYNC** to reduce GMS load at high volume. With async, `emit()` no longer raises on a rejected write and there's no read-after-write guarantee. The Python SDK, CLI, and DataHub Actions defaults are unchanged. **Action:** to restore the previous behavior (`SYNC_PRIMARY`), set `[datahub] emit_mode = SYNC_PRIMARY` in `airflow.cfg` (Airflow), `emit_mode: SYNC_PRIMARY` on the Dagster source config, `emit_mode=EmitMode.SYNC_PRIMARY` on the Prefect `DatahubEmitter` block, or `emit_mode: SYNC_PRIMARY` on the Great Expectations `DataHubValidationAction` config.
```
```suggestion
- #18014 **(Airflow / Dagster / Prefect / Great Expectations plugins)** The default emit mode changed from synchronous to **ASYNC** to reduce GMS load at high volume. With async, `emit()` no longer raises on a rejected write and there's no read-after-write guarantee. The Python SDK, CLI, and DataHub Actions defaults are unchanged. **Action:** to restore the previous behavior (`SYNC_PRIMARY`), set `[datahub] emit_mode = SYNC_PRIMARY` in `airflow.cfg` (Airflow), `emit_mode: SYNC_PRIMARY` on the Dagster source config, `emit_mode=EmitMode.SYNC_PRIMARY` on the Prefect `DatahubEmitter` block, or `emit_mode: SYNC_PRIMARY` on the Great Expectations `DataHubValidationAction` config.
```

metadata-ingestion/src/datahub/emitter/emit_mode.py
```diff
@@ -0,0 +1,25 @@
...
+    # a balance between consistency and performance. Suitable for updates that need to be immediately reflected in direct
+    # entity retrievals but where search index consistency can be slightly delayed.
+    SYNC_PRIMARY = auto()
+    # Queues the metadata change for asynchronous processing and returns immediately. The client continues execution without
+    # waiting for the change to be fully processed. Best for high-throughput scenarios where eventual consistency is acceptable.
+    ASYNC = auto()
+    # Queues the metadata change asynchronously but blocks until confirmation that the write has been fully persisted.
+    # More efficient than fully synchronous operations due to backend parallelization and batching while still providing
+    # strong consistency guarantees. Useful when you need confirmation of successful persistence without sacrificing performance.
+    ASYNC_WAIT = auto()
```
sgomezvillamor, line 21: This or a follow-up PR should also properly document this in the user-facing docs.

sgomezvillamor (approved):
Solid PR! Clean design — per-instance default with fall-through to the global env-driven default means non-plugin SDK/CLI callers are completely unaffected. The re-export preserves the existing import path, each plugin has focused unit tests, and the breaking change is clearly documented with concrete migration steps.

A few things worth addressing before merge:

**🟡 Dagster plugin mutates `DatahubClientConfig` in-place**
`datahub_sensors.py` does `self.config.datahub_client_config.default_emit_mode = self.config.emit_mode` which permanently mutates the config object passed in by the caller. `DatahubClientConfig` isn't frozen so it works at runtime, but it's a latent side-effect bug if the config is ever reused or shared. Prefer:
```python
self.graph = DataHubGraph(
    self.config.datahub_client_config.model_copy(
        update={"default_emit_mode": self.config.emit_mode}
    )
)
```

**🟡 `async_flag=False` silent behavior change on plugin emitters**
In the old code, `async_flag=False` fell through to the global `_DEFAULT_EMIT_MODE` (SYNC_PRIMARY). Now it falls through to `_default_emit_mode`, which for plugin emitters is ASYNC. So a plugin calling `emit_mcp(mcp, async_flag=False)` no longer forces sync. `async_flag` is deprecated so impact is low, but consider adding an `elif async_flag is False: emit_mode = EmitMode.SYNC_PRIMARY` guard or a comment explaining the intent.

**ℹ️ Good opportunity to add docstrings to `DataHubRestEmitter` and `DataHubGraph`**
Neither class has constructor or method docstrings today, and the inline comments added in this PR (explaining `_default_emit_mode` semantics, the `from_emitter` preservation, the global fallback) would have much better visibility as proper docstrings. Worth doing here or in a follow-up, and the `emit_mode.py` block comments could be promoted at the same time.

---

treff7es (approved):

aviraj-gour:
Thanks @sgomezvillamor for review,

> 🟡 Dagster plugin mutates DatahubClientConfig in-place
> datahub_sensors.py does self.config.datahub_client_config.default_emit_mode = self.config.emit_mode which permanently mutates the config object passed in by the caller. DatahubClientConfig isn't frozen so it works at runtime, but it's a latent side-effect bug if the config is ever reused or shared. Prefer:
> 
> self.graph = DataHubGraph(
>     self.config.datahub_client_config.model_copy(
>         update={"default_emit_mode": self.config.emit_mode}
>     )
> )

FIxed.

> 🟡 async_flag=False silent behavior change on plugin emitters
> In the old code, async_flag=False fell through to the global _DEFAULT_EMIT_MODE (SYNC_PRIMARY). Now it falls through to _default_emit_mode, which for plugin emitters is ASYNC. So a plugin calling emit_mcp(mcp, async_flag=False) no longer forces sync. async_flag is deprecated so impact is low, but consider adding an elif async_flag is False: emit_mode = EmitMode.SYNC_PRIMARY guard or a comment explaining the intent.

Added a `elif`.

metadata-ingestion/src/datahub/emitter/emit_mode.py
aviraj-gour: Will take user facing docs in follow up PR.

PR #18017, opened 2026-06-24
feat(ingestion/powerbi): separate Hive and Hadoop platforms for disti…

NehaGslab:
## Summary
Splits Hive and Hadoop into two distinct platforms in the Power BI ODBC platform resolver. Previously both were folded into a single hadoop bucket via the shared regex `r"(hadoop|hive)"`, so any Hive ODBC source (driver name or DSN mapped through `dsn_to_platform_name`) resolved to the DataHub hadoop platform.

That had two consequences:

Broken native-SQL lineage — the resolved platform is passed straight to `get_dialect()` when parsing native SQL from the M-Query. sqlglot has no hadoop dialect, so parsing aborted with Unknown dialect 'hadoop' and lineage was silently skipped.
Mismatched upstream URNs — even where parsing succeeded, upstream datasets were emitted under `urn:li:dataPlatform:hadoop,...`, which never matched the Hive tables ingested by DataHub's Hive connector (platform hive), leaving the lineage dangling.
After this change, a Hive source resolves to the hive platform, which maps to a real sqlglot dialect and matches the Hive connector's URNs.

## Changes
m_query/odbc.py: separate hive `(r"hive")` and hadoop `(r"(hadoop)")` entries in `platform_patterns`, and add the corresponding hive: "Hive" / hadoop: "Hadoop" entries in `powerbi_platform_names`. hive is ordered before hadoop, and the two regexes are now disjoint.
tests/unit/powerbi/test_odbc.py: add a Hive ODBC driver case to the driver/server/platform fixtures, plus focused tests asserting (a) Hive and Hadoop resolve to distinct platforms and (b) the platform resolved for a Hive DSN maps to a valid sqlglot dialect (regression guard for the Unknown dialect 'hadoop' failure).

## Backward compatibility

This is a **bug fix, not a breaking change** — the previous `hadoop` mapping for
Hive sources was non-functional:

- Native SQL parsing aborted with `Unknown dialect 'hadoop'`, so Hive lineage was
  silently dropped (no edges emitted) in the common case.
- Where parsing did succeed, upstreams were emitted under
  `urn:li:dataPlatform:hadoop,...`, which never matched DataHub's Hive connector
  (`hive`) URNs, leaving the lineage dangling.

After this change, Hive ODBC sources resolve to the `hive` platform: native SQL
parses against a real sqlglot dialect, and upstream URNs align with the Hive
connector.

**Effect on existing deployments:** new runs emit `hive` upstreams instead of
`hadoop`. Any previously-emitted `hadoop` upstreams were already broken/dangling;
they are retired automatically by stateful stale-entity removal on the next run,
or can be cleaned up manually. Hadoop (non-Hive) resolution is unchanged. No
config migration is required — `dsn_to_platform_name` mappings continue to work
as-is.

## Checklist

- [x] PR conforms to the Contributing Guideline (PR Title Format: feat(ingestion/powerbi): ...)
- [ ] Docs — n/a (internal platform-resolution behavior; no user-facing config change)


metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/pattern_handler.py
```diff
@@ -1834,6 +1840,13 @@ def expression_lineage(
                 break
 
         if (
+            data_platform in ODBC_TWO_TIER_PLATFORMS
+            and schema_name is not None
+            and table_name is not None
+        ):
+            # Drop the pseudo-catalog database_name for two-tier platforms.
+            qualified_table_name = f"{schema_name}.{table_name}"
```
treff7es: **Should fix — schema-less two-tier source silently emits the pseudo-catalog as the schema.**

This branch only fires when `schema_name is not None`. If a Hive ODBC source is navigated as `Database="HIVE"` + `Table="t"` with no `Schema` node, both this branch and the three-tier `elif` are skipped, and control reaches the final `elif database_name is not None and table_name is not None:` → `qualified_table_name = "HIVE.t"`, producing `urn:li:dataPlatform:hive,HIVE.t,PROD`. That's exactly the dangling, non-matching upstream URN this PR exists to eliminate — and it happens with no `report.warning()`/`logger.warning()`.

The path is reachable: the M-Query resolver reads each navigation level's `Kind` verbatim and doesn't guarantee a `Schema` level. Suggested shape:

```python
if data_platform in ODBC_TWO_TIER_PLATFORMS and table_name is not None:
    if schema_name is not None:
        qualified_table_name = f"{schema_name}.{table_name}"
    else:
        # database_name here is the pseudo-catalog (e.g. "HIVE"), not a real
        # schema; emitting it would produce a dangling URN. Surface instead.
        self.reporter.warning(
            title="Cannot build two-tier ODBC table name",
            message="Two-tier ODBC navigation had no schema level; skipping lineage.",
            context=f"table-name={self.table.full_name}, data-platform={data_platform}, database={database_name}, table={table_name}",
        )
        return Lineage.empty()
elif ...
```

Please also add a `test_odbc_two_tier_platform_with_no_schema` to lock in the intended behavior.

metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/pattern_handler.py
```diff
@@ -1522,6 +1522,12 @@ def create_lineage(
         )
 
 
+# Two-tier platforms whose connector URNs are schema.table. Their ODBC
+# navigation exposes a pseudo-catalog (e.g. Hive's constant "HIVE") that must be
+# dropped so URNs match (HiveSource is a TwoTierSQLAlchemySource).
+ODBC_TWO_TIER_PLATFORMS = {"hive"}
```
treff7es: Non-blocking (type design): two-tier-ness is a property of a data platform, but here it lives as a free-floating string set that must stay in lockstep with the `platform_patterns` key in `odbc.py` and `SupportedDataPlatform.HIVE` in `config.py`, with nothing enforcing it — a rename in one place silently disables the catalog drop. Cheapest hardening: derive the member from the enum, e.g. `ODBC_TWO_TIER_PLATFORMS = {SupportedDataPlatform.HIVE.value.datahub_data_platform_name}`. (A larger but cleaner option is an `is_two_tier` flag on `DataPlatformPair` as the single source of truth — probably out of scope here.)

metadata-ingestion/tests/unit/powerbi/test_pattern_handler.py
```diff
@@ -614,6 +616,90 @@ def test_remap_column_lineage_empty_inputs():
...
+    head: Optional[IdentifierAccessor] = None
+    for kind, name in reversed(levels):
+        head = IdentifierAccessor(
+            identifier=name, items={"Kind": kind, "Name": name}, next=head
+        )
+    assert head is not None
+    return head
+
+
+def test_odbc_two_tier_platform_drops_pseudo_catalog():
```
treff7es, line 651: These two unit tests are good (they pin the two-tier-drops-catalog contract and guard the three-tier non-regression). Two coverage gaps to consider:

1. The schema-less two-tier shape (`Database` + `Table`, no `Schema`) isn't exercised — that's the case that currently produces a wrong URN (see the comment on `pattern_handler.py`).
2. These call `expression_lineage()` directly with a hand-built accessor chain, bypassing M-Query tokenization, `Odbc.DataSource` dispatch, and driver→platform resolution. An end-to-end Hive datasource case in `tests/integration/powerbi/test_m_parser.py` (mirroring `test_mysql_odbc_regular_case`, with a golden/URN assertion) is what would actually catch the `Unknown dialect 'hadoop'` and pseudo-catalog regressions against realistic input.

metadata-ingestion/tests/unit/powerbi/test_odbc.py
```diff
@@ -108,3 +113,28 @@ def test_dsn_mapping():
...
+
+    Previously both were folded into 'hadoop' via a shared ``(hadoop|hive)``
+    regex, which (a) emitted upstream URNs under the wrong platform and
+    (b) broke native SQL parsing because sqlglot has no 'hadoop' dialect.
+    """
+    assert normalize_platform_name("hive") == ("hive", "Hive")
+    assert normalize_platform_name("hadoop") == ("hadoop", "Hadoop")
+
+
+def test_hive_platform_has_valid_sqlglot_dialect():
```
treff7es, line 129: Good regression guard — pinning the real root cause (sqlglot has no `hadoop` dialect) rather than an implementation detail. Two minor notes:

- This fixes the dialect issue for `hive` specifically; the underlying failure class (any resolved ODBC platform with no sqlglot dialect → silently dropped lineage) is still open. A systemic guard is worth a follow-up ticket.
- The parallel-list loop in `test_platform_extraction` uses `zip(..., strict=False)`; since you just appended rows to several parallel lists, `strict=True` (or a length assertion) would catch a future list-length drift cheaply.

treff7es (changes requested):
Thanks for this — the Hive/Hadoop split is a clean, well-motivated fix and the tests pin the real root causes (sqlglot dialect resolution + matching the Hive connector's `schema.table` URNs). I have one should-fix before merge plus a few non-blocking suggestions, left as inline comments.

**Should fix**
- The new two-tier branch only fires when a `Schema` level is present. A Hive ODBC source navigated as `Database="HIVE"` + `Table` (no `Schema` node) falls through to the `database.table` branch and silently emits `urn:li:dataPlatform:hive,HIVE.<table>,PROD` — re-introducing the dangling pseudo-catalog URN this PR sets out to eliminate, with no warning logged. The path is reachable (the M-Query resolver reads navigation `Kind` verbatim and doesn't force a `Schema` level) and is currently untested.

**Suggestions (non-blocking)**
- Couple `ODBC_TWO_TIER_PLATFORMS` to `SupportedDataPlatform` so the platform string can't drift.
- Add an end-to-end Hive ODBC case to `tests/integration/powerbi/test_m_parser.py` (the MySQL ODBC harness already exists) — that's the test that would catch both the dialect and pseudo-catalog issues against realistic input.
- The `get_dialect()` fix is a point fix for `hive`; a systemic guard (warn when a resolved ODBC platform has no sqlglot dialect) would prevent the next "Unknown dialect 'X'" recurrence — worth a follow-up ticket rather than expanding this PR.

Nice touches worth calling out: the disjoint regex split with deterministic ordering, the three-tier BigQuery non-regression test, and `SupportedDataPlatform.HIVE` being correctly wired into the default `dataset_type_mapping` / `known_platforms` via enum iteration.

metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/pattern_handler.py
NehaGslab: Have made the code change as per suggestion also added test case for it.

metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/pattern_handler.py
NehaGslab: Have used enum instead of hard coded string as per suggestion.

treff7es (approved):
Re-reviewed — the change-request blocker is fully resolved. ✅

- **Schema-less two-tier fix (should-fix):** `pattern_handler.py` now warns + returns empty lineage instead of emitting the `HIVE.<table>` pseudo-catalog URN, and `test_odbc_two_tier_platform_with_no_schema` pins it (asserts no upstreams + the warning fires). Verified the logic is wired correctly — `data_platform` at the call site is the DataHub platform name, so the `ODBC_TWO_TIER_PLATFORMS` membership test actually matches.
- **Enum coupling:** `ODBC_TWO_TIER_PLATFORMS` is now derived from `SupportedDataPlatform.HIVE`, so a platform-name rename can't silently disable the catalog drop. 👍

Approving. Two non-blocking follow-ups (fine to defer):
1. An end-to-end Hive ODBC case in `tests/integration/powerbi/test_m_parser.py` (mirroring `test_mysql_odbc_regular_case`) would exercise the full M-Query → dispatch → resolution path that the direct `expression_lineage()` unit tests bypass.
2. `test_platform_extraction` still uses `zip(..., strict=False)` — `strict=True` would cheaply catch parallel-list drift.

NehaGslab:
Suggestions from comment are addressed.

PR #18021, opened 2026-06-24
feat(rate-limit): multi-dimensional GMS rate limiting (adaptive capacity + per-actor/class/global scoped chain + per-resolver gates)

Dutt23:
## Summary

Adds a built-in, multi-dimensional rate-limiting subsystem for the GMS HTTP serving path (GraphQL, OpenAPI, Rest.li, `/auth/*`). It protects the Metadata Service from overload and abuse with three complementary mechanisms — adaptive in-flight **capacity**, cluster-wide **endpoint** token buckets, and a per-request **scoped chain** (per-actor → client-class → fleet-global) plus per-resolver **heavy-operation** gates. Everything is off by default and gated behind a master kill switch.

This is **not** MCP ingest throttling or Kafka-lag backpressure (`APIThrottle` / `metadataChangeProposal.throttle`); it sits in front of GMS request handling and is independent of those.

## Why

GMS had no first-class control over incoming API load. A single heavy client, a hot GraphQL resolver, or a burst of traffic could saturate threads/datastores. This adds layered protection that answers three distinct questions — "is the pod accepting more concurrent work?" (capacity), "is this path over its rate?" (endpoint), and "is this actor/class/fleet over its fair share, and is this a known-heavy resolver?" (scoped chain) — with tenant isolation so shared infrastructure stays fair across deployments.

## Architecture

Three independent layers, evaluated per request; a request must pass every layer that applies.

1. **Capacity (per-pod, adaptive).** Netflix Gradient2 adaptive-concurrency limiters cap in-flight requests per pool (`capacity.default`, `capacity.graphql`, optional operation pools), shedding load before latency/thread-pools degrade. Pools are independent (no shared parent budget). `AdaptiveCapacityLimiter` + `RuleSelector` (finest-grain match wins).
2. **Endpoint (cluster-wide, rate).** Bucket4j token buckets backed by a shared Hazelcast map — cluster totals across replicas — for path/auth limits (e.g. signup, OAuth register). `EndpointRateLimitStore`.
3. **Scoped chain (per-request).** Token buckets consumed narrow→broad with fail-fast and refund-on-deny: **per-actor** (`{tenant}:actor:{urn}`) → **client-class** (`{tenant}:browser` / `{tenant}:sdk`) → **fleet-global** (`global`, the only cross-tenant bucket). Plus **heavy-resolver** gates (`{tenant}:op:{resolver}`) for explicitly configured expensive top-level GraphQL fields. `RateLimitEngine` + `EndpointRateLimitStore`.

### Key design decisions

- **Tenant isolation.** Every tenant-scoped bucket key is prefixed `{tenantId}:` (tenantId = the deployment's `global.id`, via `RATE_LIMITS_TENANT_ID`), so a shared Hazelcast map stays isolated across deployments. `global` is intentionally un-prefixed — the single fleet-wide ceiling.
- **Single GraphQL parse.** `GraphQLOperationNameResolver.analyze()` parses the query document exactly once and returns `{queryName, identity, topLevelFields}`; the controller reuses all three. No repeated parsing in the rate-limit path, and the heavy gate stays at the controller (deliberately **not** coupled to a GraphQL instrumentation).
- **Spoof-resistant identity / heavy gate.** For unnamed operations the rate-limit identity is the sorted top-level field names (not a client-supplied name); the heavy-resolver gate matches on `topLevelFields` extracted from the AST (alias-stripped, inline-fragment and fragment-spread resolved, cycle-guarded) so it can't be dodged by renaming the operation or hiding the field in a fragment.
- **Client classification.** The Play frontend stamps `X-DataHub-Request-Source` (BROWSER vs SDK); GMS reads it via `ClientClassifier` to pick the browser/sdk class bucket. (GMS is not called directly in this topology; header-spoofing risk is accepted and documented.)
- **Fail-open + observability.** Unexpected limiter errors fail open (request allowed) and increment a `recordFailOpen` metric rather than dropping traffic; `failOpen` is configurable.
- **System actor exemption.** The internal system principal can bypass heavy-resolver buckets (`exemptSystemActor`) so background/internal traffic isn't throttled.
- **Refund-on-deny.** When a later stage of the scoped chain denies, already-consumed upstream buckets are refunded (Bucket4j `addTokens`, capped at capacity) so a rejected request doesn't permanently burn an actor's/tenant's tokens. Toggle via `scoped.refundDisabled`.
- **Config is env-driven from one file.** Policy lives in `rate-limit-config.yaml`, loaded as a Spring `@PropertySource` so it binds into `datahub.gms.rateLimits.*` and resolves `${ENV:default}` placeholders — a single shared file tuned per deployment via `RATE_LIMITS_SCOPED_*` env vars (no per-tenant copy). Startup validation in `RateLimitConfigValidator`. A file mounted at `RATE_LIMITS_CONFIG_FILE` (Spring resource URI, e.g. `file:/etc/...`) layers on top, mainly for the per-customer `scoped.heavyResolvers` map.

## Config loading: Spring @PropertySource (not a custom loader)

Rate-limit config now loads entirely through Spring: bundled defaults in `application.yaml`
plus an optional mounted override at `RATE_LIMITS_CONFIG_FILE`, both bound to `RateLimitProperties`
via relaxed `@ConfigurationProperties` — the same path the rest of GMS config uses.

### Removed (with startup WARNs, not silent)
- `RATE_LIMITS_CONFIG_FILE_ENABLED` — the override is enabled simply by pointing
  `RATE_LIMITS_CONFIG_FILE` at a file (presence = enabled); the separate flag is redundant.
- `RATE_LIMITS_CONFIG_JSON` — inline-JSON overlays required the custom Jackson loader we removed.
  Put overrides in the mounted YAML file instead.

Both were features of an earlier Jackson-based loader. `RateLimitEngineFactory` logs a clear WARN
at startup if either is still set, so no one loses config silently.

### Why this is better
- One well-understood binding + precedence model, identical to all other GMS config.
- No bespoke merge/loader code to maintain.
- OS environment variables outrank both files, so `RATE_LIMITS_*` env overrides reliably win —
  the recommended way to tune scalars per deployment. Verified by `RateLimitConfigOverridePrecedenceTest`.

## Configuration

- Master switch `RATE_LIMITS_ENABLED` (bypasses everything when false); each layer opt-in (`capacity.enabled` / `endpoint.enabled` / `scoped.enabled`), all default off.
- Hazelcast is provisioned only when endpoint/scoped limiting is enabled (`RateLimitEndpointEnabledCondition`); endpoint/scoped bucket entries use idle + LRU eviction (`bucketMaxIdleSeconds`, `bucketMaxSize`).
- Full env-var reference in `docs/deploy/environment-vars.md`; operations guide in `docs/deploy/gms-rate-limiting.md`.

## Observability & ops

- 429 responses carry `X-DataHub-RateLimit-*` headers (rule, type, source, endpoint-rule) and `Retry-After` (with configurable jitter; endpoint uses `max(minRetryAfter, bucket wait)`).
- Prometheus metrics `gms.rate_limit.*` (allow/deny counts, adaptive limit/inflight gauges, endpoint remaining).
- Inspection API (requires `Manage System Operations`): `GET /openapi/v1/rate-limits/config` (effective merged config) and `/status` (live per-pod state) — `RateLimitController`.

## Scope / non-goals

- Capacity limits are **per-pod** (no cluster-wide exact concurrency cap in v1); endpoint and scoped buckets are cluster-wide via Hazelcast.
- Does not change MCP ingest throttle, MCE/MCL consumer backpressure, or ingestion-connector behavior.
- Config changes require a pod restart; already-created Hazelcast buckets keep their sizing until idle-evicted on a rolling restart.

## Testing

Unit/integration coverage across the subsystem: `RateLimitEngineTest`, `RateLimitEngineScopedChainTest`, `RateLimitAbuseScenarioTest` (single-actor flood, SDK-vs-browser isolation, fleet-global shedding, concurrent atomic flood), `EndpointRateLimitStoreTest`, `RuleSelectorTest`, `CompiledRateLimitRuleTest`, `GraphQLOperationNameResolverTest` (single-parse + fragment-evasion guards), `GraphQLControllerTest` (front gate + heavy gate), `RateLimitFilterTest`, `RateLimitHeaderWriterTest`, `RateLimitMetricsTest`, `RateLimitControllerTest`, `RateLimitEndpointEnabledConditionTest`, `RateLimitEngineFactoryTest`, `RateLimitConfigValidatorTest`, `RateLimitPropertiesEnvBindingTest`, `PropertiesCollectorConfigurationTest`.

## Follow-ups (separate)

- Companion Helm PR (`datahub-helm-fork`): ConfigMap nests under `datahub.gms.rateLimits` and emits `RATE_LIMITS_CONFIG_FILE` with a `file:` prefix. https://github.com/acryldata/datahub-helm-fork/pull/1096
- Doc pass for residual staleness in `gms-rate-limiting.md` (legacy per-actor bundled rule; document the scoped chain end-to-end).
- Revisit possibly-orphaned `RATE_LIMITS_PER_ACTOR_*` env placeholders.
- Optional: reorder the heavy gate ahead of the scoped chain (fail-fast on the narrowest bucket; matters most with refund disabled).


metadata-service/configuration/src/main/resources/application.yaml
```diff
@@ -212,7 +212,21 @@ datahub:
       endpoint:
         enabled: ${RATE_LIMITS_ENDPOINT_ENABLED:false}
         hazelcastMapName: ${RATE_LIMITS_ENDPOINT_HAZELCAST_MAP:gmsRateLimitEndpointBuckets}
-        rules: []
+        # Idle window for per-actor bucket entries in Hazelcast (idle eviction, not TTL).
+        bucketMaxIdleSeconds: ${RATE_LIMITS_ENDPOINT_BUCKET_MAX_IDLE_SECONDS:300}
+        # LRU per-node backstop on bucket entry count.
+        bucketMaxSize: ${RATE_LIMITS_ENDPOINT_BUCKET_MAX_SIZE:100000}
+        rules:
```
david-leifker: In practice, the configuration file is almost always preferred. My thought was that you practically need to consider rate limits as a whole and across various buckets. I almost considered removing rules block here. When the configuration file is enabled, it completely replaces these. I am interested in your thoughts here.

metadata-service/configuration/src/main/resources/application.yaml
```diff
@@ -212,7 +212,21 @@ datahub:
...
+          # Per-actor GraphQL POST throttle. Active only when endpoint.enabled=true
+          # AND this rule's `enabled` is true. System actor is exempt.
+          - id: graphql-per-actor-default
+            enabled: ${RATE_LIMITS_PER_ACTOR_ENABLED:true}
+            pathPattern: ${RATE_LIMITS_PER_ACTOR_PATH:/api/graphql}
```
david-leifker: We should also consider the non-graphql endpoints, openapi and restli. Potentially consider that we generally are looking to protect against SDK usage and actually allow browser traffic for accessing the UI. I would propose inspecting the RequestContext for the browser/sdk classification and applying different rules here. I think most of the time, the browser is executing curated/tested graphql queries whereas SDK usage or custom scripts are more likely to be a problem.

david-leifker (changes requested):
* The approach seems correct, mostly I am looking for a more robust solution. 
* Needs documentation updates
* Users often run graphql without names, this is a gap in the original implementation to fix

metadata-service/configuration/src/main/resources/application.yaml
Dutt23: The browser/SDK class split already applies across all endpoints the scoped chain runs in both the GraphQL gate and the REST filter (OpenAPI/Rest.li), with separate scoped:browser and scoped:sdk sizes (via clientClassEnabled), and endpoint rules can match on client class. So we can already keep browser generous and tighten SDK everywhere. One caveat: the classification comes from the X-DataHub-Request-Source header, trusted only on the frontend hop, so direct-to-GMS callers default to non-browser. Per-actor coverage for REST is the one follow-up (GraphQL-only right now)

metadata-service/configuration/src/main/resources/application.yaml
Dutt23: Removing rules: [] won't change behavior  the field defaults to an empty list in the POJO, so with no file mounted you get empty either way, and a mounted file's rules still apply cleanly. It's cosmetic (I'd keep it just to document the shape). It also doesn't fix the index-merge footgun that only bites if the defaults carry non-empty rules. If the goal is "file wins entirely when present," that's the separate wholesale-replace change; removing [] doesn't get us there.

david-leifker (approved):

PR #18032, opened 2026-06-24
fix(): Improvements to scalability and execution of datahub-documents ingestion source + tests

jjoyce0510:
# Improve `datahub-documents` source: scalable scroll, EXTERNAL docs, and run locking

## Summary

Hardens the `datahub-documents` source for large corpora and frequent (e.g. every-15-min)
schedules. Three areas: (1) remove the document cap and make pagination tunable, (2) index
EXTERNAL documents with smart ownership handling, and (3) prevent overlapping scheduled runs
from duplicating work via a distributed lock.

### 1. Scalable fetching 

- Switched batch fetching to cursor-based `scrollAcrossEntities` — fetches the full corpus
  across pages with no 10k Elasticsearch window limit.
- Specifically, added new config:
  - `scroll_batch_size` (default `1000`) — page size per scroll request.
  - `scroll_delay_seconds` (default `0.0`) — optional delay between pages to reduce load on
    GMS/Elasticsearch.
  - `max_documents` (default `10000`, `0`/`-1` to disable) — explicit per-run safety cap that
    fails loudly when exceeded, instead of a silent hard-coded limit.

### 2. EXTERNAL document handling

- `include_external_documents` (default `True`) — EXTERNAL documents (`sourceType=EXTERNAL`)
  are now indexed by default, not just NATIVE. `platform_filter` still restricts which
  platforms are included; set to `False` to restore NATIVE-only behavior.
- `skip_external_if_semantic_content_exists` (default `True`) — ownership model for EXTERNAL
  docs:
  - Some sources (e.g. Notion, Confluence) perform their own semantic indexing and emit the
    `semanticContent` aspect themselves.
  - If an EXTERNAL document already has `semanticContent` for the active model and is **not**
    in our state, we treat the source as the owner: we skip it and do **not** record it in our
    state (never claim ownership).
  - If an EXTERNAL document has **no** `semanticContent`, we index it and record it in state
    (take ownership), then re-embed on future content changes.
  - NATIVE documents are never skipped this way. `force_reprocess` and "already owned in state"
    short-circuit the check.

### 3. Distributed run lock (prevent overlapping runs)

- New `DocumentIndexingLock` backed by a `dataHubStepState` entity used as a key/value lease
  store. Writes go through the metadata API with `EmitMode.SYNC_PRIMARY` so the lease is
  committed to the primary store (MySQL) and concurrent runners see it immediately (no reliance
  on eventually-consistent search).
- Cold start uses `CREATE_ENTITY` + `CREATE` with `If-None-Match: *` for an atomic first claim;
  releases/renewals/takeovers use UPSERT + re-read confirm.
- Lease has a TTL with periodic heartbeat renewal, so a multi-hour run keeps its lease alive
  while a crashed run's lease expires and frees the slot.
- When the lock is held by another run, the new run exits cleanly without processing (reported,
  not a failure).
- New `locking` config (`LockConfig`): `enabled` (default `True`), `lock_id`,
  `lock_ttl_seconds` (default `1800`), `lock_renewal_interval_seconds` (default `300`, validated
  to be smaller than the TTL).

## Supporting changes

- `emitter/request_helper.py`: forward `mcp.headers` to OpenAPI requests so `If-None-Match`
  works on both RestLI and OpenAPI endpoints.
- `unstructured/chunking_source.py`: expose `get_model_embedding_key()` for the
  semantic-content existence check.

## Tests

- Unit tests for scroll pagination config, EXTERNAL filtering and the source-ownership skip
  (including batch integration tests verifying skipped EXTERNAL docs are not written to state),
  and the distributed lock (no-lock/cold-start, expired-lease takeover, held-by-other,
  re-entrant, race loss, release, heartbeat renew/throttle, and lease-state predicate).


nwadams:
Is there a known documented way to manually release a stuck lock if a job fails/dies? Do locks geta uto released when the job dies?

jjoyce0510:
Yes locks get released at end of process if it completes. If it doesn't they have a TTL after which next job will simply re-acquire the lock.

There is aheart beat flow to keep the lock acquired for the duration of the job run

metadata-ingestion/src/datahub/emitter/request_helper.py
```diff
@@ -83,7 +83,8 @@ def from_mcp(
             url = f"{gms_server}/openapi/v3/entity/{mcp.entityType}/{mcp.entityUrn}"
         else:
             if mcp.aspect:
-                mcp_headers = {}
+                headers = getattr(mcp, "headers", None)
```
jjoyce0510, line 86: I needed this for Conditional writes headers.

nwadams (approved):

PR #18036, opened 2026-06-25
feat(ingest/teradata): offload view definitions to disk-backed store …

NehaGslab:
## Summary

On large Teradata installations, `_tables_cache` held every view's full `RequestText` (SQL definition) resident in memory — e.g. 2,000 views × ~100KB ≈ 200MB, scaling linearly with view count and contributing to OOM crashes. This PR moves view SQL text out of the in-memory cache and into a temporary SQLite-backed store, mirroring the pattern already used by the BigQuery connector (`FileBackedDict` for `view_definitions`).

The lightweight per-table metadata (name, description, timestamps) stays in `_tables_cache`; only the heavy SQL text is spilled to disk, keeping peak resident memory bounded regardless of the number of views.

## Changes

- Removed `request_text` from the `TeradataTable` dataclass — the field that drove the memory growth.
- Added `self._view_definitions: FileBackedDict[str]` (instance-level), populated during the existing single-threaded bulk scan in `cache_tables_and_views`, keyed via a new `_view_definition_key(schema, view_name)` helper (lowercased schema to match existing cache-lookup casing).
- Guarded reads with `self._view_definitions_lock` in `cached_get_table_properties`. This is the key difference from BigQuery: Teradata processes views across a `ThreadPoolExecutor`, and `FileBackedDict.__getitem__` mutates its in-memory LRU on read — so concurrent reads must be serialized.
- Cleanup in `close()`: `self._view_definitions.close()` removes the temp SQLite file promptly, consistent with the connector's existing sequential-recipe cache-clearing.
- Updated the (currently disabled) `optimized_get_view_definition` to read from a `Mapping[str, str]` rather than the removed field.

## Why

- Bounded memory: `FileBackedDict` keeps a small in-memory LRU and spills the rest to a temp SQLite DB, so resident view text no longer scales with view count.
- Reuses the bulk query: no extra per-view round-trips to Teradata (the SQL is already fetched in the existing scan).
- Idiomatic: matches the established BigQuery connector pattern, so reviewers recognize the shape.

## Testing

- metadata-ingestion Teradata unit/integration/performance suites pass (275 + 34), including the new coverage.

## Checklist

- [ ] Docs added/updated — N/A (internal memory optimization, no user-facing behavior change)
- [ ] Breaking changes entry in updating-datahub.md — N/A (no behavioral/API change; view definitions are still emitted identically)


metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -2759,6 +2784,15 @@ def cache_tables_and_views(self) -> None:
                             else:
                                 database_counts[table.database]["tables"] += 1
 
+                            # Offload view SQL text to disk rather than holding it in
+                            # the in-memory cache. This loop is single-threaded, so the
+                            # write does not need `_view_definitions_lock` (only the
+                            # concurrent reads later do).
+                            if table.object_type == "View" and entry.RequestText:
+                                self._view_definitions[
```
treff7es: 🔴 **Unguarded write — no error reporting.** `FileBackedDict.__setitem__` fills an in-memory LRU and flushes to the temp SQLite file once `cache_max_size` (default 900) is exceeded — so on large installs this line can raise `sqlite3.OperationalError` mid-loop. It runs inside `__init__` → `cache_tables_and_views`, so an uncaught error aborts source construction with a raw stack trace and no `self.report` entry. The old in-memory field never had this failure mode.

```python
if table.object_type == "View" and entry.RequestText:
    try:
        self._view_definitions[_view_definition_key(table.database, table.name)] = entry.RequestText.strip()
    except Exception as e:
        self.report.warning(title="Failed to store view definition", context=f"{table.database}.{table.name}", exc=e)
```

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -2131,8 +2153,16 @@ def cached_get_table_properties(
...
             if entry.name == table:
                 description = entry.description
-                if entry.object_type == "View" and entry.request_text:
-                    properties["view_definition"] = entry.request_text
+                if entry.object_type == "View":
+                    # View text lives in the disk-backed dict; reads run on parallel
+                    # worker threads and FileBackedDict mutates its LRU on read, so
+                    # the lookup must be serialized.
+                    with self._view_definitions_lock:
+                        request_text = self._view_definitions.get(
```
treff7es: 🟡 **Same silent-raise on the read.** `.get()` is not a pure read here — a cache miss goes through `__getitem__` → `_add_to_cache` → possibly `_prune_cache` (a SQLite write), which can raise. This runs on the parallel view-processing worker threads, so a single flaky read drops that view's `view_definition` with no trace. Wrap it and `self.report.warning(...)` so a missing definition is attributable (lower blast radius than the write, but same shape).

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -1829,11 +1850,12 @@ def __init__(self, config: TeradataConfig, ctx: PipelineContext):
             )
 
             # Disabling the below because the cached view definition is not the view definition the column in tablesv actually holds the last statement executed against the object... not necessarily the view definition
+            # view_definitions = self._view_definitions
```
treff7es: 🟡 **Dead code being maintained.** `optimized_get_view_definition` has no live callers — this commented-out `setattr` block is its only "caller", and the block's own comment says the cached text isn't the real view definition. The PR spends a hunk re-syncing `tables_cache=` → `view_definitions=` on code that will never run. Prefer deleting the function + this block (a one-line "removed — see git history" note preserves intent) over keeping a maintained-but-dead path.

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -655,7 +657,21 @@ class TeradataTable:
...
     last_alter_timestamp: Optional[datetime]
-    request_text: Optional[str]
+    # NOTE: View SQL text is intentionally NOT stored here. On large installations
+    # (thousands of views) keeping every view's RequestText resident drove the
+    # cache to hundreds of MB. View definitions now live in a disk-backed
+    # FileBackedDict (TeradataSource._view_definitions), keyed by
+    # _view_definition_key(schema, name).
+
+
+def _view_definition_key(schema: str, view_name: str) -> str:
```
treff7es, line 673: ℹ️ **Casing invariant is only half-enforced.** The key lowercases `schema` but not `view_name`. Fine for the live path (writer and `cached_get_table_properties` both use stored case), but `optimized_get_view_definition` keys with a caller-supplied name whose case isn't guaranteed — a mismatch silently returns `None`. Lowercasing *both* components makes the contract total and unbreakable. (Latent today since that reader is the dead code above.)

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Wrapped in try/except → report.warning as per suggestion

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Wrapped in try/except → report.warning as per suggestion

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Removed dead code, relates tests and added comment for git history

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Both schema and view_name are lowered now, added test

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
```diff
@@ -1026,27 +1043,10 @@ def grouper(fk_row):
...
-    schema_key = schema.lower()
-    if schema_key not in tables_cache:
-        return None
-
-    for table in tables_cache[schema_key]:
-        if table.name == view_name:
-            return self.normalize_name(table.request_text)
-
-    return None
+# optimized_get_view_definition was removed: the cached "view definition" from
```
treff7es: This comment should be removed

treff7es (approved):

metadata-ingestion/src/datahub/ingestion/source/sql/teradata.py
NehaGslab: Removed comment.

PR #18044, opened 2026-06-25
feat(ingest): opt-in syncIngest-marker-aware async routing for the REST emitter

rob-1019:
## Summary

Adds an opt-in routing mode to the REST emitter and sink. When enabled, a
`SYNC_PRIMARY` request is sent asynchronously **unless** one of its MCPs carries
a `syncIngest` marker in its system metadata, in which case that request is kept
synchronous.

The opt-in is explicit (client config / emitter constructor) and **defaults
off**, so the wire format and behavior are unchanged for existing callers.

## Motivation

A pipeline may want to send the bulk of its writes asynchronously for throughput
while keeping a specific subset synchronous (e.g. writes whose read-after-write
ordering matters). This adds a per-MCP escape hatch — mark the MCPs that must
stay synchronous, let the rest go async — without an all-or-nothing
`mode: SYNC` / `mode: ASYNC` choice.

## What's included

- `request_helper`: `SYNC_INGEST_KEY`, `has_sync_ingest_marker`, and a
  case/whitespace-tolerant `is_sync_ingest_marker_value`.
- `rest_emitter`: `respect_mcp_sync_marker` (the opt-in) and `_is_batch_async`,
  which resolves the wire `async` flag. `SYNC_WAIT` is never upgraded; within a
  batch, any single marked MCP keeps the whole batch synchronous
  (conservative / ordering-safe).
- `DatahubClientConfig` and the REST sink plumb the opt-in through; `mode: SYNC`
  suppresses it so an all-sync sink stays all-sync.

## What this does not do

The marker is **read, not produced**, by this change — nothing here sets
`syncIngest` automatically. A producer populates the `syncIngest`
system-metadata property on the MCPs that must remain synchronous (e.g. via a
custom aspect mutator/validator, or an upstream processing step). With the
marker absent, opting in simply routes all `SYNC_PRIMARY` writes async.

## Testing

Unit coverage for the classification (opt-in upgrades `SYNC_PRIMARY` → async),
the tolerant marker read, batch coarsening, `SYNC_WAIT` never being upgraded,
and the `mode: SYNC` suppression.

## Checklist
- [x] PR conforms to the Contributing Guideline / PR Title Format


metadata-ingestion/src/datahub/emitter/rest_emitter.py
```diff
@@ -693,10 +699,49 @@ def to_graph(self) -> "DataHubGraph":
 
         return DataHubGraph.from_emitter(self)
 
+    def respect_mcp_sync_marker(self) -> bool:
```
asikowitz: Seems redundant -- can we just define instance var `self.respect_mcp_sync_marker`

metadata-ingestion/src/datahub/ingestion/sink/datahub_rest.py
```diff
@@ -227,6 +227,14 @@ def _make_emitter(cls, config: DatahubRestSinkConfig) -> DataHubRestEmitter:
...
             disable_ssl_verification=config.disable_ssl_verification,
             openapi_ingestion=config.endpoint == RestSinkEndpoint.OPENAPI,
+            # mode: SYNC already sends everything synchronously, so marker-aware
+            # sync routing has nothing to do — keep it inert there. It only has an
+            # effect on async-capable sink modes.
+            special_sync_only_for_sync_origin=(
+                config.special_sync_only_for_sync_origin
+                if config.mode != RestSinkMode.SYNC
+                else False
+            ),
```
asikowitz: Imo would be slightly cleaner to always set this and instead have the logic that reads this be smart about it (i.e. only apply logic if `config.mode != RestSinkMode.SYNC`)

metadata-ingestion/src/datahub/ingestion/graph/config.py
```diff
@@ -36,6 +36,15 @@ class DatahubClientConfig(ConfigModel):
...
     openapi_ingestion: Optional[bool] = None
+    # Opt this client into marker-aware sync routing: emits go explicit
+    # async=true unless an MCP carries the syncIngest marker in its system
+    # metadata, which keeps that request at explicit async=false. Never
+    # implicit. The marker is read, not produced, here: a producer must
+    # populate the syncIngest system-metadata property on writes that must remain
+    # synchronous (e.g. via a custom aspect mutator/validator or an upstream
+    # processing step), otherwise opting in routes everything async. Mutually
+    # exclusive with an all-sync sink mode.
+    special_sync_only_for_sync_origin: Optional[bool] = None
```
asikowitz: Feel like we can document this in a better place... also the name `special_sync_only_for_sync_origin` confuses me. Again I see nothing wrong with something simpler like `respect_mcp_sync_marker`. Is there an advantage to having a diff name between graph and emitter?

asikowitz (approved):
Overall seems good. A few questions about naming and simplicity

metadata-ingestion/src/datahub/emitter/rest_emitter.py
rob-1019: done

metadata-ingestion/src/datahub/ingestion/sink/datahub_rest.py
rob-1019: yep, done

metadata-ingestion/src/datahub/ingestion/graph/config.py
rob-1019: Renamed to special_respect_mcp_sync_marker. I advise we retain special_ designation here to the very targeted niche use-case here and the leaky-nature of this abstraction (individual messages upgrading the entire batch).

asikowitz (approved):
LGTM

sgomezvillamor (dismissed):
Instead of a boolean `sync` marker, why not generalize it with an explicit `emitMode` field — `(SYNC_WAIT, SYNC_PRIMARY, ASYNC, ASYNC_WAIT)`?

And rather than burying it in `systemMetadata` properties, why not make it a proper field in the model? https://github.com.com/datahub-project/datahub/blob/master/metadata-models/src/main/pegasus/com/linkedin/mxe/SystemMetadata.pdl

metadata-ingestion/src/datahub/emitter/rest_emitter.py
```diff
@@ -507,6 +511,7 @@ def __init__(
         client_key_path: Optional[str] = None,
         disable_ssl_verification: bool = False,
         openapi_ingestion: Optional[bool] = None,
+        special_respect_mcp_sync_marker: Optional[bool] = None,
```
sgomezvillamor: The `special_` prefix reads oddly, why not just `respect_mcp_sync_marker` (or `respect_mcp_emit_mode`)?

metadata-ingestion/src/datahub/emitter/rest_emitter.py
```diff
@@ -526,6 +531,16 @@ def __init__(
...
         )
+        # Marker-aware sync routing: enabled only when set explicitly via config
+        # or constructor, never implicit. When enabled, a batch is upgraded to
+        # synchronous if any of its MCPs carries the syncIngest marker in its
+        # system metadata; otherwise the configured emit_mode is honored
+        # unchanged. This only ever forces more synchronicity, never less. The
+        # marker is read, not produced, here: a producer must populate the
+        # syncIngest system-metadata property on the MCPs that must remain
+        # synchronous (e.g. via a custom aspect mutator/validator, or an upstream
+        # processing step).
```
sgomezvillamor: @aviraj-gour is working on a PR that refactors some of this emit mode code: https://github.com/datahub-project/datahub/pull/18014 — expect significant conflicts, worth coordinating.

Also, user-facing docs for `RestEmitter`/`DataHubRestSink` emit modes and this event marker are becoming critical at this point.

aviraj-gour: I test-merged both branches locally to check the overlap, they merge with zero conflicts.
In the merged `emit_mcp`, the stages run in the right order:

```python
# from #18014: resolve which emit_mode to use
if async_flag is True:
    emit_mode = EmitMode.ASYNC
elif emit_mode is None:
    emit_mode = self._default_emit_mode

# from this PR: given that emit_mode, decide the wire async flag
effective_async = self._is_batch_async([mcp], emit_mode)
```
@rob-1019 Merging 18014 PR and will be raising a follow up PR for user facing docs

metadata-ingestion/src/datahub/emitter/rest_emitter.py
rob-1019: Renaming based on our discussion.

rob-1019: Done — dropped the `special_` prefix; it's `respect_mcp_sync_marker` now.

metadata-ingestion/src/datahub/emitter/rest_emitter.py
rob-1019: Rebased on master so this now sits on top of #18014 (the merge was clean), and per @aviraj-gour the user-facing emit-mode/marker docs are coming as a follow-up PR.

rob-1019:
Good call on evolvability — I've switched the boolean to a string-valued `emitModeMarker` (only `sync` today; unrecognized values are ignored, so the set can grow without a wire-format change), and kept it upgrade-only with no async value so a producer can pin a write more synchronous but never downgrade the caller's `emit_mode`. I kept it in `systemMetadata.properties` rather than a PDL field because it's a client-side transport hint that GMS never reads and shouldn't be persisted on every aspect — a model field would imply server-side semantics that don't exist, and `properties` is the documented home for exactly this kind of out-of-band signal.

PR #18046, opened 2026-06-25
feat(ingest/confluence): ingest folders as documents to preserve hierarchy

nwadams:
## Summary

Confluence **folders** are a distinct content type that the page APIs (`get_all_pages_from_space`, `get_child_pages`, `get_page_by_id`) never return, so they were never ingested. A page nested under a folder therefore lost its parent link — the folder's URN wasn't among the ingested ids, so `parentDocument` was dropped — and the page showed up as an **orphan root** in the Documents tree even though its full path is known.

This PR materializes folders as `Document` entities so nested pages keep their place in the hierarchy.

## How

Folders already appear in each page's `ancestors` array (`type == "folder"`), so they can be reconstructed with **no extra API calls**:

- `ConfluenceHierarchyExtractor.collect_folder_nodes(...)` walks the `ancestors` of every ingested page and collects each folder, deriving its parent from the preceding ancestor (page or folder).
- Each folder is emitted as a `Document` with the `"Folder"` subtype (matching how DataHub represents natively-created folders), an empty body, a parent link to its nearest ingested ancestor, and a `browsePathsV2` aspect.
- Folder URNs are registered up front so child pages (and nested folders) resolve their parent against them.
- Gated by a new `ingest_folders` config option (default `true`). Folders carry no content, so they skip the chunking/embedding step pages go through.

Result — a page nested as `Overview → Nick demo (folder) → child page` now ingests with `child page.parentDocument = folder` and `folder.parentDocument = Overview`, instead of the child becoming a root.

## Test plan

- New unit tests in `test_confluence_hierarchy.py` covering folder collection (basic, nested folders, folder directly under space, page-ancestors ignored / de-duped). All 143 Confluence unit tests pass.
- `./gradlew :metadata-ingestion:lint` passes (ruff + mypy).
- Verified end-to-end with a live CLI ingest against a test Confluence: pages nested under a folder correctly parent to the folder, the folder parents to its containing page, and the folder is tagged `subType: Folder`.

## Checklist

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/confluence/confluence_hierarchy.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/confluence/confluence_hierarchy.py
@@ -1,8 +1,23 @@
 """Hierarchy extraction logic for Confluence pages."""
 
+from dataclasses import dataclass, field
 from typing import Any, Dict, List, Optional, Set
 
 from datahub.metadata.schema_classes import BrowsePathEntryClass, BrowsePathsV2Class
+
+
+@dataclass
+class FolderNode:
+    """A Confluence folder discovered as an ancestor of an ingested page."""
+
+    id: str
+    title: str
+    # Immediate parent ancestor id (page or folder). None => directly under the space.
+    parent_id: Optional[str]
+    space_name: Optional[str]
+    space_key: Optional[str]
+    # Ancestors that precede this folder, used to build its browse path.
+    ancestor_prefix: List[Dict[str, Any]] = field(default_factory=list)
 
 
 class ConfluenceHierarchyExtractor:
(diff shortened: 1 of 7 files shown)


metadata-ingestion/src/datahub/ingestion/source/confluence/confluence_hierarchy.py
```diff
@@ -244,3 +259,44 @@ def build_browse_path_v2(
             )
 
         return BrowsePathsV2Class(path=path_entries)
+
+    @staticmethod
+    def collect_folder_nodes(pages: List[Dict[str, Any]]) -> Dict[str, FolderNode]:
```
jjoyce0510, line 267: just enusre that we gracefullly fail in case is issues with this..

metadata-ingestion/src/datahub/ingestion/source/confluence/confluence_source.py
```diff
@@ -808,6 +813,99 @@ def _build_imported_document(
...
+        """
+        if not self.config.url:
+            return
+        from urllib.parse import urlparse
+
+        parsed_url = urlparse(self.config.url)
+        instance_id = (
+            parsed_url.netloc.split(".")[0] if parsed_url.netloc else self.config.url
+        )
+        doc._set_aspect(
```
jjoyce0510, line 831: Interesting. I don't quite see the need to add this. 

Also this does NOT make it so that View in X works => That's a separate external URl field on teh document.

jjoyce0510: In general, having dpi is good to identify what platform the document comes from, tho . I thought it was already added to confluence docs tho.

metadata-ingestion/src/datahub/ingestion/source/confluence/confluence_source.py
```diff
@@ -943,26 +1041,7 @@ def _create_document_entity(
...
         )
 
-        # Add dataPlatformInstance aspect for UI features (e.g., "View in Confluence" button)
-        # TODO: This should be added to the Document SDK as an official parameter
-        # instead of using the backdoor _set_aspect() method
-        if self.config.url:
-            # Extract instance identifier from URL (e.g., "datahub-integration-testing" from the domain)
-            from urllib.parse import urlparse
-
-            parsed_url = urlparse(self.config.url)
```
jjoyce0510, line 953: nvm ok now i get it . refactored from here

jjoyce0510 (approved):

PR #18047, opened 2026-06-25
feat(frontend): scaffold Metrics page ahead of backend data model CAT-2509

annadoesdesign:
Screenshot 2026-06-25 at 2 04 59 PM

Screenshot 2026-06-25 at 1 51 23 PM

Adds a new /metrics page accessible from the sidebar under Govern. The page uses a three-column layout (browse tree, main content, entity sidebar) that mirrors the Documents page conventions, with reusable sidebar row chrome (MetricsTreeItem) that matches DocumentTreeItem exactly — 38px rows, icon→caret swap on hover when hasChildren, bgSelectedSubtle + shadowFocusBrand on selection, brand-gradient title text. Selection is purely route-driven.
All content is intentionally empty-state because the Metric / SemanticModel entity types and the metrics tree GraphQL queries are still in design. Each surface has a comment pointing at the exact swap-in once the backend lands:
- right-rail placeholder → real EntityProfileSidebar (register a MetricEntity in buildEntityRegistryV2, mirroring GlossaryTermEntity)
- sidebar tree empty state → useGetMetricsTreeQuery-backed tree
- sidebar search bar → real metric search hook
- sidebar create button → create-metric mutation
- re-add the Platform / Semantic filter selects from the Figma


datahub-web-react/src/app/metrics/MetricsSidebar.tsx
```diff
@@ -0,0 +1,260 @@
...
+    const isOverviewRouteActive = !!matchPath(location.pathname, { path: PageRoutes.METRICS, exact: true });
+
+    return (
+        <SidebarContainer
+            $width={width}
+            $isCollapsed={isCollapsed}
+            $isShowNavBarRedesign={isShowNavBarRedesign}
+            data-testid="metrics-sidebar"
+        >
+            <HeaderControls $isCollapsed={isCollapsed}>
```
ani-malgari: i see that we have a number of `isCollapsed` checks. we can cut it down to top-level check by having 2 separate components: CollapsedSidebar and ExpandedSidebar.

ani-malgari (approved):
Just have 1 non-blocking comment, but looks great. Thanks Anna!

metadata-service/configuration/src/main/resources/application.yaml
```diff
@@ -1396,6 +1396,7 @@ featureFlags:
   fineGrainedLineageNotAllowedForPlatforms: ${FINE_GRAINED_LINEAGE_NOT_ALLOWED_FOR_PLATFORMS:} # Comma separated list of platforms for which schemaFields entity edges will not be allowed to be created. for example: "hdfs, s3"
   assetSummaryPageV1: ${ASSET_SUMMARY_PAGE_V1:true} # Enables displaying the asset summary page
   datasetSummaryPageV1: ${DATASET_SUMMARY_PAGE_V1:false} # Enables displaying the dataset summary page
+  showMetricsPage: ${SHOW_METRICS_PAGE:false} # If turned on, exposes the Metrics page in the nav and enables the /metrics route. Backend Metric entity model is in development (see PR #18047); flag stays off until the entity ships.
```
ani-malgari: can we just rename it to `metricsEnabled` so we can use a single flag for all the work? showMetricsPage sounds very specific

PR #18048, opened 2026-06-25
feat(auth): add proxy header authentication support

### LINKED ISSUE #10702: Custom Authenticator using the principal from the header
**Background**
At Snap, we use proxy services that does the authentication and sends the principal in the header. This issue is to support a new authenticator to check the presence of the header with key from the ENV variable.

**Contribution**
I'm willing to contribute for this authenticator.

twthorn:
Fixes https://github.com/datahub-project/datahub/issues/10702

Add auth using headers from trusted reverse proxy (e.g., nginx, Traefik, envoy, etc.). We (Slack) also have this use case along with the reporter (Snap)

The CI failure seems flaky, it reports "Playwright E2E Tests" failing but that check succeeds for all shards when I examine the other checks. Perhaps it needs a rerun


twthorn:
@david-leifker I see you active in this repo, could you review this?

twthorn:
Perhaps @maggiehays could you take a look? Or know someone who could? Thanks!

(The CI failure appear unrelated)

twthorn:
@david-leifker ping on this. For conext, we (Slack) use this feature internally, and the issue reporter (Snap) also requests it so there is industry demand across companies for it. Let me know if I can provide any more context. Thank you for the help.

PR #18052, opened 2026-06-26
feat(ingest/teradata): categorize permission errors during profiling table sizing

NehaGslab:
## Summary

When Teradata profiling is enabled, the connector pre-filters oversized tables by
querying `DBC.TableSizeV` (gated by `profile_table_size_limit`). On locked-down
instances the ingestion user often lacks `SELECT` on that view. Previously any
failure here was reported under a single generic "Could not size tables for
profiling" warning, so a missing grant was indistinguishable from a transient
error and was not surfaced as an actionable permission problem.

This change categorizes the sizing failure: permission/access-denied errors now
emit a distinct, actionable warning and are counted separately in the report,
while still failing open (profiling proceeds without size-based filtering) so the
run never aborts.

Note: per-table profile statistics already catch permission errors and continue
(handled by the shared GE profiler with `catch_exceptions=True`); this PR only
improves reporting for the Teradata-owned sizing query and does not change that
behavior.

## Changes

- `TeradataReport`: add `profiling_permission_errors` counter and a thread-safe
  `increment_profiling_permission_error()` helper.
- `generate_profile_candidates`: on a `DBC.TableSizeV` sizing failure, classify
  the exception via the existing `_categorize_view_error(...) is ViewErrorCategory.PERMISSION`
  (numeric Teradata error code first, e.g. 3523/3524, then message substring as a
  fallback):
  - Permission error → increment the counter and emit "Unauthorized to size
    tables for profiling" (points the operator at granting `SELECT` on
    `DBC.TableSizeV`).
  - Otherwise → keep the existing generic "Could not size tables for profiling"
    warning.
  - Both paths still return `None` (fail open).
- Reused the existing categorizer rather than adding a new helper, keeping a
  single source of truth for error classification.

## Testing

- `tests/unit/test_teradata_source.py`:
  - `test_permission_failure_is_categorized_and_counted` — a `[Error 3523]`
    sizing failure increments `profiling_permission_errors` and emits the
    "Unauthorized to size tables for profiling" warning.
  - `test_non_permission_failure_uses_generic_warning` — a non-permission failure
    (e.g. timeout) keeps the generic warning and does not increment the counter.
  - Existing `generate_profile_candidates` and `_categorize_view_error` suites
    still pass.
- Ran `ruff format` / `ruff check` and the targeted unit tests locally.

## Checklist


treff7es (approved):
Reviewed with a full connector review pass (error-handling, tests, architecture, and code-quality agents). This is a clean, well-contained change and I'm approving it.

**What works well**
- Fails open correctly — profiling proceeds without size filtering rather than aborting the run, and every path emits an actionable `report.warning(...)` with `exc=e` so tracebacks are preserved.
- Reuses the existing `_categorize_view_error` classifier instead of re-implementing Teradata error-code parsing.
- Thread-safe counter increment consistent with the rest of `TeradataReport`, and good "why" comments on the new field.
- Both new tests assert the fail-open invariant (`candidates is None`) plus the counter and warning title, so the branch selection is regression-guarded. The code-less permission path is already covered by `_categorize_view_error`'s own unit tests.

**Optional polish (non-blocking — take it or leave it)**
1. `_categorize_view_error` is named/documented for view processing but is now also called from `generate_profile_candidates`. It's functionally correct today (the classification logic is phase-agnostic), but a one-line comment at the call site — or a mechanical rename to a phase-neutral name like `_categorize_teradata_error` — would prevent a future view-specific tweak from silently affecting the profiling path.
2. The two `report.warning(...)` branches share `context=`/`exc=`; you could compute just `title`/`message` in the `if/else` and call `warning(...)` once. Equally readable either way.
3. A brief comment noting that only the PERMISSION category is intentionally tracked for profiling sizing (unlike the four-way `view_*_errors` breakdown) would mark the asymmetry as deliberate.

None of these block merge. Nice work.

NehaGslab:
Suggestions from comment are addressed.

PR #18067, opened 2026-06-28
docs(quickstart): restructure into a linear first-run guide

lakshay-nasa:
Restructures the OSS quickstart so the first-run path reads top-to-bottom without interruption, and moves advanced/operational material out of the way.

## Changes

- Tighten the first-run flow so nothing sits between running the command and seeing it work.
- Group lifecycle commands under a new "Managing Your Local Instance" section.
- Remove the Poetry install tab; pip and Homebrew are the documented paths.

Back up/Restore stay as plain headings so their anchors (used by the CLI and `how/restore-indices.md`) keep resolving. No content dropped without a home, breaking-change history links to [Updating DataHub](https://github.com/datahub-project/datahub/blob/master/docs/how/updating-datahub.md) instead of duplicating it.

## Checklist


docs/quickstart.md
```diff
@@ -132,33 +102,6 @@ or head to http://localhost:9002 (username: datahub, password: datahub) to play
 Need support? Get in touch on Slack: https://datahub.com/slack/
 ```
 
-:::note Breaking changes
```
chakru-r, line 135: This probably still needs a mention -- I agree with the overall change to the structure here, but this should be either at the end either embedded or section calling out previous users of datahub and link to this breaking change in its own doc or from release notes.

chakru-r (approved):
Approving but with one comment that needs addressing

docs/quickstart.md
lakshay-nasa: Thanks Chakru. I didn't drop them, both breaking changes are still there, just moved into 'Managing Your Local Instance' with a link to Updating DataHub. To make them easier to catch before someone runs quickstart, we can add a short pointer up top.

PR #18079, opened 2026-06-29
feat(ingest): SchemaResolver casing/membership API (review slice of #18004)

puneetagarwal-datahub:
> **Draft / review aid — not for independent merge.** Isolated view of the **SchemaResolver-layer** changes from #18004 (lineage URN casing reconciliation), split out for focused review per discussion. Contains **only** the resolver-layer net diff — no work-unit processor, model, or connector changes.

## What's here

- **`SchemaResolver`** — a **disk-backed** casing/membership index (`FileBackedDict`) with `add_known_urn` / `find_by_casing` / `known_urn_count`, kept **separate from `_schema_cache`** so its `None` semantics (relied on by SQL parsing) are unchanged; plus `get_cached_schema_info`, a side-effect-free cache read used for column-casing correction.
- **`DataHubGraph._bulk_fetch_schema_info_by_filter`** — additive `include_schemaless` (default `False` = unchanged) so a single scroll can also yield existing-but-schemaless datasets as `(urn, None)`.
- **`SchemaResolverProvider`** — additive `populate_membership` (default `False`) so one bulk scroll fills both the schema cache and the casing index (no duplicate catalog scroll).

## Why it's safe

Everything is **additive and opt-in** — the new params default to today's behavior, so existing SQL-parsing consumers (bigquery, teradata, sql-parsing-aggregator, kafka-connect) are unaffected. The full `tests/unit/sql_parsing` suite passes; new tests cover the casing index (`test_schemaresolver.py`) and the `populate_membership` path (`test_schema_resolver_provider.py`).

## Deferred (not in this slice)

The sentinel-in-`_schema_cache` distinction (exists-no-schema vs not-found via the OpenAPI-v2 entity endpoint) and a configurable bulk-vs-lazy initialization remain follow-ups.

## Relationship to #18004

These changes already live in #18004; this branch is a **review aid** that shows just the resolver layer (and a head start should we split it into its own PR).

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/graph/client.py
+++ b/metadata-ingestion/src/datahub/ingestion/graph/client.py
@@ -953,10 +953,15 @@ class DataHubGraph(DatahubRestEmitter, OpenApiAPI, EntityVersioningAPI):
         status: RemovedStatusFilter = RemovedStatusFilter.NOT_SOFT_DELETED,
         batch_size: int = 100,
         extraFilters: Optional[List[RawSearchFilterRule]] = None,
-    ) -> Iterable[Tuple[str, "GraphQLSchemaMetadata"]]:
+        include_schemaless: bool = False,
+    ) -> Iterable[Tuple[str, Optional["GraphQLSchemaMetadata"]]]:
         """Fetch schema info for datasets that match all of the given filters.
 
-        :return: An iterable of (urn, schema info) tuple that match the filters.
+        :param include_schemaless: when True, also yield datasets that exist but have
+            no schema, as ``(urn, None)``. The default (False) preserves the original
+            behavior of yielding only schema-bearing datasets, so existing callers are
+            unaffected. Lets one scroll serve both schema and membership needs.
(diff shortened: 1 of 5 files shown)


puneetagarwal-datahub:
Closing — superseded by the decision to defer the SchemaResolver framework change.

Per review, the 'existing datasets with missing schema' case is an edge case (rare on Snowflake/BigQuery) and this framework is high-blast-radius, so we're not changing `SchemaResolver` now. In #18004 the casing feature resolves against the entities `SchemaResolver` already provides (schema-bearing, via `get_urns()`) with **zero** framework changes; existing-but-schemaless references are reported `UNRESOLVED`.

The design captured here (disk-backed casing index, `add_known_urn`/`find_by_casing`, `include_schemaless`/`populate_membership`, single scroll) is the basis for the **follow-up task**: enrich `SchemaResolver` to know membership — e.g. `resolve_table_and_schema(...) -> (urn, exists, schema)`, deprecating the ambiguous `resolve_table` — which then unblocks schemaless coverage, a disk-backed index, and bulk-vs-lazy init.

PR #18081, opened 2026-06-29
fix(ingest/snaplogic): Checkpoint latest received eventTime so records between runs are not skipped.

SalimAbdul-snaplogic:
### Problem

The SnapLogic lineage source uses the redundant-run-skip handler to checkpoint a time window between ingestion runs. Previously, `update_stats()` committed the **configured `end_time`** as the checkpoint's end boundary, regardless of which records were actually received.

Because SnapLogic events can lag (an event with an `eventTime` inside the window may not be returned by the API until after the window has been checkpointed), the next run would start from `end_time` and **silently skip any records that arrived late** — resulting in missing lineage.

### Fix

Track a high-water mark (`latest_event_time`) of the maximum `eventTime` actually received during a run, and checkpoint **that** value instead of the configured `end_time`.

- Added `_track_event_time()`, called per record as it's yielded, which parses `eventTime` via `parse_absolute_time` and advances the watermark.
- `update_stats()` now commits `latest_event_time` (falling back to `start_time` when no records were received, so the window does not advance past unprocessed time).
- Unparseable `eventTime` values are skipped with a source-report warning rather than crashing the run.
### Tests

Added unit tests in `test_snaplogic_lineage_extractor.py`:

- `test_update_stats_commits_latest_received_event_time` — checkpoint end = max received eventTime
- `test_update_stats_holds_window_when_no_records_received` — window held at `start_time`
- `test_track_event_time_ignores_unparseable_event_time` — bad timestamps are ignored

### Checklist

- [x] Tests added/updated


SalimAbdul-snaplogic:
Hi @maggiehays , could you please help get this PR reviewed?

lakshay-nasa:
Hi @SalimAbdul-snaplogic, sorry this sat so long, and that your August ping went unanswered! Thanks for the fix. Checkpointing on the latest received eventTime makes a lot of sense.  I've prioritized it and assigned it to a reviewer on our ingestion team, so it should get eyes soon.

PR #18088, opened 2026-06-29
feat(domains): Bring Manage Domains sidebar to parity with documents …

annadoesdesign:
Before:
Screenshot 2026-06-29 at 1 15 18 PM

After:

Screenshot 2026-06-29 at 1 04 35 PM
Screenshot 2026-06-29 at 12 58 51 PM
Screenshot 2026-06-29 at 12 58 47 PM
Screenshot 2026-06-29 at 12 58 35 PM

Restyles the Manage Domains sidebar to mirror the documents sidebar (38px rows, brand-gradient selected state, Phosphor icons, section header pattern) and introduces an owner filter sourced from the loaded domain tree.

Sidebar layout:
- Add a "Domains Home" overview row that links to /domains, visible in both expanded and collapsed sidebar states.
- Add a collapsible "All Domains" section header above the tree, matching the docs sidebar's per-platform headers.
- Move the divider above the overview row so the filter pills and the navigator share the same separator placement.
- Center icons in collapsed mode; clicking a collapsed row now navigates straight to that domain's page instead of just re-expanding the sidebar.
- Hide the child-count pill when the node or the entire sidebar is collapsed.

Owner filter:
- New DomainSidebarFiltersContext stores selected owner URNs and a deduped registry of available owners (useRef-based dedupe so registration from every node is O(1) in the steady state).
- New domainSidebarFilters.utils.ts exposes pure helpers: extractDomainOwners, extractOwnersFromDomains, matchesDomainOwnerFilter, filterDomainsByOwner. Adjacent vitest coverage for all four (19 cases) plus provider/hook tests for the context (7 cases).
- Filtering applies per level, matching how the documents sidebar filters by author.

Ant Design cleanup:
- DomainSearch swaps SearchOutlined for Phosphor MagnifyingGlass and LoadingOutlined for Alchemy Loader.
- ManageDomainsSidebar's collapse/expand button uses the semantic 'icon' color token.


datahub-web-react/src/app/domainV2/nestedDomains/domainSidebarFilters/domainSidebarFilters.utils.ts
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * Filter predicate: does `domain` match the sidebar's active owner selection?
+ *
+ *   - An empty / null `selectedOwnerUrns` means "no filter active" — every
+ *     domain passes.
+ *   - A non-empty selection means the domain must have at least one owner
+ *     whose URN is in the selection (OR semantics, matching how the
+ *     documents sidebar's author filter works).
+ */
+export function matchesDomainOwnerFilter(domain: OwnedDomainLike, selectedOwnerUrns: string[] | null): boolean {
```
jjoyce0510: The way we are filtering by owners is JUST on the first level I believe or the domains that we've loaded, not on any sub-domains? 

My expectation as a user is that if I select an owner, I'll see any parent or CHILD domains that have that owner. Currently that is not the case.

jjoyce0510: We can also delay this til later.

jjoyce0510: The safer thing to do would be to send a filter down to the backend.

datahub-web-react/src/app/domainV2/nestedDomains/domainSidebarFilters/DomainSidebarFiltersContext.tsx
```diff
@@ -0,0 +1,81 @@
...
+    //   - `availableOwners` is the rendered, ordered list — only updated
+    //     when at least one URN in the incoming batch is new.
+    // Doing it this way keeps the registration call O(1) per owner in the
+    // common case (re-renders that don't add new owners), which matters
+    // because `registerOwners` is called from every DomainNode on every
+    // render-phase effect.
+    const seenUrnsRef = useRef<Set<string>>(new Set());
+    const [availableOwners, setAvailableOwners] = useState<DomainOwnerInfo[]>([]);
+
+    const registerOwners = useCallback((owners: DomainOwnerInfo[]) => {
```
jjoyce0510: We can actually get these from the aggregations that come back in our list or search call at the top level.

jjoyce0510: INSTEAD OF CLIENT SIDE COLLECTION OF THE OWNERS FOROM EACH INDIVIDUAL LOADED DOMAIN!

jjoyce0510 (approved):

PR #18090, opened 2026-06-29
refactor(frontend/glossary): align sidebar, modals, and parent selector with domains

annadoesdesign:
Before: 
Screenshot 2026-06-30 at 2 20 20 PM
Screenshot 2026-06-30 at 2 20 22 PM
Screenshot 2026-06-30 at 2 20 25 PM

After:
Screenshot 2026-06-30 at 2 19 22 PM
Screenshot 2026-06-30 at 2 19 35 PM
Screenshot 2026-06-30 at 2 19 33 PM
Screenshot 2026-06-30 at 2 19 30 PM
Screenshot 2026-06-30 at 2 19 26 PM
Screenshot 2026-06-30 at 2 19 23 PM

## Summary

Frontend-only refactor that aligns the glossary surface with the
domains/documents surface — collapsible sidebar, Alchemy-based create/edit
modals, a tree-style parent selector — and pulls a handful of related
correctness and perf fixes along the way. No backend or GraphQL changes.

Split out of #17838 (the color-picker work continues in a stacked PR on top of
this branch); this PR is the structural cleanup that makes the color story
clean to land.

### User-facing

- **Collapsible glossary sidebar** mirroring the domains sidebar — animated
  width toggle, grouped header actions, root-level icons visible in the
  collapsed state. Also fixes a pre-existing horizontal scrollbar bug where
  `StyledDivider`'s width/margin combo overshot the sidebar by 13px on the
  right.
- **Create Glossary modal modernization (V1 + V2).** Migrates to Alchemy
  (`Modal`, `Input`, `Editor`, `Text`, `toast`), embeds the description editor
  inline (no more modal-in-modal), makes the parent picker optional, and
  inserts the optimistic sidebar entry immediately on mutation resolve
  instead of after a 2-second timeout.
- **New `EditGlossaryEntityModal`** (single Alchemy modal) replacing the
  prior pair of partial-edit dialogs, with shared name-validation and the
  same inline editor pattern.
- **Modern parent selector.** `NodeParentSelect` is rewritten on Alchemy
  `SimpleSelect` with tree-style browse + search (powered by
  `useGlossaryTreeEntities`); V1 re-exports V2 so there's one source of truth.

### Frontend (technical)

- Shared modal helpers in `glossaryEntityModal.shared.ts` (`Field`,
  `EditorContainer`, `useGlossaryNameValidation`) used by all three modals.
- Shared tree row chrome in `treeRow.styles.ts` (`RowContainer`,
  `LeftContent`, `IconSlot`, `Title`) used by `NodeItem` + `TermItem`.
- `buildOptimisticGlossaryEntity` extracted into `glossaryV2/utils.ts` so V1
  and V2 modals share one optimistic-entry factory (V1 previously skipped
  root creates). Plus `GLOSSARY_SEARCH_INDEX_REFRESH_MS` constant so both
  modals stay in sync if the refresh window changes.
- `GlossarySidebar` now passes `nodesData`/`termsData` down to
  `GlossaryBrowser` so the browser skips its own copy of the two root
  queries; sort results are memoized so `NodeItem`/`TermItem` props don't
  change identity on every render.
- Migrate touched files from antd to Alchemy (`toast`, `Modal`, `Input`,
  `Editor`, `Text`) and Phosphor icons (`PencilSimple`, `CaretDown/Right`,
  `ArrowLineLeft/Right`, `BookmarksSimple`, etc.).

### Correctness fixes

- **`useGlossaryChildren` race condition.** Switch to functional setters and
  add the missing dep so a queued multi-URN update can't strip the wrong
  subset of `urnsToUpdate`. Widens `GlossaryEntityContext.setUrnsToUpdate` to
  the full `Dispatch<SetStateAction<…>>` shape so the functional form
  type-checks. Test added in `useGlossaryChildren.test.tsx`.
- **`EntityPage` hooks-order violation.** Hoist `renderProfile` outside the
  conditional branch (V1 + V2).
- **`isCloning` prefill effect.** Now runs once via a `useRef` so a later
  `entityData` cache refresh can't overwrite what the user has typed.

## Tests

- `useGlossaryChildren.test.tsx` (4 tests) — covers the functional-setter
  race fix.
- `utils.test.ts` (13 tests) — covers `buildOptimisticGlossaryEntity` and
  `getCollapsedGlossaryItems`.
- `NodeParentSelect.test.tsx` (4 tests) — covers the move-mode filtering and
  tree-options helpers.
- `yarn type-check`, `yarn eslint --quiet`, and the touched vitest suites
  all pass.

## Checklist

- [x] PR conforms to the [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (incl. PR Title Format)
- [x] Tests added/updated
- [ ] Docs added/updated (no user-docs changes needed for this UI work)
- [ ] Breaking changes documented (none — pure refactor)


datahub-web-react/src/i18n/locales/en/governance.glossary.json
```diff
@@ -13,5 +13,9 @@
     "page.subtitle": "Classify your data assets and columns using data dictionaries",
     "page.title": "Business Glossary",
     "section.glossaryTerms": "Glossary Terms",
-    "section.termGroups": "Term Groups"
+    "section.termGroups": "Term Groups",
+    "sidebar.collapse": "Collapse Glossary",
+    "sidebar.expand": "Expand Glossary",
+    "sidebar.home": "Glossary Home",
+    "sidebar.section.allTerms": "All Glossary Terms"
```
benjiaming, line 20: We will need translations of new strings into all supported languages.

datahub-web-react/src/app/glossaryV2/BusinessGlossaryPage.tsx
```diff
@@ -122,7 +122,6 @@ const BusinessGlossaryPage = () => {
                     canCreateGlossaryEntity={!!canManageGlossaries}
                     onClose={() => setIsCreateNodeModalVisible(false)}
                     refetchData={refetchForNodes}
-                    canSelectParentUrn={false}
```
benjiaming, line 125: I would be careful here. Previously we suppressed the picker but the modal now unconditionally renders it. That means if you click "Add term group" you would see parent-node picker that was previously hidden.

benjiaming (changes requested):

datahub-web-react/src/app/glossaryV2/BusinessGlossaryPage.tsx
annadoesdesign: yes thats correct i do want that behavior i did this on purpose

benjiaming (approved):

PR #18096, opened 2026-06-30
feat(teradata): report missing tables from cache during column extraction

NehaGslab:
## Summary

When the Teradata connector extracts columns for a table, it looks the table up in the in-memory cache populated earlier by `cache_tables_and_views()`. If the table is no longer in the cache — e.g. it was dropped between table discovery and column extraction — the connector previously emitted only a bare `logger.warning` and returned no columns. This warning never surfaced in the structured ingestion report and carried no count, so an operator had no visibility into how many (or which) datasets were affected.

This PR makes that situation observable in the run report. Behavior is unchanged: the dataset is still emitted (without columns) exactly as before — only the reporting is improved.

## Changes

- Added a `num_tables_missing_from_cache` counter and `increment_tables_missing_from_cache()` to `TeradataReport`.
- Extracted a small `_report_table_missing_from_cache()` helper that increments the counter and emits a grouped report.warning (with the affected `schema.table` as context), falling back to the original `logger.warning` only when no report is attached. This reuses DataHub's existing structured-log mechanism, which deduplicates by title, aggregates the affected names, and tracks a total count — matching how Snowflake/BigQuery surface similar "listed-but-missing" data-inconsistency cases.
- Keeping the logic in a helper also keeps `optimized_get_columns` under the complexity lint threshold.

## Tests
- Added `test_optimized_get_columns_reports_table_missing_from_cache`, asserting the counter increments, the warning context includes the affected schema.table, and behavior is unchanged (returns [], no column query issued).

## Checklist


treff7es (approved):
Approving — small, well-scoped, and well-tested observability fix.

**What stands out:** beyond the stated goal (surfacing the previously-silent "table missing from cache" `logger.warning` as a counted, structured `report.warning`), the `setattr(TeradataDialect, "report", self.report)` wiring fixes a genuine latent bug: the dialect functions patched onto `TeradataDialect` read `self.report` via `hasattr(self, "report")`, but `report` was never attached — so those guards were always `False` and several metrics (column-extraction duration, primary-keys-processed, column-extraction failures) were silently staying at 0 in production. This one line un-breaks all of them, and the added comment documents why.

Tests are above the usual bar — they cover the report-present branch, the `logger.warning` fallback (with a nice explanation of why `SimpleNamespace` is used over `MagicMock`), the aggregate-don't-spam grouping behavior, and the production wiring line itself.

Two optional, non-blocking suggestions for a follow-up (not required for this PR):

1. `setattr(TeradataDialect, "report", ...)` attaches per-run state to a process-global class, so it's last-writer-wins if two Teradata sources ever run concurrently in one process. This is pre-existing (the 6 existing method patches share the identical hazard) and safe for the normal one-source-per-process model — worth a one-line note or follow-up ticket acknowledging the semantics.
2. The `delattr(TeradataDialect, "report")` cleanup lives in one test's `finally`; consider moving it into the autouse `isolate_teradata_caches` fixture teardown so the shared-class-state cleanup is handled suite-wide.

PR #18097, opened 2026-06-30
fix(ingest/powerbi): recover BigQuery ODBC lineage for views and native queries

acrylJonny:
## Summary

The PowerBI ODBC handler silently dropped two classes of BigQuery upstream lineage edges (both surfaced as `Can not determine qualified table name for ODBC data source`):

- **View leaves were ignored.** In `OdbcLineage.expression_lineage`, the navigation walk only recognized `Kind == "Table"`. PowerBI labels a view leaf with `Kind == "View"`, so `table_name` was never set and the edge was dropped. This now treats `View` the same as `Table`, matching `create_reference_table`, which already accepts either (`table_detail.get("Table") or table_detail["View"]`).
- **`is_sql_query` ignored the platform dialect.** It parsed with the default sqlglot dialect, which throws on BigQuery backtick-quoted, hyphenated project ids (e.g. `` `my-proj.dataset.table` ``). Valid native queries were misclassified as non-SQL and wrongly routed to the navigation path (which has no accessor), so they were dropped. It now parses using the source platform's dialect via `get_dialect(platform)`.

Verified against a real PowerBI tenant (file sink): BigQuery upstream edges went from 94 → 102 (7 views + 1 native query recovered) and `Can not determine qualified table name` failures went from 8 → 0, with no regressions to Hive/Databricks edges. Recovered URNs are well-formed BigQuery `project.dataset.table`.

## Checklist

- [x] Tests added (`test_odbc_view_leaf_resolves_like_table`, `test_is_sql_query_uses_platform_dialect`)
- [ ] Docs — n/a (bug fix, no user-facing config change)
- [ ] Breaking changes — none

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/pattern_handler.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/pattern_handler.py
@@ -56,6 +56,7 @@ from datahub.sql_parsing.sqlglot_lineage import (
     DownstreamColumnRef,
     SqlParsingResult,
 )
+from datahub.sql_parsing.sqlglot_utils import get_dialect
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


acrylJonny:
Superseded by a same-repo branch (not from a personal fork). Reopening from datahub-project/datahub.

PR #18098, opened 2026-06-30
fix(ingest/powerbi): recover BigQuery ODBC lineage for views and native queries

acrylJonny:
## Summary

The PowerBI ODBC handler silently dropped two classes of upstream lineage edges, both surfaced as `Can not determine qualified table name for ODBC data source`:

- **View leaves were ignored.** In `OdbcLineage.expression_lineage`, the navigation walk only recognized `Kind == "Table"`. PowerBI labels a view leaf with `Kind == "View"`, so `table_name` was never set and the edge was dropped. This now treats `View` the same as `Table`, matching `create_reference_table`, which already accepts either (`table_detail.get("Table") or table_detail["View"]`). For a three-tier platform like BigQuery this yields `project.dataset.view`.

- **`is_sql_query` ignored the platform dialect.** It parsed with the default sqlglot dialect, which throws on platform-specific syntax such as BigQuery backtick-quoted, hyphenated project ids (e.g. `` `my-proj.dataset.table` ``). Valid native queries were misclassified as non-SQL and wrongly routed to the navigation path (which has no accessor), so they were dropped. It now parses using the source platform's dialect via `get_dialect(platform)`, and falls back to the default dialect for platforms sqlglot has no dialect for (e.g. db2, vertica) so nothing else regresses.

## Tests

Unit tests in `tests/unit/powerbi/test_pattern_handler.py`:

- `test_odbc_view_leaf_resolves_like_table` — a `View` leaf resolves to a dataset URN.
- `test_odbc_three_tier_platform_keeps_database` / `test_odbc_two_tier_platform_drops_pseudo_catalog` — arity is preserved per platform.
- `test_is_sql_query_uses_platform_dialect` — dialect-specific SQL (BigQuery backtick ids) is recognized.
- `test_is_sql_query_handles_other_platforms_without_raising` — unknown platforms fall back to the default dialect rather than raising.

## Checklist

- [x] Tests added/updated
- [ ] Docs — n/a (bug fix, no user-facing config change)
- [ ] Breaking changes — none

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/pattern_handler.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/pattern_handler.py
@@ -373,8 +373,15 @@ class AbstractLineage(ABC):
             return False
         query = native_sql_parser.remove_special_characters(query)
         # Use the platform dialect so platform-specific syntax (e.g. BigQuery
-        # backtick-quoted, hyphenated project ids) parses as SQL.
-        dialect = get_dialect(platform) if platform else None
+        # backtick-quoted, hyphenated project ids) parses as SQL. Platforms
+        # sqlglot has no dialect for (e.g. db2, vertica) fall back to the
+        # default dialect rather than raising.
+        dialect: Optional[sqlglot.Dialect] = None
+        if platform:
+            try:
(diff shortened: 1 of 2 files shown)


aviraj-gour (approved):
LGTM!!

PR #18099, opened 2026-06-30
fix(ingest): allow full URNs and global tags for tags_to_ignore_sampling

kevinkarchacryl:
There are two issues with `tags_to_ignore_sampling`

The first is that if your tag is urn:li:tag:my_tag then you have to put my_tag in the recipe. I changed this so using the URN prefix will work as well. The documentation is also updated to make this clear.

The larger fix is that `tags_to_ignore_sampling` only used editableSchemaMetadata which does not account for tags that were ingested via a source. Tags for those columns are stored in the schemaMetadata.


metadata-ingestion/src/datahub/ingestion/source/ge_data_profiler.py
```diff
@@ -1658,42 +1662,70 @@ def create_bigquery_temp_table(
         raw_connection.close()
 
 
+_TAG_URN_PREFIX = "urn:li:tag:"
+
+
+def _tag_name(tag_urn: str) -> str:
+    """Return the name component of a tag URN (the part after urn:li:tag:)."""
+    return tag_urn.split(_TAG_URN_PREFIX)[1]
```
treff7es: `_tag_name` and `_normalize_tag` are two non-equivalent hand-rolled implementations of "strip the `urn:li:tag:` prefix". The codebase already has the canonical extractor — `TagUrn(x).name` (from `datahub.metadata.urns`), used this same way in the Snowflake, BigQuery, Unity Catalog and AWS tag code.

`TagUrn` accepts both a full URN and a bare name, and I verified it preserves the colon-in-name cases these tests cover:
```
TagUrn.from_string("urn:li:tag:org_common.tag:sensitive").name  -> "org_common.tag:sensitive"
TagUrn("org_common.tag:sensitive").name                         -> "org_common.tag:sensitive"
```
Switching to it lets you delete both helpers and the `_TAG_URN_PREFIX` constant. It also removes two smaller issues on this line: `split(prefix)[1]` mis-parses a name that itself contains `urn:li:tag:`, and it raises `IndexError` on a non-URN input (`_normalize_tag`'s slice doesn't). That `IndexError` is only theoretical today — `TagAssociationClass.tag` is always a full URN — so I'm not treating it as a blocker, just noting `TagUrn` resolves it for free.

One deliberate call to make: `TagUrn(...)` raises `InvalidUrnError` on a malformed recipe entry, whereas the current code passes it through silently. Failing loudly on bad `tags_to_ignore_sampling` input is arguably the better behavior, but it is a behavior change.

metadata-ingestion/src/datahub/ingestion/source/ge_data_profiler.py
```diff
@@ -1658,42 +1662,70 @@ def create_bigquery_temp_table(
...
+def _normalize_tag(tag: str) -> str:
+    """Accept either a full tag URN or just the name portion."""
+    if tag.startswith(_TAG_URN_PREFIX):
+        return tag[len(_TAG_URN_PREFIX) :]
+    return tag
+
+
+def _matching_field_paths(
+    fields: Sequence[Union[SchemaFieldClass, EditableSchemaFieldInfo]],
+    tags: set,
```
treff7es: Bare `set` — prefer a parameterized type per the type-safety convention. Callers pass a `Set[str]` (`tags_set`), so `tags: Set[str]` fits (or `AbstractSet[str]`, since only membership is used). `Set` isn't in the `typing` import block yet, so it'll need adding.

metadata-ingestion/src/datahub/ingestion/source/ge_data_profiler.py
```diff
@@ -1658,42 +1662,70 @@ def create_bigquery_temp_table(
...
+
+    # Collect from both aspects; use a set to deduplicate across them.
+    # SchemaMetadata holds ingestion-sourced column tags (e.g. from Snowflake).
+    # EditableSchemaMetadata holds tags applied via the DataHub UI.
+    columns_to_ignore: set[str] = set()
 
-    if not ignore_table:
-        metadata = datahub_graph.get_aspect(
-            entity_urn=dataset_urn, aspect_type=EditableSchemaMetadata
+    schema_metadata = datahub_graph.get_aspect(
```
treff7es, line 1705: Not a blocker for this PR — this `SchemaMetadata` fetch is genuinely needed (these are the committed DataHub tags, not the SQL-inspector schema the profiler already holds).

Correcting an earlier version of this comment: the useful optimization here is the per-table *call count*, not hoisting anything out of the loop — the lookup is inherently per-table (each dataset has its own tags/schema in the graph), so it can't be "resolved once per run". What can be improved: `get_tags` + `get_aspect(SchemaMetadata)` + `get_aspect(EditableSchemaMetadata)` are three aspects of the *same* URN fetched in three sequential round-trips. `get_aspects_for_entity(urn, [globalTags, schemaMetadata, editableSchemaMetadata])` / `get_entity_semityped(urn)` fetch them in a single `get_entity_raw` call, turning ~3N REST calls into ~1N. Fine to defer to a separate ticket.

metadata-ingestion/tests/unit/profiling/test_columns_to_ignore_sampling.py
```diff
@@ -0,0 +1,242 @@
...
+        DATASET_NAME, TAGS_TO_IGNORE, PLATFORM, ENV
+    )
+
+    assert columns.count("email") == 1
+
+
+@patch("datahub.ingestion.source.ge_data_profiler.get_default_graph")
+def test_dataset_level_tag_ignores_table(mock_get_graph):
+    """A matching tag on the dataset entity sets ignore_table=True."""
+    dataset_tags = MagicMock()
```
treff7es: Minor: this builds `dataset_tags` as a bare `MagicMock` with a `.tags` attribute, while the field-level helpers use the real `GlobalTagsClass`. Using the real class here too (e.g. `GlobalTagsClass(tags=[TagAssociationClass(tag=PII_TAG_URN)])`) keeps it consistent and would catch signature drift in what `get_tags()` returns.

treff7es (commented):
Nice fix — reading both `schemaMetadata` (source-ingested tags) and `editableSchemaMetadata` closes a real gap, and the set-based collection also removes the old double-count when a column carried two matching tags. A few suggestions inline; none of them block the correctness of the fix.

treff7es (approved):

metadata-ingestion/src/datahub/ingestion/source/ge_data_profiler.py
kevinkarchacryl: Thanks, I didn't know about that function

treff7es (approved):

kevinkarchacryl:
@treff7es I forgot that I needed to change the SQL alchemy profiler as well. After the latest changes, this worked in the actual instance with tags ingested from Snowflake.

PR #18100, opened 2026-06-30
feat(teradata): add warning for immutable `convert_urns_to_lowercase` setting to prevent duplicate entities

NehaGslab:
## Summary

`convert_urns_to_lowercase` is baked into every Teradata dataset URN, so changing
it between runs re-keys every table (e.g. `MyDb.MyTable` → `mydb.mytable`). With
stateful ingestion enabled, the old-cased URNs are then soft-deleted as stale
while the new-cased ones are created — producing duplicate or orphaned entities.

This PR adds a lightweight guardrail, following the same pattern Snowflake already
uses for this flag (`validate_convert_urns_to_lowercase` →
`add_global_warning`):

- **Config validator**: `TeradataConfig` now emits a global warning when
  `convert_urns_to_lowercase` is enabled, reminding operators to keep the value
  fixed for the life of the deployment. The warning fires only when the flag is
  set (it defaults to `False`), so default runs stay quiet.
- **Docs**: a new entry under the Teradata connector's *Limitations* section
  explains the URN-identity impact and the duplicate/orphan failure mode.
- **Tests**: two focused unit tests covering "default does not warn" and
  "enabled warns".

The default is intentionally left unchanged (`False`). Flipping it to match
Snowflake/BigQuery would itself be a URN-changing breaking change for existing
Teradata deployments, which is exactly the failure mode this PR warns about.

## Checklist


sgomezvillamor (changes requested):
`convert_urns_to_lowercase` is defined in `LowerCaseDatasetUrnConfigMixin` and `TeradataConfig` gets it from some class in the config hierarchy

so:
* This doc should not be in teradata docs, instead we may properly document it in the description of the field in `LowerCaseDatasetUrnConfigMixin` itself, so this is visible to all sources
* the same for the field validator, it should be defined closer to the `LowerCaseDatasetUrnConfigMixin`, not only for teradata

having said that: if the flag is properly documented, is the warning needed? I guess anyone adding/removing the flag would check the docs, if we warn about impact in the docs, do we need to warn at runtime too?

sgomezvillamor (approved):

PR #18102, opened 2026-06-30
fix(ingestion): prevent oversized query metadata from failing runs

alfiyas-datahub:
## Summary

Snowflake (and other SQL connectors) ingestion runs were failing with repeated GMS `400 Cannot parse request entity` errors on `queryProperties` aspects, surfacing in the UI only as "An unexpected issue occurred". The `EnsureAspectSizeProcessor` guard that is supposed to prevent this had a **unit-mismatch bug**: it computed the required reduction in serialized JSON bytes but compared it against the raw character count of the statement (`len(statement.value)`). JSON escaping (`\n`, `\"`, control chars, non-ASCII) inflates the serialized size past the raw length, so the check failed, the guard logged "Cannot truncate..." (only via `logger.warning`, never in the run report) and emitted the oversized aspect anyway. GMS rejected it; after enough rejections the run exited 1 / FAILED.

This is exercised most easily by pipelines that write to a temp table thousands of times in one session (e.g. Hightouch on Snowflake with `include_queries` enabled), where `SqlParsingAggregator` merges the whole chain into one synthetic `composite_*` query whose statement can balloon to hundreds of MB — but the same broken guard also lets any non-composite oversized `queryProperties` through, so the fix is not composite-specific.

## Approach

Keep the trimming responsibility in a single place — `EnsureAspectSizeProcessor`, which already owns aspect-size handling — so composite and non-composite queries go through the same logic. `sql_parsing_aggregator.py` is intentionally unchanged: no cap, no env var, no per-aggregator report field.

## Changes

- **`ensure_aspect_size.py`**:
  - Rewrite `ensure_query_properties_size` to binary-search the largest statement prefix that fits under the JSON-serialized budget (escape-inflation safe). Compute the aspect overhead once with an empty statement so each search step only re-encodes the candidate string, not the whole aspect.
  - Fall back to dropping `name`/`description` if non-statement fields alone exceed the limit.
  - **Re-verify after dropping fields:** if some other field (e.g. a large `customProperties`) still keeps the aspect oversized, emit a distinct "could not be truncated" warning and do **not** record a successful truncation — so a subsequent GMS 400 stays visible, not silent.
  - Extract the binary search into a reusable `_largest_fitting_prefix` helper.
  - Refactor `ensure_view_properties_size` to use `_largest_fitting_prefix` and measure serialized size the same way — the pre-existing subtractive `reduction_needed` code had the exact same unit-mismatch bug lurking for heavily-escaped view SQL.
  - Always emit a structured `source_report.warning` and record the truncation in `EnsureAspectSizeProcessorReport` (previously only `logger.warning`, which never reached the run report).
- **`test_ensure_aspect_size.py`**: add boundary/tightness/escape tests for `_largest_fitting_prefix`, a tightness assertion for the query-properties escape regression, an escape regression for `viewLogic`, and a test for the untruncatable-after-dropping-fields warning path.

## Testing

- New + existing unit tests (`tests/unit/api/source_helpers/test_ensure_aspect_size.py`): **48 passed**.
- Coverage on `ensure_aspect_size.py`: **94%** (missing lines are all pre-existing early-return guards, none in new code).


metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
```diff
@@ -478,33 +483,69 @@ def ensure_query_properties_size(
...
+                lo = mid + 1
+            else:
+                hi = mid - 1
+
+        query_properties.statement.value = candidate_value(retained_length)
+
+        # If even an empty statement can't make it fit, the other fields are over
+        # budget on their own; drop them as a last resort.
+        dropped_fields: List[str] = []
+        if self._query_properties_serialized_size(query_properties) >= max_payload_size:
```
treff7es, line 518: The last-resort field drop never re-verifies the result. If some field *other* than `name`/`description` dominates the budget (e.g. a large `customProperties`), dropping those two won't make the aspect fit — but the code below still unconditionally emits the "truncated due to size constraint" warning and calls `_record_truncation("queryProperties")`. So the run report would claim a successful truncation while GMS still rejects the aspect with a 400 — a silent-failure-shaped hole in the very guard this PR is hardening.

Rare in practice (statement/name/description normally dominate), but cheap to close: re-check `_query_properties_serialized_size(...)` after dropping and, if it's still over the limit, emit a distinct "could not truncate" warning instead of recording a successful truncation.

metadata-ingestion/src/datahub/sql_parsing/sql_parsing_aggregator.py
```diff
@@ -95,6 +96,13 @@ class QueryLogSetting(enum.Enum):
 MAX_UPSTREAM_TABLES_COUNT = 300
 MAX_FINEGRAINEDLINEAGE_COUNT = 2000
 
+# A composite query concatenates every statement in a temp-table chain with no
+# size bound; a session that writes to a temp table many times can grow the
+# merged text to hundreds of MB and overflow the GMS payload limit. Cap it here.
+MAX_COMPOSITE_QUERY_STATEMENT_CHARS = int(
```
treff7es: Two things about this constant:

1. **Diverges from the pattern it claims to follow.** The PR body says it follows the `MAX_UPSTREAM_TABLES_COUNT` pattern, but that sibling (and `MAX_FINEGRAINEDLINEAGE_COUNT`) is a plain hardcoded constant with no env override. This is the only `os.environ.get(...)` in the file — env-driven config elsewhere goes through `datahub/configuration/env_vars.py` accessors (e.g. `get_sql_agg_skip_joins()`).
2. **The env-var path is effectively untested.** Because it's read at *import time* into a module constant, the new test has to `monkeypatch.setattr(agg_module, "MAX_COMPOSITE_QUERY_STATEMENT_CHARS", cap)` rather than set the env var — so the `os.environ`/`int()` path never runs in tests (and would raise `ValueError` at import on a malformed value).

Suggestion: either drop the override and make it a plain constant like its siblings (simplest, and this cap is unlikely to need per-deployment tuning), or add a `get_max_composite_query_statement_chars()` accessor in `env_vars.py` and read it lazily.

metadata-ingestion/tests/unit/api/source_helpers/test_ensure_aspect_size.py
```diff
@@ -1114,6 +1115,67 @@ def test_ensure_size_of_too_big_query_properties(processor):
...
+        len(json.dumps(query_properties.to_obj()))
+        > QUERY_PROPERTIES_STATEMENT_MAX_PAYLOAD_BYTES
+    )
+
+    processor.ensure_query_properties_size(
+        "urn:li:query:heavy_escape", query_properties
+    )
+
+    final_size = len(json.dumps(query_properties.to_obj()))
+    assert final_size < QUERY_PROPERTIES_STATEMENT_MAX_PAYLOAD_BYTES, (
```
treff7es, line 1175: This test verifies the truncated aspect **fits** under the limit, but not that the retained prefix is the **largest** one that fits. Those are asymmetric failure modes for the binary search:

- returns *too much* → aspect still oversized → caught here;
- returns *too little* (off-by-one short, exits the loop a step early, or degenerates to `retained=0`) → aspect fits comfortably → **passes silently**.

For a binary search the off-by-one boundary is exactly where bugs hide, and it's currently unasserted — a search that always returned 0 would pass every assertion in this test. Worth adding a **tightness** check: after truncation, assert the boundary is maximal (the retained prefix fits, but keeping ~one more meaningful char would exceed the limit), and/or a small-fixed-budget case where the expected `retained_length` is computed and asserted exactly rather than just `< MAX`. A non-ASCII/astral-char variant (`"\U0001F600"`, 12 JSON bytes/char vs 2 for `\n`) would stress the char/byte accounting harder too.

This is also easier to test directly if the search loop is extracted into a small `_largest_fitting_prefix(text, fits)` helper — you could then unit-test the boundary with tiny synthetic predicates, no multi-MB fixtures, and reuse it in `ensure_view_properties_size`.

metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
```diff
@@ -478,33 +483,69 @@ def ensure_query_properties_size(
...
-                f"Cannot truncate query statement for {entity_urn} as it is smaller than or equal to the required reduction size {reduction_needed}."
+
+        def fits(retained: int) -> bool:
+            serialized = overhead_without_value + len(
+                json.dumps(candidate_value(retained))
             )
+            return serialized < max_payload_size
+
+        # Largest statement prefix that keeps the whole aspect under the limit.
+        lo, hi, retained_length = 0, len(full_statement), 0
```
treff7es: Consider extracting this loop into a small, documented helper — e.g. `_largest_fitting_prefix(text: str, fits: Callable[[int], bool]) -> int`. It's a standard, domain-agnostic algorithm (binary search for the largest value satisfying a monotonic predicate — the "rightmost True" boundary) and it knows nothing about query properties; keep the `fits`/`candidate_value` closures at the call site since those *are* queryProperties-specific.

Why it's worth it here (beyond readability):

- **Directly testable** — the correctness property is "returns the *maximal* fitting prefix", which is hard to assert through the end-to-end path (see my note on the test) but trivial to unit-test on the helper with tiny synthetic predicates, no multi-MB fixtures.
- **Reuse fixes the latent sibling bug** — `ensure_view_properties_size` needs exactly this byte-accurate truncation and currently uses the raw-char `_truncate_or_remove_field`, which has the same unit-mismatch this PR fixes. A shared helper lets both guards converge instead of duplicating (a third truncation variant already exists in the file).

Note it documents its own monotonicity precondition (`fits(k)` true ⇒ `fits(k-1)` true) — that property is load-bearing for the search and worth stating explicitly.

docs/how/updating-datahub.md
```diff
@@ -125,6 +125,8 @@ Requirements:
 
 ### Other Notable Changes
 
+- **(Ingestion framework)** Oversized query metadata no longer fails ingestion runs. Two fixes work together: (1) the SQL parsing aggregator now caps the text of synthetic "composite" queries (built by merging a temp-table chain into one query) at 5MB, configurable via `DATAHUB_MAX_COMPOSITE_QUERY_STATEMENT_CHARS`; and (2) the `EnsureAspectSizeProcessor` size guard for `queryProperties` was fixed to measure serialized (JSON) size correctly instead of comparing a raw character count against a byte-based target — the previous unit mismatch caused the guard to give up and emit oversized aspects, which GMS then rejected with a `400 Cannot parse request entity`, failing the run after enough rejections. This most often affected Snowflake sources with `include_queries` enabled against accounts with heavy temp-table query flows (e.g. Hightouch), but applies to any SQL connector that emits query entities. **No action required;** oversized statements are now truncated (with a warning and a report counter) rather than failing the run.
```
```suggestion
- #18102 **(Ingestion framework)** Oversized query metadata no longer fails ingestion runs. Two fixes work together: (1) the SQL parsing aggregator now caps the text of synthetic "composite" queries (built by merging a temp-table chain into one query) at 5MB, configurable via `DATAHUB_MAX_COMPOSITE_QUERY_STATEMENT_CHARS`; and (2) the `EnsureAspectSizeProcessor` size guard for `queryProperties` was fixed to measure serialized (JSON) size correctly instead of comparing a raw character count against a byte-based target — the previous unit mismatch caused the guard to give up and emit oversized aspects, which GMS then rejected with a `400 Cannot parse request entity`, failing the run after enough rejections. This most often affected Snowflake sources with `include_queries` enabled against accounts with heavy temp-table query flows (e.g. Hightouch), but applies to any SQL connector that emits query entities. **No action required;** oversized statements are now truncated (with a warning and a report counter) rather than failing the run.
```

metadata-ingestion/src/datahub/sql_parsing/sql_parsing_aggregator.py
sgomezvillamor: Why is this 5MB default so much lower than the usual 16MB aspect size limit? Shouldn't they be more aligned? If there is any reason, it should be documented here.

alfiyas-datahub: The system has two possible size limits: a 16 MB aspect validation limit and a 5 MB Kafka message limit. The 16 MB validator is disabled by default in the open-source deployment, while the 5 MB Kafka limit is always enforced. This means that if a metadata payload exceeds 5 MB, Kafka rejects it with a `RecordTooLargeException` before it ever reaches the aspect validator. Therefore, the truncation cap is intentionally set to 5 MB, since that is the effective limit that prevents ingestion failures. I will add a comment documenting this on the constant.

metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
alfiyas-datahub: Pulled the search into `_largest_fitting_prefix` with a short docstring on the monotonicity precondition. `ensure_query_properties_size` now just calls that helper. Added direct unit tests for the boundary cases (maximal fit, caps at text length, returns 0 when nothing fits, variable byte cost for astral chars) plus a tightness assertion on the existing escaped-statement test.

metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
alfiyas-datahub: Done, after shedding name/description, we re-check serialized size. If it's still over the budget, we emit a distinct warning ("Query properties could not be truncated below size constraint") and return without recording a successful truncation. Added `test_ensure_query_properties_reports_when_untruncatable` for this path.

metadata-ingestion/src/datahub/sql_parsing/sql_parsing_aggregator.py
alfiyas-datahub: updated with a plain constant, no env override; removed import os; added comment explaining why 5 MB (Kafka max.request.size, not the 16 MB aspect validator)

sgomezvillamor: Still concerned about the 5MB cap. Are we applying this to Query statements for non-composite queries too?

All the trimming we do is in the `ensure_aspect_size` workunit processor at 16MB. Why should composite queries have a more restrictive limit for their statements?

My proposal would be not to do any trimming in sql parsing and only do that in `ensure_aspect_size` workunit processor.

alfiyas-datahub: Dropped the sql-parsing cap. `ensure_aspect_size` now solely owns queryProperties size trimming, keeping the responsibility in one place and treating composite and non-composite queries identically.

sgomezvillamor:
Thanks for addressing my comment — keeping the trimming responsibility exclusively in the `EnsureAspectSizeProcessor` is IMO the right call; a single place owning `queryProperties` size handling for both composite and non-composite queries is much cleaner.

Two follow-ups before this is ready:

1. The PR description still describes the old two-part fix (the `sql_parsing_aggregator` composite cap / `DATAHUB_MAX_COMPOSITE_QUERY_STATEMENT_CHARS`), which no longer exists in the diff — please update it to match the current approach.
2. @treff7es's points are still pending in the code: the re-check after dropping `name`/`description` (the untruncatable case still records a successful truncation), the `_largest_fitting_prefix` helper extraction + reuse in `ensure_view_properties_size`, and the boundary/tightness tests. They're marked resolved but the changes aren't in HEAD yet.

---

metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
```diff
@@ -478,78 +478,165 @@ def ensure_query_properties_size(
...
+        # the aspect over the limit, don't claim a successful truncation — warn
+        # instead so the likely GMS 400 is visible rather than silent.
+        if self._query_properties_serialized_size(query_properties) >= max_payload_size:
             self.ctx.source_report.warning(
-                title="Query properties truncated due to size constraint",
-                message="Query properties contained too much data and would have caused ingestion to fail",
-                context=f"Query statement was truncated from {original_statement_size} to {new_statement_length} characters for {entity_urn} due to aspect size constraints",
+                title="Query properties could not be truncated below size constraint",
+                message="Query properties remained too large after truncation and may be rejected by GMS",
+                context=(
```
sgomezvillamor, line 534: in context we usually set dict-like information, textual info goes into title/message

metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
```diff
@@ -478,78 +478,165 @@ def ensure_query_properties_size(
...
+                    f"Query properties for {entity_urn} still exceed the "
+                    f"{max_payload_size} byte limit after truncating the statement "
+                    f"and dropping name/description"
+                ),
             )
-            self._record_truncation("queryProperties")
-        else:
+            return
+
+        context = (
```
sgomezvillamor, line 542: in context we usually set dict-like information, textual info goes into title/message

sgomezvillamor (approved):
LGTM 👍 Thanks for the thorough iteration.

Keeping size trimming solely in `EnsureAspectSizeProcessor` (with `sql_parsing_aggregator` untouched) is the right call — composite and non-composite queries now go through one consistent path. The review feedback is all addressed: the byte-accurate `_largest_fitting_prefix` helper is reused by both `ensure_query_properties_size` and `ensure_view_properties_size` (fixing the latent unit-mismatch in view logic too), the post-drop re-verify closes the silent-failure hole by warning instead of recording a false truncation, and the new boundary/tightness/escape tests cover the tricky cases.

Non-blocking: (1) the 5MB `queryProperties` cap rationale (Kafka `max.request.size` vs. the disabled 16MB validator) would be worth a code comment on the constant for future readers; (2) unlike the sibling guards, this only trims the statement + drops name/description rather than trimming fields in priority order — fine given the statement dominates, just noting the divergence. Neither blocks merge.

---

PR #18105, opened 2026-06-30
docs(ingest): document SDK emit modes (sync vs async) for the Python emitter

aviraj-gour:
## Summary

Adds an **Emit Modes** section to the Python Emitter guide (`metadata-ingestion/as-a-library.md`), documenting the SDK's `EmitMode` options so custom-ingestion / direct-SDK users pick the right mode for their workload.

Follow-up to #18014 (which flipped the push-based plugins — Airflow, Dagster, Prefect, Great Expectations — to `ASYNC`). The SDK / `RestEmitter` default stays `SYNC_PRIMARY`, so for direct SDK users the docs are the only safety net. Custom ingestions have been unintentionally running `sync` at high volume, severely impacting GMS/MySQL performance.

The new section covers:

- What each `EmitMode` does and when to use it (`SYNC_WAIT`, `SYNC_PRIMARY`, `ASYNC`, `ASYNC_WAIT`).
- The SDK default (`SYNC_PRIMARY`) stated explicitly, with a callout that it is **not** suited to high-throughput / bulk ingestion.
- An explicit recommendation: **high-volume / bulk ingestion should set `ASYNC`**.
- The async behavioral caveats so users aren't surprised:
  - the emit call does not raise on a rejected/invalid write — failures surface later in the Failed-MCP topic / consumer logs;
  - no read-after-write guarantee.
- A short "choosing a mode" decision guide.

Docs-only change; no code or behavior changes.

### DIFF
--- a/metadata-ingestion/as-a-library.md
+++ b/metadata-ingestion/as-a-library.md
@@ -63,6 +63,43 @@ Other examples:
 
 - [lineage_emitter_mcpw_rest.py](./examples/library/lineage_emitter_mcpw_rest.py) - emits simple bigquery table-to-table (dataset-to-dataset) lineage via REST as MetadataChangeProposalWrapper.
 
+### Emit Modes
+
+`emit()` and `emit_mcp()` accept an optional `emit_mode: EmitMode` argument that controls how long the call waits and what consistency it guarantees. The emitter default is `SYNC_PRIMARY`. You can override it per call, or set `default_emit_mode` once on the emitter.
+
+| `EmitMode`                 | Call returns after…                          | Use when                                                                  |
+| -------------------------- | -------------------------------------------- | ------------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


sgomezvillamor (approved):
LGTM

Just a couple of comments:

* Should we explain that the emit mode can be configured at instance level as well as per method?
* We may also reference/link to the experimental feature here (which by the way, the _MCPs carries the `syncIngest` marker_ is wrong, FIXME the marker is `emitModeMarker`)
https://github.com/datahub-project/datahub/blob/bc8496aed1e49e8985530e51f77adfb7fb9f657b/metadata-ingestion/sink_docs/datahub.md?plain=1#L91-L95

aviraj-gour:
Thanks for review @sgomezvillamor,

> Should we explain that the emit mode can be configured at instance level as well as per method?

I have already added a example for that.

> We may also reference/link to the experimental feature here (which by the way, the MCPs carries the syncIngest marker is wrong, FIXME the marker is emitModeMarker)

Will address this fix separately and for cross link may not be needed - I think its simple and explain types of emit mode.

PR #18108, opened 2026-06-30
feat(looker): capture explore join definitions as view logic

askumar27:
## 📋 Summary

Capture Looker explore **join definitions** (relationship semantics) and emit them as the explore dataset's `viewProperties` (`viewLogic`), so the join graph is reconstructable from ingested metadata.

## 🎯 Motivation

Looker explore joins carry the relationship semantics — which view is joined, the `sql_on` condition, the cardinality (`relationship`), the join type, and the foreign key — that are needed to reconstruct an explore's semantic model.

Today, ingestion reads the Looker API's join objects only to derive **view membership** for lineage, then discards the join conditions. As a result, explore datasets are emitted with no record of *how* their views join: downstream consumers can see that views A and B are related, but not that they join `ON a.id = b.a_id` as `many_to_one`.

This change preserves that information so the join graph (the explore's semantic model) is reconstructable from ingested metadata, symmetric with how view datasets already store their raw LookML.

## 🔁 Old State vs New State

| | Old state (before) | New state (after) |
| --- | --- | --- |
| **Join conditions** (`sql_on`) | Read, used to derive view membership, then **discarded** | Captured and preserved |
| **Relationship / cardinality** (`many_to_one`, …) | Not captured | Captured |
| **Join type** (`left_outer`, `inner`, …) | Not captured | Captured |
| **Foreign key / `from` alias** | Not captured | Captured |
| **Explore dataset metadata** | Lineage edges to joined views only | Lineage edges **plus** a reconstructed LookML `explore { … }` block in `viewProperties.viewLogic` |
| **What a consumer can tell** | Views A and B are *related* | Views A and B join **`ON a.id = b.a_id` as `many_to_one`** |
| **Join-less explores** | No `viewProperties` | No `viewProperties` (**unchanged**) |

**Before** — explore dataset carried lineage edges but no record of *how* views join:

```text
upstreamLineage: [customers, orders]   # related, but join semantics lost
(no viewProperties aspect)
```

**After** — same lineage, plus the reconstructed join graph as view logic:

```lookml
explore: customer_orders {
  label: "Customer Orders"
  join: orders {
    type: left_outer
    relationship: many_to_one
    sql_on: ${orders.customer_id} = ${customers.id} ;;
  }
}
```

## 🔧 Changes Overview

**New Features:**
- New `LookerExploreJoin` structured representation holding `name`, `from_view`, `sql_on`, `relationship`, `join_type`, and `foreign_key`, with constructors for both the Looker API and LookML code paths and a serializer back to a LookML `join` block.
- Explore datasets that have joins now emit a reconstructed LookML `explore { ... join: ... }` block as `viewProperties.viewLogic` (`viewLanguage: lookml`).

**Modifications:**
- Both explore ingestion code paths (Looker API source and LookML source) now capture the structured joins in addition to the existing field-name extraction used for lineage. The existing lineage behavior is unchanged.
- Explores **without** joins are unaffected — no `viewProperties` aspect is emitted for them.

**Dependencies:** None.

## 🏗️ Architecture/Design Notes

- The new structured join capture is **additive and orthogonal** to the existing `joins` field (which only holds field names referenced in join conditions, retained for lineage). The two are kept distinct on purpose.
- The reconstructed LookML is stored in the standard `ViewProperties` aspect (`viewLogic`/`viewLanguage`), mirroring the existing convention for view datasets — no new aspect or model change is introduced.
- The Looker API does not expose an explore's source file the way it does for views, which is why the join relationships are serialized into view logic here rather than read from a file.

## 🧪 Testing

- **Unit tests** (`tests/unit/looker/test_looker_common.py`): cover capturing joins from both the API and LookML inputs, the LookML serialization, and the no-joins → `None` case.
- **Integration goldens**: `golden_test_ingest_joins.json` and `golden_test_ingest_unaliased_joins.json` updated with the additive `viewProperties` only.
- **Live ingestion smoke test**: ran end-to-end ingestion against a real Looker instance with a file sink; explore datasets correctly emitted reconstructed `viewProperties` with `join:` blocks and their `sql_on` / `relationship` / `type` rendered as expected.
- `ruff` + `mypy` clean on changed files (via Gradle lint).

## 📊 Impact Assessment

- **Affected Components:** Looker/LookML ingestion source (`metadata-ingestion`) only.
- **Breaking Changes:** None. The change is purely additive — a new aspect on explores that have joins; join-less explores are byte-for-byte unchanged.
- **Performance Impact:** Negligible — joins are already iterated for lineage; this reuses that pass.
- **Risk Level:** Low. No model/schema changes, no new dependencies, existing lineage behavior untouched.

## 🚀 Deployment Notes

None — no configuration, migrations, or feature flags required.

## 🔗 References

- Docs: [Looker entity reference](https://datahubproject.io/docs/generated/ingestion/sources/looker)

## ❓ Questions for Reviewers

- The reconstructed `explore` block omits params we don't currently capture (e.g. only `join`-level fields are serialized). If there's appetite for capturing additional explore-level params in `viewLogic`, happy to follow up in a separate PR.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/looker/looker_common.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/looker/looker_common.py
@@ -25,6 +25,7 @@ from looker_sdk.rtl.serialize import DeserializeError
 from looker_sdk.sdk.api40.models import (
     LookmlModelExplore,
     LookmlModelExploreField,
+    LookmlModelExploreJoins,
     User,
     WriteQuery,
 )
@@ -98,6 +99,7 @@ from datahub.metadata.schema_classes import (
     TagAssociationClass,
     TagPropertiesClass,
     TagSnapshotClass,
+    ViewPropertiesClass,
 )
 from datahub.metadata.urns import TagUrn
 from datahub.sdk.dataset import Dataset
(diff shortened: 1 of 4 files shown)


alexsku (approved):
Looks good! Clean addition that captures explore join semantics (relationship, sql_on, type, foreign_key) and reconstructs them as LookML view logic via `ViewPropertiesClass`. Good separation from the existing `joins` field (retained for lineage), solid test coverage, and golden files updated.

PR #18109, opened 2026-07-01
perf(frontend): reduce initial load of FE bundle

benjiaming:
## Summary

* Move remirror out of the eager initial bundle so it loads on-demand only when an editor actually renders. 
* Initial JS download shrinks by ~164 KB and Lighthouse TTI improves by ~30 ms (1,306 → 1,276 ms, median of 3 runs against the same server).

## Drive-by
- Lazy-load Remirror
- Initial JS bundle size report on PR
- Swapped 9 `@mui/icons-material` imports for visually equivalent Phosphor CSR path imports (`@phosphor-icons/react/dist/csr/X`)

Screenshot 2026-06-30 at 5 00 07 PM


datahub-web-react/src/alchemy-components/components/Editor/Editor.tsx
```diff
@@ -1,159 +1,14 @@
...
-                </RemirrorLocaleProvider>
-            </ThemeProvider>
-        </EditorContainer>
-    );
-});
+// Thin shell that lazy-loads the remirror implementation so remirror is excluded
+// from the initial bundle. Callers keep the same <Editor /> API including ref forwarding.
+export const Editor = forwardRef<unknown, EditorProps>((props, ref) => (
+    <Suspense fallback={null}>
+        <EditorImpl {...props} ref={ref} />
```
chriscollins3456, line 11: this change doesn't cause the editor to be loaded laggily or anything? did it feel good locally when you tested it?

chriscollins3456 (approved):
def a fan of removing the old Editor component and consolidating on the one in alchemy-components

datahub-web-react/src/alchemy-components/components/Editor/Editor.tsx
benjiaming: I tested with throttling on 3G and the editor still loaded almost instantly.

PR #18113, opened 2026-07-01
fix(ingest/spark): restore Java 17 bytecode for the agent shadow jar (regressed by the Java 21 migration)

treff7es:
## Summary

The `acryl-spark-lineage` agent runs **inside customer Spark clusters**, whose default JVM today is **Java 17** on the dominant managed platforms — **Amazon EMR 7.x** (Corretto 17) and **Databricks Runtime 16–17.x LTS** (JDK 17). The repo-wide Java 17→21 toolchain migration made the agent's shadow jar **Java 21 bytecode**, so it now fails with `UnsupportedClassVersionError` on those runtimes. The previous `-PjavaClassVersionDefault=8` workaround is no longer usable (datahub-client dropped Java 8).

This keeps the agent — and every module bundled into its shadow jar — at **Java 17 bytecode**, while the rest of the repo (GMS/backend, which runs on your own Java 21) stays at 21. Java 17 bytecode loads fine on the Java 21 backend, so the shared modules are unaffected there.

## Changes

- **`build.gradle`**: `javaClassVersion()` now honors a per-module `ext.javaClassVersionOverride`, evaluated **before** the global default. (`javaClassVersionDefault` is always present via `gradle.properties`, so the override has to be checked first to take effect.)
- **Per-module pin** (`ext.javaClassVersionOverride = 17`) on exactly the modules the agent shadow jar bundles: `li-utils`, `metadata-models`, `entity-registry`, `datahub-event`, `datahub-schematron:lib`, `datahub-client`, `openlineage-converter`, `acryl-spark-lineage`.
- **`entity-registry`**: rewrite three Java 21 pattern-matching `switch` statements (`graph/cache/{AncestorWalkResult,GraphReadResult,MembershipNeighborResult}.java`) to Java 17 `instanceof` if-chains — the **only** Java 21 language features anywhere in the agent's bundled dependency chain. `sealed interface`/`record`/`instanceof` patterns are all Java 17-legal and are kept.

## Why per-module (not a global flag)

The published artifact must be Java 17 **by default**, not via a manual `-P` flag. Everything the shadow jar bundles must be ≤17 bytecode to load on a Java 17 JVM, so the pin covers the whole bundled chain; unrelated backend modules keep targeting Java 21.

## Testing

- Built `:acryl-spark-lineage:shadowJar_2_13` with **no `-P` flag** and confirmed the bundled classes are Java 17 bytecode (major version 61): the agent's own `DatahubSparkListener`, the pegasus module `metadata-models` (`ConfigEntitySpec`), and `entity-registry` (`GraphReadResult`).
- `:entity-registry:test` (537) and `:acryl-spark-lineage:test` (42) pass.

## Checklist

- [x] Tests pass for the touched modules; bytecode target verified
- [ ] No breaking changes — bytecode-target only; Java 17 classes run on the Java 21 backend

## Note

Complements #17947 (OpenLineage relocation for EMR/DataZone classpath coexistence): the relocation fixes the *collision*, this fixes the *bytecode runtime* — both are needed for the agent to actually run on EMR 7.x / Databricks 16–17.

### DIFF
--- a/.github/workflows/spark-smoke-test.yml
+++ b/.github/workflows/spark-smoke-test.yml
@@ -29,6 +29,10 @@ concurrency:
 jobs:
   spark-smoke-test:
     runs-on: ubuntu-latest
+    # Fail fast instead of hanging toward the 6h default: the docker-based smoke test (quickstart +
+    # Spark image build + spark-submit) has stalled for >1h on flaky network fetches during image
+    # builds. A bounded timeout turns such stalls into a quick, actionable red.
+    timeout-minutes: 60
     steps:
       - uses: acryldata/sane-checkout-action@186e92cc5948a9c3e1cc7a96eaff9f776f3fc8e3 # v7
       - name: Set up JDK 21
--- a/build.gradle
+++ b/build.gradle
@@ -14,8 +14,16 @@ buildscript {
       : ext.jdkVersionDefault
   }
 
-  // Main bytecode version selector
+  // Main bytecode version selector.
+  // A per-module ext.javaClassVersionOverride takes precedence over the global default: it lets the
+  // modules bundled into the acryl-spark-lineage agent shadow jar pin Java 17 bytecode so the agent
+  // runs on managed Spark whose default JVM is Java 17 (Amazon EMR 7.x, Databricks Runtime 16-17),
+  // while the rest of the repo targets 21. (javaClassVersionDefault is always present via
+  // gradle.properties, so the override must be checked first.)
   ext.javaClassVersion = { p ->
+    if (p.hasProperty('javaClassVersionOverride')) {
+      return Integer.valueOf(String.valueOf(p.getProperty('javaClassVersionOverride')))
+    }
     return p.hasProperty('javaClassVersionDefault')
       ? Integer.valueOf((String) p.getProperty('javaClassVersionDefault'))
       : ext.javaClassVersionDefault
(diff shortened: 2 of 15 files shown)


Dutt23 (approved):

sgomezvillamor (approved):
LGTM!

Just missing a mention of the requirement in the user-facing docs (or are we already covering that somewhere? 😅)

> The repo-wide Java 17→21 toolchain migration made the agent's shadow jar Java 21 bytecode, so it now fails with UnsupportedClassVersionError on those runtimes.

This sounds like a regression fix — worth reflecting that in the PR title so it gets picked up properly in the generated changelogs.

treff7es:
@sgomezvillamor both addressed:

- **Docs:** the README was actually stale here — the "How to build" section said *"Use Java 8 to build"* with the now-removed `-PjavaClassVersionDefault=8` flag. Replaced it with a **Runtime requirements** section: Java 17+ required; supported runtimes (EMR 7.x, Databricks 16.4 LTS / 17.3 LTS+, Dataproc Serverless 2.x / 3.0, Fabric 2.0) and the explicitly-unsupported Java 8/11 ones (EMR 6.x, Databricks ≤15.4, Glue ≤4.0, Dataproc-GCE 2.x, Fabric 1.3). — commit 36cb0d2
- **Title:** retitled to `fix(ingest/spark): …` — it's a regression from the repo-wide Java 17→21 migration, so it now lands in the generated changelog.

PR #18114, opened 2026-07-01
fix(ingest): accept relative start_time in generated JSON schema

hhobson:
## Summary

`BaseTimeWindowConfig.start_time` is typed as `datetime`, so the JSON schema pydantic generates for it (and which is published at `https://docs.datahub.com/schemas/datahub_ingestion_schema.json`) advertises `{"type": "string", "format": "date-time"}`. The relative-time support (`-7d`, `-7 days`) lives entirely in the `parse_start_time` `mode="before"` validator, which the schema generator cannot see.

As a result, any tool that validates a recipe against the published schema with format assertion enabled (IDE YAML/JSON validation, the ingestion UI form, CI schema linters) rejects the documented relative syntax, even though the ingestion runtime accepts it. This affects every source that inherits `BaseTimeWindowConfig` / `BaseUsageConfig` (Snowflake, BigQuery, Redshift, etc.).

## Change

Widen only the *generated* JSON schema for `start_time` via a `WithJsonSchema` annotation (`anyOf`: an ISO 8601 `date-time` branch plus a negative-relative-timespan pattern branch). The declared type stays `datetime` and the `mode="before"` validator is untouched, so runtime behaviour and downstream typing are unchanged.

`end_time` is deliberately left as strict `date-time`: it has no relative parser, so its current schema is accurate.

Notes:
- The relative branch requires a leading minus sign, matching the validator's `delta < 0` assertion.
- The `abs(delta) >= bucket_duration` check depends on the sibling `bucket_duration` field, so it cannot be expressed in a per-string JSON-schema pattern and remains a runtime-only check. The pattern is intentionally permissive (screens obvious garbage like `-42 hello`) so no valid recipe is ever blocked by the editor/CI; the runtime validator stays the source of truth for semantics.
- The pattern is lowercase-only because JSON Schema uses ECMA-262 regex, which does not support inline `(?i)` flags; recipes use lowercase units in practice.

## Checklist

- [ ] Docs related to the changes have been added/updated
- [ ] For any breaking change/potential downtime/deprecation/big changes an entry has been made in Updating DataHub (n/a — schema is strictly more permissive)

## Testing

Added `test_start_time_json_schema_allows_relative_timespans` in `tests/unit/config/test_time_window_config.py`, which asserts the generated schema keeps a `date-time` branch and that the relative-timespan branch matches `-7d` / `-7 days` while rejecting `-42 hello`. Existing runtime parsing tests continue to pass.

```
$ pytest tests/unit/config/test_time_window_config.py
6 passed
```


treff7es (approved):
lgtm, thanks

hhobson:
@sgomezvillamor @treff7es Is there anything needed before this can be merged?

PR #18120, opened 2026-07-01
feat(ingest/gc): add query cleanup pass for stale SYSTEM queries

aviraj-gour:
## Summary

Adds a `QueryCleanup` first-pass to the `datahub-gc` source that **soft-deletes** old `SYSTEM` queries by age alone.

Query entities have no first-pass cleanup today, and `query` is excluded from stateful ingestion's stale-entity removal. As a result, stale `SYSTEM` queries accumulate as permanent orphans in Elasticsearch and primary storage, causing index bloat and search/UI noise.

## Approach

- **Selection by age alone, no reference gate.** Candidates come from a single server-side search filter: `source == SYSTEM` AND `lastModifiedAt < cutoff` (`cutoff = now - retention_days`). `lastModifiedAt` is already `@Searchable` on `queryProperties`, so no backend change or reindex is needed.
- **Oldest-first ordering.** Results are sorted by `lastModifiedAt` ascending, so a capped run deletes the *most-stale* queries (not an arbitrary scroll slice) and dry-run previews reflect what a real run would delete. This required threading an optional `sort_by`/`sort_order` through `get_urns_by_filter` into `scrollAcrossEntities`'s `sortInput`.
- **`SYSTEM`-only.** `MANUAL` queries are never touched; enforced in the filter, not a config option.
- **Soft delete via the workunit stream.** Each candidate yields a `Status(removed=true)` workunit — the same mechanism as stateful ingestion's stale-entity removal — and the sink batches/writes it (the default `datahub-rest` sink is `ASYNC_BATCH`, which provides built-in backpressure). The pass keeps no concurrency or emit-mode logic.
- **Two-pass lifecycle (no hard delete in this pass).** This pass only soft-deletes. The existing `SoftDeletedEntitiesCleanup` already lists `query` in its default entity types and performs the hard-delete second pass — with `deleteReferences` and its own throttling — on a later run. This mirrors the DataProcess cleanup pattern.
- **Run-bounding.** Stops once `limit_entities_delete` or `runtime_limit_seconds` is reached (including in dry-run), mirroring `SoftDeletedEntitiesCleanup`.

## Config

**Enabled by default** with a conservative **180-day (~6 month)** retention — a large safety margin over any realistic connector ingestion window, and a major improvement over keeping `SYSTEM` queries indefinitely.

| Option | Default | Notes |
|---|---|---|
| `enabled` | `true` | Enabled by default |
| `retention_days` | `180` | Age cutoff on `lastModifiedAt` (~6 months) |
| `batch_size` | `500` | Search fetch size |
| `limit_entities_delete` | `25000` | Approx. cap per run; set to `null` to disable |
| `runtime_limit_seconds` | `7200` | Per-run wall-clock limit |

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/gc/datahub_gc.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/gc/datahub_gc.py
@@ -169,7 +169,7 @@ class DataHubGcSource(Source):
                 with self.report.new_stage("Query Cleanup"):
                     yield from self.query_cleanup.get_workunits()
             except Exception as e:
-                self.report.failure("While trying to cleanup queries ", exc=e)
+                self.report.failure("While trying to cleanup queries", exc=e)
         if self.config.soft_deleted_entities_cleanup.enabled:
             try:
                 with self.report.new_stage("Soft Deleted Entities Cleanup"):
--- a/metadata-ingestion/src/datahub/ingestion/source/gc/datahub_gc.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/gc/datahub_gc.py
@@ -47,7 +47,7 @@ logger = logging.getLogger(__name__)
 class DataHubGcSourceConfig(ConfigModel):
     dry_run: bool = Field(
         default=False,
-        description="Whether to perform a dry run or not. This is only supported for dataprocess cleanup and soft deleted entities cleanup.",
+        description="Whether to perform a dry run or not. This is only supported for dataprocess cleanup, query cleanup, and soft deleted entities cleanup.",
     )
 
     cleanup_expired_tokens: bool = Field(
--- a/metadata-ingestion/src/datahub/ingestion/source/gc/datahub_gc.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/gc/datahub_gc.py
@@ -27,6 +27,11 @@ from datahub.ingestion.source.gc.execution_request_cleanup import (
     DatahubExecutionRequestCleanupConfig,
     DatahubExecutionRequestCleanupReport,
 )
+from datahub.ingestion.source.gc.query_cleanup import (
+    QueryCleanup,
+    QueryCleanupConfig,
+    QueryCleanupReport,
+)
 from datahub.ingestion.source.gc.soft_deleted_entity_cleanup import (
     SoftDeletedEntitiesCleanup,
     SoftDeletedEntitiesCleanupConfig,
@@ -81,12 +86,18 @@ class DataHubGcSourceConfig(ConfigModel):
         description="Configuration for execution request cleanup",
     )
 
+    query_cleanup: QueryCleanupConfig = Field(
+        default_factory=QueryCleanupConfig,
+        description="Configuration for query cleanup",
+    )
+
 
 @dataclass
 class DataHubGcSourceReport(
     DataProcessCleanupReport,
     SoftDeletedEntitiesReport,
     DatahubExecutionRequestCleanupReport,
+    QueryCleanupReport,
 ):
     expired_tokens_revoked: int = 0
 
@@ -125,6 +136,9 @@ class DataHubGcSource(Source):
             graph=self.graph,
             report=self.report,
         )
+        self.query_cleanup = QueryCleanup(
+            ctx, self.config.query_cleanup, self.report, self.config.dry_run
+        )
 
     @classmethod
     def create(cls, config_dict, ctx):
@@ -150,6 +164,12 @@ class DataHubGcSource(Source):
                     self.truncate_indices()
             except Exception as e:
                 self.report.failure("While trying to truncate indices ", exc=e)
+        if self.config.query_cleanup.enabled:
+            try:
+                with self.report.new_stage("Query Cleanup"):
+                    self.query_cleanup.cleanup_queries()
+            except Exception as e:
+                self.report.failure("While trying to cleanup queries ", exc=e)
         if self.config.soft_deleted_entities_cleanup.enabled:
             try:
                 with self.report.new_stage("Soft Deleted Entities Cleanup"):
--- a/metadata-ingestion/src/datahub/ingestion/source/gc/datahub_gc.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/gc/datahub_gc.py
@@ -167,7 +167,7 @@ class DataHubGcSource(Source):
         if self.config.query_cleanup.enabled:
             try:
                 with self.report.new_stage("Query Cleanup"):
-                    self.query_cleanup.cleanup_queries()
+                    yield from self.query_cleanup.get_workunits()
             except Exception as e:
                 self.report.failure("While trying to cleanup queries ", exc=e)
         if self.config.soft_deleted_entities_cleanup.enabled:
--- a/metadata-ingestion/src/datahub/ingestion/source/gc/query_cleanup.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/gc/query_cleanup.py
@@ -1,8 +1,15 @@
 import logging
 import time
+from concurrent.futures import (
+    ALL_COMPLETED,
+    FIRST_COMPLETED,
+    Future,
+    ThreadPoolExecutor,
+    wait,
+)
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
-from typing import Iterable, Optional
+from typing import Dict, Iterable, Optional
 
 from pydantic import Field
 
@@ -47,8 +54,8 @@ class QueryCleanupConfig(ConfigModel):
             "(False) soft-deletes here and lets soft_deleted_entities_cleanup hard-delete them "
             "on a later run, which additionally calls deleteReferences to clear inbound pointers. "
             "This shortcut skips that step, so `via` pointers on lineage aspects are left "
-            "pointing at the deleted query. Hard deletes are issued one at a time; soft deletes "
-            "are emitted as workunits and batched by the sink."
+            "pointing at the deleted query. Hard deletes are parallelized across "
+            "max_workers workers; soft deletes are emitted as workunits and batched by the sink."
         ),
     )
 
@@ -57,9 +64,21 @@ class QueryCleanupConfig(ConfigModel):
         description="The number of entities to fetch in a batch from search.",
     )
 
+    max_workers: int = Field(
+        10,
+        description=(
+            "Number of worker threads for parallel hard deletion. Only applies when "
+            "hard_delete_entities is set; soft deletes go through the sink."
+        ),
+    )
+
     limit_entities_delete: Optional[int] = Field(
         25000,
-        description="Approximate max number of queries to delete in a single run.",
+        description=(
+            "Approximate max number of queries to delete in a single run. For hard delete, "
+            "the cap is checked before each submission, so concurrent workers may overshoot "
+            "it by up to max_workers entities."
+        ),
     )
 
     runtime_limit_seconds: int = Field(
@@ -96,7 +115,7 @@ class QueryCleanup:
     `lastModifiedAt < cutoff`. Soft deletes are emitted as status workunits and left to the
     sink to batch and write (the same mechanism as stale-entity removal); the existing
     SoftDeletedEntitiesCleanup completes the hard-delete second pass. Hard deletes, when
-    enabled, have no bulk endpoint and are issued imperatively one at a time.
+    enabled, have no bulk or async endpoint, so they are parallelized across a worker pool.
     """
 
     def __init__(
@@ -162,11 +181,7 @@ class QueryCleanup:
             return True
         return False
 
-    def get_workunits(self) -> Iterable[MetadataWorkUnit]:
-        if not self.config.enabled:
-            return
-        self.start_time = time.time()
-
+    def _iter_candidate_urns(self) -> Iterable[Urn]:
         for urn in self._get_urns():
             try:
                 query_urn = Urn.from_string(urn)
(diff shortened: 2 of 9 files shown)


metadata-ingestion/src/datahub/ingestion/source/gc/query_cleanup.py
```diff
@@ -0,0 +1,268 @@
...
+
+    def _collect_hard_deletes(
+        self, futures: Dict["Future[None]", Urn], return_when: str
+    ) -> Dict["Future[None]", Urn]:
+        done, not_done = wait(futures, return_when=return_when)
+        for future in done:
+            urn = futures[future]
+            exc = future.exception()
+            if exc is not None:
+                self.report.warning(
```
treff7es: **Should fix — hard-delete failures are downgraded from `failure` to `warning`.** The sibling reports the identical operation (`graph.delete_entity(hard=True)` raising in a worker) as `report.failure` — see `soft_deleted_entity_cleanup.py:235-241`. The distinction is load-bearing: `failure` drives a non-zero pipeline exit, `warning` does not.

Failure scenario: a token expires mid-run → every delete 401s → 25k warnings, `num_queries_hard_deleted=0`, and the pipeline still exits **green**. For a destructive pass whose whole job is deleting, a total failure surfacing only as warnings is the wrong default. Either switch to `self.report.failure(...)` to match the sibling, or justify the downgrade with a comment (e.g. if per-URN transient failures shouldn't fail the whole run).

metadata-ingestion/src/datahub/ingestion/source/gc/query_cleanup.py
```diff
@@ -0,0 +1,268 @@
...
+                    title="Failed to hard delete query",
+                    message="Failed to hard delete query",
+                    context=urn.urn(),
+                    exc=exc,
+                )
+            else:
+                self.report.report_query_hard_deleted(urn.urn())
+        return {f: futures[f] for f in not_done}
+
+    def _hard_delete_queries(self) -> None:
```
treff7es: **Should fix — no throttling on the hard-delete path.** The sibling `SoftDeletedEntitiesCleanup` deliberately paces itself with a `delay` field (default `0.25s`, slept every `batch_size` processed — `soft_deleted_entity_cleanup.py:39-42, 243-251`). This loop has no equivalent: it fires blocking `graph.delete_entity(hard=True)` calls across `max_workers` (10) as fast as they complete. On a 25k-query run that's a sustained synchronous-delete burst against GMS with no rate limit.

Suggest adding a `delay` config field mirroring the sibling and sleeping in `_collect_hard_deletes` (or between submissions). The soft-delete default path doesn't need this — it just yields workunits and the `datahub-rest` sink batches them in `ASYNC_BATCH` mode, which is the framework's built-in backpressure — but a one-line note saying the soft path is intentionally sink-throttled while the hard path is not would help.

metadata-ingestion/src/datahub/ingestion/source/gc/query_cleanup.py
```diff
@@ -0,0 +1,268 @@
...
+
+    max_workers: int = Field(
+        10,
+        description=(
+            "Number of worker threads for parallel hard deletion. Only applies when "
+            "hard_delete_entities is set; soft deletes go through the sink."
+        ),
+    )
+
+    limit_entities_delete: Optional[int] = Field(
```
treff7es, line 52: **Optional nit — add pydantic lower bounds to the count fields.** `Field(..., ge=1)` on `limit_entities_delete` (and `ge=1`/`ge=0` on `batch_size`/`max_workers`/`retention_days`) would move these invariants from implicit to parse-time-enforced at zero complexity cost. The one with real value is `ge=1` here: `limit_entities_delete=0` currently passes construction and makes `_deletion_limit_reached()` trip immediately → deletes nothing silently. (`Optional[int]` for the unbounded case is correct — keep it.) The sibling omits these too, so this is a pre-existing norm, not a regression — but this is a new type, so it's the cheap moment to do it right. Don't retrofit the sibling here.

metadata-ingestion/src/datahub/ingestion/source/gc/datahub_gc.py
```diff
@@ -150,6 +164,12 @@ def get_workunits_internal(
                     self.truncate_indices()
             except Exception as e:
                 self.report.failure("While trying to truncate indices ", exc=e)
+        if self.config.query_cleanup.enabled:
```
treff7es, line 167: **Should fix (one-line comment) — document the stage-ordering contract here.** `query_cleanup` must run *before* `soft_deleted_entities_cleanup` for the intended two-pass lifecycle (soft-delete queries this run → `SoftDeletedEntitiesCleanup` hard-deletes them, with `deleteReferences`, on a later run — it already includes `query` in its default `entity_types`). That rationale currently lives only in the `query_cleanup.py` docstrings, not where the ordering decision actually is. A future contributor reordering stages for an unrelated reason could silently break the contract. A short comment noting *why* the order matters (and that same-run interaction is benign because the sink writes after the source fully drains) removes that reverse-engineering risk.

metadata-ingestion/tests/unit/test_query_cleanup.py
```diff
@@ -0,0 +1,165 @@
...
+            aspect = mcp.aspect
+            assert isinstance(aspect, StatusClass)
+            assert aspect.removed
+        # Soft delete never touches the imperative hard-delete path.
+        self.mock_graph.delete_entity.assert_not_called()
+        assert self.report.num_queries_found == 3
+        assert self.report.num_queries_soft_deleted == 3
+        assert self.report.num_queries_hard_deleted == 0
+
+    def test_hard_delete_uses_graph_delete_entity(self) -> None:
```
treff7es: **Should fix — the hard-delete path is only happy-path tested.** This is the more dangerous (irreversible, parallel) code and its non-trivial branches are uncovered. Three gaps, all shareable via one `hard_delete_entities=True` + `side_effect` setup:

- **[high] Failure path** (`query_cleanup.py:219-228`): a raising `delete_entity` should be reported via `report.warning` and **not** counted in `num_queries_hard_deleted`. This is the whole reason the code reads `future.exception()` instead of `future.result()`; nothing guards a regression to `.result()` or a mis-placed counter.
- **[high] Mid-stream cap stop** (`:238`): `test_respects_deletion_limit` / `test_stops_when_runtime_limit_reached` only exercise the soft path. The hard-delete cap guards an irreversible op and is asserted nowhere. Set `limit_entities_delete=2`, feed 5+ URNs, assert `delete_entity.call_count <= 2 + max_workers` and `qc_deletion_limit_reached`.
- **[med] `FIRST_COMPLETED` draining** (`:243-245`): every hard-delete test uses 2 URNs with `max_workers=10`, so `len(futures) >= max_workers` is never true and the drain + final-sweep branch is dead in tests. Set `max_workers=2`, feed 5 URNs, assert all 5 deleted (no drops/double-counts).

metadata-ingestion/src/datahub/ingestion/source/gc/query_cleanup.py
```diff
@@ -0,0 +1,268 @@
...
+
+    def _get_urns(self) -> Iterable[str]:
+        cutoff_millis = int(
+            (
+                datetime.now(timezone.utc).timestamp()
+                - self.config.retention_days * 24 * 60 * 60
+            )
+            * 1000
+        )
+        yield from self.graph.get_urns_by_filter(
```
treff7es, line 118: **Should fix (needs a client change) — candidates aren't ordered oldest-first, so a capped run deletes an arbitrary subset.** `get_urns_by_filter` issues `scrollAcrossEntities` with no `sortCriteria` (the wrapper in `graph/client.py` doesn't expose sort at all). With `query="*"` every doc ties on `_score`, so the order is just Elasticsearch's stable scroll tiebreaker (doc-id/urn) — deterministic but **arbitrary w.r.t. `lastModifiedAt`**. The age filter decides *eligibility*, not *deletion order*. When `limit_entities_delete` or `runtime_limit_seconds` trips mid-stream (`:262`, `:238`), you delete a random 25k, not the oldest 25k — and `sample_deleted_queries` / the dry-run preview show an arbitrary slice too.

Caveat so this isn't overstated: it still *converges* — each run excludes already soft/hard-deleted candidates (`status=NOT_SOFT_DELETED`), so the eligible pool strictly shrinks and the backlog drains over ceil(total/cap) runs regardless of order. The real failure mode is the inflow case: if stale SYSTEM queries accumulate faster than one capped run clears, arbitrary ordering means the genuinely-ancient ones can linger indefinitely while newer-eligible ones churn.

Fix: sort by `lastModifiedAt` ascending so the most-stale go first (and dry-run previews become meaningful). `scrollAcrossEntities` supports `sortCriteria`, but `get_urns_by_filter` doesn't thread it through today, so this needs a small client-side addition (a `sort` param) first. Reasonable as a follow-up if you'd rather not widen this PR's scope.

metadata-ingestion/src/datahub/ingestion/source/gc/query_cleanup.py
aviraj-gour: Will remove hard delete logic from query cleanup.

metadata-ingestion/tests/unit/test_query_cleanup.py
aviraj-gour: removed hard delete.

metadata-ingestion/src/datahub/ingestion/source/gc/query_cleanup.py
```diff
@@ -0,0 +1,268 @@
...
+                f"Dry run is on, otherwise it would have deleted {query_urn} "
+                f"(hard_delete={self.config.hard_delete_entities})"
+            )
+            # Record the candidate so a dry run still previews what would go.
+            self.report.sample_deleted_queries.append(query_urn.urn())
+
+    def _hard_delete_one(self, urn: Urn) -> None:
+        self.graph.delete_entity(urn=urn.urn(), hard=True)
+
+    def _collect_hard_deletes(
```
sgomezvillamor: I understand we need a new stage for soft-deleting query entities in particular — filling the gap for Queries in stateful ingestion across sources. That makes sense.

But why does hard deletion also need to be implemented as a particular case for Queries? I'd expect that to be handled by the existing `SoftDeletedEntitiesCleanup`, with `Query` entity support added there.

https://github.com/datahub-project/datahub/blob/15462bec1cbd45eb31230778bbe9ddc16dbe2a06/metadata-ingestion/src/datahub/ingestion/source/gc/soft_deleted_entity_cleanup.py#L25-L78

Isn't that the pattern we do for DataProcess? or there do we also support direclty hard deleting on the first pass?

aviraj-gour: Yeah, removing hard_delete from query cleanup logic.

metadata-ingestion/src/datahub/ingestion/source/gc/datahub_gc.py
aviraj-gour: Ordering here doesn't actually matter — these are independent operations. query_cleanup yields soft-delete workunits (buffered by the batching sink, plus ES index lag), while soft_deleted_entities_cleanup discovers targets via ES search. So a query soft-deleted this run isn't visible to the soft-deleted pass in the same run — it gets hard-deleted on a later run regardless of stage order. 
Also there is retention_days params in soft_deleted_cleanup.

metadata-ingestion/src/datahub/ingestion/source/gc/query_cleanup.py
aviraj-gour: threaded an optional sort_by/sort_order through get_urns_by_filter (into scrollAcrossEntities's sortInput), and query cleanup now requests lastModifiedAt ascending, so the most-stale queries go first and capped runs / dry-run previews reflect the true oldest slice.

metadata-ingestion/tests/unit/sdk_v2/test_search_client.py
```diff
@@ -770,6 +771,7 @@ def test_get_urns_with_skip_cache() -> None:
                 "batchSize": unittest.mock.ANY,
                 "scrollId": None,
                 "skipCache": True,
+                "sortInput": None,
```
sgomezvillamor, line 774: may we also add a test case with non-null `sortInput` ?

sgomezvillamor:
Reviewed the implementation — the approach is clean and well-scoped. Server-side selection (`source == SYSTEM` AND `lastModifiedAt < cutoff`) with no per-entity lookups scales well, the lazy generator + early break honors the limits without over-fetching, and delegating hard-delete to the existing `SoftDeletedEntitiesCleanup` pass keeps this pass minimal. I also confirmed the sorted deep-scroll is safe: `ESUtils.buildSortOrder` appends `urn` ascending as a tiebreaker when the sort criteria don't already include it, so oldest-first paging over `scrollAcrossEntities` is stable. The `qc_`-prefixed report flags correctly avoid colliding with `SoftDeletedEntitiesReport`'s `runtime_limit_reached`/`deletion_limit_reached` in the merged `DataHubGcSourceReport`.

Two things worth addressing:

**1. The PR description doesn't match the shipped code.** The description documents `hard_delete_entities`, `max_workers`, a "Hard delete via a worker pool" path, a `get_workunits()` that "dispatches: dry-run preview, parallel hard delete, or soft-delete", "10 passing … parallel hard delete", and `retention_days (30)`. None of that is in the diff — `QueryCleanupConfig` has only `enabled`, `retention_days` (default **90**), `batch_size`, `limit_entities_delete`, `runtime_limit_seconds`; there's no hard-delete path; and the test file has 9 tests, none for hard delete. The soft-delete-only design is actually the cleaner choice, so this is a matter of updating the description to match rather than adding the missing code.

**2. Dry-run ignores `limit_entities_delete`.** In `_preview_candidates`, the stop condition calls `_deletion_limit_reached()`, which tests `self.report.num_queries_soft_deleted` — but that counter is never incremented during a dry run, so it stays `0` and the cap never triggers. A dry run therefore scrolls all matching candidates (bounded only by `runtime_limit_seconds`), and `num_queries_found` will overstate what a real capped run would delete — contradicting the intent that the dry-run preview reflect a real run. Suggest counting previewed candidates against `limit_entities_delete` in the dry-run path too.

Minor/optional: `num_queries_found` counts one past the cap in the real path (incremented before the `_deletion_limit_reached()` check, so `found == limit + 1` while `soft_deleted == limit`); and the "`limit_entities_delete=None` disables the cap" behavior is only in a code comment that points to the field description, which doesn't actually mention it.

---

sgomezvillamor (commented):
LGTM

As raised internally, just pending on final decision on default value in the bootstra mcp

metadata-ingestion/tests/unit/sdk_v2/test_search_client.py
aviraj-gour: Added a test.

metadata-service/configuration/src/main/resources/bootstrap_mcps/ingestion-datahub-gc.yaml
```diff
@@ -38,6 +38,12 @@
               max_workers: {{soft_deleted_entities_cleanup.max_workers}}{{^soft_deleted_entities_cleanup.max_workers}}10{{/soft_deleted_entities_cleanup.max_workers}}
               limit_entities_delete: {{soft_deleted_entities_cleanup.limit_entities_delete}}{{^soft_deleted_entities_cleanup.limit_entities_delete}}25000{{/soft_deleted_entities_cleanup.limit_entities_delete}}
               runtime_limit_seconds: {{soft_deleted_entities_cleanup.runtime_limit_seconds}}{{^soft_deleted_entities_cleanup.runtime_limit_seconds}}7200{{/soft_deleted_entities_cleanup.runtime_limit_seconds}}
+            query_cleanup:
+              enabled: {{query_cleanup.enabled}}{{^query_cleanup.enabled}}false{{/query_cleanup.enabled}}
+              retention_days: {{query_cleanup.retention_days}}{{^query_cleanup.retention_days}}90{{/query_cleanup.retention_days}}
```
sgomezvillamor:
```suggestion
              enabled: {{query_cleanup.enabled}}{{^query_cleanup.enabled}}true{{/query_cleanup.enabled}}
              retention_days: {{query_cleanup.retention_days}}{{^query_cleanup.retention_days}}180{{/query_cleanup.retention_days}}
```

aviraj-gour:
Thanks for the review @sgomezvillamor,

**1. PR description mismatch.** You're right, the description was stale from an earlier hard-delete design. I've rewritten it to match the shipped soft-delete-only code.

**2. Dry-run ignores `limit_entities_delete`.** Fixed. `_deletion_limit_reached` used to read `num_queries_soft_deleted`, which stays `0` in a dry run so the cap never tripped. It now takes an explicit `num_candidates_handled` count: the dry-run path counts previewed candidates (via `enumerate`) and the real path passes `num_queries_soft_deleted`. Dry-run previews now stop at `limit_entities_delete` and set `qc_deletion_limit_reached`, so the preview reflects a real capped run. Added `test_dry_run_respects_deletion_limit`.

**Minor:**
- The refactor above also makes the count consistent — the limit check now runs against the count *before* processing each candidate, so the previewed/deleted count matches the cap.
- Documented the `null`-disables-the-cap behavior directly in the `limit_entities_delete` field description (it was only in an internal comment before).

sgomezvillamor (approved):
Both items from my earlier comment are resolved:

- **Description now matches the implementation** — the hard-delete/`max_workers`/worker-pool references are gone, and the config table reflects the actual `QueryCleanupConfig` fields.
- **Dry-run now honors `limit_entities_delete`** — `_deletion_limit_reached(num_candidates_handled)` is passed the enumerate counter in the preview path and `num_queries_soft_deleted` in the real path, so both cap identically; covered by the new `test_dry_run_respects_deletion_limit`. The `null`-disables-the-cap behavior is now documented on the field and covered by `test_uncapped_deletes_all`.

The approach is sound: server-side selection (`source == SYSTEM` AND `lastModifiedAt < cutoff`), a lazy generator with early break that honors both limits, oldest-first ordering, and delegation of the hard-delete second pass to the existing `SoftDeletedEntitiesCleanup`. I confirmed the sorted deep-scroll is stable — `ESUtils.buildSortOrder` appends `urn` ascending as a tiebreaker — and the `qc_`-prefixed report flags avoid colliding with `SoftDeletedEntitiesReport` in the merged `DataHubGcSourceReport`.

The enabled-by-default behavior with a conservative 180-day retention is a reasonable default given that a live query's `lastModifiedAt` refreshes on every re-observation.

LGTM 👍

---

PR #18126, opened 2026-07-01
refactor(ui/document): render Related section entities as Alchemy Pills

annadoesdesign:
## Summary

Refactors the document Related section (`RelatedSection.tsx`) to display related assets and context documents as gray filled Alchemy `Pill` components instead of a vertical list of `EntityLink` rows.

Each pill:

- Shows the platform logo when available, falling back to the entity-type icon if the logo URL is missing or fails to load (`onError` handler).
- Navigates to the entity's profile on click (`history.push`).
- Renders an inline `X` remove button when the user has edit permissions, calling `onRemoveEntity` with `stopPropagation` so the pill-click doesn't fire.

All three visible strings in the section still go through the existing `useTranslation('entity.types')` keys (`document.relatedTitle`, `document.findRelatedPlaceholder`, `document.addRelatedEmpty`), so localized builds are unaffected.

Two small supporting changes travel with the refactor:

- `alchemy-components/components/Pills/Pill.tsx` — pass the pill's `color` prop into the inner `rightIcon` `Button` so the X icon inherits the pill's semantic color instead of always rendering in the default text color. This is what lets the gray-pill X read as gray rather than violet.
- `app/document/utils/documentUtils.ts` — extract a small `categorizeUrns(urns)` helper that splits a URN list into `{ documentUrns, assetUrns }`. Used by `handleConfirmAdd` in `RelatedSection`; replaces the previous inline `forEach` split.

## Checklist

- [x] PR conforms to the Contributing Guideline (Conventional Commits title).
- [ ] Links to related issues (n/a — UI refactor).
- [x] Existing `documentUtils` unit tests still pass (`yarn test src/app/document/utils/__tests__/documentUtils.test.ts`).
- [x] `yarn type-check` and `yarnLintFix` on the three touched files both clean.
- [x] No breaking changes.

### DIFF
--- a/datahub-web-react/src/alchemy-components/components/Pills/Pill.tsx
+++ b/datahub-web-react/src/alchemy-components/components/Pills/Pill.tsx
@@ -76,7 +76,7 @@ export function Pill({
                 : leftIcon && <Icon icon={leftIcon} size={size} onClick={onClickLeftIcon} />}
             <PillText style={customStyle}>{label}</PillText>
             {rightIcon && (
-                <Button style={{ padding: 0 }} variant="text" onClick={onClickRightIcon}>
+                <Button style={{ padding: 0 }} variant="text" color={color} onClick={onClickRightIcon}>
                     <Icon icon={rightIcon} size={size} />
                 </Button>
             )}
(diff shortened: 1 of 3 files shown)


annadoesdesign:
Reopening from within datahub-project/datahub so the base shows as just `master` (not `datahub-project:master`). New PR will link back.

PR #18133, opened 2026-07-01
fix(ingestion/bigquery): eliminate silently-ignored usage.* config in BigQuery connector

kyungsoo-datahub:
- usage.start_time, usage.end_time, usage.bucket_duration, usage.max_query_duration, usage.format_sql_queries, usage.include_top_n_queries, and usage.queries_character_limit were not working before this fix — several were silently ignored, and one (format_sql_queries) was a real bug on the default extraction path. This PR closes out the remaining gaps:

    - Time window forwarding: usage.start_time, usage.end_time, usage.bucket_duration (already fixed in a prior commit on this branch) are forwarded to their top-level equivalents with a deprecation warning, and now apply correctly under both extraction paths. usage.max_query_duration is also forwarded, but unlike the other three, it only takes effect on the legacy path (use_queries_v2: False) — the top-level field itself is legacy-only, so queries-v2 ignores it either way. Setting both the nested and top-level version of the same field is now a configuration error.

    - Queries-v2 wiring: usage.format_sql_queries, usage.include_top_n_queries, and usage.queries_character_limit already worked on the legacy path, which reads them directly. The queries-v2 aggregator (the default, use_queries_v2: True) dropped them entirely and hardcoded format_queries=False. They now flow through to the SQL parsing aggregator used by queries-v2.

    - Legacy-only fields documented: usage.include_read_operational_stats and usage.apply_view_usage_to_tables have no equivalent under queries-v2, which attributes usage via SQL parsing rather than the legacy mechanism these fields rely on. DataHub now logs a warning when either is set to a non-default value while use_queries_v2: True, and their descriptions are updated to clarify they're legacy-only.

Behavior changes:

- A recipe that previously set only usage.start_time, usage.end_time, or usage.bucket_duration will now have it applied at the top level under both extraction paths, which may widen or narrow the range of data ingested. Since BigQuery usage and lineage extraction scan audit logs, a wider window can increase query cost.
- A recipe that previously set only usage.max_query_duration will now have it forwarded to the top level, but it still only affects ingestion under the legacy path (use_queries_v2: False).
- A recipe that sets usage.format_sql_queries: true will now actually reformat SQL queries under the default queries-v2 path (previously silently skipped).
- Recipes using use_queries_v2: True with usage.include_read_operational_stats or usage.apply_view_usage_to_tables set will now see a warning explaining the field is ignored.


sgomezvillamor (approved):
Approving — this is a solid, well-scoped fix for the immediate problem: `usage.*` fields that were silently ignored now either take effect, forward to their top-level equivalents with a deprecation warning, or warn when they're inert. Conflicts raise at parse time, the `format_queries=False` hardcode bug is genuinely fixed, and the coverage + `updating-datahub.md` entries are thorough. I traced the core claim and confirmed nothing reads the nested `usage.start_time/end_time/bucket_duration/max_query_duration` on any path (legacy `usage.py`/`lineage.py` read the top-level copies), so forwarding-and-deprecating them is safe. The `start_time = Field(default=None)  # type: ignore` redeclaration also mirrors the base `BaseTimeWindowConfig` idiom, and the inherited `default_start_time` validator recomputes a real datetime — so it's never actually `None`.

Two small, non-blocking things worth a follow-up:

1. **Top-level `max_query_duration` under queries-v2 gets no warning.** It's legacy-only in effect (only `usage.py`/`lineage.py` read it; `queries_extractor.py` never references it), yet the deprecation text steers users to set it at the top level. On the default `use_queries_v2: true`, a user who follows that advice gets zero signal it's a no-op. Consider extending `warn_legacy_only_usage_fields_under_queries_v2` to also check top-level `max_query_duration != default`, mirroring the two sibling fields.

2. **Seam test gap.** `test_bigquery_source_builds_queries_extractor_config_from_usage_fields` asserts the new fields but not `include_operations=self.config.usage.include_operational_stats` or `user_email_pattern`, which the same `_build_queries_extractor_config` maps — a typo there would silently re-introduce exactly this bug class. Two extra assertions close it.

Optional nits: the two legacy-only fields warn in both `warn_legacy_only_usage_fields_under_queries_v2` (`logger.warning`) and `_warn_deprecated_configs` (`report_warning`, which also logs), so they emit twice to the log; and `f"{BaseTimeWindowConfig.model_fields['start_time'].description} ..."` would render literal `"None ..."` if a base description ever became `None` (`... or ''` guards it). Neither blocks — happy for these to be a follow-up.

---

sgomezvillamor:
Stepping back from the diff — this PR is a nice illustration of a deeper structural thing worth flagging for the roadmap (explicitly **not** for this PR, which is the right incremental fix).

The reason some `usage.*` fields apply on one path but not the other traces to two coexisting usage strategies:

- **queries-v2** (`use_queries_v2: true`, the default): a single `SqlParsingAggregator` fed from `INFORMATION_SCHEMA.JOBS`, producing lineage + usage + operations + queries together.
- **legacy** (`use_queries_v2: false`): usage goes through a separate `BigQueryUsageState` / direct-aggregation engine (`usage.py`, no aggregator), while lineage actually *already* uses `SqlParsingAggregator` (`lineage.py:247`). `use_exported_bigquery_audit_metadata` isn't a top-level path selector — it's a sub-branch of the legacy path choosing the audit-log *source* (exported `BigQueryAuditMetadata` tables vs. the GCP Logging API).

So `apply_view_usage_to_tables` and `include_read_operational_stats` are legacy-only precisely because they're features of that direct-aggregation engine, with no analog in the aggregator's SQL-parse-based usage attribution. queries-v2 is effectively the convergence target already, and it's the default.

The concrete blockers to eventually retiring the legacy path (and making every `usage.*` field apply uniformly) look like:

1. **`use_exported_bigquery_audit_metadata`** — the main one. queries-v2 is hardwired to `INFORMATION_SCHEMA.JOBS` and has no alternative audit-log source. Converging would mean sourcing `ObservedQuery` rows from exported audit metadata and feeding the same aggregator.
2. **Read/operational stats + view-usage attribution** — decide whether to implement these in the aggregator or intentionally drop them.
3. **Legacy stateful ingestion** (`RedundantLineage/UsageRunSkipHandler`) vs. queries-v2's `enable_stateful_time_window`.

Closing gap #1 in particular would let this whole nest of "applies on path X but not Y" deprecation warnings eventually collapse into "applies everywhere." None of it belongs here — just flagging the direction, since this PR is chipping away at exactly that seam. Happy to help scope it if there's interest.

---

sgomezvillamor:
@kyungsoo-datahub how [that](https://github.com/datahub-project/datahub/pull/18133/#issuecomment-4875110707) sounds for a long term enhancement?

PR #18134, opened 2026-07-02
feat(metrics): add new metric and semantic model entities

ani-malgari:
## Summary

Introduces `metric` and `semanticModel` as first-class DataHub entities. This is the
schema-layer foundation for the upcoming Metrics Catalog feature (already scaffolded
in the frontend behind the `metricsEnabled` feature flag). 

The model follows the "lean OSI-core" shape from the internal Metrics RFC: OSI-shaped
core fields plus a vendor-namespaced `customExtensions` bag for anything
platform-specific, rather than promoting every platform quirk to a first-class field.
Governance and lineage reuse existing native aspects (`ownership`, `domains`,
`upstreamLineage`, etc.).

## Entities

**`metric`** — key `(platform, id)` → `urn:li:metric:(dbt,total_revenue)`.

Aspects:
- `metricInfo` — name, description, semantic-model ref, multi-dialect SQL expression,
  AI context, recoverability, vendor extensions.
- `metricRelationships` — `parentMetric` (glossary-style tree pointer, powers the
  sidebar), `derivedFrom` (metric→metric lineage via `Edge` with `isLineage: true`),
  `relatedMetrics` (curated "see also" edges, non-lineage).
- Native aspects: `upstreamLineage`, `ownership`, `domains`, `globalTags`,
  `glossaryTerms`, `institutionalMemory`, `structuredProperties`, `status`,
  `deprecation`, `lifecycleStage`, `dataPlatformInstance`, `subTypes`, `forms`,
  `testResults`, `documentation`, `browsePaths`, `browsePathsV2`, `applications`,
  `container`, `incidentsSummary`, `displayProperties`, `assetSettings`,
  `versionProperties`, `access`.

**`semanticModel`** — key `(platform, id)` → `urn:li:semanticModel:(dbt,sales_orders)`.

Aspects:
- `semanticModelInfo` — name, description, `sourcePlatform` (URN, mirrors
  `datasetKey.platform`), AI context, `datasets[]` (with fields and dimensions),
  join `relationships[]`, vendor extensions.
- Same governance aspects as above (minus lineage/incident-oriented ones).

Both keys treat `platform` as required and immutable (part of the URN). If the source
platform is unknown at ingest time, the ingestion layer is expected to assign a stable
fallback (e.g. `datahub` for native / SDK-authored metrics).

## Design notes

- **`platform` on the key, not `dataPlatformInstance`.** Platform is a structural
  identity partition ("this is a Snowflake metric vs. a dbt metric — they are
  distinct entities per the MVP no-dedup decision"), whereas
  `dataPlatformInstance` is a decorative aspect for search faceting and policy
  scoping. Same distinction datasets already make.
- **Plain `Urn` fields, not typed URN typerefs.** Cross-entity references
  (`MetricInfo.semanticModel`, `MetricRelationships.parentMetric`, etc.) use plain
  `Urn` with `@Relationship.entityTypes` doing the type gating. Matches the
  `Domain` / `DataProduct` pattern; can promote to typed URNs later without an
  aspect version bump.

## Out of scope (follow-up PRs)

- GraphQL types + resolvers, `EntityType` enum entries, `EntityTypeMapper` /
  `UrnToEntityMapper` wiring.
- Frontend `MetricEntity` / `SemanticModelEntity` registered in
  `buildEntityRegistryV2`. Stub `MetricsPage.tsx` already tracks this.
- Ingestion sources (dbt semantic layer, Snowflake semantic views, Databricks
  metric views, OSI upload).
- SQL compiler / recoverability computation.
- Policy / access-control specifics.

## Checklist

- [x] No related public issue; internal RFC drives the design
- [x] Tests: N/A (schema-only; validated via codegen and registry build)
- [x] Docs added: `metadata-models/docs/entities/metric.md` and
      `metadata-models/docs/entities/semanticModel.md`
- [x] Not a breaking change (net-new entities; `metricsEnabled` flag defaults `false`)


metadata-models/src/main/pegasus/com/linkedin/metadata/key/MetricKey.pdl
```diff
@@ -0,0 +1,29 @@
...
+  /**
+   * The platform or semantic layer that owns this metric (e.g. "dbt",
+   * "snowflake"). Used to group metrics by platform in search facets.
+   *
+   * REQUIRED. This field is part of the metric URN and therefore
+   * immutable once written. Callers must always populate it; if the
+   * source platform is unknown at ingest time the ingestion layer is
+   * responsible for assigning a stable fallback value (e.g. "datahub"
+   * for native / SDK-authored metrics)
+   */
```
ani-malgari: this is the part i need alignment for. @gabe-lyons @asikowitz @chriscollins3456 - do you see any issues with this? Having platform preserves stateless idempotent ingestion and easy sync.

metadata-models/docs/entities/metric.md
```diff
@@ -0,0 +1,82 @@
...
+The metric entity represents a named, reusable business measurement defined in a semantic layer or
+metric store. Metrics are platform-independent: the same logical measure (e.g. `total_revenue`) can
+be ingested from dbt, Snowflake, Databricks, or any other platform as a separate entity, with
+platform identity carried by the `dataPlatformInstance` aspect.
+
+## Identity
+
+Metrics are identified by two fields:
+
+- **`namespace`** — typically the platform or project name (e.g. `dbt`, `snowflake`, `my_project`).
```
gabe-lyons: imo if we are having people truly say dbt or snowflake we should at least support an optional platformurn. it feels wrong to put in platforms as unstructured text

metadata-models/docs/entities/metric.md
```diff
@@ -0,0 +1,82 @@
...
+  Searchable as a keyword so the sidebar can group metrics by platform.
+- **`id`** — the metric name within that namespace (e.g. `total_revenue`, `daily_active_users`).
```
gabe-lyons: how are we handling nested metrics? are we letting metrics be nested? thinking about how we deal with glossary, where we support multi-layer nesting, it feels odd to not support that for metrics. is there a particular reason to not support multi layer nesting for metrics?

metadata-models/docs/entities/metric.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Relationships with Other Entities
+
+| Relationship  | Direction | Target entity   | Aspect / edge name    |
+| ------------- | --------- | --------------- | --------------------- |
+| ModeledBy     | outbound  | `semanticModel` | `metricInfo`          |
+| IsPartOf      | outbound  | `metric`        | `metricRelationships` |
+| DerivedFrom   | outbound  | `metric`        | `metricRelationships` |
+| RelatedTo     | outbound  | `metric`        | `metricRelationships` |
+| Upstream data | outbound  | `dataset`       | `upstreamLineage`     |
```
gabe-lyons: generally this would be schemaField right, not datasets?

metadata-models/docs/entities/semanticModel.md
```diff
@@ -0,0 +1,81 @@
...
+between raw datasets and the business metrics calculated over them.
+
+## Identity
+
+Semantic models are identified by two fields:
+
+- **`namespace`** — typically the platform or project name (e.g. `dbt`, `snowflake`, `my_project`).
+- **`id`** — the model name within that namespace (e.g. `orders_model`, `customer_360`).
+
+An example URN: `urn:li:semanticModel:(dbt,orders_model)`.
```
gabe-lyons: i think we either should remove data platforms like dbt or snowflake from namespace examples or use platform urns with data platform instances. this middle ground is worst of both worlds.

metadata-models/src/main/pegasus/com/linkedin/metadata/key/MetricKey.pdl
```diff
@@ -0,0 +1,29 @@
...
+  @Searchable = {
+    "fieldType": "KEYWORD"
+  }
+  platform: string
+
+  /**
+  * Unique identifier for the metric within its platform.
+  */
+  id: string
```
gabe-lyons, line 45: see my comment above

metadata-models/src/main/pegasus/com/linkedin/metadata/key/SemanticModelKey.pdl
```diff
@@ -0,0 +1,29 @@
...
+  * REQUIRED. This field is part of the metric URN and therefore
+  * immutable once written. Callers must always populate it; if the
+  * source platform is unknown at ingest time the ingestion layer is
+  * responsible for assigning a stable fallback value (e.g. "datahub"
+  * for native / SDK-authored metrics)
+  */
+  @Searchable = {
+    "fieldType": "KEYWORD"
+  }
+  platform: string
```
gabe-lyons: this should align with datasets because semantic models should always be linked to a platform

metadata-models/src/main/pegasus/com/linkedin/metadata/key/MetricKey.pdl
```diff
@@ -0,0 +1,29 @@
...
+   * REQUIRED. This field is part of the metric URN and therefore
+   * immutable once written. Callers must always populate it; if the
+   * source platform is unknown at ingest time the ingestion layer is
+   * responsible for assigning a stable fallback value (e.g. "datahub"
+   * for native / SDK-authored metrics)
+   */
+  @Searchable = {
+    "fieldType": "KEYWORD"
+  }
+  platform: string
```
gabe-lyons: we can do something like `platform: Urn` and have one option be logical, similar to what we did with logical datasets, but then we'd need another namespace string

metadata-models/src/main/pegasus/com/linkedin/metric/CustomExtension.pdl
```diff
@@ -0,0 +1,19 @@
+namespace com.linkedin.metric
+
+/**
+ * A vendor-namespaced extension blob that carries platform-specific metric metadata
+ * not expressible in the OSI core schema (Proposal B §B.5).
+ * The content field must be a valid JSON string.
+ */
+record CustomExtension {
```
gabe-lyons: why not just use structured properties? this seems like an overly complex 1-off

metadata-models/src/main/pegasus/com/linkedin/metric/Dialect.pdl
```diff
@@ -0,0 +1,13 @@
...
+ * The SQL dialect or semantic layer platform that produced a metric expression.
+ */
+enum Dialect {
+  SNOWFLAKE
+  DATABRICKS
+  DBT
+  ANSI_SQL
+  DATAHUB
+  UNKNOWN
+}
```
gabe-lyons, line 24: why do we need both platform and dialect, this seems redundant

gabe-lyons: imagine now we'll need to link dialects to icons, dialects to human readable names, etc, this is the same set as data platforms.

metadata-models/src/main/pegasus/com/linkedin/metric/DialectExpression.pdl
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * A metric expression in a specific SQL dialect or semantic layer language.
+ */
+record DialectExpression {
+
+  /**
+   * The dialect in which this expression is written.
+   */
+  dialect: Dialect
```
gabe-lyons, line 11: see comment above this can just be platform

metadata-models/src/main/pegasus/com/linkedin/metric/MetricRelationships.pdl
```diff
@@ -0,0 +1,53 @@
...
+import com.linkedin.common.Edge
+import com.linkedin.common.Urn
+
+/**
+ * Relationships from a Metric to other Metric entities (Proposal B §B.4).
+ */
+@Aspect = {
+  "name": "metricRelationships"
+}
+record MetricRelationships {
```
gabe-lyons, line 12: im concerned about us baking in the set of possible relationships. @asikowitz is doing some work to enable custom relationship types, this work is probably just a subset of that.

metadata-models/src/main/pegasus/com/linkedin/semanticmodel/Field.pdl
```diff
@@ -0,0 +1,41 @@
+namespace com.linkedin.semanticmodel
+
+import com.linkedin.metric.AiContext
+import com.linkedin.metric.CustomExtension
+import com.linkedin.metric.MetricExpression
+
+/**
+ * A named field (measure or dimension) defined inside a SemanticModel dataset.
+ */
+record Field {
```
gabe-lyons: i notice this doesn't seem to have a type? like numeric, string, time, etc?

gabe-lyons: will people want to add glossary terms, tags, etc to this?

gabe-lyons: I'm concerned we're rebuilding schemaField object here and will need to eventually keep adding all the stuff we have in that model.

metadata-models/src/main/pegasus/com/linkedin/semanticmodel/ModelDataset.pdl
```diff
@@ -0,0 +1,45 @@
+namespace com.linkedin.semanticmodel
+
+import com.linkedin.common.Urn
+import com.linkedin.metric.CustomExtension
+
+/**
+ * A logical dataset referenced by a SemanticModel, together with its schema
+ * and primary/unique key definitions.
+ */
+record ModelDataset {
```
gabe-lyons, line 8: whoa why is this not just a datasetUrn?

metadata-models/src/main/pegasus/com/linkedin/semanticmodel/SemanticModelInfo.pdl
```diff
@@ -0,0 +1,73 @@
...
+  aiContext: optional AiContext
+
+  /**
+   * The logical datasets that this semantic model exposes.
+   */
+  datasets: array[ModelDataset] = [ ]
+
+  /**
+   * Join relationships between the datasets of this semantic model.
+   */
```
gabe-lyons, line 84: i think we already model join relationships

metadata-models/src/main/pegasus/com/linkedin/semanticmodel/SemanticModelInfo.pdl
```diff
@@ -0,0 +1,73 @@
...
+  relationships: optional array[SemanticModelRelationship]
+
+  /**
+   * Vendor-namespaced extension blobs for platform-specific metadata.
+   */
+  customExtensions: optional array[CustomExtension]
```
gabe-lyons: again i think this should just be structured properties

metadata-models/src/main/pegasus/com/linkedin/semanticmodel/SemanticModelRelationship.pdl
```diff
@@ -0,0 +1,46 @@
+namespace com.linkedin.semanticmodel
+
+import com.linkedin.metric.AiContext
+import com.linkedin.metric.CustomExtension
+
+/**
+ * A join relationship between two logical datasets defined within a SemanticModel.
+ * Named SemanticModelRelationship to avoid colliding with common relationship models.
+ */
+record SemanticModelRelationship {
```
gabe-lyons, line 10: pleaes confirm no other such model exists

metadata-models/docs/entities/metric.md
ani-malgari: we will be supporting nested metrics, it's handled in the [MetricRelationships.pdl](https://github.com/datahub-project/datahub/pull/18134/changes/BASE..616b651801e265d8b1fb530cdc5489aded206b0d#diff-ec6ac2202d98ef6117466c43a3110b7bb62404b0252eea82dd5e2f458a6455e7)

metadata-models/src/main/pegasus/com/linkedin/metric/CustomExtension.pdl
ani-malgari: we're using CustomExtension for 1:1 parity with OSI model, and ease of maintenance. Also CustomExtensions are embedded inside multiple nested fields within the SemanticModel (Datasets, relationships, Fields, metrics). We want to support bidirectional sync where Datahub is the source of truth, and maintaining the OSI structure will be easy to handle. Also for ingestion, most of the OSI-supported platforms have connectors which will help get to this structure.

metadata-models/src/main/pegasus/com/linkedin/metric/Dialect.pdl
ani-malgari: `dialect` again is part of [OSI schema](https://github.com/open-semantic-interchange/OSI/blob/main/core-spec/spec.md#semantic-model) and relates to the metric expression. A metric's expression can be written in any of the above formats, and this enum only relates to the expression (no linking to icons or platform metadata). 

Also expression in multiple dialects for a single metric can be ingested.

metadata-models/src/main/pegasus/com/linkedin/metric/MetricRelationships.pdl
ani-malgari: sure, i'm happy to hear more from @asikowitz. The purpose of this aspect is to support nested metrics (tree), derived metrics (lineage) and support edge metadata between metrics.

metadata-models/src/main/pegasus/com/linkedin/metadata/key/MetricKey.pdl
sgomezvillamor, line 1: both MetricKey and SemanticModelKey, should `env: FabricType` be included in the key also? the same as we do for datasets

metadata-models/src/main/pegasus/com/linkedin/metric/MetricInfo.pdl
sgomezvillamor, line 1: what about typical audit info such as createdAt or lastUpdatedAt?

metadata-models/src/main/pegasus/com/linkedin/semanticmodel/SemanticModelInfo.pdl
sgomezvillamor, line 1: what about typical audit info such as createdAt or lastUpdatedAt?

metadata-models/src/main/pegasus/com/linkedin/semanticmodel/SemanticModelInfo.pdl
sgomezvillamor: alternatively, this could also be the generic CustomProperties aspect that most entities "inherit" in the Info/Properties aspect

sgomezvillamor:
How these new entities will be linked to the existing entities?

Eg, should `Upstream` be updated to include `SemanticModel` entities too?
https://github.com/datahub-project/datahub/blob/15462bec1cbd45eb31230778bbe9ddc16dbe2a06/metadata-models/src/main/pegasus/com/linkedin/dataset/Upstream.pdl#L26-L48

metadata-models/src/main/pegasus/com/linkedin/semanticmodel/Field.pdl
ani-malgari: Good callout. The Field here is the OSI field which is a dimension or a fact: a scalar, row-level attribute defined by its expression over one or more physical columns. 
Ex. a derived dimension `DATE_TRUNC('month', orders.order_date)` or `first_name || ' ' || last_name`. 

we don't duplicate SchemaField: a field carries only the semantic enrichment (expression, dimension/is_time, label, ai_context) and not the physical attributes. Those stay on SchemaField.

metadata-models/docs/entities/metric.md
ani-malgari: it will be datasets as well as SchemaField. Ex: `upstreamLineage: { upstreams: [orders], fineGrained: [orders.amount] }`

metadata-ingestion/src/datahub/ingestion/autogenerated/lineage.json
```diff
@@ -169,6 +169,42 @@
         ]
       }
     },
+    "metric": {
+      "upstreamLineage": {
+        "aspect": "upstreamLineage",
```
asikowitz: This is a little confusing to me in terms of language -- I was kinda expecting DerivedFrom to mean the datasets from which a metric is derived. But instead DerivedFrom == metrics from which a metric is derived, which is also different from parent metric (which could also mean metrics from which a metric is derived)

Also, the `UpstreamLineage` aspect includes fine grained lineages... will metrics have a concept of "columns" or "schema fields"? If not, I don't think this is appropriate. And note that currently the comment for upstream lineage aspect is "Upstream lineage of a dataset"

metadata-models/src/main/pegasus/com/linkedin/semanticmodel/ModelDataset.pdl
ani-malgari: ModelDataset contains semantic enrichment (alias name, dimensions etc), and it's source of truth is the semantic model.

metadata-models/docs/entities/metric.md
```diff
@@ -0,0 +1,94 @@
+# Metric
+
+The metric entity represents a named, reusable business measurement defined in a semantic layer or
+metric store. Metrics are platform-independent: the same logical measure (e.g. `total_revenue`) can
+be ingested from dbt, Snowflake, Databricks, or any other platform as a separate entity, with
+platform identity carried by the `dataPlatformInstance` aspect.
```
asikowitz: This seems in contrast with below where platform identity is on the metric urn

metadata-models/docs/entities/metric.md
```diff
@@ -0,0 +1,94 @@
...
+- **`description`** — free-text description of what the metric measures.
+- **`expression`** — the metric formula expressed in one or more SQL dialects (SNOWFLAKE,
+  DATABRICKS, DBT, ANSI_SQL, DATAHUB, UNKNOWN). Each `DialectExpression` pairs a `Dialect` enum
+  value with the raw SQL string.
+- **`aiContext`** — optional hints for AI/LLM consumers: synonyms, natural-language instructions,
+  few-shot examples, and custom instructions.
+- **`recoverability`** — searchable enum (FULL / PARTIAL / NONE) indicating whether the metric
+  value can be recomputed from raw data.
+- **`semanticModel`** — URN of the `semanticModel` entity that defines this metric's dimensional
+  context. Stored as a `ModeledBy` graph edge.
```
asikowitz: Can you discuss what it means for this to be null?

metadata-models/docs/entities/metric.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Metric Relationships
+
+Hierarchical and derivation relationships are stored in the `metricRelationships` aspect:
+
+- **`parentMetric`** — URN of the parent metric for tree-style organisation (sidebar roots are
+  metrics where `hasParentMetric=false`). Stored as an `IsPartOf` graph edge.
```
asikowitz: Imo this doc is getting a little too technical for an overview of the entity

metadata-models/docs/entities/metric.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Governance and Lineage
+
+The metric entity reuses the full set of standard governance aspects:
+
+`ownership`, `domains`, `globalTags`, `glossaryTerms`, `institutionalMemory`,
+`structuredProperties`, `status`, `deprecation`, `dataPlatformInstance`,
+`subTypes`, `forms`, `testResults`, `documentation`, `browsePaths`, `browsePathsV2`,
+`applications`, `container`, `incidentsSummary`, `displayProperties`, `assetSettings`,
+`versionProperties`, `access`, `upstreamLineage`.
```
asikowitz: Imo we should be more careful about what aspects we want to support. Are metrics getting a configurable color? Otherwise why do we need `displayProperties`. Do they need `browsePaths` when that aspect is deprecated? Are they going to be supported immediately by compliance forms or metadata tests? Will they be versioned? And I see `access` only on datasets and containers -- not sure that belongs either

metadata-models/docs/entities/metric.md
```diff
@@ -0,0 +1,94 @@
...
+## Relationships with Other Entities
+
+| Relationship                 | Direction | Target entity   | Aspect / edge name                    |
+| ---------------------------- | --------- | --------------- | ------------------------------------- |
+| ModeledBy                    | outbound  | `semanticModel` | `metricInfo`                          |
+| IsPartOf                     | outbound  | `metric`        | `metricRelationships`                 |
+| DerivedFrom                  | outbound  | `metric`        | `metricRelationships`                 |
+| RelatedTo                    | outbound  | `metric`        | `metricRelationships`                 |
+| Upstream data (table-level)  | outbound  | `dataset`       | `upstreamLineage.upstreams`           |
+| Upstream data (column-level) | outbound  | `schemaField`   | `upstreamLineage.fineGrainedLineages` |
```
asikowitz: Hmm I would prefer to just reference upstream datasets and schema fields rather than reuse this. Fine grained lineage is for column -> column which is not the case here right. Or if this is saying schema field -> semantic model dataset field then let's hash out exactly how that'll work a bit more first

metadata-models/docs/entities/metric.md
```diff
@@ -0,0 +1,94 @@
...
+
+Both granularities live in the same standard `upstreamLineage` aspect. `upstreams[]` carries
+`metric → dataset` edges (table-level), while `fineGrainedLineages[]` carries
+`metric → schemaField` edges and supports `transformOperation` and `confidenceScore` metadata.
```
asikowitz: Ok I see, these are supposed to be downstream. You can't use upstream lineage for that... that's for adding upstreams of the entity on which you're putting the aspect

metadata-models/src/main/pegasus/com/linkedin/metric/MetricRelationships.pdl
```diff
@@ -0,0 +1,53 @@
...
+   * Marked isLineage so the lineage graph includes these edges.
+   */
+  @Relationship = {
+    "/*/destinationUrn": {
+      "name": "DerivedFrom",
+      "entityTypes": [ "metric" ],
+      "isLineage": true
+    }
+  }
+  derivedFrom: array[Edge] = [ ]
```
asikowitz: So we can attach properties with this but not sure if we can replace this with another record in the future, e.g. `record MetricAssociation includes Edge { ... }` where we could add more fields. Could you ask platform team if this would be a backwards compatible? Still, think this is preferred over just an urn list.

asikowitz (commented):
Mainly confused / concerned about all the aspects being included + how lineage with datasets / schema fields is supposed to work

metadata-models/src/main/pegasus/com/linkedin/metric/CustomExtension.pdl
ani-malgari: this is updated, we're no longer using CustomExtensions.

metadata-models/src/main/pegasus/com/linkedin/semanticmodel/SemanticModelInfo.pdl
ani-malgari: this is removed

metadata-models/src/main/pegasus/com/linkedin/metric/MetricInfo.pdl
ani-malgari: makes sense, i'll account for the 2 fields

metadata-models/docs/entities/metric.md
ani-malgari: this is updated, we're using a new aspect metricUpstreams to determine source datasets and columns

metadata-models/src/main/pegasus/com/linkedin/semanticmodel/SemanticModelInfo.pdl
ani-malgari: we have first-class `ERModelRelationship` entity, while this aspect is structurally owned by semantic model and purely serves for the semantic layer. A semantic model can rename or re-scope its logical datasets independently of the underlying physical URNs. Forcing semantic-model joins to reference existing erModelRelationship entities would require two-step ingestion.

metadata-models/src/main/pegasus/com/linkedin/metadata/key/MetricKey.pdl
ani-malgari: purposefully not attaching FabricType per @gabe-lyons 's suggestion. i don't foresee to have same metric being ingested by different environments.

metadata-models/src/main/pegasus/com/linkedin/metric/MetricInfo.pdl
ani-malgari: done.

metadata-models/docs/entities/metric.md
ani-malgari: updated

metadata-models/docs/entities/metric.md
ani-malgari: when the metric was ingested without a semantic model context (e.g. thin catalog-only metrics from BI tools like Tableau). Updated the md file

metadata-models/docs/entities/metric.md
ani-malgari: agreed to the bloat, trimmed down a lot of these properties.

metadata-models/docs/entities/metric.md
ani-malgari: Yes, as discussed in the call, we have a new aspect `MetricUpstreams` for lineage to source datasets

metadata-models/src/main/pegasus/com/linkedin/metric/MetricRelationships.pdl
ani-malgari: After discussing with platform, using a custom type in place of direct Edge for future proofing. thanks for the callout!

metadata-ingestion/src/datahub/ingestion/autogenerated/lineage.json
ani-malgari: As discussed offline, we're not using `UpstreamLineage`, and replaced it with `MetricUpstreams` aspect that covers dataset and schema field lineage.

metadata-models/src/main/pegasus/com/linkedin/metric/MetricInfo.pdl
```diff
@@ -0,0 +1,87 @@
...
+  @Relationship = {
+    "name": "ModeledBy",
+    "entityTypes": [ "semanticModel" ]
+  }
+  @Searchable = {
+    "fieldName": "semanticModel",
+    "fieldType": "URN",
+    "hasValuesFieldName": "hasSemanticModel"
+  }
+  semanticModel: optional Urn
```
gabe-lyons, line 75: i think this is the wrong direction. I would instead attach a metric to semantic models. What if a metric exists in two semantic models?

metadata-models/src/main/pegasus/com/linkedin/metric/MetricRelationships.pdl
gabe-lyons: @asikowitz ?

metadata-models/src/main/pegasus/com/linkedin/metric/MetricUpstreams.pdl
```diff
@@ -0,0 +1,65 @@
...
+import com.linkedin.common.Edge
+
+/**
+ * Physical data-flow lineage from a metric to the datasets and columns it reads.
+ * Metric-to-metric derivation lineage lives on `metricRelationships.derivedFrom`.
+ */
+@Aspect = {
+  "name": "metricUpstreams"
+}
+record MetricUpstreams {
```
gabe-lyons, line 12: is this just a copy of upstreamLineage that datasets use? why do we need to create a whole new aspect for this

metadata-models/src/main/pegasus/com/linkedin/semanticmodel/Field.pdl
gabe-lyons: can we call it SemanticField or something? Field will show up in all code search

metadata-models/src/main/pegasus/com/linkedin/semanticmodel/ModelDataset.pdl
gabe-lyons: i'm not convinced by this answer...in snowflake i'm fairly sure semantic datasets are just datasets referenced by semantic models. I am really not convinced this is its own entity

metadata-models/src/main/pegasus/com/linkedin/semanticmodel/SemanticModelInfo.pdl
gabe-lyons: I'm struggling to follow this argument as well... it feels like here and above we're making nearly identical models for things that already exist but call them "Semantic" versions. this feels off.

metadata-models/src/main/pegasus/com/linkedin/metric/MetricInfo.pdl
ani-malgari: A metric can only belong to 1 semantic model, that's the assumption we've made cc: @maggiehays . If the same metric name appears in 2 semantic models, then they're treated as separate metrics, as the assumption is the underlying platform/datasets/SQL is different.

metadata-models/src/main/pegasus/com/linkedin/metric/MetricUpstreams.pdl
ani-malgari: After discussing with @asikowitz yesterday, i've understood that `upstreamLineage` was primarily created for dataset <> dataset lineage. It includes `finegrainedLineage` which only is relevant for column <> column lineage. Metric doesn't have columns like datasets, what we need is Metric <> Dataset and Metric <> Column lineage, which is what `MetricUpstreams` serve.

metadata-models/src/main/pegasus/com/linkedin/semanticmodel/ModelDataset.pdl
ani-malgari: ModelDataset is just a wrapper that pairs the dataset URN with the enrichment that belongs to this specific semantic model. It carries three things beyond the URN:

1. name: the logical alias the semantic model uses in its FACTS / DIMENSIONS / METRICS / RELATIONSHIPS clauses. You're right that in Snowflake SV grammar the alias always ends up matching the physical table name (Snowflake doesn't let you rename). But dbt semantic models let you alias freely (orders over raw.stripe_orders_v3). Even in Snowflake, the alias is the identifier the model uses in expressions, so it needs a home the compiler can resolve.
2. primaryKey / uniqueKeys: semantic-model-scoped key declarations. Two models can assert different keys over the same table.
3. fields: array[Field]: the subset of columns the semantic model chooses to expose, each with its own expression, dimension classification (regular / time / not-a-dimension), description, and AI hints. The underlying dataset can have 100 columns; the semantic view can expose 30, with renamed facts, computed expressions that aren't physical columns at all (e.g. ACTIVE_CUSTOMERS_ONLY as is_churned = FALSE in the MAU SV - that's model metadata, not dataset metadata).

All of that is semantic-model-scoped enrichment, not dataset-scoped. The same physical DIM_CUSTOMER can appear in three different semantic models with three different alias names, exposed field lists, key declarations, and time-dimension classifications. If we collapsed to datasets: array[Urn] and hung field-level enrichment off the dataset entity, the three models would clobber each other.

metadata-models/src/main/pegasus/com/linkedin/semanticmodel/Field.pdl
```diff
@@ -0,0 +1,36 @@
...
+
+  /**
+   * The name of this field as exposed by the semantic layer.
+   */
+  name: string
```
gabe-lyons: field: SchemaField

/**
   * The SQL expression that computes this field, in one or more dialects.
   */
  expression: MetricExpression

  /**
   * Dimension classification metadata; absent if this field is a measure.
   */
  dimension: optional Dimension

  /**
   * AI-specific context for improved disambiguation and retrieval.
   */
  aiContext: optional AiContext

metadata-models/src/main/pegasus/com/linkedin/semanticmodel/ModelDataset.pdl
```diff
@@ -0,0 +1,40 @@
...
+  name: string
+
+  /**
+   * URN of the DataHub Dataset entity that backs this logical dataset.
+   */
+  @Relationship = {
+    "name": "SourcedBy",
+    "entityTypes": [ "dataset" ]
+  }
+  source: Urn
```
ani-malgari, line 29: question for @kyungsoo-datahub: the source can be a physical table, view, or a derived/inline query.  For derived/inline query case, can we mint a new "query" entity during ingestion when we encounter an inline query? cc: @gabe-lyons

ani-malgari: Reference: https://docs.databricks.com/aws/en/business-semantics/metric-views/basic-modeling#use-a-sql-query-as-a-source

metadata-models/docs/entities/metric.md
```diff
@@ -0,0 +1,92 @@
...
+layer or metric store.
+
+## Identity
+
+Metrics are identified by two fields:
+
+- **`platform`** — the DataPlatform URN that owns this metric
+  (e.g. `urn:li:dataPlatform:dbt`, `urn:li:dataPlatform:snowflake`). Searchable as a URN field
+  with autocomplete and a "Platform" filter pill.
+- **`id`** — the metric name within that platform (e.g. `total_revenue`, `daily_active_users`).
```
kyungsoo-datahub: How is id uniqueness scoped? The examples use bare names like urn:li:metric:(urn:li:dataPlatform:snowflake,total_revenue). But two teams can define total_revenue in different semantic views on the same Snowflake instance, producing identical URNs.

metadata-models/src/main/pegasus/com/linkedin/metric/Dialect.pdl
```diff
@@ -0,0 +1,20 @@
+namespace com.linkedin.metric
+
+/**
+ * The SQL or expression-language dialect that produced a metric expression.
+ * Value set is aligned 1:1 with the OSI (Open Semantic Interchange) spec.
+ */
+enum Dialect {
```
kyungsoo-datahub, line 7: The enum is closed with no OTHER/UNKNOWN fallback. Dialects outside this set exist today and OSI itself will keep evolving. But adding enum symbols later is a rolling-upgrade hazard. Can we add OTHER plus an optional freeform dialectName: string on DialectExpression, so new dialects don't require a schema change? The OSI-aligned named values can stay as-is.

metadata-models/src/main/pegasus/com/linkedin/semanticmodel/ModelDataset.pdl
```diff
@@ -0,0 +1,40 @@
...
+
+  /**
+   * Logical alias used to reference this dataset within the semantic model.
+   */
+  name: string
+
+  /**
+   * URN of the DataHub Dataset entity that backs this logical dataset.
+   */
+  @Relationship = {
```
kyungsoo-datahub, line 25: We don't need `"isLineage": true` here?

metadata-models/src/main/pegasus/com/linkedin/metric/Dialect.pdl
ani-malgari: Good point. i added `OTHER` and since there are a limited number of dialects, i think can strongly enforce type safety and do not think we need to support the freeform string separately. let me know if you see any issues.

metadata-models/src/main/pegasus/com/linkedin/semanticmodel/ModelDataset.pdl
ani-malgari: the lineage only supports from metric to dataset(s)

metadata-models/docs/entities/metric.md
ani-malgari: Great point, i added `path` field to both semantic model and metric entities, similar to MLFeatureKey/MLPrimaryKey.  
Examples:
urn:li:semanticModel:(urn:li:dataPlatform:snowflake,ANALYTICS_HUB.PUBLIC_ANALYTICS,MAU_SNOWFLAKE_SUMMIT_SV)
urn:li:metric:(urn:li:dataPlatform:dbt,analytics.orders_model,revenue)

Let me know if this makes sense.

metadata-models/src/main/pegasus/com/linkedin/semanticmodel/SemanticModelInfo.pdl
ani-malgari: Why it isn't a repackaged ERModelRelationship:

1. SemanticModelRelationship joins logical aliases (ModelDataset.name) on column names, not dataset + schemaField URNs. That's load-bearing for:
- Self-joins — one semantic model can alias EMPLOYEES twice (employees_a / employees_manager) and join between them; URN-typed edges collapse both sides and lose the distinction.
- Query-backed sources — ModelDataset.source now also accepts query URNs (inline SQL like Databricks metric views), so a logical dataset may have no backing dataset URN.

2. ERModelRelationship is a standalone governed entity; a semantic-model join has no independent identity and dies with the model. Reusing it would force two-step ingestion + orphan management for pure sub-structure.

metadata-models/src/main/pegasus/com/linkedin/semanticmodel/ModelDataset.pdl
ani-malgari: @gabe-lyons - i trimmed this model, getting rid of primaryKey and uniqueKeys in favor of only using fields that will be exposed in the UI. We already have a `sourceDefinition` that supports bi-directional sync with the source platform.

gabe-lyons (approved):

metadata-models/src/main/pegasus/com/linkedin/metric/MetricRelationships.pdl
asikowitz: I have no issue with this. We do this for other entity types. Just because we'll support custom relationship types doesn't mean we shouldn't also have other ones defined in the pdl that we know we'll want

PR #18137, opened 2026-07-02
fix(ingestion): emit fieldPath add op for editableSchemaMetadata field patches

acrylJonny:
## What

`FieldPatchHelper.add_tag` / `add_term` build a patch whose target array element is keyed by `fieldPath`. During patch merge, GMS rebuilds that element from the map value and does **not** re-inject the key. As a result, patching a tag or term onto a schema field that has no existing `editableSchemaFieldInfo` element fails server-side validation:

```
HTTP 422 - fieldPath is required
```

This affects any caller that adds field-level tags/terms via `DatasetPatchBuilder` when the field is not already present in `editableSchemaMetadata` (e.g. enrichment-only connectors that never emit the full schema aspect).

## Fix

Emit an explicit `fieldPath` add op once per field, before the first tag/term op, so the element always carries its key. The op is idempotent — a repeated add of the same scalar is a harmless no-op — so existing multi-op patches are unaffected.

## Testing

- Updated `tests/unit/patch/complex_dataset_patch.json` golden (semantic comparison via `check_golden_file`).
- `pytest tests/unit/patch/test_patch_builder.py` — 9 passed.

## Checklist

- [x] Tests updated
- [ ] Breaking changes (none — additive, idempotent op)

### DIFF
--- a/metadata-ingestion/src/datahub/specific/dataset.py
+++ b/metadata-ingestion/src/datahub/specific/dataset.py
@@ -54,8 +54,25 @@ class FieldPatchHelper(Generic[_Parent]):
             else SchemaMetadataClass.ASPECT_NAME
         )
         self.aspect_field = "editableSchemaFieldInfo" if editable else "schemaFieldInfo"
+        self._field_path_op_added = False
+
+    def _ensure_field_path(self) -> None:
+        # The array element is keyed by fieldPath, but GMS's patch merge rebuilds the element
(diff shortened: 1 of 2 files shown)


acrylJonny:
Superseded by #18138, which is the same commit opened from a branch on the upstream `datahub-project/datahub` repo instead of my fork. Closing this in favor of #18138.

PR #18139, opened 2026-07-02
refactor(ingestion/dataplex): map entries to DataHub entities via per-type mapper classes

sgomezvillamor:
## Motivation

The Dataplex connector mapped each Universal Catalog entry to DataHub entities through a static declarative table (`DATAPLEX_ENTRY_TYPE_MAPPINGS`) plus one large branching build method and a set of table-driven free functions. Adding or reasoning about a single entry type meant touching several places, and per-type behavior was hard to unit-test in isolation.

## Design

Replace the table with a factory/strategy: one small `EntryMapper` class per Dataplex entry type, registered in `ENTRY_MAPPERS` (keyed by entry-type short name). Each mapper is a pure function of `(entry, EntryMappingContext)` returning an `EntryMappingResult` (main entity + additional entities + optional lineage record); shared logic lives in `build_dataset` / `build_container` helpers (composition, flat hierarchy — no family base classes).

Key modeling choices:

- **Contract as declarative data.** A mapper declares `dataplex_entry_type_short_name`, `datahub_platform`, `dataplex_fqn_regex`, an optional `dataplex_parent_entry`, and a `datahub_identity`.
- **`datahub_identity`** captures the one real difference between families — Dataset builds a name from a format string (`DatasetIdentity`), Container builds a `ContainerKey` from the parsed FQN fields (`ContainerIdentity`). `datahub_main_entity_type` is derived from it.
- **`ParentEntryLink`** bundles the parent-entry regex + parent key class as one optional unit (`None` = parent is just the owning project).
- **Cross-entry concerns stay in the orchestrator.** `DataplexEntriesProcessor` keeps the global project-container dedup and the lineage side-channel; mappers only declare/return that data, keeping them trivially testable.
- `dataplex_ids.py` is reduced to identity primitives (regexes, `ContainerKey` classes, parse/instantiate helpers); `dataplex_lineage.py` reads the registry.

Behavior-preserving: the integration golden files are unchanged and the dataplex unit + integration tests pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Pr2RizugDza6N8YT313f6c)_

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_mappers.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_mappers.py
@@ -176,6 +176,12 @@ class EntryMapper(ABC):
 
 
 def _extract_display_name(entry: dataplex_v1.Entry) -> str:
+    """Return the entry's display name, falling back to the last name segment.
+
+    Dataplex only populates ``entry_source.display_name`` for some systems, so
+    when it is missing (or not a real string) we use the trailing segment of the
+    entry resource name as a stable human-readable label.
+    """
     if entry.entry_source:
         display_name = getattr(entry.entry_source, "display_name", None)
         if isinstance(display_name, str) and display_name.strip():
@@ -184,12 +190,19 @@ def _extract_display_name(entry: dataplex_v1.Entry) -> str:
 
 
 def _extract_description(entry: dataplex_v1.Entry) -> str:
+    """Return the entry's description, or an empty string when absent."""
     if entry.entry_source and entry.entry_source.description:
         return entry.entry_source.description
     return ""
 
 
 def _extract_datetime(entry: dataplex_v1.Entry, field_name: str) -> Optional[datetime]:
+    """Read a timestamp field off ``entry_source`` as a UTC ``datetime``.
+
+    ``field_name`` is a proto timestamp attribute such as ``create_time`` or
+    ``update_time``. Returns ``None`` when there is no ``entry_source`` or the
+    field is unset.
+    """
     if not entry.entry_source:
         return None
     value = getattr(entry.entry_source, field_name, None)
@@ -199,6 +212,15 @@ def _extract_datetime(entry: dataplex_v1.Entry, field_name: str) -> Optional[dat
 
 
 def _extract_entry_group_id(entry_name: str) -> str:
+    """Extract the entry group id from a Dataplex entry resource name.
+
+    The id is the path segment immediately after ``entryGroups``. Returns
+    ``"unknown"`` when the name does not follow the expected shape.
+
+    Example:
+        ``projects/my-project/locations/us/entryGroups/@bigquery/entries/foo``
+        -> ``"@bigquery"``
+    """
     parts = entry_name.split("/")
     try:
         entry_groups_index = parts.index("entryGroups")
--- a/metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_mappers.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_mappers.py
@@ -25,7 +25,6 @@ classes, project-key lookup) live in ``dataplex_ids.py``.
 
 from __future__ import annotations
 
-import logging
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
@@ -84,9 +83,6 @@ from datahub.metadata.schema_classes import SchemaMetadataClass
 from datahub.sdk.container import Container
 from datahub.sdk.dataset import Dataset
 from datahub.sdk.entity import Entity
-
-logger = logging.getLogger(__name__)
-
 
 # ----------------------------------------------------------------------------
 # Context + result
@@ -377,13 +373,20 @@ def build_dataset(
         fqn_regex, name_format, entry.fully_qualified_name
     )
     if dataset_name is None:
-        logger.debug(
-            "Skipping dataset build for entry %s: unable to extract dataset name from FQN %s",
-            entry.name,
-            entry.fully_qualified_name,
+        ctx.report.warning(
+            title="Unparseable Dataplex fully_qualified_name",
+            message=(
+                "Recognized the entry type but could not derive a dataset name "
+                "from its fully_qualified_name. Skipping the dataset."
+            ),
+            context=(
+                f"entry_type={entry.entry_type}, "
+                f"entry_name={entry.name}, "
+                f"fully_qualified_name={entry.fully_qualified_name}"
+            ),
         )
-        # Preserve prior behavior: a project container may still be emitted even
-        # when the primary entity cannot be built.
+        # A project container may still be emitted even when the primary entity
+        # cannot be built (matches the prior independent-emission behavior).
         if additional:
             return EntryMappingResult(additional_entities=additional)
         return None
@@ -487,11 +490,20 @@ def build_container(
 
     identity_fields = parse_with_regex(fqn_regex, entry.fully_qualified_name)
     if identity_fields is None:
-        logger.debug(
-            "Skipping container build for entry %s: unable to build container key from FQN %s",
-            entry.name,
-            entry.fully_qualified_name,
+        ctx.report.warning(
+            title="Unparseable Dataplex fully_qualified_name",
+            message=(
+                "Recognized the entry type but could not build a container key "
+                "from its fully_qualified_name. Skipping the container."
+            ),
+            context=(
+                f"entry_type={entry.entry_type}, "
+                f"entry_name={entry.name}, "
+                f"fully_qualified_name={entry.fully_qualified_name}"
+            ),
         )
+        # A project container may still be emitted even when the primary entity
+        # cannot be built (matches the prior independent-emission behavior).
         if additional:
             return EntryMappingResult(additional_entities=additional)
         return None
--- a/metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_mappers.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_mappers.py
@@ -1,0 +1,907 @@
+"""Dataplex entry -> DataHub entity mappers (factory / strategy).
+
+Each Dataplex entry type maps to DataHub entities through a dedicated
+:class:`EntryMapper` implementation. A mapper is a pure function of
+``(entry, EntryMappingContext)``: given a Dataplex entry payload it returns an
+:class:`EntryMappingResult` describing the *main* DataHub entity it produces
+(Dataset or Container), any *additional* entities it may emit alongside it (the
+owning project Container), and — for dataset-producing types — a lineage record.
+
+Cross-entry / stateful concerns are deliberately kept OUT of mappers and live in
+``DataplexEntriesProcessor``:
+
+* the project Container is emitted once per project (global dedup), and
+* lineage records are appended to a shared side-channel.
+
+Mappers only *declare* those as data on the result; the orchestrator performs the
+dedup and the append. This keeps mappers trivially unit-testable
+(payload -> result) with no locks, report accumulation, or shared context.
+
+Adding support for a new Dataplex entry type means adding one small
+:class:`EntryMapper` subclass here and registering it in ``ENTRY_MAPPERS``. The
+identity primitives it needs (FQN / parent-entry regexes, ``ContainerKey``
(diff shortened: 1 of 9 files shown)


metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_mappers.py
```diff
@@ -0,0 +1,1134 @@
...
+    @property
+    def datahub_additional_entity_types(self) -> tuple[type[Entity], ...]:
+        """Entities this mapper may also emit alongside the main one.
+
+        Declarative; the orchestrator dedups them. Every mapper emits the owning
+        project Container, hence the default.
+        """
+        return (Container,)
+
+    @abstractmethod
```
treff7es, line 325: **Suggestion (optional, ~145 lines): make `map()` concrete here instead of abstract.** All 14 concrete `map()` overrides below are pure forwarding of class attributes to `build_dataset`/`build_container`; only `CloudSpannerGraphMapper` adds one extra arg. The base already has the dispatch primitive — `datahub_main_entity_type` isinstance-dispatches on `datahub_identity`. Promote `subtype` to a `datahub_subtype` class attribute and `include_graph_schema_fallback` to a defaulted flag, then implement `map()` once here:

```python
def map(self, entry, ctx):
    identity = self.datahub_identity
    if isinstance(identity, DatasetIdentity):
        return build_dataset(entry, ctx, short_name=self.dataplex_entry_type_short_name,
            platform=self.datahub_platform, subtype=self.datahub_subtype,
            fqn_regex=self.dataplex_fqn_regex, identity=identity,
            parent=self.dataplex_parent_entry,
            include_graph_schema_fallback=self.include_graph_schema_fallback)
    elif isinstance(identity, ContainerIdentity):
        return build_container(entry, ctx, platform=self.datahub_platform,
            subtype=self.datahub_subtype, fqn_regex=self.dataplex_fqn_regex,
            identity=identity, parent=self.dataplex_parent_entry)
    else:
        assert_never(identity)
```

13 of 14 mappers then shrink to ~5 declarative attribute lines with no `map()` override, without losing any type-safety.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_mappers.py
```diff
@@ -0,0 +1,1134 @@
...
+
+        Defaults to ``None``, meaning the entry has no parent container finer than
+        its owning project — parent linkage falls back to the project key. Mappers
+        with an intermediate parent override it with a single ``ParentEntryLink``,
+        keeping the parent-entry regex and schema-key class together.
+        """
+        return None
+
+    @property
+    def datahub_additional_entity_types(self) -> tuple[type[Entity], ...]:
```
treff7es, line 315: **Suggestion: remove this property — it's dead contract surface.** The docstring says "the orchestrator dedups them," but `git grep` finds no reader anywhere: `DataplexEntriesProcessor._build_entities_for_entry` dedups the actual `result.additional_entities` instances, never this class-level type declaration. A declared contract nothing consumes misleads a future maintainer into thinking overriding it changes behavior. Delete it, or wire it into an assertion (e.g. validate `additional_entities` types are a subset of the declared set) if it was meant to be defensive.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_mappers.py
```diff
@@ -0,0 +1,1134 @@
...
+    the edge case where the primary entity could not be built but an additional
+    entity still could, mirroring the previous independent-emission behavior.
+    """
+
+    main_entity: Optional[Entity] = None
+    additional_entities: list[Entity] = field(default_factory=list)
+    lineage_entry: Optional[EntryDataTuple] = None
+
+
+class DatahubIdentity:
```
treff7es: **Minor: consider dropping this empty base in favor of a type alias.** `DatahubIdentity` declares no shared method and is never used as a static type — the real contract is the `Union[DatasetIdentity, ContainerIdentity]` annotation (which already gives `assert_never` exhaustiveness). Replacing it with `DatahubIdentity = Union[DatasetIdentity, ContainerIdentity]` removes an inheritance level for zero behavioral change and aligns with the repo's "avoid tall hierarchies / prefer composition" guidance.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_mappers.py
```diff
@@ -0,0 +1,1134 @@
...
+    payload, so it is threaded in from the scan loop and used only to build the
+    lineage record. ``report`` lets mappers emit warnings for malformed payloads.
+    """
+
+    config: DataplexConfig
+    location: str
+    report: SourceReport
+
+
+@dataclass
```
treff7es: **Nit: `frozen=True` for consistency.** Its siblings `EntryMappingContext` (line 93) and `_CommonFields` (line 394) are frozen; this DTO isn't. It's short-lived so mutability is low-risk, but freezing matches the file's convention.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_mappers.py
```diff
@@ -0,0 +1,1134 @@
...
+            return EntryMappingResult(additional_entities=additional)
+        return None
+    container_key = identity.container_key(identity_fields)
+
+    # Prefer parent linkage from Dataplex parent_entry; fall back to project key.
+    container_parent_key: Optional[DataplexProjectId] = None
+    if parent is not None and entry.parent_entry:
+        container_parent_key = parent.container_key(entry.parent_entry)
+    if container_parent_key is None:
+        container_parent_key = _project_schema_key(
```
treff7es, line 702: **Follow-up (pre-existing, not introduced here — do not block):** when a `parent_entry` is present but fails to parse, `build_container` silently falls through to the project key, whereas `build_dataset` (lines 557/572) emits a `warning` for the same condition. I checked the pre-refactor code via `gh pr diff` and the old container path also fell back silently, so this is faithfully-preserved behavior. Worth mirroring `build_dataset`'s `"Unparseable Dataplex parent_entry"` warning in a later PR so a container silently reparented under its project root doesn't read as a green run.

metadata-ingestion/tests/unit/dataplex/test_dataplex_mappers.py
```diff
@@ -0,0 +1,441 @@
...
+    fqn: str,
+    parent_entry: str = "",
+    name: Optional[str] = None,
+) -> dataplex_v1.Entry:
+    entry = Mock(spec=dataplex_v1.Entry)
+    entry.name = name or f"projects/p/locations/us/entryGroups/g/entries/{short_name}"
+    entry.entry_type = f"{ENTRY_TYPE_PREFIX}{short_name}"
+    entry.fully_qualified_name = fqn
+    entry.parent_entry = parent_entry
+    entry.entry_source = None
```
treff7es, line 43: **Test gap (low priority): payload-extraction helpers lost their edge-case coverage.** Every case sets `entry_source = None` and the tests never assert the resulting `display_name`/`description`/`created`/`last_modified`/`custom_properties`. The old `test_dataplex_entries.py` had dedicated tests (deleted in this refactor) for branches that are still live in `dataplex_mappers.py`: the non-string `display_name` guard + last-segment fallback (`:348-351`), `_extract_datetime` None guards + proto→UTC conversion (`:368-373`), and `_extract_entry_group_id` → `"unknown"` fallback (`:390-391`). Happy paths are backstopped by the integration goldens, but these partial-payload branches likely aren't in any golden. ~15 lines of direct helper tests (Mock entry with a non-string display_name, missing `entry_source`, entryGroups-less name) would restore it.

metadata-ingestion/tests/unit/dataplex/test_dataplex_mappers.py
```diff
@@ -0,0 +1,441 @@
...
+    assert fmt_fields.issubset(regex_groups), (
+        f"{short_name}: format fields {fmt_fields} not all in regex groups {regex_groups}"
+    )
+
+
+CONTAINER_TYPES = {c[0] for c in CASES if c[3] == "Container"}
+
+
+@pytest.mark.parametrize("short_name", sorted(CONTAINER_TYPES))
+def test_container_fqn_regex_groups_are_key_fields(short_name: str) -> None:
```
treff7es, line 405: **Test gap (low priority): parent-entry regex↔key-field contract isn't guarded, though the docstring says it is.** `dataplex_ids.py`'s module docstring claims a registry test enforces this contract for "FQN / parent-entry regexes," but this guardrail (and its dataset-format sibling) only cover the *FQN* regexes. Since `instantiate_key` silently drops unknown fields, a superfluous/renamed `dataplex_parent_entry_regex` group would build a wrong-but-valid parent key that `test_parent_entry_link_present_iff_entry_has_parent` (URN-prefix check only) wouldn't catch. Either add a parallel parametrized guardrail over `ParentEntryLink`-bearing mappers, or narrow the docstring to admit parent-entry drift is only checked behaviorally.

treff7es (approved):

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_mappers.py
sgomezvillamor: Good call — replaced the empty base with `DatahubIdentity = Union[DatasetIdentity, ContainerIdentity]` (a `TypeAlias`), dropping the inheritance level while keeping `assert_never` exhaustiveness. 3040cda9

---

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_mappers.py
sgomezvillamor: Good call — made `EntryMappingResult` frozen to match its siblings. cda436c2

---

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_mappers.py
sgomezvillamor: Clarified the misleading docstring in 1db8091c — it no longer implies the orchestrator consumes this property (it dedups the concrete `result.additional_entities` instances). I've kept the property itself on purpose: even though nothing reads it today, it documents the mapper's output contract and keeps a mapper's outputs self-describing, which I expect to be useful as more entity families are added.

---

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_mappers.py
sgomezvillamor: Good call — mirrored the "Unparseable Dataplex parent_entry" warning in `build_container` (it still falls back to the project key, but no longer silently). 3070a9b6

---

metadata-ingestion/tests/unit/dataplex/test_dataplex_mappers.py
sgomezvillamor: Good call — added direct edge-case tests for the extraction helpers (non-string display_name / last-segment fallback, `_extract_datetime` UTC + None guards, `_extract_entry_group_id` "unknown" fallback). 29ef8c3b

---

metadata-ingestion/tests/unit/dataplex/test_dataplex_mappers.py
sgomezvillamor: Good call — added a parametrized guardrail asserting each `ParentEntryLink`'s parent-entry regex named groups are fields on its schema-key class. 29ef8c3b

---

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_mappers.py
sgomezvillamor: There's real duplication between the per-type `map()` bodies today, agreed — but I'd expect it to become anecdotal rather than grow. The current 14 types collapse into two shapes (Dataset / Container), so the bodies look near-identical *now*; as we add genuinely different entry families (VertexAI ML entities, Looker charts/dashboards, ETL dataflow/datajobs, etc.) each will carry its own `map()` logic and the shared surface shrinks in proportion.

Folding it into one concrete `map()` on the base would mean a single method branching on `datahub_main_entity_type` (and, later, on each new family) — which is exactly the big-if dispatch this refactor set out to replace. Keeping `map()` abstract lets the shared *algorithm* live in the `build_dataset` / `build_container` helpers (composition) while each type stays a thin, independently testable declaration. So I'd lean toward keeping it as-is.

---

PR #18144, opened 2026-07-02
feat(ingestion): env-based OAuth (DATAHUB_AUTH_TYPE) for the CLI, default sink, and env-resolved emitters

max-datahub:
## Summary

Follow-up to #17939. That PR added the reusable OAuth token-provider layer (`DatahubClientConfig.auth`), but only clients constructed with an explicit `auth` config could use it. This PR makes OAuth reachable from the **environment**:

- **`load_client_config()` resolves auth from env** (new `datahub/ingestion/auth/env.py`, reading `DATAHUB_AUTH_TYPE` + provider vars through the `env_vars` registry). Any process that resolves the default client config from env vars — the `datahub` CLI, the **default datahub-rest ingestion sink**, and Remote Executor recipe subprocesses that only inherit environment variables — can now authenticate with short-lived OAuth tokens instead of a static `DATAHUB_GMS_TOKEN`. Supported types mirror the provider registry: `k8s_oidc`, `azure_entra`, `oidc_client_credentials`. Missing required variables raise a `ConfigurationError` naming them; client secrets are wrapped in `SecretStr`.
- **Explicit `sink: datahub-rest` blocks inherit env OAuth** when they carry no credentials of their own — scoped to the env-configured server origin, so env-minted tokens are never sent to a different host. Explicit credentials in the sink block always win over the environment.
- **One shared token provider per sink** instead of one per worker thread — avoiding `max_threads × pipelines` token requests against the IdP.
- **The `__from_env__` emitter path honors `DATAHUB_AUTH_TYPE`** with the same precedence as `load_client_config`, and resolves the server from env even when the caller supplies explicit credentials.
- **Loud precedence, never silent**: env auth wins over `DATAHUB_GMS_TOKEN`, a stored `~/.datahubenv` token, and system-client credentials — each override is logged; the missing-server error names `DATAHUB_GMS_URL`.
- Docs: `docs/authentication/external-oauth-providers.md`.

## Known limitations

- **Authentication ≠ authorization**: an OAuth machine principal materializes as a fresh synthetic corpuser with no roles or policies — an operator must grant it a role/policy or writes fail with `403 ... is unauthorized to modify entity`.
- GMS must trust the OAuth issuer (`EXTERNAL_OAUTH_*`) — deploy-time server configuration, unchanged here.
- Browser-login sessions (`datahub init --oauth`, #17868) still refresh only at config-load time; a proposed follow-up serves them through the token-provider layer.

## Testing

- Unit: `tests/unit/auth/test_env_config.py` (env parsing per provider, precedence + warnings, sink env-merge + server scoping, shared provider, `__from_env__`), `tests/unit/cli/test_oauth_config.py`, `tests/unit/test_rest_sink.py` — green; ruff + mypy clean.
- Live end-to-end (combined pre-split branch): quickstart GMS with `EXTERNAL_OAUTH_*` trusting Keycloak — controls (anon 401, wrong-audience 401), env-auth graph round-trip, full ingest matrix, authenticated hard-delete; plus the 8-test CI canary (green in CI, ships as #18192 on top of this PR).

## Checklist

- [x] The PR conforms to DataHub's Contributing Guideline (particularly Contribution Message Format)

## The stack

| Order | PR | Base | Notes |
|---|---|---|---|


datahub-actions/src/datahub_actions/plugin/source/acryl/datahub_cloud_events_consumer.py
```diff
@@ -143,10 +142,16 @@ def _poll_events_impl(
...
-
-        response = requests.get(endpoint, params=params, headers=headers)
+        # Poll through the graph's session so authentication applies per request.
+        # Copying session.headers into a bare requests.get only worked for
+        # static tokens baked into the headers; an OAuth token provider lives in
+        # session.auth and would be silently bypassed (unauthenticated 401s).
+        # The read timeout must outlast the server-side long poll, and a timeout
+        # is required at all — sessions have no default, so a half-open
+        # connection would otherwise hang this consumer forever.
+        response = self.graph._session.get(
```
skrydal: Could we avoid using private (`_`-prefixed) fields of `self.graph`?

datahub-actions/src/datahub_actions/plugin/source/acryl/datahub_cloud_events_consumer.py
```diff
@@ -143,10 +142,16 @@ def _poll_events_impl(
...
-        headers = dict(self.graph._session.headers)
-
-        response = requests.get(endpoint, params=params, headers=headers)
+        # Poll through the graph's session so authentication applies per request.
+        # Copying session.headers into a bare requests.get only worked for
+        # static tokens baked into the headers; an OAuth token provider lives in
+        # session.auth and would be silently bypassed (unauthenticated 401s).
+        # The read timeout must outlast the server-side long poll, and a timeout
+        # is required at all — sessions have no default, so a half-open
+        # connection would otherwise hang this consumer forever.
```
skrydal: It seems to me this comment explains the recent change and not the state of the code as it is, might age badly.

datahub-actions/src/datahub_actions/plugin/source/acryl/datahub_cloud_events_consumer.py
max-datahub: Good call — fixed in 1ffe828db4. `DataHubRestEmitter` (and therefore `DataHubGraph`) now exposes the authenticated session as a public read-only `session` property (bd8f51bd6f), and both event pollers use `self.graph.session` instead of `_session`. The pre-existing `_session` usages in `action_graph.py` are untouched here to keep the diff scoped, but they can migrate to the same property.

datahub-actions/src/datahub_actions/plugin/source/acryl/datahub_cloud_events_consumer.py
max-datahub: Agreed — reworded in 1ffe828db4 to state the current invariants (per-request auth only applies to requests made through the session; the read timeout must outlast the server-side long poll) without narrating the before/after.

skrydal (approved):
LGTM

PR #18145, opened 2026-07-02
feat(sql-parsing): opt-in multiprocessing for sqlglot parsing in SqlParsingAggregator

askumar27:
## 📋 Summary

Add opt-in, per-session-safe **multiprocessing** to `SqlParsingAggregator` so sqlglot SQL parsing can run across multiple CPU cores. Default **off**; benefits every connector that uses the aggregator (Snowflake, BigQuery, Redshift, Databricks/Unity, and others).

## 🎯 Motivation

On a large production run, of ~2.45M queries scanned, roughly half (~1.25M) fell back to the expensive sqlglot parse path — and that path alone accounted for **~8.75h of the ~9.4h total run** (~25ms/query). `SqlParsingAggregator` parses strictly single-process, so this CPU-bound work could not use additional cores.

sqlglot parsing is pure-Python CPU work, so Python's GIL means threads give ~zero speedup — real **worker processes** are required. Multiprocessing had never been added because the `SchemaResolver` isn't picklable (it wraps a live SQLite connection and optionally a `DataHubGraph` client), aggregator storage is single-writer, and temp-table lineage is sequential within a session. This PR resolves all three while guaranteeing parsing output is identical to the serial path.

## 🔧 Changes Overview

**New Features**
- **Parallel SQL parsing** in `SqlParsingAggregator`, exposed via a `parallel_sql_parsing_scope()` context manager that connectors wrap their query-feeding loop in. Opt-in through two config fields (default off): `use_parallel_sql_parsing` and `sql_parsing_workers`.
- **cgroup/CPU-aware worker sizing** (`utilities/cpu_detection.py`): the worker count is the *minimum* of the cgroup CPU quota (v2 `cpu.max` / v1 `cfs_quota`), the process's scheduling affinity, and `os.cpu_count()`. This prevents over-spawning on virtualized EC2/K8s hosts where `os.cpu_count()` reports the host's cores (e.g. 64 on a node backing a 2-CPU pod). Explicit overrides are clamped down to detected capacity with a warning; runs with < 2 usable cores fall back to serial.
- **Read-only schema snapshots** (`schema_resolver.py`, `file_backed_collections.py`): `SchemaResolver.snapshot_to()` / `load_readonly()` / `for_worker()` let worker processes read schemas from a frozen, `immutable` SQLite file shared via the OS page cache (low aggregate memory), with a small writable in-memory overlay for per-worker graph hydration.
- **New report fields** for observability: whether parallelism was enabled, chosen worker count, detected CPUs, whether the request was clamped, whether it fell back to serial, and how many queries were parsed in parallel.

**Modifications**
- Wired the two config fields (via a shared `SqlParsingParallelismConfig` mixin) and the scope into **Snowflake**, **BigQuery**, **Redshift** (lineage + usage), and **Databricks/Unity Catalog**.
- Generalized `BackpressureAwareExecutor.map` with an optional, backward-compatible `executor` parameter.

**Dependencies:** none (standard library `multiprocessing`/`concurrent.futures`; no new packages, no entry-point changes).

## 🏗️ Architecture / Design Notes

- **Parse in workers, aggregate on the main process.** Only the CPU-bound `sqlglot_lineage` call runs in a spawn-based `ProcessPoolExecutor`; all mutation of aggregator state stays serialized on the main side under an apply-lock.
- **Per-session-safe ordering — the no-loss guarantee.** Temp-table lineage is session-scoped and sequential (a query can create a temp table a later same-session query consumes). Queries are routed through `PartitionExecutor` keyed by `session_id`: at most one task per session runs at a time, and the next same-session task is dispatched only after the current one is fully applied — so a temp producer is always applied before its consumer is classified/parsed. Different sessions parse concurrently (temp state is session-local). Temp-consuming queries parse inline (their in-memory temp schemas can't be shipped to a worker).
- **Graph-hydration parity.** Workers rebuild a `DataHubGraph` from a picklable `DatahubClientConfig` and use a two-tier resolver (read-only snapshot + writable overlay), so schemas not present in the snapshot are hydrated per-worker exactly as the serial path does — no lineage is lost.
- **Spawn (not fork)** start method avoids inheriting the parent's locks / SQLite file descriptors.

## 🧪 Testing

- **Serial-vs-parallel equivalence** (the core correctness gate): byte-identical emitted MCPs across no-temp-table, temp producer/consumer within a session, interleaved multi-session (including no-session queries), and query-logging-enabled scenarios, plus a repeated high-volume many-session stress test.
- Unit tests for cgroup v1/v2 parsing + clamp/fallback math; read-only snapshot round-trip and concurrent readers; worker graph-hydration parity and fetch/miss dedup; cross-process pickling parity.
- Query `latest_timestamp` merge made order-independent (max) so parallel output matches serial even for a fingerprint seen across sessions.
- **~593 unit tests pass** across the touched areas; ruff + mypy clean.
- **Follow-up:** live real-source benchmark (Snowflake/Redshift/Databricks) to measure end-to-end speedup is pending and will be run against a DataHub instance.

## 📊 Impact Assessment

- **Affected components:** `sql_parsing/` aggregator + schema resolver, two shared utilities, and the four connectors listed above.
- **Breaking changes:** none. All new constructor args and config fields are additive with safe defaults; with the feature off the serial code path is byte-for-byte unchanged.
- **Performance:** large expected reduction in wall-clock for sqlglot-heavy runs (the parse path dominates such runs), scaling with usable cores; conservative, cgroup-aware sizing avoids CPU throttling / OOM on constrained pods.
- **Risk level:** Low — opt-in (default off), guarded by CPU/serial fallback and graceful degradation if a worker pool can't be created; extensively covered by equivalence tests.

## 🚀 Deployment Notes

- Enable per connector recipe with `use_parallel_sql_parsing: true`. Worker count auto-detects a safe value; optionally pin with `sql_parsing_workers: <n>` (values above detected capacity are clamped with a warning; `< 1` is rejected at config parse).
- No migrations, no schema/model changes, no feature-flag infra. Rollback = unset the flag.
- New report fields surface the chosen parallelism and any clamping/fallback for observability.

## ❓ Questions for Reviewers

- The per-session ordering intentionally serializes queries with no `session_id` (they collapse into one bucket) to preserve the no-loss guarantee — acceptable trade-off, or worth a follow-up for connectors where session ids are frequently absent?
- Temp-session inline parses currently hold the apply-lock for the full parse (globally serializing the temp minority). Fine for now; flagged as a possible future optimization.

### DIFF
--- a/.superpowers/sdd/progress.md
+++ b/.superpowers/sdd/progress.md
@@ -1,7 +1,0 @@
-# Parallel SQL parsing — progress ledger
-Task 1: complete (cpu_detection.py, commits e8fc959..a9c4247, review clean, 18/18 tests)
-Task 2: complete (read-only ConnectionWrapper/FileBackedDict + SchemaResolver.snapshot_to/load_readonly, commits 48c566b..b20a13b, review clean incl H1 fix, 58/58 tests)
-Task 3: complete (ParallelSqlParser spawn pool + BackpressureAwareExecutor generalization, commit fcb1543c, review Spec✅/Approved, 45 tests)
-  MINOR for final review:
-  - M1: SqlParsingResult exception tracebacks dropped across pickle (parallel debug_info has no traceback vs serial); lineage unaffected. Consider a comment in _worker_parse.
-  - M2: DatahubClientConfig.token is plain str; ensure no future edit logs graph_config/initargs (no leak today).
--- a/.superpowers/sdd/progress.md
+++ b/.superpowers/sdd/progress.md
@@ -1,0 +1,7 @@
+# Parallel SQL parsing — progress ledger
+Task 1: complete (cpu_detection.py, commits e8fc959..a9c4247, review clean, 18/18 tests)
+Task 2: complete (read-only ConnectionWrapper/FileBackedDict + SchemaResolver.snapshot_to/load_readonly, commits 48c566b..b20a13b, review clean incl H1 fix, 58/58 tests)
+Task 3: complete (ParallelSqlParser spawn pool + BackpressureAwareExecutor generalization, commit fcb1543c, review Spec✅/Approved, 45 tests)
+  MINOR for final review:
+  - M1: SqlParsingResult exception tracebacks dropped across pickle (parallel debug_info has no traceback vs serial); lineage unaffected. Consider a comment in _worker_parse.
+  - M2: DatahubClientConfig.token is plain str; ensure no future edit logs graph_config/initargs (no leak today).
--- a/metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/queries_extractor.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/bigquery_v2/queries_extractor.py
@@ -49,6 +49,7 @@ from datahub.sql_parsing.schema_resolver import SchemaResolver
 from datahub.sql_parsing.sql_parsing_aggregator import (
     ObservedQuery,
     SqlParsingAggregator,
+    SqlParsingParallelismConfig,
 )
 from datahub.sql_parsing.sqlglot_utils import get_query_fingerprint
 from datahub.utilities.file_backed_collections import (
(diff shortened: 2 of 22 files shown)


askumar27:
Superseded — split into a core PR + one per connector for easier review:
- Core: #18146 (aggregator + shared utilities)
- Snowflake: #18147
- BigQuery: #18148
- Redshift: #18149
- Unity/Databricks: #18150

Connector PRs are stacked on the core branch (#18146); merge core first.

PR #18146, opened 2026-07-02
feat(sql-parsing): opt-in multiprocessing core for SqlParsingAggregator

askumar27:
## 📋 Summary
Add the **opt-in multiprocessing core** for `SqlParsingAggregator` so the CPU-bound sqlglot work (parsing **and** query formatting) runs across worker processes. Default **off**. `sql_parsing_workers` is required when enabled (no `os.cpu_count()` inference, so it can't over-spawn on virtualized hosts). Connector wiring lands in separate stacked PRs (#18147 Snowflake, #18148 BigQuery, #18149 Redshift, #18150 Unity).

## 🎯 Motivation
On a large production run, ~1.25M of ~2.45M queries hit the sqlglot path, accounting for **~8.75h of the ~9.4h total** (~25ms/query). The aggregator parsed strictly single-process; sqlglot is pure-Python (GIL-bound), so only worker *processes* help.

## 🔧 Changes
- **`schema_resolver.py` + `file_backed_collections.py`** — read-only `immutable` SQLite snapshot (shared across workers via the OS page cache) + a two-tier `for_worker` resolver (writable overlay + snapshot fallback + per-worker `DataHubGraph`) so per-worker graph hydration is not lost; single `_init_components()` construction path.
- **`parallel_sql_parser.py`** — spawn `ProcessPoolExecutor`; workers rebuilt from a picklable `DatahubClientConfig`; workers **parse and format** each query (formatting was the dominant serial term); bounded drain; frozen `ParseTask`/`ParseOutcome` (result-XOR-error); `RuntimeError` on use-after-close; `ParallelParserUnavailable` only for pool-creation failure; broken-pool report signal.
- **`sql_parsing_aggregator.py`** — `parallel_sql_parsing_scope()`; **per-session-safe** dispatch via `PartitionExecutor` keyed by `session_id` (a temp producer is always applied before its same-session consumer is parsed); apply-lock serializing all shared-state mutation; worker-preformatted query threaded into apply; report fields (parallelism, fell-back + reason, pool-broke, parsed-in-parallel); order-independent query-map merges; `SqlParsingParallelismConfig` mixin.

## ⚡ Performance
Measured on synthetic high-volume workloads (the smoke-test accounts have ~no query history, so they can't exercise this) — serial vs parallel, output byte-identical:

| workers | speedup (realistic/heavy queries) |
|--------:|:--:|
| 4 | 2.7x |
| 8 | 4.1x |
| 12 | 4.3x |

Roughly **8.75h → ~2h** on the parse phase. See the benchmark comment for methodology and the scaling curve.

## ⚠️ Known limitation — current performance bottleneck
Speedup **plateaus around 4–4.5x and does not scale linearly with workers** (8→12 workers only moves 4.1x→4.3x). This is by design, not a bug:

- Parsing + formatting are parallelized across workers, but **aggregation stays on a single main process**: every parsed result is sent back to one process, which updates the **shared** lineage/query/usage ledgers **serially** under one lock. Concurrent writers to the same ledger entry would corrupt lineage, so this serialization is required for the no-loss guarantee.
- A raw parse-only benchmark tops out at ~5.5x on 8 workers for the same reason — funneling results through one process is the ceiling, independent of the aggregator logic.
- Consequence: the single-aggregator apply/merge step is now the limiter. Beating ~5x would require **sharded aggregation** (multiple aggregator processes owning disjoint slices of the ledgers, merged at the end) — a substantial follow-up with its own correctness surface, deliberately out of scope here.

**Guidance:** enable only for high-volume, sqlglot-heavy runs (set an explicit `sql_parsing_workers`, ~8 is the sweet spot); small workloads regress due to worker spawn + snapshot overhead, which is why it's default-off.

## 🧪 Testing
Serial-vs-parallel **equivalence** (byte-identical MCPs, incl. formatted query text) across no-temp, temp producer/consumer, interleaved multi-session, query-logging, and high-volume stress; read-only snapshot round-trip + concurrent readers; worker graph-hydration parity; pool-creation serial fallback; config validation. Full `sql_parsing` suite: **521 passed, 3 skipped**; ruff + mypy clean.

## 📊 Impact
Additive, **default off** → serial path byte-for-byte unchanged. Risk: Low. No migrations, no new dependencies.

### DIFF
--- a/e2e-test/ui/playwright/pages/common/toast-component.ts
+++ b/e2e-test/ui/playwright/pages/common/toast-component.ts
@@ -5,14 +5,15 @@ import { type Locator, type Page, expect } from '@playwright/test';
  *
  * The app is mid-migration from antd's global `message.*` API (rendered in `.ant-message`)
  * to the alchemy `toast.*` API (rendered in a portal with `data-testid="toast-notification-container"`).
- * Match either container so this helper works regardless of which API a given surface uses.
+ * Some surfaces (e.g. subscriptions) still use antd's `notification.*` API, rendered in `.ant-notification`.
+ * Match any of these containers so this helper works regardless of which API a given surface uses.
  */
 export class ToastComponent {
   private readonly container: Locator;
 
   constructor(page: Page) {
-    // eslint-disable-next-line playwright/no-raw-locators -- antd's global message container has no data-testid or ARIA role; the union covers both toast systems during migration.
-    this.container = page.locator('.ant-message, [data-testid="toast-notification-container"]');
+    // eslint-disable-next-line playwright/no-raw-locators -- antd's global message/notification containers have no data-testid or ARIA role; the union covers all three toast systems during migration.
+    this.container = page.locator('.ant-message, .ant-notification, [data-testid="toast-notification-container"]');
   }
 
   getToast(text: string | RegExp): Locator {


askumar27:
### Benchmark: parallel SQL parsing speedup (updated)

**Why synthetic:** the available test accounts have ~no query history (a real Snowflake queries-v2 run parsed only 2 queries), so they can't measure speedup. To measure the parse speedup itself, it's isolated: many distinct queries fed straight through `SqlParsingAggregator`, serial vs parallel, on a 14-core box. Output equivalence checked every run (identical MCPs).

**Scaling — realistic/heavy queries (CTE + multi-join, ~12–25ms/query), serial baseline ≈ 41–46s:**

| workers | speedup | output |
|--------:|--------:|:------:|
| 2 | 1.5x | identical |
| 4 | 2.7x | identical |
| 8 | **4.1x** | identical |
| 12 | 4.3x | identical |

(Trivial queries scale worse — ~2x — because worker spawn + snapshot overhead dominates a tiny workload. Enable only for high-volume runs; ~8 workers is the sweet spot.)

**Key optimization in this PR:** query *formatting* (`try_format_query`, a full sqlglot round-trip) was originally done serially on the main thread and was the #1 serial cost after parsing — moving it into the worker lifted heavy-query speedup from ~2.6x to ~4.1x at 8 workers.

**Current bottleneck (why it plateaus ~4–4.5x, not linear):**
- Parse + format are parallel, but **aggregation is single-process**: every result funnels back to one main process that updates the shared lineage/query/usage ledgers **serially under one lock** (concurrent writers would corrupt lineage → violates the no-loss guarantee).
- A raw parse-only run (no aggregator) tops out at ~5.5x@8 for the same funnel reason — so the single-process result collection is the ceiling.
- Beyond ~8 workers, returns diminish (4.1x→4.3x from 8→12) as that funnel saturates.
- Getting materially past ~5x needs **sharded aggregation** (multiple aggregator processes over disjoint ledger slices, merged at the end) — a substantial follow-up, out of scope here.

**Bottom line:** ~4.3x on realistic workloads (≈ 8.75h → ~2h), correctness-preserving, default-off. See the PR description's "Known limitation" section.

_(Supersedes the earlier 2.1x figure, which predated the format-in-worker optimization.)_

PR #18153, opened 2026-07-02
feat(sql-parsing): cgroup/CPU-aware worker sizing for parallel parsing

askumar27:
## Summary
Add **cgroup/CPU-aware worker sizing** on top of the parallel-SQL-parsing core. Replaces the core's naive "trust the user value / default to `os.cpu_count()`" with a safe auto-detected worker count and clamping.

## Why separate
The core PR (#18146) intentionally trusts the user-provided `sql_parsing_workers` to keep its review small. This PR adds the safety layer needed for virtualized EC2/K8s hosts, where `os.cpu_count()` reports the host's cores (e.g. 64 on a node backing a 2-CPU pod) and would cause over-spawning / throttling / OOM.

## Changes
- `utilities/cpu_detection.py` — worker count = **minimum** of cgroup v2 (`cpu.max`) / v1 (`cfs_quota`) CPU quota, scheduling affinity, and `os.cpu_count()`; each signal guarded (malformed/unlimited → skipped); non-Linux falls back to `cpu_count`.
- `resolve_worker_count()` reserves a core for the main process, **clamps** an over-capacity explicit request down (with a warning), and **falls back to serial** under 2 usable cores.
- Aggregator now sizes the pool via `resolve_worker_count(sql_parsing_workers)` and re-adds report fields `sql_parsing_cpu_detected` and `sql_parsing_workers_clamped`; `sql_parsing_fell_back_to_serial` now also covers the <2-core case.

## Dependency
**Stacked on the core PR #18146** (base branch of this PR). Merge core first.

## Testing
Unit tests for cgroup v1/v2/unlimited/malformed/non-Linux + clamp + serial fallback; aggregator tests for clamped-count report fields and the detection serial-fallback path; existing equivalence/stress tests unaffected (detection pinned high in-test so explicit worker counts aren't clamped in low-CPU CI). ruff + mypy clean.

### DIFF
--- a/metadata-ingestion/src/datahub/sql_parsing/sql_parsing_aggregator.py
+++ b/metadata-ingestion/src/datahub/sql_parsing/sql_parsing_aggregator.py
@@ -4,7 +4,6 @@ import enum
 import functools
 import json
 import logging
-import os
 import pathlib
 import tempfile
 import threading
@@ -86,6 +85,7 @@ from datahub.sql_parsing.tool_meta_extractor import (
     ToolMetaExtractorReport,
 )
 from datahub.utilities.cooperative_timeout import CooperativeTimeoutError
+from datahub.utilities.cpu_detection import WorkerDecision, resolve_worker_count
 from datahub.utilities.dedup_list import deduplicate_list
 from datahub.utilities.file_backed_collections import (
     ConnectionWrapper,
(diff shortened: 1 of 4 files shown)


askumar27:
Skipping for pass 1. Core (#18146) now requires an explicit `sql_parsing_workers` when parallel parsing is enabled (no os.cpu_count() inference), which removes the virtualization over-spawn risk without needing cgroup detection yet. cgroup/CPU-aware auto-sizing can return as a follow-up on top of the merged core; branch `feat/psp-cpu-detection` is retained.

PR #18161, opened 2026-07-03
fix(kafka-connect): handle platform_instance prefix in Debezium table discovery

javabrett:
## Summary

When `platform_instance_map` is configured (e.g. `{postgres: my_instance}`), `DatasetUrn.create_from_ids` embeds the platform instance in the URN name field as `"{platform_instance}.{table_name}"` (e.g. `my_instance.my_db.public.orders`). This caused the Kafka Connect table-discovery paths to silently return 0 results, because the PI-prefixed name never matched the database prefix or the pattern filter.

**Root cause:** `_discover_tables_from_database` filters by `table_name.startswith(f"{database_name}.")`, but the `table_name` read from the SchemaResolver cache is `my_instance.my_db.public.orders` — it starts with `my_instance.`, not `my_db.`, so zero tables are discovered. The same mismatch affects the shared lineage path and the Debezium/Snowflake pattern-matching paths.

**Fix:** Adds `_strip_platform_instance_prefix()` to `BaseConnector` and calls it at every table-name extraction site across all discovery paths:

- `_discover_tables_from_database` (common.py) — used by Debezium when `table.include.list` is omitted
- `_extract_lineages_from_schema_resolver` (common.py) — shared lineage path for all connectors
- Debezium `_query_tables_from_datahub` (source_connectors.py) — pattern-matching path
- Snowflake `_query_tables_from_datahub` (source_connectors.py) — same fix for consistency

When no `platform_instance` is configured the method is a strict no-op, preserving full backwards compatibility. It uses `getattr()` for safe access to `platform_instance`, so existing callers/mocks without the attribute continue to work.

## Files Changed

- `metadata-ingestion/src/datahub/ingestion/source/kafka_connect/common.py` — adds the `_strip_platform_instance_prefix()` helper and applies it in `_discover_tables_from_database` and `_extract_lineages_from_schema_resolver`
- `metadata-ingestion/src/datahub/ingestion/source/kafka_connect/source_connectors.py` — applies PI stripping in both the Debezium and Snowflake `_query_tables_from_datahub` paths
- `metadata-ingestion/tests/unit/kafka_connect/test_kafka_connect_debezium_table_discovery.py` — adds `TestDiscoverTablesWithPlatformInstance` with focused regression tests for the PI-prefix scenario

## Test Plan

- [x] `./gradlew :metadata-ingestion:lintFix` — all checks pass
- [x] `test_kafka_connect_debezium_table_discovery.py` — 36/36 pass with the fix
- [x] Confirmed genuine regression coverage: reverting only the source-file changes (keeping the tests) turns the platform-instance tests red — `_discover_tables_from_database` returns `[]` instead of the expected tables, reproducing the "silently 0 results" bug
- [x] Reachable extraction sites are covered directly: `_discover_tables_from_database`, the Debezium `_query_tables_from_datahub`, and the Snowflake `_query_tables_from_datahub` override
- [x] The no-`platform_instance` path is covered and stays a no-op (backwards compatibility)

## Checklist

- [ ] Docs related to the changes have been added/updated (not applicable — internal fix, no user-facing surface change)
- [ ] For any breaking change an entry has been made in Updating DataHub (not applicable — no breaking change)

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/kafka_connect/common.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/kafka_connect/common.py
@@ -783,6 +783,10 @@ class BaseConnector:
                 if not table_name:
                     continue
 
+                # When platform_instance is configured, DatasetUrn embeds it as
+                # "{platform_instance}.{table_name}". Strip it before filtering.
+                table_name = self._strip_platform_instance_prefix(table_name)
+
                 # Filter by database - check if table_name starts with database prefix
                 if database_name:
                     if table_name.lower().startswith(f"{database_name.lower()}."):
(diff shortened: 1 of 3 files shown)


javabrett:
Closing as superseded by #18231, which merged to `master` and fixes the same `platform_instance`-prefix bug more centrally: it strips the prefix inside `_extract_table_name_from_urn`, so all table-discovery, pattern-matching, and lineage paths inherit the fix. That change covers the paths this PR targeted, so this one is redundant. Thanks!

PR #18163, opened 2026-07-03
fix(ingest/delta-lake): allow platform values other than delta-lake

umartin:
This patch makes the "platform" configuration property configurable. Right now the only accepted value is the literal "delta-lake"

https://docs.datahub.com/docs/generated/ingestion/sources/delta-lake#config-details


umartin:
@maggiehays Did you find a reviewer for this PR?

PR #18167, opened 2026-07-03
fix(ingest/thoughtspot): Fix ThoughtSpot ingestion with Bigquery connection

Nelvin73:
This fixes the ThoughtSpot ingestion source.

Currently the thoughtspot ingestion source contains 2 issues.

1. For BigQuery it contains the wrong data_source_type name
2. It only removes RDBMS for views, but not for tables

Both leads to not matching / invalid datahub platforms and therefore the lineage to the BigQuery objects is not available.

Furthermore, I had some rare case, where the "metadata_details" field contained an string error message, instead of the dictionary, which leads to a failed ingestion.
I fixed this edge case too.


Nelvin73:
@treff7es  
This is the PR regarding my Slack message
https://datahubspace.slack.com/archives/CUMUWQU66/p1782373666475839

treff7es:
> @treff7es This is the PR regarding my Slack message [datahubspace.slack.com/archives/CUMUWQU66/p1782373666475839](https://datahubspace.slack.com/archives/CUMUWQU66/p1782373666475839)

Thanks, this fix is super useful and looks good.

treff7es (approved):
lgtm

PR #18168, opened 2026-07-03
feat(dataplex): add Dataproc Metastore support

sgomezvillamor:
## Summary

Adds support for ingesting Google Cloud Dataproc Metastore metadata from Knowledge Catalog (Dataplex) with three entity types following the existing Cloud SQL/Spanner patterns.

### Entity Types

| Entry Type | DataHub Platform | Entity Type | Subtype | Parent |
|------------|------------------|-------------|---------|--------|
| `dataproc-metastore-service` | `dataproc-metastore` | Container | Service | Project container |
| `dataproc-metastore-database` | `dataproc-metastore` | Container | Database | Service container |
| `dataproc-metastore-table` | `dataproc-metastore` | Dataset | Table | Database container |

### Changes

**Backend:**
- Add `SERVICE` subtype to `DatasetContainerSubTypes` for service-level containers
- Add identity classes (`DataplexDataprocMetastoreProject`, `DataplexDataprocMetastoreService`, `DataplexDataprocMetastoreDatabase`) in `dataplex_ids.py`
- Add FQN regexes and parent entry regexes following the pattern: `projects/{proj}/locations/{loc}/entryGroups/@dataprocmetastore/entries/metastore.googleapis.com/projects/{proj}/locations/{loc}/services/{svc}[/databases/{db}]`
- Add three mapper classes in `dataplex_mappers.py` following Cloud SQL/Spanner patterns
- Register `dataproc-metastore` platform in `data-platforms.yaml` (version bumped to v14)
- Add platform logo: `dataprocmetastorelogo.svg`

**Frontend:**
- Register platform logo in both `ingest` and `ingestV2` source builder constants

**Documentation:**
- Update `metadata-ingestion/docs/sources/dataplex/README.md` with concept mapping table
- Update `metadata-ingestion/docs/sources/dataplex/dataplex_pre.md` with prerequisites for enabling Knowledge Catalog integration

**Testing:**
- Add 3 comprehensive test cases in `test_dataplex_mappers.py`
- All 84 unit tests passing
- lintFix and lint validation passing
- End-to-end ingestion tested successfully

### Entity Hierarchy

```
Project (Container) 
└── Service (Container, subtype: SERVICE)
    └── Database (Container, subtype: DATABASE)
        └── Table (Dataset, subtype: TABLE)
```

### Prerequisites

Dataproc Metastore services must have Knowledge Catalog integration enabled:

```sh
# For new services
gcloud metastore services create SERVICE_NAME \
  --enable-dataplex-integration \
  --location=LOCATION

# For existing services
gcloud metastore services update SERVICE_NAME \
  --enable-dataplex-integration \
  --location=LOCATION
```

### Known Limitations

**Views classified as Tables:** Google Cloud Knowledge Catalog (Dataplex) does not distinguish between tables and views for Dataproc Metastore entities. Both are classified with the same entry type `dataproc-metastore-table` and appear as "Table" in the Dataplex UI. As a result, views are ingested into DataHub with subtype "Table".

This differs from BigQuery, which has separate entry types (`bigquery-table` vs `bigquery-view`). This is a limitation of Dataplex's metadata model for Dataproc Metastore, not the DataHub connector. The connector faithfully represents what Dataplex provides.

To distinguish views from tables would require querying the Dataproc Metastore API directly or reading the underlying Hive Metastore, which is outside the scope of the Dataplex connector.

### Breaking Changes

None. This is a new feature that adds support for a new entity type.

### Test Plan

- ✅ Unit tests: 84/84 passing
- ✅ Python linting: All checks passed
- ✅ End-to-end ingestion: Verified with actual Dataplex environment (tables and views)
- ✅ Entity hierarchy validated: Project → Service → Database → Table
- ✅ Platform logo displays correctly in UI

### Screenshots

Search-2026-07-03_17_52
test_table-Table-2026-07-03_17_52


metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_mappers.py
```diff
@@ -798,7 +805,7 @@ def map(
             entry,
             ctx,
             platform=self.datahub_platform,
-            subtype=DatasetContainerSubTypes.INSTANCE,
+            subtype=DatasetContainerSubTypes.SERVICE,
```
treff7es: 🔴 **Blocker — unrelated behavior change to Cloud SQL bundled into this PR.**

This flips the emitted subtype of every existing Cloud SQL instance container from `"Instance"` → `"Service"`. Concerns:

1. **Inconsistent within Dataplex itself** — the structurally-identical `CloudSpannerInstanceMapper` and `CloudBigtableInstanceMapper` keep `INSTANCE`, so Cloud SQL becomes the lone outlier.
2. **GCP calls these "instances"** (`gcloud sql instances …`), and the parent-entry regex parses `/instances/{instance_id}`. The "Dataplex terminology" rationale doesn't hold against Cloud SQL's own naming.
3. **The PR's own docs now contradict the code** — `README.md` still lists `cloudsql-mysql-instance → Container (Instance)` and `dataplex_pre.md` says "Cloud SQL — Instances".
4. **Untested** — no test asserts the container subtype and the goldens are BigQuery-only, so this change is unguarded. It may also be the cause of the failing `datahub-connector-tests` check.

**Suggestion:** revert this line to `DatasetContainerSubTypes.INSTANCE`. The new `SERVICE` value is legitimately earned by Dataproc Metastore alone (GCP genuinely calls those "services"). If you want to standardize instance/service naming across Cloud SQL/Spanner/Bigtable, do it as a separate deliberate PR that applies one term uniformly, updates the docs, and adds a `docs/how/updating-datahub.md` entry.

metadata-ingestion/src/datahub/ingestion/source/common/subtypes.py
```diff
@@ -68,6 +68,7 @@ class GenericContainerSubTypes(StrEnum):
 class DatasetContainerSubTypes(StrEnum):
     # Generic SubTypes
     INSTANCE = "Instance"
+    SERVICE = "Service"
```
treff7es, line 71: Adding `SERVICE` here is correct and well-placed (shared enum, not a dataplex-local one) — Dataproc Metastore's top-level resource is genuinely a "service" in GCP terms. Just note this same value is also being applied to the Cloud SQL mapper (`dataplex_mappers.py:808`), which is the concern flagged there — `SERVICE` should be used by the Dataproc mapper only.

metadata-ingestion/tests/unit/dataplex/test_dataplex_mappers.py
```diff
@@ -148,6 +148,29 @@ def _ctx(include_schema: bool = False) -> EntryMappingContext:
         "Dataset",
         "my-project.us-west2.123456",
     ),
+    (
+        "dataproc-metastore-service",
```
treff7es, line 177: 🟡 The parametrized tests only assert `isinstance(main_entity, Container)`, never the container `subtype` value — so the new `SERVICE` subtype here (and the Cloud SQL `INSTANCE`→`SERVICE` change) are unguarded; a revert or typo would pass CI silently.

Consider extending `CASES` with an expected-subtype column (or adding a focused `result.main_entity.subtype` assertion in `test_mapper_builds_expected_entity`) so the emitted subtype is pinned for at least the container cases. Cheap, and it locks in the intended Dataproc `SERVICE` semantics.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_mappers.py
```diff
@@ -1044,6 +1051,78 @@ def map(
         )
 
 
+class DataprocMetastoreServiceMapper(EntryMapper):
```
treff7es, line 1066: The three new mappers faithfully mirror the Cloud SQL/Spanner pattern — declarative class attrs + delegation to the shared `build_*` helpers, registered in the factory, with the regex↔key-field contract auto-covered by the parametrized guardrail tests. Nice, minimal extension. 👍

treff7es (approved):

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_mappers.py
sgomezvillamor: ✅ Reverted in commit 1424b8d95c — Cloud SQL instances now use `INSTANCE` subtype (unchanged from before this PR). Only Dataproc Metastore uses the new `SERVICE` subtype.

metadata-ingestion/src/datahub/ingestion/source/common/subtypes.py
sgomezvillamor: ✅ Acknowledged — `SERVICE` is now used only by Dataproc Metastore mappers after reverting the Cloud SQL change in commit 1424b8d95c.

metadata-ingestion/tests/unit/dataplex/test_dataplex_mappers.py
sgomezvillamor: ✅ Fixed in commit a6318ddb33 — extended `CASES` with `expected_subtype` column and added assertion `assert result.main_entity.subtype == expected_subtype` in the Container branch of `test_mapper_builds_expected_entity`. All 98 tests pass with the new validation.

PR #18169, opened 2026-07-03
refactor(auth): move TokenType to auth-api

david-leifker:
## Summary

- Move `TokenType` from `auth-impl` to `auth-api` so it can be shared across modules
- Add `TokenClaimNames` with standard JWT claim name constants
- Update `TokenClaims` to reference the shared claim name

Part 1 of 6 in the OSS usage aggregation stack.

## Test plan

- [ ] CI spotlessJavaCheck


david-leifker:
Folded into #18170 (TokenType move is a prerequisite for the GraphQL analyzer / usage stack). Stack now starts at #18170 against master.

PR #18170, opened 2026-07-03
refactor(graphql): parse GraphQL documents once for rate limiting

david-leifker:
## Summary

- Move `TokenType` and `TokenClaimNames` to `auth-api` so shared token taxonomy is available to rate limiting and usage instrumentation
- Replace `GraphQLOperationNameResolver` with `GraphqlDocumentAnalyzer` for a single parse path
- Share operation name resolution and AST metadata for rate limiting
- Add `GraphqlHttpConstants` prerequisite (also used by usage context work in #18171)

Part 1 of 6 in the OSS usage aggregation stack.

## Test plan

- [x] `./gradlew :metadata-service:rate-limit:test`
- [ ] CI


metadata-service/rate-limit/src/main/java/com/linkedin/metadata/ratelimit/GraphqlDocumentAnalyzer.java
```diff
@@ -0,0 +1,133 @@
...
+            new GraphqlOperationMetadata(operationDefinition.getName(), kind, rootFields));
+      }
+      return List.copyOf(operations);
+    } catch (InvalidSyntaxException e) {
+      return List.of();
+    }
+  }
+
+  @Nonnull
+  private static List<String> collectFieldNames(@Nonnull SelectionSet selectionSet) {
```
Dutt23: collectFieldNames only picks up direct Field selections and ignores inline fragments and fragment spreads. The old collectTopLevelFields walked into both on purpose, since this list is what the heavy-resolver gate charges against and what the anonymous-op identity is built from.

These all execute searchAcrossEntities as a root field, but only the first reports it:

```
{ searchAcrossEntities { urn } }                                     # [searchAcrossEntities] → charged
{ ... on Query { searchAcrossEntities { urn } } }                    # [] → not charged
query { ...F } fragment F on Query { searchAcrossEntities { urn } }   # [] → not charged
```
So a client skips the heavy-resolver limit just by wrapping the field in ... on Query or a fragment spread, and anonymous queries fall back to the shared graphql identity bucket instead of the field-derived one. [docs/deploy/gms-rate-limiting.md](https://github.com/datahub-project/datahub/blob/master/docs/deploy/gms-rate-limiting.md) still says this gate is keyed on server-parsed top-level names and can't be spoofed.

The old resolver resolved nested fields collectTopLevelFields recursed into inline fragments and fragment spreads, with a fragment map (fragmentDefinitions) and a visited-set cycle guard: https://github.com/datahub-project/datahub/blob/59bc1c5ab71971ebb61536ddefe76250838b2aa9/metadata-service/rate-limit/src/main/java/com/linkedin/metadata/ratelimit/GraphQLOperationNameResolver.java#L185-L209

These four cases were tested in GraphQLOperationNameResolverTest which handles this edge case.

metadata-service/graphql-servlet-impl/src/main/java/com/datahub/graphql/SpringQueryContext.java
```diff
@@ -45,7 +45,8 @@ public SpringQueryContext(
     this.authorizer = authorizer;
 
     // operationName is an optional field only required if multiple operations are present
-    this.queryName = GraphQLOperationNameResolver.resolve(operationName, jsonQuery);
+    this.queryName =
```
Dutt23, line 49: Nit: SpringQueryContext re-parses the query document unnecessarily.

postGraphQL already resolves the operation name ([GraphQLController.java:231-233](https://github.com/datahub-project/datahub/blob/3060a5d115cd8fca8c28d2af48bcafac14803ef9/metadata-service/graphql-servlet-impl/src/main/java/com/datahub/graphql/GraphQLController.java#L231-L233)) and passes it into the constructor as operationName — which is never blank (either the real name or the graphql sentinel). So analyze(operationName, jsonQuery, null) always parses the full document just to hand back the string it was already given.

The [old resolve() path](https://github.com/datahub-project/datahub/blob/59bc1c5ab71971ebb61536ddefe76250838b2aa9/metadata-service/rate-limit/src/main/java/com/linkedin/metadata/ratelimit/GraphQLOperationNameResolver.java#L45-L48) short-circuited on named queries — one parse per request. This makes it two.

SpringQueryContext has a single caller (GraphQLController), which always passes the resolved name, so the simplest fix is:

// SpringQueryContext.java:48
this.queryName = operationName; // already resolved by GraphQLController
If you'd rather not couple the constructor to that guarantee, the skip predicate keeps a fallback and still avoids the parse when the name is present:

this.queryName =
    GraphqlDocumentAnalyzer.analyze(operationName, jsonQuery, name -> true).resolvedOperationName();

david-leifker: Fixed downstream in #18173, this is an artifact of trying to pull apart sections of a larger change.

metadata-service/rate-limit/src/main/java/com/linkedin/metadata/ratelimit/GraphqlDocumentAnalyzer.java
david-leifker: Good catch here!

Dutt23 (approved):

PR #18171, opened 2026-07-03
feat(usage): add operation context primitives and protect CorpUserInfo flags

david-leifker:
## Summary

Foundation PR for the OSS API usage aggregation stack. Adds operation-context primitives, YAML-backed usage registries, and security guardrails for privileged CorpUser flags.

**This PR (#18171)**
- Extend `RequestContext` and `OperationContext` with usage dimensions, byte measurement helpers, and `SessionContextEnricher` hooks
- Add usage enums and registries (`UsageOperation`, `AuthChannel`, `UsageActorClass`, GraphQL classification, metric/operations YAML loaders)
- Add `RequestAPI.MESSAGING` and allow `metadata_ingest` on the messaging transport in `usage_operations.yaml`
- Classify actors from `CorpUserInfo` flags (`system`, `isSupportUser`) instead of a hardcoded super-user URN
- Add `CorpUserPrivilegedFlagsValidator` to block cross-API privilege escalation when those flags are newly set to `true`:
  - `system=true` — only existing system users (or trusted internal/bootstrap writes)
  - `isSupportUser=true` — system users, support users, or trusted internal/bootstrap writes
- Drop unused `metronome_batch` from the OSS metric registry

**Already merged into this branch**
- #18172 — in-memory aggregation engine and GMS wiring

**Downstream stack (merged in chain; lands after this PR)**
- #18173 — instrument GMS APIs for usage aggregation
- #18212 — instrument queue-path `metadata_ingest` on MCE consumer
- #18174 — document aggregation metrics and add smoke coverage
- #18244 — usage registry CI gate for GraphQL, OpenAPI, and Rest.li

**Prerequisite on `master`**
- #18170 — refactor(graphql): parse GraphQL documents once for rate limiting

## Stack

```
master
  └── #18171  operation context primitives + CorpUserInfo protection  ← this PR
        └── #18172  aggregation engine + GMS wiring  (merged into branch)
              └── #18173  GMS API instrumentation  (merged)
                    └── #18212  MCE queue-path ingest  (merged)
                          └── #18174  docs + smoke tests  (merged)
                                └── #18244  usage registry CI gate  (merged)
```

Each layer is one squashed commit in the stacked branches (except #18174 + #18244, which remain separate commits on the tip).

## Review feedback addressed

- Neutralized “billing” wording in `usage_operations.yaml`; removed SCIM references (SaaS-only)
- `SecretService` uses `Constants.CORP_USER_ENTITY_NAME`
[...]

PR #18172, opened 2026-07-03
feat(usage): add in-memory aggregation engine and GMS wiring

david-leifker:
## Summary

- Add in-memory `UsageAggregationStore` with adaptive flush coordinator
- Export Micrometer metrics on flush schedule
- Wire Spring factories, YAML registries, and session enricher hooks on GMS

Stacked on #18171. Part 3 of 6 in the OSS usage aggregation stack.

## OSS master adaptations

- Added missing manifest POJOs (excluded by `.gitignore` `MANIFEST` rule on macOS)
- Added `ADMIN_ACTOR`, `METATDATA_TEST_ACTOR` constants
- Skipped billing-boundary tests where SaaS-only code is absent
- Pulled forward `GraphqlUsageClassificationFactory` needed by factory tests

## Test plan

- [x] `./gradlew :metadata-io:test --tests "*Usage*"`
- [x] `./gradlew :metadata-service:configuration:test :metadata-service:factories:test`
- [ ] CI


metadata-service/factories/src/main/java/com/linkedin/gms/factory/usage/ScimUsageSessionFilter.java
```diff
@@ -0,0 +1,69 @@
+package com.linkedin.gms.factory.usage;
```
david-leifker, line 1: Is this needed?

metadata-service/factories/build.gradle
```diff
@@ -86,6 +86,7 @@ dependencies {
   testImplementation externalDependency.testContainers
   testImplementation externalDependency.testContainersElasticsearch
   testImplementation externalDependency.testContainersCassandra
+  testImplementation testFixtures(project(':metadata-io'))
```
RyanHolstien, line 89: This comes fairly close to a circular dependency, does it make sense to split out these fixtures elsewhere?

metadata-service/schema-registry-servlet/src/test/resources/application.properties
```diff
@@ -9,5 +9,5 @@ authentication.enabled=false
 spring.main.allow-bean-definition-overriding=true
 #---
 spring.config.activate.on-profile=test
-server.port=53222
-local.server.port=53222
+server.port=0
```
RyanHolstien, line 12: Seems like an odd change, what required this?

RyanHolstien (approved):
Minor comments

metadata-service/schema-registry-servlet/src/test/resources/application.properties
david-leifker: There was a port conflict, however it was probably specific to me running tests locally in parallel. Seems like a dynamic test port is better though

metadata-service/factories/build.gradle
david-leifker: Yes it is close, but long term this god factory project needs to be removed. I removed the dependency from the upstream PR since it was noise.

PR #18179, opened 2026-07-04
feat(sql-parsing): expose is_temp_table predicate on create_lineage_from_sql_statements

acrylJonny:
## Summary

- Plumbs the `SqlParsingAggregator`'s existing `is_temp_table` predicate through the `create_lineage_from_sql_statements` convenience wrapper (defaults to `None`, so existing behavior is unchanged).
- Lets callers mark intermediate tables as temporary when the dialect's own syntax does not — e.g. Teradata `CREATE VOLATILE TABLE` parsed under another platform's dialect. Such tables are then collapsed so lineage flows through to the real base tables (table- and column-level).
- Adds focused unit tests in `test_multi_statement_lineage.py` (`TestIsTempTablePredicate`): predicate collapses a plain `CREATE TABLE`, collapses column lineage through the marked table without leaking phantom CLL, is selective (only marked tables collapse), receives the fully qualified dataset name, and a `False`-returning predicate preserves persistent-table behavior.

This is split out of the MicroStrategy connector work, where the SQL-view lineage extractor needs to treat MicroStrategy's multi-pass volatile/temp intermediates as temp tables even though they are parsed under the warehouse's dialect.

## Checklist

- [x] Tests added

### DIFF
--- a/metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
+++ b/metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
@@ -2484,14 +2484,14 @@ def create_lineage_from_sql_statements(
         default_schema: Optional default schema for unqualified table references
         graph: Optional DataHub graph client for schema resolution
         schema_aware: Whether to use schema-aware parsing
-        is_temp_table: Optional predicate, given the table name (typically
-                       ``db.schema.table``, but the exact format is platform-
-                       dependent and differs when a platform_instance is set),
-                       returning whether it is an intermediate temp table. Use it
-                       when the dialect's own syntax does not mark them (e.g.
-                       Teradata ``CREATE VOLATILE TABLE`` parsed under another
-                       platform's dialect); such tables are then collapsed so
-                       lineage flows through to the real base tables.
+        is_temp_table: Optional predicate, given the table name in
+                ``db.schema.table`` form (the platform instance, if any,
+                is stripped before the predicate is called), returning
+                whether it is an intermediate temp table. Use it when the
+                dialect's own syntax does not mark them (e.g. Teradata
+                ``CREATE VOLATILE TABLE`` parsed under another platform's
+                dialect); such tables are then collapsed so lineage flows
+                through to the real base tables.
 
     Returns:
         SqlParsingResult containing merged lineage from all statements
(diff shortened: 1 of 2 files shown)


metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
```diff
@@ -2482,6 +2484,14 @@ def create_lineage_from_sql_statements(
         default_schema: Optional default schema for unqualified table references
         graph: Optional DataHub graph client for schema resolution
         schema_aware: Whether to use schema-aware parsing
+        is_temp_table: Optional predicate, given the table name (typically
```
alokr-dhub: The docstring is a bit misleading.It says the name "differs when a platform_instance is set" but it doesn't. SqlParsingAggregator._name_from_urn strips the platform_instance prefix before invoking the predicate, so the callback always receives db.schema.table regardless of whether an instance is configured.

metadata-ingestion/tests/unit/sql_parsing/test_multi_statement_lineage.py
```diff
@@ -762,3 +768,96 @@ def test_temp_overwritten_by_second_create(self) -> None:
         assert _table_short_names(result.out_tables) == {"target"}
         # source2 should be the upstream (latest definition)
         assert "source2" in {_urn_to_table_name(u) for u in result.in_tables}
+
+
+class TestIsTempTablePredicate:
```
alokr-dhub, line 773: All six tests inherit platform_instance=None from _DEFAULTS.It does not test the instance set path. I suggest adding a test case where the platform instance is set.

alokr-dhub (changes requested):

metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
acrylJonny: Good catch — corrected. The docstring now states the predicate always receives the name in `db.schema.table` form (the platform instance is stripped before the callback). Fixed in b2112499b7.

metadata-ingestion/tests/unit/sql_parsing/test_multi_statement_lineage.py
acrylJonny: Added `test_predicate_receives_name_without_platform_instance`, which sets `platform_instance="my_instance"` and asserts the predicate still sees `dev.public.staging` (no instance prefix) and the marked table is still collapsed. In b2112499b7.

alokr-dhub (approved):

PR #845, opened 2017-11-07
Make the ETL java command configurable

mars-lan:

PR #18181, opened 2026-07-06
fix(ingest): preserve upstream column casing in BI column-level lineage (PowerBI, Mode)

puneetagarwal-datahub:
> **⚠️ Behavior change — remove upstream column-lowercasing workarounds after upgrading.**
> This fix preserves upstream column casing end-to-end. If you previously set `convert_column_urns_to_lowercase: true` on an upstream SQL source (e.g. MSSQL) to work around missing Power BI column-level lineage, **turn it off after upgrading** — Power BI now preserves the column casing, so a lowercased upstream would mismatch again. Post-fix, the matching regime is "preserve on both sides" (both flags at their defaults).

## Summary

Two BI connectors lowercased the **column/field-path** portion of upstream `schemaField` URNs in column-level lineage. Warehouse sources that lowercase URNs (`convert_urns_to_lowercase` → `lowercase_dataset_urns`) only lowercase the **dataset** portion and deliberately leave the field path untouched. Because of this asymmetry, a BI column-level edge pointed at `...,columnname)` while the upstream warehouse's `schemaField` is `...,ColumnName)`, so the column-level edge was dropped (or resolved to a lowercased ghost field) for any warehouse that stores columns in non-lowercase casing.

This PR fixes it in both PowerBI and Mode.

## Example (PowerBI)

Databricks upstream ingested with `convert_urns_to_lowercase: true`. The warehouse `schemaField` — note the dataset name is lowercased but the **column path keeps its original casing**:

```
urn:li:schemaField:(urn:li:dataset:(urn:li:dataPlatform:databricks,my_catalog.my_schema.rep_table,PROD),CustomerId)
```

PowerBI column-level lineage upstream, with `convert_lineage_urns_to_lowercase: true`:

```
before:  urn:li:schemaField:(urn:li:dataset:(...,my_catalog.my_schema.rep_table,PROD),customerid)   ❌ ghost field, edge dropped
after:   urn:li:schemaField:(urn:li:dataset:(...,my_catalog.my_schema.rep_table,PROD),CustomerId)   ✅ matches warehouse
```

The dataset portion is still lowercased (governed by `convert_lineage_urns_to_lowercase`); only the column path is now preserved — mirroring `lowercase_dataset_urns` exactly.

## Example (Mode)

A chart formula references `[Staff Last Name]`; the query dataset's schema field path is `Staff Last Name`.

```
before:  field lowercased to "staff last name" → no match against the schema → input field silently dropped
after:   matched case-insensitively → emitted as ...,Staff Last Name) → resolves to the real schema field
```

## Changes

**PowerBI**
- `powerbi.py` (`make_fine_grained_lineage_class`): stop lowercasing the upstream column name; the dataset portion is still lowercased via `lineage_urn_to_lowercase`.
- `pattern_handler.py` (`create_table_column_lineage`): the direct-table 1:1 path no longer force-lowercases the column name.

**Mode**
- `mode.py` (`get_input_fields`): match chart-formula fields case-insensitively against the query schema and emit the schema's **actual** field path, so the input-field `schemaField` URN resolves to a real field instead of a lowercased ghost (previously such fields were silently dropped when the schema used non-lowercase paths).

## Tests

- `test_cll_upstream_column_casing_preserved_when_lowercasing_lineage` — PowerBI Mapper-level: dataset portion lowercased, column path preserved.
- `test_direct_table_cll_preserves_upstream_column_casing` — PowerBI direct-table path preserves source column casing.
- `TestGetInputFields::test_preserves_schema_field_casing` — Mode input-field URN resolves to the schema's actual field-path casing.
- Mode integration golden updated (`mode_mces_golden.json`) — two previously-dropped mixed-case input fields now emitted.

## Checklist

- [x] Tests added
- [ ] Docs (not applicable)
- [ ] Breaking changes (not applicable)

## Dependencies

Paired connector-test golden refresh: **acryldata/connector-tests#845** regenerates the PowerBI goldens for the preserved CLL upstream column casing (and catches the static golden up to #17582).

Merge order: this PR → `master` first, then #845. #845 temporarily pins its validation to this PR's dev wheel and must be reverted to a released version once this ships.


treff7es (approved):

sgomezvillamor:
I miss a breaking change note in `updating-datahub.md`

sgomezvillamor (approved):
Overall, removing these arbitrary lowercasing decisions from individual sources is the right direction. That logic should be centralized in the existing auto lowercase workunit or the planned auto lineage resolver workunit.

Even, not in the scope of this PR, I think we should also eventually remove the `convert_lineage_urns_to_lowercase` flag https://github.com/search?q=repo%3Adatahub-project%2Fdatahub+convert_lineage_urns_to_lowercase+language%3APython&type=code&l=Python


Just missing a breaking change note in `updating-datahub.md`.

PR #18183, opened 2026-07-06
fix(ingestion/clickhouse): strip NUL bytes from column comments

jjjeong:
- [x] The PR conforms to DataHub's [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [Commit Message Format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#commit-message-format))

## Summary

When ingesting schema metadata from ClickHouse via the HTTP interface, column comments containing certain multibyte characters (e.g. Korean) can be corrupted, resulting in NUL bytes (`\x00`) embedded in the comment string. This causes downstream failures when attempting to store the metadata in PostgreSQL-based systems, which reject NUL bytes in text fields:

```
psycopg.DataError: PostgreSQL text fields cannot contain NUL (0x00) bytes
```

## Root Cause

ClickHouse stores the column comment correctly (e.g. `부가가치세비율`), but when DataHub reads it from `system.columns` via `clickhouse-connect`'s HTTP interface, certain multibyte character sequences are corrupted — for example, the Korean character `가` (U+AC00) becomes `\ufffd\x00`. The exact cause of this encoding corruption is unclear and may reside in `clickhouse-connect`'s HTTP response parsing rather than in DataHub's code.

## Fix

This PR is a **defensive fix** that prevents crashes caused by NUL bytes in column comments. It strips NUL bytes in `_get_column_info` before returning the column metadata:

```python
# Strip NUL bytes that may appear in column comments due to encoding
# issues when reading from ClickHouse system.columns via HTTP interface.
if comment:
    comment = comment.replace("\x00", "")
```

> **Note**: This fix prevents the crash but does not fully restore the original comment. For example, `부가가치세비율` (correct) may still be stored as `부가\ufffd치세비율` due to the upstream encoding corruption. Fully fixing the display would require identifying and resolving the root encoding bug in `clickhouse-connect`'s HTTP response parsing.

## Testing

Added two unit tests:
- `test_column_info_strips_nul_bytes`: verifies NUL bytes are removed from comments
- `test_column_info_preserves_normal_comment`: verifies normal comments are unaffected

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/sql/clickhouse.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/sql/clickhouse.py
@@ -438,6 +438,11 @@ def _get_column_info(self, name, format_type, comment):
         col_type = col_type.nested_type
         nullable = True
 
+    # Strip NUL bytes that may appear in column comments due to encoding
+    # issues when reading from ClickHouse system.columns via HTTP interface.
+    if comment:
+        comment = comment.replace("\x00", "")
+
     result = {
         "name": name,
         "type": col_type,
--- a/metadata-ingestion/tests/unit/test_clickhouse_source.py
+++ b/metadata-ingestion/tests/unit/test_clickhouse_source.py
@@ -172,3 +172,41 @@ def test_is_temp_table_custom_patterns():
     assert config.is_temp_table("test_table")
     # Default patterns no longer match with custom patterns
     assert not config.is_temp_table("_temp_table")
+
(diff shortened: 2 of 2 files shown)


jjjeong:
Hi @maggiehays, 
thank you for adding the label! 
Could you help review this PR or assign it to the right reviewer? Happy to make any changes needed!

PR #18186, opened 2026-07-06
feat(ingestion): opt-in Kafka default sink for managed ingestion (FF-gated)

Dutt23:
## Summary

Adds an **opt-in** path for ingestion to default to the `datahub-kafka` sink instead of
`datahub-rest` when a recipe does not specify a sink, plus two supporting reliability
primitives on the Kafka sink/emitter. Everything is gated behind environment variables
and **defaults to the existing REST behavior** — no change for any current deployment.

**Motivation — decrease pressure on GMS.** Today UI/managed ingestion writes metadata
synchronously over REST to GMS, so ingest volume directly loads the GMS service. Routing it
to the async MCP Kafka topic instead moves that write off GMS's synchronous request path
(GMS's always-on MCP consumer drains the topic at its own pace) — the primary lever to cut
GMS load from ingestion. This change lands only the generic, reusable primitives; enabling
it is a per-deployment decision via env vars.

## What changed

1. **FF-gated default sink selection** (`pipeline.py`)
   When a recipe omits `sink:`, the default is still `datahub-rest` unless
   `DATAHUB_INGESTION_DEFAULT_SINK=datahub-kafka` is set, in which case a
   `datahub-kafka` sink is injected via a new `_make_default_kafka_sink()` helper
   (bootstrap from `KAFKA_BOOTSTRAP_SERVER`, schema-registry from existing env). A
   single selector controls this; `acryl-datahub` carries no notion of how it is run.
   If `KAFKA_BOOTSTRAP_SERVER` is unset or the broker/schema-registry is unreachable at
   init, the run degrades to REST rather than failing. Guarding remote/REST-only
   executors (which can't reach Kafka) is done one layer up by setting an explicit
   `datahub-rest` sink — a cloud-side follow-up, kept out of the OSS SDK.

2. **Hybrid DELETE → REST fallback on the Kafka sink** (`datahub_kafka.py`)
   New optional `KafkaSinkConfig.rest_fallback`. When set, `DELETE` change-type MCPs are
   emitted synchronously via REST instead of being produced to Kafka (see Drawback below).
   Unset ⇒ the sink is pure Kafka, unchanged.

3. **Producer backpressure** (`kafka_emitter.py`)
   `produce()` is now wrapped so a full local producer queue (`BufferError`) blocks and
   retries after `poll()` instead of raising. This throttles the source to Kafka's
   sustainable rate rather than crashing the run under load. Queue bounds are tunable via
   `producer_config` (`queue.buffering.max.messages`, `queue.buffering.max.kbytes`).

4. **Tests** — FF on/off, unrecognized selector warns and stays REST, explicit sink
   overrides the default, broker/registry-unreachable degrades to REST, DELETE routing to
   the REST fallback, pure-Kafka fallback-unset behavior, oversize→REST, and BufferError retry.

## ⚠️ Main drawback: DELETE MCPs are not supported over Kafka

This is the primary limitation and the reason for the hybrid fallback.

- GMS **rejects `DELETE` change types on the async Kafka ingestion path** —
  `MCPItem.CHANGE_TYPES` only allows `UPSERT`, `UPDATE`, `CREATE`, `CREATE_ENTITY`. A
  `DELETE` MCP produced to the MCP topic fails downstream in the consumer.
- On the emitter side, `DELETE` is inherently a **synchronous REST** operation
  (`/entities?action=delete` or `DELETE /openapi/v3/entity/...`); the Kafka emitter has no
  delete semantics.

Practical impact is **low** but must be understood:

| Delete kind | Path | Kafka-safe? |
|---|---|---|
| Stateful ingestion soft-delete (stale-entity removal) | `status` aspect **UPSERT** | ✅ yes — works over Kafka |
| Hard delete (`datahub delete` CLI / graph client) | REST, bypasses the sink entirely | ✅ N/A — never used the sink |
| True `DELETE` change-type MCP through the sink | must be synchronous REST | ❌ breaks over pure Kafka |

No standard ingestion source or transformer emits a true `DELETE`/`RESTATE` change-type MCP
through the sink today, so the fallback rarely fires in practice — it's cheap insurance for
any recipe that ever does. Evidence:

- **GMS rejects DELETE/RESTATE on the async path** — `MCPItem.CHANGE_TYPES` allows only
  UPSERT/UPDATE/CREATE/CREATE_ENTITY:
  https://github.com/datahub-project/datahub/blob/master/entity-registry/src/main/java/com/linkedin/metadata/aspect/batch/MCPItem.java#L20-L22
- **PATCH is accepted server-side** (so it stays on Kafka) — `isValidChangeType` / `supportsPatch`:
  https://github.com/datahub-project/datahub/blob/master/entity-registry/src/main/java/com/linkedin/metadata/aspect/batch/MCPItem.java#L62-L81
- **The one source referencing DELETE drops it** — datahub-to-datahub source:
  https://github.com/datahub-project/datahub/blob/master/metadata-ingestion/src/datahub/ingestion/source/datahub/datahub_source.py#L187
- **Stateful-ingestion soft-deletes are UPSERTs** (Kafka-safe) — stale-entity removal emits `Status(removed=true)`:
  https://github.com/datahub-project/datahub/blob/master/metadata-ingestion/src/datahub/ingestion/source/state/stale_entity_removal_handler.py#L214-L218
- **DELETE is inherently synchronous REST** on the emitter side (`/entities?action=delete`):
  https://github.com/datahub-project/datahub/blob/master/metadata-ingestion/src/datahub/emitter/rest_emitter.py#L864

**Secondary trade-off:** the REST path benefits from GMS-side per-request rate limiting
(429 backpressure). The Kafka path bypasses GMS, so backpressure shifts to the producer
queue (handled here) and, operationally, to MCP-topic consumer-group lag monitoring.

## Delivery model & throughput (async, commit-safe)

The `datahub-kafka` sink previously flushed the producer **after every work unit** (a work
unit is ~one record), forcing a synchronous broker round-trip per record — this defeats
Kafka's async batching and is the opposite of the "offload GMS / high throughput" goal of
making it the managed default. But the per-WU flush was *incidentally load-bearing*: it was
the only pre-commit point that confirmed delivery before stateful ingestion's global state
checkpoint (`process_commits`, gated on `sink.get_report().failures`). Naively deleting it
would let checkpoints commit for writes that never landed → silent data loss / skipped
entities next run.

Fix: keep the correctness guarantee, drop the per-record cost by flushing **once**, at the
right place.
- `handle_work_unit_end` no longer flushes (that ABC hook is deprecated anyway).
- New `Sink.flush()` (default no-op; async sinks override); the pipeline calls
  `self.sink.flush()` immediately **before** `process_commits()`.
- `DatahubKafkaSink.flush()` drains the producer and reports any undelivered messages as a
  failure — so the commit gate sees them. `close()` reuses it as a final safety net.
- `flush()` is bounded by `max_queue_full_block_seconds` (no infinite hang) and its
  undelivered count is acted on (no silent success-with-loss).

| Aspect | Before | After |
|---|---|---|
| Flush cadence | after every work unit (~every record) | once before commit (+ final at close) |
| Batching / `linger.ms` | defeated / no-op | active — real async throughput |
| Throughput | effectively synchronous | async, high |
| Producer buffer | librdkafka ~1 GiB default | 128 MiB, env-tunable (`DATAHUB_KAFKA_SINK_QUEUE_MAX_*`) |
| Backpressure | implicit (blocked every flush) | `BufferError` → block + retry, 300s deadline |
| Commit correctness | via per-WU flush | via single pre-commit `sink.flush()` (same guarantee) |
| Undelivered at shutdown | flush return discarded → silent loss, run = SUCCESS | reported as failure → run fails |

Why it's still correct: the checkpoint is a single global event, so the invariant is "all
deliveries confirmed before `process_commits`", not "confirmed after every record". One
flush right before the commit gate satisfies it identically, while `produce()` + background
batching + `poll(0)` (drains callbacks) run async in between. Non-Kafka sinks inherit the
no-op `flush()` — unchanged behavior.

> **Discovery (out of scope, noted for follow-up):** the REST sink has the *same* latent
> pre-commit window — its `handle_work_unit_end` is a no-op and it writes via a bounded async
> executor (`max_pending_requests`), so up to N requests can still be in flight when
> `process_commits` runs. It is not synchronous, contrary to a common assumption. This PR
> does not change the REST sink, but the new `Sink.flush()` hook now gives it a clean place
> to close that gap later (a `flush()` override that drains its executor and reports failures).

## Safety / rollout

- **Off by default.** Unset env ⇒ byte-identical `datahub-rest` behavior.
- **Kill switch.** Unset `DATAHUB_INGESTION_DEFAULT_SINK` (or set it to anything but
  `datahub-kafka`) ⇒ instant revert to REST; no redeploy of ingestion code required.
- Only affects recipes with **no explicit `sink:`**. Any recipe that configures a sink is
  untouched.


david-leifker (approved):
Suggest getting one review from ingestion on this, otherwise lgtm!

max-datahub:
Really like the direction here — moving ingest write load off GMS's sync path is a great lever, and the off-by-default gating is thoughtful.

One thing I'd love to talk through before this backs a Kafka default: for a REST-only remote executor, the two-key gate may not be as independent as it looks. `DATAHUB_EXECUTOR_MANAGED` really means "spawned by the managed executor" — but a remote executor is one too, so the marker separates CLI-vs-executor, not in-cluster (has Kafka) vs remote (REST-only), which is the axis that matters here. Both vars can go true from two reasonable-in-isolation decisions: flipping `DATAHUB_INGESTION_DEFAULT_SINK=kafka` cluster-wide for the offload win, and treating remote executors as "managed." At that point every no-sink recipe on a REST-only executor defaults to a cluster-internal broker it can't reach.

And the failure isn't always clean: the broker isn't probed at init (the registry client, Avro serializer, and producer all connect lazily), so a small run buffers fine and only fails at `flush()` after ~300s (`max_queue_full_block_seconds`). Worse, if the broker is reachable-but-wrong (wrong cluster, or a topic nothing consumes), `produce()` and the delivery callback both succeed, `flush()` reports 0 undelivered, and the run reports SUCCESS while nothing ever lands in GMS.

Two things that'd make me comfortable here:

1. **Make it structurally impossible for a REST-only executor to inherit the Kafka default** — e.g. have the remote executor set an explicit `datahub-rest` sink. Recipes with an explicit sink are already untouched by this PR, so no combination of env vars can flip it.
2. **Fail fast when Kafka is unreachable** — a short, bounded broker + schema-registry check at pipeline init, so a misconfig dies in seconds at startup instead of a ~300s flush stall. This covers the slow-failure case; the silent reachable-but-wrong-broker case is really only closed by #1.

Happy to help with either if useful!

metadata-ingestion/src/datahub/configuration/env_vars.py
```diff
@@ -64,6 +64,66 @@ def get_gms_base_path() -> str:
...
+
+
+def get_kafka_queue_max_messages() -> int:
+    """Per-producer local queue size cap (message count) for the default Kafka sink."""
+    return _get_int_env("DATAHUB_KAFKA_QUEUE_MAX_MESSAGES", 20000)
+
+
+def get_kafka_linger_ms() -> int:
+    """Producer linger.ms (send batching window) for the default Kafka sink."""
+    return _get_int_env("DATAHUB_KAFKA_LINGER_MS", 100)
```
skrydal: In the names of the functions and env variables we should indicate those are the settings for kafka **ingestion sink**. There might be other producers present in environment where SDK is executed.

metadata-ingestion/src/datahub/emitter/kafka_emitter.py
```diff
@@ -145,6 +169,55 @@ def emit(
         else:
             return self.emit_mce_async(item, callback or _error_reporting_callback)
 
+    @staticmethod
+    def _produce_with_backpressure(
+        producer: SerializingProducer,
+        poll_timeout_seconds: float = _QUEUE_FULL_POLL_SECONDS,
+        max_block_seconds: float = _MAX_QUEUE_FULL_BLOCK_SECONDS,
+        **produce_kwargs: Any,
```
skrydal: This internal function is called only twice in this file and arguments which follow max_block_seconds are well defined (`key`, `topic`, `value`, `on_delivery`), why would we hide them behind `kwargs`? We lose typing assurances.

metadata-ingestion/src/datahub/emitter/kafka_emitter.py
```diff
@@ -145,6 +169,55 @@ def emit(
...
+        deadline = time.monotonic() + max_block_seconds
+        polls = 0
+        while True:
+            try:
+                producer.produce(**produce_kwargs)
+                return
+            except BufferError:
+                # Count the poll first so a single transient BufferError does not
+                # warn immediately (polls==0); the first warning fires after
+                # _QUEUE_FULL_WARN_EVERY_N_POLLS actual polls.
```
skrydal: there is absolutely no need for this comment

metadata-ingestion/src/datahub/emitter/kafka_emitter.py
```diff
@@ -145,6 +169,55 @@ def emit(
...
+                polls += 1
+                if polls % _QUEUE_FULL_WARN_EVERY_N_POLLS == 0:
+                    logger.warning(
+                        "Kafka producer queue full; blocking to apply backpressure. "
+                        "If this persists the broker may be unreachable."
+                    )
+                producer.poll(poll_timeout_seconds)
+                # Bail out once total blocked time exceeds the deadline, so an
+                # unreachable broker fails the run instead of blocking forever.
```
skrydal: Behavior is quite obvious, it is a simple if + raise, does not need 2 lines of comments

sgomezvillamor:
Nice work on the commit-correctness reasoning and the DELETE fallback guards. A few things before this backs a managed default:

**⚠️ `DATAHUB_EXECUTOR_MANAGED` coupling.** +1 on @max-datahub's point. This marker separates CLI-vs-executor, but the axis that actually matters for Kafka safety is in-cluster (can reach the broker) vs. remote/REST-only. A remote executor is still "managed," so flipping `DATAHUB_INGESTION_DEFAULT_SINK=kafka` cluster-wide plus treating remote executors as managed lands every no-sink recipe on a broker it can't reach. Having remote/REST-only executors set an explicit `datahub-rest` sink makes this structurally impossible and drops the confusing two-key gate.

**⚠️ Missing user-facing docs.** For an env-var-gated feature the env vars are the API, and none are documented — the selector + managed marker, `KAFKA_BOOTSTRAP_SERVER`, the `DATAHUB_KAFKA_QUEUE_MAX_*`/`LINGER_MS` knobs, the DELETE-over-Kafka limitation and `rest_fallback`, and how to monitor it (consumer lag, 300s flush deadline). (Also: the PR description says bootstrap comes from `DATAHUB_KAFKA_BOOTSTRAP`, but the code reads `KAFKA_BOOTSTRAP_SERVER`.)

**💡 Observability is weak.** The backpressure loop only warns after ~30 polls (~30s), so normal throttling is silent, and there's no counter for it. Suggest a couple of `SinkReport` fields — backpressure engagements + total blocked time, plus undelivered count — so throttling shows up in the run report rather than only in logs. That also helps distinguish "healthy but throttled" from "broker stalled."

---

metadata-ingestion/src/datahub/ingestion/run/pipeline.py
```diff
@@ -146,6 +154,67 @@ def _make_default_rest_sink(ctx: PipelineContext) -> DatahubRestSink:
...
 
+def _make_default_kafka_sink(ctx: PipelineContext) -> Sink:
+    # Imported lazily so the (heavier) Kafka/confluent dependencies are only
+    # loaded when the Kafka default sink is actually selected.
+    from datahub.cli import config_utils
+    from datahub.ingestion.sink.datahub_kafka import (
+        DatahubKafkaSink,
+        KafkaSinkConfig,
+    )
+    from datahub.ingestion.sink.datahub_rest import DatahubRestSinkConfig
```
skrydal: We are already importing `DatahubRestSink` from this module, any gain here is completely hallucinated.

metadata-ingestion/src/datahub/ingestion/run/pipeline.py
```diff
@@ -199,14 +268,47 @@ def __init__(
...
+                # This requires BOTH env vars:
+                #   DATAHUB_INGESTION_DEFAULT_SINK=kafka  (the selector)
+                #   DATAHUB_EXECUTOR_MANAGED=true         (managed-run marker)
+                # The marker is set only by the managed executor on the
+                # ingestion subprocesses it spawns. Because managed UI ingestion
+                # and an interactive `datahub ingest` share this same entrypoint,
+                # requiring the marker guarantees a manual/CLI run never flips to
+                # Kafka -- even if DATAHUB_INGESTION_DEFAULT_SINK leaks into its
+                # environment. Unset/any-other-value preserves the historical
+                # datahub-rest behavior exactly.
```
skrydal: This sounds like insanity. What does it mean that `DATAHUB_INGESTION_DEFAULT_SINK` **leaks** into the environment? How does it happen? If it happens - why wouldn't `DATAHUB_EXECUTOR_MANAGED=true` "leak" too?

This is fundamentally wrong idea, complicating maintenance and this is actually introducing a "leak" - we are leaking concerns between independent architectural layers. `acryl-datahub` shouldn't have idea how it is run. It is enough to use a single variable to control behavior introduced by this PR - keep `DATAHUB_INGESTION_DEFAULT_SINK`, with default value of `rest`. Then we will enforce this and only this variable in places where we wish to set `kafka` sink. Remove the other, superfluous variable.

metadata-ingestion/src/datahub/ingestion/run/pipeline.py
```diff
@@ -146,6 +154,67 @@ def _make_default_rest_sink(ctx: PipelineContext) -> DatahubRestSink:
     return DatahubRestSink(ctx, sink_config)
 
 
+def _make_default_kafka_sink(ctx: PipelineContext) -> Sink:
+    # Imported lazily so the (heavier) Kafka/confluent dependencies are only
+    # loaded when the Kafka default sink is actually selected.
+    from datahub.cli import config_utils
```
skrydal: This one is also effectively loaded, no reason to guard that

metadata-ingestion/src/datahub/ingestion/run/pipeline.py
```diff
@@ -236,10 +338,18 @@ def __init__(
                 flags.progress_report_max_warnings,
             )
 
-            if self.graph is None and isinstance(self.sink, DatahubRestSink):
+            if self.graph is None:
                 with _add_init_error_context("setup default datahub client"):
-                    self.graph = self.sink.emitter.to_graph()
-                    self.graph.test_connection()
+                    if isinstance(self.sink, DatahubRestSink):
+                        self.graph = self.sink.emitter.to_graph()
```
skrydal: Can we have DatahubRestSink implement `to_graph()` just like Kafka sink, so we do not need this `isinstance` check here?

metadata-ingestion/src/datahub/ingestion/run/pipeline.py
```diff
@@ -199,14 +324,47 @@ def __init__(
...
+                #   DATAHUB_EXECUTOR_MANAGED=true         (managed-run marker)
+                # The marker is set only by the managed executor on the
+                # ingestion subprocesses it spawns. Because managed UI ingestion
+                # and an interactive `datahub ingest` share this same entrypoint,
+                # requiring the marker guarantees a manual/CLI run never flips to
+                # Kafka -- even if DATAHUB_INGESTION_DEFAULT_SINK leaks into its
+                # environment. Unset/any-other-value preserves the historical
+                # datahub-rest behavior exactly.
+                default_sink = get_ingestion_default_sink().strip().lower()
+                if default_sink not in ("rest", "kafka"):
```
skrydal: I think `get_ingestion_default_sink` is the proper place for the validation of the value to live in (and decision whether to raise or not for that).

As for strings themself - `kafka` and `rest` do not match what we have in YAML `sink:` config, we use `datahub-kafka` and `datahub-rest` there. Please keep allowed values of default sink in-line with that convention.
We are clearly hit by lack of enums here (set of sinks is dynamic and open-ended). You can introduce an enum for them to be used across this file and config (e.g. DefaultSinkType enum class), but it is not strictly necessary in this PR, as unfortunately, we lack them throughout the codebase.

metadata-ingestion/src/datahub/ingestion/sink/datahub_rest.py
```diff
@@ -210,6 +210,16 @@ def __post_init__(self) -> None:
...
 
+    @classmethod
+    def make_emitter(cls, config: DatahubRestSinkConfig) -> DataHubRestEmitter:
+        """Build a REST emitter from sink config.
+
+        Public entry point so other sinks (e.g. the Kafka sink's REST fallback)
+        can reuse the exact emitter construction without reaching into a private
+        method.
+        """
+        return cls._make_emitter(config)
```
skrydal: why can't we just make `_make_emitter` public instead?

skrydal (changes requested):
General idea is great, please see my comments though.
I also am concerned with Kafka-production issues. For example, what if we want to produce an aspect bigger than 16M? Rest emitter will allow for it, with GMS failing due to the size. What will happen with Kafka?

Dutt23:
> Really like the direction here — moving ingest write load off GMS's sync path is a great lever, and the off-by-default gating is thoughtful.
> 
> One thing I'd love to talk through before this backs a Kafka default: for a REST-only remote executor, the two-key gate may not be as independent as it looks. `DATAHUB_EXECUTOR_MANAGED` really means "spawned by the managed executor" — but a remote executor is one too, so the marker separates CLI-vs-executor, not in-cluster (has Kafka) vs remote (REST-only), which is the axis that matters here. Both vars can go true from two reasonable-in-isolation decisions: flipping `DATAHUB_INGESTION_DEFAULT_SINK=kafka` cluster-wide for the offload win, and treating remote executors as "managed." At that point every no-sink recipe on a REST-only executor defaults to a cluster-internal broker it can't reach.
> 
> And the failure isn't always clean: the broker isn't probed at init (the registry client, Avro serializer, and producer all connect lazily), so a small run buffers fine and only fails at `flush()` after ~300s (`max_queue_full_block_seconds`). Worse, if the broker is reachable-but-wrong (wrong cluster, or a topic nothing consumes), `produce()` and the delivery callback both succeed, `flush()` reports 0 undelivered, and the run reports SUCCESS while nothing ever lands in GMS.
> 
> Two things that'd make me comfortable here:
> 
> 1. **Make it structurally impossible for a REST-only executor to inherit the Kafka default** — e.g. have the remote executor set an explicit `datahub-rest` sink. Recipes with an explicit sink are already untouched by this PR, so no combination of env vars can flip it.
> 2. **Fail fast when Kafka is unreachable** — a short, bounded broker + schema-registry check at pipeline init, so a misconfig dies in seconds at startup instead of a ~300s flush stall. This covers the slow-failure case; the silent reachable-but-wrong-broker case is really only closed by #1.
> 
> Happy to help with either if useful!

Thanks, this is a sharp read of the failure modes. Two updates since you wrote this:

The two-key gate is gone. Per other review feedback I dropped DATAHUB_EXECUTOR_MANAGED entirely — selection is now a single DATAHUB_INGESTION_DEFAULT_SINK (default datahub-rest). So the "two independent-looking keys both go true" path you described no longer exists. But your underlying point stands: a cluster-wide DATAHUB_INGESTION_DEFAULT_SINK=datahub-kafka can still reach a REST-only executor that can't hit the broker.

Fail-fast is in (your #2). There's now a bounded broker + schema-registry probe at pipeline init — [_kafka_endpoints_reachable](https://github.com/datahub-project/datahub/blob/e574d26a992c30d58c1dc2939bc6fca6e7b159f9/metadata-ingestion/src/datahub/ingestion/run/pipeline.py#L165): a list_topics metadata fetch against the bootstrap server and a GET on the registry's /subjects, each bounded by DATAHUB_KAFKA_SINK_INIT_PROBE_TIMEOUT (default 10s). If either fails, the run degrades to the REST default sink instead of buffering and stalling ~300s at flush() ([pipeline.py:287](https://github.com/datahub-project/datahub/blob/e574d26a992c30d58c1dc2939bc6fca6e7b159f9/metadata-ingestion/src/datahub/ingestion/run/pipeline.py#L287)). Covered by test_configure_kafka_default_unreachable_broker_falls_back_to_rest and ..._unreachable_registry_falls_back_to_rest.

The reachable-but-wrong-broker silent-success case (your #1) is not fully closed here, and I don't think it can be from inside acryl-datahub — a valid broker on the wrong cluster answers the probe, so this layer can't tell the difference (called out in the [docstring](https://github.com/datahub-project/datahub/blob/e574d26a992c30d58c1dc2939bc6fca6e7b159f9/metadata-ingestion/src/datahub/ingestion/run/pipeline.py#L175)). That's the remote-executor structural piece — as discussed in the Slack thread, the plan is to have the remote/REST-only executor set an explicit datahub-rest sink (recipes with an explicit sink are untouched by this PR, so no env combination can flip them). I'll handle that in a separate PR rather than widening this one.

Appreciate the offer to help — will tag you on the remote-executor PR.

metadata-ingestion/src/datahub/configuration/env_vars.py
Dutt23: Done — renamed to get_kafka_sink_* / DATAHUB_KAFKA_SINK_*: [env_vars.py:129](https://github.com/datahub-project/datahub/blob/e574d26a992c30d58c1dc2939bc6fca6e7b159f9/metadata-ingestion/src/datahub/configuration/env_vars.py#L129)

metadata-ingestion/src/datahub/emitter/kafka_emitter.py
Dutt23: Done — explicit keyword-only params, no **kwargs: [kafka_emitter.py:189](https://github.com/datahub-project/datahub/blob/e574d26a992c30d58c1dc2939bc6fca6e7b159f9/metadata-ingestion/src/datahub/emitter/kafka_emitter.py#L189)

metadata-ingestion/src/datahub/emitter/kafka_emitter.py
Dutt23: Done, removed.

metadata-ingestion/src/datahub/emitter/kafka_emitter.py
Dutt23: Done, removed.

metadata-ingestion/src/datahub/ingestion/run/pipeline.py
Dutt23: Done — dropped the lazy import, uses the module-level one ([pipeline.py:53](https://github.com/datahub-project/datahub/blob/e574d26a992c30d58c1dc2939bc6fca6e7b159f9/metadata-ingestion/src/datahub/ingestion/run/pipeline.py#L53)).

metadata-ingestion/src/datahub/ingestion/run/pipeline.py
Dutt23: removed DATAHUB_EXECUTOR_MANAGED. Single DATAHUB_INGESTION_DEFAULT_SINK (default datahub-rest) now, resolved in [get_ingestion_default_sink](https://github.com/datahub-project/datahub/blob/e574d26a992c30d58c1dc2939bc6fca6e7b159f9/metadata-ingestion/src/datahub/configuration/env_vars.py#L80).

metadata-ingestion/src/datahub/ingestion/run/pipeline.py
Dutt23: Done — removed; config_utils is imported at module level ([pipeline.py:19](https://github.com/datahub-project/datahub/blob/e574d26a992c30d58c1dc2939bc6fca6e7b159f9/metadata-ingestion/src/datahub/ingestion/run/pipeline.py#L19)).

metadata-ingestion/src/datahub/ingestion/run/pipeline.py
Dutt23: Done — to_graph() on the base + both sinks; pipeline just calls self.sink.to_graph() ([sink.py:180](https://github.com/datahub-project/datahub/blob/e574d26a992c30d58c1dc2939bc6fca6e7b159f9/metadata-ingestion/src/datahub/ingestion/api/sink.py#L180)). The one remaining isinstance at [pipeline.py:378](https://github.com/datahub-project/datahub/blob/e574d26a992c30d58c1dc2939bc6fca6e7b159f9/metadata-ingestion/src/datahub/ingestion/run/pipeline.py#L378) only labels sink_type after a possible Kafka→REST degrade, not behavior.

metadata-ingestion/src/datahub/ingestion/run/pipeline.py
Dutt23: Done — validation lives in [get_ingestion_default_sink](https://github.com/datahub-project/datahub/blob/e574d26a992c30d58c1dc2939bc6fca6e7b159f9/metadata-ingestion/src/datahub/configuration/env_vars.py#L80) (warn + fall back to datahub-rest, no raise); values are datahub-rest/datahub-kafka to match the sink: types. Kept as constants rather than an enum per your note it's optional here.

metadata-ingestion/src/datahub/ingestion/sink/datahub_rest.py
Dutt23: Done — public [make_emitter](https://github.com/datahub-project/datahub/blob/e574d26a992c30d58c1dc2939bc6fca6e7b159f9/metadata-ingestion/src/datahub/ingestion/sink/datahub_rest.py#L227).

Dutt23:
> General idea is great, please see my comments though. I also am concerned with Kafka-production issues. For example, what if we want to produce an aspect bigger than 16M? Rest emitter will allow for it, with GMS failing due to the size. What will happen with Kafka?

Kafka producer caps at message.max.bytes — 5 MiB, not 16, because it can't exceed the broker's, and DataHub ships the broker at KAFKA_MESSAGE_MAX_BYTES=5242880 ([docker.env:9](https://github.com/datahub-project/datahub/blob/e574d26a992c30d58c1dc2939bc6fca6e7b159f9/docker/broker/env/docker.env#L9)). A higher producer cap buys nothing — broker rejects >5 MiB regardless; 16 needs a broker+topic change (both env-tunable via DATAHUB_KAFKA_SINK_MAX_MESSAGE_BYTES).

Oversize handling: produce() raises MSG_SIZE_TOO_LARGE → mapped to MessageTooLargeError ([kafka_emitter.py:259](https://github.com/datahub-project/datahub/blob/e574d26a992c30d58c1dc2939bc6fca6e7b159f9/metadata-ingestion/src/datahub/emitter/kafka_emitter.py#L259)) → degrades to REST fallback ([datahub_kafka.py:432](https://github.com/datahub-project/datahub/blob/e574d26a992c30d58c1dc2939bc6fca6e7b159f9/metadata-ingestion/src/datahub/ingestion/sink/datahub_kafka.py#L432)). >16 MiB fails at GMS on either path — reported, never silent. No fallback → also reported, not dropped. EnsureAspectSizeProcessor trims known-large aspects upstream. Tests: test_kafka_sink.py::test_kafka_sink_oversize_*.

Dutt23:
> Nice work on the commit-correctness reasoning and the DELETE fallback guards. A few things before this backs a managed default:
> 
> **⚠️ `DATAHUB_EXECUTOR_MANAGED` coupling.** +1 on @max-datahub's point. This marker separates CLI-vs-executor, but the axis that actually matters for Kafka safety is in-cluster (can reach the broker) vs. remote/REST-only. A remote executor is still "managed," so flipping `DATAHUB_INGESTION_DEFAULT_SINK=kafka` cluster-wide plus treating remote executors as managed lands every no-sink recipe on a broker it can't reach. Having remote/REST-only executors set an explicit `datahub-rest` sink makes this structurally impossible and drops the confusing two-key gate.
> 
> **⚠️ Missing user-facing docs.** For an env-var-gated feature the env vars are the API, and none are documented — the selector + managed marker, `KAFKA_BOOTSTRAP_SERVER`, the `DATAHUB_KAFKA_QUEUE_MAX_*`/`LINGER_MS` knobs, the DELETE-over-Kafka limitation and `rest_fallback`, and how to monitor it (consumer lag, 300s flush deadline). (Also: the PR description says bootstrap comes from `DATAHUB_KAFKA_BOOTSTRAP`, but the code reads `KAFKA_BOOTSTRAP_SERVER`.)
> 
> **💡 Observability is weak.** The backpressure loop only warns after ~30 polls (~30s), so normal throttling is silent, and there's no counter for it. Suggest a couple of `SinkReport` fields — backpressure engagements + total blocked time, plus undelivered count — so throttling shows up in the run report rather than only in logs. That also helps distinguish "healthy but throttled" from "broker stalled."
> 

DATAHUB_EXECUTOR_MANAGED coupling. Removed it entirely — now a single DATAHUB_INGESTION_DEFAULT_SINK (default datahub-rest), no two-key gate ([env_vars.py:80](https://github.com/datahub-project/datahub/blob/e574d26a992c30d58c1dc2939bc6fca6e7b159f9/metadata-ingestion/src/datahub/configuration/env_vars.py#L80)). The structural "remote/REST-only executor sets an explicit datahub-rest sink" piece is a coordinator-side change — handling it in a separate PR per the Slack thread. This PR also degrades to REST if the broker/registry is unreachable at init ([pipeline.py:287](https://github.com/datahub-project/datahub/blob/e574d26a992c30d58c1dc2939bc6fca6e7b159f9/metadata-ingestion/src/datahub/ingestion/run/pipeline.py#L287)).

Docs. Added in [sink_docs/datahub.md](https://github.com/datahub-project/datahub/blob/e574d26a992c30d58c1dc2939bc6fca6e7b159f9/metadata-ingestion/sink_docs/datahub.md#L190) — all env vars, the DELETE-over-Kafka limitation + rest_fallback, ordering caveat, and failure/monitoring notes. Code + docs use KAFKA_BOOTSTRAP_SERVER; I'll fix the PR description (the DATAHUB_KAFKA_BOOTSTRAP mention was stale).

Observability. Added KafkaSinkReport fields — kafka_backpressure_engagements, kafka_backpressure_blocked_seconds, kafka_undelivered ([datahub_kafka.py:179](https://github.com/datahub-project/datahub/blob/e574d26a992c30d58c1dc2939bc6fca6e7b159f9/metadata-ingestion/src/datahub/ingestion/sink/datahub_kafka.py#L179)) — so throttling and drops show in the run report, and the backpressure loop now warns on the first block, not after ~30 polls ([kafka_emitter.py:233](https://github.com/datahub-project/datahub/blob/e574d26a992c30d58c1dc2939bc6fca6e7b159f9/metadata-ingestion/src/datahub/emitter/kafka_emitter.py#L233)).

metadata-ingestion/src/datahub/ingestion/run/pipeline.py
```diff
@@ -236,10 +410,15 @@ def __init__(
...
 
-            if self.graph is None and isinstance(self.sink, DatahubRestSink):
+            if self.graph is None:
                 with _add_init_error_context("setup default datahub client"):
-                    self.graph = self.sink.emitter.to_graph()
-                    self.graph.test_connection()
+                    # Sink.to_graph() (default None) lets a default sink provide a
+                    # GMS client so features like stateful ingestion still work:
+                    # datahub-rest returns its emitter's graph, datahub-kafka its
+                    # REST fallback's; other sinks return None (no graph).
```
skrydal: There is no need for this comment.

metadata-ingestion/src/datahub/ingestion/sink/datahub_rest.py
```diff
@@ -319,6 +331,12 @@ def emitter(self) -> DataHubRestEmitter:
             )
         return thread_local.emitter
 
+    def to_graph(self) -> Optional["DataHubGraph"]:
+        # Lets the pipeline obtain a GMS client from the default sink
+        # polymorphically (no isinstance branch): the REST emitter already
+        # wraps a graph client.
```
skrydal: No need for this comment, stating the obvious.

metadata-ingestion/tests/unit/test_rest_sink.py
```diff
@@ -467,11 +467,26 @@ def test_sync_origin_opt_in_passed_through_in_all_modes():
...
+
+def test_rest_sink_config_accepts_client_config_dump():
+    # The Kafka default sink builds its REST fallback via
+    # DatahubRestSinkConfig(**DatahubClientConfig(...).model_dump()). Under
+    # ConfigModel's extra="forbid", that round-trip breaks the moment a base
+    # client-config field isn't also a DatahubRestSinkConfig field. Guard it so
+    # such a drift fails here (a targeted unit test) instead of at runtime on
+    # every Kafka-default ingestion run.
+    from datahub.ingestion.graph.config import DatahubClientConfig
+    from datahub.ingestion.sink.datahub_rest import DatahubRestSinkConfig
```
skrydal: What's the need to have those imports inside of the function in the tests?

smoke-test/tests/test_kafka_default_sink.py
```diff
@@ -0,0 +1,69 @@
...
+):
+    """Same no-sink recipe run under the REST default and the managed Kafka
+    default: each must select the expected sink and still land metadata in GMS.
+
+    The Kafka arm exercises the exact flow UI ingestion uses; only the two env
+    markers differ from the REST arm.
+    """
+    # Clean slate so a passing existence check proves this run delivered.
+    graph_client.hard_delete_entity(DATASET_URN)
+    wait_for_writes_to_sync()
```
skrydal, line 57: Imho we should add assertion here, to make sure setup was done correctly:

```suggestion
    graph_client.hard_delete_entity(DATASET_URN)
    wait_for_writes_to_sync()
    assert not graph_client.exists(DATASET_URN)
```

skrydal (approved):
LGTM, please see the remaining comments.

metadata-ingestion/src/datahub/ingestion/sink/datahub_rest.py
Dutt23: Removed.

metadata-ingestion/tests/unit/test_rest_sink.py
Dutt23: Removed.

Dutt23: Hoisted to the top of the file: [test_rest_sink.py:14](https://github.com/datahub-project/datahub/blob/e574d26a992c30d58c1dc2939bc6fca6e7b159f9/metadata-ingestion/tests/unit/test_rest_sink.py#L14)

smoke-test/tests/test_kafka_default_sink.py
Dutt23: Added

PR #18187, opened 2026-07-06
fix(ingestion): harden the OAuth token provider (clamp, origin guard, replay guard, degradation warnings)

max-datahub:
## Summary

Hardens the OAuth token-provider layer added in #17939. Four fixes, one file:

- **Refresh-buffer clamp**: `CachingTokenProvider` clamps its refresh buffer to half the token's observed lifetime, so IdPs issuing tokens shorter than the buffer (e.g. Keycloak's default 300s) still get useful caching instead of one IdP round trip per request.
- **Origin-strict 401 retry**: the retry hook refuses to re-attach the token after a redirect off the original (scheme, host, port) origin — matching (slightly stricter than) requests' own `should_strip_auth` rules, so an `https→http` downgrade or port change can never leak a fresh bearer token.
- **Non-replayable bodies**: the 401 retry skips streamed request bodies instead of re-sending an empty body under the original `Content-Length`.
- **Loud degradation**: warn once when a token arrives already expired (every request becomes a synchronous IdP round trip) and when `TokenProviderAuth` replaces a pre-existing `Authorization` header.

Split out of #18144 (see the stack there). No behavior change for static-token users.

## Testing

`tests/unit/auth/test_token_provider.py` — 18 tests covering all four behaviors; ruff + mypy clean.

## The stack

| Order | PR | Base | Notes |
|---|---|---|---|

### DIFF
--- a/metadata-ingestion/src/datahub/emitter/token_provider.py
+++ b/metadata-ingestion/src/datahub/emitter/token_provider.py
@@ -5,7 +5,8 @@ import threading
 import time
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from typing import Callable, Optional
+from typing import Callable, Optional, Tuple
+from urllib.parse import urlparse
 
 import requests
 
@@ -83,25 +84,55 @@ class CachingTokenProvider(TokenProvider):
         self._fetch = fetch
         self._refresh_buffer_seconds = refresh_buffer_seconds
         self._lock = threading.Lock()
-        self._cached: Optional[TokenResult] = None
+        self._cached: Optional[Tuple[TokenResult, float]] = None
+        self._warned_nonpositive_lifetime = False
 
     def get_token(self) -> TokenResult:
         with self._lock:
-            if self._cached is not None and not self._is_stale(self._cached):
-                return self._cached
-            self._cached = self._fetch()
-            return self._cached
+            if self._cached is not None and not self._is_stale(*self._cached):
+                return self._cached[0]
+            result = self._fetch()
+            self._cached = (result, time.time())
+            return result
 
     def invalidate(self) -> None:
         with self._lock:
             self._cached = None
 
-    def _is_stale(self, cached: TokenResult) -> bool:
+    def _is_stale(self, cached: TokenResult, fetched_at: float) -> bool:
         # No reported expiry -> always re-fetch. The underlying acquisition is
         # cheap (a projected-token file read, or azure-identity's own caching).
         if cached.expires_at is None:
             return True
-        return time.time() >= (cached.expires_at - self._refresh_buffer_seconds)
+        # Clamp the buffer to half the token's observed lifetime: with an IdP
+        # issuing tokens whose lifetime <= the buffer (e.g. Keycloak's default
+        # 300s access tokens vs the default 300s buffer), a fixed buffer would
+        # mark every token permanently stale and turn each client request into
+        # a synchronous IdP round trip.
+        lifetime = cached.expires_at - fetched_at
+        if lifetime <= 0:
+            # Token arrived already expired (IdP expires_in <= 0, or client
+            # clock ahead of the token response): caching cannot help, so every
+            # request degrades to a synchronous IdP round trip under the
+            # provider lock. That must be loud — once — not silent.
+            if not self._warned_nonpositive_lifetime:
+                logger.warning(
+                    "Token provider returned a token with non-positive lifetime "
+                    "(%.0fs); every request will fetch a fresh token. Check the "
+                    "IdP's token lifetime configuration and client clock skew.",
+                    lifetime,
+                )
+                self._warned_nonpositive_lifetime = True
+            return True
+        buffer = min(self._refresh_buffer_seconds, lifetime / 2)
+        return time.time() >= (cached.expires_at - buffer)
+
+
+def _origin(url: Optional[str]) -> Optional[Tuple[str, Optional[str], Optional[int]]]:
+    if not url:
+        return None
+    parsed = urlparse(url)
+    return (parsed.scheme, parsed.hostname, parsed.port)
 
 
 class TokenProviderAuth(requests.auth.AuthBase):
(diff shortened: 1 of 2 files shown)


metadata-ingestion/src/datahub/emitter/token_provider.py
```diff
@@ -83,25 +84,55 @@ def __init__(
...
+            # provider lock. That must be loud — once — not silent.
+            if not self._warned_nonpositive_lifetime:
+                logger.warning(
+                    "Token provider returned a token with non-positive lifetime "
+                    "(%.0fs); every request will fetch a fresh token. Check the "
+                    "IdP's token lifetime configuration and client clock skew.",
+                    lifetime,
+                )
+                self._warned_nonpositive_lifetime = True
+            return True
```
skrydal, line 128: The comment is long, more complex than actual code and moreover - redundant. We already explain everything in the warning message, why repeating it? Please remove the comment.

Moreover, the statement in the warning is too far fetching (pun intended) - this is a single case of a token arriving with "non-positive" lifetime, we can't reason that the next token will also be arriving already expired.

metadata-ingestion/tests/unit/auth/test_token_provider.py
```diff
@@ -30,17 +31,73 @@ def fetch() -> TokenResult:
...
+        # by reporting an expiry only 40s away.
+        return TokenResult("t", time.time() + 40)
+
+    provider = CachingTokenProvider(fetch, refresh_buffer_seconds=300)
+    provider.get_token()
+    # Simulate the passage of time: shift the recorded fetch time back so the
+    # cached token sits inside its clamped refresh window.
+    assert provider._cached is not None
+    result, fetched_at = provider._cached
+    provider._cached = (result, fetched_at - 60)
```
skrydal: To avoid accessing `_cached` directly, consider using time_machine's `travel` + `shift`. Some of the other tests would benefit from that too.

skrydal (approved):
Approving, please consider the 2 comments I have left, before merging.

metadata-ingestion/src/datahub/emitter/token_provider.py
```diff
@@ -83,25 +84,55 @@ def __init__(
         self._fetch = fetch
         self._refresh_buffer_seconds = refresh_buffer_seconds
         self._lock = threading.Lock()
-        self._cached: Optional[TokenResult] = None
+        self._cached: Optional[Tuple[TokenResult, float]] = None
```
skrydal: Why can't we enhance TokenResult to contain also the time of fetch instead of building a cryptic tuple?

metadata-ingestion/src/datahub/emitter/token_provider.py
```diff
@@ -83,25 +84,55 @@ def __init__(
...
+            return result
 
     def invalidate(self) -> None:
         with self._lock:
             self._cached = None
 
-    def _is_stale(self, cached: TokenResult) -> bool:
+    def _is_stale(self, cached: TokenResult, fetched_at: float) -> bool:
         # No reported expiry -> always re-fetch. The underlying acquisition is
         # cheap (a projected-token file read, or azure-identity's own caching).
```
skrydal: I would keep first sentence but remove the second one, once again - no base for that statement.

metadata-ingestion/src/datahub/emitter/token_provider.py
```diff
@@ -128,6 +173,24 @@ def _handle_401(
...
             return response
+        # Never re-attach the token after a redirect off the original origin:
+        # requests strips Authorization not just on cross-host redirects but also
+        # on same-host scheme downgrades and port changes (should_strip_auth), and
+        # retrying here would hand the GMS credential to that other endpoint —
+        # e.g. over plaintext HTTP after an https->http redirect. Strict
+        # (scheme, host, port) equality is slightly stricter than requests' rule
+        # (it also refuses the benign http->https upgrade); the cost is only a
+        # surfaced 401, never a leaked token.
+        if _origin(response.request.url) != getattr(
```
skrydal: Do we really need a custom function to handle this case? `should_strip_auth` comparison with `_origin` logic:

| redirect | `should_strip_auth` | old `_origin` tuple |
|---|---|---|
| identical | keep | keep |
| https→http downgrade | **strip** | strip |
| port change | **strip** | strip |
| cross host | **strip** | strip |
| http→https upgrade | keep | strip (spurious 401) |
| `https://gms` vs `https://gms:443` | keep | strip (spurious 401) |

Correct me if I am wrong, but the 2 bottom cases would not lead to a token leaking, isn't that right? `_origin` is not a great function to have, but we can keep it if this difference is crucial. It would be cleaner to use mentioned function though.

metadata-ingestion/src/datahub/emitter/token_provider.py
max-datahub: Done in 6c45609602 — comment block removed, and the warning no longer predicts future tokens: it now just reports the already-expired token and points at IdP lifetime config / clock skew.

metadata-ingestion/src/datahub/emitter/token_provider.py
max-datahub: Agreed — `TokenResult` now carries `fetched_at` (stamped by `CachingTokenProvider` via `dataclasses.replace` when the result enters the cache; providers don't set it), and `_cached` is a plain `Optional[TokenResult]`. 6c45609602.

metadata-ingestion/src/datahub/emitter/token_provider.py
max-datahub: Done — kept "No reported expiry -> always re-fetch.", dropped the speculative second sentence. 6c45609602.

metadata-ingestion/src/datahub/emitter/token_provider.py
max-datahub: You're right on both counts — the two divergent cases are availability-only (spurious 401, no leak), so strictness bought nothing. Switched to requests' own rule: the guard now calls `SessionRedirectMixin().should_strip_auth(original_url, response.request.url)` and `_origin` is gone. Added a test for the http→https default-port upgrade now being retried. (The env-sink scoping in the stacked #18144 was updated to the same rule for consistency.) 6c45609602.

metadata-ingestion/tests/unit/auth/test_token_provider.py
max-datahub: Done — the clamped-buffer test now uses `time_machine.travel(..., tick=False)` + `shift()` and no longer touches `_cached`; it also gained an extra assertion at the 40s mark (still cached) that the private-poking version couldn't express. The other time-dependent tests construct their conditions via the public API only, so I left them as-is. 6c45609602.

PR #18188, opened 2026-07-06
fix(ingestion): configured auth reaches the emitter (rest sink, datapack loader, from_emitter)

max-datahub:
## Summary

Three places accepted a declarative `auth:` config (#17939) and then silently dropped it — the client emitted **unauthenticated** under OAuth:

- **`DatahubRestSink._make_emitter`** never forwarded `config.auth` to the emitter.
- **The datapack loader** hand-copied only `server`/`token` into its internal sink config; now a typed `TypedDict` that carries `auth` through.
- **`DataHubGraph.from_emitter`** rebuilt the client config with only the static token, losing the emitter's resolved session auth (now carried over; the narrow re-derivation edge is documented at the call site).

These are bug fixes to the existing #17939 contract: *if you configure `auth:`, it must be used.* Split out of #18144 (see the stack there). Stacked on the token-provider hardening PR (retarget to master once it merges).

## Testing

Regression tests in `tests/unit/auth/test_client_auth.py` (each fails without its fix) + datapack loader suite; ruff + mypy clean.

## The stack

| Order | PR | Base | Notes |
|---|---|---|---|


metadata-ingestion/tests/unit/auth/test_client_auth.py
```diff
@@ -22,3 +27,34 @@ def test_token_and_auth_mutually_exclusive():
...
+
+
+def test_rest_sink_emitter_carries_auth():
+    # Regression: DatahubRestSink._make_emitter must forward config.auth to the
+    # emitter, or the default ingestion sink silently emits unauthenticated in
+    # OAuth mode even though the sink config carries a resolved auth provider.
+    auth = AuthConfig(type="static", config={"token": "t"})
+    sink_config = DatahubRestSinkConfig(server="http://gms:8080", auth=auth)
+    emitter = DatahubRestSink._make_emitter(sink_config)
+    assert isinstance(emitter._session.auth, TokenProviderAuth)
```
skrydal: Wouldn't this assertion hold even if `DatahubRestSink._make_emitter` completely ignored provided config and created TokenProviderAuth using something completely different config? If so - this assertion does not really test behavior claimed by the comment.

skrydal (approved):
LGTM, left a minor comment on a unit test.

metadata-ingestion/tests/unit/auth/test_client_auth.py
max-datahub: Good catch — you're right, the type check alone would pass even if the config were ignored. Pushed 9b0c0fa: it now asserts the resolved provider yields the token from *this* config (`session_auth._provider.get_token().token == "t"`), so it fails if `_make_emitter` drops or substitutes the auth.

PR #18189, opened 2026-07-06
fix(ingestion): poll GMS through the authenticated session

max-datahub:
## Summary

Two event pollers copied `session.headers` into bare `requests.get` calls. Copied headers only carry a **static** token — an OAuth token provider lives in `session.auth` and was silently bypassed (401s under OAuth):

- `DataHubRestEmitter` gains a public **`session` property** — the sanctioned way to make GMS REST calls the SDK doesn't wrap, with per-request authentication applied.
- The `datahub-actions` cloud-events consumer and the unstructured-events consumer now poll through that session, with explicit timeouts (the long-poll endpoint previously had none — a half-open connection would hang the consumer forever).

Split out of #18144 (see the stack there). Independent of the other PRs in the stack.

Note on urgency: the DataHub Cloud Remote Executor imports the `datahub-actions` events consumer for its work-poll loop, so its OAuth support only works end-to-end once this fix ships in an `acryl-datahub-actions` release — ideally the next one.

## Testing

Rewritten consumer tests in both modules assert the polls go through the graph's session and cover the retry/timeout behavior, plus wire-level tests: an OAuth-style credential installed as `session.auth` on a real `requests.Session` with a recording transport adapter, asserting the poll request actually carries the credential (fails on the previous implementation, which bypassed the session). ruff + mypy clean.

## The stack

| Order | PR | Base | Notes |
|---|---|---|---|
| 1 | #18187 token-provider hardening | `master` | merged |
| 2 | #18188 configured auth reaches the emitter | #18187 | merged |
| 3 | #18144 env-based OAuth (`DATAHUB_AUTH_TYPE`) | #18188 | merged |
| 4 | #18192 OAuth CI smoke harness | #18144 | open |
| 5 | this PR | `master` | independent |

### DIFF
--- a/datahub-actions/src/datahub_actions/plugin/source/acryl/datahub_cloud_events_consumer.py
+++ b/datahub-actions/src/datahub_actions/plugin/source/acryl/datahub_cloud_events_consumer.py
@@ -2,7 +2,6 @@ import logging
 import time
 from typing import List, Optional
 
-import requests
 from pydantic import BaseModel, Field
 from requests.exceptions import (
     ChunkedEncodingError,
@@ -143,10 +142,14 @@ class DataHubEventsConsumer:
         # Remove keys where the value is None
         params = {k: v for k, v in params.items() if v is not None}
 
-        # Pass along the session headers from the graph for authentication.
-        headers = dict(self.graph._session.headers)
-
-        response = requests.get(endpoint, params=params, headers=headers)
+        # Poll through the graph's authenticated session: per-request auth (e.g.
+        # an OAuth token provider in session.auth) only applies to requests made
+        # through it. The read timeout must outlast the server-side long poll,
+        # and a timeout is required at all — sessions have no default, so a
+        # half-open connection would otherwise hang this consumer forever.
+        response = self.graph.session.get(
+            endpoint, params=params, timeout=(poll_timeout_seconds or 30) + 30
+        )
         response.raise_for_status()
 
         external_events_response = ExternalEventsResponse.model_validate(
--- a/metadata-ingestion/src/datahub/ingestion/source/unstructured/event_consumer.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/unstructured/event_consumer.py
@@ -167,12 +167,10 @@ class DocumentEventConsumer:
         params = {k: v for k, v in params.items() if v is not None and v != ""}
 
         try:
-            import requests
-
-            headers = dict(self.graph._session.headers)
-            response = requests.get(
-                endpoint, params=params, headers=headers, timeout=30
-            )
+            # Poll through the graph's authenticated session: per-request auth
+            # (e.g. an OAuth token provider in session.auth) only applies to
+            # requests made through it.
+            response = self.graph.session.get(endpoint, params=params, timeout=30)
             response.raise_for_status()
 
             data = response.json()
@@ -225,12 +223,10 @@ class DocumentEventConsumer:
         }
 
         try:
-            import requests
-
-            headers = dict(self.graph._session.headers)
-            response = requests.get(
-                endpoint, params=params, headers=headers, timeout=30
-            )
+            # Poll through the graph's authenticated session: per-request auth
+            # (e.g. an OAuth token provider in session.auth) only applies to
+            # requests made through it.
+            response = self.graph.session.get(endpoint, params=params, timeout=30)
             response.raise_for_status()
 
             data = response.json()
--- a/metadata-ingestion/src/datahub/ingestion/source/unstructured/event_consumer.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/unstructured/event_consumer.py
@@ -167,9 +167,6 @@ class DocumentEventConsumer:
         params = {k: v for k, v in params.items() if v is not None and v != ""}
 
         try:
-            # Poll through the graph's authenticated session: per-request auth
-            # (e.g. an OAuth token provider in session.auth) only applies to
-            # requests made through it.
             response = self.graph.session.get(endpoint, params=params, timeout=30)
             response.raise_for_status()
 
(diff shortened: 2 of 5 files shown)


datahub-actions/tests/unit/test_session_auth_bypass_lint.py
skrydal, line 1: This test, along with all its dependencies in this PR, is against decades of SDLC. I fundamentally disagree with  this style of testing for that particular issue. It exists here solely because Claude wanted to test his own randomly generated code. **Please remove it.** You can write proper, behavioral unit tests which will check whether auth is preserved correctly for OAuth.

skrydal (changes requested):

datahub-actions/tests/unit/test_session_auth_bypass_lint.py
max-datahub: Fair enough — removed the lint entirely (the AST checker, its self-test, and both lint-tests). What's left is the two consumer fixes, the public `session` property, and behavioral tests: the mocked-session tests for retry/timeout behavior, plus wire-level tests that install an OAuth-style credential as `session.auth` on a real session and assert the poll request actually carries it (red on the old implementation, green on the fix).

metadata-ingestion/tests/unit/test_datahub_documents_source.py
```diff
@@ -3556,3 +3554,71 @@ def test_custom_lock_id_overrides_default(self, mock_graph):
...
+        # that is not baked into session.headers is only present if the request
+        # actually goes through the session.
+        request.headers["Authorization"] = "Bearer fresh-oauth-token"
+        return request
+
+    def test_poll_events_carries_session_auth_to_the_wire(self) -> None:
+        from datahub.ingestion.graph.client import DataHubGraph
+        from datahub.ingestion.source.unstructured.event_consumer import (
+            DocumentEventConsumer,
+        )
```
skrydal: Why are those imported inside of the function?

metadata-ingestion/src/datahub/ingestion/source/unstructured/event_consumer.py
```diff
@@ -167,12 +167,10 @@ def poll_events(self, topic: str) -> list[dict[str, Any]]:
...
         try:
-            import requests
-
-            headers = dict(self.graph._session.headers)
-            response = requests.get(
-                endpoint, params=params, headers=headers, timeout=30
-            )
+            # Poll through the graph's authenticated session: per-request auth
+            # (e.g. an OAuth token provider in session.auth) only applies to
+            # requests made through it.
```
skrydal: I don't think it is reasonable to repeat the same comment _every single time we call_ `self.graph.session.get`. I think we are fine without the comments.

datahub-actions/src/datahub_actions/plugin/source/acryl/datahub_cloud_events_consumer.py
```diff
@@ -143,10 +142,14 @@ def _poll_events_impl(
...
-        headers = dict(self.graph._session.headers)
-
-        response = requests.get(endpoint, params=params, headers=headers)
+        # Poll through the graph's authenticated session: per-request auth (e.g.
+        # an OAuth token provider in session.auth) only applies to requests made
+        # through it. The read timeout must outlast the server-side long poll,
+        # and a timeout is required at all — sessions have no default, so a
+        # half-open connection would otherwise hang this consumer forever.
+        response = self.graph.session.get(
+            endpoint, params=params, timeout=(poll_timeout_seconds or 30) + 30
```
skrydal, line 151: Why are we introducing this timeout mechanism here and nowhere else? Shouldn't it be globally enforced?

In general this `.session` field is only slightly better than the previous approach (exposing `_session`). Most proper way would be hiding session object altogether. All those layers do not need to care about internals (how is auth implemented, what headers are preserved, etc.) - the fact that it did, in first place, caused us to need to change so many lines just to get OAuth working. In other way, we should go more in the direction of having something like (note we are calling graph directly):
```
self.graph.get(endpoint, params)
```
Rather than accessing internal field. In case answer to my second question would be "yes", we would just need to apply changes in a single place...

Moreover, seeing that we are polling in several places, I think graph could expose a dedicated function to poll. This way we should be able to remove a lot of repeated logic (and moreover endpoint definition) from those bottom layers.

This refactoring is a bit outside of this PR though. As long as we sort out the timeout situation, we can merge it as it is and refactor in another PR.

skrydal (approved):

metadata-ingestion/tests/unit/test_datahub_documents_source.py
max-datahub: No reason beyond matching the neighboring tests in this file, which import `DocumentEventConsumer` locally — hoisted mine to the top of the module.

metadata-ingestion/src/datahub/ingestion/source/unstructured/event_consumer.py
max-datahub: Removed.

datahub-actions/src/datahub_actions/plugin/source/acryl/datahub_cloud_events_consumer.py
max-datahub: On the timeout: `requests.Session` has no global-timeout concept — the emitter's own calls each pass their configured timeout explicitly, and these raw polls previously had **no timeout at all**, so a half-open connection would hang the consumer forever. The `(poll_timeout_seconds or 30) + 30` shape is because the client read timeout has to outlast the server-side long-poll window, which is per-call by nature.

Agreed on the direction though: a default-timeout on the shared session (applied only when a call doesn't pass one) plus a dedicated `graph.poll_events()`-style API that hides the session and the endpoint from these bottom layers would collapse most of this. That's the follow-up refactor PR — I'll keep this one scoped to the bypass fix as you suggested.

PR #18190, opened 2026-07-06
feat(ingestion): token-provider auth for metadata_change_sync and the CLI version check

max-datahub:
## Summary

Two small surfaces migrate onto the token-provider layer (#17939) outright:

- **`metadata_change_sync`** (datahub-actions): accepts a declarative `auth` config — mutually exclusive with `gms_auth_token` — installed on the emitter session, so cross-instance sync can authenticate with short-lived OAuth tokens that refresh per request.
- **CLI version check** (`upgrade.py`): fetches `/config` through `DataHubGraph` instead of hand-rolling an aiohttp request with a hand-resolved bearer token, so a static token and an `auth=` token provider both work and auth resolution stays in `DataHubGraph.__init__`. The fetch is requests-based, so it runs via `asyncio.to_thread` to keep the event loop free and the asyncio budget enforceable. Still best-effort: failures remain swallowed, the endpoint is public.

## Tradeoff worth knowing

`asyncio.wait_for` can only *abandon* a `to_thread` worker, and `concurrent.futures` joins default-executor threads at interpreter shutdown. So if the fetch outlives the budget, the process now waits for it at exit rather than cancelling cleanly the way the aiohttp coroutine did. The client config bounds it (`timeout_sec=3`, retries off), but on the OAuth path the token mint carries its own 30s timeout in `oidc_client_credentials`, which the client config does not reach.

In practice this needs a stalled GMS or IdP to bite, and the version check is best-effort by design. If we want the exit path strictly bounded too, the options are a daemon thread instead of `asyncio.to_thread`, or plumbing a timeout into the token provider. Happy to do either here or as a follow-up.

## Testing

- `metadata-ingestion/tests/unit/upgrade/` — the version check honors an `auth=` client config, not just a static token.
- `metadata_change_sync` unit tests — the auth provider lands on the emitter session, and `auth` / `gms_auth_token` are mutually exclusive.
- OAuth smoke canary (#18192) gains a case covering the version-check path end to end: env auth → `load_client_config` → `DataHubGraph` → token mint → `/config`. Caveat noted in the test — GMS serves `/config` unauthenticated, so it gates the mint-and-fetch path rather than token acceptance.
- `./gradlew :metadata-ingestion:lint` and `:smoke-test:lint` clean.

## The stack

Split of the original #18144 per review feedback (one concern per PR). Everything else has landed; this is the tail.

| PR | Status |
|---|---|
| #18187 token-provider hardening | merged |
| #18188 configured auth reaches the emitter | merged |
| #18144 env-based OAuth (`DATAHUB_AUTH_TYPE`) | merged |
| #18192 OAuth CI canary | merged |
| #18189 session-bypass fixes + lint | merged |
| #18191 TODO(oauth) markers | closed |
| **#18190 metadata_change_sync + version check** | **this PR** |


metadata-ingestion/src/datahub/upgrade/upgrade.py
```diff
@@ -126,6 +127,14 @@ async def get_github_stats():
             return (latest_server_version, latest_server_date)
 
 
+def _resolve_request_token(client_config: DatahubClientConfig) -> Optional[str]:
+    """Materialize a bearer token for the single version-check request: minted
+    fresh from the token provider under OAuth (auth=), else the static token."""
+    if client_config.auth is not None:
+        return build_token_provider(client_config.auth).get_token().token
+    return client_config.token
```
skrydal: This duplicates logic we already have in `DataHubGraph.__init__` (auth resolution), and it also exposes config retrieval via `.server_config`.

The bigger issue: `get_token()` here is a sync network call, so it blocks the event loop outside the aiohttp request — i.e. it can overrun the timeout that seems to be the whole reason this path is async. There's a hardcoded 30s timeout in the OAuth provider, so it won't hang forever, but it's not tied to the 2s budget.

I propose passing the timeout via the client config to `DataHubGraph` (with retries disabled) and using `graph.server_config` instead — that drops both the hand-rolled token resolution and the raw `/config` call.

One tradeoff to call out: the OAuth path then makes two HTTP calls (token mint + `/config`) instead of one, and only the `/config` call is bounded by the config timeout. If we want a strict 2s guarantee, the OAuth provider could take a timeout parameter too (2s per call, accepting we're no longer doing a single request).

skrydal (changes requested):
Please see the comment, no other concerns.

metadata-ingestion/src/datahub/upgrade/upgrade.py
max-datahub: > Done in 4358ad2e51 — dropped both helpers and switched to `DataHubGraph.server_config`, so auth resolution stays in one place.
> 
> One addition on top of your proposal: `server_config` is sync requests too, so calling it on the loop would block the same way `get_token()` did. I wrapped the fetch in `asyncio.to_thread` instead — with it off the loop, `asyncio.wait_for` actually enforces the ~2s budget no matter what the provider does internally, so I left the OAuth provider timeout alone. The client config gets `timeout_sec=3` / retries off just so an abandoned worker thread exits promptly.
> 
> Also added a small case to the OAuth smoke canary from #18192 covering this path (env auth → `load_client_config` → graph → token mint → `/config`), with the caveat noted in the test that `/config` is auth-exempt on GMS, so it gates the mint+fetch path rather than token acceptance.

metadata-ingestion/src/datahub/upgrade/upgrade.py
```diff
@@ -126,22 +126,17 @@ async def get_github_stats():
...
-
-    async with aiohttp.ClientSession() as session:
-        config_endpoint = f"{gms_url}/config"
-        async with session.get(config_endpoint, headers=headers) as dh_response:
-            dh_response_json = await dh_response.json()
-            return RestServiceConfig(raw_config=dh_response_json)
+def fetch_server_config_via_graph() -> RestServiceConfig:
+    """Fetch /config through DataHubGraph so auth resolution (static token or
+    auth= token provider) stays in one place. Requests-based (sync) — callers
+    on the event loop must run it in a thread."""
```
skrydal: This comment will be misleading for future readers. Let's remove it.

metadata-ingestion/tests/unit/upgrade/test_upgrade_token.py
```diff
@@ -0,0 +1,45 @@
...
+
+    result = asyncio.run(upgrade.get_server_version_stats())
+
+    server_version: Optional[Version] = result[1]
+    assert server_version == Version("1.2.0")
+    # The auth provider config must reach the graph untouched, with the
+    # best-effort bounds applied.
+    assert captured["config"].auth is not None
+    assert captured["config"].retry_max_times == 0
+    assert captured["config"].timeout_sec == 3
```
skrydal: I doubt we need to test for exactly this, what would change if I made this timeout 4 seconds by default?

smoke-test/tests/oauth/test_oauth_cli_gms.py
```diff
@@ -260,3 +260,23 @@ def test_ingest_explicit_static_token_wins_over_env_auth(
...
+    # env-configured provider on the way. GMS serves /config unauthenticated
+    # (it's in authentication.excludedPaths), so this does NOT prove GMS
+    # accepted the token — it gates that the version-check path itself
+    # (env auth -> load_client_config -> DataHubGraph -> token mint -> /config)
+    # works when auth= replaces a static token, instead of crashing or hanging.
+    upgrade = pytest.importorskip("datahub.upgrade.upgrade")
+    if not hasattr(upgrade, "fetch_server_config_via_graph"):
+        pytest.skip(
+            "installed acryl-datahub predates upgrade.fetch_server_config_via_graph"
+        )
```
skrydal: It is just sad we found ourselves in the place where we have such lines in the tests.

It doesn't appear to me we ever run this when not running straight from repo code, hence this function should be expected to be present. I think we should remove such checks, WDYT?

skrydal (approved):
LGTM

PR #18191, opened 2026-07-06
chore(ingestion): flag remaining static-token-only auth surfaces

max-datahub:
## Summary

**Comments only — no behavior change.** Greppable `TODO(oauth)` markers on every code path that still supports only a static token, so the residual map lives in the code:

- gql GraphQL stack (`api/graphql/base.py`, `api/circuit_breaker/`) + the four Airflow circuit-breaker operators/sensors
- Airflow REST hook + the Airflow 3 listener (the default lineage path — highest-priority future migration)
- GX and Prefect plugin emitters
- the shared doc-connector `DataHubConnectionConfig` and its standalone fallbacks (their primary `ctx.graph` paths are already OAuth-capable)

Draft on purpose: whether these markers land, or the map moves to a tracking issue, is a team call. Split out of #18144 (see the stack there).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## The stack

Split of the original #18144 per review feedback (one concern per PR):

| Order | PR | Base | Notes |
|---|---|---|---|
| 1 | #18187 token-provider hardening | `master` | independent |
| 2 | #18188 configured auth reaches the emitter | #18187 | retarget to master when 1 merges |
| 3 | #18144 env-based OAuth (`DATAHUB_AUTH_TYPE`) | #18188 | the feature; retarget when 2 merges |
| 4 | #18192 OAuth CI canary | #18144 | draft until 3 merges |
| — | #18189 session-bypass fixes + lint | `master` | independent |
| — | #18190 metadata_change_sync + version check | `master` | independent |
| — | #18191 TODO(oauth) markers | `master` | draft, comments only |

PR #18192, opened 2026-07-06
test(smoke): OAuth CI canary — hardened harness, 8-test suite, source-overlay tester

max-datahub:
## Summary

CI canary for the OAuth chain — GMS external-OAuth validation AND the SDK's client-side auth. Its base stack (#18187 → #18188 → #18144) has merged, so this is now based on `master` and ready for review.

- **Hardened Keycloak+quickstart harness**: correct service key, `METADATA_SERVICE_AUTH_ENABLED=true` so tests cannot pass vacuously, bind-mount anchoring; authorization disabled (`AUTH_POLICIES_ENABLED` + `REST_API_AUTHORIZATION_ENABLED`) because the canary tests *authentication* — an OAuth machine principal is a fresh synthetic corpuser and needs operator-granted roles in real deployments (documented in the README).
- **8-test suite**: audience accept/reject, anonymous-401 guard, env-based auth through `load_client_config`, ingest matrix (default sink, credential-less explicit sink env-merge, static-token-beats-env precedence), and a connector-style authenticated read via `pipeline.ctx.graph`.
- **The tester overlays the triggering ref's SDK source onto the released install** (a plain repo pip-install is impossible: read-only mount + gitignored codegen output) — so the canary gates client-side auth changes, not just the released surface.
- Destructive teardown gated on `GITHUB_ACTIONS`; assertions = pytest exit code + a no-skips guard; path-gated on the SDK auth surface + nightly.

Split out of #18144 (see the stack there); previously green end-to-end on the combined branch.

## The stack

| Order | PR | Base | Notes |
|---|---|---|---|
| 1 | #18187 token-provider hardening | `master` | ✅ merged |
| 2 | #18188 configured auth reaches the emitter | `master` | ✅ merged |
| 3 | #18144 env-based OAuth (`DATAHUB_AUTH_TYPE`) | `master` | ✅ merged (the feature) |
| 4 | #18192 OAuth CI canary | `master` | this PR — base stack merged, ready for review |


.github/workflows/oauth-smoke.yml
```diff
@@ -0,0 +1,85 @@
...
+jobs:
+  oauth-smoke:
+    name: OAuth Smoke Test
+    runs-on: ubuntu-latest
+    steps:
+      - name: Free up disk space
+        run: |
+          sudo apt-get remove 'dotnet-*' azure-cli || true
+          sudo rm -rf /usr/local/lib/android/ || true
+          sudo docker image prune -a -f || true
```
skrydal: Seems strange we need that, should we rather start with a different image?
Maybe using wolfie would be more appropriate, since it is the image we use to run actual acryl-datahub CLI?

smoke-test/tests/oauth/run-oauth-ci.sh
```diff
@@ -0,0 +1,100 @@
...
+docker exec "$TESTER" bash -c '
+  set -euo pipefail
+  pip install --quiet "acryl-datahub>=1.6.0.9" pytest requests
+  SITE_DATAHUB=$(python -c "import datahub, os; print(os.path.dirname(datahub.__file__))")
+  tar -C /repo/metadata-ingestion/src/datahub --exclude="./metadata" --exclude="*__pycache__*" -cf - . | tar -C "$SITE_DATAHUB" -xf -
+  python -c "from datahub.ingestion.auth.env import build_auth_config_from_env"  # overlay sanity check
+  pytest /smoke/test_oauth_cli_gms.py -v 2>&1 | tee /tmp/pytest-oauth.out
+  grep -qE "[0-9]+ passed" /tmp/pytest-oauth.out
+  ! grep -qE "[0-9]+ skipped|no tests ran" /tmp/pytest-oauth.out
+'
```
skrydal: This script would be much more readable simply by putting it in a separate `.sh` file.
Also - this silent dependency on `acryl-datahub>=1.6.0.9` will keep breaking when we update something. Can we source that dependency from some singular place? We already define such dependencies for smoke tests.

.github/workflows/oauth-smoke.yml
max-datahub: Replaced with `./.github/actions/free-disk-space` action so it matches the rest of CI

max-datahub: According to Claude:

> On Wolfi — I don't think it applies here: runs-on: selects a GitHub-hosted runner, not a container image, so Wolfi (our CLI's container base) can't be the runner. The disk pressure comes from the images quickstart pulls, which is independent of the runner or how we install the CLI (pip on the runner here; the tester itself is python:slim inside compose). Happy to revisit if you meant swapping the tester image specifically.

smoke-test/tests/oauth/run-oauth-ci.sh
max-datahub: > Both fixed. Pulled the inline block out into run-oauth-tester.sh (mounted at /smoke, invoked with bash /smoke/run-oauth-tester.sh) — much easier to read and it now shellchecks cleanly.

> On the version: dropped the >=1.6.0.9 literal. It's now sourced from gradle/versioning/cliVersion.gradle

.github/workflows/oauth-smoke.yml
skrydal: I understand all arguments Claude made here, my point was:
> We are using runner which ships with things which we delete to make up space (irregardless how the space was consumed). If we use image with less things, deletion would not be needed.

This is a minor point as it seems it is the approach used throughout the repository. Good call to use the script instead.

smoke-test/tests/oauth/run-oauth-tester.sh
```diff
@@ -0,0 +1,36 @@
...
+# so the installed base supplies the generated schema_classes — which must be new
+# enough for the overlaid source's imports. Pinning the base floor (==1.6.0) fails
+# with ImportError on classes added after it (e.g. DomainAssociationClass).
+pip install --quiet "acryl-datahub>=${CLI_VERSION}" pytest requests
+SITE_DATAHUB=$(python -c "import datahub, os; print(os.path.dirname(datahub.__file__))")
+tar -C /repo/metadata-ingestion/src/datahub --exclude="./metadata" --exclude="*__pycache__*" -cf - . | tar -C "$SITE_DATAHUB" -xf -
+python -c "from datahub.ingestion.auth.env import build_auth_config_from_env"  # overlay sanity check
+pytest /smoke/test_oauth_cli_gms.py -v 2>&1 | tee /tmp/pytest-oauth.out
+grep -qE "[0-9]+ passed" /tmp/pytest-oauth.out
+! grep -qE "[0-9]+ skipped|no tests ran" /tmp/pytest-oauth.out
```
skrydal: No need for such grep, check https://docs.pytest.org/en/stable/reference/exit-codes.html

skrydal (approved):
LGTM, I highly suggest adjusting the line I pointed out.

smoke-test/tests/oauth/run-oauth-tester.sh
max-datahub: Dropped both greps.
Reasoning:

> set -e + pipefail already makes pytest's exit code the pass/fail gate, so grep "passed" and the no tests ran check were redundant (exit 5 covers no-collection). The one thing the exit code can't distinguish is an all-skipped run — pytest exits 0 there — so I guard that at the source instead: assert DATAHUB_GMS_URL / KEYCLOAK_TOKEN_ENDPOINT before pytest (the module-skip trigger), and the SDK-auth skip is already caught by the import check above. No output parsing left.

PR #18195, opened 2026-07-06
ci(pre-commit): Add betterleaks secret scanning pre-commit hook

devashish2203:
# Summary

Adds a pre-commit hook that scans staged changes for hardcoded secrets using betterleaks (the actively-maintained gitleaks successor), to prevent accidental credential commits.

# Changes
- New .betterleaks.toml — extends betterleaks' default ruleset and adds a narrow native-Expr filter that allowlists known-safe documentation samples and test fixtures (a public RSA test key, sample service-account values). Any other secret, even in those same files, is still flagged.
- Registers a betterleaks hook in .github/scripts/pre-commit-override.yaml and .pre-commit-config.yaml.

# Key decisions
- language: system (native binary), not docker. A docker_image hook only mounts $PWD, so git --staged silently passes inside git worktrees (the external gitdir isn't visible) — a dangerous false-clean. The native binary resolves worktrees correctly.
- Scans the staged diff only (git --staged, pass_filenames: false) — fast and diff-scoped.
- Assumes betterleaks is available on PATH in dev/CI environments.

# Testing
- betterleaks config check → OK (325 rules loaded).
- Staged a planted key → hook blocks; allowlisted fixtures don't trigger; a new secret alongside them is still caught.

### DIFF
--- a/.betterleaks.toml
+++ b/.betterleaks.toml
@@ -1,0 +1,24 @@
+title = "DataHub betterleaks config"
+
+# Native betterleaks (Expr) config. `filter` is a global expression evaluated per
+# finding; when it returns true the finding is dropped. The patterns below match
+# specific known-safe fixtures, so any OTHER secret — even in the same files — is
+# still reported:
+#   MIIyourkey              -> dataplex docs placeholder private key
+#   MIIEvQIBADAN...HjpPX     -> archive.py documented public RSA test key (mocked replay tests)
+#   suppproject-id-1234567  -> dataplex docs sample GCP service account
+#
+# matchesAny (regex) is used rather than containsAny: containsAny misses multi-line
+# matches (e.g. the multi-line RSA key block in archive.py).
+filter = '''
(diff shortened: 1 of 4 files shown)


david-leifker (approved):
Not a blocker, however I'd spend some time adding `prefilter` to exclude common files in diffs, like lock files which are generated to cut down on the regex checks.

devashish2203:
> Not a blocker, however I'd spend some time adding `prefilter` to exclude common files in diffs, like lock files which are generated to cut down on the regex checks.

Better leaks default already handles lot of these standard files.


> useDefault also brings in betterleaks' default prefilter, which skips lock files (yarn/npm/pnpm/deno/gradle/go.sum/Pipfile/poetry), node_modules, vendored deps, binaries, and minified JS *before* regex runs — so generated files cost no regex work. We deliberately do NOT define a top-level `prefilter` here: a custom one REPLACES the default rather than extending it, and the built-in list is broader than anything we'd maintain by hand.

PR #18199, opened 2026-07-06
feat(ingestion): add S3-backed CLI version matrix source

alokr-dhub:
## Summary

Adds an S3 backend for the per-connector CLI version matrix alongside the existing HTTP backend.

### Parser extraction

`IngestionCliVersionMatrixParser` is extracted from the inline logic that previously lived entirely inside `HttpUrlIngestionCliVersionMatrixSource`. Both the HTTP and the new S3 sources delegate to this shared parser, so validation behavior (fail-closed on bad root, fail-open on bad entries) is consistent across all backends.

### New S3 source

`S3IngestionCliVersionMatrixSource` polls an S3 object on a configurable interval using the existing `S3Util` bean. It stores the last successfully parsed matrix in an `AtomicReference` and serves it on every `getMatrix()` call without re-fetching.

### Configuration

Set `ingestion.cliVersionMatrix.source: s3` and the three S3-specific properties to activate the new backend:

| Environment variable | Property | Default | Notes |
|---|---|---|---|
| `INGESTION_VERSION_MATRIX_S3_BUCKET` | `ingestion.cliVersionMatrix.s3.bucket` | _(empty)_ | S3 bucket name |
| `INGESTION_VERSION_MATRIX_S3_KEY` | `ingestion.cliVersionMatrix.s3.key` | `matrix.json` | Object key |
| `INGESTION_VERSION_MATRIX_S3_REFRESH_SECONDS` | `ingestion.cliVersionMatrix.s3.refreshSeconds` | `600` | Poll interval |

### Backward compatibility

The default `source` value is unchanged (`http`). Deployments that do not set `source: s3` are completely unaffected.

### Graceful degradation

If `source: s3` is configured but no `s3Util` Spring bean is present (AWS not configured), the factory wires a `NoOpIngestionCliVersionMatrixSource` instead of failing GMS startup. The application falls back to its built-in default connector versions.


alokr-dhub:
Closing — branch had wrong base, reopening with correct one.

PR #18202, opened 2026-07-06
feat(ingestion): add S3 and GCS-backed CLI version matrix sources

alokr-dhub:
## Summary

Lets a deployment change per-connector CLI versions without redeploying GMS, by reading the version matrix from wherever it already keeps config. A single URI selects the backend by scheme:

| `ingestion.cliVersionMatrix.uri` | Read with |
|---|---|
| `s3://bucket/matrix.json` | the credentials `datahub.objectStorage` resolves — SigV4-signed, so the bucket stays private and can be shared cross-account |
| `gs://bucket/matrix.json` | Application Default Credentials — Workload Identity on GKE, no static key file |
| `file:///path/matrix.json` | the local filesystem (mounted ConfigMap / Secret) |
| `https://host/path` | HTTP, optionally authenticated with `authToken` |
| _empty_ | nothing — connectors use `defaultCliVersion` |

## Design

One `PollingIngestionCliVersionMatrixSource` owns everything backend-independent: the refresh schedule, the atomic lock-free cache swap, JSON parsing, last-known-good retention, and failure logging. A backend is a `MatrixDocumentReader` and nothing more —

- `HttpMatrixDocumentReader` in `configuration`
- `ObjectStorageMatrixDocumentReader` in `factories`, where the cloud clients and their credential resolution already live, which keeps `configuration` free of AWS/GCP dependencies

Readers classify what they can and attach the verdict to a `MatrixReadException`, so the polling source logs an operator-actionable failure without knowing any backend's exception types. Adding a store means writing a reader, not another source.

Refresh failures log a greppable `MatrixRefreshFailure` token — `[permission]`, `[not-found]`, `[payload]`, `[transport]` — so `grep '\[permission\]'` finds exactly the refreshes that need a credentials or bucket-policy fix, and log-based alerts can route on the token instead of on message wording.

Every unusable configuration degrades to `NoOpIngestionCliVersionMatrixSource`, so connectors fall back to `defaultCliVersion` and GMS startup is never blocked: unsupported scheme, a URI naming no object, an unbuildable storage client, non-positive `refreshSeconds`, or a Spring context that never imports the object-storage factory (mae-consumer reaches this factory via `IngestionSchedulerFactory`).

## Changes to shared object-storage code

Rebased on top of #18351, and extends what it established rather than working around it:

- **`ObjectStorageClient.getObjectAsString(key)`** — a read path for small config documents alongside the existing write/presign surface. Implemented for S3, GCS and local.
- **`ObjectStorageLocation.parseDocument(uri)`** — splits a URI naming one object into the root a client is built from plus the key within it. `parse()` models a root, so without this every caller holding a document URI reinvents the split, and reinvents it differently for cloud and local.
- **`ObjectStorageClientFactory.clientFor(location)`** — the provider routing that `getInstance()` already had, parameterised by location instead of hardcoded to the configured one. `getInstance()` now calls it too, so there is one provider switch in the codebase rather than two, and no caller needs a raw `S3Client` handed back out.

## Configuration migration

**Deployments configured through Helm need no change.** The env var names are unchanged and the chart injects exactly those three, still gated on `global.datahub.managed_ingestion.versionMatrix.enabled`.

| | Before | After |
|---|---|---|
| Location | `cliVersionMatrix.http.url` | `cliVersionMatrix.uri` |
| Refresh | `cliVersionMatrix.http.refreshSeconds` | `cliVersionMatrix.refreshSeconds` |
| Auth | `cliVersionMatrix.http.authToken` | `cliVersionMatrix.authToken` |
| Kill-switch | `cliVersionMatrix.source: none` | leave `uri` empty |
| Env vars | `INGESTION_VERSION_MATRIX_URL` / `_REFRESH_SECONDS` / `_AUTH_TOKEN` | unchanged |

> [!WARNING]
> **`ingestion.cliVersionMatrix.source` (`INGESTION_VERSION_MATRIX_SOURCE`) is removed.** It was a kill-switch that won even when a URL was configured. A deployment that set `source=none` *while leaving a URL in place* will begin fetching the matrix after this change — clear the URL to keep it disabled. Deployments that disable via Helm (`versionMatrix.enabled: false`) never set the URL and are unaffected.

`authToken` now warns when set alongside an `s3://`, `gs://` or `file://` URI. Those authenticate with ambient cloud credentials, so the token was previously dropped in silence and an operator debugging a 403 was looking at the wrong end.

## Testing

Unit tests, no new integration surface:

- `PollingIngestionCliVersionMatrixSourceTest` — refresh/cache contract once for every backend, stubbed at the reader boundary: cache-and-serve, last-known-good retention on read failure, on non-JSON and on schema-invalid documents, failure classification, and that an `Error` from a bad document does not cancel the refresh schedule (`scheduleAtFixedRate` suppresses all future runs if a task lets anything escape)
- `ObjectStorageMatrixDocumentReaderTest` — S3/GCS/local failure classification through the cause chain, mocked at `ObjectStorageClient` so one suite covers all three providers
- `IngestionCliVersionMatrixServiceFactoryTest` — backend selection by scheme, plus every degrade-to-no-op path including a context with no object-storage factory
- `ObjectStorageLocationTest` — `parseDocument` for cloud, local, bucket-only and unsupported schemes
- `IngestionCliVersionMatrixParserTest` — schema validation, unchanged and now shared by all backends

## Checklist

- [x] Tests added/updated

---

> [!NOTE]
> **Overview**
> **Replaces the HTTP-only `source` + nested `http.*` config with a single `ingestion.cliVersionMatrix.uri` whose scheme picks the backend** (`s3://`, `gs://`, `file://`, `http(s)://`), plus flat `refreshSeconds` and `authToken`. `HttpMatrixSourceConfiguration` and `HttpUrlIngestionCliVersionMatrixSource` are removed; matrix JSON parsing moves to shared `IngestionCliVersionMatrixParser`, and all backends poll through `PollingIngestionCliVersionMatrixSource` + `MatrixDocumentReader` (`HttpMatrixDocumentReader`, `ObjectStorageMatrixDocumentReader`).
> 
> **Object storage is extended for small document reads:** `ObjectStorageClient.getObjectAsString`, `ObjectStorageLocation.parseDocument`, and `ObjectStorageClientFactory.clientFor(location)` so the matrix can reuse GMS ambient credentials. Refresh failures log greppable `[permission]` / `[not-found]` / `[payload]` / `[transport]` tokens via `MatrixRefreshFailure` / `MatrixReadException`.
> 
> **Operational contract:** empty or invalid URIs, bad `refreshSeconds`, missing object keys, unsupported schemes, or absent storage clients wire `NoOpIngestionCliVersionMatrixSource` instead of failing startup. **`ingestion.cliVersionMatrix.source: none` is removed** — deployments that relied on `source=none` with a URL still set will start fetching unless the URI is cleared. `authToken` on non-HTTP URIs now warns instead of being silently ignored.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b084ffbe19e5c05cac7bcfa2feec10e7f20dcb70. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


puneetagarwal-datahub:
LGTM . Can we deploy this PR on on any dev instance and test it

puneetagarwal-datahub:
LGTM . Update the code coverage part and test on dev instances for aws and gcp

puneetagarwal-datahub (approved):

david-leifker (changes requested):
Please prefer not hard coding different object storage variables/properties. Use a single set of them with a URI encoding. See the existing object storage configuration. This allows for easier management and easier support for filesystem, http, Azure, and other object storage sources.

alokr-dhub:
@david-leifker Makes sense. Collapsing to a single `cliVersionMatrix.uri` (+ `refreshSeconds`) and dropping the `source:` discriminator — `s3://` / `gs://` / `file://` / `https://` already says which backend.

I'll fold http into `ObjectStorageLocation`/`ObjectStorageClient` rather than special-casing it in the matrix factory, so there's one route instead of two.

alokr-dhub:
@david-leifker Pushed a refactor collapsing this to a single `ingestion.cliVersionMatrix.uri` (+ `refreshSeconds`) — `s3://`, `gs://`, `file://`, and `https://` all work off one property, per-backend vars are gone. Reuses `ObjectStorageClient`/`ObjectStorageLocation` for s3/gs/file (one `ObjectStorageIngestionCliVersionMatrixSource` instead of separate S3/GCS ones), and drops `GcsUtil`/`GcsUtilFactory` in favor of the existing `GcsObjectStorageClient`.

I held off adding `http`/`https` to `ObjectStorageProvider` itself — that enum is write-oriented (`putObject`, presigned URLs, `deleteObject`), and HTTP has no bucket/key, so it'd mostly be `throw unsupported(...)` plus an `authToken` field on shared config that only this caller needs. The matrix factory branches on URL scheme instead — same config either way, upload path untouched.

Happy to fold it into the enum properly if you'd rather.

alokr-dhub:
@david-leifker Please review the PR again. I have made the changes as requested.

david-leifker (approved):
LGTM! Thanks!

PR #18205, opened 2026-07-06
fix(ingestion): Make auto status aspect updater a patch MCP

chriscollins3456:
We have a bug where there's a new auto status aspect populater added by the ingestion team that automatically upserts the status aspect on assets when there is no status emitted for an entity in an ingestion run. This would overwrite semantic anchor statuses when the datahub-documents source runs (which generates semantic search aspects).

The fix: turn the auto status populator into a PATCH write so it doesn't overwrite the lifecycle stage of an entity

### DIFF
--- a/entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/template/common/StatusTemplate.java
+++ b/entity-registry/src/main/java/com/linkedin/metadata/aspect/patch/template/common/StatusTemplate.java
@@ -1,0 +1,51 @@
+package com.linkedin.metadata.aspect.patch.template.common;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.linkedin.common.Status;
+import com.linkedin.data.template.RecordTemplate;
+import com.linkedin.metadata.aspect.patch.template.Template;
+import javax.annotation.Nonnull;
+
+/**
+ * Patch template for the {@link Status} aspect.
+ *
+ * <p>Status is a flat record with no arrays, so transformFields / rebaseFields are identity
+ * transforms. Registering a template enables JSON Patch support (changeType=PATCH) for this aspect,
+ * allowing callers to set individual fields (e.g. {@code /removed}) without overwriting the entire
+ * aspect (which would wipe lifecycleStage, lifecycleState, lifecycleLastUpdated).
+ */
+public class StatusTemplate implements Template<Status> {
+
+  @Override
+  public Status getSubtype(RecordTemplate recordTemplate) throws ClassCastException {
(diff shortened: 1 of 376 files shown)


RyanHolstien (approved):

sgomezvillamor (approved):
Reviewed the `metadata-ingestion` changes. The PATCH approach is the right fix — `op=add /removed=false` preserves `lifecycleStage`/`lifecycleState`/`lifecycleLastUpdated` on existing status aspects, and for entities with no status aspect the `StatusTemplate.getDefault()` still yields `removed=false`, so it's equivalent to the old behavior. Soft-delete resurrection still works too. Tests were updated sensibly for both UPSERT- and PATCH-shaped status workunits. Approving.

A few non-blocking items:

**Question (compat):** Emitting `changeType=PATCH` for `status` only works if the receiving GMS has the new `StatusTemplate` registered (the Java half of this PR). Since the ingestion CLI is upgraded independently of GMS, a newer CLI against an older GMS will send status PATCHes it can't apply. Does this degrade gracefully, or does the run error on those MCPs? If it's the latter, might be worth a min-GMS-version note — this processor runs on basically every ingestion.

**Suggestion (encapsulation):** `patch_builder._add_patch(...)` reaches into a protected method on a bare `MetadataPatchProposal`. Everywhere else in `src/` calls `_add_patch` via `self` from inside a subclass (`ChartPatchBuilder`, the `aspect_helpers/*` mixins, etc.). Could add a small `HasStatusPatch` mixin with a public `set_removed(bool)` mirroring `tags.py`/`domains.py`, so status gets a first-class builder like every other aspect. Not required for the fix.

**Suggestion (workunit id):**
```python
for mcp in patch_builder.build():
    yield MetadataWorkUnit(id=f"{urn}-auto-status", mcp_raw=mcp)
```
`MetadataWorkUnit.from_metadata(mcp)` would auto-generate unique ids and is a touch simpler. Safe today since `build()` returns a single MCP here, but the constant id would collide if it ever returned more than one.

**Suggestion (test depth):** The new tests assert the MCP is a status PATCH but don't decode the `GenericAspect` to confirm the patch is actually `op=add /removed=false`. One small assertion on the payload would lock in the behavior this PR exists to guarantee.

---

sgomezvillamor (changes requested):
blocking to prevent this accidentally merged

since this will require an updated GMS that supports patching on the status aspect, all stateful ingestion events would effectively be lost in practice

chriscollins3456:
closing in favor of this PR which checks gms version: https://github.com/datahub-project/datahub/pull/18239

PR #18206, opened 2026-07-06
feat(context-sidebar): folder tap-target expand + per-section expand/…

annadoesdesign:
https://github.com/user-attachments/assets/f3cd3bd8-3aa6-4769-bc46-15613e2db8b6

Fixes the clunky document browse where trying to expand a folder often navigated into it instead.

- Expand/collapse: the whole left region of a folder row (indent + icon, full row height) is now the toggle target, while the title onward still navigates/opens the document.
- Per-section expand-all / collapse-all controls on the DataHub and each per-platform group header (Phosphor arrows-out/arrows-in), lazily loading children level-by-level.
- Scrollbar uses semantic color tokens and reserves its gutter so rows no longer shift when it appears.

Refactor (behavior-preserving) to keep the tree maintainable and match the codebase's util/hook conventions:
- Extract pure, unit-tested logic into document/utils/documentTreeExpansion.ts.
- Extract hooks: useSectionExpansion, useNodeChildrenLoading, useDocumentNavigation.
- Split TreeSectionHeader and ChildLoadMoreTrigger into their own files, slimming DocumentTree.tsx from 518 to 262 lines.
- Add colocated unit tests for the new util and hooks.

### DIFF
--- a/datahub-web-react/src/app/context/ContextSidebar.tsx
+++ b/datahub-web-react/src/app/context/ContextSidebar.tsx
@@ -181,7 +181,11 @@ const TreeContainer = styled.div`
     flex: 1;
     overflow-y: auto;
     overflow-x: hidden;
-    padding: 8px;
+    /* Trimmed right padding: scrollbar-gutter already reserves space on the right,
+       so the rows/labels can sit closer to the edge without shifting. */
+    padding: 8px 2px 8px 8px;
+    /* Reserve the scrollbar's space so rows don't shift left when it appears. */
+    scrollbar-gutter: stable;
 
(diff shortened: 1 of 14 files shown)


chriscollins3456 (approved):
overall i think this makes sense and the refactoring is good to split things out as opposed to what they were

i think a related thing to expand all here i've noticed is that sometimes there can be lots of documents inside a parent folder but at the bottom of the list are more folders. with infinite scroll it can be very difficult to get to these other folders. wonder if we can sort children inside of folders to have those that are parent folders appear first. not necessary for this PR but just thoughts i had related to similar functionality

PR #18207, opened 2026-07-06
fix(ingestion/dremio): avoid OversizedAllocationException and Dremio-side timeouts on large catalogs - CUS-5532

acrylJonny:
## Summary

Ingesting Dremio catalogs that contain large view definitions (or simply a lot of tables) could silently drop tables/views. Two independent causes, both addressed here:

1. **`OversizedAllocationException` from duplicated view definitions.** The per-column datasets query (`QUERY_DATASETS_*_GLOBAL`) selected `VIEW_DEFINITION` alongside every column, so a view's full SQL was duplicated once per column. A single wide view with a large definition could inflate the Arrow `VARCHAR` vector past Dremio's 2 GiB limit:

   ```
   OversizedAllocationException: Memory required for vector is (2147483648), which is overflow or more than max allowed (2147483647).
   ```

   This surfaced only as a vague "contains no tables or views" warning while silently dropping the remaining tables/views in the affected scope.

2. **Dremio-side query timeouts on large catalogs.** Even without oversized vectors, the single catalog-wide datasets query can time out (or OOM) on the Dremio side for large deployments, again dropping scopes — and the failure was previously swallowed, yielding zero datasets with no hard error.

### Changes

**View definitions fetched separately (fixes the OOM):**

- Remove `VIEW_DEFINITION` from the per-column datasets queries (CE / EE / Cloud). Definitions are now fetched by dedicated `QUERY_VIEW_DEFINITIONS_*` queries — one row per view — and merged back by `FULL_TABLE_PATH`. Since the definition is no longer duplicated per column, the vector can only overflow if a single definition approaches 2 GiB on its own, which does not happen in practice.
- The `QUERY_VIEW_DEFINITIONS_*` queries are paginated (`LIMIT`/`OFFSET`), so they order by the unique `FULL_TABLE_PATH` (EE/Cloud) / `TABLE_SCHEMA, TABLE_NAME` (CE). Ordering EE/Cloud by `TABLE_SCHEMA` alone left views with a tied sort key able to be skipped or duplicated across chunk boundaries.
- View definitions are held in an on-disk `FileBackedDict` rather than a plain in-memory dict. View SQL is the known-large payload (it's what overflows the vector), so buffering all of it for the whole run — including the memory-constrained fallback below — would reintroduce the pressure this fix removes. RAM is now bounded to the LRU cache window.
- A failed view-definition fetch now calls `report.failure(...)` (with a count of how many were retrieved) instead of silently continuing, so operators know derived lineage is incomplete.

**Automatic per-container fallback (bounds query time — no config knob):**

- The single catalog-wide query stays the default (it plans once, so it's the cheapest option). If Dremio can't run it — a planning/exec timeout or OOM on the *first* chunk, before any rows are emitted — the fetch **automatically** retries one root container (source / space / home) at a time, so the join and sort only ever cover a single container. This happens transparently; there's no user-facing setting to reason about.
- Because the fallback only triggers before anything is yielded, it can't double-count. A failure partway through a stream (rare) is now reported loudly instead of silently dropped.
- Each per-container query injects a **pushable** `TABLE_SCHEMA` predicate (`UPPER(TABLE_SCHEMA) = '<root>' OR UPPER(TABLE_SCHEMA) LIKE '<root>.%'`) directly into the `INFORMATION_SCHEMA.COLUMNS` subquery. This is deliberate: Dremio's `InfoSchemaPushFilterIntoScan` rule can push a plain `UPPER(col) = / LIKE` predicate on `TABLE_SCHEMA` down into the info-schema scan, but it does **not** push `REGEXP_LIKE` or `COLUMN_NAME` filters. The outer `REGEXP_LIKE` still bounds the `SYS`/join side and the final sort. Output is identical to the global query — partitioning only changes how the rows are fetched, not which rows.

**Robustness:**

- The chunked dataset/view-def fetches now catch transport errors (`requests` timeouts / connection resets), not just `DremioAPIException`. Those propagate raw from the lazily-streamed result pages and previously escaped the fallback/failure handling.

**Refactor:**

- All SQL — including partial predicates (schema-pattern, per-container scoping) — now lives in `DremioSQLQueries`. `dremio_api.py` only decides edition/column and calls the builders.

## Test plan

- [x] `pytest tests/unit/dremio/` — all passing
- [x] `test_dremio_view_definitions.py`: datasets queries no longer carry the definition; view queries are one-row-per-view (no `COLUMNS` join) with the correct per-edition column; paginated ordering uses a unique key; bracket-path `FULL_TABLE_PATH` normalization is consistent between the datasets and view-def queries; view definitions are backed by `FileBackedDict`; chunked map building + pagination; loud failure reporting; path-based merge
- [x] `test_dremio_partitioning.py`: global success does not fall back; a first-chunk failure falls back to the per-container path; one scoped query per container; no-containers fallback reports a failure; pushdown-shaped predicate lands inside the `COLUMNS` scan (before `ORDER BY`, no intervening join) for all three query templates; community column qualification
- [x] `ruff` + `mypy` clean on changed files

## Checklist

- [x] Tests added/updated
- [x] No breaking changes — no new config; the per-container path is an internal, automatic fallback


max-datahub (commented):
## Memory usage review

The core fix is right: splitting `VIEW_DEFINITION` out of the per-column query removes the real Dremio-side `OversizedAllocationException` (view SQL duplicated per column overflowing the 2 GiB Arrow `VARCHAR` vector). The connector's streaming design is already memory-aware — `get_datasets()` is a generator, `get_workunits_internal` does a single pass, and there's a deliberate choice not to pin full `DremioDataset` objects for large catalogs.

One thing to flag, though: the fix trades a bounded server-side OOM for an unbounded connector-side buffer.

### `view_definitions` buffers the entire catalog's view SQL in RAM

`_get_view_definitions()` chunks the query but accumulates every row into a plain `Dict[str, Optional[str]]` and returns the whole thing. That dict is then held for the entire duration of `get_all_tables_and_columns()` streaming (passed into `_get_all_tables_global_chunked`, looked up per-path in `_build_emit`), and it's also reused as-is by the per-container fallback path.

The payload here is exactly the large object the PR proved can overflow a 2 GiB vector — so the sum of all view SQL in the catalog is now resident in the connector process for the whole run, including in the memory-constrained fallback case where you least want it. Previously each definition streamed inline per chunk and was released when the chunk flushed.

This is the textbook `FileBackedDict` case (large collection of items with large per-item metadata, kept in memory across the run). The swap is nearly drop-in since the get/set interface is identical:

```python
from datahub.utilities.file_backed_collections import FileBackedDict

definitions: FileBackedDict[Optional[str]] = FileBackedDict()
```

Values are pickle-serializable strings and keys are `FULL_TABLE_PATH` strings, so it should be a clean substitution. This bounds connector RAM to ~one definition at a time (the merge lookup) at the cost of one disk read per view — a good trade given view SQL is the known-large object.

Alternative if you want to avoid disk I/O entirely: both the datasets query and the view-def query are `ORDER BY` path, so they could be merge-joined in a single streaming pass instead of buffering either side. That's heavier to write and keep correct across the three editions + community + fallback, so I'd reach for `FileBackedDict` first.

### Minor / pre-existing (not blocking, noting for context)

- `get_all_containers()` materializes the whole source/space/folder tree into a list before yielding, unlike `get_datasets()`. Containers are lighter than datasets so it's lower priority, but for deep hierarchies it's O(all containers) resident. Separate follow-up.
- `profile_targets` / `glossary_terms_collected` grow linearly with the catalog but are already slimmed to tuples / dedup-by-string with a tiny per-item payload — fine as-is.

Happy to draft the `FileBackedDict` change (it touches `_get_view_definitions` plus the two call sites that thread the dict through) if that's useful.

max-datahub (approved):
Approving — re-reviewed at `5c1f69db4`.

The memory concern from my earlier comment is fully resolved. `view_definitions` is now backed by `FileBackedDict` (commit `4b20de70b`), wrapped in `with closing(...)` so the temp DB is cleaned up, and the `with` block spans both the global and the per-container fallback paths — so the dict stays open for the whole generator lifetime and closes on exhaustion. Types thread through as `Mapping`, and the tests were updated (isinstance check + explicit `close()`).

Net effect: the large view-SQL payload is no longer held fully in RAM, so the OOM risk is now closed on both the Dremio side (the original fix) and the connector side. The one trade-off — a SQLite point lookup per dataset that misses the view dict — is indexed and sub-millisecond, which is fine at scale.

The two minor items I noted earlier (`get_all_containers` materializing the container tree, and the slim accumulator lists) are non-blocking and fine to leave for a separate follow-up.

Nice work on the fix.

PR #18210, opened 2026-07-06
fix(ece/structured-property): Refactor change event generation to always set parameters

asikowitz:
Structured property REMOVE entity change events were emitted with parameters: null whenever the removed property's URN sorted before a property still assigned to the entity. The removal branch inside the merge loop called StructuredPropertyAssignmentChangeEvent.builder(), which silently resolves to the inherited ChangeEvent.builder() and never populates parameters — so downstream consumers received REMOVE events with no propertyUrn/propertyValues.

This PR replaces the dual-pointer sorted-merge diff in StructuredPropertyChangeEventGenerator with a map-based set-difference approach (mirroring OwnershipChangeEventGenerator). All ADD/REMOVE/MODIFY events now go through a single builder helper that always sets the value assignment, so parameters can no longer be dropped. REMOVE events carry the removed property's previous values.

  Minor behavior changes:
  - MODIFY event descriptions now say "changed on entity" instead of incorrectly reusing the "added to entity" wording.
  - Inputs are no longer mutated (the old code sorted the aspects' property arrays in place).
  - Event ordering remains deterministic (sorted by property URN).


chriscollins3456 (approved):
nice this def makes sense

PR #18213, opened 2026-07-06
feat(ingest/unity): Lakehouse Federation support (foreign catalogs)

treff7es:
## Summary

Adds **Databricks Lakehouse Federation** support to the Unity Catalog source. A *foreign catalog* is a read-only mirror of an external database (PostgreSQL, SQL Server / Synapse, MySQL, Snowflake, Redshift, BigQuery, Oracle, Teradata, another Databricks workspace, Salesforce Data Cloud, or Glue / Hive metastore). This PR detects foreign catalogs and links each mirrored table to the external source dataset it copies.

For each foreign-catalog table the connector now:

- **Emits upstream COPY lineage** to the external source dataset (`include_federation_lineage`, default on), with identity **column-level lineage** (1:1, case-insensitive) when `include_column_lineage` is enabled.
- **Marks the foreign catalog container with structured properties** (`catalog_type`, `platform`, `connection`, `remote_database`) so federated catalogs are facetable in the UI (`emit_federation_structured_properties`, default on). Definitions are registered with `CREATE` + `If-None-Match: *` so a centrally-managed definition is left untouched.
- **Backfills columns from the external source** via the DataHub graph for the common case where Unity Catalog hasn't lazily synced the foreign table's schema yet (`include_federation_column_backfill`, default on; column structure only — external tags/terms are not copied).

The external dataset URN is built to byte-match the external source's own ingestion, honoring per-connection `platform_instance` / `env` / `convert_urns_to_lowercase` overrides via `federation_connection_details` (keyed by the Unity Catalog connection name). When a connector's remote database can't be determined, the connector emits nothing and reports a counted warning rather than a dangling edge.

The connection-type → DataHub-platform + `CREATE FOREIGN CATALOG OPTIONS` key mapping was validated against the official Databricks per-source documentation (e.g. Oracle uses `service_name`, BigQuery `dataProjectId`, Databricks-to-Databricks `catalog`, Salesforce Data Cloud `dataspace`).

The pre-existing Delta Lake `emit_siblings` path is unchanged.

federation_cll_edge

federation_column_lineage_clean

federation_lineage_clean

## Config

| Option | Default | Purpose |
| --- | --- | --- |
| `include_federation_lineage` | `true` | Emit an upstream COPY lineage edge from each foreign-catalog table to the external source dataset it mirrors. |
| `include_federation_column_backfill` | `true` | Backfill a foreign-catalog table's columns from the external source when Unity Catalog hasn't synced them yet (structure only; governance is not copied). |
| `emit_federation_structured_properties` | `true` | Define + assign the federation structured properties on the foreign catalog container. |
| `federation_structured_property_namespace` | `databricks.federation` | Qualified-name prefix for those structured properties. |
| `federation_connection_details` | `{}` | Per-connection overrides, keyed by the Unity Catalog connection name (see below). |

### `federation_connection_details.<connection_name>`

Overrides so the external URN byte-matches how the external source was ingested. Every field is optional; supply only what differs from the auto-detected value.

| Field | Type | Default | Purpose |
| --- | --- | --- | --- |
| `platform` | str | _auto from connection type_ | Override the DataHub platform (e.g. `mssql`, `postgres`). |
| `platform_instance` | str | `None` | The `platform_instance` the external source was ingested under. |
| `env` | str | source `env` | The environment of the external dataset (e.g. `PROD`). |
| `database` | str | _from foreign catalog OPTIONS_ | Override the remote database name. |
| `convert_urns_to_lowercase` | bool | `true` | Lower-case the external URN's name (DataHub SQL connectors lower-case identifiers). Set `false` for a source ingested case-sensitively. |

```yaml
source:
  type: unity-catalog
  config:
    include_federation_lineage: true
    include_column_lineage: true
    federation_connection_details:
      my_postgres_connection:
        platform_instance: prod-pg
        env: PROD
```

## Scalability

Backfill and column-level lineage need the external source's schema. Rather than one `graph.get_aspect` per foreign table (N round-trips), the connector bulk-loads schemas once per external `(platform, platform_instance, env)` via `SchemaResolverProvider`, scoped to the mirrored database with an exact `id START_WITH "[platform_instance.]database."` filter — so it loads only that database's datasets, not the whole external platform, and resolves each table in memory.

## Testing

Validated end-to-end against a live Databricks workspace with a foreign Snowflake catalog: table-level COPY lineage, case-insensitive column-level lineage, structured-property assignment, and the database-scoped schema fetch all landed correctly in GMS.

## Checklist

- [x] Links to related issues (N/A)
- [x] Tests added/updated — ~50 unit tests + integration golden
- [x] Docs updated — `unity-catalog_post.md`
- [ ] Breaking changes — none (new, opt-out-able feature)


metadata-ingestion/src/datahub/sql_parsing/schema_resolver_provider.py
```diff
@@ -57,22 +64,40 @@ def get(
         platform: str,
         platform_instance: Optional[str],
         env: str,
+        id_starts_with: Optional[str] = None,
     ) -> SchemaResolver:
-        """Return a bulk-initialized SchemaResolver, cached per (platform, platform_instance, env)."""
+        """Return a bulk-initialized SchemaResolver, cached per (platform, platform_instance, env, id_starts_with).
+
+        `id_starts_with` narrows the bulk fetch to datasets whose id (fully-qualified
```
sgomezvillamor: Is id here the dataset URN?

I'm a bit wary of the "id/URN starts with" filtering — it feels hacky and fragile (no clear boundary on what counts as a match). What are we actually trying to filter on? If the intent is "everything under a given database/container," could this be modeled as a parent_container filter (or similar) instead of a string-prefix check on the URN?

metadata-ingestion/src/datahub/ingestion/source/unity/federation.py
```diff
@@ -0,0 +1,243 @@
...
+    for prop in FEDERATION_PROPERTIES:
+        allowed_values = (
+            [
+                PropertyValueClass(value=platform)
+                for platform in KNOWN_FEDERATION_PLATFORMS
+            ]
+            if prop.platform_allowed_values
+            else None
+        )
+        aspect = StructuredPropertyDefinition(
```
sgomezvillamor, line 197: I'm personally a bit concerned about sources emitting property definitions — I'd expect definitions to be the responsibility of the DataHub admin. We probably lack a clear team criteria here.

In this case it can be disabled and the namespace is controlled, so it's not too bad.

My preference would be to emit these as custom properties and use transformers to map them to glossary terms, tags, or structured properties. The reasoning: custom properties are an open contract, while glossary terms and structured properties are governed assets with tighter controls.

sgomezvillamor:
Review mostly focused on the schema resolver, since it's foundational stuff that needs to be well defined.

🔴 Blocker — id START_WITH feels fragile.
The bulk-fetch scoping in schema_resolver_provider.py filters on id START_WITH "<db>.". It does work today — id is DatasetKey.name, which for SQL datasets is [platform_instance.]db.schema.table — but it's a string-prefix match on the URN name with no clear boundary on what counts as a match, and there's no other field="id" START_WITH usage anywhere in ingestion. It hard-couples to the naming convention. If the intent is "everything under a given database/container," can we model this as a parent_container (or similar) filter instead of a prefix check on the URN?

🟡 Warning — sources emitting structured-property definitions.
I'd be ok with sources emitting structured-property assignments; my concern is specifically on them emitting definitions — I'd expect definitions to be the DataHub admin's responsibility. Just a warning because we lack overall criteria / team agreement here. If this hasn't been specifically requested, it could be removed or emitted as a custom property only. Not a blocker (it's disable-able and the namespace is controlled) if it goes as it is.

Also sharing additional review notes from Claude:

🟡 Warning — federation code not individually guarded.
The table-level federation calls (_gen_federation_link, _resolve_external_schema_fields) are only caught by process_schemas' coarse handler, which continues the whole schema — so a federation exception drops all remaining real tables in that schema and mislabels it "Missed schema because of parsing issues." The catalog-level federation work in gen_catalog_containers has no guard at all. For a supplemental feature, these should degrade (count + warn) rather than affect core metadata.

🔵 Suggestion — case-folding decision duplicated (DRY/invariant).
_external_platform_key sets key.lowercase_urns and _external_dataset_urn independently recomputes detail is None or detail.convert_urns_to_lowercase, kept aligned only by a comment. Since _external_dataset_urn already holds key, use key.lowercase_urns and make _ExternalSchemaKey the single source of truth.

🔵 Suggestion — duplicate upstreamLineage aspect can clobber.
For a foreign-catalog table, process_table emits the lineage aspect and _gen_federation_link then emits a second full upstreamLineage aspect for the same URN (no patch) → last-write-wins. Low likelihood (foreign tables rarely have UC upstreams), but folding the federation COPY upstream into the single lineage aspect would avoid it.

sgomezvillamor (approved):

PR #18215, opened 2026-07-06
fix(ingestion/patch): revert: emit fieldPath add op for editableSchemaMetadata field patches (#18138)

asikowitz:
This undoes a change that would add another patch (write) call for every schema field patched with `add_tag` or `add_term`. We should instead try to handle this on the backend.


asikowitz:
Cancelled in favor of https://github.com/datahub-project/datahub/pull/18334

PR #18216, opened 2026-07-06
fix(ingest/dbt): isolate per-model failures and identify failing model in errors

kyungsoo-datahub:
## Summary

  dbt ingestion previously aborted the **entire run** when a single model hit an unhandled exception during processing (e.g. a sqlglot parsing crash on one model's SQL), and the error didn't say **which** model or file caused it — forcing operators to manually bisect across model folders.

  This wraps the per-item boundaries of dbt processing in try/except so a failing item is isolated while the rest of the run completes, and every isolated failure is reported as a structured warning containing the item's name/id **and** its file path, plus a counter, so operators can see what failed and where without digging through logs.

  ### What's isolated, and what happens to the failing item

  - **Manifest extraction** (`extract_dbt_entities`): a node whose manifest entry can't be turned into a `DBTNode` (e.g. missing a required field) is **dropped** — this is the "node itself unbuildable" case, so there's nothing left to degrade to. Counted in `node_extraction_failures`.
  - **Schema / column-level-lineage inference** (`_infer_schemas_and_update_cll`): a node that fails here **degrades**, keeping whatever dataset/schema/table-level lineage it already had — it is not dropped, and its previously-known columns are preserved (see `test_infer_schemas_and_update_cll_failure_preserves_existing_columns`). Downstream nodes are unaffected by an upstream's CLL failure (see `test_infer_schemas_and_update_cll_downstream_node_still_processes_after_upstream_failure`). Counted in `node_cll_failures`.
  - **Metadata/test-assertion/freshness-assertion/exposure emission** (`create_dbt_platform_mces`, `create_target_platform_mces`, `create_test_entity_mcps`, `create_freshness_assertion_mcps`, `create_exposure_mcps`): an item whose emission crashes partway through **keeps whatever workunits it already yielded** before the crash point (no buffer-and-rollback — see `test_create_dbt_platform_mces_partial_emission_before_failure`) and skips the rest. Counted in `node_emission_failures` (once per emission loop it failed in, so the same node can be counted more than once if multiple emission phases fail independently for it).

  All five of these are on the same `get_workunits_internal` path — a bad item in any one of them previously aborted the whole run; now each is isolated the same way.

  Note: `sqlglot_lineage()` itself already catches `Exception` internally and separately special-cases the native Rust-tokenizer `PanicException` (the sqlglot crash that originally motivated this ticket), so it needs no additional wrapper — any other failure in the surrounding CLL code (CTE mapping, post-processing, schema resolver, etc.) is caught by the per-node try/except in `_infer_schemas_and_update_cll`.

  ### Deliberate tradeoffs (not oversights)

  - **No abort/failure-rate threshold.** The run alven if every node fails (see`test_infer_schemas_and_update_cll_all_nodes_failidered escalating to `report.failure(...)` pastsome failure-rate threshold, but that flips the p(verified in `pipeline.py`'s`pretty_print_summary`), which conflicts with "altly decided against adding any such threshold/knobhere.- **No buffer-and-rollback on emission.** If an ime of its workunits were already yielded, thoseworkunits stay (partial/degraded emission) ratherm's output in memory before yielding.
  ### Known limitation: stale-entity removal                                                               
  With stateful ingestion enabled, a model that fails extraction or emission this run is not just skipped —this run's checkpoint, so DataHub's stale-entity existing metadata (docs, ownership, tags,lineage) unless the run-over-run entity change happens to exceed `fail_safe_threshold` (default 75%, whicnear-total collapse, not a handful of consistentlThis is an accepted, explicit tradeoff for this Pl in question hasn't actually been removed fromdbt, it just failed to process — a genuine gap, bes plumbing (e.g. exposing`StaleEntityRemovalHandler.add_urn_to_skip` to soreachable from a source's own code, since thehandler is owned by a generic workunit processor ()` runs) that's out of scope here and affectsevery stateful-ingestion source, not just dbt.


sgomezvillamor (approved):
Nice fix — mechanism is sound, failures are reported with the exception attached (not swallowed), and the tests cover all five wrapped boundaries with the properties that matter (isolation, degrade-vs-drop, counters, name+file_path). Approving. Two things worth addressing, one before merge:

**WARNING — isolation is incomplete: `create_freshness_assertion_mcps` and `create_exposure_mcps` are unwrapped (and untested)**

Both loops are on the same `get_workunits_internal` path (called at `dbt_common.py:1949` and `:1964`) but have no per-item try/except:
- `create_freshness_assertion_mcps` (`dbt_common.py:1650`) — a source node with malformed `freshness_info` raises in `get_urn`/`make_assertion_from_freshness`/`make_assertion_result_from_freshness`.
- `create_exposure_mcps` (`dbt_common.py:1754`) — one bad exposure raises in `get_urn`/`DashboardInfoClass`/`make_user_urn`.

A single bad item there aborts the whole run and exits non-zero — the exact failure this PR eliminates elsewhere. Either wrap them with the same `record_node_failure(kind="emission")` pattern (+ two tests), or explicitly scope them out so the "always exits 0" claim holds. For exposures, `DBTExposure` isn't a `DBTNode`, so call `self.report.record_node_failure(f"{exposure.unique_id} ({exposure.dbt_file_path})", e, ...)` directly like the `dbt_core.py` call site.

**SUGGESTION — semantic-view CLL handler doesn't attach the exception**

The pre-existing semantic-view handler reports with `report.warning(..., message=f"...{str(e)}")` + a separate `logger.debug(exc_info=True)`, but doesn't pass `exc=e`. Passing it keeps the traceback in the structured report, consistent with the convention this PR establishes.

---

PR #18218, opened 2026-07-07
fix(search): stabilize noisy reindex ETA estimate

anshbansal:
## Summary

- **Compute reindex ETA from a cumulative average, not the last delta** — `pollReindexCompletion`'s "Estimated time remaining" log previously derived the indexing rate from docs indexed since the *last observed count change*. Elasticsearch bulk-indexing is bursty and the poll interval itself grows over time, so this rate swung wildly between consecutive polls (e.g. 500 → 2800 minutes on the same steadily-progressing reindex). It now uses the average rate since the polling loop started, which is far more stable while still trending toward the correct estimate.
- **Extract `estimateMinutesRemaining` as a standalone method** — the rate/ETA arithmetic is now a small, pure `public static` helper on `ESIndexBuilder`, independently unit tested rather than buried inline in the polling loop.

Stall detection and reindex-retry-resubmission behavior are unchanged — this only affects the logged ETA on non-terminal poll iterations.

## Test plan

- [x] `ESIndexBuilderTest` — added `testEstimateMinutesRemaining` (steady rate, zero elapsed time, zero progress, already complete, sub-minute truncation); full suite passes (56 tests)
- [x] `IndexBuilderElasticSearchTest` — existing integration test exercising `pollReindexCompletion` end-to-end against real Elasticsearch still passes (42 tests, 2 pre-existing unrelated skips)
- [x] `BuildIndicesIncrementalStepTest` — existing mocks of `pollReindexCompletion` still pass unchanged (signature untouched)

### DIFF
--- a/metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/indexbuilder/ESIndexBuilder.java
+++ b/metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/indexbuilder/ESIndexBuilder.java
@@ -933,7 +933,8 @@ public class ESIndexBuilder {
     Pair<Long, Long> documentCounts =
         getDocumentCounts(opContext, expectedCountSupplier, destIndex);
     long documentCountsLastUpdated = System.currentTimeMillis();
-    long previousDocCount = documentCounts.getSecond();
+    final long pollStartTimeMillis = documentCountsLastUpdated;
+    final long pollStartDocCount = documentCounts.getSecond();
     long estimatedMinutesRemaining = 0;
 
     while (System.currentTimeMillis() < timeoutAt) {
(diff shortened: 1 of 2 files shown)


metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/indexbuilder/ESIndexBuilder.java
```diff
@@ -1083,6 +1081,20 @@ public long computeTimeoutAt() {
...
+   * cumulative average smooths out bursty ES bulk-indexing throughput and the polling loop's own
+   * growing sleep interval, both of which make a single-sample rate estimate very noisy.
+   */
+  public static long estimateMinutesRemaining(
+      long docsIndexedSinceStart, long elapsedMillisSinceStart, long remainingDocs) {
+    double indexingRate =
+        elapsedMillisSinceStart > 0 ? (double) docsIndexedSinceStart / elapsedMillisSinceStart : 0;
+    long estimatedMillisRemaining = indexingRate > 0 ? (long) (remainingDocs / indexingRate) : 0;
+    return estimatedMillisRemaining / (1000 * 60);
+  }
```
gdatahub, line 1096: It will good to calculate and log both cumulative average and last delta. 
A cumulative average reacts slowly to sustained changes in indexing speed. 

For example -- if indexing permanently slows down halfway through, the ETA may remain overly optimistic for a while because it is still influenced by the faster rate from the beginning.

gdatahub (approved):

metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/indexbuilder/ESIndexBuilder.java
anshbansal: The main time when we have to see this is when we upgrade a lot of deployments and something is taking too long. If something is slowing down I have found the slow average to be very unreliable. It has not really been helpful in such cases. 

I think what will happen is that the number will stop going down.

PR #18219, opened 2026-07-07
fix(ingestion/unity): normalize S3 external-lineage paths with partition brace-lists

treff7es:
## Summary

Unity Catalog ingestion fails when Databricks external lineage returns object-storage paths that carry a trailing partition-set component in brace-list syntax, e.g.:

```
s3://mybucket/topics/my_events/{20260410,20260411,20260412,20260413}
```

The braces and — critically — the commas are illegal in a DataHub dataset URN name segment (the comma is the URN field delimiter). The emitted upstream URN is therefore rejected by GMS:

```
Cause: ERROR :: /upstreams/0/dataset :: "Provided urn urn:li:dataset:(urn:li:dataPlatform:s3,.../{20260410,20260411,...},PROD)"
```

## Fix

Normalize such S3 lineage paths down to the parent table path before building the URN: drop every path component from the first one containing `{`, `}`, or `,` onward. Paths without those characters are unchanged (byte-identical output). A new `num_external_upstreams_partition_stripped` report counter surfaces how many paths were normalized.

## Testing

- New unit tests for `_strip_s3_partition_from_path` (brace-lists, trailing slash, no-partition passthrough, `s3a://`) and an end-to-end test asserting `_generate_lineage_aspect` yields a parseable URN with no braces for a brace-list path.
- Full Unity unit + integration suites pass.

## Checklist

- [x] Tests added


metadata-ingestion/src/datahub/ingestion/source/unity/source.py
```diff
@@ -175,6 +175,36 @@
...
+#   s3://bucket/topics/event/{20260410,20260411,20260412}
+# The braces and (crucially) the commas are illegal in a DataHub dataset URN
+# name segment — the comma is the URN field delimiter — so such a path yields a
+# URN that GMS rejects. These characters only ever appear in the partition
+# component, so we drop every path component from the first offending one
+# onward, pointing lineage at the parent table path.
+_S3_PARTITION_ILLEGAL_CHARS = re.compile(r"[{},]")
+
+
+def _strip_s3_partition_from_path(path: str) -> str:
```
sgomezvillamor, line 190: Fine/good enough to prevent GMS rejections, but we still need a proper solution for external data lake lineage — resolving S3/GCS/ABS paths to existing DataHub datasets. Out of scope for this PR, but leaving a comment about the bigger problem.

sgomezvillamor (approved):

PR #18220, opened 2026-07-07
fix(ingestion/unity): add include_ml_models toggle, enforce ml_model_max_results cap, fix ML model env

treff7es:
## Summary

Fixes three ML-model ingestion issues in the Unity Catalog source that caused ML assets to be ingested when operators wanted them off, and landed them in the wrong environment.

**1. No way to disable ML model ingestion.** There was no toggle; `ml_model_max_results: 0` did not help (see below). Adds `include_ml_models` (default `true`); when `false`, ML model discovery is skipped entirely — **no calls to the Databricks registered-models API**.

**2. `ml_model_max_results` did not actually cap ingestion.** The value was forwarded to `registered_models.list()` only as a per-page size hint, and the SDK auto-paginates, so every model was returned regardless of the setting and `0` disabled nothing. It is now enforced as a genuine per-schema cap via `itertools.islice` (`0` ingests none; the generator isn't even started, so no API call is made at `0`).

**3. `MLModel` environment was hardcoded to `PROD`.** `MLModel` entities were constructed without `env`, so the SDK default (`PROD`) was used, while their `MLModelGroup` correctly used the configured `env`. This produced mismatched URNs:

```
urn:li:mlModelGroup:(urn:li:dataPlatform:databricks,<name>,DEV)   ← correct
urn:li:mlModel:(urn:li:dataPlatform:databricks,<name>_1,PROD)     ← bug
```

`MLModel` URNs now honor `config.env`.

## Behavior change

`ml_model_max_results` (default 1000) is now a real cap. A schema with **more than** that many ML models previously got all of them (via pagination); it now stops at the cap. This aligns behavior with the field's documented meaning. Documented in `docs/how/updating-datahub.md`. Use `include_ml_models: false` to opt out of ML models entirely.

## Testing

- New unit tests: `include_ml_models: false` makes no API call and yields nothing; `ml_model_max_results: 0` ingests nothing; the cap limits the total; `MLModel` URNs use the configured `env`.
- ML-model golden integration test unchanged (its recipe uses the default `PROD` env).
- Full Unity unit + integration suites pass.

## Checklist

- [x] Tests added
- [x] Breaking/behavior change documented in `docs/how/updating-datahub.md`


docs/how/updating-datahub.md
```diff
@@ -141,6 +141,13 @@ Requirements:
 
 - **(GMS / search indexing)** Inline Elasticsearch indexing and entity-graph cache invalidation on ingest now require explicit `systemMetadata.properties.appSource = ui` (or the sync-index header), not merely a GraphQL request context. `GroupService` and `RoleService` stamp UI source via `AspectUtils.buildSynchronousMetadataChangeProposal`; other GraphQL resolvers that bypass `MutationUtils` revert to async MAE indexing. See [GMS Entity Graph Cache — Invalidation (sync writes)](../deploy/gms-entity-graph-cache.md#invalidation-sync-writes).
 
+- **(Ingestion / Unity Catalog)** ML model ingestion controls were fixed. Two things changed:
```
```suggestion
- #18220 **(Ingestion / Unity Catalog)** ML model ingestion controls were fixed. Two things changed:
```

sgomezvillamor (approved):

PR #18226, opened 2026-07-07
feat(home): add View All to asset collection modules with dynamic filters

max-datahub:
## Summary

Asset Collection home-page modules had no "View all" affordance, unlike the other large modules (Your Assets, Domains, Subscriptions, …). For teams that use a collection tile to spotlight a specific slice of the catalog — e.g. a "Data Products" tile built with the dynamic filter `Type equals Data Product` — there was no one-click path from the tile to the search page with that same slice applied. This PR adds one: dynamic-filter collections now show **View all**, which navigates to `/search` with the collection's filter reproduced as native search filters (including the standard "Type" chip).

This also gives a general answer to "users don't know how to filter for X in search": admins can curate a collection tile for X on the home page, and View all teaches the filter by landing users on a pre-filtered search page.

## Design decisions and trade-offs

**1. Correctness-first gating, derived from the tile's own query pipeline.**
The core invariant is that clicking View all must never show *different* results than the tile. The View-all params are produced from the exact same `convertLogicalPredicateToOrFilters` + `excludeEmptyAndFilters` output that drives the collection tile's query, then mapped onto the search URL — which can express exactly two shapes: a single AND group, or an OR of single-filter groups. NOT conditions round-trip as negated URL filters, and nested groups that flatten to a representable shape are supported. Anything else hides the button rather than approximating. Trade-off: a small set of collections gets no View all; in exchange, the button is trustworthy wherever it appears, and the tile and the button cannot drift apart because they share one conversion path.

Concretely, the button is hidden for:
- **Nesting with no flat representation** (an OR of multi-condition groups) — the URL's single-level structure cannot express it.
- **`exists` and valueless conditions** — these don't survive the URL serializer, while the tile's query executes them; dropping them would make search show *more* than the tile.
- **Multiple entity-type conditions under AND** — the search page's `generateOrFilters` fans entity-type values out into OR'd clauses, so `AND [Type=A, Type=B]` (zero results on the tile) would become `A OR B` on search. A single Type condition with multiple values, and duplicates under OR, remain allowed (equivalent semantics).
- **Negated or non-equality entity-type conditions** — the URL rebuild drops condition/negation flags for the entity-type portion.
- **Unknown operators** — anything the shared conversion can't translate.

When an author builds a filter that falls in the hidden set, the collection builder shows an inline info hint so the behavior is discoverable at creation time — distinguishing conditions that are merely *incomplete* (missing values; finishing them may enable the button) from filter types that aren't supported. The manual selection tab carries a small note that hand-picked collections don't support View all at this time. Empty drafts and blank rows stay silent.

**2. Entity-type conditions map to the native Type filter.**
Builder predicates store entity types as graph names (`dataProduct`); the search page expects `EntityType` enum values on the combined `_entityType␞typeNames` field. The converter maps values via `entityRegistry.getTypeFromGraphName` and targets the combined field, so the search page renders its standard "Type equals Data Products" chip rather than a raw facet. Unmappable values hide the button instead of producing a filter that matches nothing.

**3. Manual (hand-picked urn list) collections are excluded for now.**
A urn-list collection *could* navigate to search with a `urn` filter, but the search page can't preserve the tile's hand-curated ordering, renders a raw urn chip, and long lists risk URL-length limits. Rather than ship a degraded experience, v1 scopes View all to dynamic-filter collections; manual-mode support can follow if there's demand.

**4. Pure converter + thin hook, no new UI surface.**
`convertLogicalPredicateToViewAllParams` is a pure function (predicate in, `{filters, unionType}` or `null` out) so every gating rule above is unit-testable without rendering. `useAssetCollectionViewAll` wraps it with JSON parsing and navigation and returns `undefined` in every degraded case — which is already `LargeModule`'s contract for hiding the button. Button rendering, styling, and the existing `HomePageTemplateModuleViewAllClick` analytics event are all reused; the module change is one prop.

**5. No backend changes.**
Everything is derivable client-side from the module's existing `assetCollectionParams`. The only shared-code touch is exporting the query builder's existing `mapOperator` instead of duplicating its operator table.

## Example
The following screenshot shows collections that do and do not render a "View All" button based on the underlying search filter complexity:
image

UI hints make it clear for admins / users whenever a collection will not render a "View All" button:
image
image
image

## Test plan

- [x] 24 unit tests covering the converter (flat AND/OR conversion, NOT-to-negated round-trip, collapsible vs unrepresentable nesting, entity-type value mapping and its equality-only/single-under-AND guards, exists/valueless/unknown-operator/unmappable-type hide rules, blank rows), the builder hint classifier (`getViewAllSupport`: supported/incomplete/unsupported), and the hook (navigation args, manual-urn/missing/malformed/unconvertible params)
- [x] Playwright e2e (`e2e-test/ui/playwright/tests/home/homepage-collection-view-all.spec.ts`): button shown on a dynamic-filter collection and hidden on a manual one; clicking navigates to `/search` with the native Type filter applied and parsed back into the active filter chip. The suite provisions its own native user so it never touches the shared admin user's home template asserted on by the existing home specs.
- [x] End-to-end verification on a quickstart built from this branch (fresh volumes, matching frontend/backend), with a 10-tile matrix covering every gating rule:

  **Button visibility — 10/10 as designed (re-verified after the shared-pipeline refactor):**

  | Collection filter shape | View all |
  | --- | --- |
  | Flat `Type = Data Product` | ✅ shown |
  | AND: Type + tag | ✅ shown |
  | OR: two entity types | ✅ shown |
  | Nested group that flattens to a representable shape | ✅ shown |
  | NOT | ✅ shown |
  | OR of multi-condition groups, e.g. `(Type=DP AND tag) OR (owner=X)` — the general unsupported nesting case | hidden |
  | Manual urn list | hidden |
  | `exists` | hidden |
  | Valueless condition mixed with a valid one | hidden |
  | AND with two Type conditions | hidden |

  **Result parity — clicking View all shows exactly the tile's contents:**
  - Flat Type filter → `/search` with the native "Type equals Data Products" chip; 3 results = the tile's 3 data products
  - AND Type + tag → both filters in the URL; 1 result = the single tagged data product on the tile
  - OR two types → `unionType=OR` with both Type filters; 4 results (3 data products + 1 domain) = the tile's contents
  - Collapsed nested group → native Tags chip; 1 result = the tile's contents
  - NOT → negated filter round-trips through the URL (`…___true___EQUAL…`), the search page renders its native "does not equal" chip, and results exclude exactly the tile's excluded entity

  The two zero-result divergence cases confirm the gating: a valueless condition and an AND of two entity types both render an **empty tile** (match-nothing semantics), which is precisely why the button hides rather than sending users to a non-empty search page.

## Checklist

- [ ] Docs: n/a (UI affordance)
- [ ] Breaking changes: none


max-datahub:
Credit where it's due: this feature was proposed by @haeniya, who also shared a working reference implementation of the View all navigation for collection modules. That patch shaped this PR's approach — in particular navigating via `navigateToSearchUrl` with the collection's filters and targeting the combined Type filter field so the search page renders its native chip. This implementation extends it with round-trip safety guards (the button hides whenever the search page couldn't reproduce the tile's exact results — nested/NOT/exists/valueless predicates, and the AND-with-multiple-Type-conditions fan-out case) plus unit test coverage. Thanks for the contribution and the clear problem statement! 🙏

haeniya:
Big thanks to @DiogoVala for driving this feature. 🚀

max-datahub:
@haeniya and @DiogoVala, please review the PR description, functionalities, and limitations and confirm whether this aligns with your expectations :)

chriscollins3456 (commented):
i think overall this feature makes sense and the implementation is looking solid! thanks for this

datahub-web-react/src/app/homeV3/modules/assetCollection/useAssetCollectionViewAll.ts
```diff
@@ -0,0 +1,136 @@
...
+    return { field: leaf.property, values: leaf.values, condition };
+}
+
+/**
+ * Converts a collection's dynamic filter predicate into search-page URL params.
+ * Returns null when the predicate is not exactly representable in the URL
+ * (nested groups, NOT, exists, unknown operators, unmappable entity types) —
+ * callers should hide the "View all" affordance rather than show wrong results.
+ */
+export function convertLogicalPredicateToViewAllParams(
```
chriscollins3456, line 59: i think a lot of the work in this file is duplicated from AssetCollectionModule where we already import different utils or have some logic defined to convert the module params into search filters like `convertLogicalPredicateToOrFilters`

and we obviously can convert any of these into search requests. what does it look like when you have urn filters or more complex and nested filters converted into search filters and send them to the search page? is that just not possible?

I would try to reuse anything you can from AssetCollectionModule.tsx and pull stuff out from there to reuse if need be. i'd be curious if it's possible to have this View All button on any asset collection as opposed to just the simple ones.

it might get crazy with complex ones but maybe that's alright

max-datahub:
Response to the review feedback (AI-generated, quoted as such):

> Thanks for the close read — you were right about the duplication, and it's restructured now. The hook consumes the exact same `convertLogicalPredicateToOrFilters` + `excludeEmptyAndFilters` output that the module's own query executes, so the button's target and the tile's contents can't drift apart (the parallel predicate walk and the `mapOperator` export are gone). That also *expanded* coverage: NOT conditions round-trip as negated search filters, and nested groups that flatten to a representable shape get the button too.
>
> On "can we send anything to the search page": today, no — the search page's only input is the URL, and the URL filter format is structurally flat (one filter list + a single AND/OR). `orFilters` only exists internally; most DNF shapes have no URL serialization. After the refactor, what remains hidden is just: nesting that doesn't flatten, `exists`/valueless conditions (the URL drops them while the tile executes them — navigating would show *more* than the tile), and a couple of entity-type edge cases where the URL rebuild loses flags. To make that discoverable, the collection builder now shows an inline hint when a filter falls in that set. Worth noting this is fully entity-type-agnostic — any collection filter (tags-only, platform, any of the ~15 entity types) gets the button, not just the simple Type cases.
>
> If we want the button on *every* collection — agreed that would be much easier to explain — I think the right mechanism is a small search-page extension: `/search?viewAllModule=<urn>`, where the search page loads the module and applies its orFilters through this same shared converter, showing a single "From collection" chip. Exact parity for arbitrary shapes, manual urn lists included. Happy to spec that as a follow-up if you're open to it — didn't want to grow this PR into the search page.

haeniya:
Many thanks for this @max-datahub. This is exactly what we need and it addresses past user feedback we got. We would highly appreciate if we can get this feature into the July release. Thanks!

chriscollins3456 (approved):

PR #18231, opened 2026-07-07
fix(ingest/kafka-connect): strip platform_instance prefix from schema resolver URNs

treff7es:
## Summary

DataHub encodes a dataset's `platform_instance` as a leading segment **inside** the URN name (`{platform_instance}.{db}.{schema}.{table}`), not as a separate URN field. `_extract_table_name_from_urn` returned the raw name, so for sources ingested **with a platform_instance** the recovered "table name" kept the instance prefix.

That one leak broke the Kafka Connect Debezium schema-resolver path in three ways, all funneling through that single method:

1. **Table discovery returned nothing** — `_discover_tables_from_database` filters on `{database}.` but the name starts with `{platform_instance}.`, so no tables were discovered and no lineage was emitted (the reported symptom).
2. **Doubled `instance.instance.` prefix** in resolver-derived lineage URNs, since the emitter re-applies the platform_instance to `source_dataset`.
3. **`table.include.list` regex matching failed** for the same reason.

Because there is no delimiter marking where the instance ends in the flattened name, the prefix cannot be recovered from the URN string alone — the authoritative value is the instance the `SchemaResolver` was constructed with. The fix strips exactly that prefix (case-insensitively, to account for URN lowercasing) at the single parse point, read defensively via `getattr` because `platform_instance` is not part of `SchemaResolverInterface`.

## Testing

- New unit tests: prefix stripping, the no-instance case (guards against stripping a legitimate leading segment), and table discovery through a resolver carrying a platform_instance.
- Full Kafka Connect unit suite passes (531 tests).
- Verified end-to-end against a live DataHub: with the fix reverted, discovery returns `[]` on instance-prefixed Postgres datasets; with the fix it returns the tables and derives the correct topics.

## Checklist

- [x] Links to related issues (internal)
- [x] Tests added/updated
- [ ] Docs added/updated — n/a (bug fix, no behavior change for correctly-configured setups)

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/kafka_connect/common.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/kafka_connect/common.py
@@ -1155,10 +1155,23 @@ class BaseConnector:
             Extracted table name (e.g., "database.schema.table") or None if parsing fails
         """
         try:
-            return DatasetUrn.from_string(urn).name
+            name = DatasetUrn.from_string(urn).name
         except Exception as e:
             logger.debug(f"Failed to extract table name from URN {urn}: {e}")
             return None
+
+        # DataHub bakes the platform_instance into the URN name as a leading
+        # `{platform_instance}.` segment. Strip it so callers see the logical
+        # db.schema.table name; otherwise table discovery and pattern matching
+        # break for sources ingested with a platform_instance.
+        # platform_instance is not part of SchemaResolverInterface, so read it
+        # defensively for interface-only implementations that omit it.
+        platform_instance = getattr(self.schema_resolver, "platform_instance", None)
+        if platform_instance:
+            prefix = f"{platform_instance}."
+            if name.lower().startswith(prefix.lower()):
+                name = name[len(prefix) :]
+        return name
 
     def _extract_lineages_from_schema_resolver(
         self,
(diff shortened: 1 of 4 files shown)


metadata-ingestion/src/datahub/ingestion/source/kafka_connect/common.py
```diff
@@ -1222,11 +1155,24 @@ def _extract_table_name_from_urn(self, urn: str) -> Optional[str]:
...
             logger.debug(f"Failed to extract table name from URN {urn}: {e}")
             return None
 
+        # DataHub bakes the platform_instance into the URN name as a leading
+        # `{platform_instance}.` segment. Strip it so callers see the logical
+        # db.schema.table name; otherwise table discovery and pattern matching
+        # break for sources ingested with a platform_instance.
+        # platform_instance is not part of SchemaResolverInterface, so read it
+        # defensively for interface-only implementations that omit it.
+        platform_instance = getattr(self.schema_resolver, "platform_instance", None)
```
sgomezvillamor: we have `platform` in the `SchemaResolverInterface`

```
class SchemaResolverInterface(Protocol):
    @property
    def platform(self) -> str: ...

```

we could also add `platform_instance`

```
    @property
    def platform_instance(self) -> Optional[str]: ...
```

sgomezvillamor (approved):

metadata-ingestion/src/datahub/ingestion/source/kafka_connect/common.py
treff7es: Added it to the interface. Went with a plain attribute annotation instead of a @property, though: SchemaResolver and the test mocks all assign self.platform_instance in __init__, and a getter-only property on the Protocol becomes an inherited read-only descriptor, so those assignments blow up with "property has no setter". platform gets away with being a property because everyone overrides it with their own getter and nothing assigns to it. Since platform_instance is genuinely a settable instance attr everywhere, the attribute annotation is the honest contract, and the connector now reads it via typed access instead of getattr.

PR #18234, opened 2026-07-07
fix(ingestion/superset): skip unresolvable owners instead of emitting invalid corpuser urn

alokr-dhub:
## Summary
- `build_owner_urn` now skips owner ids that don't resolve in `owner_info` instead of falling back to an empty username, which produced an invalid `urn:li:corpuser:` that GMS rejects and fails the whole record.
- Dashboard/chart/dataset builders now only append the `Ownership` aspect when at least one owner resolves, instead of unconditionally appending an empty-owners aspect.
- This surfaces when the Superset `related/owners` API is unavailable (e.g. 404s on some Superset deployments), so ownership degrades gracefully instead of crashing ingestion.

## Test plan
- [x] Added `test_superset_build_owner_urn_skips_unresolved_owners` — reproduces `related/owners` returning 404 and asserts no invalid urn is produced.
- [x] Added `test_superset_build_owner_urn_resolved_owners` — mixed resolvable/unresolvable owner ids, asserts only the resolvable one is emitted.
- [x] `./gradlew :metadata-ingestion:lintFix` passes.
- [x] Full `tests/unit/test_superset_source.py` suite passes (109 passed).


metadata-ingestion/src/datahub/ingestion/source/superset.py
```diff
@@ -552,12 +552,25 @@ def parse_owner_info(self) -> Dict[str, Any]:
...
-        ]
+    def build_owner_urn(self, data: Dict[str, Any], entity_type: str) -> List[str]:
+        # owner_info can be missing an entry (e.g. the related/owners API is
+        # unavailable, as on newer Superset builds, or the owner has no email);
+        # skip unresolvable owners rather than emitting an invalid
+        # urn:li:corpuser: with no username.
+        owner_urns = []
+        for owner in data.get("owners", []):
+            owner_id = owner.get("id")
+            owner_email = self.owner_info.get(owner_id) if owner_id else None
```
puneetagarwal-datahub: simplify ownerid resolution

puneetagarwal-datahub (changes requested):

metadata-ingestion/src/datahub/ingestion/source/superset.py
alokr-dhub: refactored it to make it more readable

metadata-ingestion/src/datahub/ingestion/source/superset.py
```diff
@@ -552,12 +552,26 @@ def parse_owner_info(self) -> Dict[str, Any]:
...
+            if owner_id:
+                owner_email = self.owner_info.get(owner_id)
+                if owner_email:
+                    owner_urns.append(make_user_urn(owner_email))
+                else:
+                    self.report.warning(
+                        title="Unresolvable Superset owner",
+                        message="Skipped an owner with no resolvable email; ownership may be incomplete",
+                        context=f"Owner ID: {owner_id}, Entity Type: {entity_type}, Entity ID: {data.get('id', 'unknown')}",
+                        log=False,
```
sgomezvillamor: any reason to prevent this appearing in the logs


```suggestion
```

sgomezvillamor (approved):

puneetagarwal-datahub (approved):

PR #18239, opened 2026-07-07
feat(ingestion): Support status aspect patching with auto_status_aspect.py

chriscollins3456:
There's a bug in auto_status_aspect.py where we are always overwriting the status aspect. Instead this should be a patch MCP since there is other info in this aspect like lifecycleStage.

This supports PATCH for the status aspect in GMS and then checks the server version on the auto status aspect populator and uses patch if it's supported.


metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_status_aspect.py
```diff
@@ -19,16 +20,50 @@
...
+    """Check whether the connected GMS server supports PATCH on the status aspect.
+
+    The StatusTemplate was added in DataHub v1.7.0 / Cloud v2.1.0. GMS
+    versions without it reject PATCH MCPs for status with a 500 error.
+    Returns False when the server config is unavailable (e.g. file sink).
+    """
+    from datahub.utilities.server_config_util import (
+        RestServiceConfig,
+        get_gms_config,
+    )
```
sgomezvillamor: In general, we prefer imports at the top
Is there any reason for having these nested? circular deps doesn't seem an issue here

metadata-ingestion/src/datahub/ingestion/workunit_processors/auto_status_aspect.py
```diff
@@ -69,7 +107,23 @@ def process(self, stream: Iterable[MetadataWorkUnit]) -> Iterable[MetadataWorkUn
...
             self.report.num_status_aspects_emitted += 1
-            yield MetadataChangeProposalWrapper(
-                entityUrn=urn,
-                aspect=StatusClass(removed=False),
-            ).as_workunit()
+            if use_patch:
+                # PATCH only sets removed=false without overwriting other
+                # fields (lifecycleStage, lifecycleState, lifecycleLastUpdated).
+                patch_builder = MetadataPatchProposal(urn)
+                patch_builder._add_patch(
```
sgomezvillamor: `patch_builder._add_patch(...)` calls a private method from outside its class — it works, but it's a code smell.
Adding a patch helper or `StatusPatchBuilder` is probably out of scope for this PR, but could we leave a TODO comment for now?

metadata-ingestion/tests/unit/api/source_helpers/test_source_helpers.py
sgomezvillamor, line 1: ⚠️ Coverage gap: the version-gating in `_gms_supports_status_patch()` itself (cloud vs core thresholds, missing-config → False) isn't unit-tested — only the `patch`/`upsert` branch selection is. That gating logic is the crux of the change; a small test would be well spent.

sgomezvillamor (approved):
LGTM. The import is at module scope now, the `StatusPatchBuilder` is a much cleaner way to express the patch than reaching into `_add_patch` directly, and the version-gating logic is well covered by the new tests.

---

PR #18244, opened 2026-07-07
feat(usage): add usage registry CI gate for GraphQL, OpenAPI, and Rest.li

david-leifker:
## Summary

- Add incremental CI gates (`graphqlUsageRegistryCheck`, `openApiUsageRegistryCheck`, `restLiUsageRegistryCheck`) wired into `:metadata-service:configuration:check` that scan the full API surface and fail when handlers are uninstrumented or GraphQL operations lack `usage_operations.yaml` classification.
- Introduce reviewable exemption snapshot YAMLs (OpenAPI: `PgQueueMessagingOperationsController`; GraphQL/Rest.li: empty) plus snapshot update Gradle tasks.
- Fix OpenAPI usage instrumentation classifications (`SEARCH_QUERY`, `OTHER_READ`) and explicitly exempt pgQueue admin messaging from usage tracking.

## Test plan

- [x] `./gradlew :metadata-service:configuration:check -PgraphqlUsageCheckForce=true -PopenApiUsageCheckForce=true -PrestliUsageCheckForce=true`
- [x] Verified OpenAPI gate fails when PgQueue exemption is removed (detects uninstrumented handler)
- [x] Unit tests for scanners, classifiers, exemption reconciliation, and change-path matching


metadata-service/openapi-servlet/src/main/java/io/datahubproject/openapi/operations/messaging/PgQueueMessagingOperationsController.java
```diff
@@ -113,9 +112,7 @@ private boolean authorize(HttpServletRequest request, String operation) {
     OperationContext opContext =
         OperationContext.asSession(
             systemOperationContext,
-            RequestContext.builder()
-                .buildOpenapi(actorUrnStr, request, operation, List.of())
-                .withUsageOperation(UsageOperation.OTHER_OPERATIONS),
```
RyanHolstien, line 118: Not important to track?

RyanHolstien (approved):

metadata-service/openapi-servlet/src/main/java/io/datahubproject/openapi/operations/messaging/PgQueueMessagingOperationsController.java
david-leifker: Yeah, let me track it. This is actually no different then the kafka endpoints. It actually is not part of the transport pgQueue

PR #18245, opened 2026-07-07
fix(openapi/lineage): Refactor lineage scroll to be more lineage-specific

asikowitz:
The previous implementation had no way of request only upstreams or only downstreams for a set of urns, because the upstream for a lineage relationship can be in either the source or destination depending on the `isUpstream` flag. This. We haven't had an OSS release since this was first introduced so I see no need for backwards compatibility.

The strategy here is to query lineage edges the graph index in both directions (INCOMING + OUTGOING) for the set of urns and then filter to the appropriate lineage direction, if provided, after. I don't think we can do better than this without changing the graph index.


smoke-test/tests/openapi/v3/test_scroll_entities.py
```diff
@@ -0,0 +1,262 @@
+"""Smoke tests for the scroll_entities DataHubGraph method."""
```
asikowitz, line 1: Broke up `test_scroll` into multiple files. This should be exactly copy pasted, same as `test_scroll_relationships`

metadata-service/openapi-servlet/src/main/java/io/datahubproject/openapi/controller/ScrollUtils.java
```diff
@@ -60,114 +66,42 @@ public static ResponseEntity<GenericScrollResult<GenericRelationship>> doScrollR
...
-
-    if (!AuthUtil.isAPIAuthorized(opContext, RELATIONSHIP, READ)) {
-      throw new UnauthorizedException(
-          authentication.getActor().toUrnStr()
-              + " is unauthorized to "
-              + READ
-              + " "
-              + RELATIONSHIP);
-    }
+        getOpContext(
```
asikowitz, line 73: Refactored this method by just pulling out common utilities, like `getOpContext` or `checkAuthorized`, rather than try to combine scrollRelationships and scrollLineage into a single method

RyanHolstien (approved):

PR #18247, opened 2026-07-08
feat(agent-skills): add ml-drift-rca skill for model drift root-cause

StephenSook:
## What

Adds a new agent skill, `ml-drift-rca`, under `.agent-skills/`. It teaches an agent to root-cause silent ML model degradation using DataHub: confirm the degradation is real (not benign drift), walk the model's lineage to the upstream table that changed, identify the owning team from catalog metadata, and write the finding back onto the catalog.

## Why

The person who detects a model degradation is rarely the person who caused it or owns the upstream data. DataHub's lineage graph is the bridge between the two. This skill encodes that traversal and the write-back, including a constraint that is easy to miss: the incident metamodel does not allow incidents on `mlModel`, so the write-back is split (a structured property and a document on the model, an incident on the upstream dataset).

## Contents

Follows the same structure as the existing `test-review` skill:

- `SKILL.md`: the procedure, multi-agent compatible (Claude Code, Cursor, Codex, Copilot, Gemini CLI, Windsurf).
- `standards/drift-root-cause.md`: drift versus degradation, the write-back split, and honesty (lineage-guided correlation, not proven causation).
- `references/datahub-apis.md`: the exact Agent Context Kit reads and the structured-property, document, and `raiseIncident` writes, including the metamodel constraint.
- `scripts/trace-model-lineage.sh`: a standard-library Python-in-bash helper that walks upstream lineage to source datasets and their owners.
- `templates/drift-causation.md`: the structured-property, RCA-document, and incident content templates.

## Testing

The traversal script was run against a live self-hosted DataHub Core instance and correctly walked an `mlModel` through its `mlFeature` entities to the upstream `dataset` and its owning `corpGroup`.

## Notes

This skill was built while implementing an on-call drift-diagnosis agent on open-source DataHub Core (Agent Context Kit for reads, structured properties and documents plus `raiseIncident` for the write-back). Happy to adjust the structure, naming, or scope to match project conventions.

### DIFF
--- a/.agent-skills/ml-drift-rca/SKILL.md
+++ b/.agent-skills/ml-drift-rca/SKILL.md
@@ -1,0 +1,83 @@
+# DataHub ML Drift Root-Cause
+
+You are an expert at diagnosing silent machine-learning model degradation using DataHub. Your role is to take a drift or performance signal for a production model, walk the model's lineage in DataHub to the specific upstream data change that caused it, identify the owning team, and record the finding durably back onto the catalog so the next engineer or agent inherits the diagnosis.
+
+The core insight: the person who detects a model degradation is rarely the person who caused it or has authority to fix it. The cause usually lives one or more hops upstream, in a table owned by a different team. DataHub's lineage graph is the bridge, and its metadata surfaces are where the answer should live.
+
+---
+
+## Multi-Agent Compatibility
+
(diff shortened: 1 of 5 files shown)


StephenSook:
Closing in favor of datahub-project/datahub-skills#29, which targets the user-facing DataHub Skills registry (the right home for this skill). Thanks!

PR #18253, opened 2026-07-08
fix(ingestion): materialize TABLESAMPLE into temp tables for Snowflake SQLAlchemy profiler

sgomezvillamor:
## Problem

The Snowflake SQLAlchemy profiler is 70-800x slower than the GE profiler on large tables because it re-evaluates `TABLESAMPLE BERNOULLI(...)` on every CTE in every combined query (the query combiner batches multiple scalar queries into a single cross-joined CTE query, and each CTE re-scans the source table). GE avoids this by materializing the sample into a temp table once, then querying the small temp table.

## Solution

Materialize the TABLESAMPLE sample into a session-scoped Snowflake temporary table once, then run all profiling queries against that small materialized table. This matches what GE/GX does internally when given a `custom_sql` batch kwarg.

### Key Changes

**1. Gate `custom_sql` to GE-only** (`snowflake_profiler.py`)
- The `custom_sql` TABLESAMPLE query is now only generated when `profiling.method == "ge"`
- GX internally creates a temp table from `custom_sql`, so all its profiling queries run against that materialized table
- For the SQLAlchemy profiler path, `custom_sql` is `None` — the adapter handles sampling directly

**2. Temp table materialization** (`adapters/snowflake.py`)
- `setup_profiling()` queries `INFORMATION_SCHEMA.TABLES` for instant row count (parameterized query)
- For large tables: `CREATE OR REPLACE TEMPORARY TABLE dh_sample_<uuid> AS SELECT * FROM ... TABLESAMPLE BERNOULLI(...)`
- Two-tier BLOCK+BERNOULLI for very large tables (>50M rows), with fallback to BERNOULLI-only if BLOCK fails (e.g. on views)
- No SEED — sample is materialized once so all profiling queries see the same rows; also allows `TABLESAMPLE BERNOULLI` to work on views
- Conservative approach: when `INFORMATION_SCHEMA` row count is unavailable, assumes table is large and samples anyway
- `custom_sql` path materializes into a temp table too (same as what GX does internally)
- Temp tables are session-scoped and auto-drop when the connection closes

**3. Partition spec fix** (`sqlalchemy_profiler.py`)
- Changed `elif custom_sql:` to `elif custom_sql or context.is_sampled:` so the partition spec is set correctly when the adapter performs sampling via temp table

**4. Query combiner observability** (`sqlalchemy_query_combiner.py`)
- Added `PerfTimer` instrumentation and short query IDs to combined and fallback query execution
- Upgraded log levels from DEBUG to INFO for query execution timing
- Enables correlating slow queries in production logs

**5. Others**
- Added deprecation comment on `custom_sql` field in `ProfilingContext`
- Upgraded adapter selection log from DEBUG to INFO (`adapters/__init__.py`)

### Testing

- `./gradlew :metadata-ingestion:lintFix` — passes
- `./gradlew :metadata-ingestion:testQuick` — passes
- 19 Snowflake adapter unit tests (new: setup_profiling, sampling, row count, cleanup)
- Snowflake integration tests (`connector-tests`): `test_profiling`, `test_profiling_ge`, `test_profiling_table_level`, `test_profiling_table_level_ge` — all pass
- End-to-end validation on a production Snowflake instance (~1,000 tables) confirmed behavioral parity with the GE profiler: same entities profiled, same sampling strategy (temp table per sampled table), and same order of magnitude per-table execution time

## Notes

- Uses existing config: `use_sampling=True` (default), `sample_size=10000` (default)
- No breaking changes to API or configuration
- No new dependencies


metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/snowflake.py
```diff
@@ -26,96 +27,211 @@ class SnowflakeAdapter(PlatformAdapter):
...
+            bernoulli_pc = 100 / overgeneration_factor
+            sample_sql = (
+                f"SELECT * FROM"
+                f" (SELECT * FROM {tablename} TABLESAMPLE BLOCK ({block_pc:.8f}))"
+                f" TABLESAMPLE BERNOULLI ({bernoulli_pc:.8f})"
+            )
+        else:
+            bernoulli_pc = 100 * sample_pc
+            sample_sql = (
+                f"SELECT * FROM {tablename} TABLESAMPLE BERNOULLI ({bernoulli_pc:.8f})"
```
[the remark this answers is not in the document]
sgomezvillamor: @AikidoSec ignore: False positive. The interpolated values are not user input — `tablename` is built from `context.schema` and `context.table` which come from Snowflake's own `INFORMATION_SCHEMA.TABLES` catalog, `bernoulli_pc` is a computed float formatted with `:.8f`, and `temp_name` is a UUID we generate. Parameterized queries can't be used here because `TABLESAMPLE` and `CREATE TEMPORARY TABLE` require SQL identifiers, not bind parameter values.

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/snowflake.py
```diff
@@ -26,96 +27,211 @@ class SnowflakeAdapter(PlatformAdapter):
...
+        except SQLAlchemyError as e:
+            if use_block_presample:
+                # BLOCK sampling may fail on views; retry with BERNOULLI only
+                logger.info(
+                    f"BLOCK sampling failed for {context.pretty_name}, falling back to BERNOULLI: "
+                    f"{type(e).__name__}: {e}"
+                )
+                bernoulli_pc = 100 * sample_pc
+                fallback_sql = (
+                    f"SELECT * FROM {tablename}"
```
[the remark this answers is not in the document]
sgomezvillamor: @AikidoSec ignore: Same as above — false positive. This is the BLOCK→BERNOULLI fallback path using the same `tablename` (from `INFORMATION_SCHEMA.TABLES`) and `bernoulli_pc` (a computed float). No user input is interpolated.

metadata-ingestion/tests/unit/sqlalchemy_profiler/test_adapters.py
```diff
@@ -716,6 +717,317 @@ def test_quote_identifier_with_schema(self, adapter):
...
+
+        with patch("sqlalchemy.Table") as mock_table_class:
+            mock_table_class.return_value = MagicMock()
+            adapter._create_sampled_temp_table(context, mock_conn, row_count=100_000)
+
+        # Verify CREATE TEMPORARY TABLE with TABLESAMPLE BERNOULLI
+        executed_sql = str(mock_conn.execute.call_args[0][0])
+        assert "CREATE OR REPLACE TEMPORARY TABLE" in executed_sql
+        assert "dh_sample_" in executed_sql
+        assert "TABLESAMPLE BERNOULLI" in executed_sql
```
treff7es: **Should fix:** this test runs the real sampling arithmetic but never checks the computed percentages — only `"TABLESAMPLE BERNOULLI" in sql`. Drop the `* 100`, change a constant, or emit a >100% fraction (which Snowflake rejects) and this stays green. Since the percentage math is the whole point of the PR, assert the exact substring for `row_count=100_000` / `sample_size=10_000` → `TABLESAMPLE BERNOULLI (10.00000000)`, and in `test_sampled_temp_table_block_bernoulli_sql` assert `BLOCK (10.00000000)` + `BERNOULLI (0.10000000)`. Ideally also add a guard that `block_pc <= 100`.

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/snowflake.py
```diff
@@ -26,96 +27,211 @@ class SnowflakeAdapter(PlatformAdapter):
...
+        if not self.config.limit and self.config.use_sampling:
+            if row_count is not None and row_count <= self.config.sample_size:
+                # Table is small enough — profile directly without sampling
+                context.sql_table = self._create_sqlalchemy_table(
+                    schema=context.schema,
+                    table=context.table,
+                )
+            else:
+                # Either the table is confirmed large, or we couldn't get the
+                # row count. Be conservative and sample in both cases.
```
treff7es, line 66: **Should fix:** when the row count is unavailable this silently downgrades to a ~10% sample with only a `logger.debug`, even though `self.report` is in hand. Two real triggers: (a) views — `INFORMATION_SCHEMA.TABLES.ROW_COUNT` is NULL for views → always sampled, and for a *large* view materializing 10% undercuts the perf goal; (b) a genuinely tiny table whose lookup failed gets a 10% BERNOULLI sample that can return 0 rows → a `rowCount=0` profile stamped `partition="SAMPLE"` with no warning. Please emit a `self.report.warning`/`info` (+ a counter) when conservative sampling kicks in so operators can see "row count unknown → table X sampled".

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/snowflake.py
```diff
@@ -26,96 +27,211 @@ class SnowflakeAdapter(PlatformAdapter):
...
         """
-        # No temp resources created in Snowflake setup
-        pass
+        temp_name = f"dh_sample_{uuid.uuid4().hex[:8]}"
+        sample_pc = self.config.sample_size / row_count
 
-    # =========================================================================
-    # Table Creation
-    # =========================================================================
+        estimated_block_row_count = 500_000
```
treff7es, line 132: **Should fix (dedupe):** these constants and the block-vs-bernoulli decision are duplicated verbatim from the GE path in `snowflake_profiler.py` `get_batch_kwargs` (`estimated_block_row_count=500_000`, `overgeneration_factor=1000`, `block_profiling_min_rows`, same `use_block_presample` math). Only the SQL wrapper differs. Re-tune one and the two profiling backends silently diverge for the same connector. Extract a pure `build_tablesample_fragment(row_count, sample_size)` helper (constants + branching + `:.8f` formatting) and have both call sites keep their own SQL assembly — it also gives the math a single unit-testable home.

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/snowflake.py
```diff
@@ -26,96 +27,211 @@ class SnowflakeAdapter(PlatformAdapter):
...
+        # concrete table, same as what GX does internally.
+        if context.custom_sql:
+            logger.info(
+                f"Creating temp table from custom_sql for {context.pretty_name}"
+            )
+            context = self._create_temp_table_from_custom_sql(context, conn)
+            return context
+
+        # Determine if sampling is needed
+        row_count = self._get_row_count_from_metadata(context, conn)
```
treff7es: **Should fix (perf, on-theme):** this issues a fresh `INFORMATION_SCHEMA` round-trip per table, but the value is already known upstream — the GE path gates on `table.rows_count` (fetched during schema crawl), and `ProfilingContext.row_count` exists as a field but is never populated. `BigQueryAdapter._setup_sampling` already shows the intended pattern: `row_count = context.row_count or self._get_quick_row_count(...)`. Thread `table.rows_count` into the context and fall back to `_get_row_count_from_metadata` only when it's `None` — removes one metadata query per profiled table.

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/snowflake.py
```diff
@@ -26,96 +27,211 @@ class SnowflakeAdapter(PlatformAdapter):
...
-        )
-
-        logger.debug(
-            f"Snowflake setup for {context.pretty_name}: "
-            f"schema={context.schema}, table={context.table}"
-        )
+        # custom_sql is or may still be used for some use cases (not non-sampling).
+        # Materialize it into a temp table so profiling queries run against a small
+        # concrete table, same as what GX does internally.
+        if context.custom_sql:
```
treff7es: **Suggestion:** since `custom_sql` is now gated GE-only in `snowflake_profiler.py`, nothing populates `context.custom_sql` on the SQLAlchemy Snowflake path — this branch is only reachable from its unit test today. Either remove it until a real producer exists, or add a comment naming the future producer (e.g. anticipated partition support), so it doesn't read as a live capability. Related: `_create_temp_table_from_custom_sql` sets `is_sampled=True` for a filtered query that isn't a statistical sample, which would mislabel the profile as `SAMPLE` — worth fixing if/when this path goes live.

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/snowflake.py
```diff
@@ -26,96 +27,211 @@ class SnowflakeAdapter(PlatformAdapter):
...
+                )
+                logger.debug(f"Fallback SQL: {create_sql}")
+                conn.execute(sa.text(create_sql))
+            else:
+                raise
+
+        # Reflect the temp table as a real sa.Table
+        metadata = sa.MetaData()
+        context.sql_table = sa.Table(temp_name, metadata, autoload_with=conn)
+        context.is_sampled = True
```
treff7es, line 199: **Suggestion:** `sample_percentage` on `ProfilingContext` exists but is never set — populate it here (`bernoulli_pc`, or the effective fraction) so operators can see *how much* was sampled, not just that it was. Also, the "CREATE TEMPORARY TABLE → reflect as `sa.Table` → set `is_sampled`/`temp_table`" tail is copy-pasted 3× in this file (here, the BLOCK-fallback path, and `_create_temp_table_from_custom_sql`); a small `_materialize_temp_table(context, conn, name, sql)` helper would remove the triplication.

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/snowflake.py
```diff
@@ -26,96 +27,211 @@ class SnowflakeAdapter(PlatformAdapter):
...
+        sample_pc = self.config.sample_size / row_count
 
-    # =========================================================================
-    # Table Creation
-    # =========================================================================
+        estimated_block_row_count = 500_000
+        block_profiling_min_rows = 100 * estimated_block_row_count
+        overgeneration_factor = 1000
+
+        tablename = f'"{context.schema}"."{context.table}"'
```
treff7es: **Minor (pre-existing pattern):** `"{schema}"."{table}"` isn't escaped for embedded double-quotes — a legitimately quoted Snowflake identifier containing a `"` would break out of the quoting. Not introduced by this PR (the same pattern exists elsewhere in the profiler) and not what aikido flagged, but worth a `.replace('"', '""')` on each part while you're here.

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/snowflake.py
```diff
@@ -26,96 +27,211 @@ class SnowflakeAdapter(PlatformAdapter):
...
+        create_sql = f"CREATE OR REPLACE TEMPORARY TABLE {temp_name} AS {sample_sql}"
+
+        try:
+            logger.info(
+                f"Creating sampled temp table for {context.pretty_name} (rows: {row_count:,}, "
+                f"sample: {self.config.sample_size:,})"
+            )
+            logger.debug(f"SQL: {create_sql}")
+            conn.execute(sa.text(create_sql))
+        except SQLAlchemyError as e:
```
treff7es, line 174: **Minor:** this catches *any* `SQLAlchemyError` but the log message always blames "BLOCK sampling failed" (view incompatibility), which is misleading when the real cause is permissions / suspended warehouse / missing object. The BERNOULLI retry is uncaught so genuine failures still propagate — so low severity — but please make the message honest ("initial BLOCK+BERNOULLI sampling failed, retrying BERNOULLI-only") and route it through `self.report`.

metadata-ingestion/src/datahub/utilities/sqlalchemy_query_combiner.py
```diff
@@ -339,9 +345,17 @@ def _execute_queue(self, main_greenlet: greenlet.greenlet) -> None:
             for cte in ctes.values():
                 combined_query.append_from(cte)
 
-            logger.debug(f"Executing combined query: {str(combined_query)}")
+            query_id = SQLAlchemyQueryCombiner._generate_query_id()
             self.report.combined_queries_issued += 1
-            sa_res = _sa_execute_underlying_method(queue_item.conn, combined_query)
+            logger.info(
+                f"[{query_id}] Executing combined query ({len(pending_queue)} queries combined): {str(combined_query)}"
```
treff7es: **Suggestion (log noise):** the `[query_id]` correlation + `PerfTimer` are genuinely useful, but logging the full `str(combined_query)` at INFO on every combined query (and the fallback equivalent) is a big verbosity jump for a utility used across all SQL profilers — a large run emits one long SQL line per batch at normal verbosity. Consider keeping the SQL body at DEBUG and logging only the `[query_id]` + timing at INFO.

treff7es (approved):

metadata-ingestion/tests/unit/sqlalchemy_profiler/test_adapters.py
sgomezvillamor: Good call — addressed in a4c6143:

- Tests now assert exact percentage substrings (`TABLESAMPLE BERNOULLI (10.00000000)`, `BLOCK (10.00000000)` + `BERNOULLI (0.10000000)`) so any math regression will be caught.
- Added `min(block_pc, 100)` as a defensive cap. The `use_block_presample` guard (`row_count > sample_size * overgeneration_factor`) already guarantees `block_pc < 100`, but the explicit cap makes the invariant self-documenting.

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/snowflake.py
sgomezvillamor: Confirmed — `custom_sql` is indeed GE-only (gated by `method == "ge"` in `snowflake_profiler.py`). Replaced the dead handling branch with an assert in 3e90221:

- `if context.custom_sql:` → `assert not context.custom_sql` with a message explaining it's GE-only
- Removed the now-unreachable `_create_temp_table_from_custom_sql` method
- Updated the test to verify the assert fires when `custom_sql` is passed

This also sidesteps the `is_sampled=True` mislabeling issue you flagged — since the method no longer exists.

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/snowflake.py
sgomezvillamor: Good catch — addressed in 726b76d:

- `snowflake_profiler.py`: now passes `"row_count": table.rows_count` in `get_batch_kwargs`
- `sqlalchemy_profiler.py`: threads `row_count` through `_generate_single_profile` → `ProfilingContext`
- `snowflake.py` adapter: `row_count = context.row_count or self._get_row_count_from_metadata(context, conn)` — skips the INFORMATION_SCHEMA query when the count is already known from the schema crawl
- Added `test_setup_profiling_uses_context_row_count` that asserts `_get_row_count_from_metadata` is not called when `context.row_count` is pre-populated

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/snowflake.py
sgomezvillamor: Addressed in 95910f9 — added `self.report.warning` when conservative sampling kicks in due to unknown row count, so operators can see it in the ingestion report.

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/snowflake.py
sgomezvillamor: Valid concern — the constants and branching logic are duplicated between the two paths. Skipping the extraction for now since the GE profiler is being fully removed very soon, which will eliminate the duplication entirely (only the SQLAlchemy adapter path will remain).

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/snowflake.py
sgomezvillamor: Fixed for the SQLAlchemy path in 1023e67:

- Made `db_name` optional in `get_quoted_identifier_for_table` (returns `"schema"."table"` when `None`)
- SQLAlchemy adapter now uses `SnowflakeIdentifierBuilder.get_quoted_identifier_for_table(db_name=None, ...)` instead of raw f-string quoting
- Added two tests for the `db_name=None` variant in `TestSnowflakeIdentifierQuoting`

Not fixing the GE path (`snowflake_profiler.py:120`) since GE profiler is being removed soon.

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/snowflake.py
sgomezvillamor: Addressed in c243a32 — replaced the misleading "BLOCK sampling failed" `logger.info` with a `self.report.warning` that honestly states "BLOCK+BERNOULLI sampling failed, retrying with BERNOULLI-only" and lists the actual possible causes (views, permissions, suspended warehouse).

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/snowflake.py
sgomezvillamor: `sample_percentage` addressed in c6cb85b — now set to `bernoulli_pc` (the effective BERNOULLI percentage, whether from the initial path or the BLOCK-fallback).

The code duplication concern no longer applies — `_create_temp_table_from_custom_sql` was removed in an earlier commit (replaced with an assert), so the tail only exists in the main path + BLOCK-fallback, which is the natural shape of a try/except.

metadata-ingestion/src/datahub/utilities/sqlalchemy_query_combiner.py
sgomezvillamor: Addressed in a1a7d8b — SQL body moved to DEBUG, query_id + count/timing stays at INFO for both the combined and fallback paths.

PR #18254, opened 2026-07-08
fix(ingest/unity-catalog): honor include_views and view_pattern

max-datahub:
## Summary

The Unity Catalog (Databricks) source inherits several filter fields from `SQLCommonConfig`, so recipes can set them without a validation error — but the source **never applied them**, silently ignoring:

- `include_views: false` → views still ingested
- `view_pattern` → ignored
- `include_tables: false` → tables still ingested

This wires all three into `process_tables`, gating on the already-computed `table.is_view` / `table.is_metric_view`, mirroring how the SQLAlchemy base and the Snowflake/BigQuery sources filter these.

## Changes

- `unity/source.py` (`process_tables`):
  - **Views** (VIEW / MATERIALIZED_VIEW / HIVE_VIEW): dropped when `include_views` is false or `view_pattern` denies them.
  - **Regular tables** (neither view nor metric view): dropped when `include_tables` is false.
  - Dropped entities are not added to `view_refs` / `table_refs`, so stateful ingestion soft-deletes anything previously ingested.
- `view_pattern` and `include_views` are applied from the values **inherited from `SQLCommonConfig`** — not redeclared on the Unity config, to avoid drift from the base defaults.
- **Metric views** keep their dedicated `include_metric_views` / `metric_view_pattern` handling and are unaffected by either toggle.

## Backward compatibility

Fully compatible. `view_pattern` defaults to `table_pattern` (via the existing `SQLCommonConfig` validator); `include_views` and `include_tables` default to `true`. Existing recipes behave exactly as before.

## Tests

- **Unit** (`TestUnityCatalogViewFiltering`): views dropped via `include_views` / `view_pattern`; ordinary tables unaffected by view filters; `include_tables: false` drops only regular tables (views and metric views exempt via their own toggles).
- **Integration** (`test_unity_catalog_ingest.py`): full-pipeline assertions that views ingest by default and are skipped under `include_views: false` / `view_pattern` deny, and that `include_tables: false` skips tables while keeping views.
- Verified end-to-end against a live Unity Catalog workspace for all three toggles.

## Checklist

- [x] Tests added
- [ ] Docs (config field docs auto-generate from field descriptions)
- [x] No breaking changes (defaults preserve prior behavior)

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/unity/config.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/unity/config.py
@@ -228,6 +228,16 @@ class UnityCatalogSourceConfig(
         ),
     )
 
+    view_pattern: AllowDenyPattern = Field(
+        default=AllowDenyPattern.allow_all(),
+        description=(
+            "Regex patterns for views to filter in ingestion."
+            " Specify regex to match the entire view name in `catalog.schema.view` format."
+            " Defaults to `table_pattern` if not specified."
+            " Only applies when `include_views` is True; set `include_views: false` to skip all views."
+        ),
+    )
+
     metric_view_pattern: AllowDenyPattern = Field(
         default=AllowDenyPattern.allow_all(),
         description=(
--- a/metadata-ingestion/src/datahub/ingestion/source/unity/config.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/unity/config.py
@@ -228,15 +228,8 @@ class UnityCatalogSourceConfig(
         ),
     )
 
-    view_pattern: AllowDenyPattern = Field(
-        default=AllowDenyPattern.allow_all(),
-        description=(
-            "Regex patterns for views to filter in ingestion."
-            " Specify regex to match the entire view name in `catalog.schema.view` format."
-            " Defaults to `table_pattern` if not specified."
-            " Only applies when `include_views` is True; set `include_views: false` to skip all views."
-        ),
-    )
+    # view_pattern and include_views are inherited from SQLCommonConfig and applied
+    # in process_tables; not redeclared here to avoid drift from the base defaults.
 
     metric_view_pattern: AllowDenyPattern = Field(
         default=AllowDenyPattern.allow_all(),
--- a/metadata-ingestion/src/datahub/ingestion/source/unity/source.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/unity/source.py
@@ -726,6 +726,16 @@ class UnityCatalogSource(StatefulIngestionSourceBase, TestableSource):
                 self.report.tables.dropped(table.id, f"table ({table.table_type})")
                 continue
 
+            # Views (VIEW / MATERIALIZED_VIEW / HIVE_VIEW) are honored via
+            # include_views + view_pattern, mirroring SQL-based sources. Metric
+            # views are not is_view and keep their dedicated filtering below.
+            if table.is_view and (
+                not self.config.include_views
+                or not self.config.view_pattern.allowed(table.ref.qualified_table_name)
+            ):
+                self.report.tables.dropped(table.id, f"view ({table.table_type})")
+                continue
+
             if (
                 table.is_metric_view
                 and self.config.include_metric_views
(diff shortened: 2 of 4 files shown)


treff7es (approved):

max-datahub:
Verified end-to-end against a live Unity Catalog workspace (metadata-only run over one schema containing two tables and one view):

| Recipe | View ingested? | Tables ingested? |
|---|---|---|
| default | ✅ yes | ✅ yes |
| `include_views: false` | ❌ dropped | ✅ yes |
| `view_pattern: { deny: ['.*'] }` | ❌ dropped | ✅ yes |

In both filtered runs the view appears under `report.tables.dropped` and ordinary tables are unaffected, consistent with the unit + integration coverage added here.

metadata-ingestion/src/datahub/ingestion/source/unity/config.py
```diff
@@ -228,6 +228,16 @@ class UnityCatalogSourceConfig(
         ),
     )
 
+    view_pattern: AllowDenyPattern = Field(
```
sgomezvillamor: > Unity Catalog (Databricks) source inherits include_views and view_pattern from SQLCommonConfig

if `view_pattern` is inherited, why redefining it here in the unity source?

max-datahub: > Good question. The field is redefined purely to override the description, not the behavior.
> 
> The inherited SQLCommonConfig.view_pattern has a SQL-centric description referring to database.schema.view format. Unity Catalog uses a catalog.schema.view naming hierarchy, so the inherited wording is misleading for this source. Redeclaring lets the auto-generated config docs show Unity-correct terminology and note the include_views interaction.
> 
> This mirrors an existing pattern in the same file: Unity already redeclares catalog_pattern, schema_pattern, and table_pattern from the base config for exactly the same reason — to give them Unity-appropriate descriptions (e.g. metastore.catalog, catalog.schema.table).
> 
> The functional behavior is unchanged:
> - The default (AllowDenyPattern.allow_all()) matches the base.
> - The view_pattern → table_pattern fallback still works, because it's driven by the @model_validator on SQLFilterConfig/SQLCommonConfig, which runs regardless of where the field is declared. Overriding the Field only replaces the default/description metadata, not the validator.
> 
> So it's a documentation/UX improvement for the generated Unity docs, consistent with the sibling pattern fields — not a redundant or behavior-changing redefinition.
> 
> If you'd prefer to avoid the redeclaration, the alternative would be leaving view_pattern inherited (accepting the database.schema.view wording in the docs), the way include_views is left inherited here.

sgomezvillamor: I'm concerned on these overwrites, because drifts (eg different default values) can happen accidentally.

Comparing descriptions:

```
        description="Regex patterns for views to filter in ingestion. Note: Defaults to table_pattern if not specified. Specify regex to match the entire view name in database.schema.view format. e.g. to match all views starting with customer in Customer database and public schema, use the regex 'Customer.public.customer.*'",
````

vs

```
        description=(
            "Regex patterns for views to filter in ingestion."
            " Specify regex to match the entire view name in `catalog.schema.view` format."
            " Defaults to `table_pattern` if not specified."
            " Only applies when `include_views` is True; set `include_views: false` to skip all views."
        ),
```

description is the same but "only applies when `include_views` is True"

So wondering if this is specific behaviour that we are implementing for unity only, or standard behaviour that we do in others sources too. If the first, why? If the second, I would suggest to promote it to description in sql common. 

> If you'd prefer to avoid the redeclaration, the alternative would be leaving view_pattern inherited (accepting the database.schema.view wording in the docs), the way include_views is left inherited here.

That small diff can be addressed without a redeclaration, just providing a couple of examples in the fields' description in sql common

max-datahub:
Good call on the drift risk — addressed in 47a91d4c53. Dropped the `view_pattern` redeclaration, so it and `include_views` are now inherited straight from `SQLCommonConfig` (consistent with how `include_views` was already left inherited). Behavior is unchanged: `view_pattern` still defaults to `table_pattern` via the base validator, and deny still works.

On the one line I'd added — "only applies when `include_views` is True" — you're right that it's standard SQL-source behavior, not Unity-specific, so it shouldn't live as an override here. If we want it in the docs, better to add it to the `view_pattern` description in `sql_common` (separate PR) so every source benefits rather than just this one.

PR #18262, opened 2026-07-08
fix(i18n): add missing modules.outputPorts.moduleName translation to all locales

max-datahub:
## Summary

Adds the missing `modules.outputPorts.moduleName` translation to all five non-English locales (de, es, fr, it, pt-BR), restoring full translation parity with the English source.

## Root cause

PR #17740 ("feat: add output ports module") introduced the English key `outputPorts.moduleName` = `"Output Ports"` in `datahub-web-react/src/i18n/locales/en/modules.json` but did not add the corresponding translations. As a result, every non-English locale was exactly one key short of parity, and both parity gates reported drift:

- `node scripts/check-translations.mjs --lang <l>` → `1 missing key`
- `I18N_PARITY_MODE=fail node scripts/check-i18n-parity.mjs --lang <l>` → `1 issue`

## Change

The key is added to each locale's `modules.json` at the same position it occupies in `en/modules.json` (the parity check is order-aware). Each value matches the locale's already-established rendering of the same data-mesh concept in `entity.types.json` (`dataProduct.outputPortsCountTitle`), so terminology and casing stay internally consistent per locale.

| Locale | Value added |
| ------ | ----------- |
| de     | Output-Ports |
| es     | Puertos de salida |
| fr     | Ports de sortie |
| it     | Porte di output |
| pt-BR  | Portas de Saída |

## Verification

Both gates now report full parity for all five locales:

- `check-translations.mjs` → `coverage: 100%`, all keys present
- `check-i18n-parity.mjs` (`I18N_PARITY_MODE=fail`) → `Issues: 0`, all locales in sync with en

An independent flat-key diff across all 67 namespaces confirms the single `modules:outputPorts.moduleName` gap was the only drift (no missing/extra keys elsewhere, all namespaces present, remirror translations already at parity). Only the five `modules.json` files are modified; the English source is untouched.


acrylJonny (approved):
LGTM

PR #18266, opened 2026-07-08
perf(graphql): batch-load Dashboard.usageStats buckets and metrics

gdatahub:
## What this does

Resolving `Dashboard.usageStats` issued **three** Elasticsearch reads **per dashboard** — daily usage buckets, per-user counts, and the absolute-metrics docs — so a request covering **N** dashboards fanned out to ~**3N** queries. This PR batches and trims those reads so a page collapses toward a small constant.

> `usageStats` is not used by the DataHub UI today (no frontend `.graphql` selects it); it is for direct / external consumers. The multi-dashboard numbers below describe that pattern.

## Changes

1. **Batch the daily buckets** — new `DashboardUsageBucketsBatchLoader` issues one `batchGetAggregatedStats` per time window (`terms(urn)` + daily `date_histogram`) instead of one query per dashboard. Gated by `timeseriesAspectAggBatchLoadEnabled`; per-URN fallback preserved; single-URN requests skip the batch (nothing to collapse).
2. **Fetch absolute metrics only when requested** — those docs feed only the `metrics` field, yet were fetched on every request. Now skipped unless the caller selects `metrics`.
3. **Bound the metrics default** — `metrics` with no `limit` is now capped at **100** so it collapses to a single `top_hits` query (was up to 5000 per dashboard, per-URN). An explicit `limit > 100` still falls back to the per-URN path.
4. **Fix** — `executionsCount` was gated on the *views* cardinality column, so a bucket with executions but no views dropped its `executionsCount`; now keyed off its own column.

The **per-user aggregation stays per-URN by design** — it returns the full, uncapped user list, which would blow the `urns × users` bucket budget under a batched terms aggregation. This is the remaining `N`.

## Elasticsearch calls per request (N dashboards)

| Query shape | Before | After |
| --- | --- | --- |
| buckets + aggregations (no `metrics`) | 3N | **N + 1** |
| + `metrics`, no `limit` | 3N | **N + 2** |
| + `metrics`, `limit > 100` | 3N | **2N + 1** |

After = 1 batched buckets query + N per-user queries + {0, 1, N} metrics queries. The per-user `N` is the intentional floor (see above).

## Latency

Buckets + aggregations query (the common shape), local single-node OpenSearch over loopback. Median of 7 iterations (first dropped as cold), 5 warm-ups, 120 seeded dashboards:

| Dashboards / request | Before (per-URN) | After (this PR) | Δ |
| --- | --- | --- | --- |
| 1 | 13.3 ms | 12.6 ms | ~0 |
| 10 | 36.3 ms | 23.0 ms | **−37%** |
| 20 | 42.9 ms | 32.2 ms | **−25%** |
| 50 | 80.7 ms | 50.8 ms | **−37%** |
| 100 | 141.9 ms | 94.0 ms | **−34%** |

The improvement is driven mainly by **not fetching the unused metrics docs**. Buckets-batching *alone* is roughly neutral on a single-node loopback — there is no network round-trip to save and the per-URN queries already run concurrently — so its real benefit (fewer ES round-trips and connection/thread pressure) shows under production network latency and cluster load, which this environment does not reproduce.

## Behavior change

`usageStats { metrics }` with **no `limit`** now returns at most **100** absolute snapshots (was up to 5000). No change for dashboards with ≤100 in-window snapshots; callers needing more pass an explicit `limit`. Documented in `docs/how/updating-datahub.md`.

## Correctness

The batched and per-URN paths compute **identical responses** — verified by a smoke test running the same exact-value assertions with the flag on and off. The `executionsCount` fix is covered by a unit test that fails before the change and passes after.

## Testing

- **Unit:** `DashboardUsageBucketsBatchLoaderTest` (single-URN guard, multi-URN batch, multi-window grouping), `DashboardUsageStatsResolverTest` (flag routing + metrics-skipped-when-unselected), `DashboardUsageStatsUtilsTest` (aggregation mapping incl. the executions-but-no-views case).
- **Smoke:** `test_dashboard_usage_stats_graphql.py` — distinct per-dashboard data, exact assertions, out-of-window exclusion; idempotent (fixed-timestamp seeding); passes with the flag on and off.

## Checklist

- [x] Tests added / updated
- [x] Breaking change documented in `docs/how/updating-datahub.md`


gdatahub:
Screenshot 2026-07-24 at 7 03 28 PM
Screenshot 2026-07-24 at 7 14 59 PM
Screenshot 2026-07-24 at 7 16 44 PM

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/dashboard/DashboardUsageStatsResolver.java
```diff
@@ -39,92 +32,246 @@
 @Slf4j
 public class DashboardUsageStatsResolver
     implements DataFetcher<CompletableFuture<DashboardUsageQueryResult>> {
-  private static final String ES_FIELD_EVENT_GRANULARITY = "eventGranularity";
+
+  // Default cap for the absolute-metrics list when the caller passes no `limit`. Kept at/below
+  // top_hits' index.max_inner_result_window (OpenSearch default 100) so the null-limit fetch can be
+  // batched into one top_hits aggregation across a page of dashboards. A null (unbounded, up-to-
+  // 5000) default would exceed that cap and force the per-URN path. Callers needing more pass an
+  // explicit larger limit (which reverts to per-URN). Promote to a config property if it needs to
```
chakru-r, line 40: This is not true. See TimeseriesAspectBatchLoader. a limit > 100 crosses ES threshold and will fail. Need to pass null to skip the batch query path and fallback to per-URN limit. Change TimeseriesAspectBatchLoader to handle >100 as well to fallback.

chakru-r (changes requested):

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/dashboard/DashboardUsageStatsResolver.java
gdatahub: limit > 100 is getting falls back,  handled in the ElasticSearchTimeseriesAspectService.java

metadata-io/.../ElasticSearchTimeseriesAspectService.java:516-541
```
int perBucketLimit = timeseriesAspectServiceConfig.getTopHitsPerBucketLimit();  // 100
if (limit > perBucketLimit) {
  log.warn("batchGetAspectValues limit {} exceeds topHitsPerBucketLimit {} ...");
  for (Urn urn : urns) { result.put(urn, getAspectValues(...)); }
  return result;   // returns before any top_hits builder is constructed
}
```

chakru-r (approved):

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/dashboard/DashboardUsageStatsResolver.java
```diff
@@ -39,92 +32,246 @@
...
+        new TimeseriesAspectBatchLoader.Key(
+            dashboardUrn,
+            Constants.DASHBOARD_ENTITY_NAME,
+            Constants.DASHBOARD_USAGE_STATISTICS_ASPECT_NAME,
+            maybeStartTimeMillis,
+            maybeEndTimeMillis,
+            maybeLimit,
+            buildSharedUsageFilter(null, null, false),
+            null);
+    return loader.load(key);
```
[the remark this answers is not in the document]
gdatahub: batchGetAspectValues (ElasticSearchTimeseriesAspectService.java:635-649) failed URNs come back as empty lists, so the loader can't tell them from "no docs". 
This PR just routes metrics into that path; single-URN and limit > 100 still throw, and buckets are unaffected.

PR #18271, opened 2026-07-08
feat(i18n): make the default UI locale configurable via I18N_DEFAULT_LOCALE

max-datahub:
## Summary

Makes the DataHub UI's **default language configurable** via a new `I18N_DEFAULT_LOCALE` environment variable. Previously the default was hardcoded to English (`en`) in the frontend. Operators can now set the organization-wide default locale for users who haven't chosen their own language.

This mirrors the existing `I18N_ENABLED` flow: a GMS env var → `featureFlags` config → `appConfig` GraphQL query → frontend.

## Behavior

- **Unset** → defaults to English (`en`), unchanged from today.
- **Set to a supported locale** (e.g. `de`, `fr`, `it`) → users who haven't picked a language in **Settings → Preferences** see the UI in that locale. A user's own language choice always takes precedence.
- **Set to an invalid/unsupported value** → GMS logs a warning at config resolution, and the UI falls back to English.
- Only takes effect when i18n is enabled (`I18N_ENABLED=true`).
- i18next's missing-key fallback stays English, so any untranslated string still renders in English rather than showing a gap.

## Changes

**Backend (GMS):**
- `application.yaml` — `featureFlags.i18nDefaultLocale: ${I18N_DEFAULT_LOCALE:en}`
- `FeatureFlags.java` — new `i18nDefaultLocale` field (default `"en"`)
- `app.graphql` — `i18nDefaultLocale: String!` on `FeatureFlagsConfig`
- `AppConfigResolver` — exposes the value; `resolveI18nDefaultLocale(...)` logs a warning when the configured value isn't a recognized UI locale (validated against a documented `SUPPORTED_UI_LOCALES` set). The raw value is passed through; the web app performs the authoritative validation, so a stale backend set can only affect the warning, never override a valid locale.
- `PropertiesCollectorConfigurationTest` — classified the new property as non-sensitive

**Frontend:**
- `app.graphql` query + `appConfigContext` default — surface/consume the new field
- `useDefaultLanguage` — reads the configured locale and validates it with `isSupportedLanguage`, falling back to English otherwise (authoritative client-side guard)
- `useDefaultLanguage.test.ts` — supported value passes through; unsupported/unset → English

## Notes

- Two-layer validation by design: the server-side warning is for operator visibility (where the env var is set); the frontend owns the authoritative `SupportedLanguage` list and the final fallback. The backend `SUPPORTED_UI_LOCALES` set (used only for the warning — GMS can't import the frontend TS source of truth at runtime) is protected by a **drift-guard unit test** (`AppConfigResolverTest#testSupportedUiLocalesMatchFrontendSupportedLanguage`) that parses the frontend `SupportedLanguage` union and fails if the two diverge, mirroring the existing `PropertiesCollectorConfigurationTest` guard pattern.
- No metadata-model / PDL changes; the per-user locale mechanism is untouched.

## Verification

| Gate | Result |
| --- | --- |
| `datahub-graphql-core` + config module compile (SDL codegen picks up the new field) | ✅ BUILD SUCCESSFUL |
| `PropertiesCollectorConfigurationTest` | ✅ passed |
| `AppConfigResolverTest` (incl. new locale drift guard) | ✅ 36/36, 0 failures |
| Frontend `yarn generate` (types include `i18nDefaultLocale`) | ✅ |
| `yarn test src/app/i18n` | ✅ 26/26 (incl. new `useDefaultLanguage` tests) |
| `yarn type-check` | ✅ clean |
| eslint (changed files) | ✅ 0 errors |

### DIFF
--- a/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/config/AppConfigResolver.java
+++ b/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/config/AppConfigResolver.java
@@ -497,12 +497,14 @@ public class AppConfigResolver implements DataFetcher<CompletableFuture<AppConfi
     return isEnabledInConfig && _isS3Enabled;
   }
 
-  // Supported UI locales -- keep in sync with SupportedLanguage in
-  // datahub-web-react/src/app/i18n/types.ts. Used only to warn operators about an
-  // unrecognized I18N_DEFAULT_LOCALE value; the web app performs the authoritative
-  // validation and falls back to English, so a stale entry here only affects the warning.
-  private static final Set<String> SUPPORTED_UI_LOCALES =
-      Set.of("en", "de", "es", "pt-BR", "fr", "it");
+  // The set of UI locales the frontend can render, used ONLY to warn operators when
+  // I18N_DEFAULT_LOCALE is set to an unrecognized value. The web app performs the authoritative
+  // validation and fallback, so a stale entry here can only affect the warning, never override a
+  // valid locale. This MUST mirror the frontend's SupportedLanguage union in
+  // datahub-web-react/src/app/i18n/types.ts (the source of truth);
+  // AppConfigResolverTest#testSupportedUiLocalesMatchFrontendSupportedLanguage fails on drift.
+  // Package-private so that guard test can read it without reflection.
+  static final Set<String> SUPPORTED_UI_LOCALES = Set.of("en", "de", "es", "pt-BR", "fr", "it");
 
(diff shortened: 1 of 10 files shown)


max-datahub:
Closing in favor of a simpler approach. Rather than introducing an `I18N_DEFAULT_LOCALE` operator variable, the UI should honor the user's browser/OS language — that is already an explicit language preference, so respecting it beats asking operators to configure a separate default. Replacement is a small frontend-only change (browser-language detection with fallback to English); a new PR will follow. The per-user Settings → Preferences choice continues to take precedence.

PR #18272, opened 2026-07-08
docs: note SQLite profiling limitation (max_overflow/NullPool)

bogacsmz:
## What

Adds a one-paragraph caveat to the SQLAlchemy source docs: ingesting a **SQLite** database with
`profiling.enabled: true` fails, and points users to the working recipe (`profiling.enabled: false`).

## Why

A real rough edge. On `acryl-datahub` 1.6, profiling a SQLite source errors:

```
TypeError: Invalid argument(s) 'max_overflow' sent to create_engine(), using configuration
SQLiteDialect_pysqlite/NullPool/Engine.
```

Root cause: the profiler requests a sized connection pool (passes `max_overflow`) for parallel column
profiling, but SQLite uses `NullPool`, which rejects pool-sizing arguments. Schema and lineage ingest
fine; only profiling fails. A one-line note saves users a failed run and a confusing stack trace.

## Reproducer

```yaml
source:
  type: sqlalchemy
  config:
    platform: sqlite
    connect_uri: "sqlite:///data.db"
    profiling: { enabled: true } # errors; set false to ingest cleanly
sink: { type: datahub-rest, config: { server: "http://localhost:8080" } }
```

## Change

A single blockquote added after the "This plugin extracts" list. No code changes.

---

Found while building [Legibright](https://github.com/bogacsmz/legibright-trust-audit) for the Build
with DataHub hackathon — we ingest public SQLite demo data and hit this firsthand.

### DIFF
--- a/metadata-ingestion/docs/sources/sqlalchemy/sqlalchemy_pre.md
+++ b/metadata-ingestion/docs/sources/sqlalchemy/sqlalchemy_pre.md
@@ -11,6 +11,8 @@ This plugin extracts the following:
 - Column types associated with each table
 - Table, row, and column statistics via optional SQL profiling.
 
+> **SQLite + profiling.** SQLite uses SQLAlchemy's `NullPool`, which does not accept connection-pool sizing arguments. Because the profiler passes `max_overflow`, ingesting a SQLite database with `profiling.enabled: true` fails with `TypeError: Invalid argument(s) 'max_overflow' sent to create_engine(...)`. Ingest SQLite with `profiling.enabled: false` — schema and lineage are unaffected.
+
 ### Prerequisites
 
 Before running ingestion, ensure network connectivity to the source, valid authentication credentials, and read permissions for metadata APIs required by this module.


sgomezvillamor:
Thanks for raising this — the diagnosis is right. I reproduced it: `_add_default_options` injects `max_overflow` before the first `create_engine`, so a SQLite recipe with `profiling.enabled: true` dies in `get_inspectors()` and produces 0 events. In-memory URLs hit the same thing via `SingletonThreadPool`. One correction: schema and lineage don't ingest either — the run fails before any metadata is extracted.

That's why we'd rather fix this than document it. #18803 and #18996 both resolve the dialect's pool class and only inject `max_overflow` when it's a `QueuePool`, falling back to today's behavior if resolution fails. I patched that locally and the same recipe went from 0 events to a full `datasetProfile` — row/column counts, null/unique counts, min/max/mean/median, sample values. Only `stdev` is missing, and that already degrades to a warning.

#18803 is actively under review, so we'd prefer to land one of those two rather than merge a note about a ~6-line bug. Thanks for the clear reproducer — it made the root cause easy to confirm.

---

PR #18281, opened 2026-07-09
fix(ingest/client): Updates scroll_entities behavior when aspects=None

asikowitz:
Clarifies and fixes the aspects argument semantics in the OpenAPI client's scroll_entities:
  - aspects=None → returns urns only (no aspects fetched)
  - aspects=[] → returns all aspects
  - aspects=[...] → returns only the requested aspects

Previously, urn-only entities were silently dropped during deserialization when `aspects == None`. Adds smoke tests for this use case.

### DIFF
--- a/smoke-test/tests/openapi/v3/test_scroll_entities.py
+++ b/smoke-test/tests/openapi/v3/test_scroll_entities.py
@@ -1,6 +1,5 @@
 """Smoke tests for the scroll_entities DataHubGraph method."""
 
-import logging
 from typing import cast
 
 import pytest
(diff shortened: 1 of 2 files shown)


sgomezvillamor (approved):

smoke-test/tests/openapi/v3/test_scroll_entities.py
```diff
@@ -58,6 +58,66 @@ def test_scroll_entities_basic(graph_client: DataHubGraph) -> None:
...
+        count=10,
+    )
+    assert set(result.entities.keys()) == ALL_DATASET_URNS
+    for urn, aspects in result.entities.items():
+        assert set(aspects.keys()) == {"datasetProperties"}, (
+            f"Expected only datasetProperties for {urn}, got: {set(aspects.keys())}"
+        )
+    logger.info(
+        "scroll_entities aspects=['datasetProperties']: only that aspect returned"
+    )
```
skrydal: Typically we do not print success logs in tests, if an assertion fails, it will print whatever is put after `,` which should be precise yet not cluttering the output.

skrydal (approved):
Looks reasonable, please check my comment on tests.

PR #18284, opened 2026-07-09
ci(playwright): report individual test failures to PostHog for flake detection

priyabratadas-dh:
## Summary

- Extends existing CI test failure reporting infrastructure to cover Playwright E2E tests
- Adds content-based detection to distinguish Playwright JUnit XML from Pytest JUnit XML (both named `junit*.xml`) by peeking at `testsuite.name` for `.spec.ts` patterns
- Adds `parse_playwright_failures()` that formats test identifiers as `spec/file.spec.ts > Describe Block > test name`
- Adds a "Send failed test metrics to PostHog" step to the reusable Playwright workflow, firing on `failure()` with `run_attempt` wired in

## How flake detection works

A test is considered flaky if it appears in a `ci-test-failure` event with `run_attempt=1` but not `run_attempt=2` for the same `run_id`. No pass events needed — the existing failure-only approach is sufficient for this definition.

## Test plan

- [ ] Verify `detect_xml_type()` correctly identifies Playwright JUnit XML (`.spec.ts` in `testsuite.name`) vs Pytest JUnit XML (dotted classnames)
- [ ] Verify `parse_playwright_failures()` produces correctly formatted identifiers
- [ ] Verify PostHog step fires on Playwright test failure in CI with correct `run_attempt`, `shard`, and `shard_count` fields

### DIFF
--- a/.github/scripts/send_failed_tests_to_posthog.py
+++ b/.github/scripts/send_failed_tests_to_posthog.py
@@ -36,21 +36,42 @@ class PostHogConfig:
     timeout: int = 10
 
 
+def _detect_junit_subtype(xml_file: Path) -> str:
+    """
+    Distinguish between Playwright and Pytest JUnit XML by content.
+
+    Playwright's JUnit reporter sets testsuite.name to the spec file path
+    (e.g. "tests/domains/domains.spec.ts"), which is reliable for detection.
+    """
+    try:
+        tree = ET.parse(xml_file)
+        root = tree.getroot()
+        for ts in root.iter('testsuite'):
+            name = ts.get('name', '')
+            if '.spec.ts' in name or '.spec.js' in name or '.spec.tsx' in name:
+                return 'playwright'
+        return 'pytest'
+    except Exception:
+        return 'pytest'
+
+
 def detect_xml_type(xml_file: Path) -> str:
     """
     Detect the type of JUnit XML file.
 
-    Returns: 'cypress', 'pytest', 'java', or 'unknown'
+    Returns: 'cypress', 'playwright', 'pytest', 'java', or 'unknown'
     """
     filename = xml_file.name
 
     # Check filename patterns
     if filename.startswith('cypress-test-'):
         return 'cypress'
-    elif filename.startswith('junit') and filename.endswith('.xml'):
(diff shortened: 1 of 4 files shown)


.github/scripts/send_failed_tests_to_posthog.py
```diff
@@ -112,6 +133,56 @@ def parse_cypress_failures(xml_file: Path) -> List[FailedTest]:
...
+                error = testcase.find('error')
+
+                if failure is None and error is None:
+                    continue
+
+                classname = testcase.get('classname', '')
+                test_name = testcase.get('name', '')
+
+                parts = [p for p in [spec_file, classname, test_name] if p]
+                test_identifier = ' > '.join(parts) if parts else spec_file
```
devashish2203: nit: The name can become quite long in some cases, should we limit this to say 200 characters?

.github/workflows/resusable-playwright-tests.yml
```diff
@@ -175,6 +175,38 @@ jobs:
           path: e2e-test/ui/playwright/test-results/junit.xml
           retention-days: 5
 
+      - name: Send failed test metrics to PostHog
```
devashish2203, line 183: Do we want to send the results to posthog for workflow_dispatch runs? 
People may use this to rerun tests for stability and it'll corrupt our metrics.

Add an input (boolean) for sending the metrics, set it to true only when being called from the docker-unified.yml workflow.

devashish2203 (changes requested):

.github/scripts/send_failed_tests_to_posthog.py
priyabratadas-dh: Done — truncated to 200 chars.

.github/workflows/resusable-playwright-tests.yml
priyabratadas-dh: Added a `send_posthog_metrics` boolean input (default `false`) to the reusable workflow, and pass `send_posthog_metrics: true` only from `docker-unified.yml`. `workflow_dispatch` and standalone runs won't send metrics.

devashish2203 (approved):

PR #18286, opened 2026-07-09
chore(hooks): add pre-push large-push guard and no-verify deny hook

devashish2203:
## Summary

Adds two local git-hook guardrails plus the wiring to install them.

**Pre-push large-push guard** (`.githooks/pre-push`) — rejects any push that changes more than 500 files, to catch accidental massive pushes (vendored trees, generated output, `node_modules`, and the like). It counts unique files across every ref being pushed: a two-point diff for existing branches, and commits-not-on-any-remote for new branches. It fails open on git errors, since it's an accident guard and not a security control. A deliberate large push can still go through with `git push --no-verify`.

**no-verify deny hook** (`.claude/hooks/deny-git-no-verify.sh`) — a Claude Code `PreToolUse` hook that stops coding agents from running `git … --no-verify`, so a failing pre-commit/pre-push hook gets surfaced and fixed instead of skipped. It only fires when the command both invokes git and passes the flag. Wired up in `.claude/settings.json`.

NOTE:
The checks above are not comprehensive and there are ways to bypass the git push hool. Not doing exhaustive checks like those done in https://github.com/tupe12334/block-no-verify as this relies on the node runtime of the user and is an additional dependency.

Supporting changes:

- `settings.gradle` — installs the pre-push hook into the shared `.git/hooks` dir on build, mirroring the existing pre-commit install (skips CI and worktrees).
- `.gitignore` — un-ignore `.claude/hooks/deny-git-no-verify.sh` so it can be committed (the `.claude/*` rule was hiding it).
- `AGENTS.md` — documents the "never bypass hooks with `--no-verify`" policy that both hooks enforce.

## Testing

- Pre-push hook: exercised end-to-end against a throwaway repo — allows small pushes, allows a new branch counting only its own new files, rejects >1000-file updates and new branches, and confirms `--no-verify` bypasses.
- Deny hook: confirmed it blocks `git commit`/`git push --no-verify` (including with a quoted `-m` message before the flag) and allows normal git commands and non-git commands that merely mention the flag.

## Checklist

- [ ] Tests added/updated (hooks verified manually; no automated tests added)
- [x] Docs added/updated (`AGENTS.md`)
- [ ] Breaking change entry (n/a — local dev tooling, no runtime impact)


devashish2203:
### Known limitation: file counting on the new-branch path

Documenting a known edge for future reference. It only affects the **new-branch** branch of the hook (the `remote_oid` is all-zeros case), which counts files with:

```bash
git log --name-only --pretty=format: "$local_oid" --not --remotes --
```

`--not --remotes` = "commits reachable from the tip but not from any remote-tracking ref (`refs/remotes/*`)", i.e. what's new. Two ways that count is off:

**Over-counting — first push to an empty/fresh remote.** `--remotes` only sees remote-tracking refs that already exist locally (from a prior clone/fetch), and the hook runs before the push. In the normal cloned-repo flow this is fine — a new feature branch correctly excludes everything already on `origin/master`. But on the *very first* push to a brand-new remote (`git init` → `git remote add` → `git push -u`), there are no `refs/remotes/*` yet, so `--not --remotes` excludes nothing, `git log` walks the entire history to the root, and `sort -u` collects every file ever touched. On a repo this size that trips the limit and rejects a legitimate first push.

**Under-counting — merge commits.** `git log --name-only` prints no file list for merge commits by default (a merge has multiple parents, so its diff is omitted unless you pass `-m`/`-c`/`--cc`). Files changed only at a merge — e.g. conflict resolution / evil merges — aren't counted.

**Related:** the update path (existing branch) uses `git diff --name-only <old> <new>`, a net two-point diff, while the new-branch path uses `git log`, which counts a file touched in any included commit. So the same changes can count differently depending on whether the branch is new.

For an accident guard this is acceptable — both are edge cases — and the fail-open `|| true` only covers git *errors*, not these (they're valid results). The cleaner fix, if we want it later, is to diff the new branch against a merge-base (e.g. against `origin/HEAD`) instead of using `git log --not --remotes`: that handles merges as a net diff and avoids the whole-history blowup, at the cost of computing the right base across remotes.

kewats (approved):

PR #18287, opened 2026-07-09
fix(ingest/tableau): preserve full project hierarchy under project_pattern filtering

aviraj-gour:
## Summary

When a Tableau project matched `project_pattern` but its **ancestor** projects did not, the ingested project lost its full folder path in DataHub. Only the immediate parent survived, and it was re-rooted at the top level, so the browse tree did not reflect Tableau's actual folder structure.

Example — ingesting only a deeply nested project whose ancestors are filtered out:

```
Site
└── Project_1        (filtered out)
    └── Project_2    (filtered out)
        └── Project_3    (filtered out)
            └── Project_4    (matches filter)
```

Previously `Project_4` appeared nested only under a re-rooted `Project_3` at the top level; `Project_1`/`Project_2` were dropped.

## What changed

`emit_project_containers` now walks the full ancestor chain from the unfiltered project map and emits filtered-out ancestors as **path-only containers**:

- Ancestors are resolved from the full project map (returned by `_populate_projects_registry`) rather than only the filtered content registry, so the recursion reaches the true root.
- These ancestor containers carry **no content** — they never enter the content registry, so `project_pattern` still fully controls what is ingested.
- Path reconstruction is independent of the `extract_project_hierarchy` flag (that flag governs which content is ingested, not path correctness).
- Unreachable ancestors (e.g. insufficient permissions) degrade gracefully rather than truncating/erroring.

The full project map is threaded as an explicit argument instead of being stored on the source instance, so it is released as soon as emission completes.

## Testing

Added `TestProjectContainerHierarchy` — end-to-end tests that drive the real filtering pipeline (mocked projects fetch → `project_pattern` filtering → registry build → container emission) and assert the resulting browse tree for:

- Full reconstruction of a linear chain of unmatched ancestors
- Shared ancestors emitted exactly once across multiple matched leaves
- A matched middle project linking to both its reconstructed parent and its child
- A denied ancestor still emitted as a path-only container (path stays unbroken)
- A denied sibling excluded entirely while its sibling emits
- Path correctness preserved when `extract_project_hierarchy` is disabled

`./gradlew :metadata-ingestion:lint` and the full `tests/unit/tableau/` suite pass.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/tableau/tableau.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/tableau/tableau.py
@@ -4245,25 +4245,13 @@ class TableauSiteSource:
                 # Go to the parent project as we need to generate container first for parent
                 parent_project_key = self.gen_project_key(project_.parent_id)
 
-                # Fall back to all_project_map when the parent was filtered out of the
-                # registry, so we recurse to the root. Such ancestors become path-only
-                # containers (no content, since they're absent from the content registry).
-                parent_tableau_project: Optional[TableauProject] = (
-                    self.tableau_project_registry.get(project_.parent_id)
-                    or all_project_map.get(project_.parent_id)
-                )
-
-                if parent_tableau_project is None:
-                    # Parent unreachable (e.g. permissions); _get_all_project already
-                    # warned. Emit a single-level container and stop the upward walk.
-                    parent_tableau_project = TableauProject(
-                        id=project_.parent_id,
-                        name=project_.parent_name or project_.parent_id,
-                        description=None,
-                        parent_id=None,
-                        parent_name=None,
-                        path=[],
-                    )
+                # Resolve from the full project map, not tableau_project_registry, so
+                # we recurse to the root even through filtered-out ancestors (emitted as
+                # path-only containers). _get_all_project nulls out any parent_id absent
+                # from the map, so a non-None parent_id always resolves here.
+                parent_tableau_project: TableauProject = all_project_map[
+                    project_.parent_id
+                ]
 
                 yield from emit_project_in_topological_order(parent_tableau_project)
 
--- a/metadata-ingestion/src/datahub/ingestion/source/tableau/tableau.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/tableau/tableau.py
@@ -4245,13 +4245,17 @@ class TableauSiteSource:
                 # Go to the parent project as we need to generate container first for parent
                 parent_project_key = self.gen_project_key(project_.parent_id)
 
-                # Resolve from the full project map, not tableau_project_registry, so
-                # we recurse to the root even through filtered-out ancestors (emitted as
-                # path-only containers). _get_all_project nulls out any parent_id absent
-                # from the map, so a non-None parent_id always resolves here.
-                parent_tableau_project: TableauProject = all_project_map[
-                    project_.parent_id
-                ]
+                # Resolve from the full project map so we recurse to the root even
+                # through filtered-out ancestors. _get_all_project nulls out any
+                # parent_id absent from the map, so this always resolves; guard it so a
+                # future regression fails with context, not a bare KeyError.
+                parent_tableau_project = all_project_map.get(project_.parent_id)
+                if parent_tableau_project is None:
+                    raise ValueError(
+                        f"parent_id {project_.parent_id!r} of project "
+                        f"{project_.name!r} ({project_.id!r}) is missing from the "
+                        f"project map; expected _get_all_project to have nulled it out."
+                    )
 
                 yield from emit_project_in_topological_order(parent_tableau_project)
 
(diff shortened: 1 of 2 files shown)


metadata-ingestion/src/datahub/ingestion/source/tableau/tableau.py
```diff
@@ -4241,19 +4245,20 @@ def emit_project_in_topological_order(
...
+                    or all_project_map.get(project_.parent_id)
                 )
 
-                if (
-                    parent_tableau_project is None
-                ):  # It is not in project registry because of project_pattern
-                    assert project_.parent_name, (
-                        f"project {project_.name} should not be null"
-                    )
+                if parent_tableau_project is None:
```
treff7es, line 4253: This fallback branch appears to be unreachable, and the comment is a bit misleading — this is also the source of the Codecov "3 lines uncovered" note.

`_get_all_project` already handles the truly-unreachable-parent case: when a `parent_id` is not in `all_project_map` it emits the `"Incomplete project hierarchy"` warning **and sets `project.parent_id = None`** (lines ~1319-1324). Its post-condition assert (~1327-1336) then guarantees every non-`None` `parent_id` exists in `all_project_map`. Since `tableau_project_registry` holds those same objects, `all_project_map.get(project_.parent_id)` on line 4253 can never miss when `parent_id is not None` — so `parent_tableau_project` is never `None` here. (Same reason the old `assert project_.parent_name` never fired, and `name=project_.parent_name or project_.parent_id` on line 4261 is likewise dead.)

The comment says this is a live safety net for the permissions case, but that's exactly the case `_get_all_project` neutralizes upstream by nulling `parent_id` (so control routes to the root branch, never here).

Suggest picking one:
- **Simplify (preferred):** replace the block with `assert parent_tableau_project is not None` documenting the upstream invariant, or drop it.
- **Keep as defense-in-depth:** reword honestly (normally unreachable; guards against the invariant drifting or `-O` stripping the assert) **and** add a `self.report.warning(...)` + counter, so it isn't a silent path-only emission if it ever does fire.

metadata-ingestion/src/datahub/ingestion/source/tableau/tableau.py
```diff
@@ -4241,19 +4245,20 @@ def emit_project_in_topological_order(
                 # Go to the parent project as we need to generate container first for parent
                 parent_project_key = self.gen_project_key(project_.parent_id)
 
+                # Fall back to all_project_map when the parent was filtered out of the
+                # registry, so we recurse to the root. Such ancestors become path-only
+                # containers (no content, since they're absent from the content registry).
```
treff7es: Nit: "content registry" is a coined term with no matching identifier — the field is `tableau_project_registry`. Using the real name here saves a future reader from hunting for a "content registry" that doesn't exist.

metadata-ingestion/tests/unit/tableau/test_tableau_source.py
```diff
@@ -1315,6 +1317,270 @@ def _make_site_source() -> TableauSiteSource:
         )
 
 
+def _tsc_project(id: str, name: str, parent_id: Optional[str]) -> mock.MagicMock:
```
treff7es: Two trivial nits (non-blocking):
- `id` shadows the builtin (ruff A002) — `project_id` avoids it.
- The `List[Tuple[str, str, Optional[str]]]` project fixtures rely on positional `(id, name, parent_id)`. A tiny `class ProjectSpec(NamedTuple)` would make call sites self-documenting (`ProjectSpec("p2", "Project_2", "p1")`) and matches the repo's "dataclasses/NamedTuple over tuples" convention.

treff7es (approved):

metadata-ingestion/tests/unit/tableau/test_tableau_source.py
aviraj-gour: Fixed using ProjectSpec now.

metadata-ingestion/src/datahub/ingestion/source/tableau/tableau.py
aviraj-gour: Thanks for catching this, fixed.
Now using  `tableau_project_registry`

metadata-ingestion/src/datahub/ingestion/source/tableau/tableau.py
aviraj-gour: Agreed, dead code. Simplified it to a direct all_project_map[project_.parent_id] — dropped the Optional, the assert, and the fallback. The invariant always holds (_get_all_project nulls any dangling parent_id), so this never misses.

Added a regression test for the missing-ancestor (permissions) case so that invariant stays covered if someone touches the nulling later.

aviraj-gour:
Thanks @treff7es for a review.

metadata-ingestion/src/datahub/ingestion/source/tableau/tableau.py
aviraj-gour: Slight change raises a Value Error if key is not there
```
# Resolve from the full project map so we recurse to the root even
# through filtered-out ancestors. _get_all_project nulls out any
# parent_id absent from the map, so this always resolves; guard it so a
# future regression fails with context, not a bare KeyError.
parent_tableau_project = all_project_map.get(project_.parent_id)
if parent_tableau_project is None:
    raise ValueError(
        f"parent_id {project_.parent_id!r} of project "
        f"{project_.name!r} ({project_.id!r}) is missing from the "
        f"project map; expected _get_all_project to have nulled it out."
    )
```

PR #18289, opened 2026-07-09
fix(ingest/datalake): bound single-object JSON reads and honor max_rows for .json

acrylJonny:
## Summary

Two related fixes in the data-lake JSON schema inferrer (affects S3, GCS, and ABS sources):

- **`.json` ignored `max_rows`.** The `.json` / `application/json` inferrer was constructed without `max_rows` in both the S3 `_get_inferrer` and the ABS inline selector, so a user-tuned `max_rows` was silently ignored for `.json` files (only `.jsonl` honored it). Now all JSON branches pass `source_config.max_rows`.
- **Single-object JSON was fully materialized.** The non-array fallback did `ujson.load(file)`, loading the entire file into memory — an OOM risk for a large single object such as `{"data": [ ...millions... ]}`. Replaced with a streaming `ijson` reader (`_bounded_json_value`) that truncates every array to `max_rows` while preserving structure and native (non-`Decimal`) types.

## Behavior change: single-object `.json` files are now sampled

A `.json` file holding a single object used to be read end to end, so `construct_schema` saw every record and `is_field_nullable` inspected every array element. Both are now based on the first `max_rows` elements of each array (default **100**). For `{"rows": [ ...5000 records... ]}`, a field that first appears in record 3000 is no longer reported, and nullability is decided from the first 100 records.

This is the trade for not OOMing on large files. Top-level JSON arrays and `.jsonl` files were already sampled this way, so only single-object `.json` files change. Users who need the old exhaustive behavior can raise `max_rows`.

Objects are deliberately **not** bounded — only arrays are. Dropping object keys would drop the fields they describe, so a column-oriented dump with more than `max_rows` columns would silently lose columns. A very wide map is therefore still rebuilt in full; the docstring says so explicitly.

`max_rows` also gained `ge=1` validation on the S3, GCS, and ABS configs. At `max_rows=0` the array limit fired immediately and every array became `[]`.

## Testing

- `test_bounded_json_value_truncates_arrays` — arrays truncated, structure and native `int`/`float` preserved.
- `test_bounded_json_value_truncates_array_of_objects` and `test_bounded_json_value_skips_nested_containers_past_limit` — the skip path unwinds to the right depth through nested containers, verified with a trailing sibling key.
- `test_infer_schema_json_single_object_is_sampled` — a field that first appears past `max_rows` is not reported. This fails against the old `ujson.load` fallback, so it is an actual regression guard.
- All 13 tests in `tests/unit/data_lake/test_schema_inference.py` pass.

## Checklist

- [x] Tests added/updated
- [x] Docs added/updated — `max_rows` descriptions on the S3/GCS/ABS configs, plus a note in `docs/how/updating-datahub.md`
- [ ] Breaking changes (n/a — behavior change documented above and in `updating-datahub.md`)


metadata-ingestion/src/datahub/ingestion/source/schema_inference/json.py
```diff
@@ -34,6 +34,65 @@
 logger = logging.getLogger(__name__)
 
 
+def _bounded_json_value(file: IO[bytes], max_rows: int) -> Any:
+    """Stream-rebuild a single JSON value, truncating each array to ``max_rows``
+    elements so a huge single object (e.g. ``{"data": [ ...millions... ]}``) is
+    not fully loaded into memory just to infer its schema.
```
alokr-dhub: Arrays are bounded now, but maps still aren't — so this promise is shape-specific.

It holds for the `{"data": [...]}` case you had in mind. It doesn't hold for a wide map like `{"user_1": {...}, "user_2": {...}, ...}`, which is a pretty common way people dump JSON into a data lake. That shape still materializes completely here, and then it costs you a second time downstream: `construct_schema` emits a `SchemaField` per key, so a million-key object becomes a million fields.

Whether bounding map keys belongs in this PR is your call. If it doesn't, I'd narrow the docstring to promise only what the code actually delivers.

metadata-ingestion/src/datahub/ingestion/source/schema_inference/json.py
```diff
@@ -34,6 +34,65 @@
...
+
+    ponytail: array items beyond ``max_rows`` are dropped, so the inferred schema
+    reflects only the first ``max_rows`` items of each array.
```
alokr-dhub: The `ponytail:` prefix is a convention from another tool and won't mean anything to someone reading this file cold. Worth making it a plain comment.

The content is useful though — please keep the substance, since this sampling caveat is the part that isn't documented anywhere else (see my note on the `_bounded_json_value` call site).

metadata-ingestion/src/datahub/ingestion/source/schema_inference/json.py
```diff
@@ -34,6 +34,65 @@
...
+    """
+    root: List[object] = []
+    stack: List[Tuple[object, str]] = []  # (container, "map" | "array")
```
alokr-dhub: This is what's forcing the three `# type: ignore`s (here at L51, plus L59 and L61) — `object` as the container type means mypy can't tell that the `kind == "map"` branch implies a dict.

`Literal["map", "array"]` for the tag and `Union[dict, list]` for the container would let mypy verify the branching. For a hand-rolled parser this fiddly, that's the difference between the type checker catching a future edit and not.

metadata-ingestion/src/datahub/ingestion/source/schema_inference/json.py
```diff
@@ -34,6 +34,65 @@
...
+            root.append(value)
+            return
+        container, kind = stack[-1]
+        if kind == "map":
+            container[keys[-1]] = value  # type: ignore[index]
+        else:
+            container.append(value)  # type: ignore[attr-defined]
+
+    # use_float keeps numbers as native int/float (not Decimal), matching json.load.
+    for _prefix, event, value in ijson.parse(file, use_float=True):
```
alokr-dhub, line 74: `use_float` landed in ijson 3.2.0, but `setup.py:78` pins `"ijson<4.0.0"` with no lower bound.

Pip will almost certainly resolve to something recent, so this won't bite in normal use — but an older lockfile or a constrained resolve turns it into a parse-time crash rather than a graceful degradation. `ijson>=3.2.0,<4.0.0` closes it.

metadata-ingestion/src/datahub/ingestion/source/schema_inference/json.py
```diff
@@ -34,6 +34,65 @@
...
+            stack.pop()
+            keys.pop()
+        elif event == "end_array":
+            stack.pop()
+        else:  # scalar
+            if _at_array_limit():
+                continue
+            _attach(value)
+
+    return root[0] if root else {}
```
alokr-dhub: The `{}` branch looks unreachable — for any valid document ijson emits at least one event, so `root` always ends up with exactly one element.

If it somehow were reached, returning `{}` would quietly turn a parse anomaly into an empty schema, instead of raising and letting the caller's `except` fall through to the jsonlines path. I'd rather this fail loudly than invent an empty result.

metadata-ingestion/src/datahub/ingestion/source/schema_inference/json.py
```diff
@@ -60,7 +119,7 @@ def infer_schema(self, file: IO[bytes]) -> List[SchemaField]:
                 if not datastore:
                     # Not a top-level array — likely a single JSON object.
                     file.seek(0)
-                    datastore = [ujson.load(file)]
+                    datastore = [_bounded_json_value(file, self.max_rows)]
```
alokr-dhub: This changes inferred schemas, not just memory use — and I don't think it's called out anywhere user-visible yet.

`construct_schema` recurses into every dict inside a list, and `is_field_nullable` inspects *all* array elements to decide nullability. Before this change a single-object `.json` was read end to end, so both were exact. Now they're based on the first `max_rows` elements, and the default is only 100.

Concretely: for `{"rows": [ ...5000 records... ]}`, a field that first appears in record 3000 silently stops being reported, and nullability flips to whatever the first 100 records imply.

That's a reasonable trade against OOMing, but it deserves a line in the PR description and probably a release note — right now users would have to discover it.

metadata-ingestion/src/datahub/ingestion/source/s3/source.py
```diff
@@ -387,7 +387,7 @@ def _get_inferrer(
                 max_rows=self.source_config.max_rows, format="jsonl"
             )
         elif extension == ".json":
-            return json.JsonInferrer()
+            return json.JsonInferrer(max_rows=self.source_config.max_rows)
```
alokr-dhub, line 391: Now that `max_rows` governs JSON, its user-facing description is stale — `s3/config.py:88` and `abs/config.py:75` both still say *"Maximum number of rows to use when inferring schemas for TSV and CSV files."* Those strings render into the published connector docs, so the "Docs added/updated — n/a" checkbox needs unticking.

Separate minor thing while you're in there: `max_rows` has no `ge=1` validation. With `max_rows=0`, `_at_array_limit()` fires immediately and every array silently becomes `[]`. That's pre-existing, but this PR extends its reach from CSV row counts to JSON structure.

metadata-ingestion/tests/unit/data_lake/test_schema_inference.py
```diff
@@ -130,6 +130,61 @@ def test_infer_schema_json_single_object():
...
+                }
+            ).encode("utf-8")
+        )
+        file.seek(0)
+
+        fields = json.JsonInferrer(max_rows=10).infer_schema(file)
+
+    field_paths = [f.fieldPath for f in fields]
+    assert "integer_field" in field_paths
+    assert "big_array" in field_paths
```
alokr-dhub: These two assertions pass just as happily against the old `ujson.load(file)` implementation — nothing here observes boundedness. Revert the source change and this test stays green, so it isn't the regression guard the docstring above claims it is.

`test_bounded_json_value_truncates_arrays` is doing the real work. Either make this one assert something the old code would fail (spy on `_bounded_json_value`, or check the parsed structure directly), or soften the docstring so it isn't promising a guard it can't provide.

metadata-ingestion/tests/unit/data_lake/test_schema_inference.py
```diff
@@ -130,6 +130,61 @@ def test_infer_schema_json_single_object():
...
+def test_bounded_json_value_truncates_array_of_objects():
+    """Array elements beyond max_rows that are themselves objects are skipped
+    without being materialized (exercises the nested-container skip path)."""
+    with tempfile.TemporaryFile(mode="w+b") as file:
+        file.write(ujson.dumps({"rows": [{"a": i} for i in range(50)]}).encode("utf-8"))
+        file.seek(0)
+
+        result = json._bounded_json_value(file, max_rows=3)
+
+    assert [r["a"] for r in result["rows"]] == [0, 1, 2]
```
alokr-dhub: Nice — this covers the depth-1 skip. It only goes one level deep though, and `skip_depth` is written to handle arbitrary nesting.

A bad increment/decrement pairing on something like `{"rows": [{"a": {"b": [1, 2]}}, ...]}` past the cap wouldn't show up in the current tests. One deeper fixture would close that off cheaply.

alokr-dhub (commented):
Traced `_bounded_json_value` through nested containers, the skipped-subtree path, and the various root shapes (single object, top-level scalar, empty array) — the stack/keys bookkeeping holds up, and `use_float=True` correctly preserves the native `int`/`float` types `ujson.load` used to give us. The streaming rebuild is the right idea.

Also checked call-site coverage: only three non-test files reference `JsonInferrer` and all three are fixed here, and GCS comes along for free via `GCSSource.create_equivalent_s3_source`, so the "S3, GCS, and ABS" claim in the description holds.

Two things I'd want sorted before merge — the new regression test doesn't actually fail against the old implementation, and the schema-sampling behavior change isn't mentioned anywhere user-visible. Details inline, along with a few smaller notes you can take or leave.

metadata-ingestion/src/datahub/ingestion/source/schema_inference/json.py
acrylJonny: Narrowed the docstring rather than bounding maps. Dropping keys drops the fields they describe — a column-oriented dump (`{"col_1": {...}, ...}`) with more than `max_rows` columns would silently lose columns, which is worse than the memory it saves. The docstring now says only arrays are bounded and calls out the wide-map shape explicitly, including that it stays fully materialized.

metadata-ingestion/src/datahub/ingestion/source/schema_inference/json.py
acrylJonny: Dropped the prefix, kept the substance. It is now plain prose in the docstring, and the sampling caveat is also in the `max_rows` config description and in `updating-datahub.md`.

metadata-ingestion/src/datahub/ingestion/source/schema_inference/json.py
acrylJonny: Fixed by dropping the tag instead of typing it: the stack is now `List[Union[Dict[str, object], List[object]]]` and the branches test `isinstance(container, list)`. That removes the tuple, the tag, and all three `type: ignore`s, and mypy verifies the container handling. Pending map keys are `List[str]` with a placeholder rather than `Optional[str]`, so the dict index type checks too.

metadata-ingestion/src/datahub/ingestion/source/schema_inference/json.py
acrylJonny: Pinned `ijson>=3.2.0,<4.0.0` with a comment naming `use_float` as the reason, and regenerated `pyproject.toml` / `uv.lock` via `updateLockFile`.

metadata-ingestion/src/datahub/ingestion/source/schema_inference/json.py
acrylJonny: Now raises `ijson.common.JSONError("no JSON value found")`, which the existing `except` catches and falls through to the jsonlines reader. Also handled the adjacent case: a non-object root (bare scalar, empty array) is no longer handed to `construct_schema`, which cannot read fields off it and previously raised `AttributeError`.

metadata-ingestion/src/datahub/ingestion/source/schema_inference/json.py
acrylJonny: Documented in three places: a dedicated section in the PR description, an entry in `docs/how/updating-datahub.md` using your `{"rows": [ ...5000... ]}` example, and the `max_rows` description itself. Also added `ge=1` so `max_rows=0` can no longer truncate every array to empty.

metadata-ingestion/src/datahub/ingestion/source/s3/source.py
acrylJonny: Updated all three descriptions — `s3/config.py`, `abs/config.py`, and `gcs_source.py`, which I would have missed — and unticked the docs checkbox. `ge=1` added on all three as well.

metadata-ingestion/tests/unit/data_lake/test_schema_inference.py
acrylJonny: You are right, it was not a guard. Replaced with `test_infer_schema_json_single_object_is_sampled`: 200 records where `late_field` first appears at index 150, `max_rows=100`, asserting `rows.late_field` is absent. Verified it fails when the source is reverted to `ujson.load(file)` and passes after.

metadata-ingestion/tests/unit/data_lake/test_schema_inference.py
acrylJonny: Added `test_bounded_json_value_skips_nested_containers_past_limit` — `{"rows": [{"a": {"b": [i, i+1]}} x10], "tail": "sentinel"}` with `max_rows=2`. Asserting on the whole result plus the trailing `tail` key proves `skip_depth` unwound to exactly the top level rather than one frame off.

alokr-dhub (approved):
Approved.

PR #18290, opened 2026-07-09
fix(ingest/abs): list folders with a server-side delimiter walk

acrylJonny:
## Summary

ABS `list_folders` called `list_blobs(name_starts_with=prefix)` with no delimiter, enumerating **every descendant blob** under the prefix and de-duplicating the immediate child folders in Python — re-listing whole subtrees at each recursion level.

Switched to `container_client.walk_blobs(name_starts_with=..., delimiter="/")` so Azure returns only the immediate child prefixes server-side. The output contract is unchanged (folder paths without a trailing slash).

## Testing

- Updated `tests/unit/azure/test_abs_folder_utils.py` to mock `walk_blobs`; both tests pass and assert the delimiter walk is used and that no descendant enumeration (`list_blobs`) occurs.

## Checklist

- [x] Tests added/updated
- [ ] Breaking changes (n/a)

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/azure/abs_folder_utils.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/azure/abs_folder_utils.py
@@ -194,30 +194,15 @@ def list_folders(
 
     normalized_prefix = prefix.strip("/")
     prefix_with_separator = f"{normalized_prefix}/" if normalized_prefix else ""
-    blob_list = container_client.list_blobs(name_starts_with=prefix)
 
-    seen: set[str] = set()
-    for blob in blob_list:
-        blob_name = blob.name.strip("/")
-        if not blob_name:
+    # walk_blobs with a delimiter yields one BlobPrefix (name ends with "/") per
+    # immediate sub-folder; blobs at this level are skipped.
+    for item in container_client.walk_blobs(
+        name_starts_with=prefix_with_separator, delimiter="/"
+    ):
+        name = item.name
+        if not name.endswith("/"):
             continue
-
-        if prefix_with_separator:
-            if not blob_name.startswith(prefix_with_separator):
-                continue
-            relative_name = blob_name[len(prefix_with_separator) :]
-        else:
-            relative_name = blob_name
-
-        if "/" not in relative_name:
-            continue
-        child_folder = relative_name.split("/", 1)[0]
-        folder_name = (
-            f"{normalized_prefix}/{child_folder}" if normalized_prefix else child_folder
-        )
-
-        if folder_name in seen:
-            continue
-        seen.add(folder_name)
-
-        yield f"{folder_name}"
+        folder_name = name.strip("/")
+        if folder_name:
+            yield folder_name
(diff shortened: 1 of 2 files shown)


metadata-ingestion/src/datahub/ingestion/source/azure/abs_folder_utils.py
```diff
@@ -194,30 +194,15 @@ def list_folders(
...
     prefix_with_separator = f"{normalized_prefix}/" if normalized_prefix else ""
-    blob_list = container_client.list_blobs(name_starts_with=prefix)
 
-    seen: set[str] = set()
-    for blob in blob_list:
-        blob_name = blob.name.strip("/")
-        if not blob_name:
+    # walk_blobs with a delimiter yields one BlobPrefix (name ends with "/") per
+    # immediate sub-folder; blobs at this level are skipped.
+    for item in container_client.walk_blobs(
```
treff7es, line 200: Worth confirming this `name_starts_with` change is intentional (it is a correctness improvement, not just perf): the old code passed the raw `prefix` to `list_blobs`, which could over-match sibling prefixes (`"database"` also matches `"database2/..."`) and then filtered them out client-side with `startswith(prefix_with_separator)`. Passing `prefix_with_separator` (`"database/"`) here scopes it correctly at the source and drops the need for that filter. Good tightening.

metadata-ingestion/src/datahub/ingestion/source/azure/abs_folder_utils.py
```diff
@@ -194,30 +194,15 @@ def list_folders(
...
+        name_starts_with=prefix_with_separator, delimiter="/"
+    ):
+        name = item.name
```
treff7es, line 203: Minor behavior delta (almost certainly an improvement): some tools (ADLS Gen2 / Hadoop) create zero-byte directory-marker blobs whose name ends in `/`. The old `"/" not in relative_name` check skipped a marker-only path with no children; this `endswith("/")` branch will yield it as a folder. That's arguably more correct — a marker does denote a folder — just noting it as a deliberate difference, not a bug.

treff7es (approved):

metadata-ingestion/src/datahub/ingestion/source/azure/abs_folder_utils.py
acrylJonny: Yes, intentional — that's exactly the reasoning. Passing `prefix_with_separator` (`"database/"`) scopes the enumeration at the source so sibling prefixes like `database2/` no longer over-match, which lets us drop the client-side `startswith` filter entirely. The `delimiter="/"` walk is the perf win; the tighter prefix is the correctness win.

metadata-ingestion/src/datahub/ingestion/source/azure/abs_folder_utils.py
acrylJonny: Correct, and it's a deliberate difference. A zero-byte directory-marker blob does denote a folder, so surfacing it via the `endswith("/")` branch is the more accurate behavior — the old `"/" not in relative_name` check would silently drop a marker-only folder with no child objects.

PR #18292, opened 2026-07-09
perf(ingest/datalake): add S3 listing/schema instrumentation to the report

acrylJonny:
## Summary

Adds instrumentation to `DataLakeSourceReport` to make the relative cost of each ingestion phase visible and provide a baseline for the data-lake connector performance work:

- Counters: `objects_listed`, `tagging_api_calls`.
- Cumulative phase timers: `listing_time_taken_secs`, `schema_inference_time_taken_secs`.

These are incremented at the S3 listing loop in `get_folder_info` and the schema-inference / tagging paths in `ingest_table`. Purely additive; no behavior change.

## Testing

- Added `test_get_folder_info_records_listing_instrumentation`; all 39 tests in `tests/unit/s3/test_s3_source.py` pass.

## Checklist

- [x] Tests added/updated
- [ ] Breaking changes (n/a)

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/s3/report.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/s3/report.py
@@ -13,6 +13,15 @@ class DataLakeSourceReport(StaleEntityRemovalSourceReport):
     filtered: LossyList[str] = dataclass_field(default_factory=LossyList)
     number_of_files_filtered: int = 0
 
+    objects_listed: int = 0
+    # Number of tables for which tags were fetched (one get_s3_tags call each,
+    # which may issue multiple AWS requests internally), not a literal API count.
+    tables_tagged: int = 0
+    # Wall-clock for the whole listing phase (S3 list API + path matching +
+    # folder accumulation), not pure network I/O.
+    listing_time_taken_secs: float = 0.0
+    schema_inference_time_taken_secs: float = 0.0
+
     def report_file_scanned(self) -> None:
         self.files_scanned += 1
 
--- a/metadata-ingestion/src/datahub/ingestion/source/s3/source.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/s3/source.py
@@ -526,7 +526,11 @@ class S3Source(StatefulIngestionSourceBase):
         aspects.append(dataset_properties)
         if table_data.size_in_bytes > 0:
             try:
-                fields = self.get_fields(table_data, path_spec)
+                with PerfTimer() as schema_timer:
+                    fields = self.get_fields(table_data, path_spec)
+                self.report.schema_inference_time_taken_secs += (
+                    schema_timer.elapsed_seconds()
+                )
                 schema_metadata = SchemaMetadata(
                     schemaName=table_data.display_name,
                     platform=data_platform_urn,
@@ -558,6 +562,7 @@ class S3Source(StatefulIngestionSourceBase):
                 if table_data.full_path == table_data.table_path
                 else None
             )
+            self.report.tables_tagged += 1
             s3_tags = get_s3_tags(
                 bucket,
                 key_prefix,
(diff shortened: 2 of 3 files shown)


metadata-ingestion/src/datahub/ingestion/source/s3/report.py
```diff
@@ -13,6 +13,13 @@ class DataLakeSourceReport(StaleEntityRemovalSourceReport):
     filtered: LossyList[str] = dataclass_field(default_factory=LossyList)
     number_of_files_filtered: int = 0
 
+    # --- I/O and phase instrumentation ---
+    objects_listed: int = 0
+    tagging_api_calls: int = 0
```
treff7es: `tagging_api_calls` is incremented once per `get_s3_tags(...)` invocation (i.e. per table), but `get_s3_tags` can issue up to two AWS calls internally (bucket tags + object tags). So this counts tag-fetch invocations, not literal API calls — anyone later dividing latency by it to get per-call cost would be off. Consider `tables_tagged` or a comment clarifying the unit.

metadata-ingestion/src/datahub/ingestion/source/s3/report.py
```diff
@@ -13,6 +13,13 @@ class DataLakeSourceReport(StaleEntityRemovalSourceReport):
...
+    # cumulative time spent listing objects vs inferring schemas
+    listing_time_taken_secs: float = 0.0
```
treff7es, line 22: Minor: `listing_time_taken_secs` spans the whole listing loop, which also does `create_s3_path`, path-matching and folder accumulation — not purely the S3 list API. The name reads as pure I/O. Fine for a first baseline, but if the goal is isolating network vs. CPU cost, worth a one-word note that it's wall-clock for the whole listing+accumulation phase.

metadata-ingestion/src/datahub/ingestion/source/s3/source.py
```diff
@@ -770,9 +775,12 @@ def _is_allowed_path(path_spec_: PathSpec, s3_uri: str) -> bool:
 
         logger.info(f"Listing objects under {repr(uri)} with {prefix=}")
 
+        listing_timer = PerfTimer()
```
treff7es: `ingest_table` uses the context-manager form (`with PerfTimer() as schema_timer:`) but here it's manual `start()`/`finish()`. Beyond inconsistency, if `list_objects_recursive_path` raises mid-listing, the `finish()` at the bottom is skipped and the elapsed time is silently dropped; the `with` form records it via `__exit__` regardless. Suggest wrapping the listing loop in `with PerfTimer() as listing_timer:` for both consistency and exception-safety.

metadata-ingestion/tests/unit/s3/test_s3_source.py
```diff
@@ -332,6 +333,69 @@ def test_get_folder_info_returns_latest_file_in_each_folder(s3_resource):
...
+    bucket.create()
+    bucket.put_object(Key="my-folder/dir1/0001.csv")
+    bucket.put_object(Key="my-folder/dir1/0002.csv")
+    bucket.put_object(Key="my-folder/dir2/0001.csv")
+
+    source = _get_s3_source(path_spec)
+    list(source.get_folder_info(path_spec, "s3://my-bucket/my-folder"))
+
+    assert source.report.objects_listed == 3
+    assert isinstance(source.report.listing_time_taken_secs, float)
```
treff7es: `assert isinstance(..., float)` is vacuously true — the field is declared `float = 0.0`, so this passes even if the timer code never runs; it tests the type annotation, not behavior. `objects_listed == 3` is the meaningful assertion here. Either drop the isinstance check or make it `>= 0.0`. Same applies to `schema_inference_time_taken_secs >= 0.0` in the next test (near-trivial since it starts at 0.0) — though that test's `tagging_api_calls == 1` is genuinely meaningful.

treff7es (approved):

metadata-ingestion/src/datahub/ingestion/source/s3/report.py
acrylJonny: Good catch — renamed to `tables_tagged` (updated in the report, `ingest_table`, and the test). Added a comment noting it counts `get_s3_tags` invocations (one per table, each of which may issue multiple AWS requests internally), not literal API calls.

metadata-ingestion/src/datahub/ingestion/source/s3/report.py
acrylJonny: Agreed. Added a comment clarifying it's wall-clock for the whole listing phase (S3 list API + path matching + folder accumulation), not pure network I/O.

metadata-ingestion/src/datahub/ingestion/source/s3/source.py
acrylJonny: Done — the listing loop now uses `with PerfTimer() as listing_timer:` for consistency with `ingest_table`, and the report accumulation is in a `finally` so the elapsed time is recorded even if `list_objects_recursive_path` raises mid-stream.

metadata-ingestion/tests/unit/s3/test_s3_source.py
acrylJonny: Right — dropped the vacuous `isinstance(..., float)` check (keeping the meaningful `objects_listed == 3`) and removed the near-trivial `schema_inference_time_taken_secs >= 0.0` assertion, leaving the meaningful `tables_tagged == 1`.

PR #18293, opened 2026-07-09
perf(graphql): batch entity health resolution to eliminate N+1

gdatahub:
## Summary

**Stacked on #18585** (the `getActiveIncidentStats` search primitive). This PR is the GraphQL/health half of the original change; the search primitive was split out into #18585 for reviewability. Merge #18585 first.

The entity `health` field (`Dataset`, `Chart`, `Dashboard`, `DataFlow`, `DataJob`) is resolved by a **per-entity** `DataFetcher` (`EntityHealthResolver`). On any list/search/browse page that selects `health`, it fires once per entity — an N+1 fan-out of ES aggregations, graph lookups, and incident searches.

This PR adds **`EntityHealthBatchLoader`**, a GraphQL `DataLoader` that batches health resolution across all entities in a single request, selected by the feature flag **`entityHealthBatchLoadEnabled`** (**on by default**).

What gets batched (one call per request instead of one per entity):

- **Assertion-run aggregations** — N per-entity `getAggregatedStats` ES queries collapse to a **single** outer-terms-bucket `batchGetAggregatedStats` query (keyed on `asserteeUrn`).
- **Test results** — one `batchGetV2` per entity type instead of one `getV2` per entity.
- **Incidents** — consumes the **`getActiveIncidentStats`** primitive from #18585 (active-incident **count + latest active incident per entity** in a single ES `terms(entities.keyword)` + `top_hits` aggregation), followed by **one** `batchGetV2` for the latest incidents' `incidentInfo`. Replaces the previous N per-entity `filter()` searches + N `getV2` info fetches.

Health *interpretation* (filters, aggregation specs, row parsing, `Health` assembly) is extracted into a shared **`HealthComputationUtils`** used by both the per-entity resolver and the batch loader, so the two paths compute identical health.

**Failure isolation:** a failure in one dimension degrades to an empty result for that dimension (logged) rather than failing the page — both in the concurrent fetch phase (per-dimension `.exceptionally`) and the synchronous assemble phase (per-entity, per-dimension try/catch). Strictly more resilient than the legacy per-entity path under partial failure.

Only the graph active-assertion lookups (`getRelatedEntities`) remain per-URN — no batch relationship primitive exists yet — but they run concurrently within the loader.

## Rollout

Enabled **by default** (`entityHealthBatchLoadEnabled=true`); revert per environment with `ENTITY_HEALTH_BATCH_LOAD_ENABLED=false`, which runs the legacy per-entity path unchanged. Output is byte-identical between the two paths (see below), so this is a code-path change, not a behavior change.

## Verification

Validated end-to-end on a local instance by toggling the flag on the same build (so before/after is a pure path comparison).

- **Unit tests:** loader batch/failure-isolation tests + the `getActiveIncidentStats` request-shape test + a resolver flag-path test proving delegation to the DataLoader; the pre-existing resolver tests (unchanged) prove the shared-util refactor preserved behavior.
- **Functional parity — byte-identical.** Full ordered `health` response (type/status/message across the page) is SHA-256-identical flag-off vs flag-on at **every** page size tested (N = 1, 20, 50, 100, 200).

### Latency + DB benchmark — incident dimension (this revision)

Local instance, 200 seeded datasets each with one active incident; a single `searchAcrossEntities` request resolving `health` per row; warm JVM, 40 reps. Flag off (legacy per-entity) vs on (batched), same build.

| N (page size) | p50 before | p50 after | Δ latency | incident aspect reads before → after |
| --: | --: | --: | --: | --: |
| 1 | 34.3 ms | 32.5 ms | −5% | 2 → 1 |
| 20 | 44.2 ms | 43.9 ms | ~0% | 40 → 1 |
| 50 | 60.0 ms | 56.2 ms | −6% | 100 → 1 |
| 100 | 90.7 ms | 77.3 ms | **−15%** | 200 → 1 |
| 200 | 169.7 ms | 132.3 ms | **−22%** | 400 → 2 |

- **DB reads go O(N) → O(1):** the legacy path issues ~2·N incident-aspect reads (one `getV2` per incident); the batched path issues **one** `batchGetV2` for all incidents regardless of page size (2 at N=200 from key-chunking). The per-entity ES incident searches likewise collapse to one aggregation.
- **Scales with fan-out; no small-N regression.** At N=1 the change is within run-to-run noise (fixed GraphQL/search overhead dominates); the win emerges as the page grows and the per-entity incident work would otherwise dominate — ~15% at N=100, ~22% at N=200. This is the real-world case (search results, entity lists).

## Follow-ups (intentionally out of scope)

- Batch the graph active-assertion lookups (needs a batch relationship API) — the last remaining per-URN dimension.
- Guard the assertion aggregation against `search.max_buckets` for very high assertion cardinality (currently degrades gracefully by dropping the dimension for that sub-batch).
- Delete the now-superseded, unwired `DatasetHealthResolver`; converge and remove the legacy path once the flag has soaked.

## Checklist

- [x] Tests added/updated
- [ ] Docs added/updated (n/a — internal perf change behind a flag)
- [ ] Breaking changes documented in `updating-datahub.md` (n/a — output is byte-identical; flag revertible via env)

### DIFF
--- a/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/GmsGraphQLEngine.java
+++ b/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/GmsGraphQLEngine.java
@@ -150,6 +150,7 @@ import com.linkedin.datahub.graphql.resolvers.load.AspectResolver;
 import com.linkedin.datahub.graphql.resolvers.load.BatchGetEntitiesResolver;
 import com.linkedin.datahub.graphql.resolvers.load.DashboardStatsSummaryBatchLoader;
 import com.linkedin.datahub.graphql.resolvers.load.DatasetStatsSummaryBatchLoader;
+import com.linkedin.datahub.graphql.resolvers.load.EntityHealthBatchLoader;
 import com.linkedin.datahub.graphql.resolvers.load.EntityLineageResultResolver;
 import com.linkedin.datahub.graphql.resolvers.load.EntityRelationshipsResultResolver;
 import com.linkedin.datahub.graphql.resolvers.load.EntityTypeBatchResolver;
@@ -367,6 +368,7 @@ import com.linkedin.metadata.models.registry.EntityRegistry;
 import com.linkedin.metadata.query.filter.SortCriterion;
 import com.linkedin.metadata.query.filter.SortOrder;
 import com.linkedin.metadata.recommendation.RecommendationsService;
+import com.linkedin.metadata.search.EntitySearchService;
 import com.linkedin.metadata.service.ApplicationService;
 import com.linkedin.metadata.service.AssertionService;
 import com.linkedin.metadata.service.BusinessAttributeService;
@@ -440,6 +442,7 @@ public class GmsGraphQLEngine {
   private final GitVersion gitVersion;
   private final boolean supportsImpactAnalysis;
   private final TimeseriesAspectService timeseriesAspectService;
+  private final EntitySearchService entitySearchService;
   private final TimelineService timelineService;
   private final NativeUserService nativeUserService;
   private final GroupService groupService;
@@ -586,6 +589,7 @@ public class GmsGraphQLEngine {
     this.gitVersion = args.gitVersion;
     this.supportsImpactAnalysis = args.supportsImpactAnalysis;
     this.timeseriesAspectService = args.timeseriesAspectService;
+    this.entitySearchService = args.entitySearchService;
     this.timelineService = args.timelineService;
     this.nativeUserService = args.nativeUserService;
     this.groupService = args.groupService;
@@ -958,6 +962,15 @@ public class GmsGraphQLEngine {
             DashboardStatsSummaryBatchLoader.LOADER_NAME,
             context ->
                 DashboardStatsSummaryBatchLoader.createDataLoader(timeseriesAspectService, context))
+        .addDataLoader(
+            EntityHealthBatchLoader.LOADER_NAME,
+            context ->
+                EntityHealthBatchLoader.createDataLoader(
+                    this.entityClient,
+                    this.graphClient,
+                    timeseriesAspectService,
+                    this.entitySearchService,
+                    context))
         .addDataLoader(
             DatasetStatsSummaryBatchLoader.LOADER_NAME,
             context ->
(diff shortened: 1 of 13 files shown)


AdrianMachado:
Is there a way to break this up into multiple PRs? Not easy to review this many changes

metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/query/ESSearchDAO.java
```diff
@@ -550,6 +562,93 @@ public SearchRequest buildAggregateByValue(
...
+          }
+        },
+        MetricUtils.DROPWIZARD_NAME,
+        MetricUtils.name(this.getClass(), "getActiveIncidentStats_search"));
+  }
+
+  @VisibleForTesting
+  public SearchRequest buildActiveIncidentStatsRequest(
+      @Nonnull OperationContext opContext, @Nonnull Set<Urn> entityUrns) {
+    final String[] urnStrings = entityUrns.stream().map(Urn::toString).toArray(String[]::new);
```
chakru-r, line 618: This list is potentially unbounded -- this needs an assessment that this doesnt break ES limits and may need to be partitioned into a few requests

gdatahub:
Screenshot 2026-07-23 at 10 10 29 PM
Screenshot 2026-07-23 at 10 10 43 PM
Screenshot 2026-07-23 at 10 55 35 PM

metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/query/ESSearchDAO.java
gdatahub: You are right, it was unbounded. I have added partitioning so getActiveIncidentStats now chunks the URN set into batches of 1000 (INCIDENT_STATS_URN_BATCH_SIZE).

gdatahub: This query is primarily constrained by index.max_terms_count (for the entities.keyword terms filter) and search.max_buckets (since the by-entity terms aggregation creates one bucket per URN). Both default to 65,536, so a batch size of 1000 stays comfortably below those limits.

AdrianMachado (commented):
From a product perspective, this all seems good to me based on your benchmarks and the code for actually computing the health being almost identical to before. I would recommend getting @david-leifker or Alex Yang to review the batch resolution piece as they understand ES much better than me

chakru-r (approved):

PR #18300, opened 2026-07-09
fix(ingest/mysql): dispose reflection engines and document connection limits - CUS-9272

acrylJonny:
## Summary

- `MySQLSource.get_inspectors` created a SQLAlchemy engine per database but never disposed them, so each engine's pool kept one idle connection open for the rest of the run — one leaked connection per database.
- With profiling enabled the per-database pool can also grow to `pool_size` + `max_overflow`, and `max_overflow` defaults to `profiling.max_workers` (`5 × CPU`), so servers with a low `max_user_connections` cap were exhausted.
- Dispose the discovery engine and each per-database engine once its inspector has been consumed (reflection + profiling for that database run in the caller's loop body before the generator resumes, so disposal is safe).
- Document the connection footprint and how to bound it (`profiling.max_workers`, `options` pool sizing, or disabling profiling) for MySQL and MariaDB.

## Test plan

- Added `tests/unit/test_mysql_connection.py` asserting the discovery engine and every per-database engine are disposed.
- `ruff`, `ruff format`, `mypy`, the new unit test, and the existing `test_mysql_usage.py` pass; `mdPrettierCheck` passes.


metadata-ingestion/src/datahub/ingestion/source/sql/mysql.py
```diff
@@ -327,21 +327,33 @@ def get_inspectors(self):
...
+        for db in databases:
+            if not self.config.database_pattern.allowed(db):
+                continue
+            db_url = self.config.get_sql_alchemy_url(current_db=db)
+            db_engine = create_engine(db_url, **self.config.options)
+            self._setup_rds_iam_event_listener(db_engine, database_name=db)
+            try:
+                with db_engine.connect() as conn:
+                    inspector = inspect(conn)
+                    yield inspector
```
treff7es, line 367: This dispose-on-resume pattern is correct, but it's load-bearing on an implicit contract: all reflection + profiling for `db` must complete in the caller's loop body before it requests the next inspector — otherwise `finally: db_engine.dispose()` tears the engine down underneath in-flight work. It holds today; worth a one-line comment stating the invariant so a future refactor to lazy/deferred profiling doesn't silently break it. (The test covers disposal counts but mocks `inspect`, so it doesn't exercise this ordering — an integration-level check would, but that's optional.)

metadata-ingestion/src/datahub/ingestion/source/sql/mysql.py
```diff
@@ -327,21 +327,33 @@ def get_inspectors(self):
...
+        finally:
+            # Only used to list databases; dispose so it does not hold a pooled
+            # connection open for the whole reflection/profiling run.
+            engine.dispose()
+
+        for db in databases:
+            if not self.config.database_pattern.allowed(db):
+                continue
+            db_url = self.config.get_sql_alchemy_url(current_db=db)
+            db_engine = create_engine(db_url, **self.config.options)
```
treff7es, line 359: Note the shared surface with #18319: `self.config.options` here carries the `max_overflow` that `_add_default_options` injects when profiling is on — which is exactly why this per-DB QueuePool can grow and exhaust connections (fine here — QueuePool accepts it). #18319 fixes the mirror-image problem where that same injected option breaks the NullPool usage engine. Same root cause; worth cross-referencing the two PRs.

treff7es (approved):

metadata-ingestion/src/datahub/ingestion/source/sql/mysql.py
acrylJonny: Added a one-line invariant comment at the yield: the caller must finish all reflection + profiling for the db before requesting the next inspector, since the finally disposes the engine on resume. Flags the ordering for any future deferred-profiling refactor.

metadata-ingestion/src/datahub/ingestion/source/sql/mysql.py
acrylJonny: Added a cross-reference to #18319 at the per-DB create_engine call, noting the shared max_overflow root cause: QueuePool accepts it here, while #18319 fixes the mirror-image NullPool case.

PR #18301, opened 2026-07-09
fix(sql-parsing): filter empty schemaField columns in column-level lineage

naeemdh:
## Summary

Column-level lineage could emit an invalid `schemaField` URN with an empty field path (`urn:li:schemaField:(<dataset>,)`), which GMS rejects with a 422 during ingestion. First surfaced via Tableau custom SQL datasources, but the root cause lives in the shared `sql_parsing` module, so this fixes it for every sqlglot-derived source at once.

**Root cause:** `ColumnLineageInfo.upstreams` is built in exactly one place — `_translate_internal_column_lineage()` in `sql_parsing/sqlglot_lineage.py`. sqlglot's column-lineage resolver can emit a `ColumnLineageInfo` entry with an empty upstream column name when it can't cleanly resolve a column against the upstream table's schema (e.g. for Tableau, after `lineage_overrides.platform_override_map` swaps the SQL's original platform to an overridden one). That empty string flowed straight through into `make_schema_field_urn()`.

Tableau's own `make_fine_grained_lineage_class()` / `get_upstream_fields_from_custom_sql()` guarded the *downstream* column against `None`, but never against an empty string (the field is typed `str`, not `Optional[str]`, so it's never actually `None` — that check was dead code), and had no guard at all on the *upstream* side.

## What Changed

- **Root fix** — `sql_parsing/sqlglot_lineage.py::_translate_internal_column_lineage()`: filter out upstream `ColumnRef`s with an empty `column` at the single choke point where they're built, protecting every consumer of `ColumnLineageInfo`, not just Tableau.
- **New helpers** on `ColumnLineageInfo`: `upstream_schema_field_urns()` and `downstream_schema_field_urn()`, so callers don't need to hand-roll the empty-column filter.
- **Tableau**: `tableau_common.py::make_fine_grained_lineage_class()` and `tableau.py::get_upstream_fields_from_custom_sql()` now use the new `upstream_schema_field_urns()` helper, and their downstream guards were tightened from a no-op `is not None` check to an actual truthiness check (same empty-string gap as upstream, just less commonly hit).

## Testing

- `tests/unit/sql_parsing/test_column_lineage_info.py` (new): unit tests for the two new `ColumnLineageInfo` helpers and the `_translate_internal_column_lineage` filter.
- Tableau: regression tests for both the upstream and downstream empty-column cases, in both `make_fine_grained_lineage_class` and `get_upstream_fields_from_custom_sql`.
- Full `tests/unit/tableau/` + `tests/unit/sql_parsing/` suites (758 passed) and `./gradlew :metadata-ingestion:lintFix` / `lint` (ruff + mypy) clean.

## Related

Follow-up #18305 retrofits the other connectors that had the same unguarded inline pattern (Mode, Microstrategy, QuickSight, Grafana, PowerBI, ThoughtSpot, Looker) to use the new helper — they're now protected transitively by the root fix here regardless.

## Checklist

- [x] Tests added/updated
- [ ] Docs added/updated (n/a — internal bugfix, no user-facing config change)

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


metadata-ingestion/src/datahub/ingestion/source/tableau/tableau.py
```diff
@@ -2361,9 +2361,14 @@ def get_upstream_fields_from_custom_sql(
                 and cll_info.downstream.column is not None
                 else []
             )
+            # column_ref.column can be an empty string when sqlglot's column-lineage
+            # resolution fails to match a column against the upstream table schema.
+            # Passing that through would build an invalid schemaField URN with an
+            # empty field path and get rejected by GMS.
```
sgomezvillamor: The fix looks good, though it's quite targeted — worth thinking about pushing it higher up: does it make sense for `parsed = self.parse_custom_sql(` to return upstream `ColRef`s without columns in the first place?

Also or alternatively, I notice this pattern is quite common:

```python
upstreams = [
    builder.make_schema_field_urn(column_ref.table, column_ref.column)
    for column_ref in cll_info.upstreams
    if column_ref.column
]
```

Would it make sense to encapsulate this in a method like `cll_info.upstreams.as_schema_field_urns()`, so the filtering logic lives in one place?

The goal is to fix this consistently across all sources, not just Tableau.

naeemdh: Good call — pushed it to the root. `ColumnLineageInfo.upstreams` is built in exactly one place, `_translate_internal_column_lineage` in `sqlglot_lineage.py`, so I filtered empty columns there instead of just at the Tableau call site. Also added `ColumnLineageInfo.upstream_schema_field_urns()` / `downstream_schema_field_urn()` helpers and switched Tableau's two call sites to use them.

I didn't retrofit the other ~11 connectors with the same unguarded pattern I found while digging into this (Mode, Looker, PowerBI, Superset, ThoughtSpot, etc.) — they're now protected automatically since they all go through `_translate_internal_column_lineage`, but their call sites still have the old inline pattern. Figured that's a separate cleanup PR rather than scope-creeping this one, happy to open it as a follow-up if useful.

naeemdh: here is a follow up PR for the remaining 11 connectors to use the helper. https://github.com/datahub-project/datahub/pull/18305/changes

sgomezvillamor (approved):
LGTM — good call pushing the fix down to `_translate_internal_column_lineage` so it applies to every source at once.

Two non-blocking items:

- **Guard asymmetry**: the upstream side now filters empty columns, but the downstream guards still use `is not None` (`tableau.py:2360`, `tableau_common.py:1025`). An empty-string downstream column would slip through and build the same invalid `schemaField:(...,)` URN. Worth tightening to a truthiness check while you're here — same as you're doing in #18305.
- **Scope/title**: now that the actual fix lives in shared `sqlglot_lineage.py`, the `fix(ingest/tableau)` title undersells it. Consider updating the title/description to reflect the framework-level change.

Approving.

---

naeemdh:
Thanks! Both addressed:

- **Guard asymmetry**: tightened the downstream check in both `make_fine_grained_lineage_class` and `get_upstream_fields_from_custom_sql` from the dead `is not None` (the field is `str`, never `Optional[str]`, so that check never fired) to an actual truthiness check — same fix as the follow-up PR's connectors. Added regression tests for the empty-downstream-column case in both functions.
- **Scope/title**: updated the title to `fix(sql-parsing): filter empty schemaField columns in column-level lineage` and rewrote the description to lead with the root-cause fix in `sqlglot_lineage.py`, with Tableau as the triggering case rather than the whole story.

PR #18302, opened 2026-07-09
fix(ingest/vertexai): skip projects with the Vertex AI API disabled instead of aborting the sweep

hhobson:
## Motivation

DataHub's GCP connectors (Vertex AI, BigQuery, Dataplex, GCS) let a broad `project_id_pattern` ingest a lot of projects at once, read with a single service account. Project resolution runs against Resource Manager, a different API from the one being ingested, so a resolved project may never have had the target API (e.g. `aiplatform.googleapis.com`) enabled. The first RPC against such a project raises `PermissionDenied` (403) carrying `reason="SERVICE_DISABLED"`. Left uncaught at the per-project loop, that single project aborts the entire multi-project run, so one unused or partially-provisioned project blocks ingestion of every other project in the sweep.

A disabled target API is an expected state in a wide sweep, not a run failure: projects are provisioned unevenly and the enabled-API set varies across an org. A resilient connector must skip the disabled project with a non-fatal warning and continue, while still surfacing genuine failures (IAM denials, quota exhaustion, outages) rather than swallowing them. This is a live failure in the Vertex AI connector today, and fixing Vertex AI is the entire scope of this PR; adopting the shared helper across the other GCP connectors is a follow-up.

## Changes

- **New `source/common/gcp_errors.py` with `is_service_disabled(exc)`.**
  - Detects a disabled API via `exc.reason == "SERVICE_DISABLED"` rather than message text: the live 403 names the API "Agent Platform API" (Google's rebranded display name for `aiplatform.googleapis.com`), so a message-string match would miss it, whereas `.reason` is populated by `google-api-core` from `ErrorInfo` and is stable.
  - It lives in `common/` because the failure mode is generic to GCP sources; only the skip-and-continue application is Vertex-specific.
- **`get_workunits_internal` wraps each project in `except PermissionDenied`.**
  - If `is_service_disabled(e)` it emits a `report.warning` and continues; otherwise it re-raises.
  - Only `SERVICE_DISABLED` is skippable, so genuine IAM denials, quota, timeouts, and the separate Metadata Store `NotFound` still propagate rather than becoming silent failures.
  - It uses `report.warning` (not `report.failure`) because a swept project that never enabled the API is expected, not a run failure.
  - It is caught at the per-project loop rather than in the experiment extractor because the throw originates upstream, in the phase-1 training extractor's first Vertex call.
- **Containers are emitted only after the first resource is fetched.**
  - The per-project body moved to a new `_gen_project_resource_workunits`, and `_gen_workunits_for_project` peeks the first resource (`next(resources, None)`) before yielding `_gen_project_workunits()`.
  - A disabled project raises on that first fetch, so it no longer emits orphan project/category containers before being skipped; enabled-but-empty projects still emit their containers, unchanged.

## Testing

- `tests/unit/common/test_gcp_errors.py`: `is_service_disabled` returns true for `SERVICE_DISABLED`, false for another reason (a genuine IAM denial), and false when the exception carries no `ErrorInfo`.
- `tests/unit/vertexai/test_vertexai_source.py`: a disabled-API project is skipped while other projects still ingest; a non-`SERVICE_DISABLED` `PermissionDenied` still propagates.
- `./gradlew :metadata-ingestion:lint` (ruff + mypy) clean; the two test modules pass (25 tests).

No integration/golden tests: the change is control flow only, no new entities or aspects.

## Pre-existing issues surfaced by the refactor (out of scope)

Extracting the per-project body moved some existing code verbatim. These predate this PR and are left untouched to keep it a targeted control-flow fix; each is worth a separate follow-up:

- **Confusingly similar method names.** The pre-existing `_gen_project_workunits` (containers) now sits next to the new `_gen_workunits_for_project` and `_gen_project_resource_workunits` and is easy to misread. A clarifying rename would touch pre-existing code, so it is deferred.
- **The inner ML-metadata init `try/except` is too broad and logs via a bare logger.** It catches `GoogleAPICallError` and `ValueError` (which can mask genuine bugs) and reports failures with `logger.warning`, so they never surface in the ingestion report. It should be narrowed, switched to `self.report.warning` and caught at `get_workunits_internal` level. This block was only relocated verbatim by the extraction, not edited here.

## Checklist

- [ ] Docs (n/a — no new user-facing config or behavioural surface)
- [ ] Breaking change / updating-datahub entry (n/a — no breaking change)


sgomezvillamor (approved):
LGTM. Well-scoped control-flow fix, and it complements #17660 rather than overlapping — that one skips experiment extraction on a 404 (Metadata Store never created), this skips the whole project on a 403 `SERVICE_DISABLED` (API never enabled). Error taxonomy looks right: only `SERVICE_DISABLED` is swallowed, IAM denials / quota / timeouts / Metadata-Store `NotFound` all still propagate. The peek-before-containers refactor is behavior-preserving (containers are project-scoped, so relocating them after the first region init is safe) and avoids orphan containers on skipped projects.

One minor suggestion: `is_service_disabled(exc: Forbidden)` — the annotation works since `PermissionDenied` subclasses `Forbidden`, but `.reason` actually lives on `GoogleAPICallError`. Since the helper is framed as GCP-generic (in `common/`), widening the param to `GoogleAPICallError` would let the other GCP connectors reuse it for non-403 error families later without a type-ignore. Non-blocking.

---

sgomezvillamor:
Hi @hhobson — just checking, is there anything blocking this from being merged?

hhobson:
> Hi @hhobson — just checking, is there anything blocking this from being merged?

Hi @sgomezvillamor - No, nothing from my side. I had been meaning to ask the same question. I have a couple of PRs that have been approved and have `merge-pending-ci` tag, is there anything more I need to do at this stage?

PR #18305, opened 2026-07-09
fix(sql-parsing): retrofit connectors to use the empty-column schemaField filter

naeemdh:
## Summary

Follow-up to #18301. That PR fixed the root cause (an unguarded empty upstream column producing an invalid `schemaField` URN) at the single choke point in `sqlglot_lineage.py`, plus added `ColumnLineageInfo.upstream_schema_field_urns()` / `.downstream_schema_field_urn()` helpers so callers don't have to hand-roll the filter. This PR retrofits the other connectors that had the same unguarded inline pattern, per review discussion on #18301.

**Stacked on #18301** — this diff will shrink to just the retrofit once that one merges into `master`.

## What Changed

**Switched to the new helper(s)** (clean 1:1 fit, no per-connector URN transform in the way):
- `mode.py`, `microstrategy/lineage.py`, `quicksight/quicksight_lineage.py` → `ColumnLineageInfo.upstream_schema_field_urns()`
- `looker/lookml_source.py` → new module-level `column_refs_to_schema_field_urns()` (needed because `field.upstream_fields` is a raw `List[ColumnRef]`, not a `ColumnLineageInfo`; extracted the filtering logic out of the method into a shared free function both can call)
- `grafana/lineage.py` → same helper for upstream, plus a truthy guard on downstream (was fully unguarded on both sides)

**Added direct guards** (couldn't cleanly adopt the helper — each applies its own per-row URN transform before building the schemaField URN — alias rewriting, lowercasing, dedup):
- `powerbi/powerbi.py` — also fixes a real latent bug independent of #18301: PowerBI builds `ColumnLineageInfo` directly in `m_query/pattern_handler.py`, bypassing `sql_parsing`'s own translation step entirely, so it was never protected by the root fix.
- `thoughtspot/source.py` — guards both the upstream loop (defense-in-depth; already sqlglot-derived and covered transitively by #18301) and the downstream column (not covered by the root fix, which only filtered the upstream side).

**Left unchanged**:
- `superset.py` — its one unguarded site applies `_apply_database_alias_to_urn()` per-row, so it can't use the plain helper, but it's already sqlglot-derived (confirmed via `create_lineage_sql_parsed_result`) and therefore protected transitively by #18301's root fix. Its downstream side already had a correct truthy guard.
- `looker/view_upstream.py` — doesn't call `make_schema_field_urn` itself, just passes `ColumnRef` lists downstream to `looker_common.py`.
- `looker/looker_common.py` — already has a more detailed guard (per-ref warning logging, whitespace-stripping) than the generic helper provides; consolidating onto the plain helper would regress that.

## Testing

- New/updated unit tests: `test_grafana_lineage.py`, `tests/unit/powerbi/test_column_lineage.py` (new file), `test_thoughtspot_source.py`, `test_quicksight_lineage.py` (fixed a test that mocked `ColumnLineageInfo` with `SimpleNamespace`, which broke once the code called the new real method on it).
- Ran the full suite for every touched connector plus `sql_parsing/` and `tableau/`: 1643 passed.
- `./gradlew :metadata-ingestion:lintFix` / `lint` (ruff + mypy): clean.

## Checklist

- [x] Links to related issue (#18301, stacked)
- [x] Tests added/updated
- [ ] Docs added/updated (n/a — internal bugfix, no user-facing config change)

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


sgomezvillamor (approved):
👍

PR #18306, opened 2026-07-09
fix(ingest/sql-parsing): guard sqlglot FLATTEN crash when flattened column is unschema'd

treff7es:
## Summary

The `LATERAL FLATTEN` / `UNNEST` resolver-recursion guard in `sqlglot_lineage.py` (added to stop a sqlglot >= 30.7.0 crash) only skipped column-level lineage when a referenced **table** was entirely absent from the schema. The same sqlglot resolver recursion also fires when the base table **is** schema'd but the **flattened column itself is missing** from it — e.g. a semi-structured VARIANT/array column that was not surfaced into the schema. Those statements slipped past the guard and crashed ingestion.

On the compiled `sqlglot[c]` build the recursion overflows the native C stack into an **uncatchable SIGSEGV that kills the whole ingest process** (no traceback); on pure-Python sqlglot it surfaces as a catchable `RecursionError` and the run degrades but completes. This is why a real Snowflake `FLATTEN(<variant_col>)` query could hard-crash ingestion even on CLI builds that already contained the original guard.

## What changed

- Broadened `_statement_risks_unnest_resolver_recursion` to flag any `LATERAL FLATTEN`/`UNNEST` of an **unqualified column that is not a known column of any schema'd source**, rather than only the missing-table case. The new condition is strictly broader than the old one (every previously-detected case still fires), so existing coverage is preserved.
- Table-level lineage is unaffected; only column-level lineage is skipped for the risky statement.

## Testing

- Added two regression tests (detector-level and end-to-end) for the "base table schema'd, flattened column missing" case.
- New tests fail before the fix (detector returns `False`; end-to-end recurses) and pass after.
- Verified under the compiled `sqlglot[c]` build that the previously-crashing end-to-end case now completes cleanly (skips CLL, keeps table lineage) instead of segfaulting.
- Full `tests/unit/sql_parsing/` suite passes with no regressions.

## Checklist

- [x] Tests added/updated
- [ ] Docs — n/a (internal SQL-parsing robustness fix)
- [ ] Breaking changes — none

### DIFF
--- a/metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
+++ b/metadata-ingestion/src/datahub/sql_parsing/sqlglot_lineage.py
@@ -863,37 +863,49 @@ def _statement_risks_unnest_resolver_recursion(
     """Detect the sqlglot resolver infinite-recursion trigger (sqlglot >= 30.7.0).
 
     sqlglot's column Resolver recurses without a re-entrancy guard when it tries
-    to type a ``LATERAL FLATTEN`` / ``UNNEST`` of an *unqualified* column whose
-    base table is absent from the schema, while a schema is otherwise present:
-    typing the unnested column disambiguates it via ``get_table``, which
-    enumerates every source's columns, which re-enters the same lateral source.
-    In the compiled ``sqlglot[c]`` build this overflows the native C stack into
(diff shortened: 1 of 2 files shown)


sgomezvillamor:
superseded by #18317

PR #18311, opened 2026-07-10
feat(ece): Add source details to attribution ece parameters

asikowitz:
Standardizes behavior across ECEs for aspects with `MetadataAttribution` to store `sourceDetails` on the ECE parameters. Also standardizes parameter creation on the Event class rather than ChangeEventGenerator.

### DIFF
--- a/metadata-io/src/main/java/com/linkedin/metadata/timeline/data/entity/ChangeEventParameterUtils.java
+++ b/metadata-io/src/main/java/com/linkedin/metadata/timeline/data/entity/ChangeEventParameterUtils.java
@@ -1,12 +1,11 @@
 package com.linkedin.metadata.timeline.data.entity;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.linkedin.common.MetadataAttribution;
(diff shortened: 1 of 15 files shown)


metadata-io/src/main/java/com/linkedin/metadata/timeline/data/entity/ChangeEventParameterUtils.java
```diff
@@ -0,0 +1,25 @@
+package com.linkedin.metadata.timeline.data.entity;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.linkedin.common.MetadataAttribution;
+import javax.annotation.Nullable;
+
+/** Helpers for building ChangeEvent parameter maps shared across entity change events. */
+public class ChangeEventParameterUtils {
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
RyanHolstien: Shouldn't be creating new ObjectMappers in util classes. Should use the shared injectable unless you have a reason to create on with different settings.

RyanHolstien (approved):

PR #18316, opened 2026-07-10
refactor(ingest/sql-parsing): reconstruct struct subfield lineage without patching sqlglot

aviraj-gour:
## Summary

DataHub's SQL column-level lineage was monkey-patching the internals of the `sqlglot` library to produce **struct subfield lineage** (e.g. `col.a.b` rather than just `col`). Those patches were applied as line-numbered text diffs against sqlglot's source, so every sqlglot upgrade risked breaking them and required manual re-alignment.

This PR removes the patches and reconstructs the same subfield information in our own code instead. `sqlglot` now runs completely unmodified.

## Background: what problem are we solving?

For a query like:

```sql
SELECT min(widget.metric.metricA, widget.metric.metric_b) AS min_metric
FROM t
```

we want column lineage that records the **subfields**:

```
min_metric  <-  t.widget.metric.metricA
min_metric  <-  t.widget.metric.metric_b
```

### How sqlglot represents `widget.metric.metricA`

`sqlglot` parses the query into a tree. A struct access like `widget.metric.metricA` becomes a plain `Column` (`widget`) wrapped in `Dot` nodes that hold the field names:

```
Dot( Dot( Column(widget), "metric" ), "metricA" )
          └── the Column itself is just "widget"; ".metric.metricA" lives in the Dots
```

### Why stock sqlglot loses the subfield

`sqlglot.lineage.lineage(sql)` returns a tree of `Node`s. The bottom nodes — the **leaves** — are the real source columns. When sqlglot builds a leaf it records only the `Column` (`widget`) and **drops the surrounding `Dot`s**, so the subfield is gone.

On top of that, sqlglot stores the source columns of an expression in a `set`. Both `widget.metric.metricA` and `widget.metric.metric_b` share the same underlying `Column` (`widget`), so the set treats them as duplicates and keeps only **one** — two subfields collapse into a single leaf.

Out of the box you therefore get only `min_metric <- t.widget` (one edge, no subfield).

## Previous behavior: patch sqlglot's internals

We used to rewrite two lines *inside* sqlglot's `to_node` function via text-diff patches (applied with the `patchy` library):

1. **`set` → `list`** for the source columns, so repeated base columns survive as separate leaves instead of collapsing to one.
2. **Subfield extraction** at leaf-creation time — the one moment sqlglot still holds the `Dot`s — climbing the `Dot`s to build `"metric.metricA"` and stapling it onto the leaf as a new `subfield` field.

Our code then read `leaf.subfield`.

## New behavior: leave sqlglot alone, rebuild the subfield ourselves

Key insight: **the subfield information isn't actually lost — it's still in the query.** sqlglot drops the `Dot`s from the *leaf*, but the full expression `min(widget.metric.metricA, widget.metric.metric_b)` still sits on the node **above** the leaf. So we let sqlglot run unmodified and re-derive the subfields from that expression.

For each leaf, we look back at the expression that produced it (its immediate parent node), find every use of that column, climb its `Dot`s to recover the subfield, and emit one upstream per distinct subfield found.

Two things fall out naturally:

- The subfield comes back, because we read the `Dot`s straight from the query text instead of relying on sqlglot to carry them on the leaf.
- The `set`-collapse no longer matters — we *want* one leaf, and we re-expand it into N upstreams by finding all N subfield uses in the source expression. So the old `set`→`list` patch is unnecessary.

## Performance

Subfield reconstruction reads the leaf's parent expression, and many leaves share the same parent (e.g. a wide `COALESCE`/`CASE`). A naive per-leaf rescan that re-serializes every column is therefore quadratic in the width of a single select.

We build the leaf-name → subfields map once per distinct parent and reuse it across that parent's leaves, so `column.sql()` runs once per column occurrence instead of once per (leaf × column).

Measured on a single wide `COALESCE` over N base columns, `.sql()` calls drop from N² to N (e.g. 1600 → 40 at N=40), with identical output.

## Known limitation: subfield lineage can coarsen

Because we recover subfields from the leaf's **immediate parent** expression, accuracy depends on that parent actually carrying the subfield access. When a struct column flows through an intermediate scope, the access lives a scope higher than the leaf's parent and lineage coarsens to the base column (e.g. `asset_id <- widget` instead of `widget.asset.id`).

Two cases are covered by tests:

- `test_select_struct_subfield_through_cte_passthrough` — struct passed through a CTE, accessed in the outer query.
- `test_join_struct_subfields_shared_base_name` — a join, where sqlglot inserts an intermediate `widget AS widget` projection.

This is **not a regression**: the old patched code also derived subfields from the immediate leaf-producing expression, so accurate cross-scope subfield lineage was never supported, before or after this PR. Fully resolving it would require support on the sqlglot side; we raised it upstream and it was declined as out of scope for them, so this workaround stands and the tests above pin the current behavior.

## Tests

- `test_select_struct_subfields_from_cte` — subfield access inside a CTE that reads the table (fine-grained; guards reconstruction across nesting levels).
- `test_select_struct_subfield_through_cte_passthrough` — struct passed through a CTE, accessed in the outer query (documents coarsening).
- `test_join_struct_subfields_shared_base_name` — two joined tables sharing a struct base-column name (documents coarsening, and guards that each base column still resolves to the correct table).

## Cleanup included in this PR

With the two patches gone, their supporting scaffolding is dead and removed too:

- the pure-Python `lineage.py` reload,
- the `Node` subclass with its `subfield` field,
- the `patchy` / `apply_diff` / `_new_apply_patch` override.

`_sqlglot_patch.py` now keeps only the `__deepcopy__` cooperative-timeout wrapper (the one remaining, unrelated behavior override).


aviraj-gour:
Old approach: ran inside sqlglot at the moment the leaf was created. At that instant sqlglot was holding the exact Column object that produced the leaf. It climbed that specific object's Dots. Zero ambiguity — it's object identity.

New approach sqlglot_lineage.py:1328-1330: the leaf no longer remembers which Column it came from. So we go to the parent expression, look at every Column in it, and re-identify the right one by string comparison:

sgomezvillamor:
Nice refactor — dropping the sqlglot source patches for something we own is the right direction. Since this is a "no behavior change" refactor, the whole safety argument rests on tests, so that's where I focused.

The existing golden tests (`test_select_from_struct_subfields`, `test_select_with_full_col_name`, `test_snowflake_full_table_name_col_reference`, `test_bigquery_unnest_columns`) are good pre-existing guards and the fact that this PR touches zero golden files is a strong signal the output is byte-identical for the covered cases. But those cases are all essentially single-level selects, and I think there are a few gaps worth closing before we trust this end-to-end:

**Gap 1 — nested subquery / CTE.** This is the important one. The whole premise of the new approach is "use the leaf's *immediate* parent expression, not the root, because in a nested query the root wouldn't even mention the column." None of the current tests actually exercise a struct subfield projected out of an inner SELECT/CTE, so the exact logic the rewrite hinges on is untested. Please add one.

**Gap 2 — join / self-join with a shared base-column name.** `_leaf_column_subfields` matches occurrences by string equality (`column.sql(comments=False) != leaf_name`). If `t1.widget` and `t2.widget` both appear with different subfields, the parent-expression scan can attach subfields from the wrong table's column. We don't have a test with ambiguous base-column names — worth one.

**Gap 3 — the `parent is None` fallback** (`subfields.add("")`). Minor / defensive path, but it's currently uncovered. Fine to skip if it's hard to trigger through real SQL, just calling it out.

**Performance concern.** `_leaf_column_subfields` runs once per leaf and calls `column.sql(comments=False)` on every column returned by `find_all_in_scope`. Since multiple leaves share the same parent, that re-scan + re-serialize gets redone per leaf. For one select expression with ~M column occurrences producing ~L leaves, that's roughly **L × M `.sql()` calls — quadratic in the width of a single select**. The old code was O(1) per leaf: it just read the precomputed `node.subfield` attribute sqlglot filled in during the build. It's invisible for normal queries, but wide generated SELECTs (big CASE/COALESCE/function expressions) are exactly where it'd bite.

Idea to consider: **memoize per parent** — compute the `leaf_name → subfields` map once per distinct parent expression (bucket every column by `.sql()` in a single pass) and reuse it across that parent's leaves, instead of rescanning per leaf. That should call `.sql()` once per column occurrence total and drop the quadratic factor, while keeping output identical. Would that address the concern from your side, or is there a reason the per-leaf scan is needed?

---

aviraj-gour:
Thanks for review @sgomezvillamor,


> Gap 1 — nested subquery / CTE. This is the important one. The whole premise of the new approach is "use the leaf's immediate parent expression, not the root, because in a nested query the root wouldn't even mention the column." None of the current tests actually exercise a struct subfield projected out of an inner SELECT/CTE, so the exact logic the rewrite hinges on is untested. Please add one.

Added `test_select_struct_subfields_from_cte` — the struct subfield access lives inside the CTE, while the outer SELECT post_id, asset_id, min_metric FROM cte never mentions widget
 ```
SELECT post_id, asset_id, min_metric FROM cte never mentions widget.
WITH cte AS (
    SELECT
        post_id,
        widget.asset.id AS asset_id,
        min(widget.metric.metricA, widget.metric.metric_b) AS min_metric
    FROM data_reporting.abcde_transformed
)
SELECT post_id, asset_id, min_metric
FROM cte
```




> Gap 2 — join / self-join with a shared base-column name. _leaf_column_subfields matches occurrences by string equality (column.sql(comments=False) != leaf_name). If t1.widget and t2.widget both appear with different subfields, the parent-expression scan can attach subfields from the wrong table's column. We don't have a test with ambiguous base-column names — worth one.

Checked — not reachable. In a join sqlglot resolves each widget through an intermediate widget AS widget node, so lineage coarsens to the base column anyway (no per-subfield attribution happens), and each base column still lands on the correct table. SQL also forbids two sources sharing an alias in one scope, and the column scan is scope-local, so a subfield can't cross tables. Added a test with two joined tables sharing the base name widget to pin the correct-table behavior.

> Gap 3 — the parent is None fallback (subfields.add("")). Minor / defensive path, but it's currently uncovered. Fine to skip if it's hard to trigger through real SQL, just calling it out.

leaving it uncovered rather than adding a direct-call test for a guard.

> Performance concern. _leaf_column_subfields runs once per leaf and calls column.sql(comments=False) on every column returned by find_all_in_scope. Since multiple leaves share the same parent, that re-scan + re-serialize gets redone per leaf. For one select expression with ~M column occurrences producing ~L leaves, that's roughly L × M .sql() calls — quadratic in the width of a single select. The old code was O(1) per leaf: it just read the precomputed node.subfield attribute sqlglot filled in during the build. It's invisible for normal queries, but wide generated SELECTs (big CASE/COALESCE/function expressions) are exactly where it'd bite.

Good catch — it was genuinely quadratic. Fixed by building the leaf-name → subfields map once per parent and reusing it across that parent's leaves, instead of rescanning + re-serializing per leaf.

aviraj-gour:
Flagging a limitation:

This PR is about removing the sqlglot patch, not about improving subfield lineage coverage — so I want to be explicit about one limitation that comes with the current approach.

Subfield-level lineage isn't always fine-grained. In some cases it coarsens to the base column (`e.g. asset_id <- widget instead of widget.asset.id`). This shows up in two scenarios, both now covered by tests:

`test_select_struct_subfield_through_cte_passthrough` — the struct column passes through a CTE unchanged and the subfield access happens in the outer query.
`test_join_struct_subfields_shared_base_name` — a join, where sqlglot inserts an intermediate widget AS widget projection.

Why it happens: we reconstruct the subfield from the leaf's parent node expression. That works when the subfield access sits in the select that produces the leaf (see test_select_struct_subfields_from_cte), but the parent doesn't always carry the information we need. When a struct column flows through an intermediate scope, the actual subfield access lives a scope above the leaf's immediate parent, so we fall back to the base column.

This is not a regression. The old patched code also derived subfields from the immediate expression that created the leaf — accurate cross-scope subfield lineage was never supported, before this PR or after it. Getting it fully right would need support on the sqlglot side. We raised it with them and it was rejected; it's not on their priority list. So this workaround is what we have, and the tests above pin the current behavior so it doesn't silently regress.

sgomezvillamor (approved):
All concerns addressed — LGTM! 🎉

- **Gap 1 (nested/CTE):** `test_select_struct_subfields_from_cte` recovers full subfield lineage through the CTE — the immediate-parent path is now covered.
- **Gap 2 (join, shared base name):** confirmed no cross-table mis-attribution; coarsens to the base column and each resolves to the correct table, pinned by `test_join_struct_subfields_shared_base_name`. Nice explicit write-up of the cross-scope coarsening limitation too.
- **Gap 3:** fine to leave the defensive `parent is None` guard uncovered.
- **Perf:** per-parent memoization drops the `.sql()` work from O(L×M) to O(M), and the existing struct goldens are untouched — so provably output-neutral.

Bonus points for dropping the now-dead `patchy` dep and `unified_diff` module. CI is green. Kudos on the thorough follow-up! 👏

---

PR #18318, opened 2026-07-10
fix(ingest/teradata): release file-backed resources when init fails during table/view discovery

NehaGslab:
## Summary

The Teradata source builds three temp-file-backed resources during `__init__`
(`_view_definitions`, `schema_resolver`, `aggregator`), plus a lazily-created
pooled engine and class/module-level caches. The pipeline only registers a source
for `close()` once `__init__` returns, so if discovery (`cache_tables_and_views()`)
raised, `close()` never ran and those resources leaked across sequential recipe
runs in the same process.

This PR guarantees cleanup on both the normal and the `__init__`-failure paths by
managing every teardown on a single `contextlib.ExitStack`, matching the pattern
already used in `snowflake_v2.py`:
- The three `Closeable` resources are registered via `enter_context(...)` at
  creation, so they are torn down in reverse order.
- The non-`Closeable` teardown (pooled-engine `dispose()`, class-level table
  caches, module-level schema LRU caches) rides the same stack via `callback(...)`.
- The `__init__` failure path collapses to `self._exit_stack.close(); raise`, and
  `close()` collapses to `super().close(); self._exit_stack.close()`.

`ExitStack` provides the guarantees we previously implemented by hand: it runs
every registered cleanup even if an earlier one raises, and it does not silently
drop the original exception. As a result, `close()` now propagates a teardown
error (chained onto any in-flight exception) instead of swallowing-and-logging it —
the same behavior as the Snowflake source.

## Testing
Reworked the unit tests in tests/unit/test_teradata_source.py:
- `test_close_cleanup` — verifies `close()` releases all three resources, disposes the pooled engine (dropping the handle), and clears the table + schema LRU caches.
- `test_init_releases_resources_when_discovery_fails` — verifies an `__init__` discovery failure unwinds the stack (all resources closed, caches cleared) and re-raises the original error.
- Removed the four hand-rolled-guard tests that asserted "continue-on-failure" behavior at two separate call sites; that behavior is now provided by the stdlib `ExitStack`.

## Checklist
- [ ] Docs related to the changes have been added/updated (N/A — internal resource cleanup)
- [ ] Breaking change entry in Updating DataHub (N/A — no behavior change on the success path)


sgomezvillamor:
Thanks for tackling this — cleaning up on the `__init__` failure path is a real gap (since `close()` never runs if `__init__` throws), so the intent here is right.

My main concern is the mechanism. The three resources you're releasing — `_view_definitions` (`FileBackedDict`), `schema_resolver`, and `aggregator` — are all `Closeable` context managers (`ingestion/api/closeable.py`). The convention we use for guaranteeing these get released "in any case" is `contextlib.ExitStack`, registering each resource as it's constructed. See `snowflake_v2.py`:

```python
self._exit_stack = contextlib.ExitStack()
self.connection = self._exit_stack.enter_context(self.config.get_connection())
self.aggregator = self._exit_stack.enter_context(SqlParsingAggregator(...))
...
def close(self) -> None:
    super().close()
    self._exit_stack.close()   # unwinds everything, LIFO, keeps going if one raises
```

`ExitStack` already gives you the two properties this PR re-implements by hand: it runs *every* registered cleanup even if an earlier one raises, and it doesn't mask the exception. The non-`Closeable` teardown (clearing `_tables_cache`/`_table_creator_cache`, the three module-level LRU `cache_clear()`s, the pooled-engine `dispose()`) can ride on the same stack via `self._exit_stack.callback(...)`.

The bigger issue with the current approach is that it hand-rolls per-resource `try/except` and then **duplicates the whole teardown across two places** — the `__init__` except block and `close()`. That's a maintenance hazard: add a fourth file-backed resource later and you have to remember to update both sites, in the right order, with matching guards.

And the drift has already started, which is the tell that the duplication is a real problem and not hypothetical:

- `close()` disposes `self._pooled_engine`.
- The new `__init__` failure path does **not** dispose the pooled engine — it releases the three SQLite temp files and clears the caches, but skips the engine.

In this case it's probably harmless (discovery uses `get_metadata_engine()`, a fresh engine it disposes in its own `finally`; `_pooled_engine` is lazily created and likely doesn't exist yet at that point). But "probably harmless because of when the lazy engine happens to get created" is exactly the fragility the copy-paste introduces — it's a coincidence, not a guarantee.

Two ways to go, in order of preference:

1. **ExitStack, matching Snowflake.** Register `_view_definitions`, `schema_resolver`, `aggregator` via `enter_context(...)` at creation; register the cache-clears and pooled-engine dispose as `.callback(...)`. The init-failure path becomes just `except Exception: self._exit_stack.close(); raise`, and `close()` collapses to `self._exit_stack.close()` + `super().close()`. All the nested `try/except` and the duplication go away, and the guarantees come from the stdlib.

2. **Minimal DRY fix**, if a full refactor feels like too much for a fix PR: extract one private `_release_resources()` helper with the guarded teardown and call it from both `close()` and the `__init__` except block. Removes the duplication and the drift risk without touching resource construction, and closes the pooled-engine asymmetry for free.

Either way the added tests should shrink too, since you're no longer asserting the same guard behavior at two separate call sites.

---

sgomezvillamor (approved):
The ExitStack refactor is exactly the right direction — matches the Snowflake pattern, drops the duplicated teardown, and the pooled-engine cleanup is now consistent across the normal `close()` and the failed-`__init__` paths. Guarding construction (not just discovery) is a nice touch, actually more robust than Snowflake here. Approving.

One minor thing worth a `try/finally`, not a blocker:

⚠️ In `close()`, `super().close()` runs first and is unguarded, so if it raises (the failed-views summary / stateful-checkpoint commit can do I/O), `self._exit_stack.close()` is skipped and the temp SQLite files leak — the exact cleanup this PR exists to guarantee. Wrapping it keeps both:

```python
try:
    super().close()
finally:
    self._exit_stack.close()
```

---

PR #18319, opened 2026-07-10
fix(ingest/mysql): strip QueuePool-only options from usage NullPool engine

acrylJonny:
## Summary

- When `include_usage_statistics` and profiling are both enabled, the pipeline aborts with `TypeError: Invalid argument(s) 'max_overflow' sent to create_engine(), using configuration MySQLDialect_pymysql/NullPool/Engine`.
- Cause: the base `SQLAlchemySource._add_default_options` injects `max_overflow` (= `profiling.max_workers`) into `config.options` whenever profiling is enabled. `MySQLSource._usage_connection` forces `poolclass=NullPool` for its one-shot query-history fetch but still spread `config.options`, and `NullPool` rejects QueuePool sizing arguments.
- Fix: drop the QueuePool-only options (`pool_size`, `max_overflow`, `pool_timeout`, `pool_use_lifo`) before constructing the ephemeral `NullPool` usage engine. The long-lived inspection/profiling engines are unaffected and keep those options.

## Test plan

- Added `test_usage_connection_strips_queuepool_options_for_nullpool` asserting the usage engine is built with `NullPool` and none of the QueuePool sizing options.
- `ruff`, `ruff format`, `mypy`, and all `tests/unit/test_mysql_usage.py` tests pass.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/sql/mysql.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/sql/mysql.py
@@ -61,9 +61,11 @@ _SYSTEM_SCHEMAS = frozenset(
     {"information_schema", "performance_schema", "mysql", "sys"}
 )
 
-# QueuePool-only sizing options. SQLAlchemySource._add_default_options injects
-# `max_overflow` into `options` whenever profiling is enabled, but NullPool
-# rejects these, so they must be dropped from the ephemeral usage engine.
+# QueuePool-only sizing options that NullPool rejects, so they must be dropped
+# from the ephemeral usage engine (which forces NullPool). DataHub itself only
+# auto-injects `max_overflow` (SQLAlchemySource._add_default_options, when
+# profiling is enabled); the other three are stripped defensively in case a user
+# sets them via `config.options`.
 _QUEUE_POOL_ONLY_OPTIONS = frozenset(
     {"pool_size", "max_overflow", "pool_timeout", "pool_use_lifo"}
 )
@@ -474,8 +476,8 @@ class MySQLSource(TwoTierSQLAlchemySource):
         # NullPool + dispose() so this one-shot fetch never leaves connections
         # open. poolclass is forced last so a pooled class in options (intended
         # for the long-lived inspection engine) can't silently re-pool this
-        # ephemeral engine. QueuePool sizing options are stripped because
-        # NullPool rejects them (profiling injects max_overflow into options).
+        # ephemeral engine; QueuePool-only options are dropped (see
+        # _QUEUE_POOL_ONLY_OPTIONS) because NullPool rejects them.
         usage_options = {
             key: value
             for key, value in self.config.options.items()
--- a/metadata-ingestion/src/datahub/ingestion/source/sql/mysql.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/sql/mysql.py
@@ -61,6 +61,13 @@ _SYSTEM_SCHEMAS = frozenset(
     {"information_schema", "performance_schema", "mysql", "sys"}
 )
 
+# QueuePool-only sizing options. SQLAlchemySource._add_default_options injects
+# `max_overflow` into `options` whenever profiling is enabled, but NullPool
+# rejects these, so they must be dropped from the ephemeral usage engine.
+_QUEUE_POOL_ONLY_OPTIONS = frozenset(
+    {"pool_size", "max_overflow", "pool_timeout", "pool_use_lifo"}
+)
+
 # One row per normalized statement: DIGEST_TEXT has literals stripped to `?`,
 # COUNT_STAR counts executions since the last reset, LAST_SEEN is the most recent.
 # Low-overhead query history when performance_schema is on (vs. the general log).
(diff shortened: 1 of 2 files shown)


metadata-ingestion/tests/unit/test_mysql_usage.py
```diff
@@ -95,6 +97,23 @@ def test_usage_connection_pins_utc_and_disposes_engine(mock_create_engine):
...
+    # must not reach the ephemeral usage engine (else create_engine raises
+    # TypeError). Simulate that injected state directly.
+    mock_create_engine.return_value = _patch_rows([])
+
+    source = _source(options={"max_overflow": 10, "pool_size": 5})
+    list(source._fetch_performance_schema_queries())
+
+    kwargs = mock_create_engine.call_args.kwargs
+    assert kwargs["poolclass"] is NullPool
+    for option in ("pool_size", "max_overflow", "pool_timeout", "pool_use_lifo"):
```
treff7es: Two of these four assertions are vacuous: `options` only seeds `max_overflow` and `pool_size` (line 108), so `pool_timeout` and `pool_use_lifo` are never in `kwargs` regardless of whether the strip loop runs — their `not in` checks pass unconditionally. Either seed all four into `options` (preferred — exercises the whole `_QUEUE_POOL_ONLY_OPTIONS` set) or assert only the two that are present.

metadata-ingestion/tests/unit/test_mysql_usage.py
```diff
@@ -95,6 +97,23 @@ def test_usage_connection_pins_utc_and_disposes_engine(mock_create_engine):
     engine.dispose.assert_called_once()
 
 
+@patch("datahub.ingestion.source.sql.mysql.create_engine")
+def test_usage_connection_strips_queuepool_options_for_nullpool(mock_create_engine):
```
treff7es: This patches `create_engine`, so it asserts the kwargs the source *intends* to pass, not that they're valid for `NullPool` — it wouldn't catch SQLAlchemy renaming/adding a QueuePool-only option (prod breaks, test stays green). `create_engine` validates these kwargs eagerly *without* opening a connection, so an unmocked test is nearly free and actually reproduces the bug:

```python
# negative: the exact failure this PR fixes
with pytest.raises(TypeError):
    create_engine("mysql+pymysql://u:p@h/db", poolclass=NullPool, max_overflow=10)
# positive: the usage engine builds cleanly once the options are stripped
```

Consider adding it alongside (or in place of) the mock-level assertions.

metadata-ingestion/tests/unit/test_mysql_usage.py
```diff
@@ -95,6 +97,23 @@ def test_usage_connection_pins_utc_and_disposes_engine(mock_create_engine):
...
+    # When profiling is enabled, _add_default_options injects QueuePool sizing
```
treff7es: Optional: the first three sentences re-explain the same mechanism already documented on `_QUEUE_POOL_ONLY_OPTIONS`. The genuinely useful, non-duplicated part is the last sentence (*why* you hand-set `options` instead of enabling profiling). Consider trimming to just that.

metadata-ingestion/src/datahub/ingestion/source/sql/mysql.py
```diff
@@ -61,6 +61,13 @@
     {"information_schema", "performance_schema", "mysql", "sys"}
 )
 
+# QueuePool-only sizing options. SQLAlchemySource._add_default_options injects
```
treff7es: Minor: "injects `max_overflow` ... but NullPool rejects **these**" slightly blurs that only `max_overflow` is auto-injected by DataHub — the other three are defensively included in case a user sets them via `config.options`. A clause distinguishing the auto-injected one from the full defensive set would sharpen it. (Also, this same rationale is repeated at the `_usage_connection` call site and in the test — this constant is the natural single home for it.)

treff7es (approved):

metadata-ingestion/tests/unit/test_mysql_usage.py
acrylJonny: Done — the test now seeds all four QueuePool-only options and validates via the real create_engine, so no assertion is vacuous.

metadata-ingestion/tests/unit/test_mysql_usage.py
acrylJonny: Good call — replaced the mock-only assertions with the unmocked approach: a negative `pytest.raises(TypeError)` for `NullPool` + `max_overflow`, plus delegating to the real `create_engine` so its eager kwarg validation runs on exactly what the source builds (returns a mock to avoid a real connection). This now catches SQLAlchemy renaming/adding a QueuePool-only option.

metadata-ingestion/tests/unit/test_mysql_usage.py
acrylJonny: Done — trimmed to just the non-duplicated rationale (why options are hand-set instead of enabling profiling).

metadata-ingestion/src/datahub/ingestion/source/sql/mysql.py
acrylJonny: Done — reworded so only `max_overflow` is called out as DataHub-injected (via `_add_default_options`), with the other three stripped defensively for user-supplied `config.options`. Kept this rationale solely on the constant and reduced the call site to a reference.

PR #18326, opened 2026-07-10
perf(ingestion/dremio): make system-table fetch batch size configurable

acrylJonny:
## Summary

Dremio's system tables (`INFORMATION_SCHEMA` / `SYS`) can be slow to query, and the paged dataset/column/view-definition/query fetches used a hardcoded `LIMIT/OFFSET` page size of `1000`. On large catalogs this issued far more round-trips than necessary, and each page waits on a slow Dremio job.

This PR exposes the page size as a `batch_size` config option and raises the default.

- **New `batch_size` config** (default **10000**, was hardcoded `1000`) controlling how many rows each paged fetch requests. Applies to datasets, columns, view definitions, and query extraction.
- **`0` means "as much as possible"** → fetches pages at a safety ceiling of `1,000,000` rows.
- **Values above the ceiling are clamped** to `1,000,000`, reported as a one-time warning instead of happening silently.

### OOM tradeoff (why the old default was 1000)

The previous hardcoded `1000` carried a `# prevent OOM` comment: it was chosen specifically because Dremio can raise `OversizedAllocationException` / crash mid-iteration when materializing a large per-job result set (there's an existing test covering that OOM-mid-iteration path). Raising the default 10x deliberately trades some of that OOM headroom for far fewer round-trips against slow system tables — the right default for most catalogs, and now fully configurable. The `batch_size` description tells operators to lower the value if they hit Dremio OOM during extraction.

### About the ceiling

`DREMIO_MAX_JOB_OUTPUT_ROWS = 1_000_000` is a **defensive ceiling** on the `LIMIT` we ever issue, not an asserted hard API limit. Dremio's REST job-results endpoint paginates _without_ a documented total-row cap (it streams 500-row pages until a short page — see the [Job Results docs](https://docs.dremio.com/current/reference/api/job/job-results/)), so the ceiling simply keeps a single job's result set from growing unboundedly and keeps `batch_size=0` well-defined.

**Honest scope of the runtime warning:** `LIMIT/OFFSET` short-page pagination treats "page shorter than the requested limit" as end-of-data and cannot distinguish that from a server-truncated page. The connector emits a one-time `report.warning` when a page comes back filled exactly to the ceiling — this is a **boundary/perf nudge**, not a truncation detector. The genuinely dangerous case (Dremio capping a job's output _below_ the requested LIMIT) would surface as a _short_ page and is an accepted, undetectable limitation of this pagination scheme; the safe default of 10000 keeps ingestion far from that boundary. This is documented at the call site so the warning isn't mistaken for full truncation coverage.

## Test plan

- [x] `test_batch_size_resolves_to_chunk_size` — `10000→10000`, `500→500`, `0→1,000,000`, `5,000,000→1,000,000` (clamp).
- [x] `test_batch_size_over_ceiling_warns` / `test_batch_size_at_ceiling_does_not_warn` — locks the intentional `>` (clamp) vs `>=` (page-cap) boundary.
- [x] `test_warn_if_page_at_job_cap_fires_once` / `_silent_below_ceiling` — page-cap nudge fires once at the ceiling, silent below it.
- [x] `test_get_queries_chunked_stops_on_short_nonempty_final_page` — terminates on a non-empty short final page without an extra probe or a false warning.
- [x] Updated existing pagination assertions from `LIMIT 1000` to the new `LIMIT 10000` default.
- [x] `./gradlew :metadata-ingestion:lint` (ruff + mypy) passes.
- [x] All Dremio unit tests pass.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_api.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/dremio/dremio_api.py
@@ -41,9 +41,14 @@ DREMIO_SYSTEM_TABLES_PATTERN = [
     r"^sys\..*",
 ]
 
-# A single Dremio REST API job returns at most this many rows, so it is also the
-# largest usable LIMIT per paged fetch. Larger LIMITs would silently return only
-# this many rows, breaking the "short page means last page" pagination check.
+# Defensive ceiling on the LIMIT we ever issue per page. Dremio's REST job-results
+# endpoint paginates without a documented total-row cap (it streams 500-row pages
+# until a short page: https://docs.dremio.com/current/reference/api/job/job-results/),
+# so this is a safety bound rather than a hard API limit. It keeps the LIMIT — and
+# therefore a single job's result set — from growing unboundedly, which keeps the
+# "short page means last page" pagination check meaningful. If a page ever comes
+# back exactly this size we warn, since at that point a genuinely full page and a
+# server-truncated one are indistinguishable.
 DREMIO_MAX_JOB_OUTPUT_ROWS = 1_000_000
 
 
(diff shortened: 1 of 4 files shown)


max-datahub (approved):
LGTM. Clamp/warn boundary is clean — `>` for the clamp vs `>=` for the page-cap nudge is deliberate and well-tested, and the once-per-run latch is right.

One ask on the description: the old default carried a `# prevent OOM` comment, and there's an existing test that exists specifically because Dremio can OOM mid-iteration. Bumping the default 10x raises the LIMIT Dremio has to materialize per job, so it does trade a bit of OOM headroom for fewer round-trips. It's the right tradeoff and it's configurable now, but worth calling out that the original default was chosen for exactly that reason.

acrylJonny:
Good call — addressed. The `batch_size` description now explicitly notes the tradeoff: the old `1000` default was chosen to prevent Dremio OOM (`OversizedAllocationException` mid-iteration), and raising the default trades some of that headroom for fewer round-trips, so operators hitting OOM should lower `batch_size`. Also added an "OOM tradeoff" section to the PR description. (commit 3f3c065)

PR #18328, opened 2026-07-10
fix(ingest/mysql): keep usage lineage from emitting phantom datasets

acrylJonny:
## Summary

When `include_usage_statistics` is enabled, query-history usage/lineage attributed metadata to tables that were never ingested as datasets, producing column-less phantom entities and ignoring `database_pattern`:

- **Filtered-out databases** — the fetch-time filters only checked each query's *default schema*, not tables *referenced inside* the SQL. A query in an allowed database could resurrect a table from a database excluded by `database_pattern`.
- **Three-layer URNs on two-tier sources** — a mis-quoted `` `db.table` `` reference (a single backtick-quoted identifier with an embedded dot) parses as one table name, so the two-tier `default_schema` is prepended and the URN doubles the database (`db.db.table`).

## Fix

Mirror the Redshift pattern by wiring two callbacks into the usage `SqlParsingAggregator` (inherited automatically by MariaDB):

- `is_allowed_table` — drops system schemas and databases excluded by `database_pattern`, applied to every referenced table.
- `is_temp_table` — treats any table not in `discovered_datasets` (the tables we actually ingested) as temp. The aggregator resolves lineage *through* temp tables but never emits them, so genuine temp tables, filtered-out databases, and the phantom `db.db.table` all stop producing entities.

Also override `_save_schema_to_resolver` so `discovered_datasets` is populated whenever usage is enabled (previously gated on view lineage). It is filled during ingestion, before the usage phase runs.

## Test plan

- [x] `tests/unit/test_mysql_usage.py` — added coverage for both callbacks, a regression test asserting the phantom `db.db.table` URN is never emitted, and one asserting a filtered-out database never appears in emitted URNs. 20 passed.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/sql/mysql.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/sql/mysql.py
@@ -456,41 +456,31 @@ class MySQLSource(TwoTierSQLAlchemySource):
             generate_operations=False,
             usage_config=self.config.usage,
             eager_graph_load=False,
-            # Keep query-history lineage/usage in step with what we actually
-            # ingested. Without these, referenced tables that we never emitted as
-            # datasets (other databases, temp tables, mis-quoted identifiers) turn
-            # into phantom, column-less entities. See _is_allowed_table /
-            # _is_temp_table.
             is_allowed_table=self._is_allowed_table,
             is_temp_table=self._is_temp_table,
         )
 
     def _save_schema_to_resolver(self) -> bool:
-        # is_temp_table distinguishes real tables from phantom ones by membership
-        # in discovered_datasets, which is only populated when schemas are saved to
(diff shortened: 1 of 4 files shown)


max-datahub (approved):
LGTM. Root-cause fix in the right place — the temp-table check (not in `discovered_datasets`) is what actually kills the phantom `db.db.table`, since the aggregator runs `is_temp_table` before `is_allowed_table`. The `_process_view` no-schema branch is the correct asymmetry (tables already land in discovered_datasets). Good tests with co-assertions guarding against vacuous passes.

One thing worth noting in the description: the `is_allowed_table`/database_pattern path is mostly belt-and-suspenders next to the temp check — a filtered-out DB is never ingested, so it's never in discovered_datasets anyway. Fine to keep (mirrors Redshift), just slightly oversold as an independent fix.

PR #18332, opened 2026-07-10
perf(graphql): batch Dataset.statsSummary via cached DataLoader

gdatahub:
### Summary

`Dataset.statsSummary` resolved **once per dataset**, and each call fanned out into several Elasticsearch aggregations over the `datasetUsageStatistics` timeseries aspect. On any list view (search results, container contents, browse), this is an **N+1 explosion** against ES.

This adds `DatasetStatsSummaryBatchLoader`, a graphql-java `DataLoader` that collapses the per-dataset fan-out within a single GraphQL request into **two batched `getAggregatedStats` queries**, backed by a shared Caffeine cache.

### What changed

- **Batched aggregation** — for the datasets in a request, `queryCountLast30Days`, `uniqueUserCountLast30Days`, and `topUsersLast30Days` are computed with two batched ES aggregations (outer-terms bucket keyed on `urn`) instead of N × several.
- **Caching** — a shared, bounded Caffeine cache (24h TTL matching the per-URN `UsageClientCache`, since usage stats update daily). Cache hits are served directly; only **misses** reach ES, and all misses in a request are batched together. So a mostly-warm request queries ES only for the newly-seen datasets.
- **Parallel authorization** — per-URN `VIEW_DATASET_USAGE` auth is enforced inside the loader, run in parallel and off graphql-java's field-completion path. Unauthorized datasets resolve to `null` and are never fetched (auth runs before any cache/ES access).
- **Shared query definitions** — aggregation specs, grouping buckets, time window, and filter are extracted into `UsageServiceUtil`/`TimeseriesUtils` and shared with the per-URN path, so batched and per-entity results are **identical**.
- **Feature-flagged** — `datasetStatsSummaryBatchLoadEnabled` (`DATASET_STATS_SUMMARY_BATCH_LOAD_ENABLED`), **default on**. Disabling reverts to the per-dataset query path with no behavior change.

### Correctness

No functional change. Benchmarked with a single `entities(urns)` request over 1–500 datasets against seeded usage stats: batched results are **byte-identical** to the per-URN path (query count, unique users, and top-5 users in the same ranked order) across all sizes, in cold, warm, and partial-cache states.

### Performance (single request, seeded local instance)

| Regime | vs per-URN path |
| --- | --- |
| Cold cache (first load / N+1) | **~2.3× faster** at n=500 |
| Partial-miss (250 cached + 250 fresh) | **~1.5× faster** |
| Fully warm (100% cache hit) | on par |

The warm case was initially ~25% slower because the per-URN auth check ran serially on graphql-java's completion path; moving it into the loader's parallel path closed that gap (confirmed by thread-dump profiling).

### Test plan

- Unit tests: `DatasetStatsSummaryBatchLoaderTest` — batching + per-URN assembly, top-user capping/ordering, empty results, cache-hit-no-requery, unauthorized→null.
- `PropertiesCollectorConfigurationTest` updated for the new flag (security classification guardrail).
- End-to-end parity + latency benchmarked locally (flag on vs off) over 1–500 datasets.

### Checklist

- [x] Tests added/updated
- [ ] Docs added/updated — n/a (internal perf change behind a default-off flag)
- [ ] Breaking changes in `updating-datahub.md` — n/a (no breaking change; default-off flag)

### DIFF
--- a/metadata-service/configuration/src/main/resources/application.yaml
+++ b/metadata-service/configuration/src/main/resources/application.yaml
@@ -1527,7 +1527,7 @@ featureFlags:
   i18nEnabled: ${I18N_ENABLED:true} # If enabled, internationalization (i18n) features are active
   timeseriesAspectBatchLoadEnabled: ${TIMESERIES_ASPECT_BATCH_LOAD_ENABLED:true} # If enabled, TimeSeriesAspectResolver uses the DataLoader batch path; disable to revert to single-URN queries
   timeseriesAspectAggBatchLoadEnabled: ${TIMESERIES_ASPECT_AGG_BATCH_LOAD_ENABLED:true} # If enabled, Dashboard.statsSummary uses the batch aggregation DataLoader; disable to revert to per-URN queries
-  datasetStatsSummaryBatchLoadEnabled: ${DATASET_STATS_SUMMARY_BATCH_LOAD_ENABLED:false} # If enabled, Dataset.statsSummary uses the batch aggregation DataLoader (batches usage aggregations across a request); disable to revert to per-dataset queries
+  datasetStatsSummaryBatchLoadEnabled: ${DATASET_STATS_SUMMARY_BATCH_LOAD_ENABLED:true} # If enabled, Dataset.statsSummary uses the batch aggregation DataLoader (batches usage aggregations across a request); disable to revert to per-dataset queries
   createSchemaVersionIndex: ${CREATE_SCHEMA_VERSION_INDEX:${ZDU_STAGE_10:false}} # Creates schemaVersionIndex on metadata_aspect_v2 for ZDU schema version queries
   aspectMigrationMutatorEnabled: ${ASPECT_MIGRATION_MUTATOR_ENABLED:${ZDU_STAGE_20:false}} # Populates AspectMigrationMutatorChain with registered mutators
   browserTracingEnabled: ${BROWSER_TRACING_ENABLED:false} # If enabled, the React app emits browser (RUM) OpenTelemetry spans, exported via the frontend /otel/v1/traces proxy and correlated with backend traces.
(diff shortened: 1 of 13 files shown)


david-leifker (approved):

metadata-service/configuration/src/main/resources/application.yaml
```diff
@@ -1516,6 +1516,7 @@ featureFlags:
   i18nEnabled: ${I18N_ENABLED:true} # If enabled, internationalization (i18n) features are active
   timeseriesAspectBatchLoadEnabled: ${TIMESERIES_ASPECT_BATCH_LOAD_ENABLED:true} # If enabled, TimeSeriesAspectResolver uses the DataLoader batch path; disable to revert to single-URN queries
   timeseriesAspectAggBatchLoadEnabled: ${TIMESERIES_ASPECT_AGG_BATCH_LOAD_ENABLED:true} # If enabled, Dashboard.statsSummary uses the batch aggregation DataLoader; disable to revert to per-URN queries
+  datasetStatsSummaryBatchLoadEnabled: ${DATASET_STATS_SUMMARY_BATCH_LOAD_ENABLED:false} # If enabled, Dataset.statsSummary uses the batch aggregation DataLoader (batches usage aggregations across a request); disable to revert to per-dataset queries
```
chakru-r: Why default of false? This is only to switch to an equivalent but more performant alternative, so flag only helps if there was an unintended regression. This can default to false.  Also, smoke tests were not testing the batch loading

chakru-r (approved):
Please ensure smoke tests run with this flag enabled to confirm no regressions before merging

metadata-service/configuration/src/main/resources/application.yaml
gdatahub: I have enabled it by default now. I set it to false by default because I wasn't sure.

gdatahub:
before

after

PR #18341, opened 2026-07-11
feat(agent-context): add incident tools (list/raise/resolve) to mcp_tools

Choppaaahh:
# feat(agent-context): add incident tools (list/raise/resolve) to mcp_tools

Agents using the Agent Context Kit can read an asset's health but have no way to list, raise, or resolve incidents — so an agent that diagnoses a problem can't act on it. This adds incident tools to `mcp_tools`, following the existing module conventions (context-resolved graph client, `execute_graphql`, summary flattening via `clean_gql_response`).

- `list_incidents(urn, state="ACTIVE", start, count)` — incidents on an asset (Dataset/DataJob/DataFlow/Dashboard/Chart), filterable by state; results respect the kit's token-budget helpers (`truncate_descriptions` + `_select_results_within_budget`)
- `raise_incident(urn, title, description, incident_type="OPERATIONAL", priority=None)` — raises an ACTIVE incident via `raiseIncident`
- `resolve_incident(incident_urn, message)` — transitions to RESOLVED with a resolution message via `updateIncidentStatus`

Registered in `mcp_tools.__init__` and both builders (read tool in the base list; raise/resolve gated behind `include_mutations`, like the other mutations). The two mutation tools also honor an `INCIDENT_TOOLS_ENABLED` env kill-switch (default enabled), analogous to `SAVE_DOCUMENT_TOOL_ENABLED` — enforced in-package (rather than at MCP-server registration) so the flag also covers the LangChain/ADK builder paths. `resolve_incident` uses the `incidentStatus` subfields as declared in the current OSS GraphQL schema (no Cloud-only field markers needed). Unit tests mirror the existing mock patterns (`test_incidents.py`, 32 tests); full package suite passes (631 passed, 3 xfailed) and new files are ruff-clean.

These tools were built and battle-tested against a live DataHub instance during a DataHub hackathon submission (Agents That Do Real Work track).


AdrianMachado:
Thanks for the contribution - and apologies for the wait. I need a bit of time to review this but will get back to you with a review this week

AdrianMachado (approved):
Looks good to me

datahub-agent-context/src/datahub_agent_context/mcp_tools/incidents.py
```diff
@@ -0,0 +1,476 @@
...
+    truncate_descriptions,
+)
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_PAGE_SIZE = 20
+MAX_PAGE_SIZE = 50
+
+# Entity types that expose the `incidents` field in the GraphQL schema.
+SUPPORTED_ENTITY_TYPES = ("Dataset", "DataJob", "DataFlow", "Dashboard", "Chart")
```
AdrianMachado: I don't like maintaining static lists like this, please remove

datahub-agent-context/src/datahub_agent_context/mcp_tools/incidents.py
```diff
@@ -0,0 +1,476 @@
...
+        ... on Dashboard {
+            incidents(state: $state, start: $start, count: $count) {
+                ...entityIncidentsResultFields
+            }
+        }
+        ... on Chart {
+            incidents(state: $state, start: $start, count: $count) {
+                ...entityIncidentsResultFields
+            }
+        }
```
AdrianMachado, line 75: We recently added support for more entities, see issue: #19322 - if you can add in the new ones, that would be great
I don't quite know how to make this more maintainable of a list, but I also don't want to make that your problem

datahub-agent-context/src/datahub_agent_context/mcp_tools/incidents.py
```diff
@@ -0,0 +1,476 @@
...
+                incident_type="FIELD",
+                priority="HIGH",
+            )
+    """
+    if not _is_incident_tools_enabled():
+        return {
+            "success": False,
+            "incidentUrn": None,
+            "urn": urn,
+            "message": "Incident tools are disabled via INCIDENT_TOOLS_ENABLED",
```
AdrianMachado: Ideally use constants or builders for these reused error messages and errors

datahub-agent-context/src/datahub_agent_context/mcp_tools/incidents.py
```diff
@@ -0,0 +1,476 @@
...
+        "source": source.get("type"),
+        "createdMillis": created.get("time"),
+        "createdBy": created.get("actor"),
+    }
+    return summary
+
+
+def list_incidents(
+    urn: str,
+    state: Optional[IncidentState] = "ACTIVE",
```
AdrianMachado: nit: also use a const for the default, like you did for page size

datahub-agent-context/src/datahub_agent_context/mcp_tools/incidents.py
```diff
@@ -0,0 +1,476 @@
...
+            graph,
+            query=RAISE_INCIDENT_MUTATION,
+            variables={"input": incident_input},
+            operation_name="raiseIncident",
+        )
+
+        incident_urn = result.get("raiseIncident")
+        if not incident_urn:
+            raise RuntimeError(
+                f"Failed to raise incident on {urn} - no incident urn returned"
```
AdrianMachado: Maybe too much info on implementation detail

AdrianMachado (approved):

AdrianMachado:
**Follow-up (gate 7 gap after merge).**

#19115 landed `SchemaFieldEntity` on SDL + resolver. MCP `LIST_INCIDENTS_QUERY` still lacks `... on SchemaFieldEntity`, so agents will silently get empty lists for field URNs.

Small follow-up appreciated: add that fragment (and later `MLFeatureTable` when that PR opens).

PR #18342, opened 2026-07-11
fix(ui): Domains list reliably shows more than 25 entries via infinite scroll

### LINKED ISSUE #18015: Domains list fails to reliably show more than 25 entries
**Describe the bug**
On the Domains page, entities may be missing from the list. The issue starts becoming visible at 25+ domains.

**To Reproduce**
1. Ensure your DataHub instance has 25 or more domains.
2. Navigate to **Domains**.
3. Count the entities shown in the list and compare to the total you know to exist.
4. See that the list count is lower than the number of domains created.

**Expected behavior**
All created domains should appear in the list view.

**Desktop**
- OS: Windows, MacOs
- Browser: Edge
- Version: 1.6.0

**Additional context**
- Missing domains are fully intact, direct URL navigation works.
- The situation can be improved by adjusting to f.e. `export const DOMAIN_COUNT = 100;` in `datahub-web-react/src/app/domainV2/useScrollDomains.ts`

mvanhorn:
## Summary

On the Domains page (`/domains`), entities went missing from the list once an instance had 25 or more domains. The missing domains were intact and reachable by direct URL, so this was a list-rendering problem rather than data loss. This change makes the infinite-scroll pagination load every domain reliably, whatever the total count.

## Why this matters

Users on instances with more than 25 domains simply could not see part of their domain hierarchy in the normal browsing UI, which is a real navigability and data-visibility problem even though the underlying data was fine. The reporter narrowed it down by finding that raising the `DOMAIN_COUNT` constant helped, which points squarely at the scroll pagination rather than the query or the data.

The cause is a race in how the next page is triggered. `useScrollDomains` fetches a page at a time and advances to the next page from an effect gated on whether a sentinel element is in view. That sentinel was only mounted while `domains.length > 0`, so during the brief window when a page was loading it competed with the loading spinner and could unmount, detaching the intersection observer. Even when it stayed mounted, its trigger point sat right at the visibility threshold, so once the first page settled the observer often never re-fired and no further pages loaded. Bumping `DOMAIN_COUNT` only pushes that failure threshold higher rather than fixing the paging, so I left the constant alone.

## What changed

The fix keeps the observer reliable in two small ways. In `RootDomains.tsx` the sentinel now renders unconditionally at the tail of the list instead of only when domains already exist, so the observer stays attached across page loads. In `useScrollDomains.ts` the `useInView` call gains a `rootMargin` of `200px`, so the next page prefetches slightly before the sentinel is fully visible and the window where `inView` settled just outside the threshold and never re-triggered is closed. Together these let pagination run to exhaustion for large domain sets.

## Testing

New tests in `useScrollDomains.test.tsx` cover the paths that matter here: that the hook advances and appends the next page when a `nextScrollId` is present and the sentinel is in view so the list grows past 25 entries; that it stops paging when the result set is smaller than a full page; that duplicate URNs across pages stay deduped through the existing guard; and that a query error surfaces without wedging pagination into an infinite fetch loop.

## Checklist

Fixes #18015

### DIFF
--- a/datahub-web-react/src/app/domainV2/__tests__/useScrollDomains.test.tsx
+++ b/datahub-web-react/src/app/domainV2/__tests__/useScrollDomains.test.tsx
@@ -59,27 +59,26 @@ const createWrapper = ({ children }: { children: React.ReactNode }) => (
     </MockedProvider>
 );
 
-const createMockDomain = (index: number): ListDomainFragment =>
-    ({
-        urn: `urn:li:domain:${index}`,
-        id: `${index}`,
-        type: EntityType.Domain,
-        properties: {
-            name: `Domain ${index}`,
-            description: null,
-            customProperties: [],
-        },
-        parentDomains: null,
-        ownership: null,
-        privileges: null,
-        displayProperties: null,
-        deprecation: null,
-        entities: null,
-        dataProducts: null,
-        applicationsInDomain: null,
-        children: null,
-        institutionalMemory: null,
-    });
+const createMockDomain = (index: number): ListDomainFragment => ({
+    urn: `urn:li:domain:${index}`,
+    id: `${index}`,
+    type: EntityType.Domain,
+    properties: {
+        name: `Domain ${index}`,
+        description: null,
+        customProperties: [],
+    },
+    parentDomains: null,
+    ownership: null,
+    privileges: null,
+    displayProperties: null,
+    deprecation: null,
+    entities: null,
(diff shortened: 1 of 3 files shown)


mvanhorn:
Fixed the datahub-web-react-lint failure in 2c632d28 - the new test file wasn't formatted with the repo prettier config. Verified locally with the repo's own prettier (yarn install + npx prettier --check) before pushing.

PR #18347, opened 2026-07-11
feat(ingest/s3): profile data-lake files without PySpark or Deequ

naeemdh:
## Summary

Replaces the PySpark + AWS Deequ based data-lake profiler for S3/GCS files with a lightweight, pure-Python profiler built on `pyarrow` + `fastavro` + Apache DataSketches. This removes the JVM/Spark dependency from the `s3` and `gcs` ingestion extras entirely, and enables data profiling for GCS (previously unsupported).

## What the old thing did vs. what we built

**Before (PySpark + Deequ):**
To profile a Parquet/CSV/Avro file, DataHub would boot up an entire Apache Spark engine (a Java virtual machine, gigabytes of dependencies, JAR downloads) just to read one file and compute simple stats like min, max, mean, and "how many distinct values are in this column." Deequ is a library that runs *on top of* Spark to compute those stats.

**Now (pyarrow-based):**
We read the same files directly in Python with `pyarrow` (a lightweight, fast columnar reader — no JVM, no Spark) and compute the same stats ourselves by streaming through the data in chunks. Same inputs, same output shape, no heavyweight engine.

The big win: we deleted an enormous, awkward dependency and replaced it with a small pure-Python one.

## The uniqueCount / uniqueProportion difference

Counting **exactly** how many distinct values a column has is surprisingly expensive — to be exact, you have to remember every value you've ever seen, which blows up memory on big files. So **both** the old system and this one avoid that. Neither counts exactly.

Instead, both use a statistical trick called a **sketch** — a tiny data structure that *estimates* the distinct count using a fixed, small amount of memory (a few kilobytes), no matter if the file has 100 rows or 100 million. The tradeoff is the answer is approximate (typically within ~2%).

The catch: **the old system and this one use *different* sketch algorithms.**
- Deequ used one called **HyperLogLog**.
- We use one from Apache DataSketches called **CPC**.

They're both good, both approximate, and they're solving the same problem — but because they're different algorithms, they land on *slightly* different estimates for the same data.

## Why the golden files changed

That's exactly why the test golden files changed. For example, one column's estimated distinct count went from **9 → 10**:

- The actual true answer was **10** distinct values.
- Deequ's HyperLogLog estimated **9** (off by one).
- Our CPC sketch estimated **10** (correct here, but it's luck of the draw — on other columns ours might be the one that's slightly off).

Neither is "wrong" — they're two different approximations of the same thing. There is **no way** to make them produce byte-for-byte identical numbers, because that would require both tools to use the identical algorithm internally, which they don't.

The golden files were just frozen snapshots of what Deequ *happened* to output. Since we swapped the engine, we regenerated those snapshots to record what the new engine outputs. Most differences are these approximate-count values, plus one cosmetic thing — Deequ inserted a fake `"NullValue"` category into its frequency lists, which we dropped since null counts are already reported in their own dedicated field (`nullCount`/`nullProportion`).

Two golden changes are **correctness fixes**, not approximate-count drift:

- The `NPS…ARCN_Lakes_ChemistryData` CSV has quoted column names with embedded newlines (e.g. `"NH3-N \n(mg N/L)"`). The new reader (`newlines_in_values=True`) reads it correctly as 29 columns / 20 rows; the old Spark reader mangled it into 14 columns / 33 rows with header fragments leaking into the `Sampling Date` column as data. So this file's profile is now *right* where it was previously wrong.
- Files with a genuine mid-stream read/parse failure (e.g. duplicate empty-named columns) no longer emit a partial, polluted profile — the whole table profile is skipped with a warning instead. That's the intended honesty change from the error-handling review, not a count difference.

## One subtlety worth calling out

Deequ decides *which* stats to even compute for a column based on its "cardinality" (roughly: is this column low-variety like a status flag, or high-variety like an ID?). The old code had what looks like a bug — it fed the null-fraction (not the true unique proportion) into that decision. Rather than "fixing" it — which would silently change profiling output for existing users — this PR **deliberately replicates the same behavior** so results stay consistent for anyone already relying on them.

## Metrics parity

The new profiler reproduces the full metric set of the old one, gated by the same cardinality logic and the same `include_field_*` config flags:

- Exact: row/column/null counts, min, max, mean, stdev, and (for low-cardinality columns) distinct-value frequency tables.
- Approximate (DataSketches): unique count (CPC), median + quantiles + continuous histogram (KLL).

## Notable fixes made along the way

- **Ragged CSV rows**: pyarrow's CSV parser is stricter than Spark's about rows with a stray trailing delimiter (an extra empty field on every line). Added a fallback that reflows rows to the header width and retries, matching Spark's permissive behavior.
- **Nested JSON columns**: `list`/`struct`-typed columns can't be compared with `<`/`>`, which would crash min/max tracking; these are now correctly skipped (matching the old behavior of never surfacing min/max for nested types).

## Formats supported

Parquet, CSV, TSV, Avro, and JSON — on S3, GCS, and local filesystem. GCS reuses the same `smart_open` + boto3 path as S3 (GCS's S3-interoperability endpoint), so no GCS-specific reader code was needed; GCS data profiling is now enabled where it previously reported "Not supported".

## Packaging

The `s3` and `gcs` extras now depend on `datasketches` instead of `pyspark` + `pydeequ`. (The `abs`, `delta-lake`, and `excel` extras are out of scope and unchanged.)

## Why DataSketches (not Polars or Pandas)?

The goal here was to make profiling **lightweight**, so the driving constraint was keeping the dependency footprint minimal:

- **Minimal footprint.** The whole point was to drop the heavy PySpark + JVM stack. Polars and Pandas were both considered, but each is another sizable third-party package to take on — which works against that goal.
- **It provides exactly the estimators we need.** DataSketches gives us approximate distinct/unique counts (CPC) plus quantiles and histograms (KLL), using the same class of probabilistic estimation algorithms Deequ used on Spark (HyperLogLog). Exact stats — row/null counts, min/max/mean/stdev — are computed directly while streaming. Polars/Pandas would compute the cardinality stats *exactly*, which changes the output contract and doesn't stay memory-bounded on large files.
- **Fewest external dependencies.** Combined with `pyarrow` (already a transitive dependency), DataSketches keeps us to the bare minimum without pulling in a large new engine.

## Follow-up

PySpark/Deequ is **not fully removed from the codebase yet** — the `abs` and `delta-lake` extras still use the Spark-based profiler (`data_lake_profiling`). A dedicated follow-up PR will remove PySpark from the dependencies entirely once this one lands.

## Testing

- New unit tests for the accumulators, readers, and profiler.
- Existing S3/GCS unit tests pass unchanged.
- Full S3/GCS integration suite (including moto-mocked S3 and GCS ingestion, and the golden-file comparisons) passes.
- The `s3-slim`/`s3` no-PySpark install tests were updated to assert PySpark is no longer pulled in by either extra.
- `ruff` + `mypy` clean.

## Checklist

- [x] The PR conforms to DataHub's Contributing Guideline (particularly the Commit Message Format)
- [x] Tests added/updated
- [ ] Docs updated (follow-up)


metadata-ingestion/src/datahub/ingestion/source/data_lake_common/profiling/profiler.py
```diff
@@ -0,0 +1,286 @@
...
+
+
+def null_str(value: Any) -> Optional[str]:
+    return str(value) if value is not None else None
+
+
+class FileProfiler:
+    """
+    Profiles S3/GCS/local files (parquet/csv/tsv/avro/json) via streaming Arrow/fastavro
+    readers and Apache DataSketches, without pyspark or Deequ.
```
skrydal: As well as without a thousand other packages, I do not think listing them here is necessary.

metadata-ingestion/setup.py
```diff
@@ -775,9 +783,9 @@
     # S3 includes PySpark by default for profiling support (backward compatible)
     # Standard installation: pip install 'acryl-datahub[s3]' (with PySpark)
     # Lightweight installation: pip install 'acryl-datahub[s3-slim]' (no PySpark, no profiling)
-    "s3": {*s3_base, *data_lake_profiling},
+    "s3": {*s3_base, *file_profiling},
     "s3-slim": {*s3_base},
```
skrydal: Shouldn't we add file_profiling to `-slim` versions as well? Preferably we could do:
`"s3-slim": {*s3}`
The next step would be to remove `-slim` versions (along with CI/CD and image building references).

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/profiling/profiler.py
naeemdh: good. removed.

metadata-ingestion/setup.py
naeemdh: good callout. updated.

sgomezvillamor:
This need for a custom profiler may come up in other sources too, eg Kafka, DynamoDB, etc. Basically anything without SQL.

Worth thinking about whether we could generalize or extract the core aggregation logic so it can be reused, decoupled from the data access layer (which is file-based here).

`TableAccumulator` may be such abstraction, it may just moved somewhere else if so.

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/profiling/profiler.py
```diff
@@ -0,0 +1,286 @@
...
+    Cardinality (see `accumulators.Cardinality`) decides which fields a
+    column gets, mirroring the old Spark/Deequ profiler: low-cardinality
+    columns of any type get an exact distinctValueFrequencies table;
+    high-cardinality numeric/temporal columns get min/max/mean/median/stdev
+    plus approximate quantiles and a continuous histogram.
+    """
+
+    def __init__(
+        self,
+        aws_config: Optional[AwsConnectionConfig],
```
sgomezvillamor, line 83: are S3/GCS/local all supported with this `AwsConnectionConfig`?

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/profiling/profiler.py
```diff
@@ -0,0 +1,286 @@
...
+        ):
+            field_profile.distinctValueFrequencies = [
+                ValueFrequencyClass(value=value, frequency=frequency)
+                for value, frequency in column_stats.distinct_value_frequencies
+            ]
+
+        return field_profile
+
+    def get_table_profile(
+        self, table_data: Any, dataset_urn: str
```
sgomezvillamor: can type for `table_data` be narrowed down?
I see you are calling `full_path`, `display_name`, ...

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/profiling/profiler.py
naeemdh: Yes, all three. GCS is handled as S3 (GCSSource rewrites gs://→s3:// and uses an AwsConnectionConfig against GCS's S3-interop endpoint); local files skip aws_config, which is why it's Optional. Added a comment in _open_file to make that explicit.

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/profiling/profiler.py
naeemdh: Good call, narrowed it to TableData (via a TYPE_CHECKING import to avoid a circular import).

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/profiling/readers.py
```diff
@@ -0,0 +1,163 @@
...
+        reader = pa_csv.open_csv(
+            _reflow_ragged_rows(file_obj, delimiter),
+            parse_options=parse_options,
+            convert_options=convert_options,
+        )
+    return ColumnarSource(
+        columns=reader.schema.names,
+        column_kinds=_column_kinds_from_schema(reader.schema),
+        batches=reader,
+    )
```
sgomezvillamor: 🔴 Blocker: this reflow silently mangles data. `_reflow_ragged_rows` truncates wide rows (`row[:width]`) and pads short ones with `""`, but nothing reaches the report — the operator gets a clean-looking profile computed over altered data. The empty-named `""` column that now shows up in `golden_mces_multiple_files.json` is exactly this path firing.

Two problems beyond the missing signal:
1. `ArrowInvalid` isn't ragged-rows-specific — quoting/encoding/type errors trigger the same reflow and yield a bogus profile instead of an honest failure. We should only reflow when it's actually a row-width mismatch and re-raise otherwise.
2. `open_csv` only validates the first block for schema inference. Ragged rows deeper in a large file raise `ArrowInvalid` during streaming iteration (inside `add_batch`), which skips this fallback entirely and drops the whole file. So the fallback both masks errors it shouldn't and misses the case it was written for.

Please have `_reflow_ragged_rows` count how many rows it truncated/padded and surface that via `self.report.warning(...)` so the profile is never silently based on altered data.

---

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/profiling/profiler.py
```diff
@@ -0,0 +1,310 @@
...
+                        else:
+                            for row in source.rows:
+                                accumulator.add_row(row)
+                except Exception as e:
+                    logger.error(e)
+                    self.report.report_warning(
+                        table_data.display_name,
+                        f"unable to read table {table_data.display_name} from file {path}: {e}",
+                    )
+                    continue
```
sgomezvillamor: 🟡 The `try` wraps the whole per-file accumulation loop, so if a file fails partway through streaming, the accumulator already holds partial rows from it. We `continue`, then still finalize and emit a `DatasetProfileClass` below — the operator gets a warning that the file couldn't be read *and* a profile with wrong `rowCount`/stats built on the partial data. Track a per-table failure count and skip emission (or report a failure) when any file in the table failed, so a partial read invalidates the profile instead of quietly polluting it.

---

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/profiling/profiler.py
```diff
@@ -0,0 +1,310 @@
...
+
+                        assert accumulator is not None
+                        if isinstance(source, ColumnarSource):
+                            for batch in source.batches:
+                                accumulator.add_batch(batch)
+                        else:
+                            for row in source.rows:
+                                accumulator.add_row(row)
+                except Exception as e:
+                    logger.error(e)
```
sgomezvillamor: 🟡 `logger.error(e)` drops the traceback, and `report_warning(display_name, ...)` passes the table name as the title so every table fragments into its own log entry instead of grouping under one stable message. Prefer the structured API: `self.report.warning(title="Failed to read file during profiling", context=path, exc=e)`, and drop the redundant `logger.error` — the report already logs.

---

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/profiling/profiler.py
```diff
@@ -0,0 +1,310 @@
...
+            time_taken = timer.elapsed_seconds()
+            logger.info(
+                f"Finished profiling {table_data.full_path}; took {time_taken:.3f} seconds"
+            )
+            self.profiling_times_taken.append(time_taken)
+
+        yield MetadataChangeProposalWrapper(
+            entityUrn=dataset_urn,
+            aspect=profile,
+        ).as_workunit()
```
sgomezvillamor: 🔴 Blocker: `finalize()` and the profile build run *outside* the per-file `try`. Any error here — sketch math, histogram edge arithmetic, a stat conversion — escapes `get_table_profile`, and the driving loop in `get_workunits_internal` has no per-table guard, so one table's finalize failure aborts profiling (and ingestion) for every remaining table. Wrap this section in a try/except that reports a warning and returns, matching the per-file protection just above it.

---

metadata-ingestion/src/datahub/ingestion/source/gcs/gcs_source.py
```diff
@@ -227,7 +237,7 @@ class GCSSourceReport(DataLakeSourceReport):
     ],
 )
 @capability(SourceCapability.SCHEMA_METADATA, "Enabled by default")
-@capability(SourceCapability.DATA_PROFILING, "Not supported", supported=False)
+@capability(SourceCapability.DATA_PROFILING, "Optionally enabled via configuration")
```
sgomezvillamor, line 240: 🔴 Blocker (docs): now that profiling is capability-supported and no longer uses Spark, the user-facing profiling docs are actively wrong. `metadata-ingestion/docs/sources/s3/s3_post.md` (≈L272–285) still tells users to install Spark 3.0.3 / Hadoop 3.2, set `SPARK_HOME`/`SPARK_VERSION`, and configure `s3a://` for profiling — none of that applies anymore, and the "Avro column names must start with a letter" limitation is Spark-specific too. The GCS docs should also drop the "profiling not supported" framing to match this capability flip.

Since the checklist marks docs as a follow-up: fine for net-new docs, but this is stale content that misinforms users today — worth fixing the s3/gcs profiling sections in this PR. (`abs`/`delta-lake` still use the Spark profiler, so leave those.)

---

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/profiling/accumulators.py
```diff
@@ -0,0 +1,363 @@
...
+            self.null_count += 1
+            return
+        self.non_null_count += 1
+        self._distinct.update(str(value))
+        self._add_to_sample(value)
+        self._track_frequency(value)
+        if self.kind != ColumnKind.OTHER:
+            self._update_min_max(value, value)
+        if self.kind == ColumnKind.NUMERIC:
+            assert isinstance(value, (int, float, Decimal))
```
sgomezvillamor: 🟡 This asserts on external Avro data, so an unexpected runtime type for a NUMERIC-classified column raises a message-less `AssertionError` that the profiler's broad except turns into a blank `{e}` warning — and it's stripped entirely under `python -O`. Prefer an explicit branch that records the offending column/type via the report (or raises a `ValueError` with a message). The `self._median is not None` asserts guard real internal invariants and are fine to keep.

---

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/profiling/accumulators.py
```diff
@@ -0,0 +1,363 @@
...
+    null_count: int
+    unique_count: Optional[int]
+    min_value: Optional[ProfileValue] = None
+    max_value: Optional[ProfileValue] = None
+    mean: Optional[float] = None
+    median: Optional[float] = None
+    stdev: Optional[float] = None
+    quantiles: Optional[List[Tuple[float, float]]] = None
+    histogram: Optional[Tuple[List[str], List[float]]] = None
+    distinct_value_frequencies: Optional[List[Tuple[str, int]]] = None
```
sgomezvillamor: These three fields are passed around as bare tuples and consumed positionally in `profiler.py` (`for q, value in ...`, `boundaries, counts = ...`, `for value, frequency in ...`), which is the "never pass data as tuples" case the house style calls out. Small frozen dataclasses (`Quantile`/`Histogram`/`ValueFrequency`) would give the consumers named access, and a `Histogram.__post_init__` could enforce the `len(boundaries) == len(counts)` invariant that's currently implicit. Freezing `ColumnStats`/`TableStats` while here would make their write-once nature explicit too. (Non-blocking.)

---

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/profiling/accumulators.py
```diff
@@ -0,0 +1,363 @@
...
+        if len(non_null) == 0:
+            return
+
+        # cpc_sketch and the frequency table only accept scalar updates, so
+        # this loop (and reservoir sampling) is the one part of this method
+        # that can't be vectorized.
+        for value in non_null.to_pylist():
+            self._distinct.update(str(value))
+            self._add_to_sample(value)
+            self._track_frequency(value)
```
sgomezvillamor: Minor DRY: this scalar trio (`_distinct.update` / `_add_to_sample` / `_track_frequency`) is duplicated verbatim in `add_value`. Extracting a `_observe_scalar(value)` would collapse both call sites. (Non-blocking.)

---

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/profiling/accumulators.py
```diff
@@ -0,0 +1,363 @@
...
+            if self.non_null_count > 0
+            else None
+        )
+        row_count = self.non_null_count + self.null_count
+        # The old Spark/Deequ profiler passed null_fraction (not the unique
+        # proportion) as convert_to_cardinality's `pct_unique` argument, so
+        # `Cardinality.UNIQUE` in practice only fires for all-null columns,
+        # not fully-distinct ones. Matching that intentionally: real
+        # recipes/dashboards depend on fully-distinct numeric/string columns
+        # still getting full stats or frequency tables, not nothing.
```
sgomezvillamor: The comment says `Cardinality.UNIQUE` "in practice only fires for all-null columns" — it actually fires for *no* column. `UNIQUE` needs `pct_unique == 1.0`, but we pass `null_fraction`; `null_fraction == 1.0` implies `non_null_count == 0`, which makes `unique_count` None → `convert_to_cardinality` returns `NONE` at its earlier guard. The branch is dead. The replicated-quirk behavior is right; just worth fixing the comment so the next reader isn't misled. (Non-blocking.)

---

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/profiling/readers.py
```diff
@@ -0,0 +1,163 @@
...
+def read_json(file_obj: IO[bytes]) -> ColumnarSource:
+    # pyarrow.json has no streaming/batch API like parquet/csv (unlike Spark's
+    # read.json(), which is also a full parse under the hood); lake JSON
+    # files are typically small enough for this to be fine.
+    table = pa_json.read_json(file_obj)
+    return ColumnarSource(
+        columns=table.schema.names,
+        column_kinds=_column_kinds_from_schema(table.schema),
+        batches=table.to_batches(),
+    )
```
sgomezvillamor, line 164: `pa_json.read_json` only handles newline-delimited JSON, so a conventional top-level-array `.json` raises `ArrowInvalid` and gets reported upstream as "unable to read table … from file" — which reads like an I/O/permissions failure rather than an unsupported layout. A dedicated message (or detecting the array case) would save real debugging time. (Non-blocking.)

---

sgomezvillamor (commented):
Really nice piece of work overall — dropping the JVM/Spark/Deequ stack for a streaming pyarrow + DataSketches profiler is a big win, and the non-obvious decisions (Chan et al. variance, seeded reservoir, deliberately replicating the Deequ `null_fraction`-as-`pct_unique` quirk) are documented well.

A few things I'd like to see before merge — inline comments have the detail. Summary:

**Blockers**
- CSV ragged-row reflow silently truncates/pads data with no report signal (and produces the empty-named `""` column now in `golden_mces_multiple_files.json`).
- `finalize()` + profile emit run outside the per-file `try`, so one table's failure aborts profiling for all remaining tables.
- User-facing s3/gcs profiling docs still describe the Spark/Hadoop/`s3a://` setup — now wrong.

**Worth addressing**
- Partial-file failure still emits a profile built on partial data.
- Weak error surfacing (`logger.error(e)` loses the traceback; use `report.warning(title=…, exc=…)`).
- Bare `assert isinstance(...)` on external Avro values.

**Golden files** — the single-file goldens match the "approximate counts shifted + NullValue dropped" story exactly. `multiple_files` is broader: besides the phantom `""` column, ~14 more `distinctValueFrequencies`/`uniqueCount`/`sampleValues` blocks appear — mostly borderline columns flipping low↔high cardinality because CPC and HLL estimate differently. That's acceptable, but it's more than the description implies; worth a pass to confirm each add/remove is a cardinality-boundary flip and not accidental.

Test-coverage note in a separate comment. Suggestions are marked non-blocking.

---

sgomezvillamor:
On accumulator test coverage: the cardinality-gating matrix is well covered (numeric/string/temporal × high/low, all-null, nested-skip, the intentional quirk). The problem is the hard code isn't actually exercised:

- **Multi-batch variance merge is never meaningfully tested.** Every `add_batch` test calls it once, so it always hits the init branch — the Chan et al. merge arithmetic only runs via `add_row`, always with `count=1`/`variance=0`. The two profiler partition tests that merge two real batches use byte-identical partitions, so `delta == 0` and the correction term `delta*delta*n1*n2/total` is always zero. You could delete that term and every test still passes — and it's the entire reason we use Chan et al. over naive accumulation.
- **Numeric-stat values have zero golden coverage.** No s3 golden contains a single `mean`/`stdev`/`median`/`quantiles`/`histogram`. `test_s3_profiling_coverage.py` enables the flags but only asserts `profile_workunits > 0` and `warnings == 0` — no value checks. The numeric path is verified only at "didn't crash" level.
- **Decimal high-cardinality → histogram is untested and may throw.** `_continuous_histogram` does `Decimal` edge arithmetic and passes `Decimal` split points into the KLL float sketch's `get_pmf`; whether that raises is unverified.
- Also missing: `unique_count` tolerance at scale (only asserted `== 5` on tiny exact inputs, so the ~2% CPC claim is unprotected), reservoir determinism (seed=0), and the frequency-abandon boundary at 100/101.

Suggest a white-box test merging 2–3 batches with genuinely different means, asserting `mean`/`stdev` against numpy, plus a decimal high-cardinality case.

---

sgomezvillamor (approved):
Approving — the error-handling, type-design, and docs fixes from the last round all landed well, and the reflow/partial-read handling is much more honest now. A few things I'd still like settled, one of them before this actually merges:

**🔴 Blocker / regression — a whole dataset silently lost its profile (the golden masks it).** `datasetProfile` count in `golden_mces_multiple_files.json` dropped 7 → 6: the `NPS…ARCN_Lakes_ChemistryData` CSV is no longer profiled (Spark profiled it, and so did the earlier revision of this PR). It has quoted newlines inside its header, and pyarrow's CSV parser defaults to `newlines_in_values=False`, so `open_csv` rejects it — the old blind reflow accidentally rescued it, and the (correct) new "only reflow genuine width mismatches" logic declines, so the whole table gets skipped. CI is green only because the golden was regenerated to drop the dataset. One-line fix: `pa_csv.ParseOptions(delimiter=delimiter, newlines_in_values=True)` in `read_csv` (matches Spark, no reflow needed); then restore the dataset to the golden and confirm its columns look right.

**🟡 Warning — accumulator test-coverage gaps still open.** The new tests cover the error paths nicely, but the math gaps from my earlier comment are untouched: no multi-batch variance merge with *differing* means (you can still delete the Chan correction term and everything passes), no decimal high-cardinality → histogram, no unique-count tolerance-at-scale, reservoir determinism, or frequency-abandon boundary. That's the novel risk surface — worth adding.

**ℹ️ Suggestion — add GCS profiling docs.** Since this flips GCS profiling to supported, let's take the opportunity to add a short profiling section to `gcs_post.md` mirroring the updated S3 one; there's currently no mention there.

Thanks for the quick turnaround.

---

naeemdh:
Thanks @sgomezvillamor! Addressed all three in c8b3b80d7e.

**🔴 Blocker/regression — NPS profile restored.** One correction on the mechanism: it isn't the quoted newlines. pyarrow parses the `"NH3-N \n(mg N/L)"` header fields fine even with the default `newlines_in_values=False` (verified — `open_csv` succeeds and returns the correct field names). The actual cause is the header's three **trailing commas**, which produce three duplicate empty-named (`""`) columns. `TableAccumulator.add_batch` looked columns up by name, and `batch.column("")` raises `Field "" exists 3 times in schema`, which aborted the whole table.

Fix: access batch columns **positionally (by index)** so duplicate/empty names can no longer crash profiling. The dataset is profiled again — 28 named columns + one merged empty-named column, 20 rows — and the golden is restored (`datasetProfile` count back to 7). I also applied your `newlines_in_values=True` suggestion for Spark-parity robustness against genuine embedded newlines in *data values*; it's harmless here and no other golden changed.

**🟡 Accumulator test-coverage gaps — all added:**
- multi-batch variance merge with *differing* means (deleting the Chan correction term now fails the test)
- decimal high-cardinality → histogram / quantiles / mean
- unique-count (CPC) tolerance at scale
- reservoir-sampling determinism across runs
- frequency-table abandon at the `MAX_TRACKED_FREQUENCIES` boundary

**ℹ️ GCS profiling docs:** added a Profiling section to `gcs_post.md` mirroring the updated S3 one.

Verified locally: `ruff` + `mypy` clean, 55 profiling unit tests pass, and the full local + s3 (moto) + gcs golden suites pass — the golden diff is scoped to exactly the NPS profile restoration.

sgomezvillamor (approved):

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/profiling/profiler.py
```diff
@@ -0,0 +1,348 @@
...
+        extension = os.path.splitext(table_data.full_path)[1]
+
+        telemetry.telemetry_instance.ping("data_lake_file", {"extension": extension})
+
+        allowed_columns: Optional[List[str]] = None
+        accumulator: Optional[TableAccumulator] = None
+
+        table_read_failed = False
+        with PerfTimer() as timer:
+            for path in self._iter_table_paths(table_data):
```
treff7es: `_iter_table_paths` is a generator, so its S3 `list_objects_v2` pagination runs here as the `for` advances — outside the `try` below. For a partitioned table, a listing failure (throttling / AccessDenied) propagates up through `get_table_profile` → `ingest_table` → `get_workunits_internal` (that `yield from self.ingest_table(...)` is unguarded) and aborts the whole source run. The same error on a single-file table just warns + skips at the catch below. That's inconsistent, and it bites hardest on the many-file tables that are the common case in a lake.

Simplest fix: pull the paths inside the guarded block, e.g. `paths = list(self._iter_table_paths(table_data))` inside a try, so a listing error sets `table_read_failed` / reports a warning like everything else.

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/profiling/profiler.py
```diff
@@ -0,0 +1,348 @@
...
+        extension = os.path.splitext(table_data.full_path)[1]
+        table_path = table_data.table_path
+
+        if is_s3_uri(table_path):
+            if self.aws_config is None:
+                raise ValueError("AWS config is required to profile S3/GCS files")
+            s3_client = self.aws_config.get_s3_client(self.verify_ssl)
+            bucket = get_bucket_name(table_path)
+            prefix = get_bucket_relative_path(table_path)
+            paginator = s3_client.get_paginator("list_objects_v2")
```
treff7es: We already have `list_objects_recursive(bucket, prefix, aws_config)` in `aws/s3_boto_utils.py` — it does this same pagination (with a configurable page size and a documented GCS-cursor caveat) and returns structured `S3ObjectInfo` instead of raw `page["Contents"]` dict indexing. Reusing it drops ~15 lines and, if done inside the guarded block, also resolves the listing-abort issue above in one go.

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/profiling/profiler.py
```diff
@@ -0,0 +1,348 @@
...
+)
+from datahub.ingestion.source.data_lake_common.profiling.readers import (
+    AvroSource,
+    ColumnarSource,
+    read_avro,
+    read_csv,
+    read_json,
+    read_parquet,
+)
+from datahub.ingestion.source.s3.datalake_profiler_config import DataLakeProfilerConfig
```
treff7es, line 29: Minor layering nit (fine as a follow-up): now that the profiler lives in `data_lake_common`, importing `DataLakeProfilerConfig` / `DataLakeSourceReport` from `s3/` makes shared code depend on a connector. You already applied the right fix for `TableData` via the `TableDataLike` Protocol — these two are plain config/report classes with no s3-specific behavior, so they'd move cleanly to `data_lake_common` with a re-export from `s3/` for compat.

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/profiling/profiler.py
```diff
@@ -0,0 +1,348 @@
...
+            if table_read_failed:
+                # A file failed partway through streaming, leaving the
+                # accumulator with partial rows. Emitting now would report a
+                # wrong rowCount and stats built on partial data, so skip it.
+                return
+
+            try:
+                table_stats = accumulator.finalize()
+
+                telemetry.telemetry_instance.ping(
```
treff7es: This telemetry ping sits inside the `finalize()` try and runs before the profile object is built. If the ping ever throws, it gets reported to the operator as "Failed to compute profile" and silently drops an otherwise-valid profile. Telemetry shouldn't be able to suppress emission — move it after the profile is assembled (or outside the guard).

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/profiling/profiler.py
```diff
@@ -0,0 +1,348 @@
...
+                                accumulator.add_batch(batch)
+                        else:
+                            for row in source.rows:
+                                accumulator.add_row(row)
+                except Exception as e:
+                    # The accumulator may already hold partial rows from this
+                    # file, so the whole table profile is now untrustworthy;
+                    # record the failure and stop rather than emit polluted stats.
+                    self.report.warning(
+                        title="Failed to read file during profiling",
```
treff7es: This catch also covers the numeric off-type `ValueError` from `accumulators.add_value`, where the file was read fine but a value didn't match the column's Avro-declared type. "file could not be read" points the operator at an I/O/permissions red herring — "could not be read or profiled" would be more accurate.

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/profiling/readers.py
```diff
@@ -0,0 +1,207 @@
...
+            row = row + [""] * (width - len(row))
+        writer.writerow(row)
+    return io.BytesIO(out.getvalue().encode("utf-8")), altered
+
+
+def read_csv(file_obj: IO[bytes], delimiter: str = ",") -> ColumnarSource:
+    # newlines_in_values=True lets quoted fields span physical lines (Spark's
+    # reader tolerates this); without it a legitimate embedded newline would be
+    # misread as a row break.
+    parse_options = pa_csv.ParseOptions(delimiter=delimiter, newlines_in_values=True)
```
treff7es, line 113: `newlines_in_values=True` is doing real work: the ARCN chemistry CSV in the fixtures has quoted column names with embedded newlines (`"NH3-N \n(mg N/L)"`), and this is what makes us read it as 29 cols / 20 rows. The old Spark reader mangled it (14 cols / 33 rows, with header fragments leaking into the `Sampling Date` column as data), so that golden change is a correctness fix, not approximate-count drift. Two small asks: (1) the PR description says the golden diffs are limited to approximate counts + NullValue — worth correcting so reviewers aren't surprised; (2) add a `read_csv` test for a multiline-quoted header, since it's now load-bearing and only covered incidentally by the integration golden.

metadata-ingestion/src/datahub/ingestion/source/data_lake_common/profiling/profiler.py
```diff
@@ -0,0 +1,348 @@
...
+        config = self.profiling_config
+        field_profile = DatasetFieldProfileClass(fieldPath=column_stats.column)
+
+        field_profile.uniqueCount = column_stats.unique_count
+        if column_stats.unique_count is not None and column_stats.non_null_count > 0:
+            field_profile.uniqueProportion = (
+                column_stats.unique_count / column_stats.non_null_count
+            )
+
+        if config.include_field_null_count:
```
treff7es, line 168: The `include_field_*` gates here only ever run in the True direction in tests. `profile_table_level_only=True` is covered, but it skips `_build_field_profile` wholesale, so it doesn't exercise these individual flags. A regression that inverted or dropped one (e.g. emitting `min` when the user disabled it) would pass everything today — one parametrized test toggling a couple flags off and asserting the field is `None` would close it.

treff7es (approved):
Really nice work, thank you!

naeemdh:
Thanks for the thorough pass Tamas! All addressed.

- **Listing-abort inconsistency (profiler.py:220)** — fixed. Path enumeration is now materialized inside the guarded block (`paths = list(self._iter_table_paths(...))` in a try), so an S3 listing failure (throttling / AccessDenied) on a partitioned table reports a warning and skips the table like everything else, instead of propagating out through the unguarded `yield from` and aborting the source run. Added a unit test for the aws-config-missing case (now warn+skip, not raise).
- **Reuse `list_objects_recursive` (profiler.py:130)** — done; dropped the hand-rolled `list_objects_v2` pagination in favor of the shared helper (structured `S3ObjectInfo`, GCS-cursor aware).
- **Telemetry suppressing emission (profiler.py:312)** — moved the `profile_data_lake_table` ping out of the `finalize()` guard, after the profile is assembled, so a telemetry error can never drop a valid profile.
- **Misleading "could not be read" (profiler.py:287)** — reworded to "read or profiled", since that catch also covers the numeric off-type `ValueError` from `add_value`.
- **`newlines_in_values` is load-bearing (readers.py:113)** — (1) corrected the PR description: the ARCN chemistry golden change is a **correctness fix** (29 cols / 20 rows vs Spark's mangled 14 cols / 33 rows), not approximate-count drift; (2) added a `read_csv` unit test for a multiline-quoted header.
- **`include_field_*` gates only tested True (profiler.py:168)** — added a parametrized test that toggles each of min/max/mean/stdev/median/quantiles/histogram off and asserts the corresponding field is `None`.
- **Layering nit (profiler.py:28)** — agreed; leaving the `DataLakeProfilerConfig` / `DataLakeSourceReport` move into `data_lake_common` (with an s3 re-export) as a focused follow-up, per your note.

Verified locally: `ruff` + `mypy` clean, 63 profiling unit tests, and the full local + s3 (moto) + gcs golden suites pass (84 tests).

PR #18350, opened 2026-07-12
feat(graphql): add Metric and SemanticModel entity model, GraphQL schema, and shared edge mapper

ani-malgari:
## Summary

- Adds `metrics.graphql` with the full GraphQL contract for the `Metric` and `SemanticModel` entities (entity types, aspect types, root queries).
- Extends `entity.graphql` with `METRIC` and `SEMANTIC_MODEL` enum members and relocates `EntityEdge` to `common.graphql`.
- Adds `EdgeMapper` to map PDL `Edge` → GraphQL `EntityEdge`.
- Extends `UrnToEntityMapper`, `EntityTypeMapper`, and `EntityTypeUrnMapper` with `metric` / `semanticModel` entries.

**Resolvers and `GmsGraphQLEngine` wiring are intentionally deferred to a follow-up PR.** The new query fields in `metrics.graphql` are not registered with the engine and are therefore not served at runtime, so there is no risk of missing-resolver failures.

## Test plan

- [ ] `EdgeMapperTest` (new): resolvable destination sets `EntityEdge.destination`; unresolvable destination returns `null`; `mapList` filters out nulls; audit stamps and properties are forwarded correctly.
- [ ] `EntityTypeUrnMapperTest` (extended): `metric` and `semanticModel` round-trip via `getName` / `getEntityType` / `getEntityTypeUrn`.
- [ ] `./gradlew :datahub-graphql-core:build` passes (codegen + compile + all tests green).

## Checklist

- [x] Tests added/updated
- [ ] No breaking changes (resolvers deferred; no runtime schema change)

### DIFF
--- a/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/types/common/mappers/EdgeMapper.java
+++ b/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/types/common/mappers/EdgeMapper.java
@@ -1,0 +1,72 @@
+package com.linkedin.datahub.graphql.types.common.mappers;
+
+import com.linkedin.datahub.graphql.QueryContext;
+import com.linkedin.datahub.graphql.generated.AuditStamp;
+import com.linkedin.datahub.graphql.generated.Entity;
+import com.linkedin.datahub.graphql.generated.EntityEdge;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Maps {@link com.linkedin.common.Edge} Pegasus records to the generated GraphQL {@link
+ * EntityEdge}.
+ */
+@Slf4j
+public class EdgeMapper {
+
+  private EdgeMapper() {}
+
+  /**
+   * Maps a single PDL {@link com.linkedin.common.Edge} to a GraphQL {@link EntityEdge}. Returns
+   * {@code null} when the destination URN cannot be resolved (unknown entity type).
+   */
+  @Nullable
+  public static EntityEdge map(
+      @Nullable QueryContext context, @Nonnull com.linkedin.common.Edge edge) {
+    Entity destination = UrnToEntityMapper.map(context, edge.getDestinationUrn());
+    if (destination == null) {
+      log.warn(
+          "EdgeMapper: could not resolve destination URN {} — skipping edge",
+          edge.getDestinationUrn());
+      return null;
+    }
+
+    EntityEdge result = new EntityEdge();
+    result.setDestination(destination);
+
+    if (edge.hasCreated() && edge.getCreated() != null) {
+      AuditStamp created = AuditStampMapper.map(context, edge.getCreated());
+      result.setCreated(created);
+    }
+    if (edge.hasLastModified() && edge.getLastModified() != null) {
+      AuditStamp lastModified = AuditStampMapper.map(context, edge.getLastModified());
+      result.setLastModified(lastModified);
+    }
+    if (edge.hasProperties() && edge.getProperties() != null) {
+      result.setProperties(StringMapMapper.map(context, edge.getProperties()));
+    }
+
+    return result;
+  }
+
+  /**
+   * Maps a list of PDL {@link com.linkedin.common.Edge} objects, drops and warns any whose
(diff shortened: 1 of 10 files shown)


datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/types/common/mappers/EdgeMapper.java
```diff
@@ -0,0 +1,72 @@
...
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Maps {@link com.linkedin.common.Edge} Pegasus records to the generated GraphQL {@link
+ * EntityEdge}.
+ */
+@Slf4j
+public class EdgeMapper {
```
asikowitz, line 20: The frontend might not need all this info. We haven't created one yet so it'd prob be fine to just return the list of urns rather than all this info. But guess it doesn't really hurt

datahub-graphql-core/src/main/resources/metrics.graphql
```diff
@@ -0,0 +1,504 @@
...
+
+  """
+  Fetch a SemanticModel by URN.
+  """
+  semanticModel(urn: String!): SemanticModel
+
+  """
+  List root-level Metrics (paginated).
+  """
+  getRootMetrics(input: GetRootEntitiesInput!): GetRootMetricsResult
```
asikowitz, line 15: What are "root-level" metrics or semantic models?

datahub-graphql-core/src/main/resources/metrics.graphql
```diff
@@ -0,0 +1,504 @@
...
+  semanticModels: [SemanticModel!]!
+  start: Int!
+  count: Int!
+  total: Int!
+}
+
+"""
+A Metric entity from a semantic layer or metric store.
+"""
+type Metric implements Entity {
```
asikowitz, line 51: Can be in a follow up but this should implement `EntityWithRelationships` to go in the lineage graph

datahub-graphql-core/src/main/resources/metrics.graphql
```diff
@@ -0,0 +1,504 @@
...
+  """
+  Edges extending from this entity.
+  """
+  relationships(input: RelationshipsInput!): EntityRelationshipsResult
+}
+
+"""
+A Semantic Model entity that groups datasets and defines metrics context.
+"""
+type SemanticModel implements Entity {
```
asikowitz, line 196: Same here

datahub-graphql-core/src/main/resources/metrics.graphql
```diff
@@ -0,0 +1,504 @@
...
+
+  """
+  The SemanticModel that defines this Metric.
+  """
+  semanticModel: SemanticModel
+
+  """
+  Parent metrics in the metric hierarchy.
+  """
+  parentMetrics: ParentMetricsResult
```
asikowitz, line 105: Do you expect to extend this in the future? Could just be `parentMetrics: [Metric!]!`

asikowitz (approved):

datahub-graphql-core/src/main/resources/metrics.graphql
chriscollins3456, line 51: should both of these implement entity with relationship like most of our entities? i assume they will have relationships we want to have types for

chriscollins3456 (approved):

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/types/common/mappers/EdgeMapper.java
ani-malgari: we need edge metadata on the UI, and hence i'm adding this

datahub-graphql-core/src/main/resources/metrics.graphql
ani-malgari: That show up on level 1 of page load, similar to what we do with glossaryTerms

datahub-graphql-core/src/main/resources/metrics.graphql
ani-malgari: yes, andrew pointed out the same. i'll address it in the follow-up PR

datahub-graphql-core/src/main/resources/metrics.graphql
ani-malgari: may be not, i'll change this in the follow-up PR

PR #18353, opened 2026-07-13
feat(tracing): encode datahub timestamp trace IDs in the browser

Dutt23:
## Summary

Ports the backend `TraceIdGenerator` to the browser so frontend-initiated traces use
DataHub's timestamp-encoded trace ID format instead of OpenTelemetry's random W3C IDs.

Backend reference (what we're porting):
[`metadata-operation-context/src/main/java/io/datahubproject/metadata/context/TraceIdGenerator.java`](../blob/master/metadata-operation-context/src/main/java/io/datahubproject/metadata/context/TraceIdGenerator.java)

A DataHub trace ID is a 32-char hex string in two 8-byte halves:
- first 16 hex chars → timestamp in **micros** (`epochMillis * 1000`)
- last 16 hex chars → random, for uniqueness

Encoding wall-clock time into the ID lets GMS derive it straight from the ID
(`TraceIdGenerator.getTimestampMillis`) — powering Kafka queue-time metrics
(`MCLKafkaListener`) and async write-status classification (`TraceServiceImpl`) with no
extra metadata. Because the browser previously minted random W3C IDs, browser-originated
writes carried no decodable timestamp, so those backend signals fell back to
`createdOn`/were skipped. Matching the backend format restores them.

The browser's `traceparent` is forwarded through the Play proxy to GMS (only RFC 7230
hop-by-hop headers are stripped), so the ID the browser mints is the one GMS continues
and stores as `systemMetadata.telemetryTraceId`.

## Changes

- `datahub-web-react/src/otelTraceIdGenerator.ts` — new `OtelTraceIdGenerator`
  implementing the OTel `IdGenerator` interface. It is **generate-only**: trace IDs
  encode the timestamp; span IDs delegate to the SDK's `RandomIdGenerator`, mirroring
  Java's `IdGenerator.random()`. The browser never decodes its own trace IDs — GMS does —
  so the decode side of the port (`getTimestampMillis`) is intentionally **not** shipped
  as unused prod code. It lives in the test as the round-trip oracle (see below).
- `datahub-web-react/src/otel.ts` — wires the generator into `WebTracerProvider`.
- Unit tests covering format, round-trip (encode → decode → `epochMillis`), plausibility
  bounds, and span-ID delegation.

## Why decode is not in prod

`getTimestampMillis` is what the *backend* runs against incoming IDs; nothing in the
frontend parses a trace ID at runtime. Shipping a decode method with no caller would be
dead code. Instead the test holds a faithful port of the backend's decode logic and uses
it as an oracle: it decodes a browser-minted ID and asserts the original `epochMillis`
comes back, proving the browser format is exactly what GMS's `getTimestampMillis` will
accept. If a UI trace-correlation feature ever needs client-side parsing, lift the
function out of the test into a prod module then.

## No integer overflow

JavaScript numbers are IEEE-754 doubles: integers are only exact up to
`Number.MAX_SAFE_INTEGER` (2^53 − 1 ≈ 9.0e15). Past that, values silently snap to the
nearest representable step — no error, fabricated trailing digits.

Both halves of a trace ID exceed that:
- **Random half** — up to 2^64 − 1 ≈ 1.8e19. At that magnitude the gap between
  representable doubles is ~2048, so a `Number` would lose ~11 bits of entropy and could
  not reproduce the exact 16 hex chars.
- **Timestamp half** — `epochMillis * 1000` sits just under 2^53 today but crosses it
  around the year 2255; relying on that headroom is a latent bug.

So the ID path never uses `Number`:
- timestamp half is computed with **BigInt** (`BigInt(epochMillis) * BigInt(1000)`),
- random half is built from **raw bytes** (`crypto.getRandomValues` → hex),
- the decode oracle in the test does all comparisons in **BigInt** and only calls
  `Number()` after the value passes the plausibility window — a passing value is
  ≤ now+24h (~1.7e12), comfortably below 2^53, so that conversion is exact.

## Timekeeping

Both sides use UTC epoch time with no timezone conversion: JS `Date.now()` and Java
`Instant.now().toEpochMilli()` are both milliseconds since the Unix epoch (UTC). The
micros conversion and plausibility bounds are pure epoch arithmetic on both ends, so IDs
minted in the browser decode identically on the backend regardless of client locale.

## Backward compatibility & safety

- **Format parity with GMS** — the browser emits the same `%016x%016x` lowercase-hex
  layout as Java (`epochMillis * 1000` micros + 16 random hex). The round-trip test proves
  GMS's `getTimestampMillis` accepts browser IDs.
- **W3C validity** — output is 32 lowercase hex and never all-zero (the timestamp half is
  always non-zero post-2020), so it remains a valid `traceparent` trace-id.
- **No change to span IDs** — they still come from the SDK's `RandomIdGenerator`; only
  trace IDs change, so context propagation and downstream collectors are unaffected.
- **Randomness** — the random half uses `crypto.getRandomValues` (CSPRNG), not
  `Math.random`. `getRandomValues` is available in browsers regardless of secure context,
  and the previous SDK default generator used the same API, so there is no new failure or
  crash path.


david-leifker (approved):
Thank you!

PR #18354, opened 2026-07-13
fix(ingest/datahub): use dialect-aware JSON extraction in soft_delete…

### LINKED ISSUE #18352: datahub source: soft_deleted_urns_query still uses MySQL-only SQL (JSON_EXTRACT, double-quoted string) and fails on PostgreSQL
**Describe the bug**
The `datahub` ingestion source (DataHub-to-DataHub migration) fails on PostgreSQL backends when fetching soft-deleted URNs, because `soft_deleted_urns_query` in `metadata-ingestion/src/datahub/ingestion/source/datahub/datahub_database_reader.py` is written in MySQL-only SQL:

```python
@property
def soft_deleted_urns_query(self) -> str:
    return f"""
        SELECT DISTINCT mav.urn
        FROM {self.engine.dialect.identifier_preparer.quote(self.config.database_table_name)} as mav
        JOIN (
            SELECT *,
            JSON_EXTRACT(metadata, '$.removed') as removed
            FROM {self.engine.dialect.identifier_preparer.quote(self.config.database_table_name)}
            WHERE aspect = "status" AND version = 0
        ) as sd ON sd.urn = mav.urn
        WHERE sd.removed = true
        ORDER BY mav.urn
    """
```

Two problems on PostgreSQL:
1. `JSON_EXTRACT(text, unknown)` does not exist in PostgreSQL → `psycopg2.errors.UndefinedFunction`.
2. `aspect = "status"` — double quotes denote an *identifier* in PostgreSQL (and ANSI SQL), not a string literal → `column "status" does not exist`.

Note: the main paginated query was already made dialect-aware via `_get_json_extract_expression()` (which returns `((metadata::json)->>'removed')::boolean` for PostgreSQL and uses `aspect = 'status'`), but `soft_deleted_urns_query` was missed and still has the hardcoded MySQL expressions. This is present on current `master`.

**To Reproduce**

1. Run DataHub with a PostgreSQL metadata store (`metadata_aspect_v2` in PostgreSQL).
2. Configure a `datahub` source recipe with a PostgreSQL `database_connection`, a `kafka_connection` (so timeseries aspects are read from Kafka), and `include_soft_deleted_entities: false`:
   ```yaml
   source:
     type: datahub
     config:
       database_connection:
         scheme: postgresql+psycopg2
         host_port: '<pg-host>:5432'
         database: datahub
         username: <user>
         password: <password>
       kafka_connection:
         bootstrap: '<broker>:9092'
         schema_registry_url: '<schema-registry>'
       include_soft_deleted_entities: false
   sink:
     type: datahub-rest
     config:
       server: '<gms-url>'
   ```
3. Run `datahub ingest -c recipe.yml`.
4. The Kafka phase calls `DataHubDatabaseReader.get_soft_deleted_rows()` (see `datahub_source.py`, `_get_kafka_workunits` prerequisite), which executes `soft_deleted_urns_query` and fails with:
   ```
   sqlalchemy.exc.ProgrammingError: (psycopg2.errors.UndefinedFunction)
   function json_extract(text, unknown) does not exist
   LINE ...: JSON_EXTRACT(metadata, '$.removed') as removed
   HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
   ```

(For older releases such as v1.2.0, the main pagination query in `query()` fails with the same `UndefinedFunction` error even without `kafka_connection`; that part has since been fixed by `_get_json_extract_expression()`, but the soft-deleted query was left behind.)

**Expected behavior**
`soft_deleted_urns_query` should render dialect-appropriate SQL, like the main query already does:

- Reuse `_get_json_extract_expression()` instead of the hardcoded `JSON_EXTRACT(metadata, '$.removed')` — i.e. `((metadata::json)->>'removed')::boolean` on PostgreSQL.
- Use a single-quoted string literal `aspect = 'status'` (works on MySQL as well) instead of `aspect = "status"`.

With those two changes the query works on both MySQL and PostgreSQL. We verified this fix end-to-end against a PostgreSQL-backed DataHub instance.

**Screenshots**
N/A (CLI ingestion failure; full stack trace included above).

- OS: Linux (containerized ingestion, `acryl-datahub-actions` image)
- Browser: N/A (server-side ingestion bug)
- Version: reproduced on DataHub v1.2.0; `soft_deleted_urns_query` bug still present on current `master` of `metadata-ingestion`

**Additional context**
- Metadata store: PostgreSQL (HA/pgpool), `database_connection.scheme: postgresql+psycopg2`.
- The dialect gap is asymmetric: `query()` and `execute_server_cursor()` already branch on `self.engine.dialect.name == "postgresql"`, so PostgreSQL is clearly intended to be supported by this source — only `soft_deleted_urns_query` was left MySQL-only.

yjyspree:
…d_urns_query

The soft-deleted URNs query was written in MySQL-only SQL: JSON_EXTRACT does not exist in PostgreSQL, and double-quoted "status" is parsed as an identifier rather than a string literal. Reuse the existing \_get_json_extract_expression() helper and a single-quoted literal so the query works on both dialects.

Fixes datahub-project/datahub#18352


yjyspree:
The `testIntegrationBatch0` failure is unrelated to this change — `tests/integration/mode/test_mode_threading.py::test_threading_speedup` failed a timing assertion ("Expected >1.5x speedup but got 1.31x"), which is runner-performance dependent. This PR only touches a SQL string in the `datahub` source. Pushed an empty commit to retrigger CI.

puneetagarwal-datahub (approved):

alokr-dhub (approved):

PR #18363, opened 2026-07-13
feat(docs): reconcile documents tests and lifecycle changes to OSS

purnimagarg1:
**Linear ticket:**
https://linear.app/acryl-data/issue/PFP-4573/reconcile-changes-with-oss-fix-for-documents-tests-after-merging-to

**Description:**

Reconciles document test changes in playwright made in this PR: https://github.com/acryldata/datahub-fork/pull/10551
 To support these changes and make the tests work, lifecycle state work from this PR is also included: https://github.com/acryldata/datahub-fork/pull/9358


datahub-web-react/src/app/entityV2/summary/properties/property/properties/DocumentStatusProperty.tsx
```diff
@@ -13,76 +13,98 @@ import { SimpleSelect } from '@src/alchemy-components';
...
             return i18next.t('entity.profile.summary:documentStatus.draft');
         },
-        value: DocumentState.Unpublished,
+        value: DRAFT_LIFECYCLE_STAGE_URN,
+    },
+    {
+        get label() {
+            return i18next.t('entity.profile.summary:documentStatus.unpublished');
+        },
+        value: UNPUBLISHED_LIFECYCLE_STAGE_URN,
```
chriscollins3456, line 37: i think in the UI we only want to show Published / Unpublished and hide the draft option entirely now

metadata-service/configuration/src/main/resources/bootstrap_mcps/lifecycle-stages.yaml
```diff
@@ -1,115 +1,145 @@
-# Example lifecycle stage types for entity governance.
-# Uncomment and increment the version in bootstrap_mcps.yaml to bootstrap these stages.
+# Lifecycle stage types for entity governance.
```
chriscollins3456, line 1: let's make sure this file matches the current version of it exactly in SaaS

chriscollins3456 (commented):
nice thank you! i think the biggest thing is to make sure the files we're changing here is up to date with latest saas code and not just the Pr that was merged a few months ago. so if you wouldn't mind doing a double check on that to ensure we don't revert any changes in saas we've made in the last few months

PR #18367, opened 2026-07-13
fix(ingest/bigquery): source table stats from INFORMATION_SCHEMA.PARTITIONS

aviraj-gour:
## Summary

BigQuery sourced table row_count, size, and last_altered from the undocumented __TABLES__ Legacy SQL construct, which is unsupported by Google and may change or stop working without notice.

This changes the default to the supported `INFORMATION_SCHEMA.PARTITIONS` view. `PARTITIONS` is dataset-scoped, and is already queried for partition metadata, so sourcing stats from it adds no extra scan. It covers base tables only.

## Impact (default behavior)

| Entity | Change |
| --- | --- |
| Base tables | Unchanged (row count, size, last-altered still populated). except **empty tables** (zero rows), which are absent from `PARTITIONS` and so lose their `lastModified` timestamp. |
| External tables | Lose `lastModified` (absent from `PARTITIONS`). Row count/size were never populated regardless — BigQuery does not store their data. |
| Views | Lose `lastModified` (row count/size were never emitted for views). |
| Snapshots | Lose `lastModified` and the `rows_count` / `size_in_bytes` custom properties. |

- **Views and snapshots are no longer gated on `is_profiling_enabled()`.** Previously their `lastModified` (and, for snapshots, the `rows_count` / `size_in_bytes` custom properties) were fetched via `__TABLES__` only when profiling was enabled. But these values populate metadata aspects only — they are never consumed by the profiler — so tying them to the profiling flag was incorrect. They now have no `PARTITIONS`-based equivalent (views/snapshots aren't covered by that view), so they are gated solely on the new `use_legacy_table_stats` opt-in, independent of whether profiling is on.

- To restore the previous behavior, set `use_legacy_table_stats: true`.

## Changes

- New `tables_for_dataset_with_partition_stats` query that sources `row_count` / `size` / `last_altered` from `INFORMATION_SCHEMA.PARTITIONS` (aliased to the same output columns the mapper reads, so parsing is shared with the legacy query). External tables are absent from `PARTITIONS` and get null stats via a `LEFT JOIN`.
- New config flag `use_legacy_table_stats` (default `false`). Set it to `true` to restore the `__TABLES__` path.
- Views and snapshots have no `PARTITIONS`-based equivalent (they are not covered by that view), so their stats are gated solely on `use_legacy_table_stats`. Previously they were populated whenever profiling was enabled; those values feed metadata aspects only and are never consumed by profiling.
- Renamed the per-dataset stats queries to name their stats source, so the legacy `__TABLES__` path is obvious to future readers:
  - `tables_for_dataset` → `tables_for_dataset_with_legacy_stats`
  - `tables_for_dataset_partitions` → `tables_for_dataset_with_partition_stats`
  - `tables_for_dataset_without_partition_data` → `tables_for_dataset_without_stats`
  - `views_for_dataset` → `views_for_dataset_with_legacy_stats`
  - `views_for_dataset_without_data_read` → `views_for_dataset_without_stats`
  - `snapshots_for_dataset` → `snapshots_for_dataset_with_legacy_stats`
  - `snapshots_for_dataset_without_data_read` → `snapshots_for_dataset_without_stats`


docs/how/updating-datahub.md
```diff
@@ -49,6 +49,17 @@ Requirements:
...
 
+- **(Ingestion / BigQuery)** BigQuery now sources row count, size, and last-altered time from the supported `INFORMATION_SCHEMA.PARTITIONS` view instead of the undocumented `__TABLES__` Legacy SQL construct, which is unsupported by Google and may change or stop working without notice. `PARTITIONS` is dataset-scoped and free but covers base tables only, so the default stats change per entity type:
```
```suggestion
- #18367 **(Ingestion / BigQuery)** BigQuery now sources row count, size, and last-altered time from the supported `INFORMATION_SCHEMA.PARTITIONS` view instead of the undocumented `__TABLES__` Legacy SQL construct, which is unsupported by Google and may change or stop working without notice. `PARTITIONS` is dataset-scoped and free but covers base tables only, so the default stats change per entity type:
```

sgomezvillamor:
Couple of things, neither blocking:

**Test the NULL-stats path.** Now that base/external tables come through a `LEFT JOIN`, empty and external tables land with null `p.*` (no row count / size / last-altered). That path is exercised only implicitly today. Worth a small mapper test that feeds a row with null partition stats and asserts we get `None` back rather than an error — locks in the documented "empty tables lose lastModified" behavior against future mapper changes.

**Open for debate: restore `lastModified` for views via `INFORMATION_SCHEMA.TABLES`.** `PARTITIONS` doesn't cover views, so we drop their `lastModified` here. But `TABLES` has `last_change_time`, and the view queries already `SELECT ... FROM INFORMATION_SCHEMA.TABLES` — so we could add it back from a supported source at zero extra cost, no `__TABLES__`. Semantics are "last definition change" rather than "last data change," which for views is actually the right meaning (a view only changes when it's redefined). One thing to verify first is the column's type/units so we convert to the epoch millis the mapper expects.

And while we're here: this would be populated for all views, not gated on profiling like before. That coupling was arbitrary — probably a mistake — so decoupling it is the right call regardless.

---

aviraj-gour:
Thanks for a Review @sgomezvillamor,

> Test the NULL-stats path. Now that base/external tables come through a LEFT JOIN, empty and external tables land with null p.* (no row count / size / last-altered). That path is exercised only implicitly today. Worth a small mapper test that feeds a row with null partition stats and asserts we get None back rather than an error — locks in the documented "empty tables lose lastModified" behavior against future mapper changes.

Added a test.


> Open for debate: restore lastModified for views via INFORMATION_SCHEMA.TABLES. PARTITIONS doesn't cover views, so we drop their lastModified here. But TABLES has last_change_time, and the view queries already SELECT ... FROM INFORMATION_SCHEMA.TABLES — so we could add it back from a supported source at zero extra cost, no __TABLES__. Semantics are "last definition change" rather than "last data change," which for views is actually the right meaning (a view only changes when it's redefined). One thing to verify first is the column's type/units so we convert to the epoch millis the mapper expects.

Re restoring view lastModified via INFORMATION_SCHEMA.TABLES.last_change_time: that column doesn't exist in TABLES — querying it errors with Unrecognized name: last_change_time, and it's not in the [docs](https://docs.cloud.google.com/bigquery/docs/information-schema-tables). TABLES only has creation_time.

Supported alternatives are TABLE_STORAGE.last_modified_time or a per-table tables.get call (covers views but one API call per table, against the batched approach here). So I'd leave it out of this PR.

sgomezvillamor (approved):
LGTM. Moving base-table stats off the undocumented `__TABLES__` onto the supported `INFORMATION_SCHEMA.PARTITIONS` is the right call, and reusing the already-queried view keeps it free. Verified the aliases are preserved so the mapper is untouched, the `UNIX_MILLIS` wrap matches what `parse_ts_millis` expects, and view/snapshot stats only feed metadata aspects (not the profiler), so the default change is safe. Behavior changes are well documented in `updating-datahub.md` and the `use_legacy_table_stats` escape hatch covers anyone who needs the old path.

---

PR #15927, opened 2026-01-20
feat(powerbi): Add DirectLake lineage extraction for Microsoft Fabric

askumar27:
## 📋 Summary

Add support for extracting lineage from PowerBI DirectLake semantic models to their upstream Microsoft Fabric OneLake tables (Lakehouses and Warehouses).

## 🎯 Motivation

PowerBI DirectLake is a new storage mode that enables semantic models to read data directly from Microsoft Fabric Lakehouses and Warehouses without importing data. Users with DirectLake semantic models currently have no lineage visibility between their PowerBI tables and the underlying Fabric tables. This PR enables automatic lineage extraction from DirectLake tables to their upstream OneLake sources.

## 🔧 Changes Overview

**New Features:**
- DirectLake lineage extraction from PowerBI tables to Fabric OneLake tables
- Parsing of Fabric artifacts (Lakehouse, Warehouse, SQLAnalyticsEndpoint) from PowerBI Admin API scan results
- URN generation for `fabric-onelake` platform following PR #15888 conventions
- Support for both schemas-enabled and schemas-disabled Lakehouses

**Modifications:**
- Extended `extract_lineage()` to route DirectLake tables through dedicated extraction path
- Added `fabric_artifacts` to Workspace data class to store discovered Fabric resources (including resolved `physical_item_ids` for SQLAnalyticsEndpoint)
- Added DirectLake-specific fields to Table (`storage_mode`, `source_schema`, `source_expression`) and PowerBIDataset (`dependent_on_artifact_id`) data classes
- Fixed AttributeError when logging at DEBUG in Fabric artifact parsing (use `logger.isEnabledFor(logging.DEBUG)`)

**New Files:**
- `fabric_urn_builder.py` - URN generation for OneLake tables (temporary, to be replaced by imports from PR #15888)
- `test_powerbi_directlake.py` - Comprehensive unit tests for DirectLake functionality

### URN Format

DirectLake lineage generates URNs in the `fabric-onelake` platform using this pattern:
- With schema: `urn:li:dataset:(urn:li:dataPlatform:fabric-onelake,{workspaceGUID}.{itemGUID}.{schema}.{table},PROD)`
- Without schema: `urn:li:dataset:(urn:li:dataPlatform:fabric-onelake,{workspaceGUID}.{itemGUID}.{table},PROD)`

[...]

PR #18368, opened 2026-07-13
fix(ingest/fabric-onelake): use provisioningStatus instead of hostname pattern for SQL endpoint validation

### LINKED ISSUE #18335: [Question] fabric-onelake: SQL Analytics Endpoint connection string extraction could fail when hostname doesn't match .datawarehouse.fabric.microsoft.com
## Question

In [`schema_client.py#L295`](https://github.com/datahub-project/datahub/blob/master/metadata-ingestion/src/datahub/ingestion/source/fabric/onelake/schema_client.py#L295), the connector validates the SQL Analytics Endpoint connection string with:

```python
if ".datawarehouse.fabric.microsoft.com" in conn_str:
    return conn_str.strip()
```

The Fabric REST API (`GET /v1/workspaces/{id}/lakehouses/{id}`) returns the endpoint in `properties.sqlEndpointProperties.connectionString` with `provisioningStatus: "Success"`, but if the hostname format were to change or not contain the exact substring `.datawarehouse.fabric.microsoft.com`, the connector silently discards it and reports:

> SQL Analytics Endpoint URL is required for Lakehouse {id}

**Why is this validation needed?** If the API returns a `connectionString` and `provisioningStatus` is `"Success"`, wouldn't it be safer to trust the response rather than pattern-matching on a specific hostname format? This would make the connector resilient to any future hostname changes from Microsoft.

**Version:** `acryl-datahub==1.6.0.12`

billpratt:
## Why this change

The `get_sql_analytics_endpoint_url` method validates the SQL Analytics Endpoint connection string by checking for the substring `.datawarehouse.fabric.microsoft.com`. This fails when the Fabric environment returns a hostname with a different prefix pattern (e.g., `env-datawarehouse.fabric.microsoft.com`), causing schema extraction to be silently skipped.

The Fabric REST API already provides a `provisioningStatus` field in `sqlEndpointProperties` that explicitly indicates whether the endpoint is ready. Checking this field is more robust than pattern-matching a hostname that could change across environments.

## What changed

* `schema_client.py`: Replaced hostname substring check with `provisioningStatus == \"Success\"` validation. If the API confirms the endpoint is provisioned and a connection string is present, we trust and return it.
* `schema_client.py`: Updated stale documentation URLs (two 404 links replaced with current Microsoft Learn references).
* `test_schema_client.py`: Updated existing test to include `provisioningStatus` field, added test for non-standard hostname with `Success` status, added test verifying `InProgress` status returns `None`.

## Validation

* Unit tests pass for all scenarios: standard hostname, non-standard hostname prefix, InProgress status, missing status, empty properties, Warehouse path routing.

Fixes datahub-project/datahub#18335

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/fabric/onelake/schema_client.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/fabric/onelake/schema_client.py
@@ -186,7 +186,7 @@ def create_schema_extraction_client(
                 raise ValueError(
                     f"SQL Analytics Endpoint URL is required for {item_type} {item_id}. "
                     "The endpoint URL must be obtained from the Fabric API and cannot be constructed. "
-                    "See: https://learn.microsoft.com/en-us/fabric/data-warehouse/what-is-the-sql-analytics-endpoint-for-a-lakehouse"
+                    "See: https://learn.microsoft.com/fabric/data-engineering/lakehouse-sql-analytics-endpoint"
                 )
 
             logger.info(
@@ -262,8 +262,7 @@ class SqlAnalyticsEndpointClient:
         predicted from workspace_id alone. The endpoint must be obtained from the API.
 
         References:
-        - https://learn.microsoft.com/en-us/fabric/data-warehouse/what-is-the-sql-analytics-endpoint-for-a-lakehouse
-        - https://learn.microsoft.com/en-us/fabric/data-warehouse/warehouse-connectivity
+        - https://learn.microsoft.com/fabric/data-engineering/lakehouse-sql-analytics-endpoint
         - https://learn.microsoft.com/en-us/rest/api/fabric/lakehouse/items/get-lakehouse
         - https://learn.microsoft.com/en-us/rest/api/fabric/warehouse/items/get-warehouse
 
@@ -289,11 +288,10 @@ class SqlAnalyticsEndpointClient:
             sql_endpoint_props = properties.get("sqlEndpointProperties")
 
             if isinstance(sql_endpoint_props, dict):
-                # Lakehouse: connectionString is often just the hostname
+                provisioning_status = sql_endpoint_props.get("provisioningStatus")
                 conn_str = sql_endpoint_props.get("connectionString")
-                if conn_str and isinstance(conn_str, str):
-                    if ".datawarehouse.fabric.microsoft.com" in conn_str:
-                        return conn_str.strip()
+                if provisioning_status == "Success" and conn_str:
+                    return conn_str.strip()
 
             # Warehouse API may expose connectionString at properties level
             # (e.g. staging warehouses: properties=['connectionInfo', 'connectionString', ...])
@@ -343,7 +341,7 @@ class SqlAnalyticsEndpointClient:
             raise ValueError(
                 f"SQL Analytics Endpoint URL is required for item {item_id}. "
                 "The endpoint URL must be obtained from the Fabric API and cannot be constructed. "
-                "See: https://learn.microsoft.com/en-us/fabric/data-warehouse/what-is-the-sql-analytics-endpoint-for-a-lakehouse"
+                "See: https://learn.microsoft.com/fabric/data-engineering/lakehouse-sql-analytics-endpoint"
             )
 
         try:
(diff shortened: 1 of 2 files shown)


metadata-ingestion/src/datahub/ingestion/source/fabric/onelake/schema_client.py
```diff
@@ -289,11 +288,10 @@ def get_sql_analytics_endpoint_url(
             sql_endpoint_props = properties.get("sqlEndpointProperties")
 
             if isinstance(sql_endpoint_props, dict):
-                # Lakehouse: connectionString is often just the hostname
+                provisioning_status = sql_endpoint_props.get("provisioningStatus")
                 conn_str = sql_endpoint_props.get("connectionString")
-                if conn_str and isinstance(conn_str, str):
-                    if ".datawarehouse.fabric.microsoft.com" in conn_str:
-                        return conn_str.strip()
+                if provisioning_status == "Success" and conn_str:
```
sgomezvillamor, line 293: we may log a warning if provisioning status not "Success"

metadata-ingestion/src/datahub/ingestion/source/fabric/onelake/schema_client.py
sgomezvillamor, line 1: should we also update https://github.com/billpratt/datahub/blob/61164ef5b4cd8a579a74dbe7bf569ae748ab4269/metadata-ingestion/src/datahub/ingestion/source/fabric/onelake/schema_client.py#L48-L51 accordingly?

```
# Pattern to match Fabric SQL Analytics Endpoint hostname in connection strings
_FABRIC_ENDPOINT_HOST_PATTERN = re.compile(
    r"[a-zA-Z0-9_-]+\.datawarehouse\.fabric\.microsoft\.com"
)
```

sgomezvillamor (commented):
adding the provisioning status validation looks good to me

left some comments though

metadata-ingestion/src/datahub/ingestion/source/fabric/onelake/schema_client.py
billpratt: Looking at the [Warehouse API docs](https://learn.microsoft.com/en-us/rest/api/fabric/warehouse/items/get-warehouse), `properties.connectionString` returns a bare hostname (e.g., `qvrmbuxie7we7glrekxgy6npqu-6xgyei3x2xiejip4iime6knh5m.datawarehouse.fabric.microsoft.com`) — same format as the Lakehouse API.

If that's always the case, we could simplify the Warehouse fallback path to just return `properties.connectionString` directly (like we now do for Lakehouses), which would make `_extract_endpoint_host_from_connection_string` and `_FABRIC_ENDPOINT_HOST_PATTERN` dead code that can be removed.

However, the existing code comment mentions "staging warehouses" with `connectionInfo` — have you seen environments where the response contains a full ODBC-style connection string rather than a bare hostname? That would help me know whether to simplify or just broaden the regex.

metadata-ingestion/src/datahub/ingestion/source/fabric/onelake/schema_client.py
billpratt: Good call — added a warning log

metadata-ingestion/src/datahub/ingestion/source/fabric/onelake/schema_client.py
sgomezvillamor: That was added in this PR https://github.com/datahub-project/datahub/pull/15927 and there are even tests with containing odbc-style connection strings for `connectionInfo`, so let's keep it as it is.

sgomezvillamor (approved):
LGTM
thanks for the contrib!

billpratt:
Hi @maggiehays and @sgomezvillamor, is there anything outstanding I need to do before this can be merged? Thanks!

sgomezvillamor:
I'm merging

Next time, feel free to merge it yourself

sgomezvillamor:
Thanks for the contrib!

billpratt:
Thanks @sgomezvillamor, I didn't have the permissions to merge.

PR #18270, opened 2026-07-08
ci(smoke): enable pytest-xdist and improve matrix batch balancing

david-leifker:
## Summary

Speeds up smoke CI with **pytest-xdist** (`-n 3 --dist=loadscope`), a **two-phase runner** for policy mutators, **refreshed batch weights**, and **parallel-test hardening** (from closed #18337).

## CI results (A/B on PR matrix)

Compared on the same 7-batch matrix against `master`:

| Metric | Serial ([29193706291](https://github.com/datahub-project/datahub/actions/runs/29193706291?pr=18270)) | xdist ([29197070047](https://github.com/datahub-project/datahub/actions/runs/29197070047?pr=18270)) |
|--------|--:|--:|
| Total workflow | 40m 31s | **26m 38s (−34%)** |
| Slowest batch | Batch 4 — 33.4m | Batch 6 — 19.0m (−43%) |
| Result | green | green |

Per-batch job time with xdist: 12–19m (was 19–33m serial). Batch balancing fixed the old ~36m skew; xdist saves another ~14m on top.

## Nightly performance (`BATCH_COUNT=1`, full suite)

Manual nightly on this branch: [29265207098](https://github.com/datahub-project/datahub/actions/runs/29265207098) (`7858570`, **green**, 1h 24m wall clock).

Compare against scheduled master nightly [29237994975](https://github.com/datahub-project/datahub/actions/runs/29237994975) (`d799e80`):

| Run | Scope | Result | Slowest smoke job | Notes |
|-----|-------|--------|-------------------|-------|
| Master **attempt 1** | Full suite (serial) | **failed** | **155m** | 1077 tests; 1 failure; serial `pytest` ~108m CPU (`quickstart-postgres` x64 junit) |
| Master **attempt 2** (final) | **10-test retry only** | green | 20m | Partial re-run of failed modules — **not comparable** to full-suite timing |
| Branch [29265207098](https://github.com/datahub-project/datahub/actions/runs/29265207098) | Full suite (xdist + 2-phase) | **green** | **84m** | 1077 tests (1045 phase-1 + 32 phase-2 mutators); all 8 profile×arch jobs pass |

**Full-suite nightly wall clock (apples-to-apples):** ~115–155m serial (master attempt 1, failed) → **~59–84m** with xdist (**~45–50% faster**), now green on both x64 and arm across all 4 quickstart profiles.

**pytest-reported time** (`quickstart-postgres` x64, same ~1077 tests):

| | Serial (master attempt 1) | xdist branch |
[...]

PR #18375, opened 2026-07-13
perf(ci): speed up backend build & test — Gradle parallelism + opt-in test sharding

Dutt23:
## Summary

Speeds up the `build & test` workflow's backend path in two layers:

1. **Always-on Gradle wins** (behavior unchanged, just faster):
   - Pass `--max-workers=$(nproc)` on the heavy **CI** gradle invocations (seed, shard, coverage,
     build, frontend). This is a **per-invocation CLI override** — the committed
     `org.gradle.workers.max=2` in `gradle.properties` is **left untouched**, so local dev and every
     non-flagged invocation keep the 8GB-laptop-safe cap (no change to the setting that has caused
     memory pressure before). `--max-workers` is Gradle's cap on **concurrent tasks** (how many
     modules compile / run their `test` task at once); it does **not** change per-`Test`-task JVM
     forking — that's the separate `maxParallelForks = cores/2` knob in `build.gradle`. On CI the
     override resolves to the runner's core count: `2` on standard `ubuntu-latest` (identical to
     today) and `4` on the `depot-ubuntu-latest-4` runner (16 GB), so worst case is 4 concurrent
     tasks on a 16 GB ephemeral runner — not the constrained local machine the `=2` guard protects.
   - Bump `gradle/actions/setup-gradle` v5.0.2 → v6.2.0 (enhanced cache backend); mark matrix
     legs `cache-read-only`.
   - Remove leftover diagnostic `Disk Space Analysis` steps.

2. **Opt-in backend test sharding** (default OFF — nothing changes unless enabled). Splits the
   backend test run across parallel matrix runners, reusing the existing pytest sharding
   convention (#18270): a committed weights snapshot + LPT bin-packing.

## Sharding design (module-level, loss-proof)

- **Flag:** `datahub.ci.backendTestSharding` in `gradle.properties` (default `false`). In-repo,
  not a GitHub repo variable — portable across CI providers. When `true`: `backend-seed` +
  `backend-test-shard` run and the single `build-except-metadata-ingestion` job is skipped.
- **Whole modules only:** each shard runs bare `:module:test` (never `--tests`), so no test can
  be silently dropped by class-name discovery. A heavy "whale" module (e.g. `datahub-graphql-core`)
  lands alone on a shard via LPT and is parallelized within it by `maxParallelForks`.
- **Parity with the old job:** `backend-seed` runs `build testClasses -x test` (assemble +
  spotless/checks + main+test compile, no tests) **and** the `Ensure codegen is updated` gate, then
  warms a run-scoped Gradle build cache the shards restore. Shards run tests + upload raw JaCoCo
  `.exec`; a `backend-coverage` job merges the **union** of all shards' exec into one cross-module
  report and uploads once to Codecov (`backend` flag) — so a class covered by another module's
  tests on a different shard isn't lost. Seed + shards + coverage ≡ `build-except`'s `build`.
  Per-shard failures are reported to PostHog (parity with the old job).
- **Weights:** `generate_test_weights.py` gains `--gradle-output` (FQCN → median seconds, hardened
  against nan/inf/negative); the weekly `update-test-weights` workflow refreshes
  `.github/gradle_test_weights.json` and PRs it. Falls back to an even split when absent.
- **Shard count:** single source of truth — the `batch:` matrix length. `--split-index` /
  `--split-total` derive from GitHub's native `strategy.job-index` / `strategy.job-total` (no
  duplicated count, no extra plan job).
- **Guard:** `backend-seed` fails if <20 backend test modules are discovered (catches broken
  globs → a silently near-empty sharded run).
- **No build-config changes.** `backend-seed` runs `build testClasses -x test` with the web-react
  generators **left in** (do NOT `-x` them): they run, produce `datahub-web-react/src`, and both
  `yarnBuild`/vite and `graphqlUsageRegistryCheck` succeed against the real sources — the exact
  graph the pre-existing single-job path exercises. Excluding the generators is what breaks the
  build (yarnBuild against missing sources → vite exit 1), so the seed deliberately does not.

## Measured performance

Run `29285085530` (sharding on, cold cache, no weights → even split by class count) vs the baseline
single `build-except` job (PR #18128) ≈ **36 min**:

| stage | wall |
|---|---|
| `backend-seed` (compile everything once) | ~11 min |
| slowest of 6 shards (whale `datahub-graphql-core`) | ~12 min |
| **time to test signal** (seed → last shard; shards `needs` seed) | **~23 min** |

→ **~36 → ~23 min, ~36% faster** to a pass/fail signal — even cold-cache and even-split. Shards
were well balanced (7.6–11.75 min); the whale did **not** blow the budget.

Notes / caveats:
- **`seed` is the fixed critical-path floor** (compile once so 6 shards don't each recompile). It's
  cache-accelerated — a warm Gradle build cache shrinks it on typical few-module PRs; ~11 min is a
  coldish run. Structurally it can't go below "compile the changed modules".
- **`backend-coverage` runs AFTER the shards** (needs their `.exec`) → adds serial wall on top of
  ~23 min. That run predates the coverage job, so its wall is **not yet measured** — measure on the
  next clean run. Its merge step is `continue-on-error`, so it can't fail the PR.
- **Is `backend-coverage` a required status check?** Not determinable here (branch protection isn't
  API-readable). If protection requires the whole **`build & test`** workflow, coverage's serial
  wall gates merge-ready → accept it, or move coverage to a separate non-required workflow. If it
  requires specific job names, the new `backend-coverage` is not required → doesn't gate. **Verify
  in branch-protection settings before default-on.**
- **`.github/gradle_test_weights.json` is not committed yet** → even split by class count until the
  first `update-test-weights` run; balance tightens with real durations.

## Before flipping default on
1. Measure `backend-coverage` wall on a clean run; decide its required-check status (above).
2. Confirm the merged Codecov `backend` % matches the single-job baseline (the whole reason the
   aggregate job exists).

## Testing

- `.github/scripts/test/test_split_gradle_tests.py` — 12 unit tests (duplicate FQCNs across
  modules, nested/root project paths, exclude globs, weights parse/fallback, whole-module tasks,
  empty shard, every-module-assigned-once, LPT balance, whale isolation). Run by the existing
  `test-github-scripts.yml` (`pytest .github/scripts/test`).
- Splitter + generator smoke-tested against the full backend suite.

## Rollout

Ships **default off** — merging changes nothing. Before enabling: (1) benchmark slowest-shard +
seed wall time vs the current `build-except`, incl. the whale; (2) confirm the merged Codecov
`backend` % matches the single-job baseline on the first sharded run.

## Checklist
- [x] PR title follows Contributing Guideline
- [x] Tests added
- [ ] Docs — n/a (CI-internal)
- [ ] Breaking changes — none (opt-in, default off)


.github/scripts/split_gradle_tests.py
```diff
@@ -0,0 +1,196 @@
+#!/usr/bin/env python3
+"""Assign whole Gradle test modules to one CI shard, balanced by historical duration.
+
+Vendored + stdlib-only (py3.9+). Reads a committed weights snapshot (gradle_test_weights.json,
```
devashish2203: Shouldn't this PR also be committing the first version of gradle_test_weights.json?

.github/workflows/build-and-test.yml
```diff
@@ -50,11 +50,21 @@ jobs:
...
     steps:
       - name: Check out the repo
         uses: acryldata/sane-checkout-action@186e92cc5948a9c3e1cc7a96eaff9f776f3fc8e3 # v7
       - uses: ./.github/actions/ci-optimization
         id: ci-optimize
+      - name: Read CI flags from gradle.properties
+        id: flags
+        # In-repo, CI-agnostic flag (not a GitHub repo variable) so it survives a CI migration.
+        run: |
+          if grep -qE '^[[:space:]]*datahub\.ci\.backendTestSharding[[:space:]]*=[[:space:]]*true[[:space:]]*$' gradle.properties; then
```
devashish2203: Instead of this, let's just add an env variable that determines whether sharding is active or not. 
That way it's easier to turn this on/off atleast until we are confident of the sharded approach.

david-leifker (approved):
Please update PR description to reflect final state.

.github/workflows/build-and-test.yml
```diff
@@ -404,6 +409,296 @@ jobs:
...
+        uses: ./.github/actions/free-disk-space
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6
+        with:
+          python-version: "3.10"
+      - name: Guard — shard discovery is non-empty
+        # Fail loudly if the globs/excludes discover far too few backend test modules, instead
+        # of silently sharding a near-empty suite (green CI that tested nothing). Keep the
+        # glob/exclude set in sync with the backend-test-shard step below.
+        run: |
+          count=$(python3 .github/scripts/split_gradle_tests.py \
```
devashish2203, line 442: What happens if a new module is added or an existing module removed/renamed, but the gradle_test_weights is no refreshed? 

Those tests will never get added to the shards I believe.

.github/workflows/update-test-weights.yml
```diff
@@ -39,25 +39,41 @@ jobs:
...
-          echo "Generating test weights from downloaded artifacts..."
+          echo "Generating pytest weights..."
           python .github/scripts/generate_test_weights.py \
             --input-dir ./test-artifacts \
             --pytest-output smoke-test/pytest_test_weights.json
 
+          echo "Generating Gradle weights (keyed by FQCN)..."
+          python .github/scripts/generate_test_weights.py \
+            --input-dir ./test-artifacts-gradle \
+            --gradle-output .github/gradle_test_weights.json
```
devashish2203: nit: Since this doesn't contain all the gradle_test_weights only selective modules from the build-and-test workflow, we should rename this file to be clearer, or at the very least update the Claude.md with the clarification.

.github/scripts/split_gradle_tests.py
Dutt23: Didn't commit it initially because it's not required for correctness — when the file is absent, [load_weights](https://github.com/datahub-project/datahub/blob/d7c1afd44d1fb4b9aee03c64a0de33f2515843f3/.github/scripts/split_gradle_tests.py#L93) returns {} and the splitter falls back to an even split by class count ([module_weights](https://github.com/datahub-project/datahub/blob/d7c1afd44d1fb4b9aee03c64a0de33f2515843f3/.github/scripts/split_gradle_tests.py#L117-L119)), so shards are balanced (just less precisely) with no weights at all.

That said, agreed it's better to ship balanced from day one — adding the initial gradle_test_weights.json now, generated from a green sharded run. The weekly update-test-weights job keeps it fresh after.

.github/workflows/build-and-test.yml
Dutt23: the weights file doesn't define what runs — it's balance-only. Discovery is filesystem globbing.
discover_modules globs **/src/test/** to find modules; the weights file is loaded separately and only assigns a duration to each already-discovered module.

New module/file:picked up by the glob automatically. A class not in the weights gets a median [fallback](https://github.com/datahub-project/datahub/blob/d7c1afd44d1fb4b9aee03c64a0de33f2515843f3/.github/scripts/split_gradle_tests.py#L117-L119)(weights.get(c, fallback)), so it's still assigned to a shard. And shards run whole:module:test, never--tests ClassName(plan_for_shard), so every test in a discovered module runs regardless of the json. Nothing dropped.
Deleted module:not on disk → not discovered → never enters the task list. Stale json entry is just ignored (nothing builds tasks from the json), so no:gone:testtask and the gradle call can't fail on it.

.github/workflows/update-test-weights.yml
Dutt23: Done

devashish2203 (approved):

.github/workflows/build-and-test.yml
Dutt23: Done added env flag in repo

PR #18384, opened 2026-07-14
fix(zdu): do not delete semantic search index during orphan cleanup

chakru-r:
## Summary

During a zero-downtime `datahub-upgrade`, the `CleanIndices` step removes alias-less orphaned backing indices left behind by reindex/alias-swaps. For each managed `ReindexConfig` it deletes indices matching `<name>_*` that are alias-less and older than the build-indices retention window (default 60 days).

The base entity index config `datasetindex_v2` cleans with pattern `datasetindex_v2_*`, which **also matches the live semantic search index `datasetindex_v2_semantic`**. That index is the semantic index itself (where embeddings live), addressed directly by physical name with **no alias**. So once it ages past retention it satisfies the orphan condition (matched pattern + alias-less + older than retention) and gets deleted as if it were an orphaned backing index of the base entity index.

Because the runtime dual-write to the semantic index (`UpdateIndicesV2Strategy`) is gated on the index existing, deleting it also **silently stops the dual-write**, and the one-shot bootstrap copy (`CopyDocumentsToSemanticIndexStep`) never repopulates it — so semantic search stays broken until a manual re-bootstrap.

The base index `datasetindex_v2` is safe from its own pattern (no trailing `_<x>`); it's the `_semantic` suffix that makes the semantic index look like a base backing index.

## Fix

Skip any index recognized by `IndexConvention.isSemanticEntityIndex(...)` in `ESIndexBuilder.getOrphanedIndices`, before the retention/alias checks. The guard is **intrinsic** — derived from the context `IndexConvention` — so it protects every caller of the orphan sweep rather than depending on callers threading extra state.

- Matches only the bare `<base>index_v2_semantic` name (ends-with check).
- Backing indices (`..._v2_<ts>`, `..._semantic_<ts>`) do **not** match, so genuine orphans are still cleaned normally, including the semantic index's own backing-index churn via its `<base>_semantic_*` config.
- The bare semantic index is removed only by the reindex alias-swap (`renameReindexedIndices`, which deletes the concrete index and points the alias at the new backing) or by an intentional teardown — never by retention-based orphan cleanup.

## Testing

- New `ESIndexBuilderTest.testCleanIndex_DoesNotDeleteSemanticSiblingIndex`: the base config's clean pattern returns both a genuine backing orphan and the live semantic index (both alias-less + old); asserts only the backing orphan is deleted. Index names derived from the context `IndexConvention` so the suffix/prefix match `isSemanticEntityIndex`.
- `./gradlew :metadata-io:test --tests "*ESIndexBuilderTest"` → 57 passing.

## Checklist

- [x] Tests added
- [ ] Docs (n/a)
- [ ] Breaking changes (none)

### DIFF
--- a/metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/indexbuilder/ESIndexBuilder.java
+++ b/metadata-io/src/main/java/com/linkedin/metadata/search/elasticsearch/indexbuilder/ESIndexBuilder.java
@@ -17,6 +17,7 @@ import com.linkedin.metadata.search.utils.ESUtils;
 import com.linkedin.metadata.search.utils.RetryConfigUtils;
 import com.linkedin.metadata.search.utils.SizeUtils;
 import com.linkedin.metadata.timeseries.BatchWriteOperationsOptions;
+import com.linkedin.metadata.utils.elasticsearch.IndexConvention;
 import com.linkedin.metadata.utils.elasticsearch.SearchClientShim;
 import com.linkedin.metadata.utils.elasticsearch.responses.GetIndexResponse;
 import com.linkedin.metadata.utils.elasticsearch.responses.RawResponse;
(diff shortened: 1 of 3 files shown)


david-leifker (approved):
Nice!

PR #18385, opened 2026-07-14
feat(graphql): Add bulkEntityDataProducts resolver

asikowitz:
Adds a resolver to fetch, in bulk, the Data Products that directly contain each of a set of entities, along with which of those an entity is an output port of. Membership is resolved in one batched siblings read plus a single scroll over the DataProductContains relationship; an entity inherits the membership and output-port status of its siblings.

Also exposes `exists` on the DataProduct GraphQL type.

These back the data product lineage visualization, which must know the data product membership of every node in the graph.

### DIFF
--- a/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/dataproduct/BulkEntityDataProductsResolver.java
+++ b/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/dataproduct/BulkEntityDataProductsResolver.java
@@ -57,6 +57,9 @@ public class BulkEntityDataProductsResolver
 
   private static final String DATA_PRODUCT_CONTAINS_RELATIONSHIP = "DataProductContains";
   private static final int MAX_URNS = 100;
+  // Safety net for the sibling-expanded set that drives the relationship scroll's urn filter. Set
+  // well above MAX_URNS since each input entity may contribute several siblings.
+  private static final int MAX_URNS_WITH_SIBLINGS = 1000;
   // Page size for scrolling the DataProductContains relationship. An entity is typically in at most
   // one data product, so a single page comfortably covers the input batch and their siblings.
   private static final int RELATIONSHIP_SCROLL_COUNT = 1000;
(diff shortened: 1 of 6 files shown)


datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/dataproduct/BulkEntityDataProductsResolver.java
```diff
@@ -0,0 +1,319 @@
...
+    // escaping from the async supplier below.
+    final List<Urn> parsedUrns =
+        inputUrns.stream().map(UrnUtils::getUrn).collect(Collectors.toList());
+
+    return GraphQLConcurrencyUtils.supplyAsync(
+        () -> {
+          // 1. Read each input entity's siblings in one batched aspect read.
+          final Map<String, Set<String>> siblingsByEntity = getSiblings(opContext, parsedUrns);
+
+          // 2. Every entity plus its siblings, since a sibling's data product membership counts
```
chriscollins3456, line 97: i think this makes enough sense.. even though i did hesitate with it for a second

maybe an edge case we don't need to think too hard about but what if the siblings are in two different data products?

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/dataproduct/BulkEntityDataProductsResolver.java
```diff
@@ -0,0 +1,319 @@
...
+              QueryUtils.newRelationshipFilter(EMPTY_FILTER, RelationshipDirection.INCOMING),
+              Collections.emptyList(),
+              scrollId,
+              RELATIONSHIP_SCROLL_COUNT,
+              null,
+              null);
+
+      for (RelatedEntities related : scroll.getEntities()) {
+        // Incoming DataProductContains: source is the data product, destination is the entity.
+        final String dataProductUrn = related.getSourceUrn();
```
chriscollins3456, line 218: should we have a check to make sure the urn is indeed a data product urn and not some other entity? we could just log a warning and skip it if we want

datahub-graphql-core/src/main/resources/entity.graphql
```diff
@@ -252,6 +252,13 @@ type Query {
   """
   dataProduct(urn: String!): DataProduct
 
+  """
+  Fetch the Data Products that directly contain each of the given entities, in bulk
+  """
+  bulkEntityDataProducts(
+    input: BulkEntityDataProductsInput!
+  ): [EntityDataProducts!]!
```
chriscollins3456: i would recommend you return a `BulkEntityDataProductsResult` object in case you ever want to extend it. i've had a few endpoints where i've had to do a `v2` of it because it returned too specific of a result and not an object i could extend

chriscollins3456 (approved):
makes enough sense to me with a request to update the return type of the new endpoint to be an object we could extend later

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/dataproduct/BulkEntityDataProductsResolver.java
asikowitz: Then they'd get merged on the frontend right? We support multiple data products by default now

PR #18386, opened 2026-07-14
feat(metrics): schema registration, rename getSemanticModels, and top-level list query resolvers

ani-malgari:
## Summary

- Renames `getRootSemanticModels` → `getSemanticModels` (and `GetRootSemanticModelsResult` → `GetSemanticModelsResult`) in `metrics.graphql`; semantic models are flat so the "root" prefix was misleading
- Adds `metrics.graphql` to the runtime schema merge in `GmsGraphQLEngine` via a new `METRICS_SCHEMA_FILE` constant
- Registers `MetricType` and `SemanticModelType` as engine fields and in `entityTypes` so DataLoaders, batch loading, and `EntityInterfaceTypeResolver` work automatically
- Implements `GetRootMetricsResolver` (filters `hasParentMetric=false`) and `GetSemanticModelsResolver` (no filter) with unit tests
- Wires `metric`, `semanticModel`, `getRootMetrics`, `getSemanticModels` query fetchers in `configureQueryResolvers`, plus `LoadableTypeBatchResolver` for `GetRootMetricsResult.metrics` and `GetSemanticModelsResult.semanticModels`

### DIFF
--- a/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/metrics/GetRootMetricsResolver.java
+++ b/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/metrics/GetRootMetricsResolver.java
@@ -68,7 +68,7 @@ public class GetRootMetricsResolver
                     query,
                     filter,
                     Collections.singletonList(
-                        new SortCriterion().setField("id").setOrder(SortOrder.ASCENDING)),
+                        new SortCriterion().setField("name").setOrder(SortOrder.ASCENDING)),
                     start,
                     count);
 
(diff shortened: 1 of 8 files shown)


datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/metrics/GetRootMetricsResolver.java
```diff
@@ -0,0 +1,114 @@
...
+
+          try {
+            final SearchResult gmsResult =
+                _entityClient.search(
+                    context.getOperationContext().withSearchFlags(flags -> flags.setFulltext(true)),
+                    Constants.METRIC_ENTITY_NAME,
+                    query,
+                    filter,
+                    Collections.singletonList(
+                        new SortCriterion().setField("id").setOrder(SortOrder.ASCENDING)),
```
chriscollins3456: just curious but why sort on ID? just to be deterministic about ordering?

chriscollins3456 (approved):
nice! one question just because i'm curious

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/metrics/GetRootMetricsResolver.java
ani-malgari: ouch we should be sorting by name, i just fixed it. And yes, we want to keep it deterministic until we have more concrete idea on the order we'd like to display them.

PR #18391, opened 2026-07-14
fix(ingest/powerbi): reduce memory usage to avoid OOM on large tenants

acrylJonny:
## Summary

Large PowerBI tenants OOM'd during ingestion. This PR addresses two independent root causes, both of which held or generated large amounts of data in memory, and consolidates the PowerBI unit tests.

### 1. Everything held resident across both ingestion phases

The connector kept every `Workspace` (including its raw `scan_result` JSON payloads — the dominant memory consumer) and a global `dataset_registry` of every `PowerBIDataset` (with all tables, columns and measures) in memory across both phases. On tenants with many/large workspaces this is easily multiple GB.

- **Stream workspaces through a `FileBackedDict`** in `get_workunits_internal`. Workspaces are seeded onto disk, scanned per batch (writing scan-derived mutations back), then processed one at a time in Phase 2 and dropped after emitting work units. Only the active workspace plus a small LRU cache stay resident. Cache size is derived from `scan_batch_size` so Phase 1 batching doesn't thrash.
- **Back `PowerBiAPI.dataset_registry` with a `FileBackedDict`** so the bulk of the cross-workspace registry spills to disk.
- **Add `close()`** to `PowerBiAPI` and `PowerBiDashboardSource` to release the file-backed SQLite temp files before the base class commits state.
- **Guard the debug `json.dumps(scan_result)`** behind `logger.isEnabledFor(logging.DEBUG)` so it no longer builds a hundreds-of-MB string when debug logging is off.

### 2. M-Query parse tree rendered into logs and warning contexts

`DataAccessFunctionDetail` carries the full M-Query parse tree (`node_map`) plus the `InvokeExpression` subtree (`arg_list`). It is logged at debug level and embedded in warning contexts on nearly every table, and the default dataclass `repr` rendered the entire tree each time. With `--debug` enabled this produced multi-GB ingestion logs and heavy transient/retained memory.

- **Mark `node_map` and `arg_list` as `field(repr=False)`** so they are excluded from the repr while remaining fully accessible for lineage navigation.

### 3. Test consolidation

- Moved the four remaining top-level `test_powerbi_*.py` files into the existing `tests/unit/powerbi/` package and dropped the now-redundant `powerbi_` filename prefix to match the package convention. Pure file moves; no test-logic changes.

Net effect: memory now scales with cache size rather than tenant size, and debug logs stay bounded.

## Testing

- Added `test_dataset_registry_survives_eviction_round_trip` — forces cache eviction (`cache_max_size=1`) to verify a `PowerBIDataset` (nested avro `Table`/`Column` types and the circular `Table.dataset` back-reference) survives the pickle → SQLite → unpickle cycle the fix depends on.
- Added `test_data_access_func_detail_repr_excludes_parse_tree` — guards that the parse tree and arg list stay out of the repr.
- All existing PowerBI unit and integration tests pass (the integration test exercises the streaming refactor end-to-end).
- `./gradlew :metadata-ingestion:lint` (ruff + format + mypy) is clean.

## Checklist


metadata-ingestion/src/datahub/ingestion/source/powerbi/powerbi.py
```diff
@@ -2034,83 +2035,119 @@ def get_workunits_internal(self) -> Iterable[MetadataWorkUnit]:
...
-            allowed_workspaces, self.source_config.scan_batch_size
+        workspace_ids: List[str] = [w.id for w in allowed_workspaces]
+
+        # Workspaces (their raw scan_result payloads especially) are the dominant
+        # memory consumer and must stay available between Phase 1 (scan) and
+        # Phase 2 (per-workspace detail). Spilling them to a file-backed dict
+        # instead of holding them all resident is what prevents OOMs on large
+        # tenants. Cache size holds a scan batch plus buffer to avoid thrashing.
+        ws_cache_size = min(max(self.source_config.scan_batch_size * 2, 10), 200)
+        workspaces: FileBackedDict[powerbi_data_classes.Workspace] = FileBackedDict(
```
sgomezvillamor: Suggestion: `FileBackedDict` is a `Closeable`, i.e. an `AbstractContextManager` whose `__exit__` calls `close()`. So the `try/finally` here can just be a `with`, which drops a level of nesting and removes the manual close:

```python
with FileBackedDict[powerbi_data_classes.Workspace](
    tablename="workspaces",
    cache_max_size=ws_cache_size,
    cache_eviction_batch_size=max(ws_cache_size // 2, 1),
) as workspaces:
    ...
```

(An `ExitStack` isn't warranted — there's only one scoped resource here; `dataset_registry` is torn down separately in `PowerBiAPI.close()`.)

---

metadata-ingestion/src/datahub/ingestion/source/powerbi/powerbi.py
```diff
@@ -2034,83 +2035,119 @@ def get_workunits_internal(self) -> Iterable[MetadataWorkUnit]:
...
+                del workspaces[workspace_id]
+        finally:
+            workspaces.close()
 
     def get_report(self) -> SourceReport:
         return self.reporter
+
+    def close(self) -> None:
+        self.powerbi_client.close()
+        super().close()
```
sgomezvillamor: Warning: if `powerbi_client.close()` raises, `super().close()` never runs. `powerbi_client.close()` closes the sqlite connection and does `shutil.rmtree(temp_dir)`, both of which can fail (flush/`OperationalError` on a full or read-only disk, or the temp DB still being handle-locked on Windows → `PermissionError` from `rmtree`). `super().close()` is `StatefulIngestionSourceBase.close()`, which calls `state_provider.prepare_for_commit()` — and this connector drives per-workspace stale-entity-removal, so a skipped commit can corrupt incremental state on the next run. Best-effort temp-file cleanup shouldn't be able to abort the state commit:

```suggestion
    def close(self) -> None:
        try:
            self.powerbi_client.close()
        finally:
            super().close()
```

---

metadata-ingestion/tests/unit/powerbi/test_fill_metadata_from_scan.py
```diff
@@ -341,6 +345,53 @@ def _fake_get_workspace_datasets(workspace: Workspace) -> Dict[str, Any]:
     )
 
 
+def test_dataset_registry_survives_eviction_round_trip():
```
sgomezvillamor, line 348: Warning (please confirm): the two behaviours that are the actual point of this fix only trigger under cache eviction, and I don't think anything exercises that path:

- The write-back loop in `get_workunits_internal` (`workspaces[ws.id] = ws` after each batch) is there because a workspace reloaded from disk comes back with `dirty=False`; without re-assigning it, the in-place scan mutations are dropped on the next eviction and Phase 2 reads a stale, pre-scan copy. But every test uses a handful of workspaces with a cache of ~20, so nothing ever evicts — **the suite still passes if that write-back loop is deleted.**
- This test round-trips a standalone `PowerBIDataset`, which is great, but not the `workspaces` dict — which the code itself calls the dominant memory consumer. A `Workspace` carries the raw `scan_result`, nested datasets, `app`, endorsements, and the circular `Table.dataset` refs; if any of that doesn't survive pickle it only surfaces in prod on a large tenant.

Two small additions would close it: (a) a `get_workunits_internal` test with `scan_batch_size=1` and 12+ workspaces that mutates each `scan_result` in Phase 1 and asserts it's visible in Phase 2 (fails if the write-back is removed), and (b) a full-`Workspace` round-trip through `FileBackedDict(cache_max_size=1)`. Can you confirm whether the integration test actually hits eviction? I don't think it does (≤2 mocked workspaces, cache ≥200).

---

metadata-ingestion/src/datahub/ingestion/source/powerbi/rest_api_wrapper/powerbi_api.py
```diff
@@ -91,7 +95,18 @@ def __init__(
         # We need to store the dataset ID (which is a UUID) mapped to its dataset instance.
         # This mapping will allow us to retrieve the appropriate dataset for
         # reports and tiles across different workspaces.
-        self.dataset_registry: Dict[str, PowerBIDataset] = {}
+        #
+        # It spans every scanned workspace, so on large tenants it can hold tens
+        # of thousands of datasets. It is file-backed so the bulk spills to disk
+        # rather than staying resident, which previously caused OOMs.
+        self._file_backed_conn = ConnectionWrapper()
+        self.dataset_registry: FileBackedDict[PowerBIDataset] = FileBackedDict(
```
sgomezvillamor: Minor suggestion: worth a one-line comment here that entries are now value-copied snapshots, not shared identity. With the old plain dict, `dataset_registry[id]` and `workspace.datasets[id]` were the same instance; `FileBackedDict` returns a freshly-unpickled copy on a cache miss, so the dataset is effectively stored twice and the two are no longer the same object. It's fine today because registry datasets are consumed read-only for cross-workspace lineage, but there's no compile-time guard — a future mutation of a `.get()`-returned dataset expecting write-through would silently be lost. A note explaining the read-only-snapshot contract would save whoever hits that later.

---

sgomezvillamor: Minor suggestion: consider annotating this as `MutableMapping[str, PowerBIDataset]` rather than the concrete `FileBackedDict[PowerBIDataset]` — all consumers only use `.get`/`.update`/`__getitem__`/`__contains__`. It stops leaking the storage implementation into the attribute type, and lets `test_paginated_report_lineage.py` drop its `# type: ignore[assignment]` when it stubs in a plain dict.

---

sgomezvillamor (approved):
LGTM — the OOM fix is well-targeted and the approach is right: spill workspaces and the dataset registry to `FileBackedDict`, guard the debug `json.dumps(scan_result)`, and keep the M-Query parse tree out of `DataAccessFunctionDetail.repr`. Comments explain the non-obvious bits well.

Approving. A few inline notes, none blocking from my side:
- One `close()` ordering warning I'd fix — a temp-file cleanup failure shouldn't be able to skip the stateful-ingestion commit.
- A test-coverage gap to confirm: the eviction path (write-back persistence + `Workspace` round-trip) is the actual point of the fix but nothing seems to exercise it.
- Minor: use `with` for the local file-backed dict, document the snapshot semantics on `dataset_registry`, and type it as the interface.

---

metadata-ingestion/src/datahub/ingestion/source/powerbi/powerbi.py
acrylJonny: Done in e472280 — switched to `with FileBackedDict(...) as workspaces:` and dropped the manual close.

metadata-ingestion/src/datahub/ingestion/source/powerbi/powerbi.py
acrylJonny: Good catch — fixed in e472280. `powerbi_client.close()` now runs in a `try` with `super().close()` in `finally`, so a temp-file cleanup failure can't skip the state commit.

metadata-ingestion/tests/unit/powerbi/test_fill_metadata_from_scan.py
acrylJonny: Confirmed — the integration test doesn't hit eviction (2 mocked workspaces, cache >=200), so you're right that nothing exercised the write-back or the Workspace round-trip. Closed both gaps in e472280:

- `test_get_workunits_internal_phase1_mutations_survive_eviction`: 15 workspaces with `scan_batch_size=1` (cache ~10 evicts the early ones), mutates each `scan_result` in Phase 1 and asserts it's visible in Phase 2. I verified it fails when the write-back loop is removed and passes with it.
- `test_workspace_survives_eviction_round_trip`: full `Workspace` (raw `scan_result` + nested `PowerBIDataset` with the circular `Table.dataset` ref) through `FileBackedDict(cache_max_size=1)`.

metadata-ingestion/src/datahub/ingestion/source/powerbi/rest_api_wrapper/powerbi_api.py
acrylJonny: Added the read-only-snapshot note in e472280, spelling out that cache-miss lookups return a freshly-unpickled copy (not the instance in `workspace.datasets`) and that registry entries are consumed read-only for cross-workspace lineage.

acrylJonny: Done in e472280 — annotated as `MutableMapping[str, PowerBIDataset]` and dropped the `# type: ignore[assignment]` in test_paginated_report_lineage.py. mypy stays clean.

PR #18395, opened 2026-07-14
feat(ingest/snowflake): ingest semantic views as semanticModel/metric entities

kyungsoo-datahub:
## Summary

Ingests Snowflake semantic views as first-class **`semanticModel`** + **`metric`** entities (#18134 / #18310) instead of datasets-with-subtype. Emission is resolved **against the connected server** via a tri-state `semantic_views.emit_semantic_model_entities` flag; where the entities aren't supported, the legacy dataset representation (subtype "Semantic View") is used unchanged. Write-path only; **no `metadata-models` changes**.

## Flag (tri-state, server-aware)

`emit_semantic_model_entities: Optional[bool] = None`, resolved in the new `snowflake_semantic_model_gate.py`:

| Recipe | Behavior |
| --- | --- |
| unset (default) | Auto-resolved from the server: enabled on a managed server (e.g. DataHub Cloud) at version ≥ 2.1.0 unless Metrics is explicitly disabled; off on OSS/self-hosted, older servers, and connectionless runs (e.g. file sink) |
| `true` | Requested; honored where the server supports the new entities, else warns and falls back to legacy datasets |
| `false` | Always legacy datasets (subtype "Semantic View") |

Version capability is resolved via `server_config.supports_feature(ServiceFeature.SEMANTIC_MODEL_ENTITIES)` (the canonical registry), not an inline version check. `metricsEnabled` vetoes on an explicit `false`; an **absent** field (older server) is treated as not-disabled (fail open), but an **unreadable probe** — an operational auth/network/GraphQL error — now **fails closed** to legacy rather than auto-enabling on an unverified server.

## Emitted (new mode)

Mapping in `snowflake_semantic_model.py`.

- **semanticModel** per view: `semanticModelInfo` (join `relationships` + references to its logical datasets, DDL), plus `subTypes`/`domains`/tags/`status`.
- **logical dataset** per logical table: a `dataset` (subtype "Semantic Model Dataset") carrying `semanticModelProperties`, `schemaMetadata` (that table's dimension/fact fields), and `upstreamLineage` to its physical base table. Per-field semantic metadata lives on `semanticFieldAnnotation` aspects of the `schemaField`s; primary keys, synonyms, and column lineage are scoped per logical table.
- **metric** per METRIC column: `metricInfo` (+ `ModeledBy`) and `metricRelationships` (always emitted so `hasParentMetric` indexes; `derivedFrom` resolves table-qualified metric references). Metrics are keyed by `(logical table, name)`. No `metricUpstreams` for model-backed metrics — lineage flows `Metric → SemanticModel → Logical Dataset → Physical Dataset`.
- Structured-property `entityTypes` for `semanticModel`/`metric` are gated on **server capability** (version, not the emit flag), so the shared definition never flaps.
- No usage statistics on `semanticModel`; `include_usage` is ignored in this mode (legacy mode unchanged). `include_queries` also emits query entities when `include_usage: false` (both modes) — noted in `updating-datahub.md`.

## Known limitations

- Downstream `FROM SEMANTIC_VIEW(x)` can't declare the semanticModel as upstream (`upstreamLineage` is dataset-URN-typed).
- The `semanticModel`/`metric` URNs have no `env` — same account in two environments collides; use a distinct `platform_instance` (logical-dataset URNs are ordinary dataset URNs and keep `env`).
- Read-side rendering is follow-up work.

## Recent correctness fixes

- **Metric identity**: Snowflake allows the same metric name on different logical tables (distinct, table-qualified metrics). Metrics are keyed by `(logical table, name)` and table-bound metric URNs carry the logical table, so siblings no longer collapse into one entity. `derivedFrom` resolves table-qualified references (`ORDERS.GROSS_REVENUE`) instead of dropping every qualified reference.
- **Per-logical-table scoping**: column lineage (including recursive/chained derivation), `isPartOfKey`, and synonyms are resolved from the occurrence on the current logical table, so a same-named column defined differently on two tables no longer cross-contaminates. Lineage is routed by `(logical dataset, column)` + subtype, so a name that is a fact on one table and a metric on another can't dangle to a `schemaField` the dataset never emits.
- **Relationship cardinality** is inferred from the primary key — `ONE_ONE` only when the join columns are the *complete* primary key, else `N_ONE`.
- **Feature gating** uses `ServiceFeature.supports_feature`; an unparseable version or an unreadable `metricsEnabled` probe fails closed to legacy rather than crashing or auto-enabling.
- **Metadata-write robustness**: metric-expression parse failures are surfaced via `report.warning` (catching `SqlglotError`, so tokenizer failures don't abort the view); query/usage extraction narrows its try so downstream errors aren't mislabeled.

## Checklist

- [x] Conforms to the [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (esp. PR Title Format)
- [x] Tests added/updated
- [x] Docs added/updated
- [x] Breaking/notable change entry in [Updating DataHub](https://github.com/datahub-project/datahub/blob/master/docs/how/updating-datahub.md)

---

> [!NOTE]
> **Overview**
> Adds an optional **first-class semantic model ingestion path** for Snowflake semantic views, controlled by **`semantic_views.emit_semantic_model_entities`** (tri-state, resolved at runtime against the connected DataHub server).
> 
> When enabled, each semantic view becomes a **`semanticModel`**, each logical table a **`dataset`** (subtype Semantic Model Dataset) with **`semanticFieldAnnotation`** on columns, and each **METRIC** column a **`metric`** entity—with lineage routed **Metric → SemanticModel → logical dataset → physical table**. The legacy single **dataset** (subtype Semantic View) path remains when the flag is off or the server vetoes emission (version &lt; Cloud 2.1.0, Metrics disabled, or failed capability/probe → **fail closed** to legacy).
> 
> **Server resolution** (`snowflake_semantic_model_gate.py`) auto-enables on DataHub Cloud ≥ 2.1.0 unless `metricsEnabled` is false; OSS stays recipe-driven. **`include_usage`** is ignored in the new mode; **`include_queries`** still emits query entities (including when `include_usage: false`), with subjects pointing at **semanticModel** URNs instead of datasets. Semantic views are **excluded from `discovered_datasets`** in new mode so dataset usage/lineage pipelines do not treat them as ordinary tables.
> 
> Schema extraction gains **`SEMANTIC_RELATIONSHIPS`**, per-table PKs/unique keys, and stricter column-lineage scoping by logical table; structured-property **`entityTypes`** can include semanticModel/metric when the server supports it.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9ffc628ac7f27b533c3c492996a4f04f34dfc3dc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_semantic_model.py
```diff
@@ -0,0 +1,782 @@
...
+        self._warn_field_path_collisions(semantic_view)
+        # Metrics are entities, not fields; seed them as "placed" so the unplaced
+        # check below doesn't flag them.
+        placed_columns: Set[str] = set(metric_occurrences.keys())
+
+        for (
+            logical_name_upper,
+            physical_table,
+        ) in semantic_view.logical_to_physical_table.items():
+            physical_db, physical_schema, physical_table_name = physical_table
```
treff7es: Positional unpack of a 3-tuple. `logical_to_physical_table` is typed `Dict[str, Tuple[str, str, str]]` and its values are literally `SnowflakeTableIdentifier.as_tuple()` — and that dataclass (with `database`/`schema`/`table`) already exists. Typing the field as `Dict[str, SnowflakeTableIdentifier]` would drop the `.as_tuple()` round-trip and let this read `.database/.schema/.table`. The field pre-dates this PR so it's optional, but the new consumer is here.

metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
```diff
@@ -657,3 +660,82 @@ def _warn_view_properties_truncation(
             message="View properties contained too much data and would have caused ingestion to fail",
             context=context,
         )
+
+    def _semantic_model_info_serialized_size(
+        self, semantic_model_info: SemanticModelInfoClass
+    ) -> int:
+        return len(json.dumps(pre_json_transform(semantic_model_info.to_obj())))
+
+    def ensure_semantic_model_info_size(
```
treff7es: This is the third near-verbatim copy of the "blank the field → measure overhead once → `candidate_value`/`fits` closures → binary-search the largest fitting prefix → write back" pattern (`ensure_query_properties_size` and `_shrink_view_properties_field` are the others), and the `candidate_value` closure here is byte-for-byte identical to the view-properties one. A single `_truncate_string_field(get, set, serialize, budget, make_candidate) -> Optional[_TruncationResult]` would collapse all three and remove ~50 lines; the per-aspect orchestration (which field, warn-when-still-oversized) stays per-method.

Separately — the `process()` dispatch branch for `SemanticModelInfoClass` (~line 117) isn't exercised: every test calls `ensure_semantic_model_info_size` directly, so deleting that branch would keep the suite green. A `test_wu_processor_triggered_by_semantic_model_info` mirroring the existing `_by_query_properties_aspect` test would close that gap.

Zooming out — this whole `ensure_aspect_size.py` change (plus its ~166-line test) is generic and Snowflake-agnostic: any connector emitting `semanticModelInfo` benefits, and nothing here depends on the semantic-views feature. It would make a clean standalone PR, which would shrink this one by ~250 lines and narrow its blast radius. Not a blocker — just easier to review and to revert independently if needed.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_semantic_model.py
```diff
@@ -0,0 +1,782 @@
...
+            BrowsePathEntryClass(id=self.identifiers.snowflake_identifier(schema_name))
+        )
+        return entries
+
+    def _gen_view_tags(
+        self, semantic_view: SnowflakeSemanticView, model_urn: str
+    ) -> Iterable[MetadataWorkUnit]:
+        yield from self._emit_tags_for_entity(model_urn, semantic_view.tags or [])
+
+    def _emit_tags_for_entity(
```
treff7es, line 874: `_emit_tags_for_entity` and `_column_tags` build the identical `GlobalTags` comprehension, and `_emit_tags_for_entity` + `_gen_field_structured_property_workunits` build the identical structured-property dict. Two small helpers (`_tag_associations(tags)` and `_structured_property_values(tags)`) would de-dupe both and give one place to keep the `snowflake_identifier(tag.*_identifier())` construction correct.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_semantic_model_gate.py
```diff
@@ -0,0 +1,197 @@
...
+
+# A GraphQL error carrying both markers means the server predates metricsEnabled;
+# any other error is re-raised for the resolver, which treats a fetch failure as
+# not-disabled (fail-open), never as a veto.
+_FIELD_UNDEFINED_MARKER = "FieldUndefined"
+_METRICS_ENABLED_FIELD = "metricsEnabled"
+
+
+@dataclass
+class ResolvedEmitDecision:
```
treff7es, line 43: `ResolvedEmitDecision` is computed once and consumed once — a pure value object — so `@dataclass(frozen=True)` would lock that in. (The per-field docstrings here, especially on `entity_types_capable`, are excellent.)

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_semantic_model.py
kyungsoo-datahub: Agreed it reads better. Since the field pre-dates this PR and retyping it touches the producer (as_tuple()) and the other consumers too, I'd rather not widen the diff here. I'll take it as a follow-up.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_semantic_model.py
kyungsoo-datahub: Done

metadata-ingestion/src/datahub/ingestion/workunit_processors/ensure_aspect_size.py
kyungsoo-datahub: Split out to #18608(https://github.com/datahub-project/datahub/pull/18608) ensure_aspect_size.py + its tests are gone from this PR (this branch no longer touches the file).

That PR also adds the test_wu_processor_triggered_by_semantic_model_info you asked for. The shared _truncate_string_field dedup makes sense but touches the pre-existing query/view-properties paths, so I left it for #18608 where those methods now live.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_semantic_model_gate.py
kyungsoo-datahub: Done.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_semantic_model_gate.py
```diff
@@ -0,0 +1,197 @@
...
+
+    server_config = graph.server_config
+    is_saas = server_config.is_datahub_cloud
+    version = server_config.service_version
+
+    # Capability for the structuredPropertyDefinition entityTypes permission list.
+    # Computed before every early return so it is identical across recipe values
+    # against the same server (no definition flap). SaaS: version only; OSS: recipe.
+    entity_types_capable = (
+        server_config.is_version_at_least(*_MIN_SAAS_VERSION)
```
askumar27: `is_version_at_least` raises `ValueError` on a non-semver `service_version` (server_config_util.py:121), and this call sits outside the try/except that guards the metricsEnabled probe below. The resolver runs by default on Cloud, so a cosmetic version string (git-sha tag, 4-part, `-rc` suffix) aborts the entire Snowflake source, not just this feature. Wrap it and fail closed to legacy + `report.warning`, like the metricsEnabled block does.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_semantic_view_usage.py
```diff
@@ -317,12 +373,17 @@ def get_semantic_view_query_workunits(
...
+                canonical_identifier = identifier_lookup[normalized_name]
+                logger.debug(
+                    f"Emitting {len(queries)} query entities for {canonical_identifier}"
+                )
 
                 for query in queries:
-                    yield from self._build_query_workunits(query, view_name)
+                    yield from self._build_query_workunits(query, canonical_identifier)
 
         except Exception as e:
```
askumar27, line 385: This `except Exception` wraps `yield from self._build_query_workunits(...)`, so a downstream consumer error is caught and mislabeled as a query-extraction failure — and it only `logger.warning`s, so a real permission/query failure on QUERY_HISTORY silently yields zero query entities, indistinguishable from "no queries." Narrow the try to the query + parse, `report.warning` the failure, and run the emit loop outside it. Same yield-wrapping in `get_semantic_view_usage_workunits` above (that one does report — just move the loop out of the try).

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_semantic_model.py
```diff
@@ -0,0 +1,934 @@
...
+        if not occurrence.expression:
+            return []
+        try:
+            parsed = sqlglot.parse_one(occurrence.expression, dialect="snowflake")
+        except Exception as e:
+            logger.debug(
+                f"Failed to parse metric expression for {semantic_view.name}."
+                f"{occurrence.name} ({occurrence.expression!r}): {e}"
+            )
+            self.report.num_semantic_view_metric_expr_parse_failures += 1
```
askumar27, line 768: A metric whose expression fails to parse drops its `derivedFrom` lineage via `logger.debug` + a bare counter — invisible in the report, and a non-zero count reads as benign. Surface it with `report.warning` + the metric name, and catch `sqlglot` `ParseError` rather than bare `Exception`.

askumar27 (commented):
W1 (a version string can crash all Snowflake ingestion) and W2 (except wraps the yield) are the ones to fix before merge. Notes inline. S1-type schema.py column-fetch code is pre-existing from the master merge, so not flagged here.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_semantic_model_gate.py
askumar27: Fixed in 0ceda54. Wrapped the version check in try/except ValueError so a non-semver version can't abort the whole source — fails closed to legacy now. Added a version_unparseable flag on the decision so snowflake_v2 raises a report.warning even on the auto-enable path (recipe=None), where the existing recipe=true warning wouldn't fire.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_semantic_view_usage.py
askumar27: Fixed in 0ceda54. Narrowed the try to the fetch+grouping, moved the emit loop out, and switched to report.warning. Did the same to get_semantic_view_usage_workunits (materialize records inside the try, yield outside).

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_semantic_model.py
askumar27: Fixed in 0ceda54. report.warning with the metric name now, and catching sqlglot.errors.ParseError instead of a bare Exception. Kept the counter.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_semantic_model.py
```diff
@@ -0,0 +1,934 @@
...
+                # Uppercased to match each logical dataset's
+                # semanticModelProperties.alias, so join references resolve.
+                from_=relationship.from_table.upper(),
+                fromColumns=relationship.from_columns,
+                to=relationship.to_table.upper(),
+                toColumns=relationship.to_columns,
+                # FK joins are many-to-one; Snowflake exposes no cardinality column.
+                cardinality=ERModelRelationshipCardinalityClass.N_ONE,
+            )
+            for relationship in semantic_view.relationships
```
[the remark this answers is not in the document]
askumar27: Fixed in 0ceda54. fromColumns/toColumns now go through snowflake_identifier(col.upper()), the same normalization as the schemaField paths, so they line up under convert_urns_to_lowercase=true. Verified against a real account (WAREHOUSE_COFFEE_COMPANY, lowercasing on): 21 relationships / 82 join columns, all lowercased, 80/82 resolve to field paths. The 2 that don't are a modeling thing — the join references a column that isn't exposed as a dimension/fact on that logical table, not a casing issue.

metadata-ingestion/src/datahub/ingestion/source/snowflake/snowflake_semantic_model.py
```diff
@@ -0,0 +1,953 @@
...
+            # A metric whose expression won't parse loses its derivedFrom lineage.
+            # Surface it (not just a debug log + counter) so operators can see which
+            # metric was affected; the metric itself is still emitted without edges.
+            self.report.warning(
+                title="Could not parse semantic view metric expression",
+                message=(
+                    "A metric expression failed to parse, so its metric-to-metric "
+                    "derivedFrom lineage was skipped. The metric is still emitted."
+                ),
+                context=f"{semantic_view.name}.{occurrence.name}: {e}",
```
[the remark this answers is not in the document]
askumar27: Good catch. Fixed in 18f8da5 — widened the catch to the sqlglot.errors.SqlglotError base so TokenError (unterminated quote, etc.) is handled too, not just ParseError. Added test_derived_from_tokenizer_failure_yields_no_edges to lock it in: the bad metric still emits with an empty derivedFrom and the rest of the view is unaffected.

askumar27 (approved):

PR #18402, opened 2026-07-14
feat(graphql): add field resolvers for Metric and SemanticModel entities

ani-malgari:
## Summary

Wires up GraphQL field resolvers for the `Metric` and `SemanticModel` entity types (schema, types, and top-level list/get queries landed separately):

- `Metric.childMetrics(input: ScrollAcrossEntitiesInput!): ScrollResults` — searches metrics scoped to `parentMetric=<urn>`, hardcoded to the `METRIC` entity type.
- `Metric.parentMetric: Metric` — direct parent lookup via `LoadableTypeResolver`, reading from `MetricRelationships.parentMetric`. Also wires `MetricRelationships.parentMetric` for nested selection.
- `Metric.semanticModel`, `Metric.platform`, `Metric.dataPlatformInstance` — standard `LoadableTypeResolver` wiring.
- `SemanticModel.metrics(input: ScrollAcrossEntitiesInput!): ScrollResults` — searches metrics scoped to `semanticModel=<urn>`, hardcoded to the `METRIC` entity type.
- `SemanticModel.platform`, `SemanticModel.dataPlatformInstance` — standard `LoadableTypeResolver` wiring.
- Standard `Entity`-interface fields (`privileges`, `exists`, `relationships`) wired for both types.

### DIFF
--- a/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/GmsGraphQLEngine.java
+++ b/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/GmsGraphQLEngine.java
@@ -4074,17 +4074,6 @@ public class GmsGraphQLEngine {
                 .dataFetcher("privileges", new EntityPrivilegesResolver(entityClient))
                 .dataFetcher("exists", new EntityExistsResolver(entityService))
                 .dataFetcher("relationships", new EntityRelationshipsResultResolver(graphClient)));
-    builder.type(
-        "MetricRelationships",
-        typeWiring ->
-            typeWiring.dataFetcher(
-                "parentMetric",
-                new LoadableTypeResolver<>(
-                    metricType,
-                    env -> {
-                      final MetricRelationships rel = env.getSource();
-                      return rel.getParentMetric() != null ? rel.getParentMetric().getUrn() : null;
-                    })));
   }
 
   private void configureSemanticModelResolvers(final RuntimeWiring.Builder builder) {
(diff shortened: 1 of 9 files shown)


datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/GmsGraphQLEngine.java
```diff
@@ -4027,4 +4031,89 @@ private void configureAssetSettingsResolver(final RuntimeWiring.Builder builder)
...
+                    "childMetrics", new MetricChildMetricsResolver(entityClient, viewService))
+                .dataFetcher(
+                    "parentMetric",
+                    new LoadableTypeResolver<>(
+                        metricType,
+                        env -> {
+                          final Metric m = env.getSource();
+                          if (m.getMetricRelationships() == null
+                              || m.getMetricRelationships().getParentMetric() == null) {
+                            return null;
```
chriscollins3456: shouldn't parentMetric be an aspect fetch that we can just partially set with urn and type and use that there to hydrate? or is the relationship defined differently

just wondering in case metricRelationships doesn't get set that will block parentMetric from being hydrated

chriscollins3456: it's also a little interesting you can get parentMetric both top level and inside of metricRelationships. should it just be in one or the other?

chriscollins3456 (commented):
quick comments but in general makes sense to me

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/GmsGraphQLEngine.java
ani-malgari: sorry this is inaccurate, let me fix this. the parentMetric should just be an aspect fetch indeed

ani-malgari: @chriscollins3456 Dropped MetricRelationships.parentMetric from the GraphQL type. I have it inside metricRelationships in the PDL, moved it to the root level on GraphQL side is the cleanest way to query

chriscollins3456 (approved):

PR #18407, opened 2026-07-15
feat(metrics): add graphql fragments, metric and semantic model entity pages

ani-malgari:
## Summary

- Add GraphQL fragments and queries for `Metric`, `SemanticModel`, and metrics browse (`metric.graphql`, `semanticModel.graphql`, `metricsBrowse.graphql`)
- Register `MetricEntity` and `SemanticModelEntity` stubs in the entity registry with preview cards and placeholder profile pages
- Introduce `MetricsRoutes` as the sole owner of `MetricsSidebar`, routing `/metrics`, `/metric/:urn`, and `/semanticModel/:urn`; simplify `MetricsPage` to content-only and exclude Metric/SemanticModel from the generic entity route loop in `SearchRoutes`

## Test plan

- [x] `METRICS_ENABLED=true` set and containers restarted
- [x] Visit `/metric/<urn>` — placeholder profile renders with metric icon and name
- [x] Visit `/semanticModel/<urn>` — placeholder profile renders with platform icon and name

Screenshot 2026-07-15 at 2 27 52 PM


datahub-web-react/src/app/metrics/MetricsRoutes.tsx
```diff
@@ -0,0 +1,73 @@
...
+            <MainContent>
+                <Switch>
+                    <Route
+                        path={`${PageRoutes.METRIC_ENTITY}/:urn`}
+                        render={() => <EntityPageV2 entityType={EntityType.Metric} />}
+                    />
+                    <Route
+                        path={`${PageRoutes.SEMANTIC_MODEL_ENTITY}/:urn`}
+                        render={() => <EntityPageV2 entityType={EntityType.SemanticModel} />}
+                    />
```
chriscollins3456, line 67: i think these two routes should be handled automatically like any other entity route in SearchRoutes:

```
                    {entities.map((entity) => (
                        <Route
                            key={entity.getPathName()}
                            path={`/${entity.getPathName()}/:urn`}
                            render={() => <EntityPageV2 entityType={entity.type} />}
                        />
                    ))}

```

datahub-web-react/src/app/SearchRoutes.tsx
```diff
@@ -90,7 +94,16 @@ export const SearchRoutes = (): JSX.Element => {
                         render={() => <EntityPageV2 entityType={entity.type} />}
                     />
                 ))}
-                {metricsEnabled && <Route path={PageRoutes.METRICS} render={() => <MetricsPage />} />}
+                {metricsEnabled && (
+                    <Route
+                        path={[
+                            `${PageRoutes.METRIC_ENTITY}/:urn`,
+                            `${PageRoutes.SEMANTIC_MODEL_ENTITY}/:urn`,
```
chriscollins3456, line 101: and if we let `entities.map(...` handle these routes we will have to remove them here

datahub-web-react/src/app/metrics/MetricsRoutes.tsx
chriscollins3456: ahhh sorry now i see that we have the left sidebar for metrics routes that we need to handle like glossary terms/nodes and domains

feel free to ignore

chriscollins3456 (approved):
looking good!

PR #18414, opened 2026-07-15
fix(ingest): filter no-op delimiter nodes in Block statement handling

sgomezvillamor:
## Summary

- Filter `Semicolon` and `EndStatement` no-op delimiter nodes when unwrapping sqlglot `Block` results, fixing ~2,600 view lineage parse failures per Snowflake ingestion run
- Apply the same filter in `parse_statements_and_pick()`, which was silently picking a `Semicolon` node as the "main query" via `statements[-1]`
- Add `EndStatement` filtering for T-SQL `BEGIN/END` wrappers

## Evidence

From a Snowflake ingestion run (DataHub CLI 1.6.0):

- `sql_aggregator.num_view_definitions: 79,305`
- `sql_aggregator.num_views_failed: 2,606`
- Error: `Block contains 2 statements: ['Create', 'Semicolon']. Use parse_statements_and_pick() for multi-statement SQL.`
- All sampled failures are `CREATE OR REPLACE VIEW ... COPY GRANTS` statements

## Root cause

Verified against sqlglot 30.12.0: a bare trailing semicolon does *not* produce `Block([stmt, Semicolon])` — it parses to a plain statement node. The `Block([stmt, Semicolon])` shape only occurs when the semicolon has an **attached comment** (`; -- foo` or `; /* foo */`).

Two parsing paths were affected:
- **`_parse_statement`**: raised `"Block contains 2 statements"` — the reported production error
- **`parse_statements_and_pick`** (used by dbt, Mode): silently picked the `Semicolon` node as the "main query" via `statements[-1]` — a latent silent-corruption bug

The fix introduces a shared `_NOOP_EXPRESSION_TYPES` tuple of `(Semicolon, EndStatement)` and filters these in both paths.

## Test plan

- [x] `test_parse_statement_filters_noop_block_nodes` — parametrized: line comment, block comment, T-SQL BEGIN/END, T-SQL BEGIN/END + comment
- [x] `test_parse_statements_and_pick_filters_noop_nodes` — parametrized: dbt-style trailing comment, CTE + comment, T-SQL BEGIN/END
- [x] `test_create_view_block_semicolon_with_comment` — end-to-end lineage golden test with SQL that produces `Block([Create, Semicolon])`
- [x] `test_snowflake_create_view_copy_grants_with_casts` — COPY GRANTS lineage coverage (standalone, does not exercise the Block fix)
- [x] Existing `test_create_view_as_block_statement`, `test_block_with_all_none_expressions`, `test_block_with_multiple_statements` — still pass
- [x] `lintFix` clean

---

> [!NOTE]
> **Overview**
> Fixes Snowflake view lineage failures when sqlglot wraps a single statement in a `Block` that also includes delimiter-only nodes (e.g. `Block([Create, Semicolon])` from `; -- comment`, or T-SQL `EndStatement` from `BEGIN … END`).
> 
> **`sqlglot_utils`:** Introduces `_NOOP_EXPRESSION_TYPES` (`Semicolon`, `EndStatement`) and uses it when unwrapping `Block` in `_parse_statement` (alongside existing `None` filtering) and when building the statement list in `parse_statements_and_pick`, so delimiter nodes are not treated as extra statements or picked as the “main” query.
> 
> **Tests:** Parametrized unit tests for both parsers; lineage goldens for trailing-semicolon+comment `CREATE VIEW` and Snowflake `COPY GRANTS` view with casts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dd0cedffad9ad5d6fce4e847553584ec94c8f1f5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


metadata-ingestion/src/datahub/sql_parsing/sqlglot_utils.py
```diff
@@ -120,17 +120,23 @@ def _parse_statement(
         )
 
     # Handle Block statements from sqlglot v29+
-    # Sqlglot parses SQL with double semicolons (e.g., "CREATE VIEW ...;\n;") as
-    # Block([stmt1, None, ...]) where None represents empty statements between semicolons.
-    # We only process the non None statement, if there is 1 and only 1.
+    # Sqlglot wraps multi-expression results as Block([stmt, ...]). Two known
+    # cases produce no-op extra nodes we must discard:
+    #  1. Double semicolons ("CREATE VIEW ...;\n;") → Block([Create, None])
+    #  2. Trailing semicolons → Block([stmt, Semicolon])
```
alokr-dhub: Verified against sqlglot 30.12.0 (the pinned version) by tracing the parser's chunking logic (`parser.py:2160-2163`) and by direct execution: a bare trailing semicolon (`...;`) does **not** produce `Block([stmt, Semicolon])` — it parses straight to the statement node (e.g. `Create`), same as before this fix. `Block([stmt, Semicolon])` only occurs when the semicolon token has an **attached comment** (`...; -- foo` or `...; /* foo */`).

This comment (and the commit message) should say "trailing semicolon with an attached comment", not "trailing semicolons" generally — as written, a maintainer who tries to repro with a bare `;` will conclude the fix is unreachable/dead code (which is what happened during this review before the exact trigger was traced through the sqlglot source).

metadata-ingestion/src/datahub/sql_parsing/sqlglot_utils.py
```diff
@@ -120,17 +120,23 @@ def _parse_statement(
...
                 "Block statement must have at least one expression"
             )
 
-        # Filter out None expressions (empty statements from double semicolons)
-        non_none_expressions = [e for e in statement.expressions if e is not None]
+        # Filter out no-op artifacts: None and Semicolon nodes
+        non_none_expressions = [
+            e
+            for e in statement.expressions
+            if e is not None and not isinstance(e, sqlglot.exp.Semicolon)
```
alokr-dhub: Two smaller issues here:

1. `non_none_expressions` (line 135) is now a stale name — it filters more than `None`.
2. The filter only covers `None` and `Semicolon`. `sqlglot.exp.EndStatement` is a semantically identical no-op node that genuinely appears inside real `Block` trees for T-SQL `BEGIN...END` constructs (confirmed via direct parsing), and it isn't filtered here. If a BEGIN/END-wrapped statement lands in this function, it can still hit the "Block contains N statements" `ParseError` this PR is trying to fix for the Semicolon case.

Don't over-build for this (no registry needed per YAGNI) — but if `EndStatement` is confirmed to occur in practice, a two-element tuple `(sqlglot.exp.Semicolon, sqlglot.exp.EndStatement)` closes the gap with the same line count.

metadata-ingestion/src/datahub/sql_parsing/sqlglot_utils.py
```diff
@@ -120,17 +120,23 @@ def _parse_statement(
...
+        ]
 
         if not non_none_expressions:
             raise sqlglot.errors.ParseError(
```
alokr-dhub, line 147: The `ParseError` message on the next line — `"Block contains only None expressions - no valid SQL statement found"` — is now stale: it also fires for a `Block` containing only `Semicolon` nodes, which isn't "only None expressions." This will mislead anyone debugging a production ingestion failure by grepping this exact string. Suggest: `"Block contains no executable statements - no valid SQL statement found"` (or similar), which stays accurate regardless of which no-op node types get added to the filter later.

metadata-ingestion/tests/unit/sql_parsing/test_sqlglot_utils.py
```diff
@@ -483,3 +483,19 @@ def test_tsql_digit_leading_temp_table_lineage() -> None:
     )
     assert result.debug_info.table_error is None, result.debug_info.table_error
     assert any("real_src" in t for t in result.in_tables), result.in_tables
+
+
+def test_block_with_semicolon_node() -> None:
+    """Block([Create, Semicolon]) must unwrap to the Create node.
```
alokr-dhub: This docstring says "a trailing semicolon can cause sqlglot to produce a Block..." — per my comment on `sqlglot_utils.py:126`, that's not accurate for a bare trailing semicolon; the real trigger needs a comment attached to the semicolon token. Please correct the docstring so a future reader trying to reproduce this doesn't use a bare `;` and wrongly conclude the fix is unreachable.

Also: this test hand-constructs `Block(expressions=[Create, Semicolon])` and bypasses the `lru_cache` via `.__wrapped__` (line 500) rather than parsing real SQL. I confirmed the hand-built shape does match what real SQL produces (semicolon + trailing comment) — but since that real repro exists and is short, consider using it directly via `sqlglot.maybe_parse(...)` instead of hand-building the AST; it would also make the `__wrapped__` cache-bypass unnecessary (the reason for bypassing isn't stated — I confirmed it's not unhashability, `Expression.__hash__`/`__eq__` work fine here; it appears to be to avoid polluting the shared `lru_cache` with a synthetic object, which is reasonable but should be a one-line comment so the next reader doesn't assume `Block` is unhashable).

metadata-ingestion/tests/unit/sql_parsing/test_sqlglot_lineage.py
```diff
@@ -191,6 +191,31 @@ def test_snowflake_create_view_with_tag() -> None:
     )
 
 
+def test_snowflake_create_view_copy_grants_with_casts() -> None:
```
alokr-dhub, line 194: Verified by parsing this exact SQL string with sqlglot 30.12.0, and by running this test against the pre-fix code (HEAD~1): it passes identically either way. The SQL here has no trailing semicolon, so it parses straight to a `Create` node and never reaches the `Block` branch this PR changed — this test doesn't exercise the fix at all. It's a fine standalone lineage test for COPY GRANTS + casts, but it isn't a regression test for this bug.

The existing sibling tests just below (`test_create_view_as_block_statement`, `test_block_with_all_none_expressions`, `test_block_with_multiple_statements`) all reproduce their target `Block` shape end-to-end from real SQL via `assert_sql_result`/`sqlglot_lineage`. Recommend following that same pattern here with SQL that actually produces `Block([Create, Semicolon])` — a trailing `;` followed by a comment, e.g. `...FROM t; -- trailing comment` — which I confirmed does hit the new code path.

alokr-dhub (commented):
Reviewed the shared `sql_parsing` chokepoint change (not connector-specific) via incremental review. The mechanism (filtering a no-op `Block` node) is sound, but I verified two things directly against the pinned `sqlglot[c]==30.12.0` that affect merge-readiness — flagging as a top-level summary since they're outside this diff's hunks:

**1. `parse_statements_and_pick()` (`sqlglot_utils.py:186-189`) is a second, independent Block/no-op-unwrapping site and was not given the same guard.** It filters only `if expression` (truthy), so a stray `Semicolon` (or `EndStatement`) surviving there could become `statements[-1]` — silently "picked" as the query. It's used by `mode.py:1269` and `dbt/dbt_common.py:2502`, exactly the multi-statement-SQL callers most likely to hit this. If the premise of this PR is right, this sibling site has the identical bug, unfixed.

**2. `parse_statement`'s public signature (`sqlglot_utils.py:160`, `sql: sqlglot.exp.ExpOrStr`) accepts an `Expression` into the `lru_cache`-decorated `_parse_statement`.** `Expression` is hashable (structural, content-based hash) but is documented elsewhere in this same file as mutable (hence the `.copy()` on return at line 177). A mutable, content-hashed object as a cache key is unsound if ever mutated post-insertion — no current caller triggers it (all pass `str`), but the signature invites it.

**3. Before merging, please confirm against the actual failing SQL samples from the 2,606 production failures that they contain a comment attached to the trailing semicolon** (see inline comment on line 126) — otherwise this fix may not resolve the reported incident.

See inline comments for the rest.

alokr-dhub (approved):
Approved with minor suggestion. Must fix: `test_snowflake_create_view_copy_grants_with_casts` SQL has no trailing semicolon and parses straight to a plain Create node and does not test the fix.

metadata-ingestion/src/datahub/sql_parsing/sqlglot_utils.py
sgomezvillamor: You're right — verified empirically against sqlglot 30.12.0: a bare trailing semicolon parses straight to the statement node. `Block([stmt, Semicolon])` only occurs when the semicolon has an attached comment (`; -- foo` or `; /* foo */`).

Fixed — comment now reads:
```
#  2. Semicolons with attached comments ("...;\n-- comment") → Block([stmt, Semicolon])
```

Also added `test_create_view_block_semicolon_with_comment` as an end-to-end lineage test using real SQL that produces this exact shape (`CREATE VIEW ... ; -- end of view`).

metadata-ingestion/src/datahub/sql_parsing/sqlglot_utils.py
sgomezvillamor: Both points addressed:

1. Renamed `non_none_expressions` → `real_expressions` since it now filters more than just `None`.

2. Added `EndStatement` to the filter as a tuple: `_NOOP_TYPES = (sqlglot.exp.Semicolon, sqlglot.exp.EndStatement)`. Verified empirically that `maybe_parse("BEGIN SELECT 1; END;", dialect="tsql")` produces `Block([Command, EndStatement])` on sqlglot 30.12.0, so this is reachable. Added parametrized test cases covering both T-SQL shapes:
   - `BEGIN SELECT 1; END;` → `Block([Command, EndStatement])`
   - `BEGIN SELECT 1; END; -- done` → `Block([Command, EndStatement, Semicolon])`

metadata-ingestion/src/datahub/sql_parsing/sqlglot_utils.py
sgomezvillamor: Fixed — error message now reads: `"Block contains no executable statements - no valid SQL found"`.

metadata-ingestion/tests/unit/sql_parsing/test_sqlglot_utils.py
sgomezvillamor: Both points addressed:

1. Docstring corrected — now says "A trailing semicolon with an attached comment" rather than just "a trailing semicolon."

2. Replaced the hand-constructed AST test entirely with a parametrized `test_parse_statement_filters_noop_block_nodes` that parses real SQL through the public `parse_statement` API:

```python
@pytest.mark.parametrize("sql, dialect", [
    ("CREATE VIEW v AS SELECT 1 FROM t; -- trailing comment", "snowflake"),
    ("CREATE VIEW v AS SELECT 1 FROM t; /* trailing */", "snowflake"),
    ("BEGIN SELECT 1; END;", "tsql"),
    ("BEGIN SELECT 1; END; -- done", "tsql"),
])
```

Each case produces the actual `Block` shape end-to-end (verified against sqlglot 30.12.0), no `__wrapped__` bypass needed.

metadata-ingestion/tests/unit/sql_parsing/test_sqlglot_lineage.py
sgomezvillamor: Correct — this test doesn't exercise the Block+Semicolon fix. The SQL has no trailing semicolon so it parses straight to a `Create` node. It's standalone COPY GRANTS lineage coverage, not a regression test for this bug.

Added `test_create_view_block_semicolon_with_comment` as the proper end-to-end regression test — it uses SQL that actually produces `Block([Create, Semicolon])`:

```sql
CREATE VIEW my_view AS SELECT id, name FROM my_db.my_schema.my_table
WHERE active = TRUE; -- end of view
```

This follows the same pattern as the sibling tests (`test_create_view_as_block_statement`, etc.) — real SQL through `assert_sql_result`/`sqlglot_lineage`, with a golden file.

sgomezvillamor:
@alokr-dhub All six concerns addressed in the updated commit (c9e3363). Addressed each inline comment individually above — here's a summary of the top-level points:

**1. `parse_statements_and_pick()` sibling bug — fixed.** Verified empirically that `sqlglot.parse("SELECT 1; -- comment", dialect="snowflake")` returns `[Select, Semicolon]`. The truthy `if expression` filter doesn't remove `Semicolon` (it's truthy), so it would be silently picked as the "main query" via `statements[-1]`. Added the same `_NOOP_TYPES` filter and parametrized tests covering dbt-style and CTE SQL.

**2. Mutable Expression as lru_cache key** — acknowledged. No current caller passes an `Expression` (all pass `str`), so this is latent. The `ExpOrStr` signature is inherited from sqlglot's own API and predates this PR. Not changing here.

**3. Trailing semicolon + comment trigger — confirmed.** Verified against sqlglot 30.12.0: a bare `;` never produces `Block([stmt, Semicolon])`. The trigger requires a comment attached to the semicolon token. Comments, error messages, and docstrings updated to reflect the precise trigger.

**Also included:** `EndStatement` filtering (item 3 from inline), stale error message fix, variable rename `non_none_expressions` → `real_expressions`, and all tests rewritten to use real SQL through the public API instead of hand-constructed AST nodes.

sgomezvillamor:
Thanks for the thorough review @alokr-dhub! 🙏

PR #18415, opened 2026-07-15
test(playwright): reconcile ingestion tests with SaaS

v-tarasevich-blitz-brain:
This PR brings fixes of ingestion PW tests to OSS

### DIFF
--- a/e2e-test/ui/playwright/utils/constants.ts
+++ b/e2e-test/ui/playwright/utils/constants.ts
@@ -26,6 +26,7 @@ export const TIMEOUTS = {
   OPERATION: 1500, // Async operations (document updates, searches, input debounce delays)
   BETWEEN_OPS: 500, // Wait between sequential operations
   QUICK: 300, // Brief pause for rendering
+  INGESTION_EXECUTION: 100000, // Ingestion source execution completion
 } as const;
 
 // Interaction delays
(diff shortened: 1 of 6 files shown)


e2e-test/ui/playwright/utils/constants.ts
```diff
@@ -75,3 +76,10 @@ export const DATA_SOURCES = {
   MYSQL: 'mysql',
   SNOWFLAKE: 'snowflake',
 };
+
+// Pinned datahub CLI version for ingestion run/execute tests. Pinning a fixed
+// version keeps the executor's package resolution deterministic and fast enough
+// to finish within the test's status timeout. Must stay in sync with the
+// `config.version` pinned on the seeded execute sources in the
+// ingestion-v2/v3 fixtures.
+export const INGESTION_CLI_VERSION = '1.6.0.10';
```
devashish2203, line 85: How frequently do we need to update this? Does this need to keep up with the version changes done for the default Cli version in general?

devashish2203 (approved):

e2e-test/ui/playwright/utils/constants.ts
v-tarasevich-blitz-brain: I see that in cypress we had fixed stale version 0.15.0.5. And I don't see if we updated it ever. I think it make sense to keep it updated with the latest stable releases.

devashish2203: Hmm. Ohk. let's merge this and add a ticket to figure out how to keep this updated.

PR #18416, opened 2026-07-15
fix(ci): declare frontend task deps on graphqlUsageRegistryCheck to fix parallel-build race

Dutt23:
## Summary

`:metadata-service:configuration:graphqlUsageRegistryCheck` scans `datahub-web-react/src` for `*.graphql` files, but that directory is written by `:datahub-web-react:yarnGenerate` and `:datahub-web-react:generateLazyIconStubs`. The check never declared a dependency on those tasks.

This was latent until #18375 turned on Gradle parallelism (`--max-workers=nproc`). Under parallel scheduling, Gradle's undeclared-dependency validation fails the build when the check runs *after* the producers have registered their output:

> Task ':metadata-service:configuration:graphqlUsageRegistryCheck' uses this output of task ':datahub-web-react:yarnGenerate' without declaring an explicit or implicit dependency.

Fix: declare `dependsOn` on both producer tasks so ordering is deterministic. This is solution #2 from Gradle's own error message.

Checked manually in
Screenshot 2026-07-15 at 4 32 12 PM
 pipeline tasks run in order

## Why it only broke on master

The failure is order-sensitive, and execution order is decided by which subtree compiles first:

| Commit | Build cache | Check ran | Producers ran | Result |
|--------|-------------|-----------|---------------|--------|
| PR head ([4dbee74]) | warm — `:check UP-TO-DATE` | 07:59:01 (first) | 07:59:51 | ✅ pass |
| master squash ([bbc6f03]) | cold | 09:47:40 (after) | 09:45:44 | ❌ fail |

On PR runs the backend was cached, so `configuration:testClasses` was ready immediately and the check ran before the frontend generate tasks — the overlap never registered. The master squash-merge landed on a cold cache, the backend compiled fresh and finished *after* the frontend tasks, so the check read an already-written `datahub-web-react/src` with no declared edge and validation tripped. Same code, opposite race outcome.

The `dependsOn` edge removes the race entirely.

### DIFF
--- a/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/GraphQLEngine.java
+++ b/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/GraphQLEngine.java
@@ -47,7 +47,7 @@ import org.dataloader.DataLoader;
  * <p>This class provides a {@link Builder} builder for constructing {@link GraphQL} instances
  * provided one or more schemas, {@link DataLoader}s, & a configured {@link RuntimeWiring}.
  *
- * <p>In addition, it provides a simplified 'execute' API that accepts a 1) query string and 2) set
+ * <p>In additio, it provides a simplified 'execute' API that accepts a 1) query string and 2) set
  * of variables.
  */
 @Slf4j
--- a/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/GraphQLEngine.java
+++ b/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/GraphQLEngine.java
@@ -47,7 +47,7 @@ import org.dataloader.DataLoader;
  * <p>This class provides a {@link Builder} builder for constructing {@link GraphQL} instances
  * provided one or more schemas, {@link DataLoader}s, & a configured {@link RuntimeWiring}.
  *
- * <p>In additio, it provides a simplified 'execute' API that accepts a 1) query string and 2) set
+ * <p>In addition, it provides a simplified 'execute' API that accepts a 1) query string and 2) set
  * of variables.
  */
 @Slf4j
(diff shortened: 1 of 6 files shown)


datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/GraphQLEngine.java
```diff
@@ -47,7 +47,7 @@
  * <p>This class provides a {@link Builder} builder for constructing {@link GraphQL} instances
  * provided one or more schemas, {@link DataLoader}s, & a configured {@link RuntimeWiring}.
  *
- * <p>In addition, it provides a simplified 'execute' API that accepts a 1) query string and 2) set
+ * <p>In additio, it provides a simplified 'execute' API that accepts a 1) query string and 2) set
```
devashish2203: nit: I assume this is just to trigger CI and will be reverted.

datahub-web-react/src/App.tsx
```diff
@@ -26,7 +26,7 @@ import possibleTypesResult from '@src/possibleTypes.generated';
 import { getRuntimeBasePath, removeRuntimePath, resolveRuntimePath } from '@utils/runtimeBasePath';
 
 /*
-    Construct Apollo Client
+    Construct
```
devashish2203: nit: I assume this is just to trigger CI and will be reverted.

devashish2203 (approved):

datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/GraphQLEngine.java
Dutt23: yeah reverted now.

datahub-web-react/src/App.tsx
Dutt23: yeah reverted now.

gdatahub (approved):

PR #18418, opened 2026-07-15
fix(timeline): resolve ownership type name in Change History; fix raw diff serialization

max-datahub:
## Summary

Two small, independent fixes on the Change History / Timeline surface.

### 1. Change History shows the ownership type's display name (frontend)

The Change History entry for an ownership change rendered the type by humanizing its
URN slug (`formatOwnerTypeUrn`), so it did **not** match the name shown elsewhere in
the UI (ownership sidebar, remove-owner dialog), which resolve the ownership-type
entity's live `info.name`. Effects:

- **Custom ownership types** (opaque UUID URNs) rendered the UUID instead of the name.
- **Renamed built-in types** rendered the default slug (e.g. `__system__data_steward`
  → "Data Steward") even after the type was renamed, so the same owner showed a
  different type label in Change History vs. the ownership panel.

The name resolver (`useResolveEntityNames`) already batch-resolves every URN-valued
parameter — including `ownerTypeUrn` — into `nameMap`, and the query fragment already
selects `OwnershipTypeEntity { info { name } }`. This change just consults it, falling
back to slug humanization only when the type can't be resolved (e.g. it was deleted).

Note: this resolves the type's **current** display name (consistent with the rest of
the UI). Showing the name as it was at the time of each event would require capturing
the name with the event and is intentionally out of scope.

### 2. `/timeline` OpenAPI endpoint fails when `raw=true` (backend)

`ChangeTransaction.rawDiff` is a `jakarta.json.JsonPatch` (parsson `JsonPatchImpl`),
which is not a Jackson-serializable bean — serializing it threw
`No serializer found for org.eclipse.parsson.JsonPatchImpl` and failed the **entire**
timeline response whenever `raw=true` was requested. Added a small `JsonSerializer`
that emits the patch as its RFC-6902 JSON array. Only affects JSON serialization; all
in-process consumers use the getter/builder and are unchanged.

## Testing

- Frontend: added one focused case to `changeEventToString.test.ts` (custom-type URN
  resolves to its display name via the name map) — `47 passed`.
- Backend: added `ChangeTransactionSerializationTest` (rawDiff serializes as a JSON
  array instead of throwing) — passing.
- `spotlessApply` + eslint/prettier clean on changed files.

## Checklist

- [x] Tests added
- [ ] Docs — n/a
- [ ] Breaking changes — none

### DIFF
--- a/datahub-web-react/src/app/entityV2/shared/tabs/Dataset/Schema/history/__tests__/changeEventToString.test.ts
+++ b/datahub-web-react/src/app/entityV2/shared/tabs/Dataset/Schema/history/__tests__/changeEventToString.test.ts
@@ -422,6 +422,26 @@ describe('getChangeEventString', () => {
             expect(result).toBe('Added owner "jane_doe" (Business Owner).');
         });
 
+        it('should resolve a custom ownership type URN to its display name via the name map', () => {
+            // Custom types have opaque UUID slugs; without name resolution the label would show the
+            // UUID ("Custom Uuid 123") instead of the type's real name.
+            const changeEvent: ChangeEvent = {
+                urn: 'urn:test',
+                category: ChangeCategoryType.Ownership,
+                operation: ChangeOperationType.Add,
+                parameters: [
+                    { key: 'ownerUrn', value: 'urn:li:corpuser:jane_doe' },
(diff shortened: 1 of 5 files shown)


chriscollins3456 (approved):
lgtm!

PR #18429, opened 2026-07-16
feat(ingestion/collibra): thin API client skeleton for Collibra governance migrator [WIP]

alokr-dhub:
> **Status: Draft / WIP — do not merge.** Opening early to share direction and run CI.

## Summary

First slice of a native **Collibra** ingestion source (`type: collibra`) — a governance migrator that lifts a Collibra instance's governance layer (glossary terms/nodes, tags, domains, custom attributes, responsibilities/ownership, users/groups, status, links) into DataHub and re-attaches it to existing datasets/columns, **without** importing physical schema, tables, or lineage. It is intended to be re-runnable, stateful, and provenance-stamped so native DataHub metadata is never clobbered.

This PR adds only the **thin API client skeleton** (`client.py`). Mapping, resolver, source, and registration land in follow-ups.

## What's in this PR

`metadata-ingestion/src/datahub/ingestion/source/collibra/client.py`:

- Session factory with OAuth2 bearer auth (Basic/session fallback), TLS / private-CA support, and `urllib3.Retry` for rate-aware 429/5xx backoff.
- Cursor paginator (the core read primitive), `countLimit=0`, 1000/page.
- REST reads for operating-model type defs, communities/domains, and people.
- GraphQL Knowledge Graph entrypoint (primary path for the governance graph).
- Output Module submit → poll → download fallback for bulk pulls on large tenants.
- `ThreadPoolExecutor` partition runner for parallel extraction (parallelism is *across* partitions — cursor paging is sequential within one).

**No new runtime dependencies** — `requests` + `urllib3` + stdlib `concurrent.futures` only.

## Not done yet (follow-ups)

- Split the inline no-network self-check into `tests/` with golden files.
- Swap the dataclass stubs for `pydantic` models (`models.py`) and `Cfg` for `CollibraSourceConfig` (`config.py`).
- `mapper.py`, `urn_resolver.py`, `source.py` (StatefulIngestionSourceBase), provenance/attribution, per-aspect PATCH emission + retraction.
- Registration: setup.py entry point, `updateLockFile`, frontend source cards, logo, platform bootstrap, docs.

## Notes for reviewers

- Cursor field name and the Output Module job-status / file-download response shapes are **placeholders** to confirm against a real Collibra environment.
- Emission strategy is intentionally per-aspect (definitional → UPSERT, co-managed multi-valued → PATCH + attribution) to honor the never-clobber requirement; that logic arrives with the mapper/source, not here.

## Testing

The client carries a no-network self-check (cursor pagination, parallel fan-out, Output Module poll loop) that runs under plain `python3`; it will move into the connector's `tests/` before this leaves draft.


alokr-dhub:
Superseded by #18430 — branch renamed `ing-3056-collibra-thin-client` → `feat/collibra-migrator`. This one closed automatically when the old head branch was renamed.

PR #18432, opened 2026-07-16
fix(ingest/airflow): make asset_iolets integration test retry actually fresh

acrylJonny:
## Summary

The `v2_airflow_asset_iolets` Airflow-plugin integration test flakes in CI. A representative failure ([run 29480610906](https://github.com/datahub-project/datahub/actions/runs/29480610906/job/87563241873)) shows:

1. **Attempt 1** — an Airflow 3 scheduler task-state race marks the two `AssetAlias` tasks as `failed`, raising the `ValueError` the retry loop is designed to absorb.
2. **Attempt 2** — `airflow dags trigger` crashes with `DagRunAlreadyExists: A DAG Run already exists for DAG airflow_asset_iolets with run id manual_run_test`, surfacing only as the opaque `subprocess.CalledProcessError: ... returned non-zero exit status 1`.

**Root cause:** every retry reused the same `airflow_home` (and thus the same SQLite metadata DB), so the fixed run_id `manual_run_test` collided with the previous attempt's `DagRun`. Because that came back as a `CalledProcessError` — which the retry loop did **not** catch — the test hard-failed with no diagnostic output. The "retry with a fresh Airflow instance" was never actually fresh.

## Changes

- Give each retry attempt its own `airflow_home` (`tmp_path/attempt_<n>`) so retries use a genuinely fresh metadata DB. The run_id stays fixed because it is baked into the golden file's `DataProcessInstance` URNs.
- Capture the trigger command's stdout/stderr and retry on `CalledProcessError`, so trigger failures are both diagnosable and treated as transient (unchanged for `success=False` DAGs, which still run once).

Also cleans up two deprecation warnings the suite emits:
- `KafkaSinkConfig.parse_obj(...)` → `model_validate(...)` (Pydantic v2) in `hooks/datahub.py` and `airflow3/datahub_listener.py`.
- In the real-object adapter tests, import the canonical `airflow.sdk.Asset` instead of the deprecated `airflow.datasets.Dataset`. On Airflow 3 the latter is just an alias for the former (same class, no distinct `"Dataset"` in its MRO), so this removes the Airflow 3.2 `DeprecationWarning` at its source without losing any coverage.

## Checklist

- [x] Tests updated (test-infra reliability fix; `ruff` + `mypy` clean on changed files)


metadata-ingestion-modules/airflow-plugin/src/datahub_airflow_plugin/hooks/datahub.py
```diff
@@ -185,7 +185,9 @@ def _get_config(self) -> "KafkaSinkConfig":
             obj["connection"]["bootstrap"] = ":".join(
                 map(str, filter(None, [conn.host, conn.port]))
             )
-        config = datahub.ingestion.sink.datahub_kafka.KafkaSinkConfig.parse_obj(obj)
+        config = datahub.ingestion.sink.datahub_kafka.KafkaSinkConfig.model_validate(
```
gabe-lyons, line 188: why do we need this change

metadata-ingestion-modules/airflow-plugin/src/datahub_airflow_plugin/airflow3/datahub_listener.py
```diff
@@ -554,8 +554,10 @@ def _create_single_emitter_from_connection(self, conn_id: str) -> Optional[Emitt
                     bootstrap = ":".join(map(str, filter(None, [conn.host, conn.port])))
                     obj["connection"]["bootstrap"] = bootstrap
 
-                config = datahub.ingestion.sink.datahub_kafka.KafkaSinkConfig.parse_obj(
-                    obj
+                config = (
```
gabe-lyons, line 557: why do we need this change

gabe-lyons (approved):

metadata-ingestion-modules/airflow-plugin/src/datahub_airflow_plugin/hooks/datahub.py
acrylJonny: `parse_obj` is deprecated in Pydantic v2, and metadata-ingestion is v2-only (`pydantic>=2.4`). This was one of the last two `parse_obj` call sites in the plugin and it throws a DeprecationWarning that shows up in this suite's output. `model_validate` is the v2 equivalent and matches what `_config.py` already does with `model_validate_json`. Happy to split it out if you'd rather keep this PR strictly to the test fix.

metadata-ingestion-modules/airflow-plugin/src/datahub_airflow_plugin/airflow3/datahub_listener.py
acrylJonny: Same reason as the other one: `parse_obj` is deprecated in Pydantic v2 (metadata-ingestion is v2-only) and emits a DeprecationWarning in this suite. `model_validate` is the v2 equivalent, consistent with `_config.py`'s `model_validate_json`. Can split it out if you'd prefer this PR stay scoped to the test fix.

PR #18434, opened 2026-07-16
feat(frontend): add Dataplex ingestion source UI forms for V2

sgomezvillamor:
## Summary
- Add V2 frontend UI form support for the Google Cloud Knowledge Catalog (Dataplex) ingestion source
- Includes Connection, Filter, and Settings form sections with all config fields from the Python connector
- Uses Dataplex-specific credential fields (GCP service account) rather than reusing BigQuery's

### New files
- `datahub-web-react/src/app/ingestV2/source/conf/dataplex/dataplex.ts` — source type constant
- `datahub-web-react/src/app/ingestV2/source/builder/RecipeForm/dataplex.ts` — all form field definitions (connection, filter, advanced)

### Modified files
- `datahub-web-react/src/app/ingestV2/source/builder/constants.ts` — logo mapping
- `datahub-web-react/src/app/ingestV2/source/builder/RecipeForm/constants.ts` — RECIPE_FIELDS registry entry
- `datahub-web-react/src/app/ingestV2/source/builder/sources.json` — source metadata and default recipe

### Source picker
Screenshot 2026-07-16 at 10 43 19

### Filter section
Screenshot 2026-07-16 at 10 59 05

### Connection test
Screenshot 2026-07-16 at 11 55 22

## Test plan
- [x] Verified source appears in the V2 source picker with correct name and description
- [x] Verified filter section renders allow/deny patterns for Projects, Entries, and Entry Groups
- [x] Verified connection form renders all credential fields and connection test succeeds
- [ ] Frontend lint check (`./gradlew :datahub-web-react:yarnLint`)


puneetagarwal-datahub:
LGTM

datahub-web-react/src/app/ingestV2/source/builder/RecipeForm/dataplex.ts
```diff
@@ -0,0 +1,253 @@
...
+    placeholder: '123456789098765432101',
+    type: FieldType.TEXT,
+    fieldPath: 'source.config.credential.client_id',
+    rules: null,
+    required: true,
+};
+
+// ---- Filter Section ----
+
+const projectIdAllowFieldPath = 'source.config.project_id_pattern.allow';
```
puneetagarwal-datahub, line 73: project_ids is marked required, but the connector also accepts project_id_pattern / project_labels for project discovery. So a pattern-only setup is valid but the form blocks it. Either drop required or leave a note that pattern/label-only needs the YAML editor.

puneetagarwal-datahub (changes requested):

puneetagarwal-datahub:
For Dataplex , dont we need V1 ?

sgomezvillamor:
> For Dataplex , dont we need V1 ?

Not sure if there is a criteria on having full parity on both v1 and v2, if that's expected, there should be a sort of test/validation.

AFAIK most of users are on V2 already, which is the default.

datahub-web-react/src/app/ingestV2/source/builder/RecipeForm/dataplex.ts
sgomezvillamor: Good catch — dropped `required` and updated the tooltip to mention that `project_id_pattern` / `project_labels` are available via the YAML editor. Fixed in b4a46b63d8.

That said, the forms are intentionally scoped to the golden path (explicit project IDs + service account credentials). Covering every config permutation would require auto-generating forms from the Python config models, which is a larger UX challenge — the CLI version (and therefore the config schema) is only selected at the end of the wizard, so the form can't know the full schema upfront. For anything beyond what the form exposes, the YAML editor is the escape hatch.

For reference, BigQuery has the same pattern — `credential.project_id` is also marked required in the form even though the connector supports `project_id_pattern` as an alternative.

puneetagarwal-datahub (approved):

PR #18437, opened 2026-07-16
feat(ingest/s3): catalog object-store folders without ingesting files (emit_folders_only)

max-datahub:
## Summary

Adds an opt-in `emit_folders_only` option to `PathSpec` for the object-store file connectors (**S3**, **GCS**, **ADLS/`abs`**). When set, the connector emits each matching storage **folder** as a governable `Container` (to a glob-defined depth) and creates **no dataset entities**.

**Motivation:** catalog storage paths of unstructured files (audio, video, images, …) as governable assets — owners, tags, glossary terms, domains, documentation — while creating zero dataset entities. Folders are already `Container`s in DataHub, so this needs **no** metadata-model, GraphQL, or frontend changes. (As a side effect it also skips the sample-file reads that normal ingestion does for schema inference, but the value prop is the container-only catalog, not I/O avoidance.)

## How it works

Set `emit_folders_only: true` on a `path_spec`. Folder depth is defined by the number of `*` wildcard levels in `include`:

- `s3://bucket/media/*/` → folders one level deep
- `s3://bucket/media/*/*/` → two levels deep

```yaml
source:
  type: s3
  config:
    path_specs:
      - include: "s3://my-bucket/media/*/*/" # catalog folders two levels deep
        emit_folders_only: true
        exclude: ["s3://my-bucket/media/tmp/**"] # optional; applies to the folder walk
    aws_config:
      aws_region: us-east-1
```

- Short-circuits before any file/object listing — lists folder prefixes only (S3 `CommonPrefixes` / ADLS delimiter listing).
- Emits the bucket plus every folder component as `Container`s (subtype `Folder`); no datasets, no schema inference.
- `exclude` and `include_hidden_folders` **apply** to the folder walk. File/dataset-selection fields (`file_types`, `default_extension`, `table_name`, `tables_filter_pattern`) are **rejected** in validation (rather than silently ignored), since folders-only produces no datasets.
- Validation also rejects `{table}` and `**`; depth must be explicit `*` levels. Want multiple depths/paths? Add more `path_specs`.

## Platforms

- **S3** — core implementation.
- **GCS** — delegates to the internal S3 source; the new field is forwarded in `create_equivalent_s3_path_specs()`.
- **ADLS (`abs`)** — mirrored into the separate `abs` source class; shares the container-emission helper.

## Testing

- **Unit:** config validation (rejects `{table}`/`**` and the file/dataset fields; accepts `exclude`/`include_hidden_folders`); `folder_allowed` (hidden + exclude); `create_folder_containers` (bucket + folder chain, dedup, no dataset aspects); S3 via moto — **depth cap** (a below-glob-depth folder stays absent) and **exclude/hidden** end-to-end; ADLS via a `list_folders` mock so the real walker recursion runs.
- **Integration golden:** end-to-end for **S3 and GCS**, container-only output with the expected hierarchy.

## Follow-ups (non-blocking)

- Whether folders-only belongs on `PathSpec` vs a smaller dedicated config surface is worth a design discussion; kept on `PathSpec` for a minimal, familiar-UX diff, with unsupported fields now rejected rather than ignored.
- `List` → `Set` for `ContainerWUCreator.processed_containers` (pre-existing O(n²) membership check).

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/abs/report.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/abs/report.py
@@ -11,9 +11,13 @@ from datahub.utilities.lossy_collections import LossyList
 class DataLakeSourceReport(StaleEntityRemovalSourceReport):
     files_scanned = 0
     filtered: LossyList[str] = dataclass_field(default_factory=LossyList)
+    folders_scanned: int = 0
 
     def report_file_scanned(self) -> None:
         self.files_scanned += 1
+
+    def report_folder_scanned(self) -> None:
+        self.folders_scanned += 1
 
     def report_file_dropped(self, file: str) -> None:
         self.filtered.append(file)
--- a/metadata-ingestion/src/datahub/ingestion/source/abs/source.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/abs/source.py
@@ -404,6 +404,21 @@ class ABSSource(StatefulIngestionSourceBase):
                 container_name, f"{folder}{folder_split[1]}"
             )
 
+    def _process_folders(self, path_spec: PathSpec) -> Iterable[MetadataWorkUnit]:
+        """Emit folder Containers for a folders-only path spec, to the depth defined by
+        the wildcards in `include`. Lists only folders — never blobs."""
+        if self.source_config.azure_config is None:
+            raise ValueError("azure_config not set. Cannot browse Azure Blob Storage")
+        container_name = self.source_config.azure_config.container_name
+        relative_glob = get_container_relative_path(path_spec.glob_include)
+        logger.info(f"Processing folders-only path spec: {path_spec.include}")
+        for folder in self.resolve_templated_folders(container_name, relative_glob):
+            self.report.report_folder_scanned()
+            abs_uri = self.create_abs_path(folder.rstrip("/"))
+            yield from self.container_WU_creator.create_folder_containers(
+                abs_uri.rstrip("/")
+            )
+
     def get_dir_to_process(
         self,
         container_name: str,
@@ -574,6 +589,10 @@ class ABSSource(StatefulIngestionSourceBase):
         with PerfTimer():
             assert self.source_config.path_specs
             for path_spec in self.source_config.path_specs:
+                if path_spec.emit_folders_only:
+                    yield from self._process_folders(path_spec)
+                    continue
+
                 file_browser = (
                     self.abs_browser(
                         path_spec, self.source_config.number_of_files_to_sample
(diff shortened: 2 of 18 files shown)


sgomezvillamor:
Thanks @max-datahub — a few things I'd like to open up for discussion before we move on it:

### High-level

- `PathSpec` is fundamentally an object→dataset mapping, and a container-only mode sits awkwardly on it. Not calling it a blocker, but I'd like to think through whether this really belongs as a flag on `PathSpec` or as its own (smaller) config surface.
- The motivation says this catalogs paths "without reading file content" — but these connectors don't read file content in the first place (unless profiling is enabled). So that's not really the differentiator. The actual difference here is "emit folders as containers, and no dataset entities at all." Worth rewording the motivation so the value prop is clear.

### Concrete smell (the part I'd want addressed)

The `return self` early-exit in the validator is where it bites:

```python
if self.emit_folders_only:
    if "{table}" in self.include: raise ...
    if "**" in self.include: raise ...
    return self
```

That means a folders-only spec silently accepts almost every other field on the object — and `_process_folders` never applies them: `exclude`, `include_hidden_folders`, `tables_filter_pattern`, `file_types` / `default_extension`, etc. So e.g. someone sets `exclude` expecting some folders to be skipped and they get cataloged anyway. I'd either apply those in the folder walk or reject them explicitly, instead of silently ignoring.

### Minor

`create_folder_containers` is a near-verbatim fork of `create_container_hierarchy` (bucket key → folder chain, same prefix logic). Can we share that code instead of duplicating it?

---

max-datahub:
Pushed changes for these:

**Silent field-ignoring** — fixed. `exclude` and `include_hidden_folders` now apply in the folder walk (new `PathSpec.folder_allowed`), and the file/dataset fields (`file_types`, `default_extension`, `table_name`, `tables_filter_pattern`) are rejected in validation rather than silently accepted. Added moto tests for exclude + hidden end-to-end.

**Dedup** — `create_folder_containers` and `create_container_hierarchy` now share `_emit_bucket` / `_emit_folder_chain`; the fork is gone.

**Motivation** — reworded to lead with "folders as containers, no dataset entities." One nit on the premise though: these connectors do read sample-file bytes for schema inference by default (`get_fields` opens each file), not only when profiling — so folders-only does skip real I/O. Agreed it's not the headline; the container-only catalog is.

**PathSpec vs its own config surface** — fair question. I kept it on `PathSpec` for a minimal diff and because users already express per-path cataloging as `path_specs`. The awkward part you flagged was really the silent-ignore, which is now a hard validation error. If you'd still prefer a dedicated `folder_specs` surface, I'm happy to refactor — just say so.

sgomezvillamor (approved):
Good job addressing all the concerns! 🙌

Kept on `PathSpec` for a minimal/familiar-UX diff, with the unsupported fields now rejected rather than silently ignored — that works for me.

Minor suggestion (non-blocking): I haven't checked, but our user-facing docs likely request `s3:GetObject` (and the equivalents for GCS/ADLS) permission. Since folders-only never reads objects, it's worth mentioning the exception in the docs so folks setting up a locked-down role for this feature know they don't need object-read.

---

PR #18446, opened 2026-07-17
fix(graphql): allow delete of data products without domains

javabrett:
## Description

Since #17904, `deleteDataProduct` authorizes exclusively via **Manage Data Products** on a domain associated with the product. For a data product with **no domain**, the check short-circuits to deny before the authorizer is consulted (`isAuthorizedToManageDataProductsOnAnyDomain` over an empty domain set is unconditionally false), so no policy - including the bootstrap "Manage Data Products on All Domains" admin policies added in #17904 - can grant deletion. The entity becomes undeletable by anyone through GraphQL, including platform admins and the root user.

The UI cannot create domain-less data products (`createDataProduct` requires a domain), but the OpenAPI v3 aspect-write path, the Python SDK, and the Terraform provider legitimately can; that is how this was found (a `terraform destroy` failing with "Unauthorized to perform this action" under an admin token).

### Fix

Fall back to the `DELETE` privilege on the data product itself (`AuthorizationUtils.canDeleteEntity`), mirroring the `manage || canDeleteEntity` pattern `DeleteDomainResolver` already uses. The #17904 hardening is fully preserved: unprivileged users remain denied in every case, and the pre-#17904 fail-open behavior (any authenticated user could delete a domain-less product) is not reintroduced.

### Tests

Adds `DeleteDataProductResolverTest` (there was no existing test for this resolver):

- domain-less product + privileged user -> allowed (the regression case; fails without this fix)
- domain-less product + unprivileged user -> denied (fail-closed preserved)
- domained product + authorized / unauthorized -> unchanged behavior
- nonexistent product -> error

The full `resolvers.dataproduct` test package passes.

## Checklist

- [x] The PR conforms to DataHub's [Contributing Guideline](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md) (particularly [Commit Message Format](https://github.com/datahub-project/datahub/blob/master/docs/CONTRIBUTING.md#commit-message-format))
- [ ] For any breaking change/potential downtime/deprecation/big changes correct labels/tags (`breaking-change`/`potential-downtime`/`deprecation`/`big-changes`) are applied to the PR

### DIFF
--- a/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/dataproduct/DeleteDataProductResolver.java
+++ b/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/resolvers/dataproduct/DeleteDataProductResolver.java
@@ -4,6 +4,7 @@ import com.datahub.authentication.Authentication;
 import com.linkedin.common.urn.Urn;
 import com.linkedin.common.urn.UrnUtils;
 import com.linkedin.datahub.graphql.QueryContext;
+import com.linkedin.datahub.graphql.authorization.AuthorizationUtils;
 import com.linkedin.datahub.graphql.concurrency.GraphQLConcurrencyUtils;
 import com.linkedin.datahub.graphql.exception.AuthorizationException;
 import com.linkedin.domain.Domains;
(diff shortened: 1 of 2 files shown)


david-leifker (approved):
lgtm! thanks!

PR #18447, opened 2026-07-17
feat(web): modular asset-summary foundation for SemanticModel + Metric

ani-malgari:
## Summary
- Add 7 new `DataHubPageModuleType` values (`MODEL_DATASETS`, `METRICS_IN_MODEL`, `RELATIONSHIPS`, `DIMENSIONS`, `AI_CONTEXT`, `METRIC_SQL`, `RELATED_METRICS`) in PDL + GraphQL, and register their default URNs in `PageModuleService` / frontend constants.
- Wire frontend module registry stubs (`Module.tsx`, `useAddModuleMenu`, i18n) and ship shared `AiContextModule`; other new types render `SampleLargeModule` placeholders for PR 4/5.
- Include `ExternalReference` on `SemanticModelInfo` / `MetricInfo`, expose `externalUrl` in GraphQL, fetch `lastIngested` on SM/Metric fragments, and broaden `LastIngestedProperty` so non-native Documents (including SM/Metric) can show last ingested.


metadata-models/src/main/pegasus/com/linkedin/module/DataHubPageModuleType.pdl
```diff
@@ -60,6 +60,34 @@ enum DataHubPageModuleType {
    * Module displaying the columns of a dataset
    */
   COLUMNS
+  /**
+   * Module displaying the datasets referenced by a semantic model
+   */
+  MODEL_DATASETS
```
asikowitz: I'll be honest I wouldn't get "datasets referenced by a semantic model" from the name MODEL_DATASETS. Maybe we should at least call it SEMANTIC_MODEL_DATASETS?

metadata-models/src/main/pegasus/com/linkedin/module/DataHubPageModuleType.pdl
```diff
@@ -60,6 +60,34 @@ enum DataHubPageModuleType {
...
+  /**
+   * Module displaying the metrics defined within a semantic model
+   */
+  METRICS_IN_MODEL
```
asikowitz: Same here... and for everything below

asikowitz (approved):
Overall LGTM but as pdl changes are forever, would prefer more specific names that reference semantic models, not just MODEL or RELATIONSHIPS if these are going to be specific to semantic models only

metadata-models/src/main/pegasus/com/linkedin/module/DataHubPageModuleType.pdl
ani-malgari: updated

PR #18455, opened 2026-07-17
fix(models): don't bump schemaVersion for backward-compatible PDL changes

david-leifker:
## Summary

`schemaVersion` on a PDL `@Aspect` exists so migration tooling can detect when data stored at an older version needs transformation. Previously the auto-bumper incremented on **any** aspect change — including purely additive ones. The scheduled [PDL Aspect Change Report](https://github.com/acryldata/datahub-fork/actions/runs/29478290289) already flagged CorpUserInfo (#18278) as `bump_spurious` for exactly that reason.

This PR makes the **bumper and the report share one rule**: only **breaking** changes require a bump. Additive-only edits do not.

### Changes

- **`report_aspect_changes.py`**: Additive optional fields / new enum values no longer set `has_structural`. Bumping for them is `bump_spurious`; leaving them unbumped is `bump_not_needed` (not `bump_needed`).
- **`bump_schema_versions.py`**: Skip bumps for the same Additive set; field comparison matches the report (`type` + `optional` only — annotations/defaults ignored).
- **Drop the CorpUserInfo bump**: remove `"schemaVersion": 2` from `CorpUserInfo.pdl` and restli snapshots, keep `isSupportUser`.
- Docs + tests updated.

### Shared rule (bumper ↔ report)

| Change | Report bucket | Bump? |
| --- | --- | --- |
| Removed field / type change / optional→required / added required / enum removal | breaking | yes |
| Added optional field / added enum value | additive | **no** |
| Bump with only additive changes | noisy / `bump_spurious` | — |

## Test plan

- [x] `pytest .github/scripts/test/test_bump_schema_versions.py test/test_report_aspect_changes.py` — 241 passed
- [x] Bumper `--check` ignores CorpUserInfo as backward-compatible (exit 0)
- [x] Report `analyze_file` on CorpUserInfo-style optional add → `bump_not_needed`, `has_structural=False`
- [x] Pre-commit `bump-schema-versions` passes without re-adding the version
- [ ] CI `check-schema-versions` + restli snapshot checks

### DIFF
--- a/.github/scripts/bump_schema_versions.md
+++ b/.github/scripts/bump_schema_versions.md
@@ -1,6 +1,6 @@
 # bump_schema_versions.py
 
-A pre-commit CLI tool that automatically bumps `schemaVersion` annotations on PDL aspect files whenever their content — or the content of any record/enum/typeref they depend on (via `includes` **or** as a field type) — changes relative to the base branch.
+A pre-commit CLI tool that automatically bumps `schemaVersion` annotations on PDL aspect files whenever their content — or the content of any record/enum/typeref they depend on (via `includes` **or** as a field type) — changes **incompatibly** relative to the base branch. Purely additive, backward-compatible changes (see [Backward-compatible changes](#backward-compatible-changes-are-not-bumped)) are left untouched.
 
 ## Why this exists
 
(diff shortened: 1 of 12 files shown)


RyanHolstien (approved):
I approve of the approach. From my perspective there are two ways forward:
1. This approach, avoids unnecessary migration by avoiding version changes
2. Better detection of migration, i.e. if there is no migrator, don't go through aspect migration even if version changed.

2 makes more sense in my mind from a semantic perspective on versioning, 1 is simpler and already implemented.

gdatahub:
@david-leifker I may have misunderstood this, but I didn't fully understand what "backward compatibility pdl change" means here.  Could you please explain why we need to handle it?

gdatahub:
This make sense  --now Optional/additive PDL changes now skip the schemaVersion bump , only breaking changes (removed field, type change, optional→required, added required field, enum removal) require one.

-- It will be better,  if we factor the additive/breaking field-diff logic into one shared function/module used by both bump_schema_versions.py and report_aspect_changes.py, instead of maintaining two separate implementations that have to be kept in sync manually.

chakru-r (approved):
If an optional field needs a mutator to add a computed default value, we do need to somehow bump the version -- or make it non-optional. This is not common, so in most cases -- is more efficient.

PR #18463, opened 2026-07-18
feat(ui/lineage): Data Product Lineage

asikowitz:
## Data Product Lineage

Adds a Data Product Lineage tab. Represents the data product as a bounding box and loads all its children (up to a limit for perf reasons -- allows users to further paginate). Physical assets are expandable; if another asset in a different data product is found, another bounding box is made and all assets in that data product are grouped inside it. If an asset is in multiple data products, it appears multiple times, once in each data product bounding box.

To position nodes, we follow the following steps:
1. Organize nodes within bounding boxes
2. In one pass, place child bounding boxes and child free nodes in topological + top-down order. Bounding boxes are given highest priority. We try to place nodes by their parent and push down nodes; if there are multiple children, they start horizontally aligned with their parent and go down -- this is contrast to normal lineage where we try to vertically center all children around the parent. When placing a node, if it intersects, we either push it down or push down the other nodes.

Edges are only shown if at least one side of the edge is expanded. Right now I'm not loading all nodes once a data product is discovered -- only adjacent nodes are shown.

Includes a "output port only" filter that limits _only the root node_ to output ports. We can consider changing this in the future if people want to purely see data product -> data product lineage. Think that'll make more sense once we also support input ports.

### Other changes

**1. Removed the unused `getEntityLineageV2` query** from `lineage.graphql` (dead
after the fragment refactor).

### Feature flags

One new flag added to `FeatureFlags.java` / `application.yaml` / `app.graphql`:

| Flag | Env var | Default | Purpose |
| --- | --- | --- | --- |
| `dataProductLineageEnabled` | `DATA_PRODUCT_LINEAGE_ENABLED` | `true` | Enables the Lineage tab on the Data Product entity page (main tab + sidebar tab). |

Pagination:
Screenshot 2026-07-18 at 12 02 07 AM

Tooltip:
Screenshot 2026-07-18 at 12 02 22 AM

Layout:
Testing_2026-07-12_150333
Revenue_Dashboards_2026-07-12_150226
Sales_Dashboards_2026-07-12_150041


metadata-service/configuration/src/main/resources/application.yaml
```diff
@@ -1487,6 +1487,7 @@ featureFlags:
   editableDatasetNameEnabled: ${EDITABLE_DATASET_NAME_ENABLED:false} # Enables the ability to edit the dataset name in the UI
   showManageStructuredProperties: ${SHOW_MANAGE_STRUCTURED_PROPERTIES:true} # If turned on, show the manage structured properties button on the govern dropdown
   hideLineageInSearchCards: ${HIDE_LINEAGE_IN_SEARCH_CARDS:false} # If turned on, hides lineage information in search result cards
+  dataProductLineageEnabled: ${DATA_PRODUCT_LINEAGE_ENABLED:true} # If turned on, enables the Lineage tab on the Data Product entity page
```
chriscollins3456, line 1490: so default enabled here?

datahub-web-react/src/app/lineageV3/LineageAnnotationNode/LineageAnnotationNode.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import styled from 'styled-components';
 
-export const LINEAGE_ANNOTATION_NODE = 'lineage-annotation-node';
+export const LINEAGE_ANNOTATION_NODE_NAME = 'lineage-annotation-node';
```
asikowitz, line 4: Just a rename for consistency

datahub-web-react/src/app/lineageV3/LineageBoundingBoxNode/BoundingBoxMemberCount.tsx
```diff
@@ -0,0 +1,102 @@
...
+    /** Number of this product's members currently shown in the box. */
+    memberCount: number;
+}
+
+/**
+ * Member counter shown on the right of a data product bounding box header. The home product shows
+ * "x / y entities shown" with a "Show more" control that pages in more members; other products show
+ * how many of their assets are connected to the home product, and are not paginated.
+ */
+export default function BoundingBoxMemberCount({ urn, memberCount }: Props) {
```
asikowitz, line 60: Pagination built for the rare very large data product

datahub-web-react/src/app/lineageV3/common.ts
```diff
@@ -18,6 +18,9 @@ import { Entity, EntityType, LineageDirection, SchemaFieldRef } from '@types';
 export const TRANSITION_DURATION_MS = 250;
 export const LINEAGE_FILTER_PAGINATION = 4;
 
+// Page size for fetching/displaying a data product's members, and the initial home member limit.
+export const DATA_PRODUCT_MEMBER_PAGE_SIZE = 2;
```
chriscollins3456: so just 2 for right now? does it make sense to bump this up at all?

datahub-web-react/src/app/entityV2/dataProduct/DataProductEntity.tsx
```diff
@@ -170,6 +180,17 @@ export class DataProductEntity implements Entity<DataProduct> {
...
                 icon: AppstoreOutlined,
             },
+            {
+                name: i18next.t('entity.types:tab.lineage'),
+                // Data products show the explorer only, without the impact analysis tab — as data flows do
+                component: DAGTab,
+                icon: PartitionOutlined,
+                supportsFullsize: true,
+                display: {
+                    visible: (_, _1) => this.appconfig().config.featureFlags.dataProductLineageEnabled,
```
chriscollins3456, line 190: okay so the only place that we guard this feature is just showing/hiding this tab on the data product entity - that makes enough sense if this is the only way to view data product lineage (like you can't see it from other lineage tabs)

chriscollins3456 (approved):
didn't go too deep into the weeds of the rendering code but it seems to be structured well and makes sense. would require manual tests obviously as i know you've been doing while building.

metadata-service/configuration/src/main/resources/application.yaml
asikowitz: Yup. This just enables the tab and doesn't affect the rest of the app. I don't have too many concerns and feel good about this feature

datahub-web-react/src/app/lineageV3/common.ts
asikowitz: This was 50, forgot to change during testing, thanks!

PR #18467, opened 2026-07-18
refactor(ingest/odcs): model schema-assertion compatibility as an enum

acrylJonny:
## Summary

Follow-up to #17331. Replaces the free-form schema-assertion compatibility string with a `SchemaAssertionCompatibility` enum in the ODCS config/mapper, removing duplicated compatibility-mode literals and the latent `KeyError` path when an unrecognized mode was supplied.

## Checklist

- [x] Tests updated to use the enum
- [ ] Docs (n/a)
- [ ] Breaking change (n/a)

## Connector tests


metadata-ingestion/src/datahub/ingestion/source/odcs/odcs_config.py
```diff
@@ -83,7 +85,17 @@ def env_must_be_one_of(cls, v: str) -> str:
         return v.upper()
 
 
-_SCHEMA_ASSERTION_COMPATIBILITY_MODES = ("EXACT_MATCH", "SUPERSET", "SUBSET")
+class SchemaAssertionCompatibility(ConfigEnum):
```
AdrianMachado, line 88: Wow, I am surprised there was no existing enum

AdrianMachado (approved):
Great fix!

PR #18468, opened 2026-07-18
refactor(ingest/odcs): type quality arguments and narrow Any usage

acrylJonny:
## Summary

Follow-up to #17331. Tightens typing in the ODCS models/mapper:

- Dynamic JSON fields use `pydantic.JsonValue` instead of `Any`/`object`.
- Description fields accept a typed `ODCSDescription` model (or plain string) rather than an untyped dict.
- Quality-argument and rule fields are typed precisely; `_description_to_str` / `_is_plain_number` narrow their inputs.

No behavioural change — golden files are unaffected.

## Checklist

- [x] Tests pass (unit + golden integration)
- [ ] Docs (n/a)
- [ ] Breaking change (n/a)

## Connector tests

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/odcs/odcs_mapper.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/odcs/odcs_mapper.py
@@ -199,8 +199,7 @@ def _description_to_str(
         ("usage", description.usage),
         ("limitations", description.limitations),
     ]
-    # Some producers add non-spec prose keys (e.g. summary, description); render
-    # any string-valued extras so operator-authored text is never dropped.
+    # Render non-spec string extras (e.g. summary) so producer prose isn't dropped.
     for key, value in (description.model_extra or {}).items():
         if isinstance(value, str):
             fields.append((key, value))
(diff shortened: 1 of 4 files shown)


metadata-ingestion/src/datahub/ingestion/source/odcs/odcs_mapper.py
```diff
@@ -185,21 +187,23 @@ class SchemaBuildResult:
...
+def _description_to_str(
+    description: Optional[Union[str, ODCSDescription]],
+) -> Optional[str]:
     if description is None:
         return None
     if isinstance(description, str):
         return description.strip() or None
-    if isinstance(description, dict):
-        parts: List[str] = []
-        for key in ("purpose", "usage", "limitations", "summary", "description"):
```
AdrianMachado, line 196: so summary and description  no longer count in this case?

AdrianMachado (approved):

metadata-ingestion/src/datahub/ingestion/source/odcs/odcs_mapper.py
acrylJonny: Good catch. `summary`/`description` were never ODCS description sub-keys — the spec object is just `purpose`/`usage`/`limitations`, which is what the typed `ODCSDescription` model now covers. The original 5-key loop was speculatively pulling in two non-spec keys, so this narrowed it.

That said, dropping producer-authored prose silently isn't great, and `extra="allow"` already keeps those keys on the model. I've pushed a follow-up (885b0ae) that renders the three spec fields and then appends any string-valued extras, so `summary`/`description` (or anything else a producer tacks on) still make it into the dataset description — non-string extras like `authoritativeDefinitions` are skipped. Added a unit test covering it.

PR #18475, opened 2026-07-18
refactor(ingest): extract shared object-store file reader from dbt

acrylJonny:
## Summary

Extracts dbt's file-reading and object-store glob-expansion logic into a reusable module, `datahub.ingestion.source.common.object_store_files`, so other connectors (starting with ODCS) can source files from local paths, HTTP(S) URLs, and S3/GCS URIs without duplicating the logic. This is a pure refactor — dbt behavior is unchanged.

New shared primitives:
- `read_file_as_bytes(uri, aws_connection, gcs_connection, http_timeout_seconds)` — reads a single file from a local path, `http(s)://` URL, `s3://` URI, or `gs://` URI. Object-store URIs require the matching connection config for credentials.
- `expand_object_store_glob(uri, connection, scheme)` — prefix-scoped glob listing over an S3-compatible object store.
- `has_glob_characters` / `is_http_uri` / `expand_local_glob` helpers.

`DBTCoreSource.load_file_as_json` and the run-results glob expansion now delegate to these helpers.

## Test plan

- [x] Primitive-level tests relocated to `tests/unit/test_object_store_files.py` (glob expansion for S3/GCS, prefix scoping, multi-page, no-cross-slash matching, client errors; `read_file_as_bytes` for local/HTTP/S3 and missing-credential errors).
- [x] dbt keeps its wrapper tests (`load_file_as_json`, `_expand_run_results_paths`, config validators) — all still pass.
- [x] `./gradlew :metadata-ingestion:lint` (ruff + mypy) clean.

## Connector tests

## `max_bytes` size cap — downstream consumer

`read_file_as_bytes` accepts an optional `max_bytes` cap (rejects from `Content-Length`, or mid-stream before the body is fully buffered). **dbt does not pass it** — this is a deliberate shared-ahead-of-consumer split; the live consumer is the stacked **#18477** (ODCS remote sourcing), which threads `max_bytes=max_input_file_bytes` on every remote fetch. dbt's manifest read stays uncapped (pre-existing behaviour); adding a dbt cap would need its own config + a safe default and is out of scope here.

The HTTP branch now streams inside a `with requests.get(...) as resp:` block, so the connection is returned to the pool even when the cap trips before the body is drained (covered by `test_read_file_as_bytes_http_releases_connection_on_cap_exceeded`). BOM passthrough is locked across UTF-8/16/32 and for the S3 path.


AdrianMachado (approved):
Smart idea and good find for reuse

PR #18477, opened 2026-07-18
feat(ingest/odcs): source contracts from S3, GCS, HTTP, and Git

acrylJonny:
## Summary

Extends the ODCS ingestion source so contract YAML no longer has to live on the local filesystem. `path` now accepts:

- `s3://` / `gs://` object-store URIs — a single file or a glob (e.g. `s3://bucket/contracts/*.odcs.yaml`)
- `http(s)://` URLs to a single file
- any mix of the above (and local paths) in a list

It also adds `git_info` to shallow-clone a repository and scan it (SSH deploy key, exactly like the LookML source); non-URI `path` entries are then resolved relative to the checkout (e.g. `path: contracts/`).

This mirrors how dbt handles remote sourcing. Remote reads and object-store glob expansion reuse the shared `object_store_files` utility extracted from dbt in the base PR.

## Details

- `aws_connection` / `gcs_connection` config blocks, validated up front (an `s3://`/`gs://` URI without the matching connection fails config validation with a clear message).
- New `odcs` plugin extra pulls in boto3 (via `aws_common`) and `GitPython`; install with `pip install 'acryl-datahub[odcs]'`.
- File size guard, unknown-field warnings, and per-file error isolation all apply to remote files too.

## Stacking

Stacked on #18475 (`refactor/object-store-file-reader`) — base this PR on that branch and rebase onto `master` once it merges.

## Testing

- New unit tests: S3/GCS/HTTP URI resolution, S3 glob expansion + extension filtering, HTTP-glob rejection, remote YAML loading (size guard, non-mapping, read failure), end-to-end URI processing, and git clone-then-walk.
- `pytest tests/unit/odcs/ tests/unit/test_object_store_files.py` — 145 passed.
- `./gradlew :metadata-ingestion:lintFix` + `mypy` clean.

## Checklist

- [x] Tests added/updated
- [x] Docs added/updated (recipe + connector prerequisites)
- [ ] Breaking changes documented — n/a (additive)

## Connector tests

Part of the ODCS follow-up stack. A companion integration suite for the ODCS source is added in the private `acryldata/connector-tests` repo — acryldata/connector-tests#852 — a hermetic golden-file test that ingests a Bitol ODCS contract and asserts the full mapping (logical datasets, physical bindings, schema + data-quality assertions, and native data contracts). **This PR's behaviour is directly asserted by that suite's golden**, so the suite depends on this change landing in the validated CLI release.


AdrianMachado (changes requested):
Can we have some e2e tests for these? Like actual S3 buckets, git file, etc to prove this works and additionally we don't regress the support in the future? Unit tests are nice but e2e is best for these sort of integration tasks

acrylJonny:
Added in f2b9104 — `metadata-ingestion/tests/integration/odcs/test_odcs_remote_sourcing.py`. These drive the real transport code paths against real infra rather than mocks:

- **S3** (moto-backed bucket, real boto3 upload + read): a single `s3://` file, plus a glob that expands `contracts/*.odcs.yaml`, filters out a `.txt` decoy under the same prefix, and ingests both contracts.
- **GCS**: the same object-store read path via the real `gcs_connection.s3_compatible_connection` boto3 client, pointed at the moto mock.
- **HTTP**: a live threaded `http.server` on a random port — real `requests.get` streaming download, not a stubbed response.
- **Git**: an on-disk repo committed and cloned over `file://` via `git_info`, then the checkout is walked for `contracts/`.

Each asserts the emitted `odcs.id` on the datasets plus the report counters (`remote_files_scanned`, `git_checkout`), so a regression in any one transport fails loudly. No new dependencies — moto, gitpython and stdlib `http.server` were already available.

AdrianMachado (approved):

PR #18479, opened 2026-07-18
fix(ingestion): treat Windows drive letters as local paths in get_path_schema

kasbsquall:
## Problem

On Windows, any ingestion that resolves a local path through `FileSource` fails with:

```
KeyError: Did not find a registered class for c
```

This makes `datahub datapack load` unusable on Windows, since the loader downloads the
datapack to a local cache and hands the path to `FileSource`:

```
> datahub datapack load showcase-ecommerce
...
  File "datahub/ingestion/source/file.py", line 218, in get_filenames
    fs_class = fs_registry.get(schema)
  File "datahub/ingestion/api/registry.py", line 171, in get
    raise KeyError(f"Did not find a registered class for {key}")
KeyError: 'Did not find a registered class for c'
```

## Root cause

`get_path_schema()` derives the filesystem from `urlparse(path).scheme`. On Windows the
drive letter is parsed as the scheme, because CPython's `urlsplit` treats any run of
scheme-legal characters before the first `:` as a scheme:

```python
>>> from urllib import parse
>>> parse.urlparse(r"C:\data\file.json").scheme
'c'
```

The registry only knows `file`, `http` and `s3`, so the lookup raises. Using forward
slashes does not help: `urlparse("C:/data/file.json").scheme` is also `'c'`.

## Fix

Treat a single-character scheme as a local path. No real filesystem scheme is one
character long, so this cannot shadow a valid one.

```python
if scheme == "" or len(scheme) == 1:
    scheme = "file"
```

`LocalFileSystem` already handles Windows paths correctly, since it uses `pathlib.Path`,
so no other change is needed.

## Tests

Added `test_get_path_schema_windows_drive_letter` to
`metadata-ingestion/tests/unit/test_file_systems.py`, covering backslash paths, forward
slash paths, a non-C drive and a lowercase drive letter. The existing
`test_get_path_schema` continues to cover `s3`, `http`, `https`, absolute and relative
POSIX paths, all unaffected by this change.

## Verification

```
pytest metadata-ingestion/tests/unit/test_file_systems.py -k get_path_schema
```

Before the fix the new test fails with `assert 'c' == 'file'`; after it, both tests pass.


itxcrusher:
Independent Windows validation for the reproduction in #18493:\n\n- DataHub CLI 1.5.0.6 currently resolves both C:\\\\...\\\\file.json and C:/.../file.json to scheme c.\n- I checked out this PR's head and ran the changed function on Windows 11 / Python 3.11.\n- Uppercase and lowercase drive letters and both slash styles resolve to ile.\n- Existing s3 and https schemes remain unchanged.\n\nThis directly addresses the filesystem-registry failure reported by the datapack command.

PR #18480, opened 2026-07-18
feat(ingest/odcs): emit native DataHub data contracts for ODCS

acrylJonny:
## Summary

The ODCS connector models each contract as a logical `odcs` dataset with schema and data-quality assertions targeting it. This adds emission of a native DataHub `dataContract` entity so contracts surface on the **Contract tab** in the UI.

**Default (safe):** emit the contract only on the **logical `odcs` dataset** — the self-consistent home, since the assertions target it (`contract.entity == assertion.entity`), the URN is collision-free, and it renders under `LOGICAL_MODELS_ENABLED`. Gated by `emit_data_contract` (default `true`).

**Opt-in:** `emit_physical_data_contract` (default `false`) *also* emits the contract onto the bound physical dataset so it surfaces on the table consumers browse. This is off by default and always emitted **non-primary** because the contract URN is `guid({entity})` — identical to the hand-authored SDK convention — so a primary write here would overwrite and (on stale removal) soft-delete a user's hand-authored contract on that table. It additionally requires `emit_logical_parent` (the master switch for writing any aspect onto physical datasets).

Both contracts reference the *same* schema/data-quality Assertion URNs — no assertions are duplicated. State mirrors the ODCS `status` (`active` -> `ACTIVE`, otherwise `PENDING`). Entries that produce no assertions emit no empty contract and are tracked via a new `data_contracts_skipped_no_assertions` report counter.

## Checklist

- [x] Unit tests: logical-only by default when bound, both-when-opted-in (asserts physical is non-primary), opt-in requires `emit_logical_parent`, logical-only when unbound, skip-when-no-assertions (+ counter), state mapping, flag disable
- [x] Integration golden files updated (logical-only default)
- [x] `./gradlew :metadata-ingestion:lint` passes (ruff + mypy)
- [x] Verified locally: logical `odcs` datasets resolve `dataset.contract` via GraphQL after ingest

## Connector tests

Part of the ODCS follow-up stack. A companion integration suite for the ODCS source is added in the private `acryldata/connector-tests` repo — acryldata/connector-tests#852 — a hermetic golden-file test that ingests a Bitol ODCS contract and asserts the full mapping (logical datasets, physical bindings, schema + data-quality assertions, and native data contracts). It runs with default config (logical-only contracts), so **this PR's default behaviour is directly asserted by that suite's golden** and the suite depends on this change landing in the validated CLI release.


AdrianMachado:
I need to think a bit more about this once since we are planning to make some changes to data contracts soon

metadata-ingestion/src/datahub/ingestion/source/odcs/odcs_config.py
```diff
@@ -228,6 +235,28 @@ def owner_normalization_knobs_are_exclusive(self) -> "ODCSSourceConfig":
...
+        default=True,
+        description="Whether to emit a native DataHub `dataContract` entity for each schema entry, "
+        "on the logical `odcs` dataset. This is the self-consistent home: the assertions target "
+        "the logical dataset, so `contract.entity == assertion.entity` and results resolve "
+        "directly; it renders under the LOGICAL_MODELS_ENABLED flag. The contract references the "
+        "schema and data-quality Assertion URNs (no assertions are duplicated). State mirrors the "
+        "ODCS `status` (`active` -> ACTIVE, otherwise PENDING). Requires at least one of "
+        "`emit_assertions` / `emit_schema_assertion` to have produced an assertion for the entry.",
+    )
+    emit_physical_data_contract: bool = Field(
```
AdrianMachado, line 300: I don't think we should be handling propagation to physical assets via this ingestion source. Its a trickier subject that will have its own job that @asikowitz is working on

AdrianMachado (dismissed):

metadata-ingestion/src/datahub/ingestion/source/odcs/odcs_config.py
acrylJonny: Resolving as we had a call with this being the direction of travel

AdrianMachado:
Updating my earlier note: we agreed this connector can write the contract onto the bound physical dataset. emit_physical_data_contract can stay (default off is right — same guid({entity}) as a hand-authored contract, so enabling it overwrites that table's contract; is_primary_source=False only skips stale-delete).

The DATA_SCHEMA assertion should keep pinning the ODCS / logical-dataset schema (what the contract promises), not the bound table's catalog types. That is the useful expected schema. Don't emit a warehouse-typed sibling.

The physical path still does not give Observe something it can run for the other checks. It reuses the same assertion URNs, and those still have entity = the logical odcs dataset. The PR already calls this out (“limited detail … until assertions can target it directly”):

Freshness — DATASET_CHANGE on an odcs URN that never gets warehouse operations.
Field / SQL — Observe has no warehouse behind urn:li:dataPlatform:odcs,….
For schema, if we want pass/fail on the table, evaluation has to compare the physical schema to that logical-model assertion — not rewrite the assertion to catalog types. As written, schemaAssertion.entity is the logical URN, so a native evaluator would compare the assertion to the same YAML it was built from.

The Contract tab only rolls up assertions with a run event, so both tabs stay “no assertions run” until freshness / quality actually execute against the table (and schema is evaluated against the physical instance using the logical expected schema).

If this source is the propagator, when the flag is on it should retarget (or emit siblings of) freshness and quality at the physical URN. Schema can stay one assertion whose schema is the logical model.

Also:

emit_data_contract default-on is awkward while LOGICAL_MODELS_ENABLED is off — the table people browse stays empty unless they opt into the physical write.
The emit_data_contract Field description is stale (it omits freshness, which alone will create a contract).
Freshness SLA mapping and the fail-closed physical-URN verification are reasonable, but they are separate behavior changes — easier to review as follow-ups.

acrylJonny:
Thanks for the detailed pass — I've narrowed this PR to just native `dataContract` emission and split the rest out. Point by point:

> emit_physical_data_contract can stay (default off is right — same guid({entity}) as a hand-authored contract, so enabling it overwrites that table's contract; is_primary_source=False only skips stale-delete)

Agreed, kept as-is. Default off, `is_primary_source=False` so stale removal never touches a hand-authored contract, and the description now spells out that enabling it makes ODCS the owner of that dataset's contract.

> The DATA_SCHEMA assertion should keep pinning the ODCS / logical-dataset schema (what the contract promises), not the bound table's catalog types. Don't emit a warehouse-typed sibling.

That's the behaviour — the schema assertion stays pinned to the logical model, and there's no warehouse-typed sibling. `schemaAssertion.entity` remains the logical URN.

> If this source is the propagator, when the flag is on it should retarget (or emit siblings of) freshness and quality at the physical URN. Schema can stay one assertion whose schema is the logical model.

Fair — this is the real enhancement for the physical path, but it's a bigger behaviour change tied to the freshness work, so I've left it out of this PR. The limitation ("limited detail … until assertions can target it directly") is still called out in the config/docs. Follow-up.

> emit_data_contract default-on is awkward while LOGICAL_MODELS_ENABLED is off — the table people browse stays empty unless they opt into the physical write.

Kept the default on but documented the trade-off explicitly in `odcs_pre.md`: logical-only by default (renders under `LOGICAL_MODELS_ENABLED`), with `emit_physical_data_contract` as the opt-in for surfacing on the browsable table.

> The emit_data_contract Field description is stale (it omits freshness, which alone will create a contract).

Fixed — and moot now, since freshness is out of this PR. The description states it requires at least one schema/data-quality assertion to have been produced.

> Freshness SLA mapping and the fail-closed physical-URN verification are reasonable, but they are separate behavior changes — easier to review as follow-ups.

Done — both pulled out of this PR. Verification is back to fail-open, and the freshness/`slaProperties` mapping is removed. I'll bring them back as separate follow-ups so they can be reviewed on their own.

asikowitz (approved):

acrylJonny:
@AdrianMachado — friendly nudge on your earlier `Changes requested` review. Per your Sep 5 update, this PR was scoped down to native `dataContract` emission only and every point is addressed on the current head:

- `emit_physical_data_contract` stays default-off, emitted `is_primary_source=False` (never stale-removes a hand-authored contract), and now requires `emit_logical_parent` via a config-time validation. Its description states that enabling it makes ODCS the owner of that table's contract.
- The schema assertion stays pinned to the logical URN — no warehouse-typed sibling; the physical-render limitation is documented.
- The `emit_data_contract` description is no longer stale (freshness removed; it now states it requires at least one produced assertion).
- Freshness SLA mapping and the fail-closed physical-URN verification are pulled out into follow-ups; `_urn_exists_in_graph` is back to fail-open.

All inline threads are resolved, cubic reports everything addressed, and @asikowitz has approved. Your review is the last standing `Changes requested` blocking merge — could you re-review (approve or dismiss) when you have a moment? Thanks!

PR #18481, opened 2026-07-18
fix(ui): render custom assertion name on the Data Contract tab

acrylJonny:
## Summary

On the dataset **Contract tab**, the data-quality table rendered a CUSTOM-type assertion using only `assertion.info.description`. Custom assertions without a description (common for externally-ingested checks) therefore showed as **blank text**.

This falls back to the custom assertion's `type` — which is already fetched by the contract query (`customAssertion { type }` via `assertionDetails`) — so the check always has a visible name.

```tsx
{assertion.info?.description || assertion.info?.customAssertion?.type}
```

## Checklist

- [x] `yarnLintFix` clean on the changed file
- [x] No new GraphQL fields needed (data already fetched)

## Connector tests


AdrianMachado (approved):
This is an OK fallback but it would be nice to actually ingest names on the External assertions coming from ingestion sources.

PR #18486, opened 2026-07-19
ci(lint): combine Python CI lint jobs into a single python-lint job

devashish2203:
## Summary

Consolidate the per-module Python CI lint jobs in `.github/workflows/lint-jobs.yml` into one `python-lint` job, and add the path filters + setup outputs needed to drive it. Cuts CI runner usage (one runner instead of N) while keeping per-module isolation, per-module gating, and a shared dependency cache. The `smoke-test-python` lint job is intentionally left as its own job.

Second layer of the PFP-4982 pre-commit/CI optimization stack (stacked on `dc/pfp-4998-ci-filters`).

### `.github/path-filters/lint-jobs-filters.yml`
Add path filters for the six Python modules that were previously missing their own filter outputs: `datahub-actions`, `datahub-agent-context`, `airflow-plugin`, `gx-plugin`, `dagster-plugin`, `prefect-plugin`.

### `.github/workflows/lint-jobs.yml`
- **Setup job:** add the six new module filter outputs (each OR'"'"'d with `lint-job`), plus a single `python-lint` output = OR of all six (and `lint-job`). The `python-lint` job runs if **any** Python module'"'"'s files changed.
- **Rename** `metadata-ingestion-lint` → `python-lint`, gated by `needs.setup.outputs.python-lint == '"'"'true'"'"'`.
- **Shared uv cache:** one `UV_CACHE_DIR` cache (restore before install, save at end) keyed on `uv-python-lint-<hash of all modules'"'"' setup.py/pyproject.toml/uv.lock>`. Key omits `github.base_ref` so the cache is shared across PRs on different base branches.
- **Per-module lint steps:** one step per module, each runs `./gradlew :<module>:lint` in its own Gradle-managed venv (separate venvs, shared download cache), gated by `if: always() && needs.setup.outputs.<module> == '"'"'true'"'"'` so a failure in one module does **not** skip the others and each only runs when that module'"'"'s files changed.
- **Report step:** a final `if: always()` step collects each step'"'"'s `outcome`, prints an `::error::` listing failed modules, and exits non-zero if any module failed.

## Test plan
- [ ] PR touching only `metadata-ingestion/` → only the `metadata-ingestion lint` step runs; others skip.
- [ ] PR touching two modules → both steps run; if one fails, the other still runs and the job reports both, failing on the failed one.
- [ ] PR touching no Python module → `python-lint` job is skipped entirely.
- [ ] PR touching `.github/workflows/lint-jobs.yml` → `python-lint` runs all steps via the `lint-job` OR.
- [ ] uv cache restored on a repeat run with unchanged lockfile, across different base branches.
- [ ] `smoke-test-python` lint still runs as its own job, unchanged.

## Notes
- Stacked on `dc/pfp-4998-ci-filters` (base branch for this PR).
- Refs PFP-4999.

### DIFF
--- a/.github/path-filters/lint-jobs-filters.yml
+++ b/.github/path-filters/lint-jobs-filters.yml
@@ -34,5 +34,17 @@ datahub-web-react:
 metadata-ingestion:
   - added|modified: "metadata-ingestion/**"
   - "metadata-models/**"
+datahub-actions:
+  - added|modified: "datahub-actions/**"
+datahub-agent-context:
+  - added|modified: "datahub-agent-context/**"
+airflow-plugin:
+  - added|modified: "metadata-ingestion-modules/airflow-plugin/**"
+gx-plugin:
+  - added|modified: "metadata-ingestion-modules/gx-plugin/**"
+dagster-plugin:
+  - added|modified: "metadata-ingestion-modules/dagster-plugin/**"
+prefect-plugin:
+  - added|modified: "metadata-ingestion-modules/prefect-plugin/**"
 playwright-e2e:
   - added|modified: "e2e-test/ui/playwright/**"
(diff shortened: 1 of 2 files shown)


AdrianMachado (approved):

david-leifker (approved):
Approved after review.

devashish2203:

PR #18487, opened 2026-07-19
ci(pre-commit): use each module's pinned venv ruff for staged-file hooks (PFP-5000)

devashish2203:
## Summary

Replace the slow, module-wide Python pre-commit hooks with fast staged-file ruff hooks that only check/format the files actually staged for commit.

Previously each Python module had a single `gradle :<module>:lintFix` pre-commit hook with `pass_filenames: false`, meaning **every** commit lint-fixed the **entire** module (slow, and it mutated files outside the staged set). This PR emits two per-module staged-file ruff hooks instead — `ruff-check` (with `--fix`) and `ruff-format` — that run only on the staged `.py` files, and uses each module's own pinned ruff version + config rather than a system ruff.

This is the third layer of the PFP-4982 pre-commit/CI optimization stack (stacked on `dc/pfp-4999-python-ci-lint`).

- Replace `_generate_lint_fix_hook` (module-wide `gradle :<module>:lintFix`, `pass_filenames: false`) with `_generate_python_hooks`, which emits two staged-file hooks per module:
  - `<module>-ruff-check` → `bash .github/scripts/python_staged_ruff.sh <module> check --fix`
  - `<module>-ruff-format` → `bash .github/scripts/python_staged_ruff.sh <module> format`
  - Both with `pass_filenames: true` and `files: ^<module>/.*\.py$`.
- Gate ruff-hook generation on `build.gradle` presence: modules without a Gradle-managed venv (e.g. `perf-test/authz-perf`) have no module ruff to invoke, so they get no ruff hook (with a printed skip reason) instead of a hook that would fail at runtime.

### `.github/scripts/python_staged_ruff.sh` (new)
Wrapper that execs `<module>/venv/bin/ruff` — the ruff installed by the module's Gradle `installDev` task — so each module runs its own pinned ruff version (e.g. 0.15.18 for `metadata-ingestion`, 0.11.7 for the others) with its own `[tool.ruff]` config, which ruff auto-discovers by walking up from each staged file to the nearest `pyproject.toml`. Fails fast with an `installDev` instruction if the module venv hasn't been built yet. No remote hook repo is introduced.

### `.pre-commit-config.yaml`
Regenerated from the script above: per-module `ruff-check` / `ruff-format` staged hooks replace the old `lint-fix` hooks.

### `smoke-test/pyproject.toml`
Unrelated drive-by: bump the smoke-test pytest log file level `INFO` → `DEBUG` so failed smoke-test runs upload more verbose logs as CI artifacts.

## Test plan

- [ ] Stage a single `.py` file in `metadata-ingestion/` and commit → only the `metadata-ingestion Ruff Check` / `Ruff Format` hooks run on that file (not the whole module); commit is fast.
- [ ] Stage a `.py` file with a ruff violation → `ruff-check --fix` auto-fixes safe issues; if unfixable, the commit is blocked with the ruff error.
- [ ] Confirm each module uses its own pinned ruff version (`ruff --version` from its venv) and its own `[tool.ruff]` config (e.g. a rule disabled in one module's `pyproject.toml` is honored for that module only).
- [ ] On a module whose venv hasn't been built (`<module>/venv/bin/ruff` missing), the hook fails fast with the `./gradlew :<module>:installDev` instruction rather than a cryptic `ruff: command not found`.
- [ ] Confirm `perf-test/authz-perf` (no `build.gradle`) gets no ruff hook (skip message printed at generation time).
- [ ] Confirm Java/Spotless and Prettier hooks are unchanged.

## Notes

- Stacked on `dc/pfp-4999-python-ci-lint` (base branch for this PR).
- The staged-file ruff hooks are a strict subset of CI's full module-wide `./gradlew :<module>:lint` (which still runs in CI); pre-commit now only catches issues in the files being committed, for speed.
- Refs PFP-5000.

### DIFF
--- a/.github/scripts/generate_pre_commit.py
+++ b/.github/scripts/generate_pre_commit.py
@@ -153,7 +153,7 @@ class HookGenerator:
 
         for project in self.projects:
             if project.type == ProjectType.PYTHON:
-                hooks.append(self._generate_lint_fix_hook(project))
+                hooks.extend(self._generate_python_hooks(project))
             elif project.type == ProjectType.JAVA:
                 hooks.append(self._generate_spotless_hook(project))
             elif project.type == ProjectType.PRETTIER:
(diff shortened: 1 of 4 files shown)


david-leifker (approved):
Approved after review.

devashish2203:

PR #18488, opened 2026-07-19
ci(pre-commit): add staged-only mypy hooks at pre-push for modules that run mypy (PFP-5000)

devashish2203:
## Summary

Add staged-only **mypy** pre-push hooks for the Python modules whose Gradle lint task actually runs mypy. mypy is slow, so it runs at the **pre-push** stage (not pre-commit) and only on the files staged for the push — keeping commits fast while still catching type errors before they reach the remote. CI'"'"'s full module-wide `./gradlew :<module>:lint` (which includes mypy) is unchanged.

Fourth layer of the PFP-4982 pre-commit/CI optimization stack (stacked on `dc/pfp-5000-staged-ruff`).

- Extend `_generate_python_hooks` to also emit a `<module>-mypy` hook **only** when the module'"'"'s `build.gradle` lint task runs mypy, detected via the `"mypy` token in the Gradle file (new `_module_has_mypy` helper).
- The mypy hook is assigned `stages: [pre-push]` (mypy is slow); the existing ruff hooks are unchanged.

### `.github/scripts/python_staged_mypy.sh` (new)
Wrapper that runs the module'"'"'s own pinned mypy from its Gradle-managed venv (`<module>/venv/bin/mypy`) from the module directory, so mypy discovers the module'"'"s mypy config (`setup.cfg [mypy]` / `pyproject.toml [tool.mypy]`, including per-module overrides like `[mypy-datahub.*]`). Staged file paths are converted from repo-root-relative to module-relative. Fails fast with an `installDev` instruction if the module venv hasn'"'"'t been built. Uses `--follow-imports=silent` so only errors in the staged files themselves are reported (mypy still resolves imports for typing, but doesn'"'"'t report errors in silently-followed modules) — a fast subset of CI'"'"'s full module-wide mypy.

### `.pre-commit-config.yaml`
Regenerated: per-module `<module>-mypy` pre-push hooks added for the modules that run mypy.

## Test plan
- [ ] Stage a `.py` file in a module that runs mypy (e.g. `metadata-ingestion`) and run `git push` → the `<module> Mypy` hook runs on just that file; push is blocked on type errors in the staged file.
- [ ] Confirm each module uses its own pinned mypy version + its own `[tool.mypy]` / `setup.cfg [mypy]` config (e.g. a per-module `[mypy-datahub.*]` override is honored).
- [ ] On a module whose venv hasn'"'"'t been built, the hook fails fast with the `./gradlew :<module>:installDev` instruction.
- [ ] Module whose `build.gradle` does **not** run mypy (ruff-only) → no mypy hook emitted.
- [ ] `pre-commit` (commit-time) hooks unchanged; mypy only fires at `git push`.

## Notes
- Stacked on `dc/pfp-5000-staged-ruff` (base branch for this PR).
- The staged mypy check is a subset of CI'"'"'s full `./gradlew :<module>:lint` (which still runs in CI); pre-push now only type-checks the files being pushed, for speed.
- Refs PFP-5000, PFP-5002.

### DIFF
--- a/.github/scripts/generate_pre_commit.py
+++ b/.github/scripts/generate_pre_commit.py
@@ -219,7 +219,7 @@ class HookGenerator:
         """
         files_regex = f"^{project.path}/.*\\.py$"
         wrapper = "bash .github/scripts/python_staged_ruff.sh"
-        return [
+        hooks = [
             {
                 "id": f"{project.project_id}-ruff-check",
                 "name": f"{project.path} Ruff Check",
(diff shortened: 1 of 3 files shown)


david-leifker (approved):
Approved after review.

devashish2203:

PR #18490, opened 2026-07-19
ci(pre-commit): consolidate pre-push guard + move self-correcting formatters to pre-push (PFP-5002)

devashish2203:
## Summary

Move all self-correcting formatters (ruff format, spotlessApply, prettier, cypressLintFix, playwright lintFix) from pre-commit to the **pre-push** stage so commits stay fast, and consolidate the standalone `.githooks/pre-push-oss` 500-file push guard into the pre-commit framework as a `guard-push-file-count` hook at pre-push. Linters, codegen, and checks stay at pre-commit. Add explicit `stages` to every hook so the stage assignment is unambiguous, and wire up `pre-commit install --hook-type pre-push` in `settings.gradle` so the pre-push hooks actually run.

Sixth layer of the PFP-4982 pre-commit/CI optimization stack (stacked on `dc/pfp-5001-playwright`).

- `ruff-check` → `stages: [pre-commit]` (it applies safe lint fixes that can change code, so catch before commit).
- `ruff-format` → `stages: [pre-push]` (pure formatter, no meaning change).
- `spotlessApply` (Java) and `prettier` hooks → `stages: [pre-push]` (self-correcting formatters, slow module-wide Gradle invocations).

### `.github/scripts/pre-commit-override.yaml`
- Add `guard-push-file-count` hook (`stages: [pre-push]`, `always_run: true`, `pass_filenames: false`) — the 500-file push guard, now a pre-commit-framework hook.
- Add explicit `stages` to every existing override hook: linters/codegen/checks (update-lineage-file, check-gradle-lockfiles, check-quickstart-compose, validate-post-workflow-list, actionlint, docusaurus-markdown-lint, bump-schema-versions, betterleaks) stay `pre-commit`; self-correcting formatters (smoke-test-cypress-lint-fix, playwright-e2e-lint-and-format) move to `pre-push`.
- Comment on `guard-push-file-count` noting the ordering caveat: the generator emits generated hooks (formatters) before override hooks, so at pre-push the formatters run once before the guard. A >500-file push is still rejected, but only after the formatters have run on the diff. Making the guard fast-fail first would require changing the generator'"'"'s merge order — deferred as not worth it.

### `.github/scripts/guard_push_file_count.sh` (new)
The 500-file push guard logic, adapted for the pre-commit framework: pre-commit sets `PRE_COMMIT_FROM_REF` / `PRE_COMMIT_TO_REF` (it does not forward git'"'"'s stdin ref pairs), so the guard uses those env vars. Fail-open on missing/empty refs — this is an accident guard, not a security control, so a stale/missing object must never block a real push.

### `.githooks/pre-push-oss` (deleted)
The old standalone pre-push hook is removed; its logic now lives in `guard_push_file_count.sh` + the override entry.

### `settings.gradle`
- After `pre-commit install`, also run `pre-commit install --hook-type pre-push` so the `pre-push`-stage hooks run on `git push`.
- Remove the old `installPrePushHook()` function (which manually copied `.githooks/pre-push-oss` into `.git/hooks/`).

### `.pre-commit-config.yaml`
Regenerated with explicit `stages` on all hooks and the new `guard-push-file-count` pre-push hook.

## Test plan
- [ ] Commit a `.java` / `.py` / `.ts` file → pre-commit runs only the linters/checks (fast); formatters do **not** run at commit.
- [ ] `git push` → pre-push runs the formatters (spotlessApply, ruff format, prettier, etc.) and the `guard-push-file-count` hook.
- [ ] Push changing >500 files → rejected by `guard-push-file-count` with the clear message; humans can bypass with `git push --no-verify`.
- [ ] Push changing ≤500 files → guard passes; formatters run first, then guard.
- [ ] Confirm `pre-commit install --hook-type pre-push` runs on a fresh clone (via `settings.gradle`) and that the old `.githooks/pre-push-oss` is no longer referenced.
- [ ] In a git worktree (where `.git` is a file), pre-push hooks still install correctly via `pre-commit install --hook-type pre-push`.

## Notes
- Stacked on `dc/pfp-5001-playwright` (base branch for this PR).
- Per the agent guidance, the guard message tells agents never to bypass with `--no-verify` and to ask the user to check the commits.
- Refs PFP-5002.


AdrianMachado (approved):

david-leifker (approved):
Approved after review.

devashish2203:
merged via https://github.com/datahub-project/datahub/pull/18495

PR #18494, opened 2026-07-20
build(lint): bump ruff to 0.15.22 and resolve fallout

devashish2203:
## Summary

Bumps Ruff to **0.15.22** across all Python modules and the pre-commit Docker tag, then resolves the fallout so the repo-wide pre-commit hooks stay clean.

Stacked on top of #18491 (`dc/pfp-4982-staged-collapses`), which collapsed per-module ruff + mypy hooks into repo-wide hooks using a Docker ruff tag. This branch bumps that tag (and every module's pinned ruff) to the latest and fixes what broke.

## Changes

- `build(lint): bump ruff to 0.15.22 across all modules and the pre-commit Docker tag` — bumps the `[tool.ruff]`/dev-deps pins in every Python module's `pyproject.toml` and the `ruff_image` constant in `.github/scripts/pre-commit/generate_pre_commit.py` (Docker tag `0.15.18` → `0.15.22`).
- `ruff Format all files` / `Ruff check auto fix` / `Fix all ruff check violations` — applies `ruff format` and `ruff check --fix` fallout from the new version across all modules, plus manual fixes for violations ruff couldn't auto-fix.
- `Update pre-commit to include toml files` — adds `pyproject.toml` and `ruff.toml` to the ruff hooks' `files:` regex so config edits also trigger the hooks.
- `fix(examples): make datahub-agent-context examples mypy-clean` — adds inline `# type: ignore[import-not-found]` for optional third-party imports in the `datahub-agent-context/examples/` agents so the pre-push mypy hook passes on them.

## Testing

- Regenerated `.pre-commit-config.yaml` via `generate_pre_commit.py` and validated it.
- `ruff check .` and `ruff format --check .` clean across all Python modules on 0.15.22.
- Pre-push mypy hook passes on touched example files.

## Tracking

Linear: PFP-4982

PR #18491, opened 2026-07-19
build(pre-commit): collapse per-module ruff + mypy into repo-wide hooks

devashish2203:
## Summary

Collapse the per-module ruff and mypy pre-commit/pre-push hooks into **repo-wide** hooks:
- **ruff**: replace the per-module `<module>-ruff-check` / `<module>-ruff-format` hooks with two repo-wide hooks (`ruff-check` at pre-commit, `ruff-format` at pre-push) using the official ruff Docker image pinned to **0.15.18** (the version the modules were on at this point in the stack). ruff does per-file config discovery, so each module's `[tool.ruff]` still applies, and ruff never imports code or deps, so no module venv is needed.
- **mypy**: replace the per-module `<module>-mypy` pre-push hooks with one repo-wide `mypy` pre-push hook. mypy uses single-config (CWD-based) discovery and resolves imports against each module's installed deps, so it cannot be a single process across modules — the wrapper groups staged files by module and invokes each module's own `<module>/venv/bin/mypy` from the module dir (so each module's mypy config still applies).

Delete the now-unused `python_staged_ruff.sh` wrapper (ruff runs from the Docker image, no per-module venv ruff needed).

This branch is **collapse-only**. The ruff version bump to 0.15.22 and the resulting fallout fixes (examples, `setup.py`, toml-file watching) are split into a follow-up branch (#18494) so the collapse can be reviewed in isolation.

Seventh layer of the PFP-4982 pre-commit/CI optimization stack (stacked on `dc/pfp-5002-pre-push`).

### `.github/scripts/generate_pre_commit.py`
- Remove per-module ruff/mypy hook emission; collect Python module dirs that have a `build.gradle` (`ruff_module_dirs`) and the subset whose `build.gradle` runs mypy (`mypy_module_dirs`).
- Emit two repo-wide ruff hooks (`ruff-check` pre-commit, `ruff-format` pre-push) with `language: docker_image`, `entry: ghcr.io/astral-sh/ruff:0.15.18 check --fix` / `... format`, and a `files` regex that is a union of the module dirs (so non-module `.py` is not newly linted).
- Emit one repo-wide `mypy` pre-push hook whose entry is `bash .github/scripts/python_staged_mypy.sh --modules <comma-list>`.

### `.github/scripts/python_staged_mypy.sh`
[...]

PR #18495, opened 2026-07-20
build(lint): widen CI ruff/mypy to check all directories per Python module

devashish2203:
## Summary

Widens the Gradle `lint`/`lintFix` tasks in every Python module so Ruff and Mypy check **all** directories under the module root (`.`) instead of an explicit allowlist of `src/ tests/ examples/ scripts/`. This keeps CI and the pre-commit/pre-push hooks aligned: a file linted locally is also linted in CI, and vice versa. Also fixes the mypy/ruff fallout exposed by widening the scope, so the widened CI is green.

Top of the PFP-4982 pre-commit/CI optimization stack (stacked on `dc/pfp-4982-cleanup-and-optimizations`).

### `build(lint): widen CI ruff/mypy to check all directories per Python module`
Replaces the explicit directory lists (`src/ tests/ examples/ scripts/`) with `.` in the `ruff check`, `ruff format --check`, and `mypy` invocations of the `lint` and `lintFix` tasks across all Python modules (`metadata-ingestion`, `datahub-actions`, `datahub-agent-context`, the four `metadata-ingestion-modules/*` plugins, `smoke-test`, `scripts/dev`, `metadata-service/iceberg-catalog`). Safe because `build`/`dist`/`venv` are excluded via `pyproject.toml` (`[tool.ruff] exclude`) and `setup.cfg` (`mypy exclude`) — adds `build`/`dist` to those exclude lists where they were missing.

Also fixes the fallout exposed by the wider scope:
- `metadata-ingestion/scripts/`: `check_wheel_coverage.py` (`var-annotated`), `connector_registry.py` (`misc` — read-only `PluginRegistry.mapping`), `docgen.py` (`attr-defined`/`arg-type` for `PackageMetadata`/`Collection` typing), `generate_patch_parity_golden.py` (`union-attr` guard), `modeldocgen.py` (Avro `Schema` typing — `attr-defined`/`arg-type`, `assignment`, sort-key `None`, loop var `e` rename), `tests/test_modeldocgen.py` (`var-annotated`, `index`).
- `smoke-test/`: `test_e2e.py` (`dict-item` on `variables` dicts), `test_resources/analytics_backfill/generate_glossary_terms.py` + `generate_users.py` (`Optional` for `None` defaults, explicit `Dict`/`List` annotations, dead-code removal, corrected return type).
- `scripts/dev/tests/test_runner.py`: ruff reformatting.

### `fix(lint): place docgen type: ignore on the failing arg line`
Moves the `# type: ignore[arg-type]` in `metadata-ingestion/scripts/docgen.py` onto the argument that actually fails, so the pre-push mypy hook (which checks `docgen.py` in single-file `--follow-imports=silent` mode) agrees with the full `mypy .` run.

### `fix(examples): silence mypy dict-item error in langchain basic_agent connection config`
Adds `# type: ignore[dict-item]` on the `MultiServerMCPClient({"datahub": connection_config})` call in `datahub-agent-context/examples/langchain/basic_agent.py`, where the `connection_config: dict` (inferred `dict[Any, Any]`) doesn't match the connection union type expected by the optional `langchain_mcp_adapters` dep. Matches the existing inline-ignore pattern used for optional-dep typing in the examples.

## Testing

- `./gradlew :metadata-ingestion:lint` and the per-module `mypy .` / `ruff check .` runs are clean.
- Pre-push mypy hook passes on `scripts/docgen.py` and the langchain example in both single-file and full-module modes.

## Tracking

Linear: PFP-4982


david-leifker (approved):
Approved after review.

devashish2203:
# Pre-commit Hook Optimization — Before/After Benchmark (PFP-4982)

A controlled before/after benchmark comparing `master` vs `dc/pfp-4982-widen-ci-lint` (top of the optimization stack) for the time to make representative code changes, commit them (one commit per file type), and push the branch.

## Methodology

- One git worktree per branch, on its own test branch (`dc/master-pre-commit-test`, `dc/pfp-4982-widen-ci-lint-pre-commit-test`).
- Identical edits in both: a marker comment appended to one existing file per file type (Java, Python across 3 modules, frontend-web-react markdown, Playwright `.ts`).
  - Java: `metadata-service/configuration/.../FlagClassificationGenerator.java`
  - Python (3 modules): `metadata-ingestion`, `datahub-actions`, `datahub-agent-context` `__init__.py`
  - Frontend-web-react: `datahub-web-react/README.md` (the only frontend-module pre-commit hook is `mdPrettierWriteChanged`; source `.tsx` has no pre-commit hook on either branch — CI-only)
  - Playwright: `e2e-test/ui/playwright/tests/home/homepage-modules.spec.ts`
- Toolchains warmed up before measuring (Gradle daemon, Python venvs via `installDev`, yarn installs) so the timed run reflects steady-state iterative dev, not first-run install.
- 4 commits (one per file type) then `git push -u origin <branch>`. Each step timed with `date +%s.%N`; overall = commit-start → push-complete.
- `pre-commit` hooks installed in both worktrees; `pre-commit install --hook-type pre-push` in the widen worktree (master's standalone pre-push guard skips in worktrees, so master's push has no pre-push hooks).

## Results

### Per-step timings (seconds)

| Step                         | Master  | Widen   | Δ (widen − master) | Commit-time change |
| ---------------------------- | -------: | ------: | -----------------: | ------------------ |
| Java commit                  | 5.12     | 1.46    | −3.66              | −71%               |
| Python commit (3 modules)   | 10.20    | 2.52    | −7.68              | −75%               |
| Frontend-web-react commit   | 2.90     | 1.04    | −1.86              | −64%               |
| Playwright commit            | 10.34    | 0.87    | −9.47              | −92%               |
| **Total commit time**        | **28.56**| **5.89**| **−22.67**         | **−79%**           |
| Push                         | 23.61    | 110.93  | +87.32             | +370%              |
| **Overall (commit→push)**   | **52.51**| **117.15**| +64.64           | +123%              |

### Pre-push hook breakdown (widen push ≈ 87 s hooks + ~24 s network)

| Pre-push hook (widen)                  | Time (warm) |
| -------------------------------------- | ----------: |
| `ruff-format` (Docker, 3 py files)      | ~0.5 s      |
| `mypy` wrapper (3 modules, 3 files)    | ~1.0 s warm (cold first run is the dominant cost during the actual push) |
| Java `spotlessApply` (1 module)        | ~1.8 s      |
| `mdPrettierWriteChanged` (web-react)   | gradle/yarn (warm) |
| `playwright-e2e-lint-and-format` (`eslint . --fix` + prettier, module-wide) | dominant push cost |
| `guard-push-file-count`                 | ~0 s        |

## Interpretation

### What got faster — the commit loop (the frequent action)
- Every commit is **~5× faster**: 5–10 s each on master → ~1 s each on widen.
- Total commit time **−79%** (28.56 s → 5.89 s).
- Root causes:
  - **Python**: master ran a module-wide Gradle `lintFix` per touched module (ruff over the whole module); widen runs `ruff-check` from Docker on **staged files only**.
  - **Java / frontend / Playwright**: master ran the mutating formatters (`spotlessApply`, `mdPrettierWriteChanged`, `playwright lintFix+prettierWrite`) at **commit**; widen **moved them to pre-push**, so commit-time only runs fast non-mutating checks (e.g. `betterleaks`, `bump-schema-versions`) that don't match the staged files.
  - Playwright commit dropped 10.34 s → 0.87 s because the two separate Gradle calls (lint-fix + prettier-write) moved out of commit.

### What got slower — the push (the infrequent action)
- Push **23.6 s → 110.9 s**. Master's push is ~24 s of network with **no pre-push hooks** (its lint happens at commit; its standalone guard skips in worktrees). Widen's push runs **all formatters + mypy consolidated at pre-push** (~87 s of hooks + ~24 s network).
- The ~87 s of push-hook work is:
  1. The formatters that **moved from commit to push** (spotless, prettier, ruff-format, playwright lint+prettier) — this is the same work master did at commit, just batched into one push.
  2. **mypy at pre-push** — work master never did locally (master's `lintFix` is ruff-only; mypy runs only in CI on master). So widen adds local mypy that master defers to CI. This is the main reason **overall** wall-clock is higher.

### Net
- **Overall wall-clock is higher on widen (52.5 s → 117.2 s)** — but this is misleading as a "developer experience" metric because:
  - The **frequent** action (commit) is ~5× faster and non-blocking (~1 s vs 5–10 s per commit). A developer commits many times per branch; they push once.
  - Widen now runs **mypy locally** that master pushed to CI, so part of the push increase is "CI work pulled forward into local push," not pure overhead.
  - The push is a **one-time batch** per branch; the commit loop is the repeated cost the refactoring targets.

### Caveats
- Master's `lintFix` pre-commit hook is **ruff-only** (no mypy); mypy is CI-only on master. Widen adds mypy at pre-push. An apples-to-apples comparison excluding mypy would show widen's push ≈ master's commit-lint work relocated, not a net increase.
- First-run costs (yarn install, cold mypy cache, Docker image pull) were warmed up before measuring; both branches measured in the same warm state.
- Gradle daemons had to be run **sequentially** (one build at a time) — two concurrent Gradle builds from different worktrees deadlocked on the shared daemon registry. This affects only setup, not the measured run.
- Frontend-web-react source `.tsx` has no pre-commit hook on either branch (CI-only); the markdown hook was used to exercise the frontend-module formatter that the refactoring moved to pre-push.

## Cleanup
- Test branches pushed to origin: `dc/master-pre-commit-test`, `dc/pfp-4982-widen-ci-lint-pre-commit-test` (can be deleted).
- Worktrees at `worktrees/datahub-test-master` and `worktrees/datahub-test-widen` (can be removed with `git worktree remove`).

devashish2203:
> Approved after review.
> 
> Note (stack): #18490 formatter-on-pre-push UX is intentional but worth watching; #18491 should land with #18494 so Docker ruff and module pins stay aligned.

Since stacked PRs are not available in OSS yet, pointed the top of the stack to master and merging to ensure that all commits land together.

devashish2203:
All failures are unrelated to this. Merging the PR.

PR #18496, opened 2026-07-20
fix(ingest/fivetran): emit per-table-pair lineage to eliminate task fan-out

alfiyas-datahub:
## Summary

Fivetran modeled each connector as a single DataJob (task) that held every source table as an inlet and every destination table as an outlet. DataHub resolves table-level lineage through a task as the cross-product of its inlets and outlets, so each destination table appeared to depend on **all** source tables in the connector — a fan-out. A connector syncing ~75 tables made every output look like it had ~75 upstreams instead of its one true source.

This change:

- Emits **one DataJob per `source_table -> destination_table` pair**, each with exactly one inlet, one outlet, and only that pair's fine-grained (column) lineage. This removes the cross-product at the table level.
- Keeps a single **connector-level DataJob** to anchor sync run-history and connector metadata, but emits it with an **empty `DataJobInputOutput`**. That both keeps it out of table lineage and clears any previously ingested fan-out lineage on that URN on re-ingest.
- Uses a deterministic, length-safe per-pair DataJob id so URNs are stable across runs.

Column-level lineage was already correct; this fixes the table-level view.

Internally, the monolithic `_extend_lineage` was split into focused helpers (`_resolve_source_details`, `_build_input_dataset_urn`, `_build_fine_grained_lineages`, `_compose_lineage_properties`, `_generate_table_datajobs`).

## Test plan

- [x] New unit tests (`test_fivetran_lineage_fanout.py`): each destination has exactly one upstream, no DataJob carries multiple inlets, per-pair column lineage is scoped to its pair, per-pair job ids are unique/deterministic, and undetermined destination URNs skip edges with a single dedup'd warning.
- [x] Updated immutability + Google Sheets unit tests for the refactored helpers.
- [x] Regenerated Fivetran integration golden files.
- [x] `./gradlew :metadata-ingestion:lint` (ruff + mypy) passes.
- [x] 222 fivetran tests pass; `fivetran.py` coverage 95%.


metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran.py
```diff
@@ -762,24 +746,117 @@ def _generate_datajob_from_connector(self, connector: Connector) -> DataJob:
...
+            self._connector_custom_properties(connector, lineage_properties)
+        )
+        return datajob
 
-        # Map connector source and destination table with dataset entity
-        # Also extend the fine grained lineage of column if include_column_lineage is True
-        lineage_properties = self._extend_lineage(connector=connector, datajob=datajob)
-        # TODO: Add fine grained lineages of dataset after FineGrainedLineageDownstreamType.DATASET enabled
+    @staticmethod
+    def _table_datajob_name(connector: Connector, lineage: TableLineage) -> str:
```
treff7es, line 790: **Wire up stateful ingestion / stale-entity removal — this is now required, not optional.**

This method makes the DataJob URN churn: it's derived from the table names, and we emit up to `max_table_lineage_per_connector` (120) of them per connector. When a table pair is dropped or renamed, its per-pair DataJob URN stops being emitted and becomes an orphan carrying stale lineage. The old single-job model never had this because its URN was stable and just got overwritten each run.

The source already extends `StatefulIngestionSourceBase` and the config exposes `stateful_ingestion`, but nothing consumes it — there's no `StaleEntityRemovalHandler` and no `get_workunit_processors` override, so stale removal is dead config today. Since this PR is what makes the URNs churn, it should turn the handler on. Small, well-trodden change (odcs / github_documents are clean recent examples):

```python
from functools import partial
from datahub.ingestion.api.source import MetadataWorkUnitProcessor
from datahub.ingestion.source.state.stale_entity_removal_handler import StaleEntityRemovalHandler
from datahub.ingestion.source.state.entity_removal_state import GenericCheckpointState
from datahub.ingestion.workunit_processors.auto_stale_entity_removal import auto_stale_entity_removal

# in __init__ (self.state_provider comes from StatefulIngestionSourceBase)
self.stale_entity_removal_handler = StaleEntityRemovalHandler(
    state_provider=self.state_provider,
    report=self.report,
    config=self.config,
    state_type_class=GenericCheckpointState,
    pipeline_name=ctx.pipeline_name,
    run_id=ctx.run_id,
    platform=Constant.ORCHESTRATOR,
)

def get_workunit_processors(self) -> List[Optional[MetadataWorkUnitProcessor]]:
    return [
        *super().get_workunit_processors(),
        partial(auto_stale_entity_removal, self.stale_entity_removal_handler),
    ]
```

`auto_stale_entity_removal` tracks primary-source entities emitted this run and soft-deletes anything in the last checkpoint but absent now. Two things to confirm: (1) emission must flow through `get_workunit_processors()` — if you override `get_workunits` directly instead of `get_workunits_internal` the processor silently no-ops (Airbyte left a comment about exactly this trap); (2) referenced upstream/downstream tables are lineage endpoints, not primary-source, so they won't be touched — good; the Google Sheets `Dataset` entities are primary-source so they'll be tracked too. Worth a test: ingest two pairs, re-ingest with one gone, assert a `status.removed=true` for the dropped pair's DataJob.

metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran.py
```diff
@@ -762,24 +746,117 @@ def _generate_datajob_from_connector(self, connector: Connector) -> DataJob:
...
+        )
+        owner_email = self.log_reader.get_user_email(connector.user_id)
+        owners = [CorpUserUrn(owner_email)] if owner_email else None
 
-        datajob.set_custom_properties({**connector_properties, **lineage_properties})
+        seen_pairs: set[str] = set()
+        for lineage in connector.lineage:
+            pair_key = f"{lineage.source_table}\t{lineage.destination_table}"
+            if pair_key in seen_pairs:
+                continue
```
treff7es: The dedup is needed (two rows with the same pair would collide on the DataJob URN), but we keep the first and drop the rest with no counter or log. On master, `_extend_lineage` accumulated every row's column lineage into the one job, so if Fivetran ever returns the same pair across multiple rows (e.g. column lineage split across rows) those columns now vanish with no trace. Can we add a `report` counter (or `logger.debug`) here, or merge the column lineage of duplicate pairs instead of discarding?

metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran.py
```diff
@@ -264,120 +267,80 @@ def _extend_lineage(self, connector: Connector, datajob: DataJob) -> Dict[str, s
...
+                FineGrainedLineage(
+                    upstreamType=FineGrainedLineageUpstreamType.FIELD_SET,
+                    upstreams=(
+                        [
+                            builder.make_schema_field_urn(
+                                str(input_dataset_urn),
+                                column_lineage.source_column,
+                            )
+                        ]
+                        if input_dataset_urn
```
treff7es: When `input_dataset_urn is None` (the Google Sheets resolve-failure path) with non-empty column lineage, we still append a `FineGrainedLineage` per column with `upstreams=[]` and a populated `downstreams`. Worth confirming that's a valid aspect and not a malformed edge — I'd lean toward skipping the FGL entry entirely when there's no upstream. There's no test for None-input + non-empty column lineage (the gsheets test uses `column_lineage=[]`), so this path is currently unexercised.

metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran.py
```diff
@@ -762,24 +746,117 @@ def _generate_datajob_from_connector(self, connector: Connector) -> DataJob:
...
-            "connector_type": connector.connector_type,
-            "paused": str(connector.paused),
-            "sync_frequency": str(connector.sync_frequency),
-            "destination_id": connector.destination_id,
-        }
+        Each job has exactly one inlet and one outlet (plus that pair's column
+        lineage), so DataHub resolves table lineage precisely instead of
+        fanning every source out to every destination.
+        """
+        dataflow_urn = DataFlowUrn.create_from_ids(
```
treff7es: Nit: `dataflow_urn` and `owner_email` are computed here and again in `_generate_datajob_from_connector`. The caller `_get_connector_workunits` already resolves `source_details`/`destination_details`/`lineage_properties` once and threads them in — worth doing the same for these two. (`get_user_email` is lru-cached so it's not a perf issue, just consistency with the pattern the PR already set up.)

metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran.py
```diff
@@ -762,24 +746,117 @@ def _generate_datajob_from_connector(self, connector: Connector) -> DataJob:
...
+                            "prefix, to fix."
+                        ),
+                        context=(
+                            f"connector_id={connector.connector_id}, "
+                            f"destination_id={connector.destination_id}, "
+                            f"table={lineage.destination_table}"
+                        ),
+                        exc=e,
+                    )
+                continue
```
treff7es, line 878: Nit (pre-existing): the deduped "Destination URN could not be constructed" warning says subsequent edges are skipped silently, but there's no count, so an operator can't tell if 1 or 50k tables were dropped for a destination. A `self.report.num_lineage_edges_skipped += 1` on every `continue` would close it cheaply.

metadata-ingestion/tests/unit/fivetran/test_fivetran_lineage_fanout.py
```diff
@@ -0,0 +1,231 @@
...
+        destination_table=long_table,
+        column_lineage=[],
+    )
+    connector = _connector(0)
+
+    name_1 = source._table_datajob_name(connector, lineage)
+    name_2 = source._table_datajob_name(connector, lineage)
+
+    assert name_1 == name_2
+    assert len(name_1) < len(long_table)
```
treff7es: This assertion passes trivially — a 250-char id would also be shorter than the long table name. The invariant the fallback actually protects is the URN-length cap, so assert `len(name_1) <= 200` instead (and maybe one case confirming the sub-200 readable path is *not* hashed).

metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran.py
alfiyas-datahub: I connected the `StaleEntityRemovalHandler` so it now runs correctly during ingestion, following the same pattern used by the `odcs` and `github_documents` sources. I also verified that our source uses `get_workunits_internal()` (not `get_workunits()`), so the cleanup actually happens, avoiding the issue we previously saw with Airbyte. The new per-table-pair DataJobs and Google Sheets datasets are treated as primary entities, so they're tracked for cleanup. The upstream and downstream tables are only lineage references, so they aren't tracked. I also added a test to verify the wiring works correctly.

metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran.py
alfiyas-datahub: Done. Instead of ignoring duplicate source-to-destination table pairs, we now merge them. If the same table pair appears multiple times, all of their column lineage is combined into a single DataJob, so no column mappings are lost. I also added a metric (num_duplicate_table_pairs_merged) to track how often this happens, and a test to verify that column lineage from all duplicate rows is preserved in the final job.

metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran.py
alfiyas-datahub: Agreed, the column lineage was incorrect when there was no input dataset. I fixed it so that if there's no input URN, we don't create any column-level lineage. The table-level lineage (the output table) is still created as before. I also added a test to make sure this case works correctly when there's no input dataset but column lineage is present.

metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran.py
alfiyas-datahub: Fixed. Both are computed once in `_get_connector_workunits` and threaded into `_generate_datajob_from_connector` and `_generate_table_datajobs`.

metadata-ingestion/src/datahub/ingestion/source/fivetran/fivetran.py
alfiyas-datahub: Added `report.num_lineage_edges_skipped`, incremented on every skipped edge. The warning stays deduped per destination but now points at the counter so operators can see whether it was 1 or 50k.

metadata-ingestion/tests/unit/fivetran/test_fivetran_lineage_fanout.py
alfiyas-datahub: Updated the condition to use `len(name_1) <= 200`, so names that are 200 characters or shorter stay as readable names. Only longer names are shortened using a hash. I also added a test to verify that names under the 200-character limit are returned exactly as they are, without being hashed.

treff7es (approved):

PR #18499, opened 2026-07-20
perf(recommendations): batch candidate validation to remove per-platform N+1

gdatahub:
## Summary

On the home page, `ListRecommendationsResolver` → `RecommendationsService` runs
several `EntitySearchAggregationSource` candidate sources (Platforms, Domains,
Tags, Terms). After a single Elasticsearch aggregation, the base class
validated **each aggregated candidate one at a time** inside `getTopKValues` —
one backend call per candidate (a classic N+1).

`TopPlatformsSource` was the worst case: it overrode `isValidCandidateUrn` to
call `entityService.getLatestAspect(urn, "dataPlatformInfo")` **per candidate
platform** (up to `maxContent = 40` serial aspect fetches per home-page
request). The domain/tag/term sources hit the base default, which did a
per-candidate `exists()` check.

## Change

Replace the per-candidate validation hook with a batched
`getValidCandidateUrns(opContext, Set<Urn>)` that validates the whole
candidate set in a single call:

- **Base default** batches `entityService.exists(urns)` (one call for all urns).
- **`TopPlatformsSource`** overrides it to fetch `dataPlatformInfo` for all
  candidates via one `getLatestAspects(...)` call, keeping only logo-bearing
  platforms.
- **`getTopKValues`** now selects the top-K from an already-validated map and
  performs no I/O. The candidate set is already bounded by the aggregation
  size, so validate-all-then-top-K is equivalent to the previous
  top-K-among-valid selection.

Because the fix lives in the shared base class, Domains/Tags/Terms are batched
too (they used the per-candidate `exists` default).

## Behavior

Unchanged — identical recommendations are returned. Only the backend call
count per source drops from **O(candidates) → O(1)**.

## Testing & verification

**Unit (deterministic call counts):**
- New `TopPlatformsSourceTest` — with 30 candidate platforms the source makes
  exactly **one** `getLatestAspects` call and **zero** per-candidate
  `getLatestAspect` calls; the call count stays at 1 for both 3 and 30
  candidates (O(1)); and the same logo-bearing platforms are recommended.
- Updated `EntitySearchAggregationCandidateSourceTest` to the batched
  `exists(Collection<Urn>)` contract.

**End-to-end (local quickstart, sample metadata, 13 platforms, OpenTelemetry → Jaeger):**

| Signal | Before | After |
| --- | --- | --- |
| `metadata_aspect_v2` SELECT spans under the `TopPlatforms` `getRecommendations` span | 27 | 1 (single batched query) |
| Spans in the `listRecommendations` trace | 185 | 97 |
| Server-side `listRecommendations` span, p50 | ~20 ms | ~12 ms |

**Client latency (`curl`, 100 warm requests each, identical instance/data):**

| metric | Before (N+1) | After (batched) | improvement |
| --- | --- | --- | --- |
| p50 | 17.6 ms | 14.0 ms | −21% |
| mean | 18.4 ms | 14.7 ms | −20% |
| p90 | 21.1 ms | 16.2 ms | −24% |

The absolute latency delta is modest here because it is a warm, small dataset
(13 platforms, cached rows); the saving scales with the number of platforms
and with cache coldness (up to `maxContent = 40` serial fetches → 1 on a cold
production index). Recommendations were byte-for-byte identical before and
after in all runs.

## Checklist

- [x] Tests added/updated
- [ ] Docs — n/a (internal refactor, no user-facing or API change)

### DIFF
--- a/metadata-service/services/src/main/java/com/linkedin/metadata/recommendation/candidatesource/EntitySearchAggregationSource.java
+++ b/metadata-service/services/src/main/java/com/linkedin/metadata/recommendation/candidatesource/EntitySearchAggregationSource.java
@@ -27,6 +27,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.PriorityQueue;
+import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.annotation.Nonnull;
@@ -57,22 +58,19 @@ public abstract class EntitySearchAggregationSource implements RecommendationSou
   /** Whether the aggregate value is an urn */
   protected abstract boolean isValueUrn();
 
-  /** Whether the urn candidate is valid */
-  protected boolean isValidCandidateUrn(@Nonnull OperationContext opContext, Urn urn) {
-    return entityService.exists(opContext, urn, false);
+  /**
+   * Filters the aggregated urn candidates down to the valid ones in a single batched lookup.
+   * Subclasses that need aspect-level validation should override this and batch the aspect fetch
+   * rather than checking one urn at a time.
+   */
+  protected Set<Urn> getValidCandidateUrns(
+      @Nonnull OperationContext opContext, @Nonnull Set<Urn> candidateUrns) {
+    return entityService.exists(opContext, candidateUrns, false);
   }
 
   /** Whether the string candidate is valid */
   protected boolean isValidCandidateValue(String candidateValue) {
     return true;
-  }
-
-  /** Whether the candidate is valid Calls different functions if candidate is an Urn */
-  protected <T> boolean isValidCandidate(@Nonnull OperationContext opContext, T candidate) {
-    if (candidate instanceof Urn) {
-      return isValidCandidateUrn(opContext, (Urn) candidate);
-    }
-    return isValidCandidateValue(candidate.toString());
   }
 
   @Override
@@ -95,7 +93,11 @@ public abstract class EntitySearchAggregationSource implements RecommendationSou
 
     // If the aggregated values are not urn, simply get top k values with the most counts
     if (!isValueUrn()) {
-      return getTopKValues(opContext, aggregationResult).stream()
+      Map<String, Long> validValues =
+          aggregationResult.entrySet().stream()
+              .filter(entry -> isValidCandidateValue(entry.getKey()))
+              .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+      return getTopKValues(validValues).stream()
           .map(entry -> buildRecommendationContent(entry.getKey(), entry.getValue()))
           .collect(Collectors.toList());
     }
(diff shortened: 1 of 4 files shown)


metadata-service/services/src/main/java/com/linkedin/metadata/recommendation/candidatesource/EntitySearchAggregationSource.java
```diff
@@ -57,24 +58,21 @@ public abstract class EntitySearchAggregationSource implements RecommendationSou
...
+   * Subclasses that need aspect-level validation should override this and batch the aspect fetch
+   * rather than checking one urn at a time.
+   */
+  protected Set<Urn> getValidCandidateUrns(
+      @Nonnull OperationContext opContext, @Nonnull Set<Urn> candidateUrns) {
+    return entityService.exists(opContext, candidateUrns, false);
   }
 
   /** Whether the string candidate is valid */
   protected boolean isValidCandidateValue(String candidateValue) {
```
RyanHolstien: This seems like an oddity, if we're cleaning up here let's go ahead and remove this too

metadata-service/services/src/main/java/com/linkedin/metadata/recommendation/candidatesource/EntitySearchAggregationSource.java
```diff
@@ -95,7 +93,11 @@ public List<RecommendationContent> getRecommendations(
 
     // If the aggregated values are not urn, simply get top k values with the most counts
     if (!isValueUrn()) {
-      return getTopKValues(opContext, aggregationResult).stream()
+      Map<String, Long> validValues =
```
RyanHolstien: Since isValidCandidateValue is a no-op (always true blindly), we can remove this full stream/collect.

RyanHolstien (approved):
Minor comments, are we potentially doing more lookups (albeit batched in a single go) due to the pre-processing? I think it's likely fine and more efficient, just curious if maybe there's an opportunity to save more there. Probably not worth the complexity though since the existence checks won't hold that much in memory.

gdatahub:
> Minor comments, are we potentially doing more lookups (albeit batched in a single go) due to the pre-processing? I think it's likely fine and more efficient, just curious if maybe there's an opportunity to save more there. Probably not worth the complexity though since the existence checks won't hold that much in memory.

No, same candidate count, just batched. The ES aggregation already caps the set at getMaxContent() (40 for platforms, 10 for the rest), and the old code iterated that same bounded set calling isValidCandidate per item. 
So it's the identical set of urns, one exists/getLatestAspects call instead of N. No extra lookups, and the returned Set<Urn> is small like you said.

metadata-service/services/src/main/java/com/linkedin/metadata/recommendation/candidatesource/EntitySearchAggregationSource.java
gdatahub: Yes, isValidCandidateValue is a no-op that's never overridden, so I will remove the method and collapse the non-urn branch to just pass aggregationResult into getTopKValues.

chakru-r (approved):

gdatahub:
**Before**
Screenshot 2026-07-25 at 10 09 54 AM

**After**
Screenshot 2026-07-25 at 10 09 41 AM

PR #17332, opened 2026-05-06
feat(metadata-models): register odcs as a DataPlatform

rbrenton:
Refs #11892. Companion to #17331.

## Summary

Registers the `odcs` data platform id and display name "Open Data Contract Standard" in the bootstrap MCPs registry. Two-file change per the file-header instructions in `bootstrap_mcps/data-platforms.yaml`:

1. New entry in `metadata-service/configuration/src/main/resources/bootstrap_mcps/data-platforms.yaml` — slotted alphabetically between `mariadb` and `OpenApi`.
2. Version bump `v9` → `v10` for the `data-platforms` template in `metadata-service/configuration/src/main/resources/bootstrap_mcps.yaml` so existing GMS instances re-apply the bootstrap on next start.

This PR is split out from the larger ODCS ingestion source PR (#17331) so the platform-id and display-name decisions can be reviewed and signed off independently before they ship.

Once the [Open Data Contract Standard (ODCS)](https://bitol.io/) ingestion source merges, datasets bound by ODCS contracts will surface with this registered display name in the DataHub UI and `data_platform.json`. Without this entry, the UI falls back to the title-cased plugin id (acceptable, but inconsistent with other registered platforms like `dbt`).

## Sign-off requested from maintainers

- **Platform id `odcs`** (lowercase) — matches the source's `@platform_name(..., id="odcs")` decorator and follows the convention of other registered short-form platform ids (`dbt`, `csv`, `json-schema`).
- **Display name "Open Data Contract Standard"** — full standard name. Alternative: "ODCS" if the maintainers prefer the abbreviated form.
- **`type: OTHERS`** — matches `dbt` (a contract / metadata-overlay platform; not a `RELATIONAL_DB` / `KEY_VALUE_STORE` / `FILE_SYSTEM` / etc.).
- **No `logoUrl` for now** — follows the convention of other entries that lack a logo (`ambry`, `external`, `voldemort`, `datahub`). A logo asset can be added in a follow-up if preferred; the upstream Bitol project provides a [logo](https://github.com/bitol-io/artwork) under their license terms.

## Test plan

- [x] `metadata-service/configuration/src/main/resources/bootstrap_mcps/data-platforms.yaml` is valid YAML and parses cleanly.
- [x] The new entry follows the schema of every other entry in the file (entityUrn, entityType, aspectName, changeType, aspect).
- [x] Alphabetically slotted between `mariadb` and `OpenApi`.
[...]

PR #18506, opened 2026-07-20
chore(ingest/odcs): drop PR-number references from source comments

acrylJonny:
## Summary

- Rewords the `odcs` platform-info emission comments in `odcs_mapper.py` and `odcs_source.py` so they explain the boot-time-vs-runtime aspect rationale **without** hardcoding a PR number (`#17332`).
- Hardcoded PR references are noise in shipped source; the "why" is preserved.

Comment-only change, no behavior impact.


acrylJonny:
Folded into #18477 (commit 03bffb1) instead of carrying a separate PR — it already touches the same odcs_source.py, so the comment reword lands there. Closing.

PR #212, opened 2016-08-31
Map scm repo owner to dataset owner table

alyiwang:
Use cfg_object_name_map table to find the mapping between repo and dataset for dalids. Then merge the repo owner info from ACL files into the dataset_owner table.

PR #18509, opened 2026-07-20
feat(cli): migrate Snowflake Semantic View governance to semanticModel/metric

kyungsoo-datahub:
## Summary
- Adds `datahub migrate snowflake-semantic-views` to copy DataHub governance when flipping `semantic_views.emit_semantic_model_entities` (dataset ↔ semanticModel/metric).
- Copies entity aspects (ownership, domains, tags, terms, etc.) and fans out column tags/terms to metric / schemaField URNs; strips synthetic DIMENSION/FACT/METRIC tags.
- Migrate-before-ingest: only the **source** entity must already exist. Destination URNs (semanticModel/metric/schemaField) may not exist yet — governance aspects are written onto the mapped URNs first, and a later Snowflake ingest with `emit_semantic_model_entities` fills in the structural aspects. Does not rewrite lineage, policies, data products, or soft/hard-delete.
- Discovery defaults to **live** entities; after stateful ingest, pass `--include-soft-deleted` deliberately (CLI warns if 0 live / N soft-deleted). Soft-deleted sources are marked in the report.
- `editableSchemaMetadata` is merged field-by-field (descriptions preserved; tags/terms unioned by URN).

## Do not merge before
**Blocked on connector URN contract:** [#18395](https://github.com/datahub-project/datahub/pull/18395) (`emit_semantic_model_entities` + `SnowflakeIdentifierBuilder` / semantic model mapper). This CLI mirrors those URN shapes; stacking/merge order must keep them aligned.

## Test plan
- [x] Unit tests: `tests/unit/cli/test_snowflake_semantic_view_migration.py`
- [ ] Dry-run against an instance: migrate governance first, then run #18395 ingest, and confirm structural aspects land on the same URNs
- [ ] Confirm missing **source** entity still errors out (destination absence is now expected/allowed)


metadata-ingestion/src/datahub/cli/snowflake_semantic_view_migration.py
```diff
@@ -0,0 +1,1158 @@
...
+) -> EntityMigrationResult:
+    """Copy the governance-aspect allowlist from src_urn to dst_urn.
+
+    Source must exist; destination may be created by writing aspects (migrate
+    before ingest). Always overwrites the destination's existing aspect values
+    (last-write-wins); no merge/conflict strategy is offered here (unlike
+    dataplatform2instance/instance2instance).
+    """
+    result = EntityMigrationResult(src_urn=src_urn, dst_urn=dst_urn)
+    missing = _ensure_src_exists(graph, src_urn)
```
treff7es, line 835: 🔴 **Blocker — one transient graph error aborts the whole batch, no report printed.**

`_ensure_src_exists` (→ `graph.exists()`) here, and `_is_soft_deleted()` at L897/L947, are called **outside** the per-entity `try` that starts at L845. `run_migration`'s loop (L1137) and the CLI wrap the loop in no try/except either.

So a momentary GMS 5xx on entity #212 of a 500-URN run raises straight out: #1–211 already have writes persisted, #212–500 never run, and the operator gets a traceback instead of the report of what landed — defeating the resumable/operator-visible premise.

Fix (small): move `_ensure_src_exists` inside `migrate_entity`'s `try`, move the `_is_soft_deleted` probes inside the existing per-entity guard, and/or wrap each `run_migration` loop iteration so one URN's failure becomes an `EntityMigrationResult(error=...)` rather than killing the run.

metadata-ingestion/src/datahub/cli/snowflake_semantic_view_migration.py
```diff
@@ -0,0 +1,1158 @@
...
+        governance_aspects = _collect_governance_aspects(graph, src_urn)
+
+        if fold_editable_description_from is not None:
+            note = _maybe_fold_editable_description(
+                graph, fold_editable_description_from, dst_urn, governance_aspects
+            )
+            if note is not None:
+                result.notes.append(note)
+
+        for aspect_name, aspect in governance_aspects.items():
```
treff7es, line 849: 🟡 **Partial-write is silently misreported.** Aspects are emitted one-by-one; if `emit_mcp` fails on aspect 3 of 5, aspects 1–2 are already persisted. The `except` at L864 sets `result.error`, but the failed-branch of the report (L1116–1117) prints only `ERROR: {error}` and never `aspects_copied` — so the operator has no record that 2 aspects already landed on the destination.

Consider batching these into one MCP emit, or recording emitted-so-far into `result` before the exception can discard it. (Recoverable via idempotent re-run, hence 🟡 not 🔴.)

metadata-ingestion/src/datahub/cli/snowflake_semantic_view_migration.py
```diff
@@ -0,0 +1,1158 @@
...
+        migrated, skipped = migrate_dataset_field_governance(
+            graph,
+            src_dataset_urn,
+            dst_urn,
+            identity,
+            platform_instance,
+            convert_urns_to_lowercase,
+            dry_run,
+        )
+        result.fields_migrated = migrated
```
treff7es, line 904: 🟡 **Field-fan-out failure reports 0 fields when N were written.** `migrate_dataset_field_governance` emits per-field with no per-emit capture; a mid-loop failure propagates to the `except` at L912 which appends a "field governance migration failed" note — but the exception fires *before* `result.fields_migrated = migrated` on this line, so the report shows **0 fields migrated** when several were already persisted. Assign progress before the emit loop can throw.

metadata-ingestion/src/datahub/cli/snowflake_semantic_view_migration.py
```diff
@@ -0,0 +1,1158 @@
...
+
+
+@dataclass(frozen=True)
+class SnowflakeViewIdentity:
+    db: str
+    schema: str
+    view: str
+
+
+def snowflake_identifier(identifier: str, convert_urns_to_lowercase: bool) -> str:
```
treff7es, line 121: 🟡 **Drift risk: URN math is hand-copied from an unmerged PR, with no parity test.** These helpers (and `_semantic_model_field_path` at L453) mirror `SnowflakeIdentifierBuilder`/`SnowflakeSemanticModelMapper` — which only exist in the not-yet-merged #18395. The unit tests never compare against the connector's implementation, so a future casing/quoting change on the connector side would silently produce wrong migration writes with no CI signal.

- **Interim (this PR):** add a test pinning the exact URN string shapes this CLI produces.
- **Follow-up (after #18395 lands):** extract a dependency-free `snowflake_identifiers.py` that both the connector and this CLI import (mirroring the zero-dep `snowflake/constants.py`), preserving the "no snowflake extra in the CLI" constraint while removing the drift.

metadata-ingestion/src/datahub/cli/snowflake_semantic_view_migration.py
```diff
@@ -0,0 +1,1158 @@
...
+                    global_tags=tags if isinstance(tags, GlobalTagsClass) else None,
+                    glossary_terms=(
+                        terms if isinstance(terms, GlossaryTermsClass) else None
+                    ),
+                )
+
+    return list(by_column.values())
+
+
+def migrate_semantic_model_field_governance(
```
treff7es, line 657: 🟡 **Full-aspect overwrite of `editableSchemaMetadata` has a lost-update race.** This does fetch → merge-in-Python → re-emit-whole-aspect. A concurrent UI edit to a *different* field between the read and the write is silently clobbered.

`DatasetPatchBuilder(dst).for_field(path).add_tag(...)/.add_term(...)` (`specific/dataset.py:58-100`) already does race-safe, additive, server-side per-tag/term writes — using it here would also delete the ~60 lines of hand-rolled `_union_*`/`by_path` merge logic. (Suggestion, not blocking — overwrite is a documented choice.)

metadata-ingestion/src/datahub/cli/snowflake_semantic_view_migration.py
```diff
@@ -0,0 +1,1158 @@
...
+    _Aspect,
+)
+from datahub.metadata.urns import DatasetUrn, MetricUrn, SemanticModelUrn
+from datahub.utilities.str_enum import StrEnum
+from datahub.utilities.urns.field_paths import get_simple_field_path_from_v2_field_path
+
+log = logging.getLogger(__name__)
+
+SNOWFLAKE_PLATFORM = "snowflake"
+SEMANTIC_VIEW_SUBTYPE = "Semantic View"
```
treff7es: ℹ️ **Reuse existing constants instead of re-declaring literals** (verified both import without the snowflake driver, so the "avoid the extra" rationale still holds):
- `DatasetSubTypes.SEMANTIC_VIEW` (`ingestion/source/common/subtypes.py:39`) for `"Semantic View"`.
- `DIMENSION`/`FACT`/`METRIC` from `ingestion/source/snowflake/constants.py:78-80` for the synthetic-tag set below (keep the lowercase variants — the connector enum lacks them).

Removes two more places where a future rename could silently drift.

metadata-ingestion/src/datahub/cli/snowflake_semantic_view_migration.py
```diff
@@ -0,0 +1,1158 @@
...
+    convert_urns_to_lowercase: bool,
+    dry_run: bool,
+) -> Tuple[List[str], List[str]]:
+    """Fan out DataHub column tags/terms to metric entities or schemaField URNs.
+
+    Returns (migrated entries, skip notes). Destination metric / schemaField URNs
+    need not exist yet (migrate-before-ingest).
+    """
+    migrated: List[str] = []
+    skipped: List[str] = []
```
treff7es, line 550: ℹ️ **`skipped` is dead — always empty, but the contract promises otherwise.** It's initialized here and returned at L576 but never appended to, yet the docstring says it returns "skip notes" and the caller does `result.notes.extend(skipped)` (L911) on an always-empty list. Either populate it (there are real skip cases worth reporting) or drop it from the return. A named result type (see the type-design note on L549) would have made this dead slot obvious.

metadata-ingestion/src/datahub/cli/snowflake_semantic_view_migration.py
```diff
@@ -0,0 +1,1158 @@
...
+    migrated: List[str] = []
+    notes: List[str] = []
+    for field_gov in fields:
+        field_path, path_note = _resolve_dataset_field_path(
+            field_gov.column_name, convert_urns_to_lowercase, schema_paths
+        )
+        if path_note is not None:
+            notes.append(path_note)
+        prior = by_path.get(field_path)
+        by_path[field_path] = EditableSchemaFieldInfoClass(
```
treff7es, line 703: ℹ️ **Dead nested ternary — simplify.** `_union_global_tags(x, None)` already returns `x` unchanged (L330-333), and `_union_glossary_terms` likewise (L344-347). So the `... if field_gov.global_tags is not None else (prior...)` guards are no-ops. All four prior×field_gov combinations behave identically to calling the union helpers unconditionally:

```python
prior_tags = prior.globalTags if prior is not None else None
prior_terms = prior.glossaryTerms if prior is not None else None
by_path[field_path] = EditableSchemaFieldInfoClass(
    fieldPath=field_path,
    description=prior.description if prior is not None else None,
    globalTags=_union_global_tags(prior_tags, field_gov.global_tags),
    glossaryTerms=_union_glossary_terms(prior_terms, field_gov.glossary_terms),
)
```

metadata-ingestion/src/datahub/cli/snowflake_semantic_view_migration.py
```diff
@@ -0,0 +1,1158 @@
...
+    only_soft_deleted: bool = False,
+) -> List[str]:
+    """Find Snowflake dataset urns carrying the "Semantic View" subtype.
+
+    Defaults to live entities only. After flag-ON ingest, sources are often
+    soft-deleted — pass ``include_soft_deleted=True`` (or use the CLI flag) once
+    operators have confirmed that set is intentional. ``only_soft_deleted`` is
+    for the empty-discovery hint probe.
+    """
+    if only_soft_deleted:
```
treff7es: ℹ️ **Duplicated status ladder + CLI direction-branching.** This exact 6-line `only_soft_deleted / include_soft_deleted / else` block is repeated verbatim in `discover_semantic_model_urns` (L1034-1039). And the CLI repeats the "which discoverer for which direction" branch 3× (main discovery, the soft-only probe, and the inline message label).

Extract `_status_filter(include_soft_deleted, only_soft_deleted)` and a `discover_migration_sources(graph, direction, ...)` dispatcher — the latter removes direction knowledge from `migrate.py` entirely.

metadata-ingestion/src/datahub/cli/snowflake_semantic_view_migration.py
```diff
@@ -0,0 +1,1158 @@
...
+    "siblings",
+    "incidentsSummary",
+    "testResults",
+    "forms",
+]
+
+# Relationship types whose target aspects reference a dataset/semanticModel
+# by urn. We never rewrite these -- we only report them so operators know
+# what still points at the pre-migration urn.
+INBOUND_REFERENCE_RELATIONSHIP_TYPES: List[str] = [
```
treff7es: ℹ️ **Duplicates `migration_utils`.** This list is byte-for-byte identical to the relationship types hardcoded in `migration_utils.get_incoming_relationships`, and `_fetch_inbound_refs` (L815) reimplements that function. Promote the constant to `migration_utils` and import it; ideally refactor `get_incoming_relationships` to accept a `graph` arg and reuse it, deleting `_fetch_inbound_refs`.

metadata-ingestion/src/datahub/cli/snowflake_semantic_view_migration.py
```diff
@@ -0,0 +1,1158 @@
...
+
+def migrate_dataset_field_governance(
+    graph: DataHubGraph,
+    src_dataset_urn: str,
+    semantic_model_urn: str,
+    identity: SnowflakeViewIdentity,
+    platform_instance: Optional[str],
+    convert_urns_to_lowercase: bool,
+    dry_run: bool,
+) -> Tuple[List[str], List[str]]:
```
treff7es, line 542: ℹ️ **Type design: prefer a small result dataclass over `Tuple[List[str], List[str]]`.** This return (and `migrate_semantic_model_field_governance` L664, `filter_by_semantic_view_subtype` L1052) is exactly the case the repo convention warns against — both slots are `List[str]`, distinguished only by position, so a caller can transpose them with zero type-checker complaint. It's also what let the dead `skipped` slot (L556) go unnoticed. A `@dataclass FieldMigrationOutcome(migrated, notes)` makes the contract explicit.

metadata-ingestion/src/datahub/cli/snowflake_semantic_view_migration.py
```diff
@@ -0,0 +1,1158 @@
...
+    src_urn: str
+    dst_urn: str
+    aspects_copied: List[str] = field(default_factory=list)
+    fields_migrated: List[str] = field(default_factory=list)
+    notes: List[str] = field(default_factory=list)
+    inbound_refs: List[RelatedEntity] = field(default_factory=list)
+    error: Optional[str] = None
+
+
+@dataclass
```
treff7es, line 259: ℹ️ **Type design: `FieldGovernance` conflates two concerns.** `is_metric` is routing/classification (selects metric-URN vs schemaField-URN at L558-569) while `global_tags`/`glossary_terms` are payload. This admits the meaningless state `is_metric=True, global_tags=None, glossary_terms=None`, prevented only downstream by `_field_governance_for_emit` (L394) rather than by the type. And `is_metric` is **dead in the sm→dataset direction** — `collect_semantic_model_field_governance` sets it but `migrate_semantic_model_field_governance` never reads it. A `__post_init__` refusing the all-None payload (or splitting routing from payload) would make the emit-filter unnecessary.

metadata-ingestion/src/datahub/cli/migrate.py
```diff
@@ -742,3 +750,205 @@ def instance2instance(
         keep=keep,
         rest_emitter=graph,
     )
+
+
+def _read_urns_from_file(path: str) -> List[str]:
+    with open(path) as f:
+        return [line.strip() for line in f if line.strip() and not line.startswith("#")]
```
treff7es: ℹ️ **Nit: indented comment lines leak in as bogus URNs.** `line.startswith("#")` tests the *un-stripped* line, so `   # comment` (leading whitespace) is not skipped and `line.strip()` adds `"# comment"` as a URN. Use `line.strip().startswith("#")`. Low severity — it'd be filtered/errored downstream rather than corrupt data.

metadata-ingestion/src/datahub/cli/snowflake_semantic_view_migration.py
kyungsoo-datahub: Fixed. SEMANTIC_VIEW_SUBTYPE now aliases DatasetSubTypes.SEMANTIC_VIEW, and _SYNTHETIC_SUBTYPE_TAG_URNS's uppercase entries now use SemanticViewColumnSubtype.DIMENSION/FACT/METRIC. Lowercase variants stay as literals since the connector enum has no lowercase equivalent.

metadata-ingestion/src/datahub/cli/snowflake_semantic_view_migration.py
kyungsoo-datahub: Pin tests for gen_semantic_model_urn/gen_metric_urn/gen_dataset_urn already existed; added one for _semantic_model_field_path, the one gap. Agreed on the snowflake_identifiers.py extraction — that's post-#18395, not this PR.

metadata-ingestion/src/datahub/cli/snowflake_semantic_view_migration.py
kyungsoo-datahub: Deferring. It's good point but FieldGovernance is constructed in 3 places and consumed by the emit filter, more surface area than the concrete bugs fixed here.

metadata-ingestion/src/datahub/cli/snowflake_semantic_view_migration.py
kyungsoo-datahub: Fixed the symptom (the dead skipped slot now gets populated for real), but kept the Tuple[List[str], List[str]] shape. The @dataclass FieldMigrationOutcome refactor is deferred, touches 3 call sites + tests for a real but non-urgent gain.

metadata-ingestion/src/datahub/cli/snowflake_semantic_view_migration.py
kyungsoo-datahub: Fixed. The field-fan-out resilience patch already populates skipped for real (appends a note on per-field failure), same fix as the earlier dead-slot comment.

metadata-ingestion/src/datahub/cli/snowflake_semantic_view_migration.py
kyungsoo-datahub: Deferring. Real gain (race-safe, deletes ~60 lines) but swaps the whole write path, not a bug fix.

metadata-ingestion/src/datahub/cli/snowflake_semantic_view_migration.py
kyungsoo-datahub: Fixed. Confirmed the union helpers already no-op on a None counterpart in all four prior×field_gov cases, so the ternary was provably redundant. Simplified to unconditional calls with prior_tags/prior_terms locals, same as suggested.

metadata-ingestion/src/datahub/cli/snowflake_semantic_view_migration.py
kyungsoo-datahub: Fixed. run_migration's loop wraps each URN in try/except, converting any failure into an EntityMigrationResult(error=...) so the batch continues. Covers all fallible calls, not just the three named ones. Tested.

metadata-ingestion/src/datahub/cli/snowflake_semantic_view_migration.py
kyungsoo-datahub: Fixed. The report's failed-entity line now prints aspects copied before failure / field tags/terms copied before failure when non-empty, instead of only ERROR: {error}.

metadata-ingestion/src/datahub/cli/snowflake_semantic_view_migration.py
kyungsoo-datahub: Fixed. Each field in migrate_dataset_field_governance is now migrated inside its own try/except, so one field's failure no longer prevents the already-migrated ones from reaching the return value (rather than "assign before it can throw," which isn't possible since the final list isn't known until the loop runs).

metadata-ingestion/src/datahub/cli/migrate.py
kyungsoo-datahub: Fixed. constant promoted to migration_utils.py, both places import it now. Didn't touch get_incoming_relationships's signature or delete _fetch_inbound_refs. It's bigger change, out of scope here.

treff7es (approved):

PR #18474, opened 2026-07-18
feat(ingest/odcs): add ODCS ingestion source UI form

acrylJonny:
> The shared-reader dependency #18523 (`HTTPConnectionConfig` + `read_file_as_bytes` auth threading) has merged, and this branch has been rebased onto `master`.

## Summary

Adds frontend managed-ingestion integration for the **Open Data Contract Standard (ODCS)** source. Before this, ODCS was already in the integrations catalog (`datahub.json`) and had a platform logo, but it was **not selectable in the "Create source" UI builder** — operators had to hand-write a YAML recipe. This makes ODCS a first-class UI-configurable source, matching how the other ~70 connectors work.

- **`sources.json`**: new `odcs` entry (logo + starter recipe + `Data Catalog` category) so ODCS appears in the create-source picker.
- **`builder/constants.ts`**: wires `odcslogo.png` into the platform-logo map.
- **`RecipeForm/odcs.ts` + `RecipeForm/constants.ts`**: structured recipe form — `path` (required) plus headline toggles (emit assertions / schema assertion / logical-parent link, strict validation) and the source-location selector with its per-location credential fields. The remaining options (schema-assertion compatibility, env, verify-physical-urns, replicate-contract-metadata, tag prefix) stay configurable via the YAML editor at their defaults.
- **`conf/odcs/odcs.ts`**: connector id constant.

### Source locations + filtering (added in this branch)

- **Source location selector** — a single-select (Local / S3 / GCS / HTTP / Git) that reveals the credential fields for the chosen location: Amazon S3 (access key ID, secret access key, region), Google Cloud Storage (HMAC key ID + secret), and Git (repo, branch, deploy key). Local and HTTP(S) need no credentials. The selector is form-only and is inferred from the recipe's `path` scheme / `git_info` when editing an existing recipe, so switching from the YAML editor back to the form keeps the right fields open. Reading from more than one location at once is still supported via the YAML editor (the connector accepts a mixed `path` list).
- **`dataset_pattern` filtering** (backend + UI): new `AllowDenyPattern` on the ODCS source config, applied to the composed logical dataset name (`{contract_id}.{schema_name}`). A non-matching `schema[]` entry is skipped along with its assertions and `logicalParent` link and recorded under `report.filtered`. Surfaced in the UI Filter section via the shared dataset allow/deny fields.

## Testing

[...]

PR #18513, opened 2026-07-20
feat(ingest/ui): support multi-select in RecipeForm SELECT fields

acrylJonny:
## Summary

Adds an optional `isMultiSelect` flag to the ingestion `RecipeForm` `SELECT` field so a single form field can hold an array of values instead of one value.

- `RecipeField` gains `isMultiSelect?: boolean`.
- Both SELECT renderers honor it: the legacy `FormField` (passes `mode="multiple"`) and the multi-step builder's `SelectField` (passes `isMultiSelect` to `SimpleSelect` and preserves array values).
- Default behavior is unchanged when the flag is unset, so existing single-select fields are unaffected.

This is a small, reusable framework change split out ahead of the connector form that consumes it (the ODCS ingestion UI form, #18474), per the "shared/framework changes land in their own PR" convention.

## Checklist

- [ ] Tests added/updated (covered by the consuming connector's tests in #18474)
- [ ] Docs added/updated (if applicable)

### DIFF
--- a/datahub-web-react/src/app/ingestV2/source/builder/RecipeForm/FormField.tsx
+++ b/datahub-web-react/src/app/ingestV2/source/builder/RecipeForm/FormField.tsx
@@ -72,7 +72,11 @@ function SelectField({ field, removeMargin }: CommonFieldProps) {
             data-testid={`field-${field.name}`}
         >
             {field.options && (
-                <Select placeholder={field.placeholder} allowClear={!field.required}>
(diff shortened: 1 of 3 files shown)


acrylJonny:
Closing: the ODCS ingestion form was simplified to a single-select Source Location control, so it no longer needs the multi-select RecipeForm framework change. Reverting this framework addition rather than carrying an unused flag.

PR #18515, opened 2026-07-21
fix(ingest/powerbi): preserve downstream column casing for all SQL-parsed lineage

puneetagarwal-datahub:
## Summary

Column-level lineage from a PowerBI report to its warehouse can silently drop the **column** edges while **table**-level lineage connects fine. The downstream (PowerBI-side) `schemaField` URN is emitted with the casing of the SQL alias rather than PowerBI's real field name, so it points at a non-existent field and the arrow never renders.

`_remap_column_lineage_to_pbi_fields` already exists to correct this — it maps the parsed downstream column back to PowerBI's actual field name (case-insensitive match, exact-case output). But it was wired into **only** the Oracle lineage path. This PR moves it into the shared `AbstractLineage.parse_custom_sql`, so every SQL-parsing path benefits, and removes the now-redundant Oracle-only wrapper.

## When this bug occurs

All of these together:

1. The PowerBI report resolves lineage through a **SQL-parsing path** — the native-query path (`native_query_parsing: true` + `enable_advance_lineage_sql_construct: true`) or ODBC. (The Oracle path was already covered; the direct-table path was never affected.)
2. The report's SQL **aliases its output columns in a casing that differs from the PowerBI field**, e.g. `SELECT c AS customerid` while PowerBI names the field `CustomerId`. Because the warehouse (SQL Server, etc.) is case-insensitive, the query runs fine regardless, so the alias casing is effectively arbitrary and easy to get "wrong".
3. Column-level lineage is enabled (`extract_column_level_lineage: true`).

When those line up, DataHub — which matches `schemaField` URNs case-sensitively — cannot resolve the downstream field, and the column arrow is dropped even though the dataset (table) URNs match.

## How this was found

- Symptom: table-level lineage rendered, column-level did not — which localizes the break to the field layer (dataset URNs already match).
- The ingestion log showed CLL was generated (no parse errors) but the parsed downstream column came through lowercased (`customerid`) while the PowerBI field was `CustomerId`.
- Tracing the code, the native-query path (`NativeQueryLineage` → `parse_custom_sql`) returned the parsed lineage without the downstream remap that the Oracle path applied.

## Relationship to prior work

This is the downstream counterpart to the upstream column-casing fix (#18181). #18181 stopped lowercasing the **upstream** column; this preserves the **downstream** (PowerBI-side) column so both ends of the edge resolve.

## Testing

- **Integration** (`tests/integration/powerbi/test_m_parser.py`): new test on the native-query path asserting the downstream column is remapped to the PowerBI field's casing (`customerid` alias → `CustomerId` field). Verified it fails before the change and passes after.
- **Unit** (`tests/unit/powerbi/test_pattern_handler.py`): because the remap moved out of `OracleLineage` into the shared `parse_custom_sql`, the Oracle unit test (which mocks `parse_custom_sql`) was rescoped to its remaining responsibility — threading `default_schema`/`database`/`platform_detail` and delegating — and a new focused unit test asserts `parse_custom_sql` itself remaps the downstream column to the PowerBI field casing (keeps `testQuick` + codecov patch coverage on the moved logic).
- Full `tests/unit/powerbi/` and `tests/integration/powerbi/test_m_parser.py` suites pass (305 passed, 1 xfail).
- `./gradlew :metadata-ingestion:lint` (ruff + mypy) clean.

## Checklist

- [x] Tests added

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/pattern_handler.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/pattern_handler.py
@@ -461,10 +461,9 @@ class AbstractLineage(ABC):
         return Lineage(
             upstreams=dataplatform_tables,
             # sqlglot returns downstream columns in the SQL's alias casing, which
-            # rarely matches the casing PowerBI stores its fields in. Remap here —
-            # in the shared SQL-parsing path — so every platform's native-query and
-            # ODBC lineage gets the fix, not just the Oracle path that used to wrap
-            # this call.
+            # rarely matches the casing PowerBI stores its fields in. Remap in this
+            # shared SQL-parsing path so the downstream column resolves to the real
+            # PowerBI field regardless of the platform driving the parse.
             column_lineage=_remap_column_lineage_to_pbi_fields(
                 (
                     parsed_result.column_lineage
@@ -903,8 +902,6 @@ class OracleLineage(AbstractLineage):
 
         # `default_database` is None for the default 2-part URN shape; set, it
         # produces 3-part URNs matching `add_database_name_to_urn: true`.
-        # parse_custom_sql already remaps downstream columns to the PowerBI field
-        # casing, so no additional wrapping is needed here.
         return self.parse_custom_sql(
             query=query,
             server=server,
--- a/metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/pattern_handler.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/pattern_handler.py
@@ -188,10 +188,12 @@ def _remap_column_lineage_to_pbi_fields(
     column_lineage: List[ColumnLineageInfo],
     pbi_columns: Optional[List[Column]],
 ) -> List[ColumnLineageInfo]:
-    """sqlglot returns downstream column names in the upstream's case (Oracle is
-    lowercase), but PowerBI fields keep their original casing in the API
-    response. Without this remap, the downstream schemaField URN does not
-    resolve and the column-level edge points to a non-existent field."""
+    """sqlglot returns downstream column names in the parsed SQL's casing (driven
+    by the query's aliases and the source dialect's identifier folding), but
+    PowerBI fields keep their original casing from the API response. Without this
+    remap the downstream schemaField URN does not resolve and the column-level
+    edge points to a non-existent field. Applied for every SQL-parsing path via
+    parse_custom_sql (native-query, ODBC, and the two/three-step patterns)."""
     if not column_lineage or not pbi_columns:
         return column_lineage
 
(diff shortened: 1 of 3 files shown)


metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/pattern_handler.py
```diff
@@ -458,10 +460,18 @@ def parse_custom_sql(
...
         return Lineage(
             upstreams=dataplatform_tables,
-            column_lineage=(
-                parsed_result.column_lineage
-                if parsed_result.column_lineage is not None
-                else []
+            # sqlglot returns downstream columns in the SQL's alias casing, which
+            # rarely matches the casing PowerBI stores its fields in. Remap here —
+            # in the shared SQL-parsing path — so every platform's native-query and
+            # ODBC lineage gets the fix, not just the Oracle path that used to wrap
```
treff7es: This comment is pr specific and not future proof. Can you rephrase or drop it?

metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/pattern_handler.py
```diff
@@ -893,20 +903,15 @@ def _create_lineage_from_query(self, server: str, query: str) -> Lineage:
 
         # `default_database` is None for the default 2-part URN shape; set, it
         # produces 3-part URNs matching `add_database_name_to_urn: true`.
-        lineage = self.parse_custom_sql(
+        # parse_custom_sql already remaps downstream columns to the PowerBI field
```
treff7es: Does this comment make sense for the future?

metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/pattern_handler.py
```diff
@@ -893,20 +903,15 @@ def _create_lineage_from_query(self, server: str, query: str) -> Lineage:
...
+        # casing, so no additional wrapping is needed here.
```
treff7es:
```suggestion
```

metadata-ingestion/src/datahub/ingestion/source/powerbi/m_query/pattern_handler.py
```diff
@@ -458,10 +460,18 @@ def parse_custom_sql(
...
             upstreams=dataplatform_tables,
-            column_lineage=(
-                parsed_result.column_lineage
-                if parsed_result.column_lineage is not None
-                else []
+            # sqlglot returns downstream columns in the SQL's alias casing, which
+            # rarely matches the casing PowerBI stores its fields in. Remap here —
+            # in the shared SQL-parsing path — so every platform's native-query and
+            # ODBC lineage gets the fix, not just the Oracle path that used to wrap
+            # this call.
```
treff7es:
```suggestion
```

treff7es (approved):

PR #18517, opened 2026-07-21
fix(logical-models): make logical-parent field-path validator lenient and gate-able

max-datahub:
## Summary

Follow-up to #18498. Makes `LogicalParentFieldPathValidator` — which enforces that a **column-level** `logicalParent` edge references field paths that exist on both the child and parent schemas — safer and controllable.

Two problems with the validator as merged:

1. **It rejected writes when a referenced schema wasn't loaded.** `readFieldPaths` returned an empty set both when the `schemaMetadata` aspect was genuinely absent and when it existed but had no fields, so a column mapping written **before** the dataset's schema is ingested (a legitimate out-of-order write) would fail with "Field path not found".
2. **It had no off-switch**, unlike the sibling `LogicalParentPlatformValidator`.

## Changes

- **Leniency:** skip validation when the target `schemaMetadata` aspect is absent — a not-yet-loaded schema is indistinguishable from a genuinely missing field, so we don't reject. A real mismatch is still caught once the schema exists; a schema that is present but has no fields still validates (and rejects, since the field truly isn't there). Implemented by returning `Optional.empty()` (absent) vs `Optional.of(fieldPaths)` (present).
- **Gate flag:** `metadataChangeProposal.validation.logicalParent.fieldPathValidation.enabled` (env `MCP_VALIDATION_LOGICAL_PARENT_FIELD_PATH`), **default `true`**. Unlike the platform validator — which is default-off because it constrains where *pre-existing* logical models may live and could reject valid setups on upgrade — this one only affects *new* column-level writes and never re-validates stored data, so it stays on by default while still being disable-able.

## Scope note

This validator only fires on **column-level** (schemaField) `logicalParent` writes. Dataset-level links are untouched — notably, the ODCS ingestion source emits only dataset-level `logicalParent` edges, so it is unaffected by this validator either way.

## Testing

- `LogicalParentFieldPathValidatorTest`: added `testSkipsWhenSchemaNotYetIngested` (absent schema → no exception); existing cases still pass (both fields present → ok; child/parent field missing with schema present → rejected; unlink ignored). 5/5 green.
- `:metadata-io` and `:metadata-service:factories` compile.

## Checklist

- [x] Tests added/updated
- [x] No breaking changes


max-datahub:
Refreshed onto current `master` and re-verified. Notes for review, including one thing that differs from what we agreed in chat.

### The flag defaults to ON, deliberately

We both described this as "put it behind a flag", which reads as default-off. It ships **default-on**: `matchIfMissing = true` and `MCP_VALIDATION_LOGICAL_PARENT_FIELD_PATH:true`.

That's intentional. A default-off flag makes the validator a no-op out of the box and lets dangling column-level edges silently reappear, which defeats the reason the check was relocated to the aspect layer in the first place. The asymmetry with `LogicalParentPlatformValidator` (default-off) holds up: that one constrains where *pre-existing* logical models may live and can reject valid setups on upgrade, whereas this one only ever sees new column-level writes and never re-validates stored data. Happy to flip it if you'd rather have strict parity, but flagging it explicitly so it isn't a surprise.

### The rejection this fixes is reproducible, not theoretical

On a local quickstart, writing a column-level `logicalParent` before the parent's `schemaMetadata` exists:

```
Field path 'col_a' not found on parent dataset urn:li:dataset:(urn:li:dataPlatform:logical,tmp.parent_noschema,PROD)
```

Ingesting the parent schema and replaying the **identical** write returns 200. So it's purely a write-ordering artifact, and the message names a field that isn't actually missing — which is the part that turns into a support ticket.

### ODCS is confirmed outside this validator

`odcs_source.py` emits `odcs_to_logical_parent_mcp(physical_urn, binding.logical_urn)` — both dataset URNs. There is no schemaField-level `logicalParent` emission anywhere in the Python source tree, so the main automated producer of logical links is unaffected either way.

### Tradeoff worth naming: skipped edges are never re-validated

Skipping when the schema is absent means the edge is stored unvalidated permanently. If the schema arrives later and the field genuinely doesn't exist, nothing re-checks that stored edge — "a real mismatch is still caught once the schema exists" applies to *subsequent* writes, not the one already written. Recording it as an accepted tradeoff rather than leaving it to be discovered.

### The branch merge did not compile, now fixed

Worth knowing since the previous green CI was a month stale. `master`'s `PatchOperationUtils` consolidation (#18579) added `validatePatchItem` and reshaped `validateLogicalParent`, both declaring the old `Map<Urn, Set<String>>` cache type. Those hunks don't overlap this branch's, so the merge was textually clean but broke the build with four `incompatible types` errors. `1bd8ed204ab` threads `Map<Urn, Optional<Set<String>>>` through both signatures.

Verified locally after the fix: `LogicalParentFieldPathValidatorTest` + `LogicalParentPlatformValidatorTest` 14 passing, `:metadata-service:factories:compileJava` clean, `spotlessCheck` clean on both modules.

> - Claude (on behalf of Max)

PR #18518, opened 2026-07-21
fix(ingest/redshift): make user-info join enrichment-only, exclude rdsdb by id

treff7es:
## Summary

The Redshift usage, operation-aspect and insert-lineage queries join the query
log tables to the user-info views (`svl_user_info` / `SVV_USER_INFO`) to resolve
a username. Those views have no row for the internal `rdsdb` system user, so the
INNER join (and the name-based `user_name <> 'rdsdb'` predicate) doubled as an
**implicit** rdsdb filter — and would also silently drop any *real* user the view
cannot resolve (e.g. on Redshift editions that keep these views superuser/self-only
when ingesting as a non-superuser).

This makes the intent explicit and the extraction resilient:

- **Exclude `rdsdb` by its stable system id (`userid = 1`)**, mirroring the existing
  `stl_scan_based_lineage_query`, instead of relying on the join to drop it.
- **LEFT JOIN the user-info views for username enrichment only** — a log row is never
  gated on whether its user resolves.
- **Allow an unresolved username end to end**: `RedshiftAccessEvent.username` is now
  `Optional`, usage falls back to the existing `unknown` user, and operation aspects
  are emitted with `actor=None` (`actor` is optional on `OperationClass`) rather than
  being dropped.

## Behavior

Behavior-preserving on standard clusters — verified against a live Redshift instance:
the revised `usage_query` returns the same row count as the previous INNER form, and
all reworked queries execute as valid SQL. On clusters where the user-info views are
restricted, usage / lineage / operations are now retained (attributed to `unknown` /
no actor) instead of being silently lost.

## Checklist

- [x] Tests added/updated (query-shape tests for LEFT joins + explicit `userid <> 1`; behavioral tests for unresolved username and actor-less operations)
- [x] Docs updated (`redshift_post.md` "System table access" note)


sgomezvillamor (approved):

metadata-ingestion/src/datahub/ingestion/source/redshift/query.py
puneetagarwal-datahub, line 811: stale comments here for non-LEFT joins

puneetagarwal-datahub, line 609: do we need to replace this filter here with  " sq.userid <> 1" to keep consistent

puneetagarwal-datahub, line 705: same here si.userid <> 1

puneetagarwal-datahub:
I think PR is  missing the two provisioned copies — stl_scan_based_lineage_query and list_insert_create_queries_sql — still gate rows on usename <> 'rdsdb', so provisioned clusters keep dropping lineage for unresolved users.

PR #18519, opened 2026-07-21
fix(ingest/bigquery): sample large tables and partitions correctly in the SQLAlchemy profiler

sgomezvillamor:
## Overview

Follow-up to #18253 (which fixed Snowflake SQLAlchemy-profiler sampling), for **BigQuery**.

The GE profiler samples BigQuery via a "temp table that isn't a real temp table": it runs `SELECT * FROM t TABLESAMPLE SYSTEM (pc percent)` and reuses BigQuery's cached-results anonymous table (`query_job.destination`) as the profiling target — no `CREATE TABLE` permission, 24h auto-expire (`create_bigquery_temp_table` + `update_dataset_batch_use_sampling` in `ge_data_profiler.py`).

Unlike Snowflake, the BigQuery SQLAlchemy adapter *already* replicated this cached-results sampling. But it had three defects vs. the GE profiler, fixed here.

## What changed

**`sqlalchemy_profiler/adapters/bigquery.py`**
- **Connection leak** — `_create_temp_table_for_query` acquired a pooled raw DBAPI connection and never closed it, leaking one per profiled table and eventually exhausting the engine pool (shared engine, default QueuePool, multi-worker). It's now closed in a `finally` on every path (success / `RuntimeError` / `DatabaseError`), mirroring GE's `create_bigquery_temp_table`.
- **Redundant `COUNT(*)`** — the sampling decision queried the row count twice and never used the count the crawl already collected. The count is now resolved once via `context.row_count or self._get_quick_row_count(...)`; on the whole-table path it usually issues no `COUNT(*)` at all. (On BigQuery `SELECT COUNT(*)` is a metadata-only, 0-bytes-billed query anyway.)
- **Partitioned tables were never sampled** — sampling bailed whenever `custom_sql` was set, so a large partition was profiled in full and not labelled `SAMPLE`. It now samples the materialized partition temp table (matching GE, which samples on top of the partition), measuring the partition's own row count and setting `context.is_sampled` so the existing `_profile_row_count` `" SAMPLE (sample rows N)"` labelling applies.
- Row-count failures now surface via `report.warning(...)` instead of a silent `logger.debug`.

**`bigquery_v2/profiler.py`**
- `get_batch_kwargs` now passes `row_count=table.rows_count` so the SQLAlchemy profiler can skip the `COUNT(*)`. Safe for the GE path — its `_generate_single_profile` has `**kwargs` and ignores it.

## Notes for reviewers

- Unlike Snowflake (where #18253 gated `custom_sql` to the GE path), BigQuery's partition `custom_sql` feeds **both** the GE and SQLAlchemy engines, so its shape must stay valid for both until GE is removed. Documented in-code; once GE is gone the partition decision should move into the adapter.
- `TABLESAMPLE` must target a table (not a subquery), which is why a partition is sampled via its materialized temp table rather than by wrapping the partition SQL.
- No config/API changes, no new dependencies; uses existing `use_sampling` / `sample_size`.

## Testing

- `./gradlew :metadata-ingestion:lint` — ruff + `ruff format --check` + mypy all clean.
- `pytest tests/unit/sqlalchemy_profiler/test_adapters.py tests/unit/bigquery/test_bigquery_profiler.py` — 123 passed. New tests cover: raw connection closed (incl. error path), partition temp-table sampling, prefer-metadata vs `COUNT(*)` fallback, and `get_batch_kwargs` row_count.

## Checklist

- [x] Tests added/updated
- [ ] Docs — n/a (no user-facing config/behavior change)
- [ ] Breaking changes — none


metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/bigquery.py
```diff
@@ -216,85 +216,79 @@ def _create_temp_table_for_query(
...
-
-        # Check if table is large enough to warrant sampling
-        # Get quick row count estimate
-        row_count = self._get_quick_row_count(context, conn)
-        if row_count is None:
-            return False
-
-        return row_count > self.config.sample_size
+        finally:
+            raw_conn.close()
```
treff7es: The leak fix is right (`raw_conn` is bound before the `try`), but this `close()` is unguarded. Every meaningful error exit — the no-destination `RuntimeError` (:200), the bare `raise` (:215), and `raise ... from e` (:218) — passes through here. If the BigQuery DBAPI `close()` throws during teardown, it replaces the in-flight error as the primary exception and demotes the actionable message to a chained note. Suggest:

```python
finally:
    try:
        raw_conn.close()
    except Exception as close_err:
        logger.debug(f"Failed to close raw connection after temp-table creation: {close_err}")
```

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/bigquery.py
```diff
@@ -216,85 +216,79 @@ def _create_temp_table_for_query(
...
-                query: Any = sa.select([sa.func.count()]).select_from(context.sql_table)
-            else:
-                # Build quick count query using query builder
-                if not context.table:
-                    return None
-                # Create SQLAlchemy Table object for query construction
+            if context.temp_table:
+                # A partition/query temp table exists — count it, not the
+                # original table (its size differs from the full table).
+                table_obj: Any = sa.Table(
```
treff7es: `Any` isn't needed here — all three branches assign `sa.Table` (and `context.sql_table` is already `Optional[sa.Table]`, narrowed by the `is not None` guard). Before this PR the variable was inferred as `sa.Table`, so annotating it `Any` is a mild regression against the repo's "avoid `Any`" rule. `table_obj: sa.Table` type-checks clean (it feeds `select_from`, which takes a `FromClause`).

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/bigquery.py
```diff
@@ -303,42 +297,42 @@ def _setup_sampling(
...
+            )
+            row_count = self._get_quick_row_count(context, conn)
+        elif context.table:
+            # Prefer the row count the BigQuery source already knows
+            # (INFORMATION_SCHEMA / __TABLES__); COUNT(*) only when it is
+            # missing (e.g. views, external tables).
+            target = self.quote_identifier(
+                f"{context.schema}.{context.table}" if context.schema else context.table
+            )
+            row_count = context.row_count or self._get_quick_row_count(context, conn)
```
treff7es: Minor: `context.row_count or ...` treats a legitimate `0` as falsy, so an empty table still issues a `COUNT(*)` — harmless (0-row count is free and it short-circuits at :321) but slightly undercuts the "avoid redundant COUNT(*)" goal. `context.row_count if context.row_count is not None else ...` makes it exact.

metadata-ingestion/tests/unit/sqlalchemy_profiler/test_adapters.py
```diff
@@ -1234,6 +1237,123 @@ def mock_create_temp(ctx):
                 assert len(captured_sql) > 0
                 assert "TABLESAMPLE" in captured_sql[0].upper()
 
+    def test_create_temp_table_closes_connection_on_error(
+        self, adapter, mock_bigquery_engine
+    ):
+        """The pooled raw connection is closed even when the query fails."""
+        from google.cloud.bigquery.dbapi import exceptions as bq_exceptions
```
treff7es: Inline import — the project treats late imports in test bodies as a hard rule. The optional-dep waiver doesn't apply: `bq_exceptions` is already imported unconditionally at the top of the adapter under test, and `BigQueryAdapter` is imported at the top of this file, so the module loads at collection time regardless. Please move this to the top-level imports.

treff7es (approved):

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/bigquery.py
sgomezvillamor: Good catch — done in 28ae62a3. Wrapped the `close()` in try/except and log at debug, so a teardown failure can't clobber the in-flight `RuntimeError`/`DatabaseError`.

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/bigquery.py
sgomezvillamor: Agreed — fixed in c7c89360. Annotated `table_obj: sa.Table` (mypy is clean; it feeds `select_from`).

metadata-ingestion/src/datahub/ingestion/source/sqlalchemy_profiler/adapters/bigquery.py
sgomezvillamor: Fixed in ba5b871c. Switched to `context.row_count if context.row_count is not None else ...` so a legitimate `0` is honored. Same pattern existed in the Snowflake adapter (`setup_profiling`, guarding the INFORMATION_SCHEMA lookup) — fixed there too in the same commit.

metadata-ingestion/tests/unit/sqlalchemy_profiler/test_adapters.py
sgomezvillamor: Done in 1b9d3dfa — moved `bq_exceptions` to the top-level imports.

PR #18522, opened 2026-07-21
fix(access): update background color of request access button

v-tarasevich-blitz-brain:
Before
image

After
image


v-tarasevich-blitz-brain:
I see that another PR was merged related to this problem - https://github.com/datahub-project/datahub/pull/18364

chriscollins3456:
yeah i think we can close this with the above PR referenced fixing this already

PR #18523, opened 2026-07-21
feat(ingest): authenticated http(s) reads in shared object-store file reader

acrylJonny:
## Summary

- Adds `HTTPConnectionConfig` (shared, in `ingestion/source/common/`) supporting a **bearer token** or **HTTP basic auth** (mutually exclusive, validated) plus a `verify_ssl` toggle.
- Threads an optional `http_connection` through the shared `read_file_as_bytes` helper so `http(s)://` file sources can authenticate. Existing callers (dbt) are unaffected — the parameter defaults to `None`.
- Only bearer + basic are exposed (not arbitrary headers): `requests` strips the `Authorization` header on cross-host redirects, so bearer/basic are not leaked to a redirected origin, whereas custom-header tokens would be.
- Blank credentials (`""` / whitespace-only token, username, or password) normalize to `None`, so an empty field never emits a `Bearer ` header or an empty `("", "")` basic-auth tuple.

This is the shared-framework half of the ODCS authenticated-HTTP work, split out per the repo's "shared changes in their own PR" convention. The consumer (ODCS source + ingestion UI fields) is in #18474, which is stacked on this branch.

## Review notes

- **`verify_ssl=False` warning lives in the consumer.** `read_file_as_bytes` is a pure util with no report handle, so it can't warn. The `self.report.warning` when `http_connection.verify_ssl` is `False` — and its test — are in the ODCS source in the consumer PR #18474, where a report is available.
- **This capability is added ahead of its caller, by design.** `read_file_as_bytes(..., http_connection=)` is wired up by the ODCS config + UI form in #18474 (stacked on this branch). It is staged, not dead code.
- **Blank-string auth tightened.** A `field_validator` collapses empty/whitespace-only `token`/`username`/`password` to `None`, covered by `test_http_connection_config_blank_strings_are_unset`.

## Test plan

- [x] `pytest tests/unit/test_object_store_files.py` — bearer header, basic auth + `verify=False` passthrough, blank-string normalization, and both validator rejections (bearer+basic together; basic missing a half).
- [x] `./gradlew :metadata-ingestion:lint` (ruff + mypy) clean.

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/common/object_store_files.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/common/object_store_files.py
@@ -11,6 +11,7 @@ import requests
 from datahub.ingestion.source.aws.aws_common import AwsConnectionConfig
 from datahub.ingestion.source.aws.s3_util import is_s3_uri
 from datahub.ingestion.source.common.gcs_connection_config import GCSConnectionConfig
+from datahub.ingestion.source.common.http_connection_config import HTTPConnectionConfig
 from datahub.ingestion.source.gcs.gcs_utils import is_gcs_uri
 
 logger: logging.Logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


AdrianMachado:
This looks good to me, but would like someone from Ingestion team to also review as this is now a shared resource

kyungsoo-datahub (approved):

metadata-ingestion/src/datahub/ingestion/source/common/object_store_files.py
```diff
@@ -72,18 +73,41 @@ def _read_object_store_body(
     return data
 
 
+def _http_request_kwargs(http_connection: Optional[HTTPConnectionConfig]) -> dict:
```
sgomezvillamor: may this be a method in the HttpConnectionConfig class itself?

or is there anything specific here to object store files? if so, such a constraint should be documented

sgomezvillamor (approved):

metadata-ingestion/src/datahub/ingestion/source/common/object_store_files.py
acrylJonny: Good call — there's nothing object-store-specific about it, it's just translating the config into `requests` kwargs. Moved it onto the config as `HTTPConnectionConfig.to_request_kwargs()` in 920e2e9, so the auth logic now lives next to the fields it reads and is reusable by any http(s) caller. The call site just guards the `None` case: `**(http_connection.to_request_kwargs() if http_connection else {})`. No behavior change.

PR #18527, opened 2026-07-21
perf(graphql): batch Domain.entities count-only resolution to eliminate N+1

gdatahub:
## Summary

`Domain.entities` is a GraphQL field resolver invoked once per `Domain`, and the frontend `domainEntitiesFields` fragment selects it three times per domain (`entities` / `dataProducts` / `applicationsInDomain`). Listing or searching **N** domains therefore fired **3N** `searchAcrossEntities` calls — a classic N+1 that also multiplied the DB reads each search carries.

This PR collapses the count-only hot path into a fixed number of aggregation-only searches, growing in chunk-sized steps rather than per-domain.

## Changes

- **`DomainEntityCountsBatchLoader`** (new) — a request-scoped `DataLoader` keyed by `(domain, entity-type constraint)`. It groups requests by constraint and answers each group with one aggregation-only search per chunk of domains (`MAX_DOMAINS_PER_AGG = 25`), faceted on the `domains` field (URN bucket keys) with the entity-type constraint applied as a **search filter**, reusing the existing resolver filter semantics. graphql-java coalesces every per-domain `.load(...)` in an execution tick into one `batchLoad`.
- **`DomainEntitiesResolver`** — routes only the fully-safe count-only fast path through the loader: `count = 0`, default query, no resolved View, no caller `searchFlags`, and at most one `_entityType` filter with the default (EQUAL) condition and an explicit (non-null) value list. Everything else — hit-returning pagination, custom queries, the `listRecommendations` View context, `searchFlags`, non-default conditions, or a value-less filter — keeps the original per-domain search (`resolveDirect`), which handles those exactly as before. Nothing that reads `searchResults` can regress.
- Wired the loader into `GmsGraphQLEngine`.

The entity-type constraint is expressed as a search filter rather than an aggregation dimension, so the counts do not depend on the entity-type token format the search layer emits for the raw aggregation.

## Testing

**Unit** — `DomainEntityCountsBatchLoaderTest` (grouping-by-constraint with independent per-constraint attribution, per-domain distribution, ordering/zeroing, off-chunk isolation, failure handling, chunking) and `DomainEntitiesResolverTest` (fast-path routing vs. direct-search fallback, key construction, and fallback for the value-less `_entityType` filter and for `searchFlags`). Full `:datahub-graphql-core:test` suite green.

**Live end-to-end** — locally seeded instance of 42 domains (each with 12–18 datasets + 3 data products), querying `listDomains` with the `domainEntitiesFields` count aliases. "Before" is the per-domain path (`resolveDirect`), "after" is the batched loader, measured on the same build/data. ES search count via the OpenSearch `search.query_total` delta; DB via the MySQL `Com_select` delta.

| Metric | Before (per-domain N+1) | After (batched) | Improvement |
| --- | --- | --- | --- |
| ES searches / request | 3655 | 175 | ~21x fewer |
| DB SELECTs / request | 388 | 28 | ~14x fewer |
| Mean latency | 287 ms | 83 ms | ~3.5x faster |
| p95 latency | 331 ms | 121 ms | ~2.7x faster |

Page-size sweep on the batched build shows ES cost is **flat within a chunk and steps at the 25-domain boundary** (≤25 domains → 3 constraints × 1 chunk; 26–50 → 3 constraints × 2 chunks); 30→42 domains adds zero extra searches, versus the old path's linear 3N.

**Functional correctness — validated per call.** Across page sizes 10/25/30/42 × 20 iterations, every returned domain's `(entities, dataProducts, applications)` triple matched seeded ground truth exactly and every call was deterministic (0 failures across 2100 per-call assertions). The before/after diff confirmed per-domain counts are identical to the previous resolver for all 42 domains, and a value-less `_entityType` filter returns cleanly (no crash) via the direct-path fallback.

## Checklist

- [x] Tests added/updated
- [ ] Docs — n/a (internal perf change, no API/behavior change)
- [ ] Breaking changes — none

### DIFF
--- a/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/GmsGraphQLEngine.java
+++ b/datahub-graphql-core/src/main/java/com/linkedin/datahub/graphql/GmsGraphQLEngine.java
@@ -25,6 +25,7 @@ import com.linkedin.datahub.graphql.analytics.service.AnalyticsService;
 import com.linkedin.datahub.graphql.concurrency.GraphQLConcurrencyUtils;
 import com.linkedin.datahub.graphql.featureflags.FeatureFlags;
 import com.linkedin.datahub.graphql.generated.*;
+import com.linkedin.datahub.graphql.loaders.DomainEntityCountsBatchLoader;
 import com.linkedin.datahub.graphql.plugins.SemanticSearchPlugin;
 import com.linkedin.datahub.graphql.resolvers.MeResolver;
 import com.linkedin.datahub.graphql.resolvers.ResolverUtils;
@@ -946,6 +947,9 @@ public class GmsGraphQLEngine {
     builder
         .addDataLoaders(loaderSuppliers(loadableTypes))
         .addDataLoader("Aspect", context -> createDataLoader(aspectType, context))
+        .addDataLoader(
+            DomainEntityCountsBatchLoader.LOADER_NAME,
+            context -> DomainEntityCountsBatchLoader.create(entityClient, context))
         .addDataLoader(
             WeaklyTypedAspectsResolver.LOADER_NAME,
             context ->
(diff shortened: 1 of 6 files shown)


gdatahub:
Screenshot 2026-07-25 at 2 56 07 PM

david-leifker (approved):
Rebase onto master; call SearchUtils.getSearchEntityNames(opContext) in both paths.

PR #18530, opened 2026-07-21
fix(gms): repair non-binary metadata_aspect_v2 collation in SqlSetup

naeemdh:
## Summary

`ASYNC_BATCH` managed ingestion can finish `SUCCESS` (sink reports records written, `failures: []`) while the entities **never materialize** in GMS — intermittent, silent data loss. Re-running the identical payload with `mode: SYNC` materializes everything.

### Root cause

When `metadata_aspect_v2` is created with a **case-insensitive collation** (e.g. `utf8mb4_general_ci`, `utf8mb4_0900_ai_ci`) instead of `utf8mb4_bin`, the batch read-back in `EbeanAspectDao.getNextVersions` (`WHERE urn IN (...)`) returns rows for URNs that differ only in case from the request keys. `mergeNextVersionsFromDb` then can't find those keys in the request-keyed map, throws `IllegalStateException: URN or aspect from database did not match request keys`, and `BatchMetadataChangeProposalsProcessor.processBatch` dead-letters the **entire batch** to the FMCP topic — *after* the async sink already ack'd at Kafka-produce time. `SYNC` works because it skips that batch read-back.

Observed signature (`datahub-mce-consumer`):

```
ERROR com.linkedin.metadata.kafka.batch.BatchMetadataChangeProposalsProcessor - MCP Processor Error
java.lang.IllegalStateException: URN or aspect from database did not match request keys ...
```

### Why #16542 didn't cover it

#16542 aligned the `SqlSetup` DDL to `CHARACTER SET utf8mb4 COLLATE utf8mb4_bin`, but that only applies to **freshly created** tables. `CREATE TABLE IF NOT EXISTS` is a no-op on an existing table, so a database already provisioned with the wrong collation is never corrected, and nothing detects it.

### Fix

Add `DatabaseOperations.ensureAspectTableCollation()`, mirroring the existing `ensureAspectIndexes()` precedent that aligns already-existing tables during `SqlSetup`:

- **MySQL:** detect a non-`utf8mb4_bin` collation on the `urn`/`aspect` key columns via `information_schema`, and only when wrong, self-heal with `ALTER TABLE metadata_aspect_v2 CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_bin`, logging a `WARN` about the potential table rebuild.
- **PostgreSQL:** default no-op (its default collation is byte-exact).
- Wired into `CreateTablesStep` right after `ensureAspectIndexes`.

### Operational notes

- Existing misprovisioned databases are repaired by running `datahub-upgrade SqlSetup`; GMS boot alone does not trigger it.
- The collation repair does **not** recover already-dropped MCPs — affected runs must be re-ingested (or replayed from the FMCP topic) after the `ALTER`.

### Out of scope

The batch processor dead-lettering the *whole* batch when a single key fails (co-batched good MCPs become collateral damage) is a separate, riskier change and is left for a follow-up.

## Testing

- `./gradlew :datahub-upgrade:test --tests "*DatabaseOperationsTest"` — 30 tests pass, incl. two new (`...ConvertsWhenWrong`, `...SkipsWhenCorrect`).
- `./gradlew :datahub-upgrade:spotlessJavaCheck` passes.

### DIFF
--- a/datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/sqlsetup/CreateTablesStep.java
+++ b/datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/sqlsetup/CreateTablesStep.java
@@ -96,6 +96,9 @@ public class CreateTablesStep implements UpgradeStep {
     try (Connection connection = server.dataSource().getConnection()) {
       dbOps.ensureAspectIndexes(connection);
     }
+    try (Connection connection = server.dataSource().getConnection()) {
+      dbOps.ensureAspectTableCollation(connection);
+    }
 
     if (args.createSchemaVersionIndex()) {
       try (Connection connection = server.dataSource().getConnection()) {
--- a/datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/sqlsetup/DatabaseOperations.java
+++ b/datahub-upgrade/src/main/java/com/linkedin/datahub/upgrade/sqlsetup/DatabaseOperations.java
@@ -135,6 +135,22 @@ public interface DatabaseOperations {
   default void ensureAspectIndexes(Connection connection) throws SQLException {}
 
   /**
+   * Ensures the {@code metadata_aspect_v2} key columns use a case- and byte-exact collation. On
+   * MySQL, {@code CREATE TABLE} sets {@code utf8mb4_bin} but that only applies to freshly created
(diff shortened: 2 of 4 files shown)


naeemdh:
Superseded by https://github.com/datahub-project/datahub/pull/18531 — same change, opened from a branch on datahub-project directly instead of a fork.

PR #18535, opened 2026-07-21
fix(zdu): handle missing index case

RyanHolstien:
Handles scenarios where a partial reindex occurs and the index noted by the reindex state has been removed by some outside process.


david-leifker (approved):
Have you tested locally? At least one run where you manually delete the index at the right time?

PR #18539, opened 2026-07-22
fix(ingest/dbt): emit platform-instance aspects for target siblings

javabrett:
## Problem

When a dbt source has `target_platform_instance` set, it emits sibling and lineage MCPs that reference target-platform (warehouse) dataset URNs. For any table whose URN does not yet exist at the time of the dbt run, those references cause the server to auto-create a "stub" dataset entity with default-generated aspects:

- a `dataPlatformInstance` aspect containing the platform only, with no `instance` field, and
- a `browsePathsV2` derived from URN name segments. Because the platform instance is embedded in the URN name (`myinstance.db.schema.table`), the plain instance name becomes the browse root.

The result is a duplicate navigation folder in the Browse sidebar (a plain-name folder alongside the URN-form folder for the same instance) and entities that are invisible to platform-instance facets and filters. The defaults are generated once at entity creation and the dbt source never emits these aspects. For tables the warehouse connector ingests, its next run overwrites the defaults, so the damage is transient; for tables it never ingests (dropped or one-off tables, schemas/shares outside the warehouse recipe scope, test/temp tables) nothing corrects the stubs and they accumulate. A platform-instance URN migration (adding `platform_instance` to existing recipes) creates these stubs en masse, since every URN in the dbt manifest becomes newly-created at once.

Server-side #17263 fixes the browse root for newly-created entities when the platform instance entity already exists, but does not address the instance-less `dataPlatformInstance` aspect, entities created before that fix, or servers without it.

## Fix

When `target_platform_instance` is configured (and the new `emit_target_platform_instance_aspects` config is left enabled), `create_target_platform_mces` now emits for each target-platform entity:

1. A `dataPlatformInstance` aspect carrying the full instance URN. This is value-identical to what the warehouse connector writes for the same entity, so the upsert is a no-op for entities the warehouse connector owns and a fix for sibling-only stubs - including healing existing stubs on the next dbt run.
2. A `browsePathsV2` rooted at the platform-instance URN with database/schema segments, guarded so it never overwrites a container-based browse path written by the warehouse connector: the write only happens when a DataHub graph connection is available and the entity's existing browse path is missing or is a plain name-derived server default. This also heals previously-created stubs in place on re-ingest.

Both emissions are marked `is_primary_source=False`, consistent with the existing target-platform workunits.

## Testing

- New unit tests in `tests/unit/dbt/test_dbt_target_platform_aspects.py` covering: stub emission values, preservation of container-based browse paths, healing of plain name-derived paths, no-op without `target_platform_instance`, config kill-switch, graph-unavailable and aspect-read-failure fallbacks, and two-tier (no database) path structure. Verified red on the pre-fix code (7/8 fail) and green with the fix.
- Full dbt unit suite passes (223 tests).
- All 20 dbt integration tests pass; the `dbt-test-with-target-platform-instance` golden file is regenerated (new `dataPlatformInstance` aspects; no browse paths, as the test pipeline has no graph connection).

### DIFF
--- a/metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
+++ b/metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
@@ -577,15 +577,9 @@ class DBTCommonConfig(
     emit_target_platform_instance_aspects: bool = Field(
         default=True,
         description="When target_platform_instance is set, emit dataPlatformInstance and "
-        "browsePathsV2 aspects for target-platform sibling entities. Without these, a "
-        "target entity that is never ingested directly by the warehouse connector is "
-        "auto-created by the server with a platform-only dataPlatformInstance and a "
-        "name-derived browse path, which surfaces as a duplicate plain-name navigation "
-        "folder and hides the entity from platform-instance filters. browsePathsV2 is "
-        "only written when a DataHub graph connection is available and the entity's "
-        "existing browse path is missing or is a plain name-derived default; "
-        "container-based browse paths written by the warehouse connector are never "
-        "overwritten.",
+        "browsePathsV2 aspects for target-platform sibling entities so they are correctly "
+        "grouped under their platform instance in browse and filters. Browse paths written "
+        "by the warehouse connector are never overwritten.",
     )
     use_identifiers: bool = Field(
         default=False,
@@ -3000,12 +2994,10 @@ class DBTSourceBase(StatefulIngestionSourceBase):
                         for mcp in target_patch.build()
                     )
 
-                # Emit platform-instance metadata for the target entity. If the
-                # target entity is never ingested directly by the warehouse
-                # connector, the server auto-creates it with a platform-only
-                # dataPlatformInstance and a name-derived browse path, which
-                # surfaces as a duplicate plain-name navigation folder and hides
-                # the entity from platform-instance filters.
+                # Deliberately NOT gated by _should_create_sibling_relationships:
+                # lineage emission below also auto-creates target entities, so
+                # these aspects are needed whenever the target URN is referenced,
+                # not only when this source emits the sibling patch itself.
                 for mcp in self._create_target_platform_instance_mcps(
                     node, node_datahub_urn
                 ):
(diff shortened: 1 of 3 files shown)


metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
```diff
@@ -3026,6 +3044,87 @@ def create_target_platform_mces(
...
+        generation post datahub-project/datahub#17263 resolves the first segment
+        to a dataPlatformInstance URN when one exists. Only pre-#17263
+        server-generated defaults have a plain-name first entry, and those are
+        exactly the paths this method is meant to replace.
+        """
+        graph = self.ctx.graph
+        if graph is None:
+            return False
+        try:
+            existing = graph.get_aspect(node_datahub_urn, BrowsePathsV2Class)
```
alfiyas-datahub, line 3113: This code makes one synchronous request to GMS for every target node, every time the ingestion runs when `target_platform_instance` is configured. For large dbt projects, this can result in a lot of slow, sequential network requests, especially during a platform-instance migration where every dataset is affected. Can we improve this by caching the results, fetching them in batches, or only doing it on the first run? If not, we should at least document that this can impact performance.

metadata-ingestion/src/datahub/ingestion/source/dbt/dbt_common.py
```diff
@@ -3026,6 +3044,87 @@ def create_target_platform_mces(
...
+        instance_urn = mce_builder.make_dataplatform_instance_urn(
+            platform_urn, self.config.target_platform_instance
+        )
+        yield MetadataChangeProposalWrapper(
+            entityUrn=node_datahub_urn,
+            aspect=DataPlatformInstanceClass(
+                platform=platform_urn,
+                instance=instance_urn,
+            ),
+        )
```
alfiyas-datahub, line 3074: The "does nothing" behavior only works if target_platform_instance is the same as the warehouse connector's platform_instance. It's worth adding a note to the config saying these two values must match. If they don't, dbt will write a different instance value, and the warehouse connector won't recognize it as the same thing, which can lead to inconsistent metadata.

alfiyas-datahub (approved):

PR #18540, opened 2026-07-22
feat(ingest/dataplex): support the Dataplex Metadata Export API extraction method

### LINKED ISSUE #18538: feat(ingest/dataplex): support the Dataplex Metadata Export API for central/federated catalog projects
## Problem

The Dataplex connector discovers assets by iterating `project_ids` and calling
`list_entry_groups` → `list_entries` → `get_entry`. These APIs only return entries
**physically created** in a project's entry groups.

Many organizations run a *central catalog* architecture: tenant projects grant a central
Dataplex project read access, and assets (BigQuery datasets, DPMS, Vertex AI, etc.) surface
in the central project via Dataplex catalog linking/federation — without being physically
created there. In this setup the connector ingests nothing:

- Against the **central project**, `list_entries` returns only the auto-created
  `@bigquery_entry` / `@storage_entry` stubs (no FQN) → run ends with "No metadata was produced".
- Against the **tenant projects**, there are no Dataplex entries at all (Dataplex API not
  enabled there).

```
central-catalog  -> surfaces all BQ/DPMS/Vertex assets from tenants A, B, C (via linking)
tenant A / B / C -> no physical Dataplex entries
```

## Alternatives considered

`SearchEntries` scoped to the central project does see linked entries, but Google documents
that search recall is not guaranteed
(https://docs.cloud.google.com/dataplex/docs/search-assets#recall), making it unreliable as
a primary extraction mechanism.

## Proposal: Metadata Export API

Google's [Metadata Export API](https://docs.cloud.google.com/dataplex/docs/export-metadata)
returns a complete, consistent snapshot of catalog metadata — including linked/federated
entries. Validated against a production central-catalog environment.

Flow:

1. Submit one `MetadataJob` (type `EXPORT`) per configured location, scoped to the
   configured projects and the entry types the connector supports, writing JSONL to a GCS bucket.
2. Poll the jobs to completion, then stream the exported JSONL from GCS.
3. Each line contains a full `Entry` (aspects included) — parse it back into the
   `dataplex_v1.Entry` proto and feed it through the **existing** mapper pipeline unchanged.

## Design

- New opt-in config: `extraction_method: api | export` (default `api` — existing behavior
  fully preserved) plus an `export_config` block (job runner project, per-location GCS
  buckets, optional prefix, poll interval/timeout).
- Export mode reuses the existing mappers, schema extraction, containers, lineage
  (Data Lineage API), glossary, and stateful ingestion — only the entry-fetch layer differs.
- If an export job fails or its output can't be fully read, the run reports a failure so
  stale-entity soft-deletion is suppressed (no false tombstones from a partial export).
- Required IAM for export mode: `roles/dataplex.metadataJobOwner` on the job runner project,
  `roles/dataplex.catalogEditor` on the scoped projects, and `roles/storage.objectUser` on
  the export bucket(s).

TejasRGitHub:
## Summary

Adds an opt-in `extraction_method: export` to the Dataplex connector, using Google's [Metadata Export API](https://docs.cloud.google.com/dataplex/docs/export-metadata) to fetch entries instead of `list_entry_groups` / `list_entries` / `get_entry`.

Fixes #18538

## Why

The existing API-based discovery only returns entries **physically created** in the configured projects. In central-catalog / federated architectures — where tenant assets surface in a catalog project via Dataplex catalog linking — `list_entries` returns nothing, and the connector produces no metadata (see #18538 for full details). The Metadata Export API returns a complete snapshot including linked/federated entries.

## How

- New config: `extraction_method: api | export` (default `api`, existing behavior fully preserved) plus an `export_config` block (job runner project, per-location GCS buckets, optional prefix, poll interval/timeout) with validation.
- New `dataplex_export.py`: submits one `MetadataJob` (type `EXPORT`) per configured entries location — scoped to the configured projects and the entry types the connector supports — polls to completion, then streams the exported JSONL from GCS and parses each line back into a `dataplex_v1.Entry` proto.
- Exported entries flow through the **existing** filter + mapper pipeline unchanged (`process_exported_entries` on `DataplexEntriesProcessor`); schema extraction, containers, lineage, glossary, platform mapping, and stateful ingestion are all reused.
- If an export job fails, times out, or its output cannot be fully read, the run reports a failure so stale-entity removal is suppressed (no false soft-deletes from a partial export).
- `test_connection` validates GCS bucket access and `metadataJobs` permission in export mode.
- Docs: extraction-method guide, IAM requirements, and a recipe example added to the Dataplex source docs.

## Testing

- 15 new unit tests (config validation, bucket resolution, export scope, job submit/poll, GCS JSONL parsing incl. malformed lines and stale-job filtering).
- New integration test running the full pipeline in export mode with mocked REST/GCS and golden-file validation.
- All 310 dataplex tests pass; `lint` (ruff + mypy) clean.

## Checklist

- [x] Links to related issues
- [x] Docs related to the changes have been added/updated
- [ ] For any breaking change/potential downtime/deprecation/big changes an entry has been made in [Updating DataHub](https://github.com/datahub-project/datahub/blob/master/docs/how/updating-datahub.md) — not applicable, purely additive and opt-in


metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
```diff
@@ -0,0 +1,462 @@
...
+        "Found %d object(s) under gs://%s/%s; %d belong to export job '%s'.",
+        len(blobs),
+        target.bucket,
+        list_prefix or "",
+        len(matching),
+        target.job_id,
+    )
+    if not matching:
+        report.export_locations_with_no_output += 1
+        report.warning(
```
sgomezvillamor: ⚠️ **Partial exports can silently soft-delete a location.** Stale-entity removal is suppressed only when `report.failures` is non-empty — `report.warning()` does not suppress it (and `is_export_partial()` is currently unused in production).

Keeping *individual* errors as warning + continue is right: a single entry that fails proto parsing should be skipped, not fail the run. But *structural* errors mean the location is incomplete and should escalate to `report.failure()` so the stateful-ingestion state isn't committed and stale entities aren't wrongly soft-deleted:
- SUCCEEDED job but **no** blobs match the `job=` marker while blobs *do* exist under the prefix (here) → likely a prefix/job-id mismatch, so `len(blobs) > 0 and len(matching) == 0` should be a failure, not a warning.
- a whole blob that can't be read/parsed.
- non-`entry` lines skipped silently at L434-436 (no counter/log) → if a schema change re-keys entries, every line drops with zero warnings; add a counter so a 100%-skip run is diagnosable.

---

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
```diff
@@ -0,0 +1,462 @@
...
+    logger.info(
+        "Submitting Dataplex EXPORT job '%s' (runner_project=%s, location=%s, "
+        "projects=%s) -> %s",
+        job_id,
+        job_project,
+        location,
+        scoped_projects,
+        output_path,
+    )
+    resp = session.post(
```
sgomezvillamor: 💡 **Observability (dev + user-facing).** The POST response (`{"name": <server job resource>}`) is discarded — capturing and logging it (plus a console link) would help troubleshooting. And for user-facing observability, the submitted jobs currently live only in logs; consider a structured report field, e.g. `export_jobs: List[ExportJobInfo]` (location, job id, terminal state, duration, entries_read, output_path), so operators can see per-job status directly in the report.

---

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
```diff
@@ -0,0 +1,462 @@
...
+                    ExportTarget(
+                        location=pj.location,
+                        bucket=pj.bucket,
+                        job_id=pj.job_id,
+                        output_path=pj.output_path,
+                    )
+                )
+            else:
+                report.export_jobs_failed += 1
+                report.failure(
```
sgomezvillamor: 💡 The Dataplex `metadataJob.status` payload usually carries an error message / details, but only `state` is surfaced here. Folding that message into the failure `context` would tell the user *why* it failed, not just `state=FAILED`.

---

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
```diff
@@ -0,0 +1,462 @@
...
+        for pj in pending:
+            try:
+                job = _get_job(
+                    session,
+                    export_config.export_job_runner_project,
+                    pj.location,
+                    pj.job_id,
+                )
+            except Exception as exc:
+                logger.warning(
```
sgomezvillamor, line 453: 💡 The poll exception is logged but never captured into the report. On a *persistent* error (revoked token, 403, deleted job) the loop spins every `export_poll_seconds` until the global deadline, then reports *"timed out — increase `export_timeout_seconds`"* — misdiagnosing an auth/permission error as slowness. Suggest capturing the last poll `exc` into the timeout failure's `exc=`, and optionally failing fast after N consecutive poll errors.

---

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
```diff
@@ -0,0 +1,462 @@
...
+    logger.info(
+        "All %d export job(s) submitted; polling until each completes.",
+        len(pending),
+    )
+
+    # Phase 2: poll all pending jobs jointly.
+    targets: List[ExportTarget] = []
+    deadline = time.time() + export_config.export_timeout_seconds
+
+    while pending:
```
sgomezvillamor, line 430: 📣 **Long-running exports — worth thinking through how to address.** This blocks synchronously up to `export_timeout_seconds` (default 3600s), and the report counters aren't updated until a job reaches a terminal state, so during a multi-hour export the periodic report shows no movement — hard to distinguish from a hang for an operator.

No metadata workunits can be emitted before the export finishes, but a few options to consider: push per-cycle polling progress into the structured report (elapsed / last state per job) so it visibly isn't stalled; revisit the 1h default (large-catalog exports can legitimately exceed it); and document expected durations. A read-only "point at existing GCS output" mode (see the summary comment) would also sidestep the in-process wait entirely.

---

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
```diff
@@ -0,0 +1,462 @@
...
+    DataplexConfig,
+    DataplexExportConfig,
+)
+from datahub.ingestion.source.dataplex.dataplex_mappers import ENTRY_MAPPERS
+from datahub.ingestion.source.dataplex.dataplex_report import DataplexReport
+
+logger = logging.getLogger(__name__)
+
+DATAPLEX_API_ROOT = "https://dataplex.googleapis.com/v1"
+GCP_SCOPES = ["https://www.googleapis.com/auth/cloud-platform"]
```
sgomezvillamor, line 58: 💡 `GCP_SCOPES` is byte-for-byte identical to `_DATAPLEX_OAUTH_SCOPES` in `dataplex.py:72`, and `dataplex.py` already imports from this module — importing `GCP_SCOPES` instead of keeping two copies avoids drift. Scopes are security-relevant, so a single source is worth it.

---

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
```diff
@@ -0,0 +1,462 @@
...
+    """POST ``metadataJobs.create`` (EXPORT) scoped to the given entry types."""
+    url = (
+        f"{DATAPLEX_API_ROOT}/projects/{job_project}/locations/{location}"
+        f"/metadataJobs?metadataJobId={job_id}"
+    )
+    scoped_projects = [
+        p if p.startswith("projects/") else f"projects/{p}" for p in project_ids
+    ]
+    body = {
+        "type": "EXPORT",
```
sgomezvillamor: 💡 Per the connector convention (hoist wire literals to named constants), consider constants for `"EXPORT"` (here), the JSON keys `"status"`/`"state"` (L263-264) and `"entry"` (L434), the `".jsonl"` suffix (L368), and the content-type header (L135 — which is also likely redundant, since `requests` sets `application/json` automatically when `json=` is passed). Best done as one focused pass.

---

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
```diff
@@ -0,0 +1,462 @@
...
+
+    location: str
+    job_id: str
+    bucket: str
+    output_path: str
+    started: float = field(default_factory=time.time)
+    last_state: Optional[str] = None
+
+
+def run_exports(
```
sgomezvillamor: 💡 `run_exports` is ~165 lines and ~5-deep nested (`while → for → try → if state-changed → if terminal → if SUCCEEDED`), mixing submit / poll-loop / per-job evaluation. Extracting `_submit_all_jobs(...) -> List[_PendingJob]` and a per-job `_evaluate_job(...) -> bool` (returns "still pending") would flatten it without behavior change. The joint-polling design itself is good — this is purely about length/nesting.

---

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
```diff
@@ -0,0 +1,462 @@
...
+    stale-entity removal handler must not soft-delete them.
+    """
+    export_config = config.export_config
+    assert export_config is not None  # enforced by config validation
+    entry_types = export_scope_entry_types()
+
+    # Phase 1: submit all jobs.
+    pending: List[_PendingJob] = []
+    for location in config.entries_locations:
+        bucket = export_config.bucket_for_location(location)
```
sgomezvillamor: 💡 `bucket_for_location` / `_output_path` run *before* the per-location `try` (L202), so if bucket resolution raises it propagates out of `run_exports` and aborts every other location, rather than being recorded as just this location's failure. Is that intentional? If not, moving these two lines inside the `try` makes it consistent with the per-location failure handling right below.

---

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
```diff
@@ -0,0 +1,462 @@
...
+    storage_client: storage.Client,
+    target: ExportTarget,
+    export_config: DataplexExportConfig,
+    report: DataplexReport,
+) -> List[storage.Blob]:
+    """List this export job's ``.jsonl`` blobs under the configured prefix."""
+    prefix = (export_config.prefix or "").strip().strip("/")
+    list_prefix = f"{prefix}/" if prefix else None
+    try:
+        blobs = list(storage_client.list_blobs(target.bucket, prefix=list_prefix))
```
sgomezvillamor: 💡 Perf: this lists **all** objects under the prefix and filters by `job=` in Python. Since buckets are reused across runs (that's why the `job=` marker exists), long-lived buckets accumulate objects and each run's listing grows slower over time. Worth documenting a bucket lifecycle/retention rule, or using a per-run prefix, so the listing stays bounded.

---

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
```diff
@@ -0,0 +1,462 @@
...
+
+
+def _list_matching_blobs(
+    storage_client: storage.Client,
+    target: ExportTarget,
+    export_config: DataplexExportConfig,
+    report: DataplexReport,
+) -> List[storage.Blob]:
+    """List this export job's ``.jsonl`` blobs under the configured prefix."""
+    prefix = (export_config.prefix or "").strip().strip("/")
```
sgomezvillamor: 💡 `.strip().strip("/")` prefix normalization is duplicated here and in `_output_path` (L90). Extracting a `_clean_prefix(prefix) -> str` keeps the rule in one place.

---

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
```diff
@@ -0,0 +1,462 @@
...
+        )
+        resp.raise_for_status()
+
+
+def _get_job(
+    session: google.auth.transport.requests.AuthorizedSession,
+    job_project: str,
+    location: str,
+    job_id: str,
+) -> dict:
```
sgomezvillamor: 💡 `_get_job` returns a bare `dict`; annotate `-> Dict[str, Any]` (the connector convention prefers `typing` aliases, and `Any` is honest for boundary JSON). Optionally extract `_job_state(job) -> str` so callers don't touch the raw dict shape at L263-264.

---

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
```diff
@@ -0,0 +1,462 @@
...
+        )
+    return matching
+
+
+def iter_exported_entries(
+    storage_client: storage.Client,
+    target: ExportTarget,
+    config: DataplexConfig,
+    report: DataplexReport,
+) -> Iterable[Tuple[dataplex_v1.Entry, str]]:
```
sgomezvillamor: 💡 Per the connector convention (dataclasses over tuples), consider a small frozen `ExportedEntry(entry, location)` instead of `Tuple[dataplex_v1.Entry, str]`. It's self-documenting and removes the positional unpacking coupling between `iter_exported_entries` here and `process_exported_entries` in `dataplex_entries.py`.

---

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_config.py
```diff
@@ -108,6 +108,67 @@ class DataplexFilterConfig(ConfigModel):
...
+    bucket_base_name: Optional[str] = Field(
+        default=None,
+        description="Fallback base GCS bucket name used when a location is not "
+        "listed in 'export_bucket_config'. The bucket name is derived as "
+        "'{bucket_base_name}-{location}'.",
+    )
+
+    prefix: Optional[str] = Field(
+        default=None,
+        description="Optional folder prefix inside each export bucket (max 128 chars).",
```
sgomezvillamor: 💡 The description says "max 128 chars" but nothing enforces it. Adding `max_length=128` to the `Field(...)` lets pydantic validate it at config time instead of failing later at export time.

---

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_config.py
```diff
@@ -338,6 +418,31 @@ def validate_location_configuration(self) -> "DataplexConfig":
 
         return self
 
+    @model_validator(mode="after")
+    def validate_export_configuration(self) -> "DataplexConfig":
+        """Validate export-method configuration when extraction_method is 'export'."""
+        if self.extraction_method != "export":
+            return self
```
sgomezvillamor, line 462: 💡 This early-returns when the method isn't `export`, so an `export_config` set under `extraction_method: api` is silently accepted and ignored. A `logger.warning(...)` here would save a confusing debugging session.

---

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex.py
```diff
@@ -384,6 +396,36 @@ def test_connection(config_dict: dict) -> TestConnectionReport:
...
+                location = config.entries_locations[0]
+                runner_project = config.export_config.export_job_runner_project
+                resp = session.get(
+                    f"{DATAPLEX_API_ROOT}/projects/{runner_project}"
+                    f"/locations/{location}/metadataJobs?pageSize=1"
+                )
+                resp.raise_for_status()
+
+                bucket_name = config.export_config.bucket_for_location(location)
+                storage_client = storage.Client(
```
sgomezvillamor: 💡 This `storage.Client(project=runner_project, credentials=...)` construction is duplicated in `_get_entries_workunits_via_export` (L495). A small `build_storage_client(export_config, credentials)` helper next to `build_authed_session` would keep the "runner project owns the bucket" coupling in one spot.

---

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex.py
```diff
@@ -384,6 +396,36 @@ def test_connection(config_dict: dict) -> TestConnectionReport:
...
+            if config.extraction_method == "export":
+                # Export mode exercises the metadataJobs API on the job runner
+                # project and the export bucket instead of Catalog listing.
+                assert config.export_config is not None
+                session = build_authed_session(credentials)
+                location = config.entries_locations[0]
+                runner_project = config.export_config.export_job_runner_project
+                resp = session.get(
+                    f"{DATAPLEX_API_ROOT}/projects/{runner_project}"
+                    f"/locations/{location}/metadataJobs?pageSize=1"
```
sgomezvillamor: 💡 This probes `metadataJobs` **list** (read), but export actually needs `roles/dataplex.metadataJobOwner` (create) to submit jobs — so `test_connection` can pass while ingestion later fails on submit. Minor: the PR description says it validates "`metadataJobs` permission"; worth aligning the wording, and ideally noting that create permission isn't verified here.

---

metadata-ingestion/docs/sources/dataplex/dataplex_pre.md
```diff
@@ -4,6 +4,22 @@ The `dataplex` module ingests metadata from Google Cloud Knowledge Catalog (Data
...
+- **`export`** — submits one Dataplex [metadata export job](https://docs.cloud.google.com/dataplex/docs/export-metadata) per configured entries location (scoped to the configured projects and the entry types the connector supports), waits for the jobs to finish, and then reads the exported JSONL from a Cloud Storage bucket. Requires the `export_config` section.
+
+Use `export` when your organization runs a **central catalog** architecture: tenant projects grant a central Dataplex project permission to read their metadata, and assets surface in the central project via Dataplex catalog linking/federation without being physically created there. Those linked entries are invisible to `list_entries` (the `api` method returns nothing), but a metadata export scoped to the central project includes them.
+
+Notes on `export` mode:
+
+- One export job runs per entry in `entries_locations`, and each location must resolve to a GCS bucket via `export_config.export_bucket_config` or `export_config.bucket_base_name` (bucket name `{bucket_base_name}-{location}`). The bucket for a `global`-location job must be in a compatible (multi-)region.
```
sgomezvillamor: 💡 Typo: "One export job runs per entry in `entries_locations`" → "per entries **location**". Also worth noting the default `entries_locations` has 4 entries (incl. `global`), so export mode fans out to 4 jobs by default, and the `global` job needs a multi-region bucket.

---

sgomezvillamor (commented):
Thanks for this — overall it looks really good, kudos! 🎉 The default `api` path is untouched, exported entries flow back through the exact same filter + mapper pipeline, and the "report a failure so stale entities aren't tombstoned on a partial export" instinct is exactly the right one. Nice test suite and docs too. I left a batch of inline comments (mostly suggestions); a few higher-level things below.

### 🚧 Biggest thing: decouple job execution from reading

Right now `export` mode always submits the metadata job *and* reads its output. I think it'd be really valuable to also support a mode where the user points the connector at an existing GCS export location and we just ingest it. Three reasons:

1. **Users who orchestrate their own exports** (Cloud Scheduler / Workflows / a separate pipeline) can just hand us the output path instead of granting DataHub job-submission rights.
2. **Permissions drop sharply** — and this one's important: read-only needs only `storage.objectViewer` on the bucket, versus `dataplex.metadataJobOwner` + `dataplex.catalogEditor` on every scoped project today.
3. **Real end-to-end integration testing** — the current integration test has to mock the whole `metadataJobs` REST session; a read-only mode lets us feed JSONL fixtures straight through the pipeline with no job-submission mocking at all.

(One open question if you take this on: the reader currently scopes to *this run's* output via the `job=<id>` marker; a read-only path has no job id, so it'd need either an explicit path or "latest `job=` partition" selection.)

### ❓ On the justification for this method

Is `export` expected to be more performant than `api` for large asset counts (one bulk export vs. many `list`/`get_entry` RPCs)? If so, that's worth calling out in the user-facing docs — and it actually reframes the feature: the central-catalog/federation case becomes one important use case for `export`, rather than the whole reason for it. Curious what you've seen on real volumes.

### ✅ Tests

Clean suite, no AI-slop anti-patterns — it asserts behavior (report counters, `is_export_partial()`, `job=` scoping) rather than implementation. The gap is resilience coverage:

- **Timeout path** in `run_exports` — this is the central safety guarantee (a hung export must surface as a failure to suppress soft-deletion), and it's untested. Highest priority; `time_machine` is already a dep.
- **`test_connection` export branch** — all three outcomes uncovered (probe OK, bucket exists, bucket missing).
- **Multi-location joint polling** — the headline design feature, but every test uses a single location.
- Poll-retry recovery, `process_exported_entries` failure/filter paths (no direct unit test), mid-stream blob-read failure, the proto-parse-failure branch, and CANCELED-is-terminal.
- Highest-value single add after the timeout test: something that proves `is_export_partial()` / a reported failure **actually gates** stale-entity removal in a real pipeline run — right now that link is asserted only in isolation.

Happy to discuss any of these — thanks again for the contribution! 🙏

---

metadata-ingestion/docs/sources/dataplex/dataplex_pre.md
TejasRGitHub: @sgomezvillamor , Yes will fix the typo. Also, will add the comment with the entries_location default values

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex.py
TejasRGitHub: Will fix this @sgomezvillamor  and align the wording

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex.py
TejasRGitHub: Will do. Thanks for the suggestion

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_config.py
TejasRGitHub: Let me see if this is an actual limit with any of the GCP paths. if not I wil remove the 128 char limit and if it is then I wil add the limit in the pydantic `Field`

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_config.py
TejasRGitHub: Sure ,will add the log warning. Thanks for pointing that

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
TejasRGitHub: Thanks for pointing this. Will use the already existing scope if they work

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
TejasRGitHub: Hey @sgomezvillamor , I didn't get this. Is this comment only for documentaion ?

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
TejasRGitHub: Fixed in 54db61c60e — when the bucket/prefix contains objects but none carry this run's job id, it is now report.failure() (so stale-entity removal is suppressed); a fully empty listing stays a warning since a location can legitimately have no supported entries, and is_export_partial() now counts no-output locations too. Unit tests added for both branches.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
TejasRGitHub: Done in 54db61c60e — dataplex.py now imports GCP_SCOPES from dataplex_export, so the scopes have a single source.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
TejasRGitHub: This duplication is gone as of 54db61c60e — the reader now derives bucket/prefix from the target's output_path via a shared parse_gcs_path helper, so the strip logic lives in one place.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_config.py
TejasRGitHub: Checked the API reference and the 128-char custom prefix limit is real, so added max_length=128 on the field in 54db61c60e.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_config.py
TejasRGitHub: Added in 54db61c60e — a warning is logged when export_config is set but extraction_method is not 'export' (and likewise when job-submission settings are set in read-only mode).

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex.py
TejasRGitHub: Done in 54db61c60e — added build_storage_client() next to build_authed_session and both call sites use it.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex.py
TejasRGitHub: Aligned in 54db61c60e — the docstring and docs now say the probe verifies metadataJobs read access only and cannot verify metadataJobOwner (create). In the new read-only mode, test_connection probes the configured paths instead.

metadata-ingestion/docs/sources/dataplex/dataplex_pre.md
TejasRGitHub: Fixed in 54db61c60e — reworded to 'per entries location' and added a note that the default entries_locations fans out to 4 export jobs (with global needing a multi-region bucket).

TejasRGitHub:
Thanks a lot for the detailed review @sgomezvillamor! Pushed 54db61c60e addressing a good chunk of it — replied on the individual threads and resolved the ones that are done.

**Read-only mode** — implemented. `export_config.existing_export_paths` maps a location to the `gs://bucket[/prefix]` output of an export you orchestrate yourself; no jobs are submitted and the connector only needs `storage.objectViewer` on the buckets. On your open question about selecting output without a job id: both options work — point the path at a specific `.../job=<id>` folder to pin an exact run, or point at a parent path and the connector reads the most recently written `job=` partition. And the new integration test does exactly what you suggested: JSONL fixtures feed straight through the pipeline with zero metadataJobs mocking.

Also in that commit: the partial-export case you flagged is escalated to `report.failure()` when the prefix has objects but none from this run's job, the 128-char prefix limit turned out to be a real API constraint so it is enforced via pydantic now, plus the scopes / storage-client dedup and the test_connection + docs wording.

**On export vs api performance** — honestly the driver here was correctness (catalog-linked entries are invisible to `list_entries`), and I don't have solid numbers at scale yet. Directionally export should win for large catalogs — the api path pays one `get_entry` RPC per entry while export is one bulk job per location plus GCS reads — but I'd rather measure before putting a perf claim in the docs. Will add a note once I have real volumes to point at.

**Tests** — the read-only integration test, units for the failure-escalation branches, and test_connection coverage for the read-only mode landed with this commit. Timeout path, multi-location joint polling, and an end-to-end check that a reported failure actually gates stale-entity removal are next on my list, along with the remaining smaller suggestions (job error message in the failure context, poll-exception capture, `run_exports` split, observability fields).

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
TejasRGitHub: Done in bdadfaf65c — the status message is folded into the failure report, so a failed job now shows the server's reason instead of just state=FAILED.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
TejasRGitHub: Done in bdadfaf65c — the last poll exception is attached to the timeout failure, and after 5 consecutive poll errors the job fails fast with the underlying error instead of waiting out the deadline.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
TejasRGitHub: Done in bdadfaf65c — the server's job resource name is logged on submit, and the report now has a structured export_jobs field (location, job id, state, elapsed seconds, entries read) updated on every poll cycle.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
TejasRGitHub: Partially addressed in bdadfaf65c — the new export_jobs report field updates every poll cycle (state + elapsed), so a long run visibly isn't stalled, the poll sleep is capped to the remaining deadline, and the docs now mention expected durations. Kept the 1h default for now since it's configurable — happy to bump it if you think a different default is better.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
TejasRGitHub: Done in bdadfaf65c — hoisted EXPORT / status / state / message / entry / .jsonl to module constants and dropped the redundant Content-Type header (requests sets it for json=).

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
TejasRGitHub: Done in bdadfaf65c — extracted _submit_all_jobs() and _evaluate_job() exactly as you sketched; run_exports is now just the submit call plus the poll loop, no behavior change.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
TejasRGitHub: Good catch — moved bucket resolution inside the per-location try in bdadfaf65c, so a bad location config fails just that location.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
TejasRGitHub: Done in bdadfaf65c — annotated as Dict[str, Any] and added the _job_state() helper so callers don't touch the raw dict shape.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
TejasRGitHub: Done in bdadfaf65c — added a frozen ExportedEntry dataclass (in dataplex_helpers to avoid an import cycle) and both the reader and process_exported_entries use it now.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
TejasRGitHub: Added a docs note in bdadfaf65c recommending a GCS lifecycle/retention rule on the export buckets (plus rotating the prefix). If you'd rather see a code-level fix I'd go with a per-run subfolder in output_path so each listing is naturally scoped — let me know and I'll do that instead.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_config.py
```diff
@@ -108,6 +109,97 @@ class DataplexFilterConfig(ConfigModel):
...
+    )
+
+    prefix: Optional[str] = Field(
+        default=None,
+        max_length=128,
+        description="Optional folder prefix inside each export bucket. The "
+        "Dataplex Metadata Export API limits the custom prefix to 128 characters.",
+    )
+
+    export_poll_seconds: int = Field(
```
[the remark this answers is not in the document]
TejasRGitHub: Fixed in bdadfaf65c — the sleep is capped to the remaining deadline (min(poll, deadline - now)), so a poll interval longer than the timeout can no longer overshoot it.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_config.py
```diff
@@ -338,6 +449,75 @@ def validate_location_configuration(self) -> "DataplexConfig":
...
+                "export_config.export_job_runner_project must be set when "
+                "extraction_method is 'export' (unless "
+                "'export_config.existing_export_paths' is used)."
+            )
+        # Every entries location must resolve to a bucket up front, so a
+        # misconfiguration fails at recipe validation rather than mid-run.
+        missing = [
+            loc
+            for loc in self.entries_locations
+            if loc not in self.export_config.export_bucket_config
```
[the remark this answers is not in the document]
TejasRGitHub: Fixed in bdadfaf65c — blank export_bucket_config values are rejected at config validation.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex.py
```diff
@@ -371,19 +381,80 @@ def test_connection(config_dict: dict) -> TestConnectionReport:
...
+                    test_report.basic_connectivity = CapabilityReport(capable=True)
+                    return test_report
+
+                # Submit mode exercises the metadataJobs API on the job runner
+                # project and the export bucket instead of Catalog listing. The
+                # listing probe only proves read access — create permission
+                # (roles/dataplex.metadataJobOwner) cannot be verified without
+                # actually submitting a job.
+                session = build_authed_session(credentials)
+                location = config.entries_locations[0]
```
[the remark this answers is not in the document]
TejasRGitHub: Fixed in bdadfaf65c — test_connection now probes the metadataJobs listing and bucket for every configured entries location.

metadata-ingestion/tests/integration/dataplex/golden/dataplex_export_entries_golden.json
```diff
@@ -0,0 +1,595 @@
...
+                    "recursive": false,
+                    "isPartOfKey": false
+                },
+                {
+                    "fieldPath": "name",
+                    "nullable": true,
+                    "description": "Customer name",
+                    "type": {
+                        "type": {
+                            "com.linkedin.schema.StringType": {}
```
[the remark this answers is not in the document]
TejasRGitHub: Real bug, thanks — FLOAT64 and BIGNUMERIC now map to NumberType in bdadfaf65c and the golden is regenerated (orders.total_amount emits NumberType).

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
```diff
@@ -0,0 +1,605 @@
...
+                    except json.JSONDecodeError as exc:
+                        report.export_malformed_lines_skipped += 1
+                        report.warning(
+                            title="Malformed Dataplex export line",
+                            message="Skipping unparseable JSONL line.",
+                            context=f"blob={blob.name}",
+                            exc=exc,
+                        )
+                        continue
+                    entry_dict = obj.get("entry")
```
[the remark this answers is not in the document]
TejasRGitHub: Fixed in bdadfaf65c — non-dict JSON lines are counted as malformed and skipped instead of raising and aborting the rest of the blob.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
```diff
@@ -0,0 +1,605 @@
...
+    partitions: Dict[str, List[storage.Blob]] = {}
+    unpartitioned: List[storage.Blob] = []
+    for blob in jsonl_blobs:
+        match = JOB_PARTITION_RE.search(blob.name)
+        if match:
+            partitions.setdefault(match.group(1), []).append(blob)
+        else:
+            unpartitioned.append(blob)
+
+    if not partitions:
```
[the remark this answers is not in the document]
TejasRGitHub: Behavior kept (only the freshest job partition is read by design), but it's no longer silent — bdadfaf65c adds a report warning with the skipped count, and the docs say to point the path directly at loose objects if they should be read.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_report.py
```diff
@@ -20,6 +20,31 @@ class DataplexReport(StaleEntityRemovalSourceReport):
...
+    export_jobs_submitted: int = 0
+    export_jobs_succeeded: int = 0
+    export_jobs_failed: int = 0
+    export_blobs_read: int = 0
+    export_blobs_read_failed: int = 0
+    export_entries_read: int = 0
+    export_malformed_lines_skipped: int = 0
+    export_locations_with_no_output: int = 0
+
+    def is_export_partial(self) -> bool:
```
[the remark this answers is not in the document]
TejasRGitHub: Decision: gating stays failure-based (that's the framework's mechanism); is_export_partial() is observational and its docstring now says exactly that, including the empty-location divergence you called out (bdadfaf65c).

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_config.py
```diff
@@ -338,6 +449,75 @@ def validate_location_configuration(self) -> "DataplexConfig":
...
+                    ) from e
+            ignored = [
+                name
+                for name, value in (
+                    (
+                        "export_job_runner_project",
+                        self.export_config.export_job_runner_project,
+                    ),
+                    ("export_bucket_config", self.export_config.export_bucket_config),
+                    ("bucket_base_name", self.export_config.bucket_base_name),
```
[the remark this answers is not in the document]
TejasRGitHub: Fixed in bdadfaf65c — the warning now covers prefix, export_poll_seconds, and export_timeout_seconds when explicitly set (via model_fields_set, so defaults don't produce noise).

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
```diff
@@ -0,0 +1,605 @@
...
+                        report.export_malformed_lines_skipped += 1
+                        report.warning(
+                            title="Malformed Dataplex export line",
+                            message="Skipping unparseable JSONL line.",
+                            context=f"blob={blob.name}",
+                            exc=exc,
+                        )
+                        continue
+                    entry_dict = obj.get("entry")
+                    if not isinstance(entry_dict, dict):
```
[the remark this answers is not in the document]
TejasRGitHub: Fixed in bdadfaf65c — lines with a missing or non-dict entry field bump export_malformed_lines_skipped and emit a warning, same as unparseable lines.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
sgomezvillamor: yes, my request was for a docs note, thanks for addressing it!

sgomezvillamor:
First off — thanks for the incredibly careful pass on the previous round. Every thread got a specific fix and, in most cases, a test to go with it: the structural-vs-individual failure distinction (unmatched output now suppresses stale-removal), the structured `export_jobs` progress, the poll fail-fast, the read-only mode, the `run_exports` split, the `FLOAT64`/`BIGNUMERIC` catch — all addressed well. Really appreciate it.

One design point left before I'm comfortable merging: the shape of the export config.

### The modelling issue

The `ignored = [...]` list in `validate_export_configuration` is the tell — a config object shouldn't have to enumerate six of its own fields as "please pretend these aren't set." That smell shows up because one object (`DataplexExportConfig`) is modelling two things that don't actually overlap. If you line the fields up by mode:

- **submit** (`export`): `export_job_runner_project`, `export_bucket_config`, `bucket_base_name`, `prefix`, `export_poll_seconds`, `export_timeout_seconds`
- **read-only** (`existing_export_paths`): just `existing_export_paths`

The shared surface is empty. Everything both paths genuinely share — credentials, project/location selection, filters, the `include_*` toggles — already lives on the parent `DataplexConfig`, not in `export_config`. So there's no common "export config" to factor out; a shared base would be empty. When two variants are disjoint like this, the right tool is a **tagged/discriminated union**, not one flat object full of mutually-exclusive optionals. (There's a second, related smell: read-only mode silently overloads `entries_locations` — the entries stage reads the keys of `existing_export_paths` instead, while `entries_locations` still drives the other stages. That "same field, different meaning" also goes away once each mode owns its config.)

### Why now, not later

It's opt-in and not merged yet, so reshaping the config is free today. Once it ships, both the `extraction_method` values and the `export_config` schema become user-facing surface, and changing them means `pydantic_removed_field` deprecations plus a migration note in `updating-datahub.md`. The behavior in the PR is solid — the thing that's expensive to undo later is the config shape, so it's worth getting right up front.

### Two ways to model it

**Option 1 (preferred) — a dedicated `extraction_method`, each with its own config block:**

```python
class DataplexExportJobConfig(ConfigModel):
    """extraction_method: export — DataHub submits the metadata export jobs."""
    export_job_runner_project: str            # now required, drop the Optional
    export_bucket_config: Dict[str, str] = {}
    bucket_base_name: Optional[str] = None
    prefix: Optional[str] = Field(None, max_length=128)
    export_poll_seconds: int = 15
    export_timeout_seconds: int = 3600

class DataplexReadExportConfig(ConfigModel):
    """extraction_method: read_export — ingest export output produced elsewhere."""
    export_paths: Dict[str, str]              # location -> gs://bucket[/prefix], required

class DataplexConfig(...):
    extraction_method: Literal["api", "export", "read_export"] = "api"
    export_config: Optional[DataplexExportJobConfig] = None
    read_export_config: Optional[DataplexReadExportConfig] = None
```

Validation then collapses to one uniform rule — "the block for the selected method must be present, the others absent" — with no per-field ignore list, and `export_job_runner_project` becomes genuinely required.

**Option 2 — keep `extraction_method: export`, split `export_config` into a discriminated union on a `mode` field:**

```python
class _SubmitExport(ConfigModel):
    mode: Literal["submit"] = "submit"
    export_job_runner_project: str
    ...  # bucket / prefix / poll / timeout

class _ReadExport(ConfigModel):
    mode: Literal["read"]
    export_paths: Dict[str, str]

ExportConfig = Annotated[Union[_SubmitExport, _ReadExport], Field(discriminator="mode")]
```

Pydantic v2 discriminated unions reject fields that don't belong to the selected variant, so a misplaced field becomes a hard error instead of a silent ignore. More compact and keeps the grouping under `export_config`; the tradeoff is the `mode` discriminator is a bit more hidden than a top-level method value. I lean Option 1 since `extraction_method` is the switch users already reason about, but either removes the ignored-fields logic entirely.

### Naming nit

If you go with a third method, I'd avoid `read-config` — it reads as "reads a config file," and hyphens are awkward in a `Literal`. Naming it around *who runs the export* is clearer: `read_export` (vs `export` = "we run it"), or `export_output` / `gcs_export`.

Once the config is remodelled along these lines, I think this is good to merge. Thanks again!

---

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex.py
```diff
@@ -371,19 +381,83 @@ def test_connection(config_dict: dict) -> TestConnectionReport:
...
+                # Submit mode exercises the metadataJobs API on the job runner
+                # project and the export bucket instead of Catalog listing —
+                # for every configured entries location, since ingestion will
+                # submit one job per location. The listing probe only proves
+                # read access — create permission
+                # (roles/dataplex.metadataJobOwner) cannot be verified without
+                # actually submitting a job.
+                session = build_authed_session(credentials)
+                runner_project = config.export_config.export_job_runner_project
+                for location in config.entries_locations:
```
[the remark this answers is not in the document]
TejasRGitHub: Added in bed13578be — a failure-path test now covers the multi-location probe: the first failing bucket marks connectivity as failed, names the location, and stops probing the rest.

metadata-ingestion/src/datahub/ingestion/source/dataplex/dataplex_export.py
```diff
@@ -0,0 +1,741 @@
...
+                            exc=exc,
+                        )
+                        continue
+                    # Valid JSON that is not an ``{"entry": {...}}`` object is
+                    # skipped like a malformed line; guarding here also keeps a
+                    # stray non-object line (e.g. ``[]``) from raising and
+                    # aborting the rest of the blob.
+                    entry_dict = obj.get(ENTRY_KEY) if isinstance(obj, dict) else None
+                    if not isinstance(entry_dict, dict):
+                        report.export_malformed_lines_skipped += 1
```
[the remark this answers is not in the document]
TejasRGitHub: The report stays bounded as-is: structured logs dedupe warnings by title+message into a single entry, and the per-line contexts go into a capped LossyList — the export_malformed_lines_skipped counter carries the exact total. Leaving the per-line reporting since it's what surfaces the affected blob names.

metadata-ingestion/tests/unit/dataplex/test_dataplex_export.py
```diff
@@ -0,0 +1,817 @@
...
+        config = make_export_config(
+            export_config={
+                "export_job_runner_project": "runner-project",
+                "bucket_base_name": "my-export",
+                "export_poll_seconds": 1,
+                "export_timeout_seconds": 1,
+            }
+        )
+        report = DataplexReport()
+        session = make_session(itertools.repeat("RUNNING"))
```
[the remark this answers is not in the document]
TejasRGitHub: That one is deliberate: the loop's deadline is wall-clock, so with sleep mocked the test would just busy-spin for the same ~1s until time passes the deadline. The real 1s sleep is the cheapest deterministic way to cross it; the other tests mock sleep because they terminate by state, not by deadline.

TejasRGitHub:
Thanks for the thorough write-up @sgomezvillamor — agreed on the modelling point, and done in bed13578be following Option 1:

- `extraction_method` now has a third value, `read_export`; `export` keeps only the DataHub-submitted-jobs path.
- `export_config` (now `DataplexExportJobConfig`) holds just the submit-mode fields, with `export_job_runner_project` genuinely required again.
- New `read_export_config` block with a required, validated `export_paths` mapping.
- Validation collapsed to the one rule you suggested: the selected method's block must be present, the others absent — a misplaced block is now a hard error, and the ignored-fields list is gone entirely.
- The `entries_locations` overload is gone with it: in `read_export` the entries stage is driven purely by `export_paths` (called out on the field description).

Docs and the starter recipe cover the third method, and the branch is synced with master again — the pyproject/uv.lock conflicts are resolved by regenerating them from the merged setup.py. All dataplex tests green. Thanks again for pushing on the shape before it became frozen surface!

sgomezvillamor (approved):
Thanks for the contribution, and for so consistently working through every concern and suggestion across the rounds — the config refactor to a dedicated `read_export` method landed exactly right. LGTM 🚀

Ready to merge whenever you'd like — or let me know and I'll do it.

---

TejasRGitHub:
> Thanks for the contribution, and for so consistently working through every concern and suggestion across the rounds — the config refactor to a dedicated `read_export` method landed exactly right. LGTM 🚀
> 
> Ready to merge whenever you'd like — or let me know and I'll do it.
> 

Hey @sgomezvillamor , Thanks for all the review comments. I am ok with merging this change . I can create another PR for some additional changes I have for lineage, additional metadata, CLL ( Column level lineage )

sgomezvillamor:
CI errors are caused by a minio issue already addressed in master, so I'm updating branch with master and enabling auto-merge.

Thanks! Let's keep contribs going! 🚀

